PR #4740, opened 2020-11-05
314: Add an option to show block rewards earned so far for a miner in miner info

jennijuju:


cmd/lotus-storage-miner/info.go
```diff
@@ -158,6 +179,30 @@ func infoCmdAct(cctx *cli.Context) error {
 		}
 	}
 
+	if cctx.Bool("show-rewards") {
+		curTs := head
+		var blockRewardsEarned = abi.NewTokenAmount(0)
+		for curTs.Height() > abi.ChainEpoch(0) {
+			stout, err := api.StateCompute(ctx, curTs.Height(), nil, curTs.Key())
```
magik6k, line 186: I'm pretty sure this will take ~1 week on mainnet

jennijuju: lol😂 is there any other way to do this? Can we add a map in reward actor?

arajasek:
Yeah I think the right way to do this is a shed command that takesa miner X and a start/end tipset height (or tipsetkeys), and does the following (simpler) computation:

- load tipset
- check if any of the blocks in the tipset were mined by X
- if so, compute state to get the reward
- repeat with parent tipset until end

PR #4744, opened 2020-11-06
4399: Be able to get all the multisig wallets an address is a signer of.

### LINKED ISSUE #4399: feature request: cmd to list multisigs a wallet is part of
It would be nice 👌🏽 not urgent at all :)

jennijuju:
Resolves #4399

### DIFF
--- a/api/api_full.go
+++ b/api/api_full.go
@@ -514,6 +514,9 @@ type FullNode interface {
 	// along with the address removal.
 	MsigRemoveSigner(ctx context.Context, msig address.Address, proposer address.Address, toRemove address.Address, decrease bool) (cid.Cid, error)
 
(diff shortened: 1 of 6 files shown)


arajasek (changes requested):
I don't think this is important enough to deserve an API endpoint, we're trying to avoid adding more endpoints than necessary. I think this can either go into lotus shed, or directly into the multisig CLI (but without the API endpoint).

See [here](https://github.com/filecoin-project/lotus/blob/8c7d2efe72f864c7ebb483d047aee12138df7359/cmd/lotus-shed/balances.go#L133) for an example of how a shed util can walk the entire state tree.

arajasek:
Stale! Sad!

Issue #3749, opened 2020-09-10
Not try to send messages from addresses with 0 balance.


jennijuju:
A miner was trying to perform a `lotus-miner actor withdraw` with miner available balance being 3 FIL and worker address = 0 and got `ERROR: GasEstimateMessageGas error: estimating gas used: message execution failed: exit SysErrSenderStateInvalid(2), reason: actor balance less than needed: 0 FIL < 0.00000101 FIL (RetCode=2)`

We think the issue was b/c the first message is sent by worker address and in this case, gas cannot be paid with 0 FIL worker.

The error message is a little bit misleading here. Also, @arajasek suggested that we should improve our ux to not even  try to send messages from addresses with 0 balance.

arajasek:
Digging a little further, it looks like we already do have this check (the `checkBalance` method), it's just that the `GasEstimateMessageGas` check gets triggered and fails first. I think that's fine, a change isn't needed here.

PR #4750, opened 2020-11-06
3749:  Do not attempt to send a message if the sender balance is 0.

jennijuju:
@arajasek this is to address your suggestion for #3749, please feel free to close both pr & issue if you no longer think it's necessary!

### DIFF
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -517,6 +517,7 @@ func (mp *MessagePool) Push(m *types.SignedMessage) (cid.Cid, error) {
 	}()
 
 	mp.curTsLk.Lock()
+
 	publish, err := mp.addTs(m, mp.curTs, true, false)
 	if err != nil {
 		mp.curTsLk.Unlock()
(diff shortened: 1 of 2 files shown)


chain/messagepool/messagepool.go
```diff
@@ -631,6 +632,10 @@ func (mp *MessagePool) checkBalance(m *types.SignedMessage, curTs *types.TipSet)
 		return xerrors.Errorf("failed to check sender balance: %s: %w", err, ErrSoftValidationFailure)
 	}
 
+	if balance == abi.NewTokenAmount(0) {
```
arajasek, line 635: FYI, this isn't the right check. It will return true if the fields of the two `abi.TokenAmount` structs (`balance` and `abi.NewTokenAmount(0)`) are the same, which will be false. Use the `IsZero()` method instead.


For more: https://medium.com/learning-the-go-programming-language/comparing-values-in-go-8f7b002e767a.

arajasek:
Thanks for this, but I don't think it's needed, see note in #3749.

PR #4754, opened 2020-11-06
emit events for peer disconnections and act upon them in the blocksync tracker

### LINKED ISSUE #92: BlockSync should remove syncPeers when peer disconnects
Should be easy with libp2p eventbus

vyzo:
Closes #92

### DIFF
--- a/chain/exchange/peer_tracker.go
+++ b/chain/exchange/peer_tracker.go
@@ -11,6 +11,7 @@ import (
 	host "github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"go.uber.org/fx"
+	"go.uber.org/multierr"
 
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/lib/peermgr"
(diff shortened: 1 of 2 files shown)


Kubuxu (approved):

chain/exchange/peer_tracker.go
```diff
@@ -38,20 +39,34 @@ func newPeerTracker(lc fx.Lifecycle, h host.Host, pmgr *peermgr.PeerMgr) *bsPeer
...
+	rmSub, err := h.EventBus().Subscribe(new(peermgr.RemoveFilPeer))
+	if err != nil {
+		panic(err)
+	}
+
+	go func() {
+		for rmPeer := range rmSub.Out() {
+			bsPt.removePeer(rmPeer.(peermgr.RemoveFilPeer).Id)
+		}
+	}()
```
raulk, line 62: Creating two separate subscriptions is racy. If a peer reconnects, and depending on the system load and timing, you may end up processing the reconnection before the disconnection, which may render the internal state invalid.

A multi-type subscription will give you strict(-ish?) ordering on the subscriber side, assuming that the publisher publishes in order at origin.

Godocs: https://godoc.org/github.com/libp2p/go-libp2p-core/event#Bus

raulk (changes requested):

chain/exchange/peer_tracker.go
vyzo: I am aware of this race, but having a single emitter won't really fix the _underlying_ race: there are two seperate sources for the events, so we can get a race anyway.

I can still make the change to single emitter if that will make you happy, but it won't really change anything!

raulk: Point taken, but I think this needs a refactor on the emitter side, though. I don't think it's wise to add more fuel to the fire of something that's already racy. There is already a stability issue here; identifying it and building on it instead of fixing it doubles down on instability.

How about we:

1. Unify the "added" and "removed" events into a single event type with different statuses: "peer changed" with an "action" field.
2. Then we can have a single emitter.
3. Then we can have a single subscriber.
4. Things will be ordered, as long as in libp2p things happened in the right order, which I'm quite confident they do.

raulk (approved):

raulk (approved):
It would be unfair to ask for tests for the whole component when you're only submitting a patch, but the peer tracker looks untested.

magik6k (approved):

PR #4769, opened 2020-11-09
optimize code replace strings with constants

zgfzgf:
add ReturnType const


whyrusleeping (approved):
That oughta help us avoid typos in the future

Kubuxu (approved):

magik6k (approved):

Issue #4752, opened 2020-11-06
chain/state store improvements: introduce a block cache and optimise existing 'has' cache


raulk:
The hypothesis is that the VM and chain sync access hot values repeatedly, specifically HAMT nodes, block ranges, and actor states, over the course of various epochs.

Introducing a block cache would alleviate disk reads (sstable access), which are the largest hotspot right now.

Ideally we'd want this cache to be a CBOR object cache instead of a block cache, to avoid the cost of redeserialisation (CPU and heap allocs), but at this stage that's micro-optimisation, as there are other larger fish to fry.

Additionally, the current ARC cache takes up around 1.85% of CPU time. It's a small fraction, but moving to a different cache library like Ristretto may narrow this a sliver, while preserving or augmenting the hit ratio.

## Useful profile screenshots

<img width="319" alt="image" src="https://user-images.githubusercontent.com/1101242/98364681-23204f00-2029-11eb-9e34-e2d481e68c62.png">

<img width="1086" alt="image" src="https://user-images.githubusercontent.com/1101242/98365422-50b9c800-202a-11eb-8113-1be64eaa177b.png">

<img width="630" alt="image" src="https://user-images.githubusercontent.com/1101242/98365763-c887f280-202a-11eb-85fb-469cddba17d2.png">

arajasek:
@raulk Is this something on your radar / being handled by you & your friends?

jennijuju:
@vyzo Do you have any thoughts on this? Is this implemented or needs further implementation.

raulk:
We implemented this in https://github.com/filecoin-project/lotus/pull/4771, but the effective performance gain wasn't as pronounced as we had anticipated. The real solution is the splitstore, which keeps the hot data (latest chain finalities) in a dedicated store/region. In the current implementation, the hot and cold stores are backed by badger, and they both rely on mmap/page caches to access data efficiently. Since the hot store is accessed continuously, its indices (SStables) and vlogs should be prioritised in the kernel page cache over the cold stores, thus acting like a cache that we expect to be good enough for the purposes of this issue.

PR #4771, opened 2020-11-09
logical segregation of blockstores + freecache-cached chain and state blockstores

raulk:
Fixes https://github.com/filecoin-project/lotus/issues/4752.

This PR segregates the monolithic blockstore into two logical blockstores: the chain and the state blockstore.

Currently they're backed by the same physical store, but the abstractions introduced here allow us to front the chain and the state blockstores by different caches, each tuned for the specific data access pattern of its nature. The also pave the way for upcoming physical segregation.

The caching is based on [Freecache](https://github.com/coocood/freecache), a near LRU cache. Given our access patterns, pure LRU is estimated to behave better than ARC/2Q. We use a fork of Freecache (maintained by me) that enables zero-copy access to cached values. The total footprint of the caching layer is 416MiB.

The ARC cache is no longer used, and the deprecated `CachingBlockstore` has been removed.

This PR also introduces a `LotusBlockstore` interface. It is the union of the `Blockstore` and `Viewer` interfaces, and is required by the chain and state blockstores. This way, we remove the optionality of `Viewer` and simplify the code.

This PR also removes block size caching (Lotus never calls blockstore.GetSize).

## Benchmark

1. Imported minimal_finality_stateroots_latest.car
2. Ran for 1000 block validations.
3. Took a metrics dump.

Note that these comparisons put the cached version at a disadvantage, because the store is still light. As the store grows, the benefits of the cache are much higher. 

### Hit rate

```
lotus_blockstore_cache_hit_ratio{cache_name="chain_block_cache"} 0.8236359865598997
lotus_blockstore_cache_hit_ratio{cache_name="chain_exists_cache"} 0.789787058767283
lotus_blockstore_cache_hit_ratio{cache_name="state_block_cache"} 0.6361207867057612
lotus_blockstore_cache_hit_ratio{cache_name="state_exists_cache"} 0.765091942682962
```

### Block validation times

**Before**

Average: 1470,861553784860558 ms (91% validated under 2000ms)

```
...
lotus_block_validation_ms_bucket{le="200"} 0
lotus_block_validation_ms_bucket{le="250"} 0
lotus_block_validation_ms_bucket{le="300"} 0
lotus_block_validation_ms_bucket{le="400"} 0
lotus_block_validation_ms_bucket{le="500"} 0
lotus_block_validation_ms_bucket{le="650"} 0
lotus_block_validation_ms_bucket{le="800"} 47
lotus_block_validation_ms_bucket{le="1000"} 129
lotus_block_validation_ms_bucket{le="2000"} 923
lotus_block_validation_ms_bucket{le="5000"} 1001
lotus_block_validation_ms_bucket{le="10000"} 1001
lotus_block_validation_ms_bucket{le="20000"} 1004
lotus_block_validation_ms_bucket{le="50000"} 1004
lotus_block_validation_ms_bucket{le="100000"} 1004
lotus_block_validation_ms_bucket{le="+Inf"} 1004
lotus_block_validation_ms_sum 1.4767458445909997e+06
lotus_block_validation_ms_count 1004 
```

**After**

As you can see, the distribution has changed. Average: 1266,592261904761905ms (-14%) (95% validated under 2000ms).

```
lotus_block_validation_ms_bucket{le="300"} 0
lotus_block_validation_ms_bucket{le="400"} 0
lotus_block_validation_ms_bucket{le="500"} 8
lotus_block_validation_ms_bucket{le="650"} 61
lotus_block_validation_ms_bucket{le="800"} 144
lotus_block_validation_ms_bucket{le="1000"} 323
lotus_block_validation_ms_bucket{le="2000"} 956
lotus_block_validation_ms_bucket{le="3000"} 1003
lotus_block_validation_ms_bucket{le="4000"} 1003
lotus_block_validation_ms_bucket{le="5000"} 1003
lotus_block_validation_ms_bucket{le="7500"} 1003
lotus_block_validation_ms_bucket{le="10000"} 1003
lotus_block_validation_ms_bucket{le="20000"} 1008
lotus_block_validation_ms_bucket{le="50000"} 1008
lotus_block_validation_ms_bucket{le="100000"} 1008
lotus_block_validation_ms_bucket{le="+Inf"} 1008
lotus_block_validation_ms_sum 1.2767256901239988e+06
lotus_block_validation_ms_count 1008
```

### GC/allocs stats

**Before**

```
go_gc_duration_seconds_sum 0.044846669
go_gc_duration_seconds_count 159
lotus_process_total_memory_alloc 8.691073448e+10
process_resident_memory_bytes 9.098326016e+09
```

**After**

```
go_gc_duration_seconds_sum 0.013976411
go_gc_duration_seconds_count 86
lotus_process_total_memory_alloc 7.444904292e+10
process_resident_memory_bytes 9.624408064e+09
```

## TODO

- [x] Unit tests for the cache.
- [x] A final round of live testing.


lib/blockstore/blockstore.go
```diff
@@ -32,35 +30,32 @@ func NewTemporarySync() *SyncStore {
...
 }
 
-// Alias so other packages don't have to import go-ipfs-blockstore
+// LotusBlockstore is a standard blockstore enhanced with a view operaiton
+// (zero-copy access to values), and potentially with cache management
+// operations, or others, in the future.
+type LotusBlockstore interface {
+	Blockstore
+	Viewer
+}
```
magik6k, line 55: Could we drop the alias below, and call this `Blockstore`? Would let us avoid all the renaming everywhere.

raulk: That's what I did initially, but I reverted because it causes ambiguity in those places that still take a standard Blockstore (e.g. Graphsync and others). We would end up with two `blockstore.Blockstore` interfaces that are different because one embeds the other...

magik6k: We can still alias that blockstore here, just as something like `BasicBlockstore` (to relay the fact that the non-viewer blockstore is not the one that should be used by default).

raulk: Fixes in 66336b9.

lib/blockstore/cache_freecache.go
```diff
@@ -0,0 +1,338 @@
...
+	ac.RUnlock()
+}
+
+var (
+	HasFalse = byte(0)
+	HasTrue  = byte(1)
+
+	// Sentinel values for false and true.
+	HasFalseVal = []byte{HasFalse}
+	HasTrueVal  = []byte{HasTrue}
```
vyzo, line 38: waaaaaat?

vyzo (commented):
first pass looks ok, but there are bound to be some bugz in this monster...

lib/blockstore/cache_freecache.go
raulk: Couldn't think of a better way :-(

vyzo:
@raulk this has developed conflicts...

raulk:
@vyzo they're tiny and I believe unrelated to the work you're doing. Let's resolve them when we have something to merge.

vyzo:
absorbed into #4992

Issue #4672, opened 2020-10-31
Lotus-miner does not distribute work efficiently


k178613133:
1. The worker has been waiting for tasks, and has not received the tasks distributed by miner
2020-10-31T03:47:06.935+0800    INFO    main    lotus-seal-worker/main.go:322   Opening local storage; connecting to master
2020-10-31T03:47:07.376+0800    INFO    main    lotus-seal-worker/main.go:372   Setting up control endpoint at 127.0.0.1:3456
2020-10-31T03:47:07.409+0800    INFO    main    lotus-seal-worker/main.go:436   Waiting for tasks

2. lotus-miner sectors list。
Four tasks are stuck in precommitwait. Is the status of this task abnormal

ID  State          OnChain  Active  Expiration  Deals
0   PreCommitWait  NO       NO      n/a         CC
0   PreCommitWait  NO       NO      n/a         CC
1   PreCommitWait  NO       NO      n/a         CC
6   PreCommitWait  NO       NO      n/a         CC
7   PreCommit1     NO       NO      n/a         CC
8   PreCommit1     NO       NO      n/a         CC
9   PreCommit1     NO       NO      n/a         CC

3. How long does it take for PC1 status? It's nearly seven hours now, and it's still in progress

ID  Sector  Worker  Hostname  Task  State    Time
3   9       0       smart681  PC1   running  6h45m13.7s
4   8       0       smart681  PC1   running  6h45m13.7s
5   7       0       smart681  PC1   running  6h45m13.7s

jennijuju:
Lotus version? this may be duplicated with #4661 and may be fixed by #4666 , could you please try the latest master?

jennijuju:
Could you please try the latest master? If the issue is still there, whats your lotus version? and whats your hardware spec?

could you please also provide `lotus-miner sealing jobs` and `lotus-miner sealing workers`?

jennijuju:
Please also provide `lotus-miner info all` for investigation, thanks!

arajasek:
We'll definitely need the Lotus version to look into this

arajasek:
Maybe fixed by 4778?

PR #4778, opened 2020-11-09
Fix scheduler lockups after storage is freed

magik6k:
Works locally


arajasek (approved):
Makes sense to me. Maybe addresses #4672?

Kubuxu (approved):

PR #3730, opened 2020-09-10
fix: actor method params deserialization error exit code

alanshaw:
This fixes the exit code for deserializing params for an actor method to be [`ErrSerialization(21)`](https://github.com/filecoin-project/go-state-types/blob/bbd537e93b39321de97b17ec8d726d46c16670fa/exitcode/common.go#L17-L18) not [`SysErrSenderInvalid(1)`](https://github.com/filecoin-project/go-state-types/blob/bbd537e93b39321de97b17ec8d726d46c16670fa/exitcode/reserved.go#L10-L15).

Issue here for context: https://github.com/filecoin-project/test-vectors/issues/93#issuecomment-689593946

I have a [PR open](https://github.com/filecoin-project/test-vectors/pull/133) with interoperable test vectors I've verified will start passing when this change is made.

PR #3478, opened 2020-09-02
fix: guard against self beneficiary when deleting actor

alanshaw:
This adds a check to ensure the calling actor cannot specify themselves as the beneficiary - they're about to be deleted so this stops us from losing their funds.

More context and test vector for this edge case here: https://github.com/filecoin-project/test-vectors/pull/92#discussion_r481180625

PR #3267, opened 2020-08-24
Fix exitcode if invoker hits invalid param

Kubuxu:

PR #4783, opened 2020-11-10
Small consensus-critical fixes in 1.2.0

arajasek:
Supersedes all of #3478 and #3730, and most of #3267

### DIFF
--- a/chain/vm/invoker.go
+++ b/chain/vm/invoker.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/filecoin-project/go-state-types/network"
+
 	"github.com/filecoin-project/lotus/chain/actors/builtin"
 
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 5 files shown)


magik6k (approved):

chain/vm/invoker.go
```diff
@@ -173,17 +175,21 @@ func (*ActorRegistry) transform(instance invokee) (nativeCode, error) {
...
 				param := reflect.New(paramT)
 
+				rt := in[0].Interface().(*Runtime)
 				inBytes := in[1].Interface().([]byte)
 				if err := DecodeParams(inBytes, param.Interface()); err != nil {
-					aerr := aerrors.Absorb(err, 1, "failed to decode parameters")
+					ec := exitcode.ErrSerialization
+					if rt.NetworkVersion() < network.Version7 {
+						ec = 1
+					}
```
austinabell, line 184: Although I don't think it's possible to hit, the encoding of params also uses this faulty exit code, perhaps do the same for this or make it fatal?

arajasek: @austinabell Thanks for the suggestion! If `SerializeParams` fails, it's treated as a fatal error in the one place it's called (`TryCreateAccountActor`), so the exitcode doesn't matter.

I'm changing its returned exitcode to `ErrSerialization` (since that's objectively correct), but I'm not putting it behind network upgrade logic.

PR #4785, opened 2020-11-10
async connect to bootstrappers

strahe:
If the connection of the first bootstrap peer times out, the others cannot be connected.

This can solve this problem and make the connection faster.


magik6k (approved):
Looks good, thanks!

PR #4786, opened 2020-11-10
cmd sectors commitIDs len debug

zgfzgf:
activeSet should be sset
https://github.com/filecoin-project/lotus/blob/3ec25cbece63d929a8ea06b3572e429393e47f54/cmd/lotus-storage-miner/sectors.go#L213-L215


whyrusleeping (approved):
Hah, nice catch. pretty sure nobody will ever notice, but good for correctness

magik6k (approved):

PR #4736, opened 2020-11-05
Refactor FundManager

dirkmc:
Currently in markets we have
- a mechanism to reserve and release funds needed for a particular deal (no on-chain messages)
- a mechanism to withdraw funds that must respect the reserved amount (calls withdraw on market actor)
- a mechanism to ensure reserved funds are available when starting a deal (calls add funds on market actor)

This PR unifies the logic for these mechanisms into a single interface:
```
// Reserve adds amt to `reserved`. If there are not enough available funds for
// the address, submits a message on chain to top up available funds.
Reserve(ctx context.Context, wallet, addr address.Address, amt abi.TokenAmount) (cid.Cid, error)

// Subtract from `reserved`.
Release(addr address.Address, amt abi.TokenAmount)

// Withdraw unreserved funds. Only succeeds if there are enough unreserved
// funds for the address.
Withdraw(ctx context.Context, wallet, addr address.Address, amt abi.TokenAmount) (cid.Cid, error)
```

TODO:
- [ ] Implement the logic change in go-fil-markets https://github.com/filecoin-project/go-fil-markets/pull/445
- [ ] Implement data migration from markets DealFunds to lotus FundManager

PR #4787, opened 2020-11-10
refactor: integrate new FundManager

dirkmc:
Integrate new FundManager

Depends on
- [x] https://github.com/filecoin-project/lotus/pull/4736
- [x] https://github.com/filecoin-project/go-fil-markets/pull/445


node/modules/client.go
```diff
@@ -108,15 +107,9 @@ func NewClientDatastore(ds dtypes.MetadataDS) dtypes.ClientDatastore {
 	return namespace.Wrap(ds, datastore.NewKey("/deals/client"))
 }
 
-type ClientDealFunds funds.DealFunds
-
-func NewClientDealFunds(ds dtypes.MetadataDS) (ClientDealFunds, error) {
-	return funds.NewDealFunds(ds, datastore.NewKey("/marketfunds/client"))
```
hannahhoward, line 114: Warning: This the datastore you probably need to exact a value from for the fund manager -- also, I was thinking about this last night -- it should be an Add, rather than a Set, because in the case of a miner, you want to add both values (a bug in the previous version -- a storage miner who also made deals would not have gotten the correct behavcior our of the previous DealFunds tracker in markets)

dirkmc:
LGTM 👍

PR #4804, opened 2020-11-11
Expand sched-diag; Command to abort sealing calls

magik6k:


cmd/lotus-storage-miner/sealing.go
```diff
@@ -253,3 +261,47 @@ var sealingSchedDiagCmd = &cli.Command{
 		return nil
 	},
 }
+
+var sealingAbortCmd = &cli.Command{
+	Name:      "abort",
+	Usage:     "Abort a running job",
+	ArgsUsage: "[call id]",
```
arajasek, line 268: Wanna call this `callid` instead, it kinda sounds like it takes 2 separate arguments `call` and `id` otherwise,

arajasek (approved):
LGTM

PR #4808, opened 2020-11-11
Fix panic in FundManager

dirkmc:


jsign:
I can confirm that the pipeline I ran to discover this stopped panicking, so looks good.

hannahhoward (approved):
I am a bit anxious about the sequence of things that gets us to this condition -- how would we end up with a reserved > available but no toAdd messages? Worth possible further investigation. That said, no panic is better than panic.

arajasek (approved):
Makes sense

dirkmc:
Update the description with a scenario in which this panic can occur

PR #4822, opened 2020-11-12
increase cross boundary judgment

zgfzgf:
optimize fetchMessage code and increase judgement

### DIFF
--- a/chain/sync.go
+++ b/chain/sync.go
@@ -1618,8 +1618,8 @@ func (syncer *Syncer) fetchMessages(ctx context.Context, headers []*types.TipSet
 				}
 
 				mx.Lock()
-				if requestResult != nil {
-					copy(batch[j+offset:], requestResult)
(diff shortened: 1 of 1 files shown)


chain/sync.go
```diff
@@ -1618,8 +1618,8 @@ func (syncer *Syncer) fetchMessages(ctx context.Context, headers []*types.TipSet
 				}
 
 				mx.Lock()
-				if requestResult != nil {
-					copy(batch[j+offset:], requestResult)
+				if requestResult != nil && nextI+len(requestResult) <= lastI {
```
magik6k, line 1621: I don't think this will work with partial responses (which we want to allow)

vyzo: this just breaks things.

vyzo (changes requested):
this is plainly wrong. please stop trying to micro-optimize at the expense of breaking code.

Issue #4823, opened 2020-11-12
v1.1.3-pre2 crashes node on startup


jsign:
When testing now v.1.1.3-pre2, I got one node working wihout problems, but in other two nodes they crash with the following tail logs:

Node 1:
```
2020-11-12T13:00:04.424Z        WARN    vm      vm/vm.go:522    Send actor error        {"from": "f1tudi422iiqzb6zrk3sw3ajuhky7fxi2avnx4ghi", "to": "f1d7mq36vf6osdhcd32i6k3wyb223mdjlxnafnala", "nonce": 9, "method": "0", "height": "229305", "error": "failed to transfer funds (RetCode=6):\n    github.com/filecoin-project/lotus/chain/vm.(*VM).send.func3\n        /lotus/chain/vm/vm.go:329\n  - transfer failed when deducting funds (78.989783599103 FIL): not enough funds (RetCode=6):\n    github.com/filecoin-project/lotus/chain/vm.(*VM).transfer\n        /lotus/chain/vm/vm.go:854"}
2020-11-12T13:00:04.425Z        WARN    vm      vm/vm.go:522    Send actor error        {"from": "f1l25nnnrhenzgtk6awm4cx3h3zpemo6mssvlducy", "to": "f1d7mq36vf6osdhcd32i6k3wyb223mdjlxnafnala", "nonce": 5, "method": "0", "height": "229305", "error": "failed to transfer funds (RetCode=6):\n    github.com/filecoin-project/lotus/chain/vm.(*VM).send.func3\n        /lotus/chain/vm/vm.go:329\n  - transfer failed when deducting funds (70.26928383038 FIL): not enough funds (RetCode=6):\n    github.com/filecoin-project/lotus/chain/vm.(*VM).transfer\n        /lotus/chain/vm/vm.go:854"}
2020-11-12T13:00:04.436Z        INFO    bs:sess session/session.go:459  No peers - broadcasting {"session": 3, "want-count": 64}
2020-11-12T13:00:04.459Z        INFO    sub     sub/incoming.go:98      new block over pubsub   {"cid": "bafy2bzaceauz5g4qj33e63qejzoqb5zlrc4ojghwbsza4y2k3b5fs5kchae24", "source": "12D3KooWH59Sk5LbFatAX8h3n3rrcot7wM9y1vdScB4nm3ytbNRr", "msgfetch": 0.03994851}
[...]

PR #4827, opened 2020-11-12
Fix: dont crash on startup if funds migration fails

dirkmc:
Don't crash on startup if funds migration fails.

I'm not 100% sure this is what we want to do. It may cause some in-progress deals to fail, but is probably better than crashing.

_Edit: The correct solution here is probably for the migration to subtract the amount of completed / failed deals from the reserved amount_


jsign:
I can confirm it fixes #4823

hannahhoward (approved):
LGTM -- I think this is the best fix we have, for the moment.

magik6k (approved):

PR #4849, opened 2020-11-13
Connection Gater support

vyzo:
Hooks the libp2p connection gater in lotus and adds API/CLI for managing connection gating rules.


api/api_common.go
```diff
@@ -46,6 +47,11 @@ type Common interface {
 	// usage and current rate per protocol
 	NetBandwidthStatsByProtocol(ctx context.Context) (map[protocol.ID]metrics.Stats, error)
 
+	// ConnectionGater API
+	NetBlockAdd(ctx context.Context, acl dtypes.NetBlockList) error
+	NetBlockRemove(ctx context.Context, acl dtypes.NetBlockList) error
+	NetBlockList(ctx context.Context) (dtypes.NetBlockList, error)
```
Kubuxu, line 53: I would prefer that API did not depend on `dtypes`.

vyzo: Sure, but where should I put the ACL type definition then?

vyzo: Moved the struct to `api/types.go` and removed the dependency.

magik6k (approved):

PR #4862, opened 2020-11-16
Modify vesting schedule post-calico

### LINKED ISSUE #4660: Fix circulating supply
One of the entries in the circulating supply lookup table isn't accurate and needs to be fixed. We might also cleanup some of the circ supply code as part of this issue.

arajasek:
There was an accounting mistake in the vesting table. It was reported but not fixed because of timeline and fixing it earlier will mean a spike in collateral for miners. This PR upgrades the vesting table to more closely match the actual multisig vesting.

Fixes #4660

### DIFF
--- a/chain/stmgr/stmgr.go
+++ b/chain/stmgr/stmgr.go
@@ -72,13 +72,17 @@ type StateManager struct {
 	// ErrExpensiveFork.
 	expensiveUpgrades map[abi.ChainEpoch]struct{}
 
-	stCache              map[string][]cid.Cid
-	compWait             map[string]chan struct{}
-	stlk                 sync.Mutex
-	genesisMsigLk        sync.Mutex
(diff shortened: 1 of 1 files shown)


arajasek:
Needs ✅  from @whyrusleeping

magik6k (approved):
Upgrade logic looks good

whyrusleeping (approved):
these numbers are slightly different than i'd expect, but i'm guessing @zixuanzh has some reason for this

zixuanzh:
What do you mean? This should be as transparent as possible with no room for meddling.

arajasek:
> these numbers are slightly different than i'd expect, but i'm guessing @zixuanzh has some reason for this

the reason is likely that @whyrusleeping is bad at excel and calculates bad numbers

Issue #4850, opened 2020-11-13
Regression: node halts with fatal error when importing snapshot (recoverable)


jsign:
While updating a bunch of Lotus nodes to v1.1.3, I had to also import some chain snapshots. I feel quite confident of never seeing this in <=1.1.2.

Looks like I get the following situation consistently:
1. `lotus daemon --import-snapshot ...`
2. Snapshot importing goes fine.
3. Metadata cache clearing goes fine
4. When its about to continue starting the node:
```
ERROR: initializing node: starting node: could not build arguments for function "github.com/filecoin-project/lotus/node/modules/lp2p".StartListening.func1 (/lotus/node/modules/lp2p/addrs.go:98): failed to build host.Host: could not build arguments for function "reflect".makeFuncStu
b (/usr/local/go/src/reflect/asm_amd64.s:12): failed to build lp2p.BaseIpfsRouting: could not build arguments for function "reflect".makeFuncStub (/usr/local/go/src/reflect/asm_amd64.s:12): failed to build dtypes.NetworkName: could not build arguments for function "reflect".makeFun
cStub (/usr/local/go/src/reflect/asm_amd64.s:12): failed to build *store.ChainStore: could not build arguments for function "reflect".makeFuncStub (/usr/local/go/src/reflect/asm_amd64.s:12): failed to build dtypes.ChainBlockstore: could not build arguments for function "reflect".ma
keFuncStub (/usr/local/go/src/reflect/asm_amd64.s:12): failed to build dtypes.ChainRawBlockstore: received non-nil error from function "reflect".makeFuncStub (/usr/local/go/src/reflect/asm_amd64.s:12): failed to open badger blockstore: Cannot acquire directory lock on "/data/.lotus
/datastore/chain".  Another process is using this Badger database.: resource temporarily unavailable
```
I can't imagine any other process messing with Badger or similar... mostly because this happened to more than 4 Lotus nodes in different VMs.

After the node stops, if you `lotus daemon` (without the snapshot importing flag), the node starts again without problems and all looks good.

In summary: after importing the snapshot the node halts. You can start again and things are normal.

raulk:
Thanks for pointing out the regression! Luckily you can start the node afterwards, so this is non-blocking as you duly point out. I'll take a look soon.

raulk:
[...]

PR #4872, opened 2020-11-16
fix badger double open on daemon --import-snapshot; chainstore lifecycle

### LINKED ISSUE #4428: dangling `reorgWorker` after snapshot import
https://github.com/filecoin-project/lotus/blob/151577f04ef264a63bedb86ef6bf527a64743c5d/chain/store/store.go#L190

Normally not an issue because the repo/store is not closed until the node stops but when importing we open/close the repo and the dangling thread fails with `datastore closed` error.

raulk:
This PR fixes https://github.com/filecoin-project/lotus/issues/4850. The problem was that the IdStore did not implement `io.Closer`, therefore the store remained open after the import. This required a release of go-ipfs-blockstore, which has now been integrated here.

I also noticed this error (which I think has been there since forever; EDIT: confirmed, this was already there! #4428) just after the import is done, and before we start the node:

```
2020-11-16T15:08:17.027Z        INFO    main    lotus/daemon.go:474     accepting [bafy2bzaceckovkpb5vua5s2sbltjg2lynikeehepztyjsf7n2nrehot36bwdc] as new head
2020-11-16T15:08:17.027Z        INFO    chainstore      store/store.go:500      New heaviest tipset! [bafy2bzaceckovkpb5vua5s2sbltjg2lynikeehepztyjsf7n2nrehot36bwdc] (height=240581)
2020-11-16T15:08:17.027Z        INFO    chainstore      store/store.go:687      failed to fetch right.Parents: get block bafy2bzaceaoyzhu6i2qpsi2efx7wp2z7kta54typ5fewd52mlsvg4azjrnkjg: badger blockstore closed
2020-11-16T15:08:17.027Z        ERROR   chainstore      store/store.go:420      computing reorg ops failed: get block bafy2bzaceaoyzhu6i2qpsi2efx7wp2z7kta54typ5fewd52mlsvg4azjrnkjg: badger blockstore closed
```

This happens because the `SetHead` inside `ImportChain` triggers a reorg, but by the team the event arrives to the event loop, the badger store is closed. There are two problems here:

1. The chainstore does not accept a context, thus it the reorg runner has no way to stop. Currently, this would leak a goroutine. Fixes #4428.
2. This forced `SetHead` should not trigger a reorg; we can use the newly-added `ForceHeadSilent` operation to set the head without a reorg.

This PR fixes both those things too.

### DIFF
--- a/chain/store/store.go
+++ b/chain/store/store.go
@@ -382,7 +382,7 @@ func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipS
 // particular tipset to carry out a benchmark, verification, etc. on a chain
 // segment.
 func (cs *ChainStore) ForceHeadSilent(_ context.Context, ts *types.TipSet) error {
-	log.Warnf("(!!!) forcing a new head silently; only use this only for testing; new head: %s", ts)
+	log.Warnf("(!!!) forcing a new head silently; new head: %s", ts)
 
 	cs.heaviestLk.Lock()
 	defer cs.heaviestLk.Unlock()
--- a/chain/store/store.go
+++ b/chain/store/store.go
@@ -137,7 +137,7 @@ type ChainStore struct {
 }
 
 // localbs is guaranteed to fail Get* if requested block isn't stored locally
-func NewChainStore(bs bstore.Blockstore, localbs bstore.Blockstore, ds dstore.Batching, vmcalls vm.SyscallBuilder, j journal.Journal) *ChainStore {
+func NewChainStore(ctx context.Context, bs bstore.Blockstore, localbs bstore.Blockstore, ds dstore.Batching, vmcalls vm.SyscallBuilder, j journal.Journal) *ChainStore {
 	c, _ := lru.NewARC(DefaultMsgMetaCacheSize)
 	tsc, _ := lru.NewARC(DefaultTipSetCacheSize)
 	if j == nil {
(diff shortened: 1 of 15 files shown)


magik6k (approved):

chain/store/store.go
```diff
@@ -383,7 +382,7 @@ func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipS
 // particular tipset to carry out a benchmark, verification, etc. on a chain
 // segment.
 func (cs *ChainStore) ForceHeadSilent(_ context.Context, ts *types.TipSet) error {
-	log.Warnf("(!!!) forcing a new head silently; only use this only for testing; new head: %s", ts)
```
whyrusleeping, line 386: eh?

Kubuxu:
Personally I would prefer if chainstore had a Close method instead of relying on the context for the lifetime.

chain/store/store.go
raulk: ?

raulk:
@Kubuxu agree; addressed in 0c6072a.

Kubuxu (approved):

chain/store/store.go
magik6k: There was no valid use for this outside testing before. When importing snapshot it doesn't make sense to emit reorg notifications as nothing is listening on that chainstore, and everything that cares about chain state does check current state when subscribing

tldr this is fine

raulk:
@magik6k wanna merge this?

magik6k (approved):

PR #4873, opened 2020-11-16
Support seal proof type switching

magik6k:


magik6k (commented):
(looks good to me)

arajasek (approved):
Looks like it should work, and apparently it does.

top-org:
Could anyone share to  Filecoin Community about why Filecoin Team  introduce new proof type at v1.2.0 ? It looks like that just wrap proof type from function into Sector.StorageRef structure  ,it is better to have it but we not see any algorithm-related  change at filecoin-ffi and Rust layer(Storage Proof 5.4.0) .  so what is point/purpose to introduce those huge change ?

dkkapur:
@top-org - A bug was identified in Filecoin’s Proving Subsystem which allowed miners to exploit parallelism in SDR. Specifically, the expander graph was truncated to always fall in the first 16GiB of a sector due to an integer truncation bug in the expander parents generation. The bug was reported and fixed by the rust-fil-proofs team, resulting in a new proof type being released (5.4.0). Old sectors affected cannot be added on chain since epoch 272400. 

Thanks to @TyreeRoby for disclosing this issue to us and working with us to solve this in a timely manner.

Kubuxu:
For future reference the underlying fix is here: https://github.com/filecoin-project/rust-fil-proofs/commit/0d17d7466f40e1228a4bab25f8b4861cb0d2da4d#diff-f9e7b9c546dbd93a3981eb032af33d9f52f4e7d4403192d1f26f77c886d5fa31L158-L182

PR #4874, opened 2020-11-16
Fix cid recording issue

### LINKED ISSUE #4856: File transfers over 8GB will fail in data transfers, or less depending on data chunking strategy
**Describe the bug**

Any file transfer over 8GB, assuming a 1MB chunking strategy, will fail to complete data transfer on miner's side, even if data transfer completes.

If you are importing from an IPFS blockstore and using default 256KB chunking strategy, anything over 2GB will fail

**To Reproduce**

1. Create a deal for a file over 8GB
2. On miner side, even after data transfer completes, miner will stay stuck at StorageDealTransferring

**Expected behavior**

The data transfer completes

All lotus versions >= 1.0.0

**Additional context**

If you look in the miner logs, you find an error level message that reads contains `Executing event planner failed: Slice value in field t.ReceivedCids was too long`

What's happening here is as follows:

- We are recording CIDs received on the miner in order to be able to resume a data transfer by passing the list of CIDs already traversed to go-graphsync in the case of a restart
- We are recording them in the ChannelState, which is a CBOR encoded data structure in the data store where the CIDs received is an array data field, and the cbor encoding is done by cbor-gen
- Cbor-gen has a hard limit of 8192 on arrays

hannahhoward:
# Goals

fix #4856 

# Implementation

See corresponding dependent PR -- https://github.com/filecoin-project/go-data-transfer/pull/118

The only substantive change here is that we need to setup a directory in which to place these cid recordings, setup us as a "data-transfer" subdirectory under the repo path.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/filecoin-project/go-bitfield v0.2.3-0.20201110211213-fe2c1862e816
 	github.com/filecoin-project/go-cbor-util v0.0.0-20191219014500-08c40a1e63a2
 	github.com/filecoin-project/go-crypto v0.0.0-20191218222705-effae4ea9f03
-	github.com/filecoin-project/go-data-transfer v1.1.0
+	github.com/filecoin-project/go-data-transfer v1.2.0
 	github.com/filecoin-project/go-fil-commcid v0.0.0-20200716160307-8f644712406f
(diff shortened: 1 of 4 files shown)


arajasek (approved):
LGTM!

magik6k (approved):
Not ideal, but better

PR #4875, opened 2020-11-16
Fix init actor address map diffing

frrist:
### What
Replace init actor address map diffing with generic adt diffing logics
Optimize other diffing methods by calling Diff* iff adt CID's are different.

### DIFF
--- a/chain/events/state/predicates.go
+++ b/chain/events/state/predicates.go
@@ -1,7 +1,6 @@
 package state
 
 import (
-	"bytes"
 	"context"
 
 	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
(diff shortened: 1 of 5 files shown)


chain/events/state/predicates.go
```diff
@@ -419,179 +417,17 @@ type AddressPair struct {
...
-		err = newState.ForEachActor(func(newId abi.ActorID, newAddress address.Address) error {
-			newIdAddress, err := address.NewIDAddress(uint64(newId))
-			if err != nil {
-				return err
-			}
-
-			_, found, err := newState.ResolveAddress(newAddress)
-			if err != nil {
-				return err
-			}
```
frrist, line 575: Motivation for change lies in the bug here. This should be looking in the oldState to see if a new address has been added. New ADT logic avoids this bug.

magik6k (approved):

PR #4888, opened 2020-11-17
Avoid sending messages to the zero BLS address

arajasek:


magik6k (approved):

vyzo (commented):
code looks fine, but what's the rationale for this? bug prevention from accidental sends?

Kubuxu:
@vyzo `0` BLS key doesn't have all the properties we require from BLS keys.

Kubuxu (commented):
SGWM

PR #4890, opened 2020-11-17
Add error codes to worker return 

magik6k:
We're bumping API versions in 1.2.0 already, so getting this in now is nice because we can avoid breaking worker APIs again

This is part 1 of 2 of handling things like failing to allocate storage more gracefully. Part 2 will be handling those errors at the scheduler level, just returning tasks which got Temp errors to the scheduler instead of failing them


arajasek (approved):
LGTM

PR #4891, opened 2020-11-17
integrate upgrade codenames into upgrade schedule.

raulk:
This removes duplication and maintenance overhead. I _think_ it's OK to add the genesis "upgrade/protocol version" there, but would love a confirmation.


magik6k:
This will need to get rewritten for https://github.com/filecoin-project/lotus/pull/8158. Closing for now

PR #4896, opened 2020-11-17
Update go to 1.15.5

Kubuxu:


arajasek (approved):
Works for my Macbook and is therefore g2g.

PR #4927, opened 2020-11-19
change task priority

### LINKED ISSUE #4926: task priority may problem
https://github.com/filecoin-project/lotus/blob/648f02cbd6fcf6d41d9c513a3dc9e5a7068cd238/extern/sector-storage/sealtasks/task.go#L23-L24
should be:
TTCommit1:      3, 
 TTCommit2:      2,

zgfzgf:
fix #4926

### DIFF
--- a/extern/sector-storage/sealtasks/task.go
+++ b/extern/sector-storage/sealtasks/task.go
@@ -3,14 +3,12 @@ package sealtasks
 type TaskType string
 
 const (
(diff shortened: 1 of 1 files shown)


magik6k:
This was correct, we want to give Commit1 more priority, because it's very short, and allows us to start scheduling more C2 tasks (the scheduler can make better decisions on what C2 tasks to process first)

PR #4928, opened 2020-11-19
order resource table, put 2k last

zgfzgf:
read code better


magik6k:
Sorry, those cosmetic PRs are not helpful, and still need careful review, which takes more time than it's worth

PR #4929, opened 2020-11-20
Remove all temp file generation for deals

### LINKED ISSUE #4898: Remove temporary piece file generation from go-fil-markets for ONLINE deals
# What

Currently when we verify CommP on the provider side, we read the data for the piece out of the blockstore and write the CAR file as a temporary file to disk for the piece, stored in the .lotusminer directory, and feed it later to the sector storage interface, deleting it only later. This can cause problems for miners who only have large storage allocated towards the sectors themselves.

It would be better to:
1. Just write to an OS pipe like we do on the client side when generating commP.
2. When we handoff to the sector storage interface, regenerate the CAR file, writing to an OS pipe again but this time feeding the reader to the sector storage interface.

# Things to consider

- The rarely used feature to write an index of all CIDs in a piece, and their offsets in the CAR file, using hooks in writing the car file on each new block (future ticket: remove this feature unless someone is going to use it)
- We need to still enable the filestore flow for offline deals, at least for the moment

### LINKED ISSUE #4899: Extract Lotus CommP writer and use when calculating CommP in markets
# What

Currently when we calculate commP in go-fil-markets we pass an io.Reader of the whole piece to the FFI. The FFI seems to write a temp file the size of the whole piece to the OS temp directory during the process of calculating CommP. Instead of using the FFI directly, we can use Lotus's commP writer: https://github.com/filecoin-project/lotus/blob/master/lib/commp/writer.go which chunks up calls to the FFI when writing tempfiles so that it doesn't write a file the size of the whole piece to OS temp directory, which may not have a lot of space in a miner setup for large pieces.

# How

- extract lib/commp writer from lotus -- I think it would be nice to make a new shared repo that both go-fil-markets and lotus use -- also it may make sense to actually extract pieceIO from go-fil-markets as well and combine it. Though perhaps just extracting the commP writer as a start from lotus is a good place to be.

- on top of https://github.com/filecoin-project/go-fil-markets/pull/449, rewrite GeneratePieceCommitment in PieceIO to use the commP writer. This will probably entail not re-using the GeneratePieceReader function in GeneratePieceCommitment any more, and instead calling PrepareCar directly and passing the extract commP writer directly to PreparedCar.Dump (then calling writer.Sum)

hannahhoward:
# Goals

Reduce space required for temporary storage by generating CommP in pieces and removing use of temporary files in markets for online deals (still there for offline deals for now)

# Implementation

- Extracts commP writer (as well as zerocomm and a couple other utilities) to go-commp-utils, which also now includes pieceio
- Uses updated markets

# For discussion

I removed any internal copies of the code in go-commp-utils. However, for the ffiwrapper, which is only a partial extraction, it results in some slightly awkward imports. Alternatively, we could just live with the extra copies of the files.

fix #4898 fix #4899


extern/sector-storage/ffiwrapper/sealer_cgo.go
```diff
@@ -20,9 +20,10 @@ import (
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/specs-storage/storage"
 
+	"github.com/filecoin-project/go-commp-utils/ffiwrapper"
```
dirkmc, line 23: Let's rename this import as it shadows the name of the package this file is in

dirkmc (approved):
LGTM 👍

magik6k (approved):

Kubuxu (approved):

PR #4930, opened 2020-11-20
badger options: enable IndexCache and CompactL0OnClose.

raulk:
Unfortunately, badger assumes that no index cache means to keep all indices in memory. This is counter-intuitive and a poor choice for a default on their end, and it's what's likely causing memory pressure growth in correlation with store growth (the more data, the more SSTables to store in memory).

https://github.com/dgraph-io/badger/blob/d5a25b83fbf4f3f61ff03a9202e36f5b75544426/table/table.go#L400-L401

Profiles show a large amount of retained memory originates here as a result. (see heap profile below)

This PR sets the index cache, in order to bound the memory used for indices and prevent linear growth of memory use with datastore growth.

It also sets `CompactL0OnClose = true`, to enable read-only open by processes like Sentinel.

### heap profile

image


raulk:
Further notes. We can encourage users to run with `madv_dontneed=1`, since Go 1.16 will default to that anyway: https://go-review.googlesource.com/c/go/+/267100/.

raulk:
Absorbed into #5101.

PR #4938, opened 2020-11-20
Return total power when GetPowerRaw doesn't find miner claim

arajasek:
This is correcter, and means we don't clumsily fail on commands like `lotus state power <maddr>` when `maddr` isn't found in claims.


magik6k (commented):
Is it more correct though? I'd expect this to error if I pass some random, for example, multisig address

magik6k (approved):

Issue #4900, opened 2020-11-18
Breakup StorageDealSealing and OnDealSectorCommitted in go-fil-markets


hannahhoward:
# What

Currently, one a deal is published, it goes to StorageDealSealing and OnDealSectorCommitted is called that monitors and waits for the sealing process is complete. This gives very non-granular information about the status of sealing. Moreover, there's a bug in the logic of OnDealSectorCommitted as it relates to restarts: OnDealSectorCommitted works through the EventsAPI.Called interface and relies on seeing two seperate chain messages: the pre-commit and the prove-commit. If the pre-commit occurs, and then the node is restarted, because the code relies on seeing both messages, and after restart pre-commit has already occurred, the OnDealSectorCommitted will never see that message, meaning it can't interpret the prove-commit correctly (as it relies on extracting the sector ID from the pre-commit)

# How

Breakup StorageDealSealing into two seperate states with two seperate state entry funcs (break up VerifyDealActivated on both client and miner side). Two potential state names:

StorageDealAwaitingPreCommit
StorageDealSealing
(just rough, feel free to arrive at better ones)

Breakup OnDealSectorCommitted into OnDealSectorPreCommitted and OnDealSectorCommitted

The callback for OnDealSectorPreCommitted should take a sector ID and should save the sector ID on the stored deal.

The new OnDealSectorCommitted should take a SectorID as well as the other parameters (not sure if all or only some are needed)

There will need to be a seperate PR to break these up in Lotus

dineshshenoy:
This will
Fix a bug where after a restart the client might not notice that the miner has started sealing a deal
Give the client insight into an extra state before the miner actually starts sealing: pre-com

PR #4945, opened 2020-11-20
markets - separate watching for pre-commit from prove-commit

dirkmc:
Depends on https://github.com/filecoin-project/go-fil-markets/pull/453
Fixes https://github.com/filecoin-project/lotus/issues/4900

TODO:
- [x] Update test for OnDealSectorCommitted
- [x] Add test for OnDealSectorPreCommitted
- [x] Merge https://github.com/filecoin-project/go-fil-markets/pull/453, release and update lotus go.mod with the release version

### DIFF
--- a/markets/storageadapter/ondealsectorcommitted.go
+++ b/markets/storageadapter/ondealsectorcommitted.go
@@ -20,12 +20,7 @@ type sectorCommittedEventsAPI interface {
 	Called(check events.CheckFunc, msgHnd events.MsgHandler, rev events.RevertHandler, confidence int, timeout abi.ChainEpoch, mf events.MsgMatchFunc) error
 }
 
-type sectorPreCommitAPI interface {
-	getCurrentDealInfoAPI
-	StateSectorGetInfo(ctx context.Context, maddr address.Address, n abi.SectorNumber, tsk types.TipSetKey) (*miner.SectorOnChainInfo, error)
-}
-
-func OnDealSectorPreCommitted(ctx context.Context, api sectorPreCommitAPI, eventsApi sectorCommittedEventsAPI, provider address.Address, dealID abi.DealID, proposal market.DealProposal, publishCid *cid.Cid, cb storagemarket.DealSectorPreCommittedCallback) error {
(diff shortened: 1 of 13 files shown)


magik6k (commented):
Looks good, but it seems we want filecoin-project/go-fil-markets#453 in before this lands

markets/storageadapter/ondealsectorcommitted.go
```diff
@@ -20,110 +20,201 @@ type sectorCommittedEventsAPI interface {
...
-		case miner.Methods.ProveCommitSector:
-			if msg == nil {
-				log.Error("timed out waiting for deal activation... what now?")
-				return false, nil
+			msgDealIDs = params.DealIDs
+			sectorNumber = params.SectorNumber
+		} else {
+			// If it's a prove-commit message, the parameters don't have deal IDs,
+			// just a sector number
+			var params miner.ProveCommitSectorParams
```
hannahhoward, line 94: There is a lot of logic in here to deal with ProveCommitSector messages -- I'm not clear why we're dealing with this. As I see it, the only case of isActive = true we need to deal with is in the called function.

The goal of #4900 is designed to fix the case where a restart happens between pre-commit and prove-commit. We don't need this code for that.

What is the scenario that causes us to miss the pre-commit message entirely? Even if we shut down the node a long time, we will see the pre-commit when we sync the chain.

hannahhoward (changes requested):
There's a lot of logic in here to deal with the ProveCommitSector message in OnDealSectorPreCommitted -- why would we need this? The problem 4900 captures is that we weren't persisting the progression from PreCommit to PostCommit, so, theoretically, if node restarts in between, we lose the sector number from the pre-commit and never get to the end. It's not to cover for the possibility of missing PreCommit entirely. Unless @magik6k wants to say otherwise, I think we can remove the ProveCommitSector logic (which is complicated) from OnDealSectorPreCommitted.

hannahhoward:
Oh dear, I think I broke this. So now I need to fix the problem actually.

hannahhoward (approved):

hannahhoward:
I've fixed the history... #oops

And I downgraded my comment from a "request changes" to a suggestion, it's up to you if you want to change. Either way, you can merge at your leisure pending @magik6k's approval

markets/storageadapter/ondealsectorcommitted.go
dirkmc: Ah yes that's a good point 👍 

Nice catch @hannahhoward. I think it is worth the effort to refactor the code because it will simplify things significantly. I will do so today.

magik6k (approved):
(testing on my miner; will merge in a bit)

PR #4970, opened 2020-11-23
add send from != to

### LINKED ISSUE #4969: lotus send from == to problem
aa73d1603b665244c7d6d689965908d

zgfzgf:
fix #4969

### DIFF
--- a/cli/send.go
+++ b/cli/send.go
@@ -102,10 +102,6 @@ var sendCmd = &cli.Command{
 			fromAddr = addr
 		}
 
-		if fromAddr == toAddr {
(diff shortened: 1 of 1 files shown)


jennijuju:
There is nothing wrong for a sender do a self transfer if they want to IMO.

zgfzgf:
It is meaningless to transfer money to oneself,
which wastes gas charge and network resources

magik6k:
Self-transfers aren't invalid, and can be useful when you need to put bits of data on-chain for whatever reason (web3 apps, smart-contract integration in other blockchains, proving address ownership, etc..)

zgfzgf:
lotus send is may not
others are OK

PR #4976, opened 2020-11-23
fix get sector bug

zlhwdsz:
This PR aims to fix a panic bug when check nonexistent sectors with the following command:

`lotus state sector [miner address] [sector number]`

### DIFF
--- a/cli/state.go
+++ b/cli/state.go
@@ -1759,6 +1759,9 @@ var stateSectorCmd = &cli.Command{
 		if err != nil {
 			return err
 		}
+		if si == nil {
+			return xerrors.Errorf("sector %d for miner %s not found", sid, maddr)
+		}
 
 		fmt.Println("SectorNumber: ", si.SectorNumber)
 		fmt.Println("SealProof: ", si.SealProof)


chain/actors/builtin/miner/v2.go
```diff
@@ -83,9 +83,12 @@ func (s *state2) PreCommitDeposits() (abi.TokenAmount, error) {
 
 func (s *state2) GetSector(num abi.SectorNumber) (*SectorOnChainInfo, error) {
 	info, ok, err := s.State.GetSector(s.store, num)
-	if !ok || err != nil {
+	if err != nil {
 		return nil, err
 	}
+	if !ok {
+		return nil, xerrors.Errorf("sector %d not found", num)
+	}
```
magik6k, line 91: This isn't correct, some callers of this method expect a nil value when the sectors isn't found

zlhwdsz: Got it. I modified the code elsewhere instead, please review it again.

Kubuxu (approved):

magik6k (approved):

PR #4984, opened 2020-11-24
fix log format

yaohcn:


chain/events/events.go
```diff
@@ -129,7 +129,7 @@ func (e *Events) listenHeadChangesOnce(ctx context.Context) error {
 	}
 
 	if err := e.tsc.add(cur[0].Val); err != nil {
-		log.Warn("tsc.add: adding current tipset failed: %w", err)
+		log.Warnf("tsc.add: adding current tipset failed: %w", err)
```
Kubuxu, line 132: Can you also fix the `%w` format here, it should be `%v`.

Kubuxu (approved):
SGWM

chain/events/events.go
magik6k, line 132:
```suggestion
		log.Warnf("tsc.add: adding current tipset failed: %v", err)
```

magik6k (approved):
Thanks!

PR #4985, opened 2020-11-24
fullnode actor cmd

waynewyang:
can use independent full nodes to set owner, worker and other address.It does not have to be a miner to operate.

### DIFF
--- a/cli/actor.go
+++ b/cli/actor.go
@@ -1,0 +1,409 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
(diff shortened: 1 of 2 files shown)


cli/actor.go
```diff
@@ -0,0 +1,409 @@
...
+	Usage: "Modify actor info",
+	Subcommands: []*cli.Command{
+		workerCmd,
+		ownerCmd,
+		controlSetCmd,
+		withdrawCmd,
+	},
+}
+
+var workerCmd = &cli.Command{
```
magik6k, line 33: We should use the same code for `lotus miner ...` and `lotus-miner actor ...` commands, just with the difference that lotus-miner commands would set a default actor

magik6k (commented):
We shouldn't duplicate code

Stebalien:
Closing because it looks like there are unaddressed review comments. Ask to reopen if you want to revive this.

Issue #4701, opened 2020-11-03
Add a "compact" command


arajasek:
Inspired by #4547, our docs currently say:

```
## Compacting the chain data

It is possible to "prune" the current chain data used by Lotus to reduce the disk footprint of a Lotus node by resyncing from a minimal snapshot. This is a quick overview of the steps:

1. Stop the lotus daemon, i.e. `lotus daemon stop`)
2. Remove the contents of the `datastore/chain/` folder in the Lotus path, i.e. `rm -rf ~/.lotus/datastore/chain/*`.
3. Start the daemon using a minimal snapshot as explained above, i.e. `lotus daemon --import-snapshot https://fil-chain-snapshots-fallback.s3.amazonaws.com/mainnet/minimal_finality_stateroots_latest.car`.
```

We probably shouldn't be telling people to directly delete folders. Adding some `lotus compact` command that does this in one step is probably the best thing to do.

jennijuju:
Another solution is actually have an option to trigger regular chain datastore pruning

arajasek:
This _should_ largely get implemented by #4992

jennijuju:
Add documentation for split blockstore, which implements this feature by nature.

vyzo:
Note that as part of #5808 I am adding a `lotus-shed` command to gc the hotstore.
We can do this as an additional command in that vein, that garbage collects the coldstore.
We will have to finish #5808 first however, it still has a couple of small issues with hotstore misses.

parkan:
running the daemon with ` --import-snapshot` a second time appears to blow away the chain data and restart the sync, would that be equivalent?

arajasek:
Splitstore is the correct way to do this, we can close this.

PR #4992, opened 2020-11-24
hot/cold blockstore segregation (aka. splitstore)

vyzo:
This is the implementation of splitstore, a hot/cold segregated blockstore that can be enabled with a feature flag.

On top of https://github.com/filecoin-project/lotus/pull/5695

How to enable:
In `.lotus/config.toml` add the following section:
```
[Chainstore]
  EnableSplitstore = true
```

TODO:
- [x] dynamic bloom filter sizing
- [x] move lmdb support out to a branch, to reduce surface and keep it out of the dependency graph.
  - moved to https://github.com/filecoin-project/lotus/pull/5699
- [ ] appropriate options for badger hotstore; it currently uses as the coldstore, but we need to check if those actually make sense and tune; especially verify that value log is GC'ed and what is the correct setting for the GC threshold.
- [x] batch delete support in blockstores.
- [x] splitstore unit test.
- [ ] Properly open the Splitstore (not just the Coldstore) in lotus-shed when enabled by the configuration.

Follow up:
- optimize WalkSnapshot -- see #4994
- ~implement BatchDelete in Blockstores~ -- done in #5703


chain/store/splitstore.go
```diff
@@ -0,0 +1,334 @@
...
+
+func (s *SplitStore) HashOnRead(enabled bool) {
+	s.hot.HashOnRead(enabled)
+	s.cold.HashOnRead(enabled)
+}
+
+func (s *SplitStore) View(cid cid.Cid, cb func([]byte) error) error {
+	err := s.hot.View(cid, cb)
+	switch err {
+	case bstore.ErrNotFound:
```
Kubuxu, line 166: `ErrNotFound` is in `bstore2` (aka `lib/blockstore`) so you can use just it.
In essence, `lib/blockstore` is our adjusted version of `go-ipfs-blockstore`.

vyzo: aaah, nice; will do.

chain/store/splitstore.go
```diff
@@ -0,0 +1,334 @@
...
+	return nil
+}
+
+func (s *SplitStore) HeadChange(revert, apply []*types.TipSet) error {
+	s.curTs = apply[len(apply)-1]
+	epoch := s.curTs.Height()
+	if epoch-s.baseEpoch > CompactionThreshold && !s.compacting {
+		s.compacting = true
+		go func() {
+			defer func() { s.compacting = false }()
```
Kubuxu, line 212: data race

vyzo: yes, but I think it's totally benign; at worst we'll trigger compaction in the next head change.

vyzo: will fix anyway :)

chain/store/splitstore.go
vyzo: done.

chain/store/splitstore.go
vyzo: done.

chain/store/splitstore/snoop.go
```diff
@@ -0,0 +1,161 @@
...
+	"os"
+
+	"github.com/bmatsuo/lmdb-go/lmdb"
+
+	cid "github.com/ipfs/go-cid"
+
+	"github.com/filecoin-project/go-state-types/abi"
+)
+
+var TrackingStoreMapSize int64 = 1 << 34 // 16G
```
raulk, line 15: mmapping 16GiB might be quite a bit for this store. We should really implement step-wise incremental mmapping like bbolt. https://github.com/etcd-io/bbolt/blob/f6be82302843a215152f5a1daf652c1ee5503f85/db.go#L16-L17

chain/store/splitstore/snoop.go
```diff
@@ -0,0 +1,161 @@
...
+				if err != nil {
+					if lmdb.IsNotFound(err) {
+						return nil
+					}
+
+					return err
+				}
+
+				select {
+				case ch <- cid.NewCidV1(cid.Raw, k):
```
raulk, line 143: NewCidV1 copies the byte buffer, so passing in the mmapped key is fine here.

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+
+	// Phase 1: marking
+	log.Info("marking live objects")
+	startMark := time.Now()
+
+	// Phase 1a: mark all reachable CIDs in the hot range
+	curTs := s.curTs
+	epoch := curTs.Height()
+	coldEpoch := s.baseEpoch + build.Finality
+	err = s.cs.WalkSnapshot(context.Background(), curTs, epoch-coldEpoch+1, false, false,
```
raulk, line 273: Definitely need to address #4994 given that we're relying on WalkSnapshot in production code now!

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+	logging "github.com/ipfs/go-log/v2"
+
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/lotus/build"
+	"github.com/filecoin-project/lotus/chain/store"
+	"github.com/filecoin-project/lotus/chain/types"
+	bstore "github.com/filecoin-project/lotus/lib/blockstore"
+)
+
+const CompactionThreshold = 5 * build.Finality
```
raulk, line 24: Naming nit: we should probably rename this and `compact` to `MigrationThreshold` and `migrate`, as that's effectively what we're doing.

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+func (s *SplitStore) Has(cid cid.Cid) (bool, error) {
+	has, err := s.hot.Has(cid)
+
+	if err != nil {
+		return false, err
+	}
+
+	if has {
+		return true, nil
+	}
```
raulk, line 65: Can be simplified:

```suggestion
	if err != nil || has {
		return has, err
	}
```

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+
+	switch err {
+	case nil:
+		return blk, nil
+
+	case bstore.ErrNotFound:
+		return s.cold.Get(cid)
+
+	default:
+		return nil, err
```
raulk, line 81: When short-circuiting, I'm generally a fan of returning the original values untouched. As we add layers and layers that delegate one to another, intermediate layers making assumptions can lead to future bugs that are hard to track down.

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+
+	switch err {
+	case nil:
+		return size, nil
+
+	case bstore.ErrNotFound:
+		return s.cold.GetSize(cid)
+
+	default:
+		return 0, err
```
raulk, line 96: Same here, just return `size, err`?

chain/store/splitstore/snoop.go
```diff
@@ -0,0 +1,161 @@
...
+var TrackingStoreMapSize int64 = 1 << 34 // 16G
+
+type TrackingStore interface {
+	Put(cid.Cid, abi.ChainEpoch) error
+	PutBatch([]cid.Cid, abi.ChainEpoch) error
+	Get(cid.Cid) (abi.ChainEpoch, error)
+	Delete(cid.Cid) error
+	Keys(context.Context) (<-chan cid.Cid, error)
+	Close() error
+}
```
raulk, line 24: Does this need to mimic the blockstore interface? I don't think so. Can simplify this to:

```go
type Tracker interface {
    // called on bs.Put, bs.PutMany, bs.Get
    Track(abi.ChainEpoch, cids... cid.Cid) error
    // called on delete (actually unused)
    Untrack(cids... cid.Cid) error
    // iterator
	GetTracked(context.Context) (<-chan cid.Cid, error)
    Close() error
}
```

?

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+	baseEpoch abi.ChainEpoch
+	curTs     *types.TipSet
+
+	cs *store.ChainStore
+	ds dstore.Datastore
+
+	hot  bstore.Blockstore
+	cold bstore.Blockstore
+
+	snoop TrackingStore
```
raulk, line 40: nit: I know I proposed the pair of names "sweep / snoop", but I'm totally fine calling this tracker all throughout. There's really no good reason for having two names for it ;-)

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+	default:
+		return 0, err
+	}
+}
+
+func (s *SplitStore) Put(blk blocks.Block) error {
+	epoch := s.curTs.Height()
+	err := s.snoop.Put(blk.Cid(), epoch)
+	if err != nil {
+		return err
```
raulk, line 104: Hum, I'd prefer to have some kind of graceful degradation here if the snooper fails. Basically we should warn loudly that we've degraded, and disable GC/migration entirely when this happens, instead of failing the `Put` and interrupting service.

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+
+	ch := make(chan cid.Cid)
+	go func() {
+		defer cancel()
+		defer close(ch)
+
+		for _, in := range []<-chan cid.Cid{chHot, chCold} {
+			for cid := range in {
+				select {
+				case ch <- cid:
```
raulk, line 149: This will be problematic. Either we open the second iterator when the first one finishes, or we open both and interleave the receives, thus consuming in parallel.

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+		for s.isCompacting() {
+			time.Sleep(time.Second)
+		}
+	}
+
+	return s.env.Close()
+}
+
+func (s *SplitStore) HeadChange(revert, apply []*types.TipSet) error {
+	s.curTs = apply[len(apply)-1]
```
raulk, line 216: This is racy for two reasons.
1. Easy one: the data race. Suggestion: turn this into an atomic.Value.
2. Harder one: I _think_ the head change notification is asynchronous, so we might start receiving Puts and Gets before we update `s.curTs`.

@Kubuxu ideas to solve (2)?

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,414 @@
...
+	epoch := s.curTs.Height()
+	if epoch-s.baseEpoch > CompactionThreshold && !s.isCompacting() {
+		s.setCompacting(true)
```
raulk, line 219: This is racy. Lock is released in between read and write. An atomic CAS would work really well here.

raulk (commented):
Initial sweep; need to delve into the migration logic proper.

chain/store/splitstore/snoop.go
vyzo: dunno, I just set it to 16G to be safe, it shouldn't really matter.

We can tackle incremental mapping on a later pr.

chain/store/splitstore/splitstore.go
vyzo: nah, I like compaction for this :)

chain/store/splitstore/splitstore.go
vyzo: If the tracker fails we can't effectively do compaction and the hotstore will eventually fill leading to catastrophic failure.
I don't see how graceful degradation helps here.

chain/store/splitstore/splitstore.go
vyzo: why? They are separate stores, it shouldn't be a problem.

chain/store/splitstore/splitstore.go
vyzo: There is no data race -- reorg notifications from the chainstore are _synchornous_.

chain/store/splitstore/splitstore.go
vyzo: again no race here, the chainstore notifications are synchronous and we can have only a single compaction running.

chain/store/splitstore/splitstore.go
vyzo: I don't understand this comment :)

chain/store/splitstore/splitstore.go
vyzo: meh, it's an error -- size does not have a useful value.

chain/store/splitstore/snoop.go
vyzo: Not really, I defined the methods we need with their standard names.
I don't see what this new interface you propose adds.

chain/store/splitstore/splitstore.go
vyzo: Doesn't look thaaat critical, it's just less performant than it could be.

chain/store/splitstore/splitstore.go
vyzo: if you refer to the (1) data race as the assignment and the (single) read in put, then it's entirely inconsequential.
We also have the state mutex we can use if you fear a race detector detecing nonsense, but still.

chain/store/splitstore/snoop.go
raulk: Please add it to the TODO then

chain/store/splitstore/splitstore.go
raulk: It's O(n^2) and therefore highly sensitive to the size of the reacheable nodes. Not an algorithm we want in production, especially when we're trying to curtail allows.

chain/store/splitstore/splitstore.go
raulk: How is this a compaction and not a transfer/migration?

raulk: Compaction has specific connotations in datastores, and this is certainly not a compaction. It's not about personal preference, it's about accuracy.

chain/store/splitstore/splitstore.go
raulk: That's an assumption. And downstream code expects this to be -1 on error. That's why I suggest you just proxy the value instead of relying on assumptions (it's what I tried to explain above).

chain/store/splitstore/snoop.go
raulk: It adds simplicity and doesn't trick the reader into thinking that this is a blockstore.

chain/store/splitstore/splitstore.go
raulk: We cannot stop the world if tracking a CID fails. When that happens, we should disable GC entirely, and keep everything in the hot store. It's a safe fallback instead of a hard failure.

chain/store/splitstore/splitstore.go
raulk: Keeping an iterator open when you know you will not be making progress is not good.

chain/store/splitstore/splitstore.go
raulk: That makes unsafe assumptions about the rest of the system. And those assumptions can change any time, and it will not be noticed because there is not contract/interface breakage. Thus it will lead to a bug. Highly encourage you to synchronise here, by whichever means you want. An atomic is faster.

chain/store/splitstore/splitstore.go
raulk: Read above.

chain/store/splitstore/splitstore.go
vyzo: I just wanted to avoid writing the same code twice, but fine.

chain/store/splitstore/splitstore.go
vyzo: Writing to the hotstore then is a mistake -- it will lead to catastrophic failure further down the line or result in a leak because of an untracked CID.
If we are to write somewhere, that should be the coldstore.

chain/store/splitstore/splitstore.go
vyzo: It's not O(n^2), just not as efficient as it should be.
Optimizing it is out of scope for this PR I think, we can do it separately.

chain/store/splitstore/snoop.go
vyzo: ok. It needs to happen in the blockstore implementation as well, where we map a gigantic 256G file.

chain/store/splitstore/splitstore.go
vyzo: Well, I am not going to argue names. It's not a simple migration however, it _does_ compact the datastore because it deletes unreachable nodes.

chain/store/splitstore/splitstore.go
vyzo: No, downstream code expects _nothing_ on error, it's an undefined value.

chain/store/splitstore/splitstore.go
vyzo: It's a pointer read/write -- this things don't go in half, but fine.

chain/store/splitstore/splitstore.go
vyzo: what are you talking about, the compacting thing or the base epoch?

vyzo: the compacting thing is mutex protected.

vyzo: so it is sufficient to move the isCompacting check first, because we can't be changing the baseEpoch if we are not compacting.

chain/store/splitstore/splitstore.go
vyzo: I kind of like snoop -- it grew on me.

chain/store/splitstore/splitstore.go
vyzo: I set it to log the error and fallback to the coldstore.

chain/store/splitstore/splitstore.go
vyzo: done, I used the mutex.

chain/store/splitstore/splitstore.go
vyzo: done.

chain/store/splitstore/splitstore.go
vyzo: there is a semantic problem here, we can't report the errors from getting the iterators if we move them to the goroutine.
I am inclined to leave it as is, I don't think it's a big deal if we hold an read iterator a little longer than necessary.

chain/store/splitstore/snoop.go
vyzo: I added a TODO note.

chain/store/splitstore/splitstore.go
raulk: I updated the description of this PR. It wouldn't hurt to explain what's being contributed here -- there are a lot more people watching the repo than people that were privy to this design discussion.

chain/store/splitstore/splitstore.go
raulk: Ok, we can revisit later. A rename would be cheap.

chain/store/splitstore/splitstore.go
raulk: Yeah, that's a good point. This interface has given us a lot more pain than gain.

raulk: The problem is that it's not us who determines the speed of consumption, it's the caller. So if the caller slows down consumption during the processing of the hot set, the cold set iteration would've stayed open for too long unnecessarily. Depending on underlying database, the overhead will vary.

vyzo: Agreed, but otoh I don't think we are actually using this interface.
We can revisit if this becomes a problem.

chain/store/splitstore/splitstore.go
```diff
@@ -28,8 +28,10 @@ var baseEpochKey = dstore.NewKey("baseEpoch")
 var log = logging.Logger("splitstore")
 
 type SplitStore struct {
-	baseEpoch abi.ChainEpoch
-	curTs     *types.TipSet
+	mx         sync.Mutex
```
raulk, line 31: Let's switch this to atomic, it's close to 3x faster!

```
goos: darwin
goarch: amd64
pkg: github.com/testground/sdk-go
BenchmarkMutex
BenchmarkMutex-16          	59274573	        19.3 ns/op
BenchmarkAtomicInt64
BenchmarkAtomicInt64-16    	159052496	         7.66 ns/op
PASS
```

```go
package main

import (
	"sync"
	"sync/atomic"
	"testing"
)

func BenchmarkMutex(b *testing.B) {
	var mu sync.Mutex
	var v int64

	for i := 0; i < b.N; i++ {
		mu.Lock()
		v = int64(i)
		mu.Unlock()

		mu.Lock()
		_ = v
		mu.Unlock()
	}
}

func BenchmarkAtomicInt64(b *testing.B) {
	var v int64 // guarded by atomic

	for i := 0; i < b.N; i++ {
		atomic.StoreInt64(&v, int64(i))
		_ = atomic.LoadInt64(&v)
	}
}
```

vyzo: it doesn't really matter, this is a micro-optimization.

vyzo: also, we are accessing more complex things than integers, and things become a bit ugly with atomics.

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,422 @@
...
+}
+
+func (s *SplitStore) HeadChange(revert, apply []*types.TipSet) error {
+	s.mx.Lock()
+	s.curTs = apply[len(apply)-1]
+	epoch := s.curTs.Height()
+	s.mx.Unlock()
+
+	if !s.isCompacting() && epoch-s.baseEpoch > CompactionThreshold {
+		s.setCompacting(true)
```
raulk, line 224: Can we please make `compacting` CAS? If there's a change upstream by which two HeadChanges arrive concurrently, we may launch two compactions.

Either that, or we guard the entire method in the mutex.

chain/store/splitstore/splitstore.go
raulk: These are low-level operations, and a 3x speed up is worth, especially when the effort is close to none.

raulk: That said, the position of this mutex threw me off. It is actually guarding `curTs` and `compacting`, not `baseEpoch`. If we were to use `atomic.Value` here, the speed-up is less noticeable.

```
goos: darwin
goarch: amd64
pkg: github.com/testground/sdk-go
BenchmarkMutex
BenchmarkMutex-16          	58329135	        19.1 ns/op
BenchmarkAtomicValue
BenchmarkAtomicValue-16    	98621289	        11.1 ns/op
PASS
```

If we need to modify these state variables atomically, then we are better off using locks. But if they're independent (like compacting is), we can use atomics.

chain/store/splitstore/splitstore.go
vyzo: I don't feel strongly about it, so we can make it CAS. It doesn't really matter though, it's 10 times per minute on average.

chain/store/splitstore/splitstore.go
vyzo: We still need the mutex for curTs, there is a little more complexity there. For compacting, I don't care particularly, it can be CAS.

vyzo: done.

ribasushi:
@vyzo @raulk from a conversation elsewhere I have the following requirement for this:

> I'd want to undo the split in "capture" mode - there must be a way for both cold and hot must write to the same place without an itermediary/delay
> this is pretty critical
> this doesn't necesarily need to be an exposed config option: for the time being the capture-store needs a few codechanges in lotus anyway, so being ale to replace couple lines in `fsrepo` is enough for me
> the ask is specifically for the interfaces not to diverge too far from the basis `ipfs-blockstore` one

vyzo:
we can simply add a config option that disables gc -- you will still use the splitstore, but persist all objects without deleting unreachable ones.

vyzo:
RIPed the freecache stuff and rebased directly to master; will fix conflicts.

ribasushi:
> we can simply add a config option that disables gc -- you will still use the splitstore, but persist all objects without deleting unreachable ones.

For the specific case I need this for this is not relevant: *both* underlying blockstores need to be exchangeable with a single 3rd party one, in order to allow for synchronous-ish storage. This is why I focus on "keep the interface close enough", as "no code changes" and "use 2 stores" is not an option in my case either way.

vyzo:
Not sure I follow -- what do you want to exchange? You can simply use the coldstore.

vyzo:
Also the blockstore interface is not changing.

chain/store/splitstore/splitstore.go
```diff
@@ -62,10 +62,10 @@ func NewSplitStore(path string, ds dstore.Datastore, cold bstore.Blockstore) (*S
...
 	hot, err := lmdbbs.Open(&lmdbbs.Options{
 		Path:                 path,
-		InitialMmapSize:      256 << 20, // 256MiB.
-		MmapGrowthStepFactor: 1.25,      // scale slower than the default of 1.5
-		MmapGrowthStepMax:    512 << 20, // 512MiB.
-		MaxReaders:           32,
+		InitialMmapSize:      1 << 30, // 1GiB.
+		MmapGrowthStepFactor: 1.25,    // scale slower than the default of 1.5
+		MmapGrowthStepMax:    1 << 32, // 4GiB
+		MaxReaders:           192,
```
raulk, line 68: I'd just leave it unset; it will use the default of 254 (yes, not 256, cache lines).

vyzo: ok, sounds good; I think we should also start with a 4G map as well instead of just 1.

vyzo: done.

vyzo:
rebased on https://github.com/filecoin-project/lotus/pull/5695

vyzo:
@magik6k @Kubuxu @whyrusleeping this is at last ready for review!

raulk (commented):
Note: I'll be following up shortly to pull out the lmdb support out of this PR into a branch, in order to reduce LoC in this patch and keep it out of the dependency graph. LMDB is not used currently -- but folks can enable it as an experiment.

chain/store/splitstore/liveset_bloom.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (e *BloomLiveSetEnv) NewLiveSet(name string, sizeHint int64) (LiveSet, error) {
+	size := int64(BloomFilterMinSize)
+	for size < sizeHint {
+		size += BloomFilterMinSize
+	}
+
+	salt := make([]byte, 4)
+	_, err := rand.Read(salt) //nolint
```
raulk: We need to use `crypto/rand` here, or seed `math/rand`. Otherwise random values will repeat upon restart.

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,830 @@
...
+	}
+
+	if epoch-s.baseEpoch > CompactionThreshold {
+		go func() {
+			defer atomic.StoreInt32(&s.compacting, 0)
+
+			log.Info("compacting splitstore")
+			start := time.Now()
+
+			s.compact()
```
raulk: We should pass the current epoch here to avoid races. It could happen that by the time the compaction runs and re-reads the current epoch, it has slid.

chain/store/splitstore/liveset_bloom.go
vyzo: that's totally not a problem :)

chain/store/splitstore/splitstore.go
vyzo: ok, we can do that -- we should probably pass the tipset itself.

vyzo: done.

chain/store/splitstore/liveset_bloom.go
vyzo: but did it anyway.

magik6k (commented):
First pass, looks pretty good.

* Is there any particular reason boltdb was chosen for the tracking store?
* Do we have compaction logs from mainnet nodes? How do compaction timings look?
* How are block validation times affected? How visible is the overhead of the tracking store on puts?

raulk:
> Is there any particular reason boltdb was chosen for the tracking store?

We originally started with LMDB, but it has way too many warts that can become problems in production (e.g. mmap sizing, max readers), so we decided to ditch it. Since this dataset is going to be small, we chose bbolt for the tracking store.


We can connect the tracking store with the bloom filter, but the main right now is that the tracking store needs to survive restarts.

I think there was some limitation with some parts needing to delete items from the tracking store too, but I think we can probably find reasonable tradeoffs for those too.

> Do we have compaction logs from mainnet nodes? How do compaction timings look?

I think around 10min, but @vyzo can follow up. We need to figure out if the increased IO load that iteration+deletion places on the hotstore can impact chain sync.

> How are block validation times affected? How visible is the overhead of the tracking store on puts?

I don't want to jinx it, but tipset validation times are sub-second, and even as low as 100ms in the test machines we're using. It's too good to be true, so we're running more instances to make sure we haven't missed anything.

arajasek:
Is it safe to say this effectively enables "auto-pruning"? See https://github.com/filecoin-project/lotus/issues/4701 for more.

chain/store/splitstore/liveset_bloom.go
```diff
@@ -0,0 +1,77 @@
...
+func (e *BloomLiveSetEnv) Close() error {
+	return nil
+}
+
+func (s *BloomLiveSet) saltedKey(cid cid.Cid) []byte {
+	hash := cid.Hash()
+	key := make([]byte, len(s.salt)+len(hash))
+	n := copy(key, s.salt)
+	copy(key[n:], hash)
+	rehash := blake2b.Sum256(key)
```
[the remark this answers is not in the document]
vyzo: sure, can do.

chain/store/splitstore/liveset_test.go
```diff
@@ -0,0 +1,140 @@
...
+func TestBloomLiveSet(t *testing.T) {
+	testLiveSet(t, "bloom")
+}
+
+func testLiveSet(t *testing.T, lsType string) {
+	t.Helper()
+
+	path := "/tmp/liveset-test"
+
+	err := os.MkdirAll(path, 0777)
```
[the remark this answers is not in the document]
vyzo: sure!

chain/store/splitstore/splitstore.go
```diff
@@ -0,0 +1,830 @@
...
+	CompactionCold      = build.Finality
+)
+
+var baseEpochKey = dstore.NewKey("baseEpoch")
+
+var log = logging.Logger("splitstore")
+
+func init() {
+	// TODO temporary for debugging purposes; to be removed for merge.
+	logging.SetLogLevel("splitstore", "DEBUG")
```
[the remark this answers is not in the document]
vyzo: yeah, will remove this when it's time to merge.

blockstore/splitstore/splitstore.go
```diff
@@ -0,0 +1,831 @@
...
+	s.liveSetSize = 0
+	err := s.cs.WalkSnapshot(context.Background(), curTs, 1, s.skipOldMsgs, s.skipMsgReceipts,
+		func(cid cid.Cid) error {
+			s.liveSetSize++
+			return nil
+		})
+
+	if err != nil {
+		// TODO do something better here
+		panic(err)
```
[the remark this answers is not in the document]
vyzo: right, but it's not clear yet what do if we get thoz errors.

blockstore/splitstore/splitstore.go
```diff
@@ -0,0 +1,831 @@
...
+			return nil
+		}
+
+		// check whether it is reachable in the cold boundary
+		mark, err := coldSet.Has(cid)
+		if err != nil {
+			return xerrors.Errorf("error checkiing cold set for %s: %w", cid, err)
+		}
+
+		if mark {
```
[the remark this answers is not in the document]
vyzo: that's totally fine, we are salting the bloom filter with a random salt on every iteration so that we don't get the same false positives in every compaction.
Given that the false positives change, the garbage will get purged over time.

blockstore/splitstore/splitstore.go
```diff
@@ -0,0 +1,831 @@
...
+		}
+
+		// check whether it is reachable in the cold range
+		mark, err = coldSet.Has(cid)
+		if err != nil {
+			return xerrors.Errorf("error checkiing cold set for %s: %w", cid, err)
+		}
+
+		if s.enableGC {
+			if mark {
```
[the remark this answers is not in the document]
vyzo: see above.

blockstore/splitstore/splitstore.go
```diff
@@ -0,0 +1,831 @@
...
+
+	if err != nil {
+		// TODO do something better here
+		panic(err)
+	}
+
+	log.Infow("compaction stats", "hot", stHot, "cold", stCold, "dead", stDead)
+
+	// 2.2 copy the cold objects to the coldstore
+	log.Info("moving cold objects to the coldstore")
```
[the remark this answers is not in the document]
vyzo: sure, can do.

blockstore/splitstore/tracking.go
```diff
@@ -0,0 +1,30 @@
...
+	DeleteBatch(map[cid.Cid]struct{}) error
+	ForEach(func(cid.Cid, abi.ChainEpoch) error) error
+	Sync() error
+	Close() error
+}
+
+func NewTrackingStore(path string, trackingStoreType string) (TrackingStore, error) {
+	switch trackingStoreType {
+	case "", "bolt":
+		return NewBoltTrackingStore(filepath.Join(path, "snoop.bolt"))
```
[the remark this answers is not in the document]
vyzo: it's actually performing pretty decently for the workload; the dataset is small and we care about read performance quite a bit, and that's pretty good.

node/builder.go
```diff
@@ -586,14 +586,45 @@ func Repo(r repo.Repo) Option {
...
 
+		var cfg *config.Blockstore
+		switch settings.nodeType {
+		case repo.FullNode:
+			cfgp, ok := c.(*config.FullNode)
+			if !ok {
+				return xerrors.Errorf("invalid config from repo, got: %T", c)
+			}
+			cfg = &cfgp.Blockstore
+		case repo.StorageMiner:
```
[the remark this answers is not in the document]
vyzo: uhm ok, so what should we do here? Just create a dummy config object?

node/builder.go
```diff
@@ -586,14 +586,45 @@ func Repo(r repo.Repo) Option {
...
+				Override(new(dtypes.StateBlockstore), modules.StateSplitBlockstore),
+				Override(new(dtypes.BaseBlockstore), From(new(dtypes.SplitBlockstore))),
+				Override(new(dtypes.ExposedBlockstore), From(new(dtypes.SplitBlockstore))),
+			),
+			If(!cfg.EnableSplitstore,
+				Override(new(dtypes.ChainBlockstore), modules.ChainFlatBlockstore),
+				Override(new(dtypes.StateBlockstore), modules.StateFlatBlockstore),
+				Override(new(dtypes.BaseBlockstore), From(new(dtypes.UniversalBlockstore))),
+				Override(new(dtypes.ExposedBlockstore), From(new(dtypes.UniversalBlockstore))),
+			),
```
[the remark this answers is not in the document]
vyzo: we'd still need it because we need to pass the splitstore config to the constructor.

node/modules/blockstore.go
```diff
@@ -31,17 +36,77 @@ func UniversalBlockstore(lc fx.Lifecycle, mctx helpers.MetricsCtx, r repo.Locked
...
+
+func ChainFlatBlockstore(lc fx.Lifecycle, mctx helpers.MetricsCtx, bs dtypes.UniversalBlockstore) (dtypes.ChainBlockstore, error) {
 	return bs, nil
 }
 
-// ChainBlockstore is a hook to overlay caches for state objects, or in the
-// future, to segregate the universal blockstore into different physical state
-// and chain stores.
-func ChainBlockstore(lc fx.Lifecycle, mctx helpers.MetricsCtx, bs dtypes.UniversalBlockstore) (dtypes.ChainBlockstore, error) {
+func ChainSplitBlockstore(lc fx.Lifecycle, mctx helpers.MetricsCtx, bs dtypes.SplitBlockstore) (dtypes.ChainBlockstore, error) {
```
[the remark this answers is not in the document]
vyzo: Yes, but we did this to have the hooks ready to introduce retention polices and the like.
We can certainly change it if you feel strongly about it.

node/modules/blockstore.go
```diff
@@ -31,17 +36,77 @@ func UniversalBlockstore(lc fx.Lifecycle, mctx helpers.MetricsCtx, r repo.Locked
...
+
+	opts, err := repo.BadgerBlockstoreOptions(repo.HotBlockstore, path, r.Readonly())
+	if err != nil {
+		return nil, err
+	}
+
+	bs, err := badgerbs.Open(opts)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
vyzo: we can't, that's the coldstore.

node/modules/blockstore.go
```diff
@@ -31,17 +36,77 @@ func UniversalBlockstore(lc fx.Lifecycle, mctx helpers.MetricsCtx, r repo.Locked
...
-// future, to segregate the universal blockstore into different physical state
-// and chain stores.
-func StateBlockstore(lc fx.Lifecycle, mctx helpers.MetricsCtx, bs dtypes.UniversalBlockstore) (dtypes.StateBlockstore, error) {
+func BadgerHotBlockstore(lc fx.Lifecycle, r repo.LockedRepo) (dtypes.HotBlockstore, error) {
+	path, err := r.SplitstorePath()
+	if err != nil {
+		return nil, err
+	}
+
+	path = filepath.Join(path, "hot.badger")
```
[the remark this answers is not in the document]
vyzo: I want to keep the hotstore type in the name so that we don't get ridiculous bugs with distinguishing it.
We currently only use badger, but the experimental lmdb branch can change this.

vyzo:
Re: compaction timings

Here is a log from a mainnet test node:
```
2021-03-01T14:51:00.272Z        INFO    splitstore      splitstore/splitstore.go:316    compacting splitstore
2021-03-01T14:51:00.272Z        INFO    splitstore      splitstore/splitstore.go:339    current live set size estimate  {"size": 18865198}
2021-03-01T14:51:00.272Z        INFO    splitstore      splitstore/splitstore.go:374    running simple compaction       {"currentEpoch": "543462", "baseEpoch": "538961", "coldEpoch": "539861"}
2021-03-01T14:51:00.274Z        INFO    splitstore      splitstore/splitstore.go:384    marking reachable cold objects
2021-03-01T15:04:41.641Z        INFO    splitstore      splitstore/splitstore.go:409    marking done    {"took": 821.366804644}
2021-03-01T15:04:41.641Z        INFO    splitstore      splitstore/splitstore.go:412    collecting cold objects
2021-03-01T15:05:57.552Z        INFO    splitstore      splitstore/splitstore.go:451    collection done {"took": 75.910615065}
2021-03-01T15:05:57.552Z        INFO    splitstore      splitstore/splitstore.go:452    compaction stats        {"hot": 28134483, "cold": 7124536}
2021-03-01T15:05:57.553Z        INFO    splitstore      splitstore/splitstore.go:455    moving cold objects to the coldstore
2021-03-01T15:11:46.767Z        INFO    splitstore      splitstore/splitstore.go:502    moving done     {"took": 349.21372698}
2021-03-01T15:11:46.767Z        INFO    splitstore      splitstore/splitstore.go:506    purging cold objects from the hotstore
2021-03-01T15:40:11.060Z        INFO    splitstore      splitstore/splitstore.go:517    purging cold from hotstore done {"took": 1704.293282302}
2021-03-01T15:40:11.060Z        INFO    splitstore      splitstore/splitstore.go:521    purging cold objects from snoop
2021-03-01T15:40:51.238Z        INFO    splitstore      splitstore/splitstore.go:529    purging cold from snoop done    {"took": 40.178132121}
2021-03-01T15:40:56.934Z        INFO    splitstore      splitstore/splitstore.go:321    compaction done {"took": 2996.662441999}
```

We currently have two *very* slow operations:
- `WalkSnapshot` is ridiculous, it takes 10min for a 1 epoch walk; we need to optimize that.
- Purging data from the hotstore takes quite a bit, because we have to delete the objects one by one and it's some 7M of them; we need to add a DeleteBatch pritmitive to our blockstores.

blockstore/splitstore/liveset.go
```diff
@@ -14,14 +14,14 @@ type LiveSet interface {
 	Close() error
 }
 
-var markBytes = []byte{}
+var markBytes []byte
```
vyzo: are you sure this is a good idea? I am concerned about the behaviour of the database backed ones if we pass nil.

raulk: I'll revert just in case.

vyzo: thanks!

blockstore/splitstore/tracking.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/filecoin-project/go-state-types/abi"
+	cid "github.com/ipfs/go-cid"
+)
+
+// TrackingStore is a persistent store that tracks blocks that are added
+// within the current compaction range, including the epoch at which they are
+// written.
+//
+// On every compaction, we iterate over
```
raulk, line 16: unfinished docs

node/modules/blockstore.go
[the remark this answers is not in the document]
raulk: Yes, we will likely change the hot store library at some point (go-storethehash, lmdb, go-nudb, etc.)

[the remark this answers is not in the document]
vyzo: Important detail is that it need to be able to delete, so storethehash and nudb are not viable.

chain/store/splitstore/liveset_bloom.go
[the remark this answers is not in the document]
vyzo: done.

chain/store/splitstore/liveset_test.go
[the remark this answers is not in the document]
vyzo: done.

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: done.

node/builder.go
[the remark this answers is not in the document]
vyzo: Or do you want to not have a blockstore config section at all and put the Splitstore config in common?

node/builder.go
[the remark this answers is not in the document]
vyzo: we can drop this part of the switch then and just fall to the default.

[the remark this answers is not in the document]
vyzo: done.

node/builder.go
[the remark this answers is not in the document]
vyzo: I actually would like to keep the blockstore config section since we are planning to build nodes that operate without a coldstore in the future.

But if you feel strongly about it, we can certainly put it in ConfigCommon.

node/config/def.go
```diff
@@ -12,9 +12,10 @@ import (
 
 // Common is common config between full node and miner
 type Common struct {
-	API    API
-	Libp2p Libp2p
-	Pubsub Pubsub
+	API        API
+	Libp2p     Libp2p
+	Pubsub     Pubsub
+	Blockstore Blockstore
```
magik6k, line 18: I'd put this in the FullNode struct, and call `Chainstore` as we have some other blockstores

node/config/def.go
```diff
@@ -114,6 +115,28 @@ type Pubsub struct {
...
+	Archival             bool
+}
+
+func (s *Splitstore) GetHotStoreType() string {
+	// default is badger
+	if s.HotStoreType == "" {
+		return "badger"
+	}
+	return s.HotStoreType
+}
```
magik6k, line 138: If you set values in the default func below, it will be also used in configs which don't set this explicitly (so this wouldn't be needed)

magik6k (commented):
Looks good, other than the panics.

Some config nits

node/config/def.go
vyzo: ok, can do.

node/config/def.go
vyzo: ah good point! thanks, will do.

node/config/def.go
vyzo: done.

node/config/def.go
vyzo: done.

vyzo:
rebased on `next` and implemented DeleteMany on Union blockstore.

vyzo:
production timings from most recent compaction, with batch delete:
```
2021-03-04T05:21:00.244Z        INFO    splitstore      splitstore/splitstore.go:416    compacting splitstore
2021-03-04T05:21:00.244Z        INFO    splitstore      splitstore/splitstore.go:511    estimating mark set size
2021-03-04T05:30:08.456Z        INFO    splitstore      splitstore/splitstore.go:513    estimating mark set size done   {"took": 548.212030043, "size": 9193373}
2021-03-04T05:30:08.457Z        INFO    splitstore      splitstore/splitstore.go:546    running simple compaction       {"currentEpoch": "550962", "baseEpoch": "546461", "coldEpoch": "547361", "boundaryEpoch": "549162"}
2021-03-04T05:30:08.460Z        INFO    splitstore      splitstore/splitstore.go:556    marking reachable cold objects  {"boundaryEpoch": "549162"}
2021-03-04T05:44:08.961Z        INFO    splitstore      splitstore/splitstore.go:581    marking done    {"took": 840.501188347}
2021-03-04T05:44:08.963Z        INFO    splitstore      splitstore/splitstore.go:584    collecting cold objects
2021-03-04T05:46:27.129Z        INFO    splitstore      splitstore/splitstore.go:627    collection done {"took": 138.166176124}
2021-03-04T05:46:27.129Z        INFO    splitstore      splitstore/splitstore.go:628    compaction stats        {"hot": 51874435, "cold": 6769339}
2021-03-04T05:46:27.129Z        INFO    splitstore      splitstore/splitstore.go:631    moving cold objects to the coldstore
2021-03-04T05:50:44.469Z        INFO    splitstore      splitstore/splitstore.go:638    moving done     {"took": 257.340147821}
2021-03-04T05:50:44.470Z        INFO    splitstore      splitstore/splitstore.go:641    purging cold objects from the hotstore
2021-03-04T05:51:08.446Z        INFO    splitstore      splitstore/splitstore.go:648    purging cold from hotstore done {"took": 23.97550071}
2021-03-04T05:51:08.455Z        INFO    splitstore      splitstore/splitstore.go:652    purging cold objects from tracker
2021-03-04T05:51:31.242Z        INFO    splitstore      splitstore/splitstore.go:658    purging cold from tracker done  {"took": 22.786124416}
2021-03-04T05:51:35.510Z        INFO    splitstore      splitstore/splitstore.go:421    compaction done {"took": 1835.265712425}
```

Notes:
- BatchDelete works wonders; 23s to delete from badger and 22s to delete form bolt for 6.7M objects.
- mark set size estimation, which only happens at first compaction since restart, takes about 9min -- this is purely WalkSnapshot
- marking itself takes 14min -- again, WalkSnapshot is the culprit.
- collecting cold objects, which is iteration over the entire tracking store, takes 138s; that's iterating over about 60M objects.
- moving cold objects to the coldstore (copying) takes 257s for 6.7M objects

So the limiting factor is now clearly WalkSnapshot; the criticial section (which is the deletion) is about 45s long.

raulk:
We can get rid of the mark set size estimation step entirely by putting on a simple smart. Two possible ways:

* when we do the marking, we track the number of items we visit and we memorise it. When we mark next, we use that number * a estimated growth factor to size the bloom. On the first run, we use a fixed size which over time will likely underestimate, and maybe end up with a few more false positives, but that’s a fair tradeoff given where we are now.
* we chain rolling bloom filters and we spill over as we mark, creating them lazily. When testing, we check against all blooms in sequence.

vyzo:
Actually we can do the estimate during warm up and save it into the metadata DS.

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: done -- just returning an error to the caller and loudly logging.

chain/store/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: done.

magik6k (approved):
Looks good, just needs fixing the merge conflict (miner/fullnode views are now separate, splitstore should go the the fullnode)

vyzo:
rebased on next and fixed conflict.

magik6k (approved):

magik6k (approved):
(test lgtm too)

arajasek (approved):
Broadly makes sense, and running well on my node for a few days

PR #5022, opened 2020-11-26
Updates for new blst group check API

simonatsn:
Update blst to v0.3.1
Update blst wrappers for the new subgroup check API changes

This change should be coordinated with an update of rust-fil-proofs to avoid using differing versions of blst between the two packages.

Do infinite signatures need to be permitted up to a certain block number? See comment in the code.


dignifiedquire:
Note: this must only be merged when updating to the new version of proofs: https://github.com/filecoin-project/filecoin-ffi/pull/155, as otherwise there will be a missmatch of blst versions when linking.

magik6k:
We're now using blst through filecoin-ffi

PR #5029, opened 2020-11-27
fix: start a new line in print miner-info to avoid ambiguous display

GFZRZK:
image
This is miner info now, it really makes us confused.


magik6k (approved):
Thanks!

PR #5037, opened 2020-11-27
Fix chainstore tipset leak

Kubuxu:
It is a bit of a unortodox fix but it works.
See the comment in chainstore.


Kubuxu:
I have a node running with this, `len(cs.tipsets)` right now stabilized at ~1100.

arajasek (approved):
Maybe only trigger the deletion when `cs.tipsets` grows above a certain size?

magik6k (approved):

PR #5043, opened 2020-11-28
Feat shed balances

zgfzgf:
use ActorNameByCode func
add miner-info flag

### DIFF
--- a/cmd/lotus-shed/balances.go
+++ b/cmd/lotus-shed/balances.go
@@ -109,7 +109,7 @@ var chainBalanceCmd = &cli.Command{
 			ai := accountInfo{
 				Address: addr,
 				Balance: types.FIL(act.Balance),
-				Type:    string(act.Code.Hash()[2:]),
+				Type:    builtin.ActorNameByCode(act.Code),
 			}
 
 			if builtin.IsStorageMinerActor(act.Code) {
--- a/cmd/lotus-shed/balances.go
+++ b/cmd/lotus-shed/balances.go
@@ -113,26 +113,28 @@ var chainBalanceCmd = &cli.Command{
 				Balance: types.FIL(act.Balance),
 				Type:    builtin.ActorNameByCode(act.Code),
 			}
+			if minerInfo {
+				if builtin.IsStorageMinerActor(act.Code) {
+					pow, err := api.StateMinerPower(ctx, addr, tsk)
+					if err != nil {
+						return xerrors.Errorf("failed to get power: %w", err)
+					}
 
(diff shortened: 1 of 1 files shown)


cmd/lotus-shed/balances.go
```diff
@@ -125,6 +127,8 @@ var chainBalanceCmd = &cli.Command{
 				}
 				ai.Worker = info.Worker
 				ai.Owner = info.Owner
+				fmt.Printf("miner:%s, owner:%s, worker:%s, power:%s(%s)\n", ai.Address, ai.Owner,
```
magik6k, line 130: This command produces a CSV which is used in a number of places. This print breaks that output

cmd/lotus-shed/balances.go
```diff
@@ -32,7 +22,13 @@ import (
 	"github.com/filecoin-project/go-state-types/big"
 
 	"github.com/filecoin-project/lotus/chain/actors/adt"
+	"github.com/filecoin-project/lotus/chain/actors/builtin"
+	ainit "github.com/filecoin-project/lotus/chain/actors/builtin/init"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
+	"github.com/filecoin-project/lotus/chain/actors/builtin/multisig"
+	"github.com/filecoin-project/lotus/chain/actors/builtin/power"
+	"github.com/filecoin-project/lotus/chain/actors/builtin/reward"
+	"github.com/filecoin-project/lotus/chain/gen/genesis"
```
magik6k, line 31: Please don't commit changes to import order. Even if it's cleaner, it's annoying to review

magik6k (changes requested):

cmd/lotus-shed/balances.go
zgfzgf: add miner-info flag, default:It won't print

cmd/lotus-shed/balances.go
zgfzgf: OK

ZenGround0:
Hey @zgfzgf I'm going to close this down.  The main reason being that `./lotus-shed audits stateroot-balances --miner-info` does what it looks like you want already.

While I agree that `builtin.ActorNameByCode` is nicer than `string(act.Code.Hash()[2:])` they do have the same behavior so I'm leaving as is for now.  If you are motivated feel free to open a new PR with only the substitution of `builtin.ActorNameByCode`, ideally everywhere `string(act.Code.Hash()[2:])` is used in the codebase so we can clean these up all at once.

Issue #3263, opened 2020-08-24
Explore speedup of initial chain-sync by reducing datastore safety


ribasushi:
Filing this as an issue for further discussion. It was determined yesterday that disabling sync operations in both the Badger and Level datastores results in substantial speeding up of the initial chainsync ( see diff at the end ).

It is not clear whether running a lotus node without `fsync()` during normal operations is a bad idea. 3 conflicting solutions were proposed in a thread elsewhere:

- @magik6k: we don't keep any actually important data in badger datastores, so maybe it is safe to just disable
- @lanzafame / @placer14 / @magik6k: Maybe provide a flag which enables this "fast-sync" option to let those who don't care about resilience
- @ribasushi: have a separate instance of Badger/Level used just for initial sync/catchup and then switch back to "safe deaults"

Seeking thoughts from @Stebalien and @whyrusleeping who have ran into this previously in go-ipfs

```
diff --git a/node/repo/fsrepo_ds.go b/node/repo/fsrepo_ds.go
index e62fe17e..3a09bd0d 100644
--- a/node/repo/fsrepo_ds.go
+++ b/node/repo/fsrepo_ds.go
@@ -31,2 +31,3 @@ func badgerDs(path string) (datastore.Batching, error) {
        opts.Options = dgbadger.DefaultOptions("").WithTruncate(true).
+               WithSyncWrites(false).
                WithValueThreshold(1 << 10)
@@ -39,3 +40,3 @@ func levelDs(path string) (datastore.Batching, error) {
                Compression: ldbopts.NoCompression,
-               NoSync:      false,
+               NoSync:      true,
                Strict:      ldbopts.StrictAll,
```

placer14:
Recommend that this "fast sync" mode is sandboxed with some atomic validation/commit at the end. Want to avoid ruining your partially-synced state because your fast-sync trashed your repo and now you have to do it ALL again. This concern is even more painful once the network gets much older.

Stebalien:
I'd disable sync but make sure to call `Sync(prefix)` explicitly where required (e.g., after finishing chain sync).

heaventear:
maybe with a simple way to call linux sync command after initial chain sync.

![image](https://user-images.githubusercontent.com/2443840/91520631-560ede00-e928-11ea-8d74-44e62c50c97b.png)

Stebalien:
[...]

PR #5050, opened 2020-11-29
Bite the bullet and hard-disable sync for chainstore

ribasushi:
Given the current "state-of-the-chain" requiring folk to keep their chainstore safe is rather futile. Let's just disable sync for now.

Latest exhibit of this: https://github.com/filecoin-project/lotus/issues/3263#issue-684587473
Original: https://github.com/filecoin-project/lotus/issues/3263#issue-684587473

### DIFF
--- a/node/repo/blockstore_opts.go
+++ b/node/repo/blockstore_opts.go
@@ -11,6 +11,12 @@ func BadgerBlockstoreOptions(domain BlockstoreDomain, path string, readonly bool
 
 	opts := badgerbs.DefaultOptions(path)
 
+	// FIXME - to be revisited after tiered store work
+	// By default badger is exceptionally fsync()-happy
+	// It is less safe to run this way, but chain-loss is hardly a catastrophic event
+	// at present
+	opts.SyncWrites = false
+
 	// Due to legacy usage of blockstore.Blockstore, over a datastore, all
 	// blocks are prefixed with this namespace. In the future, this can go away,
 	// in order to shorten keys, but it'll require a migration.


raulk (changes requested):
We haven't found writes to be slow. This trades off reliability for speed. I understand the argument that the store is not 100% as reliable as is, but we certainly don't want lower reliability, which would increase the probability of corruption even further. We will likely be merging other badger config changes this week, along with the memory watchdog that I ended up drafting this weekend.

ribasushi:
> We haven't found writes to be slow.

This does not match experiences in the field. Inability to run on anything other than nvme constitutes 98% of the complaints in fil-slack, for the past 2+ months.  **Our own chain-infra** has been running with this patch manually applied since I wrote the original proposal: there's no other way to make a large badger work in my experience.

In any case: not opposed to continue applying this by hand, just stressing that writes are **extraordinarely** slow.

raulk:
So initial chain sync from a snapshot import is write intensive because we trust the data and perform no validation (read workloads).

The initial sync from the network would be read-predominant, due to the validation and the LSM amplification. Writes from validation are batched, so each batch should only fsync once. Writes should go to a single file (non-amplified), and hit the disk again when flushing the memtable to L0.

Of course, compaction is a background process all throughout. It's read intensive with a final flush and fsync.

If we want to apply the the patch herein, we can, but the tradeoff is not worth it IMO. It's likely that users will still want to use nvme anyway.

The splitstore is close to being finished. That will help a lot because it will enable folks to run the hot store of a small NVMe drive, and place the large cold store in a magnetic drive.

ribasushi:
( top-diclaimer: I am fine not merging this, addressing misconceptions )

> Writes from validation are batched, so each batch should only fsync once.

This is absolutely not the case. As proof: instrument the blockstore's `Put` ( not `Putmany` ) to `log.Warn(...)`, and observe how often we call it during normal operation.

>  likely that users will still want to use nvme anyway ... will enable folks to run the hot store of a small NVMe 

If this is a temporary state of affairs during the proof-of-concept testing of the split blockstore - that's somewhat ok. If this is the answer long-term: it is not acceptable from a product standpoint. An nvme can not be a long-term prerequisite for running a lotus node ( both for end-users and mid-tier cloud-infra users )

raulk:
> > Writes from validation are batched, so each batch should only fsync once.
> 
> This is absolutely not the case. As proof: instrument the blockstore's `Put` ( not `Putmany` ) to `log.Warn(...)`, and observe how often we call it during normal operation.
> 

Do remember if you instrumented the badger datastore or the blockstore itself? Validation definitely uses PutMany to move new objects from the temp store into the durable store, but two things come to mind that could be behind this granular Put: (1) chain objects received via chainxchg and bitswap, the (2) metadata store writes (which go directly into the datastore, hence my question).

> >  likely that users will still want to use nvme anyway ... will enable folks to run the hot store of a small NVMe 
> 
> If this is a temporary state of affairs during the proof-of-concept testing of the split blockstore - that's somewhat ok. If this is the answer long-term: it is not acceptable from a product standpoint. An nvme can not be a long-term prerequisite for running a lotus node ( both for end-users and mid-tier cloud-infra users )

That's good to know. We need to do two things here:
1. Benchmark lmdb on magnetic drives, including the workload placed by the new hot=>cold archival process.
2. Group writes from chainxchg and bitswap.

ribasushi:
>>> Writes from validation are batched, so each batch should only fsync once.

>> This is absolutely not the case. As proof: instrument the blockstore's Put ( not Putmany ) to log.Warn(...), and observe how often we call it during normal operation.

> Do you remember if you instrumented the badger datastore or the blockstore itself? 

100% the blockstore itself. Also see https://filecoinproject.slack.com/archives/C0179RNEMU4/p1606608157012600?thread_ts=1606585952.494500&cid=C0179RNEMU4 - a steady, unbroken stream of small writes.

raulk:
Thanks, will check that thread in a bit. I think it would be desirable to put the blockstore in NoSync mode and expose a Sync() method that we call when we commit a new head.

With this patch only, we risk making the blockstore even less safe than it is now. But combining it with an explicit sync on head commit, which is when the rest of Lotus expects all objects written thus far to survive a reboot, we would hit the sweet spot.

raulk:
Let's continue the discussion in https://github.com/filecoin-project/lotus/issues/5057.

PR #5055, opened 2020-11-30
Feat shed add base64

zgfzgf:
add base64
now  remove LOTUS_STORAGE_PATH  in order to remove it in lotus-miner  later 
add -d


magik6k:
This isn't needed, basically all systems have `base64` util installed

PR #5060, opened 2020-11-30
client list-asks --by-ping

magik6k:


cli/client.go
```diff
@@ -1308,9 +1327,19 @@ loop:
 					return
 				}
 
+				rt := time.Now()
+
+				_, err = api.ClientQueryAsk(ctx, *mi.PeerId, miner)
+				if err != nil {
+					return
+				}
```
magik6k, line 1335: We could implement this using libp2p ping, but that requires exposing it on the API..

With this I'm getting 20-30ms pings to peers in the EU, so I'd say it's good enough for now

dirkmc (approved):
LGTM 👍

Issue #4990, opened 2020-11-24
When a sector is terminated while sealing on the miner side, deal remains in StorageDealSealing state for client.


dineshshenoy:
From @dirkmc 

It seems that when a sector is terminated while sealing, deal remains in StorageDealSealing state.
For example a deal that went from StorageDealSealing -> StorageDealError (from the client's perspective) when the client restarted:

Nov 20 19:06:23  bafyreichan55tqeegojqjm3lnhjhtr6tqkd4l3t557tryg4kfbry4lpaqa  1367548  f02665   StorageDealSealing                Y (epoch 0)  Y (epoch 0)  baga6ea4seaqgowjicpl453l6matswfg2ftzjoxytcszrztmeaeg5z3ztfxyecoq  7.938 GiB  0.002082448 FIL   520612  false
Nov 20 19:06:23  bafyreichan55tqeegojqjm3lnhjhtr6tqkd4l3t557tryg4kfbry4lpaqa  1367548  f02665   StorageDealError                  Y (epoch 0)  Y (epoch 0)  baga6ea4seaqgowjicpl453l6matswfg2ftzjoxytcszrztmeaeg5z3ztfxyecoq  7.938 GiB  0.002082448 FIL   520612  false  error in deal activation: failed to set up called handler: called check error (h: 263798): failed to look up deal on chain: deal 1367548 not found

One idea that occurs to me is to periodically check if a deal has been terminated while it's in the StorageDealSealing state.

hannahhoward:
Follow-up @dirkmc : We should probably timeout OnDealSectorPreCommitted and OnDealSectorCommitted on the StartEpoch for the deal proposal.

hannahhoward:
^^ @magik6k do you think this will work? It seems to me if we miss the start epoch then the provecommit will fail either way?

hannahhoward:
alternatively now that we are seperating out pre-commit, is there a fixed timeout between pre-commit and prove-commit?

dirkmc:
Missing start epoch is one reason the deal will be terminated. Are there other reasons the deal might be terminated while it's in the AwaitingPreCommit / Sealing state?

dirkmc:
Fixing this will most likely solve https://github.com/filecoin-project/lotus/issues/4977

magik6k:
Yes, timeout on StartEpoch is the correct thing to do here, I believe this is when the deal will be removed from the market actor

Issue #4977, opened 2020-11-23
OnDealExpiredOrSlashed returns error "deal 12345 not found"


dirkmc:
Some miners are reporting errors like
```
applyWithConfidence is triggering events events/events_called.go:252    chain trigger (@H 251012, triggered @ 251018) failed: deal 12345 not found
```

When a deal becomes active, lotus watches until the deal expires or is slashed. A bug was introduced in the code that fetches the deal:
https://github.com/filecoin-project/lotus/blob/375b7a1c4ca52ef63446070bc03f63669750b169/chain/stmgr/utils.go#L302-L304

In fact if the deal is not found an empty deal state should be returned, as is the case further down in the same method:
https://github.com/filecoin-project/lotus/blob/375b7a1c4ca52ef63446070bc03f63669750b169/chain/stmgr/utils.go#L316-L318

dineshshenoy:
This is likely solved on the miner side but we're still seeing it on the client side.

dineshshenoy:
This could be because the sector ID for the deal  in error message didn't properly complete on the miner side (cc: @hannahhoward )

dirkmc:
We have reports that this is also occurring on the miner side in v1.2.0 & v1.2.1

kernelogic:
I have a TON of these messages on my client side.

PR #5061, opened 2020-11-30
Error out deals that are not activated by proposed deal start epoch

dirkmc:
Fixes https://github.com/filecoin-project/lotus/issues/4990

When a deal is terminated by the miner, the client can get stuck in `StorageDealSealing`.

This PR adds functionality to the functions that wait for pre-commit and prove-commit such that they will callback with an error if the chain reaches the proposed start epoch for the deal without the deal being activated.

Fixes https://github.com/filecoin-project/lotus/issues/4977

Provides a better error message for when the deal is not found.
Old: "deal 12345 not found"
New: "deal 12345 not found - deal may not have completed sealing before deal proposal start epoch, or deal may have been slashed"

### DIFF
--- a/markets/storageadapter/ondealsectorcommitted.go
+++ b/markets/storageadapter/ondealsectorcommitted.go
@@ -8,7 +8,6 @@ import (
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-fil-markets/storagemarket"
 	"github.com/filecoin-project/go-state-types/abi"
-	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/market"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
@@ -20,7 +19,6 @@ import (
 
 type sectorCommittedEventsAPI interface {
 	Called(check events.CheckFunc, msgHnd events.MsgHandler, rev events.RevertHandler, confidence int, timeout abi.ChainEpoch, mf events.MsgMatchFunc) error
-	ChainAt(hnd events.HeightHandler, rev events.RevertHandler, confidence int, h abi.ChainEpoch) error
 }
 
 func OnDealSectorPreCommitted(ctx context.Context, api getCurrentDealInfoAPI, eventsApi sectorCommittedEventsAPI, provider address.Address, dealID abi.DealID, proposal market.DealProposal, publishCid *cid.Cid, callback storagemarket.DealSectorPreCommittedCallback) error {
@@ -33,9 +31,8 @@ func OnDealSectorPreCommitted(ctx context.Context, api getCurrentDealInfoAPI, ev
 	}
 
 	// First check if the deal is already active, and if so, bail out
-	var proposedDealStartEpoch abi.ChainEpoch
 	checkFunc := func(ts *types.TipSet) (done bool, more bool, err error) {
-		deal, isActive, err := checkIfDealAlreadyActive(ctx, api, ts, dealID, proposal, publishCid)
+		isActive, err := checkIfDealAlreadyActive(ctx, api, ts, dealID, proposal, publishCid)
 		if err != nil {
 			// Note: the error returned from here will end up being returned
 			// from OnDealSectorPreCommitted so no need to call the callback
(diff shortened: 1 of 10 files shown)


markets/storageadapter/ondealsectorcommitted.go
```diff
@@ -37,6 +49,10 @@ func OnDealSectorPreCommitted(ctx context.Context, api getCurrentDealInfoAPI, ev
 			return true, false, nil
 		}
 
+		// Save the proposed deal start epoch so we can timeout if the deal
+		// hasn't been activated by that epoch
+		proposedDealStartEpoch = deal.Proposal.StartEpoch
```
hannahhoward, line 54: note: you don't need to do this in checkFun since deal proposal is already a parameter to this function

markets/storageadapter/ondealsectorcommitted.go
```diff
@@ -96,13 +112,28 @@ func OnDealSectorPreCommitted(ctx context.Context, api getCurrentDealInfoAPI, ev
 		return xerrors.Errorf("failed to set up called handler: %w", err)
 	}
 
+	// If the deal hasn't been activated by the proposed start epoch, timeout
+	// the deal
+	timeoutOnProposedStartEpoch(dealID, proposedDealStartEpoch, eventsApi, func(err error) {
```
hannahhoward, line 117: I am curious if this needs to be a seperate call, or, if you can simply use the timeout functionality of the eventsApi.Called API. Note the fifth function parameter says events.NoTimeout, but this can be a timeout. It may be a duration which would make it potentially not super workable, but getting the current epoch is fairly trival. As I understand it, if the eventsAPI times out, the `called` func will be called but the msg param will be nil -- you can see we actually have a check for that, but it probably is never met cause of the events.NoTimeout parameter. Anyway, this may be worth investigating, since it could simplify the code quite a bit.

hannahhoward (approved):
I think it's worth investingating the timeout functionality builtin to Events.Called, though, I guess I wouldn't hold up merge on it.

markets/storageadapter/ondealsectorcommitted.go
dirkmc: Ah yeah you're right, that's a better idea 👍

markets/storageadapter/ondealsectorcommitted.go
```diff
@@ -55,9 +68,11 @@ func OnDealSectorPreCommitted(ctx context.Context, api getCurrentDealInfoAPI, ev
 			}
 		}()
 
-		// Check if waiting for pre-commit timed out
+		// If the deal hasn't been activated by the proposed start epoch, the
+		// deal will timeout (when msg == nil it means the timeout epoch was reached)
 		if msg == nil {
-			return false, xerrors.Errorf("timed out waiting for deal %d pre-commit", dealID)
+			err := xerrors.Errorf("deal %d was not activated by proposed deal start epoch %d", dealID, proposal.StartEpoch)
```
magik6k, line 74: Shouldn't this be
```suggestion
			err = xerrors.Errorf("deal %d was not activated by proposed deal start epoch %d", dealID, proposal.StartEpoch)
```
for the defer thing to work?

markets/storageadapter/ondealsectorcommitted.go
```diff
@@ -92,14 +107,22 @@ func OnDealSectorPreCommitted(ctx context.Context, api getCurrentDealInfoAPI, ev
 		return nil
 	}
 
-	if err := eventsApi.Called(checkFunc, called, revert, int(build.MessageConfidence+1), events.NoTimeout, matchEvent); err != nil {
+	if err := eventsApi.Called(checkFunc, called, revert, int(build.MessageConfidence+1), timeoutEpoch, matchEvent); err != nil {
```
magik6k, line 110: Here too

magik6k (commented):
Just 1 thing

markets/storageadapter/ondealsectorcommitted.go
dirkmc: I think this one is ok because it's returned from the function directly, it's not passed back with a callback

magik6k (approved):

kernelogic:
Is it a better practice to specify a startepoch much later to allow more time to seal?

dirkmc:
It depends on the individual miner. If the miner is already overloaded and struggling to seal in time, setting the proposed deal start epoch to a later time may not help much.

PR #5085, opened 2020-12-01
Multisig API improvements

arajasek:
- Remove gas price param from MsigCreate
- Allow specifying vesting start epoch in MsigCreate


Stebalien:
@arajasek closing because this is a mess of conflicts and over 6 months old.

PR #5094, opened 2020-12-02
Allow miners to filter (un)verified deals

arajasek:


magik6k (approved):
We really need a real dynamic config thing

PR #5095, opened 2020-12-02
Optimize sched, select workers according to available space and weigh…

liyue201:
Optimize sched, select workers according to available space and weight

### DIFF
--- a/extern/sector-storage/selector_alloc.go
+++ b/extern/sector-storage/selector_alloc.go
@@ -67,9 +67,8 @@ func (s *allocSelector) Ok(ctx context.Context, task sealtasks.TaskType, spt abi
 func (s *allocSelector) Cmp(ctx context.Context, task sealtasks.TaskType, spt abi.RegisteredSealProof, a, b *workerHandle) (bool, error) {
 	if s.ptype == storiface.PathSealing {
 		return a.utilization() < b.utilization(), nil
-	} else {
-		return s.storageCmp(ctx, spt, a, b)
 	}
+	return s.storageCmp(ctx, spt, a, b)
 }
 
 func (s *allocSelector) storageCmp(ctx context.Context, spt abi.RegisteredSealProof, a, b *workerHandle) (bool, error) {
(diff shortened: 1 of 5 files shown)


extern/sector-storage/selector_alloc.go
```diff
@@ -64,8 +64,50 @@ func (s *allocSelector) Ok(ctx context.Context, task sealtasks.TaskType, spt abi
 	return false, nil
 }
 
-func (s *allocSelector) Cmp(ctx context.Context, task sealtasks.TaskType, a, b *workerHandle) (bool, error) {
-	return a.utilization() < b.utilization(), nil
+func (s *allocSelector) Cmp(ctx context.Context, task sealtasks.TaskType, spt abi.RegisteredSealProof, a, b *workerHandle) (bool, error) {
+	if s.ptype == storiface.PathSealing {
```
magik6k, line 68: The only task type which isn't using `storiface.PathSealing` for the selector is FinalizeSector. I'm not sure if this will actually a noticeable difference

liyue201: When the sealing worker and storage worker are deployed separately, it is hoped that the data on the storage worker can be evenly distributed

extern/sector-storage/selector_alloc.go
```diff
@@ -64,8 +64,50 @@ func (s *allocSelector) Ok(ctx context.Context, task sealtasks.TaskType, spt abi
...
+	bHave := map[stores.ID]struct{}{}
+	for _, path := range bPaths {
+		bHave[path.ID] = struct{}{}
+	}
+
+	ssize, err := spt.SectorSize()
+	if err != nil {
+		return true, xerrors.Errorf("getting sector size: %w", err)
+	}
+	best, err := s.index.StorageBestAlloc(ctx, s.alloc, ssize, s.ptype)
```
magik6k, line 97: Calling `StorageBestAlloc` is too expensive to do in `Cmp`, at the very least we should cache that call

magik6k (commented):
From just reading the code it's not obvious to me in what setup this improves data/load distribution, and in what ways.

If you can provide some more detailed explanation, and more information on worker/storage configuration in which this is helping, we might merge this change

magik6k:
(closing for now since this PR got very old. Feel free to reopen with better explanation)

PR #5109, opened 2020-12-03
Let max-fee support setting decimals

wenzhenxiang:


cli/mpool.go
```diff
@@ -422,12 +422,12 @@ var mpoolReplaceCmd = &cli.Command{
 
 			var mss *lapi.MessageSendSpec
 			if cctx.IsSet("max-fee") {
-				maxFee, err := types.BigFromString(cctx.String("max-fee"))
+				maxFee, err := types.ParseFIL(cctx.String("max-fee"))
```
magik6k, line 425: Previously this was attoFil. We should rename this to something like `fee-limit`, so people don't accidentally set this to a value that's 1000000000000000000x too high

wenzhenxiang: The cli command (lotus mpool replace -h ) shows that the max-fee unit is FIL , fee-limit is a good name

magik6k: Yes, but it's likely in existing scripts, and we shouldn't just be changing how the number is interpreted, even if it is more correct (changing the flag name will just break those scripts, not changing it can lead to losing funds)

ZenGround0:
Moved to #6927 to commit review suggestion

Issue #2841, opened 2020-08-05
TestAPIRPC/testMining fails randomly


Kubuxu:
I've fixed the most common case but it is still randomly failing.
Try `go test -v -run 'TestAPIRPC/testMining$' -count 100 ./node` to reproduce.

Stebalien:
I can no longer reproduce this.

PR #5112, opened 2020-12-03
Use FundManager to withdraw funds, add MarketWithdraw API

### LINKED ISSUE #5100: `wallet market withdraw` should use FundManager.Withdraw()
`wallet market withdraw` currently attempts to withdraw funds from the storage market actor regardless of how much is reserved by ongoing deals.

It should instead use [`FundManager.Withdraw()`](https://github.com/filecoin-project/lotus/blob/master/chain/market/fundmanager.go#L128) so as to not withdraw an amount that would cause deals to run into funding issues.

Probably it make sense to add a `WalletWithdraw(ctx, from, addr, amt)` API endpoint, implemented in https://github.com/filecoin-project/lotus/blob/master/node/impl/full/wallet.go

ingar:
## Summary
Use the [FundManager's Withdraw](https://github.com/filecoin-project/lotus/blob/b13226bc2f8a0233b3beb9bd18ab266547bc29c7/chain/market/fundmanager.go#L128-L130) for the CLI command, `wallet market withdraw`.  This will take into account funds reserved for in-progress deals.  Also expose this FundManager method in the API.

Resolves #5100

### DIFF
--- a/api/api_full.go
+++ b/api/api_full.go
@@ -519,6 +519,8 @@ type FullNode interface {
 	MarketReserveFunds(ctx context.Context, wallet address.Address, addr address.Address, amt types.BigInt) (cid.Cid, error)
 	// MarketReleaseFunds releases funds reserved by MarketReserveFunds
 	MarketReleaseFunds(ctx context.Context, addr address.Address, amt types.BigInt) error
+	// MarketWithdraw withdraws unlocked funds from the market actor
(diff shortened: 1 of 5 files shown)


dirkmc (approved):
LGTM 👍

dirkmc:
TestAPIRPC tests failed in CI but I ran on my local machine and they passed so I think it's unrelated to this change.
They are known to be flaky: https://github.com/filecoin-project/lotus/issues/2841

magik6k (approved):

PR #5148, opened 2020-12-08
list-asks: remove redundant code.

raulk:


magik6k (changes requested):
It's not redundant, it's how we measure ping (perhaps should have a comment here - the first call is likely to be setting up the connection to the miner which isn't 1 roundtrip)

magik6k:
See https://github.com/filecoin-project/lotus/pull/5060#discussion_r532593304

magik6k:
Going to close this - a nicer / more correct fix would be to record this in markets.

PR #5160, opened 2020-12-09
fix(events): unlock mutex before error returns

iand:


magik6k (approved):
Thanks for catching this!

PR #5163, opened 2020-12-09
feat: add export actor command

frrist:
- accepts an actor head cid and exports the actors state to a car file.

Motivation for this is creating test vectors of actor state for visor testing.

Need to do a bit of manual testing before taking this out of draft. Currently waiting for my node to sync so I can try exporting some of the states I am interested in.


frrist:
achieved via other means, closing.

PR #5166, opened 2020-12-10
fix curSealing out of MaxSealingSectors limit

liyue201:


magik6k (approved):
Thanks for catching this

PR #5167, opened 2020-12-10
fetch size according to RegisteredSealProof

acautman:
write size according to RegisteredSealProof，default 32K size may result in low performance，when sector size is 512M/32G/64G，increase write size to 1M

### DIFF
--- a/extern/sector-storage/stores/remote.go
+++ b/extern/sector-storage/stores/remote.go
@@ -3,6 +3,8 @@ package stores
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"io"
 	"io/ioutil"
 	"math/bits"
 	"mime"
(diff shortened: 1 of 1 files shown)


extern/sector-storage/stores/remote.go
```diff
@@ -123,6 +149,12 @@ func (r *Remote) AcquireSector(ctx context.Context, s storage.SectorRef, existin
 	}
 	defer releaseStorage()
 
+	var iosize uint64 = 0
+	ssize, err := GetIoSizeByProofType(s.ProofType)
```
magik6k, line 153: I'd do `iosize = 32 << 10; if s.ProofType.SectorSize() > 8 << 20 {iosize=1<<20}`

extern/sector-storage/stores/remote.go
```diff
@@ -276,12 +308,57 @@ func (r *Remote) fetch(ctx context.Context, url, outname string) error {
 	case "application/x-tar":
 		return tarutil.ExtractTar(resp.Body, outname)
 	case "application/octet-stream":
-		return files.WriteTo(files.NewReaderFile(resp.Body), outname)
+		return WriteWithSize(files.NewReaderFile(resp.Body), outname, iosize)
```
magik6k, line 311: This used the files api to make things simpler. Instead of duplicating the code below, we can just do
```suggestion
		f, err := os.Create(outname)
		defer f.Close()
		if err != nil {
			return err
		}
		_, err = io.CopyBuffer(f, resp.Body, make([]byte, iosize))
		if err != nil {
			return err
		}
		return nil
```

extern/sector-storage/stores/remote.go
magik6k: (this way we can drop GetIoSizeByProofType from above)

magik6k (commented):
Thanks for this PR. Just a couple of comments - I like this idea

magik6k:
(applied suggestions in https://github.com/filecoin-project/lotus/pull/5177)

PR #5169, opened 2020-12-10
add mpoolpending from to filter

### LINKED ISSUE #5141: Feature request - supporting fetching mempool messages to/from address
A query param to `MpoolPending` so that I can (ideally, in a somewhat performant way from a web app pov) get a list of all messages in the mempool being sent to/from a particular address.

Could use the same `MessageMatch` type that's used in `StateListMessages`, defined [here](https://github.com/filecoin-project/lotus/blob/5c37c9d8e9c6e2c8d25f77c66ef4c8ef0ab7c251/api/api_full.go#L969) maybe?

Alongside https://github.com/filecoin-project/lotus/pull/4184, this would allow the Glif wallet to show a list of a user's pending transactions, and allow the user to replace any stuck messages at a higher fee.

deaswang:
fix #5141

### DIFF
--- a/cli/mpool.go
+++ b/cli/mpool.go
@@ -49,6 +49,14 @@ var mpoolPending = &cli.Command{
 			Name:  "cids",
 			Usage: "only print cids of messages in output",
 		},
+		&cli.StringFlag{
+			Name:  "to",
+			Usage: "return messages to a given address",
+		},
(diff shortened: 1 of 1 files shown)


magik6k (approved):
Thanks!

Issue #4746, opened 2020-11-06
frequently truncating sector log


zzx234234:
fsm.go:193:
		if len(state.Log) > 8000 {
			log.Warnw("truncating sector log", "sector", state.SectorNumber)
			state.Log[2000] = Log{
				Timestamp: uint64(time.Now().Unix()),
				Message:   "truncating log (above 8000 entries)",
				Kind:      fmt.Sprintf("truncate"),
			}

			state.Log = append(state.Log[:2000], state.Log[:6000]...)     //<---------is this sth wrong???
		}

magik6k:
We're truncating the middle part of the log, as the events at the beginning/end are usually the ones worth looking at when fixing issues

PR #5180, opened 2020-12-11
fix a coding error in truncating sector log, issue #4746

zzx234234:


magik6k (approved):
Ah, good catch, that was wrong

magik6k:
(will merge after 1.3 relesase)

PR #5197, opened 2020-12-15
fix lock propose

whyrusleeping:
this was not correct.


arajasek (approved):
and is now correct, aye

magik6k (approved):

PR #5200, opened 2020-12-16
Disable gas burning for window post messages

arajasek:
Implements https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0009.md

While over-estimation fees and miner tips are still paid, gas is no longer burnt
for direct, successful window PoSt messages.

Usually, gas is burnt to prevent an attacker from spamming the network and to
allow clients to "price" messages (using the base fee cap) based on how urgently
they need them to be processed. However:

1. Window PoSt is already a "proof of work".
2. Miners need to submit WindowedPoSts on-time so all window post messages are urgent.
3. Work is already under way to move window post verification off-chain (making
it effectively free). This change simply introduces the "free" part a bit earlier.

Upgrade epoch set for Monday, Dec 21st, 18:00 PST.

### DIFF
--- a/build/params_mainnet.go
+++ b/build/params_mainnet.go
@@ -41,6 +41,9 @@ const UpgradeKumquatHeight = 170000
 const UpgradeCalicoHeight = 265200
 const UpgradePersianHeight = UpgradeCalicoHeight + (builtin2.EpochsInHour * 60)
 
+// 2020-12-22T02:00:00Z
+const UpgradeClausHeight = 343200
+
 func init() {
 	policy.SetConsensusMinerMinPower(abi.NewStoragePower(10 << 40))
 
--- a/chain/vm/burn.go
+++ b/chain/vm/burn.go
@@ -67,7 +67,7 @@ func ComputeGasOverestimationBurn(gasUsed, gasLimit int64) (int64, int64) {
 	return gasLimit - gasUsed - gasToBurn.Int64(), gasToBurn.Int64()
 }
 
-func ComputeGasOutputs(gasUsed, gasLimit int64, baseFee, feeCap, gasPremium abi.TokenAmount) GasOutputs {
+func ComputeGasOutputs(gasUsed, gasLimit int64, baseFee, feeCap, gasPremium abi.TokenAmount, chargeNetworkFee bool) GasOutputs {
 	gasUsedBig := big.NewInt(gasUsed)
 	out := ZeroGasOutputs()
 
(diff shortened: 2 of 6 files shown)


chain/vm/vm.go
```diff
@@ -599,6 +607,30 @@ func (vm *VM) ApplyMessage(ctx context.Context, cmsg types.ChainMsg) (*ApplyRet,
...
+		// the trace, but I'm not sure if that's safe?
+		if toActor, err := st.GetActor(msg.To); err != nil {
+			// If the actor wasn't found, we probably deleted it or something. Move on.
+			if !xerrors.Is(err, types.ErrActorNotFound) {
+				// Otherwise, this should never fail and something is very wrong.
+				return false, xerrors.Errorf("failed to lookup target actor: %w", err)
+			}
+		} else if builtin.IsStorageMinerActor(toActor.Code) {
+			// Ok, this is a storage miner and we've processed a window post. Remove the burn.
+			burn = false
```
Stebalien, line 627: nit: given that we've factored this out, we can just return `false, nil` here.

Stebalien: (then return `true, nil` at the bottom)

chain/vm/vm.go
```diff
@@ -545,10 +546,11 @@ func (vm *VM) ApplyMessage(ctx context.Context, cmsg types.ChainMsg) (*ApplyRet,
 		}
 	}
 
-	var errcode exitcode.ExitCode
 	var gasUsed int64
 
-	if errcode = aerrors.RetCode(actorErr); errcode != 0 {
+	errcode := aerrors.RetCode(actorErr)
+
+	if errcode != 0 {
```
Stebalien, line 553: You can revert this change to reduce the diff size.

Stebalien (approved):

chain/vm/vm.go
arajasek: done

chain/vm/vm.go
arajasek: done

anorth:
Link to the FIP?

build/params_mainnet.go
```diff
@@ -41,6 +41,8 @@ const UpgradeKumquatHeight = 170000
 const UpgradeCalicoHeight = 265200
 const UpgradePersianHeight = UpgradeCalicoHeight + (builtin2.EpochsInHour * 60)
 
+const UpgradeClausHeight = 343200
```
anorth, line 44: Nit: it would be helpful to readers to include datestamps alongside these.

arajasek:
@anorth Good call! Edited the description.

chain/vm/burn.go
```diff
@@ -76,7 +76,12 @@ func ComputeGasOutputs(gasUsed, gasLimit int64, baseFee, feeCap, gasPremium abi.
 		baseFeeToPay = feeCap
 		out.MinerPenalty = big.Mul(big.Sub(baseFee, feeCap), gasUsedBig)
 	}
-	out.BaseFeeBurn = big.Mul(baseFeeToPay, gasUsedBig)
+
+	// If burning is disabled, just skip computing the BaseFeeBurn. However,
```
anorth, line 80: "network transaction fee" would be more descriptive than burn, here and for the parameter name.

anorth (approved):
Looks good, only minor suggestions about making this more self describing. Burn is a very generic word, used for other things too.

magik6k (commented):
Lgtm

PR #5204, opened 2020-12-16
Make 'mpool replace' work on a light client

Stebalien:
Unfortunately, the MpoolPending API call doesn't work on a light client. However, this doesn't seem necessary unless one is actually trying to lookup a message by nonce/sender.

NOTE: I'm not sure how safe this is. Was there a reason we needed to check the mpool?


Stebalien:
Not the right fix.

Issue #4991, opened 2020-11-24
Miner should not try to dial the client to restart a data transfer


dineshshenoy:
From @dirkmc 

 It appears that on restart the miner is trying to dial the client to restart a data transfer.
The miner shouldn't dial the client - the client should restart when the connection goes down.

error restarting data transfer: unable to send restart request: exhausted %!d(float64=5) attempts but failed to open stream, err: failed to dial 12D3KooWHAN67HgLoqrygQnouFF5uVhwM4kP7zurpR65Powtfgsa: all dials failed
  * [/ip4/127.0.0.1/tcp/33695] dial tcp4 127.0.0.1:33695: connect: connection refused
  * [/ip6/::1/tcp/39135] dial tcp6 [::1]:39135: connect: connection refused
  * [/ip4/10.240.0.46/tcp/33695] dial tcp4 0.0.0.0:9999->10.240.0.46:33695: i/o timeout
  * [/ip4/13.68.239.146/tcp/33695] dial tcp4 0.0.0.0:9999->13.68.239.146:33695: i/o timeout

Recommendation from @magik6k  and @hannahhoward 
1) Miner shouldn't try to restart the request.
2) Client should auto-retry when it sees the network connection drop (every 5 minutes)

hannahhoward:
Follow-up for @dirkmc

Specs for this ticket:
- remove auto restart for data transfers from provider in go-fil-markets - just leave deal as is, in transferring state
- in go-fil-markets, when a transfer becomes stalled (see event / message) on the client only start a go routine with a timer that attempts to restart the data transfer every 5 minutes -- I would rather NOT put this in go-data-transfer itself - not all users (including, most importantly, the provider) will want to attempt to restart

kernelogic:
My client lotus freeze and it keep printing error messages like this

2020-11-27T08:07:27.307Z	WARN	storagemarket_network	network/libp2p_impl.go:127	exhausted 5 attempts but failed to open stream, err: dial backoff
2020-11-27T08:07:27.307Z	WARN	storagemarket_impl	clientstates/client_states.go:176	error when querying provider deal state: %!w(*xerrors.wrapError=&{failed to open stream to miner 0xc29c4054c0 {[9796235 25912814 25925750]}})
[...]

Issue #5038, opened 2020-11-27
When creating a deal and the initial data transfer fails, Lotus should resolve the situation


dineshshenoy:
Canceling a deal locally by the client is already captured by: #4093 

When creating a deal, if the data transfer doesn't initiate correctly, Lotus asks clients whether we should restart a deal or cancel an existing deal.

For example:
552   Requested           ...U3cRpimY  ...rqfxzklq  Y           0B           ...oyvxyxr3ckoqq2ry7triy223lv6je75k7e"}}  
  Message: other peer appears to have hung up. restart Channel

Ideal Behavior:
Lotus should automatically cancel the deal so that there isn't a burden on the client.

dineshshenoy:
cc: @dirkmc

dirkmc:
This `other peer appears to have hung up` message appears in `lotus client list-transfers`.
However it seems that the deal state reported by `lotus client list-deals` is `StorageDealError`.

PR #5210, opened 2020-12-16
Miner should not dial client on restart

dirkmc:
Fixes https://github.com/filecoin-project/lotus/issues/4991 https://github.com/filecoin-project/lotus/issues/5211 https://github.com/filecoin-project/lotus/issues/5224 https://github.com/filecoin-project/lotus/issues/5268 https://github.com/filecoin-project/lotus/issues/5038

- [x] Error out if there are too many consecutive restarts with no data transferred https://github.com/filecoin-project/go-data-transfer/pull/129
- [x] Assess feasibility of writing an integration test / testground test
- [x] Manual test
It wasn't a small task to write an integration or testground test, so this should be tested manually:
  - Bring up a client and miner
  - Start transferring a large file from client to miner
  - While the file is transferring, restart the miner
  - Ensure that the file completes transferring and the deal moves to the sealing state

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,9 +31,9 @@ require (
 	github.com/filecoin-project/go-cbor-util v0.0.0-20191219014500-08c40a1e63a2
 	github.com/filecoin-project/go-commp-utils v0.0.0-20201119054358-b88f7a96a434
 	github.com/filecoin-project/go-crypto v0.0.0-20191218222705-effae4ea9f03
-	github.com/filecoin-project/go-data-transfer v1.2.3
+	github.com/filecoin-project/go-data-transfer v1.2.4
 	github.com/filecoin-project/go-fil-commcid v0.0.0-20201016201715-d41df56b4f6a
-	github.com/filecoin-project/go-fil-markets v1.0.10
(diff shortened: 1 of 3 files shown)


magik6k (approved):

magik6k:
(assuming that there's some good reason that this is a draft)

dirkmc:
👍 It's a draft at the moment because I want to try to write some integration tests today to make sure the retry logic works

dirkmc:
Yesterday we performed a manual test.
The data transfer will restart correctly if the client or the miner restarts once during the transfer.
If the client or miner restarts twice during the transfer, the second time the data rate monitor in go-data-transfer may miss the restart because of https://github.com/ipfs/go-graphsync/issues/141

However I think this is a big improvement, and there is also improved logging so we can see exactly what the client and miner side of the data transfer are doing. So I favour merging this PR and then bubbling up improvements to graphsync as they are released.

For future reference, to see all the relevant data transfer activity, the following log subsystems should be set at debug level:
- `markets`
- `dt-pushchanmon`
- `dt-impl`
- `data_transfer_network`

magik6k (approved):

Issue #5212, opened 2020-12-17
v1.3.0 Miner Panicked


rjan90:
**Describe the bug**
I updated my miner to v1.3.0, and it panicked with these logs:
```
2020-12-17T02:11:04.930+0100	DEBUG	advmgr	sector-storage/sched.go:450	SCHED Acceptable win: [[] [] [] []]
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x181c920]

goroutine 28039 [running]:
github.com/filecoin-project/lotus/chain/types.(*TipSet).Height(...)
	/home/filminer/lotus/chain/types/tipset.go:151
github.com/filecoin-project/lotus/markets/storageadapter.(*ProviderNodeAdapter).OnDealExpiredOrSlashed.func2(0x0, 0x0, 0x0, 0x0, 0x503fe, 0xc0188e9aa0, 0x25b5bba0, 0xa6823f57d1dee5e9)
	/home/filminer/lotus/markets/storageadapter/provider.go:347 +0x40
github.com/filecoin-project/lotus/chain/events.(*watcherEvents).StateChanged.func1(0x0, 0x0, 0x0, 0x0, 0x503fe, 0x0, 0x0, 0x0)
	/home/filminer/lotus/chain/events/events_called.go:440 +0x62
github.com/filecoin-project/lotus/chain/events.(*hcEvents).applyTimeouts(0xc0157a8240, 0xc014f0c400)
	/home/filminer/lotus/chain/events/events_called.go:289 +0x1af
github.com/filecoin-project/lotus/chain/events.(*hcEvents).processHeadChangeEvent(0xc0157a8240, 0x0, 0x0, 0x0, 0xc02f8ea2a0, 0x1, 0x1, 0x0, 0x0)
	/home/filminer/lotus/chain/events/events_called.go:153 +0x275
github.com/filecoin-project/lotus/chain/events.(*Events).headChange(0xc0157b4090, 0x0, 0x0, 0x0, 0xc02f8ea2a0, 0x1, 0x1, 0x0, 0x0)
	/home/filminer/lotus/chain/events/events.go:190 +0x18d
github.com/filecoin-project/lotus/chain/events.(*Events).listenHeadChangesOnce(0xc0157b4090, 0x32d3040, 0xc0157961c0, 0x0, 0x0)
	/home/filminer/lotus/chain/events/events.go:165 +0x68d
[...]

PR #5216, opened 2020-12-17
markets: Fix deal expired panic

magik6k:
Fixes https://github.com/filecoin-project/lotus/issues/5212

### DIFF
--- a/markets/storageadapter/client.go
+++ b/markets/storageadapter/client.go
@@ -264,7 +264,7 @@ func (c *ClientNodeAdapter) OnDealExpiredOrSlashed(ctx context.Context, dealID a
 	// and the chain has advanced to the confidence height
 	stateChanged := func(ts *types.TipSet, ts2 *types.TipSet, states events.StateChange, h abi.ChainEpoch) (more bool, err error) {
 		// Check if the deal has already expired
-		if sd.Proposal.EndEpoch <= ts2.Height() {
+		if ts2 == nil || sd.Proposal.EndEpoch <= ts2.Height() {
 			onDealExpired(nil)
 			return false, nil
 		}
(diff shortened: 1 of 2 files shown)


dirkmc (approved):

markets/storageadapter/client.go
```diff
@@ -264,7 +264,7 @@ func (c *ClientNodeAdapter) OnDealExpiredOrSlashed(ctx context.Context, dealID a
 	// and the chain has advanced to the confidence height
 	stateChanged := func(ts *types.TipSet, ts2 *types.TipSet, states events.StateChange, h abi.ChainEpoch) (more bool, err error) {
 		// Check if the deal has already expired
-		if sd.Proposal.EndEpoch <= ts2.Height() {
+		if ts2 == nil || sd.Proposal.EndEpoch <= ts2.Height() {
```
iand, line 267: ts2 == nil when there is a timeout which is handled differently on line 273

dirkmc: Line 273 is looking at `states` not `ts2`:
`if states == nil { ...`

In reality ts2 should never be nil, it seems like an issue in the events API.
But I think it's better to make this quick fix to make sure there are no further crashes until we track down the events API issue.

iand: It's set to nil when the timeout epoch is reached without seeing the tipset you were looking for. Clunky api but at least its documented.

dirkmc: It looks like the documentation is out of date 😭  The `oldState` and `newState` parameters have since been changed to a single `states` parameter:
https://github.com/filecoin-project/lotus/blob/c11803ad55a70986c39b0d2132579ecb1a9d5b21/chain/events/events_called.go#L401

I opened an issue: https://github.com/filecoin-project/lotus/issues/5218

magik6k:
Merging as a quick fix, we can have a correct-er fix later

PR #5219, opened 2020-12-17
lotus-wallet: Add interactive mode

magik6k:
Nice for things like looking at decoded message params on a secureish device before passing to ledger

### DIFF
--- a/cmd/lotus-wallet/interactive.go
+++ b/cmd/lotus-wallet/interactive.go
@@ -31,7 +31,7 @@ type InteractiveWallet struct {
 	lk sync.Mutex
 
 	apiGetter func() (api.FullNode, jsonrpc.ClientCloser, error)
-	under api.WalletAPI
+	under     api.WalletAPI
 }
 
 func (c *InteractiveWallet) WalletNew(ctx context.Context, typ types.KeyType) (address.Address, error) {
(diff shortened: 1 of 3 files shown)


magik6k:
Example prompt:
```
ACTION: WalletSign - Sign a message/deal
ADDRESS: f1...
TYPE: message
Message JSON: {
  "Version": 0,
  "To": "f2...",
  "From": "f1....",
  "Nonce": 3,
  "Value": "0",
  "GasLimit": 427524,
  "GasFeeCap": "1637334214",
  "GasPremium": "11372",
  "Method": 2,
  "Params": "hEIAY0AAQK==",
  "CID": {
    "/": "baf..."
  }
}
Value: 0 FIL
Max Fees: 0.069999999997297 FIL
Max Total Cost: 0.069999999997297 FIL
Method: Propose
Params: {
  "To": "f099",
  "Value": "0",
  "Method": 0,
  "Params": null
}
        MultiSig Proposal Value: 0 FIL
        MultiSig Proposal Hex Params: 
        MultiSig Proposal Method: Send
        MultiSig Proposal Params: {}

Accept the above? (Yes/No):
```

cmd/lotus-wallet/interactive.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+}
+
+var yeses = []string{
+	"yes",
+	"Yes",
+	"YES",
+	"approve",
+	"Approve",
+	"sign",
```
arajasek, line 229: sign/Sign isn't a good "random yes" because you're also using these for the prompt in `WalletNew`

arajasek (approved):
Approved with one nit (and needs rebasing).

Might also be good to docs folks attention?

PR #5242, opened 2020-12-22
update  weight canSeal and canStore when attach

luqz:
update  weight canSeal and canStore when attach storage


magik6k (approved):
This makes sense, thanks!

Issue #4624, opened 2020-10-28
flaky TestSyncBadTimestamp


dirkmc:
Times out after 10 minutes:
https://app.circleci.com/pipelines/github/filecoin-project/lotus/11124/workflows/09f81bd1-2462-4395-b1a5-48256e66d2c2/jobs/84296

PR #5244, opened 2020-12-22
fix(sync): enforce fork len when changing head

schomatis:
Similar to https://github.com/filecoin-project/lotus/pull/5182 this PR is theoretically protocol breaking but the only case of interest is when we are working with forks close to finality threshold (which is orders of magnitude away from normal forks observed in the chain).

---------------------------------

Most of the syncing logic seems to assume the "heaviest" tipset it receives as input is actually the head of our synced chain even if the lock is released. This is mostly innocuous and will be handled in follow-up PRs (to make this more explicit), but there are cases when we actually want to hold the lock while doing certain checks that are not related to the validity/integrity of a chain (either heaviest or not), namely how far away it is from the actual head we will be swapping it for (there may be other cases we should look into).

### DIFF
--- a/chain/store/store.go
+++ b/chain/store/store.go
@@ -363,7 +363,7 @@ func (cs *ChainStore) PutTipSet(ctx context.Context, ts *types.TipSet) error {
 
 // MaybeTakeHeavierTipSet evaluates the incoming tipset and locks it in our
 // internal state as our new head, if and only if it is heavier than the current
-// head.
+// head and does not exceed the maximum fork length.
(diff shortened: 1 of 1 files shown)


schomatis:
Failing flaky `TestSyncBadTimestamp` https://github.com/filecoin-project/lotus/issues/4624.

Kubuxu (approved):

Kubuxu:
Uff, IDK how I feel about merging this after RC is already out.
@Magik6k your call.

schomatis:
I'm fine reverting but someone needs to own that this does get merged eventually please.

PR #5246, opened 2020-12-23
Update SECURITY.md

dkkapur:
Based on feedback from @zgfzgf and a few others - recommending that we make this change to offer more clarity on what is meant by security issues (which are not always "Critical" in severity).


schomatis (approved):
Excellent, thanks for making this more clear and avoiding confusing expectations on vulnerability levels.

PR #5249, opened 2020-12-24
add maxfee limit for mpool replace

deaswang:
add really-do-it warn and limit for great gas fee.

### DIFF
--- a/cli/mpool.go
+++ b/cli/mpool.go
@@ -378,7 +378,12 @@ var mpoolReplaceCmd = &cli.Command{
 		},
 		&cli.StringFlag{
 			Name:  "max-fee",
-			Usage: "Spend up to X attoFIL for this message (applicable for auto mode)",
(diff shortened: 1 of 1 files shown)


jennijuju:
@Kubuxu could you please review this, or see if the message management refactor may cover this (if so, we may close the pr)?

Kubuxu:
It will break existing scripts and it will return an error by default for most uses (because the primary use of `mpool replace` is to replace a message when MaxFee previously limited it) which will lead to users passing `--really-do-it` by default.

jennijuju:
We will do a survey with some miners and see if they would like this feature.

jennijuju:
> We will do a survey with some miners and see if they would like this feature.

@deaswang thank you for contributing! we did a quick survey in Slack and the community doesn't think this is useful. So im gonna close this pr.
image

However, they do think if lotus can allow adjusting the default max-fees in miner config without restarting the miner would be super helpful! Let us know iif this would be something interesting for you to hack on!

PR #5251, opened 2020-12-25
Fix deadlock in runWorker in sched_worker.go 

liismn:
Avoid possible deadlock by sh.workersLk


magik6k (approved):
Thanks, I really wonder how this managed to be observed earlier.

PR #5252, opened 2020-12-25
add fund sufficient check in send

deaswang:
add fund sufficient check in send to avoid SysErrInsufficientFunds send. like https://filfox.info/zh/address/f1o2bwv33mqejxeqmh7qtb7wob4pjdho6rgvjp5la from 2020-12-23 14:40:30.

### DIFF
--- a/cli/send.go
+++ b/cli/send.go
@@ -156,7 +156,7 @@ var sendCmd = &cli.Command{
 		totalCost := types.BigAdd(types.BigMul(msg.GasFeeCap, types.NewInt(uint64(msg.GasLimit))), msg.Value)
 
 		if fromBalance.LessThan(totalCost) {
-			fmt.Printf("From balance %s attoFIL less than total cost %s attoFIL\n", fromBalance, totalCost)
+			fmt.Printf("From balance %s less than total cost %s\n", types.FIL(fromBalance), types.FIL(totalCost))
 			if !cctx.Bool("force") {
 				return fmt.Errorf("--force must be specified for this action to have an effect; " +
 					"you have been warned")
--- a/cli/send.go
+++ b/cli/send.go
@@ -63,7 +63,7 @@ var sendCmd = &cli.Command{
 			Usage: "specify invocation parameters in hex",
 		},
 		&cli.BoolFlag{
-			Name:  "really-do-it",
+			Name:  "force",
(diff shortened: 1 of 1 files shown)


cli/send.go
```diff
@@ -143,6 +148,23 @@ var sendCmd = &cli.Command{
 			Params:     params,
 		}
 
+		// Funds insufficient check
+		fromBalance, err := api.WalletBalance(ctx, msg.From)
+		if err != nil {
+			return err
+		}
+		totalCost := types.BigMul(msg.GasFeeCap, types.NewInt(uint64(msg.GasLimit)))
+		if msg.Method == builtin.MethodSend {
```
Kubuxu, line 157: the `msg.Value` is always sent over, no matter if the method is Send or note.

magik6k (commented):
Just 1 nitpick

cli/send.go
```diff
@@ -143,6 +148,21 @@ var sendCmd = &cli.Command{
...
 
+		// Funds insufficient check
+		fromBalance, err := api.WalletBalance(ctx, msg.From)
+		if err != nil {
+			return err
+		}
+		totalCost := types.BigAdd(types.BigMul(msg.GasFeeCap, types.NewInt(uint64(msg.GasLimit))), msg.Value)
+
+		if fromBalance.LessThan(totalCost) {
+			fmt.Printf("From balance %s attoFIL less than total cost %s attoFIL\n", fromBalance, totalCost)
```
magik6k, line 159: I'd do
```suggestion
			fmt.Printf("From balance %s less than total cost %s\n", types.FIL(fromBalance), types.FIL(totalCost))
```

cli/send.go
```diff
@@ -143,6 +148,21 @@ var sendCmd = &cli.Command{
 			Params:     params,
 		}
 
+		// Funds insufficient check
+		fromBalance, err := api.WalletBalance(ctx, msg.From)
+		if err != nil {
+			return err
```
magik6k, line 154: Maybe ignore this error when force is passed

magik6k (approved):

Kubuxu (approved):

cli/send.go
deaswang: if force specified will not check insufficient fund

magik6k (approved):

PR #5263, opened 2020-12-29
update guessgas

deaswang:
update guessgas use exist real gas limit.


Kubuxu:
MinGas should be adjusted.

jennijuju:
@Kubuxu could you please review the changes?

jennijuju:
We need to have a better source to pull these values and make sure the table is always up to date.

PR #5266, opened 2020-12-29
update StatusChangedActors api

deaswang:
Update StatusChangedActors API for add/del/change actors.


api/api_full.go
```diff
@@ -416,7 +416,7 @@ type FullNode interface {
 	StateAccountKey(context.Context, address.Address, types.TipSetKey) (address.Address, error)
 	// StateChangedActors returns all the actors whose states change between the two given state CIDs
 	// TODO: Should this take tipset keys instead?
-	StateChangedActors(context.Context, cid.Cid, cid.Cid) (map[string]types.Actor, error)
+	StateChangedActors(context.Context, cid.Cid, cid.Cid) (map[string]types.Actor, map[string]types.Actor, map[string]types.Actor, error)
```
magik6k, line 419: Multi-return doesn't work over jsonrpc

deaswang: add ChangedActors struct

Stebalien:
Unfortunately, this has gone stale and no longer applies. Additionally, while the new API is unarguably better, it's unclear if it's better _enough_ to make an API breaking change.

If you'd like to pick this up again, please file an issue and we'll discuss from there. Otherwise, I understand if you've moved on by now.

Stebalien:
Actually, I'll leave this open as a draft. I believe it could go in the "v1" API to avoid breaking changes.

PR #5271, opened 2020-12-30
Gavin fixed the issue5267 which runworker can cause dead lock

gavin-ygy:
The sched_worker.go has the dead lock bug in the fun runWorker().


magik6k:
Also fixed in https://github.com/filecoin-project/lotus/pull/5251/files

PR #5273, opened 2020-12-31
add StatePledgeCollateral api

deaswang:
`StateMinerPreCommitDepositForPower` and `StateMinerInitialPledgeCollateral` has a lot of repeat calculations and same parameter/return. merge the two apis to one `StatePledgeCollateral` api should be better. If accepted, `StateMinerPreCommitDepositForPower` and `StateMinerInitialPledgeCollateral` should remove.


arajasek:
Thanks for the contribution!

Let's add this in v1 api, and get rid of `StateMinerPreCommitDepositForPower` and `StateMinerInitialPledgeCollateral` at the same time.

Also note that `make gen` needs to be run.

jennijuju:
blocked by v1 api

ZenGround0:
@deaswang any interest in picking this back up now that it is no longer blocked?

PR #5274, opened 2020-12-31
clear struct param for same style

deaswang:


raulk (commented):
We will review Lotus code style in a few weeks, and I suggest we actually go the opposite way: naming parameters to be explicit in central interfaces like this. I don't mind clearing them for now for normalisation, because the current names aren't good anyway, but ultimately I propose to name them all.

magik6k:
Also worth noting that ideally this file would be auto-generated from definitions in `api/api*.go`

magik6k:
We now have codegen for this

PR #5275, opened 2020-12-31
lotus state call will panic

xgswust:
when use lotus state call method, in lotus/cli/state.go 1620 line, the paramObj needs struct type not a ptr type,if use a ptr type, the program will panic in line 1621(paramObj.NumField())


magik6k (approved):
This is an improvement, but still broken for methods which have take non-struct params

PR #5276, opened 2020-12-31
fix network version

deaswang:
new network version should take effect after height atOrBelow.


magik6k:
I'm pretty sure that the current code is correct. Can you give an example of when it's not? (also, we can't just change this logic as that's consensus-breaking)

deaswang:
https://github.com/filecoin-project/lotus/blob/7d0296c4095df4cc1b70401a17c262ceff9af01d/chain/stmgr/stmgr.go#L116-L120

yes, these lines have done what I have intend to do.
reference for these lines before.

https://github.com/filecoin-project/lotus/blob/7d0296c4095df4cc1b70401a17c262ceff9af01d/chain/stmgr/forks.go#L58-L103

PR #4995, opened 2020-11-24
rename oni/lotus-soup imports to testplans/lotus-soup and go get lotus@master on trigger

nonsense:
This PR is:
1. Locking a specific Lotus version in `go.mod` to be used with the testplans ran on TaaS (currently based on `go get lotus@master` but could be changed to use the commit from CircleCI as well).
2. Removing imports from `filecoin-project/oni`.
3. Removing go.mod replace to Lotus as in `replace github.com/filecoin-project/lotus => ../../` since the parent folder is not included with the testplan, and this doesn't work with the test plan.
4. Upgrading sectors from 2KiB to 8MiB - computationally sealing is similarly expensive, but this allows for future tests where we cancel transfer and restart nodes/miners in the middle of a transfer.
5. Upgrade deal size from 1600 bytes to 5MB.

---

It's worth noting that without making too many changes to the tests from one version of Lotus to another, we are sometimes getting different flakyness / different success rates. These tests were very stable for a few months, then they got broken for a month, and now they seem to pass some of the time. It is worth integrating them and running them soon as part of the Lotus test suite, so that we get them to a stable state, which is not easy since most of the logic really happens behind multiple APIs and is abstracted away from the test itself.

I am not sure how much of this is timing issues due to much lower block delays / propagation delays / etc. vs actual undergoing work on Lotus / go-fil-markets / etc.

---

PRIOR TO MERGE:
- [ ] remove `lock-lotus-version-in-testplan` branch from `.circleci/config.yaml`

---

TODO:
- [ ] 1. e2e test is still flaky - sometimes clients detect fork (although not all clients detect it):

```
2020-12-07T13:45:09.016Z        WARN    chain   chain/sync.go:1418      (fork detected) synced header chain ([bafy2bzaceay2gcgqbzqumy7meq3u6yjjghyk4svdt6fetvdgpdkr6inaa7tp2] - 19) does not link to our best block ([bafy2bzacebp3ut76gfxygwqocacgykreiegg5pdqhfxqx6yfl2gsoadxtyjcc] - 1)
2020-12-07T13:45:09.017Z        WARN    chain   chain/sync.go:1418      (fork detected) synced header chain ([bafy2bzacea5dzdbsj3azff7dvdulqffedw5eekk6jbs5wcqrcadbfzhhmqsey] - 19) does not link to our best block ([bafy2bzacebp3ut76gfxygwqocacgykreiegg5pdqhfxqx6yfl2gsoadxtyjcc] - 1)
[...]

PR #5282, opened 2021-01-04
Snake a context through the Chain-blockstore creation

ribasushi:
This sends through the startup daemon context for alternative blockstores that could benefit from it.

## EDIT (by @raulk)

Context: one datastore that will benefit from this is the Postgres-backed datastore. Since it is not an embedded datastore and needs to set up a network connection on initialisation, a context is tremendously useful.


node/repo/interface.go
```diff
@@ -54,7 +55,7 @@ type LockedRepo interface {
 	Datastore(namespace string) (datastore.Batching, error)
 
 	// Blockstore returns an IPLD blockstore for the requested domain.
```
raulk, line 57:
```suggestion
	// Blockstore returns an IPLD blockstore for the requested domain.
	// The supplied context must only be used to initialize the blockstore.
	// The implementation should not retain the context for usage throughout
	// the lifecycle.
```

node/repo/interface.go
```diff
@@ -54,7 +55,7 @@ type LockedRepo interface {
 	Datastore(namespace string) (datastore.Batching, error)
```
raulk, line 55: This looks a bit unlevelled with the Datastore method. They are both backed by some kind of store that would benefit from a cancellatio handle, so we should also propagate a context here.

ribasushi: Generally fair, though not within scope of what I am trying to push out at present.
I might circle back to this at a later date when I look into migrating the datastore to be rdbms backed as wel, but I am nowhere near there yet.

@raulk do you consider lack of ctx here blocker for the PR ?

raulk: I'm happy to push this to this PR. If we don't do it now, we are piling up one more piece of technical debt IMO.

ribasushi: @raulk if you have the bandwidth - by all means, use the same PR

raulk (changes requested):
Signalling blockage until we even out the API, as per review comments. I'll take care of it.

raulk (approved):
The interfaces are levelled out now. I'm good to merge. Thanks @ribasushi.

raulk:
ping @magik6k for final call / merge.

raulk (changes requested):
Not yet, I forgot to push a commit.

raulk:
Depends on #5425 to fix the build-lotus-soup CI job.

node/modules/storageminer.go
```diff
@@ -357,8 +357,9 @@ func NewProviderPieceStore(lc fx.Lifecycle, ds dtypes.MetadataDS) (dtypes.Provid
 	return ps, nil
 }
 
-func StagingMultiDatastore(lc fx.Lifecycle, r repo.LockedRepo) (dtypes.StagingMultiDstore, error) {
-	ds, err := r.Datastore("/staging")
+func StagingMultiDatastore(lc fx.Lifecycle, mctx helpers.MetricsCtx, r repo.LockedRepo) (dtypes.StagingMultiDstore, error) {
+	ctx := helpers.LifecycleCtx(mctx, lc)
+	ds, err := r.Datastore(ctx, "/staging")
```
Kubuxu, line 362: Doc for the `Datastore` says that ctx is only for startup, here we are passing a lifecycle context from start to the end of Lotus runtime.

raulk: The `Datastore()` constructor should not retain the context. It should only be used for the start operation, and this is the only context that we have that is cancelled if the startup is cancelled.

Kubuxu: Hmm, we should be using the context from OnStart here, but it makes everything way harder.

raulk: Yeah, that would be unnecessarily complex IMO.

raulk (approved):

raulk:
Approved, but #4995 needs to land first for the tests to pass.

magik6k (approved):

Issue #4277, opened 2020-10-09
Add --chain cli feature to support between chains


mFarghaly:
Please add a CLI flag for `--chain` to switch between networks instead of compiling multiple binaries and building Docker images for each network.

```
$ lotus # connect by default to mainnet
$ lotus --chain testnet
$ lotus --chain nerpa
$ lotus --chain butterfly
```

Most blockchain protocols clients like Ethereum and Bitcoin has a flag or CLI option for switching between different networks.

dineshshenoy:
This seems like it could be a dangerous operation.  Do we need to reset the node or should it be in a config file?

rjan90:
I think this issue can be closed now as this has been merged? #rengjøring

PR #5288, opened 2021-01-05
Create a calibnet build option

arajasek:
This is a first step towards #4277. I think a `--network` option to the daemon would be nice, but this is at least better than having to switch branches to switch between mainnet and calibnet.

Could use with a bit of manual testing, but generally seems to work.


magik6k (approved):
Much nicer

Kubuxu (approved):

Issue #5125, opened 2020-12-05
Making many deals results in many small "AddBalance" messages (expensive with high gas)


jimpick:
**Describe the bug**

If there isn't enough market balance, any new deal will generate a new "AddBalance" message for only the value of the deal, which is typically very small given current pricing.

In my testing, I'm generating many deals, and I notice on Filfox that there are many, many AddBalance messages, sometimes for tiny amounts (because my deal size is small), eg.:

https://filfox.info/en/message/bafy2bzacebvfet56oiurs4mcny7gczojp7iumin25mkj4hbuaweesjcodlbvi

AddBalance amount: 0.0007188595 FIL
Gas (burn) fee: 0.026704434915442184 

So ~97% of the charge went to gas and only ~3% went to the market balance.

My test run is ongoing, but there have already been 35 such messages...

All the little messages are also consuming chain bandwidth.

**To Reproduce**

Make lots of deals.

**Expected behavior**

A better behaviour would be to fund the balance with a larger amount only when the balance goes below a certain level.

**Version (run `lotus version`):**

```
Daemon:  1.2.2+git.b13226bc2+api1.0.0
Local: lotus version 1.2.2+git.b13226bc2
```

jimpick:
My test run is over, so I'll share my results. This might be a little pathological, as it was a fresh wallet and the base fee is quite high right now and I'm making many small deals - but it illustrates the issue perfectly.

Number of attempted deals: ~2252 (executed in series, 1 for each miner I have an annotation for)
File size: 90719 bytes (a small jpeg file - the purpose of the deals is to probe the network)
Total balance added: 1.5 FIL
Remaining balance at end: 0.403445188962896898 FIL
Funds spent: 1.5 - 0.40 = ~1.1 FIL = ~ 32USD
Number of messages: 47
Total value transferred in messages: 0.2153264682 FIL
Funds spent on gas: 1.1 FIL - 0.2 FIL = ~0.9 FIL
Escrow at end: 0.02153264682 FIL
Locked at end: 0.0003131205 FIL
Number of successfully sealing deals in the end: 18

[messages.ndjson.txt](https://github.com/filecoin-project/lotus/files/5646690/messages.ndjson.txt)

jimpick:
I made a little utility with Node.js and Filecoin.js to transfer funds from a default Lotus wallet to the market balance.

https://github.com/jimpick/filecoin-market-add-balance/blob/main/index.js

dineshshenoy:
[...]

PR #5300, opened 2021-01-06
feat: lotus wallet market add (adds funds to storage market actor)

dirkmc:
Fixes https://github.com/filecoin-project/lotus/issues/5125

New CLI command to add funds to the Storage Market Actor:
`./lotus wallet market add 0.05`


magik6k (approved):
We could also warn users about low funds in the storage market actor when making deals with the `lotus client deal` cli

askender:
Besides, how to adds funds to a miner which accept deal?

dirkmc:
@askender I believe that this new CLI command will allow you to add funds to the miner's Storage Market Actor balance.
Is there a reason you can see that it would not?

PR #5307, opened 2021-01-08
Add a command to get the fees of a deal

### LINKED ISSUE #5294: Miners would like to see outstanding FIL receivables for deals.
Currently, when a storage deal is made, the total amount of FIL owed by the client is locked at the client side and not visible to the miner.  The incremental payments are only shown in the Market Balance in lotus-miner info over the stored period.  It would be very helpful to miners if outstanding FIL receivables from deals can be shown.

arajasek:
Fixes #5294

### DIFF
--- a/chain/actors/builtin/market/market.go
+++ b/chain/actors/builtin/market/market.go
@@ -1,6 +1,7 @@
 package market
 
 import (
+	"github.com/filecoin-project/go-state-types/big"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-address"
@@ -152,4 +153,20 @@ func EmptyDealState() *DealState {
 		SlashEpoch:       -1,
 		LastUpdatedEpoch: -1,
 	}
+}
+
+// returns the earned fees and pending fees for a given deal
+func (deal DealProposal) GetDealFees(height abi.ChainEpoch) (abi.TokenAmount, abi.TokenAmount) {
+	tf := big.Mul(deal.StoragePricePerEpoch, big.NewInt(int64(deal.EndEpoch-deal.StartEpoch)))
+
+	ef := big.Mul(deal.StoragePricePerEpoch, big.NewInt(int64(height-deal.StartEpoch)))
+	if ef.LessThan(big.Zero()) {
+		ef = big.Zero()
+	}
+
+	if ef.GreaterThan(tf) {
+		ef = tf
+	}
+
+	return ef, big.Sub(tf, ef)
 }
--- a/cmd/lotus-shed/main.go
+++ b/cmd/lotus-shed/main.go
@@ -34,6 +34,7 @@ func main() {
 		postFindCmd,
 		proofsCmd,
 		verifRegCmd,
+		marketCmd,
 		miscCmd,
 		mpoolCmd,
 		genesisVerifyCmd,
--- a/cmd/lotus-shed/market.go
+++ b/cmd/lotus-shed/market.go
@@ -1,0 +1,102 @@
+package main
+
+import (
+	"fmt"
+
+	lcli "github.com/filecoin-project/lotus/cli"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/urfave/cli/v2"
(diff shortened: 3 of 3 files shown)


cli/state.go
```diff
@@ -1868,3 +1826,144 @@ var stateNtwkVersionCmd = &cli.Command{
...
+	Name:  "deal",
+	Usage: "Get information about deal(s)",
+	Subcommands: []*cli.Command{
+		stateGetDealSetCmd,
+		stateDealFeesCmd,
+	},
+}
+
+var stateGetDealSetCmd = &cli.Command{
+	Name:      "get-deal",
```
magik6k, line 1840: `state deal get-deal`? I'd make it just `state deal get`.

cli/state.go
```diff
@@ -1868,3 +1826,144 @@ var stateNtwkVersionCmd = &cli.Command{
...
+			return err
+		}
+		fmt.Println(string(data))
+
+		return nil
+	},
+}
+
+var stateDealFeesCmd = &cli.Command{
+	Name:  "get-fees",
```
magik6k, line 1882: `fees`?

cli/state.go
```diff
@@ -1868,3 +1826,144 @@ var stateNtwkVersionCmd = &cli.Command{
...
+			if err != nil {
+				return err
+			}
+
+			ht = head.Height()
+		} else {
+			ht = ts.Height()
+		}
+
+		if pdrStr := cctx.String("provider"); pdrStr != "" {
```
magik6k, line 1920:
```suggestion
		if cctx.IsSet("provider") {
```

chain/actors/builtin/market/market.go
```diff
@@ -140,3 +141,15 @@ func EmptyDealState() *DealState {
 		LastUpdatedEpoch: -1,
 	}
 }
+
+// returns the earned fees and pending fees for a given deal
+func (deal DealProposal) GetDealFees(height abi.ChainEpoch) (abi.TokenAmount, abi.TokenAmount) {
+	tf := big.Mul(deal.StoragePricePerEpoch, big.NewInt(int64(deal.EndEpoch-deal.StartEpoch)))
+
+	ef := big.Mul(deal.StoragePricePerEpoch, big.NewInt(int64(height-deal.StartEpoch)))
```
magik6k, line 149: Should also be capped at total fees / deal expiration

arajasek: Oops, that was silly...thanks!

cli/state.go
arajasek: yeah, that's better

cli/state.go
```diff
@@ -1868,3 +1826,144 @@ var stateNtwkVersionCmd = &cli.Command{
...
+	},
+}
+
+var stateDealFeesCmd = &cli.Command{
+	Name:  "fees",
+	Usage: "View the storage fees associated with a particular deal or storage provider",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "provider",
+			Usage: "provider whose outstanding fees you'd like to recover",
```
Kubuxu, line 1887: recover?

arajasek: changed to `calculate`

cli/state.go
```diff
@@ -1868,3 +1826,144 @@ var stateNtwkVersionCmd = &cli.Command{
...
+			ht = ts.Height()
+		}
+
+		if cctx.IsSet("provider") {
+			p, err := address.NewFromString(cctx.String("provider"))
+			if err != nil {
+				return fmt.Errorf("failed to parse provider: %w", err)
+			}
+
+			deals, err := api.StateMarketDeals(ctx, ts.Key())
```
magik6k, line 1926: I'd avoid this API call, it will return all the milions of deals in the market actor, calling it with curl the response is 366M and takes 1.5min

jennijuju: @arajasek thoughts on this?

cli/state.go
jennijuju: @Kubuxu this looks good to you now?

arajasek:
Discussion summary: Make this a shed command (since it's gonna be slooowww). The CLI command itself should be better, perhaps using markets info as source of truth.

arajasek:
Now just a shed command

magik6k (approved):

Issue #4878, opened 2020-11-16
Make PublishStorageDeals logic clever


arajasek:
Some things we can do:

- Wait to submit a `PublishStorageDeals` message if the base fee "seems" high, and the deals don't start anytime soon
- Do not submit a `PublishStorageDeals` message if it seems like the cost of the message is (significantly?) greater than the total reward of the deal itself
- Combine multiple deals into one `PublishStorageDeals` message

jennijuju:
can this be some config value where miner say: if the base fee is greater than x FIL, stop sending out type A, B, C message?

or the other way around, if the base fee is greater than xFIL, only send out type A,B,C messages?

flyworker:
I can share something we do this morning for config the price with our miners

if  gas fee < 0.1nano : price =0.000000001
if gas fee > 0.1 nano : price = 0.000000001*gas fee/ 0.1nano

those are based on 1GB avg file size

jennijuju:
karalabe suggested: : I should be able to say I want X profit, and if the basefee skyrockets, then my deal fee should go along with it.

PatrickDeuse:
with the new control addresses in v1.2.3 there might be the possibility to at first give the pubSD messages their own control address - combined with the value in the config file this is already way cleverer than it was before. basically getting the messages out of the way to not block transfers, etc. then add config params for that message to fine grain the message values.

but the blocking issue does not disappear with all these funky param-set config values and a special lane control address. a pubSD that gets not on chain will still block a pubSD that could get on chain. and deadlines will expire... 

there might be more needed to actually improve (financial, automated) deal success than just stopping the pubSD from causing financial harm due to individual (per message) message params.

dineshshenoy:
It appears that some teams are already putting multiple deals in the same [message](https://filfox.info/en/message/bafy2bzacedokitfgain7irzvft5uluf4jzqkicscrunoq7em6r5ldp57zcbn6).   This could achieve significant cost savings especially when the base fee is high and worth doing. cc: @magik6k

KaptinLin:
See [message](https://filfox.info/zh/message/bafy2bzaceburmy7lsaxmcjq3zyy6uhyc7jhoq4iwr3f7gw2p4kpn5r43ur3ay), it merges 30 deals in one message, the fee goes down a lot. It should be the top priority to do if we want to increase the number of deals currently. Give a wait time, and merge the deals into one in that time period.

jennijuju:
[...]

PR #5309, opened 2021-01-08
batch publish deal messages

dirkmc:
Fixes https://github.com/filecoin-project/lotus/issues/4878

TODO:
- [x] Fix GetCurrentDealInfo to work with publish messages that have multiple deals
- [x] Integration tests
- [x] Remove dealID from go-fil-markets calls to OnDealSectorPreCommitted & OnDealSectorCommitted https://github.com/filecoin-project/lotus/pull/5413
- [x] [Fix storage sealing](https://github.com/filecoin-project/lotus/blob/master/extern/storage-sealing/states_failed.go#L409) so that it can handle publish messages that have multiple deals https://github.com/filecoin-project/lotus/pull/5411
- [x] Move handling of publish deals message return value from markets into lotus https://github.com/filecoin-project/lotus/pull/5428 https://github.com/filecoin-project/go-fil-markets/pull/475
- [ ] Document config params - _Will do this once the PR is approved, in case we decide on different config during review_


api/test/deals.go
```diff
@@ -8,103 +8,45 @@ import (
...
-	"github.com/filecoin-project/lotus/chain/types"
-	"github.com/filecoin-project/lotus/node/impl"
-	ipld "github.com/ipfs/go-ipld-format"
 )
 
 func TestDealFlow(t *testing.T, b APIBuilder, blocktime time.Duration, carExport, fastRet bool, startEpoch abi.ChainEpoch) {
+	s := setupOneClientOneMiner(t, b, blocktime)
+	defer s.blockMiner.Stop()
 
-	ctx := context.Background()
```
dirkmc, line 38: Most of the changes in this file are just refactoring some common code into functions

node/config/def.go
```diff
@@ -208,6 +218,11 @@ func DefaultStorageMiner() *StorageMiner {
 			ExpectedSealDuration: Duration(time.Hour * 24),
 		},
 
+		PublishMsg: PublishMsgConfig{
+			PublishPeriod:  Duration(time.Hour),
+			MaxDealsPerMsg: 8,
+		},
```
dirkmc, line 224: These are placeholder defaults for `PublishPeriod` and `MaxDealsPerMsg`, open to suggestions for these values

markets/storageadapter/getcurrentdealinfo.go
```diff
@@ -8,72 +8,107 @@ import (
...
 	"github.com/ipfs/go-cid"
 	"golang.org/x/xerrors"
 )
 
 type getCurrentDealInfoAPI interface {
+	ChainGetMessage(ctx context.Context, mc cid.Cid) (*types.Message, error)
 	StateLookupID(context.Context, address.Address, types.TipSetKey) (address.Address, error)
 	StateMarketStorageDeal(context.Context, abi.DealID, types.TipSetKey) (*api.MarketDeal, error)
-	StateSearchMsg(context.Context, cid.Cid) (*api.MsgLookup, error)
+	StateWaitMsg(ctx context.Context, msg cid.Cid, confidence uint64) (*api.MsgLookup, error)
```
dirkmc, line 23: For some reason when I use StateSearchMsg, in some of the integration tests StateSearchMsg would return an error indicating the [block was not found in the blockstore](https://app.circleci.com/pipelines/github/filecoin-project/lotus/12358/workflows/006e0e18-a7a7-4b51-a93d-d38a868d9c58/jobs/105721). But when I used StateWaitMsg I don't get this error.
This is strange because before we get to calling GetCurrentDealInfo we have already called StateWaitMsg, so the message should have been executed successfully.

s0nik42:
@dirkmc  how deal batching will be exposed to miners via lotus  ? specific command line or automatically ?

dirkmc:
It will be automatic, with two config values:
- The maximum number of deals to include in a publish message
- The amount of time to wait for more deals after the first deal arrives, before publishing the deal

node/config/def.go
s0nik42: What is the MaxDealsPerMsg supported by the network ?

PatrickDeuse: i have seen up to 30 on the chain already -but this should be proper documented with the max values

markets/storageadapter/dealpublisher.go
```diff
@@ -0,0 +1,294 @@
...
+type publishResult struct {
+	msgCid cid.Cid
+	err    error
+}
+
+type pendingDeal struct {
+	ctx    context.Context
+	deal   market2.ClientDealProposal
+	Result chan publishResult
+}
```
magik6k, line 286: Would move those definitions next to DealPublisher so you don't have to look for them

markets/storageadapter/dealpublisher.go
```diff
@@ -0,0 +1,294 @@
...
+			}
+		}()
+	}
+}
+
+// Sends the publish message
+func (p *DealPublisher) publishDealProposals(deals []market2.ClientDealProposal) (cid.Cid, error) {
+	log.Infof("publishing %d deals in publish deals queue with piece CIDs: %s", len(deals), pieceCids(deals))
+
+	provider := deals[0].Proposal.Provider
```
magik6k, line 229: I'd at least assert that the provider is the same for all deals; there may be a point where we'll want to support multiple miner actors from a single lotus-miner process

magik6k (approved):
Just a couple of nits, generally looks good.

Would also be great to have commands to see pending deals / force send a publish message

dirkmc:
> Would also be great to have commands to see pending deals / force send a publish message

Good idea, I created a ticket: https://github.com/filecoin-project/lotus/issues/5491

jennijuju:
Seems like if  I’m sending a message for deals 1,2,3,4,5, what will happen to 1,2 when the time of sending the message has passed the 3’s start epoch, the message will fail for all deals.

will one be able to republish the valid deals, say for deals 4,5?

assuming we add some checks to run the message locally first before sending it out, what will happen if by the time of  the message landed on chain, one of the deal expired?

dirkmc:
https://github.com/filecoin-project/lotus/pull/5505 implements a fix for the case where the deal proposal start epoch has already elapsed by the time the publish message is about to be sent. The fix is that we will filter this deal out of the publish message.

I'm not sure if we can do much in lotus about the case where the publish message takes a few epochs to land on chain, and in the mean time the start epoch of one of the deals expires. I believe we would need a change in specs-actors to account for this case:
- Currently the market actor [validates each deal](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_actor.go#L203)
- If one of the deals fails validation, [the whole message fails](https://github.com/filecoin-project/specs-actors/blob/57195d8909b1c366fd1af41de9e92e11d7876177/actors/builtin/market/market_actor.go#L756)

magik6k (approved):
LGTM. Also did some manual testing on my mining setup, and it seems to work well

PatrickDeuse:
can you give a quick overview on the config params until they are properly documented?

dirkmc:
@PatrickDeuse I just opened a PR against filecoin docs here: https://github.com/filecoin-project/filecoin-docs/pull/658
Please let me know if that explanation is clear.

PR #5320, opened 2021-01-10
Multicore AddPiece CommP

magik6k:
Before: `BenchmarkAddPiece512M-32    	       1	7034478758 ns/op	  75.72 MB/s`

After: `BenchmarkAddPiece512M-32    	       2	 916931366 ns/op	 580.93 MB/s`
(on AMD TR2950X)

### DIFF
--- a/extern/sector-storage/ffiwrapper/sealer_cgo.go
+++ b/extern/sector-storage/ffiwrapper/sealer_cgo.go
@@ -142,7 +142,10 @@ func (sb *Sealer) AddPiece(ctx context.Context, sector storage.SectorRef, existi
 			break
 		}
 
-		done := make(chan struct{cid.Cid; error}, 1)
+		done := make(chan struct {
+			cid.Cid
+			error
+		}, 1)
 		pbuf := <-throttle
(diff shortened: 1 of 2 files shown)


Kubuxu:
Are you 100% that pieces can be combined and the resulting CID will be the same? Do we have tests for this?

magik6k:
> Are you 100% that pieces can be combined and the resulting CID will be the same? Do we have tests for this?

Yep, It's a simple binary tree under the hood. (The CID in TestAddPiece512M was generated in a single call. Also, we are already doing this chunking on master, this is just making things run in parallel)

Kubuxu (approved):

vyzo (approved):
i neeed CPUz!!!!!

magik6k:
(tested manually on my miner, works, and it's really fast there too)

shaodan:
@magik6k Should we change resource that addpiece needs in [sector-storage/resources.go](https://github.com/filecoin-project/lotus/blob/master/extern/sector-storage/resources.go#L53)

PR #5327, opened 2021-01-12
document ParentStateRoot

schomatis:


chain/types/blockheader.go
```diff
@@ -63,8 +63,17 @@ type BlockHeader struct {
...
+	// CID of the root data structure representing the state of the *parent*
+	// tipset: the VM state generated after applying the messages from the
+	// blocks in `Parents` (not the messages from *this* block).
+	// This is redundant information which is validated to match against the
+	// state generated by our VM locally (see `ValidateBlock`). Once validated
+	// it is accepted as a reliable representation of our cached version of the
+	// state and referenced directly for performance reasons.
 	ParentStateRoot cid.Cid // 8
 
+	// Similar use as ParentStateRoot above but for the receipts of the
```
ribasushi, line 75: @schomatis the wording
> Similar use as ParentStateRoot above but for the receipts 
here implies that `ParentMessageReceipts` is
> redundant information which is validated to match 

My understanding is that there is *no* other place that links to the receipts accompanying a state-transition. Could you clarify?

schomatis: I'm going to remove the term redundant altogether as it does not apply to potential lite clients that will download the state and query it independently of the VM.

With regard to the receipt vs state distinction I'm not sure I see the difference. AFAIK the receipts are generated along with the state, meaning if I could generate one from the VM I can generate the other.

PR #5328, opened 2021-01-12
refactor Sync top level functions

schomatis:
Homogenize `Sync()` calls to better reflect the different states.

Remove `StagePersistHeaders`, absorbed by `StageHeaders` now (the first being local is of low complexity, execution time and possibilities of error compared to the second).

Extract `syncMessagesAndCheckState` out of `collectChain()` and into the main `Sync()` call flow.

### DIFF
--- a/api/api_full.go
+++ b/api/api_full.go
@@ -840,7 +840,6 @@ type SyncStateStage int
 const (
 	StageIdle = SyncStateStage(iota)
 	StageHeaders
-	StagePersistHeaders
 	StageMessages
 	StageSyncComplete
 	StageSyncErrored
@@ -851,8 +850,6 @@ func (v SyncStateStage) String() string {
 	switch v {
 	case StageHeaders:
 		return "header sync"
-	case StagePersistHeaders:
-		return "persisting headers"
 	case StageMessages:
 		return "message sync"
 	case StageSyncComplete:
--- a/api/api_full.go
+++ b/api/api_full.go
@@ -837,9 +837,13 @@ type SyncState struct {
 
 type SyncStateStage int
 
+// FIXME: Review names. Tipset instead of headers. Describe the action
+//  not the object (fetch and validate instead of headers and messages).
 const (
 	StageIdle = SyncStateStage(iota)
+	// Fetch tipsets.
 	StageHeaders
+	// Validate blocks in tipsets (and messages in those blocks).
 	StageMessages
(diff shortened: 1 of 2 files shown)


chain/sync.go
```diff
@@ -573,15 +589,40 @@ func (syncer *Syncer) Sync(ctx context.Context, maybeHead *types.TipSet) error {
...
 		return xerrors.Errorf("collectChain failed: %w", err)
 	}
 
+	ss.SetStage(api.StageMessages)
+
+	// FIXME: `StageFetchingMessages` is buried inside this function (in
+	//  `iterFullTipsets`). We should extract it out and probably request
+	//  *both* block headers and messages together through the corresponding
+	//  chain exchange interface. We can do very little validation without the
+	//  messages so it shouldn't have much impact.
```
Kubuxu, line 612: Fetching messages has quite an impact on sync speed. We should not be doing it eagerly.
One of the most important parts of validation is not syncing headers we already failed to sync (and not fetching messages for them because we did not persist them).

schomatis: Fair point. We probably shouldn't request as much messages as chain we are fetching but it still seems to me we could benefit from eagerly requesting _some_ amount of messages in advance for upcoming blocks. We are currently at the other extreme of only requesting at the same moment of validation stalling the pipeline. Anyway, just a note, not going to change any of that here.

api/api_full.go
```diff
@@ -837,10 +837,13 @@ type SyncState struct {
 
 type SyncStateStage int
 
+// FIXME: Review names. Tipset instead of headers. Describe the action
+//  not the object (fetch and validate instead of headers and messages).
 const (
 	StageIdle = SyncStateStage(iota)
+	// Fetch tipsets.
 	StageHeaders
-	StagePersistHeaders
```
Kubuxu, line 843: We should leave it here, for now, removing it will re-number the enum resulting in incompatibility.
We should open an issue for API breaking changes we want to do, mention this renumbering and link to the issue from here.

Kubuxu (changes requested):
It removes a constant from one of the few packages I would consider exported interface and also re-numbers the enum. 
Let's avoid that for now.

ZenGround0:
@magik6k would you say this no longer makes sense given your work refactoring lotus/chain?

schomatis:
Yeah, we can close this.

PR #5350, opened 2021-01-14
avoid use mp.cfg directly to avoid race

GFZRZK:
Since there exists a way to modify the value of mpool.cfg, it will cause a data race and create some unexpected cases.  Lock it with an RWMutex is good

### DIFF
--- a/chain/messagepool/config.go
+++ b/chain/messagepool/config.go
@@ -48,9 +48,13 @@ func saveConfig(cfg *types.MpoolConfig, ds dtypes.MetadataDS) error {
 }
 
 func (mp *MessagePool) GetConfig() *types.MpoolConfig {
-	mp.cfgLk.Lock()
-	defer mp.cfgLk.Unlock()
-	return mp.cfg.Clone()
+	return mp.getConfig().Clone()
+}
+
+func (mp *MessagePool) getConfig() *types.MpoolConfig {
+	mp.cfgLk.RLock()
+	defer mp.cfgLk.RUnlock()
+	return mp.cfg
 }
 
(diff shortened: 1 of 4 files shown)


vyzo (changes requested):
Well that's fine in principle, but now we are cloning the confing _every time_ we need to access something.

I would prefer if we defined accessors that leverage the lock for the various fields instead of doing that dance, it can be quite a lot of allocations.

vyzo:
Also of note, we never modify fields of the config directly, we only read them and the setting of the config copies wholesale.
So the only potential for race is in the pointer access -- so just make a method that returns the config object with the RWLock and we'll be fine.

vyzo:
Note that this is different from GetConfig which is part of the API and returns the (cloned) config for external consumption; the new accessor should be private.

chain/messagepool/config.go
```diff
@@ -48,11 +48,17 @@ func saveConfig(cfg *types.MpoolConfig, ds dtypes.MetadataDS) error {
 }
 
 func (mp *MessagePool) GetConfig() *types.MpoolConfig {
-	mp.cfgLk.Lock()
-	defer mp.cfgLk.Unlock()
+	mp.cfgLk.RLock()
+	defer mp.cfgLk.RUnlock()
 	return mp.cfg.Clone()
```
vyzo, line 53: this could be written as `getConfig().Clone()`

vyzo (approved):
Just a minor comment, it would be nice to write it that way.
Other than that, looks good.

magik6k (approved):

magik6k:
Thanks!

PR #5355, opened 2021-01-14
chore: export vm.ShouldBurn

iand:
Required for sentinel to calculate gas outputs for each message. 

There is some noise caused by gofumpt formatting of the file.

### DIFF
--- a/chain/vm/vm.go
+++ b/chain/vm/vm.go
@@ -43,9 +43,11 @@ import (
 
 const MaxCallDepth = 4096
 
-var log = logging.Logger("vm")
-var actorLog = logging.Logger("actors")
-var gasOnActorExec = newGasCharge("OnActorExec", 0, 0)
+var (
+	log            = logging.Logger("vm")
+	actorLog       = logging.Logger("actors")
+	gasOnActorExec = newGasCharge("OnActorExec", 0, 0)
+)
 
 // stat counters
 var (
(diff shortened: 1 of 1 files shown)


chain/vm/vm.go
```diff
@@ -43,9 +43,11 @@ import (
...
 const MaxCallDepth = 4096
 
-var log = logging.Logger("vm")
-var actorLog = logging.Logger("actors")
-var gasOnActorExec = newGasCharge("OnActorExec", 0, 0)
+var (
+	log            = logging.Logger("vm")
+	actorLog       = logging.Logger("actors")
+	gasOnActorExec = newGasCharge("OnActorExec", 0, 0)
+)
```
raulk, line 50: Ideally we would introduce formatting changes in separate PRs, so that future git blames can be less noisy and more accurate. But this is OK.

raulk (approved):

chain/vm/vm.go
iand: I'm happy to revert the formatting if you prefer

raulk: No worries.

iand:
Can't merge without a code owner approval.

raulk:
Need to wait for @magik6k or @Kubuxu.

Kubuxu (approved):

PR #5363, opened 2021-01-16
Integrate specs-actors v3

arajasek:
As in the title. Needs:

- some appeasing of tests
- upgrade + state tree migration logic (can go into `next` in a separate PR)
- careful review
- testing

Fixes #5363

### DIFF
--- a/api/test/window_post.go
+++ b/api/test/window_post.go
@@ -16,13 +16,17 @@ import (
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/crypto"
 	"github.com/filecoin-project/go-state-types/network"
 	"github.com/filecoin-project/lotus/extern/sector-storage/mock"
 	sealing "github.com/filecoin-project/lotus/extern/storage-sealing"
+	proof3 "github.com/filecoin-project/specs-actors/v3/actors/runtime/proof"
 	"github.com/filecoin-project/specs-storage/storage"
 
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build"
+	"github.com/filecoin-project/lotus/chain/actors"
+	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
 	"github.com/filecoin-project/lotus/chain/types"
 	bminer "github.com/filecoin-project/lotus/miner"
 	"github.com/filecoin-project/lotus/node/impl"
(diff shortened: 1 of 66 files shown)


chain/actors/builtin/verifreg/util.go
```diff
@@ -6,16 +6,18 @@ import (
...
 )
 
+// Assumes that the bitwidth for v3 HAMTs is the DefaultHamtBitwidth
 func getDataCap(store adt.Store, ver actors.Version, root cid.Cid, addr address.Address) (bool, abi.StoragePower, error) {
 	if addr.Protocol() != address.ID {
 		return false, big.Zero(), xerrors.Errorf("can only look up ID addresses")
 	}
 
-	vh, err := adt.AsMap(store, root, ver)
+	vh, err := adt.AsMap(store, root, ver, builtin3.DefaultHamtBitwidth)
```
Stebalien, line 20: I'm thinking this `adt.AsMap` needs to go away and we need to handle the different map types in every place where it matters.

node/impl/full/state.go
```diff
@@ -139,15 +139,10 @@ func (m *StateModule) StateMinerInfo(ctx context.Context, actor address.Address,
 		return miner.MinerInfo{}, xerrors.Errorf("failed to load miner actor state: %w", err)
 	}
 
-	// TODO: You know, this is terrible.
-	// I mean, we _really_ shouldn't do this. Maybe we should convert somewhere else?
```
Stebalien, line 143: :partying_face:

chain/stmgr/utils.go
```diff
@@ -207,17 +207,29 @@ func GetSectorsForWinningPoSt(ctx context.Context, nv network.Version, pv ffiwra
...
-		return nil, xerrors.Errorf("getting window proof type: %w", err)
-	}
-
 	mid, err := address.IDFromAddress(maddr)
 	if err != nil {
 		return nil, xerrors.Errorf("getting miner ID: %w", err)
 	}
 
-	ids, err := pv.GenerateWinningPoStSectorChallenge(ctx, wpt, abi.ActorID(mid), rand, numProvSect)
+	// TODO: move this to somewhere in specs-actors or the state types.
```
Stebalien, line 215: cc @anorth. Will the winning proof always be paired 1:1 with the window post proof? If so, we should have a helper function to convert one to the other (well, ideally, we'd just get rid of the difference...).

The tricky part is that we're using the _same_ type for winning and window proofs.

node/impl/full/state.go
arajasek: yeah this is the happiest outcome of it all

arajasek:
Devnet now runs past the upgrade epoch

magik6k (commented):
First pass, obviously some todos in the code, but generally looks good

chain/types/state.go
```diff
@@ -9,8 +9,10 @@ type StateTreeVersion uint64
 const (
 	// StateTreeVersion0 corresponds to actors < v2.
 	StateTreeVersion0 StateTreeVersion = iota
-	// StateTreeVersion1 corresponds to actors >= v2.
+	// StateTreeVersion1 corresponds to actors [v2, v3)
```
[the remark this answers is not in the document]
Stebalien: v2 up to but not including v3. But yeah, just v2 is simpler.

node/impl/full/state.go
anorth: 🙇

chain/stmgr/utils.go
anorth: As far as I know they will be paired 1:1, but I don't know the scope of possible futures. You need @porcuquine.

Winning PoSt has nothing to do with actors. The WinningPoSt proof types are declared there only for proximity to the other types, but arguably should be removed. It's unfortunate they use the same type, yes. We could probably make a change in Lotus in the near future to stop referencing the WinningPoSt constants altogether (by duplicating locally), opening up a path to removing them from there.

I think a mapping function in Lotus is the best immediate solution.

chain/actors/builtin/verifreg/util.go
Stebalien: Done.

node/modules/core.go
```diff
@@ -71,7 +71,7 @@ func MemoryConstraints() system.MemoryConstraints {
 // MemoryWatchdog starts the memory watchdog, applying the computed resource
 // constraints.
 func MemoryWatchdog(lr repo.LockedRepo, lc fx.Lifecycle, constraints system.MemoryConstraints) {
-	if os.Getenv(EnvWatchdogDisabled) == "1" {
+	if true || os.Getenv(EnvWatchdogDisabled) == "1" {
```
magik6k, line 74: Probably drop before merging

Stebalien: Yep. Although I'd like to figure out _why_ this is an issue before releasing.

api/test/window_post.go
```diff
@@ -602,3 +606,206 @@ loop:
...
+			Value:  types.NewInt(0),
+			From:   minerInfo.Owner, // TODO: new miner...
+		}
+		sm, err := client.MpoolPushMessage(ctx, msg, nil)
+		require.NoError(t, err)
+
+		fmt.Println("waiting dispute")
+		rec, err := client.StateWaitMsg(ctx, sm.Cid(), build.MessageConfidence)
+		require.NoError(t, err)
+		require.Zero(t, rec.Receipt.ExitCode, "dispute not accepted: %s", rec.Receipt.ExitCode.Error())
```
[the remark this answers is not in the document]
Stebalien: Done. I still need to extend that test to ensure that a correct proof can't be disputed.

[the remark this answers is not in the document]
Stebalien: Hm, wait not done. Now I remember why I hadn't committed that yet.

chain/stmgr/forks.go
```diff
@@ -704,6 +701,73 @@ func UpgradeCalico(ctx context.Context, sm *StateManager, cb ExecCallback, root
 	return newRoot, nil
 }
 
+func UpgradeActorsV3(ctx context.Context, sm *StateManager, cb ExecCallback, root cid.Cid, epoch abi.ChainEpoch, ts *types.TipSet) (cid.Cid, error) {
+	buf := bufbstore.NewTieredBstore(sm.cs.Blockstore(), bstore.NewTemporarySync())
```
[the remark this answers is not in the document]
Stebalien: What's the best way to test this? compute-state on the live network? Or just set a node to run the upgrade early?

magik6k (approved):

Issue #5097, opened 2020-12-02
Sector commit failed because ticket expire, the program should check ticket status before to do submit commit


qiusugang:
**Describe the problem**
Sector commit failed because ticket expire, you need check ticket status before to do submit commit, else it will be wasted lot of gasFee and punishment cost. 

Once sector is landed to chain, miner get fee from Owner, but after commit failed, it will be punished. 
**The program should check ticket status before doing each step!! if sector ticket already expired, donot waste next step time!!**

**Sectors status**
Event Log:
0.  2020-11-30 15:18:39 +0800 CST:  [event;sealing.SectorStartCC]   {"User":{"ID":51,"SectorType":8,"Pieces":[{"Piece":{"Size":34359738368,"PieceCID":{"/":"baga6ea4seaqao7s73y24kcutaosvacpdjgfe5pw76ooefnyqw4ynr3d2y6x2mpq"}},"DealInfo":null}]}}
1.  2020-11-30 15:18:44 +0800 CST:  [event;sealing.SectorPacked]    {"User":{"FillerPieces":null}}
2.  2020-11-30 15:18:45 +0800 CST:  [event;sealing.SectorTicket]    {"User":{"TicketValue":"zT4StplgovaBfqDY6oc0ORFqio0R2u130BaEJSe1XCU=","TicketEpoch":279577}}
[...]

PR #5366, opened 2021-01-17
storagefsm: Send correct event on ErrExpiredTicket in CommitFailed

magik6k:
Fixes https://github.com/filecoin-project/lotus/issues/5097


arajasek (approved):
seems reasonable

PR #5369, opened 2021-01-18
Use 55th percentile instead of median for gas-price

Kubuxu:
The aim is to put some negative pressure on gas-premium instead of
maintining status quo.

55th percentile instead of median should not make much difference for
block inclusion timing.

Still thinking about it a bit, don't merge yet.


magik6k (approved):

Kubuxu:
SGWM to merge it.

PR #5376, opened 2021-01-19
remove abstract map constructor/loader

Stebalien:
Different maps have different parameters now so we just construct/load them manually where needed.

### DIFF
--- a/chain/stmgr/stmgr.go
+++ b/chain/stmgr/stmgr.go
@@ -21,6 +21,9 @@ import (
 	// Used for genesis.
 	msig0 "github.com/filecoin-project/specs-actors/actors/builtin/multisig"
 
+	// we use the same adt for all receipts
+	blockadt "github.com/filecoin-project/specs-actors/actors/util/adt"
+
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build"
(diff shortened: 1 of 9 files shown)


chain/stmgr/stmgr.go
```diff
@@ -385,11 +388,7 @@ func (sm *StateManager) ApplyBlocks(ctx context.Context, parentEpoch abi.ChainEp
 		return cid.Cid{}, cid.Cid{}, err
 	}
 
-	// XXX: Is the height correct? Or should it be epoch-1?
-	rectarr, err := adt.NewArray(sm.cs.Store(ctx), actors.VersionForNetwork(sm.GetNtwkVersion(ctx, epoch)), ReceiptAmtBitwidth)
-	if err != nil {
-		return cid.Undef, cid.Undef, xerrors.Errorf("failed to create receipts amt: %w", err)
-	}
+	rectarr := blockadt.MakeEmptyArray(sm.cs.Store(ctx))
```
Stebalien, line 391: Directly using v0 should be fine here. If we want, we _can_ make this dependent on the network version, but I see no reason to directly tie it in with the specs-actors version.

Note: As of this patch, _blocks_ (messages, specifically) still use the v0 AMT.

arajasek (approved):

PR #5379, opened 2021-01-19
Build a WindowPoSt disputer

### LINKED ISSUE #5285: Build off-chain WindowPoSt checker

arajasek:
**Motivation**

[FIP-0010](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0010.md) necessitates the creation of "auditors" that try to disprove recent WindowPoSt submissions. We need this to be something easily launched, and so it makes sense to build it within a Lotus node.

**Design considerations**

There are 2 chief designs we could go with:

- **Design 1**: Periodically audit every miner's PoSt submissions (that are in the `OptimisticPoStSubmissionsSnapshot` of their deadlines). This is fairly comprehensive, but could be slow unless we have high efficiency (never re-audit a post). We could also modify this to be random, auditing some configurable number (or percentage) of miners in every period.

- **Design 2**: Subscribe to updates to the head, and listen for WindowPoSt submissions whose windows have closed (and are therefore challenge-able). Try to dispute every single one.

**Implementation**

This is a first draft that implements Design 1. 
**Testing**

Based on a quick simulation in devnets, it achieves the two key requirements: 
- audits every PoSt
- sends a dispute message for bad posts, ignores good ones.

Next big thing to test will be performance.

**TODOs**

- [ ] give it a spin on mainnet to establish a lower bound for performance
- [ ] some refactoring / code beautification is possible

Fixes #5285

### DIFF
--- a/chain/actors/builtin/miner/v0.go
+++ b/chain/actors/builtin/miner/v0.go
@@ -391,6 +391,11 @@ func (d *deadline0) PartitionsPoSted() (bitfield.BitField, error) {
 	return d.Deadline.PostSubmissions, nil
 }
 
+func (d *deadline0) DisputableProofCount() (uint64, error) {
+	// field doesn't exist until v3
+	return 0, nil
+}
+
 func (p *partition0) AllSectors() (bitfield.BitField, error) {
 	return p.Partition.Sectors, nil
 }
--- a/cli/disputer.go
+++ b/cli/disputer.go
@@ -1,0 +1,429 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/go-address"
+
+	"github.com/filecoin-project/lotus/chain/actors"
+
+	miner3 "github.com/filecoin-project/specs-actors/v3/actors/builtin/miner"
+
+	"github.com/filecoin-project/go-state-types/big"
+	lapi "github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/types"
+	builtin3 "github.com/filecoin-project/specs-actors/v3/actors/builtin"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/lotus/chain/store"
+	"github.com/urfave/cli/v2"
+)
+
+const Confidence = 10
+
+type minerDeadline struct {
+	miner address.Address
+	index uint64
+}
+
+var chainDisputeSetCmd = &cli.Command{
+	Name:  "disputer",
+	Usage: "interact with the window post disputer",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "max-fee",
+			Usage: "Spend up to X attoFIL per DisputeWindowedPoSt message",
+		},
+		&cli.StringFlag{
+			Name:  "from",
+			Usage: "optionally specify the account to send messages from",
+		},
+	},
+	Subcommands: []*cli.Command{
+		disputerStartCmd,
+		disputerMsgCmd,
+	},
+}
+
+var disputerMsgCmd = &cli.Command{
+	Name:      "dispute",
+	Usage:     "Send a specific DisputeWindowedPoSt message",
+	ArgsUsage: "[minerAddress index postIndex]",
+	Flags:     []cli.Flag{},
+	Action: func(cctx *cli.Context) error {
+		if cctx.NArg() != 3 {
+			fmt.Println("Usage: dispute [minerAddress index postIndex]")
+			return nil
+		}
+
+		ctx := ReqContext(cctx)
+
+		api, closer, err := GetFullNodeAPI(cctx)
+		if err != nil {
+			return err
+		}
+		defer closer()
+
+		toa, err := address.NewFromString(cctx.Args().First())
+		if err != nil {
+			return fmt.Errorf("given 'miner' address %q was invalid: %w", cctx.Args().First(), err)
+		}
+
+		deadline, err := strconv.ParseUint(cctx.Args().Get(1), 10, 64)
+		if err != nil {
+			return err
+		}
+
+		postIndex, err := strconv.ParseUint(cctx.Args().Get(2), 10, 64)
+		if err != nil {
+			return err
+		}
+
+		fromAddr, err := getSender(ctx, api, cctx.String("from"))
+		if err != nil {
+			return err
+		}
+
+		dpp, aerr := actors.SerializeParams(&miner3.DisputeWindowedPoStParams{
+			Deadline:  deadline,
+			PoStIndex: postIndex,
+		})
+
+		if aerr != nil {
+			return xerrors.Errorf("failed to serailize params: %w", aerr)
+		}
+
+		dmsg := &types.Message{
+			To:     toa,
+			From:   fromAddr,
+			Value:  big.Zero(),
+			Method: builtin3.MethodsMiner.DisputeWindowedPoSt,
+			Params: dpp,
+		}
+
+		rslt, err := api.StateCall(ctx, dmsg, types.EmptyTSK)
+		if err != nil {
+			return xerrors.Errorf("failed to simulate dispute: %w", err)
+		}
+
+		if rslt.MsgRct.ExitCode == 0 {
+			mss, err := getMaxFee(cctx.String("max-fee"))
+			if err != nil {
+				return err
+			}
+
+			sm, err := api.MpoolPushMessage(ctx, dmsg, mss)
+			if err != nil {
+				return err
+			}
+
+			fmt.Println("dispute message ", sm.Cid())
+		} else {
+			fmt.Println("dispute is unsuccessful")
+		}
+
+		return nil
+	},
+}
+
+var disputerStartCmd = &cli.Command{
+	Name:      "start",
+	Usage:     "Start the window post disputer",
+	ArgsUsage: "[minerAddress]",
+	Flags: []cli.Flag{
+		&cli.Uint64Flag{
+			Name:  "start-epoch",
+			Usage: "only start disputing PoSts after this epoch ",
+		},
+	},
+	Action: func(cctx *cli.Context) error {
+		api, closer, err := GetFullNodeAPI(cctx)
+		if err != nil {
+			return err
+		}
+		defer closer()
+
+		ctx := ReqContext(cctx)
+
+		fromAddr, err := getSender(ctx, api, cctx.String("from"))
+		if err != nil {
+			return err
+		}
+
+		mss, err := getMaxFee(cctx.String("max-fee"))
+		if err != nil {
+			return err
+		}
+
+		startEpoch := abi.ChainEpoch(0)
+		if cctx.IsSet("height") {
+			startEpoch = abi.ChainEpoch(cctx.Uint64("height"))
+		}
+
+		fmt.Println("checking sync status")
+
+		if err := SyncWait(ctx, api, false); err != nil {
+			return xerrors.Errorf("sync wait: %w", err)
+		}
+
+		fmt.Println("setting up window post disputer")
+
+		// subscribe to head changes and validate the current value
+
+		headChanges, err := api.ChainNotify(ctx)
+		if err != nil {
+			return err
+		}
+
+		head, ok := <-headChanges
+		if !ok {
+			return xerrors.Errorf("Notify stream was invalid")
+		}
+
+		if len(head) != 1 {
+			return xerrors.Errorf("Notify first entry should have been one item")
+		}
+
+		if head[0].Type != store.HCCurrent {
+			return xerrors.Errorf("expected current head on Notify stream (got %s)", head[0].Type)
+		}
+
+		lastEpoch := head[0].Val.Height()
+		lastStatusCheckEpoch := lastEpoch
+
+		// build initial deadlineMap
+
+		minerList, err := api.StateListMiners(ctx, types.EmptyTSK)
+		if err != nil {
+			return err
+		}
+
+		knownMiners := make(map[address.Address]struct{})
+		deadlineMap := make(map[abi.ChainEpoch][]minerDeadline)
+		for _, miner := range minerList {
+			dClose, dl, err := makeMinerDeadline(ctx, api, miner)
+			if err != nil {
+				return xerrors.Errorf("making deadline: %w", err)
+			}
+
+			deadlineMap[dClose+Confidence] = append(deadlineMap[dClose+Confidence], *dl)
+
+			knownMiners[miner] = struct{}{}
+		}
+
+		// when this fires, check for newly created miners, and purge any "missed" epochs from deadlineMap
+		statusCheckTicker := time.NewTicker(time.Hour)
+		defer statusCheckTicker.Stop()
+
+		fmt.Println("starting up window post disputer")
+
+		applyTsk := func(tsk types.TipSetKey) error {
+			log.Infof("last checked height: %d", lastEpoch)
+			dls, ok := deadlineMap[lastEpoch]
+			delete(deadlineMap, lastEpoch)
+			if !ok || startEpoch >= lastEpoch {
+				// no deadlines closed at this epoch - Confidence, or we haven't reached the start cutoff yet
+				return nil
+			}
+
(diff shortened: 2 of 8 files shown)


cli/cmd.go
```diff
@@ -317,6 +317,7 @@ var Commands = []*cli.Command{
 	WithCategory("developer", logCmd),
 	WithCategory("developer", waitApiCmd),
 	WithCategory("developer", fetchParamCmd),
+	WithCategory("developer", disputerCmd),
```
magik6k, line 320: Probably shouldn't be a top-level command, maybe under chain -> `lotus chain diputer ...`

cli/disputer.go
```diff
@@ -0,0 +1,359 @@
...
+
+		if head[0].Type != store.HCCurrent {
+			return xerrors.Errorf("expected current head on Notify stream (got %s)", head[0].Type)
+		}
+
+		fmt.Println("starting up window post disputer")
+
+		ticker := time.NewTicker(120 * time.Second)
+		defer ticker.Stop()
+		// TODO: Don't start unless fully synced (check syncstatus)
```
magik6k, line 193:
```suggestion
		if err := SyncWait(ctx, api, false); err != nil {
			return xerrors.Errorf("sync wait: %w", err)
		}
```

cli/disputer.go
```diff
@@ -0,0 +1,359 @@
...
+		for {
+			select {
+			case notif, ok := <-headChanges:
+				if !ok {
+					return xerrors.Errorf("head change channel errored")
+				}
+				for _, val := range notif {
+					switch val.Type {
+					case store.HCApply:
+						// TODO: Confirm that we're never notified to "apply" a null tipset
```
magik6k, line 203: Confirmed

chain/types/tipset.go
```diff
@@ -159,6 +159,10 @@ func (ts *TipSet) Blocks() []*BlockHeader {
 	return ts.blks
 }
 
+func (ts *TipSet) IsNull() bool {
```
magik6k, line 162: I'm quite sure we never build tipsets like this

cli/disputer.go
```diff
@@ -0,0 +1,359 @@
...
+
+						// TODO: If the newly applied tipset is built on top of n null tipset(s), we should also check the tipsets WPoStChallengeWindow+[1...n]  epochs ago
+
+						// Since we're waiting WPoStChallengeWindow epochs already, we don't have to worry about confidence of the message -- it's already pretty deep onchain
+						ts, err := api.ChainGetTipSetByHeight(ctx, val.Val.Height()-miner3.WPoStChallengeWindow, val.Val.Key())
+						if err != nil {
+							return xerrors.Errorf("failed to get tipset 60 epochs ago: %w", err)
+						}
+
+						if ts.IsNull() {
```
magik6k, line 218: If ChainGetTipSetByHeight lands on a nullblock it will get you the a tipset before the nullblock(s)

cli/disputer.go
```diff
@@ -0,0 +1,359 @@
...
+						}
+
+						dpmsgs, err := makeDisputeWindowedPosts(ctx, api, wpmsgs, fromAddr)
+						if err != nil {
+							return xerrors.Errorf("failed to check for disputes: %w", err)
+						}
+
+						for _, dpmsg := range dpmsgs {
+							fmt.Println("disputing a post at height ", ts.Height())
+							_, err := api.MpoolPushMessage(ctx, dpmsg, nil)
```
magik6k, line 241: Option for setting maxfee would be nice

cli/disputer.go
```diff
@@ -0,0 +1,359 @@
...
+				return nil
+			}
+		}
+	},
+}
+
+// for a list of messages and the tsk in which they were executed, returns the _successful_ SubmitWindowedPoSt msgs among them
+func getSubmitWindowedPoSts(ctx context.Context, fullApi api.FullNode, msgs []api.Message, tsk types.TipSetKey) ([]api.Message, error) {
+	posts := make([]api.Message, 0)
+	for _, msg := range msgs {
```
magik6k, line 279: Could run this in parallel (probably using `lib/parmap.Par(50, msgs, func(api.Message){...})`)

cli/disputer.go
```diff
@@ -0,0 +1,359 @@
...
+		if msg.Message.Method != builtin3.MethodsMiner.SubmitWindowedPoSt {
+			continue
+		}
+
+		// TODO: Might be worth building a cache of miner actors
```
magik6k, line 284: Maybe, but reorgs are annoying to handle correctly

cli/disputer.go
```diff
@@ -0,0 +1,359 @@
...
+			return xerrors.Errorf("expected current head on Notify stream (got %s)", head[0].Type)
+		}
+
+		fmt.Println("starting up window post disputer")
+
+		ticker := time.NewTicker(120 * time.Second)
+		defer ticker.Stop()
+		// TODO: Don't start unless fully synced (check syncstatus)
+		for {
+			select {
```
magik6k, line 195: I'd wrap this select into a function which catches errors, and just logs them instead of quitting

cli/disputer.go
```diff
@@ -0,0 +1,359 @@
...
+			To:         post.Message.To,
+			From:       sender,
+			Nonce:      0,
+			Value:      big.Zero(),
+			GasLimit:   0,
+			GasFeeCap:  abi.TokenAmount{},
+			GasPremium: abi.TokenAmount{},
+			Method:     builtin3.MethodsMiner.DisputeWindowedPoSt,
+			Params:     dpp,
+		}
```
magik6k, line 344:
```suggestion
		dispute := &types.Message{
			To:         post.Message.To,
			From:       sender,
			Value:      big.Zero(),
			Method:     builtin3.MethodsMiner.DisputeWindowedPoSt,
			Params:     dpp,
		}
```

cli/disputer.go
arajasek: eh, maybe? the cache would basically be remembering "this ID address corresponds / doesn't correspond to a miner actor at a depth of at least 60 epochs"...probably okay to just ignore reorgs?

cli/disputer.go
```diff
@@ -0,0 +1,397 @@
...
+				break
+			}
+		}
+
+		return nil
+	},
+}
+
+// for a list of messages and the tsk in which they were executed, returns the _successful_ SubmitWindowedPoSt msgs among them
+func getSubmitWindowedPoSts(ctx context.Context, api lapi.FullNode, msgs []lapi.Message, tsk types.TipSetKey) ([]lapi.Message, error) {
```
Stebalien, line 315: This won't work due to internal messages. A "good" attacker will prove from, e.g., a multisig to avoid this checker.

I believe the _only_ real way to do this is to:

1. Watch for new miners at every epoch, and track them. Maybe watch for entries in the power table?
2. For all tracked miners, track their proving period.
3. At the end of every epoch, check every miner that is "up" for proving. Unfortunately, this will check 1/60 of the miners every epoch, even if the miner has no live sectors. However, that check _should_ be pretty fast.

arajasek: @Stebalien Yeah, you're right...Reimplemented to use Design 1

chain/actors/builtin/miner/v0.go
```diff
@@ -391,6 +391,11 @@ func (d *deadline0) PartitionsPoSted() (bitfield.BitField, error) {
 	return d.Deadline.PostSubmissions, nil
 }
 
+func (d *deadline0) OptimisticPoStSubmissionsSnapshotLength() (uint64, error) {
+	// field doesn't exist until v3
```
arajasek, line 395: @Stebalien How angry does this make you?

arajasek:
Ran this on mainnet, and it seems pretty okay. Building the deadlineMap doesn't seem to take too long.

That's excluding the majority of the work, which is simulating the `Dispute` calls, though.

Stebalien (approved):
The approach looks like a good MVP. It needs some tests but nothing's blocking merging.

Note: this is going to be a pretty inefficient and very slow approach. I _think_ it'll be able to keep up with mainnet, but it's not taking advantage of any of the things we can do off-chain.

The next steps are to:

1. Parallelize this. Maybe call the gas estimator instead of actually trying to submit messages so we can try multiple in parallel?
2. Ideally, check _before_ actually trying to submit. We can play a _lot_ of optimization games if we can skip the VM.

cli/disputer.go
```diff
@@ -0,0 +1,433 @@
...
+
+		disputeLoop := func() (error, bool) {
+			select {
+			case notif, ok := <-headChanges:
+				if !ok {
+					return xerrors.Errorf("head change channel errored"), true
+				}
+
+				for _, val := range notif {
+					switch val.Type {
```
magik6k, line 231: I'd put each case in it's own func; this code is wide

Stebalien:
**NG Design**

1. Watch the chain for deadline changes, extracting optimistic proofs and the associated snapshotted partitions (specifically, compute the sectors that should have been proven).  Then send this back to a worker queue. Do all this in parallel.
2. In the worker queue, resolve the sector numbers to piece CIDs, _caching_ the results (we can probably just keep this cache in memory).
3. Verify proofs on multiple workers.
4. Later: use batch verification assuming that most of the proofs are correct.

The key part here is parallelism and caching. The parallelism will git us an NCORE speedup and the caching/parallel disk access will cut the time in ~half for large partitions.

cli/disputer.go
```diff
@@ -0,0 +1,433 @@
...
+								if err != nil {
+									return xerrors.Errorf("making deadline: %w", err), true
+								}
+
+								deadlineMap[dClose+Confidence] = append(deadlineMap[dClose+Confidence], *dl)
+							}
+
+							for _, dpmsg := range dpmsgs {
+								fmt.Println("disputing a PoSt from miner ", dpmsg.To)
+								_, err := api.MpoolPushMessage(ctx, dpmsg, mss)
```
magik6k, line 271: Could be run in parallel (but doesn't really need to be done now)

cli/disputer.go
```diff
@@ -0,0 +1,433 @@
...
+					}
+				}
+			case <-healthTicker.C:
+				fmt.Print("Running health check: ")
+
+				cctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+
+				if _, err := api.ID(cctx); err != nil {
+					cancel()
+					return xerrors.Errorf("health check failed"), true
```
magik6k, line 291: The API can auto-reconnect, no need to shutdown

magik6k: (will need to resubscribe to head changes)

cli/disputer.go
```diff
@@ -0,0 +1,433 @@
...
+						}
+
+						deadlineMap[dClose+Confidence] = append(deadlineMap[dClose+Confidence], *dl)
+
+						knownMiners[m] = struct{}{}
+					}
+				}
+
+				for ; lastStatusCheckEpoch < lastEpoch; lastStatusCheckEpoch++ {
+					// if an epoch got "skipped" from the deadlineMap somehow, just fry it now instead of letting it sit around forever
```
magik6k, line 320: Maybe log that

cli/disputer.go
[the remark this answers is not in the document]
Stebalien: Actually, I don't think this works. It'll fail with an error if we fail to dispute (returning and skipping everything else). We need to inspect the error).

Stebalien (changes requested):

cli/disputer.go
```diff
@@ -0,0 +1,433 @@
...
+	Name:  "disputer",
+	Usage: "interact with the window post disputer",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "max-fee",
+			Usage: "Spend up to X attoFIL for DisputeWindowedPoSt message(s)",
+		},
+		&cli.StringFlag{
+			Name:  "from",
+			Usage: "optionally specify the account to send funds from",
```
magik6k, line 44:
```suggestion
			Usage: "optionally specify the account to send messages from",
```

cli/disputer.go
```diff
@@ -0,0 +1,433 @@
...
+	index uint64
+}
+
+var chainDisputeSetCmd = &cli.Command{
+	Name:  "disputer",
+	Usage: "interact with the window post disputer",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "max-fee",
+			Usage: "Spend up to X attoFIL for DisputeWindowedPoSt message(s)",
```
magik6k, line 40:
```suggestion
			Usage: "Spend up to X attoFIL per DisputeWindowedPoSt message",
```

magik6k: I'd make this in FIL / parse with `types.FIL` below

cli/disputer.go
```diff
@@ -0,0 +1,433 @@
...
+		dispute := &types.Message{
+			To:     dl.miner,
+			From:   sender,
+			Value:  big.Zero(),
+			Method: builtin3.MethodsMiner.DisputeWindowedPoSt,
+			Params: dpp,
+		}
+
+		rslt, err := api.StateCall(ctx, dispute, types.EmptyTSK)
+		if err == nil && rslt.MsgRct.ExitCode == 0 {
```
magik6k, line 378: Probably should log errors

cli/disputer.go
```diff
@@ -0,0 +1,433 @@
...
+		}
+
+		rslt, err := api.StateCall(ctx, dmsg, types.EmptyTSK)
+		if err != nil {
+			return xerrors.Errorf("failed to simulate dispute: %w", err)
+		}
+
+		if rslt.MsgRct.ExitCode == 0 {
+			var mss *lapi.MessageSendSpec
+			if cctx.IsSet("max-fee") {
```
magik6k, line 116: Could use `getMaxFee()`

cli/disputer.go
```diff
@@ -0,0 +1,433 @@
...
+			log.Infof("last checked height: %d", lastEpoch)
+			dls, ok := deadlineMap[lastEpoch]
+			delete(deadlineMap, lastEpoch)
+			if !ok || startEpoch >= lastEpoch {
+				// no deadlines closed at this epoch - Confidence, or we haven't reached the start cutoff yet
+				return nil
+			}
+
+			dpmsgs := make([]*types.Message, 0)
+			for _, dl := range dls {
```
magik6k, line 241: Could be parallel

magik6k (approved):
Just a couple of nits, nothing really blocking

cli/disputer.go
arajasek: Pretty sure even "normal" failures (the proof is actually valid) will lead to `StateCall` returning an error...so that would log every `StateCall` unfortunately

magik6k: ah, yes

cli/disputer.go
```diff
@@ -0,0 +1,429 @@
...
+	index uint64
+}
+
+var chainDisputeSetCmd = &cli.Command{
+	Name:  "disputer",
+	Usage: "interact with the window post disputer",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "max-fee",
+			Usage: "Spend up to X attoFIL per DisputeWindowedPoSt message",
```
magik6k, line 40: Off by 1000000000000000000x
```suggestion
			Usage: "Spend up to X FIL per DisputeWindowedPoSt message",
```

magik6k (approved):

magik6k (approved):

jennijuju:
## Run a Disputer
After your node is fully synced, you can run `lotus chain disputer` with the following options:
`--max-fee` is the max amount in FIL that you are willing to spend for a `DisputeWindowedPoSt` message. Keep the amount `DisputeReward` in mind when you set this fee.
`--from` is the account you want to send the messages from.

> Note:  A DisputeWindowedPoSt costs ~300M in gas. 

### Start a Window PoSt Disputer
Simply run `lotus chain disputer start` to start a disputer. It will start dispute proof since the epoch that your disputer is started unless you specify a `--start-epoch`.

### Manual Dispute

You can also send a specific `DisputeWindowedPoSt` message by running ` .lotus chain disputer dispute [minerAddress] [index] [postIndex]`, where:
`minerAddress` is the miner id that submitted the proof you want to dispute. The same address is also the recipient of the message.
`index` is the deadline index of the proof you want to dispute for the miner, it should be in [0, 47].
`postIndex` is the post snapshot index, see the definition here.

Issue #3124, opened 2020-08-17
Reserving too much space


kernelogic:
My miner seems reserving too much space and it's causing errors like
ERROR	advmgr	sector-storage/sched.go:309	trySched(1) req.sel.Ok error: finding best alloc storage:
    github.com/filecoin-project/sector-storage.(*allocSelector).Ok
        /mnt/ipfs/lotus_src/extern/sector-storage/selector_alloc.go:49
  - no good path found:
    github.com/filecoin-project/sector-storage/stores.(*Index).StorageBestAlloc
        /mnt/ipfs/lotus_src/extern/sector-storage/stores/index.go:405

The output of storage list has this
8d0798d1-5bfa-418a-875f-543795485cc6:
	[##################################****************] 1.557 TiB/1.557 TiB 100%
	Unsealed: 2; Sealed: 1; Caches: 1; Reserved: 483.2 GiB
	Weight: 10; Use: Seal 
	URL: http://127.0.0.1:3456/remote (latency: 600µs)

And the sectors list has these (3 in PC1, which I think it's too many).
0: Proving    sSet: YES active: YES tktH: 5072  seedH: 6662  deals: [0]
1: Proving    sSet: YES active: YES tktH: 6381  seedH: 7972  deals: [0]
2: Proving    sSet: YES active: YES tktH: 7234  seedH: 8827  deals: [0]
3: Proving    sSet: YES active: YES tktH: 8074  seedH: 9667  deals: [0]
4: Proving    sSet: YES active: YES tktH: 8915  seedH: 10509 deals: [0]
5: Proving    sSet: YES active: YES tktH: 9756  seedH: 11345 deals: [0]
6: Committing sSet: NO  active: NO  tktH: 10595 seedH: 12193 deals: [0]
7: PreCommit1 sSet: NO  active: NO  tktH: 0     seedH: 0     deals: [5974 0 0 0 0 6004 0 0 0]
8: PreCommit1 sSet: NO  active: NO  tktH: 0     seedH: 0     deals: [6195 6215 0 0 0 0 0 0 0]
9: PreCommit1 sSet: NO  active: NO  tktH: 0     seedH: 0     deals: [6441 0 0 0 0 0 0 0 0 0 0 6513 0 0 0 0 0 0 0]

cwhiggins:
[...]

PR #5384, opened 2021-01-19
fix FileSize to return correct disk usage recursively

nonsense:
At the moment `FileSize` is not recursive and doesn't calculate correctly the disk usage of various paths.

This PR is fixing it, by recursively walking the path and adding together all file sizes.

---

Ultimately this is fixing a bug in the `stores` package, where we iterate over all storage reservations and then check how much disk space is used per sector:

```
	for id, ft := range p.reservations {
		for _, fileType := range storiface.PathTypes {
			if fileType&ft == 0 {
				continue
			}

			sp := p.sectorPath(id, fileType)
			used, err := ls.DiskUsage(sp)
```

`sp` in this case has the form of `./cache/s-t0127896-36` which is actually a directory with many `.dat` files, and not a single file.

---

TODO:
- [x] test with `lotus-miner`

### DIFF
--- a/extern/sector-storage/fsutil/filesize_unix.go
+++ b/extern/sector-storage/fsutil/filesize_unix.go
@@ -21,7 +21,10 @@ func FileSize(path string) (SizeInfo, error) {
 			return err
 		}
 		if !info.IsDir() {
-			stat := info.Sys().(*syscall.Stat_t)
+			stat, ok := info.Sys().(*syscall.Stat_t)
+			if !ok {
+				return xerrors.New("FileInfo.Sys of wrong type")
+			}
 
 			// NOTE: stat.Blocks is in 512B blocks, NOT in stat.Blksize		return SizeInfo{size}, nil
 			//  See https://www.gnu.org/software/libc/manual/html_node/Attribute-Meanings.html
--- a/extern/sector-storage/fsutil/filesize_unix.go
+++ b/extern/sector-storage/fsutil/filesize_unix.go
@@ -3,6 +3,7 @@ package fsutil
 import (
 	"os"
 	"path/filepath"
+	"syscall"
 
 	"golang.org/x/xerrors"
 )
@@ -12,6 +13,7 @@ type SizeInfo struct {
 }
 
 // FileSize returns bytes used by a file or directory on disk
+// NOTE: We care about the allocated bytes, not file or directory size
 func FileSize(path string) (SizeInfo, error) {
 	var size int64
 	err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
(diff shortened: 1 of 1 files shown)


extern/sector-storage/fsutil/filesize_unix.go
```diff
@@ -11,19 +11,24 @@ type SizeInfo struct {
...
-	var stat syscall.Stat_t
-	if err := syscall.Stat(path, &stat); err != nil {
-		if err == syscall.ENOENT {
+	var size int64
+	err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsDir() {
+			size += info.Size()
```
magik6k, line 22: .Size() doesn't return allocated bytes, which is what this method returned previously (Can be done by using `info.Sys().(syscall.Stat_t)...`). In case some of those files are preallocated, we should be returning allocated bytes, otherwise we may report too little reserved space relative to what the filesystem reports as free

nonsense:
I think this should be correct now. On my miner all the `Reserved` space is now gone as expected:

before
```
9c92d406-1e0b-4edd-93ba-1f413ca5431d:
        [####################################**************] 1.789 TiB/1.789 TiB 100%
        Unsealed: 2; Sealed: 1; Caches: 1; Reserved: 483.2 GiB
        Weight: 10; Use: Seal
        Local: /root/storage_sealing
        URL: http://127.0.0.1:2345/remote
```

after
```
9c92d406-1e0b-4edd-93ba-1f413ca5431d:
        [##########################################        ] 1.51 TiB/1.789 TiB 84%
        Unsealed: 2; Sealed: 2; Caches: 2; Reserved: 0 B
        Weight: 10; Use: Seal
        Local: /root/storage_sealing
        URL: http://127.0.0.1:2345/remote
```

extern/sector-storage/fsutil/filesize_unix.go
```diff
@@ -11,19 +12,29 @@ type SizeInfo struct {
...
-	var stat syscall.Stat_t
-	if err := syscall.Stat(path, &stat); err != nil {
-		if err == syscall.ENOENT {
+	var size int64
+	err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsDir() {
+			stat := info.Sys().(*syscall.Stat_t)
```
nonsense, line 24: Given that this is a `_unix.go` file, I think we must always have a value in `Sys()`, so I don't think we need to check for nil pointer here.

magik6k, line 24: I'd still defensively check this. Failing to check storage space is much less bad than crashing the miner process and potentially missing PoSts
```suggestion
			stat, ok := info.Sys().(*syscall.Stat_t)
			if !ok {
				return xerrors.New("FileInfo.Sys of wrong type")
			}
```

magik6k:
Closes #3124

extern/sector-storage/fsutil/filesize_unix.go
nonsense: Fair enough!

magik6k (commented):
(LGTM)

Issue #5370, opened 2021-01-18
Do not expose node builder internals to the consumer


schomatis:
This basically means not allowing (or at least disincentivizing) direct manipulation of the DI graph through `node.Override()` as we currently do:

https://github.com/filecoin-project/lotus/blob/476df99179157c9d0019d2bcc9a9a4fb6b31d876/cmd/lotus/daemon.go#L280-L305

In the previous example this would mean (among others):
* Avoiding boilerplate `node.Online()`/`node.Repo(r)` configuration calls (and their strict calling order that the consumer shouldn't know or care about) which should happen by default and just provide a `repo` argument (or functional option).
* Not needing to explicitly provide the `api.FullNode` to "extract" the created artifact of interest but having the API returning it directly.
* Replacing generic `Override` calls by explicit configuration options for the bootstrapper, genesis and other usually modified dependencies (i.e., white-listing what the consumer is allowed to change).

I'm fine with: 
* Retaining the functional options paradigm, but those options should have an intermediary abstraction, say, `BuildConfiguration` instead of providing direct access to the DI graph.
* Retaining the current `node.New()` API for internal tests in the `node` package which needs more advance configurations, but if some external consumer wants to access this low-level API it should be under the explicit assumption that it is manipulating the logic internals and that it is subject to change at any time.

The objective is to decouple the `node/` logic from the rest of the code to start cleaning it up and provide a more fine-grained but standardized interface to initialize only the subsystems needed during testing.

raulk:
I'm onboard with not leaking internals via the API, and obsoleting the need to directly manpulate the DI graph for simple configuration changes. In practice, I think we do need an escape hatch so that code that needs "privileged" access to the node construction can achieve so. Passing in an optional callback function to perform Overrides and such, should be good enough for that. An examle of such a privileged user would be testground test plans.

raulk:
> Replacing generic Override calls by explicit configuration options for the bootstrapper, genesis and other usually modified dependencies (i.e., white-listing what the consumer is allowed to change).

I'm a bit +1 on this, as long as we preserve the escape hatch I mentioned.

raulk:
> Not needing to explicitly provide the api.FullNode to "extract" the created artifact of interest but having the API returning it directly.

One nice feature of the current model is that you can extract components that are not necessarily exposed through the public API. We need to preserve that capability, either by exposing those components through getters, or by supplying an extractor callback.

schomatis:
[...]

PR #5392, opened 2021-01-20
Introduce a high-level node builder API

schomatis:
Introduce a `node.Builder` intermediary that encapsulates the `Override()` logic needed to instantiate a node through `node.New()` providing a new `BuildFullNode()` API. The most common configuration options are now part of the `Builder` logic while `BuildFullNode()` still accepts user-provided `Option`s to enable advanced configurations.

This is a compromise between decoupling the `Override()` logic from the consumer and still offer some flexibility. If anything this can be seen as just a way to reduce boilerplate code (like the ubiquitous `Online()`/`Repo()` options) letting the user specify directly the value they want to configure and not have to embed it in the corresponding `Override()` call.

Currently as a draft, the new API is applied only in the `daemon` command. Once better refined it will be propagated to other consumers outside the `node` package before converting it to a ready-for-review PR. I'm open to changes in the `Builder`/`BuildFullNode()` API; the only real objective here is to minimize `Override()` calls.

Closes https://github.com/filecoin-project/lotus/issues/5370.

### DIFF
--- a/cmd/lotus/daemon.go
+++ b/cmd/lotus/daemon.go
@@ -278,11 +278,11 @@ var DaemonCmd = &cli.Command{
 		}
 
 		builder := node.Builder{
-			IsLite: isLite,
+			IsLite:         isLite,
 			IsBootstrapper: isBootstrapper,
-			IsBootstrap: cctx.Bool("bootstrap"),
-			ShutdownChan: shutdownChan,
-			Repo: r,
+			IsBootstrap:    cctx.Bool("bootstrap"),
+			ShutdownChan:   shutdownChan,
+			Repo:           r,
 		}
 		if cctx.IsSet("api") {
 			builder.ApiAddress = "/ip4/127.0.0.1/tcp/" + cctx.String("api")
--- a/cmd/lotus/daemon.go
+++ b/cmd/lotus/daemon.go
@@ -18,7 +18,6 @@ import (
 	paramfetch "github.com/filecoin-project/go-paramfetch"
 	metricsprom "github.com/ipfs/go-metrics-prometheus"
 	"github.com/mitchellh/go-homedir"
-	"github.com/multiformats/go-multiaddr"
 	"github.com/urfave/cli/v2"
 	"go.opencensus.io/plugin/runmetrics"
 	"go.opencensus.io/stats"
@@ -36,7 +35,6 @@ import (
 	lcli "github.com/filecoin-project/lotus/cli"
 	"github.com/filecoin-project/lotus/extern/sector-storage/ffiwrapper"
 	"github.com/filecoin-project/lotus/journal"
-	"github.com/filecoin-project/lotus/lib/peermgr"
 	"github.com/filecoin-project/lotus/lib/ulimit"
 	"github.com/filecoin-project/lotus/metrics"
 	"github.com/filecoin-project/lotus/node"
@@ -174,6 +172,8 @@ var DaemonCmd = &cli.Command{
 			defer pprof.StopCPUProfile()
 		}
 
+		// FIXME: What is the difference between this profile option and
+		//  `cctx.Bool("bootstrap")`?
 		var isBootstrapper dtypes.Bootstrapper
 		switch profile := cctx.String("profile"); profile {
 		case "bootstrapper":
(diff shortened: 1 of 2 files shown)


node/builder.go
```diff
@@ -648,3 +648,84 @@ func Test() Option {
...
+		opts = append(opts,
+			Unset(RunPeerMgrKey),
+			Unset(new(*peermgr.PeerMgr)),
+		)
+	}
+
+	// FIXME: Apply user-supplied overrides at the end. Do we need to modify node.new?
+	opts = append(opts, userOptions...)
+
+	stop, err := New(ctx, opts...)
```
magik6k, line 728: Can be cleaned up a bit with more functional options
```suggestion
	stop, err := New(ctx, FullAPI(&api,
		Lite(b.IsLite)),

		// FIXME: Do we need to set these before Repo()/Online()?
		Override(new(dtypes.Bootstrapper), b.IsBootstrapper),
		Override(new(dtypes.ShutdownChan), b.ShutdownChan),

		Online(),
		If(b.Repo != nil, Repo(b.Repo)),

		If(b.ApiAddress != "", Override(SetApiEndpointKey, func(lr repo.LockedRepo) error {
			// FIXME: The string to multiaddr conversion should be done by
			//  the user who will be better equipped to handle the error.
			apima, err := multiaddr.NewMultiaddr(b.ApiAddress)
			if err != nil {
				return err
			}
			return lr.SetAPIEndpoint(apima)
		})),

		If(b.IsBootstrap,
			Unset(RunPeerMgrKey),
			Unset(new(*peermgr.PeerMgr)),
		),

		// Override(new(modules.Genesis), b.Genesis),

		// FIXME: For now process lite mode as user option. Should be derived from
		//  the profile builder configuration.
		// liteModeDeps,

		Options(userOptions...),
	)
```

magik6k: > // FIXME: Apply user-supplied overrides at the end. Do we need to modify node.new?

Nope, those are already processed at the end

node/builder.go
```diff
@@ -648,3 +648,84 @@ func Test() Option {
...
+
+type Builder struct {
+	// FIXME: General profile, maybe following node type for now, should
+	//  be decoupled later.
+	// Full, lite, miner.
+
+	// Configuration options that will affect the DI graph, applied at the end.
+	// FIXME: Consider making configuration options private and exposing functions
+	//  to set them. For now it doesn't seem worth it as they are mostly constants
+	//  that we don't process (maybe the ShutdownChan/Repo are the exception).
```
magik6k, line 664: We should try to design this well on the first try, things like this tend to be really painful to change.

I feel like functional options for this would work, but get a bit weird with multiple levels of them (we already have 2 levels, builder.Option and fx.Option, this would be adding a third level) - though maybe it can work nicely

schomatis: Will think about this some more. I think adding functional options for every setting will result on even more boilerplate. Most are constants with a sane default as `nil` value if left uninitialized. Maybe for the more complex ones.

cmd/lotus/daemon.go
```diff
@@ -174,6 +172,8 @@ var DaemonCmd = &cli.Command{
 			defer pprof.StopCPUProfile()
 		}
 
+		// FIXME: What is the difference between this profile option and
+		//  `cctx.Bool("bootstrap")`?
```
schomatis, line 176: Looking at this some more they do seem like they should be coupled. The bootstrap profile configures the PubSub peer manager (and the DHT as server) while the bootstrap CLI option turns off the the peer manager in the chain exchange service (and maybe others).

schomatis:
```
NAME:
   lotus daemon - Start a lotus daemon process

OPTIONS:
   --bootstrap               (default: true)
   --profile value           specify type of node
```

To clarify, it seems `--profile` would mean "am _I_ a bootstrap node?" while `--bootstrap` is "should I be connecting to a bootstrap node?".

schomatis: https://github.com/filecoin-project/lotus/pull/5395

node/builder.go
schomatis: @magik6k Looking for more feedback on new commit https://github.com/filecoin-project/lotus/pull/5392/commits/0b67f0e02d0f322b232ed4b4b7a125ce8602e3eb. I still prefer to avoid adding yet another layer of functional options. I think there is a list of options we can safely expose with sane default values and abstract away other more complex ones that may need some pre/post-processing (in the example commit here the `repo` configuration option). I don't strongly advocate for this way, open to iterating on it a some more.

schomatis: Moved settings to separate config struct https://github.com/filecoin-project/lotus/pull/5392/commits/bf2d40c19978e5993f3bd8261d41b363195e449a.

schomatis:
Per @magik6k, this refactoring is more complicated than it seems and it will be hard to move forward:

> this is not that simple, mainly because we'll (likely) soon be getting things like another lite mode (where you trust 51% of network miners instead of some external node, and pull the state with bitswap), and more modular lotus-miner, where you can deploy/connect its sub-services as separate processes

I still believe that making _some_ progress, even if not reaching the ideal, will be more beneficial here than leaving things as they stand. Achieving any type of proper decoupling seems to me will make any of the upcoming features mentioned above easier to implement than the current situation.

Issue #5311, opened 2021-01-08
Official Docker Image


ognots:
As part of the Lotus release process, create Docker image(s) containing all lotus binaries. 

## Guiding Principles
* Runtime environment(s) should have required packages and libraries installed using latest acceptable versions. For `lotus-miner` and `lotus-worker` a viable GPU runtime should be included, likely based on `nvidia/opencl`. Likely an Ubuntu based image.
* Should be deployment agnostic. No scripts or custom entry points or general opinions on how to run binaries should be included
* Be a building block for custom Docker images. Any customization should occur as a further layer on top of this image, using this image as the `FROM` source
* `lotus-miner` and `lotus-worker` binaries should produce native builds setting `RUSTFLAGS="-C target-cpu=native -g` and `FFI_BUILD_FROM_SOURCE=1` - blocked by https://github.com/filecoin-project/lotus/issues/3339
* images should be published to multiple sources, not just Dockerhub which has recently introduced restrictions on image pulls
* include all binaries from lotus. grouping of binaries in individual images to be determined

travisperson:
There is an issue open currently to update the nvidia/opencl container to support ubuntu 20.04 but it appears that nvidia might no longer be maintaining the containers any longer. https://gitlab.com/nvidia/container-images/opencl/-/issues/3

This might require that we add the opencl support our selves and maintain the images or find an alternate base image for opencl.

nonsense:
Now that #5394 is merged, it is very easy to extend the initial Dockerfile (https://github.com/filecoin-project/lotus/blob/master/Dockerfile.lotus pretty much copy-pasted from https://github.com/filecoin-project/lotus-infra/tree/master/docker) (currently contains only `lotus` and `lotus-shed` binaries) and add any other binaries we might need - either in the same image, or as a separate image pushed to a separate public registry.

mikeumus:
@nonsense https://github.com/filecoin-project/lotus-infra/tree/master/docker is 404 now.

Can we run Docker.lotus and be done or is that still just a step in setting up?

ianconsolata:
I believe with https://github.com/filecoin-project/lotus/pull/9625 and associated fix https://github.com/filecoin-project/lotus/pull/9826 this should be complete now.

Issue #5078, opened 2020-12-01
Canary deployment infra for Lotus nodes


nonsense:
Following conversations with @raulk , @dineshshenoy , @magik6k and @dirkmc , here is an initial set of tasks to provide developers with functionality to trigger canary deployments of Lotus nodes.

Most of the work on this issue happens at https://github.com/filecoin-project/tweety

## Add Kubernetes helm apps for Lotus nodes
- [x] Add `lotus-fullnode-minimal` helm chart, which supports: a) creating a new node from snapshot (`install-snapshot` rpc), and b) updating an existing node with a different version of lotus (i.e. using the existing state) (`upgrade` rpc)

## Tweety - Go microservice
- [x] Add `create`, `delete` RPC - these install Lotus on a given host within a Kubernetes namespace.
- [x] Add `acquire`, `release` RPC - developers use these to claim a given deployment for test purposes and release it when they are done testing and evaluating.
- [x] Add `install-snapshot` and `upgrade` RPC - developers use these to install a specific version of Lotus on a given deployment that they've claimed

## Tweety - Kubernetes resources
- [x] Kubernetes deployment of Tweety (with all relevant permissions)

## Monitoring
- [ ] Update `lotus-fullnode-minimal` to export metrics to InfluxDB / Grafana
- [ ] Update RPC to submit an annotation when a command is triggered, so that it is clear when a specific deployment was updated/reset;

## Slack / Github bot
- [ ] Add command to update a specific lotus node from the cluster with a version of lotus based on that branch;
- [ ] Add command to stop a specific lotus node, clear its datadir, update it to a specific version and start sync from snapshot;

## Logs
- [x] Developers should have access to the cluster to review logs of nodes.
- [x] Logs should be forwarded to a centralised logging system such as https://logz.io if deemed necessary.

## Ingress / domain
- [x] Tweety service should be publicly exposed so that Slack / Github can trigger its RPCs.

## Access to developers
- [x] Developers should have access to the cluster (Dev Kubernetes cluster) to inspect their nodes

## Images for deployments
- [ ] ~~We need some kind of build process for PRs / commits, which it out-of-scope for the Tweety service. For now I use my personal `nonsens3` Docker Hub, but we probably want a Github Action building a Docker image for every commit of Lotus.
[...]

PR #5394, opened 2021-01-20
add job to build Lotus docker image and push it to AWS ECR private (or public) repo

nonsense:
This PR is adding a CircleCI job for building a Lotus docker image to be used with the canary nodes deployments and other purposes.

## TODO

- [x] prior to merge, after PR approval: create `lotus-dev` AWS ECR private (or public) repo on AWS

---

## AWS IAM permissions

- [x] prior to merge, after PR approval: add `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_REGION`, `AWS_ECR_ACCOUNT_URL` env vars to CircleCI project

Policy for AWS user should be something along the lines of: https://docs.aws.amazon.com/AmazonECR/latest/userguide/ecr_managed_policies.html (full access or power user)

If public registry is used, additional policy permissions necessary:

```
                "ecr-public:*",
                "sts:*"
```

---

## Public vs private registry

At the moment the `circleci/aws-ecr` orb doesn't support public registries, maybe it is worth to not use it altogether so that we have full control over the `docker` commands and build and push as many images as we need. I have already added example steps for that.

For the time being I am building only a `lotus-dev` image to be used for canaries - specifically adding a `-dev` suffix to not confuse this image repository with production / stable / release images.

`.circleci/config.yml` - doesn't use `circleci/aws-ecr` orb and pushes to public ECR registry.

---

## Open questions
1. Public vs private registry - There is nothing private about these images, so we might as well use the AWS public ECR registry. This means getting rid of the `circleci/aws-ecr` orb at the moment, but I think that's fine.
_Resolution_: Use public registry

2. Build frequency - the idea behind this PR is to build an image for every commit in the Lotus repo, so that devs can quickly spin up canaries on-demand with the image.
_Resolution_: for now start with build for every commit, if this gets too much, we can scale down.

3. Build artifacts - at the moment I am building only `lotus-dev` image. We might need more images (`stats`, `chainwatch`, `opencl`, etc.) for other projects.
_Resolution_: for now we can bundle all binaries in the same image.

4. Consider adding Docker layer caching on CircleCI https://circleci.com/docs/2.0/docker-layer-caching/ in a future PR

## Related
1. https://github.com/filecoin-project/lotus/issues/5311
2. https://github.com/filecoin-project/lotus/issues/5078
3. https://github.com/CircleCI-Public/aws-ecr-orb/issues/128


travisperson:
1) I think public images are the right way to go
2) I'm okay building for every branch. I don't think it will be too crazy and we can monitor and adjust if we find things blowing up too much. Probably better to at least go with this for now and we can adjust if we think it becomes a problem and trigger buildings.
3) I don't think we need to different images for other binaries, we can just build them all into a single image. The images will get larger but I'm not sure if that will matter for dev images like this.

nonsense:
Thanks to @travisperson we have the values setup on AWS and CircleCI side, and I think this is now ready for merge. First image has been pushed, you can try it out with: `docker pull public.ecr.aws/p5f3y3g5/lotus-dev:790bcc78`

magik6k (approved):

PR #5395, opened 2021-01-20
fix(daemon): deprecate confusing bootstrap flag

schomatis:
Opinionated but _backwards-compatible_ introduction of a more descriptive flag name for the old `--bootstrap` option.

### DIFF
--- a/cmd/lotus/daemon.go
+++ b/cmd/lotus/daemon.go
@@ -107,9 +107,9 @@ var DaemonCmd = &cli.Command{
 		//  connect to *any* nodes at all (since you need the bootstrap nodes
 		//  to find the others).
 		&cli.BoolFlag{
-			Name:  "bootstrap",
-			Value: true,
-			Usage: "DEPRECATED: use 'isolated' flag instead",
+			Name:   "bootstrap",
+			Value:  true,
+			Usage:  "DEPRECATED: use 'isolated' flag instead",
 			Hidden: true,
 		},
 		&cli.BoolFlag{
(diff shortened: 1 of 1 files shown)


cmd/lotus/daemon.go
```diff
@@ -98,9 +98,24 @@ var DaemonCmd = &cli.Command{
...
 		&cli.BoolFlag{
-			Name:  "bootstrap",
-			Value: true,
+			Name:   "bootstrap",
+			Value:  true,
+			Usage:  "DEPRECATED: use 'isolated' flag instead",
+			Hidden: true,
+		},
+		&cli.BoolFlag{
+			Name:  "isolated",
```
magik6k, line 116: `--bootstrap=false` doesn't guarantee that the node won't connect to other nodes without explicitly asking it to, which `isolated`implies - if some other node connects to us we'll likely automatically connect to other nodes we get from pubsub/dht

magik6k (commented):
Given that this flag is used in a rather big number of places (devnet scripts), it will probably stick around for a while.

Maybe instead of deprecating it we can just add Usage string to it which clarifies what it does without adding tech debt

cmd/lotus/daemon.go
schomatis: Makes sense, what other name would be more appropriate? (Alternatively maybe we want to make it more isolated and should turn off other services besides the peer manager.)

schomatis:
> without adding tech debt

It looks to me like the tech debt is already here the moment this flag's name is very misleading. I don't intend to change this right away but offering a better alternative (this or whatever name is more appropriate) opens the way to (slowly) repaying that debt.

If we stick with the current state just because we don't want to change the status quo we are permanently sticking with the technical debt. Again, I'm not advocating for a violent API breakage here but I think things need to change eventually and for that we need to introduce the possibility of that change in the first place.

PR #5397, opened 2021-01-21
fix a flake in the sync manager edge case test

Stebalien:
Depending on scheduling, we may see c2 before we see c1.

### DIFF
--- a/chain/sync_manager_test.go
+++ b/chain/sync_manager_test.go
@@ -117,6 +117,13 @@ func TestSyncManagerEdgeCase(t *testing.T) {
 
 		// get the next sync target; it should be c1 as the heaviest tipset but added last (same weight as c2)
 		bop = <-stc
+		if bop.ts.Equals(c2) {
+			// there's a small race and we might get c2 first.
+			// But we should still end on c1.
+			bop.done()
+			bop = <-stc
+		}
+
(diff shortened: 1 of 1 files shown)


schomatis (commented):
LGTM I can't reproduce locally (after 1000's of runs) but this seems like it should fix the issue.

I'm deferring to @vyzo for the actual approval; accepting either `c2` or `c1` seems to defeat the purpose of the test altogether. I'm not sure how the `done()` API works but, as `Sync()` can in theory be run in parallel, to process tipsets in the proper order I suppose that this test is working under the assumption that `syncManager` is going to throttle them accordingly (coming from the same parent). If this is not the case then I have no idea what this test is doing and my green tick shouldn't count anyways.

Stebalien:
I _believe_ the goal is that c1 should be the final decision, even if c1 comes first. This still tests that (ish).

schomatis (approved):
Looking at this again, `c1` and `c2` do _not_ come from the same parent as I originally thought. `Sync()` calls will be issued in parallel for both, so this is in fact racy (not sure if by design). The fix makes sense as a temporary solution.

vyzo (approved):
yes, that's the intention! thanks for fixing this.

magik6k (approved):
:heart:

PR #5400, opened 2021-01-21
[WIP] use the correct context when starting services

Stebalien:
The lifecycle context survives the entire time, the start context may (well, should) be canceled after starting.


Stebalien:
This is never going to happen.

PR #5404, opened 2021-01-22
Retry proofs if we generate an incorrect one

Stebalien:
I'm proposing this change because lotus now submits window post proofs without locally _verifying_ the proof. This isn't strictly speaking necessary, but it should take very little time and help protect against random bit flips and bugs.

### DIFF
--- a/storage/wdpost_run.go
+++ b/storage/wdpost_run.go
@@ -20,6 +20,7 @@ import (
 	"golang.org/x/xerrors"
 
 	proof2 "github.com/filecoin-project/specs-actors/v2/actors/runtime/proof"
+	"github.com/filecoin-project/specs-actors/v3/actors/runtime/proof"
 
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build"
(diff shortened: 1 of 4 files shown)


storage/wdpost_run.go
```diff
@@ -594,10 +595,25 @@ func (s *WindowPoStScheduler) runPost(ctx context.Context, di dline.Info, ts *ty
...
 
+				// If we generated an incorrect proof, try again.
+				if correct, err := s.verifier.VerifyWindowPoSt(ctx, proof.WindowPoStVerifyInfo{
+					Randomness:        abi.PoStRandomness(rand),
+					Proofs:            postOut,
+					ChallengedSectors: sinfos,
+					Prover:            abi.ActorID(mid),
+				}); err != nil {
+					log.Errorw("window post verification failed", "post", postOut, "error", err)
+					continue
```
magik6k, line 611: Should have some cooldown (e.g. in case we're out of FDs and fail instantly)
```suggestion
					log.Errorw("window post verification failed", "post", postOut, "error", err)
					time.Sleep(5 * time.Second)
					continue
```

magik6k (commented):
1 Nit

magik6k (approved):

PR #5428, opened 2021-01-26
Move waiting for publish deals from markets into lotus

dirkmc:
Depends on
- https://github.com/filecoin-project/lotus/pull/5309
- https://github.com/filecoin-project/go-fil-markets/pull/475

Now that deals are published with more than one deal ID, move parsing of the publish message response into lotus rather than doing it in markets.


magik6k (approved):
(other than probably using the right markets commit)

magik6k (approved):

PR #5431, opened 2021-01-26
markets: OnDealExpiredOrSlashed - get deal by proposal instead of deal ID

dirkmc:
Also move common code for OnDealExpiredOrSlashed from storage adapter client and provider into one place.


markets/storageadapter/ondealexpired.go
```diff
@@ -0,0 +1,152 @@
...
+
+		// Timeout waiting for state change
+		if states == nil {
+			log.Error("timed out waiting for deal expiry")
+			return false, nil
+		}
+
+		changedDeals, ok := states.(state.ChangedDeals)
+		if !ok {
+			panic("Expected state.ChangedDeals")
```
magik6k, line 117: However unlikely I wouldn't panic if we don't have to

markets/storageadapter/ondealexpired.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func (mgr *DealExpiryManager) OnDealExpiredOrSlashed(ctx context.Context, publishCid cid.Cid, proposal market0.DealProposal, onDealExpired storagemarket.DealExpiredCallback, onDealSlashed storagemarket.DealSlashedCallback) error {
+	head, err := mgr.demAPI.ChainHead(ctx)
+	if err != nil {
+		return xerrors.Errorf("client: failed to get chain head: %w", err)
+	}
+
+	prop := market.DealProposal(proposal)
+	res, err := mgr.demAPI.GetCurrentDealInfo(ctx, head.Key().Bytes(), &prop, publishCid)
```
magik6k, line 71: Technically we can reorg after getting this info

dirkmc: We can, but in order to fix that we would need to change the way the state change API works.
I suggest we merge this improvement first and then tackle the state change API changes.

arajasek:
@dirkmc What's up with this PR?

dirkmc:
I think it was just waiting for approval - if we want to get this in let me know and I'll rebase

jennijuju:
@dirkmc we opened it for review - could you please resolve the conflict?

dirkmc:
@jennijuju I've resolved the conflict now, please review when ready

jennijuju:
still waiting https://github.com/filecoin-project/go-fil-markets/pull/476 and will be reviewed by the ignite team soon.

raulk (commented):
This needs two itests before merging to master:

* test that deal expiration works correctly
* test that deal slashing works correctly

I chatted with @dirkmc, and he'll take that on 🚀

raulk:
Here's a start for a test. Unfortunately the sector start and end are set too far out in the future (epoch 23k), so it takes a long while to run. What worried me is that despite the expiry epoch elapsing, the deal did not appear to be set as "expired" on the client side. I might have a bug in my test, or there's another problem here. Abstaining from merging until we get to the bottom of this.

`itests/deal_expiry_test.go`:

```
func TestDealExpiry(t *testing.T) {
	kit.QuietMiningLogs()

	// reset minimum deal duration to 0, so we can make very short-lived deals.
	// NOTE: this will need updating with every new specs-actors version.
	market2.DealMinDuration = 0
	market3.DealMinDuration = 0
	market4.DealMinDuration = 0
	market5.DealMinDuration = 0

	ctx := context.Background()

	var (
		client kit.TestFullNode
		miner1 kit.TestMiner
	)

	ens := kit.NewEnsemble(t, kit.MockProofs())
	ens.FullNode(&client)
	ens.Miner(&miner1, &client, kit.WithAllSubsystems())
	bm := ens.Start().InterconnectAll().BeginMining(50 * time.Millisecond)

	dh := kit.NewDealHarness(t, &client, &miner1, &miner1)

	client.WaitTillChain(ctx, kit.HeightAtLeast(5))

	_, _, _ = dh.MakeOnlineDeal(ctx, kit.MakeFullDealParams{
		Rseed:             0,
		FastRet:           true,
		StartEpoch:        0,
		MinBlocksDuration: 50,
	})

	go func() {
		ch, _ := client.ChainNotify(ctx)
		for range ch {
			bm[0].InjectNulls(10)
		}
	}()

	for {
		// sectors, err := miner.SectorsList(ctx)
		// require.NoError(t, err)

		minerDeals, err := miner1.DealsList(ctx)
		require.NoError(t, err)

		for _, d := range minerDeals {
			spew.Dump(d)
		}

		ts, err := client.ChainHead(ctx)
		require.NoError(t, err)

		fmt.Println(ts.Height())

		deals, err := client.ClientListDeals(ctx)
		require.NoError(t, err)

		fmt.Println(storagemarket.DealStates[deals[0].State])
		time.Sleep(1 * time.Second)

		fmt.Printf("%+v\n", deals[0])
	}
}
```

raulk (changes requested):
See comment above.

PR #5432, opened 2021-01-26
Refactor send command for better testability

Kubuxu:
Refactors `send` command to use an intermediate layer, called CLI Services, for handling all interactions with the Lotus node.
Establishes a pattern that allows for testing both the CLI input parsing as well as CLI Services layers.
This creates a logical split between CLI and CLI Services, where the primary function of the CLI command is to parse the input and present the output to the user. The CLI Services are meant to orchestrate the execution of the command.

Two interface mocks using `gomock` were generated, for `FullNode`  and `cli.Services` interfaces.
This allows testing that cli.Services behave correctly (by mocking out the FullNode), and testing that CLI works correctly (by mocking out Services).

As more commands are transitioned to `cli.Services`, writing integration tests will also become much easier, as the Services layer will become the toolbox the CLI is today.


Kubuxu:
Method params are not separated out, probably should be separate function.

cli/send_test.go
```diff
@@ -0,0 +1,128 @@
...
+}).Cid()
+
+func mustAddr(a address.Address, err error) address.Address {
+	if err != nil {
+		panic(err)
+	}
+	return a
+}
+
+func newMockCmd(t *testing.T, cmd *ucli.Command) (*ucli.App, *MockServicesAPI, *bytes.Buffer, func()) {
```
raulk, line 30:
```suggestion
func newMockApp(t *testing.T, cmd *ucli.Command) (*ucli.App, *MockServicesAPI, *bytes.Buffer, func()) {
```

raulk: (nit: to a lotus developer, this makes it sound like we're going to return a cli.Command)

raulk (approved):
This looks pretty good. Glad we were able to figure out how to test by layers by mocking out components.

@Kubuxu I'd really like to see more godocs around the key methods and structs here (e.g. GetFullNodeServices, ServicesAPI) before merge.

Let's merge it as a PoC, and use the structure laid out here to continue increasing test coverage.

---

For the record, `gomock` was chosen because of its expressive style and static code gen. While you can get codegen with `testify/mock` with [`mockery`](https://github.com/vektra/mockery), the resulting API is not as expressive. And some useful features are missing (e.g. asserting on the order of calls).

build/tools.go
```diff
@@ -3,5 +3,6 @@
 package build
 
 import (
+	_ "github.com/golang/mock/mockgen"
```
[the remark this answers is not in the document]
Kubuxu: Yeah, probably not required anymore as we import `mock` elsewhere now that we are using it but it is one of the tools run by `go:generate` so I would keep it here.

cli/services.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/filecoin-project/lotus/chain/stmgr"
+	types "github.com/filecoin-project/lotus/chain/types"
+	cid "github.com/ipfs/go-cid"
+	cbg "github.com/whyrusleeping/cbor-gen"
+	"golang.org/x/xerrors"
+)
+
+//go:generate go run github.com/golang/mock/mockgen -destination=servicesmock_test.go -package=cli -self_package github.com/filecoin-project/lotus/cli . ServicesAPI
+
+type ServicesAPI interface {
```
magik6k, line 24: Is there any way we can avoid having this additional layer? It's going to add maintenance overhead, and replicating this pattern in other places will not make adding tests easier. Does testing the command with the full API mock not work?

magik6k (commented):
Mockgen is nice, but I'm not too convinced about the ServicesAPI thing

cli/services.go
Kubuxu: > Is there any way we can avoid having this additional layer?

The point is to build up the pattern that will allow testing more complex CLI interactions.
Lotus will be getting more complex commands in near future (interactive sending, interactive deal-making as examples).
I don't think integration testing a command with interactive UI is the right way to do it.

This is the point of CLI Services, to separate the user integration and CLI parsing from actions the commands make.
This way we can clearly test the parsing or UI of the command, and we can characterise Services calls it should make.
Then we can test services themselves (if they are more complex) with mocked backend and also use the Services layer to perform very clean integration tests with real Lotus API. Using normal function calls with typed inputs and outputs, instead of serializing inputs and invoking the CLI and parsing the output.

> It's going to add maintenance overhead, and replicating this pattern in other places will not make adding tests easier.

I disagree, simple commands can be mocked out directly (`chain get head` for example) introducing this pattern does not prevent it. More complex commands will gain a lot from this pattern because of the responsibility separation allowing tests to focus on either only the parsing or behaviour.

Case in point, our hardly tested, mostly because 1. testing CLI is painful because stringifying, parsing and interacting with it; 2. no mocking

Introducing mocking is only a small step as it doesn't make writing tests for complex commands any nicer.

raulk: It's not just about testing, it's about code structure and separation of concerns.

This separation keeps the CLI layer light and well-scoped. Its job is limited to parsing of flags/args, validating them, transforming into upper types, calling something, and presenting the information back to the user. It is no longer tasked with business logic and orchestration. The CLI layer is "dumb".

There are nice effects that arise from that:
1. Testing becomes more focused. Now you can test the behaviour of the CLI separately from the behaviour of the orchestration.
2. You can choose to expose the orchestration as a new JSON-RPC API operation.

Kubuxu: Also, the `send` command is right on the edge of when I would consider adding the Services layer abstraction. I wouldn't do it for much simpler commands.

raulk: Yep, for commands that simply expose a JSON-RPC call over the CLI, you wouldn't create a services operation.

@Kubuxu how would you feel about embedding the `FullAPI` inside the full node `Services`?

magik6k (approved):

Kubuxu:
Test short failure is unrelated.
I'll merge tomorrow morning if there are no comments.

PR #5433, opened 2021-01-26
Add miner config to always keep ensealed deal copies

magik6k:
Allows miners to set
```
[Sealing]
AlwaysKeepUnsealedCopy = true
```

in lotus-miner configs, to keep copies of deals even if those deals don't have the FastRetrieval flag set. By doing this miners can avoid having to unseal sectors while still being able to provide retrieval

### DIFF
--- a/extern/storage-sealing/sealiface/config.go
+++ b/extern/storage-sealing/sealiface/config.go
@@ -15,4 +15,6 @@ type Config struct {
 	MaxSealingSectorsForDeals uint64
 
 	WaitDealsDelay time.Duration
+
+	AlwaysKeepUnsealedCopy bool
 }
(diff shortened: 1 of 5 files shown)


PatrickDeuse:
https://github.com/filecoin-project/lotus/issues/5435

a "never keep" flag, the opposite of `AlwaysKeepUnsealedCopy = true` would be nice, not sure if it needs a second flag or another one for "let the client choose" then

PatrickDeuse:
in addition: it might be worth thinking about an extra folder for these "sealed"-unsealed sectors so miners can link storage to the folder that doesn't need to be local and does not interfere with the "normal" unsealed folder

magik6k:
> in addition: it might be worth thinking about an extra folder for these "sealed"-unsealed sectors so miners can link storage to the folder that doesn't need to be local and does not interfere with the "normal" unsealed folder

Those unsealed files get moved to long term storage (--store) in the Finalize step

arajasek (approved):

magik6k:
(will add the other option in a separate PR)

PR #5434, opened 2021-01-26
Implement pre-migration framework

### LINKED ISSUE #5415: Integrate network upgrade pre-migration
The spec-actors v3 migration supports "pre-migration" and caching. That is, the migration logic can be run _before_ the network upgrade epoch. When the upgrade epoch finally arrives, actors that have not changed since the re-migration do not need to be "re-migrated". Instead, the result from the pre-migration will be reused.

**Motivation**

Network state upgrades take time as they need to re-write most of the state tree. However, there is a long tail of actors that change infrequently. We can take advantage of this by running the migration ahead of time, and re-using the results for actors that don't change.

**Design**

120 and 30 epochs (TBC) before the upgrade epoch, the migration logic will be executed (in the background) against the current state. After these "pre-migrations", lotus will record a mapping of actor addresses to `(old_state_root, new_state_root)` tuples.

When running a migration, lotus (well, spec-actors) will, for every actor:

1. Lookup the `(old_state_root, new_state_root)` tuples.
2. If the current state root for the actor matches `old_state_root`, the state root will be replaced with the `new_state_root` and the migration logic will _not_ be run again.

Stebalien:
This PR adds:

1. A PreMigrations slice to every Upgrade. This can be filled with functions that should be executed
   some number of epochs before the upgrade happens.
2. A MigrationCache that's passed to both the final Migration function, and any PreMigration functions.
3. An _optional_ Start/Stop method on the StateManager that starts/stops the pre-migration scheduler.

Finally, this PR wires up pre-migrations for the actors V3 upgrade.

fixes #5415 

TODO:

* [x] Tests.
* [ ] Consider extracting into a new service. I really don't like shoving this into the state manager along with everything else, but the state manager _is_ responsible for upgrades.
* [x] Merge https://github.com/filecoin-project/specs-actors/pull/1368

### DIFF
--- a/chain/stmgr/forks.go
+++ b/chain/stmgr/forks.go
@@ -292,7 +292,7 @@ func (sm *StateManager) preMigrationWorker(ctx context.Context) {
 						// migration to use the cache may assume that
 						// certain blocks exist, even if they don't.
 						tmpCache := cache.Clone()
-						err := migrationFunc(preCtx, sm, cache, ts.ParentState(), ts.Height(), ts)
+						err := migrationFunc(preCtx, sm, tmpCache, ts.ParentState(), ts.Height(), ts)
 						if err != nil {
 							log.Errorw("failed to run pre-migration",
 								"error", err)
--- a/chain/stmgr/forks.go
+++ b/chain/stmgr/forks.go
@@ -67,7 +67,7 @@ type PreUpgradeFunc func(
 	sm *StateManager, cache MigrationCache,
 	oldState cid.Cid,
 	height abi.ChainEpoch, ts *types.TipSet,
-)
+) error
 
 // PreUpgrade describes a pre-migration step to prepare for a network state upgrade. Pre-migrations
 // are optimizations, are not guaranteed to run, and may be canceled and/or run multiple times.
@@ -76,13 +76,11 @@ type PreUpgrade struct {
 	// expected to run asynchronously and must abort promptly when canceled.
 	PreUpgrade PreUpgradeFunc
 
-	// After specifies that this pre-migration should be started _after_ the given epoch.
-	//
-	// In case of chain reverts, the pre-migration will not be canceled and the state will not
-	// be reverted.
-	After abi.ChainEpoch
+	// When specifies that this pre-migration should be started at most When epochs before the upgrade.
+	When abi.ChainEpoch
 
-	// NotAfter specifies that this pre-migration should not be started after the given epoch.
+	// NotAfter specifies that this pre-migration should not be started NotAfter epochs before
+	// the final upgrade epoch.
 	//
 	// This should be set such that the pre-migration is likely to complete at least 5 epochs
 	// before the next pre-migration and/or upgrade epoch hits.
@@ -170,6 +168,15 @@ func DefaultUpgradeSchedule() UpgradeSchedule {
 		Height:    build.UpgradeActorsV3Height,
 		Network:   network.Version10,
 		Migration: UpgradeActorsV3,
+		PreUpgrades: []PreUpgrade{{
+			PreUpgrade: PreUpgradeActorsV3,
+			When:       120,
+			NotAfter:   60,
+		}, {
+			PreUpgrade: PreUpgradeActorsV3,
+			When:       30,
+			NotAfter:   20,
+		}},
 		Expensive: true,
 	}}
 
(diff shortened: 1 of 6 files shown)


chain/stmgr/forks.go
```diff
@@ -135,14 +192,20 @@ func DefaultUpgradeSchedule() UpgradeSchedule {
...
-	// Make sure we're not trying to upgrade to version 0.
+	// Make sure each upgrade is valid.
 	for _, u := range us {
 		if u.Network <= 0 {
 			return xerrors.Errorf("cannot upgrade to version <= 0: %d", u.Network)
 		}
+
+		for _, m := range u.PreMigrations {
+			if m.When <= m.NotAfter {
+				return xerrors.Errorf("pre-migration cannot end before it starts: %d <= %d", m.When, m.NotAfter)
```
Kubuxu, line 203: Do we want to enforce the ordering of PreMigrations in the array, as in `n.When < (n+1).When`

chain/stmgr/forks.go
```diff
@@ -180,6 +250,96 @@ func (sm *StateManager) hasExpensiveFork(ctx context.Context, height abi.ChainEp
...
+	defer cancel()
+
+	type op struct {
+		after    abi.ChainEpoch
+		notAfter abi.ChainEpoch
+		run      func(ts *types.TipSet)
+	}
+
+	var wg sync.WaitGroup
+	defer wg.Wait()
```
Kubuxu, line 282: This wg.Wait() will block if we don't run pre-migration and try to shutdown.

chain/stmgr/stmgr.go
```diff
@@ -103,16 +113,21 @@ func NewStateManagerWithUpgradeSchedule(cs *store.ChainStore, us UpgradeSchedule
...
 		// If we have any upgrades, process them and create a version
 		// schedule.
 		for _, upgrade := range us {
-			if upgrade.Migration != nil {
-				stateMigrations[upgrade.Height] = upgrade.Migration
+			if upgrade.Migration != nil || upgrade.PreMigrations != nil {
+				migration := &migration{
+					upgrade:       upgrade.Migration,
+					preMigrations: upgrade.PreMigrations,
+					cache:         nv10.NewMemMigrationCache(),
```
Kubuxu, line 128: a bit messy but I guess it is fine

Stebalien: Yeah, we should probably move it elsewhere, but this works for now. Really, it was that or copy it.

chain/stmgr/forks.go
```diff
@@ -121,6 +169,15 @@ func DefaultUpgradeSchedule() UpgradeSchedule {
...
 		Network:   network.Version10,
 		Migration: UpgradeActorsV3,
+		PreMigrations: []PreMigration{{
+			PreMigration: PreUpgradeActorsV3,
+			When:         120,
+			NotAfter:     60,
+		}, {
+			PreMigration: PreUpgradeActorsV3,
+			When:         30,
+			NotAfter:     20,
```
magik6k: Probably should get those epochs from `build/`

chain/stmgr/forks.go
```diff
@@ -702,12 +884,45 @@ func UpgradeCalico(ctx context.Context, sm *StateManager, cb ExecCallback, root
 	return newRoot, nil
 }
 
-func UpgradeActorsV3(ctx context.Context, sm *StateManager, cb ExecCallback, root cid.Cid, epoch abi.ChainEpoch, ts *types.TipSet) (cid.Cid, error) {
+func UpgradeActorsV3(ctx context.Context, sm *StateManager, cache MigrationCache, cb ExecCallback, root cid.Cid, epoch abi.ChainEpoch, ts *types.TipSet) (cid.Cid, error) {
+	// TODO: tune this.
+	config := nv10.Config{MaxWorkers: 1}
```
magik6k, line 889: Do we have good/any idea on what this should be?

chain/stmgr/forks_test.go
```diff
@@ -317,3 +318,149 @@ func TestForkRefuseCall(t *testing.T) {
...
+	})
+
+	cg.SetStateManager(sm)
+
+	for i := 0; i < 50; i++ {
+		_, err := cg.NextTipSet()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
```
magik6k, line 465: shouldn't we assert that things actually got invoked?

chain/stmgr/stmgr.go
```diff
@@ -62,15 +63,24 @@ type versionSpec struct {
 	atOrBelow      abi.ChainEpoch
 }
 
+type migration struct {
+	upgrade       MigrationFunc
+	preMigrations []PreMigration
+	cache         *nv10.MemMigrationCache
```
magik6k, line 69: Do we know how much memory this will use?

chain/stmgr/forks.go
```diff
@@ -180,6 +266,102 @@ func (sm *StateManager) hasExpensiveFork(ctx context.Context, height abi.ChainEp
...
+	sort.Slice(schedule, func(i, j int) bool {
+		return schedule[i].after < schedule[j].after
+	})
+
+	// Finally, when the head changes, see if there's anything we need to do.
+	for change := range sm.cs.SubHeadChanges(ctx) {
+		for _, head := range change {
+			for len(schedule) > 0 {
+				op := &schedule[0]
+				if head.Val.Height() < op.after {
```
magik6k, line 351: We could add some confidence here to make reverts less likely to have impact here (if they even matter)

chain/stmgr/forks.go
```diff
@@ -180,6 +266,102 @@ func (sm *StateManager) hasExpensiveFork(ctx context.Context, height abi.ChainEp
...
+	}
+
+	// Then sort by epoch.
+	sort.Slice(schedule, func(i, j int) bool {
+		return schedule[i].after < schedule[j].after
+	})
+
+	// Finally, when the head changes, see if there's anything we need to do.
+	for change := range sm.cs.SubHeadChanges(ctx) {
+		for _, head := range change {
```
magik6k, line 348: Should we also drop the migration caches here after say upgradeEpoch+finality to not keep a bunch of stuff in memory?

chain/stmgr/stmgr.go
```diff
@@ -148,6 +163,33 @@ func cidsToKey(cids []cid.Cid) string {
...
+	var ctx context.Context
+	ctx, sm.cancel = context.WithCancel(context.Background())
+	sm.shutdown = make(chan struct{})
+	go sm.preMigrationWorker(ctx)
+	return nil
+}
+
+// Stop starts the state manager's background processes.
+//
+// This is method is not safe to invoke concurrently with Start.
```
arajasek, line 180:
```suggestion
// This method is not safe to invoke concurrently with Start.
```

chain/stmgr/stmgr.go
```diff
@@ -148,6 +163,33 @@ func cidsToKey(cids []cid.Cid) string {
 	return out
 }
 
+// Start starts the state manager's optional background processes. At the moment, this schedules
+// pre-migration functions to run ahead of network upgrades.
+//
+// This is method is not safe to invoke from multiple threads or concurrently with Stop.
```
arajasek, line 169:
```suggestion
// This method is not safe to invoke from multiple threads or concurrently with Stop.
```

chain/stmgr/forks.go
```diff
@@ -36,20 +39,66 @@ import (
...
+// are optimizations, are not guaranteed to run, and may be canceled and/or run multiple times.
+type PreMigration struct {
+	// PreMigration is the pre-migration function to run at the specified time. This function is
+	// run asynchronously and must abort promptly when canceled.
+	PreMigration PreMigrationFunc
+
+	// When specifies that this pre-migration should be started at most When epochs before the upgrade.
+	When abi.ChainEpoch
+
+	// NotAfter specifies that this pre-migration should not be started NotAfter epochs before
```
arajasek, line 84: It's a bit more than that, right -- it also means that an in-process migration will be cancelled at this epoch?

arajasek (approved):

chain/stmgr/forks.go
Stebalien: Maybe? I mean, these are just saying "do the pre-migration N epochs before the upgrade height". I'm not sure if being able to configure these in the build is all that useful.

chain/stmgr/forks.go
Stebalien: I'm setting the pre-migration to half the CPUs, and the migration to all the CPUs. I'll also try at 2x CPUs just to see if it matters.

chain/stmgr/forks_test.go
Stebalien: We can't pass the upgrade height without that, but I can add an additional assertion.

chain/stmgr/stmgr.go
Stebalien: The cache? Maybe a few hundred bytes per actor. It doesn't cache any actual data, just a mapping of `<actor-address>-<old-state-root>` to `<new-state-root>`.

chain/stmgr/forks.go
Stebalien: Reverts don't matter.

chain/stmgr/forks.go
Stebalien: Hm. Yes, we should. I'll nil out the "ops" as we pass them.

Stebalien: Well, actually, it's not _quite_ that simple. I'll see if there's a better way to do this.

chain/stmgr/forks.go
Stebalien: Yeah, I'm going to to fix that. Really, we need _three_ points in time:

1. Start after.
2. Don't start after.
3. Stop at.

arajasek: Yeah, that sounds better

Stebalien: Could you take a look now?

chain/stmgr/forks.go
Stebalien: Can you take a look at the current values?

chain/stmgr/forks_test.go
Stebalien: Done.

arajasek (approved):
LGTM

PR #5457, opened 2021-01-28
feat: avoid computing the tip-set state when looking up addresses

Stebalien:


arajasek (approved):
Yeah, this should be good!

PR #5473, opened 2021-01-28
1.5.x staging branch

magik6k:
Branch for features which can land on master after the v3 actor upgrade epoch


magik6k:
(merging since 1.5.0 is now postponed, there will be 1.4.2)

PR #5476, opened 2021-01-29
fix verified registry migration

Stebalien:


Stebalien:
Note: this is incompatible with the current calibration net.

magik6k (approved):

PR #5492, opened 2021-02-01
ipfsbstore: Fix has for non-existing blocks

### LINKED ISSUE #5361: retrieval hangs when using IpfsOnlineMode=true
In the context of using Lotus with a `go-ipfs` node as the underlying blockstore, I've faced what seems to be a bug.
When running with the default `[Client] IpfsOnlineMode = false`, retrievals works fine, but with `true` they hang.

If I run with "online"=true, the retrieval hangs with the last retrieval-event being:
`retrieval event {"name": "ClientEventLastPaymentRequested", "deal ID": "0", "state": "DealStatusFundsNeededLastPayment", "message": ""}`
After that, it simply hangs.
(BTW, that "deal ID": 0 might be another thing to fix, it also prints that on successful retrievals. It isn't the correct Deal ID.)

I came up with a simple way to let anyone reproduce this.

Preparation:
- `git clone https://github.com/jsign/lotus && cd lotus && git checkout jsign/onlinefailretrieval && make deps`
(it's based on Lotus master)

Test-flow:
1. `docker run -p 5001:5001 --rm --name ipfs ipfs/go-ipfs:v0.7.0`
2. `go test ./node -race -v -run=TestAPIDealFlow/TestFastRetrievalDealFlow -count 1`
3.  Wait for the result (PASS or hangs (more on this below)
4.  After that, Ctrl+C the docker run ... window, as to clean up things.
The cloned branch will test a setup with `onlineMode := true`, so you should see the test hanging.

To check that running in `onlineMode := false` pass the test, simply change the `onlineMode :=` value in line 406 node/test/builder.go. Nothing else needs to be changed. Then just do Test-Flow steps again.

This is a somewhat modified existing test. It has minimal changes to create this scenario, and reproduce the problem. Obviously, doing git diff with Louts master can show changes.

magik6k:
Fixes #5361

Discovered this when doing something unrelated on top of this blockstore

Basically the implementation of `blockstore.Has` would call `block/stat` on the ipfs node - doing that in online mode will obviously try to fetch that node from the network. This then compounds with the default blockservice implementation, which calls `Has` before calling `Put`, which is basically guaranteed to hang in online mode when trying to retrieve data into ipfs which can't be fetched from the ipfs network (also means that data would be fetched from ipfs first before being retrieved from filecoin..)

### DIFF
--- a/lib/ipfsbstore/ipfsbstore.go
+++ b/lib/ipfsbstore/ipfsbstore.go
@@ -21,7 +21,7 @@ import (
 )
 
 type IpfsBstore struct {
-	ctx context.Context
+	ctx             context.Context
 	api, offlineAPI iface.CoreAPI
 }
 
@@ -43,8 +43,8 @@ func NewIpfsBstore(ctx context.Context, onlineMode bool) (*IpfsBstore, error) {
 	}
 
 	return &IpfsBstore{
-		ctx: ctx,
-		api: api,
+		ctx:        ctx,
+		api:        api,
 		offlineAPI: offlineAPI,
 	}, nil
 }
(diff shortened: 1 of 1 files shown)


dirkmc (approved):
🎉

PR #5500, opened 2021-02-02
refactor and simplify code

schomatis:
Not ready to be merged; it's a simplification of the original (and it's probably broken at the moment). The objective of this PR is to gather feedback to see if I understand the original correctly. Most of the simplification has been reflected in the heavily-refactored `add()` function.

Most notably:

* We pack in contiguous order. Once we create a new box we don't pack anything else in the previous ones (we don't reuse boxes). This is clearly sub-performant but for now it helps to simplify the logic (keeping track only of the last/current box in use and not of different box IDs across `add()` calls). Will reimplement that part in a follow-up patch after the current changes here are validated.

* Removed the internal links, now implicit in the encoding (if a link is not in the external array then it has been packed in the same box).

* Encapsulated raw node extraction logic into `extractIntoRawNode()`. I'm still not sure why do we need to create a new node here, would like some clarifications to add to the comments. I'm assuming that the raw node format helps to cleanly decouple the parent node from its children when separating in different boxes.


cmd/lotus-shed/dagspliter/dagsplitter.go
```diff
@@ -41,33 +42,36 @@ type Edge struct {
...
 type Box struct {
 	// Nodes with external links
 	// root always 0-th; nodes don't have to depend on root; can be empty
-	Nodes []*BoxedNode
-
-	// Subgraphs of boxed nodes which fit here
-	// if no Nodes above, root is 0-th; can be empty
-	Internal []cid.Cid
+	// FIXME: What are the >1th nodes? Children? Maybe we could decouple that.
+	Nodes []cid.Cid
```
magik6k, line 51: Nodes which have both internal and external links (if they only had one of those, those nodes would be internal/external)

cmd/lotus-shed/dagspliter/dagsplitter.go
```diff
@@ -122,104 +120,132 @@ func (b *builder) getSize(nd ipld.Node) (uint64, error) {
...
-				}
+	// Too big for the current box. We need to examine its links to see
+	// how to partition it.
+
+	// First the parent node.
+	// FIXME: How to check the size of the parent node without taking into
+	//  account the children? The Node interface doesn't seem to account for
+	//  that so we are going directly to the Block interface for now.
+	//  (If this is indeed the correct way reuse the raw data for the extract
+	//   call next.)
```
magik6k, line 204: That's a bit annoying, and something that could be better about dag-pb, but it's likely not changing any time soon.. maybe with unixfs v2 we can get node count / total raw byte size entries.

For now we can probably get away with not accounting non-file data well, and just have some % overhead when accounting space (obviously that will break horribly with small files, but it should be good enough in the average case)

cmd/lotus-shed/dagspliter/dagsplitter.go
```diff
@@ -298,3 +325,10 @@ var Cmd = &cli.Command{
 		return nil
 	},
 }
+
+// FIXME: Add real test. For now check that the output matches across refactors.
```
magik6k, line 329: A (manual) test I'm using is adding the rather large pond node_modules directory (`make lotus-pond-app; ipfs add -r lotuspond`), then chunking with 31M chunk size, and then making sure that no resulting car files are above 32M

cmd/lotus-shed/dagspliter/dagsplitter.go
```diff
@@ -41,33 +42,36 @@ type Edge struct {
...
-
-	// Subgraphs of boxed nodes which fit here
-	// if no Nodes above, root is 0-th; can be empty
-	Internal []cid.Cid
+	// FIXME: What are the >1th nodes? Children? Maybe we could decouple that.
+	Nodes []cid.Cid
 
 	// References to external subgraphs
 	// if no Nodes nor Internal above, root is 0-th; can be empty
+	// FIXME: Maybe replace by array of boxes CID.
```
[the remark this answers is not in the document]
schomatis: Going to go with that second option.

cmd/lotus-shed/dagspliter/dagsplitter.go
```diff
@@ -122,104 +120,132 @@ func (b *builder) getSize(nd ipld.Node) (uint64, error) {
...
+func (b *builder) packNodeCid(root cid.Cid, size uint64) {
+	// FIXME: Maybe assert size (`fits`).
+	b.box().nodes = append(b.box().nodes, root)
+	b.box().used += size
+}
+
+func (b *builder) extractIntoRawNode(ctx context.Context, root cid.Cid, rootNode ipld.Node) (ipld.Node, error) {
+	p := root.Prefix()
+
+	// TODO: wasn't the ipfs blockstore supposed to operate on multihashes?
```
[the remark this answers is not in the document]
schomatis: That's your own `TODO`  (but it's good to know you're consistent in your thought :grin:)

cmd/lotus-shed/dagspliter/dagsplitter.go
schomatis: That's valid but I was more trying to make sure I don't change the output at all during some parts of the refactoring (matching CIDs apart from size restriction).

cmd/lotus-shed/dagspliter/dagsplitter.go
schomatis: Removed internal links.

PR #5505, opened 2021-02-02
Dont publish expired deals

dirkmc:
@jennijuju [pointed out](https://github.com/filecoin-project/lotus/pull/5309#issuecomment-771493626) that we should not include deals in a batch publish message if the deal has already expired (the current epoch is greater than the deal proposal start epoch). This PR implements a fix.

### DIFF
--- a/markets/storageadapter/dealpublisher.go
+++ b/markets/storageadapter/dealpublisher.go
@@ -25,6 +25,7 @@ import (
 )
 
 type dealPublisherAPI interface {
+	ChainHead(context.Context) (*types.TipSet, error)
 	MpoolPushMessage(ctx context.Context, msg *types.Message, spec *api.MessageSendSpec) (*types.SignedMessage, error)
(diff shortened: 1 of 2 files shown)


markets/storageadapter/dealpublisher.go
```diff
@@ -223,32 +224,66 @@ func (p *DealPublisher) publishReady(ready []*pendingDeal) {
...
 }
 
+// validateDeal checks that the deal proposal start epoch hasn't already
+// elapsed
+func (p *DealPublisher) validateDeal(deal market2.ClientDealProposal) error {
+	head, err := p.api.ChainHead(p.ctx)
+	if err != nil {
+		return err
+	}
+	if head.Height() > deal.Proposal.StartEpoch {
```
magik6k, line 273: We could add some config here to allow setting some minimum window

magik6k (approved):

PR #5507, opened 2021-02-02
refactor coalescing logic into its own function, take both cancellation sets into account

vyzo:


vyzo:
So this is a small refactoring and an additional sanity check -- we thought there was a bug, but it was a case of Garbage In, Garbage Out.
I think we should still merge it.

Kubuxu (approved):

magik6k (approved):

PR #5512, opened 2021-02-03
dagsplitter: write car files

schomatis:
*This command writes to disk, beware.* Undocumented: the created boxes will be written as CAR files to a `dagsplitter-car-files/` directory in the `pwd`.

*Correctness is not the aim of this PR yet* (talking to you @magik6k). The objective is still to gather feedback and refine the design before focusing on correctness or performance.

With a concrete use case of writing the split DAGs (packaged in `Box`es) to CAR files the design has been reviewed and greatly simplified. The `add()` function in particular should be reviewed from scratch as an entirely new function.

Most notably:
* We only need to store *roots* of (sub-)graphs in the boxes, which is what the CAR file logic needs. From those roots we simply walk the DAG and if the link is *not* in the `External` list then we write to the current CAR file (as belonging to the same box). If it is on that list it means it will be a root of another box and hence a root of another CAR file processed later.
* We can now use the good old `merkledag.Walk()` logic in the box generation. We just pack the boxing logic in the `Visit` function where we keep track of the box size and determine if the node we are visiting fits in the current box (so do nothing and keep walking) or doesn't (add it as an external link and add it to the queue of new roots that need to be processed in subsequent `Walk`s).
* We only handle CIDs in the `Box` as this is the only thing the CAR generation logic needs. We do not wrap nodes in `RawNode`; we store their CIDs as is.

Still underperformant. We still pack in the most naive way possible and we check links iterating an array. Possibly other issues. None of it is a deal-breaker to stop progress at this point.

As discussed during design it's looking more and more like most of this logic should belong in the `go-merkledag` (and related) package/repo.


cmd/lotus-shed/dagspliter/dagsplitter.go
```diff
@@ -325,13 +390,63 @@ var Cmd = &cli.Command{
...
+	return func(nd ipld.Node) (out []*ipld.Link, err error) {
+		for _, link := range nd.Links() {
+
+			// Do not walk into nodes external to the current `box`.
+			if box.isExternal(link) {
+				//_, _ = fmt.Fprintf(os.Stderr, "Found external link, skipping from CAR generation: %s\n", link.Cid.String())
+				continue
+			}
+
+			// FIXME: Taken from the original `gen.CarWalkFunc`, what are these?
```
magik6k, line 434: Filecoin sector commitment CIDs (CommD (padded/truncated sha256 binary tree), CommR (basically magic tree))

Those are linked directly in the chain state, so this avoids trying to accidentally walk over a few exabytes of data

magik6k (approved):

cmd/lotus-shed/dagspliter/dagsplitter.go
```diff
@@ -325,13 +390,63 @@ var Cmd = &cli.Command{
...
+				// FIXME: We could just retain the created boxes but trying
+				//  to make it more decoupled (we might not have them available
+				//  by the time this is called).
+				return xerrors.Errorf("failed retrieving box: %w", err)
+			}
+
+			//_, _ = fmt.Fprintf(os.Stderr, "Creating car with roots: %v\n", box.Nodes)
+
+			out := new(bytes.Buffer)
+			if err := car.WriteCarWithWalker(context.TODO(), dag, box.Nodes, out, BoxCarWalkFunc(&box)); err != nil {
```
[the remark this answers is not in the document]
schomatis: Agreed. This was just a PoC to spit the CAR files but the end product should have this logic properly decoupled.

cmd/lotus-shed/dagspliter/dagsplitter.go
schomatis: Ok, so we leave this check. (Ideally this walk function should reuse the code from the original one to avoid duplicating these).

PR #5537, opened 2021-02-05
Update sched_worker.go

dafsic:
fixed ap, c1 and fin not to be processed in time

We want to handle all tasks that can todo in all windows, not the first window. The "break assignLoop" will return, not loop.

### DIFF
--- a/extern/sector-storage/sched_worker.go
+++ b/extern/sector-storage/sched_worker.go
@@ -341,9 +341,7 @@ func (sw *schedWorker) processAssignedWindows() {
 	worker := sw.worker
 
 	// process windows in order
-	widx := 0
-	for len(worker.activeWindows) > widx {
-		firstWindow := worker.activeWindows[widx]
+	for widx, window := range worker.activeWindows {
 
 		// process tasks within a window, preferring tasks at lower indexes
 		tidx := 0
--- a/extern/sector-storage/sched_worker.go
+++ b/extern/sector-storage/sched_worker.go
@@ -340,14 +340,15 @@ func (sw *schedWorker) workerCompactWindows() {
 func (sw *schedWorker) processAssignedWindows() {
 	worker := sw.worker
 
-assignLoop:
 	// process windows in order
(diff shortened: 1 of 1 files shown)


extern/sector-storage/sched_worker.go
```diff
@@ -340,14 +340,15 @@ func (sw *schedWorker) workerCompactWindows() {
 func (sw *schedWorker) processAssignedWindows() {
 	worker := sw.worker
 
-assignLoop:
 	// process windows in order
-	for len(worker.activeWindows) > 0 {
-		firstWindow := worker.activeWindows[0]
+	widx := 0
+	for len(worker.activeWindows) > widx {
+		firstWindow := worker.activeWindows[widx]
```
magik6k, line 346: I'd rename `firstWindow` to just `window`, and just use a foreach loop

```suggestion
	for widx, window := range worker.activeWindows {
```

magik6k (changes requested):
Thanks for the PR!

This breaks one of the main functions of the windowed scheduling, which is not starting smaller tasks before bigger tasks - the issue is that, for example, a constant stream of PC1 tasks can block PC2 tasks for a very long time.

What I'd do:
* Make the Order map in https://github.com/filecoin-project/lotus/blob/master/extern/sector-storage/sealtasks/task.go#L19 public
  * (also while there make Commit1 have -1 priority)
* `if widx > 0`, only assign tasks with negative priority (tasks which are very short)

extern/sector-storage/sched_worker.go
dafsic: Sorry for not watching this for a long time.

This still has a problem, when the first activeWindow has a P2 task and the second activeWindow has a C1 task. But the resources not enough to do P2, but enough to do C1. This code makes C1 wait until P2 is complete.

https://github.com/filecoin-project/lotus/blob/97207df705fc475ead184c14f90c161c8bca5c64/extern/sector-storage/sched_worker.go#L362-L365

Issue #5491, opened 2021-02-01
Add CLI commands for deal publishing to see pending deals / force send a publish message


dirkmc:
### Background

https://github.com/filecoin-project/lotus/pull/5309 implements publishing of multiple deals per publish deals message.
When a deal is submitted to be published, lotus will wait for a configurable amount of time for other deals to include in the publish message.

### Proposal
Add a command to see the list of deals that are pending publish, and a command to force publishing of the pending deals.

```
# lists deals that are in the queue to be published
./lotus-miner storage-deals pending-publish

# send a publish message for all deals that are currently in the queue
./lotus-miner storage-deals pending-publish --publish-now
```

PR #5538, opened 2021-02-05
add commands to list pending deals and force publish

dirkmc:
Fixes https://github.com/filecoin-project/lotus/issues/5491


api/apistruct/struct.go
```diff
@@ -301,6 +301,8 @@ type StorageMinerStruct struct {
 		MarketDataTransferUpdates func(ctx context.Context) (<-chan api.DataTransferChannel, error)                                                                                                            `perm:"write"`
 		MarketRestartDataTransfer func(ctx context.Context, transferID datatransfer.TransferID, otherPeer peer.ID, isInitiator bool) error                                                                     `perm:"read"`
 		MarketCancelDataTransfer  func(ctx context.Context, transferID datatransfer.TransferID, otherPeer peer.ID, isInitiator bool) error                                                                     `perm:"read"`
+		MarketPendingDeals        func(ctx context.Context) (api.PendingDealInfo, error)                                                                                                                       `perm:"admin"`
```
arajasek, line 304: does this need `admin` perm?

arajasek (approved):
Seems wholly reasonable, just one non-blocking q.

api/apistruct/struct.go
magik6k: Not really

magik6k:
(tested on my miner; it works!)

PR #5539, opened 2021-02-05
dagsplitter: bfs

schomatis:
Not tested at all. Incorporates new BFS Walk from https://github.com/ipfs/go-merkledag/pull/63 that needs to be reviewed.

Usage:
```
./lotus-shed dagsplit --breadth-first QmRLzQZ5efau2kJLfZRm9Guo1DxiBp3xCAVf6EuPCqKdsB 1M
```

(The min subgraph size is also a command option now.)


schomatis:
Definitely broken, the first CAR is above max size even if the Box isn't.

schomatis:
`WriteCarWithWalker` also uses `dag.Walk()`, maybe that the BFS options should be forwarded there as well.

Stebalien:
(unlikely to be reviewed at this point)

PR #5540, opened 2021-02-05
Remove unnecessary database reads in validation check

austinabell:
Is there any possible way that this genesis will ever be nil and it would otherwise load from store?


magik6k (approved):
Great catch! Reading through the code I don't think there's any case where this can be nil

PR #5554, opened 2021-02-08
fix `build-lotus-soup` ci job

nonsense:
Currently `build-lotus-soup` job is failing on CI, due to missing step to fetch latest lotus dependency from master.

### DIFF
--- a/testplans/lotus-soup/testkit/role_miner.go
+++ b/testplans/lotus-soup/testkit/role_miner.go
@@ -165,7 +165,7 @@ func PrepareMiner(t *TestEnvironment) (*LotusMiner, error) {
 			return nil, err
 		}
 
-		ds, err := lr.Datastore("/metadata")
+		ds, err := lr.Datastore(context.Background(), "/metadata")
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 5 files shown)


testplans/lotus-soup/testkit/role_miner.go
```diff
@@ -165,7 +165,7 @@ func PrepareMiner(t *TestEnvironment) (*LotusMiner, error) {
 			return nil, err
 		}
 
-		ds, err := lr.Datastore(context.Background(), "/metadata")
```
magik6k, line 168: It looks like this isn't using lotus from master - https://github.com/filecoin-project/lotus/blob/master/node/repo/interface.go#L58

nonsense: That's right, sorry about this.

The problem was that `build-lotus-soup` was missing a step to `go get github.com/filecoin-project/lotus@master` and this was indeed building an outdated commit from master.

I've added the step, so now on every CI run, we build against Lotus from master, which is what we want.

magik6k (approved):

PR #5571, opened 2021-02-10
lotus-seed command to generate lotus block

coryschwartz:
Currently devnet deployment requires spinning up a full daemon.
This would move the daemon `lotus-make-genesis` logic to lotus-seed
so it can be more easily executed in scripts, etc.


cmd/lotus-seed/genesis.go
```diff
@@ -9,6 +9,11 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/filecoin-project/lotus/chain/vm"
+	"github.com/filecoin-project/lotus/extern/sector-storage/ffiwrapper"
+	"github.com/filecoin-project/lotus/journal"
+	"github.com/filecoin-project/lotus/lib/blockstore"
+	"github.com/filecoin-project/lotus/node/modules/testing"
```
coryschwartz, line 16: This is the same module used in the lotus daemon CLI currently. Maybe this should be moved away from a "testing" module.

magik6k (approved):

PR #5575, opened 2021-02-11
An faster database Proof of Concept

hannahhoward:
This PR swaps out Badger as the backing store for the chain with github.com/hannahhoward/go-storethehash, an append only key value store based on @vmx https://github.com/vmx/storethehash. It appears in initial testing to sync the chain about 30-40% faster than badger, without introducing splitstore.

How I tested:
- import a chain snapshot
- wait for it to get to specific block height on message sync
- time the time it takes to validate 60 tipsets (importantly, the same 60 tipsets — this felt important, which is why i always reused a snapshot)

| Store      | Min Time To Validate 60 Blocks | Max Time To Validate 60 Blocks |
| ----------- | ----------- | ------------|
| Badger        | 4:59      | 5:19     |
| go-storethehash   | 3:35        | 4:04 |

Note this is running on my local machine which is below the minimum recommended system requirements for a node (has 16GB ram)

- Interestingly, by default go-storethehash was slower for the snapshot import (I suspect cause it produces a ton of disk reads for inserting into the index)
- I tried various different improvements, including memory-mapping the index. None of them made a huge difference on sync speed, though memory-mapping vastly improved the snapshot import so it was about 50% faster than badger. This PR includes the memory-mapped version, though it probably needs work to be production ready
- I believe this will perform even better with splitstore in place
- go-storethehash while functional is certainly not production ready in the sense of actually merging this -- but hopefully is a useful proof of concept if we want to move forward with a new database


ZenGround0:
Since this is a PoC and not an active area of work I am moving this to draft.

PR #5609, opened 2021-02-16
Verify Voucher locks in VoucherValidUnlocked

Kubuxu:


paychmgr/manager.go
```diff
@@ -344,6 +344,7 @@ func (pm *Manager) trackInboundChannel(ctx context.Context, ch address.Address)
 	return pm.store.TrackChannel(stateCi)
 }
 
+// TODO: secret vs proof doesn't make sense, there is only one, not two
```
magik6k, line 347: I believe we had both at some point in the past

magik6k (approved):
Looks good assuming it doesn't break anything, lint is unhappy though

Kubuxu:
Lint is not happy due to there being extra function in tests. In future, it might be used.

Kubuxu:
@jennijuju can you take a look at getting this merged. I thought it was merged ages ago.

paychmgr/manager.go
Stebalien: We did. But it wasn't usable so we removed it from actors.

paychmgr/paych.go
```diff
@@ -180,6 +180,20 @@ func (ca *channelAccessor) checkVoucherValidUnlocked(ctx context.Context, ch add
 		return nil, xerrors.Errorf("voucher ChannelAddr doesn't match channel address, got %s, expected %s", sv.ChannelAddr, ch)
 	}
 
+	// check voucher is unlocked
```
Stebalien, line 183: There are two different senses of unlocked here. `checkVoucherValidUnlocked` means "check the voucher is valid without taking the lock because we're already holding it".

Stebalien (changes requested):

Kubuxu:
@Stebalien are you still requesting changes here or did we clear it up?

Stebalien (approved):
We'll need a bit of followup work, but this is a strict improvement.

Stebalien:
@Kubuxu please followup with a patch to test this.

PR #5375, opened 2021-01-18
storagefsm: Rewrite input handling

magik6k:
Closes #5356

The main goal here is to enable ingesting deal data in parallel - the old storagefsm interface had a global lock which meant that only one deal could be added to a sector at a time. It also wasn't very robust and could create corrupted sectors if the miner process was restarted at a wrong time (this PR doesn't fully solve this issue, but it makes it much better. Full fix will come in a followup PR and will require changing some worker APIs)

With this PR up to configured `Sealing.MaxWaitDealsSectors` deals can be processed in parallel (before it was only possible to process one deal in parallel per lotus-miner)

Status
- [x] CI Tests
- [x] Test on mainnet miner
- [ ] Review

(Probably should land soon after v1.5.0 as it's a rather big refactor, so putting it in a mandatory release may not be a good idea)

PR #5612, opened 2021-02-16
storagefsm: Cleanup CC sector creation

### LINKED ISSUE #3896: MaxSealingSectors setting seemingly ignored
I've set `MaxSealingSectors = 4` (because I've queued up too many PCs and it OOMs my node). However, even like this, lotus starts all of the 7 PCs queued up. Is there a way to make it only do 3-4 concurrently without removing some of them and leaving the gap in the sector list?

magik6k:
On top of #5375

With this calling `lotus-miner sectors pledge` in a loop with very short interval should respect the `MaxSealingSectors` config.

Fixes #3896

### DIFF
--- a/api/api_storage.go
+++ b/api/api_storage.go
@@ -36,7 +36,7 @@ type StorageMiner interface {
 	MiningBase(context.Context) (*types.TipSet, error)
 
 	// Temp api for testing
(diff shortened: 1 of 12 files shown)


arajasek (approved):
Much nicer!

PR #5620, opened 2021-02-17
miner: Config to disable owner/worker address fallback

magik6k:
TODO:
- [x] Test

### DIFF
--- a/node/config/def.go
+++ b/node/config/def.go
@@ -91,6 +91,15 @@ type MinerFeeConfig struct {
 type MinerAddressConfig struct {
 	PreCommitControl []string
 	CommitControl    []string
+	TerminateControl []string
+
+	// DisableOwnerFallback disables usage of the owner address for messages
+	// sent automatically
+	DisableOwnerFallback bool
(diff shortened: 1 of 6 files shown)


node/config/def.go
```diff
@@ -91,6 +91,15 @@ type MinerFeeConfig struct {
 type MinerAddressConfig struct {
 	PreCommitControl []string
 	CommitControl    []string
+	TerminateControl []string
+
+	// DisableOwnerFallback disables usage of the owner address for messages
+	// sent automatically
+	DisableOwnerFallback bool
+	// DisableWorkerFallback disables usage of the worker address for messages
+	// sent automatically if control addresses are configured, even if those
```
arajasek, line 100: // DisableWorkerFallback disables usage of the worker address for messages
// sent automatically, **if** control addresses are configured.
// A control address that doesn't have enough funds will still be chosen over the worker address if this flag is set.

arajasek: (found the comment confusing)

arajasek (approved):
LGTM

PR #5622, opened 2021-02-18
fix drop sector

yaohcn:


magik6k (approved):
Thanks for catching this!

PR #5608, opened 2021-02-16
Revert "feat: deals - show data transfer %"

dirkmc:
This reverts commit b6c9ddccfff30a47b3cb1d29f1d99b7a29dd01b4
PR: https://github.com/filecoin-project/lotus/pull/5553

Adding the `RawBlockSize` field to the `DataRef` created a backwards-incompatibile unmarshalling issue: https://github.com/filecoin-project/lotus/issues/5601

PR #5629, opened 2021-02-19
update markets with pricing fix, and improve docs

whyrusleeping:
pulling in https://github.com/filecoin-project/go-fil-markets/pull/491 for less nonsensical deal pricing

### DIFF
--- a/cli/client.go
+++ b/cli/client.go
@@ -299,8 +299,20 @@ var clientLocalCmd = &cli.Command{
 }
 
 var clientDealCmd = &cli.Command{
-	Name:      "deal",
-	Usage:     "Initialize storage deal with a miner",
+	Name:  "deal",
+	Usage: "Initialize storage deal with a miner",
+	Description: `Make a deal with a miner.
+
+dataCid comes from running 'lotus client import'.
+
+miner is the address of the miner you wish to make a deal with.
+
+price is measured in FIL/GB/Block. Miners usually don't accept a bid
+lower than their advertised ask. You can check a miners listed price
+with 'lotus client query-ask <miner address>'.
+
+duration is how long the miner should store the data for, in blocks.
+The minimum value is 518400 (6 months).`,
 	ArgsUsage: "[dataCid miner price duration]",
 	Flags: []cli.Flag{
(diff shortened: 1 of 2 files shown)


cli/client.go
```diff
@@ -299,8 +299,20 @@ var clientLocalCmd = &cli.Command{
...
-	Usage:     "Initialize storage deal with a miner",
+	Name:  "deal",
+	Usage: "Initialize storage deal with a miner",
+	Description: `Make a deal with a miner.
+
+dataCid comes from running 'lotus client import'.
+
+miner is the address of the miner you wish to make a deal with.
+
+price is measured in FIL/GB/Block. Miners usually don't accept a bid
```
dirkmc, line 310:
```suggestion
price is measured in FIL/GB/Epoch. Miners usually don't accept a bid
```

dirkmc (approved):
LGTM 👍 

The current version of markets v1.1.8 caused issues when lotus clients on a previous version of markets tried to propose a deal with lotus miners on v1.1.8: https://github.com/filecoin-project/lotus/pull/5608

https://github.com/whyrusleeping/cbor-gen/pull/50 should fix the issue. @whyrusleeping could you review it please?

magik6k (commented):
This is API breaking, at the very least we should change the field name, maybe adding fallback to current behavior. Keeping it this way can cause people to make deals for way more FIL than they intended to

(interactive mode price math on [L815](https://github.com/filecoin-project/lotus/pull/5629/files#diff-2f58bf379b16a8226af249f73e0c214fc9b15bbe9af18d03ca84c734fa76e1e1R815) needs fixing)

whyrusleeping:
@magik6k i can't imagine anyone actually uses an interface this broken for anything. The worst case is that someone pays 32 times more than they were expecting for a 32GB deal. I dont even think anyone has successfully made a 32gb deal using this interface.

One easy thing we can do is to put a little warning if the users price is above the miners listed ask price by 2x, and force them to add a flag if they really want to pay that much

arajasek:
Let's extract out the helptext from this, but drop the (now-irrelevant) markets update

PR #5666, opened 2021-02-23
Command to extend sector expiration

### LINKED ISSUE #5011: Expose a CMD/API to extend sector lifetime
This is already implemented on the spec-actor, we just need to expose it in Lotus to call `ExtendSectorExpiration`.

`lotus-miner sectors extend-sector <sector id> <term>`

magik6k:
Resolves #5011

### DIFF
--- a/cmd/lotus-storage-miner/sectors.go
+++ b/cmd/lotus-storage-miner/sectors.go
@@ -13,10 +13,13 @@ import (
 	"github.com/urfave/cli/v2"
 	"golang.org/x/xerrors"
 
+	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
+	miner0 "github.com/filecoin-project/specs-actors/actors/builtin/miner"
 
 	"github.com/filecoin-project/lotus/api"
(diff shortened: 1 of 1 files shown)


magik6k:
(tested manually on my miner)

cmd/lotus-storage-miner/sectors.go
```diff
@@ -403,6 +407,100 @@ var sectorsRefsCmd = &cli.Command{
...
+		}
+
+		params := &miner0.ExtendSectorExpirationParams{}
+		for l, numbers := range sectors {
+
+			params.Extensions = append(params.Extensions, miner0.ExpirationExtension{
+				Deadline:      l.Deadline,
+				Partition:     l.Partition,
+				Sectors:       bitfield.NewFromSet(numbers),
+				NewExpiration: abi.ChainEpoch(cctx.Int64("new-expiration")),
```
arajasek, line 472: Would be nice to sanity-check this param (positive / greater than all the current expiration epochs / whatever actors code expects)

arajasek (approved):
Approved with an optional suggestion

cmd/lotus-storage-miner/sectors.go
magik6k: Pretty sure that MpoolPush will fail with a nice error message when that happens

PR #5669, opened 2021-02-23
Improve error message with importing a chain

ghoneycutt:
Without this commit the error message does not provide much value as it
lacks context.

### DIFF
--- a/cmd/lotus/daemon.go
+++ b/cmd/lotus/daemon.go
@@ -400,7 +400,7 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool) 
 		defer resp.Body.Close() //nolint:errcheck
 
 		if resp.StatusCode != http.StatusOK {
-			return xerrors.Errorf("non-200 response: %d", resp.StatusCode)
+			return xerrors.Errorf("fetching chain CAR failed with non-200 response: %d", resp.StatusCode)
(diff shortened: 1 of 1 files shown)


cmd/lotus/daemon.go
```diff
@@ -400,7 +400,7 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
 		defer resp.Body.Close() //nolint:errcheck
 
 		if resp.StatusCode != http.StatusOK {
-			return xerrors.Errorf("non-200 response: %d", resp.StatusCode)
+			return xerrors.Errorf("Import chain failed with non-200 response: %d", resp.StatusCode)
```
Kubuxu, line 403:
```suggestion
			return xerrors.Errorf("fetching chain CAR failed: %d", resp.StatusCode)
```

ghoneycutt: Keeping the reason (non-200 response) would help users diagnose the problem.

ghoneycutt:
@Kubuxu I have made the change you requested.

Kubuxu (approved):

magik6k (approved):

magik6k:
(merging into #5679)

ghoneycutt:
Thank you!

PR #5672, opened 2021-02-24
Implement scaffolding for sentinel integration

frrist:
### What
- add sentinel module scaffolding and API
  - allows lotus to be started in sentinel mode and start a watch that logs
tipsets as they are received. closes https://github.com/filecoin-project/sentinel/issues/179
  - Define Sentinel API and embed it in FullNodeAPI.
- add ldflag to build lotus in "SentinelMode" with make command `lotus-sentinel`.

### TODO
- [x] Figure out how to make docs-check green.
  - currently fails `panic: no perms for method: WatchStart` but WatchStart has perms admin

### DIFF
--- a/api/api_sentinel.go
+++ b/api/api_sentinel.go
@@ -7,6 +7,6 @@ import (
 type Sentinel interface {
 	// MethodGroup: Sentinel
 
-	// WatchStart start a watch against the chain
+	// SentinelWatchStart start a watch against the chain
 	SentinelWatchStart(context.Context) error
 }
--- a/api/api_sentinel.go
+++ b/api/api_sentinel.go
@@ -1,0 +1,12 @@
+package api
+
+import (
+	"context"
+)
+
+type Sentinel interface {
+	// MethodGroup: Sentinel
+
+	// WatchStart start a watch against the chain
+	WatchStart(context.Context) error
+}
--- a/api/api_sentinel.go
+++ b/api/api_sentinel.go
@@ -8,5 +8,5 @@ type Sentinel interface {
 	// MethodGroup: Sentinel
 
 	// WatchStart start a watch against the chain
-	WatchStart(context.Context) error
+	SentinelWatchStart(context.Context) error
 }
--- a/api/apistruct/struct.go
+++ b/api/apistruct/struct.go
@@ -1861,7 +1861,7 @@ func (c *WalletStruct) WalletDelete(ctx context.Context, addr address.Address) e
 	return c.Internal.WalletDelete(ctx, addr)
 }
 
-func (s *SentinelStruct) WatchStart(ctx context.Context) error {
+func (s *SentinelStruct) SentinelWatchStart(ctx context.Context) error {
 	return s.Internal.WatchStart(ctx)
 }
 
--- a/api/apistruct/struct.go
+++ b/api/apistruct/struct.go
@@ -473,7 +473,7 @@ type WalletStruct struct {
 
 type SentinelStruct struct {
 	Internal struct {
-		WatchStart func(ctx context.Context) error `perm:"admin"`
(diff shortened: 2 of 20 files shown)


api/apistruct/struct.go
```diff
@@ -79,6 +79,7 @@ type CommonStruct struct {
 // FullNodeStruct implements API passing calls to user-provided function values.
 type FullNodeStruct struct {
 	CommonStruct
+	SentinelStruct
```
magik6k, line 82: Can avoid the sentinel.SentinelUnavailable thing in node builder with
```suggestion
	SentinelStruct `optional:"true"`
```

api/api_sentinel.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"context"
+)
+
+type Sentinel interface {
+	// MethodGroup: Sentinel
+
+	// WatchStart start a watch against the chain
+	WatchStart(context.Context) error
```
magik6k, line 11: If the method group is `Sentinel`, the method should be prefixed with `Sentinel`

api/apistruct/struct.go
```diff
@@ -470,6 +471,12 @@ type WalletStruct struct {
 	}
 }
 
+type SentinelStruct struct {
```
magik6k, line 474: I think it might be easier to have the sentinel methods be defined on the FullNodeStruct struct, so it doesn't have to be registered in all the random places

magik6k (commented):
Looks pretty good, just a couple of comments/discussion points.

Docsgen is unhappy, probably because it expects flat interfaces (`make docsgen` / likely broken somewhere here - https://github.com/filecoin-project/lotus/blob/master/api/docgen/docgen.go#L411-L504)

build/sentinel.go
```diff
@@ -0,0 +1,3 @@
+package build
+
+var SentinelMode string = "false"
```
iand, line 3: Setting and parsing this seems really clunky. Is this the pattern already in use by Lotus? If so then let's leave it (but it's still clunky!)

In theory the Go compiler could optimize away code that isn't needed for sentinel mode but I think this would have to be a const bool for that to work.

I feel that fx must have better support for compile time switching of modules.

iand (commented):
Looks good and not too intrusive. 

But I'd like us to explore how we could make it even less intrusive, mainly to hedge our bets while we continue to explore possibilities for less coupling between sentinel and Lotus.

What would the changes look like if we had a separate binary to run that imported all the necessary parts of Lotus daemon? Could we inject the module from that code instead of adding it to the various Lotus packages?

api/apistruct/struct.go
frrist: While I agree it would be easier, I believe separation will make future work simpler to land as changes will only occur in sentinel-related structures and files. I think this is a bit cleaner too. But, if you feel strongly about defining methods on the FullNodeStruct, we can do it that way instead.

build/sentinel.go
frrist: The only prior art I saw for this was passing the version via a build flag. @magik6k wants to avoid adding more flags to the `daemon` and suggested a build flag or a go plugin. If, as you've already suggested, I can pull the required bits into a separate binary, then I think this flag can be removed.

hsanjuan: Unless separate binary, using [build tags](https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags) might be cleaner: `go build -tags sentinel`. It also forces to separate everything sentinel-specific to its own files.

frrist: I've implemented a standalone binary here: https://github.com/filecoin-project/lotus/pull/5693

frrist:
going with https://github.com/filecoin-project/lotus/pull/5693

PR #5685, opened 2021-02-25
V1 API staging branch

magik6k:
Not meant to be merged right now, just a place to start working on API changes with a backwards compatibility wrapper.

(the first commit renames the CreateBackup endpoint to demonstrate this working, second one reverts that to keep things clean)

Smaller change wishlist:
* [ ] Return tipset keys from ChainHead - https://github.com/filecoin-project/lotus/issues/5944
* [x] Explicit 'allow replaced' param in `StateWait/SearchMsg` - https://github.com/filecoin-project/lotus/pull/5970
* [x] Remove StateGetReceipt - https://github.com/filecoin-project/lotus/pull/5970


api/v0/api.go
```diff
@@ -0,0 +1,558 @@
+package v0
+
+import (
+	"context"
+	datatransfer "github.com/filecoin-project/go-data-transfer"
+	"github.com/filecoin-project/go-state-types/network"
+	"github.com/filecoin-project/lotus/api" // todo: alias types from here
```
magik6k, line 7: It may also make more sense to move types here, then alias from newer versions, kind of like we do in actors - then the newer versions only need to declare new/changing types, and it's harder to break older apis by changing things in new apis

The drawback is that we can't easily disable building of those wrapper layers, potentially making binaries a bit bigger than ideal, but I don't think that's a big issue

cmd/lotus/rpc.go
```diff
@@ -38,15 +39,23 @@ func serveRPC(a api.FullNode, stop node.StopFunc, addr multiaddr.Multiaddr, shut
...
+		}
+
+		http.Handle(path, ah)
 	}
 
-	http.Handle("/rpc/v0", ah)
+	pma := apistruct.PermissionedFullAPI(metrics.MetricedFullAPI(a))
+
+	serveRpc("/rpc/v1", pma)
+	serveRpc("/rpc/v0", &v0.WrapperV1{Common: pma, Internal: pma})
```
magik6k, line 58: Is it worth having an `/rpc/latest` (or expose the latest api on `/rpc` too?)

api/apistruct/types.go
```diff
@@ -0,0 +1,12 @@
...
+import "github.com/filecoin-project/lotus/api"
+
+type ChainIO = api.ChainIO
+type Common = api.Common
+type FullNode = api.FullNode
+type Gateway = api.Gateway
+type Signable = api.Signable
+type StorageMiner = api.StorageMiner
+type Wallet = api.Wallet
+type Worker = api.Worker
```
magik6k, line 12: This is here only for the generated type checks at the bottom of struct.go

I'm not sure why `apistruct` is a separate package, maybe it's worth merging it with `api`

magik6k:
Some open questions:
* Should we move `api/api_*.go` to `api/v1api`, then alias v1 things in the `api` package (I'd say yes)?
* Should we keep types in the first version they were used in (actors style) or the newest (I think doing it actors-style should be cleaner, but makes it harder to drop support for older versions)
* Should we merge apistruct into the main interface package (can't recall what was the reason to have it be a separate package)

vyzo (commented):
so we are moving the old api to `api/v0api` and keeping v1 flat in `api`?

magik6k:
> so we are moving the old api to `api/v0api` and keeping v1 flat in `api`?

For now yes, although this is one of the things to figure out what's the best approach. We could keep types and v1 interface in `api/v1api`, then alias to that from `api` - and I think that this would be a bit cleaner - I'll open a PR on top of this one to see how that would look

magik6k:
Ok, tried the thing with moving things to `api/v1api` and aliasing top-level - it's not really clean, and results in a massive diff (even compared to this one)

arajasek (approved):
Re-approved

arajasek (approved):

magik6k:
:ship:

PR #5693, opened 2021-02-26
Implement Event observer and Settings for 3rd party dep injection

frrist:
### What
- Implement TipSetObserver interface for observing tipset events (apply/revert)
- Implement Settings opts for 3rd party dep injection (allows Visor to instantiate a Lotus node)

### DIFF
--- a/node/builder.go
+++ b/node/builder.go
@@ -729,3 +729,19 @@ func Test() Option {
 		Override(new(*storageadapter.DealPublisher), storageadapter.NewDealPublisher(nil, storageadapter.PublishMsgConfig{})),
 	)
 }
+
+// For 3rd party dep injection.
+
+func WithRepoType(repoType repo.RepoType) func(s *Settings) error {
+	return func(s *Settings) error {
+		s.nodeType = repoType
+		return nil
+	}
+}
+
+func WithInvokesKey(i invoke, resApi interface{}) func(s *Settings) error {
+	return func(s *Settings) error {
+		s.invokes[i] = fx.Populate(resApi)
+		return nil
+	}
+}
--- a/chain/events/events.go
+++ b/chain/events/events.go
@@ -55,11 +55,11 @@ type Events struct {
 
 	heightEvents
 	*hcEvents
+
+	observers []TipSetObserver
 }
 
-func NewEvents(ctx context.Context, api eventAPI) *Events {
-	gcConfidence := 2 * build.ForkLengthThreshold
-
+func NewEventsWithConfidence(ctx context.Context, api eventAPI, gcConfidence abi.ChainEpoch) *Events {
 	tsc := newTSCache(gcConfidence, api)
 
 	e := &Events{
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -225,6 +225,12 @@ lotus-keygen:
 .PHONY: lotus-keygen
 BINS+=lotus-keygen
 
+lotus-sentinel: $(BUILD_DEPS)
+	rm -f lotus-sentinel
+	go build -o lotus-sentinel ./cmd/lotus-sentinel
+.PHONY: lotus-keygen
```
magik6k, line 231:
```suggestion
.PHONY: lotus-sentinel
```

cmd/lotus-sentinel/daemon.go
```diff
@@ -0,0 +1,252 @@
...
+		if err != nil {
+			return xerrors.Errorf("getting api endpoint: %w", err)
+		}
+
+		// TODO: properly parse api endpoint (or make it a URL)
+		return serveRPC(api, stop, endpoint, shutdownChan, int64(cctx.Int("api-max-req-size")))
+	},
+}
+
+func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool) (err error) {
```
magik6k, line 143: I'd try to deduplicate this with the 'normal' daemon

node/builder.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/chain/vm"
 	"github.com/filecoin-project/lotus/chain/wallet"
+	api2 "github.com/filecoin-project/lotus/cmd/lotus-sentinel/api"
```
magik6k, line 17:
```suggestion
	sentinelapi "github.com/filecoin-project/lotus/cmd/lotus-sentinel/api"
```

magik6k (commented):
First pass - looks pretty good

(I'm assuming you want a review from @iand before merging)

cmd/lotus-sentinel/daemon.go
frrist: Any preference for where this should live? In order for other packages to use it, we'll have to move it out of this file since it's in `package main` -- my first thought was the repo package.

iand: Would prefer not to have to make the user switch between binaries for this task. Let's find a new home for this function in a suitable lotus package.

iand (commented):
Pretty happy with how this turned out. Now the question is whether the binary needs to live in the lotus repo or can it live in sentinel's? The extension points and exposure of things like ImportChain obviously have to stay in Lotus but once we have those it seems by moving the binary elsewhere we could avoid Lotus depending on the all the visor and postgres packages. Plus we get to decouple the development processes of the two projects.

frrist:
I'm going to extract the lotus-sentinel binary code to the visor repo and modify this PR only to contain the changes required to make that work -- I suspect they will be minimal (sorry for the churn)

frrist:
Converting to a draft as this will require more changes than I anticipated. The other half lives here: https://github.com/filecoin-project/sentinel-visor/pull/408

magik6k (approved):

PR #5484, opened 2021-01-29
refactor blockstores

raulk:
- [X] Place all blockstores into a single flattened package. Subpackages are only created for blockstores that depend on third-party components (e.g. badger).
- [X] Make the `blockstore.Blockstore` interface the union of the basic IPFS blockstore + Viewer, for cleanness and to remove type assertions. This interface will be enhanced soon with a `Sync()` method: 5057.
- [X] Introduce a `blockstore.Adapt(BasicBlockstore)` method to adapt a basic IPFS blockstore into a Lotus blockstore.
- [x] Rename blockstores for consistency.
- [ ] ~Add godocs to all blockstores.~ Still TODO.

---

Once we merge this, we'll be in a better position to add support for the `Sync()` method (#5057), and to continue extending the blockstore with specialised operations for better efficiency.

---

Partly addresses #5455.

PR #5695, opened 2021-02-28
segregate chain and state blockstores

raulk:
Builds on #5484.

---

This paves the way for better object lifetime management.

Concretely, it makes it possible to:

- have different stores backing chain and state data.
- having the same datastore library, but using different parameters.
- attach different caching layers/policies to each class of data, e.g. sizing caches differently.
- specifying different retention policies for chain and state data.

This separation is important because:

- access patterns/frequency of chain and state data are different.
- state is derivable from chain, so one could never expunge the chain store, and only retain state objects reachable from the last finality in the state store.

---

I originally submitted this in #4771, but we @vyzo and I decided to do away with the caching for now. That was in the context of LMDB experiments (which uses the page cache aggressively). It may make sense to reintroduce Freecache on top of the hot badger store.


vyzo (approved):

Kubuxu (approved):
SGWM

chain/store/store.go
```diff
@@ -139,30 +139,30 @@ type ChainStore struct {
 	wg       sync.WaitGroup
 }
 
-// localbs is guaranteed to fail Get* if requested block isn't stored locally
-func NewChainStore(bs bstore.Blockstore, localbs bstore.Blockstore, ds dstore.Batching, vmcalls vm.SyscallBuilder, j journal.Journal) *ChainStore {
-	mmCache, _ := lru.NewARC(DefaultMsgMetaCacheSize)
-	tsCache, _ := lru.NewARC(DefaultTipSetCacheSize)
+// chainLocalBlockstore is guaranteed to fail Get* if requested block isn't stored locally
```
magik6k, line 142:
```suggestion
```

chain/store/store.go
```diff
@@ -1519,13 +1503,17 @@ func (cs *ChainStore) WalkSnapshot(ctx context.Context, ts *types.TipSet, inclRe
 
 		if b.Height == 0 || b.Height > ts.Height()-inclRecentRoots {
 			if walked.Visit(b.ParentStateRoot) {
-				cids, err := recurseLinks(cs.bs, walked, b.ParentStateRoot, []cid.Cid{b.ParentStateRoot})
+				cids, err := recurseLinks(cs.chainBlockstore, walked, b.ParentStateRoot, []cid.Cid{b.ParentStateRoot})
```
magik6k, line 1506: Not stateBlockstore?

chain/store/store.go
```diff
@@ -1444,8 +1428,8 @@ func (cs *ChainStore) Export(ctx context.Context, ts *types.TipSet, inclRecentRo
 		return xerrors.Errorf("failed to write car header: %s", err)
 	}
 
-	return cs.WalkSnapshot(ctx, ts, inclRecentRoots, skipOldMsgs, func(c cid.Cid) error {
-		blk, err := cs.bs.Get(c)
+	return cs.WalkSnapshot(ctx, ts, inclRecentRoots, skipOldMsgs, true, func(c cid.Cid) error {
+		blk, err := cs.chainBlockstore.Get(c)
```
magik6k, line 1432: This can also be getting cids for the stateBlockstore, will this be an issue?

chain/sync.go
```diff
@@ -1172,7 +1172,7 @@ func (syncer *Syncer) checkBlockMessages(ctx context.Context, b *types.FullBlock
 	}
 
 	// Finally, flush.
-	return vm.Copy(ctx, tmpbs, syncer.store.Blockstore(), mrcid)
+	return vm.Copy(ctx, tmpbs, syncer.store.ChainBlockstore(), mrcid)
```
magik6k, line 1175: This is also flushing the state (pretty sure)

chain/stmgr/stmgr.go
```diff
@@ -430,7 +430,8 @@ func (sm *StateManager) ApplyBlocks(ctx context.Context, parentEpoch abi.ChainEp
 		return cid.Cid{}, cid.Cid{}, err
 	}
 
-	rectarr := blockadt.MakeEmptyArray(sm.cs.Store(ctx))
+	// XXX: Is the height correct? Or should it be epoch-1?
+	rectarr := blockadt.MakeEmptyArray(sm.cs.ActorStore(ctx))
```
magik6k, line 434: ?

chain/store/store.go
raulk: Not right now, but we need some union abstraction for these cases.

chain/stmgr/stmgr.go
raulk: The comment looks like a merge error.

chain/store/store.go
raulk: I implemented a union blockstore in 2047a74. This concern should be solved now.

blockstore/union.go
```diff
@@ -0,0 +1,108 @@
...
+				outCh <- cid
+			}
+		}
+	}()
+
+	return outCh, nil
+}
+
+func (m unionBlockstore) HashOnRead(enabled bool) {
+	panic("implement me")
```
magik6k, line 107: Missed method

chain/store/store.go
```diff
@@ -1561,7 +1549,7 @@ func (cs *ChainStore) WalkSnapshot(ctx context.Context, ts *types.TipSet, inclRe
 }
 
 func (cs *ChainStore) Import(r io.Reader) (*types.TipSet, error) {
-	header, err := car.LoadCar(cs.Blockstore(), r)
+	header, err := car.LoadCar(cs.StateBlockstore(), r)
```
magik6k, line 1552: I'd at least leave some comment here noting that when chain/state blockstores split, this should be taken care of

chain/sync.go
```diff
@@ -1172,7 +1172,7 @@ func (syncer *Syncer) checkBlockMessages(ctx context.Context, b *types.FullBlock
 	}
 
 	// Finally, flush.
-	return vm.Copy(ctx, tmpbs, syncer.store.Blockstore(), mrcid)
```
magik6k, line 1175: Pretty sure this should be the chain blockstore

magik6k: (and my comment [here](https://github.com/filecoin-project/lotus/pull/5695#discussion_r584992695) was wrong, I just looked at `vm.Copy` and thought this must be touching the state)

travisperson:
Is there a migration path for existing datastores to splitstore?

vyzo:
It should happen automagically; on first run with the splistore, it will warm the hotstore by fetching all reachable objects from the old blockstore, which becomes the coldstore.

The directory structure looks like this:
```
.lotus
  datastore
    chain -- the old blockstore, now coldstore.
    splitstore
      hot.badger -- the hot blockstore, badger.
      tracker.bolt -- the tracking db, which is a BoltDB file.
```

chain/sync.go
raulk: I knew I was right the first time :-)

raulk:
@magik6k ready again.

magik6k (approved):

PR #5699, opened 2021-03-01
[EXPERIMENTAL] lmdb support for splitstore

raulk:
This PR introduces experimental LMDB support to the splitstore.


vyzo (approved):
given that I actually wrote the code, I say LGTM!

raulk:
We decided against LMDB because it introduces many edge cases, and the performance gain for the dataset sizes we're dealing with is lower than expected. Therefore, the tradeoff is not justified.

PR #5700, opened 2021-03-01
Add simple command for benching StateCompute

Kubuxu:


rjan90:
@Kubuxu should we keep this PR/land it, or should we close it now?

PR #5703, opened 2021-03-02
add DeleteMany to Blockstore interface

vyzo:


vyzo:
seems I missed the idstore, fixing.

raulk (approved):

blockstore/blockstore.go
```diff
@@ -18,25 +18,40 @@ var ErrNotFound = blockstore.ErrNotFound
...
 // blockstore.
 func WrapIDStore(bstore blockstore.Blockstore) Blockstore {
-	return blockstore.NewIdStore(bstore).(Blockstore)
+	if is, ok := bstore.(*idstore); ok {
+		// already wrapped
+		return is
+	}
+
+	if bs, ok := bstore.(Blockstore); ok {
+		// we need to wrap our own becase we don't want to neuter the DeleteMany method
```
raulk, line 45:
```suggestion
		// we need to wrap our own because we don't want to neuter the DeleteMany method
		// the underlying blockstore has implemented an (efficient) DeleteMany
```

blockstore/blockstore.go
```diff
@@ -18,25 +18,40 @@ var ErrNotFound = blockstore.ErrNotFound
...
+		return NewIDStore(bs)
+	}
+
+	return NewIDStore(Adapt(bstore))
```
raulk, line 49:
```suggestion
    // The underlying blockstore does not implement DeleteMany, so we need to shim it.
    // This is less efficient as it'll iterate and perform single deletes.
	return NewIDStore(Adapt(bstore))
```

blockstore/blockstore.go
```diff
@@ -18,25 +18,40 @@ var ErrNotFound = blockstore.ErrNotFound
...
 }
 
 // FromDatastore creates a new blockstore backed by the given datastore.
 func FromDatastore(dstore ds.Batching) Blockstore {
-	return WrapIDStore(blockstore.NewBlockstore(dstore))
+	return WrapIDStore(Adapt(blockstore.NewBlockstore(dstore)))
```
raulk, line 54: The `Adapt` here is not necessary, and it will lead to double adapting.

vyzo: ah yes, indeed.

vyzo: fixed.

raulk (approved):

magik6k (commented):
Late LGTM

PR #5704, opened 2021-03-02
Log block CID in the large delay warning

arajasek:


chain/sub/incoming.go
```diff
@@ -101,7 +101,7 @@ func HandleIncomingBlocks(ctx context.Context, bsub *pubsub.Subscription, s *cha
 					[]tag.Mutator{tag.Insert(metrics.MinerID, blk.Header.Miner.String())},
 					metrics.BlockDelay.M(delay),
 				)
-				log.Warnf("Received block with large delay %d from miner %s", delay, blk.Header.Miner)
+				log.Warnf("Received block %s with large delay %d from miner %s", blk.Cid(), delay, blk.Header.Miner)
```
raulk, line 104: Can we take this opportunity to switch this line to a structured log?

```suggestion
				log.Warnw("received block with large delay from miner", "block", blk.Cid(), "delay", delay, "miner", blk.Header.Miner)
```

magik6k (approved):

PR #5709, opened 2021-03-03
perf: add cache for gas permium estimation

Kubuxu:


magik6k (commented):
Doesn't build in tests
```
node/impl/full/gas_test.go:15:62: cannot use []gasMeta literal (type []gasMeta) as type []GasMeta in argument to medianGasPremium
```

magik6k (approved):

PR #5718, opened 2021-03-04
fix(multisig): The format of the amount is not correct in msigLockApp…

Rennbon:
I discovered this when I ported the multisig-wallet to Venus


magik6k (approved):
Thanks for catching this

PR #5721, opened 2021-03-04
digitalocean pass test

coryschwartz:
In order to be put on the digitalocean marketplace, digitalocean has requirements.

In particular, this script needs to pass.

https://raw.githubusercontent.com/digitalocean/marketplace-partners/master/scripts/img_check.sh. This PR will put the digitalocean snapshot into compliance so it can be marketplace-ready.


coryschwartz:
I gave a snapshot created by this PR to @johndmulhausen to use for the digitalocean marketplace application.

tools/packer/repo/config.toml
```diff
@@ -0,0 +1,36 @@
+[API]
```
coryschwartz, line 1: The reason this file was added rather than sticking to the default config is so I could enable the UFW firewall with predictable libp2p port numbers.

coryschwartz:
If not earlier, I have a meeting with digitalocean Thursday, but this PR solves a couple of things they would see as problems, particularly, the apt upgrade and the firewall.

tools/packer/lotus.pkr.hcl
```diff
@@ -58,7 +58,7 @@ source "digitalocean" "lotus" {
 
 build {
   sources = [
-    "source.amazon-ebs.lotus",
+    /* "source.amazon-ebs.lotus", */
```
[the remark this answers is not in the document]
coryschwartz: You're right, thanks. Uncommenting!

tools/packer/setup.sh
```diff
@@ -55,3 +58,11 @@ done
 # Enable services
 systemctl daemon-reload
 systemctl enable lotus-daemon
+
+# Setup firewall
+ufw enable
+ufw default deny incoming
+ufw default allow outgoing
+ufw allow ssh
+ufw allow 1234   #lotus api
```
magik6k, line 67: Making the api open to the internet by default (or at all) is a pretty bad idea

tools/packer/etc/motd
```diff
@@ -52,6 +52,13 @@ You only need to do this once, after which, you can enable and start the miner.
 
 
 
+Do you want to access your lotus daemon remotely? Learn how to setup token authentication
+and use client libraries from lotus docs.
+
+https://docs.filecoin.io/build/lotus/enable-remote-api-access/
```
magik6k, line 58: Note: we should add some warnings about the risks of exposing lotus API to the internet to this doc

tools/packer/setup.sh
coryschwartz: Thanks, I'll remove this rule.

magik6k (approved):

PR #5729, opened 2021-03-05
Feat/api no dep build

whyrusleeping:
This removes the api packages dependency on the build package, which transitively removes its dependency on the chain/actors/policy package which depends on every single version of the specs actors, which is a fairly bad time for people who just want to use the lotus api.

further extractions coming as my quest to have a very lightweight api package continues.


Stebalien (approved):
This has a slight breaking change by renaming Version to APIVersion, but avoiding that is a bit tricky unless we want to call what's now called Version something like BasicVersion.

magik6k (approved):

PR #5743, opened 2021-03-08
Default the AlwaysKeepUnsealedCopy flag to true

dirkmc:


Stebalien:
I guess this is an OK stop-gap but something is fishy:

1. The retrieval cost should be set such that it covers unsealing, if required.
2. Clients should be specifying that they want fast retrieval in their deals.

Given that we don't have 1 and this is therefore hurting miners more than it helps, we might as well enable this by default.

Stebalien (approved):

arajasek (approved):
Yeah, I continue to be uncomfortable with relying on defaults to massage things into the shape we want, but this is fine.

PR #5744, opened 2021-03-08
garbage collect hotstore after compaction

vyzo:
so that we reclaim space, otherwise the hotstore keeps balooning despite the deletions.

### DIFF
--- a/blockstore/badger/blockstore.go
+++ b/blockstore/badger/blockstore.go
@@ -131,6 +131,15 @@ func (b *Blockstore) Close() error {
 	return b.DB.Close()
 }
 
+// GC runs garbage collection on the value log
+func (b *Blockstore) GC() error {
+	if atomic.LoadInt64(&b.state) != stateOpen {
+		return ErrBlockstoreClosed
+	}
+
+	return b.DB.RunValueLogGC(0.125)
+}
+
 // View implements blockstore.Viewer, which leverages zero-copy read-only
 // access to values.
 func (b *Blockstore) View(cid cid.Cid, fn func([]byte) error) error {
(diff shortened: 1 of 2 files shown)


blockstore/badger/blockstore.go
```diff
@@ -131,6 +131,15 @@ func (b *Blockstore) Close() error {
 	return b.DB.Close()
 }
 
+// GC runs garbage collection on the value log
+func (b *Blockstore) GC() error {
```
magik6k, line 135: I'd call this CollectGarbage to match the go-datastore method

blockstore/badger/blockstore.go
```diff
@@ -131,6 +131,15 @@ func (b *Blockstore) Close() error {
...
+	if atomic.LoadInt64(&b.state) != stateOpen {
+		return ErrBlockstoreClosed
+	}
+
+	return b.DB.RunValueLogGC(0.125)
```
magik6k, line 140: We need to ignore the `badger.ErrNoRewrite` error (see https://github.com/ipfs/go-ds-badger/blob/master/datastore.go#L420)

blockstore/badger/blockstore.go
vyzo: ok, will do.

blockstore/badger/blockstore.go
vyzo: good point, will do.

blockstore/badger/blockstore.go

blockstore/badger/blockstore.go

magik6k (approved):

PR #5750, opened 2021-03-09
test: attempt to make the splitstore test deterministic

Stebalien:
At a minimum, make it thread-safe.

Unfortunately, this patch doesn't actually fix it. The test is still flaky and I can't seem to find the correct sleeps to fix that.


vyzo:
maybe we can query the `compacting` atomic?

vyzo:
yeah, that fixes it and makes the test a lot faster.

vyzo (approved):

magik6k (approved):

PR #5755, opened 2021-03-09
Metadata datastore log

magik6k:
This makes us write 'continuous' backups to `/.lotus[miner]/kvlog/metadata`, which means that when the metadata leveldb gets corrupted, it's still very easy to recover the latest state


arajasek (approved):
Nice...I assume it doesn't slow anything down, right?

magik6k:
It doesn't have measurable impact, the metadata datastore isn't written all that much

lib/backupds/datastore.go
```diff
@@ -146,11 +176,24 @@ func (d *Datastore) Sync(prefix datastore.Key) error {
...
+
+	return nil
+}
+
+func (d *Datastore) Close() error {
+	if err := d.child.Close(); err != nil {
+		return xerrors.Errorf("closing child datastore: %w", err)
+	}
+
+	return d.CloseLog()
```
vyzo, line 196: so here, things can go bad if we fail to close the `child`, because we will fail to close the log.
How about doing both in parallel and multierror combining the results?

lib/backupds/log.go
```diff
@@ -0,0 +1,221 @@
...
+					log.Errorw("failed to sync log", "error", err)
+				}
+			case <-d.closing:
+				if err := l.Close(); err != nil {
+					log.Errorw("failed to close log", "error", err)
+				}
+				return
+			}
+		}
+	}()
```
vyzo, line 89: nit: I would put this in a named method.

vyzo (commented):
looks ok, but is there are replay procedure to restore from the log?

PR #5756, opened 2021-03-09
Fix bootstrapper profile setting

magik6k:
node.Online()->LibP2P would override it again, setting it to false - https://github.com/filecoin-project/lotus/blob/master/node/builder.go#L194


vyzo (approved):
ouch!

PR #5759, opened 2021-03-10
Check liveness of sectors when processing termination batches

arajasek:
`lzusaga` somehow ran into this problem on Slack: https://filecoinproject.slack.com/archives/CPFTWMY7N/p1615315583105100.

I'm not entirely sure how this happened, since the liveness check [here](https://github.com/filecoin-project/lotus/blob/32885e1129178582fbf75711b249f5b8d99663c3/extern/storage-sealing/states_proving.go#L48-L65) should be preventing putting such sectors into the termination queue. Maybe that liveness check isn't 100% correct? I'm not sure.

But, regardless, re-checking liveness and removing out any not-live sectors should be a fine thing to do?

### DIFF
--- a/extern/storage-sealing/terminate_batch.go
+++ b/extern/storage-sealing/terminate_batch.go
@@ -143,6 +143,18 @@ func (b *TerminateBatcher) processBatch(notif, after bool) (*cid.Cid, error) {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


magik6k (approved):
This makes sense; It's likely that some sectors can get terminated while waiting for the batch to get filled/submitted

PR #5761, opened 2021-03-10
Fix TestUnpadReader on Go 1.16

magik6k:
UnpadReader can't deal with small reads right now, and `ioutil.ReadAll` in Go 1.16 starts with some tiny reads


Kubuxu (approved):

Kubuxu:
It is a hack, maybe add a comment on it.

PR #5764, opened 2021-03-10
fix: wait a bit before starting to compute window post proofs

magik6k:
Due to a (now consensus critical) bug in randomness selection, "beacon" randomness depends on whether the epoch in which the beacon occurred was a null block. Unfortunately, this means:

1. If a challenge lands on a non-null round but that non-null round gets re-orged to a null round, the randomness will change to the beacon in the first preceding non-null block.
2. If a challenge lands on a null round but that null round gets re-orged to a non-null round, the randomness will change to the beacon in the new non-null round.

The correct solution is to change the way the beacon is selected:

1. Wait until a non-null round.
2. The block in this round will contain all beacons from the null rounds, select the beacon from the target epoch.

Unfortunately, this requires a mandatory network upgrade.

Instead, this patch just waits a bit before beginning to compute window post after passing the challenge round to wait-out any reorgs.

### DIFF
--- a/storage/wdpost_run.go
+++ b/storage/wdpost_run.go
@@ -600,9 +600,18 @@ func (s *WindowPoStScheduler) runPost(ctx context.Context, di dline.Info, ts *ty
 					return nil, xerrors.Errorf("received no proofs back from generate window post")
 				}
 
+				checkRand, err := s.api.ChainGetRandomnessFromBeacon(ctx, types.EmptyTSK, crypto.DomainSeparationTag_WindowedPoStChallengeSeed, di.Challenge, buf.Bytes())
+				if err != nil {
+					return nil, xerrors.Errorf("failed to get chain randomness from beacon for window post (ts=%d; deadline=%d): %w", ts.Height(), di, err)
+				}
+
+				if !bytes.Equal(checkRand, rand) {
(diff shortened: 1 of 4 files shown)


Kubuxu (approved):

storage/wdpost_run.go
```diff
@@ -600,9 +600,23 @@ func (s *WindowPoStScheduler) runPost(ctx context.Context, di dline.Info, ts *ty
...
+					return nil, xerrors.Errorf("getting current head: %w", err)
+				}
+
+				checkRand, err := s.api.ChainGetRandomnessFromBeacon(ctx, headTs.Key(), crypto.DomainSeparationTag_WindowedPoStChallengeSeed, di.Challenge, buf.Bytes())
+				if err != nil {
+					return nil, xerrors.Errorf("failed to get chain randomness from beacon for window post (ts=%d; deadline=%d): %w", ts.Height(), di, err)
+				}
+
+				if !bytes.Equal(checkRand, rand) {
+					log.Warnw("windowpost randomness changed", "old", rand, "new", checkRand, "ts-height", ts.Height(), "challenge-height", di.Challenge, "tsk", ts.Key())
```
Stebalien, line 614: We need to retry here, not just warn.

PR #5767, opened 2021-03-10
node status api

### LINKED ISSUE #5762: NodeStatus API Endpoint for checking node health
Initially, it would be used to confirm that node is in a good state before producing a message.

Checks proposed:
- the sync status: how far out of sync the node is.
   - Use the difference between the current time and genesis to figure out which epoch we should be at.
- peering status: do we have more than 5 peers with positive pubsub scores?
   - i.e. are we confident that if we send this message now, it will be properly propagated to the network?

vyzo:
Closes #5762

### DIFF
--- a/node/impl/full.go
+++ b/node/impl/full.go
@@ -9,6 +9,8 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 
 	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/build"
+	"github.com/filecoin-project/lotus/chain/stmgr"
 	"github.com/filecoin-project/lotus/node/impl/client"
 	"github.com/filecoin-project/lotus/node/impl/common"
 	"github.com/filecoin-project/lotus/node/impl/full"
@@ -51,15 +53,21 @@ func (n *FullNodeAPI) NodeStatus(ctx context.Context) (status api.NodeStatus, er
 	delta := time.Since(timestamp).Seconds()
 	status.SyncStatus.Behind = uint64(delta / 30)
 
-	// get connected peers
-	pis, err := n.NetPeers(ctx)
+	// get peers in the messages and blocks topics
+	peersMsgs := make(map[peer.ID]struct{})
+	peersBlocks := make(map[peer.ID]struct{})
+
+	netName, err := stmgr.GetNetworkName(ctx, n.StateManager, curTs.ParentState())
 	if err != nil {
 		return status, err
 	}
 
-	peers := make(map[peer.ID]struct{})
-	for _, pi := range pis {
-		peers[pi.ID] = struct{}{}
+	for _, p := range n.PubSub.ListPeers(build.MessagesTopic(netName)) {
+		peersMsgs[p] = struct{}{}
+	}
+
+	for _, p := range n.PubSub.ListPeers(build.BlocksTopic(netName)) {
+		peersBlocks[p] = struct{}{}
 	}
 
 	// get scores for all connected and recent peers
@@ -69,14 +77,16 @@ func (n *FullNodeAPI) NodeStatus(ctx context.Context) (status api.NodeStatus, er
 	}
 
 	for _, score := range scores {
-		// we only care about connected peers
-		_, connected := peers[score.ID]
-		if !connected {
-			continue
-		}
-
 		if score.Score.Score > -1000 {
-			status.PeerStatus.PeersToPublish++
+			_, inMsgs := peersMsgs[score.ID]
+			if inMsgs {
+				status.PeerStatus.PeersToPublishMsgs++
+			}
+
+			_, inBlocks := peersBlocks[score.ID]
+			if inBlocks {
+				status.PeerStatus.PeersToPublishBlocks++
+			}
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


vyzo:
@magik6k what do you want to do with the chain health thing?

node/impl/full.go
```diff
@@ -37,4 +40,40 @@ func (n *FullNodeAPI) CreateBackup(ctx context.Context, fpath string) error {
...
+	}
+
+	scoreMap := make(map[peer.ID]float64)
+	for _, score := range scores {
+		scoreMap[score.ID] = score.Score.Score
+	}
+
+	for _, pi := range pis {
+		score := scoreMap[pi.ID]
+		if score > -1000 {
```
raulk, line 71: Where is this threshold coming from?

node/impl/full.go
```diff
@@ -37,4 +40,40 @@ func (n *FullNodeAPI) CreateBackup(ctx context.Context, fpath string) error {
...
+	}
+
+	scores, err := n.NetPubsubScores(ctx)
+	if err != nil {
+		return status, err
+	}
+
+	scoreMap := make(map[peer.ID]float64)
+	for _, score := range scores {
+		scoreMap[score.ID] = score.Score.Score
```
raulk, line 66: This map could be large, and it looks like too much detail for most use cases. Can we put it behind a flag?

node/impl/full.go
```diff
@@ -37,4 +40,40 @@ func (n *FullNodeAPI) CreateBackup(ctx context.Context, fpath string) error {
...
+	}
+
+	for _, pi := range pis {
+		score := scoreMap[pi.ID]
+		if score > -1000 {
+			status.PeerStatus.PeersToPublish++
```
raulk, line 72: Can we break down peers to publish by topic? The message management CLI cares about mpool, not about blocks (I think?).

vyzo: I guess we could list peers by topic and only use those; but in practice all of them subscribe to the same topics as us.

node/impl/full.go
vyzo: Hrm, maybe; it's kind of important however, at least for the logical health checks.

node/impl/full.go
vyzo: our score configuration, no good way to reference it.

node/impl/full.go
vyzo: We can also break it into two, peers to publish blocks and peers to publish messages.

node/impl/full.go
raulk: Can we create a constant in right place and reference it from here? I'm afraid that if we change the threshold at origin, we will forget to update it here.

node/impl/full.go
vyzo: also, how large is large? It won't be more than 1k in practice.

vyzo: we don't report specific peers, just the count of peers above the threshold.

node/impl/full.go
vyzo: probably.

node/impl/full.go
```diff
@@ -37,4 +40,40 @@ func (n *FullNodeAPI) CreateBackup(ctx context.Context, fpath string) error {
...
+
+	status.SyncStatus.Epoch = uint64(curTs.Height())
+	timestamp := time.Unix(int64(curTs.MinTimestamp()), 0)
+	delta := time.Since(timestamp).Seconds()
+	status.SyncStatus.Behind = uint64(delta / 30)
+
+	pis, err := n.NetPeers(ctx)
+	if err != nil {
+		return status, err
+	}
```
raulk, line 57: This obtains the currently _connected_ peers, which is a subset of peers with scores, right? Can we add a comment. It's not immediately obvious why we obtain peers separately, when `NetPubsubScores` already returns peers.

node/impl/full.go
raulk: Oh, I see. It's a bit confusing to follow due to the mixture of assignments between local vars and result struct fields.

node/impl/full.go
vyzo: right, NetPubScores is all peers within the retention window; will add comment.

node/impl/full.go
vyzo: done.

node/impl/full.go
vyzo: added comment.

node/impl/full.go
vyzo: crap now I have to fiddle with  DI :)

vyzo: ah, stmgr seems to be able to give it to me; hooray!

vyzo: done; there are now two separate counts -- one for blocks and once for msgs, and I use the peers listed in each topic by pubsub itself.

node/impl/full.go
vyzo: done.

vyzo:
we briefly discussed the chain health status with @magik6k and I will add some statistics on average number of blocks in the last finality (and also last 100 epochs).

I will make this optional with a boolean argument, as it is potentially expensive and we might be calling status a lot for message sending.

vyzo:
rebased on master

PR #5768, opened 2021-03-10
fix: retry proof when randomness changes

Stebalien:
Simply warning isn't enough.


magik6k (commented):
We still would retry, ~15 lines below

Stebalien:
Ah, because we check with `checkRand`? I see. But that's _really_ confusing and brittle. Can we just be explicit?

magik6k (approved):

PR #5773, opened 2021-03-11
fix window post rand check

IPFSMain-Official:
GenerateWindowPoSt func will modifies the value of rand (randomness[31] &= 0x3f), this will result in false in the bytes.Equal(checkRand, rand)


magik6k (approved):
Thanks for catching this

GFZRZK:
the usage of rand [here](https://github.com/filecoin-project/lotus/pull/5773/files#diff-4dbac165573efccdd31749a14b4c4f662da912af52776371b6743a6000ba98b4L620) should be changed also

GFZRZK:
@magik6k

PR #5776, opened 2021-03-11
backupds: Allow larger values in write log

magik6k:
Datastore entries can be larger than a few kb, which would lead us to create corrupted logs


Kubuxu (approved):
We really need to collect a set of use-cases and requirements for cbor-gen and make it work for them

PR #5790, opened 2021-03-11
polish(adt_diff): only diff if CID changed

frrist:
:facepalm: easy optimization I missed when implementing


frrist:
Test failures are due to calls to `Root()` resulting in a `Put` call on the underlying blockstore, which isn't supported... https://github.com/filecoin-project/lotus/blob/master/blockstore/api.go#L52

PR #5791, opened 2021-03-12
move DI stuff for paychmgr into modules

whyrusleeping:
This makes it easier to use the paychmgr package outside of lotus.

### DIFF
--- a/chain/stmgr/rpc/rpcstatemanager.go
+++ b/chain/stmgr/rpc/rpcstatemanager.go
@@ -1,4 +1,4 @@
-package modules
+package rpcstmgr
 
 import (
 	"context"
--- a/node/builder.go
+++ b/node/builder.go
@@ -307,9 +307,9 @@ var ChainNode = Options(
 	Override(new(api.WalletAPI), From(new(wallet.MultiWallet))),
 
 	// Service: Payment channels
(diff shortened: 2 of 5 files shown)


Stebalien (approved):

Stebalien (approved):

whyrusleeping:
I need one or two more things here... the paych manager thing still depends on some funky fx nonsense.  manageAPIImpl is an unexported struct that requires an embedded other type thats fulfilled via the constructor with a magic fx.In PaychAPI *struct* (not interface).

This needs some untangling...

whyrusleeping:
@magik6k youre gonna have to take a look at that last commit, I dont think I really understand what I was doing there with the dependency injection.

magik6k:
Just needed to use the magic `From`, the difference it makes is that if you pass a non-function value as the second param to `Override`, it just gets put into the DI container.

The `From` will instead wrap it into a function returning the value, so it is pulled from DI (`From(*T) -> func(t T) T {return t}`) - which in case of structs with `fx.In` means that all the fields internal to the struct will get pulled from DI (and without the `From` they will just be nil)

(This should be better documented)

magik6k (approved):

whyrusleeping:
Aha! that makes sense, thanks :)

PR #5794, opened 2021-03-12
fix: make sure atomic 64bit fields are 64bit aligned

Stebalien:
Otherwise, this won't work on 32bit ARM.

See https://golang.org/pkg/sync/atomic/#pkg-note-BUG

### DIFF
--- a/blockstore/badger/blockstore.go
+++ b/blockstore/badger/blockstore.go
@@ -84,10 +84,10 @@ const (
 // operation calls after Close() has returned, but it may not happen for
 // operations in progress. Those are likely to fail with a different error.
 type Blockstore struct {
-	DB *badger.DB
-
-	// state is guarded by atomic.
+	// state is accessed atomically
 	state int64
+
+	DB *badger.DB
 
 	prefixing bool
 	prefix    []byte


Stebalien:
Or we can just fix https://github.com/filecoin-project/lotus/issues/5795.

magik6k (commented):
Yeah, not sure how much we save by not using a RWMutex, it seems like the cleaner solution

We should also review other uses of atomics in the codebase (maybe check  if there are any linters which could detect this)

ribasushi:
@Stebalien afaik filecoin-ffi has a hard-dependency on 64bit instruction sets, so this is moot in the context of lotus

magik6k (approved):

Stebalien:
TL;DR: There's no reason _not_ to fix this. While we never plan on supporting 32bit platforms, someone else might come along and do that.

Alternatively, someone might try to create a proof-less "lotus lite", extract this code into a separate package, fork, etc.

yaohcn:
> Yeah, not sure how much we save by not using a RWMutex, it seems like the cleaner solution
> 
> We should also review other uses of atomics in the codebase (maybe check if there are any linters which could detect this)

There are many uses of atomics in the codebase.

PR #5800, opened 2021-03-12
shed command to unpack miner info dumps

magik6k:
(also makes the `info all` a bit more robust)

### DIFF
--- a/cmd/lotus-shed/miner.go
+++ b/cmd/lotus-shed/miner.go
@@ -1,0 +1,112 @@
+package main
+
+import (
+	"bufio"
+	"github.com/mitchellh/go-homedir"
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+var minerCmd = &cli.Command{
+	Name:  "miner",
+	Usage: "miner-related utilities",
+	Subcommands: []*cli.Command{
+		minerUnpackInfoCmd,
+	},
+}
+
+var minerUnpackInfoCmd = &cli.Command{
+	Name:  "unpack-info",
+	Usage: "unpack miner info all dump",
+	ArgsUsage: "[allinfo.txt] [dir]",
+	Action: func(cctx *cli.Context) error {
+		if cctx.Args().Len() != 2 {
+			return xerrors.Errorf("expected 2 args")
+		}
+
+		src, err := homedir.Expand(cctx.Args().Get(0))
+		if err != nil {
+			return xerrors.Errorf("expand src: %w", err)
+		}
+
+		f, err := os.Open(src)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/lotus-shed/miner.go
```diff
@@ -0,0 +1,113 @@
...
+	Usage: "miner-related utilities",
+	Subcommands: []*cli.Command{
+		minerUnpackInfoCmd,
+	},
+}
+
+var minerUnpackInfoCmd = &cli.Command{
+	Name:      "unpack-info",
+	Usage:     "unpack miner info all dump",
+	ArgsUsage: "[allinfo.txt] [dir]",
```
Stebalien, line 26: nit: call `dir` `output direcctory` (i.e., make it clear that this is the output).

cmd/lotus-shed/miner.go
```diff
@@ -0,0 +1,113 @@
...
+				}
+			}
+			if err != nil {
+				return xerrors.Errorf("read line: %w", err)
+			}
+			sl := string(l)
+
+			if strings.HasPrefix(sl, "#") {
+				if strings.Contains(sl, "..") {
+					return xerrors.Errorf("bad name %s", sl)
```
Stebalien, line 65: Are we trying to prevent directory traversal? (maybe a comment)

cmd/lotus-shed/miner.go
```diff
@@ -0,0 +1,113 @@
...
+
+		var outf *os.File
+
+		r := bufio.NewReader(f)
+		for {
+			l, _, err := r.ReadLine()
+			if err == io.EOF {
+				if outf != nil {
+					return outf.Close()
+				}
```
Stebalien, line 56: Maybe return an explicit "we did nothing" error?

cmd/lotus-shed/miner.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+
+			if outf != nil {
+				if _, err := outf.Write(l); err != nil {
+					return xerrors.Errorf("write line: %w", err)
+				}
+				if _, err := outf.Write([]byte("\n")); err != nil {
+					return xerrors.Errorf("write line end: %w", err)
+				}
```
Stebalien, line 109: nit: `fmt.Fprintln(outf, l)` is a bit nicer.

cmd/lotus-storage-miner/info_all.go
```diff
@@ -35,88 +35,88 @@ var infoAllCmd = &cli.Command{
 
 		fmt.Println("#: Version")
 		if err := lcli.VersionCmd.Action(cctx); err != nil {
```
Stebalien, line 37: It would be nice to annotate these errors with _what_ failed (even just a line number).

Stebalien (approved):
A few comments but nothing important/blocking.

PR #5801, opened 2021-03-12
storagefsm: Trigger input processing when below limits

magik6k:
Currently when a miner starts processing more deals in parallel than the configured sealing limits, we wouldn't start processing more deals after the first batch of sectors got sealed.

This should fix that case

### DIFF
--- a/extern/storage-sealing/fsm.go
+++ b/extern/storage-sealing/fsm.go
@@ -412,7 +412,7 @@ func (m *Sealing) onUpdateSector(ctx context.Context, state *SectorInfo) error {
 	// trigger more input processing when we've dipped below max sealing limits
 	if shouldUpdateInput {
 		go func() {
-			m.inputLk.Unlock()
+			m.inputLk.Lock()
 			defer m.inputLk.Unlock()
 
 			if err := m.updateInput(ctx, sp); err != nil {
--- a/extern/storage-sealing/fsm.go
+++ b/extern/storage-sealing/fsm.go
@@ -394,6 +394,10 @@ func (m *Sealing) plan(events []statemachine.Event, state *SectorInfo) (func(sta
 }
 
 func (m *Sealing) onUpdateSector(ctx context.Context, state *SectorInfo) error {
+	if m.getConfig == nil {
+		return nil // tests
+	}
+
 	cfg, err := m.getConfig()
 	if err != nil {
 		return xerrors.Errorf("getting config: %w", err)
--- a/extern/storage-sealing/fsm.go
+++ b/extern/storage-sealing/fsm.go
@@ -300,7 +300,9 @@ func (m *Sealing) plan(events []statemachine.Event, state *SectorInfo) (func(sta
 
 	*/
 
-	m.stats.updateSector(m.minerSectorID(state.SectorNumber), state.State)
+	if err := m.onUpdateSector(context.TODO(), state); err != nil {
+		log.Errorw("update sector stats", "error", err)
+	}
 
 	switch state.State {
 	// Happy path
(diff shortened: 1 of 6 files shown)


extern/storage-sealing/fsm.go
```diff
@@ -391,6 +393,37 @@ func (m *Sealing) plan(events []statemachine.Event, state *SectorInfo) (func(sta
...
+	}
+	sp, err := m.currentSealProof(ctx)
+	if err != nil {
+		return xerrors.Errorf("getting seal proof type: %w", err)
+	}
+
+	shouldUpdateInput := m.stats.updateSector(cfg, m.minerSectorID(state.SectorNumber), state.State)
+
+	// trigger more input processing when we've dipped below max sealing limits
+	if shouldUpdateInput {
```
magik6k, line 413: changing this to `shouldUpdateInput && false` makes `TestBatchDealInput` fail reliably

extern/storage-sealing/fsm.go
```diff
@@ -300,7 +300,9 @@ func (m *Sealing) plan(events []statemachine.Event, state *SectorInfo) (func(sta
 
 	*/
 
-	m.stats.updateSector(m.minerSectorID(state.SectorNumber), state.State)
+	if err := m.onUpdateSector(context.TODO(), state); err != nil {
+		log.Errorw("update sector stats", "error", err)
```
Stebalien, line 304: shouldn't we be doing something more than just logging?

extern/storage-sealing/stats.go
```diff
@@ -23,10 +24,14 @@ type SectorStats struct {
 	totals   [nsst]uint64
 }
 
-func (ss *SectorStats) updateSector(id abi.SectorID, st SectorState) {
+func (ss *SectorStats) updateSector(cfg sealiface.Config, id abi.SectorID, st SectorState) (updateInput bool) {
 	ss.lk.Lock()
 	defer ss.lk.Unlock()
 
+	preSealing := ss.totals[sstStaging] + ss.totals[sstSealing] + ss.totals[sstFailed]
+	preStaging := ss.totals[sstStaging]
```
Stebalien, line 32: maybe define a helper function?

extern/storage-sealing/fsm.go
```diff
@@ -391,6 +393,37 @@ func (m *Sealing) plan(events []statemachine.Event, state *SectorInfo) (func(sta
...
+
+	shouldUpdateInput := m.stats.updateSector(cfg, m.minerSectorID(state.SectorNumber), state.State)
+
+	// trigger more input processing when we've dipped below max sealing limits
+	if shouldUpdateInput {
+		go func() {
+			m.inputLk.Lock()
+			defer m.inputLk.Unlock()
+
+			if err := m.updateInput(ctx, sp); err != nil {
```
Stebalien, line 418: Could we end up triggering an avalanche of input updates (there's no deduping)? I assume nothing is going to really happen until we call `updateInput` but I figured I'd check.

Stebalien (approved):

extern/storage-sealing/fsm.go
magik6k: We could do some retrying, but all the errors we can get here are of the "something is horribly broken" nature, so it probably wouldn't even help

extern/storage-sealing/fsm.go
magik6k: This should only be getting called when we  go from 'too many sectors sealing' to 'can seal more sectors', which should happen once per sealed sector, which means that even if you are a really, really big miners, you likely won't call this more than a few times per epoch (and at that scale it's highly likely that you're not even using this code)

extern/storage-sealing/fsm.go
Stebalien: Yeah, makes sense.

Stebalien (approved):

extern/storage-sealing/input.go
```diff
@@ -400,6 +400,9 @@ func (m *Sealing) tryCreateDealSector(ctx context.Context, sp abi.RegisteredSeal
 		return xerrors.Errorf("initializing sector: %w", err)
 	}
 
+	// update stats early, fsm planner would do that async
+	m.stats.updateSector(cfg, m.minerSectorID(sid), UndefinedSectorState)
```
magik6k, line 404: Without this we could end up creating a bunch of sectors at once, and start more parallel work than specified by the max sealing sector config.

Stebalien (approved):

Stebalien (approved):

PR #5804, opened 2021-03-12
fix: StateManager.Replay

zl03jsj:
in the `statemanager.replay` method, 
after get the msg(specified by 'mcid' param) execution result,
the callback function returns an 'errHaltExecution' to break  out the loop in `computeTipSetState`
but some times `computeTipsetstate`  warps the `errHaltExecution` with `xerrors`
that would cause following expression:

(err != errHaltExecution)  to be `true` .

which  causes `StateManager.Replay` returns an error with following message: 

`unexpected error during execution: callback failed on reward message: halt`


Stebalien (approved):

magik6k (approved):
Thanks!

PR #5805, opened 2021-03-12
mock sealer: Simulate randomness sideeffects

magik6k:
Ideally we would change the ffiwrapper to not have sideeffects, but we need to make sure that we don't use this randomness anywhere after it's modified by the ffi calls.

This at least makes the mock and ffiwrapper do the same thing


Stebalien (approved):
I've updated it to mock side effects on window post verification as well.

PR #5819, opened 2021-03-16
Add --actor flag in lotus-shed sectors terminate

chadwick2143:
Can specify miner actor address when terminate sectors,
so that you can terminate sectors without miner api.


magik6k (approved):
Thanks!

PR #5834, opened 2021-03-18
Nerpa reset based on 1.5.2

liviudm:


ribasushi (approved):
Validated to work past v3 upgrade, releasing the ~hounds~ miners

PR #5837, opened 2021-03-18
Build more binaries by default when a devnet tag is active

ribasushi:
@arajasek without doing this I need to manually build shed, which is super annoying


arajasek (approved):
yeah, why the hell not

magik6k (approved):

PR #5838, opened 2021-03-18
api: Document StateSearchMsg replaced message behavior

magik6k:
This is a fairly sharp edge which we should fix in the v1 API, but for now it should at least get better docs

### DIFF
--- a/api/api_full.go
+++ b/api/api_full.go
@@ -403,6 +403,20 @@ type FullNode interface {
 	// StateSectorPartition finds deadline/partition with the specified sector
 	StateSectorPartition(ctx context.Context, maddr address.Address, sectorNumber abi.SectorNumber, tok types.TipSetKey) (*miner.SectorLocation, error)
 	// StateSearchMsg searches for a message in the chain, and returns its receipt and the tipset where it was executed
+	//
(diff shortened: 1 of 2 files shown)


api/api_full.go
```diff
@@ -70,7 +70,8 @@ type FullNode interface {
 	ChainGetBlockMessages(ctx context.Context, blockCid cid.Cid) (*BlockMessages, error)
 
 	// ChainGetParentReceipts returns receipts for messages in parent tipset of
-	// the specified block.
+	// the specified block. The receipts in the list returned is one-to-one with the
```
arajasek, line 73: Someone confirm i'm not lying.

arajasek (approved):

api/api_full.go
magik6k: Correct

arajasek (approved):

magik6k:
> There is a lot of copy/paste. Is there a way to reduce this?
In v1 API we can consolidate those methods more

> Related, is the content in api-methods.md generated from api_full.go automatically?
Yep, that file is generated

> Should we maybe state that some of our APIs are "replacement message sensitive"
This is basically what we do in this PR, I don't thing we have more APIs which are replace-sensitive

PR #4711, opened 2020-11-03
OpenRPC Support

meowsbits:
Rel https://github.com/filecoin-project/devgrants/pull/165

This is an implementation adding OpenRPC Document support for the Filecoin API(s). 

Note that the `docsgen` program (easily accessed via `make docsgen`) also uses reflection to generate Markdown documentation, and that program has been refactored to share some logic with the OpenRPC documentation to avoid code duplication (eg. examples and comments).

Would love to get some feedback on the general architecture (ie Is plugging directly in to the `rpc` package a best approach?), as well as a few other things, like Is re-using the given examples from the docsgen acceptable?, Is it desirable to add more detailed schemas for common data types (which has been exemplified by a tentative `cid.Cid`)? 

We'd be happy and eager to schedule a meeting to talk through some of these ideas and questions if interest and time allows. 

---

Here's a gist of what the rendered `rpc.discover` response can look like at the OpenRPC Playground: https://playground.open-rpc.org/?url=https://gist.githubusercontent.com/meowsbits/2f22b61ce46f6897d343f850ac86b540/raw/openrpc_doc.json

image

PR #5843, opened 2021-03-19
OpenRPC Support

magik6k:
This some remaining issues with https://github.com/filecoin-project/lotus/pull/4711, should be mergable in this shape now to unblock more V1 API work


Kubuxu:
What are the two `goros` and `goross` files?

magik6k:
Those were the result of running `go mod tidy` and `git add go*`, fixed now

Kubuxu (approved):
SGWM but IMO we should squash it due to non-building commits in the middle.

PR #5841, opened 2021-03-18
shed: Command to list duplicate messages in tipsets

magik6k:

PR #5847, opened 2021-03-19
shed: command to list duplicate messages in tipsets (steb)

Stebalien:
Modification of #5841.

This one:

1. Let's you specify a start/stop height.
2. Prints out ndjson to stdout, and status/errors to stderr.
3. Prints one entry per set of duplicates.
4. Allows filtering messages by method number.
5. Allows configuring the number of threads.

Usage:

```bash
> cd cmd/lotus-shed
> go build
> ./lotus-shed audits duplicate-messages --start=148888 | tee output.ndjson
```

That will walk back to the mainnet epoch and write the results to stdout and output.ndjson.

**TODO**

* [x] Make it possible to _start_ at some height.
* [ ] Make it faster (parallel fetch and/or direct datastore access)?
* [ ] Make the method number filter customizable.
* [x] Add better docs

### DIFF
--- a/cmd/lotus-shed/balances.go
+++ b/cmd/lotus-shed/balances.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/filecoin-project/lotus/build"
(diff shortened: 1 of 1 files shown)


Stebalien:
I sincerely apologize for the git commit messages, I needed to send the code to a different machine for testing, and got tired of writing nice git commit messages. I'll rebase/squash later but don't want to mess with history right now.

Kubuxu:
@Stebalien for this I use git over ssh with bare repo on the other side and one working repo.
So for example:
```
devbox1> git init --bare ~/lotus-bare
local> git remote add  devbox ssh://devbox1/~/lotus-bare; git push master
devbox1> git clone ~/lotus-bare ~/lotus
```
Then if you want to push some branch you push to devbox1 and in `~/lotus` dir in devbox you just pull and check it out.
The bare repo in the middle is there because you cannot push to checked out branches.

cmd/lotus-shed/balances.go
```diff
@@ -70,6 +73,204 @@ var auditsCmd = &cli.Command{
...
+
+		methods := make(map[abi.MethodNum]bool)
+		for _, m := range cctx.IntSlice("method") {
+			if m < 0 {
+				return fmt.Errorf("expected method numbers to be non-negative")
+			}
+			methods[abi.MethodNum(m)] = true
+		}
+
+		target := abi.ChainEpoch(cctx.Int("start"))
```
arajasek, line 156: Might as well do some sanity-checking of `start` here (non-negative, < `end`, etc.)

cmd/lotus-shed/balances.go
```diff
@@ -70,6 +73,204 @@ var auditsCmd = &cli.Command{
...
+					for i, m := range bms.SecpkMessages {
+						if len(methods) > 0 && !methods[m.Message.Method] {
+							continue
+						}
+						c, ok := msgs[anonce(&m.Message)]
+						if !ok {
+							c = make(map[cid.Cid]*types.Message, 1)
+							msgs[anonce(&m.Message)] = c
+						}
+						c[bms.Cids[len(bms.BlsMessages)+i]] = &m.Message
```
arajasek, line 214: Ew. But yes.

cmd/lotus-shed/balances.go
```diff
@@ -70,6 +73,204 @@ var auditsCmd = &cli.Command{
...
+			DefaultText: "genesis",
+		},
+		&cli.IntFlag{
+			Name:        "end",
+			Usage:       "the last epoch to check",
+			DefaultText: "the current head",
+		},
+		&cli.IntSliceFlag{
+			Name:        "method",
+			Usage:       "Filter results by method number.",
```
arajasek, line 110: Yeah, I'm not sure how useful this is? Given that different MethodNums mean different things, I feel like this will practically come down to {all methods, only balance transfers (0), everything except balance transfers (non-zero)}.

arajasek (approved):
LGTM, maybe confirm it works with latest changes.

cmd/lotus-shed/balances.go
Stebalien: We used this before to only look for specific message types (sends, likely multisig calls, etc.). It's not perfect, but it can filter out a lot of noise.

cmd/lotus-shed/balances.go
Stebalien: Done.

Stebalien:
> confirm it works with latest changes

Done. I've also added support for `{exclude,include}-{from,to}` (which also seems to work).

Stebalien:
(going to assume this is good to go, it's just a shed command after all).

PR #5850, opened 2021-03-21
expose NextID from nice market actor interface

whyrusleeping:
I needed it, so i coded it.


Kubuxu:
Why do you need NextID?

magik6k (approved):

Issue #5829, opened 2021-03-17
[BUG] Retrieval deal fails with "deal data transfer failed: incomplete response" - Client Side


ethanwu155:
**Describe the bug**
Client cannot get stored file on FileCoin from miner. Stored file has `StorageDealActive` Status.

**Version (run `lotus version`):**
lotus version 1.5.0+mainnet+git.191a05da4

**To Reproduce**
1. Run `lotus client retrieve --miner <minerID> <data CID> <output file>`
2. Get the following error: 

```
ERROR: retrieval failed: Retrieve: Retrieval Error: error generated by data transfer: deal │ data transfer failed: incomplete response
```

**Expected behavior**
Expect to be able to retrieve stored file.

**Logs**
```
$ lotus client retrieve --miner f066596 bafyaa6qsgafcmalqudsaeiekkmiypfxq6vgdwepdafg6jyacye46gurjichuapbzxwqhoywq7yjaagelucbyabasgafcmalqudsaeif3e6tsyc7c3anlmobix3744duq7cyasjsmrsmbkdn4besizu5ffyjaagf6zcc36aykcqeaeghh3sbl6bzaqcaibaaeedt5zav7am output.mp4                                                                                       
> Recv: 0 B, Paid 0 FIL, ClientEventOpen (DealStatusNew)                                  
> Recv: 0 B, Paid 0 FIL, ClientEventDealProposed (DealStatusWaitForAcceptance)            
> Recv: 122 B, Paid 0 FIL, ClientEventBlocksReceived (DealStatusWaitForAcceptance)        
> Recv: 122 B, Paid 0 FIL, ClientEventDealAccepted (DealStatusAccepted)                   
> Recv: 122 B, Paid 0 FIL, ClientEventPaymentChannelAddingFunds (DealStatusPaymentChannelAllocatingLane)                                                                            
> Recv: 122 B, Paid 0 FIL, ClientEventLaneAllocated (DealStatusOngoing)                   
> Recv: 122 B, Paid 0 FIL, ClientEventDataTransferError (DealStatusErrored)               
> Recv: 0 B, Paid 0 FIL, ClientEventOpen (DealStatusNew)                                  
ERROR: retrieval failed: Retrieve: Retrieval Error: error generated by data transfer: deal data transfer failed: incomplete response 
 ```
 
**Screenshots**
See attachment

**Additional context**
None
[...]

PR #5851, opened 2021-03-22
better logging when unsealing fails

dirkmc:
This should help miners identify the underlying cause of errors during unsealing. See https://github.com/filecoin-project/lotus/issues/5829

### DIFF
--- a/markets/retrievaladapter/provider.go
+++ b/markets/retrievaladapter/provider.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"io"
 
+	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
+
 	"github.com/filecoin-project/lotus/api"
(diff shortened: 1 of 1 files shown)


markets/retrievaladapter/provider.go
```diff
@@ -61,13 +64,20 @@ func (rpn *retrievalProviderNode) UnsealSector(ctx context.Context, sectorID abi
...
 
+	// Set up a pipe so that data can be written from the unsealing process
+	// into the reader returned by this function
 	r, w := io.Pipe()
 	go func() {
 		var commD cid.Cid
 		if si.CommD != nil {
 			commD = *si.CommD
 		}
+		// Unseal the piece into the pipe's writer
```
magik6k, line 75: Technically this doesn't always unseal, if we have the unsealed copy around, we'll just read that

magik6k (commented):
1 Nit about the comment

magik6k (approved):

PR #5852, opened 2021-03-22
Rework message checks

vyzo:
As we discussed:
- We now have a generic check structure, that includes hints
- The message check struct aggregates the generic check structure, together with a message Cid
- The message check api:
   - Returns all checks for a given message, grouped by message instead of a flat array
   - Provides hints for resolving failing checks


documentation/en/api-methods.md
```diff
@@ -2435,9 +2421,7 @@ Response:
 ```
 
 ### MsigCreate
-MsigCreate creates a multisig wallet
-It takes the following params: <required number of senders>, <approving addresses>, <unlock duration>
-<initial balance>, <sender address of the create msg>, <gas price>
+INVERT
```
Kubuxu, line 2424: Looks like my `//INVERT` comments broke docks.

api/types.go
```diff
@@ -137,29 +137,34 @@ type NodeChainStatus struct {
...
+	CheckStatusMessageBaseFeeUpperBound
+	CheckStatusMessageGetStateNonce
+	CheckStatusMessageNonce
+	CheckStatusMessageGetStateBalance
+	CheckStatusMessageBalance
 )
 
+type CheckStatus struct {
+	Code CheckStatusCode
+	OK   bool
```
Kubuxu, line 162: I'm not sure if I would like it to be bool or enum with: success, fail, not-applicable (or something like that).

Kubuxu (approved):

api/types.go
```diff
@@ -137,29 +137,34 @@ type NodeChainStatus struct {
...
+	Err  string
+	Hint map[string]interface{}
+}
+
 type MessageCheckStatus struct {
```
Kubuxu, line 167: We need to figure out better names.
I'm thinking about:
for the top-level `CheckResult`
for the message `MessageCheckResult`
for the enum `CheckID`.

Hmm, if we move everything about checks to a separate package then we can strip the `Check` prefix from everything.
So it would be `check.Result`. `check.ID` or `check.Code` and the enum would have shorter prefix.
WDYT?
@vyzo

api/types.go
vyzo: I think we will be fine with a bool; non applicable checks can be simply skipped.

api/types.go
vyzo: We can certainly do that, I am not fond of comically long names either :)

PR #5854, opened 2021-03-22
API proxy struct codegen

magik6k:
Includes https://github.com/filecoin-project/lotus/pull/5844

The main part is just 300LOC in ` gen/api/proxygen.go`, rest of the diff is moving permissions and codegen

Like docsgen, for now it's hardcoded for the `api/` package, that will get fixed when porting to the versioned api


api/api_common.go
```diff
@@ -19,60 +19,60 @@ type Common interface {
 
 	// MethodGroup: Auth
 
-	AuthVerify(ctx context.Context, token string) ([]auth.Permission, error)
-	AuthNew(ctx context.Context, perms []auth.Permission) ([]byte, error)
+	AuthVerify(ctx context.Context, token string) ([]auth.Permission, error) // perm:read
```
Kubuxu, line 22: directives, in contrast to comments,  are written without leading space.

gen/api/proxygen.go
```diff
@@ -0,0 +1,300 @@
...
+{{end}}
+
+`)
+	return err
+}
+
+func doTemplate(w io.Writer, info interface{}, templ string) error {
+	t := template.Must(template.New("").
+		Funcs(template.FuncMap{
+			"ReadHeader": func(rdr string) string {
```
Kubuxu, line 294: Unused.

magik6k:
Ah, openrpc was getting param names from the struct, which is why those files are now smaller - diff from master:
```
10c10
<       "description": "```go\nfunc (c *FullNodeStruct) BeaconGetEntry(ctx context.Context, epoch abi.ChainEpoch) (*types.BeaconEntry, error) {\n\treturn c.Internal.BeaconGetEntry(ctx, epoch)\n}\n```",
---
>       "description": "```go\nfunc (s *FullNodeStruct) BeaconGetEntry(p0 context.Context, p1 abi.ChainEpoch) (*types.BeaconEntry, error) {\n\treturn s.Internal.BeaconGetEntry(p0, p1)\n}\n```",
15c15
<           "name": "epoch",
---
>           "name": "p1",
66c66
<         "url": "https://github.com/filecoin-project/lotus/blob/master/api/apistruct/struct.go#L893"
---
>         "url": "https://github.com/filecoin-project/lotus/blob/master/api/apistruct/struct.go#L885"

```

Makefile
```diff
@@ -325,7 +325,11 @@ type-gen:
 method-gen:
 	(cd ./lotuspond/front/src/chain && go run ./methodgen.go)
 
-gen: type-gen method-gen
+api-gen:
+	go run ./gen/api > api/apistruct/struct.go
+	goimports -w api/apistruct
+	goimports -w api/apistruct
```
vyzo, line 331: why twice?

gen/api/proxygen.go
```diff
@@ -0,0 +1,297 @@
...
+	case *ast.ChanType:
+		subt, err := typeName(t.Value)
+		if err != nil {
+			return "", err
+		}
+		if t.Dir == ast.SEND {
+			subt = "->chan " + subt
+		} else {
+			subt = "<-chan " + subt
+		}
```
vyzo, line 112: how do we handle bidirectional channels? Or are they just not a thing in the api?

vyzo (commented):
this looks good to me and it's definitely an improvement over having to do it by hand.

I suppose the v1 scaffolding will come at a later pr or is it planned for here?

Makefile
magik6k: Apparently calling it once won't fully fix imports

gen/api/proxygen.go
magik6k: Those don't work over jsonrpc

Kubuxu:
It was also my mistake re binary files, I through the 99% meant 99% smaller, it probably means 99% different.

vyzo (approved):

Kubuxu (approved):
Let's squash as it doesn't build in the middle.

PR #5864, opened 2021-03-23
Export developer and network commands for consumption by derivatives of Lotus

frrist:
Sentinel-Visor embeds and instantiates a lotus node. The commands exported here may be reused in that implementation.


arajasek (approved):
No objection from me

magik6k (approved):

PR #5866, opened 2021-03-24
Feat/batch estimate message

hunjixin:


magik6k (commented):
Needs `make gen`

node/impl/full/gas.go
```diff
@@ -349,3 +353,113 @@ func (m *GasModule) GasEstimateMessageGas(ctx context.Context, msg *types.Messag
...
+	}
+
+	pending, ts := m.Mpool.PendingFor(fromA)
+	priorMsgs := make([]types.ChainMsg, 0, len(pending))
+	for _, m := range pending {
+		priorMsgs = append(priorMsgs, m)
+	}
+
+	var msgs []*types.Message
+	for _, estimateMessage := range estimateMessages {
```
magik6k, line 382: Should deduplicate this code with `gasEstimateGasLimit`

node/impl/full/gas.go
```diff
@@ -349,3 +353,113 @@ func (m *GasModule) GasEstimateMessageGas(ctx context.Context, msg *types.Messag
 
 	return msg, nil
 }
+
+func (m *GasModule) GasBatchEstimateMessageGas(ctx context.Context, estimateMessages []*api.EstimateMessage, selectCount int, tsk types.TipSetKey) ([]*types.Message, error) {
+	if len(estimateMessages) == 0 {
+		return nil, nil
+	}
+	if selectCount == 0 {
```
magik6k, line 361: What's the use of this parameter? Why not just pass fewer messages?

hunjixin:
@magik6k 
Now that this code doesn't work very well, here's a description of the problem I'm having. 
As we write the messenger component, we need to select messages in batches, assign nonce, estimate gas charges, and now the practice is to assign nonce one by one, request GasEstimateMessageGas, request signatures, request the MpoolPush interface. The problem with this is that：
1.  Api takes a long time. During the process of estimating 100 message gas, the cumulative time of the GasEstimateMessageGas  is at the second level
2. Too many requests caused the request to fail

eg:
If there are 1000 messages in the message pool (which is not a large amount for the mine pool), select 20 message .But because there may be an error in the middle of the message, I can't just pass in 10, and here I'm thinking about 2 times the number of incoming estimate messages, choosing 10 out of 20. 

Another problem is nonce allocation, and if there is an error in the middle,  the nonce of the error message can be assigned to the next message,  Otherwise, to get enough messages, back to the messenger to reassign nonce  and request daemon again. If have many error msg, myabe repeat this many times

so I hope to have an interface that can do this at once, and after careful consideration, such a function signature will meet my needs. If you think it's okay, I'm going to re-implement this API along these idea
```sh
GasBatchEstimateMessageGas(ctx context.Context, estimateMessages []*api.EstimateMessage, selectCount int, fromNonce uint64, tsk types.TipSetKey) ([]*api.EstimateResult, error)
```

jennijuju:
may partially resolve https://github.com/filecoin-project/lotus/issues/5964

hunjixin:
> may partially resolve #5964

it will be better if CallWithGas can run one by one msg

jennijuju:
Triage note: review this after the v1 api update

arajasek:
Stale :(

Issue #5870, opened 2021-03-24
[Feature Request] lotus CLI command to cancel retrieval deal


dirkmc:
Sometimes retrieval deals get stuck in a state where the data transfer has failed, but the deal is still in Ongoing: https://github.com/filecoin-project/go-fil-markets/issues/513

We should add a command to explicitly cancel a retrieval deal, and any associated data transfer.

PR #5871, opened 2021-03-24
add `lotus client cancel-retrieval` cmd to lotus CLI

nonsense:
Fixes https://github.com/filecoin-project/lotus/issues/5870

---

Added by @raulk:

This command might seem redudant with `lotus client cancel-transfer`. However:
1. This cancels the retrieval deal no matter which stage it's at (transfer or not).
2. Retrieval currently has some edge cases where cancelling the transfer will not cancel the retrieval deal.
3. We should probably get rid of `lotus client cancel-transfer`, and instead rely on the commands to cancel the deals themselves, which would terminate the deal regardless of the stage.


dirkmc (approved):

nonsense:
1. `test` and `test-short` seem flaky and I think failures are not related.

2. After running `make docsgen`, I get three files that are modified:

```
	modified:   build/openrpc/full.json.gz
	modified:   build/openrpc/miner.json.gz
	modified:   build/openrpc/worker.json.gz
```

Not sure if I should be replacing these as part of the PR?

dirkmc (approved):

cli/client.go
```diff
@@ -1975,6 +1976,32 @@ var clientCancelTransfer = &cli.Command{
 	},
 }
 
+var clientCancelRetrievalDeal = &cli.Command{
+	Name:  "cancel-retrieval-deal",
```
magik6k, line 1980:
```suggestion
	Name:  "cancel-retrieval",
```

magik6k: Or maybe even make this a sub-command of the `retrieve` command -- 
* `lotus client cancel-retrieval-deal` vs
* `lotus client cancel-retrieval` vs
* `lotus client retrieve cancel`

cli/client.go
```diff
@@ -1975,6 +1976,32 @@ var clientCancelTransfer = &cli.Command{
...
+	Action: func(cctx *cli.Context) error {
+		api, closer, err := GetFullNodeAPI(cctx)
+		if err != nil {
+			return err
+		}
+		defer closer()
+		ctx := ReqContext(cctx)
+
+		if cctx.Int64("dealid") == 0 {
+			return errors.New("deal id cannot be 0")
```
magik6k, line 1998: Are you sure we don't start that counter at 0?

magik6k (commented):
2 comments

(needs `make gen`)

cli/client.go
dirkmc: hmm turns out it *does* start at zero: https://github.com/filecoin-project/go-storedcounter/blob/master/storedcounter.go#L23
So we should only return an error if the deal id is < 0

cli/client.go
dirkmc: I like `lotus client retrieve cancel` 👍

raulk: Agree.

aarshkshah1992:
@magik6k @dirkmc  Have addressed the review comments.

But, on running `make gen`, I seem to be running into some weird import error:

```go run ./gen/main.go
# github.com/filecoin-project/lotus/api/apistruct
api/apistruct/struct.go:6:1: imported and not used: "bytes"
api/apistruct/struct.go:8:1: imported and not used: "encoding/json"
api/apistruct/struct.go:9:1: imported and not used: "fmt"
api/apistruct/struct.go:19:1: imported and not used: "github.com/filecoin-project/go-state-types/big"
api/apistruct/struct.go:23:1: imported and not used: "github.com/filecoin-project/lotus/chain/actors/builtin"
api/apistruct/struct.go:27:1: imported and not used: "github.com/filecoin-project/lotus/chain/actors/builtin/power"
api/apistruct/struct.go:35:1: market redeclared as imported package name
	previous declaration at api/apistruct/struct.go:24:1
api/apistruct/struct.go:35:1: imported and not used: "github.com/filecoin-project/specs-actors/v2/actors/builtin/market"
api/apistruct/struct.go:43:1: imported and not used: "github.com/libp2p/go-libp2p-pubsub" as pubsub
api/apistruct/struct.go:44:1: imported and not used: "github.com/multiformats/go-multiaddr" as multiaddr
api/apistruct/struct.go:44:1: too many errors
make: *** [type-gen] Error 2
```

What should I do to fix this ?

dirkmc:
I was able to successfully run `make gen` and pushed the changes

dirkmc:
It wasn't possible to add the command as a subcommand of `lotus client retrieve` because `lotus client retrieve` is itself a command that takes arguments.
Instead I've named it `lotus client retrieve-cancel`

raulk (approved):

raulk:
This has been squashed into https://github.com/filecoin-project/lotus/tree/minerx/staging for minerX early testing.

arajasek:
Fixes #5870

magik6k (approved):

magik6k:
An integration test would be great to have

magik6k (approved):

nonsense:
> An integration test would be great to have

Note that this is not forgotten, I just got pulled into something else. I think when we add `list retrieval deals` command, it should be trivial to add an integration test for this cmd as well.

PR #5876, opened 2021-03-24
Take latency into account when making interactive deals

magik6k:
In manual testing this results in drastically higher success rate for me.

After setting max 150ms latency, making 20 deals, 12 got rejected immediately, 8 started transferring with decent speed right away.


arajasek (approved):
SGTM

PR #5884, opened 2021-03-26
remote calc winningpost proof

hunjixin:
This pr is intended to provide compatible venus mining, venus mining to integrate the existing miner of lotus


magik6k (commented):
This makes sense, just need to run `make gen`, and commit

magik6k (approved):

PR #5887, opened 2021-03-26
Testground checks on push

coryschwartz:


arajasek:
We need to figure out which of these test plans we'll actually want (concerns around stability / we already fail some of these). Need to poke @nonsense probably.

.github/workflows/testground-on-push.yml
```diff
@@ -0,0 +1,33 @@
...
+  testground:
+    runs-on: ubuntu-latest
+    name: ${{ matrix.composition_file }}
+    strategy:
+      matrix:
+        include:
+          - backend_addr: ci.testground.ipfs.team
+            backend_proto: https
+            plan_directory: testplans/graphsync
+            composition_file: testplans/graphsync/_compositions/stress-k8s.toml
```
nonsense, line 16: This one probably doesn't belong to Lotus. I think if we have the graphsync repo first the graphsync tests, it'd be better!

nonsense:
> We need to figure out which of these test plans we'll actually want (concerns around stability / we already fail some of these). Need to poke @nonsense probably.

I believe we want both testplans related to Lotus - the end-to-end deals one, as well as the payment channels stress-test. The graphsync test plan should live in the graphsync repo I believe. There is no reason for any of these to be flaky or to not work ever, so I think it is great that we are adding the status checks! Thanks @coryschwartz !!!

.github/workflows/testground-on-push.yml
coryschwartz: Thanks, I'll remove this one.

BigLep:
@magik6k is going to open a GitHub issue to track the work for fixing the failing/flaky tests and then relate to this PR, since that is what's blocking this.

nonsense:
I think this is a great start. Ideally Testground itself could handle this (this functionality is buggy and not working on Testground unfortunately), and not have an additional Github Action handle the status checks updates, but I think for now this is fine.

Given that we will probably have more test plans in the future, It'd be nice to collect all Testground testplans related to Lotus under one status check, and not as individual status checks, however we only have 2 tests right now, so this is fine as is for a start.

Once `graphsync` is removed, I think we could merge.

nonsense:
> @magik6k is going to open a GitHub issue to track the work for fixing the failing/flaky tests and then relate to this PR, since that is what's blocking this.

I think both the `deals end-to-end test` (i.e. `baseline-k8s-3-1.toml`) and `payment channel stress tests` (i.e. paych-stress-k8s.toml`) are no longer flaky, given that we merge the following PRs:
- https://github.com/filecoin-project/lotus/pull/6122
- https://github.com/filecoin-project/lotus/pull/6124
- https://github.com/filecoin-project/lotus/pull/6126

.github/workflows/testground-on-push.yml
magik6k, line 16:
```suggestion
```

magik6k (approved):

PR #5888, opened 2021-03-26
create comments on release

coryschwartz:
This submits a bunch of testground tests and then reports them back to you in a comment.

To see an example of what this looks like, you can see https://github.com/coryschwartz/lotus/issues/20

### DIFF
--- a/.github/workflows/testground-release-issue.yml
+++ b/.github/workflows/testground-release-issue.yml
@@ -1,0 +1,115 @@
+---
+
+name: Testground Issue Checker
+
+on:
+  issues:
+    types:
+      - opened
+      - reopened
+
+jobs:
+  initial-comment:
+    if: contains(github.event.issue.labels.*.name, 'RELEASE')
+    runs-on: ubuntu-latest
+    steps:
+      - name: initial-comment
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          issue-number: ${{ github.event.issue.number }}
+          body: |
+            hey there!
+
+            To prepare for this release, I'm kicking off a few testground jobs.
+            Once they finish, I'll post an update in this comment.
+
+            __finished tests below this line__
+            
+            |composition|testground id|status|outcome|
+            |-----------|-------------|------|-------
+
+  testground-run:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - backend_addr: ci.testground.ipfs.team
+            backend_proto: https
+            plan_directory: testplans/graphsync
+            composition_file: testplans/graphsync/_compositions/stress-k8s.toml 
+          - backend_addr: ci.testground.ipfs.team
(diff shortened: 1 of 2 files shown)


magik6k:
We already have https://github.com/filecoin-project/lotus/blob/master/.circleci/config.yml#L302, would it maybe makes sense to trigger this in circle too?

coryschwartz:
@magik6k The benefit this has over CircleCI is just the comment it makes, so you can see the result of the tests in in the issue. It's just to make it easier when following the release checklist to see that tests are passing.

jennijuju:
@magik6k could you please take a look at this when your time permits?

.github/workflows/testground-release-issue.yml
```diff
@@ -0,0 +1,115 @@
...
+            plan_directory: testplans/lotus-soup
+            composition_file: testplans/lotus-soup/_compositions/baseline-k8s-3-3.toml
+          - backend_addr: ci.testground.ipfs.team
+            backend_proto: https
+            plan_directory: testplans/lotus-soup
+            composition_file: testplans/lotus-soup/_compositions/drand-outage-k8s.toml
+          - backend_addr: ci.testground.ipfs.team
+            backend_proto: https
+            plan_directory: testplans/lotus-soup
+            composition_file: testplans/lotus-soup/_compositions/deals-stress-concurrent-natural-k8s.toml
```
nonsense, line 73: @coryschwartz I suggest we have only two tests here:
- `baseline-k8s-3-1.toml` - 1 miner 3 clients
- `paych-stress-k8s.toml`

The rest are not stable, hit various bugs - either in tests or in Lotus, and at the moment should not be part of a release checklist.

.github/workflows/testground-release-issue.yml
```diff
@@ -0,0 +1,115 @@
...
+
+  testground-run:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - backend_addr: ci.testground.ipfs.team
+            backend_proto: https
```
nonsense, line 39: Graphsync testplan should not be part of the Lotus release I think.

Stebalien:
@jennijuju the call is yours here. We already test these in CI, the question is whether or not we want them tracked on the release issue itself. Given that we make a release _pr_, I think that's not really critical. But up to you.

jennijuju:
Will make a decision by Friday.

coryschwartz:
Thanks @jennijuju

jennijuju:
@coryschwartz thanks for the pr - I will close this for now but would definitely revisit this after we have a more stable and automated release process. 
With the process we have today, I think we'll add some overhead with trying to link all pr to a dedicated release issue so we can get the test results there. Also, the way we use release issue now is more for external communication. As we slowly fix all flakey tests and set the expectation to only merge PR and make a release when all tests pass + we dont have a high test ground test coverage atm - this comment in the release issue won't provide too much meaning info.

Issue #5647, opened 2021-02-21
To use a multi-signature wallet as the miner owner, use the command: lotus-miner actor control list | actor withdraw 0.001 err:owner f02222: error getting account key: failed to get account actor state for f02222: unknown actor code bafkqadtgnfwc6mrpnv2wy5djonuwo


upengs:
**Describe the bug**
To use a multi-signature wallet as the miner owner, use the command: lotus-miner actor control list | actor withdraw 0.001 err:

**To Reproduce**
1. Run 'lotus-miner actor withdraw 0.01'   'lotus-miner actor control list'
2. See error
```
owner f02222: error getting account key: failed to get account actor state for f02222: unknown actor code bafkqadtgnfwc6mrpnv2wy5djonuwo
```

**Expected behavior**
success
**Screenshots**

**Version (run `lotus version`):**

**Additional context**

arajasek:
Three options here:

- create a new CLI command to do this (easy, but somewhat unsustainable)
- add docs to explain how to do this (akin to what we have for the set-owner) (terrible UX)
- actually improve the UX of encoding params & sending arbitrary messages (not easy)

arajasek:
for now we might as well include docs akin to https://github.com/filecoin-project/lotus/pull/5342#issue-554009129

modytor:
I am waiting for this feature, and my customers are also very anxious. When will we release this feature?

jennijuju:
blocked by #1136

stuberman:
Would love to see progress on this! Needed by SPs @Reiers @ribasushi

rjan90:
Hey @stuberman! You can withdraw the balance with a msig following this guide https://lotus.filecoin.io/tutorials/lotus-miner/msig-as-owner/#withdraw-balance-with-multisig-owner

alvin-reyes:
We can wrap the steps above and introduce the following

```
lotus-miner actor propose --from=<multisignature member 1> <multisignature address> <minerid> 0 16 <params>
lotus-miner actor approve --from=<multisignature member 2> <multisignature address> <transactionId> <address of proposer of txnId> <minerid> 0 16 <params>
```

We set the encoding as `hex` by default but the user can specify `--encoding=base64` if an encoding needs to be enforced.

alvin-reyes:
Adding here the related slack discussion(s): 

- https://filecoinproject.slack.com/archives/CP50PPW2X/p1655153138579989

Issue #1136, opened 2020-01-22
Add created multisig addresses to local wallet store


whyrusleeping:
We should try to treat multisigs we create as 'accounts' as much as we can. There are a number of places where we can make the UX nice here:

- Include multisig addresses in the wallet
  - This means `lotus wallet list` shows multisig addresses
  - It also makes it easier to remember your addresses
- Allow `lotus send` to send from multisigs
- Allow a multisig wallet to be your 'default' address
  - We technically get this for free if we include them in the wallet, but this has implications

jennijuju:
@Kubuxu will work on a design and have some AI for later this week.

whyrusleeping:
This would be really great to do, improving how people use and interact with multisigs in lotus will have significant gains all across the board in terms of security and convenience of using the protocol.

Filecoin does make custodying funds safely pretty easy, but if we make the UX here better and *seamless*, i really think a lot of people will benefit.

jennijuju:
Uplifting the Priority to P1 - having a better multisig experience is important for the Filecoin network/ecosystem.

This is a great project for tributes to take on:
- [ ] Design the architecture ~ 1week
- [ ] Implementation ~1 week

stuberman:
Any updates to this? @Reiers @ribasushi  This is becoming more critical to SPs.

PR #5893, opened 2021-03-27
introduce message prototypes

vyzo:
This introduces message prototypes to applicable API endpoints, which
allows us to invert control of message sending and give the user a
chance to intervene with an interactive ui.

### DIFF
--- a/cli/services.go
+++ b/cli/services.go
@@ -22,22 +22,24 @@ import (
 //go:generate go run github.com/golang/mock/mockgen -destination=servicesmock_test.go -package=cli -self_package github.com/filecoin-project/lotus/cli . ServicesAPI
 
 type ServicesAPI interface {
+	FullNodeAPI() api.FullNode
+
 	GetBaseFee(ctx context.Context) (abi.TokenAmount, error)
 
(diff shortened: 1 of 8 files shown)


arajasek:
We should think about solutions to https://github.com/filecoin-project/lotus/issues/1136 (which is _quite_ desired, see #5647 as an example) as part of this.

Kubuxu:
@arajasek this is quite unrelated.

cli/services.go
```diff
@@ -102,15 +108,24 @@ type CheckInfo struct {
...
+		if err != nil {
+			return nil, xerrors.Errorf("mpool get nonce: %w", err)
+		}
+		prototype.Message.Nonce = nonce
+		prototype.ValidNonce = true
+	}
+
 	var outChecks [][]api.MessageCheckStatus
-	checks, err := s.api.MpoolCheckMessages(ctx, []*types.Message{prototype})
+	checks, err := s.api.MpoolCheckMessages(ctx, []*types.Message{&prototype.Message})
```
Kubuxu, line 122: I think it would be better to pass the prototype to MpoolCheckMessages

PR #5918, opened 2021-03-30
Fix Makefile for apigen and docs

Kubuxu:
Make it now fail with multithreading


magik6k:
Was fixed in https://github.com/filecoin-project/lotus/pull/5685

PR #5924, opened 2021-03-31
Util for miners to extend all v1 sectors

arajasek:
Run `lotus-miner sectors extend --v1-sectors`. 

- The `tolerance` flag can be passed to indicate what durations aren't "worth" extending. It defaults to one week, which means that sectors whose current lifetime's are within one week of the maximum possible lifetime will not be extended.
- The `expiration-cutoff` flag can be passed to skip sectors whose expiration is past a certain point from the current head. It defaults to infinity (no cutoff), but if, say, `28800` was specified, then only sectors expiring in the next 10 days would be extended (2880 epochs in 1 day).

- [x] Tested manually on butterfly (kinda)
- [x] Should be tested on a large mainnet miner

### DIFF
--- a/cmd/lotus-storage-miner/sectors.go
+++ b/cmd/lotus-storage-miner/sectors.go
@@ -434,11 +434,6 @@ var sectorsExtendCmd = &cli.Command{
 		&cli.StringFlag{},
 	},
 	Action: func(cctx *cli.Context) error {
-		nodeApi, closer, err := lcli.GetStorageMinerAPI(cctx)
-		if err != nil {
-			return err
-		}
-		defer closer()
 
 		api, nCloser, err := lcli.GetFullNodeAPI(cctx)
 		if err != nil {
@@ -447,9 +442,10 @@ var sectorsExtendCmd = &cli.Command{
 		defer nCloser()
 
 		ctx := lcli.ReqContext(cctx)
-		maddr, err := nodeApi.ActorAddress(ctx)
+
+		maddr, err := getActorAddress(ctx, cctx)
 		if err != nil {
-			return xerrors.Errorf("getting miner actor address: %w", err)
+			return err
 		}
 
 		var params []miner0.ExtendSectorExpirationParams
@@ -512,7 +508,6 @@ var sectorsExtendCmd = &cli.Command{
 					}
 				}
 
-				// TODO: Count added sectors, if it exceeds 10k, split messages
 				p := &miner0.ExtendSectorExpirationParams{}
 				scount := 0
 				for l, exts := range sectors {
@@ -578,8 +573,8 @@ var sectorsExtendCmd = &cli.Command{
 			return xerrors.Errorf("getting miner info: %w", err)
 		}
 
-		for _, p := range params {
-			sp, aerr := actors.SerializeParams(&p)
+		for i := range params {
+			sp, aerr := actors.SerializeParams(&params[i])
 			if aerr != nil {
 				return xerrors.Errorf("serializing params: %w", err)
 			}
@@ -836,12 +831,6 @@ var sectorsCapacityCollateralCmd = &cli.Command{
 	},
 	Action: func(cctx *cli.Context) error {
 
-		mApi, mCloser, err := lcli.GetStorageMinerAPI(cctx)
-		if err != nil {
-			return err
-		}
-		defer mCloser()
-
 		nApi, nCloser, err := lcli.GetFullNodeAPI(cctx)
 		if err != nil {
 			return err
@@ -850,7 +839,7 @@ var sectorsCapacityCollateralCmd = &cli.Command{
 
 		ctx := lcli.ReqContext(cctx)
 
-		maddr, err := mApi.ActorAddress(ctx)
+		maddr, err := getActorAddress(ctx, cctx)
 		if err != nil {
 			return err
 		}
--- a/cmd/lotus-storage-miner/sectors.go
+++ b/cmd/lotus-storage-miner/sectors.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/filecoin-project/lotus/chain/actors/builtin"
+
 	"github.com/docker/go-units"
 	"github.com/fatih/color"
 	"github.com/urfave/cli/v2"
@@ -422,7 +424,12 @@ var sectorsExtendCmd = &cli.Command{
 		&cli.Int64Flag{
 			Name:     "new-expiration",
 			Usage:    "new expiration epoch",
-			Required: true,
+			Required: false,
+		},
+		&cli.BoolFlag{
+			Name:     "v1-sectors",
+			Usage:    "renews all v1 sectors up to the maximum possible lifetime",
+			Required: false,
 		},
 		&cli.StringFlag{},
 	},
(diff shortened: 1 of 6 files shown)


cmd/lotus-storage-miner/sectors.go
```diff
@@ -440,69 +444,155 @@ var sectorsExtendCmd = &cli.Command{
...
+		var params []miner3.ExtendSectorExpirationParams
 
-		for i, s := range cctx.Args().Slice() {
-			id, err := strconv.ParseUint(s, 10, 64)
-			if err != nil {
-				return xerrors.Errorf("could not parse sector %d: %w", i, err)
+		if cctx.Bool("v1-sectors") {
+			extensions := map[miner.SectorLocation]map[abi.ChainEpoch][]uint64{}
+			// are given durations within a week?
+			closeEnough := func(a, b abi.ChainEpoch) bool {
```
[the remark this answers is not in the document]
Stebalien: Also, document.

cmd/lotus-storage-miner/sectors.go
```diff
@@ -440,69 +444,155 @@ var sectorsExtendCmd = &cli.Command{
...
 
-			p, err := api.StateSectorPartition(ctx, maddr, abi.SectorNumber(id), types.EmptyTSK)
+			sis, err := api.StateMinerActiveSectors(ctx, maddr, types.EmptyTSK)
 			if err != nil {
-				return xerrors.Errorf("getting sector location for sector %d: %w", id, err)
+				return xerrors.Errorf("getting miner sector infos: %w", err)
+			}
+
+			for _, si := range sis {
+				if si.SealProof < abi.RegisteredSealProof_StackedDrg2KiBV1_1 {
```
Stebalien, line 473: maybe invert and use a continue?

cmd/lotus-storage-miner/sectors.go
```diff
@@ -440,69 +444,155 @@ var sectorsExtendCmd = &cli.Command{
...
+
+					ml := builtin3.SealProofPoliciesV11[si.SealProof].SectorMaxLifetime
```
Stebalien, line 475: Let's move this into `chain/actors/policy` and make it a function over the version.

cmd/lotus-storage-miner/sectors.go
```diff
@@ -440,69 +444,155 @@ var sectorsExtendCmd = &cli.Command{
...
+					// if the sector's missing less than a week of its maximum possible lifetime, don't bother extending it
+					if closeEnough(si.Expiration-si.Activation, ml) {
+						continue
+					}
+
+					newExp := ml - (miner3.WPoStProvingPeriod * 2) + si.Activation
```
Stebalien, line 481: What's this?

cmd/lotus-storage-miner/sectors.go
```diff
@@ -440,69 +444,155 @@ var sectorsExtendCmd = &cli.Command{
...
-				return xerrors.Errorf("getting sector location for sector %d: %w", id, err)
+				return xerrors.Errorf("getting miner sector infos: %w", err)
+			}
+
+			for _, si := range sis {
+				if si.SealProof < abi.RegisteredSealProof_StackedDrg2KiBV1_1 {
+
+					ml := builtin3.SealProofPoliciesV11[si.SealProof].SectorMaxLifetime
+					// if the sector's missing less than a week of its maximum possible lifetime, don't bother extending it
+					if closeEnough(si.Expiration-si.Activation, ml) {
```
Stebalien, line 477: maybe don't bother extending if the sector doesn't expire for X days?

Stebalien: (configurable)

cmd/lotus-storage-miner/sectors.go
```diff
@@ -440,69 +444,155 @@ var sectorsExtendCmd = &cli.Command{
...
+						continue
+					}
+
+					newExp := ml - (miner3.WPoStProvingPeriod * 2) + si.Activation
+					p, err := api.StateSectorPartition(ctx, maddr, si.SectorNumber, types.EmptyTSK)
```
Stebalien, line 482: This is really inefficient (searches all partitions/deadlines). OK for a first pass, but this could be _very_ slow for some miners.

On the other hand, may be not... let's wait and see.

cmd/lotus-storage-miner/sectors.go
```diff
@@ -440,69 +444,155 @@ var sectorsExtendCmd = &cli.Command{
...
+						}
+					}
+				}
 			}
+			p := &miner3.ExtendSectorExpirationParams{}
+			scount := 0
+			for l, exts := range extensions {
+				for newExp, numbers := range exts {
+					scount += len(numbers)
+					if scount > miner.AddressedSectorsMax || len(p.Extensions) == miner3.DeclarationsMax {
```
Stebalien, line 517: nit: abstract over policy.

cmd/lotus-storage-miner/sectors.go
```diff
@@ -440,69 +444,155 @@ var sectorsExtendCmd = &cli.Command{
...
+							}
+						}
+
+						if !added {
+							es[newExp] = []uint64{uint64(si.SectorNumber)}
+						}
+					}
+				}
 			}
+			p := &miner3.ExtendSectorExpirationParams{}
```
Stebalien, line 512: why by reference? We're dereferencing everywhere.

cmd/lotus-storage-miner/sectors.go
```diff
@@ -440,69 +444,155 @@ var sectorsExtendCmd = &cli.Command{
...
-			return xerrors.Errorf("serializing params: %w", err)
+				if p == nil {
+					return xerrors.Errorf("sector %d not found in any partition", id)
+				}
+
+				sectors[*p] = append(sectors[*p], id)
+			}
+
+			p := &miner3.ExtendSectorExpirationParams{}
+			for l, numbers := range sectors {
```
Stebalien, line 558: Can we dedup this with the above loop? It's doing the exact same thing (except now we aren't sending multiple messages).

Stebalien (commented):
No blocking changes (fine to approve as-is) but we could make this slightly nicer.

cmd/lotus-storage-miner/sectors.go
arajasek: Added a comment, it sets the new expiration to 2 days under the theoretical maximum lifetime. We do the same when sealing sectors (not entirely sure why, but seems reasonable).

cmd/lotus-storage-miner/sectors.go
arajasek: Yeah, I tried it for the largest miners, and it is...not fast. Easy optimization is to only load the actor state once, but that is not gonna be the critical path. Alternative is to iterate over deadlines, and pre-compute one very large map of sectors -> locations, and then hit that up as needed.

This is a "needs to be run once" command, though, so eh?

cmd/lotus-storage-miner/sectors.go
arajasek: Added an expiration-cutoff option.

cmd/lotus-storage-miner/sectors.go
arajasek: Yes, but eh.

cmd/lotus-storage-miner/sectors.go
Stebalien: I _think_ it may have something to do with how long it might take a provecommit to land on-chain. But it's not likely to hurt.

cmd/lotus-storage-miner/sectors.go
Stebalien: Eh == we aren't sending multiple messages here so it'll fail in some cases (although I guess we can say that that's pretty unlikely).

Stebalien (approved):

magik6k (approved):

PR #5925, opened 2021-03-31
Wait for a while after SyncSubmitBlock

diwufeiwen:


whyrusleeping:
Could you describe the specific edge case that causes the problem this fixes? This is fairly fiddly code, with a good number of tests. And some tests that rely on mining blocks are failing here...

whyrusleeping:
(reading related issue now)

whyrusleeping:
Okay, so to my eyes, it seems that the problem here is that when we submit a block, GetBestMiningCandidate doesnt necessarily return a tipset with our newly mined block right away, so we trip the break condition on line 182, which then results in line 219 triggering and causing us to sleep for an entire block cycle? If we are already late-ish submitting our block, then this causes us to be even later for the next block. This is a problem, makes sense. 

I think this isnt the right fix though, Instead, we should probably fix GetBestMiningCandidate to have knowledge of the block we just submitted, and have it spit that back to us if its better than whatever ChainHead is showing us (this might happen for some reason if block processing is slow)

diwufeiwen:
> Could you describe the specific edge case that causes the problem this fixes? This is fairly fiddly code, with a good number of tests. And some tests that rely on mining blocks are failing here...

For example：
10:00:06.000   received MiningBase T01{... height: 1000, time: 10:00:00 ...}
Run mineOne  -> isWinner -> generate blockB{height: 1001, time: 10:30:00}
10:00:30.000  Run SyncSubmitBlock
10:00:30.100(about)  Entering a new cycle of mining,Run GetBestMiningCandidate  ---> Most likely not to receive the latest Tipset{height: 1001},Will still receive T01{height: 1000}  ---> L272 m.niceSleep(time.Duration(build.BlockDelaySecs) * time.Second)
10:01:00.100  Entering a new cycle of mining,Run GetBestMiningCandidate  ---> received MiningBase T02{... height: 1001, time: 10:30:00 ...}
... It looks like it takes at least two rounds to get on track

diwufeiwen:
> Okay, so to my eyes, it seems that the problem here is that when we submit a block, GetBestMiningCandidate doesnt necessarily return a tipset with our newly mined block right away, so we trip the break condition on line 182, which then results in line 219 triggering and causing us to sleep for an entire block cycle? If we are already late-ish submitting our block, then this causes us to be even later for the next block. This is a problem, makes sense.
> 
> I think this isnt the right fix though, Instead, we should probably fix GetBestMiningCandidate to have knowledge of the block we just submitted, and have it spit that back to us if its better than whatever ChainHead is showing us (this might happen for some reason if block processing is slow)

The reason you said is correct!

diwufeiwen:
> Could you describe the specific edge case that causes the problem this fixes? This is fairly fiddly code, with a good number of tests. And some tests that rely on mining blocks are failing here...

10:00:06.000   received MiningBase T01{... height: 1000, time: 10:00:00 ...}
Run mineOne  -> isWinner -> generate blockB{height: 1001, time: 10:30:00}
10:00:30.000  Run SyncSubmitBlock
10:00:30.100(about)  Entering a new cycle of mining,Run GetBestMiningCandidate  ---> Most likely not to receive the latest Tipset{height: 1001},Will still receive T01{height: 1000}  ---> L272 m.niceSleep(time.Duration(build.BlockDelaySecs) * time.Second)
10:01:00.100  Entering a new cycle of mining,Run GetBestMiningCandidate  ---> received MiningBase T02{... height: 1001, time: 10:30:00 ...}
... It looks like it takes at least two rounds to get on track

if wait 6s after PropagationDelaySecs, the process is:
10:00:30.000  Run SyncSubmitBlock
Sleep 6s
10:00:36.100(about)  Entering a new cycle of mining,Run GetBestMiningCandidate  ---> Most likely receive the latest Tipset{height: 1001,time: 10:30:00}
This seems logical, of course I may not understand this thoroughly,this is just a suggestion

whyrusleeping:
The propagation delay is really about *other* blocks coming in from the network, In general I like to have fewer random sleeps in code like this. Forcibly retaining knowledge of our own block seems like a better approach than an extra sleep, though both would have the same effect. 

cc @magik6k @Kubuxu

diwufeiwen:
> Forcibly retaining knowledge of our own block

Forcibly keeping our own blocks without waiting for a while(PropagationDelaySecs) may be problematic: If there are multiple blocks in the last cycle, even if a block is produced according to its own base, its weight is less than that including all bases, so it becomes an orphan block. I think the introduction of PropagationDelaySecs is quite reasonable

magik6k:
If you mined a block in parent epoch, according to consensus rules you have to include it in your parent tipset, otherwise you can be slashed ("parent-grinding fault") - when that happens, both the miner node, and the chain node will prevent you from sending that tipset to the network anyways -> https://github.com/filecoin-project/lotus/blob/master/chain/gen/slashfilter/slashfilter.go#L52

diwufeiwen:
> If you mined a block in parent epoch, according to consensus rules you have to include it in your parent tipset, otherwise you can be slashed ("parent-grinding fault") - when that happens, both the miner node, and the chain node will prevent you from sending that tipset to the network anyways

Yes, but the current problem is that it is impossible to continuously produce blocks. After SyncSubmitBlock, it will enter the next cycle and will not receive any blocks, including your own  in parent epoch, so waiting for 6s(PropagationDelaySecs) may be a reasonable choice.

jennijuju:
This needs a careful review from @magik6k & @whyrusleeping  - may take some time to get to it.

arajasek:
Closing this so related discussion can occur in https://github.com/filecoin-project/lotus/issues/10754

PR #5926, opened 2021-03-31
fix storage reservations

yaohcn:


magik6k (approved):
Thanks for catching this!

PR #5930, opened 2021-03-31
other network images

coryschwartz:
"1-click" images for other network images


coryschwartz:
This came up out of the need to spin up nerpanet nodes for dealbot testing.  With this PR, nerpanet, calibrationnet, mainnet, and butterfly all have public AMIs to enable speedy testing/dev

magik6k (approved):

PR #5943, opened 2021-04-02
fix 4857: show help for set-addrs

### LINKED ISSUE #4857: Show help menu when `lotus-miner actor set-addrs` is called without argument
Most lotus commands print a help screen if no argument is provided. The `lotus-miner actor set-addrs` does not, rather instantly drops a transaction on chain that sets the address to `nil`. I'd say this is probably not what a user wants to do :)

deaswang:
fix #4857

### DIFF
--- a/cmd/lotus-storage-miner/actor.go
+++ b/cmd/lotus-storage-miner/actor.go
@@ -53,11 +53,20 @@ var actorSetAddrsCmd = &cli.Command{
 			Usage: "set gas limit",
 			Value: 0,
 		},
+		&cli.BoolFlag{
+			Name:  "unset",
+			Usage: "unset address",
+			Value: false,
+		},
 	},
 	Action: func(cctx *cli.Context) error {
 		args := cctx.Args().Slice()
-		if len(args) == 0 {
+		unset := cctx.Bool("unset")
+		if len(args) == 0 && !unset {
 			return cli.ShowSubcommandHelp(cctx)
+		}
+		if len(args) > 0 && unset {
+			return fmt.Errorf("unset should with no arguments")
 		}
 
 		nodeAPI, closer, err := lcli.GetStorageMinerAPI(cctx)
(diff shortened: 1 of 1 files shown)


cmd/lotus-storage-miner/actor.go
```diff
@@ -70,7 +75,10 @@ var actorSetAddrsCmd = &cli.Command{
 		ctx := lcli.ReqContext(cctx)
 
 		var addrs []abi.Multiaddrs
-		for _, a := range cctx.Args().Slice() {
+		for _, a := range args {
```
arajasek, line 78: Is this to enable a way to unset all addresses? If so, I'd rather do the following:

- Add a `--unset` boolean flag
- If it's supplied with no arguments, unset the miner addresses (what this currently does)
- If it's supplied with arguments, fail 
- If it isn't supplied, and yet no arguments are provided, fail

arajasek (commented):
Thanks for taking the initiative! I have one question / suggestion.

cmd/lotus-storage-miner/actor.go
```diff
@@ -53,8 +53,22 @@ var actorSetAddrsCmd = &cli.Command{
...
+		},
 	},
 	Action: func(cctx *cli.Context) error {
+		args := cctx.Args().Slice()
+		unset := cctx.Bool("unset")
+		if len(args) == 0 && !unset {
+			return cli.ShowSubcommandHelp(cctx)
+		}
+		if len(args) > 0 && unset {
+			return fmt.Errorf("unset should with no arguments")
```
magik6k, line 69:
```suggestion
			return fmt.Errorf("unset can only be used with no arguments")
```

arajasek (approved):
Looks good with Magik's suggested change

magik6k (approved):

magik6k:
Thanks!

PR #5961, opened 2021-04-05
Introduce stateless offline dealflow, bypassing the FSM/deallists

ribasushi:
( read diff under `-w` )

This is aproposal for an additional flag --manual-stateless-deal and a
corresponding API endpoint ClientStatelessDeal. This allows firing off
an offline-style deal against a miner without keeping further track of
it locally.

Not keeping any local state introduces the limitation of requiring free
storage deals, as there is nothing to tie the payment channel setup to.

Rationale/need for this type of flow is the case of incredibly large
sets of data nd deals, where the client and providers have prearranged
payment ahead of time, and the client has a separate-from-lotus database
of deal inventory. This way the client can use their lotus node merely
as a network gateway, without running into any limitations currently
present in both lotus as a whole and go-fil-markets in particular.

Specific context for this work is filecoin-discover, where the requirement
is to onboard ~ 12,000,000 individual deals against a pool of miners
with whom the client has prearranged a relationship.


magik6k (commented):
Needs `make gen`

BigLep:
Are there any additional tests we should be adding?  During w3dt project leads meeting, I saw @raulk post that we should have a testground test that stresses this path.

ribasushi:
Mmmmm this codepath shuts off pretty much all of client lotus, reducing everything to a single fire-and-forget libp2p message: there's nothing to stress.

Stress testing would be required on the miner side, since we can't do the same "short circuit" there. But that's a way larger endeavor for which a lot of ignite-like domain knowledge is required.

That said, a basic test can't hurt, provided there is interest in this functionality in core lotus the first place.

pooja:
Hey team! Now that we're post-upgrade (congrats!!), wanted to check in on this PR again -- cc @magik6k @BigLep would you mind taking a look and merging if it looks ok?

node/impl/client/client.go
```diff
@@ -171,25 +188,110 @@ func (a *API) ClientStartDeal(ctx context.Context, params *api.StartDealParams)
...
+		return nil, xerrors.Errorf("failed to sign proposal : %w", err)
+	}
+
+	dealProposalSigned := &market.ClientDealProposal{
+		Proposal:        *dealProposal,
+		ClientSignature: *dealProposalSig,
+	}
+	dStream, err := network.NewFromLibp2pHost(a.Host,
+		network.RetryParameters(0, 0, 0, 0),
+	).NewDealStream(ctx, *mi.PeerId)
```
ribasushi, line 258: @magik6k @whyrusleeping: I am unsure how to configure lp2p here properly, so I left it bare. Suggestions welcome!

BigLep:
@pooja : we missed talking about this during triage.  It was a truncated session and Magik is also out this week and next.

Can this wait until his return or does it need to get done sooner?  I'm bummed that we have PRs open for so long (not good).  The team is also being asked to make Lotus 1.10 (with FIP8/13 integration) the top priority.

ribasushi:
@BigLep rebased/updated for master / apiv1

node/impl/client/client.go
ribasushi: @raulk perhaps you have input here? ☝️

cli/client.go
```diff
@@ -471,7 +475,18 @@ The minimum value is 518400 (6 months).`,
 			FastRetrieval:      cctx.Bool("fast-retrieval"),
 			VerifiedDeal:       isVerified,
 			ProviderCollateral: provCol,
-		})
+		}
+
+		var proposal *cid.Cid
+		if cctx.Bool("manual-stateless-deal") {
+			if ref.TransferType != storagemarket.TTManual {
```
arajasek, line 482: also enforce that the price is 0?

node/impl/client/client.go
arajasek: Not my lane, but `smnet.RetryParameters(time.Second, 5*time.Minute, 15, 5)` seems reasonable (see `marketsRetryParams` and `dtRetryParams`).

arajasek (commented):
Seems reasonable to me. Have we / can we manually test out the flow (assuming the testground test isn't happening)?

That's probably the easiest way to have enough confidence to ship.

cli/client.go
ribasushi: Whoops, fixed.

arajasek:
> > Have we / can we manually test out the flow
> 
> @arajasek I have been using this on mainnet / in production since I opened the PR, so it has about a month of mileage now.

That's all I need to hear :)

PR #5962, opened 2021-04-05
storagefsm: Apply global events even in broken states

### LINKED ISSUE #3982: Be able to remove sectors that are in invalid state.
This is for removing the sectors caused by #3626 

Sectors that are being updated to a invalid state via `updated-state` due to a typo in the new status/mis-usage, they cannot be removed or updated to a right state (both operation lead state machine shutdown).

It will be really nice if we can have a CLI to delete the sector and its data regardless the fsm.

magik6k:
Fixes #3982


arajasek (approved):

arajasek:
Give us tests, we crave tests

arajasek (approved):
Reapproved now that test craving has been satisfied

PR #5968, opened 2021-04-06
remove duplicate ask and calculate ping before lock

deaswang:


magik6k (commented):
The first `ClientQueryAsk` usually needs to start a new connection to the miner, which can be many network round trips, and will not give you accurate ping. The second call should use existing libp2p connection, just starting a new stream, and sending the query, which in theory is just 1 network roundtrip, which means that it's good enough to give us accurate ping to the miner.

Ideally we'd just use the libp2p ping protocol, but that would need to be exposed on the API first.

jennijuju:
> The first `ClientQueryAsk` usually needs to start a new connection to the miner, which can be many network round trips, and will not give you accurate ping. The second call should use existing libp2p connection, just starting a new stream, and sending the query, which in theory is just 1 network roundtrip, which means that it's good enough to give us accurate ping to the miner.
> 
> Ideally we'd just use the libp2p ping protocol, but that would need to be exposed on the API first.

@deaswang Could you please address the review?

deaswang:
Understand. And the `time.Now().Sub(rt)` before `lk.Lock()` should be more accurate.

magik6k (approved):

PR #5970, opened 2021-04-06
V1 API: Support api changes, cleanup msg wait/search api 

magik6k:
On top of https://github.com/filecoin-project/lotus/pull/5685

This PR:
* Makes the changes needed to be made across the codebase to actually support different API versions
  * CLI uses v0 api by default
  * The rest of the codebase should be using v1
* Collapses `State[Wait/Search]MsgLimited,StateGetReceipt` into `State[Wait/Search]Msg` whth added params which should make it much harder to use incorrectly


api/wrap.go
```diff
@@ -0,0 +1,32 @@
+package api
+
+import (
+	"reflect"
+)
+
+// Wrap adapts partial api impl to another version
+// proxyT is the proxy type used as input in wrapperT
+// Usage: Wrap(new(v1api.FullNodeStruct), new(v0api.WrapperV1Full), eventsApi).(EventAPI)
+func Wrap(proxyT, wrapperT, impl interface{}) interface{} {
```
Kubuxu, line 10: What does the `impl` do?

magik6k: It provides a partial implementation of the interface, which is used to populate the struct

Kubuxu (commented):
SGWM

arajasek (approved):
Ship it!

magik6k:
(merging into the v1 api branch, so things are in one place)

PR #5971, opened 2021-04-06
Convert the chainstore lock to RW

ribasushi:
High read-API concurrency facilitates multiple calls to GetHeaviestTipSet
which in turn could slow down chainsync.


magik6k (approved):
I'm not sure what can be calling this method so much to cause sync issues, but I guess this is making things very slightly safer (esp. given that Go RWLock is write-preferring)

PR #5974, opened 2021-04-06
Centralize everything on ipfs/go-log/v2

ribasushi:
I am not entirely sure this is right, but everything seems to build...

@Stebalien I didn't manage to find anything substantially different between v1 and v2, even the zap-frame-skip seems to be the same. Is there anything I am missing?

### DIFF
--- a/chain/beacon/beacon.go
+++ b/chain/beacon/beacon.go
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	"github.com/filecoin-project/go-state-types/abi"
-	logging "github.com/ipfs/go-log"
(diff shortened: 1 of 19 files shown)


magik6k (approved):

Stebalien:
> I didn't manage to find anything substantially different between v1 and v2, even the zap-frame-skip seems to be the same. Is there anything I am missing?

We got rid of tracing support.

ribasushi:
> We got rid of tracing support.

@Stebalien so does this mean this PR is good to go then...?

Stebalien (approved):

Stebalien:
Yep!

PR #5976, opened 2021-04-06
localstorage: Demote reservation stat error to debug

magik6k:
This can be a bit spammy, and can happen normally, for example when finalizing the sector:
* We'll reserve space for the sealed sector and cache, 
* We'll start fetching sealed data
* Cache will not start to be fetched until we fetch the sealed sector file
* Since that takes a while, we'll spam this error when periodically checking storage health / getting disk usage when scheduling tasks


arajasek (approved):
Needs tests.

PR #5983, opened 2021-04-07
Support standalone miner-market process

magik6k:
This PR adds support for running the miner market module as a separate process
* A lotus-miner market process connects to the "main" lotus process for all sealing/storage interactions
* It can pull data directly from workers
* It has it's own libp2p node

TODO:
- [ ] Fix api deps
  - (don't depend on ffi)
- [ ] Fix node tests
- [ ] Delegate market node auth to the main node (storage subsystem assumes that)
  - Put [API Secret](https://github.com/filecoin-project/lotus/blob/3fc23a785c2fa0d75fa8d737c635c01f9ae256fe/node/impl/common/common.go#L37) behind an interface, in DI replace the implementation by the main node API
- [ ] Figure out what to do about local market-node storage
  - [ ] Either drop completely (likely preferred initially)
    - This likely just means making the local storage implementation ignore paths, or (a bit cleaner) making it possible to run Remote store without a Local store
  - [ ] Or register storage properly (and handle disconnects, a bit like workers do)
- [ ] Use http-mode jsonrpc for talking to the main Sterage/Sealing node (for reliability)
- [ ] Return nicer errors from unsupported api calls when subsystems are disabled
  - Not sure if we can do much better than `if x == nil { error(module x is disabled) }`
- [ ] Don't construct miner Libp2p node when not needed
  - Mostly just shifting things around in the node builder
- [ ] Bump API versions which need bumping
  - Miner version at least
- [ ] Check API versions on reconnect
- [ ] Setup docs
- [ ] Write integration test
  - Copy one of the `api/test/deal.go` tests, and run in this setup


extern/sector-storage/stores/remote.go
```diff
@@ -293,6 +295,148 @@ func (r *Remote) fetch(ctx context.Context, url, outname string) error {
...
+
+		if !has {
+			if err := pf.Close(); err != nil {
+				return nil, xerrors.Errorf("close partial file: %w", err)
+			}
+
+			return nil, nil
+		}
+
+		return pf.Reader(storiface.PaddedByteIndex(offset), size)
```
magik6k, line 433: This part _technically_ should live on the Local store

magik6k:
Superseded by https://github.com/filecoin-project/lotus/pull/6356

PR #6003, opened 2021-04-09
Fix fallback chainstore

magik6k:
Fixes https://github.com/filecoin-project/lotus/issues/5975


arajasek (approved):
My node no longer crashes when doing this, so approved.

ribasushi (approved):
Let's 🚢

ribasushi:
@arajasek can we backport this to 0.7? it's fixing a regression arguably...

firesWu:
I use this PR and add env `LOTUS_ENABLE_CHAINSTORE_FALLBACK =1`
When I exec `delete-obj some cid` and restart lotus-daemon, lotus still report error  which is 'block not found'.
I fallback to 1.5.0 version, lotus sync normally.

PR #6012, opened 2021-04-11
Feature: start and stop lotus processes in systemd style.

PatHoo:
Feature: start and stop lotus processes in systemd style.


Stebalien:
Unfortunately, these configs are going to be very specific to the individual miner and don't generalize well.

PR #6021, opened 2021-04-12
Create codeql-analysis.yml

jfcg:
Add Github CodeQL security checks for Go.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -17,8 +17,6 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ master ]
-  schedule:
-    - cron: '37 14 * * 5'
 
 jobs:
   analyze:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go', 'javascript', 'python' ]
```
magik6k, line 31: We can drop python / js as those languages aren't used for anything beyond some dev scripts / tools

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '37 14 * * 5'
```
magik6k, line 21: Does it make sense to run this periodically if we already run this on all PR / master commits?

magik6k (commented):
Thanks for the PR.

This looks good, but I'd like to see what we can get from this, and after doing a couple of searches, only things I find is that this gives us "security checks" by analyzing the code, and makes it possible to run quires against the code.

Are there any concrete examples of what security checks this does, and how are they made visible to us? (just want to understand how this is useful to us before merging)

jfcg:
In my fork under `/security/code-scanning` I see:
lotus

You can see the alerts of the merge [here](https://github.com/filecoin-project/lotus/pull/6021/checks?check_run_id=2327208253)

magik6k (approved):
Looks good, thanks for the PR!

PR #6033, opened 2021-04-14
fix extend v1 sector

TerenceClark:


cmd/lotus-storage-miner/sectors.go
```diff
@@ -507,6 +507,9 @@ var sectorsExtendCmd = &cli.Command{
 
 				// Set the new expiration to 48 hours less than the theoretical maximum lifetime
 				newExp := ml - (miner3.WPoStProvingPeriod * 2) + si.Activation
+				if withinTolerance(si.Expiration, newExp) {
```
arajasek, line 510: This is somewhat redundant due to the check on line 504, but happy to allow it.

cmd/lotus-storage-miner/sectors.go
```diff
@@ -524,7 +527,8 @@ var sectorsExtendCmd = &cli.Command{
 				} else {
 					added := false
 					for exp := range es {
-						if withinTolerance(exp, newExp) {
+						// newExp < exp, can't extend
+						if withinTolerance(exp, newExp) && newExp >= exp {
```
arajasek, line 531: Nice catch, thank you! The optimal thing to do would be to repack the extensions, replacing `exp` with the smaller `newExp` in this case, but I don't think it's too important.

arajasek (approved):

arajasek:
Closing in favour of #6066

PR #6062, opened 2021-04-19
Allow external wallet to sign BlockHeader

diwufeiwen:
If my cluster deploys an independent wallet service, first judge whether there is a corresponding from when generating a block, and if not, let the wallet service sign.

### DIFF
--- a/chain/gen/mining.go
+++ b/chain/gen/mining.go
@@ -116,19 +116,26 @@ func MinerCreateBlock(ctx context.Context, sm *stmgr.StateManager, w api.Wallet,
 	}
 	next.ParentBaseFee = baseFee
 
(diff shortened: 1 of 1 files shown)


Stebalien (changes requested):
The code doesn't appear to do what the description says. It does not allow signing blocks with an external wallet (that _should_ already be possible with the `lotus-wallet` command and a normal external lotus wallet). Instead, the code seems to just _not_ sign blocks if they key can't be found.

I understand you probably have some form of setup where you have some node that sits between your miner and the network, signing blocks. But for every user without that setup, this code will just silently emit bad blocks when the key isn't available.

As-is, this isn't something we can accept into the core lotus implementation.

PR #6064, opened 2021-04-19
polish(adt_diff): optimize adt diff array and map logic

frrist:
- only walk the map/array if they have different root CID's

This will fail CI since calling `Root` on a hamt or amt node will cause a write to occur on the underlying blockstore. The blockstore used by all lotus tests is read-only, and thus this operation always fails.

### DIFF
--- a/chain/actors/adt/diff_adt.go
+++ b/chain/actors/adt/diff_adt.go
@@ -27,6 +27,18 @@ type AdtArrayDiff interface {
 // - All values that exist in preArr and in curArr are passed to AdtArrayDiff.Modify()
 //  - It is the responsibility of AdtArrayDiff.Modify() to determine if the values it was passed have been modified.
 func DiffAdtArray(preArr, curArr Array, out AdtArrayDiff) error {
+	preRoot, err := preArr.Root()
+	if err != nil {
+		return err
+	}
+	curRoot, err := curArr.Root()
+	if err != nil {
+		return err
+	}
+	if curRoot.Equals(preRoot) {
+		return nil
+	}
+
 	notNew := make(map[int64]struct{}, curArr.Length())
 	prevVal := new(typegen.Deferred)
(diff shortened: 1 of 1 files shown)


magik6k (commented):
Tests need fixing before this can be merged

frrist:
I'm not sure if this can be easily fixed. The tests are failing because of a write on a blockstore that doesn't support writes. Here is the code path that leads to the failure:

`StatePredicates` uses an `APIBlockstore` to access state:
https://github.com/filecoin-project/lotus/blob/2e6369012517972eb3f4f24f8a28e5a834cc66ad/chain/events/state/predicates.go#L37-L42

`APIBlockstore` is basically a read-only blockstore: 
https://github.com/filecoin-project/lotus/blob/2e6369012517972eb3f4f24f8a28e5a834cc66ad/blockstore/api.go#L52-L58

The calls to `Root` [added in this PR](https://github.com/filecoin-project/lotus/pull/6064/files#diff-cbc12e112e8452aed24afded90d0a85692852ee78b558e83459031e31988045cR30-R37) will ultimately call the AMT Node's `Flush()` method which will attempt to write the "new" root to the underlying store backing the AMT (the `APIBlockstore`):
https://github.com/filecoin-project/go-amt-ipld/blob/5e04cfcbd10549e8b6ef2f39dedf7a76b0474d1d/amt.go#L331-L345

These calls fail because writing to a read-only blockstore isn't allowed.

I suspect `StatePredicates` uses an `APIBlockstore` because it is intended to be used over the API for read-only state inspection. In order to include this optimization and keep the `StatePredicates` usable over the API, the (AMT wrapper) Array's `Root()` method needs to return a CID without writing to its underlying store...not sure that is even do-able.

frrist:
Landing this change looks to be complicated. Going to explore alternatives.

Issue #6027, opened 2021-04-13
[BUG]  sectors extend --v1-sectors bugs


benjaminh83:
## Edit (@jennijuju)

List of edge cases to handle in https://github.com/filecoin-project/lotus/blob/50b4ea30836618d199e24024f4f591de3dfbb516/cmd/lotus-storage-miner/sectors.go#L428:

- [ ] If the sector is expired by the the time the extension message lands on chain, the message will fail. Sectors that are legally allowed to be extended should be extended successfully, which is not the case right now as all sectors will fail extension.  example message bafy2bzaceawk4q2q52gr5ewouszkzl5plhp5lxkax6kk3axi4wkskqs24ricg
- [ ] For sectors that have deals and lifetime is over 540 days, we should not try to extend them.
- [ ] Is trying to extend the sector with an illegal extension duration 
       ```lotus-miner sectors extend --v1-sectors` exec with error below
ERROR: mpool push message: GasEstimateMessageGas error: estimating gas used: message execution failed: exit 16, reason: invalid expiration 1801153, total sector lifetime (1556194) cannot exceed 1555200 after activation 244959 (RetCode=16)```

**Describe the bug**
I was trying to run the command to extend my v1 sectors, as the network v11 was rolled out. I did the command: 

```
~$ lotus-miner sectors extend --v1-sectors
bafy2bzacear3hoscrxlkf3l6bfkahzijxwee2o7npgi3bdkuavghzpiv5h6gc
```
This did not get on chain when executed. 7h later, I restarted lotus-daemon and I also pushed some messages stuck in local mpool. It seems I now got the message on chain:
https://filfox.info/en/message/bafy2bzaceawk4q2q52gr5ewouszkzl5plhp5lxkax6kk3axi4wkskqs24ricg

**Two problems:** 
**1.** One sector expired in the meantime, before the message landed on chain, causing the whole batch message to be rejected. 
Status: ErrForbidden
Error Message: cannot extend expiration for expired sector 1131, expired at 665992, now 666145 (RetCode=18)
**2.** Some kind of overestimation happened on this batch deal, causing it to burn 2 FIL! 

**Version (run `lotus version`):**
```
Daemon:  1.6.0+mainnet+git.3fc23a785.dirty+api1.0.1
Local: lotus-miner version 1.6.0+mainnet+git.3fc23a785.dirty
```
Dirty because of GPU2...

flyworker:
lotus-miner sectors extend --v1-sectors
[...]

PR #6066, opened 2021-04-20
Fix bugs in sectors extend --v1-sectors

chadwick2143:
Fix bugs described in #6027 
Based on the work of #6033

### DIFF
--- a/cmd/lotus-storage-miner/sectors.go
+++ b/cmd/lotus-storage-miner/sectors.go
@@ -538,7 +538,7 @@ var sectorsExtendCmd = &cli.Command{
 				} else {
 					added := false
 					for exp := range es {
-						if withinTolerance(exp, newExp) && newExp > exp && exp > si.Expiration {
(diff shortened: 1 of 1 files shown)


cmd/lotus-storage-miner/sectors.go
```diff
@@ -524,7 +538,7 @@ var sectorsExtendCmd = &cli.Command{
 				} else {
 					added := false
 					for exp := range es {
-						if withinTolerance(exp, newExp) {
+						if withinTolerance(exp, newExp) && newExp > exp && exp > si.Expiration {
```
arajasek, line 541:
```suggestion
						if withinTolerance(exp, newExp) && newExp >= exp && exp > si.Expiration {
```

arajasek (changes requested):

arajasek (approved):
Thanks for the PR! I've made one necessary change and this is good to go.

arajasek:
> this could address #6180 too

It doesn't address it, but the solution to 6180 will share similar logic to the v1 sector extension logic.

PR #6076, opened 2021-04-21
wrap gateway api

coryschwartz:
This fixes a bug in which the gateway rpc api registered all the jsonrpc method for the FullNode and causes the gateway to hit nil pointer errors when one of those methods was called.


Stebalien:
This appears to have been fixed. The v0 wrapper is registered in:

https://github.com/filecoin-project/lotus/blob/b584f446619f7fc4b17073799eb46969e89a5a3e/gateway/handler.go#L28-L29

PR #6079, opened 2021-04-21
Support for checking Unsealing status

aarshkshah1992:
@magik6k 

- Merged origin/master into https://github.com/filecoin-project/lotus/pull/5983/commits/59d59d548281fc669a81c07811de1adb1ba4f9c0.

- Added Support to check Unsealing Status.

- Pulled fixes from https://github.com/filecoin-project/lotus/pull/5983/commits/cf0757b347412a0d31f58b4313c66d0776d34063.

- fmt. lint, gen etc..

However, how do we fix the test failures here ?


magik6k:
I think it's fine to drop bca702a 4ef03e5 from here as those changes are related to market node separation, but don't really do anything we need (and it should make it much easier to debug what's going on with DI)

PR #6085, opened 2021-04-21
set buildtype in nerpa and butterfly

coryschwartz:
Add "Build Type" to nerpanet and butterfly.

I am not sure if these were intentionally left out. If they were, we can just close this.

The impetus of this PR is to allow people doing tests on nerpanet to be able to see what network they are on by running `lotus version`

This includes all the networks in our docs https://network.filecoin.io/

### DIFF
--- a/build/version.go
+++ b/build/version.go
@@ -6,12 +6,14 @@ var CurrentCommit string
 var BuildType int
 
 const (
-	BuildDefault    = 0
-	BuildMainnet    = 0x1
-	Build2k         = 0x2
-	BuildDebug      = 0x3
-	BuildCalibnet   = 0x4
-	BuildInteropnet = 0x5
+	BuildDefault      = 0
+	BuildMainnet      = 0x1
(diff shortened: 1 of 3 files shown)


build/version.go
```diff
@@ -4,11 +4,13 @@ var CurrentCommit string
 var BuildType int
 
 const (
-	BuildDefault  = 0
-	BuildMainnet  = 0x1
-	Build2k       = 0x2
-	BuildDebug    = 0x3
-	BuildCalibnet = 0x4
```
magik6k, line 11: I'd keep this as explicit hex values - we're doing some bit-hacky things with these (in 2k/debug builds)

coryschwartz: hm, okay. in either case, they are type int, I think it's equivilent to use iota in this case. Will change back to explicit hex anyway.

arajasek (approved):

Stebalien (approved):

PR #6090, opened 2021-04-22
Add more debug logging for unsealing

dirkmc:
Adding some logging to help track down issue with retrieval unsealing


magik6k (approved):
Seems quite verbose, but doesn't happen that often on most nodes

dirkmc:
It is quite verbose - we're trying to figure out where things are going wrong when making retrievals, somewhere between the markets code and the sealing code

PR #6094, opened 2021-04-23
feat: add the ability to "pin" a head after setting it

Stebalien:
This works by simply rejecting reverts passed the pinned height, treating it as the new "zero".


api/api_full.go
```diff
@@ -110,6 +110,9 @@ type FullNode interface {
 	// between the two objects.
 	ChainStatObj(ctx context.Context, obj cid.Cid, base cid.Cid) (ObjStat, error) //perm:read
 
+	// ChainPinHead forcefully sets current chain head, forbidding forks passed the new head. Use with caution.
+	ChainPinHead(context.Context, types.TipSetKey) error //perm:admin
```
Stebalien, line 114: It's either this or modify ChainSetHead to take a pin parameter.

Stebalien (commented):
Note: this doesn't currently persist across restarts. IMO, that's probably fine, but it would be easy to make it persist if desired.

arajasek:
I don't think this adds any functionality not covered by "checkpointing" tipsets, unfortunately (see https://github.com/filecoin-project/lotus/blob/1ac0c9a926269c790950cdd79214c4c545aa896f/chain/sync.go#L1443 for impl).

The key limitation with the checkpointing functionality is that you can only sync things that you have already validated and are already in your main chain (the main chain restriction is side-steppable by doing a sethead first).

AFAIC this has the same key limitation, which makes me think it's...unhelpful?

Stebalien:
Are you saying we could just call `SetHead` then `SetCheckpoint` to do the same thing?

Stebalien:
Ok, apparently this is just totally not what we want.

PR #6122, opened 2021-04-28
upgrade `lotus-soup` testplans and reduce deals concurrency to a single miner

nonsense:
This PR is bumping up Go to 1.16.3 for all docker images used in Testground test plans.

It is also reduce concurrency for deals towards a single miner, until we fix the underlying issues with deals being blocked in `StorageDealAwaitingPreCommit` stage.

Screenshot 2021-04-28 at 14 03 35

### DIFF
--- a/testplans/lotus-soup/deals_e2e.go
+++ b/testplans/lotus-soup/deals_e2e.go
@@ -75,7 +75,9 @@ func dealsE2E(t *testkit.TestEnvironment) error {
 	// give some time to the miner, otherwise, we get errors like:
 	// deal errored deal failed: (State=26) error calling node: publishing deal: GasEstimateMessageGas
 	// error: estimating gas used: message execution failed: exit 19, reason: failed to lock balance: failed to lock client funds: not enough balance to lock for addr t0102: escrow balance 0 < locked 0 + required 640297000 (RetCode=19)
-	time.Sleep(50 * time.Second)
+	time.Sleep(40 * time.Second)
+
+	time.Sleep(time.Duration(t.GlobalSeq) * 5 * time.Second)
 
 	// generate 1600 bytes of random data
 	data := make([]byte, 5000000)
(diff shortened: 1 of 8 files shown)


testplans/lotus-soup/deals_e2e.go
```diff
@@ -75,7 +75,9 @@ func dealsE2E(t *testkit.TestEnvironment) error {
 	// give some time to the miner, otherwise, we get errors like:
 	// deal errored deal failed: (State=26) error calling node: publishing deal: GasEstimateMessageGas
 	// error: estimating gas used: message execution failed: exit 19, reason: failed to lock balance: failed to lock client funds: not enough balance to lock for addr t0102: escrow balance 0 < locked 0 + required 640297000 (RetCode=19)
-	time.Sleep(50 * time.Second)
+	time.Sleep(40 * time.Second)
+
+	time.Sleep(time.Duration(t.GlobalSeq) * 5 * time.Second)
```
nonsense, line 80: The idea here is that we give 5sec. in-between of each deal from various clients, which is certainly more than 1 block time.

dirkmc (approved):

testplans/lotus-soup/manifest.toml
```diff
@@ -9,8 +9,8 @@ enabled = true
 
 [builders."docker:go"]
 enabled = true
-build_base_image = "iptestground/oni-buildbase:v13-lotus"
-runtime_image = "iptestground/oni-runtime:v8-debug"
+build_base_image = "iptestground/oni-buildbase:v14-lotus"
+runtime_image = "iptestground/oni-runtime:v9-debug"
```
nonsense, line 13: I've upgraded docker images with latest `lotus` and `lotus-miner` binaries (aliased to `lll` and `lm` within the images), so that we can debug deadlocked devnets, when a deal ends up in the `StorageDealAwaitingPreCommit` state.

magik6k (approved):

0racl3z (commented):
Thx!

PR #5769, opened 2021-03-10
[WIP] FIP0013 integration for Actors v5

magik6k:
Currently this is pulling in https://github.com/filecoin-project/filecoin-ffi/pull/169, https://github.com/filecoin-project/go-paramfetch/pull/13 and https://github.com/filecoin-project/specs-actors/pull/1381, and implementing some plumbing logic for proof aggregation

TODO:
- [x] implement mock proof api to unblock some integration work
- [ ] Write some unit tests for CommitBatcher
- [ ] Write an integration test for commit batching
  - [ ] Test edge-cases
    - [ ] One porep fails
    - [ ] One commit expires? (deal start?)
    - [ ] Batch fails for whatever reason (e.g. out of gas)
- [x] Integrate nicer proof aggregation apis when they land in filecoin-ffi
- [x] Expose CommitBatcher in the miner api / cli (similar to termination batcher)
- [x] Test in pond
- [x] Switch to v5 actors when ready in specs-actors

PR #6128, opened 2021-04-28
retrieval: Direct piecereader

magik6k:
Cherry-picks parts of https://github.com/filecoin-project/lotus/pull/5769, and makes them work

Did some testing in a pond devnet with a worker, it seems to work, but needs testing on a real node, and ideally some integration tests


extern/sector-storage/fr32/readers.go
```diff
@@ -51,13 +51,13 @@ func (r *unpadReader) Read(out []byte) (int, error) {
 
 	r.left -= uint64(todo)
 
-	n, err := r.src.Read(r.work[:todo])
+	n, err := io.ReadAtLeast(r.src, r.work[:todo], int(todo))
 	if err != nil && err != io.EOF {
 		return n, err
 	}
 
 	if n != int(todo) {
```
dirkmc, line 59: Should this be
```
if n < int(todo)
```

Although the check may not even be necessary. Looking at the docs for ReadAtLeast it seems like it will return an error if it doesn't read at least `todo` bytes:
```
// ReadAtLeast reads from r into buf until it has read at least min bytes.
// It returns the number of bytes copied and an error if fewer bytes were read.
// The error is EOF only if no bytes were read.
// If an EOF happens after reading fewer than min bytes,
// ReadAtLeast returns ErrUnexpectedEOF.
// If min is greater than the length of buf, ReadAtLeast returns ErrShortBuffer.
// On return, n >= min if and only if err == nil.
// If r returns an error having read at least min bytes, the error is dropped.
```

markets/retrievaladapter/provider.go
```diff
@@ -64,27 +66,21 @@ func (rpn *retrievalProviderNode) UnsealSector(ctx context.Context, sectorID abi
...
-		// Read the piece into the pipe's writer, unsealing the piece if necessary
-		log.Debugf("read piece in sector %d, offset %d, length %d from miner %d", sectorID, offset, length, mid)
-		err := rpn.sealer.ReadPiece(ctx, w, ref, storiface.UnpaddedByteIndex(offset), length, si.TicketValue, commD)
-		if err != nil {
-			log.Errorf("failed to unseal piece from sector %d: %s", sectorID, err)
-		}
-		// Close the reader with any error that was returned while reading the piece
-		_ = w.CloseWithError(err)
-	}()
+	// Read the piece into the pipe's writer, unsealing the piece if necessary
```
dirkmc, line 77:
```suggestion
	// Get a reader for the piece, unsealing the piece if necessary
```

extern/sector-storage/stores/remote.go
```diff
@@ -293,6 +295,148 @@ func (r *Remote) fetch(ctx context.Context, url, outname string) error {
...
+
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
+		resp.Body.Close() // nolint
+		return nil, xerrors.Errorf("non-200 code: %d", resp.StatusCode)
+	}
+
+	return resp.Body, nil
+}
+
+// Reated gets a reader for unsealed file range. Can return nil in case the requested range isn't allocated in the file
```
dirkmc, line 360:
```suggestion
// Reader gets a reader for unsealed file range. Can return nil in case the requested range isn't allocated in the file
```

dirkmc (commented):
Looks good, just one question

dirkmc (approved):

dirkmc:
When we tried to run this with our miner we got the following error when doing a retrieval:
```
2021-05-04T18:32:57.675+0300    INFO    markets loggers/loggers.go:30   retrieval provider event        {"name": "ProviderEventCancelComplete", "deal ID": "1620120981223977232", "receiver": "12D3KooWHMGTcNbS64arHyaPoUv6NrZi9V63kuVbnE6jbBjm5EUS", "state": "DealStatusErrored", "message": "failed to unseal piece from sector 647: opening partial file: file '/root/storage_sealing/unsealed/s-t0127896-647' has inconsistent length; has 34359738377 bytes; expected 2057 (9 trailer, 2048 sector data)"}
```

magik6k:
^ likely caused by not setting the seal proof type in the http handler / remote store on either the worker or miner node

aarshkshah1992:
@magik6k Thanks for that pointer. I'll grok this PR, factor in your suggestion and create a new PR to take this work to completion,

magik6k:
> Btw, do you have any pointers to existing integration tests that we can use as reference to write integration tests for this feature ?

Api tests which do retrieval (like [this one](https://github.com/filecoin-project/lotus/blob/72fc16e8a794f6d552c6002e4159820362fe62a9/api/test/deals.go#L440)) should be exercising most of the now code here (I'd run them with coverage to see if that's actually the case).

Testing the remote parts / new http code is a bit more complex as it requires setting up workers in the test framework, which is currently not there, and will probably require refactoring the worker constructor to be usable in tests (`cmd/lotus-seal-worker/main.go`), possibly by making it use the DI framework (I could help with this - if not, having it be non-DI based for now is also fine, the worker isn't that complicated)

magik6k:
(for worker tests, basically we'd need to extract the whole worker thing from the `run` command function, and make it possible to provide things which are now taken from `cctx` (miner API, miner store address, temp repo) from a test)

magik6k:
I think this is the issue:
* UnsealSector is getting sector info from lower-level `secb.SectorsStatus` (storage-fsm), which now returns api.SectorInfo (so that the interface matches the RPC interface for when we separate the market node)
* SectorInfo.SealProof is an optional field filled out in higher-level RPC impl, set to zero in storage-fsm -> https://github.com/filecoin-project/lotus/pull/6128/files#diff-0dd21b9cbecc38ac24b8e03f24c12f67fe8e5620af9bcca340a647017897135dR80
* We then pass that zero-value in retrieval adapter -> https://github.com/filecoin-project/lotus/pull/6128/files#diff-3b8c5ae92101b90e9b36a390ea8bcd193b7d519a667006a3d514ac163a76afaeR69
* That value eventually gets used in the http_handler to get the sector size -> https://github.com/filecoin-project/lotus/pull/6128/files#diff-578f861bfe2bd7f0ffb0f6b853915dd65fb4ffdc0dcf0c2825948010fcb321e3R166 (sealProof=0 is old 2k sectors, which is why this worked for me in pond..)

I think the correct thing to do here is to just set the proof type in the fsm adapter from `info.SectorType` (which is the first field set after sectors are created, so it should be set no matter at which stage of the sealing pipeline sectors are)

magik6k:
Superseded by #6280

PR #6142, opened 2021-04-29
fix: use a consistent tipset in commands

Stebalien:
It's very easy to write an incorrect command that operates over different heads by using the "empty" tipset. This change makes the `LoadTipSet` command helper get the latest head from the lotus daemon if its unset.

The cost is an extra call to get the head. That should be trivial in most cases.

### DIFF
--- a/cli/chain.go
+++ b/cli/chain.go
@@ -723,12 +723,6 @@ var ChainGetCmd = &cli.Command{
 				return err
 			}
 
-			if ts == nil {
-				ts, err = api.ChainHead(ctx)
-				if err != nil {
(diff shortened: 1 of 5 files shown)


cli/chain.go
```diff
@@ -723,12 +723,6 @@ var ChainGetCmd = &cli.Command{
 				return err
 			}
 
-			if ts == nil {
```
Stebalien, line 726: We now _always_ do this.

magik6k (approved):

arajasek (approved):

PR #2467, opened 2020-07-18
Add script to generate lotus cli document.

MitchellSoo:
Generate lotus command lines documents as text and markdown in folder "lotus/documentation/en".

PR #6145, opened 2021-04-29
Generate CLI docs

magik6k:
Based on https://github.com/filecoin-project/lotus/pull/2467

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -368,7 +368,13 @@ docsgen-openrpc-worker: docsgen-openrpc-bin
 .PHONY: docsgen docsgen-md-bin docsgen-openrpc-bin
 
 gen: type-gen method-gen docsgen api-gen
+	@echo ">>> IF YOU'VE MODIFIED THE CLI, REMEMBER TO ALSO MAKE docsgen-cli"
 .PHONY: gen
+
+# separate from gen because it needs binaries
+docsgen-cli: lotus lotus-miner lotus-worker
+	python ./scripts/generate-lotus-cli.py
+.PHONY: docsgen-cli
 
 print-%:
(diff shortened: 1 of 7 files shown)


documentation/en/cli-lotus.md
```diff
@@ -0,0 +1,2659 @@
+# lotus
+```
+NAME:
+   lotus - Filecoin decentralized storage network client
+
+USAGE:
+   lotus [global options] command [command options] [arguments...]
+
+VERSION:
+   1.11.0-dev+mainnet
```
arajasek, line 10: I'd also drop the `buildType()` from `UserVersion()` here if LOTUS_VERSION_IGNORE_COMMIT is set.

Makefile
```diff
@@ -368,7 +368,13 @@ docsgen-openrpc-worker: docsgen-openrpc-bin
 .PHONY: docsgen docsgen-md-bin docsgen-openrpc-bin
 
 gen: type-gen method-gen docsgen api-gen
+	@echo ">>> IF YOU'VE MODIFIED THE CLI, REMEMBER TO ALSO MAKE docsgen-cli"
```
arajasek, line 371: Is doscgen-cli slow? Why not just always run it?

arajasek (approved):
LGTM, 2 non-blockers.

Makefile
magik6k: It's not in gen because I'm not sure how to force make to execute one target after another when running in parallel without making one target depend on another - docsgen-cli depends on lotus binaries, which can't build if we're in the middle of generating api code, but we don't want the lotus build to depend on gen

PR #6148, opened 2021-04-30
mpool: Cleanup pre-nv12 selection logic

magik6k:


arajasek (approved):
I'm kinda tempted to leave the mute logic (as a no-op for now), but up to you.

PR #6164, opened 2021-04-30
drand: fix beacon cache

magik6k:


arajasek (approved):

chain/beacon/drand/drand.go
```diff
@@ -168,8 +168,8 @@ func (db *DrandBeacon) getCachedValue(round uint64) *types.BeaconEntry {
 	if !ok {
 		return nil
 	}
-	e, _ := v.(*types.BeaconEntry)
-	return e
+	e, _ := v.(types.BeaconEntry)
```
Kubuxu, line 171: Maybe switch to storying pointers in the cache. Now we are pulling out a pointer (`interface{}`) out of the cache, dereferencing it to struct, and then we take a pointer of that struct which escapes causing allocation.

PR #6169, opened 2021-05-03
Add a shed util to count miners by post type

arajasek:


ZenGround0 (approved):
General observation: we could group together utils that want direct datastore access for speed.  Maybe a `lotus-basement` to go with the `lotus-shed`? The active cron util I just wrote should do it this way instead.

PR #6175, opened 2021-05-03
Dynamic Retrieval pricing

aarshkshah1992:
Implement Dynamic Retrieval Pricing.

This PR supports two modes of dynamically pricing retrieval deals:

1. Lotus will ship with a "default" pricing policy that will:
 - Use the ask configured in the Ask Store(set via `lotus retrieval-deals set-ask`) to price deals.
 - However, it will NOT charge for Unsealing if we already have an Unsealed sector we can serve the retrieval from.
 - It will NOT charge for the data transfer if there exists a verified storage deal for the payload being retrieved. This behaviour can be turned off by setting the `VerifiedDealsFreeTransfer` flag to false in the config.
 
2. Users can configure an external shell script(similar to the deal filter mechanism) that  takes a json marshalled [PricingInput](https://github.com/filecoin-project/go-fil-markets/blob/d32334ae625914d4c05bb602a80a53a29c2967bc/retrievalmarket/types.go#L387) as the input and outputs a json marshalled Ask.  One of the input parameters is the ask configured in the Ask Store via the  retrieval `set-ask` CLI command. To use this mode instead of the default above, the `UseDynamicRetrievalPricing` flag needs to be set to true in the config.

Depends on https://github.com/filecoin-project/go-fil-markets/pull/542.

#### Mainnet
- [x] Default pricing: Transfer price is zero for retrieval of payload contained in a verified storage deal. 
- [x] Default pricing: Transfer price is quoted from the ask store for retrieval of payload NOT contained in a verified storage deal.
- [x] Default pricing: Change miner's ask in the `AskStore` using `set-ask` and then the correct transfer price is quoted from the ask store for retrieval of payload NOT contained in a verified storage deal.
- [x] Default pricing: Change the default pricing config to charge data-transfer fee for Verified deals as well and then the transfer price is also quoted from the ask store for retrieval of a payload contained in a verified storage deal.
- [x] Default pricing: Do not charge for Unsealing if we have already have an Unsealed sector file containing the payload.
- [ ] External pricing script:  Quote prices as expected for verified and unverified deals.

#### Devnet
- [x] External pricing script: Quotes prices as expected based on an externally configured script using the ask configured in the ask-store and deal pricing params such as Sealed/Unsealed deal etc.
- [x] Default pricing: Charge for Unsealing if we do NOT have an Unsealed sector file for the payload.

### TODO 
- [x] Merge https://github.com/filecoin-project/go-fil-markets/pull/542 and update deps.

Note: This Lotus PR depends on `go-fil-markets` [v1.5.0](https://github.com/filecoin-project/go-fil-markets/releases/tag/v1.5.0) which is the Markets version that contains the[ changes needed for ](https://github.com/filecoin-project/go-fil-markets/commit/e1a462e9ffd15e273f21cfb5ecb0ed29b64168e0) Dynamic Retrieval Pricing.


cmd/lotus-storage-miner/retrieval-deals.go
```diff
@@ -160,80 +155,8 @@ var retrievalDealsListCmd = &cli.Command{
...
-			}
-			ask.PaymentIntervalIncrease = uint64(v)
-		}
-
-		return api.MarketSetRetrievalAsk(ctx, ask)
-	},
-}
-
-var retrievalGetAskCmd = &cli.Command{
+// TODO FIX THIS
```
raulk, line 158: If this PR is incomplete, better mark it as draft ;-)

cmd/lotus-storage-miner/retrieval-deals.go
```diff
@@ -160,80 +155,8 @@ var retrievalDealsListCmd = &cli.Command{
 	},
 }
 
-var retrievalSetAskCmd = &cli.Command{
```
raulk, line 163: Why are we removing this command? Shouldn't this retrieval ask be the fallback price? It could be sent to the pricing strategy, or the pricing strategy could return a boolean saying whether it has calculated a price, or if it wants to fall back to the fixed price.

markets/pricing/cli.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func CliRetrievalPricingFunc(cmd string) dtypes.RetrievalPricingFunc {
+	return func(ctx context.Context, pricingInput retrievalmarket.DealPricingParams) (retrievalmarket.Ask, error) {
+		return runPricingFunc(ctx, cmd, pricingInput)
+	}
+}
+
+func runPricingFunc(_ context.Context, cmd string, params interface{}) (retrievalmarket.Ask, error) {
+	j, err := json.MarshalIndent(params, "", "  ")
```
raulk, line 21: Why MarshalIndent? A single-line JSON would be easier to process and debug.

markets/retrievaladapter/provider.go
```diff
@@ -38,6 +38,50 @@ func NewRetrievalProviderNode(maddr dtypes.MinerAddress, secb sectorblocks.Secto
 	return &retrievalProviderNode{address.Address(maddr), secb, pp, full}
 }
 
+func (rpn *retrievalProviderNode) GetDealPricingParams(ctx context.Context, storageDeals []abi.DealID) (retrievalmarket.DealPricingParams, error) {
+	resp := retrievalmarket.DealPricingParams{}
```
raulk, line 42: Can we pack as much data as possible as input to the retrieval pricing function? PieceCID, PieceSize, at least.

markets/retrievaladapter/provider.go
```diff
@@ -38,6 +38,50 @@ func NewRetrievalProviderNode(maddr dtypes.MinerAddress, secb sectorblocks.Secto
 	return &retrievalProviderNode{address.Address(maddr), secb, pp, full}
 }
 
+func (rpn *retrievalProviderNode) GetDealPricingParams(ctx context.Context, storageDeals []abi.DealID) (retrievalmarket.DealPricingParams, error) {
```
raulk, line 41:
```suggestion
func (rpn *retrievalProviderNode) GetRetrievalPricingInput(ctx context.Context, storageDeals []abi.DealID) (retrievalmarket.RetrievalPricingInput, error) {
```

node/config/def.go
```diff
@@ -66,6 +66,20 @@ type DealmakingConfig struct {
...
+	DynamicRetrievalPricingFunc string
+
+	DefaultRetrievalPricingParams DefaultRetrievalPricingParams
+}
+
+type DefaultRetrievalPricingParams struct {
+	PricePerByte            types.FIL
+	UnsealPrice             types.FIL
+	PaymentInterval         uint64
+	PaymentIntervalIncrease uint64
```
raulk, line 82: Let's keep the set retrieval ask command and remove this from configuration. We can't break that interface. 

That allows for setting a "fixed" ask for when:

1. clients are querying the retrieval ask (get retrieval ask)
2. a dynamic pricing function (if enabled) refuses to calculate a dynamic price, or it errors

And miners will want to adjust the fixed price at runtime, so having these in configuration removes flexibility.

raulk (changes requested):

node/config/def.go
```diff
@@ -66,6 +66,20 @@ type DealmakingConfig struct {
 
 	Filter          string
 	RetrievalFilter string
+
+	// If UseDynamicRetrievalPricing is set to false(default), we will use the DefaultRetrievalPricingParams
+	// for pricing retrieval deals, otherwise we will use the DynamicRetrievalPricingFunc configured here.
+	UseDynamicRetrievalPricing  bool
+	DynamicRetrievalPricingFunc string
```
dirkmc, line 73: Suggestion: if DynamicRetrievalPricingFunc is the empty string then we use the DefaultRetrievalPricingParams.
That way we only need one variable.

raulk (commented):
Follow-up comment to clarify what I think the UX should be here.

I think we need to support three policies for dynamic retrieval pricing.

1. `none` => no dynamic pricer is configured, all retrievals use the fixed ask, configured via set-ask.
2. `default` => (default we ship with) makes unsealed pieces (potentially) free to retrieve. We need to provide configuration values here because miners should be able to charge for data transfer if they want to (they just would rank lower in the reputation system)
3. `external` => an external script.

What the retrieval protocol would do is:

1. If `none` is configured, no changes to current behaviour.
2. If `default` is configured, call the default pricing function (which has been instantiated with some configuration values, maybe). The function should return a tuple `(price: float, dynamic: bool)`.
    * if `dynamic==true`, just return the `price`.
    * if `dynamic==false`, the pricing policy has refused to calculate a price (e.g. the piece is sealed), and we fall back to the configured fixed ask.
    * if it errors, we fall back to the fixed ask.
3. If `external` is configured, call the script and use the same logic above.

raulk:
After chatting with @aarshkshah1992, we can simplify this.

* retrieval pricing function is always a **pure function** that receives the relevant storage deal properties, (including whether it was a verified deal), sector status (sealed/unsealed), **and** the current configured ask, as input.
* there are two retrieval pricing function types:
    * default
    * external (script)
* the default/built-in retrieval pricing function sets the unseal price to 0 if the sector is unsealed (this is actually a bugfix).
	* it takes a configuration value `verified_deals_free_transfer=true/false` to decide whether it provides free retrieval for verified deals or not.
		* if true and the deal was verified => price per byte == 0.
		* if false => price per byte is taken from the ask.

cmd/lotus-storage-miner/retrieval-deals.go
```diff
@@ -234,8 +234,9 @@ var retrievalSetAskCmd = &cli.Command{
 }
 
 var retrievalGetAskCmd = &cli.Command{
-	Name:  "get-ask",
-	Usage: "Get the provider's current retrieval ask",
+	Name: "get-ask",
+	Usage: "Get the current retrieval ask configured by the provider in the ask-store. No-op if user has configured an external script to " +
+		"dynamically price retrieval deals.",
```
raulk, line 239: Why a no-op in this case? The external script should also receive the configured ask, that way the system is symmetric in all its possible pricing function strategies.

markets/pricing/cli.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"encoding/json"
+	"os/exec"
+
+	"github.com/filecoin-project/go-fil-markets/retrievalmarket"
+	"github.com/filecoin-project/lotus/node/modules/dtypes"
+	"golang.org/x/xerrors"
+)
+
+func CliRetrievalPricingFunc(cmd string) dtypes.RetrievalPricingFunc {
```
raulk, line 14:
```suggestion
func ExternalRetrievalPricingFunc(cmd string) dtypes.RetrievalPricingFunc {
```

markets/pricing/cli.go
```diff
@@ -0,0 +1,48 @@
...
+		resp := retrievalmarket.Ask{}
+
+		if err := json.Unmarshal(bz, &resp); err != nil {
+			return resp, xerrors.Errorf("failed to parse pricing output %s, err=%s", string(bz), err)
+		}
+		return resp, nil
+	case *exec.ExitError:
+		return retrievalmarket.Ask{}, xerrors.Errorf("pricing func exited with error: %s", errb.String())
+	default:
+		return retrievalmarket.Ask{}, xerrors.Errorf("filter cmd run error:%s", err)
```
raulk, line 46:
```suggestion
		return retrievalmarket.Ask{}, xerrors.Errorf("pricing func cmd run error: %s", err)
```

markets/pricing/cli.go
```diff
@@ -0,0 +1,48 @@
...
+func runPricingFunc(_ context.Context, cmd string, params interface{}) (retrievalmarket.Ask, error) {
+	j, err := json.Marshal(params)
+	if err != nil {
+		return retrievalmarket.Ask{}, err
+	}
+
+	var out bytes.Buffer
+	var errb bytes.Buffer
+
+	c := exec.Command("sh", "-c", cmd)
```
raulk, line 29: I wonder if we should do payload validation here to defend against potential injection attacks. It's possible that the input is already validated somewhere previous, though. My concern is that we're taking untrusted input (user provided input) and piping it to an shell job...

markets/retrievaladapter/provider.go
```diff
@@ -38,6 +38,53 @@ func NewRetrievalProviderNode(maddr dtypes.MinerAddress, secb sectorblocks.Secto
...
+			return resp, xerrors.Errorf("failed to look up deals on chain: %w", err)
+		}
+		if ds.Proposal.VerifiedDeal {
+			resp.VerifiedDeal = true
+		}
+
+		if ds.Proposal.PieceCID.Equals(pieceCID) {
+			resp.PieceSize = ds.Proposal.PieceSize.Unpadded()
+		}
+	}
```
raulk, line 61: I'm not fully grasping this logic. This takes a list of deals, and scans through them, never breaking out of the loop. I think this leads to inconsistent data, for example, if we have 3 deals [0..2]:

1. If 1 was verified, but 2 isn't, we will run with verified=true.
2. If 1 didn't match the piece CID, but 2 does, we will take the piece size of 2.

This seems a bit off.

node/config/def.go
```diff
@@ -66,6 +66,21 @@ type DealmakingConfig struct {
 
 	Filter          string
 	RetrievalFilter string
+
+	// If `UseDynamicRetrievalPricing` is set to true(default is false), we will use the `DynamicRetrievalPricingFunc` script
+	// to dynamically price retrieval deals.
+	UseDynamicRetrievalPricing  bool
```
raulk, line 72: Suggestion:


```go
RetrievalPricing *RetrievalPricing

type RetrievalPricing struct {
    Strategy string // possible values: "default", "external"

    External *RetrievalPricingExternal
    Default  *RetrievalPricingDefault
}

type RetrievalPricingExternal struct {
    Path string
}

type RetrievalPricingDefault struct {
    VerifiedDealsFreeTransfer bool
}
```

This would lead to this configuration:

```toml
[DealmakingConfig]
RetrievalPricing.Strategy = "default"
RetrievalPricing.Default.VerifiedDealsFreeTransfer = true
```

or

```toml
[DealmakingConfig]
RetrievalPricing.Strategy = "external"
RetrievalPricing.External.Path = "/path/to/script"
```

node/modules/storageminer.go
```diff
@@ -631,6 +633,30 @@ func RetrievalDealFilter(userFilter dtypes.RetrievalDealFilter) func(onlineOk dt
...
+			ask.UnsealPrice = big.Zero()
+		}
+
+		// don't charge for data transfer for verified deals if it's been configured to do so.
+		if pricingInput.VerifiedDeal && params.VerifiedDealsFreeTransfer {
+			ask.PricePerByte = big.Zero()
+		}
+
+		return ask, nil
+	}
```
raulk, line 657: Better to put this function in go-fil-markets, and reference it here.

markets/retrievaladapter/provider.go
aarshkshah1992: @raulk 

1. These are storage deals made for the `payloadCID` the retrieval client is looking for.
2. For the `VerifiedDeal ` flag in the `PricingInput`, I am looking to answer the question "does there exist any verified storage deal for this `payloadCID` " ?
3. Now, as per the current data model in markets, a `payloadCID` can belong to many Pieces and each Piece can correspond to multiple storage deals for that Piece, each deal being sealed in a different sector. We first select a Piece we want to retrieve the payload from (prioritising a Piece which has a storage deal on an Unsealed sector) and then pass the PieceCID to this function. Now, since the storage deals could be for different pieces(with these different pieces containing the same `payloadCID`), I wanna match the PieceCIDs to then determine the PieceSize for the retrieval deal. 

Not breaking the loop here after we've determined both pieces of information is a mistake, that can be fixed. Does this make sense ?

aarshkshah1992: cc @dirkmc for the Markets context.

node/config/def.go
aarshkshah1992: This is no longer needed after the config refactor mentioned above.

extern/sector-storage/stores/remote.go
```diff
@@ -357,6 +357,67 @@ func (r *Remote) readRemote(ctx context.Context, url string, spt abi.RegisteredS
...
+		return false, xerrors.Errorf("CheckAllocated only supports unsealed files")
+	}
+
+	paths, _, err := r.local.AcquireSector(ctx, s, ft, storiface.FTNone, storiface.PathStorage, storiface.AcquireMove)
+	if err != nil {
+		return false, xerrors.Errorf("acquire local: %w", err)
+	}
+
+	path := storiface.PathByType(paths, ft)
+	if path == "" {
```
nonsense, line 371: When is `path == ""` ? This function always returns and string based on the `ft`, no?

extern/sector-storage/stores/remote.go
```diff
@@ -357,6 +357,67 @@ func (r *Remote) readRemote(ctx context.Context, url string, spt abi.RegisteredS
 	return resp.Body, nil
 }
 
+func (r *Remote) CheckAllocated(ctx context.Context, s storage.SectorRef, offset, size abi.PaddedPieceSize, ft storiface.SectorFileType) (bool, error) {
+	if ft != storiface.FTUnsealed {
+		return false, xerrors.Errorf("CheckAllocated only supports unsealed files")
```
dirkmc, line 362: If the method only supports unsealed files, then maybe we don't need to pass the ft parameter at all (and we can change the method name to something like CheckUnsealedAllocated)

markets/pricing/cli.go
```diff
@@ -0,0 +1,48 @@
...
+	c.Stdout = &out
+	c.Stderr = &errb
+
+	switch err := c.Run().(type) {
+	case nil:
+		bz := out.Bytes()
+		resp := retrievalmarket.Ask{}
+
+		if err := json.Unmarshal(bz, &resp); err != nil {
+			return resp, xerrors.Errorf("failed to parse pricing output %s, err=%s", string(bz), err)
```
dirkmc, line 40:
```suggestion
			return resp, xerrors.Errorf("failed to parse pricing output %s: %w", string(bz), err)
```

markets/pricing/cli.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+		return resp, nil
+	case *exec.ExitError:
+		return retrievalmarket.Ask{}, xerrors.Errorf("pricing func exited with error: %s", errb.String())
+	default:
+		return retrievalmarket.Ask{}, xerrors.Errorf("pricing func cmd run error: %s", err)
```
dirkmc, line 46:
```suggestion
		return retrievalmarket.Ask{}, xerrors.Errorf("pricing func cmd run error: %w", err)
```

node/builder.go
```diff
@@ -566,6 +566,13 @@ func ConfigStorageMiner(c interface{}) Option {
 		return Error(xerrors.Errorf("invalid config from repo, got: %T", c))
 	}
 
+	pricingConfig := cfg.Dealmaking.RetrievalPricing
+	if pricingConfig.Strategy == config.ExternalRetrievalPricing {
+		if pricingConfig.External == nil || pricingConfig.External.Path == "" {
+			return Error(xerrors.New("invalid external retrieval pricing config"))
```
dirkmc, line 572: Let's separate out the error messages for External == nil and the path being the empty string - probably users are going to make these mistakes so we want to give them as much information as possible about what is wrong so they can fix it

dirkmc (commented):
I left a couple of suggestions.

When we test this out on a real miner we should measure the performance of checking if a piece in unsealed - it will be called as part of an externally facing API call so we need to make sure it's not too expensive.

extern/sector-storage/stores/remote.go
aarshkshah1992: This is done.

extern/sector-storage/piece_provider.go
```diff
@@ -40,6 +43,26 @@ func NewPieceProvider(storage *stores.Remote, index stores.SectorIndex, uns Unse
...
+	if err := size.Validate(); err != nil {
+		return false, xerrors.Errorf("size is not a valid piece size: %w", err)
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	if err := p.index.StorageLock(ctx, sector.ID, storiface.FTUnsealed, storiface.FTNone); err != nil {
+		cancel()
+		return false, xerrors.Errorf("acquiring read sector lock: %w", err)
+	}
+	defer cancel()
```
nonsense, line 61: Suggestion: maybe move this `defer cancel()` below line 56, and then remove the `cancel()` on line 58?

extern/sector-storage/stores/remote.go
```diff
@@ -484,6 +484,78 @@ func (r *Remote) readRemote(ctx context.Context, url string, offset, size abi.Pa
...
+
+	paths, _, err := r.local.AcquireSector(ctx, s, ft, storiface.FTNone, storiface.PathStorage, storiface.AcquireMove)
+	if err != nil {
+		return false, xerrors.Errorf("acquire local: %w", err)
+	}
+
+	path := storiface.PathByType(paths, ft)
+	if path != "" {
+		// if we have the unsealed file locally, check if it has the unsealed piece.
+		log.Infof("Read local %s (+%d,%d)", path, offset, size)
```
nonsense, line 501:
```suggestion
		log.Debugf("Read local %s (+%d,%d)", path, offset, size)
```

aarshkshah1992: Hey @nonsense This is just one line during reading a piece and having as Info dosen't hurt.

extern/sector-storage/piece_provider.go
aarshkshah1992: This is done.

aarshkshah1992:
@nonsense Please can you take a look at the `build-lotus-soup` failure in the CI ? Looks like a dependency problem but I'm not sure how to fix that.

aarshkshah1992:
@magik6k Please can you give this a review ?

extern/sector-storage/stores/remote.go
nonsense: Obviously that's a nit, and I won't go into arguing about it, but getting an INFO line for every read piece in Lotus, seems like debug information to me.

markets/pricing/cli.go
magik6k: Well, this is coming straight from the miner config, so it's not-so-untrusted user input. Obviously you can still do dumb things with it, but the same can be said about `sudo`

markets/retrievaladapter/provider.go
```diff
@@ -99,3 +99,69 @@ func (rpn *retrievalProviderNode) GetChainHead(ctx context.Context) (shared.TipS
...
+		if err != nil {
+			return resp, xerrors.Errorf("failed to look up deal %d on chain: err=%w", dealID, err)
+		}
+		if ds.Proposal.VerifiedDeal {
+			resp.VerifiedDeal = true
+		}
+
+		if ds.Proposal.PieceCID.Equals(pieceCID) {
+			resp.PieceSize = ds.Proposal.PieceSize.Unpadded()
+		}
```
magik6k, line 155: Reading the comment above, this should always be true, no?

node/builder.go
```diff
@@ -563,6 +570,17 @@ func ConfigStorageMiner(c interface{}) Option {
...
+	pricingConfig := cfg.Dealmaking.RetrievalPricing
+	if pricingConfig.Strategy == config.ExternalRetrievalPricing {
+		if pricingConfig.External == nil {
+			return Error(xerrors.New("retrieval pricing policy has been to set to external but external policy config is nil"))
+		}
+
+		if pricingConfig.External.Path == "" {
+			return Error(xerrors.New("retrieval pricing policy has been to set to external but external script path is empty"))
+		}
+	}
```
magik6k, line 582: I'd do `elseif Strategy != default -> error "only default or external are valid values"`

node/config/def.go
```diff
@@ -66,6 +74,29 @@ type DealmakingConfig struct {
...
 	Filter          string
 	RetrievalFilter string
+
+	RetrievalPricing *RetrievalPricing
+}
+
+type RetrievalPricing struct {
+	Strategy string // possible values: "default", "external"
+
+	Default  *RetrievalPricingDefault
```
magik6k, line 84: How does that look in default toml config? (there is a test in this dir to print default configs)

extern/sector-storage/stores/remote.go
magik6k, line 501:
```suggestion
		log.Debugf("Check local %s (+%d,%d)", path, offset, size)
```

Debug because this will be spammy (`stores` logger is debug by default, but miners who don't want their logs to be 10s of GBs can still easily disable those)

magik6k (commented):
First pass, the general approach makes a lot of sense. Should be easy to land after landing the direct piecereader PR.

extern/sector-storage/stores/remote.go
aarshkshah1992: Done.

node/config/def.go
aarshkshah1992: @magik6k 

Output of `TestDefaultMinerRoundtrip`:

```
# Default config:
[API]
  ListenAddress = "/ip4/127.0.0.1/tcp/2345/http"
  RemoteListenAddress = "127.0.0.1:2345"
  Timeout = "30s"

[Backup]
  DisableMetadataLog = false

[Libp2p]
  ListenAddresses = ["/ip4/0.0.0.0/tcp/0", "/ip6/::/tcp/0"]
  AnnounceAddresses = []
  NoAnnounceAddresses = []
  ConnMgrLow = 150
  ConnMgrHigh = 180
  ConnMgrGrace = "20s"

[Pubsub]
  Bootstrapper = false
  RemoteTracer = "/dns4/pubsub-tracer.filecoin.io/tcp/4001/p2p/QmTd6UvR47vUidRNZ1ZKXHrAFhqTJAD27rKL9XYghEKgKX"

[Dealmaking]
  ConsiderOnlineStorageDeals = true
  ConsiderOfflineStorageDeals = true
  ConsiderOnlineRetrievalDeals = true
  ConsiderOfflineRetrievalDeals = true
  ConsiderVerifiedStorageDeals = true
  ConsiderUnverifiedStorageDeals = true
  PieceCidBlocklist = []
  ExpectedSealDuration = "24h0m0s"
  PublishMsgPeriod = "1h0m0s"
  MaxDealsPerPublishMsg = 8
  MaxProviderCollateralMultiplier = 2
  Filter = ""
  RetrievalFilter = ""
  [Dealmaking.RetrievalPricing]
    Strategy = "default"
    [Dealmaking.RetrievalPricing.Default]
      VerifiedDealsFreeTransfer = true
    [Dealmaking.RetrievalPricing.External]
      Path = ""

```

aarshkshah1992:
@magik6k Have rebased & addressed your review. Please merge if happy.

raulk:
@aarshkshah1992 do you really mean to update the FFI here?

go.sum
```diff
@@ -283,8 +283,8 @@ github.com/filecoin-project/go-fil-commcid v0.0.0-20200716160307-8f644712406f/go
 github.com/filecoin-project/go-fil-commcid v0.0.0-20201016201715-d41df56b4f6a h1:hyJ+pUm/4U4RdEZBlg6k8Ma4rDiuvqyGpoICXAxwsTg=
 github.com/filecoin-project/go-fil-commcid v0.0.0-20201016201715-d41df56b4f6a/go.mod h1:Eaox7Hvus1JgPrL5+M3+h7aSPHc0cVqpSxA+TxIEpZQ=
 github.com/filecoin-project/go-fil-markets v1.0.5-0.20201113164554-c5eba40d5335/go.mod h1:AJySOJC00JRWEZzRG2KsfUnqEf5ITXxeX09BE9N4f9c=
-github.com/filecoin-project/go-fil-markets v1.4.0 h1:J4L6o+FVOmS7ZWV6wxLPiuoDzGC7iS3S5NRFL1enEr0=
-github.com/filecoin-project/go-fil-markets v1.4.0/go.mod h1:7be6zzFwaN8kxVeYZf/UUj/JilHC0ogPvWqE1TW8Ptk=
+github.com/filecoin-project/go-fil-markets v1.5.0 h1:3KEs01L8XFCEgujZ6ggFjr1XWjpjTQcmSSeo3I99I0k=
+github.com/filecoin-project/go-fil-markets v1.5.0/go.mod h1:7be6zzFwaN8kxVeYZf/UUj/JilHC0ogPvWqE1TW8Ptk=
```
raulk, line 287: This is dragging in version upgrade on go-fil-markets. Just pointing out that the code effectively merged in this PR is not just _this PR_, but also the upstream changes. Could you please detail the highlights of go-fil-markets v1.5.0 in the PR description, and include a link to the release?

itests/kit/deals.go
```diff
@@ -43,6 +43,29 @@ func NewDealHarness(t *testing.T, client api.FullNode, miner TestMiner) *DealHar
...
+	time.Sleep(time.Second)
+	dh.WaitDealSealed(ctx, deal, false, false, nil)
+
+	// Retrieval
+	info, err := dh.client.ClientGetDealInfo(ctx, *deal)
+	require.NoError(dh.t, err)
+
+	return data, info, fcid
+}
+
```
raulk, line 68: This is going to change, but one thing we should try to avoid doing is adding bespoke methods to the test kit that do overly specific things. In this case, an alternative could be to wrap the make `MakeFullDeal` method parameters in a struct, and provide an option "PerformRetrieval" to decide whether we do a retrieval or not, instead of duplicating the code here and removing a few lines.

markets/pricing/cli.go
raulk: I'm not referring to the miner configuration (i.e. the script, or the path to it) -- I'm more referring to the fact that what's being passed as input is a serialized `retrievalmarket.PricingInput`. But that's a constrained type that's populated in our code, i.e. it's not _actually_ user-provided input. So my considerations don't really apply here.

image

markets/retrievaladapter/provider.go
```diff
@@ -99,3 +99,69 @@ func (rpn *retrievalProviderNode) GetChainHead(ctx context.Context) (shared.TipS
...
+			Number: sectorID,
+		},
+		ProofType: si.SectorType,
+	}
+
+	log.Debugf("will call IsUnsealed now sector=%+v, offset=%d, size=%d", sectorID, offset, length)
+	return rpn.pp.IsUnsealed(ctx, ref, storiface.UnpaddedByteIndex(offset), length)
+}
+
+// `storageDeals` param here is the list of storage deals made for the `payloadCID` the retrieval client is looking for.
```
raulk, line 126: Nit: godoc here should start with a description of what this function does, in the form:

```
GetRetrievalPricingInput takes a set of candidate storage deals that can serve a retrieval request,
and returns an minimally populated PricingInput. This PricingInput should be enhanced
with more data, and passed to the pricing function to determine the final quoted price.
```

markets/retrievaladapter/provider.go
```diff
@@ -99,3 +99,69 @@ func (rpn *retrievalProviderNode) GetChainHead(ctx context.Context) (shared.TipS
...
+		}
+		if ds.Proposal.VerifiedDeal {
+			resp.VerifiedDeal = true
+		}
+
+		if ds.Proposal.PieceCID.Equals(pieceCID) {
+			resp.PieceSize = ds.Proposal.PieceSize.Unpadded()
+		}
+
+		if resp.VerifiedDeal && resp.PieceSize != 0 {
```
raulk, line 157: This is the most "optimal" case, right? That's why we consider ourselves satisfied and stop the search. Maybe add a comment of what's the rationale here.

markets/retrievaladapter/provider.go
```diff
@@ -99,3 +99,69 @@ func (rpn *retrievalProviderNode) GetChainHead(ctx context.Context) (shared.TipS
...
+			break
+		}
+	}
+
+	if resp.PieceSize == 0 {
+		return resp, xerrors.New("failed to find matching piece, PieceSize is zero")
+	}
```
raulk, line 164: Hmmm... logically speaking, the piece size being zero and not actually finding a match are two different outcomes. What you are doing here is really expressing and asserting a system invariant ("no deal can be made with piece size zero"). I think this is OK? But please add a comment acknowledging this situation.

extern/sector-storage/stores/remote_test.go
```diff
@@ -326,6 +327,244 @@ func TestReader(t *testing.T) {
...
+		needHttpServer bool
+
+		getAllocatedReturnCode int
+
+		serverUrl string
+
+		// expectation
+		errStr            string
+		expectedIsUnealed bool
+	}{
```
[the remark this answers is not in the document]
raulk: I defer to @magik6k to review the unsealing logic and the test coverage. I'm not so familiar with this part of the codebase.

node/builder.go
raulk: @aarshkshah1992 I think you addressed this, but could you comment and mark as resolved?

node/config/def.go
```diff
@@ -14,6 +14,14 @@ import (
 	sectorstorage "github.com/filecoin-project/lotus/extern/sector-storage"
 )
 
+const (
+	// DefaultRetrievalPricing configures the node to use the default retrieval pricing policy.
+	DefaultRetrievalPricing = "default"
+	// ExternalRetrievalPricing configures the node to use the external retrieval pricing script
+	// configured by the user.
+	ExternalRetrievalPricing = "external"
+)
```
raulk, line 23: ubernit: invert the naming here, as doing so groups these consts alphabetically in godocs.

```suggestion
const (
	// RetrievalPricingDefault configures the node to use the default retrieval pricing policy.
	RetrievalPricingDefault = "default"
	// RetrievalPricingExternal configures the node to use the external retrieval pricing script
	// configured by the user.
	RetrievalPricingExternal = "external"
)
```

node/config/def.go
raulk: @magik6k we settled on this as it's clean, enables extensibility, and follows best practices.

node/config/def.go
```diff
@@ -322,6 +353,16 @@ func DefaultStorageMiner() *StorageMiner {
...
 			MaxProviderCollateralMultiplier: 2,
+
+			RetrievalPricing: &RetrievalPricing{
+				Strategy: DefaultRetrievalPricing,
+				Default: &RetrievalPricingDefault{
+					VerifiedDealsFreeTransfer: true,
+				},
+				External: &RetrievalPricingExternal{
+					Path: "",
+				},
```
raulk, line 364: I think we should omit this altogether. Is it not possible?

raulk:
@aarshkshah1992 we need to write docs for this. Can you draft a section for the "Mine >> Managing retrievals" subpage? https://docs.filecoin.io/mine/lotus/manage-retrieval-deals/ A markdown gist is fine. Possible structure for this section:

```md
## Retrieval pricing

Lotus allows you to set different policies to calculate the quoted price of a retrieval deal:

* default: <explain>
* external: <explain>

### Using the default retrieval pricing policy

<explain>

### Using the external retrieval pricing policy

<explain>
```

@atopal pinging you here as your team will likely own the integration of the content ;-)

raulk (commented):
Actually, am I right that we're missing converage for the interface with the external pricing function? I'd expect these test cases:

- OK: verify that we generate the right input, and that we can process well generated output.
- ERR: path doesn't exist.
- ERR: path is not executable.
- ERR: we can handle badly structured output, e.g. not JSON.
- ERR: we can handle bad output content, e.g. output is JSON, but all fields are nil values, and/or numeric fields overflow.

node/builder.go
aarshkshah1992: This is done.

extern/sector-storage/stores/remote.go
```diff
@@ -507,7 +579,7 @@ func (r *Remote) Reader(ctx context.Context, s storage.SectorRef, offset, size a
 	if path != "" {
 		// if we have the unsealed file locally, return a reader that can be used to read the contents of the
 		// unsealed piece.
-		log.Infof("Read local %s (+%d,%d)", path, offset, size)
+		log.Debugf("Check local %s (+%d,%d)", path, offset, size)
```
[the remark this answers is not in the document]
aarshkshah1992: Yes, we don't need it except for when we are chasing down hairy bugs.

markets/pricing/cli.go
aarshkshah1992: So, this is settled.

go.sum
aarshkshah1992: This is done.

node/config/def.go
aarshkshah1992: @raulk Some tests complain that this is needed because of how the toml gets serialized/de-serialised. No harm in keeping it around unless you want me to dig deeper into why those tests complain.

itests/kit/deals.go
aarshkshah1992: This is done.

markets/retrievaladapter/provider.go
aarshkshah1992: This is done.

markets/retrievaladapter/provider.go
aarshkshah1992: This is done.

markets/retrievaladapter/provider.go
aarshkshah1992: This is done.

node/config/def.go
aarshkshah1992: This is done.

aarshkshah1992:
## Retrieval pricing

Lotus allows you to set different policies to calculate the quoted price of a retrieval deal:

* **default:** Lotus is shipped with an in-built default pricing policy that offers free retrievals for Verified Unsealed deals. Miners can still charge clients for data-transfer if they wish to by turning off free data-transfer for verified deals in the Miner config. Miners can also charge clients for Unsealing if they don't have an Unsealed sector file to retrieve the deal payload from.

* **external:** Miners can configure an external pricing script that will be used instead of the default pricing policy to price retrieval deals. The script will be called with the relevant deal parameters and the output quote will be used by Lotus to price deals.

### Using the default retrieval pricing policy

- The default pricing policy uses the `Ask` configured in the Ask Store (set via the `lotus retrieval-deals set-ask` CLI command) to price deals.
- However, it will not charge for data-transfer if there exists a verified storage deal for the payload being retrieved. This behaviour can be turned off by setting the `VerifiedDealsFreeTransfer` flag to false in the `DealMaking.RetrievalPricing` section of the config as shown below.
```
 [Dealmaking.RetrievalPricing]
    Strategy = "default"
    [Dealmaking.RetrievalPricing.Default]
      VerifiedDealsFreeTransfer = false
```
 
- It will also not charge for Unsealing if we already have an Unsealed sector file containing the retrieval payload.

### Using the external retrieval pricing policy

- Users can configure an external pricing script(similar to the deal filter mechanism) that takes a json marshalled `PricingInput` as the input and outputs a json marshalled `Ask` i.e. the quote.  The `PricingInput` struct is defined as follows:

```
type PricingInput struct {
	// PayloadCID is the cid of the payload to retrieve.
	PayloadCID cid.Cid
	// PieceCID is the cid of the Piece from which the Payload will be retrieved.
	PieceCID cid.Cid
	// PieceSize is the size of the Piece from which the payload will be retrieved.
	PieceSize abi.UnpaddedPieceSize
	// Client is the peerID of the retrieval client.
	Client peer.ID
	// VerifiedDeal is true if there exists a verified storage deal for the PayloadCID.
	VerifiedDeal bool
	// Unsealed is true if there exists an unsealed sector from which we can retrieve the given payload.
	Unsealed bool
	// CurrentAsk is the ask configured in the ask-store via the `lotus retrieval-deals set-ask` CLI command.
	CurrentAsk Ask
}
```


- To use this mode instead of the default above, the `DealMaking.RetrievalPricing` section of the config needs to be configured to use the "external" pricing policy and needs to be given the absolute path of the pricing script as shown below:

```
[Dealmaking.RetrievalPricing]
    Strategy = "external"
    [Dealmaking.RetrievalPricing.External]
      Path = "/var/script"
```

aarshkshah1992:
@raulk I'd love to write tests for the external pricing script but it's not clear to me how instantiate the nodes with a desired config. For now, all nodes are instantiated with the default config. Does the `testkit` package have anything for this ?

aarshkshah1992:
@magik6k Please can you approve the PR if happy ?

jennijuju:
> se can you approve the PR if happy ?

@raulk @dirkmc do you guys ✅ this pr before Magik and aayush should take a look?

dirkmc (approved):

magik6k (approved):
Just one nitpick, looks good otherwise

extern/sector-storage/stores/remote.go
```diff
@@ -484,6 +484,78 @@ func (r *Remote) readRemote(ctx context.Context, url string, offset, size abi.Pa
...
+		has, err := r.pfHandler.HasAllocated(pf, storiface.UnpaddedByteIndex(offset.Unpadded()), size.Unpadded())
+		if err != nil {
+			return false, xerrors.Errorf("has allocated: %w", err)
+		}
+
+		if err := r.pfHandler.Close(pf); err != nil {
+			return false, xerrors.Errorf("failed to close partial file: %s", err)
+		}
+		log.Debugf("checked if local partial file has the piece %s (+%d,%d), returning answer=%t", path, offset, size, has)
+		return has, nil
```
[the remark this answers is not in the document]
aarshkshah1992: @magik6k Address this with some tests. Please take a look.

magik6k (approved):

raulk:
@aarshkshah1992 could you link the issue for the extra test we wanted to add (external pricing function)?

aarshkshah1992:
Filed an issue to write missing e2e tests for the external pricing script feature at https://github.com/filecoin-project/lotus/issues/6508.

PR #6192, opened 2021-05-05
build on jdcloud

coryschwartz:
Build packer images on jdcloud.


Stebalien:
@coryschwartz keep or close?

PR #6195, opened 2021-05-06
Introduce v5 actors

arajasek:
Actor wrappers generated by codegen `make actors-gen`.


build/params_2k.go
```diff
@@ -39,6 +39,8 @@ var UpgradeNorwegianHeight = abi.ChainEpoch(40)
 
 var UpgradeActorsV4Height = abi.ChainEpoch(45)
 
+var UpgradeHyperdriveHeight = abi.ChainEpoch(50)
```
Stebalien, line 42: ActorsV5? I mean, I know hyperdrive is nicer... but it's kind of inconsistent.

That or we can go back and name all the other heights. I'd actually prefer to do it that way.

Stebalien (approved):
The code-gen makes this kind of thing a lot less stressful.

build/params_2k.go
arajasek: Yeah, I'll rename the old ones -- the inconsistency was bothering me too. I think we want to go with Hyperdrive for now (and names from this point on).

Stebalien: `UpgradeHyperspaceBypassHeight`? We can totally print some bad poetry at the time of the upgrade.

arajasek:
Test failures will be resolved by https://github.com/filecoin-project/specs-actors/pull/1401

chain/stmgr/forks.go
```diff
@@ -1053,7 +1057,7 @@ func upgradeActorsV3Common(
 	// Perform the migration
 	newHamtRoot, err := nv10.MigrateStateTree(ctx, store, stateRoot.Actors, epoch, config, migrationLogger{}, cache)
 	if err != nil {
-		return cid.Undef, xerrors.Errorf("upgrading to actors v2: %w", err)
+		return cid.Undef, xerrors.Errorf("upgrading to actors v3: %w", err)
```
ZenGround0, line 1060: Nice find

ZenGround0 (approved):
LGTM

chain/stmgr/forks.go
arajasek: Haha, thanks!

arajasek:
I _think_ the failures are flaky at this point. I'm going to merge it into the staging branch for now, but we will need to gain confidence in this.

PR #6215, opened 2021-05-08
add rpc for publish message

hunjixin:
mpool can only republish message of local address, but in venus the daemon has no address, so msg in mpool from other tool like venus-messager hard to publish to network.  add this api for other tool to publish their messages


jennijuju:
maybe better to just add a param flag for publishing in `MpoolPush`

hunjixin:
> @hunjixin could you please take a look at the above comment and see if you want to make the change?

This maybe cause API incompatibility issues， its okay？

PR #6226, opened 2021-05-10
Improve the cli state call command 

wadeAlexC:
`./lotus state call` is currently unusable for most actor methods, as parameter parsing does not support many types.

This PR changes the `call` command to accept parameters as base64 or hex strings, as these formats are used extensively in other CLI commands.

Additionally, `call` has been changed to support multiple formats for message return:

* `decoded` (default): Returns JSON based on the return type of the method called
* `raw`: Returns the raw `[]byte` found in the receipt
* `hex`: Converts the returned data to hex
* `base64` Converts the returned data to base64

Here's an example:

```
alex-pc$ lotus state call --from f0127573 f0127595 23 RACipzE=
Call receipt:
Exit code: 0
Gas Used: 0
Return:
{}
alex-pc$ lotus state call f04 9
Call receipt:
Exit code: 0
Gas Used: 0
Return:
{
  "RawBytePower": "5903779890947162112",
  "QualityAdjPower": "5905615863821533184",
  "PledgeCollateral": "71023842284078169744154578",
  "QualityAdjPowerSmoothed": {
    "PositionEstimate": "2009447455836142781845833807839954701917245286790497884694",
    "VelocityEstimate": "5306208777810336014187026111611315363244696189173411"
  }
}
```

### DIFF
--- a/cli/state.go
+++ b/cli/state.go
@@ -3,6 +3,8 @@ package cli
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"html/template"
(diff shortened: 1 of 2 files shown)


magik6k (approved):
This is much better, thanks!

Issue #6182, opened 2021-05-04
[BUG] Lotus tries to acquire ~500GB for sectors in FinalizeSector state


nonsense:
**Describe the bug**
For some reason I ended up with 4 sectors in FinalizeSector state on my miner:

```
639  Proving         YES      YES     2275725 (in 1 year 24 weeks)    CC
640  FinalizeSector  YES      NO      2275725 (in 1 year 24 weeks)    CC
  RecoveryTimeout: 766665 (in 1 week 6 days)
641  Proving         YES      YES     2275725 (in 1 year 24 weeks)    CC
642  FinalizeSector  YES      NO      2278605 (in 1 year 24 weeks)    CC
  RecoveryTimeout: 766665 (in 1 week 6 days)
643  FinalizeSector  YES      NO      2278605 (in 1 year 24 weeks)    CC
  RecoveryTimeout: 766665 (in 1 week 6 days)
644  FinalizeSector  YES      NO      2278605 (in 1 year 24 weeks)    CC
645  WaitDeals       NO       NO      n/a                             1
```

Sectors never moved to `Proving` state, so after rebooting the miner, I noticed that Lotus tries to acquire ~520GB disk space which is not available in order to process the sectors:

```
2021-05-04T13:26:27.384+0300    ERROR   advmgr  sector-storage/sched.go:404     trySched(1) req.sel.Ok error: finding best alloc storage:
    github.com/filecoin-project/lotus/extern/sector-storage.(*allocSelector).Ok
        /root/lotus/extern/sector-storage/selector_alloc.go:55
  - no good path found:
    github.com/filecoin-project/lotus/extern/sector-storage/stores.(*Index).StorageBestAlloc
        /root/lotus/extern/sector-storage/stores/index.go:421
2021-05-04T13:26:27.384+0300    DEBUG   stores  stores/index.go:403     not allocating on d3f534f4-e820-419d-8405-3debfaa4fa25, out of space (available: 512476852224, need: 518832049356)
```

Even though I had 1.2TB free on sealing location, none of the sectors could progress.

The workaround was to hack Lotus and override the `stat.Available` variable in order to move forward with finalizing the sectors:

```
diff --git a/extern/sector-storage/stores/local.go b/extern/sector-storage/stores/local.go
index 5a10b21b9..0166dde8c 100644
--- a/extern/sector-storage/stores/local.go
+++ b/extern/sector-storage/stores/local.go
[...]

PR #6233, opened 2021-05-11
consider storiface.PathStorage when calculating storage requirements

nonsense:
Addresses: https://github.com/filecoin-project/lotus/issues/6182

When a sector gets to the `FinalizeSector` state, currently Lotus tries to acquire ~500GB of storage on the location for storage, while it needs much less for the sector.

This code was proposed by @moremorefun (Thank you!!) and I think it is solving the issue.

### DIFF
--- a/extern/sector-storage/stores/index.go
+++ b/extern/sector-storage/stores/index.go
@@ -3,6 +3,7 @@ package stores
 import (
 	"context"
 	"errors"
+	"math"
 	"net/url"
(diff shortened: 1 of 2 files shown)


nonsense:
This has been tested on `sofiaminer` and I think it works fine.

dirkmc (approved):
LGTM 👍

magik6k (approved):

Issue #6030, opened 2021-04-13
[Feature Request] Offer pre-built binaries for Lotus


atopal:
**Is your feature request related to a problem? Please describe.**
Currently people who are interesting in giving Filecoin a try have to build it from scratch, which is quite time consuming and unusual.

**Describe the solution you'd like**
Offer pre-built binaries for common OSes to download

**Describe alternatives you've considered**
We've written instructions for how to build it from scratch

f8-ptrk:
@atopal the problem i see with this is that the binaries are highly processor architecture dependent. there would need to be a wide array of binaries to build or at least a target subset of architectures.

and all those binaries multiplied by 3 for the 3 different networks

atopal:
I think it would be fine to offer a subset of all possible binaries, limited to just the most popular ones we see on our site, and also to limit to just the network we want to use in our "getting started" guide. We'd still have docs for how you can build it yourself, if you don't have a matching binary.

travisperson:
We have binaries for linux and macos on x86, the macos builds look to have been broken for a bit though when the hwloc dependency was introduced. This was resolved in PR https://github.com/filecoin-project/lotus/pull/5934 though, so we should see build going through now.

atopal:
@travisperson Do you have a link? I couldn't find the binaries.

terichadbourne:
@f8-ptrk For some clarity, our goal is specifically to the needs of storage clients who we're currently teaching to use Lotus Lite on mainnet, with greatly reduced hardware requirements. Totally understand that miners with intense hardware setups will want a lot of customization, and should continue to have it. 

@travisperson If we indeed are already offering binaries, where would a user go to find them? 

cc @dkkapur re golden path backlog - we expect that using binaries would drop 20-30 minutes from our current hour-long path to making starting one's first deal

BigLep:
Per verbal standup on 2021-04-27, @travisperson is going to respond with:
1. Where to find the binaries?
2. Steps that would need to be taking to get binaries created for Lotus 1.8
3. Steps to take for adding additional hardware types (e.g., arm_64)

With 2 and 3, a maintenance tribute can pick this up.

travisperson:
[...]

PR #6238, opened 2021-05-12
chore(ci): Enable build on RC tags

placer14:
Enables test, build, publish for all OS and network targets for release candidate tags in CI.

Re: #6030


arajasek (approved):
Thanks, this should work...I think

PR #6240, opened 2021-05-12
Fix randomness fetching around null blocks

arajasek:
This is the larger but "better" fix to the critical portion of #3613. 

- [x] Needs tests
- [ ] Can we come up with a creative way to avoid the height check in the Chain API impl? That's the only bit I'm unhappy about.
- [ ] Confirm we should actually be changing ChainRandomness fetching too.

### DIFF
--- a/chain/sync_test.go
+++ b/chain/sync_test.go
@@ -7,6 +7,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/filecoin-project/go-state-types/crypto"
+
+	"github.com/filecoin-project/go-state-types/network"
+
+	"github.com/filecoin-project/lotus/chain/stmgr"
+
 	"github.com/ipfs/go-cid"
 
 	ds "github.com/ipfs/go-datastore"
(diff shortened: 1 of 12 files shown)


node/impl/full/chain.go
```diff
@@ -95,7 +97,12 @@ func (a *ChainAPI) ChainGetRandomnessFromTickets(ctx context.Context, tsk types.
 		return nil, xerrors.Errorf("loading tipset key: %w", err)
 	}
 
-	return a.Chain.GetChainRandomness(ctx, pts.Cids(), personalization, randEpoch, entropy)
+	// Doing this here is slightly nicer than doing it in the chainstore directly, but it's still bad for ChainAPI to reason about network upgrades
+	if randEpoch > build.UpgradeHyperdriveHeight {
```
arajasek, line 101: Would love a creative suggestion that allows us to not do this

Kubuxu:
This version seems better to me. The Chain API having to check network version is annoying, especially if we consider that in future upgrades might depend on state and not a pure chain.

arajasek:
Added a test which will pass once we make a minor tweak to the tipset skipcache index (waiting on some input from Why)

chain/sync_test.go
```diff
@@ -833,3 +891,55 @@ func TestSyncCheckpointEarlierThanHead(t *testing.T) {
 	require.Equal(tu.t, p1Head, a.TipSet())
 	tu.assertBad(p1, b.TipSet())
 }
+
+func TestDrandNull(t *testing.T) {
```
magik6k, line 895: Seems to be failing on CI - https://app.circleci.com/pipelines/github/filecoin-project/lotus/14479/workflows/adc1f7a9-40b6-41f2-a72c-b9cb00f20c70/jobs/151579

arajasek: Yes, because of the tipset cache thing i mentioned above^

Kubuxu (approved):
SGWM, regarding the ugliness: the randomness just does not belong in Chain API,
as choosing randomness turns out to be a consensus rule built on top of the content of the chain.

chain/sync_test.go
```diff
@@ -833,3 +891,55 @@ func TestSyncCheckpointEarlierThanHead(t *testing.T) {
...
+	H := 10
+	v5h := abi.ChainEpoch(50)
+	build.UpgradeHyperdriveHeight = v5h
```
arajasek, line 898: mental note to reset this at end of test

magik6k (approved):

PR #6249, opened 2021-05-13
Upgrade nerpa to actor v4 and bump the version to rc4

jennijuju:


arajasek (approved):
Could also bump the CHANGELOG.md

PR #6256, opened 2021-05-14
fix(ci): Updates to lotus CI build process

placer14:
- updates build environment to latest Ubuntu LTS (20.04).
- creates discussion in "announcements" category when tag doesn't contain `-rc[0-9]+$`
- only marks tags which do have `-rc[0-9]+$` as pre-release releases
- includes more output in build during release posting for future troubleshooting

Contributions toward #6030

### DIFF
--- a/scripts/publish-release.sh
+++ b/scripts/publish-release.sh
@@ -29,13 +29,20 @@ RELEASE_ID=`echo "${RELEASE_RESPONSE}" | jq '.id'`
 if [ "${RELEASE_ID}" = "null" ]; then
   echo "creating release"
 
+  COND_CREATE_DISCUSSION=""
+  PRERELEASE=true
+  if [[ ${CIRCLE_TAG} =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+    COND_CREATE_DISCUSSION="\"discussion_category_name\": \"announcement\","
+    PRERELEASE=false
+  fi
+
   RELEASE_DATA="{
     \"tag_name\": \"${CIRCLE_TAG}\",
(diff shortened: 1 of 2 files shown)


scripts/publish-release.sh
```diff
@@ -32,6 +32,7 @@ if [ "${RELEASE_ID}" = "null" ]; then
   RELEASE_DATA="{
     \"tag_name\": \"${CIRCLE_TAG}\",
     \"target_commitish\": \"${CIRCLE_SHA1}\",
+    \"discussion_category_name\": \"announcement\",
```
jennijuju, line 35: is creating a discussion mandatory to triger the build? it is fine, i just don't see why

scripts/publish-release.sh
```diff
@@ -29,12 +29,20 @@ RELEASE_ID=`echo "${RELEASE_RESPONSE}" | jq '.id'`
 if [ "${RELEASE_ID}" = "null" ]; then
   echo "creating release"
 
+  COND_CREATE_DISCUSSION=""
+  PRERELEASE=true
+  if [[ ${CIRCLE_TAG} =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+    COND_CREATE_DISCUSSION="\"discussion_category_name\": \"announcement\","
+    PRERELEASE=false
+  fi
```
placer14, line 37: If the tag doesn't have `-rcNNN` at the end, then create a discussion in the `announcement` category and mark the release as _not_ a pre-release.

scripts/publish-release.sh
placer14: No, it is not required. Updated to make it conditional.

jennijuju (approved):

PR #6264, opened 2021-05-15
rollout on release

coryschwartz:
This is an automaticly-executed trigger of the "rollout" workflow. It should be have the same as the trigger-lotus-rollout job, except this one will be executed when a release is created.

See here for the manual deployment.

see https://github.com/filecoin-project/lotus/pull/6263 for the manual trigger for this workflow.


coryschwartz:
Closing, this is just a UI nicety.

Issue #4069, opened 2020-09-28
Fetch failure due to lack of disk space


hyunmoon:
**Describe the bug**

```
WARN	rpc	go-jsonrpc@v0.1.2-0.20200822201400-474f4fdccc52/handler.go:241	error in RPC call to 'Filecoin.Fetch': allocate local sector for fetching:
    github.com/filecoin-project/lotus/extern/sector-storage/stores.(*Remote).AcquireSector
        /home/downloads/lotus/extern/sector-storage/stores/remote.go:111
  - couldn't find a suitable path for a sector:
    github.com/filecoin-project/lotus/extern/sector-storage/stores.(*Local).AcquireSector
        /home/downloads/lotus/extern/sector-storage/stores/local.go:402
```

If a worker happened to get more sectors than it can handle, it runs out of space.
When it runs out of space, all sectors in it fail to be finalized because they can't be fetched.

When this happens, the worker usually has sector files that should no longer have.
I found 682G worth of cache files in `cache/fetching` directory.
They are all in `Proving` state so they should have been deleted but they didn't.
Files like these lead to lack of disk space.

**To Reproduce**
1. A worker dedicated to PC2 takes sectors while it has some sectors in WaitSeed, Commiting, and Finalize state.
2. Some sectors that are now proving didn't get deleted from the worker properly.
3. The worker runs out of disk space.
4. The sectors it had can't be finalized.
5. From that point, the worker does nothing but outputting errors.

**Expected behavior**
1. Any obsolete sector files should always be deleted from the worker.
2. If the worker can't do anything due to lack of disk space, it should try to move some sectors to the miner or other workers to make some space.

**Screenshots**
![report_fetch](https://user-images.githubusercontent.com/2242149/94411375-7514c000-01b3-11eb-9060-b18728cad327.JPG)
![report_fetch_2](https://user-images.githubusercontent.com/2242149/94411378-76de8380-01b3-11eb-8a4b-b1a309413382.JPG)
(All sectors in fetching directory are in Proving state so they should've deleted)

**Version (run `lotus version`):**
Tag v0.8.0

**Additional context**
I think this is happening because I simply don't have enough disk space on each workers. In theory, it should be enough but sometimes, some workers get more sectors than others and run out of disk space.

Related to:
https://github.com/filecoin-project/lotus/issues/3969
https://github.com/filecoin-project/lotus/issues/4015

shotcollin:
[...]

PR #6268, opened 2021-05-17
document sector store subsystem

aarshkshah1992:
Documents parts of the sector storage subsystem, Worker functionality and the Miner <-> Worker RPC subsystem.


extern/sector-storage/selector_existing.go
```diff
@@ -28,6 +28,10 @@ func newExistingSelector(index stores.SectorIndex, sector abi.SectorID, alloc st
 	}
 }
 
+// existingSelector only allows selecting a worker which supports the given tasktype and
+// 1. If `s.allowFetch` is false -> already has the sector file of type `s.alloc` for sector `s.sector` locally.
+// 2. If `s.allowFetch` is true -> if it dosen't have the sector file locally, it should have enough space in it's local sealing scratch space
```
nonsense, line 33:
```suggestion
// 2. If `s.allowFetch` is true -> if it doesn't have the sector file locally, it should have enough space in its local sealing scratch space
```

extern/sector-storage/stores/local.go
```diff
@@ -537,6 +545,8 @@ func (st *Local) Local(ctx context.Context) ([]StoragePath, error) {
 	return out, nil
 }
 
+// Remove removes the given sector file types  for the given sector from the local sealing scratch space and also from the local
```
nonsense, line 548:
```suggestion
// Remove removes the given sector file types for the given sector from the local sealing scratch space and also from the local
```

extern/sector-storage/storiface/filetype.go
```diff
@@ -51,10 +59,12 @@ func (t SectorFileType) String() string {
 	}
 }
 
+// Has returns true if the bitmask contains the given single file type.
```
nonsense, line 62:
```suggestion
// Has returns true if the bitmask contains the given file type.
```

extern/sector-storage/storiface/filetype.go
```diff
@@ -73,6 +83,7 @@ func (t SectorFileType) SealSpaceUse(ssize abi.SectorSize) (uint64, error) {
 	return need, nil
 }
 
+// All returns all the file types contained in the bitmask.
```
nonsense, line 86:
```suggestion
// All unpacks a SectorFileType bitmask into an array of booleans, one for each file type.
```

extern/sector-storage/storiface/filetype.go
```diff
@@ -113,6 +128,7 @@ func SectorName(sid abi.SectorID) string {
 	return fmt.Sprintf("s-t0%d-%d", sid.Miner, sid.Number)
 }
 
+// PathByType returns the  storageID/absolute sector store path  of the given fileType in the SectorPaths.
```
nonsense, line 131:
```suggestion
// PathByType returns the  storageID/absolute sector store path of the given fileType in the SectorPaths.
```

extern/sector-storage/storiface/filetype.go
```diff
@@ -126,6 +142,7 @@ func PathByType(sps SectorPaths, fileType SectorFileType) string {
 	panic("requested unknown path type")
 }
 
+// PathByType sets the  storageID/absolute sector store path of the given fileType in the SectorPaths.
```
nonsense, line 145:
```suggestion
// PathByType sets the storageID/absolute sector store path of the given fileType in the SectorPaths.
```

extern/sector-storage/storiface/storage.go
```diff
@@ -1,15 +1,21 @@
 package storiface
 
+// PathType is the type of a path in the sector store.
 type PathType string
 
 const (
+	// PathStorage means the path type that is used for long term storage of sealed sector files.
 	PathStorage PathType = "storage"
+	// PathSealing means the sealing sratch space that is used a staging area for sealing sector files.
```
nonsense, line 9:
```suggestion
	// PathSealing means the sealing scratch space that is used a staging area for sealing sector files.
```

extern/sector-storage/storiface/storage.go
```diff
@@ -1,15 +1,21 @@
...
 	PathSealing PathType = "sealing"
 )
 
+// AcquireMode is the move/copy semantics that is used by workers when fetching sector files from another worker.
 type AcquireMode string
 
 const (
+	// AcquireMove means move semantics that acquire a sector file from another location and them remove the original copy.
```
nonsense, line 17:
```suggestion
	// AcquireMove means move semantics that acquire a sector file from another location and then remove the original copy.
```

extern/sector-storage/storiface/worker.go
```diff
@@ -16,19 +16,30 @@ import (
 	"github.com/filecoin-project/lotus/extern/sector-storage/sealtasks"
 )
 
+// WorkerInfo contains information about the worker such as it's hostname and the
```
nonsense, line 19:
```suggestion
// WorkerInfo contains information about the worker such as its hostname and the
```

extern/sector-storage/storiface/worker.go
```diff
@@ -77,33 +93,85 @@ var _ fmt.Stringer = &CallID{}
...
+// WorkerCalls is the Worker's RPC interface that the miner calls to schedule sealing/unsealing tasks on the worker.
+// Tasks are executed asynchronously by the worker and the response is later returned to the Miner by using the Miner's `WorkerReturn` RPC interface.
+// Miner can use the `CallID` returned by the worker to track the status of the task.
 type WorkerCalls interface {
+	// AddPiece schedules a task to add a Piece to an Unsealed sector.
+	// If this is the first Piece being added to the given Sector, the worker will create an Unsealed Sector file on
+	// one of the sealing scratch spaces in the local sector store to add the given Piece.
+	// If the given Sector already contains existing Pieces, the worker will add the given Piece to the corresponding Unsealed Sector file.
+	// If the worker does NOT have the Unsealed sector file, it will fetch it from a worker that has it before adding the Piece.
+	// The call will fail of the Worker is NOT able to allocate space to store the Unsealed sector file.
```
nonsense, line 105:
```suggestion
	// The call will fail if the Worker is NOT able to allocate space to store the Unsealed sector file.
```

extern/sector-storage/storiface/worker.go
```diff
@@ -77,33 +93,85 @@ var _ fmt.Stringer = &CallID{}
...
+	// The call will fail if the Worker is NOT able to allocate space to store the Sealing files.
 	SealCommit1(ctx context.Context, sector storage.SectorRef, ticket abi.SealRandomness, seed abi.InteractiveSealRandomness, pieces []abi.PieceInfo, cids storage.SectorCids) (CallID, error)
+
+	// SealCommit2 schedules a task to perform the `SealCommit2` stage of a sector being sealed.
 	SealCommit2(ctx context.Context, sector storage.SectorRef, c1o storage.Commit1Out) (CallID, error)
+
+	// FinalizeSector schedules a task to perform the `FinalizeSector` stage of a sector being sealed.
+	// As a part of this task, the worker fetches(if required) and stores the cached sealing file on one of the long term
+	// storage directories in the local sector store.
+	// It fetches/keeps around the  Unsealed file in the  Sealing scratch space if `keepUnsealed` is non-zero.
```
nonsense, line 133:
```suggestion
	// It fetches/keeps around the Unsealed file in the Sealing scratch space if `keepUnsealed` is non-zero.
```

extern/sector-storage/storiface/worker.go
```diff
@@ -77,33 +93,85 @@ var _ fmt.Stringer = &CallID{}
...
 	ErrUnknown ErrorCode = iota
 )
 
 const (
-	// Temp Errors
+	// ErrTempUnknown represents Temp Errors.
 	ErrTempUnknown ErrorCode = iota + 100
+	// ErrTempWorkerRestart represents an error that occurred due to a worker restarting.
 	ErrTempWorkerRestart
+	// ErrTempAllocateSpace means that the worker dosen't have enough space in it's sector store for a new allocation.
```
nonsense, line 169:
```suggestion
	// ErrTempAllocateSpace means that the worker doesn't have enough space in its sector store for a new allocation.
```

node/impl/storminer.go
```diff
@@ -377,6 +377,8 @@ func (sm *StorageMinerAPI) SectorMarkForUpgrade(ctx context.Context, id abi.Sect
 	return sm.Miner.MarkForUpgrade(id)
 }
 
+// WorkerConnect is called by a worker to register itself with the miner.
+// The worker sends the url at which it's running it's RPC server.
```
nonsense, line 381:
```suggestion
// The worker sends the url at which it's running its RPC server.
```

nonsense (approved):
added a few fixes and suggestions. overall lgtm!

extern/sector-storage/manager.go
```diff
@@ -31,16 +31,25 @@ var ErrNoWorkers = errors.New("no suitable workers found")
...
+	// The storage directories located at these paths are managed by the Worker's local sector store.
 	Paths(context.Context) ([]stores.StoragePath, error)
 
+	// Info returns information about the worker's resources that is used by the Miner's scheduler
+	// to determine which worker to schedule tasks on.
 	Info(context.Context) (storiface.WorkerInfo, error)
 
+	// Session returns a random uuid generated by a worker on startup.
+	// It uniquely identifies a run of the worker and can be used to identify worker restarts.
 	Session(context.Context) (uuid.UUID, error)
```
raulk, line 53: Is this all it does? I would imagine the goal of this method is to actually start a new session with the worker, which could do other things. The UUID is just the identifier of the session.

extern/sector-storage/manager.go
```diff
@@ -187,6 +196,14 @@ func (m *Manager) AddLocalStorage(ctx context.Context, path string) error {
...
 }
 
+// AddWorker registers a worker with the miner's work scheduler.
+// We query the worker for it's current sessionID (a uuid generated by the worker on each startup)
+// to determine if we already know this current run of the worker and therefore discard it as a duplicate or if it's a new run and we need to add it
+// to our list of known workers.
+// We then repeatedly poll the worker's sessionID to detect restarts by identifying sessionID mismatches.
+// If we detect a worker restart, we will remove the current worker run from the list of known workers.
+// Upon restart, a worker will register itself again by calling `AddWorker` after generating a new session ID.
+// This polling basically acts as a heartbeat in the system.
```
raulk, line 206: I would suggest to separate implementation details from the godoc that we want exposed publicly. This reads too imperative for a godoc. It feels like you are describing logic, and not function. The logic may change any time, so it's better described _inside_ as inline comments in the method body. What we want to describe in a godoc is function, expectations, prerequisites, and results.

Consider this style:

```
AddWorker registers a worker with the miner's work scheduler.

This method safely deduplicates workers we already know about. Thus,
a given worker will never be able to doubly register. Workers are
identified by their runtime session ID, which is randomly generated
upon restart.

Once added, the Manager will monitor a Worker through a polling
heartbeat. If the worker is gone, or there's a session ID mismatch,
the manager drops the worker. If the worker was restarted, it will
register again under the new session ID.
```

extern/sector-storage/selector_alloc.go
```diff
@@ -26,6 +26,8 @@ func newAllocSelector(index stores.SectorIndex, alloc storiface.SectorFileType,
 	}
 }
 
+// allocSelector only allows selecting a worker which supports the given tasktype and has enough space in the configured path type(`s.ptype`) in
+// the sector store to store a sector file of type/s `s.alloc` for a sector with the given size(`spt.SectorSize()`).
```
raulk, line 30: This is too detailed / implementation specific for godocs. Private code references should be _inside_ the method body.

It also doesn't explain what allocSector actually does, it just expresses a constraint.

extern/sector-storage/selector_existing.go
```diff
@@ -28,6 +28,10 @@ func newExistingSelector(index stores.SectorIndex, sector abi.SectorID, alloc st
 	}
 }
 
+// existingSelector only allows selecting a worker which supports the given tasktype and
```
raulk, line 31: Same as with `allocSelector`.

extern/sector-storage/stores/index.go
```diff
@@ -51,15 +51,33 @@ type SectorStorageInfo struct {
 	Primary bool
 }
 
+// SectorIndex is the Miner's sector store Index/Registry that keeps track of all storage directories
+// and sector files in the sector stores of all Workers.
+// The Miner exposes this Index API over JSON RPC to all remote workers. Local workers can of-course call it from within the process.
```
raulk, line 56:
```suggestion
// The Miner exposes this Index API over JSON RPC to all remote workers. Local workers can query it too within the process, without resorting to a network call.
```

extern/sector-storage/stores/index.go
```diff
@@ -51,15 +51,33 @@ type SectorStorageInfo struct {
...
 type SectorIndex interface { // part of storage-miner api
+	// StorageAttach is called by workers to register a new storage directory or update information about an existing storage directory
+	// with the Index.
 	StorageAttach(context.Context, StorageInfo, fsutil.FsStat) error
+	// StorageInfo is used to get detailed information about a storage directory with the given ID.
```
raulk, line 61:
```suggestion
	// StorageInfo returns detailed information about a storage directory with the given ID.
```

extern/sector-storage/stores/index.go
```diff
@@ -51,15 +51,33 @@ type SectorStorageInfo struct {
 	Primary bool
 }
 
+// SectorIndex is the Miner's sector store Index/Registry that keeps track of all storage directories
+// and sector files in the sector stores of all Workers.
+// The Miner exposes this Index API over JSON RPC to all remote workers. Local workers can of-course call it from within the process.
 type SectorIndex interface { // part of storage-miner api
+	// StorageAttach is called by workers to register a new storage directory or update information about an existing storage directory
+	// with the Index.
```
raulk, line 59: This is great info to have, but let's separate function from usage. Consider:

```
// StorageAttach registers a new storage directory, or updates
// information about an existing storage directory within the Index.

// It is called by workers to provision storage directories.
```

extern/sector-storage/stores/index.go
```diff
@@ -51,15 +51,33 @@ type SectorStorageInfo struct {
...
 	StorageAttach(context.Context, StorageInfo, fsutil.FsStat) error
+	// StorageInfo is used to get detailed information about a storage directory with the given ID.
 	StorageInfo(context.Context, ID) (StorageInfo, error)
+	// StorageReportHealth is called by workers to report the health of a registered storage directory.
+	// This is the mechanism my which the Index tracks healthy/unhealthy/available/unavailable storage directories across workers.
```
raulk, line 64: Same as above, let's separate function from usages. A problem with this style is that this is a public API, and it may get used in other ways in the future, which would render this documentation obsolete immediately. So the better style here is:

```
StorageReportHealth records the health ... (function)

It is used by ..., at least. (usage known now)
```

extern/sector-storage/stores/index.go
```diff
@@ -51,15 +51,33 @@ type SectorStorageInfo struct {
...
 	StorageReportHealth(context.Context, ID, HealthReport) error
 
+	// StorageDeclareSector is called by workers to register new sector files or update information about
+	// existing sector files with the Index.
+	// `primary` is set to true if the sector files are in the long term storage of the worker.
+	// `primary` is set to false if the sector files are in the sealing scratch space of the worker.
```
raulk, line 70: This reads imperative and expresses usage. It doesn't explain what is the function of `primary`. What does it stand for?

```
`primary` specifies the nature of storage. If true, it indicates the sector files are in the long-term storage of the worker. Otherwise, it indicates they're in its scatch space.
```

extern/sector-storage/stores/local.go
```diff
@@ -353,6 +353,8 @@ func (st *Local) reportStorage(ctx context.Context) {
 	}
 }
 
+// Reserve reserves space for the given sector file types on the given local storage directories after accounting for the overhead incurred by storing
```
raulk, line 356: This is good style!

nit: line wrapping?

extern/sector-storage/stores/local.go
```diff
@@ -414,7 +416,13 @@ func (st *Local) Reserve(ctx context.Context, sid storage.SectorRef, ft storifac
 	return done, nil
 }
 
-func (st *Local) AcquireSector(ctx context.Context, sid storage.SectorRef, existing storiface.SectorFileType, allocate storiface.SectorFileType, pathType storiface.PathType, op storiface.AcquireMode) (storiface.SectorPaths, storiface.SectorPaths, error) {
+// AcquireSector is used to perform two functions for a sector with the given ID:
```
raulk, line 419:
```suggestion
// AcquireSector performs two operations for a sector with the given ID:
```

extern/sector-storage/manager.go
magik6k: Yes, it only returns the current worker session ID, which the worker generates on process startup (kind of like `/proc/sys/kernel/random/boot_id`)

extern/sector-storage/ffiwrapper/types.go
```diff
@@ -42,6 +42,7 @@ type Verifier interface {
 }
 
 type SectorProvider interface {
+	// TODO This documentation is probably incorrect.
 	// * returns storiface.ErrSectorNotFound if a requested existing sector doesn't exist
```
aarshkshah1992, line 46: @magik6k 

```
returns storiface.ErrSectorNotFound if a requested existing sector doesn't exist
```

This is probably wrong because the Remote Sector store implementation fetches the `existing` sector files types from other workers if it does NOT have them locally.

extern/sector-storage/stores/index.go
magik6k: `primary` means the "main" copy of the sector, and can be true for sector files still in stealing storage.

This is mostly used for cleanup purposes - the `store.RemoveCopies` method will look at sector files, and if there is a primary sector file - remove the non-primary files (normally there shouldn't be any stray sector files, but copies can be left in sealing paths when things get shutdown forcefully)

extern/sector-storage/stores/index.go
```diff
@@ -51,15 +51,33 @@ type SectorStorageInfo struct {
...
+	// StorageDropSector is called by workers to drop sector files from the Index.
 	StorageDropSector(ctx context.Context, storageID ID, s abi.SectorID, ft storiface.SectorFileType) error
+	// StorageFindSector is called by workers to lookup storage directories containing the given sector files.
+	// If `allowFetch` is set to true, the Index will also return storage directories that do NOT yet have the sector files
+	// but can store them after fetching them from other storage directories that have them.
 	StorageFindSector(ctx context.Context, sector abi.SectorID, ft storiface.SectorFileType, ssize abi.SectorSize, allowFetch bool) ([]SectorStorageInfo, error)
-
+	// StorageBestAlloc is called by workers to lookup storage directories that can store sector files of the given types for a given sector.
+	// `pathType` indicates the storage capabilities(sealing/long term storage) that a storage directory should have to be considered eligible.
 	StorageBestAlloc(ctx context.Context, allocate storiface.SectorFileType, ssize abi.SectorSize, pathType storiface.PathType) ([]StorageInfo, error)
```
magik6k, line 81: Worth noting that the results are returned sorted by preference (first result is best)

extern/sector-storage/stores/local.go
```diff
@@ -560,6 +570,8 @@ func (st *Local) Remove(ctx context.Context, sid abi.SectorID, typ storiface.Sec
 	return nil
 }
 
+// RemoveCopies removes all copies of the given sector file types for the given sector from the local sealing scratch space.
+// However, it does NOT remove the "primary" copies of those sector files i.e. sector files that are stored in the local long term storage space.
```
magik6k, line 574: Primary copies can also live in sealing storage while the sector is going through the sealing pipeline. "Primary" basically just marks the copy which shouldn't be removed by this method.

extern/sector-storage/storiface/worker.go
```diff
@@ -64,9 +75,14 @@ type WorkerJob struct {
 	Hostname string `json:",omitempty"` // optional, set for ret-wait jobs
 }
 
+// CallID uniquely identifies a task submitted by a miner to a worker.
+// It is generated by the worker when it receives a RPC request to perform a task and is returned
+// to the miner as a response to the RPC call so it can use it to track the status of the task.
```
magik6k, line 80:
```suggestion
// CallID uniquely identifies a task submitted by a miner to a worker.
// It is generated by the worker when it receives a RPC request to perform a task and is returned
// to the miner as a response to the RPC call so it can use it to track the status of the task across node restarts.
```

extern/sector-storage/storiface/worker.go
magik6k: > The call will fail of the Worker is NOT able to allocate space to store the Unsealed sector file

Eh, that's a bug - https://github.com/filecoin-project/lotus/issues/4069 (probably don't want to document it)

magik6k: (applies below as well)

magik6k: Ah, I guess this is correct, we're just not handling this error in the main scheduler correctly

extern/sector-storage/storiface/worker.go
```diff
@@ -77,33 +93,85 @@ var _ fmt.Stringer = &CallID{}
...
+
+	// FinalizeSector schedules a task to perform the `FinalizeSector` stage of a sector being sealed.
+	// As a part of this task, the worker fetches(if required) and stores the cached sealing file on one of the long term
+	// storage directories in the local sector store.
+	// It fetches/keeps around the  Unsealed file in the  Sealing scratch space if `keepUnsealed` is non-zero.
+	// The Unsealed file will be moved to long term storage later when the Miner calls `MoveStorage`.
+	// If `keepUnsealed` is set to zero, it will remove the Unsealed file locally and from all other workers that have it.
 	FinalizeSector(ctx context.Context, sector storage.SectorRef, keepUnsealed []storage.Range) (CallID, error)
+
+	// TODO: ReleaseUnsealed has still not been implemented.
```
magik6k, line 138: This is meant to either remove the unsealed file if no ranges to be kept are passed or deallocate parts of the file which we don't need (using `FALLOC_FL_PUNCH_HOLE` / partialfile.Free / https://github.com/filecoin-project/lotus/blob/master/extern/sector-storage/fsutil/dealloc_linux.go#L14)

magik6k (commented):
This is great!

extern/sector-storage/ffiwrapper/types.go
magik6k: Yeah, the doc isn't correct, it probably has been here before we had remote storage support

magik6k:
Probably reasonable to say that this won't land at this point

PR #6282, opened 2021-05-18
FIP 13 addenda: correctly handle commit batch timer

vyzo:


extern/storage-sealing/commit_batch.go
```diff
@@ -114,6 +118,79 @@ func (b *CommitBatcher) run() {
...
+
+	wait := deadline.Sub(now)
+	if wait > maxWait {
+		wait = maxWait
+	}
+
+	return wait
+}
+
+func (b *CommitBatcher) getSectorDeadline(sn abi.SectorNumber) time.Time {
```
magik6k, line 158: I'd cache those as we can have 100s of sectors each with 100s of pieces, this will be doing a lot of RPC queries on each loop

extern/storage-sealing/commit_batch.go
```diff
@@ -192,12 +270,14 @@ func (b *CommitBatcher) processBatch(notif, after bool) (*cid.Cid, error) {
 }
 
 // register commit, wait for batch message, return message CID
-func (b *CommitBatcher) AddCommit(ctx context.Context, s abi.SectorNumber, in AggregateInput) (mcid cid.Cid, err error) {
+func (b *CommitBatcher) AddCommit(ctx context.Context, s SectorInfo, in AggregateInput) (mcid cid.Cid, err error) {
+	sn := s.SectorNumber
 	b.lk.Lock()
-	b.todo[s] = in
+	b.sectors[sn] = s
```
magik6k, line 276: As mentioned above, maybe store sector deadline instead of sector info (the deadline shouldn't change)

magik6k (commented):
Looks quite good, just 1 note

The config values also need to be wired through in `node/modules/storageminer.go ~L820~850`

extern/storage-sealing/commit_batch.go
vyzo: yeah, good point!

extern/storage-sealing/commit_batch.go
vyzo: totally.

extern/storage-sealing/commit_batch.go
```diff
@@ -114,6 +119,74 @@ func (b *CommitBatcher) run() {
...
+
+		proposal, err := b.api.StateMarketStorageDealProposal(b.mctx, p.DealInfo.DealID, tok)
+		if err != nil {
+			log.Errorf("getting deal proposal for %d: %s", p.DealInfo.DealID, err)
+			continue
+		}
+
+		if proposal.StartEpoch < deadlineEpoch {
+			deadlineEpoch = proposal.StartEpoch
+		}
```
magik6k, line 180: We should be able to use `p.DealInfo.DealSchedule.StartEpoch` instead of calling RPC

magik6k (commented):
One more thing

extern/storage-sealing/commit_batch.go
vyzo: ah great!

magik6k (approved):

PR #6304, opened 2021-05-20
fix ticket expired

yaohcn:


magik6k (approved):
Thanks! I think this looks good.

Can you try adding a test case to `fsm_test.go`?

magik6k (approved):

PR #6305, opened 2021-05-20
Allow starting networks from arbitrary actor versions

### LINKED ISSUE #5817: Allow starting networks from arbitrary actor versions
First step is to support generating genesis files from not-v0 actors.

Motivation:
The primary motivation here is to be able to start networks from arbitrary actor versions, which allows other implementations of the Filecoin protocol to be able to sync a chain even if they only have support from, say, v3 actors and onwards. More broadly, this advances the goal of making the Lotus codebase and utils less mainnet-specific. One could start a chain from any actor's version and so should be able to, even if the actual Filecoin mainnet started from v0.
Design and steps:
- Get familiar with the genesis creation code in `chain/gen/genesis.go`. The key methods are `MakeGenesisBlock` and `MakeInitialStateTree`.
- Tweak the code to receive an actors version as a parameter (`chain/actors/version.go` for actors versions). Maybe this goes into the genesis template in `genesis/types.go`.
- Based on provided actors version, use the appropriate version of specs-actors when constructing state tree (`bultion0.AccountActorCodeID -> builtin4.AccountActorCodeID`, for example)
- Start the network from the appropriate network version (e.g. network v14 for actors v4). Bit unclear how to do this, probably requires tweaking how we set up the stage in the build `params` file (see the `NewStateManagerWithUpgradeSchedule` method).
- Iron out remaining kinks
Testing:
The best way to test this out would be to run a devnet that starts from a non-v0 actors version. Doing so will require tweaking the upgrade epoch heights in params_2k.go. The next step will be to kick off the desired `interopnet` using this new feature, and seeing if that works well.

arajasek:
Fixes #5817 

Allowed me to get to height 1500 on a devnet starting from v4 actors, so I'm gonna call this a win.


build/params_2k.go
```diff
@@ -24,20 +24,20 @@ var UpgradeIgnitionHeight = abi.ChainEpoch(-2)
 var UpgradeRefuelHeight = abi.ChainEpoch(-3)
 var UpgradeTapeHeight = abi.ChainEpoch(-4)
 
-var UpgradeActorsV2Height = abi.ChainEpoch(10)
-var UpgradeLiftoffHeight = abi.ChainEpoch(-5)
+var UpgradeActorsV2Height = abi.ChainEpoch(-5)
```
arajasek, line 27: Changed these thinking we could make devnets start from latest nv by default, but can undo that.

chain/actors/builtin/miner/state.go.template
```diff
@@ -366,6 +376,45 @@ func (s *state{{.v}}) decodeSectorPreCommitOnChainInfo(val *cbg.Deferred) (Secto
 	return fromV{{.v}}SectorPreCommitOnChainInfo(sp), nil
 }
 
+func (s *state{{.v}}) EraseAllUnproven() error {
```
arajasek, line 379: This is the only objectively horrible hack I had to add to make this work.

arajasek: At least the only one that isn't super easy to remove later.

chain/vm/vm.go
```diff
@@ -662,37 +662,6 @@ func (vm *VM) Flush(ctx context.Context) (cid.Cid, error) {
 	return root, nil
 }
 
-// MutateState usage: MutateState(ctx, idAddr, func(cst cbor.IpldStore, st *ActorStateType) error {...})
-func (vm *VM) MutateState(ctx context.Context, addr address.Address, fn interface{}) error {
```
arajasek, line 666: Quite happy that this is dead.

chain/gen/genesis/util.go
```diff
@@ -49,29 +46,3 @@ func doExecValue(ctx context.Context, vm *vm.VM, to, from address.Address, value
 
 	return ret.Return, nil
 }
-
-// TODO: Get from build
-// TODO: make a list/schedule of these.
-var GenesisNetworkVersion = func() network.Version {
```
arajasek, line 55: Happy this is gone

arajasek:
This is against `master` and so has v4 actors as latest, will probably open the same against `feat/nv13` to start up networks from v5.

chain/actors/builtin/account/actor.go.template
```diff
@@ -34,8 +35,30 @@ func Load(store adt.Store, act *types.Actor) (State, error) {
...
+    }
+
+	return cid.Undef, xerrors.Errorf("unknown actor version %d", av)
+}
+
 type State interface {
 	cbor.Marshaler
 
 	PubkeyAddress() (address.Address, error)
+    GetState() interface{}
```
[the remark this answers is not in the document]
arajasek: I hate you

chain/actors/agen/main.go
```diff
@@ -8,31 +8,30 @@ import (
...
-
 var versionImports = map[int]string{
-	0:             "/",
-	2:             "/v2/",
-	3:             "/v3/",
-	latestVersion: "/v4/",
+	0:                         "/",
+	2:                         "/v2/",
+	3:                         "/v3/",
+	lotusactors.LatestVersion: "/v4/",
```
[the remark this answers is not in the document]
arajasek: Done

chain/actors/builtin/account/actor.go.template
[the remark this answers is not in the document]
arajasek: Also done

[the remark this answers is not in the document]
ec2: ugh spaces...

magik6k (approved):

Issue #6260, opened 2021-05-14
[Mining Issue] mineOne: unexpectedly nil MiningBaseInfo


mezerotm:
### Problem

```
{"level":"warn","ts":"2021-05-14T18:25:06.070Z","logger":"miner","caller":"miner/miner.go:429","msg":"mineOne: unexpectedly nil MiningBaseInfo for round 757011, off tipset 757010/{bafy2bzacebrvdoawvmkr4e7qnddkcsojgmcgvm4np6dn2y6e2kqtl3q7laxu,bafy2bzaceajnd4335cmh4ufbiouvjvelehjgkw7xwr2w6jyxz7egiy43pyk2g,bafy2bzaced7ybzp77gpilvbdrimkypkwbtndyqzp4cj2447bwkekvzvq6mmei,bafy2bzaceds6s7m7auavpvwpz47wzerliyajofybrvq3nvg3uknrrvsk7im5o,bafy2bzaceakmvn7gqxxa7dyjeagsctchwhbq2cotlzfi27indecbjrn5cqf36,bafy2bzacebubw2jklzu4s4ewmhqx4lk3gamxp57wom2buyfos2hjin25dejc}"}
```

### Version
lotus version 1.11.0-dev+mainnet+git.bf10b051f
lotus-miner version 1.11.0-dev+mainnet+git.bf10b051f
lotus-worker version 1.11.0-dev+mainnet+git.bf10b051f

### Setup

CPU: 16 Core AMD Epyc 7281 16c/32t
RAM: 256 gb 2666 mhz DDR4 ECC Ram (8x32gb)
Boot: 1tb Western Digital NVME SSD
GPU: Nvidia GT 710
Motherboard: ASUS KNPA U-16
LAN: 10 gbps
WAN: 1gbps down, 50 mbps up
RAID1: 2x 8tb Seagate Exos 7e8

**notes**
- The port forward is set up and working
- The net reachability will change from Public to Private seemingly at random, but this issue still exists regardless of the state of net reachability

**lotus-miner actor control list**
```
name       ID        key           use        balance
owner      f027197  f3w4x6ugc...             0.210965460834167905 FIL
worker     f082765  f3vjzy3ep...  other      0.582264902988507581 FIL
control-0  f082734  f3w2rkq5y...  precommit  1.2 FIL
control-1  f083206  f3wxorh6x...  commit     0.5 FIL
```

---
**environment**
```
export PATH=$PATH:/usr/local/go/bin
[...]

PR #6310, opened 2021-05-21
Fix logging around mineOne

ribasushi:
- A nil MiningBaseInfo is *NOT* unexpected: it happens when one is in penalty
  https://github.com/filecoin-project/lotus/blob/v1.9.0/chain/stmgr/utils.go#L500-L502

- Remove the log from IsRoundWinner(): all we care about is the randbase epoch

Addresses https://github.com/filecoin-project/lotus/issues/6260

### DIFF
--- a/chain/gen/gen.go
+++ b/chain/gen/gen.go
@@ -3,7 +3,6 @@ package gen
 import (
 	"bytes"
 	"context"
-	"encoding/base64"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


magik6k (commented):
Linter complains

jennijuju:
Thanks for the quick fix! @ribasushi  We should get this in as soon as possible as it's confusing some people.

jennijuju:
This should get merged to master and also cherry-picked to the v1.10.0

magik6k (approved):

PR #6311, opened 2021-05-21
consolidate integration tests into `itests` package; create test kit; cleanup

raulk:
Integration tests in Lotus are pretty coarse-grained. They exercise the whole Lotus stack top to bottom... i.e. they act like end-to-end tests.

They are quite a labyrinth too, as they are scattered between packages, and there's quite a bit of indirection. Folks that wrote new test cases ended up reinventing the wheel (mining loops, builders, etc.) because existing reusable code was hard to discover.

This PR aims to bring some initial order to the chaos.

This PR:
- moves all integration tests to the `itests` package.
- separates the testing utilities from the test cases themselves (introducing an `itests/kit`) package.


raulk:
This is ready for final review and merge @magik6k 🎉  The merge was quite a pain, but finally done. Not sure if the `test` test suite will pass, but it's been running for 15min now and it looks like it's going to time out. I've checked master builds and this seems to be general flakiness. Happy to look at that once this is on master.

raulk:
🟢🟢🟢🟢🟢

magik6k (approved):
Just one comment, not blocking merge on it as it looks like currently we're not including this package outside of tests, but it should be addressed a followup PR

PR #6280, opened 2021-05-18
Bypass task scheduler for reading unsealed pieces

aarshkshah1992:
Picks the minimal set of changes we need from @magik6k's PR at https://github.com/filecoin-project/lotus/pull/6128.

### TODO
- [x] Unit Tests.
  - [x] Test changes to the http handler to check presence of unsealed pieces and to get sector files.
  - [x] Test the `Reader` API on the Remote Store that tries to read an unsealed piece from an already unsealed file on any of the workers.
  - [x] Unit tests for the newly introduced `SectorUnseal` API in the Manager.
  
- [x] Integration Tests.
  - [x] ReadPiece for piece in unsealed file on local worker
  - [x] ReadPiece for piece in sealed file on local worker
  - [x] ReadPiece for piece in unsealed file on remote worker
  - [x] ReadPiece for piece in sealed file on remote worker

- [x] Testing on a devnet/pond with 8MB sector size.
  - [x] Test storage / retrieval
    - Create storage deal
    - Retrieve deal
  - [x] Test storage / retrieval with unsealing
    - Create storage deal
    - Remove unsealed file
    - Retrieve deal (so it unseals)
    - Retrieve deal (so it's freshly unsealed)
  - [x]  Test workers
    - Disable unsealing on miner
    - Enable all tasks for worker (so unsealing must happen on worker)
    - Remove unsealed file
    - Retrieve deal (so it unseals)
    - Retrieve deal (so it's freshly unsealed)

- [x] Manual test against mainnet miner (running single worker)

PR #6313, opened 2021-05-21
Read Pieces by bypassing the scheduler and add extra logging for retrievals

dirkmc:
Adds extra logging to retrieval deals on miner.

Branched from https://github.com/filecoin-project/lotus/pull/6280


raulk:
@dirkmc I'm tentatively closing this PR because I suspect that the extra logging added here was also cherry-picked onto #6280 (already merged). If it's not the case, let's open a new PR with only the logging changes, since #6280 is already on master 🎉

PR #6314, opened 2021-05-21
decouple rpc server and shutdown logic from cmd.

raulk:
The RPC setup logic is duplicated in various places, concretely in various tests suites.

This PR places the RPC setup logic inside the `node` package, and decouples the process lifecycle from it.

I believe this PR respects the previous exit code interface (which is easy to break because not elicited).


raulk:
Putting this back into draft mode temporarily, as I'd like to perform further testing on it.

PR #6316, opened 2021-05-24
Fix shell completions

Aloxaf:
Fix completions according to https://github.com/urfave/cli/blob/master/docs/v2/manual.md#distribution-and-persistent-autocompletion.


magik6k (approved):

magik6k:
Thanks for the PR, works great!

Issue #5151, opened 2020-12-08
Allow an address to receive more than one DataCap assignment


jnthnvctr:
Client's can currently receive DataCap from a Notary at a given address. After spending that DataCap, if a Client wants more they have to provide the Notary with a new address in order to receive a new allocation of DataCap. 

This introduces complexity for apps / users that require "top-ups" of DataCap. As an example - for an app like Slate they would need to behind the scenes monitor DataCap and usage for users (and swap out the addresses receiving and storing DataCap).

### EDIT(@jennijuju)
Currently, datacap is assigned by notaries by using `./lotus-shed verifreg`, which should be change to the `./lotus-shed notary` terminalogy and also the cmd should be moving to a lotus cmd.

Instead of making a client verified client and then send a one-time upfront allocation, simply do a `./lotus notary send-datacap` to allocate DataCap to a client account.
 
blocked by https://github.com/filecoin-project/specs-actors/issues/1325

jennijuju:
Reopen the ticket when [FIP 0012](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0012.md) is accepted and implemented in spec actor

jennijuju:
Reopening the issue as fip is now accepted and the actor implementation is ready https://github.com/filecoin-project/specs-actors/pull/1390

jennijuju:
Blocked by getting v4 actor dependency.

jennijuju:
blocked by https://github.com/filecoin-project/lotus/issues/6022

BigLep:
Remaining work:
1. Unit test in Lotus
2. Get UX feedback from Product people and do anything requested (https://github.com/filecoin-project/lotus/issues/6194)

We can't close this until actors v5 is in master.

PR #6317, opened 2021-05-24
Add test for AddVerifiedClient

jennijuju:
In preparation of https://github.com/filecoin-project/lotus/issues/5151#issuecomment-841416830 for v5

### DIFF
--- a/api/test/verifreg.go
+++ b/api/test/verifreg.go
@@ -2,9 +2,9 @@ package test
 
 import (
 	"context"
-	"fmt"
-	"os"
 	"strings"
+
+	lapi "github.com/filecoin-project/lotus/api"
 
 	"github.com/filecoin-project/lotus/chain/actors"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/verifreg"
@@ -15,12 +15,8 @@ import (
 	"time"
 
 	"github.com/filecoin-project/go-state-types/big"
-	lapi "github.com/filecoin-project/lotus/api"
-	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain/types"
-	logging "github.com/ipfs/go-log/v2"
 )
-
 
 func AddVerifiedClient(t *testing.T, b APIBuilder) {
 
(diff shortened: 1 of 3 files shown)


node/test/builder.go
```diff
@@ -276,12 +280,24 @@ func mockBuilderOpts(t *testing.T, fullOpts []test.FullNodeOpts, storage []test.
...
+		vrk = gen.DefaultVerifregRootkeyActor
+	}
+
 	templ := &genesis.Template{
 		Accounts:         genaccs,
 		Miners:           genms,
 		NetworkName:      "test",
 		Timestamp:        uint64(time.Now().Unix() - 10000), // some time sufficiently far in the past
-		VerifregRootKey:  gen.DefaultVerifregRootkeyActor,
+		VerifregRootKey:  vrk,
```
magik6k, line 300: We could just make this the first miner key (`genaccs[0]`), this way we don't need the separate constructors

api/test/verifreg.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+func init() {
+	logging.SetAllLoggers(logging.LevelInfo)
+	err := os.Setenv("BELLMAN_NO_GPU", "1")
+	if err != nil {
+		panic(fmt.Sprintf("failed to set BELLMAN_NO_GPU env variable: %s", err))
+	}
+	build.InsecurePoStValidation = true
+}
```
magik6k, line 31: This will mess with other tests

```suggestion
```

node/node_test.go
```diff
@@ -259,3 +259,9 @@ func TestDeadlineToggling(t *testing.T) {
 
 	test.TestDeadlineToggling(t, builder.MockSbBuilder, 2*time.Millisecond)
 }
+
+func TestVerifiedClientTopUp(t *testing.T) {
+	logging.SetLogLevel("storageminer", "FATAL")
+	logging.SetLogLevel("chain", "ERROR")
+	test.AddVerifiedClient(t, builder.Builder)
```
magik6k, line 266: This makes it use fake (much faster) proofs
```suggestion
	test.AddVerifiedClient(t, builder.MockSbBuilder)
```

magik6k (approved):

PR #6323, opened 2021-05-25
separate tracing environment variables

coryschwartz:
The reason for this PR is to make it easier to setup metrics in kubernetes environments.

in kubernetes, the IP address of the agent is not known until the pod starts, after which it can be made available as an environment variable. By that time, it is too late to append the port number.  When the agent endpoint is read in with a single environment variable, the best you can do is combine the host and port numbers as a bash script prior to starting lotus.

With this PR, the agent host and port can be read in with two environment variables without any bash shenanigans. 

Also, I added environment variables to support jaeger authentication, or sending directly to the collector.


magik6k (commented):
(lint fails)

We should also update `documentation/en/jaeger-tracing.md`

magik6k (approved):

PR #6333, opened 2021-05-26
Allow starting networks from arbitrary actor versions

### LINKED ISSUE #5817: Allow starting networks from arbitrary actor versions

arajasek:
Fixes #5817. Again.


node/impl/client/client.go
```diff
@@ -96,7 +96,13 @@ func calcDealExpiration(minDuration uint64, md *dline.Info, startEpoch abi.Chain
 	minExp := startEpoch + abi.ChainEpoch(minDuration)
 
 	// Align on miners ProvingPeriodBoundary
-	return minExp + md.WPoStProvingPeriod - (minExp % md.WPoStProvingPeriod) + (md.PeriodStart % md.WPoStProvingPeriod) - 1
+	exp := minExp + md.WPoStProvingPeriod - (minExp % md.WPoStProvingPeriod) + (md.PeriodStart % md.WPoStProvingPeriod) - 1
+	// Should only be possible for miners created around genesis
+	for exp < minExp {
```
arajasek, line 101: I...guess?

magik6k (approved):
Seems to actually work in devnets this time

PR #6337, opened 2021-05-27
Add a command to list retrievals

hannahhoward:
# Goals

Especially in automated environments with lots of concurrent retrievals (see dealbot), the lack of a command monitor all retrievals becomes a pain point. My recollection is we didn't build this original cause we assumed "well retrievals will be fast so you wont want to monitor them" -- but this isn't neccesarily true -- large retrievals on slow connections can take a while, especially if you're running lots in parallel

# Implementation

-- v1API (assuming I should add here cause it's a new, untested API) --
   -- Add ClientListRetrievals
   -- Add ClientGetRetrievalUpdates
 -- cli
   -- add `lotus client list-retrievals` -- based on `lotus client list-deals` and `lotus client list-transfers`
 
 Do we have a way we are testing commands right now (other than by hand)? If not I will run against some lotus nodes I am working with.


jennijuju:
@hannahhoward should we add it to the v0 API package too?

jennijuju:
@hannahhoward is this an API that you want to be available for most users? If so, i thinkk it should be added to the v0 api package as v0 will still be the default for several months!

dirkmc:
I'd suggest we show deals that errored out by default (with a flag to hide).

It seems that people often get confused that their failed deal doesn't show up in list-deals, list-transfers etc. eg see https://github.com/filecoin-project/lotus/issues/6346

hannahhoward:
updated to have show-failed as default true

dirkmc (approved):

magik6k (approved):

PR #6350, opened 2021-05-28
Extend the default deal start epoch delay

### LINKED ISSUE #6277: Extend the default deal start epoch delay
With MaxProveCommitDuration got extended 1 day -> 6 days in actor v5
https://github.com/filecoin-project/specs-actors/blob/a389ccae8d5df456ee6b93f3fc836717b0a7c8d9/actors/builtin/miner/policy.go#L176

We may presume miners may wait longer before submiiting a provecommit message to aggreate as many as many proofs as possible, therefore,  lotus should extend the default deal start epoch delay to reduce the chances of deal getting expired by the time provecommitment lands on chain.

Suggested value to be 8 days (publishstoragedeal -> seal duration + pre commit + MaxProveCommitDuration + 8 hour buffer)

arajasek:
Closes #6277

### DIFF
--- a/node/impl/client/client.go
+++ b/node/impl/client/client.go
@@ -62,7 +62,8 @@ import (
 
 var DefaultHashFunction = uint64(mh.BLAKE2B_MIN + 31)
 
-const dealStartBufferHours uint64 = 49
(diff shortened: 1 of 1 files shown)


magik6k (approved):
(we should get this configurable)

PR #6354, opened 2021-05-29
[wip] switch networks with the same binary

coryschwartz:
This isn't in a mergable state, but comments are welcome.

This change would make lotus easier to manage in a few ways.

1. newcomers trying out lotus can switch networks without recompiling
2. lotus packaging (as docker images, AMIs, or digitalocean droplets, for example) is simplified.
3. The upgrade and testing processes are improved because the exact same image can be used in test networks as would run in production networks.
4. If, one day, lotus is to be in package managers or otherwise distributed, it should be called `lotus` and not `lotus-mainnet`, `lotus-butterfly`...

wihtout this change, network parameters are `const` to provided protection from changes. The protection from some of these consts is superficial. For example, `GenesisFile` is const, but this is only the filename. The genesis block itself is mutable after it is returned from the rice box.

Other parameters such as the upgrade schedule for block heights are also const.

Moving these parameters to a getter pattern, I think, little downside by comparison. Outside of the "build" package, the upgrade schedule, genesis filename and other parameters are unexported and there is no way to set them by accident. Also, the return value of the `NetworkParams()` func in parameters.go cannot be changed so the entire set of network parameters cannot be easily changed. This is pretty good protection agasint any accidental changes to the upgrade schedule, and I think the protection is similar to const values.


build/parameters.go
```diff
@@ -1,7 +1,147 @@
...
+
+type sharedNetworkPrams interface {
+}
+
+type combinedNetworkParams interface {
+	perNetworkParams
+	sharedNetworkPrams
+}
+
+func init() {
```
coryschwartz, line 61: This init() has all of the original init() functions in it.

jennijuju:
Hey @coryschwartz = @magik6k and @arajasek believe this will bring us to moon and will provide feedback soon!

PR #6041, opened 2021-04-14
storagefsm: Fix batch deal packing behavior

magik6k:
Fixes #6013
Fixes #5077
Fixes #6010

PR #6355, opened 2021-05-30
events: Fix handling of multiple matched events per epoch

### LINKED ISSUE #2249: Storage deals stuck in `StorageDealSealing`
**Describe the bug**
While stress testing storage and retrieval deals with 300 concurrent deals per client, we _always_ end up with the clients stuck in `StorageDealSealing` without making any progress

**To Reproduce**
https://github.com/filecoin-project/oni/pull/81

**Expected behavior**
Storage deals should make progress!

**Version (run `lotus --version`):**
lotus master as of this morning: `lotus v0.4.2-0.20200703005127-9d56dabb316f`

**Additional context**
Investigating the goroutine dumps from one of the miners, there is an inordinate number of goroutines stuck in `github.com/filecoin-project/go-statemachine.(*StateMachine).run`, stuck in `github.com/filecoin-project/go-statemachine@v0.0.0-20200612181802-4eb3d0c68eba/machine.go:53`.
There is also a number of goroutines stuck in `github.com/filecoin-project/sector-storage.(*scheduler).Schedule`, line `github.com/filecoin-project/sector-storage@v0.0.0-2020063018
0318-4c1968f62a8f/sched.go:108`

Will attach full goroutine dump

### LINKED ISSUE #5377: deal state stuck in StorageDealAwaitingPreCommit,miner state is StorageDealActive
**Describe the bug**
deal state stuck in StorageDealAwaitingPreCommit,miner state is StorageDealActive

**To Reproduce**
lotus client deal ...

**Expected behavior**
expect lotus deal state: StorageDealAwaitingPreCommit -> StorageDealActive

**Screenshots**

image

1.4.0+git.e9989d0e4+api1.0.0

**Additional context**
thanks

### LINKED ISSUE #5890: Online deals stuck in StorageDealStaged (batching 2 deals)
**Describe the bug**
I have two cases where I have batched the publishdeals message with two 32GiB deals. After the message is published, it starts sealing one of the two deals in a new sector. The other deal is left in StorageDealStaged. It will stay there until the next deal is pushed in a publishdeal message. Next deal is then waiting in StorageDealStaged, while the older second deal from the batch is now sealing. So the miner is getting one deal behind everytime deals a batched. 

Only workaround is to publish every deal as one, so it does not pile up. I now have to outstanding deals, but they are all still moving forward, just with an offset of two deals. 

```
Daemon:  1.5.3+mainnet+git.3e6dcb578.dirty+api1.0.1
Local: lotus-miner version 1.5.3+mainnet+git.3e6dcb578.dirty
```
(dirty because of GPU2)

**To Reproduce**
1. Create two 32GiB size deals
2. Wait for both deals to be ready for publish
3. Publish both deals

**Expected behavior**
First deals should move to AP and PC1, then the next deal should move to AP and PC1. 

**Screenshots**
```
...lozcljj4  1635933  StorageDealAwaitingPreCommit             f3qpxctorx5cvsfz4emzl5tgraiqavzkdazul475hpe4tdjc6mydnunt4betsweons4fwyd7d54vcvbcskx4la  32GiB     0 FIL                     1458472
...ada3jiya  1636650  StorageDealStaged                        f3qpxctorx5cvsfz4emzl5tgraiqavzkdazul475hpe4tdjc6mydnunt4betsweons4fwyd7d54vcvbcskx4la  32GiB     0 FIL                     1457964
...ufl7nkce  1636651  StorageDealStaged                        f3qpxctorx5cvsfz4emzl5tgraiqavzkdazul475hpe4tdjc6mydnunt4betsweons4fwyd7d54vcvbcskx4la  32GiB     0 FIL                     1457373
```
**Additional context**
https://filfox.info/en/message/bafy2bzacebgsg655ujibbygurkakkzmkwyvfqpbxwdxfvsq2j5h4bgmj42m5c
https://filfox.info/en/message/bafy2bzaceboiciopeu5hlvy5pnk2wtftcp2z3rdgi3l6iopxxcsmkqtjq5w2e

### LINKED ISSUE #6123: [BUG] Deal stuck in StorageDealAwaitingPreCommit on miner, even though sector that includes the deal is Proving
**Describe the bug**
The `lotus-soup` testground end-to-end test has to ability to send concurrent deals from multiple clients to the same miner. When that happens, sometimes some of the deals are blocked in StorageDealAwaitingPreCommit state, even though the sectors that the miner included those deals in, passed through all stages, up to the final Proving state.

It appears that the problem is with the `SectorCommittedManager.OnDealSectorPreCommitted` function as well as the `messageEvents` type. It appears that the miner processes PreCommit messages for some deals (as there are multiple concurrent deals submitted to it), but not for all deals, and the messages do not match the expected parameters, for example:

For every deal we inspect messages on chain, but somehow we don't see all messages at:
```
		// Check through the deal IDs associated with this message
		for _, did := range params.DealIDs {
			if did == res.DealID {
```

---

At https://github.com/filecoin-project/lotus/blob/master/markets/storageadapter/ondealsectorcommitted.go , the `matched` callback is called for every PreCommit message, however the `called` callback is not - so if the DealID does not match one of the `params.DealIDs`, the deal gets stuck.

---

https://github.com/filecoin-project/lotus/commit/4688da51781a1f3249b534e962f8824f323607e3

**To Reproduce**
1. Trigger testground test plan (part of CI), with removed randomized sleep for different clients: https://github.com/filecoin-project/lotus/blob/master/testplans/lotus-soup/deals_e2e.go#L80
2. Note that some client deals are blocked in the StorageDealAwaitingPreCommit state, and never succeed, resulting in the test timing out.

**Expected behavior**
It is expected that all client deals succeed within 4-5min.

magik6k:
Got a reliable repro for the deals getting stuck in StorageDealAwaitingPreCommit issue when trying to debug https://github.com/filecoin-project/lotus/pull/6041 with rr (which runs with one thread, and makes everything slower).

Basically, we would ignore (Pre)Commit messages if there was more than one matching message from a miner in a given epoch. Normally we wouldn't hit this very often, but it was pretty common case of test flakiness. 

Fixes #6123
Fixes #5890
Fixes #5377
Fixes #2249 (probably)

### DIFF
--- a/chain/events/events_called.go
+++ b/chain/events/events_called.go
@@ -144,8 +144,10 @@ func (e *hcEvents) processHeadChangeEvent(rev, app []*types.TipSet) error {
 
 		// Queue up calls until there have been enough blocks to reach
 		// confidence on the message calls
-		for tid, data := range newCalls {
-			e.queueForConfidence(tid, data, nil, ts)
+		for tid, calls := range newCalls {
+			for _, data := range calls {
+				e.queueForConfidence(tid, data, nil, ts)
+			}
 		}
 
 		for at := e.lastTs.Height(); at <= ts.Height(); at++ {
@@ -474,7 +476,7 @@ func newMessageEvents(ctx context.Context, hcAPI headChangeAPI, cs EventAPI) mes
 }
 
 // Check if there are any new actor calls
-func (me *messageEvents) checkNewCalls(ts *types.TipSet) (map[triggerID]eventData, error) {
+func (me *messageEvents) checkNewCalls(ts *types.TipSet) (map[triggerID][]eventData, error) {
 	pts, err := me.cs.ChainGetTipSet(me.ctx, ts.Parents()) // we actually care about messages in the parent tipset here
 	if err != nil {
 		log.Errorf("getting parent tipset in checkNewCalls: %s", err)
@@ -485,7 +487,7 @@ func (me *messageEvents) checkNewCalls(ts *types.TipSet) (map[triggerID]eventDat
 	defer me.lk.RUnlock()
 
 	// For each message in the tipset
-	res := make(map[triggerID]eventData)
+	res := make(map[triggerID][]eventData)
 	me.messagesForTs(pts, func(msg *types.Message) {
 		// TODO: provide receipts
 
@@ -500,7 +502,7 @@ func (me *messageEvents) checkNewCalls(ts *types.TipSet) (map[triggerID]eventDat
 			// If there was a match, include the message in the results for the
 			// trigger
 			if matched {
-				res[tid] = msg
+				res[tid] = append(res[tid], msg)
 			}
 		}
 	})
--- a/chain/events/events_test.go
+++ b/chain/events/events_test.go
@@ -1323,3 +1323,62 @@ func TestStateChangedTimeout(t *testing.T) {
 	fcs.advance(0, 5, nil)
 	require.False(t, called)
 }
+
+func TestCalledMultiplePerEpoch(t *testing.T) {
+	fcs := &fakeCS{
+		t: t,
+		h: 1,
+
+		msgs:    map[cid.Cid]fakeMsg{},
+		blkMsgs: map[cid.Cid]cid.Cid{},
+		tsc:     newTSCache(2*build.ForkLengthThreshold, nil),
+	}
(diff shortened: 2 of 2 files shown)


whyrusleeping (approved):
Nice find. Excited to see how much flakiness all over this improves.

nonsense:
Great stuff! This was happening on occasion on Testground as well, awesome that it is fixed now!

PR #6356, opened 2021-05-31
Support standalone miner-market process

### LINKED ISSUE #5149: concept of a miner DMZ
link: https://en.wikipedia.org/wiki/DMZ_(computing)

## Problem

The lotus miner is a specialised process with a several purposes that seem to be at odds with each other when it comes to network/external exposure/isolation requirements:

1. To produce new blocks in the blockchain (by running leader election, winning PoSt, and broadcasting the block).
2. Managing storage by onboarding new sectors, and proving them when challenged (Window PoSt).
3. Accepting and sealing deals.

For 1 and 2, the miner doesn't need to be exposed openly on the p2p network. These processes can be carried out in relative isolation, proxying via a companion daemon, which is what happens now. With this architecture, the miner process enjoys certain protection from the p2p network.

But for 3, **the protocol currently expects the miner to be directly reachable/diallable**. This is at odds with 1 and 2 above.

Preliminary network diagnostics show that many miners choose to not participate in the DHT nor advertise multiaddrs in their miner actor, therefore they opt to be not diallable and, thus, accept no deals.

## Discussion

I think we need an official concept of a [miner DMZ](https://en.wikipedia.org/wiki/DMZ_(computing)), which is a specialised node that takes the role of the companion daemon, but also enhances it with a deals bridge/depot functionality.

This way, the DMZ would take the responsibility of being exposed in the network, accepting deals on behalf of the miner with a predefined policy, participating in the storage market, etc. while the **core miner** would stay isolated from the network.

The miner DMZ would also own keeping the storage market actor updated with the deal status.

This would also protect the miner from bugs/inefficiencies in the deals, markets, data transfer components, like #4877. For scaling deals, a miner could run many replicated miner DMZ nodes fronting a single miner.

nonsense:
Associated docs: https://github.com/filecoin-project/filecoin-docs/pull/890

---

This PR is a cherry-picked version of https://github.com/filecoin-project/lotus/pull/5983. Closes #5149.

---

1. It introduces the concept of `Subsystems` within the `lotus-miner` process - currently:
- StorageMarket
- Mining
- Sealing
- SectorStorage

2. It adds support for running the miner market module as a separate process.
3. A `lotus-miner market service` connects to the `main` lotus process for all sealing/storage interactions, or:
- `market` service runs `StorageMarget` subsystem
- `main miner` service runs `Mining`, `Sealing` and `SectorStorage` subsystems.

---

Depending on configuration options, this PR provides flexibility for the the miner to be split into multiple processes, or to be run as one process as is currently the case.

---

Currently the deployment flow for running a `market` module as a separate process is the following:
1. Create backup of the original miner repo / metadata (i.e. something like: `export LOTUS_BACKUP_BASE_PATH=~/.lotusbackup && export LOTUS_MINER_PATH=~/.lotusminer && export LOTUS_PATH=~/.devlotus && lotus-miner backup ~/.lotusbackup/backupfile`)
2. Initialise the market service based on the backup from 1, and generate a new libp2p identity (i.e. something like:
```
APISEALER=`./lotus-miner auth api-info --perm=admin`

export LOTUS_PATH=~/.devLotus && export LOTUS_MINER_PATH=~/.restoredrepo && ./lotus-miner init service --enable-market --api-sealer=$APISEALER --api-sector-index=$APISEALER --config=~/.lotusmarket/config.toml --storage-config=~/.lotusmarket/storage.json ~/.lotusbackup/backupfile
```
This would also send a message on chain and update the address for the given miner, so that the markets process is the point of entry from the on-chain DHT.
3. Run market service as well as main miner service (i.e. something like:
```
./lotus-miner run --nosync
LOTUS_MINER_PATH=~/.restoredrepo ./lotus-miner run --nosync
```

---

TODO:
- [x] update documentation - https://github.com/filecoin-project/filecoin-docs/pull/890
- [x] resolve conflicts with dynamic pricing PR
- [x] resolve conflicts and update tests based on `itests/kit`
- [x] fix unit tests

TODO FROM PARENT:
- [x] Fix api deps
  - (don't depend on ffi)
- [x] Fix node tests
- [x] Delegate market node auth to the main node (storage subsystem assumes that)
  - Put [API Secret](https://github.com/filecoin-project/lotus/blob/3fc23a785c2fa0d75fa8d737c635c01f9ae256fe/node/impl/common/common.go#L37) behind an interface, in DI replace the implementation by the main node API
- [x] Figure out what to do about local market-node storage
  - [x] Either drop completely (likely preferred initially)
    - This likely just means making the local storage implementation ignore paths, or (a bit cleaner) making it possible to run Remote store without a Local store
  - [ ] Or register storage properly (and handle disconnects, a bit like workers do)
- [x] Use http-mode jsonrpc for talking to the main Sterage/Sealing node (for reliability)
- [x] Don't construct miner Libp2p node when not needed
  - Mostly just shifting things around in the node builder
- [x] Bump API versions which need bumping
  - Miner version at least
- [ ] Check API versions on reconnect
- [x] Setup docs - https://github.com/filecoin-project/filecoin-docs/pull/890
- [x] Write integration test
  - Copy one of the `api/test/deal.go` tests, and run in this setup

FOLLOW UP / NICE TO HAVE:
- [ ] Add Testground testplan, so that we support both mode of operation - separate processes for miner and markets, as well as one process for miner and markets
- [ ] Return nicer errors from unsupported api calls when subsystems are disabled
  - Not sure if we can do much better than `if x == nil { error(module x is disabled) }`
- [ ] Improve dev experience with respect to setup


magik6k:
This will likely conflict with https://github.com/filecoin-project/lotus/pull/6280 a bunch (it also cherry-picks from my PR)

raulk:
@nonsense will follow up with specific questions to untangle the TODO items here.

api/api_storage.go
```diff
@@ -135,8 +142,8 @@ type StorageMiner interface {
 	StorageBestAlloc(ctx context.Context, allocate storiface.SectorFileType, ssize abi.SectorSize, pathType storiface.PathType) ([]stores.StorageInfo, error)           //perm:admin
 	StorageLock(ctx context.Context, sector abi.SectorID, read storiface.SectorFileType, write storiface.SectorFileType) error                                          //perm:admin
 	StorageTryLock(ctx context.Context, sector abi.SectorID, read storiface.SectorFileType, write storiface.SectorFileType) (bool, error)                               //perm:admin
+	StorageList(ctx context.Context) (map[stores.ID][]stores.Decl, error)                                                                                               //perm:admin
```
dirkmc, line 145: Is it ok to remove `//perm:admin` here?

cmd/lotus-storage-miner/init_restore.go
```diff
@@ -30,250 +31,260 @@ import (
...
+	}
 
-		log.Info("Checking full node version")
+	if !v.APIVersion.EqMajorMinor(lapi.FullAPIVersion1) {
+		return xerrors.Errorf("Remote API version didn't match (expected %s, remote %s)", lapi.FullAPIVersion1, v.APIVersion)
+	}
 
-		v, err := api.Version(ctx)
+	if !cctx.Bool("nosync") {
+		if err := lcli.SyncWait(ctx, &v0api.WrapperV1Full{FullNode: api}, false); err != nil {
```
dirkmc, line 60: Suggest adding a log line here as sync wait can probably take a while

extern/storage-sealing/states_proving.go
```diff
@@ -126,3 +126,22 @@ func (m *Sealing) handleRemoving(ctx statemachine.Context, sector SectorInfo) er
...
+	// TODO: track sector health / expiration
+	log.Infof("Proving sector %d", sector.SectorNumber)
+
+	cfg, err := m.getConfig()
+	if err != nil {
+		return xerrors.Errorf("getting sealing config: %w", err)
+	}
+
+	if err := m.sealer.ReleaseUnsealed(ctx.Context(), m.minerSector(sector.SectorType, sector.SectorNumber), sector.keepUnsealedRanges(true, cfg.AlwaysKeepUnsealedCopy)); err != nil {
+		log.Error(err)
```
dirkmc, line 140: Why not return the error here?

markets/retrievaladapter/provider.go
```diff
@@ -48,14 +51,12 @@ func (rpn *retrievalProviderNode) GetMinerWorkerAddress(ctx context.Context, min
 }
 
 func (rpn *retrievalProviderNode) UnsealSector(ctx context.Context, sectorID abi.SectorNumber, offset abi.UnpaddedPieceSize, length abi.UnpaddedPieceSize) (io.ReadCloser, error) {
-	log.Debugf("get sector %d, offset %d, length %d", sectorID, offset, length)
```
dirkmc, line 51: Could we leave this debug line here, it's useful for tracking down where things are going wrong during retrievals 🙏

dirkmc (commented):
It seems like most of the PR is refactoring, just a few small comments so far.

It also seems like the init service is still WIP, so I didn't review that closely, let me know if you'd like me to do so.

cmd/lotus-storage-miner/init_restore.go
nonsense: When we don't have `nosync`, we output the result of `lotus sync wait` and it is clear that the miner is waiting for sync.

extern/storage-sealing/states_proving.go
nonsense: I don't know, but this is just being moved from one file to the other.

api/api_storage.go
nonsense: We are just moving this, I think we need the annotation so that only users with correct JWT tokens can call this.

raulk:
This PR fixes https://github.com/filecoin-project/lotus/issues/5149.

node/enable_libp2p.go
```diff
@@ -0,0 +1,40 @@
+package node
+
+import (
+	"github.com/filecoin-project/lotus/node/config"
+	"github.com/filecoin-project/lotus/node/repo"
+)
+
+func ApplyIfEnableLibP2P(r repo.Repo, opts ...Option) Option {
```
magik6k, line 8: This looks really hacky, I'd love to avoid having this if at all possible

node/impl/common/net.go
```diff
@@ -0,0 +1,34 @@
...
+	metrics "github.com/libp2p/go-libp2p-core/metrics"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	protocol "github.com/libp2p/go-libp2p-core/protocol"
+
+	"github.com/filecoin-project/lotus/api"
+	apitypes "github.com/filecoin-project/lotus/api/types"
+)
+
+type NetAPI interface {
```
magik6k, line 15: Should be much cleaner with https://github.com/filecoin-project/lotus/pull/6627

node/impl/storminer.go
```diff
@@ -49,54 +48,63 @@ import (
...
-	Host          host.Host
-	AddrSel       *storage.AddressSelector
-	DealPublisher *storageadapter.DealPublisher
-
-	Epp gen.WinningPoStProver
+	Full        api.FullNode
+	LocalStore  *stores.Local
+	RemoteStore *stores.Remote
+
+	// Markets
```
magik6k, line 55: We should split submodule APIs like https://github.com/filecoin-project/lotus/pull/6627, but I'm ok with that being done after this PR is merged

node/modules/storageminer.go
```diff
@@ -210,14 +209,15 @@ type StorageMinerParams struct {
 }
 
 func StorageMiner(fc config.MinerFeeConfig) func(params StorageMinerParams) (*storage.Miner, error) {
+	fmt.Printf("setting up storage miner with %#v \n", fc)
+
```
magik6k, line 213:
```suggestion
```

node/modules/storageminer.go
```diff
@@ -724,6 +721,19 @@ func StorageAuth(ctx helpers.MetricsCtx, ca v0api.Common) (sectorstorage.Storage
 	return sectorstorage.StorageAuth(headers), nil
 }
 
+func StorageAuthWithURL(url string) func(ctx helpers.MetricsCtx, ca v0api.Common) (sectorstorage.StorageAuth, error) {
+	log.Infow("Setting auth token based on URL", "url", url)
```
magik6k, line 725: This might log API secrets, which we don't want as miners need to be able to share logs with not-so-trusted people (e.g. us)

node/rpc.go
```diff
@@ -117,18 +118,21 @@ func MinerHandler(a api.StorageMiner, permissioned bool) (http.Handler, error) {
 		mapi = api.PermissionedStorMinerAPI(mapi)
 	}
 
-	rpcServer := jsonrpc.NewServer()
+	//_, _ = a.ActorAddress(context.Background())
```
magik6k, line 121:
```suggestion
```

storage/sectorblocks/blocks.go
```diff
@@ -48,17 +47,22 @@ func DsKeyToDealID(key datastore.Key) (uint64, error) {
 	return dealID, nil
 }
 
+type SectorBuilder interface { // todo: apify, make work remote
```
magik6k, line 50: todo looks done

```suggestion
type SectorBuilder interface {
```

magik6k (commented):
First pass, things look pretty good, mostly needs some cleanup around node construction.

Do we have a related filecoin-docs PR already?

markets/retrievaladapter/provider.go
```diff
@@ -102,12 +105,13 @@ func (rpn *retrievalProviderNode) GetChainHead(ctx context.Context) (shared.TipS
 }
 
 func (rpn *retrievalProviderNode) IsUnsealed(ctx context.Context, sectorID abi.SectorNumber, offset abi.UnpaddedPieceSize, length abi.UnpaddedPieceSize) (bool, error) {
-	si, err := rpn.miner.GetSectorInfo(sectorID)
+	//TODO(anteva): maybe true? show on chain info??
```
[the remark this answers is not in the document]
nonsense: @magik6k sorry, this was buggy until today. We added `sectorsStatus` today. Basically `IsUnsealed` should check the on-chain data as well, in order to get the `ProofType` and the sector size. I think it is correct now.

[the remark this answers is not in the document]
nonsense: Until today both `IsUnsealed` and `UnsealSector` were broken, as they were directly using the `rpn.miner.GetSectorInfo` which is the in-process impl.

I have refactored these to use `sectorsStatus` now, which augments correctly `si` to include onchain data as well, as it is needed due to `ProofType`.

node/enable_libp2p.go
nonsense: I have removed `ApplyIfEnableLibP2P` so no unnecessary abstraction now. Looking into passing config into `StorageMiner` to see if we can simplify further.

node/builder_miner.go
```diff
@@ -0,0 +1,222 @@
...
+			})),
+			Override(new(storagemarket.StorageProviderNode), storageadapter.NewProviderNodeAdapter(&cfg.Fees, &cfg.Dealmaking)),
+		),
+
+		Override(new(sectorstorage.SealerConfig), cfg.Storage),
+		Override(new(*storage.AddressSelector), modules.AddressSelector(&cfg.Addresses)),
+	)
+}
+
+func StorageMiner(out *api.StorageMiner) Option {
```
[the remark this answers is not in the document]
nonsense: That's right, thanks for the hint, this is much cleaner now.

node/enable_libp2p.go
nonsense: Having passed the Subsystems config, and adding an `enableLibp2pNode bool` in `Settings`, this is no longer needed.

extern/sector-storage/stores/remote.go
```diff
@@ -297,6 +298,32 @@ func (r *Remote) fetch(ctx context.Context, url, outname string) error {
...
 }
 
+func (r *Remote) checkAllocated(ctx context.Context, url string, spt abi.RegisteredSealProof, offset, size abi.PaddedPieceSize) (bool, error) {
+	url = fmt.Sprintf("%s/%d/allocated/%d/%d", url, spt, offset.Unpadded(), size.Unpadded())
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return false, xerrors.Errorf("request: %w", err)
+	}
+	req.Header = r.auth.Clone()
+	fmt.Printf("req using header: %#v \n", r.auth)
```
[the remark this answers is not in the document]
magik6k: Github still shows this here

node/builder.go
```diff
@@ -742,7 +375,7 @@ func New(ctx context.Context, opts ...Option) (StopFunc, error) {
 		fx.Options(ctors...),
 		fx.Options(settings.invokes...),
 
-		fx.NopLogger,
+		//fx.NopLogger,
```
magik6k, line 378:
```suggestion
		fx.NopLogger,
```

node/builder_miner.go
```diff
@@ -0,0 +1,223 @@
...
+			Override(new(modules.MinerStorageService), modules.ConnectStorageService(cfg.Subsystems.SectorIndexApiInfo)),
+			Override(new(sectorstorage.Unsealer), From(new(modules.MinerStorageService))),
+			Override(new(sectorblocks.SectorBuilder), From(new(modules.MinerStorageService))),
+		),
+		If(!cfg.Subsystems.EnableSealing,
+			Override(new(modules.MinerSealingService), modules.ConnectSealingService(cfg.Subsystems.SealerApiInfo)),
+			Override(new(stores.SectorIndex), From(new(modules.MinerSealingService))),
+		),
+
+		If(cfg.Subsystems.EnableStorageMarket,
```
magik6k, line 130: Why don't we include the `LibP2P` option here (and in the client options)? This should allow us to get rid of `settings.enableLibp2pNode`

node/modules/storageminer.go
```diff
@@ -669,29 +672,21 @@ func RetrievalProvider(h host.Host,
...
 var ManagerWorkPrefix = datastore.NewKey("/stmgr/calls")
 
-func LocalStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, ls stores.LocalStorage, si stores.SectorIndex, urls sectorstorage.URLs) (*stores.Local, error) {
+func LocalStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, ls stores.LocalStorage, si stores.SectorIndex, urls stores.URLs) (*stores.Local, error) {
 	ctx := helpers.LifecycleCtx(mctx, lc)
 	return stores.NewLocal(ctx, ls, si, urls)
 }
 
 func RemoteStorage(lstor *stores.Local, si stores.SectorIndex, sa sectorstorage.StorageAuth, sc sectorstorage.SealerConfig) *stores.Remote {
+	fmt.Printf("setting RemoteStorage: %#v \n", sa) // TODO(anteva): remove me prior to merge to master
```
magik6k, line 689: todo

magik6k (commented):
Looks almost mergable, just a couple of small comments

node/builder_miner.go
nonsense: I will give this a try, but I am not sure it is possible, as there are some configs that need to be applied earlier.

raulk: Agree. Now that we've separated `Libp2p` from `Online`, `enableLibp2pNode` maps directly to enabling the `Libp2p` aggregate. This should be feasible.

raulk: Oh wait, I think this may have to do with the fact that we need to read the Config to know whether to enable or not the libp2p node, right?

raulk: Only when Markets is enabled in the miner config, should we enable libp2p. So it's a bit of a chicken and egg, and that's why it's convoluted.

raulk: Actually, no. It should be possible!

cmd/lotus-storage-miner/init_service.go
```diff
@@ -0,0 +1,136 @@
...
+		},
+		&cli.BoolFlag{
+			Name:  "nosync",
+			Usage: "don't check full-node sync status",
+		},
+
+		&cli.BoolFlag{
+			Name:  "enable-market",
+			Usage: "enable market module",
+		},
```
raulk, line 70: Can we prepare this for future services by making it a string enum?

```
lotus init service --name=markets
```

node/config/def.go
```diff
@@ -50,13 +50,24 @@ type Backup struct {
...
 	Storage    sectorstorage.SealerConfig
 	Fees       MinerFeeConfig
 	Addresses  MinerAddressConfig
 }
 
+type MinerSubsystemConfig struct {
+	EnableMining        bool
+	EnableSealing       bool
+	EnableSectorStorage bool
+	EnableStorageMarket bool
```
raulk, line 65:
```suggestion
	EnableMarkets bool
```

node/impl/common/mock/mock.go
```diff
@@ -0,0 +1,101 @@
+package mock
```
raulk, line 1: Can we rename the file to `net.go`?

node/impl/storminer.go
```diff
@@ -49,54 +48,63 @@ import (
...
+	PieceStore        dtypes.ProviderPieceStore         `optional:"true"`
+	StorageProvider   storagemarket.StorageProvider     `optional:"true"`
+	RetrievalProvider retrievalmarket.RetrievalProvider `optional:"true"`
+	DataTransfer      dtypes.ProviderDataTransfer       `optional:"true"`
+	DealPublisher     *storageadapter.DealPublisher     `optional:"true"`
+	SectorBlocks      *sectorblocks.SectorBlocks        `optional:"true"`
+	Host              host.Host                         `optional:"true"`
```
raulk, line 62: We should move these out to a `MarketsAPI` struct and embed it here.

node/impl/storminer.go
```diff
@@ -49,54 +48,63 @@ import (
 type StorageMinerAPI struct {
```
raulk, line 48: We should consider renaming this to Provider, as the term "Miner" becomes overloaded. It refers both to _any_ process on the miner-side deployment, as well as the specific process that does storage mining.

raulk: However, this is clearly a follow-up PR, possibly for 1.13 where we bring more consistency to this rearchitecture and we rename even commands and binaries (`lotus-provider`).

raulk (commented):
Initial comments. We probably want to bring consistency between the terms "subsystem" and "service" -- we use them a bit interchangeably?

extern/sector-storage/manager.go
```diff
@@ -725,3 +723,4 @@ func (m *Manager) Close(ctx context.Context) error {
 
 var _ Unsealer = &Manager{}
 var _ SectorManager = &Manager{}
+var _ Unsealer = &Manager{}
```
raulk, line 726: Duplicated type check.

extern/storage-sealing/input.go
```diff
@@ -236,34 +237,34 @@ func (m *Sealing) handleAddPieceFailed(ctx statemachine.Context, sector SectorIn
 	return nil
 }
 
-func (m *Sealing) AddPieceToAnySector(ctx context.Context, size abi.UnpaddedPieceSize, data storage.Data, deal DealInfo) (abi.SectorNumber, abi.PaddedPieceSize, error) {
+func (m *Sealing) SectorAddPieceToAny(ctx context.Context, size abi.UnpaddedPieceSize, data storage.Data, deal api.PieceDealInfo) (api.SectorOffset, error) {
```
raulk, line 240: This seems to be like an API breaking change, externally. Are we OK making it? Do we need to document it in release notes? @magik6k

node/modules/storageminer.go
```diff
@@ -658,9 +660,10 @@ func RetrievalPricingFunc(cfg config.DealmakingConfig) func(_ dtypes.ConsiderOnl
...
-func RetrievalProvider(h host.Host,
-	miner *storage.Miner,
-	full v1api.FullNode,
+func RetrievalProvider(
+	maddr dtypes.MinerAddress,
+	adapter retrievalmarket.RetrievalProviderNode,
+	netwk rmnet.RetrievalMarketNetwork,
 	ds dtypes.MetadataDS,
 	pieceStore dtypes.ProviderPieceStore,
 	mds dtypes.StagingMultiDstore,
```
raulk, line 669: pieceProvider argument is unused.

node/modules/storageminer.go
```diff
@@ -724,6 +718,18 @@ func StorageAuth(ctx helpers.MetricsCtx, ca v0api.Common) (sectorstorage.Storage
...
+	return func(ctx helpers.MetricsCtx, ca v0api.Common) (sectorstorage.StorageAuth, error) {
+		s := strings.Split(url, ":")
+		if len(s) != 2 {
+			return nil, errors.New("unexpected format of URL")
+		}
+		headers := http.Header{}
+		headers.Add("Authorization", "Bearer "+s[0])
+		return sectorstorage.StorageAuth(headers), nil
+	}
+}
```
raulk, line 731: What kind of URL is this? It seems wrong to call it a URL, based on what we're doing here. Can you give an example value of what will be passed here?

cmd/lotus-storage-miner/init_restore.go
```diff
@@ -30,250 +31,259 @@ import (
...
+	}
 
-		log.Info("Checking proof parameters")
+	return after(api, maddr, peerid, mi)
+}
 
-		if err := paramfetch.GetParams(ctx, build.ParametersJSON(), build.SrsJSON(), uint64(mi.SectorSize)); err != nil {
-			return xerrors.Errorf("fetching proof parameters: %w", err)
-		}
+var restoreCmd = &cli.Command{
```
raulk, line 250: Can we keep this at the top, as it serves as the subcommand definition/entrypoint, and also to minimise the diff noise?

cmd/lotus-storage-miner/init_restore.go
```diff
@@ -30,250 +31,259 @@ import (
...
-		log.Info("Trying to connect to full node RPC")
+	api, closer, err := lcli.GetFullNodeAPIV1(cctx) // TODO: consider storing full node address in config
+	if err != nil {
+		return err
+	}
+	defer closer()
 
-		if err := checkV1ApiSupport(ctx, cctx); err != nil {
-			return err
-		}
```
raulk, line 63: We don't need this any longer?

cmd/lotus-storage-miner/init_restore.go
```diff
@@ -30,250 +31,259 @@ import (
...
-			Usage: "storage paths config (storage.json)",
-		},
-	},
-	ArgsUsage: "[backupFile]",
-	Action: func(cctx *cli.Context) error {
-		log.Info("Initializing lotus miner using a backup")
-		if cctx.Args().Len() != 1 {
-			return xerrors.Errorf("expected 1 argument")
-		}
+func restore(ctx context.Context, cctx *cli.Context, manageConfig func(*config.StorageMiner) error, after func(api lapi.FullNode, addr address.Address, peerid peer.ID, mi miner.MinerInfo) error) error {
```
raulk, line 34: FYI, I wasn't able to review this file because the diff is too noisy and I'm not familiar with this command. I trust @magik6k here.

cmd/lotus-storage-miner/run.go
```diff
@@ -142,14 +163,18 @@ var runCmd = &cli.Command{
...
-			return xerrors.Errorf("connecting to full node (libp2p): %w", err)
+			// Bootstrap with full node
+			remoteAddrs, err := nodeApi.NetAddrsListen(ctx)
+			if err != nil {
+				return xerrors.Errorf("getting full node libp2p address: %w", err)
+			}
+
+			if err := minerapi.NetConnect(ctx, remoteAddrs); err != nil {
+				return xerrors.Errorf("connecting to full node (libp2p): %w", err)
+			}
```
raulk, line 177: Why do we need the markets node connected to the full node via libp2p? I think that shouldn't be necessary.

cmd/lotus-storage-miner/init_service.go
```diff
@@ -0,0 +1,136 @@
...
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:     "config",
+			Usage:    "config file (config.toml)",
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:     "storage-config",
+			Usage:    "storage paths config (storage.json)",
+			Required: true,
```
raulk, line 60: Why is this required? Can't this be omitted for markets nodes?

raulk (approved):
LGTM, besides the comments! This was a heroic effort. Awesome work here, @nonsense and @magik6k ❤️  This is the start of a more modular Lotus!

Some top-level comments:

- Let's standardise on the name "subsystem" or "service". We use both interchangeably and it's confusing (service in the cli tooling, subsystem in the config). Either is fine with me, unless we do want to use both because they refer to different things now, or we expect them to (e.g. service as a collection of subsystems?). If so, we should document the difference.
- OK to merge the changes in the testkit as is. I can follow-up with a new PR to bring more order/structure to that and de-entropise.
- We may need commands to query a lotus node to know which subsystems is running, and to know which other APIs it's interacting with, but that can be a follow-up.

extern/storage-sealing/input.go
magik6k: This is just an internal API

cmd/lotus-storage-miner/init_service.go
magik6k: Market nodes need storage access for retrieval

cmd/lotus-storage-miner/run.go
magik6k: This is done to help bootstrap the node with the network (dht, autonat, etc.)

node/modules/storageminer.go
magik6k: The string is `[jwtToken]:[url]`, the parameter should probably be called something like `apiInfo`

cmd/lotus-storage-miner/init_service.go
nonsense: I have also been confused about that. An empty `storage.json` used for initialising the `market` node seems to be working - this is something that I should understand better and also update the docs, see: https://github.com/filecoin-project/filecoin-docs/pull/890/files#r665537068

node/impl/storminer.go
nonsense: Noted this in a list with follow-ups, and will mark as resolved, so that we concentrate on outstanding issues for this PR!

cmd/lotus-storage-miner/init_service.go
nonsense: I removed the `--enable-market` flag, and added a `--name` flag for services to be enabled.

cmd/lotus-storage-miner/init_service.go
```diff
@@ -63,12 +39,10 @@ var serviceCmd = &cli.Command{
 			Name:  "nosync",
 			Usage: "don't check full-node sync status",
 		},
-
-		&cli.BoolFlag{
-			Name:  "enable-market",
-			Usage: "enable market module",
+		&cli.StringSliceFlag{
+			Name:  "name",
```
magik6k, line 43: Maybe `type`? `lotus-miner init service --name markets ...` looks a bit weild

magik6k (approved):
Well, I think this is in good enough shape to land in master, just one cosmetic suggestion.

cmd/lotus-storage-miner/init_service.go
nonsense: Renamed to `type`.

nonsense:
@magik6k @raulk in the last commit (https://github.com/filecoin-project/lotus/pull/6356/commits/2af02af1c1909168cc5e0ef5b3562c8374cdfc09), I renamed `--name` to `--type` and also addressed the UX problems with `storage-config.json`:

`restore` is used both for backup restoration and for initialisation of new services, so it needs to support 3 modes:
- restoring from backup where the user wants to restore a given storage-config.json
- restoring from backup where the user wants to restore other parts from their backup, and keep their existing storage-config.json
- initialising a new markets type service, where we need a blank storage-config.json (with empty set of StoragePaths), since at the moment `Remote` requires a `Local` store, which requires a proper config.

So parsing of the storage-config.json is extracted away, and we call `restore` with the necessary *stores.StorageConfig

cmd/lotus-storage-miner/init_service.go
nonsense: @raulk this is addressed in the latest commit - see https://github.com/filecoin-project/lotus/pull/6356#issuecomment-878998621

magik6k (approved):
:boat:

PR #6367, opened 2021-06-01
Robustify commit batcher

magik6k:
The main point of this PR is to support submitting commits individually when the batch is too small, the second point is making error handling / propagation in the commit batcher much better.

(also fixes (implements) collateral support in the batcher)

((also does some this with the precommit batcher))


arajasek (approved):
SGWM

PR #6384, opened 2021-06-03
build appimage in CI

coryschwartz:
Build appimage and include it in the release.

The CI executor now is 20.04 so I updated the AppImage yaml to work  there.

I updated the release scripts to include the appimage. I have not run through the release to verify that the bundle and release scripts work so be extra cautious reviewing that.

EDIT:

I ran a dry run for the release step; looks like it will succeed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -428,6 +428,41 @@ jobs:
             - "~/.rustup"
             - "~/.cargo"
 
+  build-appimage:
+    machine:
+      image: ubuntu-2004:202104-01
+    steps:
+      - checkout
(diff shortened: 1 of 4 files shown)


coryschwartz:
I just did a dry run. I nerfed the publish script to only echo the publish commands and watched the circleci run. It looks like it works provided the CIRCLE_TAG is set.

dkkapur:
Discussion in lotus standup today - Stewards do not need to block on merging this in. @travisperson can do the review and we should merge this accordingly.

magik6k (approved):

PR #6388, opened 2021-06-03
build snapcraft

coryschwartz:
publish snap to the snapcraft store


coryschwartz:
I made some changes to the snapchat build aside from adding the CI logic.

Origionally, I was using the pre-built binaries and just copying them into the snap, and was using ubuntu "core18" as the snap base. With this base, the linking was compatible with the golang docker image (which is debian). However, but this turned out to be a headache to get all the linking right with the golang-16.4 docker image. The binaries built in the build-all step have enough incompatibilty with package versions that it was easier to simply re-build it for the snap. 

So now, the binary is built as part of the snap construction and is based on ubuntu 20.04 -- although it will be installable on any linux OS running snap.

coryschwartz:
I made some improvements to make sure this works with strict confinement, and added more detail for https://snapcraft.io/lotus-filecoin

.circleci/config.yml
```diff
@@ -861,3 +889,29 @@ workflows:
...
+          filters:
+            branches:
+              ignore:
+                - /.*/
+            tags:
+              only:
+                - /^v\d+\.\d+\.\d+(-rc\d+)?$/
+      - publish-snapcraft:
+          name: publish-snapcraft-tmp
+          channel: edge
```
magik6k, line 904: Do we want to keep this?

magik6k (approved):
Looks good, just one question.

.circleci/config.yml
coryschwartz: oh, no thanks for catching that.

magik6k (approved):

PR #6375, opened 2021-06-01
Update go-graphsync v0.8.0

hannahhoward:
# Goals

Good news: The latest version of go-graphync and go-ipld-prime contains an important optimization to remove a bunch of unneccesary hashing when reading from local blockstores, which should cut down significantly on CPU usage
Bad news: The latest versions of these libraries contain a some important, but very breaking changes to syntax.
Good news: I'm updating them all the way to Lotus! And also, in most cases, from the point of view of the consuming libraries, they're simplifications.

# Implementation

- Update to go-graphsync v0.8.0 with go-ipld-prime linksystem branch & trusted store.
- Update other dependent libraries with SHA updates
- had to also update libp2p-quic-transport to v0.10.0 cause of libp2p updates
- Fix syntax changes which are mostly to:
   - dagcbor Encode / Decode function name
   - use of linksystem over loader + storer.
   
# Background

- More about the LinkSystem: https://github.com/ipld/go-ipld-prime/pull/143
- Trusted store branch: https://github.com/ipld/go-ipld-prime/pull/149
- PR to merge LinkSystem into go-graphsync: https://github.com/ipfs/go-graphsync/pull/161

# For Discussion

Merge Blocked ON: 
https://github.com/filecoin-project/go-data-transfer/pull/212
https://github.com/filecoin-project/go-multistore/pull/3
https://github.com/filecoin-project/go-fil-markets/pull/550

PR #6393, opened 2021-06-04
Feat/datamodel selector retrieval

ribasushi:
~( This PR depends on and includes https://github.com/filecoin-project/lotus/pull/6375 )~ ✅ 

Introduce a new `RetrievalOrder`-struct field and a CLI option that takes a string representation as understood by https://pkg.go.dev/github.com/ipld/go-ipld-selector-text-lite#SelectorSpecFromPath . Allows for partial retrieval of any sub-DAG of a deal *provided* the user knows the exact low-level shape of the deal contents.

As an example with this patch one can retrieve the first entry of a UnixFS directory by executing:
`lotus client retrieve --miner f0XXXXX --datamodel-path-selector 'Links/0/Hash' bafyROOTCID ~/output`

See top of [itests/deals_partial_retrieval_test.go](https://github.com/filecoin-project/lotus/pull/6393/files#diff-7cecd455116bb046634b469a4e86e3c143b53241750c85af87dfd3e799951ac4) for a more elaborate example.


ribasushi:
Review note: this is the first user-facing selector interface that I know of. Also selector execution is not something we have a lot of prior art for. Request for @warpfork @mvdan @willscott and @hannahhoward to very carefully examine these parts:
- https://github.com/filecoin-project/lotus/blob/d539169121/markets/utils/selectors.go#L23-L96
- https://github.com/filecoin-project/lotus/blob/d53916912168f66b64db8beb28b5344eb77aac4c/node/impl/client/client.go#L832-L855
- https://github.com/filecoin-project/lotus/blob/d539169121/node/impl/client/client.go#L737-L753

mvdan:
Nit: in the first link, you use `context.TODO()`, but you should probably use the `context.Context` that's part of the `LinkContext` parameter you get passed.

I've raised github.com/ipld/go-ipld-prime/issues/185 about making the "basicnode chooser" fewer lines.

I'm not familiar with the selector/traversal API, so I'll leave that to Eric :)

cli/client.go
```diff
@@ -1044,6 +1045,10 @@ var clientRetrieveCmd = &cli.Command{
 			Name:  "miner",
 			Usage: "miner address for retrieval, if not present it'll use local discovery",
 		},
+		&cli.StringFlag{
+			Name:  "datamodel-path-selector",
+			Usage: "a rudimentary (DM-level-only) text-path selector, allowing for sub-selection within a deal",
```
willscott, line 1050: do we imagine extending this to ADL-extended selectors soon?
would that be a different flag?
it seems fine for this to just be a `path-selector` rather than including all the `datamodel-only` warnings on the interface.

node/impl/client/client.go
```diff
@@ -730,9 +735,27 @@ func (a *API) clientRetrieve(ctx context.Context, order api.RetrievalOrder, ref
 			return err
 		}*/
 
+		// FIXME - this is a direct copy from https://github.com/filecoin-project/go-fil-markets/blob/v1.4.0/shared/selectors.go#L11-L16
+		// Unable to use it because we need the SelectorSpec, and markets exposes just a reified node
+		ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
+		selspec := ssb.ExploreRecursive(
+			selector.RecursionLimitNone(),
+			ssb.ExploreAll(ssb.ExploreRecursiveEdge()),
```
willscott, line 743: do we want a matcher here as well?

willscott (approved):

cli/client.go
ribasushi: Right... this goes back to the mega-thread, where @aschmahmann and me were pushing against transparent-unixfs in the low-level selection.
I personally would not use a heuristic-engaging selector on principle, and left the door open for "do not try to be clever here" with the `datamodel` prefix. I am open to switching this out, but we would need signaling "I want to use this as a low level selector" in some fashion.

node/impl/client/client.go
ribasushi: I opted for https://github.com/filecoin-project/lotus/blob/efeaea76b97fd52203122ee33a7633da74ba9631/node/impl/client/client.go#L738-L739
Diverging didn't make sense... @hannahhoward should comment here: either we export the spec, or we adjust both

kernelogic:
lotus -v
lotus version 1.11.0-dev+mainnet+git.7f1712267

Hi Peter, I tried locally but it doesn't seem to do anything. Maybe I did it wrong.

First, I use ipfs to add a folder:
ipfs add --nocopy=true --pin=false -r IRIS_313-11199/
added QmP9vWnVVajeG2NBGYRXDwuPnG6jfXe8d1y52767uGzTw9 IRIS_313-11199/IRIS_313-11199.realigned.bam
added QmW6PSiW9nGWqU8j1jRifD1uwvYrSCUsyqdrCWD31yCDK3 IRIS_313-11199/IRIS_313-11199.realigned.bam.bai
added bafkreigy6gqpcqdb6eqvjok3dcur6kpoa22dajxu5j4stal6ir6ipl25ly IRIS_313-11199/IRIS_313-11199.realigned.bam.stats
added QmaKpQCZxUGtVhqPYzL6mL34UtMZsDnCijq43Q8JvW9e2P IRIS_313-11199/IRIS_313-11199.snp.vcf.gz
added QmNgyyKTdSPRmGMDTNdKXBZi3m3gcH7aEfsr23T2NQwmJA IRIS_313-11199/IRIS_313-11199.snp.vcf.gz.tbi
added QmRUzQCpixgFCbfJsKsn9KfQAE7jWpAmVMQwergzVewNiG IRIS_313-11199

Then I export the dag as a car:
ipfs dag export QmRUzQCpixgFCbfJsKsn9KfQAE7jWpAmVMQwergzVewNiG > IRIS_313-11199.car

Third I imported the car into lotus
lotus client import --car ~/IRIS_313-11199.car 
Import 3, Root QmRUzQCpixgFCbfJsKsn9KfQAE7jWpAmVMQwergzVewNiG

Last I try to retrieve a single file
lotus client retrieve --allow-local --datamodel-path-selector 'Links/0/Hash' QmRUzQCpixgFCbfJsKsn9KfQAE7jWpAmVMQwergzVewNiG ~/output

I expect to get a single file but instead I got a folder containing all files from the dag. I also tried 'Links/1/Hash'  and 'Links/0/Hash/Links/1/Hash', no difference. The "output" is always a folder with all files. 
ll ~/output
total 5425240
drwxrwxr-x  2 dev dev       4096 Jun  6 05:51 ./
drwxr-xr-x 12 dev dev       4096 Jun  6 05:51 ../
-rw-rw-r--  1 dev dev 3597111237 Jun  6 05:50 IRIS_313-11199.realigned.bam
-rw-rw-r--  1 dev dev    1797448 Jun  6 05:50 IRIS_313-11199.realigned.bam.bai
-rw-rw-r--  1 dev dev        382 Jun  6 05:50 IRIS_313-11199.realigned.bam.stats
-rw-rw-r--  1 dev dev 1955950650 Jun  6 05:51 IRIS_313-11199.snp.vcf.gz
-rw-rw-r--  1 dev dev     558311 Jun  6 05:51 IRIS_313-11199.snp.vcf.gz.tbi

ribasushi:
Debugged in fil-slack - operator error:
```
lotus version
Daemon: 1.9.0+mainnet+git.ada7f97ba+api1.3.0
Local: lotus version 1.11.0-dev+mainnet+git.7f1712267
```

kernelogic:
Yes it totally works now, after restart daemon.

kernelogic:
Hi @ribasushi when will this be merged and released? Would be really handy to demo it for slingshot 2.4.

ribasushi:
@kernelogic this PR hard-depends on https://github.com/filecoin-project/lotus/pull/6375 , which in turn depends on ensuring backwarrds compatibility between old and new nodes. I do not have good visibility into the near-term roadmap :( 

Note, however, that the changes/functionality in this PR are client-side only. Means you can just build that and demo it ( I just updated it to master-after-hyperdrive so one can continue using it on mainnet  )

kernelogic:
Thank you thank you! I can definitely use this branch for demoing this feature now.

ribasushi:
Converting to draft, since I need to do a couple things after https://github.com/filecoin-project/lotus/pull/6375 merges

ribasushi:
Ok, this is now again ready, rebased on latest dagstore work, with exhaustive tests including working with a partial-dag car and a full real miner storage/dealpublish/retrieval cycle ( 👋 @mikeal ).

The original seed of this code is nearly half a year old at this stage, with folks like @kernelogic using it in prod. We should make a call whether we want to ship it, or close this PR officially for good.
/cc @BigLep 

@magik6k: this adds one extra field to the API, but the field is default-nullable. If how I did it is not ok - let me know how to fix ( or feel free to push a fix to the branch if it is faster than explaining )

kernelogic:
I would like this to get released. I provide support for this feature in my recent slingshot projects.

BigLep:
Thanks @ribasushi .  I'm not up on all the needs/usecases here.  Assuming the change needs code quality / testing standards, what would be the reasons not to ship this?  Is it moving us in a direction counter to long term plans?  Are we worried about the maintenance overhead?

ribasushi:
@BigLep I have no worries personally, and maintenance overhead should be trivial. The reason I raised a question of "do we want this" is due to:
- the general difficulty of getting this shipped for protractd reasons outside of this PR
- It is a feature similar to https://github.com/filecoin-project/lotus/pull/7313, with the core difference of not requiring anything from the miner, pushing all indexing/record-keeping work onto the client ( in fact this code works against any unmodified miner currently attached to the network, even `1.10.0` )

api/api_full.go
```diff
@@ -931,9 +932,10 @@ type MarketDeal struct {
 
 type RetrievalOrder struct {
 	// TODO: make this less unixfs specific
-	Root  cid.Cid
-	Piece *cid.Cid
-	Size  uint64
+	Root                  cid.Cid
+	Piece                 *cid.Cid
+	DatamodelPathSelector *textselector.Expression
```
ribasushi, line 937: @magik6k poke that this still needs your 👍 or a recipe how to do it correctly

rvagg:
Hey, so a few items to throw in for consideration now I'm getting my head into this space and catching up on this PR:

1. go-ipfs uses the `/ipld/` prefix for paths to get _down to the datamodel_ and this is becoming much clearer with 0.10 which essentially says "strip out the UnixFS ADL loader" before traversing, which is what's going on here. Perhaps this could be repurposed here so as to avoid the `datamodel` term in the command? Strictly require that all requests here have an `/ipld/` prefix and reject those that don't—with the implication that a future iteration will support bare paths that will walk over ADLs (in fact that's possible now, here, with UnixFS at least). I'm not sure everyone's comfortable with this `/ipld/` thing, but it's already in there and it's so far the only other mechanism in play for selecting DM vs ADL (this PR being the second proposed mechanism).
2. Does the mixing of "path" and "selector" concern anyone else here? They seem like concepts we may want to retain a small barrier between. Strictly speaking, a "path" can be a subset of a "selector", but do we want them jumbled into the same command name?
3. Does this take into account paths that terminate inside a block? My understanding of the code here is that it'll give you the "last loaded block" which means that it probably works just fine, but does it make sense that I can do `/Links/10/Hash/Links/50` and get the same result as `/Links/10/Hash/Data`, or the more strictly correct `/Links/10/Hash`? Should there be an error if your path doesn't terminate at a CID (or "block boundary" because `/` would also be correct I assume).

So, if acting on both of those items, this command could be `--root-path`, or `--subroot-path`. Then it's (a) strictly understood to be an IPLD node path (hopefully to a block boundary) and with a mandatory `/ipld/` prefix it's known to be operating at the datamodel layer.

ribasushi:
@rvagg as mentioned in https://github.com/filecoin-project/lotus/pull/6393#discussion_r645743719 I have very strong reservations about this part:

> a future iteration will support bare paths that will walk over ADLs

There **must be a way** for me to say "under no circumstances attempt a heuristic - just do what I asked". Hence the `datamodel` part to both
1. imply in the api-contract an ADL-parse will not be attempted
2. leaves room for "proper" JSON selector specifications as currently used by @willscott in the `go-car` CLI

> Should there be an error if your path doesn't terminate at a CID

it already does error: https://github.com/filecoin-project/lotus/pull/6393/files#diff-b947764ffb1599ab13eb67261ce286ab799a34ed90124c1d1352b1a48fb1766aR1041-R1045

ribasushi:
@rvagg another more subtle consideration is that the API and CLI here are strictly for unixfs, nothing else ( it is the lotus poof-of-concept for retrievals, if you are doing anything more involved you should be piping into an ipfs instance anyway ). The lotus API in this area is very firmly centered aroud a "root CID", so it was expedient to say "and also this path off of that CID you just got".

Even if I "fix" the CLI ([hard, because positional mandatory args](https://github.com/filecoin-project/lotus/blob/0444435589d0c6817072278e222346829ed637a4/cli/client.go#L1090)), the API is even more difficult to retrofit.

rvagg:
> Should there be an error if your path doesn't terminate at a CID
> 
> > it already does error: https://github.com/filecoin-project/lotus/pull/6393/files#diff-b947764ffb1599ab13eb67261ce286ab799a34ed90124c1d1352b1a48fb1766aR1041-R1045

Does that actually error if you go deeper than a block boundary though? It looks to me like it'll allow you to `/Links/10/Hash/Links/50` and `/Links/10/Hash/Data` (and more) to match what should be just `/Links/10/Hash` because it's `p.LastBlock` that you're picking up, regardless of what `n` happens to end up as. Maybe not a show-stopper (and I'm not sure how you easily address this) but could be cause for API surprise by someone?

ribasushi:
Ahhhhh I see what you mean, yes. It will not error, no.
@warpfork what would be the hand-wavey code to detect "where am I right now boundary-wise" ?

ribasushi:
This PR has been updated once again to track latest master. In fact it had to be merged on top of (and to thus include)  https://github.com/filecoin-project/lotus/pull/7441, because `go-ipld-prime` interfaces changed in a patch version *again* 🤦‍♂️ 

There are now [negative-tests](https://github.com/filecoin-project/lotus/commit/af0d9b650#diff-7cecd455116bb046634b469a4e86e3c143b53241750c85af87dfd3e799951ac4R127-R159) addressing most of the comments above, everything is green, in fact coverage went up.

Also an identical CLI flag with the same functionality got [merged into the `filc` standalone client](https://github.com/application-research/filclient/pull/12). For folks who need the feature today - you can try to switch to the above.

mvdan:
>  because `go-ipld-prime` interfaces changed in a patch version _again_

Which interface, out of curiosity? Nothing came to mind between v0.12.0 and v0.12.3, and I just skimmed the list of commits and didn't spot anything obvious.

ribasushi:
`CommonSelector_MatchAllRecursively` became something else

node/impl/client/client.go
```diff
@@ -995,6 +1025,46 @@ func (a *API) clientRetrieve(ctx context.Context, order api.RetrievalOrder, ref
...
+		if err := utils.TraverseDag(
+			ctx,
+			ds,
+			root,
+			selspec.Node(),
+			func(p traversal.Progress, n ipld.Node, r traversal.VisitReason) error {
+				if r == traversal.VisitReason_SelectionMatch {
+
+					if p.LastBlock.Path.String() != p.Path.String() {
+						return xerrors.Errorf("unsupported selection path '%s' does not correspond to a node boundary (a.k.a. CID link)", p.Path.String())
```
rvagg, line 1044:
```suggestion
						return xerrors.Errorf("unsupported selection path '%s' does not correspond to a block boundary (a.k.a. CID link)", p.Path.String())
```

rvagg:
Ahh, sorry, I suppose that was me. Trying to move away from exposing bare `Selector` instances and push the `Node` form everywhere it's used externally, there'll probably be more breakages around that to come at some point, hopefully in a non-patch release!

node/impl/client/client.go
ribasushi: God catch, fixed!

go.mod
```diff
@@ -101,7 +101,9 @@ require (
 	github.com/ipfs/interface-go-ipfs-core v0.4.0
 	github.com/ipld/go-car v0.3.2-0.20211001225732-32d0d9933823
 	github.com/ipld/go-car/v2 v2.0.3-0.20210811121346-c514a30114d7
+	github.com/ipld/go-codec-dagpb v1.3.0
 	github.com/ipld/go-ipld-prime v0.12.3
+	github.com/ipld/go-ipld-selector-text-lite v0.0.0-20210817134355-4c190a2bb825
```
Stebalien, line 106: @ribasushi or @rvagg please tag this.

PR #6403, opened 2021-06-07
upgrade testground action version

coryschwartz:
github action v1.0 is not including the --metadata flags so the version information is displayed in testground. Upgrading the github action to a version that does.


coryschwartz:
@travisperson I just cut a release for this action v1.1. The difference is that v1.1 sends `--metadata` flag, which is responsible for the "created" column you see here https://ci.testground.ipfs.team/tasks.

magik6k (approved):

PR #6405, opened 2021-06-07
implement a command to export a car

Stebalien:
Add a lotus-shed command for exporting a car of an arbitrary dag.

Currently, we allow exporting tipsets, etc. But we don't have any convenient commands for simply exporting an arbitrary dag for analysis. This is useful for, e.g., analyzing bad actor states.

### DIFF
--- a/cmd/lotus-shed/export-car.go
+++ b/cmd/lotus-shed/export-car.go
@@ -1,0 +1,103 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/ipfs/go-blockservice"
+	"github.com/ipfs/go-cid"
+	offline "github.com/ipfs/go-ipfs-exchange-offline"
+	format "github.com/ipfs/go-ipld-format"
+	"github.com/ipfs/go-merkledag"
+	"github.com/ipld/go-car"
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+
+	lcli "github.com/filecoin-project/lotus/cli"
+	"github.com/filecoin-project/lotus/node/repo"
+)
+
+func carWalkFunc(nd format.Node) (out []*format.Link, err error) {
+	for _, link := range nd.Links() {
+		if link.Cid.Prefix().Codec == cid.FilCommitmentSealed || link.Cid.Prefix().Codec == cid.FilCommitmentUnsealed {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


cmd/lotus-shed/export-car.go
```diff
@@ -0,0 +1,104 @@
...
+			continue
+		}
+		out = append(out, link)
+	}
+	return out, nil
+}
+
+var exportCarCmd = &cli.Command{
+	Name:        "export-car",
+	Description: "Export a car from repo (requires node to be offline)",
```
magik6k, line 34: We could support online-mode with the API blockstore (`blockstore.NewAPIBlockstore`)

magik6k (approved):

cmd/lotus-shed/export-car.go
Stebalien: Is there a good way to detect this?

magik6k: I guess do something like what we do in https://github.com/filecoin-project/lotus/blob/master/cli/wait.go

Could also just be an explicit --offline flag

cmd/lotus-shed/export-car.go
```diff
@@ -0,0 +1,104 @@
...
+		var roots []cid.Cid
+		for _, arg := range cctx.Args().Tail() {
+			c, err := cid.Decode(arg)
+			if err != nil {
+				return err
+			}
+			roots = append(roots, c)
+		}
+
+		ctx := context.TODO()
```
magik6k, line 56: cctx.Context()

magik6k:
(comments not blocking merge, this is already pretty useful)

cmd/lotus-shed/export-car.go
Stebalien: Eh, I'd rather not spend time on this for now.

Stebalien:
@magik6k I'm not able to merge this.

cmd/lotus-shed/export-car.go
```diff
@@ -0,0 +1,104 @@
+package main
+
+import (
+	"context"
```
magik6k, line 4: CI says this is unused

Stebalien: "just one little change, what could go wrong"

magik6k (approved):

Issue #6156, opened 2021-04-30
Large scale testing post FIP13/FIP8


Stebalien:
When FIP13 & FIP8 land, onboarding throughput should increase. We need to verify that lotus can handle this.

1. Estimate maximum onboarding rate given FIP13 & FIP8 and current gas limits.
2. Given this onboarding rate, estimate the number of on-chain sectors in 1, 2, 5 years.
3. Verify that lotus can support this number of sectors:
    1. Window PoSt messages can be submitted.
    2. Sector expiration/termination won't cause performance issues.
    3. Sector faults/missed window PoSt messages won't cause performance issues.
    4. State doesn't grow too large.

Outputs:

- Spreadsheet with estimates/calculations for chain size, timing, object size, maximum miner sizes, etc.
- A tool that creates a large state snapshot so we start a devnet from a "large" state.
- Timing analysis from a devnet using that tool.

Related items:
- [x] #6478 
- [ ] #6476 
- [ ] #6424

BigLep:
For 1, @ZenGround0 has a spreadsheet for the estimation.  

Scaling:
1. Individual object size.  Programatically construct a state tree of some size.  
- Devnet too slow
- Manually generate the state tree that is massive
 - Potentially let this be a devnet seed?  (lotus-shed tool)

Other conversation notes:
- Minimum: add a step to our release process
- Look into testground

Kubuxu:
We should also evaluate the scaling cost of cron as well as state churn (amount of state that gets rewritten per epoch on average) as this determines size of full nodes and snapshots even more so than state size itself.

Stebalien:
Luckily, in the "good" case, cron only touches partitions if something goes wrong, and never touches sectors. But yeah, we need to consider the "everyone faults all at once" case.

Stebalien:
So, I think the whole dev-net part of this is a no-go, unfortunately, because we'd actually need to prove the 2k sectors.

Instead, I'm going to go for a local simulation.

Acumes:
I want to deploy 2K internal network in the test network to verify this function. What should I do?

Stebalien:
Figuring that out is part of the point of this project. However, actually running this instead of just simulating it may not be possible. Assuming a 10x growth (50EiB target), you'd need:

1. At least 3TiB of storage. Plus overhead so let's call it 10TiB.
2. Enough compute to compute ~500 window posts a minute (or fake it).

So my current plan is to simulate parts of it.

[...]

PR #6406, opened 2021-06-07
feat: implement lotus-sim

Stebalien:
Implements #6156.

**Usage**

```bash
> make build
> cd cmd/lotus-sim
> go build .
> ./lotus-sim create
> ./lotus-sim run
```

Limitations: to run this, you need 2 weeks worth of state-trees to get power lookback (used to determine per-miner onboarding rates). Ideally, we'd do something else (maybe just look at the number of sectors and make assumptions?

**Goals**

- Test how the lotus VM and actors as the network grows.
- Find potential bugs that only appear at scale.
- Determine state growth rate.
- Test network upgrades.

**Non Goals**

This simulation does not:

1. Use real proofs/sectors.
2. Sign anything.
3. Verify any signatures.
4. Use a real miner. It just keeps using the last miner to mine a block (from when the network started).
5. Do any "network" related operations. It just creates and validates (skipping proofs) a fake chain.

**Design**

This simulation:

1. Forks the network at the current head (or at a desired point, see `./lotus-sim create --help`).
2. At every epoch:
    1. Funds the "funding" account (f0100) if necessary (by taking from random actors).
    2. Submits window posts for every "healthy" sector.
    3. Submits (batch) prove commit messages for any outstanding pre-commits that are ready to be proven.
    4. Submits (batch) pre-commit messages in the remaining space.

When necessary, the "funding" account will send funds to actors.


Stebalien:
To the reviewer: Most of the changes here are just a part of the simulation, but not all:

1. It adds some methods to the actors shims.
2. It exposes the ActorStore from the VM (so we can read intermediate state and, importantly, nonces).
3. It sets ActorErr in one VM.ApplyMessage case where it should have been set but wasn't (not consensus critical).
4. It modifies StateTree.ForEach to iterate over un-flushed actors.
5. It fixes the genesis logic to set the balance (to zero) for some actors. This won't actually affect anything as we default to zero on serialize anyways. Unfortunately, that doesn't help when looking at unseralized actors in `ForEach` (see 4).

Please pay careful attention to 4. It shouldn't cause any problems (mostly used in genesis and some specific upgrades), and brings the function in-line with expectations. But it's still something that's good to look at carefully.

cmd/lotus-sim/simulation/stages/commit_queue.go
```diff
@@ -0,0 +1,200 @@
...
+	"sort"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
+	"github.com/filecoin-project/lotus/chain/actors/policy"
+)
+
+// pendingCommitTracker tracks pending commits per-miner for a single epohc.
```
magik6k, line 13:
```suggestion
// pendingCommitTracker tracks pending commits per-miner for a single epoch.
```

cmd/lotus-sim/simulation/stages/funding_stage.go
```diff
@@ -0,0 +1,316 @@
...
+	return res, err
+}
+
+func (fs *FundingStage) fund(bb *blockbuilder.BlockBuilder, target address.Address, times int) error {
+	amt := TargetFunds
+	if times > 0 {
+		if times >= 8 {
+			times = 8 // cap
+		}
+		amt = big.Lsh(amt, uint(times))
```
magik6k, line 90: `times` sounds like a weird param name for something we put into Lsh

cmd/lotus-sim/upgrade.go
```diff
@@ -0,0 +1,109 @@
...
+	Name:        "set",
+	ArgsUsage:   "<network-version> [+]<epochs>",
+	Description: "Set a network upgrade height. Prefix with '+' to set it relative to the last epoch.",
+	Action: func(cctx *cli.Context) (err error) {
+		args := cctx.Args()
+		if args.Len() != 2 {
+			return fmt.Errorf("expected 2 arguments")
+		}
+		nvString := args.Get(0)
+		networkVersion, err := strconv.ParseInt(nvString, 10, 64)
```
magik6k, line 75: CodeQL complains about this not being
```suggestion
		networkVersion, err := strconv.ParseUint(nvString, 10, 32)
```

magik6k (approved):
Mostly cosmetic nits.

It would also be really great to have at least a basic integration test, so that we know when it would get completely broken by some change.

Should also have a make target, and be added to `BINS` so it's built in `make buildall`

chain/actors/builtin/miner/state.go.template
```diff
@@ -255,6 +282,42 @@ func (s *state{{.v}}) GetProvingPeriodStart() (abi.ChainEpoch, error) {
...
+	}
+
+	sectors := make([]abi.SectorNumber, 0, count)
+	for iter.HasNext() && len(sectors) < count {
+		nextNo, err := iter.Next()
+		if err != nil {
+			return nil, err
+		}
+		sectors = append(sectors, abi.SectorNumber(nextNo))
+	}
```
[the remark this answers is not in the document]
Stebalien: Yes, but I have a bitfield iterator, not a bitfield (and I'd rather not serialize back to a bitfield unless you feel strongly about this).

chain/vm/vm.go
```diff
@@ -439,6 +439,8 @@ func (vm *VM) ApplyMessage(ctx context.Context, cmsg types.ChainMsg) (*ApplyRet,
 			},
 			GasCosts: &gasOutputs,
 			Duration: time.Since(start),
+			ActorErr: aerrors.Newf(exitcode.SysErrSenderInvalid,
+				"message gas limit does not cover on-chain gas costs"),
```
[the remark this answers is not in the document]
Stebalien: Yes, yes it should...

magik6k (approved):

PR #6416, opened 2021-06-08
Commit and Precommit batcher cannot share a getSectorDeadline method

### LINKED ISSUE #6412: CommitBatcher getSectorDeadline should check the earlier PreCommit expiry in the batch
Currently, PreCoomitBatcher and CommitBatcher shares getSectorDeadline. https://github.com/filecoin-project/lotus/blob/fadc79a4875bf965faa9c749856fb5038527d8b2/extern/storage-sealing/commit_batch.go#L455

however, in additional to deal expirations, commit batcher should also make sure it pushes the aggregation message before any of the sectors is going to expired for MaxProveCommitDuration. Therefore, instead of checking `si.TicketEpoch + policy.MaxPreCommitRandomnessLookback` CommitBatcher should be checking [SectorPreCommitOnChainInfo.PreCommitEpoch](https://github.com/filecoin-project/specs-actors/blob/0c296bb386bf18065e604e07eaff3c9807ec0757/actors/builtin/miner/miner_state.go#L156) + [MaxProveCommitDuration](https://github.com/filecoin-project/specs-actors/blob/0c296bb386bf18065e604e07eaff3c9807ec0757/actors/builtin/miner/policy.go#L152)

arajasek:
Fixes #6412 

Two commits:

- Rename "deadlines" as requested by @jennijuju to avoid confusion with window post deadlines
- Split the `getSectorCutoff` methods between precommit and commit batchers

### DIFF
--- a/extern/storage-sealing/commit_batch.go
+++ b/extern/storage-sealing/commit_batch.go
@@ -51,9 +51,9 @@ type CommitBatcher struct {
 	getConfig GetSealingConfigFunc
 	prover    ffiwrapper.Prover
 
-	deadlines map[abi.SectorNumber]time.Time
-	todo      map[abi.SectorNumber]AggregateInput
-	waiting   map[abi.SectorNumber][]chan sealiface.CommitBatchRes
+	cutoffs map[abi.SectorNumber]time.Time
+	todo    map[abi.SectorNumber]AggregateInput
+	waiting map[abi.SectorNumber][]chan sealiface.CommitBatchRes
 
 	notify, stop, stopped chan struct{}
 	force                 chan chan []sealiface.CommitBatchRes
@@ -70,9 +70,9 @@ func NewCommitBatcher(mctx context.Context, maddr address.Address, api CommitBat
 		getConfig: getConfig,
 		prover:    prov,
 
-		deadlines: map[abi.SectorNumber]time.Time{},
-		todo:      map[abi.SectorNumber]AggregateInput{},
-		waiting:   map[abi.SectorNumber][]chan sealiface.CommitBatchRes{},
+		cutoffs: map[abi.SectorNumber]time.Time{},
+		todo:    map[abi.SectorNumber]AggregateInput{},
+		waiting: map[abi.SectorNumber][]chan sealiface.CommitBatchRes{},
 
 		notify:  make(chan struct{}, 1),
 		force:   make(chan chan []sealiface.CommitBatchRes),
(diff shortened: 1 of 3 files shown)


extern/storage-sealing/types.go
```diff
@@ -88,7 +88,7 @@ type SectorInfo struct {
 	PreCommitInfo    *miner.SectorPreCommitInfo
 	PreCommitDeposit big.Int
 	PreCommitMessage *cid.Cid
-	PreCommitTipSet  TipSetToken
```
magik6k, line 91: SectorInfo is an on-disk struct, and we probably shouldn't break it if we can.

(It's PreCommitTipSet instead of just height because we used to use it in lookups in later states, but that wasn't reorg-proof enough. Now it's here mostly for debugging)

extern/storage-sealing/commit_batch.go
```diff
@@ -452,24 +456,36 @@ func (b *CommitBatcher) Stop(ctx context.Context) error {
...
+		return time.Now(), err
+	}
+
+	nv, err := b.api.StateNetworkVersion(b.mctx, tok)
+	if err != nil {
+		log.Errorf("getting network version: %s", err)
+		return time.Now(), err
+	}
+
+	cutoffEpoch := si.PreCommitHeight + policy.GetMaxProveCommitDuration(actors.VersionForNetwork(nv), si.SectorType)
```
magik6k, line 472: If we've just added the PreCommitHeight field, it won't be present on existing sectors in existing miners

arajasek:
@magik6k Yeah, makes sense. I'll make the CommitBatcher look up the height.

extern/storage-sealing/commit_batch.go
```diff
@@ -452,24 +456,43 @@ func (b *CommitBatcher) Stop(ctx context.Context) error {
...
+	tok, curEpoch, err := b.api.ChainHead(b.mctx)
+	if err != nil {
+		log.Errorf("getting chain head: %s", err)
+		return time.Now(), nil
+	}
+
+	nv, err := b.api.StateNetworkVersion(b.mctx, tok)
+	if err != nil {
+		log.Errorf("getting network version: %s", err)
+		return time.Now(), err
```
magik6k, line 469: I'd just wrap these with xerrors.Errorf, and drop the log - errors returned from AddCommit are saved in sector logs, which makes it much easier to debug individual sectors

extern/storage-sealing/commit_batch.go
```diff
@@ -452,24 +456,43 @@ func (b *CommitBatcher) Stop(ctx context.Context) error {
 	}
 }
 
-func getSectorDeadline(curEpoch abi.ChainEpoch, si SectorInfo) time.Time {
-	deadlineEpoch := si.TicketEpoch + policy.MaxPreCommitRandomnessLookback
+func (b *CommitBatcher) getCommitCutoff(si SectorInfo) (time.Time, error) {
```
magik6k, line 459: We should consider changing this function to return epochs, which would make it much easier to do testing with (not in this PR tho)

arajasek: Yeah, agreed...adding a TODO in the code

extern/storage-sealing/commit_batch.go
arajasek: seems reasonable

magik6k (approved):

Issue #6338, opened 2021-05-27
[NV13] Mpool will reject sufficiently large batch prove commits


Stebalien:
As of network version 13, batch prove commit proofs can be up to 41KiB (last I checked?). However, the message pool rejects messages over 32KiB. We're going to need to up this limit.

https://github.com/filecoin-project/lotus/blob/2b93bcde72b869e8936ce7254acf6e0ad7c937bf/chain/messagepool/messagepool.go#L550-L552

PR #6419, opened 2021-06-08
Increase message size limit

Kubuxu:
Adds test for it
Resolves https://github.com/filecoin-project/lotus/issues/6338
Needs to be picked up by feat/nv13

### DIFF
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -59,6 +59,8 @@ var MaxUntrustedActorPendingMessages = 10
 
 var MaxNonceGap = uint64(4)
 
+const MaxMessageSize = 64 << 10 // 64KiB
+
 var (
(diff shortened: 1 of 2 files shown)


chain/messagepool/messagepool.go
```diff
@@ -665,7 +665,7 @@ func (mp *MessagePool) Push(ctx context.Context, m *types.SignedMessage) (cid.Ci
 
 func (mp *MessagePool) checkMessage(m *types.SignedMessage) error {
 	// big messages are bad, anti DOS
-	if m.Size() > 32*1024 {
+	if m.Size() > 64*1024 {
```
vyzo, line 668: maybe we should make it a named constant?

vyzo (approved):

Stebalien (approved):

magik6k (approved):

Issue #6407, opened 2021-06-07
Add Miner message cost configuration for ProveCommitAggregated and PreCommitBatches


jennijuju:
Ideally, we have a per sector cost setting.

PR #6420, opened 2021-06-08
Fee config for sector batching

magik6k:
Fixes https://github.com/filecoin-project/lotus/issues/6407


arajasek:
Can we send stuff like this into `release/v1.10.0` instead, and then just periodically merge that branch into `master`?

PR #6428, opened 2021-06-09
Fund miners with the aggregate fee when ProveCommitting

arajasek:


extern/storage-sealing/commit_batch.go
```diff
@@ -285,7 +286,14 @@ func (b *CommitBatcher) processBatch(cfg sealiface.Config) ([]sealiface.CommitBa
 		return []sealiface.CommitBatchRes{res}, xerrors.Errorf("couldn't get miner info: %w", err)
 	}
 
-	goodFunds := big.Add(b.feeCfg.MaxCommitGasFee, collateral)
+	bf, err := b.api.ChainBaseFee(b.mctx, tok)
+	if err != nil {
+		return []sealiface.CommitBatchRes{res}, xerrors.Errorf("couldn't get base fee: %w", err)
+	}
+
+	aggFee := miner5.AggregateNetworkFee(len(infos), bf)
```
arajasek, line 294: This should be added to the actor policy wrappers and abstracted, but I was lazy...Can do so if people feel strongly.

Stebalien: I'd just do it. Otherwise, we'll miss it if it changes (and this is something I can definitely see changing.

extern/storage-sealing/commit_batch.go
```diff
@@ -285,7 +286,14 @@ func (b *CommitBatcher) processBatch(cfg sealiface.Config) ([]sealiface.CommitBa
...
+
+	goodFunds := big.Add(b.feeCfg.MaxCommitGasFee, big.Add(collateral, aggFee))
```
Stebalien, line 296: Hm.... shouldn't we have a new field for this? Or shouldn't the max fee be a function of the number of sectors, or something like that?

magik6k: https://github.com/filecoin-project/lotus/pull/6420 does that

extern/storage-sealing/commit_batch.go
arajasek: kk done

arajasek:
Depends on #6416 now

magik6k (approved):

PR #6430, opened 2021-06-09
Fix: correct the change of message size limit

GFZRZK:
the change of MaxMessageSize should take effect in all place you used


magik6k (approved):

vyzo (approved):

arajasek:
Thanks for the PR!

PR #6432, opened 2021-06-09
Unit tests for sector batchers

magik6k:
Could use slightly more cases, but this should cover most of the logic


arajasek (approved):
SGWM

PR #6445, opened 2021-06-10
itests/kit: add guard to ensure imports from tests only.

raulk:


magik6k (approved):
Looks like a hack, but it seems to work

raulk:
Unfortunately it seems to be the only accurate way to detect when we're running inside a test.

Issue #6390, opened 2021-06-04
[BUG] CommitBatch should clean FaildSector


GFZRZK:
Assmuing a sector failed in [sc, err := b.getSectorCollateral(id, tok)](https://github.com/filecoin-project/lotus/blob/v1.10.0-rc1/extern/storage-sealing/commit_batch.go#L242), it will be add to [FailedSector](https://github.com/filecoin-project/lotus/blob/v1.10.0-rc1/extern/storage-sealing/commit_batch.go#L244), but after processBatch, this field seems not be cleanned corretly.

GFZRZK:
The code is based on v1.10-rc1, if you already fix it then just ignore it.

PR #6451, opened 2021-06-11
clean failed sectors in batch commit

GFZRZK:
solve the problem https://github.com/filecoin-project/lotus/issues/6390


magik6k (commented):
Oh, good catch, but the bug is that those failed sectors aren't in `res.Sectors`, which was meant to have all sectors, including failed sectors

GFZRZK:
Yeah, putting failed or success sectors into `res` seems like a better fix-up

magik6k (approved):
Thanks for the PR!

Issue #6431, opened 2021-06-09
Support early finalization before commit


jennijuju:
Currently lotus finalizing the sectors locally after sectors are onchain, however if a something goes wrong  in the FinalizeSector step, miners won’t be able to proving it and will have to terminate the sectors.

Allow early finalization can prevent it bit.

This is also important with ProveCommitAggregate, it may trigger a large amount of transfers from worker to the miner and can be too heavy for miners current setup to handle. Although `ParallelFetchLimit` can help a little with that.. it’s not ideal. It will be ideal if we can have #5339 , but for now, have an option for early finalization helps too.

hxg8202:
It is important

duzy555666:
if sector wait ProCommitAggregate  ,sector Will it go to the next step, or is it just waiting ?

PR #6452, opened 2021-06-11
sealing: Early finalization option

magik6k:
Closes https://github.com/filecoin-project/lotus/issues/6431


magik6k:
(will add more tests after this gets backported to master (mostly to avoid conflicts - integration tests are passing with early finalization enabled in the default config))

arajasek (approved):

PR #6467, opened 2021-06-14
convert paych tests to kit2

dirkmc:
Test updated and passing

### DIFF
--- a/itests/paych_api_test.go
+++ b/itests/paych_api_test.go
@@ -38,14 +38,13 @@ func TestPaymentChannelsAPI(t *testing.T) {
 		miner           kit2.TestMiner
 	)
 
-	//n, sn := kit2.MockMinerBuilder(t, kit2.TwoFull, kit2.OneMiner)
 	ens := kit2.NewEnsemble(t, kit2.MockProofs()).
 		FullNode(&paymentCreator).
 		FullNode(&paymentReceiver).
 		Miner(&miner, &paymentCreator).
 		Start().
 		InterconnectAll()
-	bms := ens.BeginMining(blockTime, &miner)
+	bms := ens.BeginMining(blockTime)
 	bm := bms[0]
 
 	// send some funds to register the receiver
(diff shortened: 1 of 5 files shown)


itests/paych_api_test.go
```diff
@@ -24,44 +23,38 @@ import (
...
+	)
+
+	//n, sn := kit2.MockMinerBuilder(t, kit2.TwoFull, kit2.OneMiner)
+	ens := kit2.NewEnsemble(t, kit2.MockProofs()).
+		FullNode(&paymentCreator).
+		FullNode(&paymentReceiver).
+		Miner(&miner, &paymentCreator).
+		Start().
+		InterconnectAll()
+	bms := ens.BeginMining(blockTime, &miner)
```
raulk, line 48: No need to supply the miner. If not supplied, by default we'll begin mining on all miners ;-)

itests/paych_api_test.go
```diff
@@ -24,44 +23,38 @@ import (
...
 	receiverAddr, err := paymentReceiver.WalletNew(ctx, types.KTSecp256k1)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	kit.SendFunds(ctx, t, paymentCreator, receiverAddr, abi.NewTokenAmount(1e18))
+	kit2.SendFunds(ctx, t, paymentCreator, receiverAddr, abi.NewTokenAmount(1e18))
 
 	// setup the payment channel
 	createrAddr, err := paymentCreator.WalletDefaultAddress(ctx)
```
raulk, line 60: Can we replace all the error checks + failure with require?

raulk:
Mining is stopping halfway through this test because when we inject nulls, the chain advances too quickly and eventually it reaches the present, so we start waiting on natural time instead of doing catch-up mining. Looking into it.

raulk:
I pushed a few commits to fix the problem.

raulk (approved):

PR #6469, opened 2021-06-14
Refactor verifreg test to kit2

dirkmc:
TODO: currently the test fails with
```
    verifreg_test.go:64: AddVerifier failed:  getting key address: failed to get account actor state for f080: unknown actor code bafkqadtgnfwc6mrpnv2wy5djonuwo
```
when calling the AddVerifier method


raulk (commented):
Let's migrate to require 🙏

raulk (approved):
I fixed what was happening here. The issue is that when I implemented kit2, I set up the **default** verifier root key in the genesis, which is a multisig of 1 with a hardcoded address. Therefore, the test had no permissions to add a verifier!

I fixed this by adding a new ensemble option that enables us to specify the root verifier key at genesis, also assigning it some balance (so it can actually send messages to add verifiers). The option is called `RootVerifier`.

I also took the opportunity to add an `Account` functional option with which one can set up arbitrary accounts with arbitrary balances in genesis.

We use this to set up the verifier account.

@dirkmc -- there was some other test that was sending balance from a miner to a new account. That test can be migrated to the `Account()` option 🙏

raulk (approved):

itests/kit2/node_opts_nv.go
```diff
@@ -1,36 +1,64 @@
...
-	}, {
-		Network:   network.Version12,
-		Height:    3,
-		Migration: stmgr.UpgradeActorsV4,
-	}, {
-		Network:   network.Version13,
-		Height:    4,
-		Migration: stmgr.UpgradeActorsV5,
-	}}
+	fullSchedule := stmgr.UpgradeSchedule{}
```
dirkmc, line 61: Should this be
```
fullSchedule := DefaultTestUpgradeSchedule
```

Otherwise I think the `for` loop is a noop

PR #6470, opened 2021-06-14
Refactor ccupgrade test to use kit2

dirkmc:
Tests passing ✅

### DIFF
--- a/itests/ccupgrade_test.go
+++ b/itests/ccupgrade_test.go
@@ -35,7 +35,8 @@ func runTestCCUpgrade(t *testing.T, upgradeHeight abi.ChainEpoch) {
 	ctx := context.Background()
 	blockTime := 5 * time.Millisecond
 
-	client, miner, ens := kit2.EnsembleMinimal(t, kit2.MockProofs(), kit2.LatestActorsAt(upgradeHeight))
+	opts := kit2.ConstructorOpts(kit2.LatestActorsAt(upgradeHeight))
+	client, miner, ens := kit2.EnsembleMinimal(t, kit2.MockProofs(), opts)
 	ens.InterconnectAll().BeginMining(blockTime)
 
 	maddr, err := miner.ActorAddress(ctx)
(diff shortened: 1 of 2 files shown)


itests/kit2/node_opts_nv.go
```diff
@@ -7,12 +7,12 @@ import (
...
 )
 
-func LatestActorsAt(upgradeHeight abi.ChainEpoch) node.Option {
+func LatestActorsAt(upgradeHeight abi.ChainEpoch) NodeOpt {
 	// Attention: Update this when introducing new actor versions or your tests will be sad
 	return NetworkUpgradeAt(network.Version13, upgradeHeight)
 }
 
-func NetworkUpgradeAt(version network.Version, upgradeHeight abi.ChainEpoch) node.Option {
+func NetworkUpgradeAt(version network.Version, upgradeHeight abi.ChainEpoch) NodeOpt {
```
raulk, line 15: Makes does sense as it makes things more concise. But I think we'll have to revert it eventually because it removes the possibility of accompanying this with multiple `ConstructorOpts`, unless we make those mergeable.

raulk: I would feel better about reverting this now and being OK with the slightly longer-hand form.

dirkmc: Makes sense 👍

raulk (approved):

itests/ccupgrade_test.go
```diff
@@ -3,21 +3,20 @@ package itests
 import (
 	"context"
 	"fmt"
-	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/filecoin-project/lotus/itests/kit"
```
raulk, line 9: Why do we still have this import here? 🤔

raulk (changes requested):

itests/ccupgrade_test.go
dirkmc: Fixed

raulk (approved):

Issue #6418, opened 2021-06-08
PubSub Booster network integration


vyzo:
For context, see https://github.com/protocol/web3-dev-team/pull/104

Code related Action Items:
- [ ] Create bootster pubsub profile.
- [ ] Create nil coldstore abstraction for splitstore.
- [ ] Ensure splitstore works with a nil coldstore; complete or cherrypick #5808 .

Deployment:
- [ ] Coordinate with infra to get machines with DNS and deploy code.
- [ ] Reach out to miners in China and establish direct peering agreements.
- [ ] Reach out to exchanges and establish peering agreements.

f8-ptrk:
>  Reach out to miners in China and establish direct peering agreements.
>  Reach out to exchanges and establish peering agreements.

@vyzo can you elaborate a bit how there agreements would look like and what exactly they are peering?

vyzo:
@f8-ptrk sure.

We want to peer the miner's full node (one of them suffices if they have multiple, they can do their own internal peering for their full nodes) with one of the boosters, and maintain this peering agreement indefinitely.

Basically we need to get their full node multiaddr and have them configure a respective pubsub booster for direct peering.

So let's say that we have full node `/ip4/.../p2p/QmA` peering with booster `/dns4/.../p2p/QmBooster`.
Then the full node will have to add this to its configuration:
```
[Pubsub]
  DirectPeers = ["/dns4/.../p2p/QmBooster"]
```

and the booster will have to add this to its config:
```
[Pubsub]
  DirectPeers = ["/ip4/.../p2p/QmA"]
```

that's all it takes code wise.

NSC-FIL:
If this helps with increasing successful block rewards it is a very welcome change. Seems like lucky value for small and medium miners has dropped since Hyperdrive upgrade.

jennijuju:
> Seems like lucky value for small and medium miners has dropped since Hyperdrive upgrade.

Could you please provide the data source that drives up to this statement?

f8-ptrk:
> If this helps with increasing successful block rewards it is a very welcome change. Seems like lucky value for small and medium miners has dropped since Hyperdrive upgrade.

i honestly doubt it. the math behind it hasn't changed. and if you are winning or not is a local calculations. 

[...]

PR #5808, opened 2021-03-13
Splitstore optimizations

vyzo:
This optimizes the splitstore so that
- we get rid of the full compaction option, as it is not exercised or terribly useful. It was written with an eye to gc, but that one is dangerous.
- we use an optimized version of chain walking that doesn't walk the entire chain and only visits state roots; this keeps the hotstore size effectively constant, instead of slowly growing as it the chain grows.
- we mark from the current epoch to the boundary epoch when we detect sync gaps, in order to avoid pathologies if the node stays offline for an extended period of time.
- we reduce the CompactionThreshold to 3 finalities; that means that we retain only 2-3 finalities in the hot range, reducing the size of the hotstore by 40% (50% right after a compaction).

Also adds a lotus-shed command for aggressive offline gc.

See #5789 

The new chain walking warms up the splitstore 3x faster, while it makes marking 4x faster.

PR #6474, opened 2021-06-14
Splitstore Enhanchements

vyzo:
- Subsumes #6458 (see #6418)
- Cherry-picks and subsumes #5808 

... TBD: Add More Detail ...

### DIFF
--- a/blockstore/noop.go
+++ b/blockstore/noop.go
@@ -1,0 +1,66 @@
+package blockstore
+
+import (
+	"context"
+	"io"
+
+	blocks "github.com/ipfs/go-block-format"
+	cid "github.com/ipfs/go-cid"
+)
+
+var _ Blockstore = (*noopstore)(nil)
+
+type noopstore struct {
+	bs Blockstore
+}
+
+func NewNoopStore(bs Blockstore) Blockstore {
+	return &noopstore{bs: bs}
+}
+
+func (b *noopstore) Has(cid cid.Cid) (bool, error) {
+	return b.bs.Has(cid)
+}
+
+func (b *noopstore) HashOnRead(hor bool) {
+	b.bs.HashOnRead(hor)
+}
+
+func (b *noopstore) Get(cid cid.Cid) (blocks.Block, error) {
+	return b.bs.Get(cid)
+}
+
+func (b *noopstore) GetSize(cid cid.Cid) (int, error) {
+	return b.bs.GetSize(cid)
+}
+
+func (b *noopstore) View(cid cid.Cid, f func([]byte) error) error {
+	return b.bs.View(cid, f)
+}
+
+func (b *noopstore) Put(blk blocks.Block) error {
+	return nil
+}
+
+func (b *noopstore) PutMany(blks []blocks.Block) error {
+	return nil
+}
+
+func (b *noopstore) DeleteBlock(cid cid.Cid) error {
+	return nil
+}
+
+func (b *noopstore) DeleteMany(cids []cid.Cid) error {
+	return nil
+}
+
+func (b *noopstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
+	return b.bs.AllKeysChan(ctx)
+}
+
+func (b *noopstore) Close() error {
+	if c, ok := b.bs.(io.Closer); ok {
+		return c.Close()
+	}
+	return nil
+}
--- a/blockstore/splitstore/splitstore.go
+++ b/blockstore/splitstore/splitstore.go
@@ -42,7 +42,7 @@ var (
 	// === :: cold (already archived)
 	// ≡≡≡ :: to be archived in this compaction
 	// --- :: hot
-	CompactionThreshold = 5 * build.Finality
+	CompactionThreshold = 3 * build.Finality
 
 	// CompactionCold is the number of epochs that will be archived to the
 	// cold store on compaction. See diagram on CompactionThreshold for a
--- a/blockstore/splitstore/splitstore.go
+++ b/blockstore/splitstore/splitstore.go
@@ -1,6 +1,7 @@
 package splitstore
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"errors"
@@ -15,6 +16,7 @@ import (
 	cid "github.com/ipfs/go-cid"
 	dstore "github.com/ipfs/go-datastore"
 	logging "github.com/ipfs/go-log/v2"
+	cbg "github.com/whyrusleeping/cbor-gen"
 
 	"github.com/filecoin-project/go-state-types/abi"
 
@@ -48,7 +50,7 @@ var (
 	CompactionCold = build.Finality
 
 	// CompactionBoundary is the number of epochs from the current epoch at which
-	// we will walk the chain for live objects
+	// we will walk the chain for live objects.
 	CompactionBoundary = 2 * build.Finality
 )
 
@@ -73,7 +75,6 @@ const (
 	batchSize = 16384
 
 	defaultColdPurgeSize = 7_000_000
-	defaultDeadPurgeSize = 1_000_000
 )
 
 type Config struct {
@@ -94,7 +95,6 @@ type ChainAccessor interface {
 	GetTipsetByHeight(context.Context, abi.ChainEpoch, *types.TipSet, bool) (*types.TipSet, error)
 	GetHeaviestTipSet() *types.TipSet
 	SubscribeHeadChanges(change func(revert []*types.TipSet, apply []*types.TipSet) error)
-	WalkSnapshot(context.Context, *types.TipSet, abi.ChainEpoch, bool, bool, func(cid.Cid) error) error
 }
 
 type SplitStore struct {
@@ -104,6 +104,7 @@ type SplitStore struct {
 
 	baseEpoch   abi.ChainEpoch
 	warmupEpoch abi.ChainEpoch
+	warm        bool
 
 	coldPurgeSize int
 
@@ -340,6 +341,7 @@ func (s *SplitStore) Start(chain ChainAccessor) error {
 	switch err {
 	case nil:
 		s.warmupEpoch = bytesToEpoch(bs)
+		s.warm = true
 
 	case dstore.ErrNotFound:
 	default:
@@ -396,7 +398,7 @@ func (s *SplitStore) HeadChange(_, apply []*types.TipSet) error {
 		return nil
 	}
 
-	if s.warmupEpoch == 0 {
+	if !s.warm {
 		// splitstore needs to warm up
 		go func() {
 			defer atomic.StoreInt32(&s.compacting, 0)
@@ -404,7 +406,17 @@ func (s *SplitStore) HeadChange(_, apply []*types.TipSet) error {
 			log.Info("warming up hotstore")
 			start := time.Now()
 
-			s.warmup(curTs)
+			baseTs, err := s.chain.GetTipsetByHeight(context.Background(), s.baseEpoch, curTs, true)
+			if err != nil {
+				log.Errorf("error warming up hotstore: error getting tipset at base epoch: %s", err)
+				return
+			}
+
+			err = s.warmup(baseTs)
+			if err != nil {
+				log.Errorf("error warming up hotstore: %s", err)
+				return
+			}
 
 			log.Infow("warm up done", "took", time.Since(start))
 		}()
@@ -432,14 +444,16 @@ func (s *SplitStore) HeadChange(_, apply []*types.TipSet) error {
 	return nil
 }
 
-func (s *SplitStore) warmup(curTs *types.TipSet) {
+func (s *SplitStore) warmup(curTs *types.TipSet) error {
 	epoch := curTs.Height()
 
 	batchHot := make([]blocks.Block, 0, batchSize)
 	batchSnoop := make([]cid.Cid, 0, batchSize)
 
 	count := int64(0)
-	err := s.chain.WalkSnapshot(context.Background(), curTs, 1, true, true,
+	xcount := int64(0)
+	missing := int64(0)
+	err := s.walk(curTs, epoch,
 		func(cid cid.Cid) error {
 			count++
 
@@ -454,8 +468,14 @@ func (s *SplitStore) warmup(curTs *types.TipSet) {
 
 			blk, err := s.cold.Get(cid)
 			if err != nil {
+				if err == bstore.ErrNotFound {
+					missing++
+					return nil
+				}
 				return err
 			}
+
+			xcount++
 
 			batchHot = append(batchHot, blk)
 			batchSnoop = append(batchSnoop, cid)
@@ -478,39 +498,41 @@ func (s *SplitStore) warmup(curTs *types.TipSet) {
 		})
 
 	if err != nil {
-		log.Errorf("error warming up splitstore: %s", err)
-		return
+		return err
 	}
 
 	if len(batchHot) > 0 {
 		err = s.tracker.PutBatch(batchSnoop, epoch)
 		if err != nil {
-			log.Errorf("error warming up splitstore: %s", err)
-			return
+			return err
 		}
 
 		err = s.hot.PutMany(batchHot)
 		if err != nil {
-			log.Errorf("error warming up splitstore: %s", err)
-			return
+			return err
 		}
 	}
+
+	log.Infow("warmup stats", "visited", count, "cold", xcount, "missing", missing)
 
 	if count > s.markSetSize {
 		s.markSetSize = count + count>>2 // overestimate a bit
 	}
 
 	// save the warmup epoch
+	s.warm = true
 	s.warmupEpoch = epoch
 	err = s.ds.Put(warmupEpochKey, epochToBytes(epoch))
 	if err != nil {
-		log.Errorf("error saving warmup epoch: %s", err)
+		log.Warnf("error saving warmup epoch: %s", err)
 	}
 
 	err = s.ds.Put(markSetSizeKey, int64ToBytes(s.markSetSize))
 	if err != nil {
-		log.Errorf("error saving mark set size: %s", err)
+		log.Warnf("error saving mark set size: %s", err)
 	}
+
+	return nil
 }
 
 // Compaction/GC Algorithm
@@ -540,8 +562,10 @@ func (s *SplitStore) compact(curTs *types.TipSet) {
 }
 
 func (s *SplitStore) estimateMarkSetSize(curTs *types.TipSet) error {
+	epoch := curTs.Height()
+
 	var count int64
-	err := s.chain.WalkSnapshot(context.Background(), curTs, 1, true, true,
+	err := s.walk(curTs, epoch,
 		func(cid cid.Cid) error {
 			count++
 			return nil
@@ -578,7 +602,7 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 	}
 
 	var count int64
-	err = s.chain.WalkSnapshot(context.Background(), boundaryTs, 1, true, true,
+	err = s.walk(boundaryTs, boundaryEpoch,
 		func(cid cid.Cid) error {
 			count++
 			return coldSet.Mark(cid)
@@ -592,7 +616,7 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 		s.markSetSize = count + count>>2 // overestimate a bit
 	}
 
-	log.Infow("marking done", "took", time.Since(startMark))
+	log.Infow("marking done", "took", time.Since(startMark), "marked", count)
 
 	// 2. move cold unreachable objects to the coldstore
 	log.Info("collecting cold objects")
@@ -698,6 +722,93 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 	}
 
 	return nil
+}
+
+func (s *SplitStore) walk(ts *types.TipSet, boundary abi.ChainEpoch, f func(cid.Cid) error) error {
+	walked := cid.NewSet()
+	toWalk := ts.Cids()
+
+	walkBlock := func(c cid.Cid) error {
+		if !walked.Visit(c) {
+			return nil
+		}
+
+		blk, err := s.Get(c)
+		if err != nil {
+			return xerrors.Errorf("error retrieving block (cid: %s): %w", c, err)
+		}
+
+		var hdr types.BlockHeader
+		if err := hdr.UnmarshalCBOR(bytes.NewBuffer(blk.RawData())); err != nil {
+			return xerrors.Errorf("error unmarshaling block header (cid: %s): %w", c, err)
+		}
+
+		// don't walk under the boundary
+		if hdr.Height < boundary {
+			return nil
+		}
+
+		if err := f(c); err != nil {
+			return err
+		}
+
+		if err := s.walkLinks(hdr.Messages, walked, f); err != nil {
+			return xerrors.Errorf("error walking messages (cid: %s): %w", hdr.Messages, err)
+		}
+
+		if err := s.walkLinks(hdr.ParentStateRoot, walked, f); err != nil {
+			return xerrors.Errorf("error walking state root (cid: %s): %w", hdr.ParentStateRoot, err)
+		}
+
+		toWalk = append(toWalk, hdr.Parents...)
+		return nil
+	}
+
+	for len(toWalk) > 0 {
+		walking := toWalk
+		toWalk = nil
+		for _, c := range walking {
+			if err := walkBlock(c); err != nil {
+				return xerrors.Errorf("error walking block (cid: %s): %w", c, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (s *SplitStore) walkLinks(c cid.Cid, walked *cid.Set, f func(cid.Cid) error) error {
+	if !walked.Visit(c) {
+		return nil
+	}
+
+	if c.Prefix().Codec != cid.DagCBOR {
+		return nil
+	}
+
+	if err := f(c); err != nil {
+		return err
+	}
+
+	blk, err := s.Get(c)
+	if err != nil {
+		return xerrors.Errorf("error retrieving linked block (cid: %s): %w", c, err)
+	}
+
+	var rerr error
+	err = cbg.ScanForLinks(bytes.NewReader(blk.RawData()), func(c cid.Cid) {
+		if rerr != nil {
+			return
+		}
+
+		rerr = s.walkLinks(c, walked, f)
+	})
+
+	if err != nil {
+		return xerrors.Errorf("error scanning links (cid: %s): %w", c, err)
+	}
+
+	return rerr
 }
 
 func (s *SplitStore) moveColdBlocks(cold []cid.Cid) error {
--- a/blockstore/splitstore/splitstore.go
+++ b/blockstore/splitstore/splitstore.go
@@ -310,7 +310,11 @@ func (s *SplitStore) View(cid cid.Cid, cb func([]byte) error) error {
 	err := s.hot.View(cid, cb)
 	switch err {
 	case bstore.ErrNotFound:
-		return s.cold.View(cid, cb)
+		err = s.cold.View(cid, cb)
+		if err == nil {
+			stats.Record(context.Background(), metrics.SplitstoreMiss.M(1))
+		}
+		return err
 
 	default:
 		return err
--- a/blockstore/splitstore/splitstore.go
+++ b/blockstore/splitstore/splitstore.go
@@ -86,17 +86,6 @@ type Config struct {
 	//
 	// Supported values are: "bloom" (default if omitted), "bolt".
 	MarkSetType string
-	// perform full reachability analysis (expensive) for compaction
-	// You should enable this option if you plan to use the splitstore without a backing coldstore
-	EnableFullCompaction bool
-	// EXPERIMENTAL enable pruning of unreachable objects.
-	// This has not been sufficiently tested yet; only enable if you know what you are doing.
-	// Only applies if you enable full compaction.
-	EnableGC bool
-	// full archival nodes should enable this if EnableFullCompaction is enabled
-	// do NOT enable this if you synced from a snapshot.
-	// Only applies if you enabled full compaction
-	Archival bool
 }
 
 // ChainAccessor allows the Splitstore to access the chain. It will most likely
@@ -113,16 +102,10 @@ type SplitStore struct {
 	critsection int32 // compaction critical section
 	closing     int32 // the split store is closing
 
-	fullCompaction  bool
-	enableGC        bool
-	skipOldMsgs     bool
-	skipMsgReceipts bool
-
 	baseEpoch   abi.ChainEpoch
 	warmupEpoch abi.ChainEpoch
 
 	coldPurgeSize int
-	deadPurgeSize int
 
 	mx    sync.Mutex
 	curTs *types.TipSet
@@ -165,16 +148,7 @@ func Open(path string, ds dstore.Datastore, hot, cold bstore.Blockstore, cfg *Co
 		tracker: tracker,
 		env:     env,
 
-		fullCompaction:  cfg.EnableFullCompaction,
-		enableGC:        cfg.EnableGC,
-		skipOldMsgs:     !(cfg.EnableFullCompaction && cfg.Archival),
-		skipMsgReceipts: !(cfg.EnableFullCompaction && cfg.Archival),
-
 		coldPurgeSize: defaultColdPurgeSize,
-	}
-
-	if cfg.EnableGC {
-		ss.deadPurgeSize = defaultDeadPurgeSize
 	}
 
 	return ss, nil
@@ -465,7 +439,7 @@ func (s *SplitStore) warmup(curTs *types.TipSet) {
 	batchSnoop := make([]cid.Cid, 0, batchSize)
 
 	count := int64(0)
-	err := s.chain.WalkSnapshot(context.Background(), curTs, 1, s.skipOldMsgs, s.skipMsgReceipts,
+	err := s.chain.WalkSnapshot(context.Background(), curTs, 1, true, true,
 		func(cid cid.Cid) error {
 			count++
 
@@ -556,11 +530,7 @@ func (s *SplitStore) compact(curTs *types.TipSet) {
 	}
 
 	start := time.Now()
-	if s.fullCompaction {
-		err = s.compactFull(curTs)
-	} else {
-		err = s.compactSimple(curTs)
-	}
+	err = s.doCompact(curTs)
 	took := time.Since(start).Milliseconds()
 	stats.Record(context.Background(), metrics.SplitstoreCompactionTimeSeconds.M(float64(took)/1e3))
 
@@ -571,7 +541,7 @@ func (s *SplitStore) compact(curTs *types.TipSet) {
 
 func (s *SplitStore) estimateMarkSetSize(curTs *types.TipSet) error {
 	var count int64
-	err := s.chain.WalkSnapshot(context.Background(), curTs, 1, s.skipOldMsgs, s.skipMsgReceipts,
+	err := s.chain.WalkSnapshot(context.Background(), curTs, 1, true, true,
 		func(cid cid.Cid) error {
 			count++
 			return nil
@@ -585,12 +555,12 @@ func (s *SplitStore) estimateMarkSetSize(curTs *types.TipSet) error {
 	return nil
 }
 
-func (s *SplitStore) compactSimple(curTs *types.TipSet) error {
+func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 	coldEpoch := s.baseEpoch + CompactionCold
 	currentEpoch := curTs.Height()
 	boundaryEpoch := currentEpoch - CompactionBoundary
 
-	log.Infow("running simple compaction", "currentEpoch", currentEpoch, "baseEpoch", s.baseEpoch, "coldEpoch", coldEpoch, "boundaryEpoch", boundaryEpoch)
+	log.Infow("running compaction", "currentEpoch", currentEpoch, "baseEpoch", s.baseEpoch, "coldEpoch", coldEpoch, "boundaryEpoch", boundaryEpoch)
 
 	coldSet, err := s.env.Create("cold", s.markSetSize)
 	if err != nil {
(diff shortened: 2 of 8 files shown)


vyzo:
rebased on master.

vyzo:
rebased on master.

vyzo:
TL;DR: I have marked https://github.com/filecoin-project/lotus/pull/6474 as ready for review; SpitStore v1 is considered ready for the prying eyes of @raulk, @stebalien, and (later on) @Magik6k

This PR is ready after exactly 21 days in draft; it ended with a hacking spree during the weekend, when the core of the splitstore was rewritten to ditch the ill-fated tracking store.
We have also gained the ability to run without a coldstore, using a configuration option.

The SplitStore is now free-standing and works based solely on the chain.
Compaction works transactionally with the following algorithm:
- We prepare a transaction, whereby all i/o referenced objects through the API are tracked.
- We walk the chain and mark reachable objects, keeping 4 finalities of state roots and messages and all headers all the way to genesis.
- Once the chain walk is complete, we begin full transaction protection with concurrent marking; we walk and mark all references created during the chain walk. On the same time, all I/O through the API concurrently marks objects as live references.
- We collect cold objects by iterating through the hotstore and checking the mark set; if an object is not marked, then it is candidate for purge.
- When running with a coldstore, we next copy all cold objects to the coldstore.
- At this point we are ready to begin purging:
  - We sort cold objects heaviest first, so as to never delete the consituents of a DAG before the DAG itself (which would leave dangling references)
  - We delete in small batches taking a lock; each batch is checked again for marks, from the concurrent transactional mark, so as to never delete anything live
- We then end the transaction and compact/gc the hotstore.

Overall I am quite happy with this direction, and it is definitely inching closer to production ready. We will have to test the shit out of it of course, but a future where the splitstore is the default is not so far away.

Next Steps:
- Implement online garbage collection for the coldstore; see https://github.com/filecoin-project/lotus/issues/6577
- Implement a suitable trait/optional interface so that the exposed blockstore abstraction can make API calls that don't pin references as live even though they are coming through the network. This will ensure that network requests cannot cause us to keep unwanted objects hot.

Big Win:
- The need for https://github.com/protocol/web3-dev-team/pull/120 is now obviated! We don't need to track write epochs any more, references use concurrent marking that short-circuits, and the last remaining item, network-request awareness can be implemented in an alternative manner.

vyzo:
Fixed some issues with sorting.
Specifically, it was broken because cids were coming out of the blockstore and were thus raw, so the DAG check was short-circuiting.
Furthermore, computing weights by generic walking is ridiculously inefficient, so I wrote a specialized function for computing weights (after a few iterations).

vyzo:
Did some clean up and housekeeping while awaiting review.

vyzo:
Moved transaction preparation before starting the compaction goroutine, so as subsequent state computations are certain to be included in the transaction.

vyzo:
And also added transactional protection for applied tipsets during compaction.

vyzo:
Did some more clean up, we don't actually need to store the head!
Also renamed some txn related methods for better readbility.

node/builder.go
```diff
@@ -643,6 +643,10 @@ func Repo(r repo.Repo) Option {
 			Override(new(dtypes.UniversalBlockstore), modules.UniversalBlockstore),
 
 			If(cfg.EnableSplitstore,
+				If(cfg.Splitstore.ColdStoreType == "universal",
+					Override(new(dtypes.ColdBlockstore), From(new(dtypes.UniversalBlockstore)))),
```
raulk, line 647: I find this naming confusing here. The term `universal` was used to represent that the store contains both categories of data (chain or state data), not the technology choice. We might need to revisit how these things are wired at the "LockedRepo" level. Maybe we're doing too much at DI and we could offload the choice to the repo itself?

blockstore/splitstore/splitstore.go
```diff
@@ -191,27 +197,49 @@ func (s *SplitStore) DeleteMany(_ []cid.Cid) error {
...
 func (s *SplitStore) Get(cid cid.Cid) (blocks.Block, error) {
+	s.txnLk.RLock()
+	defer s.txnLk.RUnlock()
+
 	blk, err := s.hot.Get(cid)
 
 	switch err {
 	case nil:
-		return blk, nil
+		err = s.trackTxnRef(cid)
```
raulk, line 226: Are we sure we want to propagate errors from the tx tracking, even if the block was accessed correctly?

blockstore/splitstore/splitstore.go
```diff
@@ -569,110 +766,189 @@ func (s *SplitStore) compact(curTs *types.TipSet) {
...
+
+					// we also short-circuit on old blocks, as these can come from a network request
+					// and cause us to fail because we have purged its consistituents (or missing from
+					// the beginning in case of snapshot sync, e.g. parent message receipts or old messages)
+					// if these blocks are on our chain, they would have been marked but they might be
+					// from a fork.
+					//
+					// Ideally, we would have API options to preclude us from trcking references to such
+					// objects, but we don't so we have to do this check
+					isOldBlock, err := s.isOldBlockHeader(c, lookbackEpoch)
```
raulk, line 854: This is OK for now but it's definitely a hack. I'm not sure we want API options: they're going to be close to impossible to apply correctly at all relevant places (e.g. imagine a StateCompute JSON-RPC call for a historical tipset -- we would need to percolate the "no tracking" option down to the blockstore used by the VM...).

Instead, I'd suggest using the `ExposedBlockstore` DI wiring here (which we need to extend to the JSON-RPC layer).

To avoid dirtying the API with options, one idea is to add a `func (s *Splitstore) WithTrackingOmitted() blockstore.Blockstore)` method that returns the union of the hot and cold stores, without any tracking logic.

We can then set this as the `ExposedBlockstore`, and off we go.

raulk (commented):
I continue reviewing, releasing some feedback.

blockstore/splitstore/splitstore.go
vyzo: Agreed and I certainly don't want to have this code here permanently.
I fully agree that doing it through the exposed blockstore is the right approach, and the simplified view sounds like a good approach.

Plan is to do this in a subsequent pr.

blockstore/splitstore/splitstore.go
vyzo: I am not sure either, we will probably be fine with just warning.

node/builder.go
vyzo: you named those things :)

blockstore/splitstore/splitstore.go
vyzo: Going further, I think we can do this with a trait that defines `Expose() Blockstore`.
The exposed blockstore implementation wont have to know about splitstore then and just check if it implements the trait.

raulk (commented):
More comments as I keep reviewing.

node/builder.go
raulk: Yeah, `UniversalBlockstore` makes sense as the blockstore that the system uses as a whole, but not as a backing implementation for one of the splitstore stores.

raulk: Anyway, a lot of things are messy in Lotus' DI, so I don't think getting this better structured is a blocker for now.

blockstore/splitstore/debug.go
```diff
@@ -0,0 +1,360 @@
...
+	"time"
+
+	"go.uber.org/multierr"
+	"golang.org/x/xerrors"
+
+	blocks "github.com/ipfs/go-block-format"
+	cid "github.com/ipfs/go-cid"
+)
+
+type debugLog struct {
```
raulk, line 22: This is a beautiful piece of code ;-)

blockstore/splitstore/markset.go
```diff
@@ -28,8 +28,14 @@ type MarkSetEnv interface {
...
-	case "", "bloom":
-		return NewBloomMarkSetEnv()
+	case "bloom":
+		return NewBloomMarkSetEnv(false)
+	case "bloomts":
+		return NewBloomMarkSetEnv(true)
+	case "map":
+		return NewMapMarkSetEnv(false)
+	case "mapts":
+		return NewMapMarkSetEnv(true)
```
raulk, line 38: Can we document that `ts` stands for "thread-safe"? I was trying to find the association with Tipsets all this time!

blockstore/splitstore/markset_bolt.go
```diff
@@ -1,6 +1,7 @@
 package splitstore
```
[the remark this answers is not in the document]
vyzo: yeah, I've been thinking about it; besides it's pretty slow.

blockstore/noop.go
```diff
@@ -0,0 +1,66 @@
...
+	cid "github.com/ipfs/go-cid"
+)
+
+var _ Blockstore = (*noopstore)(nil)
+
+type noopstore struct {
+	bs Blockstore
+}
+
+func NewNoopStore(bs Blockstore) Blockstore {
```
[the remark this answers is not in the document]
vyzo: dunno, the blockstore passed in doesn't necessary implement it.

blockstore/splitstore/markset_map.go
```diff
@@ -0,0 +1,61 @@
...
+		s.mx.Lock()
+		defer s.mx.Unlock()
+	}
+
+	_, ok := s.set[string(cid.Hash())]
+	return ok, nil
+}
+
+func (s *MapMarkSet) Close() error {
+	return nil
```
raulk, line 60: Probably want to make the set nil (optionally under the lock), so it can be reclaimed.

blockstore/splitstore/markset_bloom.go
```diff
@@ -64,12 +67,20 @@ func (s *BloomMarkSet) saltedKey(cid cid.Cid) []byte {
...
 
 func (s *BloomMarkSet) Has(cid cid.Cid) (bool, error) {
-	return s.bf.Has(s.saltedKey(cid)), nil
+	if s.ts {
+		return s.bf.HasTS(s.saltedKey(cid)), nil
+	}
+	return s.bf.HasTS(s.saltedKey(cid)), nil
 }
 
 func (s *BloomMarkSet) Close() error {
```
raulk, line 86: Probably want to make the bloom nil (optionally under the lock), so it can be reclaimed.

node/builder.go
vyzo: punts aside, I don't think strongly about this. I am fine with doing it this way for now, and we can revisit later.

raulk (commented):
Need to do one more pass, and we'd be good to go on my end ;-)

blockstore/splitstore/markset.go
vyzo: haha, ok will do!

blockstore/splitstore/markset_map.go
vyzo: yeah, good point.

blockstore/splitstore/markset_bloom.go
vyzo: sure, will do.

blockstore/splitstore/splitstore.go
vyzo: done, turned them to warnings and stopped leaking them through the API.

blockstore/splitstore/markset_map.go
vyzo: done.

blockstore/splitstore/markset_bloom.go
vyzo: done.

blockstore/splitstore/markset.go
vyzo: done.

blockstore/splitstore/markset_bolt.go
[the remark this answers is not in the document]
vyzo: DONE!

raulk (commented):
This looks pretty good ❤️ 

It's also a large PR to study, and there may be latent bugs and edge cases, so I think we should keep in experimental state for a while.

@vyzo One area that I'm concerned about is with clean/graceful shutdown. We don't check for a closure during lengthy processes (recursive marking, purging, etc.) I think we should check the `closing` flag more frequently and yield sooner when it's true.

blockstore/splitstore/splitstore.go
```diff
@@ -730,330 +1011,534 @@ func (s *SplitStore) compactSimple(curTs *types.TipSet) error {
...
 		}
 
-		// remove the tracker tracking
-		startPurge := time.Now()
-		log.Info("purging dead objects from tracker")
-		err = s.purgeTracking(dead)
+		err := deleteBatch(cids[start:end])
 		if err != nil {
-			return xerrors.Errorf("error purging tracking for dead blocks: %w", err)
+			return xerrors.Errorf("error deleting batch: %w", err)
```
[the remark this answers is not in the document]
vyzo: we could, but unfortunately it's unsafe; we can break our promise of never leaving a danglint reference.

blockstore/splitstore/splitstore.go
```diff
@@ -730,330 +1011,534 @@ func (s *SplitStore) compactSimple(curTs *types.TipSet) error {
...
-	// 3. if we have dead objects, delete them from the hotstore and remove the tracking
-	if len(dead) > 0 {
-		log.Info("deleting dead objects")
-		err = s.purgeBlocks(dead)
-		if err != nil {
-			return xerrors.Errorf("error purging dead blocks: %w", err)
+	// we don't delete one giant batch of millions of objects, but rather do smaller batches
+	// so that we don't stop the world for an extended period of time
+	done := false
+	for i := 0; !done; i++ {
```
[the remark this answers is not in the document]
vyzo: certainly, will add a check.

blockstore/splitstore/splitstore.go
```diff
@@ -730,330 +1011,534 @@ func (s *SplitStore) compactSimple(curTs *types.TipSet) error {
...
+}
+
+// I really don't like having this code, but we seem to have some occasional DAG references with
+// missing constituents. During testing in mainnet *some* of these references *sometimes* appeared
+// after a little bit.
+// We need to figure out where they are coming from and eliminate that vector, but until then we
+// have this gem[TM].
+// My best guess is that they are parent message receipts or yet to be computed state roots; magik
+// thinks the cause may be block validation.
+func (s *SplitStore) waitForMissingRefs() {
```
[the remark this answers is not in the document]
vyzo: yeah, I don't like seeing warnings in my logs. So far they seem to be inocuous, but we we ant to be absolutely certain.

blockstore/splitstore/splitstore.go
```diff
@@ -730,330 +1011,534 @@ func (s *SplitStore) compactSimple(curTs *types.TipSet) error {
...
-	}
+	return nil
+}
 
-	if coldCnt > 0 {
-		s.coldPurgeSize = coldCnt + coldCnt>>2 // overestimate a bit
-	}
-	if deadCnt > 0 {
-		s.deadPurgeSize = deadCnt + deadCnt>>2 // overestimate a bit
+func (s *SplitStore) getObjectWeight(c cid.Cid, weights map[string]int, key func(cid.Cid) string) int {
```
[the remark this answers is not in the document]
vyzo: I don't think so, it's a DAG after all!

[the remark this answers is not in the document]
vyzo: But certainly worth asking the question!

vyzo:
> One area that I'm concerned about is with clean/graceful shutdown. We don't check for a closure during lengthy processes (recursive marking, purging, etc.) I think we should check the closing flag more frequently and yield sooner when it's true.

We only need to do it during the critical section, otherwise `Close` doesn't block.

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: done.

blockstore/splitstore/markset_map.go
vyzo: this is actually unsafe, as a splitstore Close would lead us to panic; i have reverted.

blockstore/splitstore/markset_bloom.go
vyzo: ditto here, actually unsafe so I have reverted.

raulk (approved):
Leaving my projected LGTM here -- OK from me to merge once you feel good about addressing the comments!

blockstore/splitstore/markset_map.go
vyzo: after some thought, re-added the niling but I also added explicit nil checks and error returns.

blockstore/splitstore/markset_bloom.go
vyzo: readded with explicit nil checks, see above.

vyzo:
A couple of changes per raul's comments:
- Sprinkled some frequent closure checks for graceful shutdown in lengthy processes
- Made marksets more intelligent and utilized the error returns to immediately release memory on shutdown and preclude further operations.

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: Gave it some more thought, I think it is sound for DAGs.
It would be nice if we could find something faster of course.

blockstore/noop.go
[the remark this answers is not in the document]
vyzo: I don't think we should add it actually; it implies efficient iterator support and any implementation would have to fall back to `AllKeysChan` if the trait is not implemented.
So we might as well simply not implement it.

blockstore/badger/blockstore.go
```diff
@@ -442,6 +443,55 @@ func (b *Blockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 	return ch, nil
 }
 
+// Implementation of BlockstoreIterator interface
+func (b *Blockstore) ForEachKey(f func(cid.Cid) error) error {
```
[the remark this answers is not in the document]
vyzo: Yeah, it doesn't seem to block anything from happening.

vyzo:
I will add a README with some documentation and then invoke MAGIK!

blockstore/splitstore/splitstore.go
vyzo: Follow up issue: https://github.com/filecoin-project/lotus/issues/6705

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: Follow up issue: https://github.com/filecoin-project/lotus/issues/6706

vyzo:
Added README and opened follow up issues (#6705, #6706); it's time to use the wand and summon @magik6k

blockstore/splitstore/splitstore.go
```diff
@@ -569,155 +793,235 @@ func (s *SplitStore) compact(curTs *types.TipSet) {
...
-
-		// check whether it is reachable in the cold boundary
-		mark, err := coldSet.Has(cid)
+	cold := make([]cid.Cid, 0, s.coldPurgeSize)
+	err = s.hot.(bstore.BlockstoreIterator).ForEachKey(func(c cid.Cid) error {
+		// was it marked?
+		mark, err := markSet.Has(c)
 		if err != nil {
-			return xerrors.Errorf("error checkiing cold set for %s: %w", cid, err)
+			return xerrors.Errorf("error checkiing mark set for %s: %w", c, err)
```
magik6k, line 937:
```suggestion
			return xerrors.Errorf("error checking mark set for %s: %w", c, err)
```

magik6k (commented):
Mostly just a couple of typos (and some questions) - generally looks good, but this definitely is the kind of thing that deserves lots of real-world testing

blockstore/splitstore/splitstore.go
```diff
@@ -67,36 +73,34 @@ var (
...
 
 	log = logging.Logger("splitstore")
+
+	// used to signal end of walk
+	errStopWalk = errors.New("stop walk")
+
+	// set this to true if you are debugging the splitstore to enable debug logging
+	enableDebugLog = false
+	// set this to true if you want to track origin stack traces in the write log
+	enableDebugLogWriteTraces = false
```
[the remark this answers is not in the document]
vyzo: yeah, that's definitely a good idea, but it will also mean that we might miss older writes.

I'll add the env vars.

blockstore/splitstore/splitstore.go
```diff
@@ -322,10 +376,34 @@ func (s *SplitStore) HashOnRead(enabled bool) {
 }
 
 func (s *SplitStore) View(cid cid.Cid, cb func([]byte) error) error {
+	s.txnLk.RLock()
+	defer s.txnLk.RUnlock()
+
 	err := s.hot.View(cid, cb)
 	switch err {
+	case nil:
+		err = s.trackTxnRef(cid)
```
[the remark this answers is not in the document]
vyzo: I don't think it matters, we hold the RLock anyway.

blockstore/splitstore/splitstore.go
```diff
@@ -322,10 +376,34 @@ func (s *SplitStore) HashOnRead(enabled bool) {
 }
 
 func (s *SplitStore) View(cid cid.Cid, cb func([]byte) error) error {
+	s.txnLk.RLock()
+	defer s.txnLk.RUnlock()
+
 	err := s.hot.View(cid, cb)
```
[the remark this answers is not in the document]
vyzo: OUCH, that's a very good observation; it sucks that those things aren't recursive.
We might have to write our own RWLock .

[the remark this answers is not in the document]
vyzo: Let me write a quick program to verify the problem.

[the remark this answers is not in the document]
vyzo: Yep verified the deadlock scenario; this totally sucks.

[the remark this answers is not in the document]
vyzo: implementation of custom txnlock that allows recursive RLocking: https://gist.github.com/vyzo/b1f6e2922270f6dfcd410a6bb06a88e5

If we actually need it, we can use it; for now let's rely on our usage patterns to not deadlock.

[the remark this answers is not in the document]
vyzo: Following sync discussion, the _right_ solution is to protect the references optimistically _before_ calling `s.host.View` and do the latter without holding the RLock.
This will work even if the block is missing, because protection uses `walkObjectIncomplete` which occurs checks and handles missing references.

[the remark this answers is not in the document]
vyzo: implemented in https://github.com/filecoin-project/lotus/pull/6474/commits/60dd97c7fc5f8f114ebebf9fe78cbb39b6672e76

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: See https://github.com/filecoin-project/lotus/pull/6474/commits/60dd97c7fc5f8f114ebebf9fe78cbb39b6672e76 -- we now track the reference optimistically before calling view.

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: done.

vyzo:
@magik6k ready for another round of review.

And definitely agreed on the real world testing -- we need to test the shit out of it.

vyzo:
Inspiration while sleeping; thread-safety with RWMutexes had noticeable impact in initial marking time; and the "ts" suffixes confused @raulk.

So I removed the distinction and made thread-safety a property is turned on on demand, when it's time to start concurrent marking, with `SetConcurrent`.

blockstore/noop.go
```diff
@@ -0,0 +1,66 @@
...
+	"context"
+	"io"
+
+	blocks "github.com/ipfs/go-block-format"
+	cid "github.com/ipfs/go-cid"
+)
+
+var _ Blockstore = (*noopstore)(nil)
+
+type noopstore struct {
```
Stebalien, line 13: nit: discardstore? It's not exactly a no-op.

blockstore/splitstore/README.md
```diff
@@ -0,0 +1,66 @@
...
+regular badger blockstore (the default), which can be periodically
+garbage collected according to configurable user retention policies.
+
+To enable the splitstore, edit `.lotus/config.toml` and add the following:
+```
+[Chainstore]
+  EnableSplitstore = true
+```
+
+If you want to use the noop coldstore, also add the following:
```
Stebalien, line 23: Can we change this to reflect intent?

1. The user doesn't care about what happens under the covers.
2. If the option specifies _intent_, we can _change_ what happens under the covers without messing with the options.

blockstore/splitstore/debug.go
Stebalien: Maybe a bit less beautiful?

```go
type debugLogOp struct {
    path string

    mu     sync.Mutex
    count  int64
}

type debugLog struct {
    read, write, delete, stack debugLogOp
    stackMap                   map[string]string
}
```

(and then DRY up the constructor as well)

blockstore/splitstore/debug.go
```diff
@@ -0,0 +1,360 @@
...
+	d.stackMx.Lock()
+	err4 := d.stackLog.Close()
+	d.stackMx.Unlock()
+
+	return multierr.Combine(err1, err2, err3, err4)
+}
+
+func (d *debugLog) getStack() string {
+	sk := d.getNormalizedStackTrace()
+	hash := sha256.Sum256([]byte(sk))
```
Stebalien, line 317: Do we _really_ need a sha256 hash here? Just a _tad_ slow. Just a bit.

(probably not critical)

blockstore/splitstore/debug.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	arxivPath := fmt.Sprintf("%s-%d", path, time.Now().Unix())
+	err = os.Rename(path, arxivPath)
+	if err != nil {
+		return xerrors.Errorf("error moving file: %w", err)
+	}
+
+	go func() {
+		cmd := exec.Command("gzip", arxivPath)
```
Stebalien, line 281: https://golang.org/pkg/compress/gzip/?

blockstore/splitstore/debug.go
Stebalien: Would also let you delete a bunch of these functions. You could even have a "forAll" helper.

Stebalien (changes requested):
Getting some of the small stuff out of the way.

blockstore/blockstore.go
```diff
@@ -30,6 +30,11 @@ type BatchDeleter interface {
 	DeleteMany(cids []cid.Cid) error
 }
 
+// BlockstoreIterator is a trait for efficient iteration
+type BlockstoreIterator interface {
+	ForEachKey(func(cid.Cid) error) error
```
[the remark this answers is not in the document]
vyzo: Probably like 2x?

blockstore/badger/blockstore.go
```diff
@@ -442,6 +443,55 @@ func (b *Blockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
...
+	if atomic.LoadInt64(&b.state) != stateOpen {
```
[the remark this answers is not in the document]
vyzo: Yeah, I can see how this can be problematic; I'll add a RWMutex.

blockstore/noop.go
vyzo: Dunno, I am not married to the name. If you think it's better, then we can rename :)

blockstore/splitstore/README.md
vyzo: Ok. Any suggestions for wording?

blockstore/splitstore/debug.go
vyzo: No, even md5 will do -- but in the grand scheme of things we are taking a stack trace, which is a lot slower than the hash!

blockstore/splitstore/debug.go
vyzo: it is less code this way, but sure we can do it in process.

blockstore/splitstore/debug.go
vyzo: sure we can do that. I didn't put that much thought in code cosmetics here, it was purely pragmatic while debugging the misses.

blockstore/badger/blockstore.go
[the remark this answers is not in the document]
vyzo: damn, we have the View deadlock problem there too!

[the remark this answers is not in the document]
vyzo: Fixed in https://github.com/filecoin-project/lotus/pull/6474/commits/909f7039d47c351c2581de0c39bb1c9b5d8a721d

blockstore/blockstore.go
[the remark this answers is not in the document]
Stebalien: Ah, probably because you don't have a buffer on that channel which forces context switches. They're faster in go, but not _that_ fast.

I'm not really _against_ this version, but internal iteration has some... issues as we've noticed.

blockstore/noop.go
Stebalien: I don't _really_ care. I'd just expect a noop store to always do nothing (i.e., never return anything)

blockstore/splitstore/README.md
Stebalien: I'm not entirely sure why a user might want to do this, but I _assume_ it's because they're going to put a snapshot there but don't want to keep old state otherwise? Unclear?

blockstore/splitstore/debug.go
Stebalien: Yeah, probably.

blockstore/noop.go
vyzo: Yeah, agreed. I'll call it `discardstore` and change the config option name to "discard".

blockstore/splitstore/debug.go
Stebalien: You know, it's debug code. Up to you. I just cringe at external deps because they tend to lead to random bug reports.

blockstore/blockstore.go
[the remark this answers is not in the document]
vyzo: See https://github.com/filecoin-project/lotus/commit/909f7039d47c351c2581de0c39bb1c9b5d8a721d -- finally solved our issues :)

blockstore/splitstore/debug.go
vyzo: Yeah, it's immaterial. If a user doesn't have gzip in his system... what can I say? He'll have to live with some garbage during debugging.

Regardless, If I don't have anything better to do, I will turn it to internal code.

blockstore/splitstore/README.md
vyzo: If they want to run with not much of a disk, it is very handy; it is also critical for the boosters.

blockstore/splitstore/splitstore.go
vyzo: So I killed the `isBlockHeader` hack in https://github.com/filecoin-project/lotus/pull/6474/commits/4f89d260b05e7da546b82b7e0e2f42e50851c3a3
It was moribund code anyway, and we ought to resolve #6705 before shipping v1.11.1

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: Actually it is little tricker than that, the View can theoretically be so long that it starts before a compaction starts and be active when the purge begins... which could be catastrophic.

I'll add some code to track views when a transaction is not active and block starting a new transaction while there are active views.
We can do that with a waitgroup.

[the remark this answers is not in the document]
vyzo: Implemented in https://github.com/filecoin-project/lotus/pull/6474/commits/acc4c374ef4226f953cfedf93121ca711f70cc4c

I think we can call the issue resolved.

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: See https://github.com/filecoin-project/lotus/commit/acc4c374ef4226f953cfedf93121ca711f70cc4c which fixes some issues with the previous attempt.

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: Well, add this to the pile: https://github.com/filecoin-project/lotus/pull/6474/commits/095d7427bab6063aba1f31f6cb44acb2e5ade6e9

I am leaving the conversation unresolved until we are all happy.

blockstore/noop.go
vyzo: Done in https://github.com/filecoin-project/lotus/pull/6474/commits/c0a1cfffa1f1fbea767af4f214fbefde14187622

blockstore/splitstore/README.md
vyzo: Updated the wording in https://github.com/filecoin-project/lotus/pull/6474/commits/b9a5ea8f7bed48942e85dbf0fae0976402bb4f6b

Does it look better?

blockstore/splitstore/debug.go
vyzo: refactored to DRY in https://github.com/filecoin-project/lotus/pull/6474/commits/f5ae10e3d1b8447f814c7c44082a390cbc4a094d

vyzo:
Follow up issue for testing: https://github.com/filecoin-project/lotus/issues/6725

Stebalien (approved):
Please address my final _final_ review in https://github.com/filecoin-project/lotus/pull/6718 before merging. But I'm going to give this a :+1: anyways because the change shouldn't be too difficult.

vyzo:
Addressed the final _final_ review issue in https://github.com/filecoin-project/lotus/pull/6474/commits/257423e917ffb929907f2309df77c2ed3f62f32a, and finetuned in https://github.com/filecoin-project/lotus/pull/6474/commits/af399529ecca4c6e082474cada76c9b96a3d8f86

vyzo:
Summoning @magik6k -- this is ready for you!

magik6k (approved):
Can't spot anything to be nitpicky about, I guess this means we can see if it floats.

blockstore/splitstore/splitstore.go
```diff
@@ -458,266 +616,528 @@ func (s *SplitStore) HeadChange(_, apply []*types.TipSet) error {
...
+func (s *SplitStore) viewDone() {
+	s.txnViewsMx.Lock()
+	defer s.txnViewsMx.Unlock()
 
-	if err != nil {
-		log.Errorf("error warming up splitstore: %s", err)
-		return
+	s.txnViews--
+	if s.txnViews == 0 && s.txnViewsWaiting {
+		s.txnViewsCond.Signal()
```
Stebalien, line 656: There's no reason to use `Signal` instead of `Broadcast` here. It _shouldn't_ matter, but think of `Signal` as an optimization when you _only_ want to wake up one waiter (when you expect multiple).

Stebalien: Not critical... but still.

blockstore/splitstore/splitstore.go
```diff
@@ -458,266 +616,528 @@ func (s *SplitStore) HeadChange(_, apply []*types.TipSet) error {
...
-	}
-	defer coldSet.Close() //nolint:errcheck
+		if s.txnProtect != nil {
+			mark, err := s.txnProtect.Has(c)
+			if err != nil {
+				if !quiet {
+					quiet = true
+					log.Warnf("error checking markset: %s", err)
+				}
+				continue
```
Stebalien, line 720: This is still inconsistent with trackTxnRef.

blockstore/splitstore/splitstore.go
```diff
@@ -458,266 +616,528 @@ func (s *SplitStore) HeadChange(_, apply []*types.TipSet) error {
...
+	s.txnViewsMx.Lock()
+	defer s.txnViewsMx.Unlock()
 
-	if count > s.markSetSize {
-		s.markSetSize = count + count>>2 // overestimate a bit
+	s.txnViewsWaiting = true
+	for s.txnViews > 0 {
+		s.txnViewsCond.Wait()
 	}
+	s.txnViewsWaiting = false
```
Stebalien, line 668: So, we're waiting for all the views to end, but then we're not _doing_ anything atomically. That can't be right.

Should we be waiting for all the non-tracked views to end? I.e., should we have a return at https://github.com/filecoin-project/lotus/pull/6474/files#diff-eac9e730a0594047de6e81aa421dcd33aac194e505a18945ca45e72db789687cR642?

blockstore/splitstore/splitstore.go
vyzo: ok, I'll fix it in the open gc pr to use Broadcast.

blockstore/splitstore/splitstore.go
vyzo: wait, what?  we have already started the transaction.
All the views are tracked now as they always increment the rlock.
The bariier just ensures that there is no view that started before the transaction that hasn't ended.

blockstore/splitstore/splitstore.go
vyzo: in what way? 
Are you concerned about an error not causing the transaction to abort?
Currently the only way the markset errors is if it has been closed, ie the transaction has been aborted.

blockstore/splitstore/splitstore.go
vyzo: fixed in https://github.com/filecoin-project/lotus/pull/6728/commits/86d7d1472d37003a63d08bb451dc70bb5a3d5f1d

blockstore/splitstore/splitstore.go
Stebalien: In trackTxnRef, we track the ref even when we error. _Here_, we skip when we error.

blockstore/splitstore/splitstore.go
Stebalien: Ok. This _should_ be fine.

Ideally, we'd track views started after the transaction starts separately, but that's not strictly necessary.

blockstore/splitstore/splitstore.go
vyzo: ah, yes you are right; let me fix it.

vyzo: fixed in https://github.com/filecoin-project/lotus/pull/6728/commits/77854670f02aff7e5d294e11b3e64e89f5c40e5a

blockstore/splitstore/splitstore.go
vyzo: Yeah.

To summarize our sync conversation: a long running view is a bug and i'd very much rather hang the compaction (which is something we'll see in the logs) than do a potentially catastrophic delete.

PR #6479, opened 2021-06-14
Test multicore SDR support

magik6k:
This test works when run individually, but broken when it runs with any other tests as proofs expect the call to FilInitLogFd to be the very first proofs call.

Probably acceptable to work around by just running this test individually in a separate step on Circle

### DIFF
--- a/extern/sector-storage/ffiwrapper/sealer_test.go
+++ b/extern/sector-storage/ffiwrapper/sealer_test.go
@@ -882,11 +882,11 @@ func setupLogger(t *testing.T) *bytes.Buffer {
 }
 
 func TestMulticoreSDR(t *testing.T) {
-	rustLogger := setupLogger(t)
-
 	if os.Getenv("TEST_RUSTPROOFS_LOGS") != "1" {
 		t.Skip("skipping test without TEST_RUSTPROOFS_LOGS=1")
 	}
+
+	rustLogger := setupLogger(t)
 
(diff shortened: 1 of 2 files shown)


BigLep:
Great bias for action here @magik6k! Thanks for finding a way to catch this early and prevent more manual effort.

arajasek (approved):
Not the most unhacky thing I've ever seen, but eh.

jennijuju (approved):

extern/sector-storage/ffiwrapper/sealer_test.go
```diff
@@ -853,3 +854,87 @@ func TestAddPiece512MPadded(t *testing.T) {
...
+	s := seal{ref: si}
+
+	// check multicore
+	_ = os.Setenv("FIL_PROOFS_USE_MULTICORE_SDR", "1")
+	rustLogger.Reset()
+	s.precommit(t, sb, si, func() {})
+
+	ok := false
+	for _, s := range strings.Split(rustLogger.String(), "\n") {
+		if strings.Contains(s, "create_label::multi") {
```
jennijuju, line 933: can this be used as a hard ✅ ? when the things was down during the weekend, a miner reported that in his log its showing multi-core however, single-core was used in reality

magik6k: Yeah, this tests that the multicore feature works. If it uses only a single core, then that's a separate bug, and I'm not sure if we have a good way to test that from Go

PR #6483, opened 2021-06-15
convert multisig and gateway tests to kit2

dirkmc:
Depends on https://github.com/filecoin-project/lotus/pull/6464


itests/kit2/node_opts.go
```diff
@@ -31,6 +32,10 @@ var DefaultNodeOpts = nodeOpts{
 	sectors: DefaultPresealsPerBootstrapMiner,
 }
 
+// OptBuilder is used to create an option after some other node is already
+// active. Takes all active nodes as a parameter.
+type OptBuilder func(activeNodes []*TestFullNode) node.Option
```
raulk, line 37: I don't think we need this. I think this is solved in the new kit because it has the ability to add nodes _after_ the ensemble has been created. So this ensemble should be created in two phases:

1. We should start the ensemble once without the full node.
2. Then we add the lite client to the ensemble, passing in a normal `ConstructorOpt` with the gateway configuration. And call `Start()` again. Unless there are bugs, this should start only the lite node.

raulk (changes requested):

itests/kit2/node_opts.go
dirkmc: Unfortunately that doesn't work, because the wallet is created at Genesis:
https://github.com/filecoin-project/lotus/blob/305e3245b68c4b30e319c143e8779c08b2c46442/itests/kit2/ensemble.go#L136-L150

So if we start the other nodes first, then try to start the lite client, we get a nil pointer on this line:
https://github.com/filecoin-project/lotus/blob/305e3245b68c4b30e319c143e8779c08b2c46442/itests/kit2/ensemble.go#L287

raulk: Aha, I'll extend the testkit. We did say that we want to be able to assign arbitrary balances at genesis.

raulk (approved):
I've updated this PR using a two instantiation stages.

PR #6484, opened 2021-06-15
refactor paych cli tests to use kit2

dirkmc:


raulk (approved):
Concurrent deals tests failing. @dirkmc is looking into those and will in the base branch.

PR #6490, opened 2021-06-15
Add utils to use multisigs as miner owners

arajasek:
Does the thing


magik6k (approved):
So much better than having to do this manually

btw, there are 2 other methods which need to be called from the owner address:
* ChangeWorkerAddress
* ConfirmUpdateWorkerKey

arajasek:
@magik6k Yeah, I did this as a one-off for a particular use-case, but I'll add those methods for completeness (eventually).

PR #6309, opened 2021-05-21
Adjust various CLI display ratios to arbitrary precision

ribasushi:
Originally the deviations from using float64 were insignificant, but at
exabyte scale they start to show up. Cleanup all displays, and clarify
the expectation text, adding an extra 99.9% probability calculator to
`lotus-miner info`

Before:
image

After:
image

/cc @Kubuxu @nikkolasg

PR #6501, opened 2021-06-16
Fix soup

arajasek:
Use latest Lotus, fixes build failure caused by #6309


ribasushi (approved):
:cryingbear:

At least now I'll know what a mysterious k8s fail actually is

PR #6507, opened 2021-06-17
refactor sector terminate test to use kit2

dirkmc:
Depends on https://github.com/filecoin-project/lotus/pull/6506


itests/kit2/chain.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/stretchr/testify/require"
+)
+
+func WaitTillChainHeight(ctx context.Context, t *testing.T, node *TestFullNode, blocktime time.Duration, height int) abi.ChainEpoch {
```
raulk, line 12: We should have this as a `TestFullNode` method. Then we can remove the `t` and `node` arguments.

raulk (changes requested):

itests/kit2/chain.go
raulk: Also, let's document the testkit. We are really trying to up the quality bar to make it easy for future test writers to use this framework!

itests/kit2/node_full.go
```diff
@@ -33,3 +36,50 @@ func (f *TestFullNode) CreateImportFile(ctx context.Context, rseed int, size int
 	require.NoError(f.t, err)
 	return res, path
 }
+
+// WaitTillChain waits until a specified chain condition is met. It returns
+// the first tipset where the condition is met.
+func (f *TestFullNode) WaitTillChain(ctx context.Context, pred ChainPredicate) *types.TipSet {
```
raulk, line 42: I introduced this method, and generalised it with a little bit of functional magic. I was also able to simplify the deals itest.

raulk (approved):

dirkmc:
Nice LGTM 👍

PR #6511, opened 2021-06-17
Refactor deadline toggling test to kit2

dirkmc:
Fails with
```
2021-06-17T14:56:17.610+0200	WARN	vm	vm/runtime.go:350	Abortf: failed to construct initial miner info: invalid partition sectors: unsupported proof type: 0
```

I'm not sure why because proof type 0 is RegisteredSealProof_StackedDrg2KiBV1, which is registered in the init method:
https://github.com/filecoin-project/lotus/blob/2c30ac394daa1706a1c75fcedeea4df90556dd95/itests/kit2/init.go#L17

### DIFF
--- a/itests/deadlines_test.go
+++ b/itests/deadlines_test.go
@@ -26,7 +26,6 @@ import (
 	miner2 "github.com/filecoin-project/specs-actors/v2/actors/builtin/miner"
 	"github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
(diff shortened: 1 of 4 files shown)


raulk (approved):
I fixed this issue in [`5548541` (#6511)](https://github.com/filecoin-project/lotus/pull/6511/commits/5548541e1a08c313c5e1d148ea739293329ced88). The problem is that non-genesis miners need to be instantiated with a _concrete_ proof type. Will watch CI and merge if all looks good.

PR #6524, opened 2021-06-18
Update to proofs v8.0.2 with correct aggregate lengths

magik6k:


extern/sector-storage/ffiwrapper/sealer_test.go
```diff
@@ -544,6 +544,7 @@ func TestSealAndVerifyAggregate(t *testing.T) {
 
 	avi.Proof, err = ProofProver.AggregateSealProofs(avi, toAggregate)
 	require.NoError(t, err)
+	require.Len(t, avi.Proof, 11188)
```
magik6k, line 547: This number is less than 11220 which we expect in the mock impl, but I'm going to assume that that's caused by the fact that 2k sectors have 10x less snarks

Kubuxu:
We might want to increment the filecoin-ffi soft version, as this does not break binary compatibility but is an important change.https://github.com/filecoin-project/lotus/blob/master/Makefile#L45

Kubuxu (approved):

arajasek (approved):

PR #6535, opened 2021-06-20
Remove specific lotus version from lotus-soup testplan

ribasushi:
This is probably a better version of https://github.com/filecoin-project/lotus/pull/6501/files#diff-19fb69201c76df493a44cc5d2eeade37a663809b8a571eb7325f4274dbdf90ddR16

@arajasek thoughts?

### DIFF
--- a/testplans/lotus-soup/go.mod
+++ b/testplans/lotus-soup/go.mod
@@ -9,11 +9,11 @@ require (
 	github.com/drand/drand v1.2.1
 	github.com/filecoin-project/go-address v0.0.5
 	github.com/filecoin-project/go-data-transfer v1.6.0
-	github.com/filecoin-project/go-fil-markets v1.4.0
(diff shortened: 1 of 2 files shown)


ribasushi:
> probably a better version

Well, based on the test failures it clearly isn't. Converting to draft in the meantime

@nonsense thoughts why this doesn't "just work" ? ( it compiles locally as expected )

nonsense:
@ribasushi at the moment Testground doesn't have support for custom `context`. We use the directory of the test plan as the context for the test plan, in this case this is `./testplans/lotus-soup`. We don't upload the `../../` alongside the test plan, so TaaS has no way of compiling and running the test plan.

ribasushi:
killing PR/branch, determined in a sync conversation this is *much* harder to tackle correctly

PR #6537, opened 2021-06-21
Test/pledge batching msg prop

### LINKED ISSUE #6352: Expand TestPledgeBatching to test to test message propagation.
We need to end-to-end test batch sector proving in a devnet to make sure we don't have any hidden issues anywhere in the pipeline (e.g., message propagation because we'll have larger messages).

This will help us catch bugs like #6338.

ZenGround0:
Closes #6352

### DIFF
--- a/itests/sector_pledge_test.go
+++ b/itests/sector_pledge_test.go
@@ -7,12 +7,17 @@ import (
 	"testing"
 	"time"
 
+	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/lotus/itests/kit"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build"
+	"github.com/filecoin-project/lotus/chain/types"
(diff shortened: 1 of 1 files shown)


itests/sector_pledge_test.go
```diff
@@ -104,6 +109,83 @@ func TestPledgeBatching(t *testing.T) {
...
+
+		// Ensure that max aggregate message has propagated to the other node by checking current state
+		sectorInfosAfter, err := client.StateMinerSectors(ctx, miner.ActorAddr, nil, types.EmptyTSK)
+		require.NoError(t, err)
+		assert.Equal(t, miner5.MaxAggregatedSectors+kit.DefaultPresealsPerBootstrapMiner, len(sectorInfosAfter))
+	}
+
+	t.Run("Force max prove commit aggregate size", func(t *testing.T) {
+		runTest(t)
+	})
```
magik6k, line 186: Could be
```suggestion
	t.Run("Force max prove commit aggregate size", runTest)
```

magik6k (commented):
Don't we want to also have another client, and check that it keeps in sync after we publish those batches?

Stebalien (approved):

PR #6542, opened 2021-06-21
extern/storage: add ability to ignore worker resources when scheduling.

raulk:
This deflakes tests on some platforms. In some cases, the dev/test workstation may be overcommitted and swapping, and the memory available check performed during scheduling may fail. On macOS this reports a MemoryReserved larger than MemoryPhysical, and Lotus fails to schedule the task.

This introduces a worker configuration option to instruct the scheduler to ignore resources when scheduling tasks. It is intended to be used only in testing.

### DIFF
--- a/extern/sector-storage/manager.go
+++ b/extern/sector-storage/manager.go
@@ -96,6 +96,10 @@ type SealerConfig struct {
 	AllowPreCommit2 bool
 	AllowCommit     bool
 	AllowUnseal     bool
+
+	// IgnoreResourceFiltering instructs the system to ignore available
+	// resources when assigning tasks to the local worker.
+	IgnoreResourceFiltering bool
 }
 
 type StorageAuth http.Header
--- a/extern/sector-storage/sched.go
+++ b/extern/sector-storage/sched.go
@@ -349,24 +349,24 @@ func (sh *scheduler) trySched() {
 	defer sh.workersLk.RUnlock()
 
 	windowsLen := len(sh.openWindows)
-	queuneLen := sh.schedQueue.Len()
+	queueLen := sh.schedQueue.Len()
 
-	log.Debugf("SCHED %d queued; %d open windows", queuneLen, windowsLen)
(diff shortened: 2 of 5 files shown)


extern/sector-storage/sched.go
```diff
@@ -349,24 +349,24 @@ func (sh *scheduler) trySched() {
 	defer sh.workersLk.RUnlock()
 
 	windowsLen := len(sh.openWindows)
-	queuneLen := sh.schedQueue.Len()
+	queueLen := sh.schedQueue.Len()
```
raulk, line 352: typo fix

extern/sector-storage/sched.go
```diff
@@ -393,7 +393,8 @@ func (sh *scheduler) trySched() {
 				}
 
 				// TODO: allow bigger windows
-				if !windows[wnd].allocated.canHandleRequest(needRes, windowRequest.worker, "schedAcceptable", worker.info.Resources) {
+				ignoringResources := worker.info.IgnoreResources
+				if !ignoringResources && !windows[wnd].allocated.canHandleRequest(needRes, windowRequest.worker, "schedAcceptable", worker.info.Resources) {
```
magik6k, line 397: Would pass full `worker.info` into `canHandleRequest` and handle this condition there

extern/sector-storage/manager.go
```diff
@@ -155,9 +169,12 @@ func New(ctx context.Context, lstor *stores.Local, stor *stores.Remote, ls store
 		localTasks = append(localTasks, sealtasks.TTUnseal)
 	}
 
-	err = m.AddWorker(ctx, NewLocalWorker(WorkerConfig{
-		TaskTypes: localTasks,
-	}, stor, lstor, si, m, wss))
+	wcfg := WorkerConfig{
+		IgnoreResourceFiltering: sc.ResourceFiltering == ResourceFilteringHardware,
```
magik6k, line 173: Not `== ResourceFilteringDisabled`?

magik6k (approved):

PR #6544, opened 2021-06-21
add docker-compose file

coryschwartz:
* fileocin/lotus
  - includes an initialization that pulls form the latest snapshot, similarly to the snap package and DO droplet
  - Runs the lotus daemon by default, which is the most common scenario, and most likely the expected behavior
* filecoin/lotus-all-in-one
  - includes all the lotus utilities you might want
* docker-compose
  - help devs get up and running quckly for local development, just run `docker-compose up -d` and away you go.
  - Has a hint about how to run a miner, but not uncommented by default since many people wouldn't want that.
  - Developers can attach to the fullnode or to the gateway to test against the limited gateway API, test locally before using public gateways.


Dockerfile.lotus
```diff
@@ -56,19 +56,230 @@ COPY --from=builder /usr/lib/x86_64-linux-gnu/libOpenCL.so.1 /lib/
...
+VOLUME /var/tmp/filecoin-proof-parameters
+
+USER fc
+
+EXPOSE 1234
 
-RUN mkdir /var/lib/lotus /var/tmp/filecoin-proof-parameters && chown fc /var/lib/lotus /var/tmp/filecoin-proof-parameters
+ENTRYPOINT ["/docker-lotus-entrypoint.sh"]
+
+CMD ["-help"]
```
travisperson, line 85: `/docker-lotus-entrypoint.sh` doesn't have a help flag

scripts/docker-lotus-entrypoint.sh
```diff
@@ -0,0 +1,24 @@
...
+	echo importing minimal snapshot
+	/usr/local/bin/lotus daemon --import-snapshot "$DOCKER_LOTUS_IMPORT_SNAPSHOT" --halt-after-import
+
+	# Block future inits
+	date > "$GATE"
+fi
+
+if [ ! -z DOCKER_LOTUS_WALLET_IMPORT ]; then
+	mkdir $LOTUS_PATH/keystore
+	cp "${DOCKER_LOTUS_WALLET_IMPORT}" "${LOTUS_PATH}/keystore"
```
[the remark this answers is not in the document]
coryschwartz: Thanks, good point.

Dockerfile.lotus
coryschwartz: > `/docker-lotus-entrypoint.sh` doesn't have a help flag

It only does because it exec's to lotus at the end.

travisperson: oh nice

BigLep:
@coryschwartz : per updated Lotus CR policy, can you get an approval on your team, and then merge if maintainers haven't replied within 2 business days?  Ping me on Slack if you need more info here.

coryschwartz:
Thanks @BigLep

@travisperson let me know if you see any problems here, I had this sitting in draft for a while so it's already pretty ancient.

travisperson (approved):

PR #6563, opened 2021-06-22
Always flush when timer goes off

### LINKED ISSUE #6541: Send commits using ProveCommitSectors when time configs hit
when [cfg.CommitBatchWait is hit and # of sector is < minbatch ](https://github.com/filecoin-project/lotus/blob/715176698ffa1ac089963765d65a2dd355364114/extern/storage-sealing/commit_batch.go#L193-L195)we are doing nothing, but we should  processIndividually in this case to send the commits out

arajasek:
Might add a test, but here's a PR.

Fixes #6541

### DIFF
--- a/extern/storage-sealing/commit_batch.go
+++ b/extern/storage-sealing/commit_batch.go
@@ -108,7 +108,7 @@ func (b *CommitBatcher) run() {
 		}
 		lastMsg = nil
 
-		var sendAboveMax, sendAboveMin bool
+		var sendAboveMax bool
 		select {
 		case <-b.stop:
 			close(b.stopped)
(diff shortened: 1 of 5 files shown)


magik6k (approved):

BigLep:
Is there a reason we wouldn't add a test here?

jennijuju:
I think the same logic change should be applied to pre commit batcher too.

jennijuju (changes requested):
I think the same logic change should be applied to pre commit batcher too.

magik6k:
> I think the same logic change should be applied to pre commit batcher too.

In the precommit batcher the min batch size has a slightly different meaning as we haven't committed to publishing those sectors yet, and sending individual precommits makes no sense, so the only use for MinBatch is not sending anything

We might want to have Commit / PreCommitMinBatch config vars have different names tho, to make sure it's clear that they are doing quite different things

arajasek:
> > I think the same logic change should be applied to pre commit batcher too.
> 
> In the precommit batcher the min batch size has a slightly different meaning as we haven't committed to publishing those sectors yet, and sending individual precommits makes no sense, so the only use for MinBatch is not sending anything
> 
> We might want to have Commit / PreCommitMinBatch config vars have different names tho, to make sure it's clear that they are doing quite different things

Aren't there two cases when we _do_ want to publish sectors, though:

- if a ticket is expiring (and we're risking losing the PC work)
- if a deal included in the sector is approaching activation

arajasek:
Summary: makes sense to just remove MinPreCommitBatch entirely -- if the timer goes off, we just batch and precommit whatever we have.

Thanks @jennijuju!

magik6k (approved):

PR #6566, opened 2021-06-23
scale up sector expiration to avoid sector expire in batch-pre-commit waitting

GFZRZK:
Here, 10 epoch is no longer satisfied because it mostly will sleep more than 10 epochs in batch submit pre-commit, so I suggested it should scale up to the longest time that pre-commit ticket expired at

### DIFF
--- a/extern/storage-sealing/states_sealing.go
+++ b/extern/storage-sealing/states_sealing.go
@@ -274,7 +274,7 @@ func (m *Sealing) preCommitParams(ctx statemachine.Context, sector SectorInfo) (
 
 	msd := policy.GetMaxProveCommitDuration(actors.VersionForNetwork(nv), sector.SectorType)
 
-	if minExpiration := height + msd + miner.MinSectorExpiration + 10; expiration < minExpiration {
+	if minExpiration := sector.TicketEpoch + policy.MaxPreCommitRandomnessLookback + msd + miner.MinSectorExpiration; expiration < minExpiration {
 		expiration = minExpiration
 	}
 	// TODO: enforce a reasonable _maximum_ sector lifetime?


extern/storage-sealing/states_sealing.go
```diff
@@ -274,7 +274,7 @@ func (m *Sealing) preCommitParams(ctx statemachine.Context, sector SectorInfo) (
 
 	msd := policy.GetMaxProveCommitDuration(actors.VersionForNetwork(nv), sector.SectorType)
 
-	if minExpiration := height + msd + miner.MinSectorExpiration + 10; expiration < minExpiration {
+	if minExpiration := sector.TicketEpoch + policy.MaxPreCommitRandomnessLookback + msd + miner.MinSectorExpiration; expiration < minExpiration {
```
magik6k, line 277: Do we need the `+ msd` here? I'd think that `sector.TicketEpoch + policy.MaxPreCommitRandomnessLookback` covers the entire period in which the sector can be procommitted.

magik6k (commented):
Great catch, just one question.

extern/storage-sealing/states_sealing.go
GFZRZK: Yes. When PreCommit sector [validateExpiration](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L721) in specs-actors, it uses `curEpoch +  MaxProveCommitDuration[precommit.SealProof]` as its activation, so I think we'd better add msd here. `sector.TicketEpoch + policy.MaxPreCommitRandomnessLookback` only covers precommit expire time

magik6k (approved):

Issue #6157, opened 2021-04-30
E2E Testing for FIP8 & FIP13


Stebalien:
FIP8 and FIP13 will allow batched pre-commits and batched prove-commits. This process needs end-to-end testing at scale (100s of sectors, test match batch sizes).

1. We need unit/integration tests to verify pre/prove-commit batching.
    1. What happens when a single sector fails?
2. We need to test this in a dev-net.

BigLep:
Discussion notes:
1. Scale
 - 100's of sectors all at once

2. Assertions
 - What happens if 1 of 100 sectors fail
 - Pipeline is able to notice that we don't have enough sectors in a batch and thus wait.

BigLep:
@magik6k and @vyzo : I want to make sure we're clear about the testing work for FIP8 and 13.  I want to make sure we're being precise about what we expect tested, that we have owners, and that we're tracking it (whether as a checklist or separate issues).  I took a first stab below based on what I saw in this issue and in PR #6235.  Please correct and ideally set as the subject of this issue or another.

In addition to making sure we don't lose track of anything, it'll be important as we update our estimates of the remaining work on 2021-05-20. 

<table>
<tr>
	<th>Item
	<th>Precommit Status
	<th>Postcommit Status
	<th>Comment
<tr>
	<td>Unit tests for CommitBatcher
	<td>[ ]
	<td>[ ]
	<td>Owner: @mention | Issue/PR: #XXXX  
<tr>
	<td>Commit Batching integration test: one porep fails
	<td>[ ]
	<td>[ ]
	<td>Owner: @mention | Issue/PR: #XXXX  
<tr>
	<td>Commit Batching integration test: one commit expires
	<td>[ ]
	<td>[ ]
	<td>Owner: @mention | Issue/PR: #XXXX  
<tr>
	<td>Commit Batching integration test: Batch fails (e.g., out of gas)
	<td>[ ]
	<td>[ ]
	<td>Owner: @mention | Issue/PR: #XXXX  
<tr>
	<td>Scale testing
	<td>[ ]
	<td>[ ]
	<td>This is automated verification that we can handle 100's of sectors

Owner: @mention | Issue/PR: #XXXX  
</table>

magik6k:
(updated the comment above to have a better list of what needs to be done)

arajasek:
#6432 adds unit tests for both the precommit and commit batchers

arajasek:
- Unit tests landed in #6432 
[...]

PR #6570, opened 2021-06-23
[wip] Test for deals expiring in batchers

magik6k:
This will address https://github.com/filecoin-project/lotus/issues/6157

Draft because the test is not triggering the expected logic yet / has some incorrect asserts, but already shows that the thing works at least ~80% as expected

### DIFF
--- a/extern/storage-sealing/precommit_batch.go
+++ b/extern/storage-sealing/precommit_batch.go
@@ -181,8 +181,12 @@ func (b *PreCommitBatcher) maybeStartBatch(notif bool) ([]sealiface.PreCommitBat
 		return nil, xerrors.Errorf("getting config: %w", err)
 	}
 
-	if notif && total < cfg.MaxPreCommitBatch {
-		return nil, nil
+	if notif {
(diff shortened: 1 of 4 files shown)


extern/storage-sealing/precommit_batch.go
```diff
@@ -171,8 +171,12 @@ func (b *PreCommitBatcher) maybeStartBatch(notif, after bool) ([]sealiface.PreCo
...
 	}
 
-	if notif && total < cfg.MaxPreCommitBatch {
-		return nil, nil
+	if notif {
+		if total < cfg.MaxPreCommitBatch {
+			return nil, nil
+		}
+
+		log.Info("processing batch: queued more than Max")
```
arajasek, line 179: can this not happen at exactly Max?

arajasek:
Nice work!

arajasek:
@gagizoned I don't think this should be in the milestone, but def good to get this across the finish line

magik6k:
After fighting with this test some more I'm pretty sure that this specific case can't actually happen, which is why it's basically impossible to make this test do what it says in the title.

Issue #6575, opened 2021-06-23
Add MaxStartEpochDelay to miner configuration 


jennijuju:
BasicDealFilter rejects deals that are >7 days in the future, ideally miners can set this value themselves in the configuration and we pull from it https://github.com/filecoin-project/lotus/blob/release/v1.10.0/node/modules/storageminer.go#L568-L573

PR #6576, opened 2021-06-23
Add miner-side MaxDealStartDelay config

magik6k:
Fixes https://github.com/filecoin-project/lotus/issues/6575


ribasushi (approved):
@magik6k perhaps add a comment in `node/impl/client/client.go` that if the default gets raised further it needs to be synchronized with what's in the config/def

PR #6581, opened 2021-06-24
fix handleSubmitPreCommitBatch() preCommitParams error

johnli-helloworld:
preCommitParams() will send event ， and two events will be sent.

### DIFF
--- a/extern/storage-sealing/states_sealing.go
+++ b/extern/storage-sealing/states_sealing.go
@@ -360,7 +360,7 @@ func (m *Sealing) handleSubmitPreCommitBatch(ctx statemachine.Context, sector Se
 
 	params, deposit, _, err := m.preCommitParams(ctx, sector)
 	if params == nil || err != nil {
-		return ctx.Send(SectorChainPreCommitFailed{xerrors.Errorf("preCommitParams: %w", err)})
+		return err
 	}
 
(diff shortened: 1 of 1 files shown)


extern/storage-sealing/states_sealing.go
```diff
@@ -360,7 +360,7 @@ func (m *Sealing) handleSubmitPreCommitBatch(ctx statemachine.Context, sector Se
 
 	params, deposit, _, err := m.preCommitParams(ctx, sector)
 	if params == nil || err != nil {
-		return ctx.Send(SectorChainPreCommitFailed{xerrors.Errorf("preCommitParams: %w", err)})
+		return err
```
magik6k, line 363: I think it's more correct to still send the event if we see the error
```suggestion
		if err != nil {
			return ctx.Send(SectorChainPreCommitFailed{xerrors.Errorf("preCommitParams: %w", err)})
		}
		return nil // event was sent in preCommitParams
```

magik6k (commented):
Thanks for the PR, just 1 comment.

There is also the same `if` in `handlePreCommitting`, should fix that one too.

jennijuju:
@johnli-helloworld  - Could you please see Magik's comment and address it accordingly?

magik6k:
Merged in https://github.com/filecoin-project/lotus/pull/6763, thanks!

PR #6582, opened 2021-06-24
Fix an error in msigLockCancel

IPFSUnion:
Only 5 parameters are needed to call lock-cancel command, which can be seen in line 1432


IPFSUnion:
截屏2021-06-24 14 18 54

magik6k (approved):

PR #6588, opened 2021-06-24
WalletVerify should support ID addresses

### LINKED ISSUE #6586: [BUG] CLI `wallet sign` and `wallet verify` differ in `f0...` address support
**Describe the bug**

While `lotus wallet sign` supports/resolves short addresses to their actual robust versions, `lotus wallet verify` fails to do the same:

# Works
```
~$ lotus wallet verify \
  f3qh7m6ul2n6pbko6bovtthogzv4qdu4bh6hlmbjw55khic6wegintl5ma55fwbxjzh5g4hweyj3n5j3lbd3ra \
  $( echo "boo" | xxd -p ) \
  $( lotus wallet sign f022197 $( echo "boo" | xxd -p ) )
valid
```
# Fails
```
~$ lotus wallet verify \
  f022197 \
  $( echo "boo" | xxd -p ) \
  $( lotus wallet sign f022197 $( echo "boo" | xxd -p ) )
invalid
ERROR: CLI Verify called with invalid signature
```

v1.10.0

arajasek:
Fixes #6586

### DIFF
--- a/node/impl/full/wallet.go
+++ b/node/impl/full/wallet.go
@@ -71,7 +71,11 @@ func (a *WalletAPI) WalletSignMessage(ctx context.Context, k address.Address, ms
 }
 
 func (a *WalletAPI) WalletVerify(ctx context.Context, k address.Address, msg []byte, sig *crypto.Signature) (bool, error) {
-	return sigs.Verify(sig, k, msg) == nil, nil
+	keyAddr, err := a.StateManagerAPI.ResolveToKeyAddress(ctx, k, nil)
+	if err != nil {
+		return false, xerrors.Errorf("failed to resolve ID address: %w", keyAddr)
+	}
+	return sigs.Verify(sig, keyAddr, msg) == nil, nil
 }
 
 func (a *WalletAPI) WalletDefaultAddress(ctx context.Context) (address.Address, error) {


ribasushi (approved):
❤️

jennijuju:
This seems weird to me. I’d consider Id addresses are just some index of wallet account/actor account  - not wallet itself which is this api is suggesting to verify

raulk (commented):
Are we sure we want this? ID addresses are subject to reorg, and thus this function is really impure.

EDIT: the motivation appears to be to bring symmetry w.r.t. https://github.com/filecoin-project/lotus/issues/6586, but I'm not even sure `lotus wallet sign f<id>` is a good idea.

arajasek:
We could overhaul both sign and verify to use ResolveToKeyAddressAtFinality. Or just refuse to use ID addresses here. I don't feel too strongly.

ribasushi:
@raulk @jennijuju this is CLI, not API territory. We generally support `f0xxx` in most places ( notably in all variations of `send`), as we **pint** `f0xxx` addresses in most `info`-style places, knowing full well they are unstable. 

If we want to reject this change - we have to also commit to overhauling the CLI to make it self-consistent, otherwise the exact error I ran into would keep popping up.

jennijuju:
> We could overhaul both sign and verify to use ResolveToKeyAddressAtFinality. Or just refuse to use ID addresses here. I don't feel too strongly.

+1 on the latter option.

jennijuju:
> @raulk @jennijuju this is CLI, not API territory. We generally support `f0xxx` in most places ( notably in all variations of `send`), as we **print** `f0xxx` addresses in most `info`-style places, knowing full well they are unstable.
> 
> If we want to reject this change - we have to also commit to overhauling the CLI to make it self-consistent, otherwise the exact error I ran into would keep popping up.

I think both should refuse Id address

Kubuxu:
In my opinion, we should be moving forward to supporting ID addresses for all actors after the finality passed, not move in the reverse direction.

raulk:
Resolving to ID addresses at the last finalised epoch SGTM

raulk:
It's a bit complicated from the UX perspective though. As it binds the success or failure of the operation to time the ID address was created, which isn't info the user can obtain easily.

ribasushi:
> It's a bit complicated from the UX perspective though. As it binds the success or failure of the operation to time the ID address was created, which isn't info the user can obtain easily.

It's something we can make a part of the error:
`f0xxxx resolving to f1yyyy was created less than 900 epochs ago: use the long form once you have validated its legitimacy`

Stebalien:
Converting back to draft until the feedback is addressed.

arajasek:
Stale

PR #6593, opened 2021-06-25
handleCommitFailed remove checkPrecommit

johnli-helloworld:
If the sector commits a COMMIT message and fails, for example, there is not enough money. After retry (enough money) the expectation is raised, but this logic will make it fail


johnli-helloworld:
especially ticket expired

johnli-helloworld:
Sorry, there are some problems with this submission, it is too rough. but ticket expired this judgment is unreasonable

magik6k (approved):
Thanks!

* checkPrecommit was also dropped in https://github.com/filecoin-project/lotus/pull/6634
* The `SectorRetrySubmitCommit` on batch error will be merged in https://github.com/filecoin-project/lotus/pull/6638

Issue #6602, opened 2021-06-25
[BUG] libp2p yamux bug showing up in deal-making


jennijuju:
The current market in master may have trouble with making deals with nodes that on v1.10.0 and below - miner is observing that when deals being initiated, graphsync doesn't send any blocks.

Before the issue got fixed in go-libp2p, the immediate thing action we can takeon lotus is to downgrade just the yamux dependency (github.com/libp2p/go-libp2p-yamux v0.4.1) via a go mod replace,

raulk:
This is not a market protocol incompatibility, but rather an issue with the latest go-libp2p dependencies. Downgrading go-libp2p-yamux to `github.com/libp2p/go-libp2p-yamux v0.4.0` via a replace directive fixes this. @magik6k will post a PR/patch soon. 

We will need a integration test reproducing this issue reliably so we can debug what's the regression in yamux.

raulk:
Reproduction with testground here: https://github.com/raulk/whitenoise/compare/repro/lotus-6602?expand=1

arajasek:
1.11 works around this by downgrading

gagizoned:
@arajasek @raulk anything else we need to consider for this one or we can close it, since It’s classified as P0?

PR #6605, opened 2021-06-25
downgrade libp2p/go-libp2p-yamux to v0.5.1.

raulk:
This should fix the windowing issue seen in https://github.com/filecoin-project/lotus/issues/6602. It is a temporary fix to make master happy, while we investigate.


magik6k (approved):
Seems to help on my miner

PR #6612, opened 2021-06-28
Miner SimultaneousTransfers config

magik6k:


dirkmc (approved):

raulk (changes requested):
Can we add an integration test with the new testkit?

raulk:
Suggestion: capture all transfer start/stop events (by subscribing to markets events), set a limit of –say– 2 simultaneous transfers, perform 100 concurrent transfers, and then go over the events to ensure that only 2 transfers were ever active at a time.

itests/deals_concurrent_test.go
```diff
@@ -47,3 +57,71 @@ func TestDealCyclesConcurrent(t *testing.T) {
...
+						maxOngoing = len(ongoing)
+					}
+				case <-ctx.Done():
+					return
+				}
+			}
+		}()
+
+		dh.RunConcurrentDeals(kit.RunConcurrentDealsOpts{
+			N:             1, // TODO: set to 20 after https://github.com/ipfs/go-graphsync/issues/175 is fixed
```
magik6k, line 115: With this set to 10, when the test would not hang because of this issue, the assert below would fail if the config wasn't set, and it does pass when it's set

arajasek (approved):

PR #6620, opened 2021-06-29
Envvar to disable slash filter

arajasek:
Maybe we merge this? Maybe we don't.

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -326,7 +326,7 @@ minerLoop:
 
 			if err := m.sf.MinedBlock(b.Header, base.TipSet.Height()+base.NullRounds); err != nil {
 				log.Errorf("<!!> SLASH FILTER ERROR: %s", err)
-				if os.Getenv("LOTUS_MINER_NO_FILTER") != "_yes_" {
+				if os.Getenv("LOTUS_MINER_NO_FILTER") != "_yes_i_know_i_can_and_probably_will_lose_all_my_fil_and_power_" {
 					continue
 				}
 			}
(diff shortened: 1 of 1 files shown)


miner/miner.go
```diff
@@ -325,7 +326,9 @@ minerLoop:
 
 			if err := m.sf.MinedBlock(b.Header, base.TipSet.Height()+base.NullRounds); err != nil {
 				log.Errorf("<!!> SLASH FILTER ERROR: %s", err)
-				continue
+				if os.Getenv("LOTUS_MINER_NO_FILTER") != "_yes_" {
```
magik6k, line 329:
```suggestion
				if os.Getenv("LOTUS_MINER_NO_FILTER") != "_yes_i_know_i_can_and_probably_will_lose_all_my_fil_and_power_" {
```

magik6k (commented):
I'd make the env var a bit more explicit in what can happen when this is set

miner/miner.go
arajasek: Seems reasonable

miner/miner.go
```diff
@@ -325,7 +326,9 @@ minerLoop:
 
 			if err := m.sf.MinedBlock(b.Header, base.TipSet.Height()+base.NullRounds); err != nil {
 				log.Errorf("<!!> SLASH FILTER ERROR: %s", err)
-				continue
+				if os.Getenv("LOTUS_MINER_NO_FILTER") != "_yes_i_know_i_can_and_probably_will_lose_all_my_fil_and_power_" {
```
ribasushi, line 329: Perhaps?

```suggestion
				if os.Getenv("LOTUS_MINER_NO_SLASHFILTER") != "_yes_i_know_i_can_and_probably_will_lose_all_my_fil_and_power_" {
```

W have a number of other `FILTER`s

ribasushi (approved):
LGTM

Suggested varname change

Can be merged as-is though

Issue #6622, opened 2021-06-29
[BUG] when we build a `ProveCommitAggregate ` msg,  we should set value is `collateral + aggFee` 


firesWu:
> Note: For security-related bugs/issues, please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).

**Describe the bug**
https://github.com/filecoin-project/lotus/blob/ba7e730d2465f639e4edbe41ad1b96dbc74cf9b3/extern/storage-sealing/commit_batch.go#L317

why only send collateral to miner, it should send `collateral  + aggFee`. 
if we only send collateral, `ProveCommitAggregate` was report this error.

https://github.com/filecoin-project/specs-actors/blob/f7d531232aa2b5dfe6c040d1703eaaaf9dad4668/actors/builtin/miner/miner_actor.go#L967

if people initialize a new miner, `ProveCommitAggregate` will exec failed until people send some fil to miner

**Version (run `lotus version`):**

v1.10.0

**Expected behavior**

it should send `collateral  + aggFee` to miner when send `ProveCommitAggregate` msg

arajasek:
Closed by #6625

PR #6623, opened 2021-06-29
fix: miner balance is not enough, so that ProveCommitAggregate msg ex…

firesWu:
solve https://github.com/filecoin-project/lotus/issues/6622


magik6k (approved):
Thanks for catching this!

magik6k:
(marking as backport in case we decide to do 1.10.1)

magik6k:
Merged in https://github.com/filecoin-project/lotus/pull/6625

PR #6625, opened 2021-06-29
fix prove commit aggregate send token amount

magik6k:
This is #6623, but with conflicts resolved


magik6k:
(https://github.com/filecoin-project/lotus/pull/6623 was approved already)

PR #6627, opened 2021-06-29
api: Separate the Net interface from Common

magik6k:
This implements the basic separation of the Net api from CommonAPI, which should make it much easier to construct nodes without it

(this is for https://github.com/filecoin-project/lotus/pull/6356, but can be merged separately to reduce PR sizes)


api/api_storage.go
```diff
@@ -41,6 +41,7 @@ import (
 // StorageMiner is a low-level interface to the Filecoin network storage miner node
 type StorageMiner interface {
 	Common
+	Net
```
magik6k, line 44: I don't think we have to do this in initial MRA split PR, but ideally:
* We'd have the miner interface split into more, per service interfaces
* In the API serve logic we'd only construct the RPC server with the services a node actually supports

node/builder.go
```diff
@@ -363,6 +364,7 @@ var ChainNode = Options(
 var MinerNode = Options(
 	// API dependencies
 	Override(new(api.Common), From(new(common.CommonAPI))),
+	Override(new(api.Net), From(new(net.NetAPI))),
```
magik6k, line 367: In no-market mode this can be provided with `Override(new(api.Net), api.NetStub{})` until we do more of the api-splitting work

node/impl/storminer.go
```diff
@@ -39,15 +40,17 @@ import (
...
-	"github.com/filecoin-project/lotus/node/impl/common"
 	"github.com/filecoin-project/lotus/node/modules/dtypes"
 	"github.com/filecoin-project/lotus/storage"
 	"github.com/filecoin-project/lotus/storage/sectorblocks"
 	sto "github.com/filecoin-project/specs-storage/storage"
 )
 
 type StorageMinerAPI struct {
-	common.CommonAPI
+	fx.In
```
nonsense, line 50: @magik6k why is this necessary now?

magik6k: It's no longer provided by the embedded structs (which Fx apparently was picking up)

magik6k: (`fx.In` embeds `dig.In`, which is a private interface, so if it was embedded in deeper structs, the parent struct would 'inherit' it)

api/proxy_util_test.go
```diff
@@ -0,0 +1,36 @@
...
+	require.Len(t, sts, 2)
+
+	sa := sts[0].(*struct{ A int })
+	sa.A = 3
+	sb := sts[1].(*struct{ B int })
+	sb.B = 4
+
+	require.Equal(t, 3, proxy.Internal.A)
+	require.Equal(t, 4, proxy.StrB.Internal.B)
+}
```
ZenGround0, line 36: Something maybe a bit useful would be to have a test case with a nested Internal:
```
type StrC struct {
   Internal struct {
      Internal struct {}
   }
}
```

to codify that the inner Internals don't get picked up.

ZenGround0 (approved):
LGTM

PR #6629, opened 2021-06-29
Config for collateral from miner available balance

### LINKED ISSUE #6603: Allow PreCommit collateral to be moved from miner available balance
Precommit collateral is moved from miner available balance in miner actor. Currently, in lotus, we are "forcing" the collateral to be paid from the worker/control address of the miner with the precommitsector message which leads miners to need to monitor the address balance closely so to maintain their miner operations = even they have enough funds in available balance to cover the collateral.

Two proposals we have:
A:  Add a check to see if miner's enough available balance to cover `collateral + buffer`, if it does, don't send any more funds with the message, if not, send the collateral with the message. 
B: make it configuration for miner to chose if they want to pay off using available balance or worker/control address.

magik6k:
Closes #6603

### DIFF
--- a/extern/storage-sealing/utils.go
+++ b/extern/storage-sealing/utils.go
@@ -10,7 +10,6 @@ import (
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
 
-	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/extern/storage-sealing/sealiface"
 )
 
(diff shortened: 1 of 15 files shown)


node/config/def.go
```diff
@@ -312,12 +315,13 @@ func DefaultStorageMiner() *StorageMiner {
...
-			WaitDealsDelay:            Duration(time.Hour * 6),
-			AlwaysKeepUnsealedCopy:    true,
-			FinalizeEarly:             false,
+			MaxWaitDealsSectors:        2, // 64G with 32G sectors
+			MaxSealingSectors:          0,
+			MaxSealingSectorsForDeals:  0,
+			WaitDealsDelay:             Duration(time.Hour * 6),
+			AlwaysKeepUnsealedCopy:     true,
+			FinalizeEarly:              false,
+			CollateralFromMinerBalance: false,
```
magik6k, line 324: Defaults to false as that won't work on fresh miners without sending them some funds first. We could add a separate option to prefer miner-available funds, and only send more if needed, but that requires a bit more advanced accounting logic (here we can rely on gas estimation failing in case of not having enough funds)

arajasek: Is the accounting any more advanced? Why is it not just a matter of setting `value = AvailableBalance - CollateralNeeded - Cfg.Buffer`, where `Buffer` is some minimum we don't want available balance to fall below.

Gas estimation should work exactly the same and catch any insufficient funds?

arajasek (commented):
This looks correct, but I would definitely prefer a minimum amount instead of just a boolean use / don't use.

node/config/def.go
```diff
@@ -324,6 +331,10 @@ func DefaultStorageMiner() *StorageMiner {
 			AlwaysKeepUnsealedCopy:    true,
 			FinalizeEarly:             false,
 
+			CollateralFromMinerBalance: false,
+			AvailableBalanceBuffer:     types.FIL(big.Zero()),
```
arajasek, line 335: Is 0 the best default here? I guess it's okay because `CollateralFromMinerBalance` is defaulting to false anyway.

extern/storage-sealing/utils.go
```diff
@@ -55,3 +63,30 @@ func (m *Sealing) GetSectorInfo(sid abi.SectorNumber) (SectorInfo, error) {
...
 	return out, err
 }
+
+func collateralSendAmount(ctx context.Context, api interface {
+	StateMinerAvailableBalance(context.Context, address.Address, TipSetToken) (big.Int, error)
+}, maddr address.Address, cfg sealiface.Config, collateral abi.TokenAmount) (abi.TokenAmount, error) {
+	if cfg.CollateralFromMinerBalance {
+		avail := types.TotalFilecoinInt
+
+		if !cfg.DisableCollateralFallback {
```
arajasek, line 73: Seems correct as-is, but I feel like readability is improved by saying `if !cfg.DisableCollateralFallback { return big.Zero() }` at the start of this method.

arajasek (approved):
LGTM

extern/storage-sealing/utils.go
magik6k: Maybe, but that isn't correct as when we have `!cfg.DisableCollateralFallback`, collateral still can be non-zero

magik6k: (unless you meant that the other way)

arajasek: i did mean that the other way

PR #6633, opened 2021-06-29
run digitalocean erase script

coryschwartz:
Last time we pushed, we were rejected, and with good reason. Packer left
a ssh installed on the user account. DigitalOcean has a "cleanup"
script they like people to use. This PR will have us use their clean-up
script.

This  is the rejection message from DO:
rejection

I've re-submitted with this change to the build, This PR is marked draft. Will mark ready for review once DO approves the image.


coryschwartz:
I'm going to close this, but I'm opening a similar PR. Need to make sure this is based just on 1.11 and nothing else from master.

PR #6634, opened 2021-06-30
remove precommit check in handleCommitFailed

GFZRZK:
it makes no sense to check pre-commit info in`handleCommitFailed`, in the test we find that `checkPrecommit` will throw an `ErrExpiredTicket` easily because the expiration of proveCommit change to 30, that will cause the good sector to be removed. 
I've noticed these are old codes. Is there any other reason to check pre-commit now?

### DIFF
--- a/extern/storage-sealing/states_failed.go
+++ b/extern/storage-sealing/states_failed.go
@@ -182,7 +182,7 @@ func (m *Sealing) handleComputeProofFailed(ctx statemachine.Context, sector Sect
 }
 
 func (m *Sealing) handleCommitFailed(ctx statemachine.Context, sector SectorInfo) error {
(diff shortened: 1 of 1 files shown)


extern/storage-sealing/states_failed.go
```diff
@@ -216,32 +216,6 @@ func (m *Sealing) handleCommitFailed(ctx statemachine.Context, sector SectorInfo
 		}
 	}
```
magik6k, line 218: gofmt complains about this

```suggestion
```

magik6k (approved):
Thanks for the fix.

Ah yes, this made more sense on older actors, and before `checkCommit` was also doing deal checks

(marking as backport as this should probably be in 1.10.1 if we decide to release it)

PR #6635, opened 2021-06-30
fix ticket expiration check

llifezou:
fix ticket expiration check, otherwise it may cause a large number of loops to retry GetTicket when retrying PreCommit1；

Because there may be C1 and C2 execution failures, and after they retries more than a certain number of times,  will cause PreCommit1 to be re-executed, and now it is the submission period of 30, so it will cause an endless loop of repeated execution of GetTicket.

### DIFF
--- a/extern/storage-sealing/states_sealing.go
+++ b/extern/storage-sealing/states_sealing.go
@@ -151,7 +151,7 @@ func (m *Sealing) getTicket(ctx statemachine.Context, sector SectorInfo) (abi.Se
 	}
 
 	if allocated { // allocated is true, sector precommitted but expired, will SectorCommitFailed or SectorRemove
-		return nil, 0, allocated, xerrors.Errorf("Sector %s precommitted but expired", sector.SectorNumber)
+		return nil, 0, allocated, xerrors.Errorf("sector %s precommitted but expired", sector.SectorNumber)
 	}
 
 	rand, err := m.api.ChainGetRandomnessFromTickets(ctx.Context(), tok, crypto.DomainSeparationTag_SealRandomness, ticketEpoch, buf.Bytes())
--- a/extern/storage-sealing/states_sealing.go
+++ b/extern/storage-sealing/states_sealing.go
@@ -109,22 +109,30 @@ func checkProveCommitExpired(preCommitEpoch, msd abi.ChainEpoch, currEpoch abi.C
 	return currEpoch > preCommitEpoch+msd
 }
 
-func (m *Sealing) getTicket(ctx statemachine.Context, sector SectorInfo) (abi.SealRandomness, abi.ChainEpoch, error) {
+func (m *Sealing) getTicket(ctx statemachine.Context, sector SectorInfo) (abi.SealRandomness, abi.ChainEpoch, bool, error) {
 	tok, epoch, err := m.api.ChainHead(ctx.Context())
 	if err != nil {
-		log.Errorf("handlePreCommit1: api error, not proceeding: %+v", err)
-		return nil, 0, nil
+		log.Errorf("getTicket: api error, not proceeding: %+v", err)
+		return nil, 0, false, nil
+	}
+
+	// the reason why the StateMinerSectorAllocated function is placed here, if it is outside,
+	//	if the MarshalCBOR function and StateSectorPreCommitInfo function return err, it will be executed
+	allocated, aerr := m.api.StateMinerSectorAllocated(ctx.Context(), m.maddr, sector.SectorNumber, nil)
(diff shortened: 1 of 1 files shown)


llifezou:
this large number of cycles will cause the kvlog to increase rapidly

extern/storage-sealing/states_sealing.go
```diff
@@ -182,14 +193,35 @@ func (m *Sealing) handlePreCommit1(ctx statemachine.Context, sector SectorInfo)
...
 
 	if checkTicketExpired(sector.TicketEpoch, height) {
-		return ctx.Send(SectorOldTicket{}) // go get new ticket
+		pci, err := m.api.StateSectorPreCommitInfo(ctx.Context(), m.maddr, sector.SectorNumber, tok)
+		if err != nil {
+			log.Errorf("handlePreCommit1: StateSectorPreCommitInfo: api error, not proceeding: %+v", err)
+			return nil
+		}
+
+		if pci == nil {
```
magik6k, line 209: It is possible that after the precommit has expired on-chain, `pci` can be nil, but the sector number is allocated (`StateMinerSectorAllocated` returns true, but `StateSectorPreCommitInfo` returns nil - this means that we won't be able to precommit this sector again). In that case the only thing that we can do is to remove the sector.

We should probably call StateMinerSectorAllocated here, and if it returns true, just do `return ctx.Send(SectorRemove{})`

magik6k (commented):
I'm not 100% sure if this will fully fix the issue, it likely can still happen after the precommit expires fully

extern/storage-sealing/states_sealing.go
llifezou: yes, you are right. thx

extern/storage-sealing/states_sealing.go
```diff
@@ -105,48 +105,66 @@ func checkTicketExpired(ticket, head abi.ChainEpoch) bool {
...
 		}
+
+		msd := policy.GetMaxProveCommitDuration(actors.VersionForNetwork(nv), sector.SectorType)
+
+		if checkProveCommitExpired(pci.PreCommitEpoch, msd, epoch) {
+			return nil, 0, allocated, xerrors.Errorf("ticket expired for precommitted sector")
+		}
+	}
+
+	if allocated { // allocated is true, sector precommitted but expired, will SectorCommitFailed or SectorRemove
```
magik6k, line 153: I think this should be
```suggestion
	if pci == nil && allocated { // allocated is true, sector precommitted but expired, will SectorCommitFailed or SectorRemove
```

If the precommit is still not expired, we can retry sealing with the same ticket (this is what L139 - `ticketEpoch = pci.Info.SealRandEpoch` - is for)

magik6k (commented):
Looks good with 1 comment

extern/storage-sealing/states_sealing.go
```diff
@@ -105,48 +105,66 @@ func checkTicketExpired(ticket, head abi.ChainEpoch) bool {
...
+		return nil, 0, allocated, xerrors.Errorf("Sector %s precommitted but expired", sector.SectorNumber)
```
arajasek, line 154:
```suggestion
		return nil, 0, allocated, xerrors.Errorf("sector %s precommitted but expired", sector.SectorNumber)
```

extern/storage-sealing/states_sealing.go
```diff
@@ -105,48 +105,66 @@ func checkTicketExpired(ticket, head abi.ChainEpoch) bool {
 	return head-ticket > MaxTicketAge // TODO: allow configuring expected seal durations
 }
 
-func (m *Sealing) getTicket(ctx statemachine.Context, sector SectorInfo) (abi.SealRandomness, abi.ChainEpoch, error) {
+func checkProveCommitExpired(preCommitEpoch, msd abi.ChainEpoch, currEpoch abi.ChainEpoch) bool {
```
arajasek, line 108: should be called `checkPreCommitExpired`?

arajasek (commented):
First pass looks good but I haven't fully reasoned about all the cases...Will do a second pass later (but okay to merge for now).

extern/storage-sealing/states_sealing.go
magik6k: This is checking if we're past the time to prove (not precommit) a sector, so I think this name is correct

extern/storage-sealing/states_sealing.go
llifezou: You are right, I missed this piece of logic.

magik6k (approved):
Thanks!

PR #6636, opened 2021-06-30
Reuse timers in sealing batch logic

johnli-helloworld:
The original batchwait would cause a timer instance to be created for each round of the for loop。

### DIFF
--- a/extern/storage-sealing/commit_batch.go
+++ b/extern/storage-sealing/commit_batch.go
@@ -106,6 +106,7 @@ func (b *CommitBatcher) run() {
 		panic(err)
 	}
 
+	timer := time.NewTimer(b.batchWait(cfg.CommitBatchWait, cfg.CommitBatchSlack))
 	for {
 		if forceRes != nil {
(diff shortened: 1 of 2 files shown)


extern/storage-sealing/commit_batch.go
```diff
@@ -132,17 +133,26 @@ func (b *CommitBatcher) run() {
...
-func (b *CommitBatcher) batchWait(maxWait, slack time.Duration) <-chan time.Time {
+func (b *CommitBatcher) batchWait(maxWait, slack time.Duration) time.Duration {
 	now := time.Now()
 
 	b.lk.Lock()
 	defer b.lk.Unlock()
 
 	if len(b.todo) == 0 {
-		return nil
+		return maxWait
```
magik6k, line 155: Ideally we wouldn't be resetting the timer here, but I don't think this micro-optimization is worth the added complexity

magik6k (approved):
Thanks for the PR!

PR #6639, opened 2021-06-30
polish(statetree): accept a context in statetree diff for timeouts

frrist:
- this operation can take a while, this change gives users the option
to abort if it takes "too long".

Motivation: this operation takes a while for sentinel to perform, we'd like to abort if it takes too long.

@iand if this looks alright to you I'll tag lotus devs for a second look

### DIFF
--- a/chain/state/statetree.go
+++ b/chain/state/statetree.go
@@ -547,7 +547,7 @@ func (st *StateTree) Version() types.StateTreeVersion {
 	return st.version
 }
 
-func Diff(oldTree, newTree *StateTree) (map[string]types.Actor, error) {
+func Diff(ctx context.Context, oldTree, newTree *StateTree) (map[string]types.Actor, error) {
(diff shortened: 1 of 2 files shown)


iand:
This is the pattern I expected, so :+1: from me

chain/state/statetree.go
```diff
@@ -547,41 +547,46 @@ func (st *StateTree) Version() types.StateTreeVersion {
...
-		addr, err := address.NewFromBytes([]byte(k))
-		if err != nil {
-			return xerrors.Errorf("address in state tree was not valid: %w", err)
-		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			var act types.Actor
+
```
iand, line 563: I would normally close the select here with the default case being essentially a no-op. The context check then becomes like an early return.

magik6k (approved):

PR #6649, opened 2021-07-01
fixed bug for sector FinalizeFailed when set FinalizeEarly=true

yangjian102621:


extern/storage-sealing/fsm.go
```diff
@@ -123,7 +123,7 @@ var fsmPlanners = map[SectorState]func(events []statemachine.Event, state *Secto
 		on(SectorRetrySubmitCommit{}, SubmitCommit),
 	),
 	CommitAggregateWait: planOne(
-		on(SectorProving{}, FinalizeSector),
```
magik6k, line 126: That should be ok, calling finalize for the second time should just see that the sector is already finalized, and go to Proving that way

What errors are you seeing?

lionsoul2014: before SubmitCommitAggregate and everything is find and the sectors were finalized, also the data were downloaded to the final storage server, After SubmitCommitAggregate the sectors state keep firing the following steps:
```
SubmitCommitAggregate(on.SectorCommitAggregateSent)->CommitWait(on.SectorProving)->FinalizeSector
```
since the sectors were deleted before the SubmitCommitAggregate finalize operation and the Finalize call now will fire SectorFinalizeFailed CUZ of sector not found error, it then fire FinalizeFailed and 
```
	FinalizeFailed: planOne(
		on(SectorRetryFinalize{}, FinalizeSector),
	),
``

it keeps firing FinalizeSector and it keeps retry finalize and the sectors state will always be FinalizedFailed.

that's what happened during our local tests.

lionsoul2014: also this PR will cause a new issue for when if the FinalizeEarly were set to false and it will not call the finalize after SectorProving

magik6k:
https://github.com/filecoin-project/lotus/pull/6653 should fix the underlying Finalize issue, it would be really great if you were able to see if it fixes the issue for you

extern/storage-sealing/fsm.go
yangjian102621: > That should be ok, calling finalize for the second time should just see that the sector is already finalized, and go to Proving that way
> 
> What errors are you seeing?

The second FinalizeSector will cause the FinalizeFailed. Now it fixed in PR #6653 , Thanks.

yangjian102621:
This bug is fixed in   PR #6653

yangjian102621:
> #6653 should fix the underlying Finalize issue, it would be really great if you were able to see if it fixes the issue for you

Yep, i see. and your solution is indeed better, i had tested it in local dev-net and it running well, thanks.

PR #6651, opened 2021-07-01
Fix the bug of SectorRetryFinalize dead loop after SubmitCommitAggregate while FinalizeEarly is true

lionsoul2014:
with FinalizeEarly set to true:

Before SubmitCommitAggregate and everything is find and sectors were finalized:

commit-finalize-02

Also the data were all downloaded to the final storage server and all the sectors were on state SubmitCommitAggregate:

commit-aggregate-02

After SubmitCommitAggregate by force publish and the sectors state keep firing the following state changes:
```golang
SubmitCommitAggregate(on.SectorCommitAggregateSent)->CommitWait(on.SectorProving)->FinalizeSector
```
since the sectors were deleted before the SubmitCommitAggregate finalize call and the call to Finalize now will fire SectorFinalizeFailed CUZ of sector not found, it then fire FinalizeFailed and enter the following SectorRetryFinalize dead loop.
```golang
FinalizeSector: planOne(
    on(SectorFinalized{}, Proving),
    on(SectorFinalizeFailed{}, FinalizeFailed),
),
FinalizeFailed: planOne(
    on(SectorRetryFinalize{}, FinalizeSector),
),
```

BUG: **it keeps retrying finalize sectors and the sectors state will always be FinalizedFailed**.

**About this PR:**
1, Add handleCommitFinalizeSector to handler the commit finalize and it keeps retry until the finalize finished.
2, Ignore the finalize failed during handleFinalizeSector if cfg.FinalizeEarly is true since the commit finalize will handle this situation, the second call of finalize will return update the sector state to proving (For FinalizeEarly is set to true and ignore finalize retry).
3, everything as usual if cfg.FinalizeEarly is set to false.

now second calling of finalize failed:
finalize-failed

And all sectors state were updated to Proving:

state-proving

### DIFF
--- a/extern/storage-sealing/fsm.go
+++ b/extern/storage-sealing/fsm.go
@@ -388,7 +388,7 @@ func (m *Sealing) plan(events []statemachine.Event, state *SectorInfo) (func(sta
 	case CommitWait:
 		return m.handleCommitWait, processed, nil
 	case CommitFinalize:
-		fallthrough
+		return m.handleCommitFinalizeSector, processed, nil
 	case FinalizeSector:
 		return m.handleFinalizeSector, processed, nil
 
--- a/extern/storage-sealing/states_sealing.go
+++ b/extern/storage-sealing/states_sealing.go
@@ -686,7 +686,15 @@ func (m *Sealing) handleCommitWait(ctx statemachine.Context, sector SectorInfo) 
 	return ctx.Send(SectorProving{})
 }
 
+func (m *Sealing) handleCommitFinalizeSector(ctx statemachine.Context, sector SectorInfo) error {
+	return m.doFinalizeSector(ctx, sector, true)
+}
+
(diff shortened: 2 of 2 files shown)


magik6k:
https://github.com/filecoin-project/lotus/pull/6653 should fix the underlying Finalize issue, it would be really great if you were able to see if it fixes the issue for you

lionsoul2014:
> #6653 should fix the underlying Finalize issue, it would be really great if you were able to see if it fixes the issue for you

i read your code and it should be fixed. No matter how many times we invoke the finalize and it should be executed as long as the sector can be found, that is great solution.

lionsoul2014:
> #6653 should fix the underlying Finalize issue, it would be really great if you were able to see if it fixes the issue for you

i am going to test it in a local 2k network.

PR #6653, opened 2021-07-01
storage: Fix FinalizeSector with sectors in stoage paths

magik6k:
This fixes the case where sectors could be moved in one call to Finalize to a long-term storage, and then fail in second call to Finalize because we expected to find those files still in sealing storage

Related to https://github.com/filecoin-project/lotus/pull/6651 and https://github.com/filecoin-project/lotus/pull/6649

### DIFF
--- a/extern/sector-storage/manager.go
+++ b/extern/sector-storage/manager.go
@@ -528,10 +528,25 @@ func (m *Manager) FinalizeSector(ctx context.Context, sector storage.SectorRef, 
 		}
 	}
 
+	pathType := storiface.PathStorage
+	{
+		sealedStores, err := m.index.StorageFindSector(ctx, sector.ID, storiface.FTSealed, 0, false)
+		if err != nil {
+			return xerrors.Errorf("finding sealed sector: %w", err)
+		}
+
+		for _, store := range sealedStores {
+			if store.CanSeal {
+				pathType = storiface.PathSealing
+				break
+			}
+		}
+	}
+
 	selector := newExistingSelector(m.index, sector.ID, storiface.FTCache|storiface.FTSealed, false)
 
 	err := m.sched.Schedule(ctx, sector, sealtasks.TTFinalize, selector,
-		m.schedFetch(sector, storiface.FTCache|storiface.FTSealed|unsealed, storiface.PathSealing, storiface.AcquireMove),
+		m.schedFetch(sector, storiface.FTCache|storiface.FTSealed|unsealed, pathType, storiface.AcquireMove),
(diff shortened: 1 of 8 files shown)


extern/sector-storage/manager.go
```diff
@@ -528,10 +528,25 @@ func (m *Manager) FinalizeSector(ctx context.Context, sector storage.SectorRef,
...
+				break
+			}
+		}
+	}
+
 	selector := newExistingSelector(m.index, sector.ID, storiface.FTCache|storiface.FTSealed, false)
 
 	err := m.sched.Schedule(ctx, sector, sealtasks.TTFinalize, selector,
-		m.schedFetch(sector, storiface.FTCache|storiface.FTSealed|unsealed, storiface.PathSealing, storiface.AcquireMove),
+		m.schedFetch(sector, storiface.FTCache|storiface.FTSealed|unsealed, pathType, storiface.AcquireMove),
```
ZenGround0, line 549: I am brand new to this code so I have some very basic questions.

If we have already finalized this sector id then all of the `sealedStores` above will have `store.CanSeal` equal to `false` ?

Why do we need the worker to finalize again if the pathType is `PathStorage`?  Do sectors in long term storage need to be finalized again during subsequent finalize calls?  Similarly why do we need to do the `MoveStorage` below in this case?  Aren't all the sectors already in long term storage?

ZenGround0 (approved):
Confused about some of the basics but what I understand looks good

itests/deals_test.go
```diff
@@ -14,7 +14,7 @@ func TestDealsWithSealingAndRPC(t *testing.T) {
 
 	kit.QuietMiningLogs()
 
-	var blockTime = 1 * time.Second
+	var blockTime = 50 * time.Millisecond
```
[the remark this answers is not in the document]
magik6k: Correct

extern/sector-storage/manager.go
magik6k: > If we have already finalized this sector id then all of the sealedStores above will have store.CanSeal equal to false ?

In some setups storage paths can be used both for sealing and storage

> Why do we need the worker to finalize again if the pathType is PathStorage?

It's possible that the sector wasn't properly finalized for a number of other reasons (sealing path switched to storage, sectors moved manually, etc.). The finalize operation is quite cheap (removing/trimming some files), so there is little downside to calling it twice for the added robustness. 

>  Similarly why do we need to do the MoveStorage below in this case? Aren't all the sectors already in long term storage?

MoveStorage already has that check internally

PR #6650, opened 2021-07-01
commit batch: AggregateAboveBaseFee config

magik6k:
Closes https://github.com/filecoin-project/lotus/issues/6645

This PR adds `AggregateAboveBaseFee` to the Sealing config. The value is used to decide whether or not to do aggregation based on the current network basefee. Defaults to 0.15nFIL based on estimates from https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0013.md#batch-incentive-alignment

(also adds a regression test to https://github.com/filecoin-project/lotus/pull/6647)

PR #6656, opened 2021-07-02
Default AggregateCommits to false

Fatman13:
https://filecoinproject.slack.com/archives/CPFTWMY7N/p1625205022475600
Many small miners are unaware of AggregateCommits may waste their money when basefee is low. It might be better that people choose to aggregate instead of aggregate by defaut. 

Currently, on [filscout](https://filscout.com/en/gas/gas), 99% of the gas burn are from ProveCommitAggregate, which is a bit of a waste given basefee at 100 aFil.

alt text


wangyancun:
after base fee soars, you will say that it should be aggregated.
capricious human.

magik6k:
https://github.com/filecoin-project/lotus/pull/6650 adds a config for minimum network basefee below which aggregation will be disabled

PR #1843, opened 2020-05-26
Allow remote IPFS node support & IPFS for data retrieval & automatic env wiring

jsign:
This PR has three changes.

**Automatic environment variables wiring for .toml config file**
Currently, a user can only change configuration attributes in a .toml file. Now a `LOTUS_` prefix is wired to be considered for changing these attributes.
The env variable name respects the logical grouping, some examples of use:
- `LOTUS_API_LISTENADDRESS=...`
- `LOTUS_CLIENT_USEIPFS=...`
- `LOTUS_METRICS_NICKNAME=...`

Environment variables _takes precedence_ over configurations in the .toml file, as usually expected.

**Allow using a remote IPFS node as the underlying blockstore**
This extends the work done #1641. That meant using a local IPFS node as the underlying block storage for getting data to start deals.
This PRs adds two extra features:
- It allows to use of a remote IPFS node providing its multiaddress API.
- It allows to optionally also use IPFS for storing retrieved data from deals.

This is mapped in the following attributes in the .toml file:
```toml
[Client]
# UseIpfs = false
# IpfsMAddr = ""
# IpfsUseForRetrieval = false
```
(by default setup).

The `UseIpfs` flag was done in #1641. Now, if `UseIpfs=true` and want to use a remote IPFS node, its corresponding multiaddr should be set, e.g: `IpfsMAddr=/ip4/172.0.9.1/tcp/5001`.
This will use the IPFS node only as a source of retrieving data to make new deals.
In case you also want to store data in IPFS when data is retrieved from deals, set `IpfsUseForRetrieval=true` (saying it differently, IPFS is the Get&Put blockstore for Lotus).

Considering previous point feature, you can also set via envs: `LOTUS_CLIENT_USEIPFS=true`, `LOTUS_CLIENT_IPFSMADDR=...`, etc.

**Allow "fetching" data instead of necessarily retrieving it to files**
Considering the new feature `IpfsUseForRetrieval`, now allow to "fetch" a Deal data without necessarily dealing with storing the result in files but just feeding IPFS with it.
This basically means that after the call ends, the data will be available in the underlying configured IPFS node for retrieval. This is convenient to avoid dealing with files or duplicating data that we already "fed" into the IPFS node.

To use this feature, call the `ClientRetrieve` API with a `nil` `*apil.FileRef` (now a pointer).

From birds-eye, now Lotus can entirely rely on IPFS for data storage to make deals and retrieve them.

cc @whyrusleeping  @magik6k

PR #6671, opened 2021-07-03
integrate DAG store and CARv2 in deal-making

aarshkshah1992:
**_(Description added by @raulk)_**

## Technical description

This commit removes badger from the deal-making processes, and moves to a new architecture with the [dagstore](https://github.com/filecoin-project/dagstore/) as the cental component on the miner-side, and CARv2s on the client-side.

Every deal that has been handed off to the sealing subsystem becomes a shard in the dagstore. Shards are mounted via the `LotusMount`, which teaches the dagstore how to load the related piece when serving retrievals.

When the miner starts the Lotus for the first time with this patch, we will perform a **one-time migration** of all active deals into the dagstore. This is a lightweight process, and it consists simply of registering the shards in the dagstore.

Shards are backed by the unsealed copy of the piece. This is currently a CARv1. However, the dagstore keeps CARv2 indices for all pieces, so when it's time to acquire a shard to serve a retrieval, the unsealed CARv1 is joined with its index (safeguarded by the dagstore), to form a read-only blockstore, thus taking the place of the monolithic badger.

Data transfers have been adjusted to interface directly with CARv2 files. On inbound transfers (client retrievals, miner storage deals), we stream the received data into a CARv2 `ReadWrite` blockstore. On outbound transfers (client storage deals, miner retrievals), we serve the data off a CARv2 `ReadOnly` blockstore.

Client-side imports are managed by the refactored `*imports.Manager` component (when not using IPFS integration). Just like it before, we use the go-filestore library to avoid duplicating the data from the original file in the resulting UnixFS DAG (concretely the leaves). However, the target of those imports are what we call "ref-CARv2s": CARv2 files placed under the `$LOTUS_PATH/imports` directory, containing the intermediate nodes in full, and the leaves as positional references to the original file on disk.

Client-side retrievals are placed into CARv2 files in the location: `$LOTUS_PATH/retrievals`.

A new set of `Dagstore*` JSON-RPC operations and `lotus-miner dagstore` subcommands have been introduced on the miner-side to inspect and manage the dagstore.

Despite moving to a CARv2-backed system, the IPFS integration has been respected, and it continues to be possible to make storage deals with data held in an IPFS node, and to perform retrievals directly into an IPFS node.

---

## Client imports

Because the "client" Badger blockstores are no longer used, existing imports on the client will be rendered useless.

On startup, Lotus will enumerate all imports and print WARN statements on the log for each import that needs to be reimported. These log lines contain these messages:

- import lacks carv2 path; import will not work; please reimport
- import has missing/broken carv2; please reimport

At the end, we will print a "sanity check completed" message indicating the count of imports found, and how many were deemed broken.

---

### About this PR

We believe the test coverage to be pretty satisfactory, with a substantial portion of the LoC diff corresponding to tests.

This commit is a squashed version of 200+ commits developed by @aarshkshah1992, @dirkmc, and @raulk over the course of months.

This contribution was thoroughly tested in the M1 milestone with minerX.2: filecoin-project/lotus#6852.

Read more about the motivation here:
- https://github.com/protocol/web3-dev-team/pull/116
- https://github.com/filecoin-project/dagstore/blob/master/docs/design.

Co-authored-by: @aarshkshah1992 @dirkmc @raulk.

---

### TODO

- [x] Migration of existing storage client imports in Badger to CARv2 files.
    - added a warning in logs.
- [ ] Add tests for Dagstore* JSON-RPC operations.
- [ ] Add tests for expired deals and slashed deals (coverage for OnDealExpiredOrSlashed).
- [ ] Try to get rid of `StagingBlockstore`, `StagingGraphsync`.
- [ ] Try to get rid of `/client` datastore and `/staging` datastore.
- [ ] If the above is successfuly, remove the datastore muxing (try to remove the datastore muxing).


node/impl/client/client.go
```diff
@@ -483,84 +466,103 @@ func (a *API) makeRetrievalQuery(ctx context.Context, rp rm.RetrievalPeer, paylo
...
-
-	db, err := params.New(chunker.NewSizeSplitter(file, int64(build.UnixfsChunkSize)))
+	defer os.Remove(tmpPath) //nolint:errcheck
+	tmpF, err := os.Open(tmpPath)
 	if err != nil {
 		return cid.Undef, err
 	}
-	nd, err := balanced.Layout(db)
-	if err != nil {
+	defer tmpF.Close() //nolint:errcheck
```
dirkmc, line 543: I don't think we need this `defer tmpF.Close()` as we call `tmpF.Close()` immediately below

node/impl/client/client.go
```diff
@@ -763,72 +755,74 @@ func (a *API) clientRetrieve(ctx context.Context, order api.RetrievalOrder, ref
...
+			finish(err)
+			return
+		}
+
 		finish(f.Close())
 		return
 	}
 
-	nd, err := rdag.Get(ctx, order.Root)
+	rw, err := blockstore.OpenReadOnly(carV2FilePath, false)
```
dirkmc, line 811: supernit: change var name from rw to readOnly?

node/repo/importmgr/mgr.go
```diff
@@ -98,14 +116,50 @@ func (m *Mgr) Info(id multistore.StoreID) (*StoreMeta, error) {
...
+func (m *Mgr) CARV2FilePathFor(dagRoot cid.Cid) (string, error) {
+	importIDs, err := m.List()
+	if err != nil {
+		return "", xerrors.Errorf("failed to fetch import IDs: %w", err)
+	}
+
+	for _, importID := range importIDs {
+		info, err := m.Info(importID)
+		if err != nil {
+			continue
```
dirkmc, line 136: We should probably log out the error here

node/repo/importmgr/mgr.go
```diff
@@ -98,14 +116,50 @@ func (m *Mgr) Info(id multistore.StoreID) (*StoreMeta, error) {
...
+		}
+		if info.Labels[LRootCid] == "" {
+			continue
+		}
+		c, err := cid.Parse(info.Labels[LRootCid])
+		if err != nil {
+			continue
```
dirkmc, line 143: We should probably log out the error here

dirkmc (commented):
Looking really good so far 👍

api/docgen/docgen.go
```diff
@@ -88,7 +87,7 @@ func init() {
 	addExample(pid)
 	addExample(&pid)
 
-	multistoreIDExample := multistore.StoreID(50)
+	storeIDExample := uint64(50)
```
magik6k, line 90: This isn't needed if it's not its own type

api/api_full.go
```diff
@@ -712,16 +711,17 @@ type MinerSectors struct {
 
 type ImportRes struct {
 	Root     cid.Cid
-	ImportID multistore.StoreID
+	ImportID uint64
```
magik6k, line 714: Would be nice to have this still be a type, makes it more clear which things are what in all the places

node/builder_chain.go
```diff
@@ -168,9 +168,10 @@ func ConfigFullNode(c interface{}) Option {
 
 		If(cfg.Client.UseIpfs,
 			Override(new(dtypes.ClientBlockstore), modules.IpfsClientBlockstore(ipfsMaddr, cfg.Client.IpfsOnlineMode)),
-			If(cfg.Client.IpfsUseForRetrieval,
+			// TODO Fix this when we use IPFS for retrieval here.
+			/*If(cfg.Client.IpfsUseForRetrieval,
 				Override(new(dtypes.ClientRetrievalStoreManager), modules.ClientBlockstoreRetrievalStoreManager),
-			),
+			),*/
```
magik6k, line 174: This feature is used by some users (afaik powergate uses this)

node/builder_miner.go
```diff
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/filecoin-project/go-fil-markets/shared_testutil"
```
magik6k, line 7: (that's a weird package name for a DAGStore)

node/impl/client/import.go
```diff
@@ -0,0 +1,174 @@
...
+		return cid.Undef, xerrors.Errorf("failed to create temp CARv2 file: %w", err)
+	}
+	defer os.Remove(tmpCARv2Path) //nolint:errcheck
+
+	tempCARv2Store, err := blockstore.NewReadWrite(tmpCARv2Path, []cid.Cid{})
+	if err != nil {
+		return cid.Undef, xerrors.Errorf("failed to create rootless temp CARv2 rw store: %w", err)
+	}
+	defer tempCARv2Store.Finalize() //nolint:errcheck
+	bsvc := blockservice.New(tempCARv2Store, offline.Exchange(tempCARv2Store))
```
magik6k, line 46: We could avoid a lot of write i/o by using an (in-memory) filestore here. This should make imports significantly faster - We'd only do sequential reading here, and likely get most of the file cached in RAM, so in the second pass we'd mostly do sequential writing (yes, SSDs are pretty good at random io too, but most are really bad at mixing reading and writing)

node/impl/client/import.go
```diff
@@ -0,0 +1,174 @@
...
+	switch hd.Version {
+	case 2:
+
+		// This is a CARv2, we can import it as it is by simply copying it.
+		outF, err := os.Open(outputCARv2Path)
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("failed to open output CARv2 file: %w", err)
+		}
+		defer outF.Close() //nolint:errcheck
+		_, err = io.Copy(outF, inputF)
```
magik6k, line 158: We should try to avoid copying if we don't have to

node/repo/importmgr/mgr.go
```diff
@@ -98,14 +116,50 @@ func (m *Mgr) Info(id multistore.StoreID) (*StoreMeta, error) {
...
+		if c.Equals(dagRoot) {
+			return info.Labels[LCARv2FilePath], nil
+		}
+	}
+
+	return "", nil
+}
+
+func (m *Mgr) NewTempFile(id uint64) (string, error) {
+	file, err := ioutil.TempFile(m.repoPath, fmt.Sprintf("%d", id))
```
magik6k, line 154: `ioutil.TempFile` already appends a random number at the end, I'd just do
```suggestion
func (m *Mgr) NewTempFile() (string, error) {
	file, err := ioutil.TempFile(m.repoPath, fmt.Sprintf("import-temp-", id))
```

magik6k: This should also be a subdirectory in the repo, so users can mount it to a separate volume easily

magik6k (commented):
((looks like `itests/fstmp889313799` got committed by accident))

node/builder_miner.go
aarshkshah1992: @magik6k This is a mocked out DAG Store. See the TODO. We will soon be integrating the [actual DAGStore]([url](https://github.com/filecoin-project/dagstore)).

node/repo/importmgr/mgr.go
aarshkshah1992: @magik6k Should we allow users to pass in their subdir path for this via the Config ?

node/impl/client/import.go
aarshkshah1992: @magik6k Can you point me to some code that uses Filestores to do create a Unixfs DAG ? Not sure what you mean here.

node/impl/client/import.go
aarshkshah1992: @magik6k Unfortunately, we need to copy the file here as we need to take ownership of the imported data and need to have our own CARv2 copies of the data the user is importing as we can't rely on them to keep the original around.

This is similar to the existing flow, except that instead of copying and importing to Badger, we're copying to a CARv2 file.

node/impl/client/client.go
aarshkshah1992: Done.

node/repo/importmgr/mgr.go
aarshkshah1992: Done.

node/repo/importmgr/mgr.go
aarshkshah1992: Done.

api/docgen/docgen.go
aarshkshah1992: Removed.

api/api_full.go
aarshkshah1992: Have fixed this by introducing a new `ImportID` type.

node/builder_miner.go
aarshkshah1992: Having said that, even this has been fixed.

node/builder_chain.go
aarshkshah1992: @magik6k Have fixed this. We allow a retrieval client to write the retrieved CARv2 files to the IPFS node backed blockstore confgiured here. However, this "write retrieved CARv2 to the blockstore" is only done for this special `Client.IpfsUseForRetrieval` case.

node/impl/client/import.go
aarshkshah1992: @magik6k I've addressed this comment by using a filestore backed by an in-memory blockstore in the first pass here. However, I have some concerns around cleaning up this temporary filestore that I've marked as a `TODO` in the code. 
For the cleanup, should we simply iterate over all the keys the filestore has and delete them or is there more to it ?

node/impl/client/import.go
magik6k: In the existing flow we're using filestore, which is relying on users keeping original files around. I think it's ok to keep this behavior. We could add a param to the deal api/cli to let users decide whether or not to copy the data.

node/repo/importmgr/mgr.go
magik6k: It can just be a hardcoded directory in the repo - users will be able to link/mount it to different volumes

node/impl/client/import.go
```diff
@@ -0,0 +1,174 @@
...
+func (a *API) importNormalFileToCARv2(ctx context.Context, importID importmgr.ImportID, inputFilePath string, outputCARv2Path string) (c cid.Cid, finalErr error) {
+
+	// TODO: We've currently put in a hack to create the Unixfs DAG as a CARv2 without using Badger.
+	// We first create the Unixfs DAG using a filestore to get the root of the Unixfs DAG.
+	// We can't create the UnixfsDAG right away using a CARv2 read-write blockstore as the blockstore
+	// needs the root of the DAG during instantiation to write out a valid CARv2 file.
+	//
+	// In the second pass, we create a CARv2 file with the root present using the root node we get in the above step.
+	// This hack should be fixed when CARv2 allows specifying the root AFTER finishing the CARv2 streaming write.
+	// TODO How do we clean this up ?
```
magik6k, line 41: Given that it's all in-memroy, I'd think this just gets GCed (assuming that things aren't accidentally referenced anywhere)

hannahhoward (commented):
This seems to be writing CAR v2 files to the hard drive on the client side. I thought we'd discussed avoiding storing a second copy of the clients data on the HDD.  @magik6k are you ok with every import for regular files on the client having a CarV2 copy on the hard drive (which will of course be 32 additional GB in many cases). 

The current code jumped through a lot of hoops to avoid ever storing a CAR file on the client's machine, and always using OS-pipes when we did file i/o needed for CommP calculation. So I was wondering if the intent is in fact to no longer do that.

itests/deals_concurrent_test.go
```diff
@@ -63,7 +63,7 @@ func TestDealWithMarketAndMinerNode(t *testing.T) {
 		ns := fmt.Sprintf("%d", n)
 		t.Run(ns+"-fastretrieval-CAR", func(t *testing.T) { runTest(t, n, true, true) })
 		t.Run(ns+"-fastretrieval-NoCAR", func(t *testing.T) { runTest(t, n, true, false) })
-		t.Run(ns+"-stdretrieval-CAR", func(t *testing.T) { runTest(t, n, true, false) })
+		t.Run(ns+"-stdretrieval-CAR", func(t *testing.T) { runTest(t, n, false, true) })
```
aarshkshah1992, line 66: @dirkmc Why have we made this change ?

itests/deals_concurrent_test.go
```diff
@@ -102,7 +102,7 @@ func TestDealCyclesConcurrent(t *testing.T) {
...
```
aarshkshah1992, line 105: @dirkmc Why have we made this change ?

raulk (commented):
@aarshkshah1992 @dirkmc I think there's a misunderstanding in the changes to `client import`; this part should be mostly unchanged, as @hannahhoward and @magik6k pointed out above.

IUC, the **current** state of the system is as follows:

1. User provides a raw file.
2. Lotus instantiates a new Filestore using the multistore. The multistore combines two things: an actual blockstore (Badger in this case) and a CID => file offset/length tracker. See https://github.com/filecoin-project/go-multistore/blob/master/store.go#L33. There's special-casing for a special type of block called [`PosInfo`](https://github.com/ipfs/go-ipfs-posinfo) that encodes the offset of the block within the file, and its length.
    - What's not clear to me is how Lotus client import leverages that special casing, because we seem to use a normal DAG service that feeds normal IPLD blocks to the blockstore. @hannahhoward -- illuminate?
3. When chunking, we do not import blocks into the Badger blockstore, but somehow the PosInfo route is used, so we record CID => (offset, length) in the datastore, thus avoiding creating a copy of the file.

Later, when performing the deal, we simulate forming the CARv1 and feed it to the commP writer as a stream, to calculate the commP. But the CARv1 is never actually written on disk.

The footprint of this operation is just the cost of the CID => (offset, length) positional tracking, which likely wouldn't exceed 3MiB for a 64GiB file chunked with the 1MiB UnixFS chunking policy.

The Graphsync transfer is conducted using the positional mapping (it does not use a Badger store), which is why the original file is needed.

Conversely, this PR **introduces** these changes:

1. User provides a raw file.
2. File is chunked into a CARv2 to calculate its root.
3. Another CARv2 is generated with the right root.

For a 64GiB raw file, the footprint of the new code is in the order of 128GiB, which is a regression we can't afford.

---

I suggest to revert the changes in the `client import` process.

itests/deals_concurrent_test.go
dirkmc: This fixes a bug in the test - the name of the test is `stdretrieval-CAR` but the parameters are wrong so the test duplicates `fastretrieval-NoCAR`

node/impl/client/client.go
aarshkshah1992: The copy statement before that could fail.

node/impl/client/import.go
aarshkshah1992: This has been done.

node/impl/client/import.go
aarshkshah1992: This has been fixed. We no longer copy the file.

node/builder_miner.go
```diff
@@ -148,7 +149,9 @@ func ConfigStorageMiner(c interface{}) Option {
 			Override(new(dtypes.RetrievalPricingFunc), modules.RetrievalPricingFunc(cfg.Dealmaking)),
 
 			// DAG Store
-			Override(new(dagstore.DagStore), modules.DAGStore),
+			Override(new(*mount.Registry), modules.DAGStoreRegistry),
```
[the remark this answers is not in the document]
aarshkshah1992: This no longer exists here.

node/builder_miner.go
```diff
@@ -148,7 +149,9 @@ func ConfigStorageMiner(c interface{}) Option {
...
+			Override(new(*dagstore.DAGStore), modules.DAGStore),
+			Override(new(mktdagstore.DagStoreWrapper), modules.DAGStoreWrapper),
```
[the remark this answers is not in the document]
aarshkshah1992: This is done.

node/modules/storageminer.go
```diff
@@ -596,15 +575,32 @@ func BasicDealFilter(user dtypes.StorageDealFilter) func(onlineOk dtypes.Conside
...
+	dagStoreDS := namespace.Wrap(ds, datastore.NewKey("/dagstore/provider"))
+	cfg := mktdagstore.MarketDAGStoreConfig{
+		TransientsDir: filepath.Join(dagStoreDir, "transients"),
+		IndexDir:      filepath.Join(dagStoreDir, "index"),
+		Datastore:     dagStoreDS,
+	}
+	mountApi := mktdagstore.NewLotusMountAPI(pieceStore, rpn)
+	return mktdagstore.NewDagStoreWrapper(cfg, mountApi)
+}
+
```
raulk, line 594: - I think Lotus should be owning and instantiating the DAG store via dependency injection. 
    - Having markets own this important part of the system makes it escape dependency injection, and it also makes it way harder for other parts of the system (e.g. CLI, JSON-RPC) to interact with the DAG store.
    - The DAG store should be a top-level Lotus component on the markets node.
- We can get rid of the `MarketDAGStoreConfig` struct, and just use `dagstore.Config`.
- The DAG store transients and index dir should be provided by Lotus configuration, falling back to defaults if unset. It's very likely that users will want to set these paths manually.
- We need to create a Lotus config section for DAG store parameters (including throttling levels!) 
- I _think_ that the Lotus mount should live in lotus because it's Lotus-specific (miner APIs are private and not standardised AFAIK), so this component is not reusable outside of Lotus.

aarshkshah1992: @dirkmc is working on PRs for these.

dirkmc: PR to move lotus-specific code to lotus:
- [x] go-fil-markets: https://github.com/filecoin-project/go-fil-markets/pull/578
- [ ] lotus

dirkmc: @raulk please take a look at the above two PRs

I haven't made many changes to the code that is copied over the lotus. In particular I've left `MarketDAGStoreConfig` in place. I'm not sure if it makes sense to replace it with `dagstore.Config` because if we use the intermediate struct then we can hide the details of the DAG store from the rest of the codebase. eg the registry and failure channel are created inside `NewDagStoreWrapper`.

Let's continue the discussion in the PR at https://github.com/filecoin-project/lotus/pull/6793

dirkmc (approved):

node/impl/client/import.go
```diff
@@ -0,0 +1,122 @@
...
+	// We can't create the UnixfsDAG right away using a CARv2 read-write blockstore as the blockstore
+	// needs the root of the DAG during instantiation to write out a valid CARv2 file.
+	//
+	// In the second pass, we create a CARv2 file with the root present using the root node we get in the above step.
+	// This hack should be fixed when CARv2 allows specifying the root AFTER finishing the CARv2 streaming write.
+	fm := filestore.NewFileManager(ds_sync.MutexWrap(datastore.NewMapDatastore()), "/")
+	fm.AllowFiles = true
+	fstore := filestore.NewFilestore(bstore.NewMemorySync(), fm)
+	bsvc := blockservice.New(fstore, offline.Exchange(fstore))
+	defer bsvc.Close() //nolint:errcheck
```
raulk, line 42: I'm worried about the scalability of the in-memory import, even if we are just storing refs and intermediate nodes. In practice it won't be more than a few MBs I think, but since we have writeable CAR blockstore, why not use those to reduce the memory footprint?

node/impl/client/import.go
raulk: This part was changed such that the first step is no longer carried out in memory.

markets/storageadapter/provider.go
```diff
@@ -57,10 +58,12 @@ type ProviderNodeAdapter struct {
 	maxDealCollateralMultiplier uint64
 	dsMatcher                   *dealStateMatcher
 	scMgr                       *SectorCommittedManager
+
+	rpn retrievalmarket.RetrievalProviderNode
 }
 
-func NewProviderNodeAdapter(fc *config.MinerFeeConfig, dc *config.DealmakingConfig) func(mctx helpers.MetricsCtx, lc fx.Lifecycle, dag dtypes.StagingDAG, secb *sectorblocks.SectorBlocks, full v1api.FullNode, dealPublisher *DealPublisher) storagemarket.StorageProviderNode {
-	return func(mctx helpers.MetricsCtx, lc fx.Lifecycle, dag dtypes.StagingDAG, secb *sectorblocks.SectorBlocks, full v1api.FullNode, dealPublisher *DealPublisher) storagemarket.StorageProviderNode {
+func NewProviderNodeAdapter(fc *config.MinerFeeConfig, dc *config.DealmakingConfig) func(mctx helpers.MetricsCtx, lc fx.Lifecycle, dag dtypes.StagingDAG, secb *sectorblocks.SectorBlocks, full v1api.FullNode, dealPublisher *DealPublisher, rpn retrievalmarket.RetrievalProviderNode) storagemarket.StorageProviderNode {
```
hannahhoward, line 65: So, as best I understand, we have the following:

Intention: RetrievalProviderNode is intended to be a facade interface for the retrieval provider code in go-fil-markets to talk to either a markets process or miner process for functionality not provided in the markets library. It's just an interface definition. It should include only methods used by retrieval code in go-fil-markets.

Currently: RetrievalProviderNode provides two functions:
IsUnsealed
UnsealSector

UnsealSector is not used by retrieval market code, and thus, should not be part of RetrievalProviderNode.

UnsealSector is however required by markets/dagstore/MinerAPI

IsUnsealed is used by markets/dagstore/MinerAPI AND by the implementation of StorageProviderNode -- but IsUnsealed is not actually used on the StorageProviderNode in the markets implementation -- it's not clear why it's added.

Both MinerAPI and StorageProviderNode depend on RetrievalProviderNode.... that seems super awkward.

It seems like we have some untangling to do.

First, RetrievalProviderNode, and StorageMarketNode are just interfaces -- they're defined at the site of use, in go-fil-markets, per Go lang best practices (see https://github.com/golang/go/wiki/CodeReviewComments#interfaces). There's no need to get attached to that collection of methods as having any inherent meaning for code in this repo, other than you have to put together something that implements everything requires by markets code in order to initialize a storage provider / retrieval provider. 

My recommendations are:
- Remove UnsealSector from RetrievalProviderNode
- Remove IsUnsealed from StorageProviderNode

- From markets/retrievaladapter/provider, extract IsUnsealed and UnsealSector implementations into a common dependency that can embed in MinerAPI & markets/retrievaladapter/provider (you can just have it be an embedded struct to avoid the one line methods)
- You can just use this in StorageProviderNode if you need to add IsUnsealed at some later point.

hannahhoward (changes requested):
In general, awesome! Yay for this huge change. Most of my comments are suggestions.

However, I have three blocking comments:
1. The use of RetrievalProviderNode as a dependency to get the IsUnsealed/UnsealSector methods is super awkward and the dependencies need untangling. I've put suggestions on how to do it.
2. The IPFS blockstore integration is super awkward for retrieval, and I'm pretty sure it's broken completely for storage. We either need to fix it or work closely with partners before we ship with this feature removed. I know PowerGate is the main user, I don't know how important that is.
3. Unless I am reading the code wrong, it appears that we are still writing second full copies of files out to the hard drive when we import them -- see comments on `doImport` in `node/impl/client/import.go`, despite the existence of ReadWriteFilestore which certainly appears like it could be used to avoid doing these writes. This isn't blocking if someone can explain to me why I'm reading the code wrong -- it's fine if I just am not seeing it.

I wonder: Blocking issues 2 & 3 have to do with the person on the client side of deals, not the miner. What is our testing plan for working with this code with people who are making deals? (cause I know we've been focused on beta testing with miners). I realize the dealbot is testing the dealmaking, but it also has space do deal with extra file copies AND it doesn't test the IPFS blockstore functionality.

blockstore/badger/blockstore.go
```diff
@@ -296,68 +297,96 @@ func (b *Blockstore) movingGC() error {
...
 	log.Info("copying blockstore")
 	err = b.doCopy(b.db, b.dbNext)
 	if err != nil {
-		return fmt.Errorf("error moving badger blockstore to %s: %w", path, err)
+		return fmt.Errorf("error moving badger blockstore to %s: %w", newPath, err)
 	}
 
 	b.lockMove()
-	db1 := b.db
+	dbOld := b.db
```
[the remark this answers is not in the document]
raulk: This is a side-effect from merging master => release/1.11.x => this branch, and GitHub not being able to distinguish that this code is already in master. It's an artefact of the diff. We will open a new squashed PR to clean up the interface once we address these comments and resolve all threads here.

cli/auth.go
```diff
@@ -113,7 +113,7 @@ var AuthApiInfoToken = &cli.Command{
 
 		ti, ok := cctx.App.Metadata["repoType"]
 		if !ok {
-			log.Errorf("unknown repo type, are you sure you want to use GetAPI?")
+			log.Errorf("unknown repo type, are you sure you want to use GetCommonAPI?")
```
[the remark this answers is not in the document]
raulk: This is a side-effect from merging master => release/1.11.x => this branch, and GitHub not being able to distinguish that this code is already in master. It's an artefact of the diff. We will open a new squashed PR to clean up the interface once we address these comments and resolve all threads here.

api/api_storage.go
```diff
@@ -332,3 +379,34 @@ type DealSchedule struct {
...
+// DagstoreShardInfo is the serialized form of dagstore.DagstoreShardInfo that
+// we expose through JSON-RPC to avoid clients having to depend on the
+// dagstore lib.
+type DagstoreShardInfo struct {
+	Key   string
+	State string
+	Error string
+}
+
+// DagstoreShardResult enumerates results per shard.
```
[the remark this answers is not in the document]
raulk: Correct. Context: this was shared with the InitializeAll operation, but then that op required a richer interface. Can rename back.

cli/util/api.go
```diff
@@ -2,6 +2,7 @@ package cliutil
```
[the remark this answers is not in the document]
raulk: This is a side-effect from merging master => release/1.11.x => this branch, and GitHub not being able to distinguish that this code is already in master. It's an artefact of the diff. We will open a new squashed PR to clean up the interface once we address these comments and resolve all threads here.

node/modules/client.go
```diff
@@ -173,13 +151,13 @@ func NewClientDatastore(ds dtypes.MetadataDS) dtypes.ClientDatastore {
 	return namespace.Wrap(ds, datastore.NewKey("/deals/client"))
 }
 
-func StorageClient(lc fx.Lifecycle, h host.Host, ibs dtypes.ClientBlockstore, mds dtypes.ClientMultiDstore, r repo.LockedRepo, dataTransfer dtypes.ClientDataTransfer, discovery *discoveryimpl.Local, deals dtypes.ClientDatastore, scn storagemarket.StorageClientNode, j journal.Journal) (storagemarket.StorageClient, error) {
+func StorageClient(lc fx.Lifecycle, h host.Host, dataTransfer dtypes.ClientDataTransfer, discovery *discoveryimpl.Local, deals dtypes.ClientDatastore, scn storagemarket.StorageClientNode, j journal.Journal) (storagemarket.StorageClient, error) {
```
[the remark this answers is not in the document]
raulk: Hum, do they? Are they instantiating Lotus programmatically and overriding this with their own blockstore through DI? Do we have concrete users that use this? It's a very hacky thing to tap into implementation details through DI to serve an external use case (we should be having a dedicated API for this if we expect this case to be supported externally going forward).

We thought this blockstore was used only internally. We absolutely don't want to break existing clients. Can we identify who these clients are?

node/impl/client/client.go
```diff
@@ -777,40 +779,58 @@ func (a *API) clientRetrieve(ctx context.Context, order api.RetrievalOrder, ref
...
-		// local retrieval
-		st, err := ((*multistore.MultiStore)(a.Mds)).Get(*order.LocalStore)
+		carV2FilePath = order.LocalCARV2FilePath
+	}
+
+	// TODO We only support this currently for the IPFS Retrieval use case
+	// where users want to write out filecoin retrievals directly to IPFS.
+	// If users haven' configured the Ipfs retrieval flag, the blockstore we get here will be a "no-op" blockstore.
+	// write out the CARv2 file to the retrieval block-store (which is really an IPFS node behind the scenes).
+	rs, err := a.RetrievalStoreMgr.NewStore()
```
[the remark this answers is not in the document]
raulk: @hannahhoward we thought about that (it was my preferred approach too), but it won't work well with restarts. On a restart and a transfer resumption, go-fil-markets needs to re-open the CAR (that's why it also saves the path in state). We _could_ choose to keep the multistore abstraction as a "router", and have go-fil-markets call into it when it needs the blockstore, but that feels even more awkward.

node/modules/client.go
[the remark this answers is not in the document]
ribasushi: @raulk I believe Hannah is referring to this: https://github.com/filecoin-project/lotus/blob/v1.11.1-rc2/node/builder_chain.go#L169-L174

Textile virtually never uses the lotus-local blockstore, and @dkkapur and I use the IPFS linkage in various cases for slingshot work ( the flow is `lotus client retrieve ... /dev/null` and then pulling whatever one needs from the IPFS instance )

node/builder_chain.go
```diff
@@ -169,7 +169,7 @@ func ConfigFullNode(c interface{}) Option {
 		If(cfg.Client.UseIpfs,
 			Override(new(dtypes.ClientBlockstore), modules.IpfsClientBlockstore(ipfsMaddr, cfg.Client.IpfsOnlineMode)),
```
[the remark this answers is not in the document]
ribasushi: Nod, we can't lose this, good catch!

node/modules/client.go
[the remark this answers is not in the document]
ribasushi: Original context: https://github.com/filecoin-project/lotus/pull/1843

markets/storageadapter/provider.go
dirkmc: Agree with all of this 👍

node/modules/client.go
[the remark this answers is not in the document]
raulk: Thoughts here after discussing with @dirkmc off band:

1. We attempt to make the go-fil-markets API deal with blockstores only for storage and retrievals, removing all path and CAR file management from that library. This was already identified in previous reviews as desirable.
2. This means removing all CAR paths from deal state objects in go-fil-markets.
3. This complicates resumption on restart; before go-fil-markets used to have the CAR path so it could resume autonomously. Now it will need to ask to reopen the blockstore through some interface (possibly a retrieval/storage node adaptor, or the existing getter/fetcher?).
4. When the node is configured to use the IPFS Blockstore, the interface would always return the blockstore instead of CAR files.
    - We assume that `lotus client retrieve` calls with an `<output_path>` do not use the path in this case? Is this even handled now? Calls with an output path should be rejected.
5. To avoid breakage of existing clients, `--car` should unpack the CARv1 from within the CARv2. Right now we'll need to do this through an inefficient copy, until https://github.com/ipld/go-car/issues/207 is implemented.
    - Please add a TODO in code pointing to that issue.
    - Alternatively, we could just choose to break backwards compatibility and return a CARv2.
    - If we choose not to, we should add a `--carv2` flag that just returns a CARv2.

@dirkmc ^^ can you confirm if this captures what we discussed?

markets/storageadapter/provider.go
dirkmc: Implemented the change in
- [x] markets https://github.com/filecoin-project/go-fil-markets/pull/601
- [ ] lotus

cmd/lotus-miner/dagstore.go
```diff
@@ -0,0 +1,215 @@
...
+	Subcommands: []*cli.Command{
+		dagstoreListShardsCmd,
+		dagstoreInitializeShardCmd,
+		dagstoreRecoverShardCmd,
+		dagstoreInitializeAllCmd,
+		dagstoreGcCmd,
+	},
+}
+
+var dagstoreListShardsCmd = &cli.Command{
```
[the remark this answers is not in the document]
raulk: > (otherwise default is color)

That's not how I'm reading that reference...

markets/dagstore/miner_api.go
```diff
@@ -0,0 +1,184 @@
+package dagstore
```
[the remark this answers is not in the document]
raulk: The reason why it's here it's because it's specific to Lotus and we have been working under the assumption that go-fil-markets is **not** Lotus specific.

node/impl/client/import.go
```diff
@@ -0,0 +1,144 @@
...
+		return cid.Undef, xerrors.Errorf("failed to import file to store to compute root: %w", err)
+	}
+
+	if err := fstore.Close(); err != nil {
+		return cid.Undef, xerrors.Errorf("failed to finalize car filestore: %w", err)
+	}
+
+	// Step 2. We now have the root of the UnixFS DAG, and we can write the
+	// final CAR for real under `dst`.
+	bs, err := blockstore.OpenReadWrite(dst, []cid.Cid{root},
```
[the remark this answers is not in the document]
raulk: Yeah, that looks wrong (my fault), and thanks for catching.

cmd/lotus-miner/dagstore.go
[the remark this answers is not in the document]
raulk: So this sent me down an investigation on how we do terminal detection in Lotus. What you thought is not quite correct. This flag __forces__ colouring, as a way of escaping the terminal detection we already do in the `init()` function of the `cli` package. I think this is all a bit messed up, but I'll play by the game anyway.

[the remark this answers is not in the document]
raulk: Fixed in [`608f6d1` (#6671)](https://github.com/filecoin-project/lotus/pull/6671/commits/608f6d1fc616bfc313d308fdb92ff6e76652fa33).

cmd/lotus-miner/dagstore.go
```diff
@@ -0,0 +1,215 @@
...
+
+			case <-ctx.Done():
+				return fmt.Errorf("aborted")
+			}
+		}
+	},
+}
+
+var dagstoreGcCmd = &cli.Command{
+	Name:  "gc",
```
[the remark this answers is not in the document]
raulk: Fixed in [`608f6d1` (#6671)](https://github.com/filecoin-project/lotus/pull/6671/commits/608f6d1fc616bfc313d308fdb92ff6e76652fa33).

api/api_storage.go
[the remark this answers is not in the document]
raulk: Actually, I'm going to keep it this way since in the future we may add operations that do require reporting per shard.

node/modules/client.go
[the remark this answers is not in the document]
raulk: @dirkmc ^^ since this got implemented quickly, are we ready to adopt it?

node/impl/client/client.go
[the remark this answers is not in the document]
raulk: Update here: yes, @dirkmc is refactoring this circuitry.

hannahhoward (approved):
Pre-emptive approval to unblock merge while I'm on vacation. Will review tomorrow to see if there are any changes before I leave.

dirkmc:
@raulk need to audit if blockstore provider to loader and storer in node/modules/graphsync.go are actually being used

node/impl/client/client.go
```diff
@@ -495,84 +484,120 @@ func (a *API) makeRetrievalQuery(ctx context.Context, rp rm.RetrievalPeer, paylo
...
-
-	id, st, err := a.imgr().NewStore()
+func (a *API) ClientRemoveImport(ctx context.Context, importID imports.ID) error {
+	info, err := a.imgr().Info(importID)
 	if err != nil {
-		return cid.Undef, err
+		return xerrors.Errorf("failed to fetch import info: %w", err)
 	}
-	if err := a.imgr().AddLabel(id, "source", "import-local"); err != nil {
-		return cid.Cid{}, err
```
raulk, line 538: @aarshkshah1992 removing this label causes a functional regression.

node/impl/client/client.go
```diff
@@ -495,84 +484,120 @@ func (a *API) makeRetrievalQuery(ctx context.Context, rp rm.RetrievalPeer, paylo
...
 	}
 
-	bufferedDS := ipld.NewBufferedDAG(ctx, st.DAG)
+	return a.imgr().Remove(importID)
+}
 
-	prefix, err := merkledag.PrefixForCidVersion(1)
+func (a *API) ClientImportLocal(ctx context.Context, r io.Reader) (cid.Cid, error) {
+	// write payload to temp file
+	tmpPath, err := a.imgr().NewTempFile(imports.ID(rand.Uint64()))
```
raulk, line 569: We should be using an actual ID generated by the import manager.

node/impl/client/client.go
```diff
@@ -495,84 +484,120 @@ func (a *API) makeRetrievalQuery(ctx context.Context, rp rm.RetrievalPeer, paylo
...
-	db, err := params.New(chunker.NewSizeSplitter(file, int64(build.UnixfsChunkSize)))
+	defer os.Remove(tmpPath) //nolint:errcheck
+	tmpF, err := os.Open(tmpPath)
 	if err != nil {
 		return cid.Undef, err
 	}
-	nd, err := balanced.Layout(db)
-	if err != nil {
+	defer tmpF.Close() //nolint:errcheck
+	if _, err := io.Copy(tmpF, r); err != nil {
```
raulk, line 579: This is a performance regression. We just replaced direct UnixFS chunking into a "heavy" CAR file (containing leaf data) with two files (the temp file where we wrote the upload, and the position CAR file). I don't think that's right.

raulk: I think we should import straight into a heavy CAR.

node/impl/client/client.go
```diff
@@ -1046,113 +1119,39 @@ func (a *API) ClientDealPieceCID(ctx context.Context, root cid.Cid) (api.DataCID
 }
 
 func (a *API) ClientGenCar(ctx context.Context, ref api.FileRef, outputPath string) error {
-	id, st, err := a.imgr().NewStore()
+	id := imports.ID(rand.Uint64())
+	tmp, err := a.imgr().NewTempFile(id)
```
raulk, line 1123: Why is this even on the ImportMgr, if it's not managed by it nor it uses an actual ID.

raulk (approved):

cli/client.go
```diff
@@ -174,18 +174,18 @@ var clientDropCmd = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
-		var ids []multistore.StoreID
+		var ids []uint64
```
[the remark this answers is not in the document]
raulk: Yeah, it's not worth creating a dependency from the CLI package to something more deeply nested.

node/builder_chain.go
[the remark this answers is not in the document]
raulk: Fixed.

node/modules/client.go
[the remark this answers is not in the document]
raulk: Fixed.

node/impl/client/client.go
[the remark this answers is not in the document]
raulk: Fixed.

node/impl/client/client.go
raulk: Fixed.

node/impl/client/client.go
raulk: Fixed.

node/impl/client/client.go
raulk: Fixed.

node/impl/client/client.go
raulk: Fixed.

raulk (approved):

go.mod
```diff
@@ -127,39 +128,40 @@ require (
...
+	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
 	golang.org/x/tools v0.1.5
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
 	gopkg.in/cheggaaa/pb.v1 v1.0.28
 	gotest.tools v2.2.0+incompatible
-	honnef.co/go/tools v0.0.1-2020.1.3 // indirect
 )
 
+replace github.com/multiformats/go-multihash => github.com/multiformats/go-multihash v0.0.14
```
Stebalien, line 163: Please do _not_ do this without a very good reason. If you do, leave a comment explaining why. As far as I can tell, there was no reason to do this and it makes it impossible to update other dependencies.

magik6k (commented):
Post merge review pass - looks really good, only 3 nitpicks

Issue #6678, opened 2021-07-04
potential edge case in dynamic retrieval pricing


raulk:
A few retrieval deals made by the dealbot with the same miner failed with the following. Recording here to analyse further:

```
got error in offer: retrieval query offer errored: failed to price deal: GetRetrievalPricingInput:
failed to look up deal 1912670 on chain: err=deal 1912670 not found - deal may not have completed
sealing before deal proposal start epoch, or deal may have been slashed: deal 1912670 not found -
deal may not have completed sealing before deal proposal start epoch, or deal may have been slashed`
```

This may be a legitimate and genuine case of the miner failing to seal the storage deal on time, but let's analyse it.

aarshkshah1992:
@raulk Have filed https://github.com/filecoin-project/lotus/pull/6679. We shouldn't error out the retrieval pricing input function if we fail to lookup the state of a storage deal because we could have other storage deals for the Piece we're trying to retrieve.

PR #6679, opened 2021-07-05
Get retrieval pricing input should not error out on a deal state fetch

aarshkshah1992:
Closes https://github.com/filecoin-project/lotus/issues/6678.

We iterate over all storage deals for a payload to see if we have a verified storage deal with the required PieceCID to fetch the input for the retrieval pricing function. This PR introduces a fix so we don't error out in we fail to look up a deal state and just continue iterating over the other deals.

### DIFF
--- a/markets/retrievaladapter/provider.go
+++ b/markets/retrievaladapter/provider.go
@@ -5,6 +5,7 @@ import (
 	"io"
 
 	"github.com/filecoin-project/lotus/api/v1api"
+	"github.com/hashicorp/go-multierror"
 	"golang.org/x/xerrors"
 
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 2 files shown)


markets/retrievaladapter/provider.go
```diff
@@ -138,7 +138,8 @@ func (rpn *retrievalProviderNode) GetRetrievalPricingInput(ctx context.Context,
 	for _, dealID := range storageDeals {
 		ds, err := rpn.full.StateMarketStorageDeal(ctx, dealID, tsk)
 		if err != nil {
-			return resp, xerrors.Errorf("failed to look up deal %d on chain: err=%w", dealID, err)
+			log.Warnf("failed to look up deal %d on chain: err=%w", dealID, err)
```
dirkmc, line 141: I suggest that if attempting to fetch the state for all deals fails, we return the error (instead of just logging it) so it's clearer what the root cause is

aarshkshah1992: Fixed this. Thanks.

markets/retrievaladapter/provider.go
```diff
@@ -135,10 +135,14 @@ func (rpn *retrievalProviderNode) GetRetrievalPricingInput(ctx context.Context,
 	}
 	tsk := head.Key()
 
+	var lastErr error
```
raulk, line 138: Let's have this be a multierror, so we can accumulate all errors instead of just the last one.

raulk:
@aarshkshah1992 lint fails.

aarshkshah1992:
@raulk Have addressed your review and lint is now green.

aarshkshah1992:
Note that the CI test failure is a known flaky.

dirkmc (approved):

magik6k (approved):

PR #6688, opened 2021-07-06
fix "lotus-seed genesis car" error "merkledag: not found"

turuslan:
`actor.code` cids of type `v1 - raw - identity` were breaking car generation.


magik6k (approved):

magik6k:
Thanks for the PR!

PR #6696, opened 2021-07-07
Enable color by default only if os.Stdout is a TTY

ribasushi:


cli/util/color.go
```diff
@@ -0,0 +1,9 @@
+package cliutil
+
+import (
+	"os"
+
+	"github.com/mattn/go-isatty"
+)
+
+var DefaultColorUse = isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd())
```
magik6k, line 9: Can we also have an env var for this? (`LOTUS_COLOR=1`, or sth like that, unless there is already a standard env var for that kind of thing)

Just `isatty.IsTerminal` will return false when running stuff in `watch --color ..` for example.

ribasushi: So... we could introduce `LOTUS_COLOR=1` *OR* we could recognize `GOLOG_LOG_FMT=color` which is [a thing from `ipfs/go-log`](https://github.com/ipfs/go-log/pull/116#issuecomment-873779808)

I can go either way, just suggesting an alternative for smaller amount of options to keep track of...

@magik6k say the word

ribasushi: @Stebalien please be weighing in as well, I surely missed other stuff here too ;)

magik6k: `GOLOG_LOG_FMT` works for me

ribasushi: @magik6k please be checking https://github.com/filecoin-project/lotus/pull/6696/commits/4dd59b566ced

magik6k (approved):

Issue #6676, opened 2021-07-04
[BUG] test-itest-ccupgrade has a >50% spurious falure rate in CI


ribasushi:
**Describe the bug**

The `test-itest-ccupgrade` CI task has a >50% spurious faliure rate. Specific error is:
```
...
2021-07-04T07:39:25.130Z	�[34mINFO�[0m	markets	loggers/loggers.go:25	retrieval client event	{"name": "ClientEventPaymentChannelReady", "deal ID": "1625384357364895259", "state": "DealStatusPaymentChannelAllocatingLane", "message": ""}
2021-07-04T07:39:25.134Z	�[34mINFO�[0m	markets	loggers/loggers.go:25	retrieval client event	{"name": "ClientEventLaneAllocated", "deal ID": "1625384357364895259", "state": "DealStatusOngoing", "message": ""}
2021-07-04T07:39:25.134Z	�[34mINFO�[0m	markets	loggers/loggers.go:25	retrieval client event	{"name": "ClientEventLastPaymentRequested", "deal ID": "1625384357364895259", "state": "DealStatusFundsNeededLastPayment", "message": ""}
2021-07-04T07:39:25.134Z	�[34mINFO�[0m	markets	loggers/loggers.go:25	retrieval client event	{"name": "ClientEventSendFunds", "deal ID": "1625384357364895259", "state": "DealStatusSendFundsLastPayment", "message": ""}
2021-07-04T07:39:25.140Z	�[33mWARN�[0m	data_transfer_network	network/libp2p_impl.go:277	error setting deadline: set pipe: deadline not supported
2021-07-04T07:39:25.140Z	�[33mWARN�[0m	data_transfer_network	network/libp2p_impl.go:293	error resetting deadline: set pipe: deadline not supported
2021-07-04T07:39:25.141Z	�[34mINFO�[0m	markets	loggers/loggers.go:25	retrieval client event	{"name": "ClientEventPaymentSent", "deal ID": "1625384357364895259", "state": "DealStatusFinalizing", "message": ""}
2021-07-04T07:39:25.141Z	�[33mWARN�[0m	paych	paychmgr/paych.go:349	AddVoucher: voucher re-added
[...]

PR #6697, opened 2021-07-07
Config for deal publishing control addresses

magik6k:
This might be useful when running with a separate market miner node - will let miners run a separate lotus daemon without the worker key

Also fixes https://github.com/filecoin-project/lotus/issues/6676


itests/kit/control.go
```diff
@@ -0,0 +1,43 @@
...
+		To:     w.ActorAddr,
+		Method: miner.Methods.ChangeWorkerAddress,
+
+		Value:  big.Zero(),
+		Params: sp,
+	}, nil)
+	require.NoError(t, err)
+
+	WaitMsg(ctx, t, client, smsg.Cid())
+}
```
raulk, line 43: Can we move this to the TestMiner type? Then we can get rid of t, client, and w.

chain/actors/policy/policy.go.template
```diff
@@ -132,6 +132,17 @@ func GetMaxProveCommitDuration(ver actors.Version, t abi.RegisteredSealProof) ab
 	}
 }
 
+func SetProviderCollateralSupplyTarget(num, denom big.Int) {
```
arajasek, line 135: Can we throw 

```
Can we throw 

```
// SetProviderCollateralSupplyTarget sets the percentage of normalized circulating
// supply that must be covered by provider collateral in a deal. This should
// only be used for testing.
```

File: chain/actors/policy/policy.go.template [hunk 3]
```diff
@@ -132,6 +132,17 @@ func GetMaxProveCommitDuration(ver actors.Version, t abi.RegisteredSealProof) ab
...
+{{range .versions}}
+	{{if (ge . 2)}}
```
arajasek, line 137: Guessing we're not doing this for earlier versions just because it isn't that helpful and the `ProvCollateralPercentSupplyNumV0` / `ProvCollateralPercentSupplyNumV1` thing is messy?

arajasek (approved):

chain/actors/policy/policy.go.template
magik6k: We didn't have supply target in v0 actors

magik6k: (and v1 actors didn't exist)

arajasek: isn't that what `ProvCollateralPercentSupplyNumV0` and `ProvCollateralPercentSupplyNumV1` did in v0 actors?

magik6k: Oh, maybe, but somehow those don't seem to matter

PR #6702, opened 2021-07-08
Refactor retrieval commands

### LINKED ISSUE #6415: [Proposal] Simplify ClientRetrieveAPIs to mirror storage
**Is your feature request related to a problem? Please describe.**

Currently, we have apis for initiating client retrievals:

```
	// ClientRetrieve initiates the retrieval of a file, as specified in the order.
	ClientRetrieve(ctx context.Context, order api.RetrievalOrder, ref *api.FileRef) error //perm:admin
	// ClientRetrieveWithEvents initiates the retrieval of a file, as specified in the order, and provides a channel
	// of status updates.
	ClientRetrieveWithEvents(ctx context.Context, order api.RetrievalOrder, ref *api.FileRef) (<-chan marketevents.RetrievalEvent, error) //perm:admin
```

These functions do A LOT:
- They initiate the retrieval
- They also consume events relating to retrieval until the deal is done
- They export the output file to the passed in file ref.
- They transparently clear out any blockstore space used

The APIs also convert the stream of events generated by go-fil-markets to a custom stream, in the process dropping some important data (like the retrieval deal ID). The only difference between the two methods is one version returns the custom event stream while the other consumes said stream to then end and returns the final error.

Finally, these APIs look quite different from storage APIs, which provide:
- function to import data
- function to simply initiate a deal (but not wait for it do finish)
- function to get all deal updates
- function to list deals

**Describe the solution you'd like**

Per https://github.com/filecoin-project/lotus/pull/6337 we now have functions that mirror two of the four storage functions:
- function to get all retrieval deal updates - `ClientGetRetrievalUpdates`
- function to list all retrievals - `ClientListRetrievals`

My proposal is to remove the two above APIs from the V1 API and provide two much simpler functions:

```golang
	// ClientExport exports a file stored in the local filestore to a system file
	ClientExport(ctx context.Context, exportRef ExportRef, fileRef FileRef) error //perm:admin
	ClientRetrieve(ctx context.Context, params RetrievalParams) (*RestrievalRes, error) //perm:admin
```

The types referenced are:

```golang	

// RetrievalParams is RetrievalOrder minus the "LocalStore" param
type RetrievalParams struct {
	// TODO: make this less unixfs specific
	Root                    cid.Cid
	Piece                   *cid.Cid
	Size                    uint64
	Total                   types.BigInt
	UnsealPrice             types.BigInt
	PaymentInterval         uint64
	PaymentIntervalIncrease uint64
	Client                  address.Address
	Miner                   address.Address
	MinerPeer               *retrievalmarket.RetrievalPeer
}

type RestrievalRes struct {
	DealID  retrievalmarket.DealID
	StoreID *multistore.StoreID
}

type ExportRef struct {
	Root    cid.Cid
	StoreID *multistore.StoreID
}
```

The V0 API could continue to offer the old methods, backed by the functions above, plus `ClientGetRetrievalUpdates` and `ClientRemoveImport` (to clear out blockstore data)

**Describe alternatives you've considered**
Adding a DealID to the event stream returned by ClientRetrieveWithEvents, a simpler solution one way my team is block, but one which really doesn't fix the API asymmetry and how complicated the retrieval methods currently are

**Additional context**

There is one potential gotcha:
`ClientGetRetrievalUpdates` returns a struct that is missing one field from the custom stream returned by ClientRetrieveWithEvents who data type is marketevents.Event -- an "Event" field which references the internal go-fil-markets event that triggered the update.  This is currently included in  the CLI output, though I don't think it's very informative and I'd be surprised if anyone else is using it.  We could accept the breaking change of dropping this parameter, or add it to the return channel for `ClientGetRetrievalUpdates`

This also allows us to remove the retrieval store manager.

Also, this does cause a number of changes to integration tests and lotus-soup, plus we can optionally rewrite the CLI command. I've actually implemented a lot of this change but don't want to spend more time without general approval of the direction

hannahhoward:
# Goals

fix #6415

# Implementation

This implements the proposed retrieval changes per the referenced Github issue.

Note that this most likely will conflict with https://github.com/filecoin-project/lotus/pull/6671 and #6415 does not have consensus. I also know we're invested in de-risking prior to the 1.11.1 release.

That said, I wanted to "get this done" when I had a moment to provide an demonstration of what it looks like and so that others can hopefully finish it off.


jennijuju:
blocked by https://github.com/filecoin-project/lotus/pull/6671

magik6k:
Already lives in master, merged in one of the recent API refactors

Issue #6705, opened 2021-07-08
Implement Exposed Blockstores


vyzo:
With splitstore v1 (see #6474) there is now a real distinction between the exposed blockstore abstraction and the underlying blockstore.
Network requests can pin references into the hotstore during compaction, which can result in keeping unwanted data hot even though they are unreachable from the chain.

In order to communicate the origin of network requests, we propose an `ExposableBlockstore` trait/optional interface which provides a view of the blockstore with network request awareness in relevant codepaths.
The trait looks like this:
```
type ExposableBlockstore interface {
  Expose() Blockstore
}
```

The splitstore can implement the trait and provide a view that bypasses transactional protection for all reads.

The exposed blockstore abstraction is currently hooked through DI to chainxchg/bitswap, but we also need to hook it to the blockstore exposed through json-rpc.

PR #6718, opened 2021-07-09
Splitstore Review

Stebalien:
Reviewing #6474 

Unfortunately, github makes it difficult to review rewrites because the patches are nasty. So I'm just adding comments to the code directly.

To respond, please review _this_ PR. I'll continue to push new comments inline.

### DIFF
--- a/blockstore/splitstore/splitstore.go
+++ b/blockstore/splitstore/splitstore.go
@@ -693,6 +693,8 @@ func (s *SplitStore) trackTxnRefMany(cids []cid.Cid) {
 					quiet = true
 					log.Warnf("error checking markset: %s", err)
 				}
+				// This is inconsistent with trackTxnRef (where we track it
+				// anyways).
 				continue
 			}
 
@@ -703,8 +705,6 @@ func (s *SplitStore) trackTxnRefMany(cids []cid.Cid) {
 
 		s.txnRefs[c] = struct{}{}
 	}
-
-	return
 }
 
 // protect all pending transactional references
@@ -936,6 +936,12 @@ func (s *SplitStore) doWarmup(curTs *types.TipSet) error {
 func (s *SplitStore) compact(curTs *types.TipSet, wg *sync.WaitGroup) {
 	log.Info("waiting for active views to complete")
 	start := time.Now()
+	// I'm not holding the txnLk here. Is it not possible for:
+	//
+	// 1. The waitgroup to temporarily hit 0.
+	// 2. protectView to be called, raising it back to 1.
+	//
+	// _Reusing_ a waitgroup is legal. Hitting 0 then going back to 1 while "waiting" is not.
 	wg.Wait()
 	log.Infow("waiting for active views done", "took", time.Since(start))
 
@@ -1457,6 +1463,7 @@ func (s *SplitStore) sortObjects(cids []cid.Cid) error {
 		wi := weights[key(cids[i])]
 		wj := weights[key(cids[j])]
 		if wi == wj {
+			// nit: consider just comparing the keys?
 			return bytes.Compare(cids[i].Hash(), cids[j].Hash()) > 0
 		}
 


blockstore/splitstore/splitstore.go
```diff
@@ -618,13 +620,27 @@ func (s *SplitStore) trackTxnRefMany(cids []cid.Cid) error {
...
 	// cannot be deleted before the object itself.
 	err := s.walkObjectIncomplete(root, cid.NewSet(),
 		func(c cid.Cid) error {
+			// Let's make sure we _completely_ match vm.Copy. We don't need this _now_,
+			// but should try to not to introduce inconsistencies.
+
+			if root.Prefix().MhType == 0 {
+				// identity cid, skip
+				return errStopWalk
+			}
```
vyzo, line 642: let's move this into `isFilCommitment` and rename that function, so that we don't have to sprinkle it everywhere.

blockstore/splitstore/splitstore.go
```diff
@@ -1282,6 +1298,7 @@ func (s *SplitStore) checkClosing() error {
 	return nil
 }
 
+// This is very dangerious and could go horribly wrong if we have anything else with a structure that matches a block header.
```
vyzo, line 1301: Yeah I hate it too.

I've been considering removing it and just live with extra garbage from network requests until https://github.com/filecoin-project/lotus/issues/6705 is implemented.

I want to resolve that before v1.11.1 so we might as well already remove this code.

blockstore/splitstore/splitstore.go
```diff
@@ -618,13 +620,27 @@ func (s *SplitStore) trackTxnRefMany(cids []cid.Cid) error {
 	return nil
 }
 
+// This could get _real_ slow (many minutes to hours) and will block tipset verification. Can we do this lazily and/or force
+// the GC thread to handle this?
+//
+// I.e., take a lock to make sure GC is paused, then stick this in a "to traverse" queue. Then, when
+// GC "resumes", it first checks that queue.
```
vyzo, line 627: yeah, I have been thinking about this too.

It hasn't been observed in testing, but then again it's about 30 compactions so far and it will probably happen very infrequently.

blockstore/splitstore/splitstore.go
```diff
@@ -228,6 +228,8 @@ func (s *SplitStore) Has(cid cid.Cid) (bool, error) {
 		return true, nil
 	}
 
+	// Didn't we need to somehow ressurect everything under this in case we decide to GC? Or are
+	// we handling that in a followup patch.
```
vyzo, line 232: we do, but first we have to implement #6705.
So subsequent pr, TBD before code freeze for v1.11.1.

vyzo: Otherwise anyone syncing from us will blow the hotstore with old messages and stuff.

blockstore/splitstore/splitstore.go
vyzo: done in https://github.com/filecoin-project/lotus/pull/6474/commits/de5e21bf1a86746166ee9cb637d4121ccde10bba
I also added code to handle identity cids at the interface boundary so as to be consistent with wrapped idstores.

blockstore/splitstore/splitstore.go
vyzo: Ok, I killed it at https://github.com/filecoin-project/lotus/pull/6474/commits/4f89d260b05e7da546b82b7e0e2f42e50851c3a3

At this point, the worst that could happen is that we keep some useless blocks from forks due to some network request.
I think we can live with that until #6705 is done.

blockstore/splitstore/markset_bloom.go
```diff
@@ -62,6 +62,11 @@ func (s *BloomMarkSet) saltedKey(cid cid.Cid) []byte {
 	key := make([]byte, len(s.salt)+len(hash))
 	n := copy(key, s.salt)
 	copy(key[n:], hash)
+	// FYI, you probably don't need a cryptographic hash function here given that we're salting
+	// anyways.
+	// You could use https://golang.org/pkg/hash/maphash/
+	// (probably)
+	// (if it's a perf issue)
```
vyzo, line 69: I'd rather have a crypto hash here, as a collision would totally kill debugability.

blockstore/splitstore/splitstore.go
vyzo: Implemented concurrent marking -- see https://github.com/filecoin-project/lotus/pull/6474/commits/da0feb3fa43a78bd7be3d8aa49495cd31e1a87c2

blockstore/splitstore/markset_bloom.go
vyzo: But yeah, if we have perf problems with bloom filters we can totally change the hash function.

blockstore/splitstore/splitstore.go
Stebalien: Eh, block headers are _tiny_ anyways.

blockstore/splitstore/splitstore.go
vyzo: and it works pretty well -- there is a compaction running right now, and it cleans up txn refs in at most two iterations with concurrent marking.

blockstore/splitstore/splitstore.go
vyzo: yeah, but they point to giant trees behind state roots :)
Anyway, it's not a big deal since we have marked the chain and it will should fixed in time for release so not really a problem.

Stebalien:
Have we tried exercising this code by syncing the chain? I assume most of our "realistic" tests have been one offs so it would be nice to have a test that rapidly and repeatedly purges.

vyzo:
> Have we tried exercising this code by syncing the chain? I assume most of our "realistic" tests have been one offs so it would be nice to have a test that rapidly and repeatedly purges.

Yes, I have two live nodes running latest code for about a week now, one with discard and the other with a coldstore. Both are syncing fine and exhibit no ill effects, with block validation unaffected by compaction.

However, as we've been discussing with @raulk, we do need a harness test so that we can reliably and reproducibly do automated testing. In my opinion, this is critical and necessary before we declare the code production ready. I will create follow up issue as it is unlikely we'll get to it before code freeze for v1.11.1.

blockstore/splitstore/splitstore.go
vyzo: I will create a follow up issue for this, as I am not sure we'll get to it before code freeze for v1.11.1 and we also need to think a bit more about it -- i know @raulk has opinions on the matter.

blockstore/splitstore/splitstore.go
```diff
@@ -1522,24 +1582,31 @@ func (s *SplitStore) purge(cids []cid.Cid, markSet MarkSet) error {
 		func(cids []cid.Cid) error {
 			deadCids := deadCids[:0]
 
-		again:
-			if err := s.checkClosing(); err != nil {
-				return err
-			}
+			for {
```
vyzo, line 1585: haha, the allergy to goto's strikes again!

blockstore/splitstore/splitstore.go
```diff
@@ -1259,6 +1318,7 @@ func (s *SplitStore) walkObject(c cid.Cid, walked *cid.Set, f func(cid.Cid) erro
 	return nil
 }
 
+// Maybe check closing in here? This could run for a while.
```
vyzo, line 1321: sure, that's probably a good idea.

blockstore/splitstore/splitstore.go
```diff
@@ -1126,6 +1184,7 @@ func (s *SplitStore) beginTxnProtect() *sync.WaitGroup {
 }
 
 func (s *SplitStore) beginTxnMarking(markSet MarkSet) {
+	// No lock? Have we tested this with the golang race detector?
```
vyzo, line 1187: it's only invoked by compaction exactly once, so it is protected effectively by the `compacting` lock.

blockstore/splitstore/splitstore.go
```diff
@@ -1093,6 +1150,7 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 	log.Infow("purging cold objects from hotstore done", "took", time.Since(startPurge))
 
 	// we are done; do some housekeeping
+	// We have a bunch of ways to return early without invoking this. Maybe we need to defer it? Or set some kind of dirty bit?
```
vyzo, line 1153: it is already deferred when we `beginTxnProtect` before we invoke `compact`.

blockstore/splitstore/splitstore.go
```diff
@@ -751,6 +806,7 @@ func (s *SplitStore) protectTxnRefs(markSet MarkSet) error {
 			}
 			wg.Wait()
 		} else {
+			// This really isn't a necessary optimization.
```
vyzo, line 809: ha ok, I will remove it.

blockstore/splitstore/splitstore.go
```diff
@@ -1063,6 +1119,7 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 	}
 	log.Infow("sorting done", "took", time.Since(startSort))
 
+	// Can we add a comment explaining why we might want to do this here instead of just waiting till we purge?
```
vyzo, line 1122: sure; basically I want to minimize the invocations inside purge, but we can also ellide it.

blockstore/splitstore/splitstore.go
```diff
@@ -737,6 +789,9 @@ func (s *SplitStore) protectTxnRefs(markSet MarkSet) error {
 			for c := range workch {
 				err := s.doTxnProtect(c, markSet)
 				if err != nil {
+					// I would _NOT_ proceed if we fail here. Any errors like
+					// this in GC should cause us to abort and (to avoid
+					// deleting anything).
```
vyzo, line 794: yeah, good point. will fix.

blockstore/splitstore/splitstore.go
```diff
@@ -676,6 +725,7 @@ func (s *SplitStore) trackTxnRefMany(cids []cid.Cid) {
 
 		s.txnRefs[c] = struct{}{}
 	}
+	// defer?
```
vyzo, line 728: sure.

blockstore/splitstore/splitstore.go
```diff
@@ -551,6 +591,15 @@ func (s *SplitStore) HeadChange(_, apply []*types.TipSet) error {
 
 	if !atomic.CompareAndSwapInt32(&s.compacting, 0, 1) {
 		// we are currently compacting -- protect the new tipset(s)
+
+		// race: we call beginTxnProtect _after_ setting compacting.
+		// 1. HeadChange triggers a compaction but stops immediately after setting this
+		//    flag.
+		// 2. HeadChange tries ties to trigger another compaction, sees one is already in
```
vyzo, line 598: wait, that's not possible. HeadChange notifications are synchronous, we can't have a second one before the first one has returned.

blockstore/splitstore/splitstore.go
```diff
@@ -529,6 +568,7 @@ func (s *SplitStore) Start(chain ChainAccessor) error {
 func (s *SplitStore) Close() error {
 	atomic.StoreInt32(&s.closing, 1)
 
+	// I'm _pretty_ sure we can just use a lock here now. Doesn't even need to be a rwmutex.
```
vyzo, line 571: I was using a lock way back when the code was first written but raul was adamant that we ought to use an atomic here as it is exactly the pattern that calls for it.

blockstore/splitstore/splitstore.go
```diff
@@ -435,14 +470,18 @@ func (s *SplitStore) View(cid cid.Cid, cb func([]byte) error) error {
...
+	// ends then re-opens while a view is open? Highly unlikely... but I wonder if we should
+	// just use the wait-group regardless?
+
+	// ultra-nit: you _can_ make this _slightly_ simpler by returning a possibly no-op function:
+	defer s.protectView(cid)()
 
 	err := s.hot.View(cid, cb)
 	switch err {
 	case bstore.ErrNotFound:
+		// I see we repeat this pattern a lot. Can we put it in a helper?
```
vyzo, line 484: sure

blockstore/splitstore/splitstore.go
```diff
@@ -435,14 +470,18 @@ func (s *SplitStore) View(cid cid.Cid, cb func([]byte) error) error {
...
 	//    lead to deadlock with recursive views.
-	wg := s.protectView(cid)
-	if wg != nil {
-		defer wg.Done()
-	}
+
+	// What happens if the transaction ends while a view is open? What happens if a transaction
+	// ends then re-opens while a view is open? Highly unlikely... but I wonder if we should
+	// just use the wait-group regardless?
+
```
vyzo, line 477: that's fine, we'll just track the references normally.

blockstore/splitstore/splitstore.go
```diff
@@ -435,14 +470,18 @@ func (s *SplitStore) View(cid cid.Cid, cb func([]byte) error) error {
...
+	// ultra-nit: you _can_ make this _slightly_ simpler by returning a possibly no-op function:
```
vyzo, line 478: a little prettier but it also allocates a closure; not sure if it is worth it.

blockstore/splitstore/splitstore.go
```diff
@@ -393,20 +402,46 @@ func (s *SplitStore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 		return nil, err
 	}
 
-	ch := make(chan cid.Cid)
+	// arbitrary but >= 1 helps avoid context switching in this case
+	ch := make(chan cid.Cid, 8)
```
vyzo, line 406: sure, will add a buffer.

blockstore/splitstore/splitstore.go
```diff
@@ -263,6 +271,7 @@ func (s *SplitStore) Get(cid cid.Cid) (blocks.Block, error) {
 
 		blk, err = s.cold.Get(cid)
 		if err == nil {
+			// nit: use the stored context?
```
vyzo, line 274: sure.

blockstore/splitstore/splitstore.go
```diff
@@ -181,6 +187,8 @@ func Open(path string, ds dstore.Datastore, hot, cold bstore.Blockstore, cfg *Co
 		cold:       cold,
 		markSetEnv: markSetEnv,
 
+		// nit: this doesn't need to be allocated as long as we don't move it.
+		// fyi, you _can_ return &s.txnViews.
```
vyzo, line 191: yeah, it doesn't move. I hadn't realized you can reuse the waitgroup with a new `Wait` after a previous `Wait` is done.

So yeah, no need to allocate indeed.

blockstore/splitstore/splitstore.go
```diff
@@ -163,6 +163,12 @@ func init() {
 // attached to the ChainStore with Start in order to trigger compaction.
 func Open(path string, ds dstore.Datastore, hot, cold bstore.Blockstore, cfg *Config) (*SplitStore, error) {
 	// hot blockstore must support BlockstoreIterator
+	// nit: can we just define a local type?
+	// So we can store this instead of casting multiple times?
+	type hotStore interface {
+		bstore.Blockstore
+		bstore.BlockstoreIterator
+	}
```
vyzo, line 171: sure, although it doesn't cost anything in the grand scheme of things.

blockstore/splitstore/splitstore.go
```diff
@@ -393,20 +402,46 @@ func (s *SplitStore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
...
+			case <-ctx.Done():
+				return
 			}
 		}
+
+		// Also note: We're _supposed_ to guarantee that CIDs are only returned once. We
+		// _probably_ need to put the CIDs in a set to check that.
+		//
+		// That might actually be a reason to read through the hotstore before the
+		// coldstore?
```
vyzo, line 444: ah yeah, we might have some dupes in the coldstore.
I'll add a markset to make sure we don't return keys twice.

blockstore/splitstore/splitstore.go
vyzo: also note, when a view has started before a transaction we will actually defer starting compaction until it's done.

blockstore/splitstore/splitstore.go
```diff
@@ -1699,6 +1768,7 @@ func bytesToUint64(buf []byte) uint64 {
 	return i
 }
 
+// So, I get the name... I think? But maybe flip it and say `isStateObject` (or `notStateObject`)?
```
vyzo, line 1771: hrm... maybe; not married to the name, but not sure the alternatives are any better either!

blockstore/splitstore/splitstore.go
vyzo: the only reason I am doing it seperately is there is likely a large number of references accumulated during sort (it's slow!) and I didn't want to do that in purge context.

blockstore/splitstore/splitstore.go
vyzo, line 571: it will also need a wait group to use instead of the poll. I don't feel strongly about this code and either way works. 
The polll does have the advantage that it spams the lock so that the operator will see it.
Other than that, lock+waitgroup is probably nicer.

I don't feel strongly about this, so up to you.

blockstore/splitstore/splitstore.go
vyzo: I think I will add a comment about this, so as to make the invariant assumption explicit.

Note that this is very unlikely to change, as a lot of things depend on this and all hell will break loose if it ever changes.

blockstore/splitstore/splitstore.go
vyzo: Also note it's idempotent, so this basically ends the transaction early.

blockstore/splitstore/splitstore.go
```diff
@@ -1522,24 +1582,31 @@ func (s *SplitStore) purge(cids []cid.Cid, markSet MarkSet) error {
...
+					// keep the lock when we hit the desired condition.
+					break
+				}
 				s.txnLk.Unlock()
 
+				// Doesn't really matter but...
+				//
+				// It's slightly strange that we're effectively looping _twice_.
+				// Maybe break protectTxnRefs into two functions? One for the loop,
+				// and one to do the actual thing?
```
vyzo, line 1601: well, there is also logging involved, and I wanted to see those timings and numbers in the logs.

blockstore/splitstore/splitstore.go
```diff
@@ -1635,6 +1702,8 @@ func (s *SplitStore) waitForMissingRefs(markSet MarkSet) {
 		}
 	}
 
+	// Do we really want to proceed here? Maybe we should just abort and try again later?
+	// Also, aren't these errors?
```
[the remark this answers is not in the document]
vyzo: Also, someone syncing off us and reading old headers will trigger this as well; this vector will go away once #6705 is done though.

blockstore/splitstore/splitstore.go
vyzo: done, also in `walkObject` -- i put it right before scanning links for recursion.

blockstore/splitstore/splitstore.go
vyzo: done.

blockstore/splitstore/splitstore.go
vyzo: removed it.

blockstore/splitstore/splitstore.go
vyzo: Rewrote to use errgroup and return an error if it fails to protect.

... and yeah, errgroup is very neat indeed!

blockstore/splitstore/splitstore.go
vyzo: added comment.

blockstore/splitstore/splitstore.go
vyzo: and we do this to reduce memory pressure while garbage collecting the hotstore.

blockstore/splitstore/splitstore.go
vyzo: rewrote without the goto.

vyzo (commented):
Addressed review comments in https://github.com/filecoin-project/lotus/pull/6474/commits/0c5e336ff19c0530cf48459151c5e6d2ef55ad6b

vyzo:
Created follow up issue for testing in https://github.com/filecoin-project/lotus/issues/6725
It is not urgent, but we definitely need to resolve this in the next few months.

blockstore/splitstore/splitstore.go
vyzo: follow up issue in https://github.com/filecoin-project/lotus/issues/6726

blockstore/splitstore/splitstore.go
Stebalien: I see. My main concern was:

1. Start compaction.
2. Start view.
3. Finish compaction.
4. Start compaction.

From what I could tell, the second compaction wouldn't wait for the view. Is that correct or am I missing something?

blockstore/splitstore/splitstore.go
Stebalien: Well.... that can always be solved. But that just moves ugly code from one place to another, so meh.

Stebalien: (really, it makes it overall worse)

blockstore/splitstore/splitstore.go
Stebalien: Ok, fair enough. I'd still feel better with a lock in case someone decides to change that, but I guess it's a reasonable constraint on HeadChange.

blockstore/splitstore/splitstore.go
Stebalien: Hm. But it's _read_ elsewhere, right? Have you run tests with the race detector?

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
Stebalien: > TL;DR They don't seem to be errors at all and there is a legitimate scenario where they can arise: a lookback for randomness references an old block header whose messages and state root have been garbage collected (or were never there in the first place because we synced from snapshot).

Ah, I see. Because we're keeping everything we've touched. We _really_ need VM integration.

blockstore/splitstore/splitstore.go
Stebalien: Well, I assume "unitary" refers to "unit", but these aren't really "units". _Really_, we're looking to recurs through all "state" objects, skipping anything that's not a part of the state.

vyzo: Yeah, it refers exactly to that; it's a unit, no state attached in the blockstore.
Really, I don't particularly care how it's called, as long as it conveys enough information :)

blockstore/splitstore/splitstore.go
vyzo: well, it's not bigie to put a lock just for head changes! Just pointing out it's not necessary, but it will take 30s to put one.

blockstore/splitstore/splitstore.go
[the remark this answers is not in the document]
vyzo: yeah, agreed.

blockstore/splitstore/splitstore.go
vyzo: yes, I have, but we have just one simple unit test so it's not that great of a confidence builder.

blockstore/splitstore/splitstore.go
Stebalien: Well, it's a unit from the perspective of the state-tree. But I had to think _real_ hard to figure out why we were calling it that.

(but this also really doesn't matter)

blockstore/splitstore/splitstore.go
vyzo: Ah yes, you are right. Unlikely to happen but possible, so I guess we should just always use the WaitGroup (now that it is persistent).

vyzo: done in https://github.com/filecoin-project/lotus/pull/6474/commits/759594d01c22a63cc920138f7c67285eea3ff051

blockstore/splitstore/splitstore.go
vyzo: done in https://github.com/filecoin-project/lotus/pull/6474/commits/60212c86cbd855ba531fb9e7a8303cf57eb25b67

Stebalien:
Ok, final final review.

blockstore/splitstore/splitstore.go
```diff
@@ -936,6 +936,12 @@ func (s *SplitStore) doWarmup(curTs *types.TipSet) error {
 func (s *SplitStore) compact(curTs *types.TipSet, wg *sync.WaitGroup) {
 	log.Info("waiting for active views to complete")
 	start := time.Now()
+	// I'm not holding the txnLk here. Is it not possible for:
+	//
+	// 1. The waitgroup to temporarily hit 0.
+	// 2. protectView to be called, raising it back to 1.
+	//
+	// _Reusing_ a waitgroup is legal. Hitting 0 then going back to 1 while "waiting" is not.
```
Stebalien, line 944: I.e., https://pkg.go.dev/sync#WaitGroup.Add

One solution here is to keep a separate count (under a lock)? Or maybe using a waitgroup was a bad idea and we should just use mutex, condition, and count.

"Wait"

```go
mu.Lock()
waitingToCompact = true
for activeViews > 0 {
    cond.Wait()
}
waitingToCompact = false
// do any other state updates we need to do atomically (e.g., begin the transaction?)
mu.Unlock()
```

"View" enter:

```go
mu.Lock()
defer mu.Unlock()
if waitingToCompact && activeViews == 0 {
    // abort
}
activeViews++
```

"View" exit

```go
mu.Lock()
defer mu.Unlock()
activeViews--
if activeViews == 0 && waitingToCompact {
    cond.Broadcast()
}
```

Stebalien: But maybe there are easier solutions.

vyzo: and finetuned in https://github.com/filecoin-project/lotus/pull/6474/commits/af399529ecca4c6e082474cada76c9b96a3d8f86

jacobheun:
Closing this as #6474 was merged.

PR #6721, opened 2021-07-09
add TestDealCycleProviderRestart

dirkmc:


aarshkshah1992 (approved):

aarshkshah1992:
@nonsense Does this test make sense to you as well ?

itests/deals_concurrent_test.go
```diff
@@ -107,6 +111,79 @@ func TestDealCyclesConcurrent(t *testing.T) {
...
+	// Make a storage deal against the first provider
+	deal, res, inPath := dh.MakeOnlineDeal(context.Background(), kit.MakeFullDealParams{
+		Rseed:      5,
+		StartEpoch: startEpoch,
+	})
+
+	// Set up the second provider
+	provider2Opts := append(opts,
+		kit.OwnerAddr(client.DefaultKey),
+		kit.MainMiner(&provider1),
```
nonsense, line 175: @aarshkshah1992 this doesn't make sense to me. AFAICT you want to create a second provider with the same DAG store, but this is not what this test is doing. Here we are creating a second provider and linking it to the first provider, as if you have a market process and a miner process (this is what the `MainMiner` option is for).

---

I think the idea is OK, why don't you just start a new ensemble, same as you started the first one, and just link to the same DAG store?

nonsense: Also I don't see us stopping the initial miner - not sure if we want to, or not, but I think we want to.

dirkmc: I tried following the suggestion of creating a second ensemble but unfortunately this is complicated to do because there are multiple pieces that need to be shared in order to perform a full deal cycle, for example the piece store.
I think the correct solution is what Anton suggests - we need to be able to stop and start nodes.

I'm going to close this for now and we can come back to it when it's possible to stop and start nodes.

Issue #6577, opened 2021-06-23
Splitstore: Online Garbage Collection for the Coldstore


vyzo:
# Need for Space

Once the splitstore has been deployed (see also #6474) we have the ability to perform online garbage collection for the coldstore, as we control writes.
Specifically, we only write to the coldstore during compaction, when we move newly cold objects, protected by the compaction lock.
That means we can perform gc on the coldstore without disrupting regular node operations or requiring downtime.

Garbage collecting the coldstore is an essential operation for keeping space usage bounded in non-archival nodes -- see also #4701.

## Design Considerations

Garbage collection must effectively reclaim space; hence we can't use native badger gc which is horrible at reclaiming space and requires hacks in order to convince the thing to reclaim as much space as possible.
Furthermore, even if we do manage to reclaim all space possible, the gc'ed blockstore has the tendency to quickly balloon back up in size.

Instead, we propose a _moving_ garbage collector for the coldstore, which also allows us to tune the gc process to the user's needs.

Fundamentally, the gc operation will instantiate a new (empty) coldstore, walk the chain for live objects in the coldstore according to user retention policies, and then move live objects to the new coldstore.

Once the move is complete, the new coldstore becomes the actual coldstore and the old coldstore is deleted.

### Retention Policies:

At a minimum, we must ensure that we retain chain headers (all the way to genesis), as it is not currently safe to discard them due to unlimited randomness lookback (this may change in the future, but we still want to keep them in order to be able to navigate the chain).

Apart from that it is up to the user:
- whether to retain messages, and for how many finalities (or all of them)
- whether to retain message receipts
- how deep to retain state roots and associated objects.

So the garbage collection interface must allow the user to specify preferences/policies that match their own demand.

A possible sane default:
- retain all messages and receipts to allow other nodes to sync using our node.
- retain state roots for up to a few finalities (this could perhaps be none) to allow resets.

### Additional considerations:
- the coldstore is by design likely to be housed in a separate disk and be symlinked in the `.lotus/datastore/chain`  path.
- we must not run out of space during the move; the user must ensure enough available space. Generally this should be less than the current coldstore, but more precise estimtes can be made by the size of the hotstore and the number of finalities we want to retain for space objects.

In order to address these issues, we propose that the user supplies the new coldstore path at the time of move, which is then symlinked by the system itself into `~/.lotus/datastore/chain`.

## Interface

We propose to introduce a new (v1) API which can be invoked to trigger gc on demand, perhaps through a cron job.
[...]

Issue #6725, opened 2021-07-10
Splitstore Harness Test


vyzo:
During the development of Splitstore v1 in #6474, we laregely relied on manual testing using live nodes in the mainnet.
While this is acceptable for development, and indeed very beneficial and confidence building to have the nodes running in mainnet, it is not sufficient for testing of production code.

We propose to create a harness test for splitstore, which includes a sufficient subset of the chain (at least 6 finalities) so as to initialize a splitstore node and trigger compaction, possibly starting from a snapshot.
We can't easily do that with an integration test, as we need a lot of (real world!) blocks and a sufficiently long chain to trigger at least one compaction.
We also need a sufficiently complex part of lotus so as to validate blocks and compute state during the test to ensure that everything is nominal.

This is a non-trivial project, but it is necessary in order to mark the splitstore production ready.

PR #6728, opened 2021-07-11
Splitstore Garbage Collection

vyzo:
Implements online coldstore chain pruning and garbage collection with the `lotus chain prune` command.

See #6577  for discussion.

From the README:
The coldstore can be pruned and garbage collection using the `lotus chain prune` command.
Note that the command initiates pruning, but runs asynchronously as it can take a long time to
complete.

By default, pruning keeps all chain reachable object; the user however has the option to specify
a retention policy for old state roots and message receipts with the `--retention` option.
The value is an integer with the following semantics:
- If it is -1 then all state objects reachable from the chain will be retained in the coldstore.
  This is the (safe) default.
- If it is 0 then no state objects that are unreachable within the compaction boundary will
  be retained in the coldstore.
  This effectively throws away all old state roots and it is maximally effective at reclaiming space.
- If it is a positive integer, then it's the number of finalities past the compaction boundary
  for which chain-reachable state objects are retained.
  This allows you to keep some older state roots in case you need to reset your head outside
  the compaction boundary or perform historical queries.

During pruning, unreachable objects are deleted from the coldstore. In order to reclaim space,
you also need to specify a gargage collection policy, with two possible options:
- The `--online-gc` flag performs online garbage collection; this is fast but does not reclaim all
  space possible.
- The `--moving-gc` flag performs moving garbage collection, where the coldstore is moved,
  copying only live objects. If your coldstore lives outside the `.lotus` directory, e.g. with
  a symlink to a different file system comprising of cheaper disks, then you can specify the
  directory to move to with the `--move-to` option.
  This reclaims all possible space, but it is slow and also requires disk space to house the new
  coldstore together with the old coldstore during the move.


vyzo:
rebased for changes in the base branch.

vyzo:
rebased on master.

vyzo:
I'll rebase and resolve the gen conflicts when we are ready to merge.

Stebalien:
This needs integration tests. I.e., tests where we trigger GC _while_ computing a tipset or at least something that _approximates_ computing a tipset (i.e. a bunch of reads, then `vm.Copy`). Ideally:

1. Start GC before computing, and during.
2. End GC during and after.

VM execution doesn't _just_ write blocks, it makes a bunch of assumptions about what we have and what we don't have.

vyzo:
Agreed; for now I am testing in a live node, but we definitely want a proper integration test.
I propose we add it to the Harness test effort outlined in #6725 

How does it sounds to you? 
It something we can reasonably handle without time pressure after the v1.11.1 release code freeze.

Stebalien:
It's really a waste of time to review until there are tests. Testing on a node isn't likely to trigger the conditions we care about.

Stebalien:
(also, let's please break this into two PRs: one that refactors (splits files), then one that makes the actual change)

vyzo:
Agreed with the sentiment, following our sync discussion.
I'll cherry-pick the refactor out for easy review, and then the plan is to write a good enough integration test. 
If we can get it done before the deadline, great.
If not, we don't have to ship it now, and people who are willing to take the risk can always use the pr directly.

BlocksOnAChain:
@vyzo should we close this one since it's related to 1.11.1 which is out?
FYI: @jennijuju

vyzo:
no, its in draft for a reason.

magik6k:
Done in https://github.com/filecoin-project/lotus/pull/9056

PR #6739, opened 2021-07-12
ClientRetrieve stops on cancel

### LINKED ISSUE #6738: [BUG] ClientRetrieve & ClientRetrieveWithEvents hang when ClientCancelRetrievalDeal is called
> Note: For security-related bugs/issues, please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).

**Describe the bug**
There is no check for a deal being cancelled in ClientRetrieve or ClientRetrieveWithEvents. As a result if you called ClientCancelRetrievalDeal while another call for ClientRetrieve or ClientRetrieveWithEvents is in progress, they will hang indefinitely.

any

**To Reproduce**
1. With no other deals running, in one terminal window, create a retrieval deal with `lotus client retrieve`
2. While other terminal is running, in seperate window, use `lotus client list-retrievals` to extract deal id (or inspect logs)
2. Cancel that retrieval deal (lotus client cancel-retrievel --deal-id ~extracted-deal-id~)

First window will hang indefinitely. It will see the cancel event but channel will never terminate, causing deal to hang.

**Expected behavior**
A cancel should terminate a call to Lotus ClientRetrieve or ClientRetrieveWithEvents

**Logs**
Discovered using dealbot

hannahhoward:
# Goals

fix #6738

# Implementation

Check for deal status cancelled when inspecting deal events in ClientRetrieve and ClientRetrieveWithEvents, and close channel when that happens.

### DIFF
--- a/node/impl/client/client.go
+++ b/node/impl/client/client.go
@@ -680,6 +680,8 @@ func readSubscribeEvents(ctx context.Context, dealID retrievalmarket.DealID, sub
 			return nil
 		case rm.DealStatusRejected:
 			return xerrors.Errorf("Retrieval Proposal Rejected: %s", state.Message)
+		case rm.DealStatusCancelled:
+			return xerrors.Errorf("Retrieval was cancelled externally: %s", state.Message)
 		case
 			rm.DealStatusDealNotFound,
 			rm.DealStatusErrored:


raulk (approved):
Should probably implement an integration test at some point.

arajasek:
is it not bad go practice for errors to start with a capital letter? (i don't really know what i'm talking about)

hannahhoward:
@arajasek I think it might be, and also, I seem to chafe against this cause I hate the idea of not using proper human grammar, so no matter how many times someone reminds me, I forget this rule. (unintentionally I swear)

raulk:
oOpSiE

raulk:
wOULD tHIS bE a vALID eRROR

raulk:
Just kidding :-) Agree we should settle on a rule and revisit all errors to comply with it. And that rule should probably be what's idiomatic (starts with lower case, doesn't end with punctuation)

PR #6743, opened 2021-07-13
Handle the --color flag via proper global state

ribasushi:
Use `color.NoColor` as a global progressively modified as one descends
down command flags. Allows non-surprising operation when --color is
accepted at multiple layers, and even makes this behave predictably:

    lotus-miner --color actor control list --color=false
 
Thanks to @stuberman for [poking me about the current state being non-ideal](https://github.com/filecoin-project/lotus/commit/4dd59b566ced9712a3af3e8accb6aaa612ab058b#commitcomment-53382350), and to @iand for prompting me to reexamine how we do things

image

P.S. GNU-getopt 4EVAH‼️


cli/util.go
```diff
@@ -15,6 +18,13 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 )
 
+// Set the global default, to be overridden by individual cli flags in order
+func init() {
+	color.NoColor = os.Getenv("GOLOG_LOG_FMT") != "color" &&
+		!isatty.IsTerminal(os.Stdout.Fd()) &&
```
iand, line 24: Strictly I think you should check `GOLOG_OUTPUT` to see if logging is being redirected to `stderr` here

iand (approved):
LGTM but see the comment about GOLOG_OUTPUT

cli/util.go
ribasushi: Urgh... that's a bit of a can of worms...
Perhaps we should just make this a public method and inspect the result?
https://github.com/ipfs/go-log/blob/5b4daa4491c2cc9e0ba85ce4b5fcc913f6fcdf6b/setup.go#L274-L377

iand: Sensible suggestion to avoid duplicating that logic

ribasushi: Need to think about this more, that'll have to come via another PR

magik6k (approved):

PR #6754, opened 2021-07-14
Splitstore: Some small fixes

vyzo:
Cherry-picked from #6728 

- Fixes a couple of inconsistencies identified post merge
- Introduces BlockstoreGC and BlockstoreMover interfaces in preparation for moving GC in #6728
- Adds a compactionIndex state variable to use as decision for moving gc (where applicable) and also a warmup indicator for splitstore v0 upgrades.

### DIFF
--- a/blockstore/splitstore/splitstore_gc.go
+++ b/blockstore/splitstore/splitstore_gc.go
@@ -5,39 +5,12 @@ import (
 	"time"
 
 	bstore "github.com/filecoin-project/lotus/blockstore"
-	cid "github.com/ipfs/go-cid"
 )
 
 func (s *SplitStore) gcHotstore() {
-	// we only perform moving gc every 20 compactions (about once a week) as it can take a while
-	if s.compactionIndex%20 == 0 {
-		if err := s.gcBlockstoreMoving(s.hot, "", nil); err != nil {
-			log.Warnf("error moving hotstore: %s", err)
-			// fallthrough to online gc
-		} else {
-			return
-		}
-	}
-
 	if err := s.gcBlockstoreOnline(s.hot); err != nil {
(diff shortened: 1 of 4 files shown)


blockstore/splitstore/splitstore_gc.go
```diff
@@ -7,20 +7,27 @@ import (
 )
 
 func (s *SplitStore) gcHotstore() {
+	// we only perform moving gc every 10 compactions as it can take a while
+	if s.compactionIndex%10 != 0 {
+		goto online_gc
```
vyzo: note for steb: this goto is very short-lived; it disappears a couple of commits later.

Kubuxu (approved):

vyzo:
don't rush mering this yet, we are discussing the MoveTo interface with @stebalien and we might make a small change here.

Stebalien:
We're not going to be able to discuss the semantics of `MoveTo` and whether or not it's the right choice without taking other uses into account. IIRC, you said we were using it to do something with the cold store?

vyzo:
Yes, it is used in two ways in #6728 
- implement moving GC for the hotstore (the code is already here)
- implement moving GC for the coldstore, allowing the user to specify an external path. 
  this supports use cases where the coldstore is soft-linked to a separate file system comprising of cheaper disks.

vyzo:
Now, given all the grief that this is causing us, we can just drop it altogether from this pr and move it to the gc pr, where we can decide on what to do.
We will still have to make a decision as to its suitability as an interface, but it doesn't need to be made now.
Besides, we are not shipping the GC stuff until we have tests, and this is looking increasingly likely to happen after v1.11.1.

Stebalien (approved):

vyzo:
rebased on master.

PR #6760, opened 2021-07-15
cache loaded block messages

hunjixin:
GetBlockMessages Is a high-load API, use cache to reduce network traffic

### DIFF
--- a/chain/events/events_called.go
+++ b/chain/events/events_called.go
@@ -5,6 +5,9 @@ import (
 	"math"
 	"sync"
 
+	"github.com/filecoin-project/lotus/api"
+	lru "github.com/hashicorp/golang-lru"
+
 	"github.com/filecoin-project/lotus/chain/stmgr"
 
 	"github.com/filecoin-project/go-state-types/abi"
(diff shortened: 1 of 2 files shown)


chain/events/events_called.go
```diff
@@ -464,14 +467,20 @@ type messageEvents struct {
 
 	lk       sync.RWMutex
 	matchers map[triggerID]MsgMatchFunc
+
+	blsMsgLk    sync.Mutex
+	blsMsgCache *lru.ARCCache
```
magik6k, line 472: Since BlockMessages include bls and secp messages, I'd call this something different than `bls..` (e.g `blockMsg..`)

magik6k (commented):
Thanks for the PR! Just one naming nitpick.

It would also be great to have a test which asserts that the cache works, for example by asserting that the `ChainGetBlockMessages` method is only called once per tipset

chain/events/events_called.go
hunjixin: resolved

hunjixin:
@magik6k

magik6k (approved):
Thanks!

PR #6763, opened 2021-07-15
sealing: Handle preCommitParams errors more correctly

magik6k:
This is addressing review on https://github.com/filecoin-project/lotus/pull/6581

### DIFF
--- a/extern/storage-sealing/states_sealing.go
+++ b/extern/storage-sealing/states_sealing.go
@@ -358,8 +358,11 @@ func (m *Sealing) handlePreCommitting(ctx statemachine.Context, sector SectorInf
 	}
 
 	params, pcd, tok, err := m.preCommitParams(ctx, sector)
-	if params == nil || err != nil {
-		return err
+	if err != nil {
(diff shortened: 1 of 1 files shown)


extern/storage-sealing/states_sealing.go
```diff
@@ -359,7 +359,10 @@ func (m *Sealing) handlePreCommitting(ctx statemachine.Context, sector SectorInf
 
 	params, pcd, tok, err := m.preCommitParams(ctx, sector)
 	if params == nil || err != nil {
-		return err
+		if err != nil {
```
GFZRZK, line 362: this code is so wired, why not:
```go
if err != nil {
    return ctx.Send(SectorChainPreCommitFailed{xerrors.Errorf("preCommitParams: %w", err)})
}
if params == nil {
    return nil
}
```

Stebalien (approved):

Kubuxu (approved):

Kubuxu:
Auto merge won't mere it as the diff is not tested at all.

Issue #6673, opened 2021-07-04
[Meta Issue] Replace Badger in Deal Making with CARv2 and the Sharded DAG Store and test with MRA 


aarshkshah1992:
This is a meta-issue to track the work of replacing  Badger from Filecoin deal making with [CARv2](https://github.com/ipld/ipld/blob/master/specs/transport/car/carv2/index.md) and the [Sharded DAG Store](https://github.com/filecoin-project/dagstore/pull/20) and do some solid testing along with the [MRA ](https://github.com/filecoin-project/lotus/pull/6356)changes to ensure a smooth landing.

---
### Pending CARv2 features
- [x] [CARv2 read-write blockstore should support resumption](https://github.com/ipld/go-car/issues/98).
---
### Sharded DAG Store
- [x] TODO.
---
### Markets
- [x] https://github.com/filecoin-project/go-fil-markets/pull/558.
- [x] https://github.com/filecoin-project/go-fil-markets/pull/560.
- [x] Integrate Sharded DAG Store into the Storage Market.
- [x] Integrate Sharded DAG Store into the Retrieval Market.
---
### Lotus
- [x] https://github.com/filecoin-project/lotus/pull/6671 (PR is against the MRA changes).
- [x] Integrate the Sharded DAG Store in the Lotus PR above.       
--- 
### Migration 
- [ ] ~Migration of existing storage client data imports in Badger to CARv2 files:~
-  ~Iterate over all the imported payloads in the storage client's Badger store and migrate them to CARv2 files so clients don't have to re-import them. **_Maybe_** don't clean up the Badger store right away so client can downgrade without losing imports.~

- [ ] Register Shards for all active storage deals on the Miner with the Sharded DAG Store. https://github.com/filecoin-project/go-fil-markets/pull/570
- Iterate over all active storage deals on a Miner and register shards for all of them with the Sharded DAG Store so we can serve retrievals for deals made before the upgrade.
- Raise PR against https://github.com/filecoin-project/lotus/pull/6671.

---
~### Testground testing **_(MRA + CARv2 changes)_**~
- Upgraded client -> upgraded miner (Online storage deal and retrieval).
- Upgraded client -> upgraded miner (Offline storage deal and retrieval).
- Upgraded client-> old miner (Online storage deal and retrieval).
- Upgraded client-> old miner (Offline storage deal and retrieval).
- Old client -> upgraded miner  (Online storage deal and retrieval).
[...]

PR #6765, opened 2021-07-15
initial lotus-compat testplan

nonsense:
This PR is adding an initial testplan that automates the compatibility tests we want to run on Testground for Lotus, for example:
1. upgraded miner / old client
2. old client / upgraded miner

... and other similar compatibility tests, detailed under "Testground testing" at https://github.com/filecoin-project/lotus/issues/6673

---

You can run this testplan with: `testground run composition -f ./_compositions/lotus-compat.toml` assuming that your environment is connected to TaaS.

Logs from `lotus daemon` and `lotus miner` as well as the testplan are recorded in the testplan's outputs directory, so can be downloaded from TaaS after the test concludes.

---

Known issues:
- [ ] the two versions of Lotus that we are running compatibility tests on are currently hardcoded in the `Dockerfile` - these need to be dynamic
- [ ] ideally we want a real-time access to logs - we could achieve this by exposing an http server on TaaS pointing to the relevant directory under /outputs - for now we have to wait for the test plan to conclude in order to view logs.


magik6k:
Feel free to reopen if there is still need for this; today there seems to not be much need for testground tests

PR #6766, opened 2021-07-15
PriceListByVersion

### LINKED ISSUE #6652: PricelistByEpoch Technical Debt
From @Stebalien 
"
PricelistByEpoch has some pretty serious problems and I'm not sure how to cleanly fix them.
- It takes an epoch, not a network version.
- It applies the price-list at the upgrade epoch, instead of after the upgrade epoch. This is off by one.
- The VM calls PricelistByEpoch(currentEpoch)
- Sync/mpool uses the parent tipset's height.
The first step here would be to try to fix the off-by-one and test it around the calico upgrade. If that fails... this is going to be annoying to maintain going forward.
"

Additionally this has made running new auto-gen test-vectors on lotus hard because these assume the latest actors version is valid at height 0 with current price list.  We could remove that assumption during test-vector generation and use the epoch 0 pricelist but this is a step in the wrong direction.

ZenGround0:
Closes #6652


chain/sync.go
```diff
@@ -1054,14 +1054,15 @@ func (syncer *Syncer) checkBlockMessages(ctx context.Context, b *types.FullBlock
 		return xerrors.Errorf("failed to load base state tree: %w", err)
 	}
 
-	pl := vm.PricelistByEpoch(baseTs.Height())
+	nv := syncer.sm.GetNtwkVersion(ctx, b.Header.Height)
```
ZenGround0, line 1057: I'm changing this from the base tipset's version to the block including the message's version which I think is the desired behavior.

magik6k (approved):
Looks better than what we had, obviously https://github.com/filecoin-project/go-state-types/pull/30 should land first (probably with the bigger version number changes) (also need the api version bump)

chain/stmgr/forks.go
```diff
@@ -292,6 +296,18 @@ func (us UpgradeSchedule) Validate() error {
 	return nil
 }
 
+func (us UpgradeSchedule) GetNtwkVersion(e abi.ChainEpoch) (network.Version, error) {
```
ZenGround0, line 299: @magik6k I added this so that the mpool could get current network version without having to construct a new stmgr and depend on chainstore.  If you think this redundancy is unacceptable I can try a deeper refactor.

api/version.go
```diff
@@ -54,8 +54,8 @@ func VersionForType(nodeType NodeType) (Version, error) {
 
 // semver versions of the rpc api exposed
 var (
-	FullAPIVersion0 = newVer(1, 3, 0)
-	FullAPIVersion1 = newVer(2, 1, 0)
+	FullAPIVersion0 = newVer(1, 3, 1)
```
ZenGround0, line 57: @Stebalien @magik6k is there anything else needed to finish bumping the api version?

chain/sync.go
ZenGround0: Good call out.  I checked this against mainnet and this code syncs past the calico upgrade (265060 to 265241 so far) so we should be all good.

api/version.go
Kubuxu: I think that is it.

chain/stmgr/forks.go
Kubuxu: Seems fine with me

magik6k:
(marking as draft until we land https://github.com/filecoin-project/go-state-types/pull/30)

PR #6768, opened 2021-07-16
Exposed Splitstore: protect Has invocations 

vyzo:
It is conceivable that exposed blockstore users may want to optimize their outbound requests for data by invoking Has on the blockstore.
This protects such invocation so as not to break such optimizations.


vyzo:
Following sync discussion with @magik6k we decided that it is unnecessary and even detrimental; bitswap (the heavy exposed blockstore user) is used only for fetching messages we didn't get by pubsub while following the chain, and it also calls `Has` a lot in response to network requests.

PR #6771, opened 2021-07-16
upgrade data transfer stack; fix test.

raulk:
These new versions propagate queued transfers in Graphsync as a go-fil-markets state. `datatransfer.Ongoing` was not the right thing to check, as the data transfer is marked as ongoing _even if_ it's actually queued. What happens is as follows:

- data transfer adds a "request queued" hook to graphsync, which gets notified every time a request is queued in the graphsync layer.
- in data transfer, when the hook is called by graphsync, we emit an event, but we DO NOT transition the state (it stays as Ongoing).
- in go-fil-markets, we subscribe to data transfer events and DO respond to that event with a state transition to `ClientEventDataTransferQueued`.

IMO, we are skipping a level here and it should be the data tranfser channel that moves to "queued", instead of the client markets state machine.

But I'm sure @aarshkshah1992 @dirkmc have some thoughts here.


raulk:
This is incorrect. Closing in favour of https://github.com/filecoin-project/lotus/pull/6770.

PR #6777, opened 2021-07-17
Splitstore: add support for protecting out of chain references in the blockstore

vyzo:
This adds a mechanism for protecting out-of-chain references during compaction (and in the future gc).

The problem is that we actually have live references in the blockstore, that are not (yet) reachable from the chain. This arises from pending messages in the mpool, and can result in archiving/discarding such messages during compaction even though they are live.

The solution is to do what garbage collecting programming languages do since the beginning of time: add a hook to protect external references.
The hook is hooked[sic] to the mpool through DI.

Open question: are there other such references that need to be protected?


vyzo:
Following sync discussion with @Stebalien , we decided that we shouldn't indirect through the chainstore and instead hook the gc hooks directly through DI.
Less invasive this way, even if it is more indirect.

node/modules/chain.go
```diff
@@ -58,7 +58,7 @@ func ChainBlockService(bs dtypes.ExposedBlockstore, rem dtypes.ChainBitswap) dty
 	return blockservice.New(bs, rem)
 }
 
-func MessagePool(lc fx.Lifecycle, mpp messagepool.Provider, ds dtypes.MetadataDS, nn dtypes.NetworkName, j journal.Journal) (*messagepool.MessagePool, error) {
+func MessagePool(lc fx.Lifecycle, mpp messagepool.Provider, ds dtypes.MetadataDS, nn dtypes.NetworkName, j journal.Journal, protector dtypes.GCReferenceProtector) (*messagepool.MessagePool, error) {
```
Stebalien, line 61: This isn't really a _dependency_, so I assume this would be a separate "invoke" step (next to the splitstore step). But I don't have strong opinions.

Stebalien: Well, I have strong opinions. Just not about this.

chain/messagepool/messagepool.go
```diff
@@ -426,6 +426,27 @@ func New(api Provider, ds dtypes.MetadataDS, netName dtypes.NetworkName, j journ
 	return mp, nil
 }
 
+func (mp *MessagePool) ProtectMessages(protect func(cid.Cid) error) error {
```
Stebalien, line 429:
```suggestion
func (mp *MessagePool) ForEachPending(mcid func(cid.Cid) error) error {
```

Or something like that? Basically, the message pool isn't actually protecting _anything_. It's just iterating over its set of messages.

Stebalien (approved):
Some nits but this is much nicer.

node/modules/chain.go
vyzo: We could do it with an invoke step, but doing it here avoids all possible races I think.
No strong opinions on this either.

vyzo:
flaky test failure in https://app.circleci.com/pipelines/github/filecoin-project/lotus/15905/workflows/5d4be844-ce96-4922-ac4c-09503c0d3b3e/jobs/205203

Opened issue https://github.com/filecoin-project/lotus/issues/6799

node/modules/chain.go
vyzo: basically I want to make sure that `AddProtector` invocations all happen before `Start` in the splitstore.

Stebalien: I assume that `Start` gets called from a hook, which runs after all invokes (IIRC).

Issue #6792, opened 2021-07-19
[BUG] Runtime Segregation cause panic and crash


tmyuu:
**Describe the bug**
Runtime Segregation cause panic with subsystems of mining/sealing/proving node when publishing deals.

**Version (run `lotus version`):**
```
Daemon:  1.11.1-dev+mainnet+git.1c30d07d6+api1.3.0
Local: lotus version 1.11.1-dev+mainnet+git.1c30d07d6
```

**To Reproduce**
- Separating processes according to the [introduction](https://github.com/nonsense/filecoin-docs/blob/split-markets-miners/docs/mine/lotus/split-markets-miners.md).
- Running both nodes on the same machine with different ports.
- Accepting a deal and then publishing it.
- *All sealing processes are undertaken by lotus-worker as same as before the separation.

**Expected behavior**
Only lotus-miner mining/sealing/proving node shuts down.

**Logs**
miner with mining/sealing/providing log: 
```
2021-07-19T19:27:31.794+0900  DEBUG  advmgr sector-storage/sched.go:450   SCHED Acceptable win: [[2 3]]
2021-07-19T19:27:31.794+0900  DEBUG  advmgr sector-storage/sched.go:465   SCHED try assign sqi:0 sector 672 to window 2
2021-07-19T19:27:31.794+0900  DEBUG  advmgr sector-storage/sched.go:472   SCHED ASSIGNED sqi:0 sector 672 task seal/v0/addpiece to window 2
2021-07-19T19:27:31.794+0900  DEBUG  advmgr sector-storage/sched_worker.go:368   assign worker sector 672
2021-07-19T19:27:31.794+0900  DEBUG  advmgr sector-storage/sched.go:354   SCHED 0 queued; 4 open windows
2021-07-19T19:27:31.795+0900  DEBUG  advmgr sector-storage/sched_worker.go:278   task done    {"workerid": "f2ba1640-f9ce-41b6-8146-1c85ae48404c"}
2021-07-19T19:27:31.795+0900  DEBUG  advmgr sector-storage/sched.go:354   SCHED 0 queued; 4 open windows
[...]

PR #6800, opened 2021-07-20
wrap close(wait) with sync.Once to avoid panic

nonsense:
Closes https://github.com/filecoin-project/lotus/issues/6792

---

Fixes the panic described in the issue. Clients are reading up until EOF error, and at that point if `Close()` is called, the process was crashing.

### DIFF
--- a/lib/rpcenc/reader.go
+++ b/lib/rpcenc/reader.go
@@ -80,19 +80,24 @@ func ReaderParamEncoder(addr string) jsonrpc.Option {
 
 type waitReadCloser struct {
 	io.ReadCloser
-	wait chan struct{}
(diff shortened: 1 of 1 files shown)


lib/rpcenc/reader.go
```diff
@@ -134,7 +145,9 @@ func ReaderParamDecoder() (http.HandlerFunc, jsonrpc.ServerOption) {
 		}
 
 		select {
-		case <-wr.wait:
+		case <-wr.watch:
+			// TODO should we check if we failed the Read, and if so
```
Stebalien, line 149: Yes, but a followup patch sounds reasonable.

PR #6803, opened 2021-07-20
Configurable CC Sector Expiration

### LINKED ISSUE #4760: Allow miners to set expiration duration for sectors
The sectors expiration duration can be any time from 180 days - 540 days. Currently, all the cc sectors' expiration duration is set to 540 days by default, and we do not expose any cli for miners to customize it. Sectors with deals are expired when the last deal is ended.

Ideally, we should allow miners to set the below values in the configuration:

- [x] #6836

- [ ] minimum expiration duration for deal sectors before sealing

### LINKED ISSUE #6836: default expiration duration for cc sectors
Child task of #4760.

placer14:
(fixes #6836)

Adds config:
```
 	// CommittedCapacitySectorLifetime is the duration a Committed Capacity (CC) sector will
	// live before it must be extended or converted into sector containing deals before it is
	// terminated. Value must be between 180-540 days inclusive
	CommittedCapacitySectorLifetime Duration
```
Related #4760


node/config/def.go
```diff
@@ -131,6 +131,12 @@ type SealingConfig struct {
 
 	WaitDealsDelay Duration
 
+	// CommittedCapacityDefaultLifetime is the default duration a Committed Capacity (CC)
+	// sector will live before it must be extended or converted into sector containing deals
+	// before it is terminated.
+	// Value must be between 180-540 days inclusive.
+	CommittedCapacityDefaultLifetime Duration
```
placer14, line 138: Should I consider including some validation around this value's range on-load? (I assume the actor protects its inputs and produces a runtime error, but I imagine catching that on startup might be a better experience?) 

If so, where is the best place to put this check? I quickly walked config init back to the node builders which had a smattering of config validations, but didn't seem like the right place for it. Suggestions?

storage/miner.go
```diff
@@ -203,6 +203,16 @@ func (m *Miner) Run(ctx context.Context) error {
 	return nil
 }
 
+func getDefaultSectorExpirationExtension(cfg sealing.GetSealingConfigFunc) abi.ChainEpoch {
```
magik6k, line 206: I think we should wire the config getter into the precommit policy, and use the up-to-date value on each call to get sector expiration (also do the validation there).

node/config/def.go
magik6k: As noted in the other comment, we should check this in the precommit policy.

(we also need to set a default value below in this file)

placer14:
Rebased onto latest `master` branch. ^

extern/storage-sealing/precommit_policy.go
```diff
@@ -79,11 +83,41 @@ func (p *BasicPreCommitPolicy) Expiration(ctx context.Context, ps ...Piece) (abi
...
+		return 0, xerrors.Errorf("sealing config load error: %w", err)
+	}
+
+	var ccLifetimeEpochs = abi.ChainEpoch(uint64(c.CommittedCapacitySectorLifetime.Truncate(builtin.EpochDurationSeconds).Seconds()) / builtin.EpochDurationSeconds)
+	// if zero value in config, assume maximum sector extension
+	if ccLifetimeEpochs == 0 {
+		ccLifetimeEpochs = policy.GetMaxSectorExpirationExtension()
+	}
+
+	if minExpiration := abi.ChainEpoch(miner.MinSectorExpiration); ccLifetimeEpochs < minExpiration {
```
placer14, line 113: I tried to put this into the `policy` package, but it appears to be autogen'd code. LMK if you'd like for me to abstract this upstream.

extern/storage-sealing/precommit_policy_test.go
```diff
@@ -38,21 +58,49 @@ func fakePieceCid(t *testing.T) cid.Cid {
...
+	exp, err := pcp.Expiration(context.Background())
+
+	require.NoError(t, err)
+
+	// as set when there are no deal pieces
+	expected := h + policy.GetMaxSectorExpirationExtension() - (pBuffer * 2)
+	// as set just before returning within Expiration()
+	expected += miner.WPoStProvingPeriod - (expected % miner.WPoStProvingPeriod) + pBoundary - 1
+	assert.Equal(t, int(expected), int(exp))
+}
```
placer14, line 75: The math is a little obtuse here so I put literal `expected`s for both paths (zero config for CC and non-CC sectors) as well as show how the custom value is applied. I kept the later units with literals in the assert and simply updated the values.

extern/storage-sealing/precommit_policy_test.go
```diff
@@ -112,13 +162,14 @@ func TestBasicPolicyIgnoresExistingScheduleIfExpired(t *testing.T) {
 	exp, err := policy.Expiration(context.Background(), pieces...)
 	require.NoError(t, err)
 
-	assert.Equal(t, 2879, int(exp))
+	assert.Equal(t, 1558079, int(exp))
```
placer14, line 165: I didn't bother showing the math in this (and later) cases as the above units cover those paths. (I did manually check the math on these changes.)

storage/miner.go
```diff
@@ -179,19 +178,17 @@ func (m *Miner) Run(ctx context.Context) error {
...
 
 		// Instantiate a precommit policy.
-		defaultDuration = policy.GetMaxSectorExpirationExtension() - (md.WPoStProvingPeriod * 2)
+		cfg             = sealing.GetSealingConfigFunc(m.getSealConfig)
 		provingBoundary = md.PeriodStart % md.WPoStProvingPeriod
+		provingBuffer   = md.WPoStProvingPeriod * 2
 
 		// TODO: Maybe we update this policy after actor upgrades?
-		pcp = sealing.NewBasicPreCommitPolicy(adaptedAPI, defaultDuration, provingBoundary)
+		pcp = sealing.NewBasicPreCommitPolicy(adaptedAPI, cfg, provingBoundary, provingBuffer)
```
placer14, line 186: Arg list is getting a bit long? LMK if you'd like me to extract an options struct out of this.

extern/storage-sealing/precommit_policy_test.go
```diff
@@ -38,21 +58,49 @@ func fakePieceCid(t *testing.T) cid.Cid {
...
 
-	exp, err := policy.Expiration(context.Background())
 	require.NoError(t, err)
 
-	assert.Equal(t, 2879, int(exp))
+	// as set when there are no deal pieces
+	expected := h + customLifetimeEpochs - (pBuffer * 2)
+	// as set just before returning within Expiration()
+	expected += miner.WPoStProvingPeriod - (expected % miner.WPoStProvingPeriod) + pBoundary - 1
+	assert.Equal(t, int(expected), int(exp))
```
placer14, line 94: I'm pretty sure this math is unchanged, but please double-check this against the impl. 🙇

extern/storage-sealing/precommit_policy.go
magik6k: `policy.go` is generated form  `policy.go.template`, so you'd just put this in that file, and run `make actors-gen`

(I'm also not sure if it's worth putting this code there since it doesn't do anything actor-version specific, and likely won't be used outside of the sealing package)

extern/storage-sealing/precommit_policy.go
```diff
@@ -79,11 +83,41 @@ func (p *BasicPreCommitPolicy) Expiration(ctx context.Context, ps ...Piece) (abi
...
 	return *end, nil
 }
+
+func (p *BasicPreCommitPolicy) getCCSectorLifetime() (abi.ChainEpoch, error) {
+	c, err := p.getSealingConfig()
+	if err != nil {
+		return 0, xerrors.Errorf("sealing config load error: %w", err)
+	}
+
+	var ccLifetimeEpochs = abi.ChainEpoch(uint64(c.CommittedCapacitySectorLifetime.Truncate(builtin.EpochDurationSeconds).Seconds()) / builtin.EpochDurationSeconds)
```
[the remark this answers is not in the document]
magik6k: (also pretty sure we don't need this truncate as the integer division should take care of that anyways)

magik6k (commented):
Looks good, just one tiny issue

extern/storage-sealing/precommit_policy.go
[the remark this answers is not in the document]
placer14: Not sure what you were meaning re cast to Duration. But good catch on the extraneous Truncate.

placer14:
@magik6k This should be ready to merge, despite some testground PRs which are hanging. I've kicked them off again to see if they'll resolve.

arajasek (approved):

PR #6804, opened 2021-07-20
test: fix flaky window post tests

Stebalien:
Wait until the network upgrade is finished. If we try to seal while it's happening, we have a few annoying edge cases that can fail if we try to submit some messages right on the upgrade epoch (which is why everyone turns that kind of stuff off for the upgrade epoch).

### DIFF
--- a/itests/wdpost_test.go
+++ b/itests/wdpost_test.go
@@ -213,11 +213,17 @@ func TestWindowPostBaseFeeNoBurn(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	sched := kit.DefaultTestUpgradeSchedule
+	lastUpgradeHeight := sched[len(sched)-1].Height
+
 	och := build.UpgradeClausHeight
-	build.UpgradeClausHeight = 10
+	build.UpgradeClausHeight = lastUpgradeHeight + 1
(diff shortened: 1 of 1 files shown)


Kubuxu (approved):

itests/wdpost_test.go
```diff
@@ -268,6 +274,12 @@ func TestWindowPostBaseFeeBurn(t *testing.T) {
 	client, miner, ens := kit.EnsembleMinimal(t, kit.MockProofs(), opts)
 	ens.InterconnectAll().BeginMining(blocktime)
 
+	// Ideally we'd be a bit more precise here, but getting the information we need from the
+	// test framework is more work than it's worth.
+	//
+	// We just need to wait till all upgrades are done.
+	client.WaitTillChain(ctx, kit.HeightAtLeast(20))
```
raulk, line 281: @Stebalien we could add a predicate that tests against `StateNetworkVersion`? We'd need to expand the signature of `ChainPredicate` so we can pass in the FullNodeAPI.

Stebalien: Yeah. That would work. I guess I could also just pass a closure here and capture it.

raulk: Good idea.

PR #6805, opened 2021-07-21
fix: on randomness change, use new rand

Stebalien:


storage/wdpost_run.go
```diff
@@ -659,6 +659,7 @@ func (s *WindowPoStScheduler) runPoStCycle(ctx context.Context, di dline.Info, t
 
 				if !bytes.Equal(checkRand, rand) {
```
ZenGround0, line 660: I'm trying to understand how this can happen.  The challenge epoch should be a constant and my limited understanding of beacon randomness is that the epoch to beacon mapping never changes.  Chain reorgs should also never change which randomness goes with which epoch. Is the last condition not being met because of the beacon bug we didn't fully fix in hyperdrive?

ZenGround0 (approved):
LGTM

storage/wdpost_run.go
Stebalien: > Is the last condition not being met because of the beacon bug we didn't fully fix in hyperdrive?

Yes.

Kubuxu (approved):

magik6k (approved):

Kubuxu:
Automerge won't merge it as diff is not covered by tests.

PR #6806, opened 2021-07-21
capitalize lotus mount fields for deserialization

dirkmc:


raulk (commented):
I think only `Api` needs to be public, so the templating logic can copy it. `PieceCid` is set by `Deserialize()`, so it could stay private. But it doesn't hurt to have them both as exported ;-)

PR #6807, opened 2021-07-21
ClientFindData: always fetch peer id from chain

nonsense:
This PR is fixing `ClientFindData` when data about peers in the local store of a node is out-dated.

---

If a client makes a deal with a miner, it keeps the miner's peer id in a local store. In the future if the miner changes its peer id, and the client tries to retrieve the content (datacid), `ClientFindData` would still be trying to reach the miner on its old peer id, ultimately failing.

This is manifested with `Failed to find file` error in the CLI:

```
➜  lotus git:(master) lotus client retrieve $datacid /tmp/file1234
Failed to find file
```

### DIFF
--- a/node/impl/client/client.go
+++ b/node/impl/client/client.go
@@ -436,7 +436,19 @@ func (a *API) ClientFindData(ctx context.Context, root cid.Cid, piece *cid.Cid) 
 		if piece != nil && !piece.Equals(*p.PieceCID) {
 			continue
 		}
-		out = append(out, a.makeRetrievalQuery(ctx, p, root, piece, rm.QueryParams{}))
+
+		// do not rely on local data with respect to peer id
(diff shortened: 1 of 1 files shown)


dirkmc (approved):
LGTM 👍

Kubuxu (approved):

raulk (approved):

raulk (approved):
Can we have a test for this in this PR, or an issue to track? 🙏

magik6k (approved):

BigLep:
@nonsense : did the test get added?

nonsense:
@BigLep I don't think there is a test for this case.

We will have one as soon as we finish the lotus-compatibility tests part of M1 (initial PR at https://github.com/filecoin-project/lotus/pull/6765 )

BigLep:
Beautiful - thanks.

PR #6811, opened 2021-07-21
splitstore shed utils

### LINKED ISSUE #6737: [Feature Request] Splitstore: Copy hot store data to cold store for rollbacks
The split store treats the monolith store as the coldstore, user will have trouble syncing if they disable splitstore after have it enabled for syncing before as data that's in the hotstore are missing. (Related #6088 - won't fix)

That being said, it will be nice if we can have a tool, probably in lotus-shed, to copy hotstore data to the cold store when a rollback is necessary so user won't have to resync from a snapshot.

vyzo:
Adds a `lotus-shed splitstore` command with some utilities:
- [X] `rollback` command, which rolls back the splitstore.
- [x] `check` which performs a health check.
- [x] `info` which prints some basic information.

Closes #6737


magik6k:
Would be nice to also have a light `info` command which would get you blockstore sizes, and maybe some state (e.g. epochs in hotstore + maybe some things which could be useful for debugging user nodes)

vyzo:
sure!

cmd/lotus-shed/splitstore.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/ipfs/go-datastore/query"
+	logging "github.com/ipfs/go-log/v2"
+
+	"github.com/filecoin-project/lotus/node/config"
+	"github.com/filecoin-project/lotus/node/repo"
+)
+
+var splitstoreCmd = &cli.Command{
+	Name:        "splitstore",
+	Description: "splitstore utiities",
```
raulk, line 27:
```suggestion
	Description: "splitstore utilities",
```

cmd/lotus-shed/splitstore.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	// compact + gc the coldstore
+	fmt.Println("compacting coldstore...")
+	nworkers := runtime.NumCPU()
+	if nworkers < 2 {
+		nworkers = 2
+	}
+
+	err = cold.Flatten(nworkers)
```
raulk, line 180: How long can this take? Because if it takes more than a couple minutes, it's probably best to have it as an option as it's not functionally necessary.

cmd/lotus-shed/splitstore.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return xerrors.Errorf("error compacting coldstore: %w", err)
+	}
+
+	fmt.Println("garbage collecting coldstore...")
+	for err == nil {
+		err = cold.RunValueLogGC(0.0625)
```
raulk, line 187: Ditto.

cmd/lotus-shed/splitstore.go
```diff
@@ -0,0 +1,238 @@
...
+			return xerrors.Errorf("splitstore is not enabled")
+		}
+
+		fmt.Println("copying hotstore to coldstore...")
+		err = copyHotstoreToColdstore(lr)
+		if err != nil {
+			return xerrors.Errorf("error copying hotstore to coldstore: %w", err)
+		}
+
+		fmt.Println("deleting splistore directory...")
```
raulk, line 84:
```suggestion
		fmt.Println("deleting splitstore directory...")
```

cmd/lotus-shed/splitstore.go
```diff
@@ -0,0 +1,238 @@
...
+		err = deleteSplitstoreDir(lr)
```
raulk, line 85: Are we sure we don't want this behind an option?

raulk (commented):
Looks sane.

cmd/lotus-shed/splitstore.go
vyzo: no, it shouldn't take more than a couple of minutes normally but it might for big coldstores.
So yeah, I'll put it behind an option.

cmd/lotus-shed/splitstore.go
vyzo: they should go together behind the same option.

cmd/lotus-shed/splitstore.go
vyzo: There is no point in keeping this once we've copied the hotstore to the coldstore.

cmd/lotus-shed/splitstore.go
vyzo: done.

cmd/lotus-shed/splitstore.go
vyzo: done.

jacobheun:
@vyzo can you ensure that https://github.com/filecoin-project/lotus/blob/master/blockstore/splitstore/README.md is updated to include a section on how to rollback when this is ready?

vyzo:
ah yes, thanks for reminding me -- will do.

vyzo:
rebased on master.

vyzo:
rebased on master.

api/api_full.go
```diff
@@ -164,6 +164,13 @@ type FullNode interface {
 	// If oldmsgskip is set, messages from before the requested roots are also not included.
 	ChainExport(ctx context.Context, nroots abi.ChainEpoch, oldmsgskip bool, tsk types.TipSetKey) (<-chan []byte, error) //perm:read
 
+	// ChainCheckBlockstore performs an (asynchronous) health check on the chain/state blockstore
+	// if supported by the underlying implementation.
+	ChainCheckBlockstore(context.Context) error //perm:read
```
magik6k, line 169: Since this writes to a file in splitstore dir, and isquite expensive, I'd make the perm 'admin' ('write' could also work, but it's more of a node maintenance call, so admin fits better)

magik6k (commented):
Looks good, just one nit about perms

api/api_full.go
vyzo: ah yes, very good point; fixing.

vyzo: done!

magik6k (approved):

PR #6817, opened 2021-07-21
Introduce the LOTUS_CHAIN_BADGERSTORE_DISABLE_FSYNC envvar

ribasushi:
This is a redux of https://github.com/filecoin-project/lotus/pull/5050

Allows the user to control the opts.SyncWrites option of BadgerDs, with the tri-state environment variable `LOTUS_CHAIN_BADGERSTORE_DISABLE_FSYNC`

- unset == the default (currently fsync enabled)
- set with a false-y value ( `""`, `"0"`, `"no"`, `"false"` ) == fsync enabled no matter what a future default is
- set with any other value == fsync is disabled ignored defaults (recommended for day-to-day use)

Applying this brought @whyrusleeping's average chain validation times from "impossible to sync" ( > 30s ) down to 1~3s/tipset

### DIFF
--- a/node/repo/fsrepo.go
+++ b/node/repo/fsrepo.go
@@ -327,6 +327,21 @@ func (fsr *fsLockedRepo) Blockstore(ctx context.Context, domain BlockstoreDomain
 			return
 		}
 
+		//
+		// Tri-state environment variable LOTUS_CHAIN_BADGERSTORE_DISABLE_FSYNC
+		// - unset == the default (currently fsync enabled)
+		// - set with a false-y value == fsync enabled no matter what a future default is
+		// - set with any other value == fsync is disabled ignored defaults (recommended for day-to-day use)
+		//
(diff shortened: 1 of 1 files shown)


magik6k (commented):
It would be nice if we could use the config for this.

(also should apply to the splitstore hotstore badger)

ribasushi:
@magik6k there is opposition to this option on the team ( see previous issue ), so the suggested way forward was an undocumented envvar. I can certainly plumb this through the config, but I am not sure exposing this functionality is generally desired....

> (also should apply to the splitstore hotstore badger)

@vyzo can you help with a pointer which spots should be covered? ( having a bit of trouble navigating through the new codebase )

vyzo:
The spllitstore badger uses the fsrepo option constructors, so we should be covered by this.

magik6k (approved):

PR #6833, opened 2021-07-22
Splitstore: support on-disk marksets using badger

vyzo:
Very useful if you run out of memory during compaction.


vyzo:
Notes:
- marking is about twice as slow than with the in-memory map, but we have no memory pressure.
- block validation time is unaffected, thanks to the `FileIO` badger options in the markset.
- removed the markset occurs check from `trackTxnRefs` so as not to hit the markset; that's a non-critical optimization that doesn't really affect anything.
- added some documentation for configuration options in the README.

blockstore/splitstore/README.md
```diff
@@ -27,9 +27,34 @@ If you intend to use the discard coldstore, your also need to add the following:
     ColdStoreType = "discard"
```
raulk, line 27: Can we add a big warning saying that the usage of `discard` is discouraged unless you absolutely know what you're doing?

blockstore/splitstore/README.md
```diff
@@ -27,9 +27,34 @@ If you intend to use the discard coldstore, your also need to add the following:
...
+
+- `HotStoreType` -- specifies the type of hotstore to use.
+  The only currently supported option is `"badger"`.
+- `ColdStoreType` -- specifies the type of coldstore to use.
+  The default value is `"universal"`, which will use the initial monolith blockstore
+  as the coldstore.
+  The other possible value is `"discard"`, as outlined above, which is specialized for
+  running without a coldstore. Note that the discard store wraps the initial monolith
+  blockstore and discards writes; this is necessary to support syncing from a snapshot.
+- `MarkSetType` -- specifies the type of markset to use during compaction.
```
raulk, line 47: Have we defined what a "markset" is? If not, it makes this hard to follow.

blockstore/splitstore/markset_badger.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("error creating markset directory: %w", err)
+	}
+
+	opts := badger.DefaultOptions(path)
+	opts.SyncWrites = false
+	opts.CompactL0OnClose = false
+	opts.Compression = options.None
+	opts.TableLoadingMode = options.FileIO
+	opts.ValueLogLoadingMode = options.FileIO
```
raulk, line 68: Can we add a comment for our future selves here, on why we're using `FileIO`?

blockstore/splitstore/markset_badger.go
```diff
@@ -0,0 +1,225 @@
...
+	err := os.MkdirAll(msPath, 0755) //nolint:gosec
+	if err != nil {
+		return nil, xerrors.Errorf("error creating markset directory: %w", err)
+	}
+
+	return &BadgerMarkSetEnv{path: msPath}, nil
+}
+
+func (e *BadgerMarkSetEnv) Create(name string, sizeHint int64) (MarkSet, error) {
+	path := filepath.Join(e.path, name)
```
raulk, line 50: Do we need to add a salt to this filename on the caller side? `Create` always get called with a fixed name `live`. This means that:

1. on restart, we will open an _existing_ markset. However, this is unsafe because `SyncWrites` below is false!
2. a failure to `Close` and `Remove` will end up opening a previously existing database.

If we always want to operate on the same database, we need to ensure that reopening it will always be safe. However, there's probably little value in doing that because we do not resume marking on a process restart. So the recommendation here is to add a salt.

OTOH, an interruption could lead to leftover/abandoned markset databases.

blockstore/splitstore/markset_badger.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, xerrors.Errorf("error clearing markset directory: %w", err)
+	}
+
+	err = os.MkdirAll(path, 0755) //nolint:gosec
+	if err != nil {
+		return nil, xerrors.Errorf("error creating markset directory: %w", err)
+	}
+
+	opts := badger.DefaultOptions(path)
+	opts.SyncWrites = false
```
raulk, line 64: Consider flipping to `true` depending on the above.

blockstore/splitstore/markset_badger.go
```diff
@@ -0,0 +1,225 @@
...
+
+	err := db.Close()
+	if err != nil {
+		return xerrors.Errorf("error closing badger markset: %w", err)
+	}
+
+	err = os.RemoveAll(s.path)
+	if err != nil {
+		return xerrors.Errorf("error deleting badger markset: %w", err)
+	}
```
raulk, line 210: Close is not idempotent. If called multiple times or concurrently, only the caller that gets to execute the close logic will see any errors. Subsequent callers would not see this error. Consider storing a `closeErr` and returning that when `s.pend == nil` above.

blockstore/splitstore/splitstore_compact.go
```diff
@@ -184,16 +184,6 @@ func (s *SplitStore) trackTxnRef(c cid.Cid) {
...
-	if s.txnProtect != nil {
-		mark, err := s.txnProtect.Has(c)
-		if err != nil {
-			log.Warnf("error checking markset: %s", err)
-			// track it anyways
-		} else if mark {
-			return
-		}
-	}
-
```
raulk, line 196: Why remove this?

blockstore/splitstore/splitstore_compact.go
```diff
@@ -209,27 +199,11 @@ func (s *SplitStore) trackTxnRefMany(cids []cid.Cid) {
...
-					log.Warnf("error checking markset: %s", err)
-				}
-				// track it anyways
-			}
-
-			if mark {
-				continue
-			}
-		}
-
```
raulk, line 232: Why remove this?

blockstore/splitstore/README.md
vyzo: well, it's not exactly discouraged, but you should definitely know what you are doing!
I'll add something along this lines.

blockstore/splitstore/README.md
vyzo: Sure, I'll define it.

blockstore/splitstore/markset_badger.go
vyzo: Yeah, good call -- this is kind of critical.

blockstore/splitstore/markset_badger.go
vyzo: No, on restart we just delete everything -- marksets are transient.

blockstore/splitstore/markset_badger.go
vyzo: Nope, not needed -- the existing markset is deleted on restart.

blockstore/splitstore/markset_badger.go
vyzo: Maybe, but It doesn't really matter, it's a throwaway datastore!

blockstore/splitstore/splitstore_compact.go
vyzo: so that we don't hit the markset and potentially stall an I/O operation.

blockstore/splitstore/splitstore_compact.go
vyzo: same as above.

blockstore/splitstore/markset_badger.go
vyzo: Actually when we reopen it.

blockstore/splitstore/markset_badger.go
vyzo: Correction: when we reopen the markset, it's deleted.

So basically there is no safety concerns from close errors. The close implementation deletes the thing, and when we reopen it we delete again just to be sure we didn't have left overs.

Basically the on-disk markset is completely transient, we don't care at all about persisting it once we are done using it.

blockstore/splitstore/README.md
vyzo: Added a warning.

blockstore/splitstore/README.md
vyzo: Added a definition.

blockstore/splitstore/markset_badger.go
vyzo: On the other hand, using a salt would leave us with leftover marksets we can't consistently delete on restart, basically garbage on disk.
So using a salt is not the right thing to do here, and the code behaves as desired afaict.

blockstore/splitstore/markset_badger.go
vyzo: Also, it would slow down things significantly.

blockstore/splitstore/markset_badger.go
vyzo: done.

blockstore/splitstore/markset_badger.go
raulk: Ack. Makes sense.

blockstore/splitstore/markset_badger.go
raulk: Ack.

blockstore/splitstore/markset_badger.go
raulk: Happy to go this way.

raulk (approved):

magik6k (approved):

PR #6840, opened 2021-07-22
lotus-shed: initial export cmd for markets related metadata

nonsense:
This PR is adding commands to `lotus-shed` so that miners can rollback from a multi-process architecture - one where part of the repo metadata datastore is located on the `markets` node, and part of it on the sealing/proving node.

During rollback, we need to copy back all the relevant metadata keys/values back to the monolith repo metadata datastore.

---

TODO:
- [x] 1. a backup command for the markets-relevant datastore keys/values (already done by @dirkmc )
- [x] 2. a restore command that would write all keys/values from 1. on top of an existing metadata datastore
- [x] 3. update docs on how to use these commands: https://github.com/filecoin-project/filecoin-docs/pull/972

### DIFF
--- a/cmd/lotus-shed/market.go
+++ b/cmd/lotus-shed/market.go
@@ -31,6 +31,7 @@ var marketCmd = &cli.Command{
 	Subcommands: []*cli.Command{
 		marketDealFeesCmd,
 		marketExportDatastoreCmd,
+		marketImportDatastoreCmd,
 	},
 }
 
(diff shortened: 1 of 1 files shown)


cmd/lotus-shed/market.go
```diff
@@ -100,3 +114,196 @@ var marketDealFeesCmd = &cli.Command{
...
+	Name:        "import-datastore",
+	Description: "import markets datastore key/values from a backup file",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "repo",
+			Usage: "path to the repo",
+		},
+		&cli.StringFlag{
+			Name:     "backup-path",
+			Usage:    "path to the backup directory",
```
arajasek, line 247:
```suggestion
			Usage:    "path to the backup file",
```

arajasek (approved):
LGTM, only suggestion would be to perhaps add a verbose flag to the export -- I'm not sure if it's currently very noisy.

raulk (approved):

PR #6775, opened 2021-07-17
Splitstore: add retention policy option for keeping messages in the hotstore

vyzo:
It appears that deals have a longish lookback for messages; this adds an option to accommodate them.

cc @ribasushi

Issue #6834, opened 2021-07-22
splitstore: Add retention policy test


magik6k:
Was promised in https://github.com/filecoin-project/lotus/pull/6775#issuecomment-882771136

PR #6841, opened 2021-07-22
fix warmup by decoupling state from message receipt walk

vyzo:
this was accidentally broken in #6775.

### DIFF
--- a/blockstore/splitstore/splitstore_compact.go
+++ b/blockstore/splitstore/splitstore_compact.go
@@ -436,7 +436,7 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 	startMark := time.Now()
 
 	var count int64
-	err = s.walkChain(curTs, boundaryEpoch, inclMsgsEpoch, boundaryEpoch,
+	err = s.walkChain(curTs, boundaryEpoch, inclMsgsEpoch,
 		func(c cid.Cid) error {
 			if isUnitaryObject(c) {
 				return errStopWalk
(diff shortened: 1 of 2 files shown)


Kubuxu (approved):
SGWM but could use tests

vyzo:
Yeah, tests forthcoming soon -- we hav #6834 open.
I probably won't get to it today, as there are a couple of things ahead of it in the queue, but I'll get to it asap.

blockstore/splitstore/splitstore_compact.go
```diff
@@ -631,14 +631,16 @@ func (s *SplitStore) endTxnProtect() {
 	s.txnMissing = nil
 }
 
-func (s *SplitStore) walkChain(ts *types.TipSet, inclState abi.ChainEpoch, inclMsgs abi.ChainEpoch,
+func (s *SplitStore) walkChain(ts *types.TipSet, inclState, inclMsgs, inclMsgReceipts abi.ChainEpoch,
```
magik6k, line 634: Why separate height for receipts?

vyzo: well, I wanted to decouple it from state as it was the source of the bug -- when we sync from snapshot we dont have message receipts at all!

We could ostensibly couple them with the messages, but I don't think we need them at all in order to support the use cases -- please correct me if I am wrong :)

vyzo: following sync discussion, we discussed to collate them with message retention.

magik6k (approved):

Issue #5291, opened 2021-01-05
Provide storage miner with configurable path for offline deal data when staging for sealing.


f8-ptrk:
Issue1: When importing data for an offline deal the fstmp files are placed directly in the LOTUS_MINER_PATH (possibly in a sealing folder, not sure - but for sure not in a subfolder) folder. that robs miners of the possibility to organize offline deals well.

Issue2: the fstmp filenames are not useful. 

**To Reproduce**
1. Run 'lotus-miner storage-deals import-data someCID /path/to/import/data.car'
2. See LOTUS_MINER_PATH

**Expected behavior**
The files in a folder one could link another disk/partition/location to. Files with a name connected to the deal/imported file-name

**Version (run `lotus version`):**
Daemon:  1.4.0+git.e9989d0e4+api1.0.0
Local: lotus-miner version 1.4.0+git.e9989d0e4

**Additional context**
Nothing critical...

s0nik42:
Can these files being removed safely ?

PatrickDeuse:
@s0nik42 they should get removed as soon as the deal goes to Active. just from watching terminal windows while sealing offline deals.

s0nik42:
@ribasushi @jennijuju , is this going to be solved at some point ? my miner is now creating 15GB of logs a day!, millions of useless connections to peers and with slingshot ending, in 3 days it's almost going to double as I've got a lot of deals that will never be served. 

Thank you,

frrist:
Exploring this code for the first time, going to leave some bread crumbs for anyone else interested in following along as I suspect some thoughtful design decisions will need to be made before proceeding with a solution.

1. The `lotus-storage-miner` has a single repo that is initialized in the location specified by `LOTUS_MINER_PATH`([code](https://github.com/filecoin-project/lotus/blob/7a78527fdeaa948b1fa4631d2738d2538db863f/cmd/lotus-storage-miner/init.go#L176-L205))

2. When the storage miner is instructed to `run`, a storage miner node is instantiated using the aforementioned repo ([code](https://github.com/filecoin-project/lotus/blob/7a78527fdeaa948b1fa4631d2738d2538db863fa/cmd/lotus-storage-miner/run.go#L145-L156))

[...]

PR #6853, opened 2021-07-23
feat(miner deals): create subdir to miner repo for staged deals

frrist:
Closes https://github.com/filecoin-project/lotus/issues/5291

### DIFF
--- a/node/modules/storageminer.go
+++ b/node/modules/storageminer.go
@@ -607,7 +607,7 @@ func StorageProvider(minerAddress dtypes.MinerAddress,
 	df dtypes.StorageDealFilter,
 ) (storagemarket.StorageProvider, error) {
 	net := smnet.NewFromLibp2pHost(h)
-	store, err := piecefilestore.NewLocalFileStore(piecefilestore.OsPath(r.Path()))
+	store, err := piecefilestore.NewLocalFileStore(piecefilestore.OsPath(filepath.Join(r.Path(), "deal-staging")))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


node/modules/storageminer.go
```diff
@@ -607,7 +607,7 @@ func StorageProvider(minerAddress dtypes.MinerAddress,
 	df dtypes.StorageDealFilter,
 ) (storagemarket.StorageProvider, error) {
 	net := smnet.NewFromLibp2pHost(h)
-	store, err := piecefilestore.NewLocalFileStore(piecefilestore.OsPath(r.Path()))
+	store, err := piecefilestore.NewLocalFileStore(piecefilestore.OsPath(filepath.Join(r.Path(), "deal-staging")))
```
frrist, line 610: Next steps could be to make this path value configurable via the miner config (I'm sure this would make @travisperson's life a bit easier). If we think it's worth doing I can add it in.

node/repo/memrepo.go
```diff
@@ -107,6 +107,12 @@ func (lmem *lockedMemRepo) Path() string {
 	}
 
 	if lmem.t == StorageMiner {
+		// this is required due to the method makeDealStaging from cmd/lotus-storage-miner/init.go
+		// deal-staging is the directory deal files are staged in before being sealed into sectors
+		// for offline deal flow.
+		if err := os.MkdirAll(filepath.Join(t, "deal-staging"), 0755); err != nil {
```
frrist, line 113: This is kinda ugly, but I didn't want to make the Repo responsible for creating this directory since it is used by things other than the miner.

ZenGround0 (approved):

PR #6848, opened 2021-07-23
Config UX improvements

magik6k:
This PR:
* Moves config types to a separate file
* Adds codegen for docstring map
* Adds docstring comments to default config
* Adds `lotus config default` and `lotus-miner config default` commands to print up-to-date default config
* Adds `lotus config updated` and `lotus-miner config updated` commands which print current node config with filled in, commented out default values (which might have been added in new node version)

Example miner 'updated' config output:
```toml
[API]
  # Binding address for the Lotus API
  #
  # type: string
  ListenAddress = "/ip4/10.x.x.x/tcp/2345/http"
  # type: string
  RemoteListenAddress = "10.x.x.x:2345"
  # type: Duration
  #Timeout = "30s"


[Backup]
  # Note that in case of metadata corruption it might be much harder to recover
  # your node if metadata log is disabled
  #
  # type: bool
  #DisableMetadataLog = false


[Libp2p]
  # Binding address for the libp2p host - 0 means random port.
  # Format: multiaddress; see https://multiformats.io/multiaddr/
  #
  # type: []string
  ListenAddresses = ["/ip4/0.0.0.0/tcp/x", "/ip6/::/tcp/0"]
  # Addresses to explicitally announce to other peers. If not specified,
  # all interface addresses are announced
  # Format: multiaddress
  #
  # type: []string
  AnnounceAddresses = ["/ip4/x.x.x.x/tcp/x"]
  # Addresses to not announce
  # Format: multiaddress
  #
  # type: []string
  #NoAnnounceAddresses = []

  # type: uint
  ConnMgrLow = 3500
  # type: uint
  ConnMgrHigh = 4800
  # type: Duration
  #ConnMgrGrace = "20s"


[Pubsub]
  # Run the node in bootstrap-node mode
  #
  # type: bool
  #Bootstrapper = false

  # type: string
  #RemoteTracer = "/dns4/pubsub-tracer.filecoin.io/tcp/4001/p2p/QmTd6UvR47vUidRNZ1ZKXHrAFhqTJAD27rKL9XYghEKgKX"


[Subsystems]
  # type: bool
  #EnableMining = true

  # type: bool
  #EnableSealing = true

  # type: bool
  #EnableSectorStorage = true

  # type: bool
  #EnableMarkets = true

  # type: string
  #SealerApiInfo = ""

  # type: string
  #SectorIndexApiInfo = ""


[Dealmaking]
  # When enabled, the miner can accept online deals
  #
  # type: bool
[...]

PR #6854, opened 2021-07-23
Moving GC for badger

vyzo:
Cherry-picked from #6728.

This adds the option to perform moving GC on a badger blockstore; this effectively reclaims all space.
It is also hooked into the splitstore, to perform moving GC once every 20 compactions (about once a week, user configurable).

Rationale: badger supports online GC, which is quite fast but ultimately ineffective at reclaiming space. It has been observed that the hotstore size slowly creeps up, despite frequent garbage collections.
This addresses the situation with real moving GC.


vyzo:
dunno what's up with the gen-check failure, there is no information there.

magik6k:
> dunno what's up with the gen-check failure, there is no information there.

When you expand the second-to-last thing on circle it shows you the diff. The usual solution is to just run `make gen` which should regenerate whatever it's complaining about (in this case config doc-gen stuff which just landed in https://github.com/filecoin-project/lotus/pull/6848 - so just running `make cfgdoc-gen` should also be enough, and faster)

node/config/types.go
```diff
@@ -282,7 +282,8 @@ type Splitstore struct {
 	HotStoreType  string
 	MarkSetType   string
 
-	HotStoreMessageRetention uint64
+	HotStoreMessageRetention  uint64
+	HotStoreMovingGCFrequency uint64
```
magik6k, line 286: Would be nice to add docstrings to those since we now expose those comments in user configs automagically

vyzo: Ah nice, I didn't know about this. I will add a one-liner for them, as we have detailed documentation in the README.

vyzo: done.

vyzo:
Effectiveness of moving GC:

Before moving GC, after about a week with online GC in every compaction:
```
132G    ./datastore/splitstore/hot.badger
```

An hour after moving GC:
```
57G     ./datastore/splitstore/hot.badger.1627141269
```

vyzo:
rebased on master.

vyzo:
rebased on master and fixed conflicts.

blockstore/badger/blockstore.go
```diff
@@ -80,13 +83,26 @@ const (
 	stateClosed
 )
 
+const (
+	moveStateNone = iota
+	moveStateMoving
+	moveStateCleanup
+	moveStateLock
+)
```
raulk, line 91: 1. let's make this an int type
2. godocs on what each of these states means?

blockstore/badger/blockstore.go
```diff
@@ -165,26 +183,342 @@ func (b *Blockstore) isOpen() bool {
...
+	log.Infof("moving blockstore from %s to %s", b.opts.Dir, path)
+
+	opts := b.opts
+	opts.Dir = path
+	opts.ValueDir = path
+
+	db2, err := badger.Open(opts.Options)
+	if err != nil {
+		return fmt.Errorf("failed to open badger blockstore in %s: %w", path, err)
+	}
```
raulk, line 291: What would happen if we fail to open a new badger store?

blockstore/badger/blockstore.go
```diff
@@ -349,18 +698,36 @@ func (b *Blockstore) Put(block blocks.Block) error {
...
+
+	if err := put(b.db); err != nil {
+		return err
+	}
+
+	if b.db2 != nil {
+		if err := put(b.db2); err != nil {
+			return err
+		}
+	}
```
raulk, line 728: What is this for?

blockstore/badger/blockstore.go
```diff
@@ -165,26 +183,342 @@ func (b *Blockstore) isOpen() bool {
...
+		b.deleteDB(oldpath)
+	}
+
+	log.Info("moving blockstore done")
+	return nil
+}
+
+// doCopy copies a badger blockstore to another, with an optional filter; if the filter
+// is not nil, then only cids that satisfy the filter will be copied.
+func (b *Blockstore) doCopy(from, to *badger.DB, filter func(cid.Cid) bool) error {
```
raulk, line 341: What do we plan to use this filter for?

blockstore/badger/blockstore.go
```diff
@@ -165,26 +183,342 @@ func (b *Blockstore) isOpen() bool {
...
+	putPooled := func() {
+		for _, buf := range pooled {
+			pool.Put(buf)
+		}
+		pooled = pooled[:0]
+	}
+	defer putPooled()
+
+	var buf []byte
+	for iter.Rewind(); iter.Valid(); iter.Next() {
```
raulk, line 368: Did we consider using https://pkg.go.dev/github.com/dgraph-io/badger/v2#readme-stream, which is more efficient and automatically concurrent, instead of an iterator? It also supports selecting keys by predicates (to satisfy the filter).

raulk (commented):
Quite a lot to digest -- tricky code. Still reviewing.

blockstore/badger/blockstore.go
```diff
@@ -165,26 +183,342 @@ func (b *Blockstore) isOpen() bool {
...
+			return ErrBlockstoreClosed
+		}
+
+		item := iter.Item()
+
+		kk := item.Key()
+		if filter != nil {
+			k := kk
+			if b.prefixing {
+				k = k[b.prefixLen:]
```
Stebalien, line 379: I'd just check the length first, just to be safe.

blockstore/badger/blockstore.go
```diff
@@ -165,26 +183,342 @@ func (b *Blockstore) isOpen() bool {
...
+		return batch.Flush()
+	}
+
+	return nil
+}
+
+func (b *Blockstore) deleteDB(path string) {
+	// follow symbolic links, otherwise the data wil be left behind
+	lpath := path
+	for {
```
Stebalien, line 437: EvalSymlinks?

node/config/types.go
```diff
@@ -278,11 +278,23 @@ type Chainstore struct {
...
+	// Only currently supported value is "badger".
+	HotStoreType string
+	// MarkSetType specifies the type of the markset.
+	// It can be "map" (default) for in memory marking or "badger" for on-disk marking.
+	MarkSetType string
+
+	// HotStoreMessageRetention specifies the retention policy for messages, in finalities beyond
+	// the compaction boundary; default is 0.
 	HotStoreMessageRetention uint64
+	// HotStoreMovingGCFrequency specifies how often to perform moving GC on the hotstore.
```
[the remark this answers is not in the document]
vyzo: sure.

blockstore/badger/blockstore.go
vyzo: sure!

blockstore/badger/blockstore.go
vyzo: it would just error out and signal an error in the `CollectGarbage` call -- I don't think we need to do anything more here.

blockstore/badger/blockstore.go
vyzo: so that we don't lose data during the move -- the put (also putMany) has to go to both dbs during the move.

blockstore/badger/blockstore.go
vyzo: The filter is there to support a `CopyTo` interface which would take a filter as we've been discussing with steven.
We can also remove it for now and add it later, when we have a concrete use case for it.

blockstore/badger/blockstore.go
vyzo: My only concern with the streaming interface is that we also have writes happening concurrently with the move.
If that's not a problem, then by all means, let's do streaming, I would be very happy to go faster :)

blockstore/blockstore.go
```diff
@@ -37,9 +37,21 @@ type BlockstoreIterator interface {
...
+type blockstoreMovingGCKey struct{}
+type blockstoreMovingGCPath struct{}
+
+// BlockstoreMovingGC is a garbage collection option that instructs the blockstore
+// to use moving GC if supported.
+var BlockstoreMovingGC = blockstoreMovingGCKey{}
+
+// BlockstoreMovingGCPath is a garbage collection option that specifies an optional
+// target path for moving GC.
+var BlockstoreMovingGCPath = blockstoreMovingGCPath{}
```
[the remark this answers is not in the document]
vyzo: Well the use-case is a symlinked badger in a separate file system, where the user specifies the path.

blockstore/blockstore.go
```diff
@@ -37,9 +37,21 @@ type BlockstoreIterator interface {
 
 // BlockstoreGC is a trait for blockstores that support online garbage collection
 type BlockstoreGC interface {
-	CollectGarbage() error
+	CollectGarbage(options map[interface{}]interface{}) error
```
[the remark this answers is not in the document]
vyzo: The difficulty is that the functional option would be defined in the base type and used in a subtype.
I guess we could make a `GCConfig` object and apply there so that we avoid the parsing.
So yeah, I'll change it to use functional options, it's cleaner.

blockstore/blockstore.go
```diff
@@ -37,9 +37,21 @@ type BlockstoreIterator interface {
...
 }
 
+// garbage collection options
+type blockstoreMovingGCKey struct{}
+type blockstoreMovingGCPath struct{}
+
+// BlockstoreMovingGC is a garbage collection option that instructs the blockstore
+// to use moving GC if supported.
+var BlockstoreMovingGC = blockstoreMovingGCKey{}
```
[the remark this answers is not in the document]
vyzo: Good point, I'll rename to `FullGC` or something similar.

blockstore/splitstore/README.md
```diff
@@ -59,6 +59,17 @@ These are options in the `[Chainstore.Splitstore]` section of the configuration:
   nodes beyond 4 finalities, while running with the discard coldstore option.
   It is also useful for miners who accept deals and need to lookback messages beyond
   the 4 finalities, which would otherwise hit the coldstore.
+- `HotStoreMovingGCFrequency` -- specifies how frequenty to garbage collect the hotstore
```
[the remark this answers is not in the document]
vyzo: yup, will rename to `FullGCFrequency`.

blockstore/badger/blockstore.go
```diff
@@ -35,6 +36,8 @@ var (
 	log = logger.Logger("badgerbs")
 )
 
+const moveBatchSize = 16384
```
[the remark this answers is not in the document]
vyzo: sure.

blockstore/badger/blockstore.go
```diff
@@ -165,26 +183,342 @@ func (b *Blockstore) isOpen() bool {
...
+		log.Warnf("error closing badger db: %s", err)
+	}
+
+	dbpath := b.opts.Dir
+	oldpath := fmt.Sprintf("%s.old.%d", dbpath, time.Now().Unix())
+
+	ok := true
+	err = os.Rename(dbpath, oldpath)
+	if err != nil {
+		// this is bad, but not catastrophic; new data will be written in db2 and user can fix
```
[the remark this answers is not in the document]
vyzo: then we should panic I guess.

blockstore/badger/blockstore.go
```diff
@@ -80,13 +83,26 @@ const (
...
 	viewers sync.WaitGroup
 
-	DB   *badger.DB
+	moveMx    sync.Mutex
+	moveCond  sync.Cond
+	moveState int
+	rlock     int
+
+	db   *badger.DB
+	db2  *badger.DB // when moving
```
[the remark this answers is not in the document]
vyzo: hrm, maybe, just more characters to type :)

blockstore/badger/blockstore.go
```diff
@@ -165,26 +183,342 @@ func (b *Blockstore) isOpen() bool {
...
+		// this is bad, but not catastrophic; new data will be written in db2 and user can fix
+		log.Errorf("error renaming badger db dir from %s to %s; USER ACTION REQUIRED", dbpath, oldpath)
+		ok = false
+	}
+
+	if ok {
+		err = os.Symlink(path, dbpath)
+		if err != nil {
+			// ditto, this is bad, but not catastrophic; user can fix
+			log.Errorf("error symlinking badger db dir from %s to %s; USER ACTION REQUIRED", path, dbpath)
```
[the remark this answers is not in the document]
vyzo: It's not catastrophic in the sense that we haven't lost any data.
It is bad however, since as you point out the db is not in the target path because the symlinking failed.
I guess we should panic to make the user fix the situation instead of leaving him with a ticking bomb.

blockstore/badger/blockstore.go
```diff
@@ -165,26 +183,342 @@ func (b *Blockstore) isOpen() bool {
...
+	defer batch.Cancel()
+
+	txn := from.NewTransaction(false)
+	defer txn.Discard()
+
+	opts := badger.IteratorOptions{PrefetchSize: moveBatchSize}
+	iter := txn.NewIterator(opts)
+	defer iter.Close()
+
+	pooled := make([][]byte, 0, 2*moveBatchSize)
```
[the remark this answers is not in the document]
vyzo: Ah yes, good idea.

blockstore/badger/blockstore.go
vyzo: ok.

blockstore/badger/blockstore.go
vyzo: aha! I didn't know this method existed; I'll use it.

blockstore/badger/blockstore.go
```diff
@@ -165,26 +183,342 @@ func (b *Blockstore) isOpen() bool {
...
+		err = os.Symlink(path, dbpath)
+		if err != nil {
+			// ditto, this is bad, but not catastrophic; user can fix
+			log.Errorf("error symlinking badger db dir from %s to %s; USER ACTION REQUIRED", path, dbpath)
+		}
+
+		b.deleteDB(oldpath)
+	}
+
+	log.Info("moving blockstore done")
```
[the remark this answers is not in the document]
vyzo: at this point we have swapped the blockstores in the internal db variables and using the new one.
I guess we should panic in the two bad conditions to not even get here.

blockstore/badger/blockstore.go
vyzo: Ok, I checked the docs and there is no mention of concurrent writes issues; I would assume that this is fine.
I'll try to convert to use the streaming interface.

blockstore/blockstore.go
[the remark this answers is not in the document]
Stebalien: By _telling_ badger where to move the blockstore on GC? I'd expect that to be a _static_ config option (or maybe the moving GC would follow the symlink on start).

Let's talk at standup, I'm confused.

blockstore/blockstore.go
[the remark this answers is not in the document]
Stebalien: Basically, we have two cases:

1. This doesn't generalize. Why are we pretending it does?
2. It does generalize. So generalize it.

One-off abstractions are just a pain.

blockstore/blockstore.go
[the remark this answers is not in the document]
vyzo: Yeah, I'll drop it for now and if/when we need it we can re-add it, and we can talk on standup.

blockstore/blockstore.go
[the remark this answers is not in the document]
vyzo: yup, I respect your strong feelings on the matter!

blockstore/badger/blockstore.go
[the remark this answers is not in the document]
Stebalien: Yeah... that may be the only real option. Although I'd prefer to handle the error better (at least maybe retry this operation once?).

blockstore/badger/blockstore.go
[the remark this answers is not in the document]
Stebalien: Yeah. I guess, not catastrophic but also not "can continue".

blockstore/badger/blockstore.go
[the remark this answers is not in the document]
Stebalien: Yeah, you're right.

blockstore/badger/blockstore.go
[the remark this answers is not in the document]
Stebalien: Yeah, just slightly more descriptive. Also, https://github.com/joaotavora/eglot or https://github.com/emacs-lsp/lsp-mode.

blockstore/badger/blockstore.go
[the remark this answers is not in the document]
vyzo: I implemented it, and then I did the streaming interface; turns out it's completely unnecessary.

blockstore/badger/blockstore.go
vyzo: Turns out the code is way simpler with the streaming interface; only concern is allocations, as it will make a hole bunch of intermediate garbage with the protobuf objects.

blockstore/splitstore/README.md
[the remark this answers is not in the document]
raulk: IMO "moving" is already the most pessimistic/common-denominator form of GC. I do think it generalises nicely because it's the ultimate form of "GC" that databases that do not support efficient GC will fall back to. It's basically a backstop/catch-all that is applicable to all and any database.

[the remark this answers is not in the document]
vyzo: That's a good point, but I think that steven's "FullGC" nomenclature is right -- the user shouldn't care about the mechanics, as long as it really is a full gc; besides I have already renamed it :)
(but I can rename back if we deem so)

blockstore/badger/blockstore.go
raulk: I'd be surprised if they don't use object pooling. Maybe check their source? Otherwise Lotus is lucky to have a memory watchdog...

vyzo: yeah, good point, We'll see how well it works in practice, but I expect it to be quite a bit faster.

vyzo:
Changes per review comments (quite a few!):
- Updated the interface to use functional options.
- Renamed MovingGC to FullGC in the nomenclature.
- Removed the target path option and made the new path resolve links.
- Dropped the filter from the copy; we are not currently using it anyway.
- Implemented the copy using the streaming interface.
- Made path manipulation failures that require action panic; that seems to be the only sane way to proceed.
- Made the state variables typed.
- Renamed db2 to dbNext in the data structure.
- Used EvalSymlinks where appropriate

vyzo:
ok, this is ready for the next round of review; in the meantime I'll deploy it in my discard node to get some timings and real-world burn.

node/config/types.go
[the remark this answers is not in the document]
vyzo: done.

blockstore/badger/blockstore.go

blockstore/badger/blockstore.go
vyzo: removed it.

blockstore/blockstore.go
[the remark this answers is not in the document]
vyzo: done.

blockstore/blockstore.go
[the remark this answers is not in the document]
vyzo: done, functional options it is!

blockstore/splitstore/README.md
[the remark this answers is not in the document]
vyzo: done.

blockstore/badger/blockstore.go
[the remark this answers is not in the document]
vyzo: done.

blockstore/badger/blockstore.go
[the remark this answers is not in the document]
vyzo: done, it now panics.

blockstore/badger/blockstore.go
vyzo: filter is gone.

blockstore/badger/blockstore.go

blockstore/badger/blockstore.go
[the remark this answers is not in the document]

vyzo:
Timings are looking really good:
```
2021-07-27T11:31:33.039Z        INFO    splitstore      splitstore/splitstore_gc.go:30  garbage collecting hotstore done        {"took": 535.387156281}
```

previously it was taking about 40min; that's 4 times faster.

blockstore/badger/blockstore.go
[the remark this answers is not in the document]
vyzo: it's gone!

blockstore/badger/blockstore.go
vyzo: bingo! It's 4.5 times faster (4.5min -vs- 9min) and didn't notice any deleterious effects from memory usage.

raulk (approved):

jennijuju:
We really want this in, however, it's not a hard blocker for v1.11.1-rc1

blockstore/badger/blockstore.go
```diff
@@ -165,26 +194,239 @@ func (b *Blockstore) isOpen() bool {
...
+	if err != nil {
+		return fmt.Errorf("error resolving symlink %s: %w", basePath, err)
+	}
+
+	if basePath == linkPath {
+		path = basePath
+	} else {
+		name := filepath.Base(basePath)
+		dir := filepath.Dir(linkPath)
+		path = filepath.Join(dir, name)
```
Stebalien, line 303: Why are we doing this instead of just using `linkPath`? Assuming `/a/b/foo -> /x/y/bar`, this will create `/x/y/foo.$time`. Don't we want to create `/x/y/bar.$time`?

blockstore/badger/blockstore.go
```diff
@@ -165,26 +194,239 @@ func (b *Blockstore) isOpen() bool {
...
+	b.moveState = moveStateLock
+	for b.rlock > 0 {
+		b.moveCond.Wait()
+	}
+
+	b.moveState = moveStateMoving
+	b.moveCond.Broadcast()
+	b.moveMx.Unlock()
+
+	var path string
```
Stebalien, line 261: `newPath`? Just calling this path is confusing.

blockstore/badger/blockstore.go
```diff
@@ -165,26 +194,239 @@ func (b *Blockstore) isOpen() bool {
...
+		return fmt.Errorf("error moving badger blockstore to %s: %w", path, err)
+	}
+
+	b.lockMove()
+	db1 := b.db
+	b.db = b.dbNext
+	b.dbNext = nil
+	b.unlockMove(moveStateCleanup)
+
+	err = db1.Close()
```
Stebalien, line 334: dbOld, please! All these 1's and 2's make the code very difficult to follow.

blockstore/badger/blockstore.go
```diff
@@ -165,26 +194,239 @@ func (b *Blockstore) isOpen() bool {
...
+	if err != nil {
+		log.Warnf("error closing old badger db: %s", err)
+	}
+
+	dbpath := b.opts.Dir
```
Stebalien, line 339: _which_ db path is this? Please chose a variable name.

Stebalien: As far as i can tell, this is the _old_ path.

blockstore/badger/blockstore.go
```diff
@@ -165,26 +194,239 @@ func (b *Blockstore) isOpen() bool {
...
+	oldpath := fmt.Sprintf("%s.old.%d", dbpath, time.Now().Unix())
```
Stebalien, line 340: Backup path?

blockstore/badger/blockstore.go
```diff
@@ -165,26 +194,239 @@ func (b *Blockstore) isOpen() bool {
...
+
+	if err = os.Rename(dbpath, oldpath); err != nil {
+		// this is not catastrophic in the sense that we have not lost any data.
+		// but it is pretty bad, as the db path points to the old db, while we are now using to the new
+		// db; we can't continue and leave a ticking bomb for the next restart.
+		// so a panic is appropriate and user can fix.
+		panic(fmt.Errorf("error renaming old badger db dir from %s to %s: %w; USER ACTION REQUIRED", dbpath, oldpath, err)) //nolint
+	}
+
+	if err = os.Symlink(path, dbpath); err != nil {
```
Stebalien, line 350: What if I wasn't using symlinks?

Stebalien (commented):
Mostly naming comments but also some questions around file names.

TL;DR: When I have 2 databases and 4 different paths (old, new, link target, backup), the names really matter.

blockstore/badger/blockstore.go
```diff
@@ -165,26 +194,239 @@ func (b *Blockstore) isOpen() bool {
...
+
+	err = db1.Close()
+	if err != nil {
+		log.Warnf("error closing old badger db: %s", err)
+	}
+
+	dbpath := b.opts.Dir
+	oldpath := fmt.Sprintf("%s.old.%d", dbpath, time.Now().Unix())
+
+	if err = os.Rename(dbpath, oldpath); err != nil {
```
Stebalien, line 342: Won't this just rename the symlink? Don't we need to move the actual file?

We need integration tests for this.

Stebalien: Ah, I see... we are renaming the symlink intentionally. Why not just delete it then delete `linkPath`?

Stebalien:
Other feedback: The renaming logic with the symlink dance is strange. We should be preserving symlinks if the user is using them, and _not_ using symlinks if the user isn't already using them.

1. Not all filesystems support them.
2. Are we sure we're not creating absolute symlinks? That could cause trouble when moving repos.

blockstore/badger/blockstore.go
vyzo: Well, the problem with that is that when we have multiple moves we'll have `foo.$time` already and we'll end up linking in `foo.$time.$time....`.

blockstore/badger/blockstore.go
vyzo: sure, will fix the names to make the logic easier to follow.

blockstore/badger/blockstore.go
vyzo: yup! sorry about that.

blockstore/badger/blockstore.go
vyzo: this is the canonical path where we want the db to live.

blockstore/badger/blockstore.go
vyzo: sure.

blockstore/badger/blockstore.go
vyzo: is that a problem? 
The issue with not using symlinks is moving across filesystems, which can take a long time.

blockstore/badger/blockstore.go
vyzo: yeah, it's fine -- we'll delete deep in deleteDB which follows symlinks.

Also, I don't think we need an integration test for this, we can test pretty much everything with unit tests as the changes are pretty self-contained.

What do you want to add to the unit test?

vyzo:
> Not all filesystems support them.

Wait, is there a realistic situation where lotus would run in a filesystem that doesn't support symlinks?
Even windows support them these days.

> Are we sure we're not creating absolute symlinks? That could cause trouble when moving repos.

Gah, you are right about this. We _are_ creating absolute links, I'll fix in follow up.

blockstore/badger/blockstore.go
vyzo: I'll add a comment here to make the intent clear.

Stebalien: I see. I was just thinking, if the user linked to a directory on a remote drive...

blockstore/badger/blockstore.go
Stebalien: Thank you! (sorry to be a bit anal about this)

blockstore/badger/blockstore.go
Stebalien: Ah, got it. Leaving this as it is sounds reasonable then.

blockstore/badger/blockstore.go
Stebalien: Basically, what if I had the database at "foo". Now I'll end up with a symlink at "foo" pointing to a random directory.

If I started out with plain directories, I'd rather end up with plain directories. It's not critical, but the current version is going to confuse users.

1. They initialize their repo.
2. They see where their hotstore is.
3. The hotstore gets moved and replaced by a symlink.

blockstore/badger/blockstore.go
vyzo: not a problem, really. We will want to read this code again in the future, and I can see how the 1s and 2s could confuse pretty much everyone (including me :)

blockstore/badger/blockstore.go
Stebalien: > yeah, it's fine -- we'll delete deep in deleteDB which follows symlinks.

It's fine, but a _bit_ weird.

> What do you want to add to the unit test?

The unit test is fine. I thought there was a bug here which would have been caught by a basic test. I revisited this because I then _did_ look at the test, and it clearly checked this case.

Your tests are good.

Stebalien:
> Wait, is there a realistic situation where lotus would run in a filesystem that doesn't support symlinks?

Well... probably not. Given that NFS does. Although someone probably does it anyways...

Sorry, I've had some recent encounters with symlinks.

blockstore/badger/blockstore.go
vyzo: Hrm, it complicates things to try and keep it symlink-free, as we'll need two versions of things.

We should probably be clear to users that the splitstore _owns_ the splitstore directory; they shouldn't muck with it.
The coldstore is another matter though.

Stebalien: Fair enough.

vyzo:
Follw up in #6905

blockstore/badger/blockstore.go
vyzo: done in https://github.com/filecoin-project/lotus/pull/6905/commits/950ded454d2a6101090097f4e3bb60c722875497

blockstore/badger/blockstore.go
vyzo: added comment about intent in https://github.com/filecoin-project/lotus/pull/6905/commits/950ded454d2a6101090097f4e3bb60c722875497

blockstore/badger/blockstore.go
vyzo: done in https://github.com/filecoin-project/lotus/pull/6905/commits/950ded454d2a6101090097f4e3bb60c722875497

blockstore/badger/blockstore.go
vyzo: done in https://github.com/filecoin-project/lotus/pull/6905/commits/950ded454d2a6101090097f4e3bb60c722875497

PR #6857, opened 2021-07-23
feat: Don't panic when api impl is nil

### LINKED ISSUE #6632: Dont panic when calling a method that is not supported by lotus gateway
Currently, there are a list of methods thats supported by lotus full node api, but not the lotus gateway apis. So for someone who uses lotus lite and called a methods thats not in lotus gateway, they will get a panic like 
```{"jsonrpc":"2.0","id":0,"error":{"code":0,"message":"fatal error calling 'Filecoin.': panic in rpc method 'Filecoin.: runtime error: invalid memory address or nil pointer dereference"``` .
this should be handled more properly with a clearer error message.

### PR DESCRIPTION
**Pull request by placer14**
(fixes #6632)

```
curl -X POST -i -H "Content-Type: application/json" --data '{ "jsonrpc": "2.0", "id":1, "method": "Filecoin.MpoolGetNonce", "params": [ "t1mps3lifbqregwqnfbcyjlllectulzgchbhsvomy" ] }' 'http://127.0.0.1:2346/rpc/v0'
HTTP/1.1 200 OK
Date: Fri, 23 Jul 2021 23:17:42 GMT
Content-Length: 77
Content-Type: text/plain; charset=utf-8

{"jsonrpc":"2.0","id":1,"error":{"code":1,"message":"method not supported"}}
```


placer14:
This removes the panicking, BUT it's unclear how to manage the HTTP status code and the error code returned in the payload. This also checks in the artifacts from `make gen`. LMK if another process handles this and should be removed from this PR.

magik6k:
Http 200 code is correct even for method-not-found jsonrpc errors. We could use https://github.com/filecoin-project/go-jsonrpc/pull/36/files to return error codes, but I don't think that's necessary

> This also checks in the artifacts from `make gen`

That's good, those are meant to be checked in

magik6k (approved):
Looks like a nice improvement

PR #6859, opened 2021-07-24
feat(repoversion): implement repo versioning

### LINKED ISSUE #543: Repo should contain version file

frrist:
- closes #543

### DIFF
--- a/node/repo/fsrepo.go
+++ b/node/repo/fsrepo.go
@@ -39,6 +39,7 @@ const (
 	fsDatastore     = "datastore"
 	fsLock          = "repo.lock"
 	fsKeystore      = "keystore"
+	fsVersion       = "repo.version"
 )
 
 type RepoType int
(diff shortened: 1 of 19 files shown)


node/repo/fsrepo.go
```diff
@@ -172,6 +242,27 @@ func (fsr *FsRepo) initKeystore() error {
 	return os.Mkdir(kstorePath, 0700)
 }
 
+func (fsr *FsRepo) initVersion(t RepoType) error {
```
magik6k, line 245: Since most repos don't have version, we should be calling this on open too, which should gradually get users the version file in their repos (and eventually after next network upgrade or two we should be able to drop that)

magik6k (commented):
We should be calling `Version` on repo open, and checking that it either returns the current version, or ErrNoVersion (and call initVersion in that case). This way we at least guard from opening newer-versioned repos.

node/repo/fsrepo.go
frrist: The closest thing to a repo `Open()` method is `Exists()`, which is where we perform the check that the version file is present, and seems like the obvious place to create the version file is one is not already present - as you've described. 
However, since the lotus repo and miner repo can be versioned independently (unclear if this is a requirement) this would require a `RepoType` to be passed to the `Exists()` method, which doesn't seem ideal.

Possible Solutions:
- lotus and miner repo always have the same versions: we can init the version in the Exists method as you've described
- A new `Migrate` method is created and called if exists returns `ErrNoVersion`
- Exists accepts a repotype

magik6k: I think it's ok to init the version in Exists for now

ZenGround0:
@magik6k could you take a quick look at the last commit initializing version file during Exists()?

magik6k:
Converting to draft until the review gets addressed

rjan90:
Closing this PR as outdated/stale

BigLep:
2024-09-03 maintainer conversation: this work will get done if/when we first need to do a migration.

PR #6867, opened 2021-07-25
Improve splitstore warmup

vyzo:
While testing the `check` command in #6811 it was observed (to my great surprise) that my discard node actually has some cold references, dating back to the original snapshot.
So it appears that the warmup is not capturing all references that it should.

This improves the situation by traversing a full finality worth of state during warmup.

### DIFF
--- a/blockstore/splitstore/splitstore_warmup.go
+++ b/blockstore/splitstore/splitstore_warmup.go
@@ -9,6 +9,7 @@ import (
 	blocks "github.com/ipfs/go-block-format"
 	cid "github.com/ipfs/go-cid"
 
+	"github.com/filecoin-project/go-state-types/abi"
 	bstore "github.com/filecoin-project/lotus/blockstore"
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain/types"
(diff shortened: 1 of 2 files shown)


blockstore/splitstore/splitstore_warmup.go
```diff
@@ -44,11 +50,12 @@ func (s *SplitStore) warmup(curTs *types.TipSet) error {
 // objects are written in batches so as to minimize overhead.
 func (s *SplitStore) doWarmup(curTs *types.TipSet) error {
 	epoch := curTs.Height()
+	boundaryEpoch := epoch - WarmupBoundary
```
raulk, line 53: This might underflow to negative in integration tests where we use the splitstore. Maybe we should add a Ceil?

raulk (approved):
This is probably safer in the face of reorgs immediately after initialization.

blockstore/splitstore/splitstore_warmup.go
vyzo: ah yes, good point -- I'll add a ceiling check.

vyzo: done.

vyzo:
rebased on master and added undefrlow check.

PR #6868, opened 2021-07-25
Splitstore: Implement cold object reification

### LINKED ISSUE #6726: Splitstore: reification of cold objects
Normally, with #6474, a node should not refer to the coldstore during operation.
In particular, there should be no hot objects with references to cold state.

It is possible however to refer to coldstore objects by operator intervention: e.g. a state-compute or set-head from the API or resolution of a fork.
If the references are more than 4 finalities away they will likely be in the coldstore, and newly created objects will reside in the hotstore but refered to cold objects.

Also, moving forward, we eventually want to make splitstore IPLD-aware so that we can invoke something like `Pin(reference)` and have everything reified to the hotstore.

The issue here is that we don't currently reify state when there is a miss. There is a good reason for this currently (see #6705), but it is something we want to tackle eventully.

vyzo:
On top of #6854 
Closes #6726 

This adds configurable cold object reification in the splitstore. We can do that now that #6705 is resolved.

Rationale: It has been observed that the hot range in my discard store still contains some references to the cold store (containing only the original snapshot data), even after rewarming up with #6867; afacit these are objects that get recreated by the VM and occurs-checked with Has, which avoids a write.

This patch adds a policy for cold object reification, with the following options:
- 0/default no cold object reification.
- 1 reification when there is a view.
- 2 reification when there is a view or a get.

The separation of view from get comes from the observation that state computations access objects with view to unmarshal, while chain exchange uses get to transmit over the network. Thus a setting of 1 should be safe; a setting of 2 becomes relevant only when we have gc for the coldstore, but it's there to future proof.

### DIFF
--- a/blockstore/badger/blockstore.go
+++ b/blockstore/badger/blockstore.go
@@ -96,7 +96,7 @@ type Blockstore struct {
 	viewers sync.WaitGroup
 
 	moveMx    sync.RWMutex
-	moveCond  *sync.Cond
+	moveCond  sync.Cond
 	moveState int
 	rlock     int
 
@@ -135,7 +135,7 @@ func Open(opts Options) (*Blockstore, error) {
 		bs.prefixLen = len(bs.prefix)
 	}
 
-	bs.moveCond = sync.NewCond(&bs.moveMx)
+	bs.moveCond.L = &bs.moveMx
 
 	return bs, nil
 }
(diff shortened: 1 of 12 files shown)


vyzo:
Turning this back to draft, as I am not convinced this is the right approach.

vyzo:
Following discussion with @stebalien it appears that this is the correct course of action:
- Implement a specialized view of the splitstore that recursively reifes all hotstore misses
- Use the view in the vm (and other places that need to reify objects)

This will ensure that we reify objects only when we really need to, and avoid accidental reification.
It will also obviate the need for a policy and the (hacky) distinction of get-vs-view.

In this light I am closing this pr; will open a new one with the desired implementation.

PR #6888, opened 2021-07-27
add a very verbose -vv flag to lotus and lotus-miner.

raulk:
This helps debug issues like the Miner Runtime Architecture segregation issues reported here: https://github.com/filecoin-project/lotus/discussions/6861, where there might be a mix up of environment variables, repos, etc. leading to the wrong API endpoints being used.


jennijuju (dismissed):

jennijuju:
missing `make docsgen-cli`

jennijuju:
ideally id suggest a separate cli, like `lotus version -v` or adding `lotus api-info` instead of adding a global option

raulk:
@jennijuju I've thought this as a general CLI diagnostics flag. Right now we're printing API endpoints, but we can print other things, like file/directory paths, polling intervals, counts, etc. 

For the API endpoint tracing, I don't think putting it behind a command is the right call, because we want to trace what's being used _in the command we are running right now_.

ZenGround0 (approved):
I think of "VeryVerbose" as the standard name for this and it has the benefit of matching "vv" but this is not very important.

raulk:
@ZenGround0 agree. Changed the terminology.

PR #6892, opened 2021-07-27
update go-libp2p-pubsub to v0.5.1

vyzo:
fixes a potential panic related to blacklisting.


vyzo:
ok, this needs some deeper go-libp2p stack udpates; retracting for now.

PR #6895, opened 2021-07-28
Eliminate inefficiency in markets logging

hannahhoward:
# Goals

Whenever we have a data transfer event we currently read in the entire list of received CIDs off disk just to get their length.
We have a function that can return length without reading the CID list

# Implementation

- Use ReceivedCidsLen() rather than len(ReceivedCids)


magik6k (approved):

Kubuxu (approved):

magik6k:
(Will need to be ported / PRad to master as well)

raulk (approved):
This is already fixed in the m1 branches.

jennijuju:
this needs to be in master

PR #6905, opened 2021-07-28
Moving GC Followup

vyzo:
Follow up from #6854 

This improves code readability with better variable names.
It also fixes a (minor) bug, whereby we always create absolute symlinks even though the link might be in the same directory


blockstore/badger/blockstore.go
```diff
@@ -296,68 +297,96 @@ func (b *Blockstore) movingGC() error {
...
 
-	b.deleteDB(oldpath)
+	// the delete follows symlinks
+	b.deleteDB(backupPath)
 
 	log.Info("moving blockstore done")
 	return nil
 }
 
+// symlink creates a symlink from path to linkPath; the link is relative if the two are
```
Kubuxu, line 370:
```suggestion
// symlink creates a symlink from path to linkTo; the link is relative if the two are
```

blockstore/badger/blockstore.go
```diff
@@ -296,68 +297,96 @@ func (b *Blockstore) movingGC() error {
...
+// in the same directory
+func (b *Blockstore) symlink(path, linkTo string) error {
+	resolvedPathDir, err := filepath.EvalSymlinks(filepath.Dir(path))
```
Kubuxu, line 373: shouldn't you first eval symlinks and then take the dir?

vyzo: No, linkTo doesn't exist yet.

vyzo: We could do it for path however, just kept it like this for symmetry.

vyzo: But we know that it is not a symlink itself, as we created, so moot.

blockstore/badger/blockstore.go
```diff
@@ -296,68 +297,96 @@ func (b *Blockstore) movingGC() error {
...
+	// the delete follows symlinks
+	b.deleteDB(backupPath)
 
 	log.Info("moving blockstore done")
 	return nil
 }
 
+// symlink creates a symlink from path to linkTo; the link is relative if the two are
+// in the same directory
+func (b *Blockstore) symlink(path, linkTo string) error {
```
Stebalien, line 372: make this a free function?

Stebalien (approved):

blockstore/badger/blockstore.go
vyzo: sure!

jennijuju:
This should be backported to release/v1.11.1

Issue #6887, opened 2021-07-27
lotus-miner info doesn't work after MRA service separation


nonsense:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

lotus miner - mining and block production

### Lotus Version

$ lotus version
Daemon:  1.11.1-dev+mainnet+git.48e8f54c6+api1.3.0
Local: lotus version 1.11.1-dev+mainnet+git.48e8f54c6

`feat/replace-multistore-carv2` branch

### Describe the Bug

MRA feature allows miners to split their monolith miner node into multiple service - proving/sealing miner process, and markets process.

This creates a problem for `lotus-miner info` command - it is currently trying to fetch data from APIs that are available exclusively on one or the other process, so it will always return an error, for example:

```bash
$ LOTUS_MINER_PATH=~/lotus-miner-path ./lotus-miner info
Chain: [sync ok] [basefee 137.976 pFIL]
Miner: f0127896 (32 GiB sectors)
Power: 25.7 Ti / 8.15 Ei (0.0003%)
        Raw: 22.72 TiB / 8.149 EiB (0.0003%)
        Committed: 22.72 TiB
        Proving: 22.72 TiB
Projected average block win rate: 0.30/week (every 554h56m13s)
[...]

PR #6906, opened 2021-07-28
adding an RuntimeSubsystems API to storage miner; fix `lotus-miner info`

nonsense:
Fixes https://github.com/filecoin-project/lotus/issues/6887

---

This PR is adding `RuntimeSubsystems` API endpoint, so that CLI commands such as `lotus-miner info` can detect what kind of miner node they interact with, and issue the correct calls for information, depending on available subsystems.


api/api_subsystems.go
```diff
@@ -0,0 +1,61 @@
+package api
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+type MinerSubsystem int
```
raulk, line 8: Can we name this plural?

raulk: Why do we want to make this a bitfield instead of just a simple []string?

documentation/en/api-v0-methods-miner.md
```diff
@@ -533,6 +535,18 @@ Inputs:
...
+
+### GetEnabledSubsystems
+
+
+Perms: read
+
+Inputs: `null`
+
+Response: `null`
+
```
raulk, line 549: There's something wrong here

node/builder_miner.go
```diff
@@ -72,6 +72,7 @@ func ConfigStorageMiner(c interface{}) Option {
 	return Options(
 		ConfigCommon(&cfg.Common, enableLibp2pNode),
 
+		Override(new(*api.MinerSubsystem), modules.AddMinerSubsystems(cfg.Subsystems)),
```
raulk, line 75: Let's create a dtype here: dtypes.EnabledSubsystems = api.MinerSubsystems.

node/impl/storminer.go
```diff
@@ -48,6 +48,8 @@ import (
 type StorageMinerAPI struct {
 	fx.In
 
+	Subsystems *api.MinerSubsystem `optional:"true"`
```
raulk, line 51:
```suggestion
	EnabledSubsystems *api.MinerSubsystem `optional:"true"`
```

raulk: Is this truly optional? I would imagine it's not.

node/impl/storminer.go
```diff
@@ -703,4 +705,15 @@ func (sm *StorageMinerAPI) ComputeProof(ctx context.Context, ssi []builtin.Secto
...
+	index := 1
+	for i := 0; i < api.MinerSubsystems; i++ {
+		if sm.Subsystems.Has(api.MinerSubsystem(index)) {
+			res = append(res, api.MinerSubsystem(index))
+		}
+		index *= 2
+	}
+	return
+}
+
```
raulk, line 718: This seems very unnecessary. This should just be a:

```
type MinerSubsystems []string
```

And we should just return a []string.

api/api_storage.go
```diff
@@ -166,6 +166,8 @@ type StorageMiner interface {
 	MarketPendingDeals(ctx context.Context) (PendingDealInfo, error)                                                             //perm:write
 	MarketPublishPendingDeals(ctx context.Context) error                                                                         //perm:admin
 
+	GetEnabledSubsystems(ctx context.Context) ([]MinerSubsystem, error) //perm:read
```
raulk, line 169: This does not follow the naming pattern for JSON-RPC methods. Should begin with a domain name (e.g. market, deals, etc.)

raulk: There are a few exceptions for top-level operations, but I wonder if we want to put this into a `Runtime` category. Choices:

- `RuntimeGetEnabledSubystems`
- `EnabledSubsystems`

I think the second is fine as it's shorter and this is definitely one of those global cases.

node/impl/storminer.go
nonsense: It's not, removed the `optional` tag.

node/impl/storminer.go
nonsense: I think it is nicer to be able to just have this API `typed` when we use it in `lotus-miner info` for example.

api/api_storage.go
nonsense: `RuntimeSubsystems` seems good to me - it is a category + name, and `Enabled` is really implicit in this context, given `runtime`.

api/api_subsystems.go
nonsense: Added `MinerSubsystems []MinerSubsystem`.

cmd/lotus-miner/info.go
```diff
@@ -95,284 +102,289 @@ func infoCmdAct(cctx *cli.Context) error {
...
-	if err != nil {
-		return err
-	}
-
-	tbs := blockstore.NewTieredBstore(blockstore.NewAPIBlockstore(api), blockstore.NewMemory())
-	mas, err := miner.Load(adt.WrapStore(ctx, cbor.NewCborStore(tbs)), mact)
-	if err != nil {
-		return err
-	}
+	if subsystems.Has(api.SectorStorageSubsystem) {
```
nonsense, line 105: @raulk this is why I prefer to have the API typed, and not `[]string` - it is serialised as []string already.

cmd/lotus-miner/info.go
```diff
@@ -95,284 +102,289 @@ func infoCmdAct(cctx *cli.Context) error {
...
+		//  * Sealed sectors (count / bytes)
+		//  * Power
 	}
 
-	var retrComplete dealStat
-	for _, retrieval := range retrievals {
-		if retrieval.Status == retrievalmarket.DealStatusCompleted {
-			retrComplete.count++
-			retrComplete.bytes += retrieval.TotalSent
+	if subsystems.Has(api.MarketsSubsystem) {
```
magik6k, line 296: Can we extract bodies of those ifs to separate functions?

nonsense:
Marking this as ready for review, as it is already in working state, and addresses the underlying issue with miners getting panics on `lotus-miner info`, and I know we want to issue a new RC tag today. Nevertheless we should address all outstanding comments.

api/api_subsystems.go
```diff
@@ -0,0 +1,63 @@
...
+	for _, v := range ms {
+		if v == entry {
+			return true
+		}
+
+	}
+	return false
+}
+
+type MinerSubsystem int
```
raulk, line 20: Missing godocs.

api/api_subsystems.go
```diff
@@ -0,0 +1,63 @@
...
+
+const (
+	MarketsSubsystem MinerSubsystem = iota
+	MiningSubsystem
+	SealingSubsystem
+	SectorStorageSubsystem
+)
```
raulk, line 27: Missing docs. The convention is to use a stable prefix to group these together lexicographically for godocs and discoverability purposes.

api/api_subsystems.go
```diff
@@ -0,0 +1,63 @@
...
+func (ms MinerSubsystem) String() string {
+	return MinerSubsystemToString[ms]
+}
+
+var MinerSubsystemToString = map[MinerSubsystem]string{
+	MarketsSubsystem:       "Markets",
+	MiningSubsystem:        "Mining",
+	SealingSubsystem:       "Sealing",
+	SectorStorageSubsystem: "SectorStorage",
+}
```
raulk, line 38: Doesn't hurt too much in this case, but we should use a static array since the underlying type is an int, thus avoiding a map lookup every time we want to return the string representation.

api/api_subsystems.go
```diff
@@ -0,0 +1,63 @@
...
+	"Sealing":       SealingSubsystem,
+	"SectorStorage": SectorStorageSubsystem,
+}
+
+func (ms MinerSubsystem) MarshalJSON() ([]byte, error) {
+	buffer := bytes.NewBufferString(`"`)
+	buffer.WriteString(MinerSubsystemToString[ms])
+	buffer.WriteString(`"`)
+	return buffer.Bytes(), nil
+}
```
raulk, line 52: Can just delegate to the json package to marshal the string: `json.Marshal(string)`. No need to write the raw output here.

documentation/en/api-v0-methods-miner.md
```diff
@@ -1522,6 +1524,18 @@ Inputs:
...
+### RuntimeSubsystems
+
+
+Perms: read
+
+Inputs: `null`
+
+Response: `null`
+
```
raulk, line 1538: Need to fix this.

node/impl/storminer.go
```diff
@@ -48,6 +48,8 @@ import (
 type StorageMinerAPI struct {
 	fx.In
 
+	Subsystems api.MinerSubsystems
```
raulk, line 51:
```suggestion
	EnabledSubsystems api.MinerSubsystems
```

api/api_subsystems.go
```diff
@@ -0,0 +1,63 @@
+package api
```
raulk, line 1: Should not prefix this file with `api` because it doesn't actually define an API.

api/api_storage.go
```diff
@@ -166,6 +166,8 @@ type StorageMiner interface {
 	MarketPendingDeals(ctx context.Context) (PendingDealInfo, error)                                                             //perm:write
 	MarketPublishPendingDeals(ctx context.Context) error                                                                         //perm:admin
 
+	RuntimeSubsystems(ctx context.Context) (MinerSubsystems, error) //perm:read
```
raulk, line 169: Missing godoc.

raulk (commented):
I'll push a commit with these nits.

raulk (approved):

raulk:
Will merge into our branch and we'll backport into master, to unblock the m1.3 release.

node/modules/storageminer.go
```diff
@@ -1007,3 +1007,19 @@ func mutateCfg(r repo.LockedRepo, mutator func(*config.StorageMiner)) error {
 
 	return multierr.Combine(typeErr, setConfigErr)
 }
+
+func PopulateEnabledMinerSubsystems(cfg config.MinerSubsystemConfig) (res api.MinerSubsystems) {
```
Stebalien, line 1011:
```suggestion
func EnabledMinerSubsystems(cfg config.MinerSubsystemConfig) (res api.MinerSubsystems) {
```

Populate implies "fill in" (i.e., in the config).

node/modules/storageminer.go
```diff
@@ -1007,3 +1007,19 @@ func mutateCfg(r repo.LockedRepo, mutator func(*config.StorageMiner)) error {
...
+	if cfg.EnableSealing {
+		res = append(res, api.SubsystemSealing)
+	}
+	if cfg.EnableSectorStorage {
+		res = append(res, api.SubsystemSectorStorage)
+	}
+	if cfg.EnableMarkets {
+		res = append(res, api.SubsystemMarkets)
+	}
+	return
```
Stebalien, line 1024:
```suggestion
	return res
```

Always better to be explicit.

Stebalien (approved):

PR #6914, opened 2021-07-28
Implement splitstore cold object reification

vyzo:
See #6726 for context.

This implements a specialized view for the splitstore which reifies cold objects on hotstore misses.

~The view is hooked to the VM through the constructor _for now_ but I am not sure that's the right place to do it.~

~One alternative would be to change `Chainstore.StateBlockstore()` to take a hotview parameter, but that would touch too many call sites.~
~But maybe that's the right thing to do, open to discussion about this.~

~Another alternative is to make a `Chainstore.StateBlockstoreHotView()` and call that when we want to use the hotview.~
~We'd have to go through the vm constructor call sites and see whether we want to use it.~

Edit: the hotview is presented through the `StateBlockstore` accessor in `Chainstore`. This ensure that state accesses are reifying. Maybe that's too broad though and we need to introduce and alternate accessor which uses the reifying view, and hunt down callsites to decide what's appropriate.

### DIFF
--- a/blockstore/blockstore.go
+++ b/blockstore/blockstore.go
@@ -60,6 +60,11 @@ type BlockstoreSize interface {
 	Size() (int64, error)
 }
 
+// BlockstoreHotView is a trait for blockstores that support a notion of hot views
+type BlockstoreHotView interface {
+	HotView() Blockstore
+}
+
 // WrapIDStore wraps the underlying blockstore in an "identity" blockstore.
 // The ID store filters out all puts for blocks with CIDs using the "identity"
 // hash function. It also extracts inlined blocks from CIDs using the identity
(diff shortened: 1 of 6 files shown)


vyzo:
rebased on master.

vyzo:
rebased on master.

vyzo:
did a rebase and some commit management; there is no need to change the VM, the view should be provided by `Chainstore.StateBlockstore`.

vyzo:
rebased on master.

vyzo:
rebased on master.

vyzo:
following sync discussion with @magik6k we decided to use a context option for reification and only do it for block validation.

Will rework this after the blockstore upgrade work is done.

vyzo:
Closing to cherry pick and rework on top of #8008.

PR #6917, opened 2021-07-28
Hack to handle old network versions from old apis

ZenGround0:
I don't know if this is a good idea but it should work and will make things work out when new code reads from old api calls.  

@whyrusleeping hit the panic in the wild reading from the old api.

### DIFF
--- a/chain/actors/version.go
+++ b/chain/actors/version.go
@@ -22,6 +22,10 @@ const (
 
 // Converts a network version into an actors adt version.
 func VersionForNetwork(version network.Version) Version {
+	// Handle previous api version (1 - 13)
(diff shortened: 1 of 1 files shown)


Kubuxu:
If we want it, then it should be done on the API implementation side instead of internally.

magik6k (commented):
Really needs a unit test asserting that newer versions are going to work with this hack (in lotus or state-types)

ZenGround0:
@whyrusleeping resolution from lotus standup is that this is working as intended because the api version was bumped.  The recommendation to avoid the panic is to check api versions when communicating between lotus and lotus lite.

PR #4954, opened 2020-11-21
print runtime send from

zgfzgf:
fix #4953

PR #6918, opened 2021-07-28
Log more call context during errors

ZenGround0:
This carries forward some nice work from @zgfzgf in #4954 

Logging call context from Abortf as in the original PR turns out to be redundant with the logging from shimCall because all Abortf calls panic up to shimCall immediately.  Logging shimCall failures is not redundant with internal and vm send actor error logging in the edge case where an internal send errors but the error is dropped which can occur in the case of programmer error in calls from reward and cron.


chain/vm/runtime.go
```diff
@@ -391,7 +391,7 @@ func (rt *Runtime) Send(to address.Address, method abi.MethodNum, m cbor.Marshal
 		if err.IsFatal() {
 			panic(err)
 		}
-		log.Warnf("vmctx send failed: to: %s, method: %d: ret: %d, err: %s", to, method, ret, err)
+		log.Warnf("vmctx send failed: from: %s to: %s, method: %d: ret: %d, err: %s", rt.Receiver(), to, method, ret, err)
```
arajasek, line 394: What does this `ret` look like when printed? Is that coherent?

ZenGround0: Since this is currently on latest code we should check the mainnet logs and find out :) 

I think %x will probably be better so we can get hex of cbor bytes and we could make that change here

ZenGround0: Looking more closely `internalSend` should only return a non empty ret value in the case send returns a non empty ret value. This will only return non empty ret if `shimCall` returns non empty ret which can only happen if shimCall does not return an error (either from marshalling ret or an abort from within the actor method being shimmed).  This means printing ret should be strictly redundant since it is always empty.

There are only a few types of errors I see on mainnet but they all print `[]` for ret so that supports this line of reasoning.

Stebalien (approved):

PR #6936, opened 2021-07-29
support MARKETS_API_INFO env var, and markets-repo, markets-api-url CLI flags.

### LINKED ISSUE #6935: Dedicated env var for lotus markets
### Checklist

### Lotus component

Other

### Improvement Suggestion

Currently with a split markets process and to run a cmd, a process path must be identified like `LOTUS_MINER_PATH=~/markets-repo-location ./lotus-miner`

which means, miners need to provide this path specifically for each operation they are about to run.

ideally, we have dedicated evn var, for example, `LOTUS_MINER_X` and `LOTUS_MARKETS_X` for each process and will be read automatically according to the operation.

raulk:
Closes #6935.

Adds support for the `MARKETS_API_INFO` env variable, `markets-api-url` flag, and `markets-repo` flag.

The markets API endpoint/client is now exposed through `GetMarketsAPI()`, which works with both monolith and split deployments by handling multiple environment variables in order of precedence.

Command actions have been refactored to use the appropriate API (`GetStorageMinerAPI`, `GetMarketsAPI`).

In the near future, we want to create a `lotus-markets` binary entrypoint. Until then, markets and storage miner client commands have a single entrypoint `lotus-miner`, which creates ambiguity when calling common commands (e.g. log, auth, pprof, ~net~).

To solve for that, I've added a flag to select the markets node:

```
lotus-miner --call-on-markets log list
```

This is marked as temporary, because we ultimately don't want it here, but we also don't have the bandwidth to create the `lotus-markets` entrypoint just now.

---

This PR does not affect the miner server-side, only the `lotus-miner` client CLI UX.


cli/auth.go
```diff
@@ -128,7 +128,8 @@ var AuthApiInfoToken = &cli.Command{
 
 		// TODO: Log in audit log when it is implemented
 
-		fmt.Printf("%s=%s:%s\n", cliutil.EnvForRepo(t), string(token), ainfo.Addr)
+		currentEnv, _ := cliutil.EnvsForRepo(t)
+		fmt.Printf("%s=%s:%s\n", currentEnv, string(token), ainfo.Addr)
```
nonsense, line 132: We have a bug here, the resulting string looks like:
```
[MINER_API_INFO]=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.X0ujyPv7C1O-Xy88jpFnaPQDUeKdcYHtVvPqlFuzB_0:/ip4/127.0.0.1/tcp/2345/http
```

and it should look like
```
MINER_API_INFO=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.X0ujyPv7C1O-Xy88jpFnaPQDUeKdcYHtVvPqlFuzB_0:/ip4/127.0.0.1/tcp/2345/http
```

nonsense: So probably just picking the first item from that slice is good.

raulk: Fixed by changing the signature of EnvForAPIInfos to return a single primary var.

Kubuxu (approved):

Issue #6921, opened 2021-07-28
lotus-miner storage-deals get-ask/set-ask issues


Meatball13:
### Checklist

- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] I **am** reporting a bug w.r.t one of the [M1 tags](https://github.com/filecoin-project/lotus/discussions/6852#discussioncomment-1043951). If not, choose another issue option [here](https://github.com/filecoin-project/lotus/issues/new/choose).
- [X] I am **not** reporting a bug around deal making. If yes, create a [M1 Bug Report For Deal Making](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fbug%2CM1-release&template=m1_bug_report_deal.yml).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.

### Lotus component

Other

### Lotus Tag and  Version

```text
Lotus: lotus version 1.11.1-dev+mainnet+git.1757ffcb2
Miner: lotus-miner version 1.11.1-dev+mainnet+git.1757ffcb2
```

### Describe the Bug

Attempting to do a lotus-miner storage-deals get-ask and lotus-miner storage-deals set-ask command after setting up split mining.  The 'get-ask' command appears to work successfully if you point it at the lotus-miner with the markets-repo env variable.  'set-ask' does not appear to work regardless if you point it at the lotus-miner with the markets repo or the .lotusminer env variable.  Here's the output from the  attempts:

```
tbtminer@hdfil-dm:~$ LOTUS_MINER_PATH=~/.lotusminer lotus-miner storage-deals get-ask
ERROR: fatal error calling 'Filecoin.MarketGetAsk': panic in rpc method 'Filecoin.MarketGetAsk': runtime error: invalid memory address or nil pointer dereference

tbtminer@hdfil-dm:~$ LOTUS_MINER_PATH=~/markets-repo-location lotus-miner storage-deals get-ask
Price per GiB/Epoch  Verified  Min. Piece Size (padded)  Max. Piece Size (padded)  Expiry (Epoch)  Expiry (Appx. Rem. Time)  Seq. No.
0.00005 FIL          0 FIL     4 GiB                     32 GiB                    1047164         612h56m30s                122

[...]

PR #6939, opened 2021-07-29
fix `lotus-miner storage-deals set-ask` - make sure that `ActorAddress` works on all miner nodes

nonsense:
This PR is fixing https://github.com/filecoin-project/lotus/issues/6921

### DIFF
--- a/node/impl/storminer.go
+++ b/node/impl/storminer.go
@@ -15,6 +15,7 @@ import (
 	"github.com/filecoin-project/lotus/build"
 	"github.com/google/uuid"
 	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"go.uber.org/fx"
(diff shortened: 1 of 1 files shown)


magik6k (commented):
I'd expect `ActorAddress` to work on the market node, with that we can call the full node for sector size -> `StateMinerInfo().SectorSize`

nonsense:
@raulk @magik6k given that we discussed that this is an immutable value, I think this is good solution. You okay with it, or you have something else in mind?

raulk (commented):
Honestly I'm not comfortable doing this. The reason why this works is because it's a side-effect from having copied the miner's metadata store into the markets process. I'm not even sure that was the right thing to do in the first place, and if we merge this, we will be marrying to that approach forever.

raulk (commented):
This is already solved by https://github.com/filecoin-project/lotus/pull/6936, as CLI commands are now intelligent to route to the miner node or the markets node. I don't see the need to merge a miner-side change.

jacobheun:
Closing as this is no longer needed with #6936.

PR #5487, opened 2021-01-30
Update sealing.go

shihuajin111:
Consider the case where the int(barCols)-cpuBars is negative,solve  negative Repeat count bug 

panic: strings: negative Repeat count
goroutine 1 [running]:
strings.Repeat(0x2e33731, 0x1, 0xffffffffffffffd8, 0x0, 0x0)
	/usr/local/go/src/strings/strings.go:529 +0x5e5
main.glob..func46(0xc0006628c0, 0x0, 0x0)
	/mnt/filecoin-lvm/labs/labs/lotus/cmd/lotus-storage-miner/sealing.go:97 +0x9ea
github.com/urfave/cli/v2.(*Command).Run(0x58c65e0, 0xc0006627c0, 0x0, 0x0)
	/mnt/filecoin-lvm/labs/labs/go/pkg/mod/github.com/urfave/cli/v2@v2.2.0/command.go:164 +0x4ed
github.com/urfave/cli/v2.(*App).RunAsSubcommand(0xc000524480, 0xc0006625c0, 0x0, 0x0)
	/mnt/filecoin-lvm/labs/labs/go/pkg/mod/github.com/urfave/cli/v2@v2.2.0/app.go:427 +0xa9f

PR #6942, opened 2021-07-29
Improve formatting of workers CLI

arajasek:
Subsumes #5487

### DIFF
--- a/cmd/lotus-miner/sealing.go
+++ b/cmd/lotus-miner/sealing.go
@@ -91,7 +91,10 @@ var sealingWorkersCmd = &cli.Command{
 
 			var barCols = uint64(64)
 			cpuBars := int(stat.CpuUse * barCols / stat.Info.Resources.CPUs)
-			cpuBar := strings.Repeat("|", cpuBars) + strings.Repeat(" ", int(barCols)-cpuBars)
+			cpuBar := strings.Repeat("|", cpuBars)
+			if int(barCols)-cpuBars >= 0 {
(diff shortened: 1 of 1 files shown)


cmd/lotus-miner/sealing.go
```diff
@@ -91,7 +91,10 @@ var sealingWorkersCmd = &cli.Command{
 
 			var barCols = uint64(64)
 			cpuBars := int(stat.CpuUse * barCols / stat.Info.Resources.CPUs)
-			cpuBar := strings.Repeat("|", cpuBars) + strings.Repeat(" ", int(barCols)-cpuBars)
+			cpuBar := strings.Repeat("|", cpuBars)
+			if int(barCols)-cpuBars >= 0 {
+				cpuBar = cpuBar + strings.Repeat(" ", int(barCols)-cpuBars)
```
Stebalien, line 96: nit `cpuBars += ...`

Stebalien (approved):

cmd/lotus-miner/sealing.go
arajasek: we don't know that in many places in lotus so i assumed it was bad go-ing

arajasek: but happy to agree

arajasek:
```suggestion
				cpuBar += strings.Repeat(" ", int(barCols)-cpuBars)
```

PR #6944, opened 2021-07-29
Lotus release process  - how we make releases

BlocksOnAChain:
Lotus release philosophy explained

### DIFF
--- a/documentation/Lotus release flow
+++ b/documentation/Lotus release flow
@@ -1,0 +1,138 @@
+
+# `lotus` Release Flow
+
+## Table of Contents
+- [Purpose](#Purpose)
+- [High-level summary](#High-level summary)
+- [Motivation and Requirements](#Motivation and Requirements)
+- [Adopted Conventions](#Adopted Conventions)
+  - [Major releases](#Major releases)
+  - [Mandatory releases](#Mandatory releases)
+  - [Feature releases](#Feature releases)
+  - [Examples Scenarios(#Examples Scenarios)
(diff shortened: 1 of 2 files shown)


documentation/Lotus release flow
```diff
@@ -0,0 +1,138 @@
...
+
+## Table of Contents
+- [Purpose](#Purpose)
+- [High-level summary](#High-level summary)
+- [Motivation and Requirements](#Motivation and Requirements)
+- [Adopted Conventions](#Adopted Conventions)
+  - [Major releases](#Major releases)
+  - [Mandatory releases](#Mandatory releases)
+  - [Feature releases](#Feature releases)
+  - [Examples Scenarios(#Examples Scenarios)
```
BlocksOnAChain, line 12: - [Examples Scenarios](#Examples Scenarios)

documentation/Lotus release flow
```diff
@@ -0,0 +1,138 @@
+
+# `lotus` Release Flow
+
+## Table of Contents
```
BigLep, line 4: Should we use a tool to generate this to ensure there are no mistakes: https://stackoverflow.com/questions/18244417/how-do-i-create-some-kind-of-table-of-content-in-github-wiki ?

Also, TOC will be automatically rendered in menu by GitHub: https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/

BigLep (commented):
General: is there a standard line length we should enforce so the diff is easier to work with?

PR #6948, opened 2021-07-30
Dagstore throttle fetch and copy

aarshkshah1992:


markets/dagstore/lotusaccessor.go
```diff
@@ -58,6 +59,55 @@ func (m *lotusAccessor) Start(_ context.Context) error {
...
+
+		if isUnsealed {
+			return true, nil
+		}
+	}
+
+	// we don't have an unsealed sector containing the piece
+	return false, nil
+}
+
```
raulk, line 110: Can't we deduplicate this logic with the similar logic in FetchUnsealedPiece? We could call this `LoadBestDeal` and have it return a `(bool, deal, error)`. `FetchUnsealedPiece` would then take the deal as a parameter.

The mount should call `LoadBestDeal` once, and memoize the result.

raulk: Postponed this consolidation in favour of shipping m1.3.3.

markets/dagstore/lotusaccessor.go
```diff
@@ -58,6 +59,54 @@ func (m *lotusAccessor) Start(_ context.Context) error {
 	return m.readyMgr.FireReady(nil)
 }
 
+func (m *lotusAccessor) IsUnsealed(ctx context.Context, pieceCid cid.Cid) (bool, error) {
```
raulk, line 62: The code duplication will be consolidated later, while we clean up the throttling primitives and we potentially collapse the `lotusAccessor` into the `LotusMount`.

PR #6949, opened 2021-07-30
Reduce splitstore memory usage during chain walks

vyzo:
Walking the chain was using a `cid.Set`, which can grow quite big (30M+ objects, maybe more) leading to signficant _transient_ memory usage.

This adds an ObjectVisitor interface to use instead of a blanket `cid.Set`. This allows us to use a noop set during compaciton (as we already have the markset to lean on, which will be on-disk with `MarkSetType = "badger"`) and markset-backed ones during warmup/check.

We also reset the walked block set on every iteration, to avoid building up a big one with all the blocks; we can do that because the walk is BFS.


vyzo:
Opening as draft until I have tested timing/memory effects in my nodes.

vyzo:
warmup is 2x slower, takes about an hour now.

vyzo:
marking is somewhat slower also; but on the bright side, there was no watchdog action whatsoever (which was a usual occurence).

vyzo:
Marked as ready for review, as node burn was successful.

vyzo:
Unified the markset/visitor dichotomy, which opens up interesting possibilities: we can now check+mark atomically using `Visit`.

Follow up: parallelize `walkChain`; yes we can now!

vyzo:
rebasd on master.

blockstore/splitstore/markset_badger.go
```diff
@@ -46,39 +48,12 @@ func NewBadgerMarkSetEnv(path string) (MarkSetEnv, error) {
 	return &BadgerMarkSetEnv{path: msPath}, nil
 }
 
-func (e *BadgerMarkSetEnv) Create(name string, sizeHint int64) (MarkSet, error) {
+func (e *BadgerMarkSetEnv) create(name string, sizeHint int64) (*BadgerMarkSet, error) {
 	path := filepath.Join(e.path, name)
 
-	// clean up first
-	err := os.RemoveAll(path)
```
Stebalien, line 53: Not related to this patch, but this seems scary...

Stebalien: Maybe fine, but I'd consider giving all "transient" datastores a common "tmp-" prefix.

blockstore/splitstore/markset_badger.go
```diff
@@ -188,6 +136,140 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
+		s.mx.RUnlock()
+		return visit, err
+	}
+
+	err = s.db.View(func(txn *badger.Txn) error {
+		_, err := txn.Get(key)
+		return err
+	})
+
+	switch err {
```
Stebalien, line 174: We could _always_ copy the version, then unlock, then do the rest of this function to simplify the locking a bit.

blockstore/splitstore/markset_badger.go
```diff
@@ -188,6 +136,140 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
 	}
 }
 
+func (s *BadgerMarkSet) Visit(c cid.Cid) (bool, error) {
```
Stebalien, line 139: Maybe break this into multiple functions so we can defer unlocks?

blockstore/splitstore/markset_badger.go
```diff
@@ -188,6 +136,140 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
+		return err
+	})
+
+	switch err {
+	case nil:
+		s.mx.RUnlock()
+		return false, nil
+
+	case badger.ErrKeyNotFound:
+		// need to upgrade the lock to exclusive in order to write; take the version count to see
```
Stebalien, line 180: I'd move this case out of the switch statement and use an early return.

blockstore/splitstore/markset_badger.go
```diff
@@ -188,6 +136,140 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
+		// if there was another write while we were upgrading
+		version := s.version
+		s.mx.RUnlock()
+
+		s.mx.Lock()
```
Stebalien, line 185: If we move this out of the switch statement, we can defer the unlock.

blockstore/splitstore/markset_badger.go
```diff
@@ -188,6 +136,140 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
 	}
 }
 
+func (s *BadgerMarkSet) Visit(c cid.Cid) (bool, error) {
+	key := c.Hash()
+	pendKey := string(key)
+
+	checkPending := func() (bool, error) {
```
Stebalien, line 143: the behavior here is flipped. I'd expect `checkPending` to return _true_ if a write is pending.

blockstore/splitstore/markset_badger.go
```diff
@@ -188,6 +136,140 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
+	visit, err := checkPending()
+	if !visit {
+		s.mx.RUnlock()
+		return visit, err
+	}
+
+	err = s.db.View(func(txn *badger.Txn) error {
+		_, err := txn.Get(key)
+		return err
+	})
```
Stebalien, line 172: Introduce a `BadgerMarkSet.has` helper?

blockstore/splitstore/markset_badger.go
```diff
@@ -188,6 +136,140 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
+}
+
+// writer holds the lock
+func (s *BadgerMarkSet) putPending() error {
+	pend := s.pend
+	seqno := s.seqno
+	s.seqno++
+	s.writing[seqno] = pend
+	s.pend = make(map[string]struct{})
+	s.writers++
```
Stebalien, line 239: Isn't this redundant with `len(s.writing)`?

blockstore/splitstore/markset_badger.go
```diff
@@ -188,6 +136,140 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
+		return false, xerrors.Errorf("error checking badger markset: %w", err)
+	}
+}
+
+// writer holds the lock
+func (s *BadgerMarkSet) putPending() error {
+	pend := s.pend
+	seqno := s.seqno
+	s.seqno++
+	s.writing[seqno] = pend
```
Stebalien, line 237: So, with a background worker, we can get rid of this and just have a single "writing" and "pending" set. When the background worker is done flushing the last set, it would check to see if it should flush the next one (e.g., use a channel as a flag).

That _should_ help simplify things a bit.

blockstore/splitstore/splitstore_compact.go
```diff
@@ -1079,27 +1132,43 @@ func (s *SplitStore) waitForMissingRefs(markSet MarkSet) {
...
 					}
 
-					mark, err := markSet.Has(c)
-					if err != nil {
-						return xerrors.Errorf("error checking markset for %s: %w", c, err)
-					}
-
-					if mark {
-						return errStopWalk
+					visitor, ok := markSet.(MarkSetVisitor)
```
Stebalien, line 1145: Can we somehow avoid this cast?

Stebalien (commented):
The general logic looks good, but I feel like we can simplify a bit.

blockstore/splitstore/markset.go
```diff
@@ -21,8 +21,15 @@ type MarkSet interface {
...
 
+type MarkSetVisitor interface {
+	ObjectVisitor
+	Close() error
+	SetConcurrent()
+}
+
 type MarkSetEnv interface {
 	Create(name string, sizeHint int64) (MarkSet, error)
+	CreateVisitor(name string, sizeHint int64) (MarkSetVisitor, error)
```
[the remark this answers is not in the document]
vyzo: Well, the name is what to call this markset if applicable (for debuging purposes, and also restart cleanup purposes).
The sizeHint is a hint on how many elements to expect; the map markset can pre-allocate the map for example.

I guess I should add some docstring.

blockstore/splitstore/markset_badger.go
vyzo: Probably better to just put it in the markSetEnv so that the entire directory is tmp labelled.

blockstore/splitstore/markset_badger.go
vyzo: yeah, the name is perhaps unfortunate; the return signature is (visit, err) -- perhaps I should just name the return variables to make it clear what it returns.

blockstore/splitstore/markset_badger.go
vyzo: maybe, let me think a bit about that.

blockstore/splitstore/markset_badger.go
vyzo: yes it is.

blockstore/splitstore/markset_badger.go
vyzo: The disadvantage of the background worker is that we can do only one write at the time -- this can do multiple concurrent writes, which is strictly better i think.

blockstore/splitstore/splitstore_compact.go
vyzo: Yes we can -- I can widen the `MarkSetVisitor` type to be `ObjectVisitor` + `MarkSet` and then we don't need to cast or bother with the else branch code.

blockstore/splitstore/markset_badger.go
vyzo: I want to be able to do the write without holding the lock.

blockstore/splitstore/markset_badger.go
vyzo: let me think if it can be somehow be rewritten while maintaining the nice optimistic concurrency properties.

blockstore/splitstore/markset_badger.go
vyzo: Also, I extract the open/close to precisly have the `Transient` in the function name to make it clear that this really is temporary data.

blockstore/splitstore/markset_badger.go
vyzo: we have to hold the lock while doing the fast checks on the maps.

blockstore/splitstore/markset_badger.go
vyzo: I guess I could rename it to `visitPending` so that it is consistent with its return signature.

blockstore/splitstore/markset.go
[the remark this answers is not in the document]
vyzo: done.

blockstore/splitstore/splitstore_compact.go
vyzo: done.

blockstore/splitstore/markset_badger.go
vyzo: It also loses the timely error return property.

I guess we could have a worker that spawns goroutines to do writes, and somehow bubble up the error to be checked in the next operation, but this gets more complicated than I'd like it to be.

blockstore/splitstore/markset_badger.go
vyzo: I made `tryPending` and `tryDB` helpers.

blockstore/splitstore/markset_badger.go
vyzo: made it a method called `tryPending` and reversed the signature to return `(has, error)`.

blockstore/splitstore/markset_badger.go
vyzo: removed.

blockstore/splitstore/markset_badger.go
vyzo: hid the switch inside the `tryDB` helper method.

blockstore/splitstore/markset_badger.go
vyzo: I broke it into `tryPending` and `tryDB` methods, reused by `Has` as well.
Still not deferring the unlocks, but it is much simpler now.

vyzo: It doesn't look like it needs to be broken out into multiple functions now imo, as all the logic is in the helper functions and what this method does is lock management and gluing.

blockstore/splitstore/markset_badger.go
vyzo: I added the `".tmp"` suffix to all markset names.

vyzo:
Did a number of simplifications:
- The `MarkSetVisitor` interface was widened to be the sum of `ObjectVisitor` and `MakrSet`, thus obviating the need for a cast in splitstore code.
- Removed redundant `writers` state variable from badger markset.
- Simplified and deduplicated the code for `Has` and `Visit` by introducing `tryPending` and `tryDB` helpers.
- Added `".tmp"` suffix to all transient dbs in the markset directory.

vyzo:
pushed forgotten commit.

blockstore/splitstore/markset_badger.go
Stebalien: I _think_ badger will handle write parallelism under the covers.

vyzo: hrm not sure -- I prefer to be explicitly parallel and also keep the ability to return an error at some writer so that we can abort.

On the same time, I don't feel terribly strongly about this and we could use a background worker who spawns writers and bubbles up the error in some way.

Stebalien:
I don't feel too strongly either. I'll take a look at the new code.

blockstore/splitstore/markset_badger.go
```diff
@@ -184,8 +186,53 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
-		return false, xerrors.Errorf("error checking badger markset: %w", err)
+		return false, err
+	}
+}
+
+// writer holds the exclusive lock; put releases it
+func (s *BadgerMarkSet) put(key string) error {
+	s.pend[key] = struct{}{}
+	if len(s.pend) < badgerMarkSetBatchSize {
+		s.mx.Unlock()
```
Stebalien, line 197: :bomb:

blockstore/splitstore/markset_badger.go
```diff
@@ -184,8 +186,53 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
+	s.writing[seqno] = pend
+	s.pend = make(map[string]struct{})
+	s.mx.Unlock()
+
+	defer func() {
+		s.mx.Lock()
+		defer s.mx.Unlock()
+
+		s.version++
+		delete(s.writing, seqno)
```
Stebalien, line 213: Do we really want to do this on failure?

blockstore/splitstore/markset_badger.go
Stebalien: Folding this function back into the caller is an acceptable solution. Although I'd consider breaking it up into the part that needs the lock, and the part that doesn't.

blockstore/splitstore/markset_badger.go
```diff
@@ -184,8 +186,53 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
+		if len(s.writing) == 0 {
+			s.cond.Broadcast()
```
Stebalien, line 215: waitgroup?

Stebalien: This seems like a case where we're never going to call back into this after the user tries to close, as far as I can tell at least.

blockstore/splitstore/markset_badger.go
```diff
@@ -104,74 +87,93 @@ func (s *BadgerMarkSet) Mark(c cid.Cid) error {
...
+		return false, err
+	}
 
-	for k := range pend {
-		if err := batch.Set([]byte(k), empty); err != nil {
-			return err
+	// we need to upgrade the lock to exclusive in order to write; take the version count to see
+	// if there was another write while we were upgrading
+	version := s.version
+	s.mx.RUnlock()
```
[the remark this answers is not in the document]
Stebalien: (well, once we avoid dropping the lock in put)

blockstore/splitstore/markset_badger.go
Stebalien: Not critical, but I prefer to avoid conditions unless I need them as most people don't know about them (let alone understand how to use them correctly).

Stebalien (changes requested):
So, _most_ of this patch looks fine. But the locking is still a bit tricker than I'd like. I can live with everything _except_ taking the lock in one function and dropping it in another. That's really the only blocking change.

blockstore/splitstore/markset.go
```diff
@@ -21,8 +21,19 @@ type MarkSet interface {
...
+	MarkSet
+	ObjectVisitor
+}
+
 type MarkSetEnv interface {
+	// Create creates a new markset within the environment.
+	// name is a unique name for this markset, mapped to the filesystem in disk-backed environments
+	// sizeHint is a hint about the expected size of the markset
 	Create(name string, sizeHint int64) (MarkSet, error)
+	// CreateVisitor is like Create, but returns a wider interface that supports atomic visits.
```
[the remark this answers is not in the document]
vyzo: well, we need this if we want to keep bloom marksets around, as it's pointless to use `Visit` on them.

blockstore/splitstore/markset_badger.go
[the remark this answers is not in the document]
vyzo: the problem with the deferred lock here is that it covers the scope of the actual write in the db, so it is _not_ correct; this is a potentially slow operation that can block other readers/writers.
But ok, I will make it release the lock within the function and use another one for doing the actual write.

blockstore/splitstore/markset_badger.go
vyzo: we have to break it and scope the lock externally; the whole reason for releasing the lock in here is to avoid doing the actual write while holding the lock.

blockstore/splitstore/markset_badger.go
vyzo: should we let them accumulate? It probably doesn't matter at all however, since a failure will quickly bubble up and stop the logic that is using the markset.

blockstore/splitstore/markset_badger.go
vyzo: maybe, but this is cleaner imho -- readers should be educated about condition variables, it's an invaluable tool in concurrency.

blockstore/splitstore/markset_badger.go
Stebalien: Not holding the lock while writing is good. Releasing a lock across a function boundary is not good.

blockstore/splitstore/markset_badger.go
Stebalien: It's more correct that way, right?

blockstore/splitstore/markset_badger.go
Stebalien: Eh, doesn't really matter and it does mean we only track the count in one place.

blockstore/splitstore/markset_badger.go
Stebalien: I mean, we _should_ just end up failing out and not making _any_ changes. But I'd feel _slightly_ safer this way (and it can't hurt).

blockstore/splitstore/markset_badger.go
vyzo: understood, allergic reaction :)
I am changing it to release the lock locally and use a separate write function.

blockstore/splitstore/markset_badger.go
vyzo: it also means that we can't close and we need a separate count for writers (as it was before).
But fine, I will check the error and not remove it if we fail, and also add a writers count to disambiguate for closing.

blockstore/splitstore/markset.go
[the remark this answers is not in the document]
Stebalien: So, what happens if I specify that I want to use a bloom filter? Does lotus start spamming my log with random errors?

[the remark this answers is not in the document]
Stebalien: Basically, we need to better distinguish between "something went wrong" and "this isn't supported, move along". That can either be done with multiple interfaces/types, sentinel errors, or a boolean "supported".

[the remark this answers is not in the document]
vyzo: yeah, it will fail the compaction altogether. Perhaps we should kill the bloom code.

blockstore/splitstore/markset_badger.go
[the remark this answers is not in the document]
Stebalien: Well, the point here was less about the deferred unlock and more about reducing the distance between the lock and the unlock calls. I _completely_ missed the fact that we weren't calling unlock after put the first time through.

blockstore/splitstore/markset_badger.go
vyzo: done.

blockstore/splitstore/markset_badger.go
vyzo: done.

blockstore/splitstore/markset.go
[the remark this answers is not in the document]
vyzo: Arguably, the best solution here is a `SupportsVisitor` method in the `MarkSetEnv` so that we can fail early (in constructor) if the user specifies the bloom filter for markset.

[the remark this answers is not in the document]
Stebalien: I'd kill it if we're not using it. We can always revive it from git history.

blockstore/splitstore/markset_badger.go
```diff
@@ -184,10 +195,70 @@ func (s *BadgerMarkSet) Has(c cid.Cid) (bool, error) {
...
+func (s *BadgerMarkSet) put(key string) (write bool, seqno int) {
+	s.pend[key] = struct{}{}
+	if len(s.pend) < badgerMarkSetBatchSize {
+		return false, 0
+	}
+
+	seqno = s.seqno
+	s.seqno++
+	s.writing[seqno] = s.pend
+	s.pend = make(map[string]struct{})
```
Stebalien, line 214: We don't _actually_ have to do this here. Instead, we can return "full" (or even just return the number of pending items). Then, we can call `flush` with no parameters. We can then double check to see if the pending set is over the limit and, if so, flush. Otherwise, someone else wins and we walk away.

blockstore/splitstore/markset.go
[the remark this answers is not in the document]
vyzo: done and done -- have cake, eat it too.

So I added a `SupportVisitor` interface for future extensibility, and kill the unused bloom filter code.

blockstore/splitstore/markset_badger.go
```diff
@@ -92,86 +68,123 @@ func (e *BadgerMarkSetEnv) Create(name string, sizeHint int64) (MarkSet, error)
...
+		s.mx.Unlock()
+		return false, err
+	}
+
+	if version != s.version {
+		// something was written to the db, we need to check it
+		has, err = s.tryDB(key)
+		if has || err != nil {
+			s.mx.Unlock()
+			return false, err
```
Stebalien, line 153: I'd still consider that code I posted as it should simplify this function a bit (even if you can't defer). But that's not critical.

Stebalien (approved):
Two non-blocking nits.

blockstore/splitstore/markset_badger.go
vyzo: actually, that's a pretty good idea and it will simplify things quite a bit -- no need for separate `pend` and `writing` sets this way.

vyzo: wait, we can't -- failure semantics. I'll keep it as is as the improvement would be marginal.

blockstore/splitstore/markset_badger.go
vyzo: yeah, very small delta; I'll leave it as is.

vyzo:
rebased on master for merge, as it was a bit behind.

PR #6961, opened 2021-07-31
make epoch time for mainnet more intuitive

zlhwdsz:


cli/util.go
```diff
@@ -45,6 +45,12 @@ func parseTipSet(ctx context.Context, api v0api.FullNode, vals []string) (*types
 }
 
 func EpochTime(curr, e abi.ChainEpoch) string {
+	if build.BuildType == build.BuildMainnet {
+		start := time.Date(2020, 8, 24, 22, 0, 0, 0, time.UTC)
+		eTime := start.Add(time.Second * time.Duration(int64(build.BlockDelaySecs)*int64(e)))
+		return fmt.Sprintf("%d (%s)", e, eTime.Format("2006-01-02 15:04 MST"))
```
ZenGround0, line 51: @zlhwdsz I think the notion of relative time is actually more intuitive than including the timestamp.  I can see the timestamp helping though so I'll approve if you print both for mainnet Epoch time: `(%d, %s ([now | %s ago | in %s]))`

ZenGround0 (changes requested):

cli/util.go
zlhwdsz: df95d151fec0ddc5f50173edc16d5cf

zlhwdsz: This way the check-expire lines will be lengthy and hard to read:

df95d151fec0ddc5f50173edc16d5cf

I will close this PR and provide a cusomized epochTime function for check-expire cmd in a new PR.

PR #6967, opened 2021-08-01
lotus-shed splitstore clear command

vyzo:
This is very useful when the splitstore/chain is linked in volatile storage and the user wants to restart from a new snapshot.


jennijuju:
Lint is sad

vyzo:
fixed, it was a bug too!

ZenGround0 (approved):
LGTM

PR #6970, opened 2021-08-02
Remove forgotten non-functioning config from the pre-mainnet days

ribasushi:


tools/packer/repo/config.toml
```diff
@@ -13,7 +13,7 @@ ListenAddresses = ["/ip4/0.0.0.0/tcp/5678", "/ip6/::/tcp/5678"]
 #
 [Pubsub]
 #  Bootstrapper = false
-#  RemoteTracer = "/dns4/pubsub-tracer.filecoin.io/tcp/4001/p2p/QmTd6UvR47vUidRNZ1ZKXHrAFhqTJAD27rKL9XYghEKgKX"
+#  RemoteTracer = ""
```
ZenGround0, line 16: What's the reason for keeping this dead config field around?  What happens if we just remove it instead of zeroing it out?

ZenGround0 (approved):

tools/packer/repo/config.toml
ribasushi: @ZenGround0 I do not know enough about this... @whyrusleeping thoughts?

whyrusleeping:
⛵

PR #6978, opened 2021-08-03
repro backup restore error

dirkmc:
On a local devnet:
- start a miner and daemon process
- create a backup of the datastore:
  `./lotus-miner backup /tmp/backup.cbor`
- restore the backup into a new markets repo (see https://docs.filecoin.io/mine/lotus/split-markets-miners/#splitting-the-lotus-miner-monolith)

The resulting error is
```
 7.68 KiB / 13.71 KiB [==================================================================>---------------------------------------------------]  56.01% 6.04 MiB/s 0s
ERROR: restoring metadata: reading backup: unmarshaling log entry: cbor input should be of type array
```

This is an example backup file generated through the above process:
[backup.cbor.zip](https://github.com/filecoin-project/lotus/files/6925148/backup.cbor.zip)

To repro unzip the above file, put it at /tmp/backup.cbor and run TestRestore (included in this PR)

### DIFF
--- a/cmd/lotus-miner/init_restore_test.go
+++ b/cmd/lotus-miner/init_restore_test.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"os"
+	"testing"
+
+	ds "github.com/ipfs/go-datastore"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/lotus/lib/backupds"
+	"golang.org/x/xerrors"
+)
+
+func TestRestore(t *testing.T) {
+	require.NoError(t, runBackup())
+}
+
+func runBackup() error {
+	bf := "/tmp/backup.cbor"
+
+	f, err := os.Open(bf)
+	if err != nil {
+		return xerrors.Errorf("opening backup file: %w", err)
+	}
+	defer f.Close() // nolint:errcheck
+
+	dstore := ds.NewMapDatastore()
+	err = backupds.RestoreInto(f, dstore)
+
+	if err != nil {
+		return xerrors.Errorf("restoring metadata: %w", err)
+	}
+
+	return nil
+}


rvagg:
Here's what's in the file and the cbor error: https://gist.github.com/rvagg/5135dc579e4c5a1315402cafd997a5d5

At the start it matches what's in lib/backupds/datastore.go in that it's an `array(2)`, with the first entry being an `array(indefinite)` full of `array(2)`s, each with two `byte(x)` entries. The indefinite length array breaks, then there's a `byte(32)` which I assume is the checksum .. but then there's more terminals, which there shouldn't be because it was only an `array(2)` at the top:

```
    74                                            #     string(20)
      6368efbfbd4a2f736563746f72732f325902efbfbd  #       "ch�J/sectors/2Y�"
      efbfbd1a65                                  #       "�e"
    53                                            #     bytes(19)
      746174656750726f76696e676c536563746f72      #       "tategProvinglSector"
    4e                                            #     bytes(14)
      756d626572026a536563746f7254                #       "umber\x02jSectorT"
```

and after that there's this `string(28776)` but it only has 6119 bytes remaining.

It looks well-formed at the beginning, but then there's a bunch of cruft at the end. Note that there's no use of `string` majors anywhere in backup, so the `string(20)` above is coming from somewhere else, and the `string(28776)` is certainly from somewhere else and possibly just a random byte being interpreted as cbor.

rvagg:
If it helps, the 32-byte sha2-256 checksum at byte position 7826 matches the hash of the bytes it's supposed to - 1 to 7826 (just the indefinite-length array portion). So the data is correct as far as was written, but there's some trailing cruft.

And, if you trim the bytes to the end of the checksum, 7860, then it's properly formed CBOR as expected - a 2 element array, first element being indefinite-length (with 22 elements, each of which is a [bytes, bytes] tuple) and the second element being the 32-byte checksum.

It's these trailing bytes beyond 7860 that are a problem, invalid CBOR and looking kind of random.

rvagg:
[backup.cbor.trimed.zip](https://github.com/filecoin-project/lotus/files/6935454/backup.cbor.trimed.zip)

^ there's the trimmed backup file if you want to try reimporting it to see if it works as expected.

raulk (commented):
This would benefit from @magik6k's attention. It's not clear if this happens only on fresh miners with no activity/sectors, or if it happens on any miner as long as it has _zero_ deals. The latter would be more worrying.

@dirkmc agreed to test which of the two situations it is on a devnet.

cc @jennijuju

dirkmc:
Sorry everyone, the problem was that I wasn't removing the existing backup file before creating a new one. So the second backup would corrupt the original backup.
Resolution: delete any existing backup file before creating a new one.

dirkmc:
Ticket to improve CLI to avoid this issue in future:
Calling lotus-miner backup twice with the same output path corrupts the backup file: https://github.com/filecoin-project/lotus/issues/7028

PR #596, opened 2019-11-13
Add rebup for local mesages

Kubuxu:

PR #6979, opened 2021-08-03
dagstore + carv2 refactors; integrate with config

raulk:
This PR integrates the dagstore + CARv2 solution with the configuration system in Lotus (`config.toml`).

It also exports the DAGStore to the DI context in preparation for implementing the upcoming JSON-RPC operations and cli commands.

Finally, it also performs renames and refactors for code cleanup/consolidation and consistency.


go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/elastic/gosigar v0.12.0
 	github.com/etclabscore/go-openrpc-reflect v0.0.36
 	github.com/fatih/color v1.9.0
-	github.com/filecoin-project/dagstore v0.4.0
+	github.com/filecoin-project/dagstore v0.4.1-0.20210803144142-2ef26aaf8e18
```
raulk, line 29: Need to make a release.

node/modules/storageminer_dagstore.go
```diff
@@ -0,0 +1,118 @@
...
+		if err == nil {
+			cfg.MaxConcurrencyStorageCalls = concurrency
+		}
+	}
+
+	mountApi := mdagstore.NewMinerAPI(pieceStore, rpn, cfg.MaxConcurrencyStorageCalls)
+	ready := make(chan error, 1)
+	pieceStore.OnReady(func(err error) {
+		ready <- err
+	})
```
dirkmc, line 46: Do we want to wait for ready inside a go routine?

node/modules/storageminer_dagstore.go
```diff
@@ -0,0 +1,118 @@
...
+		cfg.TransientsDir = filepath.Join(defaultDir, "transients")
+	}
+	if cfg.IndexDir == "" {
+		cfg.IndexDir = filepath.Join(defaultDir, "index")
+	}
+	if cfg.DatastoreDir == "" {
+		cfg.DatastoreDir = filepath.Join(defaultDir, "datastore")
+	}
+
+	v, ok := os.LookupEnv(EnvDAGStoreCopyConcurrency)
```
dirkmc, line 83: Why does the copy concurrency come from from an env var instead of repo config?

markets/dagstore/wrapper.go
```diff
@@ -3,89 +3,73 @@ package dagstore
...
 	failureCh  chan dagstore.ShardResult
 	traceCh    chan dagstore.Trace
 	gcInterval time.Duration
 }
 
-var _ shared.DagStoreWrapper = (*Wrapper)(nil)
+var _ stores.DAGStoreWrapper = (*Wrapper)(nil)
 
-func NewDagStoreWrapper(cfg MarketDAGStoreConfig, mountApi LotusAccessor) (*Wrapper, error) {
+func NewDAGStore(cfg config.DAGStoreConfig, mountApi MinerAPI) (*dagstore.DAGStore, *Wrapper, error) {
```
dirkmc, line 46: Why do we need to return *dagstore.DAGStore here? Is it just so we can call close on the DAGStore?
If so I'd suggest adding a Close method to the Wrapper instead. That way the caller only has to deal with one interface, and we could avoid needing the DI trickery with the "special" type

node/config/def.go
```diff
@@ -188,6 +188,15 @@ func DefaultStorageMiner() *StorageMiner {
 			TerminateControl:   []string{},
 			DealPublishControl: []string{},
 		},
+
+		DAGStore: DAGStoreConfig{
+			// zero value paths for transients, indices, datastore.
+			// by default they'll end up under the node repo.
```
dirkmc, line 194: It wasn't immediately clear to me what this comment was referring to (until I went and looked at the DAGStoreConfig struct)

node/impl/client/import.go
```diff
@@ -2,75 +2,96 @@ package client
...
+	bsvc = blockservice.New(bs, offline.Exchange(bs))
+	dags = merkledag.NewDAGService(bsvc)
+	finalRoot, err := buildUnixFS(ctx, src, dags)
 	if err != nil {
-		return cid.Undef, xerrors.Errorf("failed to create Unixfs DAG with CARv2 blockstore: %w", err)
+		_ = bs.Close()
+		return cid.Undef, xerrors.Errorf("failed to create UnixFS DAG with carv2 blockstore: %w", err)
+	}
+
+	if err := bs.Finalize(); err != nil {
```
dirkmc, line 80: Having both Close() and Finalize() on this interface is confusing - the go-car team is going to eliminate Close right?

node/impl/client/import.go
```diff
@@ -2,75 +2,96 @@ package client
...
-	root, err := importNormalFileToUnixfsDAG(ctx, inputFilePath, merkledag.NewDAGService(bsvc))
+	root, err := buildUnixFS(ctx, src, dags)
 	if err != nil {
-		return cid.Undef, xerrors.Errorf("failed to import file to store: %w", err)
+		_ = fstore.Close()
+		return cid.Undef, xerrors.Errorf("failed to import file to store to compute root: %w", err)
+	}
+
+	if err := fstore.Close(); err != nil {
+		return cid.Undef, xerrors.Errorf("failed to finalize car filestore: %w", err)
```
dirkmc, line 60: Should this be fstore.Finalize()?

node/config/types.go
```diff
@@ -50,6 +50,45 @@ type StorageMiner struct {
...
+	MaxConcurrentReadyFetches int
+
+	// The maximum number of simultaneous inflight API calls to the storage
+	// subsystem.
+	// Default value: 100.
+	MaxConcurrencyStorageCalls int
+
+	// The number of milliseconds between calls to periodic dagstore GC.
+	// Default value: 60000 (60 seconds = 1 minute).
+	GCIntervalMillis int
```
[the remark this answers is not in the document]
raulk: Nice, didn't realise we had a `Duration` type!

node/config/types.go
```diff
@@ -50,6 +50,45 @@ type StorageMiner struct {
 	Storage    sectorstorage.SealerConfig
 	Fees       MinerFeeConfig
 	Addresses  MinerAddressConfig
+	DAGStore   DAGStoreConfig
+}
+
+type DAGStoreConfig struct {
```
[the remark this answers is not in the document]
raulk: I thought of that, but miners want flexibility here because transients, index and datastore have different sizing expectations, and also require different durability and access guarantees (e.g. transients can be wiped out, but index and datastore must remain consistent and durable). In practice, miners will place transients in volatile/cheap storage, and index and datastore in durable, replicated, fast-access storage.

Granted that miners can use symlinks, but I consider that a hack to overcome a limitation of the system.

> "if anything goes wrong just delete the dagStore directory"

I think this is a valuable property to have for testing, but as we move to production, destroying your dagStore directory won't be something we ask for very frequently (or at all).

node/modules/storageminer_dagstore.go
raulk: We're waiting for ready below, in the OnStart hook.

node/modules/storageminer_dagstore.go
raulk: It actually comes from the configuration, and _is overridden_ by the env var. I just wanted to respect the env variable we introduced, but we can also kill it since it's probably something we don't want to perpetuate?

markets/dagstore/wrapper.go
raulk: Ah no, we need `*dagstore.DAGStore` to be outjected/exported to the DI context because we'll inject it in the JSON-RPC layer to serve operations like `DagstoreShardList`, `DagstoreShardInitialize`.

raulk: An alternative would be to _embed_ the `*dagstore.DAGStore` into the `*Wrapper`, but that is more non-obvious and requires a deeper refactor (there are name collisions).

raulk: Oh, another thing. fx has pretty nice / first-class support for multiple outjections from a constructor: https://pkg.go.dev/go.uber.org/fx?utm_source=godoc#Out. It's just that the utils we've created on the Lotus side make this a "special" case.

node/config/def.go
raulk: I'll revise!

node/impl/client/import.go
raulk: The thing here is that the go-car library manages the file descriptor -- we provide a path on `OpenRead{Only,Write}`, and it opens the file internally. `Close()` just closes that file descriptor (which can be used to pause a CAR being written, or to abort it entirely -- which is what we do here), and `Finalize()` computes the final index, appends it, and closes the file descriptor. They aren't going to eliminate `Close()`, I think -- they overlooked that there's an type embed on their side that makes that difficult.

About this particular usage: this is the error case, so we don't want to finalize, as we are aborting.

node/impl/client/import.go
raulk: Nope, because we use `stores.ReadWriteFilestore()` which returns a ClosableBlockstore whose `Close` method calls `Finalize`. This design was already in place in the wip branch, we certainly can review it because it's rather footgunny.

(In a previous review of go-fil-markets I had removed these utilities because they created obscurity IMO -- see commit [`4d3607a` (#596)](https://github.com/filecoin-project/go-fil-markets/pull/596/commits/4d3607a44c8d9f7ce14def8d21bf2436fcb5cff9), but then readded them to keep code DRY -- happy to revisit!)

node/config/types.go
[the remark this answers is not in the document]
dirkmc: Ok fair points 👍

node/modules/storageminer_dagstore.go
dirkmc: Right my bad 👍

node/modules/storageminer_dagstore.go
dirkmc: Yeah I think it would be simpler to get rid of the env var and just have the value come from config

markets/dagstore/wrapper.go
dirkmc: I see, makes sense 👍 

It _might_ be simpler to have a separate method that returns the `*dagstore.DAGStore`, and then inject that into DAGStoreWrapper through DI, but defer to your judgment as it sounds like you have plans for where to use the `*dagstore.DAGStore` in the near future.

raulk:
Looks like this is reviewed and ready to be merged after I upgrade the go-fil-markets dep.

PR #6983, opened 2021-08-04
dagstore: add JSON-RPC operations and cli.

raulk:
Three operations and with their corresponding CLI commands added:

* `DagstoreInitializeShard`
* `DagstoreInitializeAll`
* `DagstoreListShards`
* `DagstoreGC`

The first two are a prerequisite for externalizing bulk shard initialization as step 2 of the migration.


documentation/en/api-v0-methods-miner.md
```diff
@@ -345,6 +349,46 @@ Inputs:
...
+
+Perms: admin
+
+Inputs: `null`
+
+Response: `null`
+
+### DagstoreInitializeShard
+DagstoreInitializeShard initializes an uninitialized shard by acquiring
+it and releasing it as soon as it's ready.
```
dirkmc, line 367: I think it's not that clear why you would want to use this command if you don't know how the DAG store works internally. Could we add some context to this description? (and maybe the description in the CLI help)

dirkmc (approved):

dirkmc:
LGTM 👍 

One suggestion I have is to make pagination available for the shard list API (not necessarily for the CLI but that would be a nice to have). But that could be done in a follow-up PR.

documentation/en/api-v0-methods-miner.md
raulk: Ack, I didn't feel happy writing this. Will revise.

node/impl/storminer.go
```diff
@@ -616,6 +616,36 @@ func (sm *StorageMinerAPI) DagstoreInitializeShard(ctx context.Context, key stri
...
+
+	k := shard.KeyFromString(key)
+
+	info, err := sm.DAGStore.GetShardInfo(k)
+	if err != nil {
+		return fmt.Errorf("failed to get shard info: %w", err)
+	}
+	if st := info.ShardState; st != dagstore.ShardStateErrored {
+		return fmt.Errorf("cannot recover shard; expected state ShardStateErrored, was: %s", st.String())
+	}
```
dirkmc, line 632: We may just want to return nil when the shard is not in an error state - otherwise it could be hard to avoid a race where you check the shard state, see that it's error and call recover shard, but in the mean time another process has already recovered the shard.

raulk: It won't race, since the second call will fail.

PR #6998, opened 2021-08-06
Test with latest actors version

Stebalien:
We were already trying to do this, but had some issues:

1. We were forced to run all upgrades early, in-sequence.
    1. This actually has some issues as can be seen in https://github.com/filecoin-project/specs-actors/issues/1453.
    2. We had to carefully wait for upgrades to finish before moving on.
    3. This slows down the tests.
2. We weren't consistently testing with the latest version.
    1. Lots of tests would just run the default upgrade schedule, and would sometimes fail if the upgrade did something unexpected.
    2. We weren't always testing the latest code.

This patch:

1. Sets the genesis version, where applicable.
2. Defaults to an empty upgrade schedule with genesis at the latest network version.
3. Makes some of the tests more reliable (e.g., wait a _full_ proving period instead of the _remainder_ of the proving period).
4. Fixes genesis at actors v2.

~TODO figure out why it's failing:~

1. I'm seeing errors about missing t_aux files in tests that try to use the actual prover. I have no idea how these tests worked before.
2. ~When I remove the check for t_aux files (do we need them with fauxseals?), post still doesn't work. I'm seeing an error about trying to prove deadline 0 when we're already in deadline 1, which is suspiciously like https://github.com/filecoin-project/specs-actors/issues/1453.~

Answer: in case 2, my machine was just too slow. CI and my new machine are faster.


Stebalien:
Rebasing on #7011. The tests now pass locally when doing this. I believe this was a timing issue (I'm now using a faster computer).

However, this has me a bit concerned (for the tests, not the network). Now that we're starting from a later actors version, we may have to prove pre-seal sectors _immediately_ where _before_ the proving epoch would start later. In theory, we could start in the middle of of an active proving window. I'm pretty sure that's what was happening, actually...

Stebalien:
Given that the tests pass reliably in CI, I think we're probably fine.

Stebalien:
Hm. So I _still_ see this issue if I enable TestDealWithMarketAndMinerNode (at least locally). Now, this test _was_ disabled for a reason, but this is still interesting.

chain/gen/genesis/miners.go
```diff
@@ -584,14 +587,23 @@ func currentEpochBlockReward(ctx context.Context, vm *vm.VM, maddr address.Addre
 	}
 
 	// TODO: This hack should move to reward actor wrapper
```
ZenGround0, line 589: Does this TODO still represent our current plans?  If so I can look into scoping this work on the actors side.  However it seems like our idea of this has changed and this is no longer considered a hack.  If so let's remove the TODO.

itests/kit/ensemble.go
```diff
@@ -208,11 +216,15 @@ func (n *Ensemble) Miner(miner *TestMiner, full *TestFullNode, opts ...NodeOpt)
 			genm    *genesis.Miner
 		)
 
+		// Default 2KiB sector for the network version
```
ZenGround0, line 219: nit: seems like this should be an ensemble option so the default can be overridden, though maybe this should be left to whoever first wants to use non-2KiB sectors in an itest

itests/ccupgrade_test.go
```diff
@@ -17,10 +17,9 @@ func TestCCUpgrade(t *testing.T) {
 	kit.QuietMiningLogs()
 
 	for _, height := range []abi.ChainEpoch{
-		-1,   // before
-		162,  // while sealing
-		530,  // after upgrade deal
-		5000, // after
```
Stebalien, line 23: There wasn't really a point to this. It was just testing with the previous actors version.

itests/kit/ensemble_opts.go
```diff
@@ -19,10 +21,16 @@ type ensembleOpts struct {
 	verifiedRoot genesisAccount
 	accounts     []genesisAccount
 	mockProofs   bool
+
+	upgradeSchedule stmgr.UpgradeSchedule
 }
 
 var DefaultEnsembleOpts = ensembleOpts{
 	pastOffset: 10000000 * time.Second, // time sufficiently in the past to trigger catch-up mining.
+	upgradeSchedule: stmgr.UpgradeSchedule{{
```
ZenGround0, line 30: nit: the fact that this is so nicely expressed as `GenesisNetworkVersion(build.NewestNetworkVersion)` makes me wonder if we'd be slightly better off with a DefaultEnsembleOpts of type `[]EnsembleOpts` applied to a zero valued `ensembleOpts`.  The name `ensembleOpts` could be changed to something different, `ensembleCfg`?, to make this a little less confusing.

Not very important just riffing.

itests/wdpost_test.go
```diff
@@ -63,7 +65,7 @@ func testWindowPostUpgrade(t *testing.T, blocktime time.Duration, nSectors int,
 	require.NoError(t, err)
 
 	t.Log("Running one proving period")
-	waitUntil := di.PeriodStart + di.WPoStProvingPeriod + 2
```
ZenGround0, line 66: Was this what fixed the "proving deadline 0 in deadline 1" issue?

ZenGround0 (approved):
LGTM

chain/gen/genesis/miners.go
Stebalien: Ideally this would move into the reward actor _wrapper_. That is, the shim in chain/actors/builtin/reward. We have some abstractions there around message sending.

itests/kit/ensemble.go
Stebalien: Yeah, I'm going to leave it to them.

itests/kit/ensemble_opts.go
Stebalien: I agree. I'd much rather have a set of default options (also avoids cases where the default "config" gets modified accidentally). But that'll be a future cleanup.

itests/wdpost_test.go
Stebalien: You can see that issue in the "+20". Termination happens at the end of the deadline _always_. If we need "slack", something is wrong.

PR #6974, opened 2021-08-02
Unsubscribe slow chain head subscribers

Stebalien:
This way, slow readers can just "resubscribe" when they're able to catch up. That way, a single slow leader doesn't block all notifications.

fixes #6883, #6947

PR #7000, opened 2021-08-06
Refactor events subsystem

Stebalien:
This builds on #6974.

## Changes

1. Pass contexts everywhere instead of storing them.
2. Handle "catch up" if the chain notify channel is closed.
3. Unify events around a single "observer" system. To simplify things, observers are passed both a to and a from tipset instead of just a "target" tipset. On reverts, this "target" tipset was actually the from tipset (the tipset being reverted) which was confusing.
4. Improve the cache:
    1. Validate inputs. Given the "catch up" handling, this may not be necessary.
    2. Expose the normal lotus API. This removes the temptation to use the cache as a source of truth.
    3. Ensure that we're always able to call through the cache when necessary (instead of just failing when something isn't in the cache).
    4. Use this caching API everywhere, instead of explicitly using the cache.
5. Improve the height system.
    1. Simplify  handler management.
    2. Implement handler GC.
6. Improve the hcEvents system.
    1. Correctly handle reverts (I think?). Previously, a revert from A to B, then an apply from B to C, would check for changes from A to C, not changes from B to C (it would store the last tipset as A instead of B).
    2. Store the tipsets instead of the heights for queued events (faster, less error cases, etc.).
7. Propegate errors when constructing the events system.
8. Handle pure reverts. Technically, a user can call SetHead to revert to a parent without applying any new blocks. Before, we would have restarted and likely would have corrupted the event state.
9. Optimized message handling (i.e., avoided re-hashing messages).

## Architecture

The architecture isn't significantly different from the current one.

1. Observer: subscribes to head changes and calls apply/revert on event subsystems.
2. Cache: Caches messages (LRU); tipsets and head on the current chain (by subscribing to the observer).
3. Event Subsystems: Hook into the observer to react to chain changes.

```
[Lotus Node]
   ^     | (head change notifications)
   |     |
  cache<-| -------------------+
  ^ ^    |                    |
  | |    v                    |
  |observer                   |
  |   | (apply/revert events) |
  |  / \______________________+
  | v
[Event Subsystems] # height events, message events, state events, etc...
```

The main problem with this architecture is that it's entirely synchronous (callbacks). I'd prefer to use channels, but that's a larger refactor and I assume there are assumptions being made about everything happening in lock-step.

## Future

Future improvements:
- Reduce locking further.
- Simplify the "ChainAt" function. The lock dance we're playing there is likely overkill.
- We could probably re-implement much of hcEvents in-terms of more generalized events (e.g., we might be able to use the height event system).
- Propagate contexts through `StateChanged`
- Maybe remove the cache? We only _need_ it for looking up "lookback" tipsets on timeout in the hcEvents system. Everywhere else, we just remember the tipsets we need for later as we get them.

## TODO (before merge)

Improve test coverage (at least test the fixed bugs/edge cases).frob


chain/events/events_test.go
```diff
@@ -1334,20 +1264,15 @@ func TestStateChangedTimeout(t *testing.T) {
...
-	}
-	require.NoError(t, fcs.tsc.add(fcs.makeTs(t, nil, 1, dummyCid)))
+	fcs = newFakeCS(t)
+	events, err = NewEvents(context.Background(), fcs)
+	require.NoError(t, err)
 
-	events = NewEvents(context.Background(), fcs)
+	// XXX: Needed to set the latest head so "check" succeeds". Is that OK? Or do we expect
+	// check to work _before_ we've received any events.
+	fcs.advance(0, 1, nil)
```
Stebalien, line 1273: @magik6k is this OK?

Previously: if we hadn't yet seen a block, we'd ask for the latest head and call check on that.
Now: if we haven't yet seen a block, we don't call check.

I made this change because there was no guarantee that the caller would see events for blocks between that latest head and the new head.

Possible solutions:

1. Leave this as it is.
2. Provide a way to get the latest head that guarantees that the caller will also get all events after that head.

magik6k: I think this is ok, the main point of the check method is providing atomicity guarantees (state change can't be missed).

(though we should check that nothing depends on the check being synchronous)

Stebalien: So, currently it _won't_ call the check function (at all) if no events have been received. I'll fix that.

Stebalien: Ok, I've now made registering an observer return the current head. That should make this atomic and avoid any potential deadlocks.

Stebalien:
> Simplify the "ChainAt" function. The lock dance we're playing there is likely overkill.

I'm happy to take a crack at this now if deemed necessary.

chain/events/cache.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/ipfs/go-cid"
+)
+
+type uncachedAPI interface {
+	ChainNotify(context.Context) (<-chan []*api.HeadChange, error)
+	ChainGetPath(ctx context.Context, from, to types.TipSetKey) ([]*api.HeadChange, error)
```
hunjixin, line 15: its better to reorg in local . ChainGetPath need the same node. but if have multi nodes behind gateway. maybe error 
 here

chain/events/events_called.go
```diff
@@ -516,47 +465,32 @@ func (me *messageEvents) checkNewCalls(ts *types.TipSet) (map[triggerID][]eventD
...
-			msgsI, err = me.cs.ChainGetBlockMessages(context.TODO(), tsb.Cid())
-			if err != nil {
-				log.Errorf("messagesForTs MessagesForBlock failed (ts.H=%d, Bcid:%s, B.Mcid:%s): %s", ts.Height(), tsb.Cid(), tsb.Messages, err)
-				// this is quite bad, but probably better than missing all the other updates
-				me.blockMsgLk.Unlock()
-				continue
-			}
-			me.blockMsgCache.Add(tsb.Cid(), msgsI)
+	for i, tsb := range ts.Cids() {
+		msgs, err := me.cs.ChainGetBlockMessages(context.TODO(), tsb)
```
hunjixin, line 476: get message block by block. but in fact there are many messages not to be executed or duplicate. could use ChainGetMessagesInTipset here?

chain/events/observer.go
```diff
@@ -0,0 +1,255 @@
...
+			log.Warn("listenHeadChanges quit")
+		}
+		select {
+		case <-build.Clock.After(time.Second):
+		case <-ctx.Done():
+			log.Warnf("not restarting listenHeadChanges: context error: %s", ctx.Err())
+			return
+		}
+
+		log.Info("restarting listenHeadChanges")
```
vyzo, line 69: this will likely spam a lot, it's once a second with no additional information.

vyzo (commented):
First pass, this looks sane so far.

I'll have to do a second pass to focus on the details, as this is a rather complex pass.

chain/events/cache.go
Stebalien: ChainGetPath is exposed over the API, but not the gateway. If we did this locally, we'd have to go back and forth to the API for every head change which would be _very_ slow.

Are you using the events subsystem from behind a _gateway_ (i.e., do you have custom code that relies on this)?

chain/events/events_called.go
Stebalien: That's a good point, but something for a followup patch. I'm replicating the current behavior here.

chain/events/observer.go
Stebalien: Only if we get disconnected because we're reading slowly. Basically, `ChainNotify` buffers internally (16 blocks?), so we'll only see this if we fail to keep up with the chain for 8 minutes.

chain/events/events_called.go
Stebalien: Hm. Actually, I _think_ this code is correct as-is. We _want_ to know if a message was included in the tipset, even if it wasn't executed. That way we can check to see if the message successfully applied. Otherwise, we'll keep waiting for the message to show up even though it never will.

Any context here @magik6k?

chain/events/cache.go
hunjixin: use nginx to proxy market.

chain/events/events_called.go
magik6k: I think this is fine for now as it doesn't seem change the current behavior

magik6k (approved):

PR #7011, opened 2021-08-09
fix: don't check for t_aux when proving

Stebalien:
We don't need it (I think?).


Stebalien:
Note: this needs a _very_ careful review (by @cryptonemo or @porcuquine ?), and ideally a quick test in a dev net with real sectors.

porcuquine (commented):
`t_aux` is 'temporary aux'. `p_aux` is persistent. You need `p_aux` for PoST, but you only need `t_aux` until the vanilla PoRep proof has been created. I can't tell from the context here when the check happens, so I can't tell whether it can be removed here.

It's not true that you *never* need `t_aux`, but the need is only temporary. If you want to gain high confidence in this change in a pragmatic way, try changing the check to instead delete that file, then see if you can get all the way to a verifying PoRep. If so, then it's safe.

The risk is that you remove the check and observe that this doesn't break anything in testing. However, if the file is still there (though unchecked), this might be a false negative. In that case, you would have failed to exercise the check in the case that it was needed.

(Alternately, if you want to test the candidate code directly, you can enhance the test to ensure the file has been deleted before the check is run — but in that case, make sure you don't do that before it is even generated for the sector.)

Stebalien:
>  I can't tell from the context here when the check happens, so I can't tell whether it can be removed here.

Digging through the code, this check is _only_ for window post.

porcuquine:
> Digging through the code, this check is only for window post.

Based on that assumption, and the general statements I made above, this should be safe to remove then. Again, the temporary/persistent distinction pertains exactly to whether the auxiliary data is required to complete PoRep over over time indefinitely as checked by PoST.

Stebalien:
Ok, this definitely fixes https://github.com/filecoin-project/lotus/pull/6998.

magik6k (approved):

PR #7033, opened 2021-08-11
move client-side handling of blockstores/CARs to Lotus from go-fil-markets

dirkmc:
WIP: still need to
- work out the  dependency injection details
- reconcile the paths in the blockstore accessor implementations with the paths in the import manager (node/repo/importmgr/mgr.go)


node/repo/importmgr/mgr.go
```diff
@@ -35,11 +35,11 @@ const (
 	LFileStoreCARv2FilePath = "CARv2Path" // path of the full CARv2 file or a CARv2 file that can serve as the backing store for a Filestore.
 )
 
-func New(ds datastore.Batching, repoPath string) *Mgr {
+func New(ds datastore.Batching, tmpDir string) *Mgr {
 	return &Mgr{
-		repoPath: repoPath,
-		ds:       datastore.NewLogDatastore(namespace.Wrap(ds, datastore.NewKey("/stores")), "storess"),
-		counter:  shared.NewTimeCounter(),
+		tmpDir:  tmpDir,
```
raulk, line 40: Technically speaking, this doesn't need to be a tempdir. Can we call `baseDir`?

markets/retrievaladapter/blockstoremanager.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+// BlockstoreManager provides an interface to get a blockstore and perform
+// transformations based on a DAG root CID.
+// It is used to provide the retrieval client with a blockstore backed by
+// either an IPFS or CARv2 file.
+type BlockstoreManager interface {
+	retrievalimpl.BlockstoreAccessor
+	WriteCarV1(ctx context.Context, rootCid cid.Cid, out string) error
+	WriteUnixFSFile(ctx context.Context, rootCid cid.Cid, out string) error
```
raulk, line 32: This seems odd. Why do these methods _need_ to be here? This logic for writing a CARv1 or a UnixFS file will not change across implementations AFAIK.

The input for those methods is a root CID and a **blockstore**. The caller can get the blockstore from the `BlockstoreAccessor`, and it has the root CID in its local scope.

IMO these should be top-level utility functions taking a blockstore, not members of an interface.

node/repo/importmgr/mgr.go
dirkmc: It's only used in NewTempFile(), but in any case happy to change the name

raulk:
Superseded by https://github.com/filecoin-project/lotus/pull/7079, which solves the TODOs and takes a slightly different approach.

Issue #7020, opened 2021-08-10
Revert 6766


jennijuju:
Need to backport to v1.11.1

- [x] integrate the spec actor release that has reverted go state network version change 
- [ ] Revert 6766

jennijuju:
this is blocking the v1.11.1 release

jennijuju:
resolved by #7022

PR #7038, opened 2021-08-11
Strict major minor version checking on v0 and v1 apis

ZenGround0:
Related to #7012 #7020 I've been thinking that a step in the right direction is to enforce that new nodes are talking to the same API version both v0 and v1.  This won't help the problem of needing to support backwards compatibility for the v0 api for older coder versions but it will prevent nodes from this point onwards from getting too far and erroring in confusing ways when attempting to work against different api versions.

I put this up (1) to see if it breaks CI in interesting ways (2) to get the opinion of @filecoin-project/lotus-maintainers @whyrusleeping and @raulk.

### DIFF
--- a/cli/util/api.go
+++ b/cli/util/api.go
@@ -236,19 +236,7 @@ func GetFullNodeAPI(ctx *cli.Context) (v0api.FullNode, jsonrpc.ClientCloser, err
 		_, _ = fmt.Fprintln(ctx.App.Writer, "using full node API v0 endpoint:", addr)
 	}
 
-	v0API, closer, err := client.NewFullNodeRPCV0(ctx.Context, addr, headers)
-	if err != nil {
-		return nil, nil, err
-	}
-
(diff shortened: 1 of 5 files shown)


Stebalien:
Hm. I think strict version checking makes sense for v1, but not v0. v1 is "unstable", so every version can be breaking. v0 is stable, so new versions should be backwards compatible.

Stebalien (approved):

Stebalien:
(well, approved once tests pass)

And we can even use even/odd versioning here as well. I.e., when we're ready to "stabilize" v1, we can cut a v2 that's stable and a v3 that's unstable. That should make it easier to track where we need strict version checks and where we don't.

magik6k (commented):
Looks like there is some weirdness with Version on lotus-gateway

ZenGround0:
Yeah the test kit cli testing tool doesn't set the `api.RunningNodeType` global.  Should be fixed now.

PR #7042, opened 2021-08-12
make `lotus-miner net` commands hit markets subsystem.

raulk:
The libp2p host on the miner is now only started where the markets subsystem runs. This PR makes `lotus-miner net` commands hit the node where the markets subsystem is running.

Honestly, `Net` no longer belongs in the `Common` commands, nor should the endpoint be resolved through `GetCommonAPI`/`GetAPI`, but there is some larger untangling to do to tidy this up, and it's associated with pre-existing tech debt.

Choosing not to do it now, as there are other decisions to make first with regards to providing the optimal UX experience for miners, now and as we introduce more miner services.

---

Testing: unfortunately we don't have a decent framework/suite for CLI testing. All itests hit the JSON-RPC API or the native API. So I can't add a test for this. Instead, I performed manual testing and all looks good.

### DIFF
--- a/cli/cmd.go
+++ b/cli/cmd.go
@@ -58,7 +58,6 @@ var GetMarketsAPI = cliutil.GetMarketsAPI
 var GetWorkerAPI = cliutil.GetWorkerAPI
 
 var CommonCommands = []*cli.Command{
-	// NetCmd: no longer a common command, since lotus-miner nodes without a markets subsystem no longer run a libp2p node
 	AuthCmd,
 	LogCmd,
 	WaitApiCmd,
--- a/cli/cmd.go
+++ b/cli/cmd.go
@@ -58,7 +58,7 @@ var GetMarketsAPI = cliutil.GetMarketsAPI
 var GetWorkerAPI = cliutil.GetWorkerAPI
 
 var CommonCommands = []*cli.Command{
-	NetCmd,
+	// NetCmd: no longer a common command, since lotus-miner nodes without a markets subsystem no longer run a libp2p node
 	AuthCmd,
 	LogCmd,
 	WaitApiCmd,
--- a/cmd/lotus-miner/main.go
+++ b/cmd/lotus-miner/main.go
@@ -5,16 +5,17 @@ import (
 	"fmt"
 
 	"github.com/fatih/color"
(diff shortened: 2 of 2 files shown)


raulk:
**Before**

```sh
$ lotus-miner net peers
ERROR: method not supported
```

**After**

```sh
$ lotus-miner net peers
12D3KooW..., [/ip4/1.2.3.4/tcp/1]
12D3KooW..., [/ip4/2.3.4.5/tcp/2]
12D3KooW..., [/ip4/3.4.5.6/tcp/3]
12D3KooW..., [/ip4/4.5.6.7/tcp/4]
```

Note how I don't need to use `--call-on-markets`.

dirkmc (approved):
Nice 👍

Kubuxu (approved):

cli/cmd.go
```diff
@@ -58,7 +58,7 @@ var GetMarketsAPI = cliutil.GetMarketsAPI
 var GetWorkerAPI = cliutil.GetWorkerAPI
 
 var CommonCommands = []*cli.Command{
-	NetCmd,
+	// NetCmd: no longer a common command, since lotus-miner nodes without a markets subsystem no longer run a libp2p node
```
magik6k, line 61: Would just remove from here

magik6k (changes requested):
We also need to add `NetCmd` to `cmd/lotus`, otherwise we'll lose `net` commands on the chain process

raulk:
@magik6k `cmd/lotus` uses `lcli.Commands`, which already has `lcli.NetCmd`, that's why it's not a problem. The only thing using `lcli.CommonCommands` is actually `cmd/lotus-miner`. That's why I say that this part of the codebase needs a deeper refactor.

https://github.com/filecoin-project/lotus/blob/5f6166435c5106cf4c3920c0636fa1bc3e02c1ca/cli/cmd.go#L70

cli/cmd.go
raulk: Done.

jennijuju (approved):

jennijuju:
Nice! thanks for this - would be really nice to get in v1.11.1.

magik6k (approved):

PR #7066, opened 2021-08-14
 Correct the config comment for SimultaneousTransfers, where it only applies to retrieval deals

jennijuju:
@hannahhoward has confirmed that Simultaneous transfers in graphsync ONLY affect the responder, and for storage deals the storage provoider is the receiving side so the configuration wont be effective.

### DIFF
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -92,7 +92,7 @@ type DealmakingConfig struct {
 	// as a multiplier of the minimum collateral bound
 	MaxProviderCollateralMultiplier uint64
 
-	// The maximum number of parallel online data transfers (storage+retrieval)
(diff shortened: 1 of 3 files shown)


node/config/types.go
```diff
@@ -307,8 +307,7 @@ type Client struct {
 	IpfsOnlineMode      bool
 	IpfsMAddr           string
 	IpfsUseForRetrieval bool
-	// The maximum number of simultaneous data transfers between the client
-	// and storage providers
+	// The maximum number of simultaneous data transfers between the storage provider and retrieval client
```
raulk, line 310: This would only be for storage deals AFAIK.

jennijuju:
Closing for now - will reopen soon after  figuring out a better and clearer explanation

PR #7079, opened 2021-08-15
clean up car handling, IPFS integration, import manager, retrieve RPC.

raulk:
The goal of this commit was to move the CAR handling logic from go-fil-markets to Lotus via the BlockstoreAccessor abstractions introduced recently.

However, as I began to do that, I realised that the ImportMgr component had ended up quite messy. So I cleaned it up.

I also noticed that we were doing **a lot of** CAR wrangling and copying. Related code was MVP quality and not production-grade. CARs were being placed in temporary locations, and copied several times, imported into blockstores, etc. All of that resulting in overhead that was not acceptable in master.

So I cleaned that up by introducing Blockstore-level abstractions. The system now detects when it's running with IPFS integration, and uses the old strategy of retrieving directly into the IPFS blockstore, instead of using a CARv2 as a buffer.

I also ended up refactoring the `lotus client retrieve` operation. It was quite messy and had a bunch of pathways, which are easier to follow.


node/impl/client/car_helpers.go
```diff
@@ -0,0 +1,91 @@
+package client
```
raulk, line 1: This needs to go and be replaced by go-car support.

PR #7099, opened 2021-08-17
Shed: Create a verifreg command for when VRK isn't a multisig

arajasek:


cmd/lotus-shed/verifreg.go
```diff
@@ -110,6 +111,71 @@ var verifRegAddVerifierCmd = &cli.Command{
...
+			return err
+		}
+
+		allowance, err := types.BigFromString(cctx.Args().Get(2))
+		if err != nil {
+			return err
+		}
+
+		// TODO: ActorUpgrade: Abstract
+		params, err := actors.SerializeParams(&verifreg2.AddVerifierParams{Address: verifier, Allowance: allowance})
```
Stebalien, line 139: Consider requiring a non-id address for the actor, in case of reorgs.

Stebalien (approved):

PR #7108, opened 2021-08-17
Simple alert system; FD limit alerts

magik6k:
This PR adds a new 'Alerting' system, which is built on top of the existing Journal. The basic idea is to have a central place for reporting things which need user action, but aren't necessarily so critical that they would warrant stopping the miner process.

It allows subsystems to define alerts which can be raised or resolved. Alerts can then be viewed by users with `lotus[-miner] log alerts.`

Also adds a basic 'too low file descriptor limit' alert (this has caused me to fail windowpost multiple times (and happens when starting the miner directly from the shell, which I sometimes have to do when trying new things)).

Example `log alerts` output with bad limit, and after fixing the limit, and restarting the node:
2021-08-17-151249_1164x157_scrot

Example `lotus-miner info` when some alerts are active:
2021-08-17-175548_633x164_scrot

TODO:
- [ ] Docstrings
- [ ] Open issues with more alert ideas
- [ ] Tests


lanzafame:
@magik6k This feature would be more useful if there was a way to have the alert information extracted from the machine that the daemon/miner is running on. Most mining operations don't perform monitoring via sshing into a box and running lotus[-miner] commands but rather scrape logs or metrics. If the alerting journal was logged to a file like the standard journal that would be super useful.

magik6k:
> This feature would be more useful if there was a way to have the alert information extracted from the machine that the daemon/miner is running on

This definitely can be done when we have more alert types. Do you mind opening an issue (maybe proposing some specific mechanism, I'm not too familiar with alerting systems)?

> If the alerting journal was logged to a file like the standard journal that would be super useful.

Alerts are logged to the journal

lanzafame:
> This definitely can be done when we have more alert types. Do you mind opening an issue (maybe proposing some specific mechanism, I'm not too familiar with alerting systems)?

Sure thing. 
   
> Alerts are logged to the journal

I tried looking for them there, but I couldn't find them, granted the only alert that I was getting was the 'low fd limit' alert but I saw no mention of it in the journal. I will reconfirm that today though.

lib/ulimit/ulimit.go
```diff
@@ -17,7 +17,7 @@ var (
 	supportsFDManagement = false
 
 	// getlimit returns the soft and hard limits of file descriptors counts
-	getLimit func() (uint64, uint64, error)
+	GetLimit func() (uint64, uint64, error)
```
Stebalien, line 20: I'd consider exposing a _separate_ `GetLimit` function instead of making this variable public.

node/builder.go
```diff
@@ -146,6 +150,9 @@ func defaults() []Option {
 		// global system journal.
 		Override(new(journal.DisabledEvents), journal.EnvDisabledEvents),
 		Override(new(journal.Journal), modules.OpenFilesystemJournal),
+		Override(new(*alerting.Alerting), alerting.NewAlertingSystem),
+
+		Override(CheckFDLimit, modules.CheckFdLimit(16<<10)),
```
Stebalien, line 155: Can we put the limit in a named variable/constant?

node/builder_miner.go
```diff
@@ -74,6 +74,8 @@ func ConfigStorageMiner(c interface{}) Option {
 	return Options(
 		ConfigCommon(&cfg.Common, enableLibp2pNode),
 
+		Override(CheckFDLimit, modules.CheckFdLimit(100_000)), // recommend at least 100k FD limit to miners
```
Stebalien, line 77: Also variable/constant? Not critical, but a nice thing to group. Maybe even drop it in build?

node/modules/alerts.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"github.com/filecoin-project/lotus/journal/alerting"
+	"github.com/filecoin-project/lotus/lib/ulimit"
+)
+
+func CheckFdLimit(min uint64) func(al *alerting.Alerting) {
+	return func(al *alerting.Alerting) {
+		if ulimit.GetLimit == nil {
+			return
```
Stebalien, line 11: No alert in this case?

cli/log.go
```diff
@@ -100,3 +103,51 @@ var LogSetLevel = &cli.Command{
...
 	},
 }
+
+var LogAlerts = &cli.Command{
+	Name:  "alerts",
+	Usage: "Get alert states",
+	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "all",
+			Usage: "get all (active and inactive) alerts",
```
Stebalien, line 113: note that it defaults to active?

journal/alerting/alerts.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	a.alerts[at] = upd(alert, rawMsg)
+}
+
+// Raise marks the alert condition as active and records related event in the journal
+func (a *Alerting) Raise(at AlertType, message interface{}) {
+	log.Errorw("alert raised", "type", at, "message", message)
+
+	a.update(at, message, func(alert Alert, rawMsg json.RawMessage) Alert {
```
Stebalien, line 107: This API is a bit strange because `update` doesn't actually need the message. I'd consider making a separate `marshalMessage` helper. I.e.:

```go
func marshalMessage(msg interface{}) json.RawMessage {
	rawMsg, err := json.Marshal(message)
	if err != nil {
		log.Errorw("marshaling alert message failed", "type", at, "error", err)
		rawMsg, err = json.Marshal(&struct {
			AlertError string
		}{
			AlertError: err.Error(),
		})
		log.Errorw("marshaling marshaling error failed", "type", at, "error", err)
	}
    return rawMsg
}
```

journal/alerting/alerts.go
```diff
@@ -0,0 +1,161 @@
...
+
+// GetAlerts returns all registered (active and inactive) alerts
+func (a *Alerting) GetAlerts() []Alert {
+	a.lk.Lock()
+	defer a.lk.Unlock()
+
+	out := make([]Alert, 0, len(a.alerts))
+	for _, alert := range a.alerts {
+		out = append(out, alert)
+	}
```
Stebalien, line 151:
```suggestion
	out := make([]Alert, len(a.alerts))
	copy(out, a.alerts)
```

Stebalien (approved):
Some nits you can ignore if you just want to merge it.

PR #7112, opened 2021-08-18
journal: current log file does not contain dynamic date

### LINKED ISSUE #7050: Make current journal log file distinguishable from rotated journal log files
### Checklist

### Lotus component

Other

### Improvement Suggestion

I am trying to get Loki to scrape journal logs atm and it is difficult with the current naming of the journal log files as there is no way to distinguish from the current log file and which files have been rotated.

Currently, they all look like the following:
```
lotus-journal-2021-08-04T112641+1000.ndjson
lotus-journal-2021-08-04T164634+1000.ndjson
lotus-journal-2021-08-05T105711+1000.ndjson
lotus-journal-2021-08-05T111201+1000.ndjson
lotus-journal-2021-08-05T113522+1000.ndjson
lotus-journal-2021-08-07T135550+1000.ndjson
lotus-journal-2021-08-10T140251+1000.ndjson
lotus-journal-2021-08-10T163444+1000.ndjson
lotus-journal-2021-08-10T163858+1000.ndjson <-- current file
```

Ideally, they would look like the following:
```
lotus-journal-2021-08-04T112641+1000.ndjson
lotus-journal-2021-08-04T164634+1000.ndjson
lotus-journal-2021-08-05T105711+1000.ndjson
lotus-journal-2021-08-05T111201+1000.ndjson
lotus-journal-2021-08-05T113522+1000.ndjson
lotus-journal-2021-08-07T135550+1000.ndjson
lotus-journal-2021-08-10T140251+1000.ndjson
lotus-journal-2021-08-10T163444+1000.ndjson
lotus-journal.ndjson
```
This way a simple pattern match can be used to capture only the current log file.

lanzafame:
Resolves #7050.

### DIFF
--- a/journal/fs.go
+++ b/journal/fs.go
@@ -109,9 +109,17 @@ func (f *fsJournal) rollJournalFile() error {
 		_ = f.fi.Close()
 	}
 
-	nfi, err := os.Create(filepath.Join(f.dir, fmt.Sprintf("lotus-journal-%s.ndjson", build.Clock.Now().Format(RFC3339nocolon))))
+	// check if journal file exists
+	if _, err := os.Stat(filepath.Join(f.dir, "lotus-journal.json")); err == nil {
+		err := os.Rename(filepath.Join(f.dir, "lotus-journal.json"), filepath.Join(f.dir, fmt.Sprintf("lotus-journal-%s.ndjson", build.Clock.Now().Format(RFC3339nocolon))))
+		if err != nil {
+			return xerrors.Errorf("failed to roll journal file: %w", err)
+		}
+	}
+
+	nfi, err := os.Create(filepath.Join(f.dir, "lotus-journal.json"))
 	if err != nil {
-		return xerrors.Errorf("failed to open journal file: %w", err)
+		return xerrors.Errorf("failed to create journal file: %w", err)
 	}
 
 	f.fi = nfi


journal/fs.go
```diff
@@ -109,9 +109,17 @@ func (f *fsJournal) rollJournalFile() error {
 		_ = f.fi.Close()
 	}
 
-	nfi, err := os.Create(filepath.Join(f.dir, fmt.Sprintf("lotus-journal-%s.ndjson", build.Clock.Now().Format(RFC3339nocolon))))
+	// check if journal file exists
+	if _, err := os.Stat(filepath.Join(f.dir, "lotus-journal.json")); err == nil {
+		err := os.Rename(filepath.Join(f.dir, "lotus-journal.json"), filepath.Join(f.dir, fmt.Sprintf("lotus-journal-%s.ndjson", build.Clock.Now().Format(RFC3339nocolon))))
```
raulk, line 114: nit: that's a long line, can we break it out in local vars?

journal/fs.go
```diff
@@ -109,9 +109,17 @@ func (f *fsJournal) rollJournalFile() error {
 		_ = f.fi.Close()
 	}
 
-	nfi, err := os.Create(filepath.Join(f.dir, fmt.Sprintf("lotus-journal-%s.ndjson", build.Clock.Now().Format(RFC3339nocolon))))
+	// check if journal file exists
+	if _, err := os.Stat(filepath.Join(f.dir, "lotus-journal.json")); err == nil {
```
raulk, line 113: Technically, this could be a directory, but even if it is we want to rename it, so it's fine that we don't use the stat output.

raulk (commented):
So this will lead to slightly different semantics on the naming of journal files. Right now we stamp journals with the timestamp at which they were _opened_. This will stamp them with the date at which they were _closed/rotated_.

I'd honestly prefer to have a date range in there, even if it results in an excessively long filename.

Also to consider is the behaviour on restart -- this will stamp the **existing** journal with the timestamp of the restart.

lanzafame:
@raulk All these comments should be resolved. I also realised that I was using `.json` instead of `.ndjson` for the 'current' log file.

journal/fs.go
```diff
@@ -108,14 +112,30 @@ func (f *fsJournal) rollJournalFile() error {
 	if f.fi != nil {
 		_ = f.fi.Close()
 	}
+	end := build.Clock.Now()
+	current := filepath.Join(f.dir, "lotus-journal.ndjson")
+	rolled := filepath.Join(f.dir, fmt.Sprintf(
+		"lotus-journal-%s-%s.ndjson",
+		f.start.Format(RFC3339nocolon), end.Format(RFC3339nocolon),
```
raulk, line 119: This breaks when we restart, because `f.start` will no longer have the original start timestamp, so the stamping would be inaccurate. We could stamp the file with the start timestamp, and then append the end timestamp on rotation, but that would make the `current` filename dynamic again. I'm OK just accepting the semantic change to stamp the rotated file with the end date.

journal/fs.go
```diff
@@ -108,14 +112,30 @@ func (f *fsJournal) rollJournalFile() error {
...
+	// check if journal file exists
+	if fi, err := os.Stat(current); err == nil && !fi.IsDir() {
+		err := os.Rename(current, rolled)
+		if err != nil {
+			return xerrors.Errorf("failed to roll journal file: %w", err)
+		}
+	}
 
-	nfi, err := os.Create(filepath.Join(f.dir, fmt.Sprintf("lotus-journal-%s.ndjson", build.Clock.Now().Format(RFC3339nocolon))))
+	nfi, err := os.Create(filepath.Join(f.dir, "lotus-journal.json"))
```
raulk, line 130:
```suggestion
	nfi, err := os.Create(filepath.Join(f.dir, "lotus-journal.ndjson"))
```

raulk:
@lanzafame Apologies for back and forth, but I can't imagine a simple way of remembering the original start date on restart. I'm happy to accept the original semantic change where the timestamp is now the end/rotation timestamp. This is also how I believe frameworks like log4j handle rotation.

Also, a bugfix.

Issue #7103, opened 2021-08-17
[BUG] Piece Cid mismatch between deal proposal in Actor state and deal Piece added to Sector (online and offline deals)


aarshkshah1992:
### Checklist

- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] I **am** reporting a bug w.r.t one of the [M1 tags](https://github.com/filecoin-project/lotus/discussions/6852#discussioncomment-1043951). If not, choose another issue option [here](https://github.com/filecoin-project/lotus/issues/new/choose).
- [X] I am **not** reporting a bug around deal making. If yes, create a [M1 Bug Report For Deal Making](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fbug%2CM1-release&template=m1_bug_report_deal.yml).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.

### Lotus component

lotus miner/worker - sealing

### Lotus Tag and  Version

```text
Unknown
```

### Describe the Bug

@stuberman reported an issue where the deal Piece CID stored in the Actor subsystem does not match the PieceCID of the same deal as known by the Sector being sealed.

### Logging Information

```text
021-08-17T01:02:55.149Z	WARN	sectors	storage-sealing/states_sealing.go:194	invalid deals in sector 1507: piece 0 (of 14) of sector 1507 refers deal 2285896 with wrong PieceCID: baga6ea4seaqijqccdoqgqwqbx54vui2eazh6ijf5kku5eq3xwokp6tclivuoqei != baga6ea4seaqlopksik42wyueute5eiyrge6nfqga54o4shsa24va23tz74fcacq
2021-08-17T01:02:55.172Z	WARN	sectors	storage-sealing/states_failed.go:357	piece 0 (of 14) of sector 1507 refers deal 2285896 with wrong PieceCID: baga6ea4seaqijqccdoqgqwqbx54vui2eazh6ijf5kku5eq3xwokp6tclivuoqei != baga6ea4seaqlopksik42wyueute5eiyrge6nfqga54o4shsa24va23tz74fcacq
[...]

PR #7117, opened 2021-08-18
sealing: Fix RecoverDealIDs loop with changed PieceCID

magik6k:
(note: 95% of the diff is testing, the bugfix is the few lines changed at the bottom of `states_failed.go`)

I'm not exactly sure what has to happen to make the FSM put wrong piece into a sector, but this has happened on my miner after it has crashed with OOM while processing a whole bunch of deal, so it's probably a rather unlikely edge-case.

One likely explanation is that a call to `AddPiece` didn't add the correct data (e.g. the reader stream dropped early), which got the wrong PieceCID into `SectorInfo.Pieces[].Piece.PieceCID`. If that's the case (which I pretty strongly suspect it is), we can only do 2 things:
1. Drop the piece from the sector (but that's rather complicated, requires new storage methods, and will require non-trivial shuffling data around in sectors to fix padding if the removed piece was not at the end of the sector)

2. Remove the sector
  * Given that we'll detect this before even attempting PC1 (we call `checkPieces` as the first step), we don't loose much work (only putting deals into the sector)
  * Deals which fail because of being put into that sector should auto-retry adding to a new sector, so effectively we end up doing 1. by reassembling the sector from scratch


neondragon:
"rather unlikely edge-case" -- I have 7 sectors and 63 deals stuck with RecoverDealIDs on f019551.

jennijuju:
id consider to get this in a lotus v1.11.2-rc if possible

raulk (commented):
I'm not super familiar with this part of the codebase. I think @aarshkshah1992 can provide a better review here.

* If what we want to do is remove the sector if we find a piece CID mismatch, then this looks good. 

* Would the deals originally packed into that sector be reassigned a new sector automatically? If so, should we write a test for that?

aarshkshah1992:
@jennijuju @magik6k  This should close https://github.com/filecoin-project/lotus/issues/7103.

aarshkshah1992:
@magik6k 

```
_Deals which fail because of being put into that sector should auto-retry adding to a new sector, 
so effectively we end up doing 1. by reassembling the sector from scratch_
```

I see we are completely dropping the sector with the bad pieces. But, where is the code to do the above i.e. put all the deals in the dropped sector into a new sector ?

raulk:
@aarshkshah1992 

> Deals which fail because of being put into that sector should auto-retry adding to a new sector, so effectively we end up doing 1. by reassembling the sector from scratch

^^ @magik6k do we have a test for that behaviour. Rr can we write one, since that's the most useful outcome that folks are going to expect?

magik6k:
> do we have a test for that behaviour. Rr can we write one, since that's the most useful outcome that folks are going to expect?

No, it's

magik6k:
(missclick on the close thing)

> I see we are completely dropping the sector with the bad pieces. But, where is the code to do the above i.e. put all the deals in the dropped sector into a new sector ?

My assumption is that we do the retrying in markets.

> do we have a test for that behaviour. Rr can we write one, since that's the most useful outcome that folks are going to expect?

No, I'm assuming that markets are going to retry putting deals into a sector when they see that sealing failed (I'm not 100% sure if this is the current behavior, but I'm pretty sure it was at some point).

aarshkshah1992:
@magik6k Discussed this offline but putting it here for Github record.

There is no such code in Markets that detects when a deal is dropped from a sector and re-attempts adding it to another sector. In-fact, I am not even sure if there's a mechanism by which Markets can detect that.

arajasek (approved):
SGWM, but what actually happens to these deals now?

raulk (approved):
I think this is just a first step, but it's not a complete solution as stated in comments above.

IUC, by the point we are here, markets has already handed off the deal to the sealing subsystem, and responsibility is transferred.

AFAIK, there is no notification back to markets that the deal has been excluded from a sector a posteriori, or the sector is deleted altogether (like in this PR).

Even if there were, all that markets could do is call `SectorAddPieceToAny` again.

In the case we're addressing immediately, it seems like the piece transfer was interrupted, and the miner node ended up with a partial piece which, of course, doesn't compute up to the expected PieceCID. 

In other words, the handoff failed but nobody noticed until later. IMO we need to focus on making sure that failures on handoff are detected **immediately**.

The thing that complicates it is the `io.Reader` JSON-RPC encoder: it is incapable of knowing if the transfer was sent in full or not!

Possible solution:

1. allow the caller of the JSON-RPC client to pass special readers with a `Len() int64` method.
2. encoder: type assert the `io.Reader`, if it contains a `Len() int64` method, call it to obtain the length of the transfer and set it in the `Content-Length` header of the POST/PUT.
3. decoder: get the content length, and assert that the `io.Copy` copies that many bytes. If not, declare the transfer as failed, which should fail the original JSON-RPC invocation (thus the client -- markets -- would notice and would retry immediately).

PR #7135, opened 2021-08-19
Fix error handling in SectorAddPieceToAny api impl

magik6k:
A value+error can get routed there from https://github.com/filecoin-project/lotus/blob/master/extern/storage-sealing/input.go#L214, which will make go-jsonrpc hide the real error, and instead it will complain about value+error being returned


aarshkshah1992 (approved):
Looks harmless.

PR #7140, opened 2021-08-19
miner: Command to list/remove expired sectors

magik6k:
We should make this automatic at some point, but this is at least a good tool to manage expired sectors

`--show-removed` output:
2021-08-19-170424_792x730_scrot

The 'actually remove things' output:
2021-08-19-170441_897x1123_scrot

### DIFF
--- a/cmd/lotus-miner/sectors.go
+++ b/cmd/lotus-miner/sectors.go
@@ -44,6 +44,7 @@ var sectorsCmd = &cli.Command{
 		sectorsUpdateCmd,
 		sectorsPledgeCmd,
 		sectorsCheckExpireCmd,
+		sectorsExpiredCmd,
 		sectorsRenewCmd,
 		sectorsExtendCmd,
 		sectorsTerminateCmd,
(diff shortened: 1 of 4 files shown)


Stebalien:
Rebased to try to make CI green.

cmd/lotus-miner/sectors.go
```diff
@@ -1515,6 +1516,192 @@ var sectorsUpdateCmd = &cli.Command{
...
+				return xerrors.Errorf("getting network version: %w", err)
+			}
+
+			lbEpoch = head.Height() - policy.GetWinningPoStSectorSetLookback(nv)
+			if lbEpoch < 0 {
+				return xerrors.Errorf("too early to terminate sectors")
+			}
+		}
+
+		if cctx.IsSet("confirm-remove-count") && !cctx.IsSet("expired-epoch") {
```
arajasek, line 1592: Wanna move this (and any other possible quickchecks) to the top so that you don't iterate all the sectors before failing trivially?

cmd/lotus-miner/sectors.go
```diff
@@ -1515,6 +1516,192 @@ var sectorsUpdateCmd = &cli.Command{
...
+		ctx := lcli.ReqContext(cctx)
+
+		maddr, err := nodeApi.ActorAddress(ctx)
+		if err != nil {
+			return xerrors.Errorf("getting actor address: %w", err)
+		}
+
+		// toCheck is a working bitfield which will only contain terminated sectors
+		toCheck := bitfield.New()
+		{
```
arajasek, line 1563: Just for my education what's the point of having code blocks like this?

arajasek (approved):
Looks like it does the thing

cmd/lotus-miner/sectors.go
magik6k: Generally just for not leaking variables into the wider function context (which makes it easier to see when we are done with those vars)

PR #7147, opened 2021-08-20
Codecov Projects

ZenGround0:
WIP

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,7 +1,62 @@
 coverage:
   status:
     project:
-      default:
+      tools-and-tests:
         target: auto
         threshold: 0.1%
+        informational: true
+        paths:
+          - "testplans"
+          - "tools"
+          - "system"
+          - "snap"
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -1,29 +1,65 @@
...
+        target: auto
+        threshold: 0.1%
+        informational: true        
+        paths: 
+          - "miner"
+          - "storage"
+      chain:
+        target: auto
+        threshold: 0.1%
+        informational: true
```
arajasek, line 35: Probably sufficient to just set this to `false` -- we could subdivide into smaller projects, but I think they'd all still be "required" instead of "informational"

arajasek (approved):
This _should_ do what we want it to do. Whether the changes accurately reflect what happens in a PR is TBD...

BigLep:
For context, this is part of https://github.com/filecoin-project/lotus/issues/7164

PR #7148, opened 2021-08-20
itests: support larger sector sizes; add large deal test.

raulk:
Up until now, itests has been limited by the 2KiB sector size. It has been impractical to use larger sizes because we would go download the proof parameters, even if using mock proofs.

This PR:
1. introduces the ability to specify the sector size as a `Miner` `NodeOpt`.
2. introduces sugar to enable large sectors in the policy (concretely 512MiB sectors).
3. adds a test that performs a 64MiB deal.
4. allows us to specify arbitrary sector sizes on genesis (before, we would ignore the relevant NodeOpt).

## Future

Now that we can perform deals with more interesting sizes, we should look into the following:

- performing deals without writing files on disk (constructing a random DAG in memoroy, for example)
- adjusting the test kit to introduce deal size as a first-class parameter -- right now it's not really adjustable.

### DIFF
--- a/itests/deals_512mb_test.go
+++ b/itests/deals_512mb_test.go
@@ -15,7 +15,7 @@ func TestStorageDealMissingBlock(t *testing.T) {
 	ctx := context.Background()
 
 	// enable 512MiB proofs so we can conduct larger transfers.
-	kit.EnableLargeSectors()
+	kit.EnableLargeSectors(t)
 	kit.QuietMiningLogs()
 
 	client, miner, ens := kit.EnsembleMinimal(t,
--- a/itests/deals_512mb_test.go
+++ b/itests/deals_512mb_test.go
@@ -1,0 +1,44 @@
+package itests
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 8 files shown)


itests/kit/sectors.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/lotus/chain/actors/policy"
+)
+
+// EnableLargeSectors enables 512MiB sectors. This is useful in combination with
+// mock proofs, for testing larger transfers.
+func EnableLargeSectors() {
```
magik6k, line 11: We should probably use `t.Cleanup` here to return the global policy state back to the default.

itests/deals_512mb_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	client, miner, ens := kit.EnsembleMinimal(t,
+		kit.MockProofs(),
+		kit.SectorSize(512<<20), // 512MiB sectors.
+	)
+	ens.InterconnectAll().BeginMining(50 * time.Millisecond)
+
+	dh := kit.NewDealHarness(t, client, miner, miner)
+
+	client.WaitTillChain(ctx, kit.HeightAtLeast(5))
```
Stebalien, line 29: Do we need this? I don't think we need it anymore.

Stebalien (approved):

PR #7150, opened 2021-08-20
api/command for encoding actor params

magik6k:
Example use:
```
$ lotus-shed msg encode-params --to f02 --method 4 '"35496873496"'                                                                                                                                                                                                                                                                                                   
RgAIQ8dOGA==
```

(`encode-params-code` is similar, but useful when the actor of a given type doesn't exist on-chain (e.g. for serializing constructor params))


arajasek:
`lotus chain encode params` does the same thing as the new `encode-params`

arajasek (approved):
lgtm

PR #7173, opened 2021-08-24
fix TestDealPublisher

Stebalien:
This test _should_ be mostly deterministic, but is still failing. I _believe_ this is because:

1. We record the "start" time before starting the timer.
2. In the test, we increment the clock till right after the timer should fire, if the timer start time matches the recorded start time.
3. If we happen to advance the mock clock between recording the time and starting the timer, the timer will never fire.

Unfortunately, I can't reproduce this locally (at all) so I'm mostly just guessing.

### DIFF
--- a/markets/storageadapter/dealpublisher.go
+++ b/markets/storageadapter/dealpublisher.go
@@ -228,11 +228,15 @@ func (p *DealPublisher) waitForMoreDeals() {
 	// Set a timeout to wait for more deals to arrive
 	log.Infof("waiting publish deals queue period of %s before publishing", p.publishPeriod)
 	ctx, cancel := context.WithCancel(p.ctx)
+
+	// Create the timer _before_ taking the current time so publishPeriod+timeout is always >=
+	// the actual timer timeout.
+	timer := build.Clock.Timer(p.publishPeriod)
+
 	p.publishPeriodStart = build.Clock.Now()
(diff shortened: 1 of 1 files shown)


markets/storageadapter/dealpublisher.go
```diff
@@ -384,12 +388,12 @@ func pieceCids(deals []market2.ClientDealProposal) string {
 
 // filter out deals that have been cancelled
 func (p *DealPublisher) filterCancelledDeals() {
-	i := 0
+	filtered := p.pending[:0]
```
Stebalien, line 391: This was just a drive-by fix to make the logic here easier to reason about.

markets/storageadapter/dealpublisher.go
```diff
@@ -257,7 +261,7 @@ func (p *DealPublisher) publishAllDeals() {
 
 	// Filter out any deals that have been cancelled
 	p.filterCancelledDeals()
-	deals := p.pending[:]
+	deals := p.pending
```
Stebalien, line 264: Also a drive-by fix. There's no need to do this unless `p.pending` is an array.

Stebalien:
codecove -> :facepalm:

Stebalien:
Flaky test disabled in https://github.com/filecoin-project/lotus/pull/7176.

arajasek (approved):

Issue #7146, opened 2021-08-20
offline deals: m1.3.5 PC1 error pieces and comm_d do not match


stuberman:
After AddPiece completes, PC1 fails at start with message "pieces and comm_d do not match"

**Worker log**
2021-08-20T14:09:13.541 INFO filcrypto::proofs::api > generate_data_commitment: start
2021-08-20T14:09:13.541 INFO filecoin_proofs::api::seal > compute_comm_d:start
2021-08-20T14:09:13.541 INFO filecoin_proofs::pieces > verifying 8192 pieces
2021-08-20T14:09:13.542 INFO filecoin_proofs::api::seal > compute_comm_d:finish
2021-08-20T14:09:13.542 INFO filcrypto::proofs::api > generate_data_commitment: finish
2021-08-20T14:09:13.622Z	DEBUG	advmgr	sector-storage/worker_local.go:137	acquired sector {{1278 1526} 8} (e:1; a:0): {{0 0} /seal/worker/unsealed/s-t01278-1526  }
2021-08-20T14:09:13.628Z	DEBUG	stores	stores/local.go:117	getting disk usage of '/seal2/worker/sealed/s-t01278-1526': file does not exist
2021-08-20T14:09:13.628Z	DEBUG	advmgr	sector-storage/worker_local.go:137	acquired sector {{1278 1526} 8} (e:1; a:6): {{0 0} /seal/worker/unsealed/s-t01278-1526 /seal2/worker/sealed/s-t01278-1526 /seal2/worker/cache/s-t01278-1526}
2021-08-20T14:09:13.628 INFO filcrypto::proofs::api > seal_pre_commit_phase1: start
2021-08-20T14:09:13.628 INFO filecoin_proofs::api::seal > seal_pre_commit_phase1:start: SectorId(1526)
2021-08-20T14:09:32.732 INFO filecoin_proofs::api::seal > building merkle tree for the original data
2021-08-20T14:10:45.104 INFO filecoin_proofs::api::seal > verifying pieces
2021-08-20T14:10:45.104 INFO filecoin_proofs::pieces > verifying 1 pieces
2021-08-20T14:10:45.104 INFO filcrypto::proofs::api > seal_pre_commit_phase1: finish

**Miner log**
[...]

PR #7175, opened 2021-08-24
Add expected CID param to AddPiece

magik6k:
- [ ] Depends on https://github.com/filecoin-project/specs-storage/pull/16

Should address https://github.com/filecoin-project/lotus/issues/7146 by making it impossible to put corrupted pieces into sectors (note that after one AddPiece call fails, subsequent calls should overwrite the bad data)


documentation/en/api-v0-methods-worker.md
```diff
@@ -167,6 +167,7 @@ Inputs:
   },
   null,
   1024,
+  null,
```
Stebalien, line 170: This shouldn't change. Do we need to stub out a version of the v0 API that passes nil?

Stebalien (approved):
Logic looks good, just making sure this isn't breaking the v0 API.

Stebalien:
Flaky test disabled in #7176.

extern/storage-sealing/states_sealing.go
```diff
@@ -88,7 +89,9 @@ func (m *Sealing) padSector(ctx context.Context, sectorID storage.SectorRef, exi
 
 	out := make([]abi.PieceInfo, len(sizes))
 	for i, size := range sizes {
-		ppi, err := m.sealer.AddPiece(ctx, sectorID, existingPieceSizes, size, NewNullReader(size))
+		expectCid := zerocomm.ZeroPieceCommitment(size)
```
aarshkshah1992, line 92: @magik6k  Why are we calculating the Cid of a "size" here ?

extern/storage-sealing/input.go
```diff
@@ -187,10 +188,13 @@ func (m *Sealing) handleAddPiece(ctx statemachine.Context, sector SectorInfo) er
 		offset += padLength.Unpadded()
 
 		for _, p := range pads {
+			expectCid := zerocomm.ZeroPieceCommitment(p.Unpadded())
```
aarshkshah1992, line 191: Shouldn't this be `deal.deal.DealProposal.PieceCID` ? Why are we self calculating the expected PieceCid here ?

documentation/en/api-v0-methods-worker.md
Stebalien: Yeah... but is there any real reason to? Alternatively, if it's internal, is there any way to make sure it's always in the "unstable" API, and never gets stabilized?

Stebalien: (doesn't really block merge)

Issue #4611, opened 2020-10-27
TestBatchDealInput randomly hangs


magik6k:
https://app.circleci.com/pipelines/github/filecoin-project/lotus/11093/workflows/5716db76-1252-4142-b84a-aae3a0264388/jobs/83821

```
goroutine 48560 [chan receive, 29 minutes]:
github.com/filecoin-project/lotus/api/test.testRetrieval(0xc0011d2c60, 0x34b96c0, 0xc00019e010, 0x3501ac0, 0xc001d32600, 0xc004bcc480, 0x26, 0xc001631870, 0x3501a00, 0xc0059ec000, ...)
	/home/circleci/project/api/test/deals.go:410 +0x680
github.com/filecoin-project/lotus/api/test.MakeDeal(0xc0011d2c60, 0x34b96c0, 0xc00019e010, 0x7, 0x3501ac0, 0xc001d32600, 0x34fd740, 0xc003dcc1a0, 0x0, 0x0, ...)
	/home/circleci/project/api/test/deals.go:127 +0x441
github.com/filecoin-project/lotus/api/test.TestDoubleDealFlow(0xc0011d2c60, 0x31c6310, 0x989680)
	/home/circleci/project/api/test/deals.go:101 +0x488
github.com/filecoin-project/lotus/node_test.TestAPIDealFlow.func3(0xc0011d2c60)
	/home/circleci/project/node/node_test.go:48 +0x40
testing.tRunner(0xc0011d2c60, 0x31c6330)
	/usr/local/go/src/testing/testing.go:1039 +0xdc
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1090 +0x372
```

magik6k:
Looking at logs, last retrieval log is
```
2020-10-27T11:22:13.855Z	�[34mINFO�[0m	markets	loggers/loggers.go:24	retrieval event	{"name": "ClientEventLaneAllocated", "deal ID": "1", "state": "DealStatusOngoing", "message": ""}
```

dirkmc:
More examples:
https://app.circleci.com/pipelines/github/filecoin-project/lotus/11124/workflows/09f81bd1-2462-4395-b1a5-48256e66d2c2/jobs/84286
https://app.circleci.com/pipelines/github/filecoin-project/lotus/11124/workflows/09f81bd1-2462-4395-b1a5-48256e66d2c2/jobs/84296

dirkmc:
[...]

PR #7176, opened 2021-08-24
test: disable flaky TestBatchDealInput

Stebalien:
See #4611.


Stebalien:
Now we see just how much code coverage we lose.

arajasek (approved):

PR #7169, opened 2021-08-24
Ignore nil throttler

aarshkshah1992:
Closes #7048 .

Issue #7138, opened 2021-08-19
add tests for `Dagstore*` JSON-RPC operations.


raulk:
These operations are tested in https://github.com/filecoin-project/dagstore and have been thouroughly tested by minerX.2 miners during M1. But we should add coverage in Lotus for the RPCs concretely.

jennijuju:
@raulk could you please help define the scope so that we can have bloxico to work on this potentially?

rjan90:
Closing this issue as outdated on the lotus-miner side of things. This issue is related to the legacy markets code in the Lotus-Miner which will be deprecated.

PR #7177, opened 2021-08-25
dagstore/InitializeAll: Fix throttling bug for concurrency=0

aarshkshah1992:
Address https://github.com/filecoin-project/lotus/pull/7169#discussion_r695597163.


raulk (approved):
This is better. We need tests for these operations! https://github.com/filecoin-project/lotus/issues/7138

PR #5290, opened 2021-01-05
add produced blocks info

deaswang:
fix #314 add produced blocks info.

PR #7202, opened 2021-08-27
Add optional mined block list to miner info

magik6k:
This is https://github.com/filecoin-project/lotus/pull/5290 with conflicts resolved, and with some tweaks (mostly added a  progress indicator which is useful for miners who only mine a block or two per month, where this gets a little slow (2-3 min per month))


Shekelme:
It would be nice if this command also displayed the date of finding for each block in it's output.

PR #7210, opened 2021-08-27
fix: correctly handle null blocks when detecting an expensive fork

### LINKED ISSUE #7192: TestForkRefuseCall has a flake
https://app.circleci.com/pipelines/github/filecoin-project/lotus/16950/workflows/d2cf84ff-8a53-4b26-a42f-0fe68d513192/jobs/251127

```
2021-08-26T20:03:18.025Z	INFO	chainstore	store/store.go:628	New heaviest tipset! [bafy2bzacecmsdath2uqfavzhwiucjoxpw4gfqg3bv2i2husu7ovpfkeoiryps] (height=42)
2021-08-26T20:03:18.027Z	WARN	statemgr	stmgr/forks.go:187	STARTING migration	{"height": "40", "from": "bafy2bzacecvgncx7qudezlqojvirxsykyz3hcq5kdstqfewvncw3meilk4q3i"}
2021-08-26T20:03:18.027Z	WARN	statemgr	stmgr/forks.go:200	COMPLETED migration	{"height": "40", "from": "bafy2bzacecvgncx7qudezlqojvirxsykyz3hcq5kdstqfewvncw3meilk4q3i", "to": "bafy2bzacecvgncx7qudezlqojvirxsykyz3hcq5kdstqfewvncw3meilk4q3i", "duration": 0.000044774}
    forks_test.go:310: 
        	Error Trace:	forks_test.go:310
        	Error:      	Not equal: 
        	            	expected: *errors.errorString(&errors.errorString{s:"refusing explicit call due to state fork at epoch"})
        	            	actual  : <nil>(<nil>)
        	Test:       	TestForkRefuseCall
--- FAIL: TestForkRefuseCall (0.95s)
```

Like last time, this is probably due to null blocks. But I've spent hours trying to make this test impervious to null blocks and have no idea why it's failing now. Running it 1000 times doesn't seem to change anything.

Stebalien:
Also improve/fix documentation to reflect the _actual_ tipset that's passed into upgrades.

And update some comments.

fixes #7192

### DIFF
--- a/chain/stmgr/call.go
+++ b/chain/stmgr/call.go
@@ -49,7 +49,7 @@ func (sm *StateManager) Call(ctx context.Context, msg *types.Message, ts *types.
 			}
 			ts = pts
 		}
-	} else {
+	} else if ts.Height() > 0 {
 		pts, err := sm.cs.LoadTipSet(ts.Parents())
 		if err != nil {
 			return nil, xerrors.Errorf("failed to load parent tipset: %w", err)
@@ -58,6 +58,9 @@ func (sm *StateManager) Call(ctx context.Context, msg *types.Message, ts *types.
 		if sm.hasExpensiveFork(pheight) {
 			return nil, ErrExpensiveFork
 		}
+	} else {
+		// We can't get the parent tipset in this case.
+		pheight = ts.Height() - 1
 	}
 
 	bstate := ts.ParentState()
--- a/chain/stmgr/call.go
+++ b/chain/stmgr/call.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/crypto"
 	"github.com/ipfs/go-cid"
 	"go.opencensus.io/trace"
@@ -20,41 +21,49 @@ import (
 
 var ErrExpensiveFork = errors.New("refusing explicit call due to state fork at epoch")
 
+// Call applies the given message to the given tipset's parent state, at the epoch following the
+// tipset's parent. In the presence of null blocks, the height at which the message is invoked may
+// be less than the specified tipset.
+//
+// - If no tipset is specified, the first tipset without an expensive migration is used.
+// - If executing a message at a given tipset would trigger an expensive migration, the call will
+//   fail with ErrExpensiveFork.
 func (sm *StateManager) Call(ctx context.Context, msg *types.Message, ts *types.TipSet) (*api.InvocResult, error) {
 	ctx, span := trace.StartSpan(ctx, "statemanager.Call")
 	defer span.End()
 
+	var pheight abi.ChainEpoch = -1
+
 	// If no tipset is provided, try to find one without a fork.
 	if ts == nil {
 		ts = sm.cs.GetHeaviestTipSet()
-
 		// Search back till we find a height with no fork, or we reach the beginning.
-		for ts.Height() > 0 && sm.hasExpensiveFork(ctx, ts.Height()-1) {
-			var err error
-			ts, err = sm.cs.GetTipSetFromKey(ts.Parents())
+		for ts.Height() > 0 {
+			pts, err := sm.cs.GetTipSetFromKey(ts.Parents())
 			if err != nil {
 				return nil, xerrors.Errorf("failed to find a non-forking epoch: %w", err)
 			}
+			if !sm.hasExpensiveFork(pts.Height()) {
+				pheight = pts.Height()
+				break
+			}
+			ts = pts
+		}
+	} else {
+		pts, err := sm.cs.LoadTipSet(ts.Parents())
+		if err != nil {
+			return nil, xerrors.Errorf("failed to load parent tipset: %w", err)
+		}
+		pheight = pts.Height()
+		if sm.hasExpensiveFork(pheight) {
+			return nil, ErrExpensiveFork
 		}
 	}
 
 	bstate := ts.ParentState()
-	pts, err := sm.cs.LoadTipSet(ts.Parents())
-	if err != nil {
-		return nil, xerrors.Errorf("failed to load parent tipset: %w", err)
-	}
-	pheight := pts.Height()
-
-	// If we have to run an expensive migration, and we're not at genesis,
-	// return an error because the migration will take too long.
-	//
-	// We allow this at height 0 for at-genesis migrations (for testing).
-	if pheight > 0 && sm.hasExpensiveFork(ctx, pheight) {
-		return nil, ErrExpensiveFork
-	}
 
 	// Run the (not expensive) migration.
-	bstate, err = sm.handleStateForks(ctx, bstate, pheight, nil, ts)
+	bstate, err := sm.handleStateForks(ctx, bstate, pheight, nil, ts)
 	if err != nil {
 		return nil, fmt.Errorf("failed to handle fork: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


chain/stmgr/forks.go
```diff
@@ -208,7 +211,21 @@ func (sm *StateManager) handleStateForks(ctx context.Context, root cid.Cid, heig
 	return retCid, nil
 }
 
-func (sm *StateManager) hasExpensiveFork(ctx context.Context, height abi.ChainEpoch) bool {
+// Returns true if executing the current tipset would trigger an expensive fork.
+//
+// - If the tipset is the genesis, this function always returns false.
+// - If inclusive is true, this function will also return true if applying a message on-top-of the
```
ZenGround0, line 217: This comment looks out of date with the current solution, `inclusive` is no longer a parameter and there are now two functions which should probably each have a comment.

chain/stmgr/forks.go
```diff
@@ -208,7 +211,21 @@ func (sm *StateManager) handleStateForks(ctx context.Context, root cid.Cid, heig
...
+//   tipset would trigger a fork.
+func (sm *StateManager) hasExpensiveForkBetween(parent, height abi.ChainEpoch) bool {
+	for h := parent; h < height; h++ {
+		if _, ok := sm.expensiveUpgrades[h]; ok {
```
ZenGround0, line 221: This probably doesn't show up unless you force it but I believe the number of epochs between parent and height can be made arbitrarily big from the cli and it would be nice for this not to hang by taking advantage of sparseness of upgrade heights in the `expensiveUprages` map.

chain/stmgr/execute.go
```diff
@@ -99,7 +99,6 @@ func (sm *StateManager) ApplyBlocks(ctx context.Context, parentEpoch abi.ChainEp
 		}
```
ZenGround0, line 99: nit: not at all your bug but reading the above code it looks like line 82's error message is wrong and confusing and makes sense to fix now.

chain/stmgr/forks.go
```diff
@@ -41,8 +41,11 @@ type MigrationCache interface {
 // - The oldState is the state produced by the upgrade epoch.
 // - The returned newState is the new state that will be used by the next epoch.
 // - The height is the upgrade epoch height (already executed).
-// - The tipset is the tipset for the last non-null block before the upgrade. Do
-//   not assume that ts.Height() is the upgrade height.
+// - The tipset is the first non-null tipset after the upgrade height (the tipset in
+//   which the upgrade is executed). Do not assume that ts.Height() is the upgrade height.
+//
+// NOTE: In StateCompute and CallWithGas, the passed tipset is actually the tipset _before_ the
```
ZenGround0, line 47: Do the historical upgrades all work if we consistently pass in the tipset before the upgrade? It looks like a lot of the recent ones don't use this param at all. We should file an issue for checking this and if so refactoring the upgrade functions to take in this tipset instead so that Call and ComputeState never have access to weird state.

chain/stmgr/forks.go
ZenGround0: Oh ok I see now that the only uses of this are constrained so that the number of iterations is bounded by number of null blocks.  There's potential for misuse from future code maybe (though parameter names help) but no hanging possible.

chain/stmgr/call.go
```diff
@@ -140,25 +149,33 @@ func (sm *StateManager) CallWithGas(ctx context.Context, msg *types.Message, pri
 		// run the fork logic in `sm.TipSetState`. We need the _current_
```
ZenGround0, line 149: Is there a good reason why `Call` happens against tipset parent state and `CallWithGas` happens against tipset state?  As far as I understand is what requires the introduction of `hasExpensiveForkBetween`.  Beyond removing extra work it would also just be nice to standardize these two similar calls

chain/stmgr/forks.go
```diff
@@ -208,7 +211,21 @@ func (sm *StateManager) handleStateForks(ctx context.Context, root cid.Cid, heig
 	return retCid, nil
 }
 
-func (sm *StateManager) hasExpensiveFork(ctx context.Context, height abi.ChainEpoch) bool {
+// Returns true if executing the current tipset would trigger an expensive fork.
+//
+// - If the tipset is the genesis, this function always returns false.
```
ZenGround0, line 216: As written now this invariant depends on the sm.expensiveUpgrades table which in reasonable scenarios could potentially contain "expensive" upgrades if say we did all migrations at the genesis epoch. 

It might be better to override the table to always return false on 0 to maintain this invariant.  Any network doing this would probably be doing very little work in the expensive migration since nothing has happened in state yet.

Also all existing callers seem to implicitly assume that this is the case anyway by only checking heights > 0 so why not make it official.

chain/stmgr/forks.go
Stebalien: Ah, yes. Let me fix that.

chain/stmgr/forks.go
Stebalien: I can have a switch if the number of epochs exceeds the number of upgrades.

chain/stmgr/forks_test.go
```diff
@@ -292,14 +307,20 @@ func TestForkRefuseCall(t *testing.T) {
 		GasFeeCap:  types.NewInt(0),
 	}
 
+	nullStart := abi.ChainEpoch(testForkHeight - nullsBefore)
+	nullLength := abi.ChainEpoch(nullsBefore + nullsAfter)
+
 	for i := 0; i < 50; i++ {
```
ZenGround0, line 313: Not your bug: I don't love that testForkHeight is a global and that this test relies on knowing that its value is 40 in order to actually test anything.  Maybe make `testForkHeight` not a global, or change 50 to something like `2 * testForkHeight`?

ZenGround0 (approved):

chain/stmgr/forks.go
Stebalien: Hm. Actually, I'd rather not. It introduces an additional code path that won't be well tested.

Stebalien: (I could write tests, but I'd rather stick with one code path)

chain/stmgr/execute.go
Stebalien: I'd rather not touch that for now. I don't think that error is consensus critical, but I'd rather not touch it.

chain/stmgr/forks.go
Stebalien: As far as I can tell, it's used for lookbacks. But it may also be used for randomness in some cases?

Stebalien: https://github.com/filecoin-project/lotus/issues/7226

chain/stmgr/call.go
Stebalien: CallWithGas applies the message on-top-of the tipset to provide better gas estimation, I think. By executing it on-top-of the last tipset, instead of on-top-of the parent tipset state, we get the best gas estimates.

I'd prefer to always use the parent state, but I'm not sure what might break. I'm especially concerned that, e.g., it might (rarely) cause issues with window post:

1. See we're at height X.
2. Try to push the proof message.
3. Fail because the window isn't open yet.

chain/stmgr/forks.go
Stebalien: Because the function is called `hasExpensiveFork`, not `hasExpensiveForkOrIsGenesis` :smile:. Basically, I don't want to surprise users.

At the moment, genesis matters to all callers because there's really no other option. But I'd like to keep that as an explicit decision.

chain/stmgr/forks_test.go
Stebalien: Well, being a global is intentional. But yeah, 50 being hard-coded is a problem.

chain/stmgr/forks.go
Stebalien: But I've fixed the issue in Call where I wasn't handling this consistently.

magik6k:
Lint fails
```
chain/stmgr/forks.go:215:32: `occuring` is a misspelling of `occurring` (misspell)
// migration. NOTE: migrations occuring _at_ the target height are not included, as they're executed
```

PR #7211, opened 2021-08-27
test: re-enable disabled tests

Stebalien:
1. Rewrite VRF test to actually test win counts, and enable it.
2. Stop skipping some tests that now pass.

### DIFF
--- a/chain/vectors/vectors_test.go
+++ b/chain/vectors/vectors_test.go
@@ -26,7 +26,6 @@ func LoadVector(t *testing.T, f string, out interface{}) {
 }
 
 func TestBlockHeaderVectors(t *testing.T) {
-	t.Skip("we need to regenerate for beacon")
 	var headers []HeaderVector
 	LoadVector(t, "block_headers.json", &headers)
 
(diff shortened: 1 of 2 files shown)


magik6k:
Pretty sure `TestWinCounts` was written to just output to an external file for analyzing manually, cc @Kubuxu

chain/vectors/vectors_test.go
```diff
@@ -26,7 +26,6 @@ func LoadVector(t *testing.T, f string, out interface{}) {
 }
 
 func TestBlockHeaderVectors(t *testing.T) {
-	t.Skip("we need to regenerate for beacon")
```
ZenGround0, line 29: Have we done the vector regeneration described here?  
Do we even run vector tests any more?

chain/vectors/vectors_test.go
```diff
@@ -65,8 +64,6 @@ func TestMessageSigningVectors(t *testing.T) {
 }
 
 func TestUnsignedMessageVectors(t *testing.T) {
```
ZenGround0, line 66: Have the vectors been fixed?

chain/vectors/vectors_test.go
Stebalien: The tests pass.

chain/vectors/vectors_test.go
Stebalien: The tests pass.

Stebalien: Although I'm not seeing any updates.

Stebalien:
> Pretty sure TestWinCounts was written to just output to an external file for analyzing manually, cc @Kubuxu

They were but it should be automatic now.

ZenGround0 (approved):

PR #7235, opened 2021-08-31
RFC: Add lotus-miner is-stop-safe to check if stopping miner is safe

elliot-u410:
This is an attempt at automating the checks listed in https://docs.filecoin.io/mine/lotus/miner-lifecycle/#safely-restarting-the-miner-daemon

I still need to do testing and polish this work, but wanted to request feedback in parallel.

### DIFF
--- a/cli/util.go
+++ b/cli/util.go
@@ -44,14 +44,19 @@ func parseTipSet(ctx context.Context, api v0api.FullNode, vals []string) (*types
 	return types.NewTipSet(headers)
 }
 
+// Returns negative duration when second argument is in the past.
+func DurationBetweenEpochs(now, future abi.ChainEpoch) time.Duration {
+	return time.Second * time.Duration(int64(build.BlockDelaySecs)*int64(future-now))
+}
+
 func EpochTime(curr, e abi.ChainEpoch) string {
 	switch {
 	case curr > e:
-		return fmt.Sprintf("%d (%s ago)", e, durafmt.Parse(time.Second*time.Duration(int64(build.BlockDelaySecs)*int64(curr-e))).LimitFirstN(2))
+		return fmt.Sprintf("%d (%s ago)", e, durafmt.Parse(DurationBetweenEpochs(e, curr)).LimitFirstN(2))
 	case curr == e:
 		return fmt.Sprintf("%d (now)", e)
 	case curr < e:
-		return fmt.Sprintf("%d (in %s)", e, durafmt.Parse(time.Second*time.Duration(int64(build.BlockDelaySecs)*int64(e-curr))).LimitFirstN(2))
+		return fmt.Sprintf("%d (in %s)", e, durafmt.Parse(DurationBetweenEpochs(curr, e)).LimitFirstN(2))
 	}
 
 	panic("math broke")
--- a/cmd/lotus-miner/stop.go
+++ b/cmd/lotus-miner/stop.go
@@ -1,7 +1,10 @@
 package main
 
 import (
+	"encoding/json"
+	"fmt"
 	_ "net/http/pprof"
+	"os"
 
 	"github.com/urfave/cli/v2"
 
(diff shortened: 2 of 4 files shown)


noch3:
Any recommendations on how to resolve Circle CI failures?

jennijuju:
> Any recommendations on how to resolve Circle CI failures?

`make gen` should do

Stebalien:
You may need to run `make gen`. Also make sure you're up-to-date with master.

noch3:
@jennijuju @Stebalien Thanks for the tip! I ran that and `make docsgen-cli`

noch3:
@jennijuju @Stebalien  I'll hold off on adding test coverage until I get some confirmation that the general approach is sound.

cmd/lotus-miner/is_stop_safe.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+type DeadlineFaultCutoff struct {
+	CurrentEpoch         abi.ChainEpoch `json:"current-epoch"`
+	CutoffEpoch          abi.ChainEpoch `json:"cutoff-epoch"`
+	EstimatedSecondsLeft float32        `json:"estimated-seconds-left"`
+}
+
+var isStopSafeCmd = &cli.Command{
+	Name:  "is-stop-safe",
```
magik6k, line 35: I'd make this be 2 sub-commands of the `stop` command:
* `lotus-miner stop check` which just tells you if it's ok to stop
* `lotus-miner stop safe` which stops when it's safe

cmd/lotus-miner/is_stop_safe.go
```diff
@@ -0,0 +1,318 @@
...
+var SafeStorageDealStates = map[storagemarket.StorageDealStatus]struct{}{
+	storagemarket.StorageDealUnknown:          {},
+	storagemarket.StorageDealProposalNotFound: {},
+	storagemarket.StorageDealProposalRejected: {},
+	// storagemarket.StorageDealProposalAccepted:             {},
+	// storagemarket.StorageDealAcceptWait:                   {},
+	// storagemarket.StorageDealStartDataTransfer:            {},
+	// storagemarket.StorageDealStaged:                       {},
+	// storagemarket.StorageDealAwaitingPreCommit:            {},
+	// storagemarket.StorageDealSealing:                      {},
```
magik6k, line 237: It should be ok to stop when deals are in the sealing pipeline, sealing should be able to restart cleanly

magik6k (commented):
Thanks for the PR!

I like this idea, but would put the command in a slightly different place to not add more commands to the top-level `lotus-miner` command.

cmd/lotus-miner/is_stop_safe.go
noch3: I didn't like the command placement either. Thanks for the direction!

Should `lotus-miner stop safe` subscribe to/poll state changes until it's safe and then stop? If so, then maybe `lotus-miner stop check --wait` should also exist? Or did you have in mind that it would fail immediately if it's not safe.

Either way, it could be argued that `lotus-miner stop` should always perform checks by default and maybe `lotus-miner stop --force` should be the current behavior. This is, however, a breaking UI change.

magik6k: > subscribe to/poll state changes until it's safe and then stop?

Yes, it can just poll changes

> maybe lotus-miner stop check --wait should also exist?

Good idea, and should be easy to add when we have `stop safe` (also, `safe` can just be a `--safe` flag if we go with `--force` bolew)

> it could be argued that lotus-miner stop should always perform checks by default and maybe lotus-miner stop --force should be the current behavior. This is, however, a breaking UI change.

I thought about that, but would prefer to avoid the breaking change for now. Maybe we could:
* Add a warning when normal `stop` is called saying something like 'performing unsafe node shutdown. (NOTE: in near future safe shutdown will become the default. Specify --force to keep current behavior)'
* Not print the warning if `--force` is specified
* At some point in te future (in one or two versions) change the default behavior to safe

noch3: @magik6k Thinking about this more:

I'm not sure a `wait` feature is actually useful. If you're stopping a miner because you want to take it down permanently then you likely don't care how safe it is. If you're stopping a miner because you plan to upgrade/fix/move it, then you aren't going to want it to stop at an unknown point in the future. You'd want to have planned down time where you are available to do the transition at the most opportune time (hence the check command gives you the time to the next deadline fault cutoff).

What's more, if all we are doing is polling, then it's trivial to re-implement that manually:
```bash
while ! lotus-miner stop --check; do sleep 1; done; lotus-miner stop
```

As such I'd propose we simply move this check into `lotus-miner stop --check` and add a `lotus-miner stop --force` command that does the current behavior. Then `lotus-miner stop` would also do the current behavior with a deprecation warning. In the future `lotus-miner stop` will simply fail if the checks fail. It won't wait.

Thoughts?

noch3:
How can I improve test coverage of this MR?

rjan90:
Hey! 👋

rjan90:

Closing

PR #7236, opened 2021-08-31
sectors expired: Handle precomitted and unproven sectors correctly

magik6k:
Without this we'd say that sectors which were not PoSt-ed yet were already expired.

### DIFF
--- a/chain/actors/builtin/miner/state.go.template
+++ b/chain/actors/builtin/miner/state.go.template
@@ -549,6 +549,10 @@ func (p *partition{{.v}}) RecoveringSectors() (bitfield.BitField, error) {
 	return p.Partition.Recoveries, nil
 }
 
+func (p *partition{{.v}}) UnprovenSectors() (bitfield.BitField, error) {
+	return {{if (ge .v 2)}}p.Partition.Unproven{{else}}bitfield.New(){{end}}, nil
+}
+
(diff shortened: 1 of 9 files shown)


chain/actors/builtin/miner/state.go.template
```diff
@@ -549,6 +549,10 @@ func (p *partition{{.v}}) RecoveringSectors() (bitfield.BitField, error) {
 	return p.Partition.Recoveries, nil
 }
 
+func (p *partition{{.v}}) UnprovenSectors() (bitfield.BitField, error) {
+	return {{if (ge .v 2)}}p.Partition.Unproven{{else}}bitfield.New(){{end}}, nil
```
ZenGround0, line 553: The fact that you can specify `if (ge .v 2)` highlights how nice this autogen template format is 👌

cmd/lotus-miner/sectors.go
```diff
@@ -1628,12 +1628,31 @@ var sectorsExpiredCmd = &cli.Command{
 				}
 
 				toCheck, err = bitfield.SubtractBitField(toCheck, live)
```
ZenGround0, line 1630: super nit: if we change the generic name `toCheck` to something more descriptive i.e. `deadSectors` it will make this slightly more readable.

ZenGround0 (approved):
LGTM

PR #7261, opened 2021-09-02
Show deal sizes is sealing sectors

magik6k:
2021-09-02-214304_983x126_scrot

(review commit-by-commit)

PR #7263, opened 2021-09-02
Add --unproven flag to the sectors list command

magik6k:
Depends on https://github.com/filecoin-project/lotus/pull/7261

This PR adds an `--unproven` flag to `lotus-miner sectors list`, which makes it print only sectors which are being sealed. It also is a lot faster than printing all sectors.

### DIFF
--- a/cmd/lotus-miner/sectors.go
+++ b/cmd/lotus-miner/sectors.go
@@ -266,8 +266,9 @@ var sectorsListCmd = &cli.Command{
 	Usage: "List sectors",
 	Flags: []cli.Flag{
 		&cli.BoolFlag{
-			Name:  "show-removed",
-			Usage: "show removed sectors",
+			Name:    "show-removed",
+			Usage:   "show removed sectors",
+			Aliases: []string{"r"},
 		},
 		&cli.BoolFlag{
 			Name:        "color",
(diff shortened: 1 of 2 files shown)


cmd/lotus-miner/sectors.go
```diff
@@ -314,17 +322,33 @@ var sectorsListCmd = &cli.Command{
...
+			states = make([]api.SectorState, len(sList))
 			for i := range sList {
-				ss[i] = api.SectorState(sList[i])
+				states[i] = api.SectorState(sList[i])
+			}
+		}
+
+		if cctx.Bool("unproven") {
+			for state := range sealing.ExistSectorStateList {
+				if state == sealing.Proving {
```
ZenGround0, line 341: One minor point of confusion that I think is possible with this code and terminology is that there is a brief window where the sector is marked in onchain data as "unproven" but the sealing FSM sees the sector as "Proving".  If I understand the FSM correctly the sector transitions to FinalizeSector and then Proving after successful ProveCommit message without waiting for any window posts but miner state keeps the sector in "unproven" state until first post.

If there is another name you could give this flag (unfinalized, uncommitted, unporeped?) it would prevent confusion with the on chain concept.

ZenGround0 (approved):

PR #7269, opened 2021-09-03
Scheduler Enhancements

clinta:
This PR adds valuable functionality to the scheduler. It makes the worker aware of cgroup memory limits to prevent overloading of a worker that has been memory limited by cgroups. It changes the GPU resource from a bool to a float so that tasks can be represented as requiring multiple, or a portion of a GPU. And it permits workers to override the resource table with environment variables.

Tests have not yet been added, but this is being opened as a draft to get feedback.

### DIFF
--- a/extern/sector-storage/resources.go
+++ b/extern/sector-storage/resources.go
@@ -1,9 +1,12 @@
 package sectorstorage
 
 import (
+	"strconv"
+
 	"github.com/filecoin-project/go-state-types/abi"
 
 	"github.com/filecoin-project/lotus/extern/sector-storage/sealtasks"
+	"github.com/filecoin-project/lotus/extern/sector-storage/storiface"
 )
 
 type Resources struct {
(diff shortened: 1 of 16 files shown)


magik6k (commented):
Very quick first read through the diff - I like this, though we should probably try to use the config instead of env vars (though we can also change this after the initial prototype lands).

The ffi submodule update may be committed by accident

clinta:
Thanks for the feedback. I've fixed the accidental ffi commit.

As for config file vs environment. Personally I find environment files easier to deal with, especially when running nodes from systemd and using templated units. If we add the ability to get these from the config file, I'd like to preserve the ability to get them from the environment as well. If we do both I have no strong opinions about which should have priority.

clinta:
Another thought about having this in a config file.

Currently there is an expectation that a user can change resources.go, and compile just lotus-miner, and all workers will get their new defaults from what is in the compiled miner binary. This PR as written does not break this expectation. If a user does not specify environment variables on their worker, this behavior will continue.

If these settings are available in a worker config file, it would take care to preserve this expectation, as the default config in the compiled worker might override the defaults from the miner.

jennijuju:
Id agree config settings are preferred than env var settings for sealing scheduler  and ideally we keep everything in the same section https://docs.filecoin.io/mine/lotus/miner-configuration/#sealing-section

extern/sector-storage/resources.go
```diff
@@ -268,7 +316,7 @@ var ResourceTable = map[sealtasks.TaskType]map[abi.RegisteredSealProof]Resources
 			MinMemory: 1 << 20,
 
 			MaxParallelism: 0,
-			CanGPU:         false,
+			GPUUtilization: 1.0,
```
magik6k, line 319: This should be 0 / not present (and a few below as well)

extern/sector-storage/worker_local.go
```diff
@@ -482,41 +488,164 @@ func (l *LocalWorker) Paths(ctx context.Context) ([]stores.StoragePath, error) {
...
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		fields := bytes.Fields(scanner.Bytes())
+		if len(fields) >= 9 && bytes.Equal(fields[8], []byte("cgroup2")) {
+			return string(fields[4]), nil
+		}
+	}
+	return "", nil
+}
```
magik6k, line 521: We already use `github.com/containerd/cgroups` for getting memory info in go-watchdog, so it would be nice to use this here as well.

See https://github.com/raulk/go-watchdog/blob/master/watchdog_linux.go#L29-L46

Should also be disabled on non-linux builds (though I see we're ignoring the error below and falling back to host info, so that's maybe fine.)

extern/sector-storage/worker_local.go
```diff
@@ -482,41 +488,164 @@ func (l *LocalWorker) Paths(ctx context.Context) ([]stores.StoragePath, error) {
...
 
-	memSwap := mem.VirtualTotal
-	if l.noSwap {
-		memSwap = 0
+	resourceOpts := make(map[string]string)
+	for tt := range l.acceptTasks {
+		ttShort := tt.Short()
+		for _, res_opt := range []string{"_MAX_MEMORY", "_MIN_MEMORY", "_MAX_PARALLELISM", "_BASE_MIN_MEMORY", "_GPU_UTILIZATION"} {
+			n := ttShort + res_opt
+			if val := os.Getenv(n); val != "" {
```
magik6k, line 620: Nit: Could use `os.LookupEnv(key string) (string, bool)`

magik6k (commented):
Thanks for the PR! Changes look good, I missed the part that we pull these settings from workers, which have no config right now (and most worker settings are passed with env vars anyways), so that part is ok.

We also need to bump the minor API versions of Worker/Miner API in `api/version.go` since those changes are breaking.

Multi-GPU support is going to be great, especially with big improvements coming to gpu scheduling in rust-proofs soon.

extern/sector-storage/resources.go
clinta: Got overly aggressive in replacing lines with `CanGPU`. Fixed.

extern/sector-storage/worker_local.go
clinta: I've changed around this logic in the latest commit. My method only supported cgroups v2, now it supports v1 and v2. Unfortunately containerd/cgroups v2 support is missing a few features, like getting the mount point so I had to still implement that myself.

clinta:
Experimenting with this has revealed another issue. Now that the scheduler is respecting my memory limits in cgroups, the scheduler failed to assign tasks that could run. The idea of reporting "reserved memory" has a problem because so called "reserved memory" includes the memory used by running tasks.

To correct this I've changed the worker API to individually report the system used memory and used swap. Then the scheduler, which has knowledge of running tasks and their memory requirements can compare the system's used memory with the memory used by running tasks and operate accordingly.

I've rebased to make this the second commit in the PR.

magik6k (commented):
Re-reviewed the changes, everything looks reasonable. Any reason to keep this in draft other than tests?

clinta:
I've been running this for about a week now and am pretty happy with it. So
yes, from my view the only thing remaining is adding tests.

jennijuju:
@clinta Thank you for submitting the PR! and we would love to get this tested out with MinerX fellows (lotus early testers). Im wondering if we can get a breif user guide on the evn var usages and how to set them?

extern/sector-storage/worker_local.go
```diff
@@ -1,16 +1,23 @@
...
+	"path/filepath"
 	"reflect"
 	"runtime"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
 
+	"github.com/containerd/cgroups"
+	cgroupv2 "github.com/containerd/cgroups/v2"
```
magik6k, line 20: We'll need to put cgroup stuff behind some build flag so that OSX build still works

magik6k:
Hey, sorry for taking so long on this PR.

I've opened https://github.com/filecoin-project/lotus/pull/7703 to finally land this (rebased with some tests added, and some smaller changes for maintainability (one notable change is adding sector size to the resource env-vars - this was needed because we want to have shared workers which may work on different sector sizes at the same time))

magik6k:
Merged in #7703!

PR #7273, opened 2021-09-06
fix index out of range

frankffenn:


ZenGround0 (approved):

magik6k (approved):
Oh, interesting, I guess this can happen when using the --publish-now cli with no sectors queued?

Good catch, thanks for the PR!

Issue #7241, opened 2021-08-31
Throttle staging deals on the lotus-miner markets service and reject new deals, until staged deals are added to sectors.


nonsense:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated improvement to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [X] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Improvement Suggestion

Currently Lotus has no way to throttle incoming storage deals. This results in the `deal-staging` directory in the `markets` service repo to become increasingly large without any limits, which results in disk space getting filled to 100% and essentially halting the miner/fullnode.

For example on a small miner configuration, where the sealing area and the full node are hosted on a 2TB SSD, there is room only for a few deals and 1-2 sectors being sealed. However the markets subsystem is accepting and storing/staging too many deals successfully, which fills the SSD very quickly.

Ideally we need a mechanism to specify that we want no more than N deals or M GBs to be in the markets node staging area.

jennijuju:
is it possible to do a dt rate limiting configuration by size (like byte/sec) based on network speed + ongoing traffic?

nonsense:
@jennijuju we considered doing this in Lotus, but decided that it is probably best handled by external to Lotus tools, such as tc and other connection traffic shaping tools.

nonsense:
@jennijuju I think we should close this issue, because it was created for the stopgap tasks of rejecting deals when the staging are is full.

[...]

PR #7276, opened 2021-09-06
introduce MaxStagingDealsBytes - reject new deals if our staging deals area is full

nonsense:
This PR is introducing a `MaxStagingDealsBytes` config - if the markets repository `deals-staging` directory gets too large, the markets node can now be configured to reject new incoming deals.

A stopgap for https://github.com/filecoin-project/lotus/issues/7241

---

TODO:
- [x] add a test as part of the `itests` suite (this PR has already been tested on a devnet)

### DIFF
--- a/node/modules/storageminer.go
+++ b/node/modules/storageminer.go
@@ -542,9 +542,8 @@ func BasicDealFilter(cfg config.DealmakingConfig, user dtypes.StorageDealFilter)
 				return false, "miner error", err
 			}
 
-			diskUsageGiB := diskUsageBytes / 1024 / 1024 / 1024
-			if cfg.MaxStagingDealsGiB != 0 && diskUsageGiB >= cfg.MaxStagingDealsGiB {
-				log.Errorw("proposed deal rejected because there are too many deals in the staging area at the moment", "MaxStagingDealsGiB", cfg.MaxStagingDealsGiB, "DiskUsageGiB", diskUsageGiB)
+			if cfg.MaxStagingDealsBytes != 0 && diskUsageBytes >= cfg.MaxStagingDealsBytes {
+				log.Errorw("proposed deal rejected because there are too many deals in the staging area at the moment", "MaxStagingDealsBytes", cfg.MaxStagingDealsBytes, "DiskUsageBytes", diskUsageBytes)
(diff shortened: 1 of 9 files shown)


benjaminh83:
I'm not sure this fix will be really effective on making sure the market node does not crash. My .lotusmarket folder is currently able to handle 18TiB of data, and I would have no issue with it being filled if it was needed as a buffer. Issue here is the market node that stops working probably when it has x number of transfers all stuck at 0B transferred. 
So setting a limit at x MiB on the deals-staging directory will be a pretty arbitrary number for a miner... Like how will we know how high we can set the limit before the market node gets shaky?

Also this is a pretty easy way to do a DoS on our miner. Estuary sends a miner 100 deals like a railgun and transfers 0B or very few and stream reset. Our folder filles up and we are now rejecting all deals from all clients until we purge the market node. 
Wouldn't be better to look at what client IDs are actually transferring and if it keeps on initiating new deal transfers while it already has transfers at 0B, then reject the individual ID, and not just reject all incoming deals..

benjaminh83:
Off cause, solving the issue of a single machine miner getting the disk full that also houses .lotus and .lotusminer, then it makes perfectly sense. 
But tmpfs files has prior to the market split always lived inside the .lotusminer path, so its not a new thing that we should watch out for the available storage. 
So maybe the throttle I'm referring to is more about making sure the lotus market node does not crash because of too many transfers, where this is more about not running out of space because of deals.

nonsense:
@benjaminh83 this config is about preventing the miner from getting out of disk space. If you run a sealing node and a markets node on the same host, at the moment it is very easy for the markets node to fill the disk space by accepting too many deals.

This PR is not attempting to address all the issues you are describing.

jennijuju:
Would be really nice to get this in v1.11.3 as a stop gap, but okay to postpone to a later release

magik6k (approved):

node/modules/storageminer.go
```diff
@@ -533,6 +537,16 @@ func BasicDealFilter(user dtypes.StorageDealFilter) func(onlineOk dtypes.Conside
 				return false, "miner error", err
 			}
 
+			diskUsageBytes, err := r.DiskUsage(r.Path() + "/deal-staging")
```
magik6k, line 540: should be `filepath.Join`, also ideally we'd not hardcode 'deal-staging'

PR #7308, opened 2021-09-09
Add --unproven flag to the sectors list command

magik6k:
This is https://github.com/filecoin-project/lotus/pull/7263, but this time opened to the right branch


magik6k:
(Merging as #7263 was approved)

PR #7309, opened 2021-09-09
remove job to install jq

ognots:
jq is already installed now in either a newer version of CircleCI's MacOS VMs or in a previous CI step.
I ran a failing macos job with ssh enabled, and inspected '/usr/local/bin' and found found the following output

lrwxr-xr-x    1 distiller  admin    23 Jun 22 14:50 jq -> ../Cellar/jq/1.6/bin/jq

the existing symlink causes the 'Install jq' job to fail.
removing this job should resolve the issue

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -368,11 +368,6 @@ jobs:
           command: |
             curl https://sh.rustup.rs -sSf | sh -s -- -y
       - run:
-          name: Install jq
-          command: |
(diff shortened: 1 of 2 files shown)


magik6k (commented):
Circle config is generated from `.circleci/template.yml`, so the change also needs to be done there

ognots:
@magik6k updated template.yaml and pushed

ognots:
sorry previous comment jumped the gun. change is there now

magik6k (approved):

jennijuju:
thanks @ognots ! backported to release/v1.11.3

PR #7310, opened 2021-09-10
fix: check padSector Cid

swift-mx:


magik6k (approved):
Great catch! Thanks for the PR!

PR #7312, opened 2021-09-10
add `lotus-miner storage-deals list --format=json` with transfers

nonsense:
This PR is updating the `lotus-miner storage-deals list` cmd to print out storage deals together with their respective transfers, for easier debugging - only when using `--format=json`. At the moment fetching storage deals and transfers is done via two separate commands and somewhat cumbersome to use given the tab-delimited tables.

`--format=json` prints out results in line-delimited JSON, which I think is much easier to read than the `table` approach we use in many commands, because:

1. Field type is right next to every value (given that we generally have 100s-1000s of deals):
```
{"datetime":"2021-09-10T00:57:41+03:00","verified-deal":true,"proposal-cid":"bafyreif5yiywi6njej3nz2uwrj6yo2dnrc7mnokk3qhhfc6m46efpurmay","deal-id":0,"deal-status":"StorageDealTransferring","client":"f3vnq2cmwig3qjisnx5hobxvsd4drn4f54xfxnv4tciw6vnjdsf5xipgafreprh5riwmgtcirpcdmi3urbg36a","piece-size":"16GiB","price":"0 FIL","duration-epochs":1494720,"transfer-id":1631214107052231430,"transfer-status":"Ongoing","transferred-data":"0B"}
```

2. Searching and formatting is very easy with `grep` and `jq`:
```
$ LOTUS_MARKETS_PATH=~/markets-repo-location ./lotus-miner storage-deals --format=json | grep 7vsgutuiem | jq
{
  "datetime": "2021-09-10T11:36:48+03:00",
  "verified-deal": true,
  "proposal-cid": "bafyreie2e66s3ztkdrfoqxm7hrch5d75i3nbrtp6jnsdcyrr7vsgutuiem",
  "deal-id": 0,
  "deal-status": "StorageDealTransferring",
  "client": "f3vnq2cmwig3qjisnx5hobxvsd4drn4f54xfxnv4tciw6vnjdsf5xipgafreprh5riwmgtcirpcdmi3urbg36a",
  "piece-size": "8GiB",
  "price": "0 FIL",
  "duration-epochs": 1494720,
  "transfer-id": 1631214107052231700,
  "transfer-status": "Ongoing",
  "transferred-data": "0B"
}
```

### DIFF
--- a/cmd/lotus-miner/market.go
+++ b/cmd/lotus-miner/market.go
@@ -21,6 +21,8 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/multiformats/go-multibase"
 	"github.com/urfave/cli/v2"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"golang.org/x/xerrors"
 
 	cborutil "github.com/filecoin-project/go-cbor-util"
@@ -28,6 +30,7 @@ import (
 	"github.com/filecoin-project/go-fil-markets/storagemarket"
 	"github.com/filecoin-project/go-state-types/abi"
 
+	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain/types"
 	lcli "github.com/filecoin-project/lotus/cli"
@@ -342,6 +345,7 @@ var storageDealsCmd = &cli.Command{
 	Subcommands: []*cli.Command{
 		dealsImportDataCmd,
 		dealsListCmd,
+		dealsListWithTransfersCmd,
 		storageDealSelectionCmd,
 		setAskCmd,
 		getAskCmd,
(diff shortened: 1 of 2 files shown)


dirkmc:
Thanks for putting this together 👍 

I agree the JSON output is clearer and easier to parse.

Would it make sense to add the output format and a flag for whether to include deal transfers onto the existing `list` command?
Something like
`./lotus-miner storage-deals list --format=json --with-transfers`

nonsense:
> Would it make sense to add the output format and a flag for whether to include deal transfers onto the existing list command? Something like
./lotus-miner storage-deals list --format=json --with-transfers

I actually started with that, and it gets messy very quickly to support --with-transfers in the `--format=table` so I decided to just go with a separate command and keep things simpler. I don't have hard feelings about that though, so I could bundle it together with `list`, but maybe add `--with-transfers` only for `--format=json`? WDYT?

nonsense:
It is quite hard to support `--watch` also for `data transfers`, so I opted for:
1. Not adding `--with-transfers` flag at all - always try to load transfers when using `--format=json`
2. Leave `--format=table` as is, as we might want to remove `--watch` altogether.

cmd/lotus-miner/market.go
```diff
@@ -396,63 +403,74 @@ var dealsListCmd = &cli.Command{
...
+			return listDealsWithJSON(cctx)
 		}
-		defer closer()
 
-		ctx := lcli.DaemonContext(cctx)
+		return fmt.Errorf("unknown format: %s; use `table` or `json`", cctx.String("format"))
+	},
+}
+
+func listDealsWithTable(cctx *cli.Context) error {
```
nonsense, line 417: For easier review - this is not modified.

dirkmc (approved):
LGTM 👍

noch3:
In #7235 I used a `--json` flag. Should there be a standard approach to this? I'd like to point out that AFAIK the table output doesn't actually use TSV. The tabs get converted to spaces. This makes parsing it harder than it should be IMO.

nonsense:
> In #7235 I used a `--json` flag. Should there be a standard approach to this? I'd like to point out that AFAIK the table output doesn't actually use TSV. The tabs get converted to spaces. This makes parsing it harder than it should be IMO.

Yes, ideally we should be consistent across various command line tools. Having said that I don't have any preference over `--format=json` or just `--json` and `--format=table` and `--table`, but probably better to be explicit.

For now I will leave the other reviewers to chime in.

magik6k (approved):

PR #7313, opened 2021-09-10
feat: markets: Integrate index ingest protocol and retrieve by any CID

dirkmc:
## Related Issues

The motivation for this work is to solve content routing in Filecoin network, and facilitate retrieval of content by any CID. 

Fixes:
- https://github.com/filecoin-project/lotus/issues/8015

## Proposed Changes
The work here integrates the [indexing ingestion protocol](https://github.com/filecoin-project/storetheindex/blob/main/doc/ingest.md) into lotus using the [default implementation](https://github.com/filecoin-project/index-provider).

As a result, the multihashes contained in successful deals are announced onto indexer ingest gossipsub topic. Similarly, multihashes that belonged to slashed or expired deals are advertised as no longer available. The work exposes an endpoint on the markets process to allow indexer nodes to sync the list of multihashes for processing. The indexers use this information to then provider a service where a user can look up which SP provides a given CID or Multihash.

The work also introduces a peer protection API that allows dynamic change to the list of peer IDs the connection to which is protected by the daemon process. This is introduced so that: 1) there is no need to restart the daemon in order to make changes to this config parameter, and 2) automatically protect connections from markets to daemon process to allow indexing gossip messages to propagate through the network via Damon's connection to Filecoin bootstrap peers.

The `lotus` CLI is expanded to provide a set of utility commands that allow bulk announcements of all deals to indexers, along with functionality to list and modify peer IDs protected by the daemon process.

## Additional Info

Discussion:
- https://github.com/filecoin-project/lotus/discussions/8087

Integrates indexing integration work done in `go-fil-markets`:
- https://github.com/filecoin-project/go-fil-markets/pull/629
- https://github.com/filecoin-project/go-fil-markets/pull/673

Additionally, a range of work is done in [`filecoin-project/dagstore`](https://github.com/filecoin-project/dagstore) to accommodate the indexing integration as well as rolling out a new CARv2 index, that stores the multihash code as well as the multihash digest in the CARv2 indices. 

The core index provider functionality integrated is implemented in [`filecoin-project/index-provider`](https://github.com/filecoin-project/index-provider). The Indexing protocol and indexer node can be found in [`filecoin-project/storetheindex`](https://github.com/filecoin-project/storetheindex).

## Checklist

- [x] CI is green


itests/deals_anycid_test.go
```diff
@@ -0,0 +1,143 @@
...
+		bsa = modules.StorageBlockstoreAccessor(importmgr)
+		return bsa
+	}
+	bsaOpt := kit.ConstructorOpts(node.Override(new(storagemarket.BlockstoreAccessor), bsaFn))
+
+	// Allow 512MB sectors
+	eightMBSectorsOpt := kit.SectorSize(8 << 20)
+
+	// Create a client, and a miner with its own full node
+	_, client, miner, ens := kit.EnsembleTwoOne(t, kit.MockProofs(), bsaOpt, eightMBSectorsOpt)
```
nonsense, line 71: Note here for @raulk and @aarshkshah1992  - we normally use the `EnsembleMinimal` which re-uses full nodes between `miner` and `client` - I believe this is one (if not the only) example, which tests deals properly, as we are NOT reusing the full node between the miner and the client.

cmd/lotus-miner/dagstore.go
```diff
@@ -265,3 +238,114 @@ var dagstoreGcCmd = &cli.Command{
...
+		"ShardStateErrored":   color.FgRed,
+		"ShardStateNew":       color.FgYellow,
+	}
+
+	for _, s := range shards {
+		m := map[string]interface{}{
+			"Key": s.Key,
+			"State": func() string {
+				if c, ok := colors[s.State]; ok {
+					return color.New(c).Sprint(s.State)
```
magik6k, line 265: Supernit: can we trim the `ShardState` prefix? Makes it easier to visually scan, and the prefix itself doesn't really provide any information to the user.

willscott:
This PR description lists a remaining outstanding item of 'CLI command to publish all deals.' Is that something @aarshkshah1992  you're still going to work on?
Can you update the PR description with the remaining tasks that need to happen for it to be merge-able?

node/config/types.go
```diff
@@ -158,6 +161,23 @@ type DealmakingConfig struct {
 	RetrievalPricing *RetrievalPricing
 }
 
+type IndexerProviderConfig struct {
+	config.Ingest
```
masih, line 165: Looking at the output changes in `doc_gen.go` looks like `cfgdocgen/gen.go` does not support embedded structs. 

Consider either fixing the generator or fork `config.Ingest` here?

willscott (approved):

markets/idxprov/mesh.go
```diff
@@ -21,16 +25,25 @@ type Libp2pMeshCreator struct {
...
-	addrs, err := mc.fullnodeApi.NetAddrsListen(ctx)
+	faddrs, err := mc.fullnodeApi.NetAddrsListen(ctx)
 	if err != nil {
 		return err
 	}
 
-	if err := mc.idxProvHost.Connect(ctx, addrs); err != nil {
+	// if we already have a connection to the full node, there's nothing to do here
+	if mc.idxProvHost.Network().Connectedness(faddrs.ID) == network.Connected {
+		return nil
```
willscott, line 35: is there a race here where we're connected, but don't protect?
an extra protect shouldn't hurt since it should be idempotent if done with the same tag.

aarshkshah1992: Addressed this.

node/modules/storageminer_idxprov.go
```diff
@@ -0,0 +1,176 @@
...
+
+	helpers.MetricsCtx
+	fx.Lifecycle
+	Repo      repo.LockedRepo
+	Datastore dtypes.MetadataDS
+	PeerID    peer.ID
+	peerstore.Peerstore
+}
+
+func IndexProviderHost(cfg config.IndexProviderConfig) func(IdxProv) (idxprov.Host, error) {
```
magik6k, line 52: Why do we need a separate host instance?

willscott: we don't and have now gotten rid of the additional host.

go.mod
```diff
@@ -168,9 +168,10 @@ require (
...
 	gopkg.in/cheggaaa/pb.v1 v1.0.28
 	gotest.tools v2.2.0+incompatible
-	lukechampine.com/blake3 v1.1.7 // indirect
 )
 
 replace github.com/filecoin-project/filecoin-ffi => ./extern/filecoin-ffi
 
 replace github.com/filecoin-project/test-vectors => ./extern/test-vectors
+
+replace github.com/ipfs/go-path => github.com/ipfs/go-path v0.0.7
```
willscott, line 177: Is this needed?

documentation/en/default-lotus-miner-config.toml
```diff
@@ -257,6 +257,27 @@
       #Path = ""
 
 
+[IndexProvider]
```
willscott, line 260: should there be a config option example here for how to disable the IndexProvider?

markets/dagstore/wrapper.go
```diff
@@ -398,6 +405,33 @@ func (w *Wrapper) markRegistrationComplete() error {
...
+		return nil, xerrors.Errorf("getting pieces containing block %s: %w", blockCID, err)
+	}
+
+	// Convert from shard key to cid
+	pieceCids := make([]cid.Cid, 0, len(shardKeys))
+	for _, k := range shardKeys {
+		c, err := cid.Parse(k.String())
+		if err != nil {
+			prefix := fmt.Sprintf("getting pieces containing block %s:", blockCID)
+			return nil, xerrors.Errorf(prefix+" converting shard key %s to piece cid: %w", k, err)
```
willscott, line 422: prefix should be a %s format string rather than mixed format + concatenation?

jennijuju:
please apply the checklist in the pr template to this pr description please https://github.com/filecoin-project/lotus/blob/master/.github/pull_request_template.md

documentation/en/default-lotus-miner-config.toml
magik6k: This file is auto-generated from `node/config/types.go`, and will contain whatever defaults are set in the default config

itests/kit/client.go
```diff
@@ -111,9 +111,14 @@ func RunClientTest(t *testing.T, cmds []*lcli.Command, clientNode *TestFullNode)
...
-	fmt.Println("retrieve:\n", out)
-	require.Regexp(t, regexp.MustCompile("Success"), out)
+
+	for {
+		out = clientCLI.RunCmd("client", "retrieve", dataCid.String(), path)
+		fmt.Println("retrieve:\n", out)
+		if strings.Contains(out, "Success") {
+			break
+		}
+	}
```
magik6k, line 121: Why the loop? If there is a good reason, it would be nice to have a comment explaining.

cmd/lotus-miner/index_provider.go
```diff
@@ -0,0 +1,86 @@
...
+
+		ctx := lcli.ReqContext(cctx)
+
+		return marketsApi.IndexerAnnounceDeal(ctx, proposalCid)
+	},
+}
+
+var indexProvAnnounceAllCmd = &cli.Command{
+	Name:  "announce-all",
+	Usage: "Announce all active deals to indexers so they can download its indices",
```
magik6k, line 63: nit: `its` doesn't sound right here

node/config/types.go
magik6k: Probably want to copy config.Ingest into here so config comments get generated correctly

node/modules/storageminer_idxprov.go
```diff
@@ -0,0 +1,71 @@
...
+		log.Infof("Using extra gossip data in index provider engine: %s", ma)
+
+		e, err := engine.New(cfg.Ingest, pkey, dt, marketHost, ipds, retAdds, engine.WithExtraGossipData(ma.Bytes()))
+		if err != nil {
+			return nil, xerrors.Errorf("creating indexer provider engine: %w", err)
+		}
+
+		args.Lifecycle.Append(fx.Hook{
+			OnStart: func(ctx context.Context) error {
+				if err := e.Start(ctx); err != nil {
```
magik6k, line 57: I don't know how this context is used inside e.Start, but note that `OnStart` contexts are cancelled after startup is finished (or if it takes too long - 30s iirc)

magik6k:
Are there any setup docs you could link in this PR?

willscott:
https://github.com/filecoin-project/lotus/discussions/8087 has a lot of docs around it, currently

node/modules/services.go
```diff
@@ -198,6 +199,36 @@ func HandleIncomingMessages(mctx helpers.MetricsCtx, lc fx.Lifecycle, ps *pubsub
...
+	}
+
+	topicHandle, err := ps.Join(topicName)
+	if err != nil {
+		return xerrors.Errorf("failed to join pubsub topic %s: %w", topicName, err)
+	}
+	cancelFunc, err := topicHandle.Relay()
+	if err != nil {
+		return xerrors.Errorf("failed to relay to pubsub messages for topic %s: %w", topicName, err)
+	}
```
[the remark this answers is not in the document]
willscott: the goal here is get gossip from those that participate to the indexer. if we don't have other nodes forwarding messages through the mesh I'm worried it won't be reliable. We implemented a pretty strict validator that @vyzo i think is okay with.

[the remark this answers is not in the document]
vyzo: yeah, the validator is pretty good, i am not worried about it.

node/modules/storageminer_idxprov.go
masih: Left comment to note that it is I believe safe to reuse the context. TLDR; `e.Start` does not hold onto the given context and uses for rehydrating cache etc.

Please unresolve if there is something I might have missed.

go.mod
masih: Looks like this is needed otherwise node chain API implementation would need an upgrade. 

Resolving. Please unresolve if this is a blocker.

documentation/en/default-lotus-miner-config.toml
masih: Added `Enable` flag to this config section, set to false by default until this PR is merged to master.

go.mod
magik6k: Yeah, we shouldn't have this here unless it means a big change. I can be convinced to keep this if we'll fix this soon in a followup PR

magik6k (commented):
Feels very close to landing. CI is failing.

Looks like there is an unintentional ffi update pushed.

chain/store/snapshot.go
```diff
@@ -197,7 +197,7 @@ func recurseLinks(ctx context.Context, bs bstore.Blockstore, walked *cid.Set, ro
 			rerr = err
 		}
 	})
-	if err != nil {
+	if err != nil && err != io.EOF {
```
magik6k, line 200: This is a new bug in cbor-gen.

Fixed in 37c43cae8e145618a0ca6f2c8c1e3e5ef39a7d40 / https://github.com/whyrusleeping/cbor-gen/pull/63

go.mod
masih: I have upgraded the code here: https://github.com/filecoin-project/lotus/pull/7313/commits/1bb65189af87631b2adc32ccc728f1af1ffee554

Please review carefully as I am not well rehearsed with lotus codebase. 
Resolving.
Thanks

node/impl/full/chain.go
```diff
@@ -557,21 +559,16 @@ func (a *ChainAPI) ChainGetNode(ctx context.Context, p string) (*api.IpldObject,
...
 
-	dag := merkledag.NewDAGService(bsvc)
-
-	r := &resolver.Resolver{
-		DAG:         dag,
-		ResolveOnce: resolveOnce(bs, a.TsExec),
-	}
-
-	node, err := r.ResolvePath(ctx, ip)
+	node, lnk, err := r.ResolvePath(ctx, ip)
```
magik6k, line 565: This needs to use the `resolveOnce` we implement above; It's what implements most of the functionality of this API

magik6k (approved):

testplans/lotus-soup/go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/drand/drand v1.3.0
 	github.com/filecoin-project/go-address v0.0.6
 	github.com/filecoin-project/go-data-transfer v1.14.0
-	github.com/filecoin-project/go-fil-markets v1.19.1
+	github.com/filecoin-project/go-fil-markets v1.20.0
```
willscott, line 12: this should be bumped to v1.20.1 before merge

magik6k (approved):

PR #7321, opened 2021-09-12
Mempool msg selection should respect block message limits

arajasek:


vyzo (commented):
ugh, this looks ugly... not passing judgement yet though.

magik6k (approved):

chain/messagepool/selection.go
```diff
@@ -46,24 +51,155 @@ func (mp *MessagePool) SelectMessages(ctx context.Context, ts *types.TipSet, tq
...
 
-	if len(msgs) > build.BlockMessageLimit {
-		msgs = msgs[:build.BlockMessageLimit]
+	if sm == nil {
+		return nil, nil
+	}
+
+	// one last sanity check
+	if len(sm.msgs) > build.BlockMessageLimit {
+		sm.msgs = sm.msgs[:build.BlockMessageLimit]
```
Kubuxu, line 72: Log warning or error because this means that the code is wrong.

Kubuxu (approved):
SGTM, 1 nit

chain/messagepool/selection.go
arajasek: Yessir!

PR #7327, opened 2021-09-13
fix: Enforce verification key integrity check regardless of TRUST_PARAMS=1

### LINKED ISSUE #6640: [BUG] `lotus daemon` should refuse to run unless all verification params are present
**Describe the bug**

During the recent upgrade a number of folks ended up running their nodes with the new aggregate-proof validation param missing. This led to the formation of an uncomfortably long-lived side-chain.

Since the validation set is really small ( less than 300MiB ), we should just check its presence and integrity on every startup, regardless of `TRUST_PARAMS` or any other escape hatches.

placer14:
closes #6640 

See change in https://github.com/filecoin-project/go-paramfetch/pull/17 for details.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/filecoin-project/go-fil-markets v1.12.0
 	github.com/filecoin-project/go-jsonrpc v0.1.4-0.20210217175800-45ea43ac2bec
 	github.com/filecoin-project/go-padreader v0.0.0-20210723183308-812a16dc01b1
-	github.com/filecoin-project/go-paramfetch v0.0.2-0.20210614165157-25a6c7769498
+	github.com/filecoin-project/go-paramfetch v0.0.2
 	github.com/filecoin-project/go-state-types v0.1.1-0.20210810190654-139e0e79e69e
 	github.com/filecoin-project/go-statemachine v1.0.1
 	github.com/filecoin-project/go-statestore v0.1.1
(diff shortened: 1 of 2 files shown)


placer14:
Please check that I did the module dependency dance properly. A `go mod tidy` seemed to reverse the update to `go-fetchparam` which was not what I expected (as it is neither unused or missing). 🤷

placer14:
ATTENTION: HOLD THIS PR FROM BEING MERGED!

(I realize that the dependent package may squash-merge causing the final commit SHA to change. This would cause the build to break later once dangling branches are garbage-collected and that would be bad.)

placer14:
Holding this PR as I now realize the dependent package could be squash-merged causing the commit SHA to change. I'll ready this again once the dependent PR is merged and SHA updated.

go.mod
```diff
@@ -39,7 +39,7 @@ require (
 	github.com/filecoin-project/go-fil-markets v1.12.0
 	github.com/filecoin-project/go-jsonrpc v0.1.4-0.20210217175800-45ea43ac2bec
 	github.com/filecoin-project/go-padreader v0.0.0-20210723183308-812a16dc01b1
-	github.com/filecoin-project/go-paramfetch v0.0.2-0.20210614165157-25a6c7769498
+	github.com/filecoin-project/go-paramfetch v0.0.2-0.20210913223022-0cca2e511a58
```
magik6k, line 42: Nit: use a tagged version

magik6k (approved):

PR #7329, opened 2021-09-14
Add caches to lotus-stats and splitcode

travisperson:


arajasek (approved):
I think this is fine as is

magik6k (approved):

Issue #7333, opened 2021-09-15
Can we use the new features in v1.11.2 on devnet?


dayou5168:
### Discussed in https://github.com/filecoin-project/lotus/discussions/7323

<div type='discussions-op-text'>

<sup>Originally posted by **dayou5168** September 13, 2021</sup>
I tried to renew sectors.
`lotus-miner sectors renew --sector-file /tmp/sector 1555200`
and i got this response
> Renewing 1 sectors: 
{
  "Extensions": [
    {
      "Deadline": 21,
      "Partition": 0,
      "Sectors": "46",
      "NewExpiration": 1577324
    }
  ]
}
1 sectors renewed

my sectorfile content is:
> cat /tmp/sector 
46

sector 46 still have old lifecycle. and if i use lotus-miner sectors extend command. i got success.
> 46  Proving  YES      YES     703377 (in 33 weeks 5 days)   2      1.5KiB

and also when i change CommittedCapacitySectorLifetime = "46.656ms" to anytime . seems it will not effect the sectors lifecycle in devnet.</div>

Reiers:
**Fix merged !**

Closing ticket.

PR #7337, opened 2021-09-16
fix bug for CommittedCapacitySectorLifetime

ldoublewood:
This PR is fixing the follow two bugs related to #7333 
 
- Wrong default CommittedCapacitySectorLifetime config value
- CommittedCapacitySectorLifetime config item is not working


magik6k (commented):
Thanks for catching this!

Needs `make gen`

magik6k (approved):

Issue #7315, opened 2021-09-10
automatically export helpful info needed for debugging upon crash


jennijuju:
This can be a shed cmd or some feature miner can enable in the configuration. The goal would be auto-saving all the information can be used for debugging whenever there is a node crash, hang, error and so on.  Info to save can be similar to `lotus-miner info all`, logs, goroutines, dmesg. etc.

placer14:
Putting some concreteness around this, I'm going to use a panic as the most obvious trigger for this behavior. I will write it such that it can be called from anywhere should we decide to add this to outher high-value failure cases.

Will start with a stacktrace, goroutine, and journal+log snippet from prior $DEBUG_PANICNET_LOGLOOKBACK=1m (or something) and we can go from there.

f8-ptrk:
the journal does what the journal does. people just need to know about it in the first place. it's kind of a hidden feature. it rotates properly upon restart already i think

maybe have a debug log that get rotated ever 2nd restart or some like that. 

i wouldn't connect this feature to the normal log people are running

f8-ptrk:
logs connected to a crash can date back hours @placer14 

thats my main concern with $DEBUG_PANICLOG_LOOKBACK=1m

gh3990:
maybe have a little separation in there like :

lotus-miner info all (for everything mentioned above by Jenn)
lotus-miner info WinPost (for everything related to troubleshooting WinPoST)
lotus-miner markets info all (everything relate to the market node)
lotus-miner tech info all (will also take env, storage config, miner config, api config, network config)

just an idea :-)

TippyFlitsUK:
I would opt top have this as a manually triggered additional command rather than something that is fired automatically.

Meatball13:
Yeah, agreed that it might be nice to have it manually triggered.

placer14:
@f8-ptrk 
> the journal does what the journal does. people just need to know about it in the first place. it's kind of a hidden feature. it rotates properly upon restart already i think

The journal eventually rotates that away if you're not looking for the panic info at that moment. We're trying to make it easy for operators to gather this information easily. The journal is still going to journal and we'll snapshot the last N lines for this report to save for later, as needed.

> i doubt it makes sense to keep extensive logs in memory just to write them out in case of a crash
I'm leaving the logs alone expecting that will already be stored by most rudimentary deployments. The example above keeps nothing in memory until triggered (and then allocs the bare minimum in case the panic is due to OOM conditions).

[...]

PR #7341, opened 2021-09-17
feat: Catch panic to generate report and reraise

placer14:
an attempt at #7315

Room for improvement?
- ~expose [journal lookback param](https://github.com/filecoin-project/lotus/pull/7341/files#diff-3c77b768688af24630531611dce1d0271ad9f14cbbb2b4fc893ffaa5fec406a4R25) as envvar~ done
- report rotation param and exposing as envvar?
- dump more things?
- maybe don't catch such a broad surface and find more narrow scopes to recover from?
- make each report optional (similar to GOLOG_LOG_LEVEL_NAMED)?

### DIFF
--- a/build/panic_reporter.go
+++ b/build/panic_reporter.go
@@ -79,7 +79,7 @@ func writeAppVersion(file string) {
 	}
 	defer f.Close() //nolint:errcheck
 
-	versionString := []byte(BuildVersion + buildType() + CurrentCommit + "\n")
+	versionString := []byte(BuildVersion + BuildTypeString() + CurrentCommit + "\n")
 	if _, err := f.Write(versionString); err != nil {
 		panicLog.Error(err.Error())
 	}
--- a/build/panic_reporter.go
+++ b/build/panic_reporter.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"runtime/debug"
 	"runtime/pprof"
+	"strconv"
 	"syscall"
 	"time"
 
@@ -14,15 +15,18 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 )
 
-var panicLog = logging.Logger("panic-reporter")
+var (
+	panicLog           = logging.Logger("panic-reporter")
+	defaultJournalTail = 500
+)
 
 // PanicReportingPath is the name of the subdir created within the repoPath path
 // provided to GeneratePanicReport
-var PanicReportingPath = "lotus_panic_reports"
+var PanicReportingPath = "lotus-panic-reports"
 
 // PanicReportJournalTail is the number of lines captured from the end of
 // the lotus journal to be included in the panic report.
-var PanicReportJournalTail = 500
+var PanicReportJournalTail = defaultJournalTail
 
 // GeneratePanicReport produces a timestamped dump of the application state
 // for inspection and debugging purposes. Call this function from any place
@@ -39,6 +43,14 @@ func GeneratePanicReport(persistPath, repoPath, label string) {
 		panicLog.Error("persist path is empty, aborting panic report generation")
 		return
 	}
+	tl := os.Getenv("LOTUS_PANIC_JOURNAL_LOOKBACK")
+	if tl != "" && PanicReportJournalTail == defaultJournalTail {
+		i, err := strconv.Atoi(tl)
+		if err == nil {
+			PanicReportJournalTail = i
+		}
+	}
+
 	reportPath := filepath.Join(persistPath, PanicReportingPath, generateReportName(label))
 	panicLog.Warnf("generating panic report at %s", reportPath)
 	syscall.Umask(0)
--- a/build/panic_reporter.go
+++ b/build/panic_reporter.go
@@ -79,11 +79,7 @@ func writeAppVersion(file string) {
 	}
 	defer f.Close() //nolint:errcheck
 
-	ignoreCommitBefore := os.Getenv("LOTUS_VERSION_IGNORE_COMMIT")
-	os.Setenv("LOTUS_VERSION_IGNORE_COMMIT", "")                       //nolint:errcheck
-	defer os.Setenv("LOTUS_VERSION_IGNORE_COMMIT", ignoreCommitBefore) //nolint:errcheck
-
-	versionString := []byte(UserVersion() + "\n")
+	versionString := []byte(BuildVersion + buildType() + CurrentCommit + "\n")
 	if _, err := f.Write(versionString); err != nil {
 		panicLog.Error(err.Error())
 	}
--- a/build/panic_reporter.go
+++ b/build/panic_reporter.go
@@ -8,7 +8,7 @@ import (
 	"runtime/debug"
 	"runtime/pprof"
 	"strconv"
-	"syscall"
+	"strings"
 	"time"
 
 	"github.com/icza/backscanner"
@@ -58,8 +58,7 @@ func GeneratePanicReport(persistPath, repoPath, label string) {
 		}
 	}
 
-	syscall.Umask(0)
-	err := os.MkdirAll(reportPath, 0755)
+	err := os.MkdirAll(reportPath, 0644)
 	if err != nil {
 		panicLog.Error(err.Error())
 		return
@@ -157,11 +156,7 @@ func writeJournalTail(tailLen int, repoPath, file string) {
 			}
 			break
 		}
-		if _, err := f.Write(line); err != nil {
-			panicLog.Error(err.Error())
-			break
-		}
-		if _, err := f.Write([]byte("\n")); err != nil {
+		if _, err := f.Write(append(line, "\n"...)); err != nil {
 			panicLog.Error(err.Error())
 			break
 		}
@@ -179,5 +174,6 @@ func getLatestJournalFilePath(repoPath string) (string, error) {
 }
 
 func generateReportName(label string) string {
-	return fmt.Sprintf("report_%s_%s", label, time.Now().Format("2006-01-02T150405Z0700"))
+	label = strings.ReplaceAll(label, " ", "")
+	return fmt.Sprintf("report_%s_%s", label, time.Now().Format("2006-01-02T150405"))
 }
--- a/build/panic_reporter.go
+++ b/build/panic_reporter.go
@@ -1,0 +1,146 @@
+package build
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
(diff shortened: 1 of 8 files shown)


build/panic_reporter.go
```diff
@@ -0,0 +1,146 @@
...
+	"runtime/debug"
+	"runtime/pprof"
+	"syscall"
+	"time"
+
+	"github.com/icza/backscanner"
+	logging "github.com/ipfs/go-log/v2"
+)
+
+var panicLog = logging.Logger("panic_reporter")
```
travisperson, line 17:
```suggestion
var panicLog = logging.Logger("panic-reporter")
```

build/panic_reporter.go
```diff
@@ -0,0 +1,146 @@
...
+
+// PanicReportingPath is the name of the subdir created within the repoPath path
+// provided to GeneratePanicReport
+var PanicReportingPath = "panic_reports"
```
travisperson, line 21:
```suggestion
var PanicReportingPath = "panic-reports"
```

cli/helper.go
```diff
@@ -32,6 +33,13 @@ func ShowHelp(cctx *ufcli.Context, err error) error {
 }
 
 func RunApp(app *ufcli.App) {
+	defer func() {
+		if r := recover(); r != nil {
+			// Generate report in LOTUS_PATH and re-raise panic
+			build.GeneratePanicReport(os.Getenv("LOTUS_PATH"), "app_panic")
```
travisperson, line 39: This isn't guaranteed to be set, and each binary may have it's own location eg `$LOTUS_MINER_PATH`, `$LOTUS_WORKER_PATH`, etc.

Have you tried to see if you can run this recovery in the `After` method of urfave cli? This will give you access to the urfave cli context which will allow you to use the cli flags and avoid using `os.Getenv`.

Being able to customize the location where the reports are generated would also be amazing. Something like `LOTUS_PANIC_REPORTS` or similar that is accepted on every binary would be great. It could default into the `LOTUS_*_PATH`. You can use `app.Name` to do a lookup for the correct cli flag to read from to get the correct repo path for defaults + overrides using env vars.

The second argument could then also be the application name as well. This in conjunction with a `LOTUS_PANIC_REPORTS` would mean someone could have all their panics send to a single location and then identified by the label.

build/panic_reporter.go
travisperson: eg:
https://github.com/filecoin-project/lotus/blob/6df17bceb1908683cb83ee40423ab767a6eb399e/chain/wallet/ledger/ledger.go#L24
https://github.com/filecoin-project/lotus/blob/6df17bceb1908683cb83ee40423ab767a6eb399e/chain/consensus/filcns/filecoin.go#L47

build/panic_reporter.go
travisperson: eg:
https://github.com/filecoin-project/lotus/blob/6df17bceb1908683cb83ee40423ab767a6eb399e/node/modules/client.go#L116

cli/helper.go
placer14: I used the local context instead of a lookup based on `app.Name`. LMK if you think this is sufficient. (Was trying to avoid brittle lookup regressions if names change.) I think I got all of your suggestions in here now.

build/panic_reporter.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	tl := os.Getenv("LOTUS_PANIC_JOURNAL_LOOKBACK")
+	if tl != "" && PanicReportJournalTail == defaultJournalTail {
+		i, err := strconv.Atoi(tl)
+		if err == nil {
+			PanicReportJournalTail = i
+		}
+	}
+
+	reportPath := filepath.Join(persistPath, PanicReportingPath, generateReportName(label))
```
magik6k, line 54: Can we drop `PanicReportingPath` from here, and just use persistPath - then when we call this function we'd:
* Use --panic-reports if set
* Use `Join(repoPath, PanicReportingPath)` otherwise

build/panic_reporter.go
```diff
@@ -0,0 +1,164 @@
...
+	err := os.MkdirAll(reportPath, 0755)
+	if err != nil {
+		panicLog.Error(err.Error())
+		return
+	}
+
+	writeStackTrace(filepath.Join(reportPath, "stacktrace.dump"))
+	writeProfile("goroutines", filepath.Join(reportPath, "goroutines.pprof.gz"))
+	writeProfile("heap", filepath.Join(reportPath, "heap.pprof.gz"))
+	writeJournalTail(PanicReportJournalTail, repoPath, filepath.Join(reportPath, "journal.ndjson"))
```
magik6k, line 66: Would be really great to also add version information to the report (using `build.UserVersion()`)

build/panic_reporter.go
```diff
@@ -51,27 +58,43 @@ func GeneratePanicReport(persistPath, repoPath, label string) {
...
+func writeAppVersion(file string) {
+	f, err := os.Create(file)
+	if err != nil {
+		panicLog.Error(err.Error())
+	}
+	defer f.Close() //nolint:errcheck
+
+	ignoreCommitBefore := os.Getenv("LOTUS_VERSION_IGNORE_COMMIT")
+	os.Setenv("LOTUS_VERSION_IGNORE_COMMIT", "")                       //nolint:errcheck
+	defer os.Setenv("LOTUS_VERSION_IGNORE_COMMIT", ignoreCommitBefore) //nolint:errcheck
```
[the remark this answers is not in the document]
placer14: I opted to keep version func calls private until we know we want it to live external to `build` package. Didn't want to expose more if we don't really need to.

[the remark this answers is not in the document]
magik6k: There was no good reason for it to be private other than nothing needed it to be public before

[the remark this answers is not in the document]
placer14: Okay, sounds like you really want it. https://github.com/filecoin-project/lotus/pull/7341/commits/7525e74894a415db5163995c26414a9eb5f88f8c

build/panic_reporter.go
```diff
@@ -0,0 +1,183 @@
...
+
+	tl := os.Getenv("LOTUS_PANIC_JOURNAL_LOOKBACK")
+	if tl != "" && PanicReportJournalTail == defaultJournalTail {
+		i, err := strconv.Atoi(tl)
+		if err == nil {
+			PanicReportJournalTail = i
+		}
+	}
+
+	syscall.Umask(0)
```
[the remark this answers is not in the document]
placer14: I was debugging something and accidentally committed it. Removed.

build/panic_reporter.go
```diff
@@ -0,0 +1,179 @@
...
+
+	tl := os.Getenv("LOTUS_PANIC_JOURNAL_LOOKBACK")
+	if tl != "" && PanicReportJournalTail == defaultJournalTail {
+		i, err := strconv.Atoi(tl)
+		if err == nil {
+			PanicReportJournalTail = i
+		}
+	}
+
+	err := os.MkdirAll(reportPath, 0644)
```
magik6k, line 61: Dirs need the +x bit to be usable (so this should be 755)

build/panic_reporter.go
```diff
@@ -157,11 +156,7 @@ func writeJournalTail(tailLen int, repoPath, file string) {
 			}
 			break
 		}
-		if _, err := f.Write(line); err != nil {
-			panicLog.Error(err.Error())
-			break
-		}
-		if _, err := f.Write([]byte("\n")); err != nil {
+		if _, err := f.Write(append(line, "\n"...)); err != nil {
```
magik6k, line 159: backscanner docs say that the slice returned from LineBytes shouldn't be modified -> https://pkg.go.dev/github.com/icza/backscanner?utm_source=godoc#Scanner.LineBytes

placer14: Good catch.

cmd/lotus-seal-worker/main.go
```diff
@@ -75,6 +75,12 @@ func main() {
 				Value:   "~/.lotusworker", // TODO: Consider XDG_DATA_HOME
 				Usage:   fmt.Sprintf("Specify worker repo path. flag %s and env WORKER_PATH are DEPRECATION, will REMOVE SOON", FlagWorkerRepoDeprecation),
 			},
+			&cli.StringFlag{
+				Name:    "panic-reports",
+				EnvVars: []string{"LOTUS_PANIC_REPORT_PATH"},
+				Hidden:  true,
+				Value:   "~/.lotus", // should follow --repo default
```
magik6k, line 82:
```suggestion
				Value:   "~/.lotusworker", // should follow --repo default
```

cmd/lotus-miner/main.go
```diff
@@ -113,6 +113,12 @@ func main() {
 				Usage:       "use color in display output",
 				DefaultText: "depends on output being a TTY",
 			},
+			&cli.StringFlag{
+				Name:    "panic-reports",
+				EnvVars: []string{"LOTUS_PANIC_REPORT_PATH"},
+				Hidden:  true,
+				Value:   "~/.lotus", // should follow --repo default
```
magik6k, line 120:
```suggestion
				Value:   "~/.lotusminer", // should follow --repo default
```

cmd/lotus-miner/main.go
```diff
@@ -146,6 +152,14 @@ func main() {
 			}
 			return nil
 		},
+		After: func(c *cli.Context) error {
+			if r := recover(); r != nil {
+				// Generate report in LOTUS_PATH and re-raise panic
+				build.GeneratePanicReport(c.String("panic-reports"), c.String("repo"), c.App.Name)
```
magik6k, line 158:
```suggestion
				build.GeneratePanicReport(c.String("panic-reports"), c.String(FlagMinerRepo), c.App.Name)
```

magik6k (commented):
panic_reporter.go looks good, just some nits/questions on the report paths in miner/worker

build/panic_reporter.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func writeAppVersion(file string) {
+	f, err := os.Create(file)
+	if err != nil {
+		panicLog.Error(err.Error())
+	}
+	defer f.Close() //nolint:errcheck
+
+	versionString := []byte(BuildVersion + BuildTypeString() + CurrentCommit + "\n")
```
Kubuxu, line 81: You can use `build.UserVersion()`.

magik6k (approved):
Would be great to have a test for this someday, but it looks like it should work

PR #7344, opened 2021-09-17
fix Local Multi Sign Wallet #298

kaola526:


arajasek:
Thanks for the PR!

The idea was that this flag could only be supplied for read-only actions...I'd be happy to accept this change, though!

Issue #7345, opened 2021-09-17
1.11.3-rc1: Deal problem caused mine crashed and not able to get restarted


WillTechMusing:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [X] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [X] lotus miner - proving(WindowPoSt)
- [X] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon:  1.11.3-rc1+mainnet+git.5db02ca05+api1.3.0
Local: lotus version 1.11.3-rc1+mainnet+git.5db02ca05
```

### Describe the Bug

miner crashed after running hours. when tried to restarted, it crashed after printing the following errors. At this point, the miner is not able to start so missing all window post deadlines already.

from the miner log:
[...]

PR #7346, opened 2021-09-17
GetCurrentDealInfo err: handle correctly err case

nonsense:
Fixes: https://github.com/filecoin-project/lotus/issues/7345

---

When `GetCurrentDealInfo` returns an error, such as:

```
		return nil, xerrors.Errorf(
			"deal %d not found "+
				"- deal may not have completed sealing before deal proposal "+
				"start epoch, or deal may have been slashed",
			dealID)
```

we should just note that we can't recover that deal, mark it as failed, and later remove the sector.

At the moment when `GetCurrentDealInfo` returns an error, we continue below the if-condition check for err and we throw a nil pointer exception and crash.


ribasushi (approved):

jennijuju:
backported in v1.11.3-rc2

PR #7355, opened 2021-09-20
lotus shed: fr32 utils

magik6k:


cmd/lotus-shed/fr32.go
```diff
@@ -0,0 +1,62 @@
...
+	Name:        "fr32",
+	Description: "fr32 tools",
+	Subcommands: []*cli.Command{
+		fr32WriteCmd,
+		fr32ReadCmd,
+	},
+}
+
+var fr32WriteCmd = &cli.Command{
+	Name:  "write",
```
Kubuxu, line 24: Maybe give it similar semantics as `base64` and other encoding decoding commands.
Encode by default, decode with `-d` flag.
Not sure.

Kubuxu (approved):

PR #7361, opened 2021-09-21
CLI tools for the DAGStore inverted index

nonsense:
This PR is adding 2 CLI tools to the DAGStore inverted index:
- Lookup the pieces that a CID belongs to
- Inspecting the size of the inverted index

It is not ready for merging, because we first need to:
- [ ] merge parent PR and upgrade libp2p as per the parent PR: https://github.com/filecoin-project/lotus/pull/7313
- [x] upgrade DAGStore and its APIs to the latest version (current version still refers to CIDs and not multihashes, and `Size()` is missing).

TODO:
- [ ] rename `inverted index` to `piece index` so it is easier for users to understand what this is about.


cmd/lotus-miner/dagstore.go
```diff
@@ -265,3 +236,108 @@ var dagstoreGcCmd = &cli.Command{
...
+		defer closer()
+
+		ctx := lcli.ReqContext(cctx)
+
+		size, err := marketsApi.DagstoreInvertedIndexSize(ctx)
+		if err != nil {
+			return err
+		}
+
+		fmt.Println(size)
```
dirkmc, line 302: I suggest we "humanize" the size here, eg "1k" instead of "1024"
I think somewhere in the code base we're importing a library to do that.

cmd/lotus-miner/dagstore.go
```diff
@@ -265,3 +236,108 @@ var dagstoreGcCmd = &cli.Command{
...
+		&cli.BoolFlag{
+			Name:        "color",
+			Usage:       "use color in display output",
+			DefaultText: "depends on output being a TTY",
+		},
+		&cli.BoolFlag{
+			Name:        "cid",
+			Usage:       "cid to lookup",
+			DefaultText: "",
+		},
```
dirkmc, line 321: I think this should be a string rather than a bool.
But I'd suggest we pass it as a required argument, rather than a flag.

dirkmc (approved):

PR #7370, opened 2021-09-22
fix Local Multi Sign Wallet

kaola526:
fix Local Multi Sign Wallet issue [#298 ](https://github.com/filecoin-project/community/issues/298)


magik6k:
Looking at this briefly it looks really cool.

(to fix the gen/docs CI, run `make gen` and `make docsgen-cli`, and commit changes)

kaola526:
> Looking at this briefly it looks really cool.
> 
> (to fix the gen/docs CI, run `make gen` and `make docsgen-cli`, and commit changes)

Running `make gen` and `make docsgen-cli` locally succeeded, but the following task failed on circleci.What do I need to do next

chain/wallet/multi.go
```diff
@@ -152,6 +152,17 @@ func (m MultiWallet) WalletImport(ctx context.Context, info *types.KeyInfo) (add
 	return w.WalletImport(ctx, info)
 }
 
+func (m MultiWallet) WalletMsigImport(ctx context.Context, idAddress address.Address, robustAddress address.Address) error {
+	var local getif = m.Local
+
```
jennijuju, line 157: there should be a check to make sure that whoever is importing the msig has the key to one of the msig signers, because later we also have to pick which of the signer addresses to use for msig operations

chain/wallet/wallet.go
```diff
@@ -160,6 +161,24 @@ func (w *LocalWallet) WalletImport(ctx context.Context, ki *types.KeyInfo) (addr
 	return k.Address, nil
 }
 
+func (w *LocalWallet) WalletMsigImport(ctx context.Context, idAddress address.Address, robustAddress address.Address) error {
+	log.Infof("WalletMsigImport idAddress:%s, robustAddress:%s", idAddress.String(), robustAddress.String())
+
+	w.lk.Lock()
+	defer w.lk.Unlock()
+
```
jennijuju, line 169: >there should be a check to make sure that whoever is importing the msig has the key to one of the msig signers, because later we also have to pick which of the signer addresses to use for msig operations

same here

cli/send.go
```diff
@@ -145,6 +150,52 @@ var sendCmd = &cli.Command{
...
 
+		if msig := cctx.String("msig"); msig != "" {
+			msigAddr, err := address.NewFromString(msig)
+			if err != nil {
+				return err
+			}
+
+			api2 := srv.FullNodeAPI()
+
+			proto, err := api2.MsigPropose(ctx, msigAddr, params.To, params.Val, params.From, 0, []byte(""))
```
jennijuju, line 161: how are we choosing which signer address we want to use to propose the message?

jennijuju (changes requested):

chain/wallet/multi.go
kaola526: 创建多签地址，创建人并不一定作为多签地址的签名人哦。
另外考虑到多签地址跟签名者没有强关联关系。
这个修改可能会导致原创建接口进行相应的修改。
当然我觉得统一修改也会更简化使用者的操作，您可以再考虑一下，需不需要优化

magik6k (commented):
First pass, the daemon side looks pretty good. Some comments on the CLI parts.

(sorry for the delay, last few weeks were pretty busy for us)

kaola526:
1，Add the 'walletapprove' command, considering that ordinary wallet is a main role. Of course, as a basic instruction, it is indeed repeated. Agree to remove；
2，“can we add an '--import' flag to `lotus msig create`”,Good idea
3，If the robustaddress address can pass well from the idaddress. It can be removed. I didn't find a good way at that time. Agree to remove. Please tell me how to get robustaddress。

magik6k:
you can get an ID address from a robust address with the `StateLookupID` api. Going the other way is a bit harder (we don't have an api for this right now)

kaola526:
image
`lotus msig create --import=true` and `lotus wallet list` Command test

kaola526:
image
`lotus send --from xxx --msig xxx` Comment test
image
`lotus msig approve --from xxx multisigAddress messageId ` Commend test

kaola526:
How about the progress? Are there any more advice? Thanks?

rjan90:
@magik6k Is this worth bringing over the line? Or is this so stale that it is better to close, and start over again?

rjan90:

Closing

PR #7376, opened 2021-09-22
Fix Drand fetching around null tipsets

### LINKED ISSUE #3613: Lotus fetches drand from chain backwards with null blocks while specs fetches forward

arajasek:
Closes #3613


chain/store/rand.go
```diff
@@ -79,16 +69,27 @@ func (cs *ChainStore) GetBeaconRandomness(ctx context.Context, blks []cid.Cid, p
...
+	if err != nil {
+		return nil, err
+	}
 
-func (cs *ChainStore) GetChainRandomnessLookingForward(ctx context.Context, blks []cid.Cid, pers crypto.DomainSeparationTag, round abi.ChainEpoch, entropy []byte) ([]byte, error) {
-	return cs.GetChainRandomness(ctx, blks, pers, round, entropy, false)
+	be, err := cs.ExtractBeaconEntryForRound(randTs, uint64(round))
+	if err != nil {
+		log.Errorf("failed to get beacon entry as expected: %w", err)
+		// If this failed for some reason, just use the latest entry (lest we halt the chain)
```
arajasek, line 81: Bad idea? Better to just fail & stop?

chain/sync_test.go
```diff
@@ -1037,81 +1035,6 @@ func TestSyncCheckpointEarlierThanHead(t *testing.T) {
 	require.True(tu.t, p1Head.Equals(b.TipSet()))
 }
 
-func TestDrandNull(t *testing.T) {
```
arajasek, line 1040: This test was pretty bad...better test in rand_test now that also covers this case

chain/store/rand.go
nikkolasg: If a node can't retrieve the beacon at this round for network reason, that means he'll try to get the one before, is that right ? In that case, it can cause a split with him and the rest of the chain no ?

arajasek: Yes...but if every node runs into the same problem, this would keep the network going (violating the spec in the process)

I'm almost certain we don't wanna do this, but wanted to get feedback

chain/store/rand.go
```diff
@@ -146,6 +143,56 @@ func (cs *ChainStore) GetLatestBeaconEntry(ts *types.TipSet) (*types.BeaconEntry
 	return nil, xerrors.Errorf("found NO beacon entries in the 20 latest tipsets")
 }
 
+func (cs *ChainStore) ExtractBeaconEntryForRound(ts *types.TipSet, round uint64) (*types.BeaconEntry, error) {
+	cbe := ts.Blocks()[0].BeaconEntries
+	for _, v := range cbe {
+		if v.Round == round {
```
Kubuxu, line 149: You are passing ChainEpoch as round and comparing it with Drand round.

chain/store/store.go
```diff
@@ -1189,3 +1189,38 @@ func breakWeightTie(ts1, ts2 *types.TipSet) bool {
...
 }
+
+func (cs *ChainStore) GetTipSetFromKey(tsk types.TipSetKey) (*types.TipSet, error) {
+	if tsk.IsEmpty() {
+		return cs.GetHeaviestTipSet(), nil
+	}
+	return cs.LoadTipSet(tsk)
+}
+
+func (cs *ChainStore) GetLatestBeaconEntry(ts *types.TipSet) (*types.BeaconEntry, error) {
```
arajasek, line 1200: This could be in the new stateRand instead, doesn't really make a difference

Kubuxu (approved):

magik6k (commented):
Post-merge lgtm

chain/store/rand.go
nicola: wait, does this create a world where a block at an epoch could have two different beacons? if yes, then this is a problem.

nikkolasg: Yes the default behaviour should be that we can't validate this block (yet), we can't assume "everyone on the network has the same problem accessing drand".

arajasek: This is outdated, the code has been removed

PR #7378, opened 2021-09-22
ChainStore: Add a tiebreaker rule for tipsets of equal weight

### LINKED ISSUE #2981: "Tie breaker rule": implement a method to decide between chains with the same weight
**Background**
Two (or more) forks may have the same exact weight and in this case we need a rule to decide which one to extend. 

**Proposal**
Implement the following rule already described in the spec [link](https://filecoin-project.github.io/specs/#selecting-between-tipsets-with-equal-weight):

> When selecting between Tipsets of equal weight, a miner chooses the one with the smallest final ticket.
> 
> In the case where two Tipsets of equal weight have the same min ticket, the miner will compare the next smallest ticket (and select the Tipset with the next smaller ticket). This continues until one Tipset is selected.

Note: we should use the ticket used as election proof (`VRF(drand)`) for this ([updated spec](https://hackmd.io/@rCqYwHC7RQeS4hwgARV_tg/SJSOb7GyP#Tie-breaker-rules)).

CC @sa8 @Kubuxu

arajasek:
See https://github.com/filecoin-project/FIPs/pull/156

Closes #2981

### DIFF
--- a/chain/store/store.go
+++ b/chain/store/store.go
@@ -424,7 +424,13 @@ func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipS
 		return err
 	}
 
-	if w.GreaterThan(heaviestW) {
+	heavier := w.GreaterThan(heaviestW)
+	if w.Equals(heaviestW) && !ts.Equals(cs.heaviest) {
+		log.Errorw("weight draw", "currTs", cs.heaviest, "ts", ts)
+		heavier = breakWeightTie(ts, cs.heaviest)
+	}
+
+	if heavier {
 		// TODO: don't do this for initial sync. Now that we don't have a
 		// difference between 'bootstrap sync' and 'caught up' sync, we need
 		// some other heuristic.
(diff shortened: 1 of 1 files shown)


chain/store/store.go
```diff
@@ -424,7 +424,13 @@ func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipS
 		return err
 	}
 
-	if w.GreaterThan(heaviestW) {
+	heavier := w.GreaterThan(heaviestW)
+	if w.Equals(heaviestW) && !ts.Equals(cs.heaviest) {
+		log.Errorw("weight draw", "currTs", cs.heaviest, "ts", ts)
```
magik6k, line 429: Could log the result of breakWeightTie

magik6k (approved):

chain/store/store.go
arajasek: yessir will do

Kubuxu: probably could downgrade it, it was error due to the lack of tie breaker

PR #7382, opened 2021-09-23
improve systemd

coryschwartz:
This adds an pre-exec script that runs prior to starting lotus for the first time, which I think makes it more useful for people to use.

Additionally, I made some edits to the Makefile so that "make install" and "make uninstall" take a destination directory. This makes it useful for installing multiple versions (if you really want to do that) but also it will be useful for packaging scripts.

https://www.gnu.org/prep/standards/html_node/DESTDIR.html#DESTDIR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,13 @@ $(warning Your Golang version is go$(shell expr $(GOVERSION) / 1000000).$(shell 
 $(error Update Golang to version to at least 1.16.0)
 endif
 
+DESTDIR?=/
+PREFIX?=/usr
+
+BINDIR=$(DESTDIR)$(PREFIX)/bin
+SYSTEMDDIR=$(DESTDIR)/usr/lib/systemd/system
+SHAREDIR=$(DESTDIR)$(PREFIX)/share
+
 # git modules that need to be loaded
 MODULES:=
 
@@ -54,6 +61,15 @@ $(MODULES): build/.update-modules ;
 build/.update-modules:
 	git submodule update --init --recursive
 	touch $@
+
+$(BINDIR):
+	mkdir -p $(BINDIR)
+
+$(SYSTEMDDIR):
+	mkdir -p $(SYSTEMDDIR)
+
+$(SHAREDIR):
+	mkdir -p $(SHAREDIR)
 
 # end git modules
 
@@ -122,19 +138,28 @@ an existing lotus binary in your PATH. This may cause problems if you don't run 
 
 .PHONY: build
 
-install: install-daemon install-miner install-worker
+install: install-daemon install-miner install-worker install-services install-completions
 
-install-daemon:
-	install -C ./lotus /usr/local/bin/lotus
+install-daemon: lotus $(BINDIR)
+	install -o root -g root -m 755 -s ./lotus $(BINDIR)/lotus
+	install -C -o root -g root -m 755 ./scripts/lotus-init.sh $(BINDIR)/lotus-init.sh
 
-install-miner:
-	install -C ./lotus-miner /usr/local/bin/lotus-miner
+install-miner: lotus-miner $(BINDIR)
+	install -o root -g root -m 755 -s ./lotus-miner $(BINDIR)/lotus-miner
 
-install-worker:
-	install -C ./lotus-worker /usr/local/bin/lotus-worker
+install-worker: lotus-worker $(BINDIR)
+	install -o root -g root -m 755 -s ./lotus-worker $(BINDIR)lotus-worker
 
-install-app:
-	install -C ./$(APP) /usr/local/bin/$(APP)
+uninstall-daemon:
+	rm $(BINDIR)/lotus
+
+uninstall-lotus-miner:
+	rm $(BINDIR)/lotus-miner
+
+uninstall-lotus-worker:
+	rm $(BINDIR)/lotus-worker
+
+uninstall-daemons: uninstall-daemon uninstall-lotus-miner uninstall-lotus-worker
 
 # TOOLS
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -55,6 +62,15 @@ build/.update-modules:
...
 
+$(BINDIR):
+	mkdir -p $(BINDIR)
+
+$(SYSTEMDDIR):
+	mkdir -p $(SYSTEMDDIR)
+
+$(SHAREDIR):
+	mkdir -p $(SHAREDIR)
+
```
Stebalien, line 73: modes?

Stebalien: And do we even need these? Can we not use `install -D`?

Makefile
```diff
@@ -122,19 +138,28 @@ an existing lotus binary in your PATH. This may cause problems if you don't run
 
 .PHONY: build
 
-install: install-daemon install-miner install-worker
+install: install-daemon install-miner install-worker install-services install-completions
+
+install-daemon: lotus $(BINDIR)
+	install -o root -g root -m 755 -s ./lotus $(BINDIR)/lotus
+	install -C -o root -g root -m 755 ./scripts/lotus-init.sh $(BINDIR)/lotus-init.sh
```
Stebalien, line 145: We shouldn't be specifying the user/group.

scripts/lotus-daemon.service
```diff
@@ -1,19 +1,17 @@
...
 Description=Lotus Daemon
-After=network-online.target
-Requires=network-online.target
+After=network.target
 
 [Service]
-Environment=GOLOG_FILE="/var/log/lotus/daemon.log"
-Environment=GOLOG_LOG_FMT="json"
+User=fc
+Group=fc
```
Stebalien, line 7: We can either:

- Use a dynamic user.
- Install a systemd-sysusers.

Also, we should call it lotus, not just "fc" (which could mean anything).

scripts/lotus-daemon.service
```diff
@@ -1,19 +1,17 @@
...
+ExecStartPre=/usr/local/bin/lotus-init.sh
 ExecStart=/usr/local/bin/lotus daemon
```
Stebalien, line 9: Wrong path?

scripts/lotus-daemon.service
```diff
@@ -1,19 +1,17 @@
...
+ExecStop=/usr/local/bin/lotus daemon stop
```
Stebalien, line 10: Is sending a signal not just as safe?

scripts/lotus-daemon.service
```diff
@@ -1,19 +1,17 @@
...
+Environment=LOTUS_PATH=/var/lib/lotus
 Restart=always
-RestartSec=10
-
-MemoryAccounting=true
-MemoryHigh=8G
-MemoryMax=10G
-LimitNOFILE=8192:10240
+RestartSec=30
+TimeoutSec=infinity
```
Stebalien, line 14: We definitely shouldn't set an infinite timeout sec

scripts/lotus-daemon.service
```diff
@@ -1,19 +1,17 @@
 [Unit]
 Description=Lotus Daemon
-After=network-online.target
-Requires=network-online.target
```
Stebalien, line 4: This was the correct version.

scripts/lotus-daemon.service
```diff
@@ -1,19 +1,17 @@
...
+After=network.target
 
 [Service]
-Environment=GOLOG_FILE="/var/log/lotus/daemon.log"
-Environment=GOLOG_LOG_FMT="json"
```
Stebalien, line 8: Why are we removing these?

scripts/lotus-daemon.service
```diff
@@ -1,19 +1,17 @@
...
+User=fc
+Group=fc
+ExecStartPre=/usr/local/bin/lotus-init.sh
```
Stebalien, line 8: Can we just pass `--import-chain` to the daemon? I believe that should import iff not already initialized.

scripts/lotus-daemon.service
Kubuxu: it is

scripts/lotus-daemon.service
Kubuxu: --import-chain always imports

Stebalien:
> --import-chain always imports afaik

`--import-snapshot` does, but the help says `--import-chain` doesn't?

scripts/lotus-daemon.service
coryschwartz: Sure, this can change it to lotus. There is already "fc" being used elsewhere, so I was continuing to do that, but I have no preference for it and I do think calling it lotus is better. I would not be opposed to systemd-sysuers either.

scripts/lotus-daemon.service
coryschwartz: It has to be pretty long to take into account the chain import. Maybe infinity isn't right, but I am not sure how long is reasonable before it becomes clear that an error has occurred, so instead I set this to infinity and rely on the start script to indicate if there's been an error.


I considered just not including long ExecStartPre at all, but I think do remove it could result in a poorer experience for people, particularly if when they install through packages. If you install an application, you expect it to be working correctly within a reasonable time-frame. Without importing the chain at the start, users are left either to wait for the chain to sync over the network or to find in the documentation that chain imports are possible and do it on their own. 

I think the way to do this might be to handle the import differently. Maybe --import-snapshot/--import-chain would only do do the import if it wasn't already done, or force to import again with --force-import-snapshot flags. 


Actually, I'll move this PR back into draft and come up with a better fix for this that doesn't require the intiialization script, because I think it does cause some problems, some of which are described in these comments.

scripts/lotus-daemon.service
coryschwartz: I suppose this is just personal preference, since systemd already has a mechanism for handling journal output, and this doesn't include a logrotate configuration. I can add this back, though it might eventually fill up disk space with the log file un-handled.

scripts/lotus-daemon.service
coryschwartz: import-chain will force a validation back to ts-0 
https://github.com/filecoin-project/lotus/blob/master/cmd/lotus/daemon.go#L529

running --import-chain against minimal_finality_stateroots does result in a validation error. Although aside from this, I don't think there is a difference between the two import procedures.

scripts/lotus-daemon.service
lanzafame: @coryschwartz Why not perform a oneshot service that does the import and have that be a prerequisite for starting the daemon service?

scripts/lotus-daemon.service
Stebalien: Fair enough. You're right, we might as well keep it simple in the default config. The user can always use a drop-in.

scripts/lotus-daemon.service
Stebalien: Ah... That's, unfortunate.

scripts/lotus-daemon.service
Stebalien: Could even use a condition to determine if it needs to run.

scripts/lotus-daemon.service
Stebalien: Hm. Where are we using "fc"? I'm just concerned that users will be confused if they later look at their passwd file and find "fc".

Makefile
Stebalien: The deb package should be run with fakeroot, right? Is the default user not correct?

rjan90:
Closing this PR as stale/outdated.

PR #7389, opened 2021-09-27
Fix calculation of Drand round from Filecoin epochs

### LINKED ISSUE #2170: Drand rounds are not computed as drand computes them
**Describe the bug**

Lotus doesn't use the same "timestamp-to-round" mapping than drand:
Drand does the following ([code](https://github.com/drand/drand/blob/master/chain/time.go#L25)):
```go
math.Floor((now - genesis) / drandPeriod) + 1 
```
because round 1 happens at genesis time.
Lotus however does the following ([code](https://github.com/filecoin-project/lotus/blob/master/chain/beacon/drand/drand.go#L188)) :
```go
math.Floor((now - genesis) / drandPeriod) 
```
It's a breaking change to move to drand, but it would be the preferred way. 

**Consequences**: the lotus chain uses a round before the one it should use, having an offset of 2 instead of 1 (i.e. normally it looks at a drand round for one epoch before, but it's in fact using the randomness of one more previous drand round).

arajasek:
Closes #2170

Opened against dev branch for now, will move to be against feat/nv14 when #7376 lands


chain/beacon/beacon.go
```diff
@@ -42,10 +44,10 @@ type BeaconPoint struct {
 type RandomBeacon interface {
 	Entry(context.Context, uint64) <-chan Response
 	VerifyEntry(types.BeaconEntry, types.BeaconEntry) error
-	MaxBeaconRoundForEpoch(abi.ChainEpoch) uint64
+	MaxBeaconRoundForEpoch(network.Version, abi.ChainEpoch) uint64
```
nikkolasg, line 47: Shouldn't we use https://github.com/filecoin-project/lotus/pull/7389/files#diff-587eaf0df6b60dbf741d19bbe39439f6f48ffe171e82a71c76b82484ba48f386R23-R38 for determining the right version ? Now this PR is putting the both logic (1) select the right version (2) return the right randomness in the same method impacting all functions, while the Schedule seems to have been made to clearly separate the logics so it's easy to upgrade later on (without changing necessarily the API of  RandomBeacon or the rest of the code)

chain/beacon/drand/drand.go
```diff
@@ -201,11 +204,32 @@ func (db *DrandBeacon) VerifyEntry(curr types.BeaconEntry, prev types.BeaconEntr
...
+
+	return db.maxBeaconRoundV2(latestTs)
+}
+
+func (db *DrandBeacon) maxBeaconRoundV1(latestTs uint64) uint64 {
 	dround := (latestTs - db.drandGenTime) / uint64(db.interval.Seconds())
 	return dround
 }
 
+func (db *DrandBeacon) maxBeaconRoundV2(latestTs uint64) uint64 {
```
nikkolasg, line 223: While the function is correct, I'm simply suggesting that it may make sense to simply take the one from drand implementation ? https://github.com/drand/drand/blob/master/chain/time.go#L52 but no worries

nikkolasg (approved):
non-blocking comments so LGTM !

chain/beacon/beacon.go
Kubuxu: Link doesn't work :/

Kubuxu: but reflecting nikkolasg comment, should we just use beacon schedule and upgrade the beacon to another "version" of itself, instead of adding this in?

arajasek:
Closing for now -- Chocolate has enough changing that we shouldn't be squeezing this in-

PR #7397, opened 2021-09-28
cli command to get retrieval ask

dirkmc:
This PR needs to be reworked to use the "Query" protocol to get the dynamic retrieval ask (instead of using a new ask protocol, which is unnecessary)


dirkmc:
Superseded by https://github.com/filecoin-project/lotus/pull/7814

PR #7398, opened 2021-09-28
Lotus extended pubsub tracer

mpetrun5:
Separates lp2p tracer and lotus tracer and adds ability to track peer scores with lotus tracer.

Lotus tracer has support for generic transports and implemented are: 
 - json file transport that saves locally
 - elasticsearch transport that sends to default or custom elasticsearch instance index

Couldn't figure out how to override config value, with the value passed through CLI flag and wondering is it even necessary? The idea was to have some property defined as part of the configuration (in node/config/types) and then to add an additional CLI flag in cmd/lotus/daemon.go that will override config value if present. I thought that an example of this behavior is the bootstrapper property that calls override function when creating a new node - node.Override(new(dtypes.Bootstrapper), isBootstrapper), but I wasn't able to implement it this way.

Kibana dashboards configurations can be exported as JSON-like files, but we are not sure if we should keep these files as part of the repo or they should be saved separately, I would like to hear your thoughts on this.

Example of dashboards:

Block propagation displayed as forced layout graph where link force is block propagation average time between peers:

image

Peer score table:

The weighted score is calculated via [Kibana runtime fields](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/runtime.html) when correct source auth value is sent in trace event.

image


cmd/lotus/daemon.go
```diff
@@ -153,6 +154,18 @@ var DaemonCmd = &cli.Command{
...
+			Usage: "starts tracer and outputs to json file defined with this flag",
+		},
+		&cli.StringFlag{
+			Name:  "trace-to-elasticsearch",
+			Usage: "starts tracer and outputs to elasticsearch, flag must contain connection string for elasticsearch",
+		},
+		&cli.StringFlag{
+			Name:  "trace-source-auth",
+			Usage: "auth token for trusted source of traces",
+		},
```
magik6k, line 168: We should use the node config instead of defining special flags for this. If node operators really want to pass those through the CLI, we have envconfig - https://github.com/filecoin-project/lotus/blob/964df1a2ecc38bf5ab721b7bb6a03e603eb1c98e/node/config/load.go#L41 - https://github.com/kelseyhightower/envconfig

node/config/types.go
```diff
@@ -309,6 +309,10 @@ type Pubsub struct {
 	DirectPeers           []string
 	IPColocationWhitelist []string
 	RemoteTracer          string
+	JsonTracer            string
+	ElasticSearchTracer   string
+	ElasticSearchIndex    string
+	TracerSourceAuth      string
```
magik6k, line 315: Needs to be hooked up to things in node/builder. Also would be great to add docstrings for those, as we then use those to generate doc comments in the config file

node/modules/tracer/tracer.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (lt *lotusTracer) PeerScores(scores map[peer.ID]*pubsub.PeerScoreSnapshot) {
+	now := time.Now().UnixNano()
+	for pid, score := range scores {
+		evt := &LotusTraceEvent{
+			Type:       *TraceEvent_PEER_SCORES.Enum(),
+			PeerID:     lt.pid.Pretty(),
+			Timestamp:  &now,
+			SourceAuth: lt.sa,
+			PeerScore:  TraceEvent_PeerScore{PeerID: pid.Pretty(), Score: float32(score.Score)},
```
vyzo, line 60: We can get some richer data here, it would be valuable to see the score components as well.

cmd/lotus/daemon.go
MakMuftic: I agree with you, I removed CLI flags (and we already had all these prop. as part of node config)

vyzo (approved):

PR #7405, opened 2021-09-30
Update to graphsync to v0.10.0, enable seperate storage and retrieval transfer limits

hannahhoward:
# Goals

Update go-graphsync v0.10.0

Add support for seperate transfer limits across storage and retrieval, aiming to provide initial solutions for https://github.com/filecoin-project/lotus/pull/7276

Blockers:
- [x] update to final go-graphsync v0.10.0 (pending use and validation of graphsync/do-not-send-first-blocks in go-data-transfer)
- [x] verify no issues with renaming SimultaneousTransfers, or clear upgrade path

# Implementation

- update go-graphsync to v0.10.0 with outgoing request limits
- rename configuration SimultaneousTransfers -> SimultaneousTransfersForStorage, SimultaneousTransfersForRetrieval
- configure request limists accordingly for graphsync (note these are opposite between client + miner see table below)

# For discussion

Mapping role and deal type to graphsync request type:

Role | DealType | GS Request Type
---- | --------- | ------------------
Client | Storage | Incoming
Client | Retrieval | Outgoing
Provider | Storage | Outgoing
Provider | Retrieval | Incoming


rvagg (approved):
@jennijuju: it'd be worth calling out in release notes that `SimultaneousTransfers` stops working entirely here and there's no migration or repurposing. If you were using that then you need to update to have both `SimultaneousTransfersForStorage` and `SimultaneousTransfersForRetrieval`.

rvagg:
Rebased to current master. We've lost a bit of codecov here and it's red, but it comes back up via #7427 which is on top of this branch.

jennijuju:
> @jennijuju: it'd be worth calling out in release notes that `SimultaneousTransfers` stops working entirely here and there's no migration or repurposing. If you were using that then you need to update to have both `SimultaneousTransfersForStorage` and `SimultaneousTransfersForRetrieval`.

Ack! Should we recommend SP to finish/stop all transfers before the upgrade then?

rvagg:
> Ack! Should we recommend SP to finish/stop all transfers before updating then?

Oh, I'm not sure about that; I suppose it won't matter, the upgrade will bring better restart functionality as well as these limits, but the change in config options shouldn't be a major change in behaviour, just an annoying config change if it's been configured. Maybe @dirkmc can chime in on this.

dirkmc:
I don't think we need to explicitly ask storage providers to stop data transfers before upgrading

jennijuju:
> I don't think we need to explicitly ask storage providers to stop data transfers before upgrading

So when they restart the node with X ongoing transfer, where X  > SimultaneousTransfersForStorage, new deals will just be rejected and the previously ongoing transfer can be resumed?

rvagg:
@jennijuju I don't know the answer to that but I'm assuming that would be correct. Old default is 20, new default is 20, so the only surprises here would be if an SP has changed the number on their end, they would need to set both `SimultaneousTransfersForStorage` and `SimultaneousTransfersForRetrieval` to that number to achieve the same effect as their old config.

dirkmc (approved):

Issue #5137, opened 2020-12-07
Expired deals continue to cause new sectors to be removed


stuberman:
**Describe the bug**
Many older deals still in a status in lotus-miner sectors list where they are causing sealing problems.
An example is an expired deal from Dec 4 continues to be added to new sectors with other older deals and the sector is removed at the end of PC2

**To Reproduce**
1. While my miner is idle, it will create a new sector and add old expired deals.
2. See error below:

`lotus-miner sectors status --on-chain-info --log 365`
SectorID:	365
Status:		Removed
CIDcommD:	baga6ea4seaqfzritndxvdh6semu7aqk3snjvtmehftfatveisw3xjn2egr4t6gq
CIDcommR:	bagboea4b5abcb4fgqxzpqancc4ir4hl4gbtboujwvz4xfqngf4ooqqerq3tdduqf
Ticket:		91967c6a19678b0ae9d7e210e03e1063dfbb256b0dcf21acad81866ef5e56b20
TicketH:	300445
Seed:		
SeedH:		0
Precommit:	<nil>
Commit:		<nil>
Proof:		
Deals:		[1431209 1431396 1431415 0 0 0]
Retries:	0

Sector On Chain Info
SealProof:		0
Activation:		0
Expiration:		0
DealWeight:		0
VerifiedDealWeight:		0
InitialPledge:		0

Expiration Info
OnTime:		0
Early:		0
--------
Event Log:
0.	2020-12-07 08:06:34 +0000 UTC:	[event;sealing.SectorStart]	{"User":{"ID":365,"SectorType":8}}
1.	2020-12-07 08:07:17 +0000 UTC:	[event;sealing.SectorAddPiece]	{"User":{"NewPiece":{"Piece":{"Size":2147483648,"PieceCID":{"/":"baga6ea4seaqjnupg2f6aawhbsslxfdexlwzvkjtc2gslqsaa4zcxgqykmacrkiy"}},"DealInfo":{"PublishCid":{"/":"bafy2bzaceabvuanwxzw4qawrlr4rnldy4utuevsyjnb6rwyik7uji5oxpyfzy"},"DealID":1431209,"DealSchedule":{"StartEpoch":301874,"EndEpoch":1341228},"KeepUnsealed":false}}}}
[...]

PR #7407, opened 2021-09-30
check for deal start epoch on SectorAddPieceToAny

nonsense:
Should fix: https://github.com/filecoin-project/lotus/issues/5137

---

This PR is:
1. Adding a check for deal start epoch in `SectorAddPieceToAny`.
2. Adding an epoch buffer `StartEpochSealingBuffer` to both `SectorAddPieceToAny`, as well as to `validateDeal` prior to sending a `PublishStorageDeals` message, to account for some time necessary to seal a sector.

At the moment I have added `StartEpochSealingBuffer` to be 480 by default == which is 4 hours, and I think this is reasonable. I am not sure if there are miners out there that go from `PublishStorageDeals` to a sealed sector faster than 4 hours...

### DIFF
--- a/documentation/en/default-lotus-miner-config.toml
+++ b/documentation/en/default-lotus-miner-config.toml
@@ -207,6 +207,12 @@
   # env var: LOTUS_DEALMAKING_SIMULTANEOUSTRANSFERS
   #SimultaneousTransfers = 20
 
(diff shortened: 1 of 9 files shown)


magik6k:
(rebased to update config docs)

magik6k (approved):

PR #7414, opened 2021-09-30
shed: simple wallet balancer util

magik6k:
Small util to automatically manage balances in multiple wallets. Useful with multiple control addresses.

Maybe someday this will become it's own thing / live in lotus, but for now shed seems to be a good enough place.

### DIFF
--- a/cmd/lotus-shed/balancer.go
+++ b/cmd/lotus-shed/balancer.go
@@ -1,0 +1,222 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+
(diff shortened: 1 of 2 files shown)


cmd/lotus-shed/balancer.go
```diff
@@ -0,0 +1,222 @@
...
+				}
+
+				if bal.Balance.LessThan(req.low) {
+					toSend = append(toSend, &send{
+						to:  req.addr,
+						amt: big.Sub(req.high, bal.Balance),
+					})
+				}
+			}
+
```
arajasek, line 169: Short-circuit if no `toSend`s?

arajasek (approved):

cmd/lotus-shed/balancer.go
magik6k: Computers are fast

PR #7427, opened 2021-10-02
Limit maximum number of DAG links to traverse

hannahhoward:
# Goals

Set reasonable limits on number of links traversed while calculating commP & doing go-graphsync transfers

Blockers:
- [x] update to tagged go-ipld-prime 0.12.3
- [x] merge https://github.com/filecoin-project/go-fil-markets/pull/633
- [x] Add https://github.com/filecoin-project/lotus/pull/7428
- [x] update to tagged go-fil-markets
- [x] rebase on master and change the base for this PR when https://github.com/filecoin-project/lotus/pull/7405 merges with tagged go-graphsync
- [x] Validate the error condition locally by setting the environment variable low.  The goals here are: 1) Confirm the limit works 2) See what the actual error message in the logs is when we hit the limit.  Paste the error message in this PR.
- [x] Sanity check by running a manual storage and retrieval deal against production

# Implementation

- update go-graphsync and go-car and go-fil-markets
- set a config.MaxTraversalLinks
- allow it to be overridden via ENV
- pass to initialization code for markets & go-graphsync

### DIFF
--- a/node/modules/client.go
+++ b/node/modules/client.go
@@ -36,6 +36,7 @@ import (
 	"github.com/filecoin-project/lotus/markets"
 	marketevents "github.com/filecoin-project/lotus/markets/loggers"
 	"github.com/filecoin-project/lotus/markets/retrievaladapter"
+	"github.com/filecoin-project/lotus/node/config"
 	"github.com/filecoin-project/lotus/node/impl/full"
 	payapi "github.com/filecoin-project/lotus/node/impl/paych"
 	"github.com/filecoin-project/lotus/node/modules/dtypes"
@@ -182,7 +183,7 @@ func StorageClient(lc fx.Lifecycle, h host.Host, dataTransfer dtypes.ClientDataT
 	marketsRetryParams := smnet.RetryParameters(time.Second, 5*time.Minute, 15, 5)
 	net := smnet.NewFromLibp2pHost(h, marketsRetryParams)
 
-	c, err := storageimpl.NewClient(net, dataTransfer, discovery, deals, scn, accessor, storageimpl.DealPollingInterval(time.Second))
+	c, err := storageimpl.NewClient(net, dataTransfer, discovery, deals, scn, accessor, storageimpl.DealPollingInterval(time.Second), storageimpl.MaxTraversalLinks(config.MaxTraversalLinks))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


node/modules/client.go
```diff
@@ -182,7 +183,7 @@ func StorageClient(lc fx.Lifecycle, h host.Host, dataTransfer dtypes.ClientDataT
 	marketsRetryParams := smnet.RetryParameters(time.Second, 5*time.Minute, 15, 5)
 	net := smnet.NewFromLibp2pHost(h, marketsRetryParams)
 
-	c, err := storageimpl.NewClient(net, dataTransfer, discovery, deals, scn, accessor, storageimpl.DealPollingInterval(time.Second))
+	c, err := storageimpl.NewClient(net, dataTransfer, discovery, deals, scn, accessor, storageimpl.DealPollingInterval(time.Second), storageimpl.MaxTraversalLinks(config.MaxTraversalLinks))
```
rvagg, line 186: Is this the only place that graphsync is initiated? It's the only one I could find but do we have other avenues, like go-datatransfer? Otherwise we're relying on unbounded graphsync

rvagg:
https://github.com/filecoin-project/lotus/pull/7428 was intended to be in here as well, there's a couple of `SelectiveCar` uses locally that aren't deferred via gfm.

rvagg:
Rebased onto current #7405 which has been rebased to master.

rvagg:
Pulled in go-fil-markets@v1.13.0 which pulls in the newly tagged go-data-transfer@v1.11.0 which therefore incorporates #7441 into this branch.

rvagg:
So, the easiest test of this I think is to start the daemon with `LOTUS_MAX_TRAVERSAL_LINKS=10 lotus daemon`

Importing a file of decent size (in this case just the built `lotus` binary) using: `lotus client generate-car <input> <output>` yields:

```
ERROR: failed to write CAR to output file: traversal budget exceeded: budget for links reached zero while on path "Links/10/Hash" (link: "bafk2bzacecdzw7j4fcqxwcgad7zb4rehej7uqa4a2263dhoglkcgt465p32qy")
```

There are some other places that are a bit more difficult to run manual tests on without an operating storage network.

dirkmc:
#### Version Compatibility Testing
Old: lotus v1.11.3
New: _This branch_

- New client / Old server
  - Storage ✅ 
  - Retrieval ✅ 
- New client / New server
  - Storage ✅ 
  - Retrieval ✅ 
- Old client / New server
  - Storage ✅ 
  - Retrieval ✅

PR #7428, opened 2021-10-02
traversals: limit link load count for SelectiveCar use as well

rvagg:
@hannahhoward I forgot this bit was meant to be in the original changeset too


rvagg:
pulled in to #7427

PR #7440, opened 2021-10-04
Chocolate to master

jennijuju:


arajasek (approved):
Idk whether the changelog should be going in (and if it is, should it be 1.12.0-rc1?)

but idc :P

PR #7444, opened 2021-10-04
sealing: Recover sectors after failed AddPiece

magik6k:


extern/storage-sealing/input.go
```diff
@@ -245,9 +245,7 @@ func (m *Sealing) handleAddPiece(ctx statemachine.Context, sector SectorInfo) er
 }
 
 func (m *Sealing) handleAddPieceFailed(ctx statemachine.Context, sector SectorInfo) error {
-	log.Errorf("No recovery plan for AddPiece failing")
-	// todo: cleanup sector / just go retry (requires adding offset param to AddPiece in sector-storage for this to be safe)
```
magik6k, line 249: Review note: we do have offset in AddPiece already, calculated from the existingPieces thing which only has valid pieces

arajasek (approved):

PR #7441, opened 2021-10-04
use go-fil-markets v1.13.0

dirkmc:
This version of go-fil-markets has [a fix](https://github.com/filecoin-project/go-data-transfer/pull/257) that makes restarting data transfers more reliable.

TODO:
- [x] Storage deal restart testing
   OLD: lotus v1.11.3
   NEW: _this branch_
  - [x] New client / Old server  ✅
  - [x] New client / New server  ✅
  - [x] Old client / New server  ✅
- [x] Release go-ipld-prime v0.12.3 and bubble up to lotus
- [ ] Release go-graphsync v0.10.0 (currently this branch is using v0.10.0-rc3)

PR #7446, opened 2021-10-04
Expose basic text-based datamodel selector on retrieval

ribasushi:
This PR is effectively identical to https://github.com/filecoin-project/lotus/pull/6393, but **does not** include https://github.com/filecoin-project/lotus/pull/7441 - it only pulls in the newest go-ipld-prime.

Syntaxt of selection is located at
https://pkg.go.dev/github.com/ipld/go-ipld-selector-text-lite#SelectorSpecFromPath

Example use, assuming that:
  - The root of the deal is a plain dag-pb unixfs directory
  - The directory is not sharded
  - The user wants to retrieve the first entry in that directory

  lotus client retrieve --miner f0XXXXX --datamodel-path-selector 'Links/0/Hash' bafyROOTCID ~/output

For a much more elaborate example see the top of ./itests/deals_partial_retrieval_test.go


ribasushi:
Closing as per conversation with @jennijuju: we are going with https://github.com/filecoin-project/lotus/pull/6393 as soon as he GS upgrade lands.

PR #7450, opened 2021-10-06
StateSearchMsg: Correct usage of the allowReplaced flag

### LINKED ISSUE #7009: If miner replaces publish deal message, it does not update that in its deal state
If a miner publishes a deal, but then does an mpool replace on that message before it lands (to increase the gas limit, or something) the cid recorded in their markets state store for the deal does not get updated. This leads to clients never really finding out that their data has made it into a deal, and leads to a lot of faults in estuary (estuary will mark the deal as failed on its end because it has no way of knowing that it actually landed on chain).

We can work around this by just waiting for the miner to update the deal ID in their deal status response, but this requires a good deal of extra logic to validate that everything is correct.

arajasek:
Fixes #7009 

Also fixes all uses of `StateSearchMsg` with `allowReplaced` set to false.

### DIFF
--- a/chain/gen/gen.go
+++ b/chain/gen/gen.go
@@ -297,6 +297,10 @@ func NewGeneratorWithUpgradeSchedule(us stmgr.UpgradeSchedule) (*ChainGen, error
 	return NewGeneratorWithSectorsAndUpgradeSchedule(1, us)
 }
 
+func (cg *ChainGen) Blockstore() blockstore.Blockstore {
+	return cg.bs
+}
+
 func (cg *ChainGen) StateManager() *stmgr.StateManager {
 	return cg.sm
 }
--- a/chain/stmgr/searchwait.go
+++ b/chain/stmgr/searchwait.go
@@ -252,22 +252,27 @@ func (sm *StateManager) tipsetExecutedMessage(ts *types.TipSet, msg cid.Cid, vmm
 
 		if m.VMMessage().From == vmm.From { // cheaper to just check origin first
 			if m.VMMessage().Nonce == vmm.Nonce {
-				if allowReplaced && m.VMMessage().EqualCall(vmm) {
-					if m.Cid() != msg {
-						log.Warnw("found message with equal nonce and call params but different CID",
-							"wanted", msg, "found", m.Cid(), "nonce", vmm.Nonce, "from", vmm.From)
-					}
-
-					pr, err := sm.cs.GetParentReceipt(ts.Blocks()[0], i)
-					if err != nil {
-						return nil, cid.Undef, err
-					}
-					return pr, m.Cid(), nil
+				if !m.VMMessage().EqualCall(vmm) {
+					// this is an entirely different message, fail
+					return nil, cid.Undef, xerrors.Errorf("found message with equal nonce as the one we are looking for that is NOT a valid replacement message (F:%s n %d, TS: %s n%d)",
+						msg, vmm.Nonce, m.Cid(), m.VMMessage().Nonce)
 				}
 
-				// this should be that message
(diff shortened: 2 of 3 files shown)


jennijuju (approved):

arajasek:
> Can use a test 🐱🐱

done

testplans/lotus-soup/go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/ipfs/go-log/v2 v2.3.0
 	github.com/ipfs/go-merkledag v0.3.2
 	github.com/ipfs/go-unixfs v0.2.6
-	github.com/ipld/go-car v0.3.1-null-padded-files
+	github.com/ipld/go-car v0.3.2-0.20211001225732-32d0d9933823
```
Stebalien, line 29: :rage:

Stebalien: This is not a commit on master, and is not tagged.

chain/stmgr/searchwait.go
```diff
@@ -252,22 +252,27 @@ func (sm *StateManager) tipsetExecutedMessage(ts *types.TipSet, msg cid.Cid, vmm
...
+					// this is an entirely different message, fail
+					return nil, cid.Undef, xerrors.Errorf("found message with equal nonce as the one we are looking for that is NOT a valid replacement message (F:%s n %d, TS: %s n%d)",
+						msg, vmm.Nonce, m.Cid(), m.VMMessage().Nonce)
+				}
 
-					pr, err := sm.cs.GetParentReceipt(ts.Blocks()[0], i)
-					if err != nil {
-						return nil, cid.Undef, err
+				if m.Cid() != msg {
+					log.Warnw("found message with equal nonce and call params but different CID",
```
Stebalien, line 262: should only warn if we _don't_ allow replaced. Otherwise, the user will see scary messages.

chain/gen/gen.go
```diff
@@ -297,6 +297,10 @@ func NewGeneratorWithUpgradeSchedule(us stmgr.UpgradeSchedule) (*ChainGen, error
 	return NewGeneratorWithSectorsAndUpgradeSchedule(1, us)
 }
 
+func (cg *ChainGen) Blockstore() blockstore.Blockstore {
+	return cg.bs
+}
```
Stebalien, line 302: Eh?

Stebalien (commented):
Mostly LGTM. Except for the deps and the unrelated change.

testplans/lotus-soup/go.mod
arajasek: then ytf does it magically get changed i hate gomod idk how anything works is it my ide doing this??

chain/stmgr/searchwait.go
arajasek: fair

chain/gen/gen.go
Stebalien: Ah. SGTM.

testplans/lotus-soup/go.mod
Stebalien: :shrug:? What go version?

Stebalien: Ah, I see... It was already broken, the go mod just wasn't updated.

arajasek: what was broken? i honestly dunno how any of this works, i'm probably the one who broke it.

Stebalien: Updated to a non-tagged commit not on master. It's not broken _yet_, but will be if the branch the commit lives on gets a force-push (which happens _frequently_ in this particular branch).

Stebalien (approved):

PR #7453, opened 2021-10-06
Sector storage groups

magik6k:
This PR adds two new optional fields to `sectorstore.json`:
* `Groups []string` - list of group names the storage path belongs to.
* `AllowTo []string` - list of group names to which sectors can be fetched to from this storage path.

Those two should allow for creating worker groups, and avoiding unnecesarily moving data between multi-purpose workers. For example in the following setup:

* Storage miner
  * Path with `Groups: ["example-storage-group-1"]` 
* Worker 1 (PC1, PC2):
  * Path with `Groups: ["example-seal-group-1"], AllowTo: ["example-seal-group-1"]`
* Worker 2 (PC1, PC2):
  * Path with `Groups: ["example-seal-group-2"], AllowTo: ["example-seal-group-2"]`
* Worker 3 (PC1):
  * Path with `AllowTo: ["example-seal-group-1""]`

Without storage groups, PC2 tasks on workers 1/2 could be scheduled with sector data from other workers, which would often waste bandwidth and occasionally block processing on fetching data.
With storage groups configured as above, sectors which had PC1 done on worker1 / worker2 will always execute PC2 on the same worker. Sectors from worker3 will only go to worker1 for PC2

Groups can be setup in two ways:
* For now storage paths, when using the `storage attach --init` command with the new `--groups` / `--allow-to` flags
* For existing storage paths - by modifying `[path]/sectorstore.json`, then restarting lotus-miner/worker

Groups:
```
{
  "ID": "74e1d667-7bc9-49bc-a9a6-0c30afd8684c",
  "Weight": 10,
  "CanSeal": false,
  "CanStore": true,
  "MaxStorage": 0,
  "Groups": ["storage0"]
}
```

AllowTo:
```
{
  "ID": "54f090bf-9fa9-4a47-9f67-0cc4f24c810e",
  "Weight": 10,
  "CanSeal": true,
  "CanStore": false,
  "AllowTo": ["storage0"]
}
```

TODO:
* [x] CLI Utils
  * [x] Show in storagelist
  * [x] Storage attach --init flags
* [x] Test on some miner setups
* [x] Test with minerx miners


arajasek (approved):

cmd/lotus-miner/storage.go
```diff
@@ -95,6 +95,14 @@ over time
 			Name:  "max-storage",
 			Usage: "(for init) limit storage space for sectors (expensive for very large paths!)",
 		},
+		&cli.StringSliceFlag{
+			Name:  "groups",
+			Usage: "path group names",
+		},
+		&cli.StringSliceFlag{
+			Name:  "allow-to",
+			Usage: "group names to which data from this path can be sent (allow all if not specified)",
```
jennijuju, line 104:
```suggestion
			Usage: "group names to which can pull data from this path (allow all if not specified)",
```

cmd/lotus-seal-worker/storage.go
```diff
@@ -51,6 +51,14 @@ var storageAttachCmd = &cli.Command{
 			Name:  "max-storage",
 			Usage: "(for init) limit storage space for sectors (expensive for very large paths!)",
 		},
+		&cli.StringSliceFlag{
+			Name:  "groups",
+			Usage: "path group names",
+		},
+		&cli.StringSliceFlag{
+			Name:  "allow-to",
+			Usage: "group names to which data from this path can be sent (allow all if not specified)",
```
jennijuju, line 60:
```suggestion
			Usage: "group names to which can pull data from this path (allow all if not specified)",
```

Issue #7102, opened 2021-08-17
Allow manual retry of deal publishing


f8-ptrk:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

lotus miner/market - storage deal

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Aug 17 07:37:02  true      bafyreictx7dwv4u5wdrqakwcfyahslpgrnccilr3kmz4k3gwhi6vz4wt4a  0       StorageDealError  f1wdxdpqh3hirrhp353i4o6ld7bsw6evh3v7i5jtq                                               32GiB   0 FIL                     520216    12D3KooWQKcgS3UBAG9WNJiV57iXRMx9bY2jD86aPPb5Y
NWvVEJe-12D3KooWMyP6mT7SC1AqnLqxzQQASuWooqnCrzBGXgUtQPsaERqq-1629176818622218181  error calling node: publishing deal: mpool push: failed to push message: not enough funds (required: 0.049999999994707184 FIL, balance: 0.010645432316105853 FIL): not enough funds to execute transaction

this could be recovered easily - right now its a 100% fail

### Describe the solution you'd like

a retry failed publish deals/deal command

neondragon:
+1
It would be great if Lotus could better differentiate between recoverable and unrecoverable storage deal errors. Currently when a PublishStorageDeals message fails for any reason, all deals published in the message are destroyed by Lotus, even when only one deal in the message has caused an error (such as insufficient DataCap). Deals and data are lost unnecessarily.

Unrecoverable errors -> StorageDealFailing -> StorageDealError
Potentially recoverable publishing errors -> perhaps transition to StorageDealPublishFailed and provide a 'storage-deals recover' and 'storage-deals cancel' command to allow the operator to push the deal back into the publish queue.

Automatically re-queueing and retrying publishing deals would be even better, but only where there is no chance of a loop of failing messages.

jennijuju:
ideally lotus should do #7368 by default

and then give the option for a manual retry

PR #7454, opened 2021-10-06
itests: retry deal when control addr is out of funds

nonsense:
Should fix: https://github.com/filecoin-project/lotus/issues/7102

Related: https://github.com/filecoin-project/go-fil-markets/pull/638

---

I am trying to come up with a solution that doesn't involve adding another 1-2 states to `go-fil-transfer` and another 1-2 events and adding more state transitions.

I think a deal being stuck in the `Publish` state is the natural thing to happen in case we fail to transition to the `Publishing` state, i.e. we fail to initiate the PSD message.

Having discussed within the #ignite team, we decided against automatically trying to retry PSD messages. With this PR, a PSD message would be retried only if the node is bounced, as a `Restart` event is sent to deals that are not in a terminal state.

---

TODO:
- [x] add test case to handle case where we have very little FIL and actor fails differently
- [x] add test case to confirm that deal blocks in `Publish` state, unless we call `MarketRetryPublishDeal`
- [ ] update `go-fil-markets` ref when related PR is merged so that we don't refer to a commit that's not on `master`

### DIFF
--- a/itests/deals_retry_deal_no_funds_test.go
+++ b/itests/deals_retry_deal_no_funds_test.go
@@ -169,7 +169,7 @@ func TestDealsRetryLackOfFunds_blockInPublishDeal(t *testing.T) {
 
 	dealSealed := make(chan struct{})
 	go func() {
-		dh.WaitDealSealed(ctx, deal, false, false, nil)
+		dh.WaitDealSealedQuiet(ctx, deal, false, false, nil)
 		dealSealed <- struct{}{}
 	}()
 
--- a/itests/kit/deals.go
+++ b/itests/kit/deals.go
@@ -177,6 +177,41 @@ loop:
 	}
 }
 
+// WaitDealSealedQuiet waits until the deal is sealed, without logging anything.
+func (dh *DealHarness) WaitDealSealedQuiet(ctx context.Context, deal *cid.Cid, noseal, noSealStart bool, cb func()) {
+loop:
+	for {
+		di, err := dh.client.ClientGetDealInfo(ctx, *deal)
+		require.NoError(dh.t, err)
+
+		switch di.State {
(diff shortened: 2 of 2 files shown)


cmd/lotus-miner/market.go
```diff
@@ -910,6 +911,34 @@ var dealsPendingPublish = &cli.Command{
...
 }
 
+var dealsRetryPublish = &cli.Command{
+	Name:  "retry-publish",
+	Usage: "retry publishing a deal",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name: "proposal-cid",
+		},
+	},
```
dirkmc, line 921: Maybe simpler to make this a parameter rather than a flag.
```
lotus-miner storage-deals retry-publish <proposal cid>
```
instead of
```
lotus-miner storage-deals retry-publish --proposal-cid=<proposal cid>
```

itests/deals_retry_deal_no_funds_test.go
```diff
@@ -0,0 +1,107 @@
...
+	dp := dh.DefaultStartDealParams()
+	dp.Data.Root = res.Root
+	dp.FastRetrieval = true
+	dp.EpochPrice = abi.NewTokenAmount(62500000) // minimum asking price.
+	deal := dh.StartDeal(ctx, dp)
+
+	propcid := *deal
+
+	go func() {
+		time.Sleep(20 * time.Second)
```
dirkmc, line 96: Can we subscribe to deal events, and watch for publish, rather than using a sleep here? (or maybe subscribe to deal events and use a shorter sleep)

nonsense: I will certainly use a shorter sleep. We can monitor for Publish, but you can't really tell if the Publish handler did something or not, and this is really what we are interested in.

PR #7467, opened 2021-10-07
Normlize selector use within lotus

ribasushi:


rvagg (approved):
I've been slightly hesitant to do because it's not precisely the same selector, but functionally should be.

We could simplify further by cleaning up the textselector API I reckon, but that's not critical and I think we should work on integrating that functionality into go-ipld-prime and clarifying the API when doing that.

magik6k (approved):

PR #7468, opened 2021-10-07
chore(deps): use go-car master version

rvagg:
Ref: https://github.com/ipld/go-car/issues/254

Depends on getting null-padding added as an option to go-car master, getting a proper tag for this, and getting https://github.com/filecoin-project/go-fil-markets/pull/639 merged (and tagged I suppose).

This is not critical, it's clean-up because Lotus has been depending on a branch of go-car for a long time.


rvagg:
I believe that we _don't_ need the zero=eof for go-car v0 usage here anymore, so we just need an upstream v0.3.2 tag and to get that merged in here along with a matching go-fil-markets and go-commp-utils.

rvagg:
resolved in current master

PR #7469, opened 2021-10-07
Remove obsolete GS testplan - it now lives in go-graphsync

ribasushi:


ribasushi:
A conversation with @nonsense revealed this is no longer an actively-relied-upon test. The motivation of purging it from lotus is the ancient go-ipld-prime dependency it relies on.

PR #7473, opened 2021-10-07
Stop indirectly depending on deprecated github.com/prometheus/common

ribasushi:
Gets us out of https://github.com/prometheus/common/pull/306


ribasushi:
@whyrusleeping this is why you can't upgrade Estuary

@jennijuju we need to backport this to 1.13.0, it's really annoying to build with the API otherwise.

PR #7478, opened 2021-10-08
Dep upgrade pass

ribasushi:
Touching just the seemingly non-controversial ones


Stebalien (approved):
Don't think any of these should be breaking.

PR #7481, opened 2021-10-08
peerstore@v0.2.9 was withdrawn, let's not depend on it directly

ribasushi:
As per slack conversation - while this is probably harmless, we should just fix go.mod


ribasushi:
@marten-seemann could you please also confirm I am correct here: https://github.com/filecoin-project/lotus/pull/7481#issuecomment-939112769

> 0.2.9 and 0.3.0 are identical inside, there's no need to retest anything

marten-seemann:
> 0.2.9 and 0.3.0 are identical inside, there's no need to retest anything

That is correct, see https://github.com/libp2p/go-libp2p-peerstore/compare/v0.2.9...v0.3.0.

Kubuxu (approved):

PR #7483, opened 2021-10-10
Bugfix: withdraw CLIs should depend on network version

arajasek:


jennijuju (approved):

jennijuju:
Lgtm

jennijuju:
(Also learnt today we actually take max(available balance, requested amount) in actor 😅😅😅

cmd/lotus-miner/actor.go
```diff
@@ -291,14 +294,21 @@ var actorWithdrawCmd = &cli.Command{
...
-		if withdrawn != amount {
-			fmt.Printf("Note that this is less than the requested amount of %s FIL\n", amount)
+		if nv >= network.Version14 {
+			var withdrawn abi.TokenAmount
+			if err := withdrawn.UnmarshalCBOR(bytes.NewReader(wait.Receipt.Return)); err != nil {
+				return err
+			}
+
+			fmt.Printf("Successfully withdrew %s FIL\n", withdrawn)
+			if withdrawn != amount {
```
jennijuju, line 309: In lotus - we are checking if the requested amount is > balance. if so, we will return an error directely 

https://github.com/filecoin-project/lotus/blob/1d2f84322d4fc1c0d5357476aa693c309097a8c8/cmd/lotus-miner/actor.go#L243-L259

so either we need to change that behavior or get rid of this check

cli/wallet.go
```diff
@@ -634,14 +636,21 @@ var walletMarketWithdraw = &cli.Command{
...
-			fmt.Printf("Note that this is less than the requested amount of %s FIL\n", amt)
+		if nv >= network.Version14 {
+			var withdrawn abi.TokenAmount
+			if err := withdrawn.UnmarshalCBOR(bytes.NewReader(wait.Receipt.Return)); err != nil {
+				return err
+			}
+
+			fmt.Printf("Successfully withdrew %s FIL\n", withdrawn)
+			if withdrawn != amt {
+				fmt.Printf("Note that this is less than the requested amount of %s FIL\n", amt)
```
jennijuju, line 652: same, we are checking if the balance is < requested.  if so, we will return an error directly and wont get here

https://github.com/filecoin-project/lotus/blob/1d2f84322d4fc1c0d5357476aa693c309097a8c8/cli/wallet.go#L612-L614

cmd/lotus-miner/actor.go
jennijuju: in actor - it does take the smaller value and process tho seems like 
https://github.com/filecoin-project/specs-actors/blob/855f60561dfc014fe7e601402de3621d3219df47/actors/builtin/miner/miner_actor.go#L2052

cmd/lotus-miner/actor.go
```diff
@@ -280,6 +282,7 @@ var actorWithdrawCmd = &cli.Command{
 		fmt.Printf("Requested rewards withdrawal in message %s\n", smsg.Cid())
 
 		// wait for it to get mined into a block
+		fmt.Println("using ", uint64(cctx.Int("confidence")))
```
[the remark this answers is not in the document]
jennijuju: (during testing, i found output `using 5` makes me confused and not sure what its about

cmd/lotus-miner/actor.go
arajasek: Hmm, that's interesting...i'd prefer to leave both actually. 

I think if someone is explicitly trying to withdraw more than they can, we should fail and tell them. If a message winds up withdrawing less than expected, we should also yell at them about that. 

The latter case should only happen in unusual edge cases (the available balance changed between message signing and execution).

Kubuxu (changes requested):
AFAIK the comparison needs to be explicit instead of via `!=`.

cmd/lotus-miner/actor.go
jennijuju, line 309:
```suggestion
			if withdrawn.LessThan(amount) {
```

cli/wallet.go
```diff
@@ -634,14 +636,21 @@ var walletMarketWithdraw = &cli.Command{
...
-		if withdrawn != amt {
-			fmt.Printf("Note that this is less than the requested amount of %s FIL\n", amt)
+		if nv >= network.Version14 {
+			var withdrawn abi.TokenAmount
+			if err := withdrawn.UnmarshalCBOR(bytes.NewReader(wait.Receipt.Return)); err != nil {
+				return err
+			}
+
+			fmt.Printf("Successfully withdrew %s FIL\n", withdrawn)
+			if withdrawn != amt {
```
jennijuju, line 651:
```suggestion
			if withdrawn.LessThan(amount) {
```

Kubuxu (approved):

PR #7495, opened 2021-10-11
Inline codegen

magik6k:
The general idea is that we can use some magic template comments + shared json to add codegen to any part of the code.

TODO:
* [x] Decide if we want this
* [x] Maybe figure out a nicer comment format
* [x] Add to some more places
* [x] Attach to `make gen`
* [x] Fix lint

To use right now, run `go run ./gen/inline-gen . gen/inlinegen-data.json; go fmt ./...`

### DIFF
--- a/chain/actors/version.go
+++ b/chain/actors/version.go
@@ -10,7 +10,7 @@ type Version int
 
 /* inline-gen template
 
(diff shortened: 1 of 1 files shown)


arajasek (approved):
niiiice

PR #7513, opened 2021-10-14
lotus-miner: expose Config in StorageMiner; add SealingPipelineState API endpoint

nonsense:
This PR is exposing the lotus-miner StorageMiner Config, as well as adding a SealingPipelineState API to be consumed by the markets node to determine if a miner is available to accept new deals.

The idea is to improve the `deal acceptance filter` to take into account the state of the sealing pipeline when accepting deals.


dirkmc (commented):
Looks good - just one nit.

Could you post an example of what the output looks like here too

PR #7555, opened 2021-10-25
checkCommit should return SectorCommitFailed

yaohcn:


magik6k (approved):

magik6k:
(we should eventually just drop the ComputeProofFailed state)

PR #7557, opened 2021-10-26
expose msig locked balance api

jennijuju:


jennijuju:
turns out we can just use a Actor balance — avails balance

PR #7559, opened 2021-10-26
add missing NodeType tag

yaohcn:


magik6k (approved):

coder-chenda:
Does this bug fix have any effect on those using v13.0? If not fixed, will v13.0 still work fine? @yhio

PR #7578, opened 2021-10-28
feat(graphsync): allow setting of per-peer incoming requests for miners

rvagg:
@jennijuju to get this started. Exposes graphsync 0.10.3's `MaxInProgressIncomingRequestsPerPeer` as `SimultaneousTransfersForStoragePerClient` for storage providers only (the other gs options are exposed for fullnodes/clients too, I'm assuming this is appropriate behaviour).

Will need @hannahhoward's eye, and tests would be nice but I have no idea what that might look like.

### DIFF
--- a/documentation/en/default-lotus-miner-config.toml
+++ b/documentation/en/default-lotus-miner-config.toml
@@ -207,6 +207,17 @@
   # env var: LOTUS_DEALMAKING_SIMULTANEOUSTRANSFERSFORSTORAGE
   #SimultaneousTransfersForStorage = 20
 
+  # The maximum number of simultaneous data transfers from any single client
+  # for storage deals.
+  # Unset by default (0), and values higher than SimultaneousTransfersForStorage
+  # will have no effect; i.e. the total number of simultaneous data transfers
+  # across all storage clients is bound by SimultaneousTransfersForStorage
+  # regardless of this number.
+  #
+  # type: uint64
+  # env var: LOTUS_DEALMAKING_SIMULTANEOUSTRANSFERSFORSTORAGEPERCLIENT
+  #SimultaneousTransfersForStoragePerClient = 0
+
   # The maximum number of parallel online data transfers for retrieval deals
   #
   # type: uint64
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -131,6 +131,13 @@ type DealmakingConfig struct {
 	MaxStagingDealsBytes int64
 	// The maximum number of parallel online data transfers for storage deals
 	SimultaneousTransfersForStorage uint64
+	// The maximum number of simultaneous data transfers from any single client
+	// for storage deals.
+	// Unset by default (0), and values higher than SimultaneousTransfersForStorage
+	// will have no effect; i.e. the total number of simultaneous data transfers
+	// across all storage clients is bound by SimultaneousTransfersForStorage
+	// regardless of this number.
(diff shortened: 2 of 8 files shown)


hannahhoward (commented):
LGTM as long as folks are ok with the slightly confusing behavior of 0 = no per peer limit.

The 0 made sense as a hidden default, but is perhaps a take confusing as explict config. I think we documented it though so it's probably fine?

jennijuju:
@rvagg @hannahhoward  is the total amount of `SimultaneousTransfersForStoragePerClient` bounded by  `SimultaneousTransfersForStorage `?  so for example, if `SimultaneousTransfersForStorage` is set to 20, and `SimultaneousTransfersForStoragePerClient` is set to 10 and the SP has incoming dt from 5 clients. does SP will only have 20  ongoing dt across these 5 clients from a first come first serve basis?

rvagg:
@jennijuju yes, `SimultaneousTransfersForStorage` is the upper limit regardless. `SimultaneousTransfersForStoragePerClient` is bound by it (docs in here says that) and your example is correct - capped at 20 simultaneous in total and 10 per dt. Do you think it could be made more clear in the docs here perhaps?

documentation/en/default-lotus-miner-config.toml
```diff
@@ -207,6 +207,15 @@
   # env var: LOTUS_DEALMAKING_SIMULTANEOUSTRANSFERSFORSTORAGE
   #SimultaneousTransfersForStorage = 20
 
+  # The maximum number of simultaneous data transfers from any single client
+  # for storage deals.
+  # Unset by default (0), and values higher than SimultaneousTransfersForStorage
+  # will have no effect.
+  #
+  # type: uint64
```
jennijuju, line 215:
```suggestion
  # for storage deals.
  # Unset by default (0), and values higher than SimultaneousTransfersForStorage will have no effect.
  # Total amount of simultaneous data transfer across all storage clients is also bounded by `SimultaneousTransfersForStorage`.
  #
  # type: uint64
```

node/config/types.go
```diff
@@ -131,6 +131,11 @@ type DealmakingConfig struct {
 	MaxStagingDealsBytes int64
 	// The maximum number of parallel online data transfers for storage deals
 	SimultaneousTransfersForStorage uint64
+	// The maximum number of simultaneous data transfers from any single client
+	// for storage deals.
+	// Unset by default (0), and values higher than SimultaneousTransfersForStorage
+	// will have no effect.
```
jennijuju, line 137:
```suggestion
	// will have no effect. Total amount of simultaneous data transfer across all storage clients is also bounded by `SimultaneousTransfersForStorage`.
```

jennijuju:
> @jennijuju yes, `SimultaneousTransfersForStorage` is the upper limit regardless. `SimultaneousTransfersForStoragePerClient` is bound by it (docs in here says that) and your example is correct - capped at 20 simultaneous in total and 10 per dt. Do you think it could be made more clear in the docs here perhaps?

thanks @rvagg ! made one small suggestion on the doc - do you think it make sense?

rvagg:
@jennijuju yep, I've added your words (fairly closely anyway), rebased and squashed this branch so it's up to date and just one commit now.

jennijuju (approved):

magik6k (approved):

PR #7595, opened 2021-11-03
Shed: Add a util to create miners more easily

arajasek:


ZenGround0 (approved):

jennijuju:
is there a reason why this is a hiding shed tool?

PR #7610, opened 2021-11-09
Partial retrieval ux improvements

magik6k:
TODO:
- [x] Cleanup api (keep the path thing in v0 api, use json in v1)
- [x] Add helptext to retrieve command
- [ ] <s>`retrieve stat` command</s>
- [x] `retrieve ls` command
- [x] `retrieve cat` command
- [x] Merge with https://github.com/filecoin-project/lotus/issues/6415


magik6k:
Examples of the new commands:
2021-11-17-202028_1948x1260_scrot

api/types.go
```diff
@@ -194,4 +193,45 @@ type RetrievalInfo struct {
...
+// - otherwise the string is interpreted as ipld-selector-text-lite (simple ipld path)
+//   see https://github.com/ipld/go-ipld-selector-text-lite
+type Selector string
+
+type DagSpec struct {
+	// RootSelector specifies root node
+	// - when using textselector, the path specifies the root node
+	// - if nil then RootSelector is inferred from DataSelector
+	// - must match a single node
+	RootSelector *Selector
```
warpfork, line 217: Did you actually want the powerlevel of selectors here?  If not, this could perhaps also just be [Path](https://pkg.go.dev/github.com/ipld/go-ipld-prime@v0.14.0/datamodel#Path).

magik6k: Probably? The main focus of lotus API is not constraining what users can do, and it feels like you can't use paths to express everything you can express with the 'full' selectors (will be even more true after ADL stuff lands)

node/impl/client/client.go
```diff
@@ -760,325 +763,366 @@ func (a *API) ClientCancelRetrievalDeal(ctx context.Context, dealID rm.DealID) e
...
+				rsn,
+				func(p traversal.Progress, n ipld.Node, r traversal.VisitReason) error {
+					if r == traversal.VisitReason_SelectionMatch {
+						if rootOnNodeBoundary && p.LastBlock.Path.String() != p.Path.String() {
+							return xerrors.Errorf("unsupported selection path '%s' does not correspond to a block boundary (a.k.a. CID link)", p.Path.String())
+						}
+
+						if p.LastBlock.Link == nil {
+							// this is likely the root node that we've matched here
+							// todo: is this a correct assumption
```
arajasek, line 1061: remove

arajasek (approved):

PR #7615, opened 2021-11-10
Feat/snap deals storage

ZenGround0:
WIP introducing snap deals support for sector storage by adding a new file type.

Needs https://github.com/filecoin-project/specs-storage/pull/17


extern/sector-storage/manager_test.go
```diff
@@ -65,11 +68,11 @@ func newTestStorage(t *testing.T) *testStorage {
...
-	for _, path := range t.StoragePaths {
-		if err := os.RemoveAll(path.Path); err != nil {
-			fmt.Println("Cleanup error:", err)
-		}
-	}
+	// 	for _, path := range t.StoragePaths {
+	// 		if err := os.RemoveAll(path.Path); err != nil {
+	// 			fmt.Println("Cleanup error:", err)
+	// 		}
+	// 	}
```
magik6k, line 75: Bring back before merging (could add an env var to turn this cleanup off)

extern/sector-storage/manager_test.go
```diff
@@ -162,6 +165,132 @@ func TestSimple(t *testing.T) {
...
+
+	w := NewLocalWorker(WorkerConfig{TaskTypes: localTasks}, stor, lstor, idx, m, statestore.New(wds))
+	err := m.AddWorker(ctx, w)
+	require.NoError(t, err)
+
+	proofType := abi.RegisteredSealProof_StackedDrg2KiBV1
+	ptStr := os.Getenv("LOTUS_TEST_SNAP_DEALS_PROOF_TYPE")
+	switch ptStr {
+	case "2k":
+	case "8M":
```
magik6k, line 209: Missing 2k?

extern/sector-storage/manager.go
```diff
@@ -601,6 +605,151 @@ func (m *Manager) Remove(ctx context.Context, sector storage.SectorRef) error {
...
+		return storage.ReplicaUpdateOut{}, err
+	}
+	return out, waitErr
+}
+
+func (m *Manager) ProveReplicaUpdate1(ctx context.Context, sector storage.SectorRef, sectorKey, newSealed, newUnsealed cid.Cid) (out storage.ReplicaVanillaProofs, err error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	wk, wait, cancel, err := m.getWork(ctx, sealtasks.TTProveReplicaUpdate1, sector, sectorKey, newSealed, newUnsealed)
```
magik6k, line 661: Wait, doesn't this cancel shadow the cancel above, and don't we end up calling the `getWork` cancel twice, and not calling `context.WithCancel`?

magik6k: (I realize we're doing the same above)

magik6k (approved):
Don't see much to complain about, looks pretty clean

magik6k:
(we also should bump minor worker/miner API versions since we're adding new APIs, but do that after rebasing on latest master because conflicts)

ZenGround0:
> (we also should bump minor worker/miner API versions (api/version.go) since we're adding new APIs, but do that after rebasing on latest master because conflicts)

Good call out.  I don't know the workflow for rebasing next on master but will do this when that time comes.

extern/sector-storage/manager_test.go
ZenGround0: It should be covered by the default above

PR #7617, opened 2021-11-10
Integrate v7 actors

arajasek:
Draft skeleton, basically does all the v7 integration except for FFI call for verification.

Blocked by actors work getting in (https://github.com/filecoin-project/specs-actors/pull/1529)


arajasek:
Needs to be changed to a `next` or some such branch.

Or we can just land in master with an upgrade epoch set to after the rapture.

extern/sector-storage/ffiwrapper/verifier_cgo.go
```diff
@@ -120,6 +122,11 @@ func (proofVerifier) VerifyAggregateSeals(aggregate proof5.AggregateSealVerifyPr
 	return ffi.VerifyAggregateSeals(aggregate)
 }
 
+func (proofVerifier) VerifyReplicaUpdate(update proof7.ReplicaUpdateInfo) (bool, error) {
+	v := ffi.FunctionsSectorUpdate{}
```
arajasek, line 126: I see conversation about this [here](https://github.com/filecoin-project/filecoin-ffi/pull/208#issuecomment-966457279), but I'm not sure why this breaks the pattern (why isn't the verify call in proofs.go like the others).

@ZenGround0 maybe you know?

chain/consensus/filcns/upgrades.go
```diff
@@ -1170,7 +1186,97 @@ func upgradeActorsV6Common(
...
+func PreUpgradeActorsV7(ctx context.Context, sm *stmgr.StateManager, cache stmgr.MigrationCache, root cid.Cid, epoch abi.ChainEpoch, ts *types.TipSet) error {
+	// Use half the CPUs for pre-migration, but leave at least 3.
+	workerCount := runtime.NumCPU()
+	if workerCount <= 4 {
+		workerCount = 1
+	} else {
+		workerCount /= 2
+	}
+
+	//TODO: nv15
```
magik6k, line 1247: Followup PR?

chain/vm/gas_v0.go
```diff
@@ -228,8 +227,14 @@ func (pl *pricelistV0) OnVerifyAggregateSeals(aggregate proof5.AggregateSealVeri
 	return newGasCharge("OnVerifyAggregateSeals", perProof*num+step.Lookup(num), 0)
 }
 
+// OnVerifyReplicaUpdate
+func (pl *pricelistV0) OnVerifyReplicaUpdate(update proof7.ReplicaUpdateInfo) GasCharge {
+	// TODO: do the thing
```
magik6k, line 232: Guessing also a followup PR?

extern/sector-storage/ffiwrapper/verifier_cgo.go
magik6k: Yeah, we need to add those, see https://github.com/filecoin-project/filecoin-ffi/blob/master/proofs.go#L27 for example. (we want it there, so non-lotus things can use go types more easily without importing lotus)

extern/sector-storage/mock/mock.go
```diff
@@ -547,6 +549,11 @@ func (m mockVerifProver) VerifyAggregateSeals(aggregate proof5.AggregateSealVeri
 	return ok, nil
 }
 
+// TODO: do the thing
+func (m mockVerifProver) VerifyReplicaUpdate(update proof7.ReplicaUpdateInfo) (bool, error) {
+	return false, nil
```
magik6k, line 554: `return true, nil` should be reasonable for testing, we can do fancier things later

itests/ccupgrade_test.go
```diff
@@ -32,7 +32,7 @@ func runTestCCUpgrade(t *testing.T, upgradeHeight abi.ChainEpoch) {
 	ctx := context.Background()
 	blockTime := 5 * time.Millisecond
 
-	client, miner, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.LatestActorsAt(upgradeHeight))
+	client, miner, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.TurboUpgradeAt(upgradeHeight))
```
magik6k, line 35: (We'll probably end up transforming ccupgrade tests into snapdeals tests, but for now this is ok)

magik6k (commented):
Mostly some notes/nits

Also, wdpost test is unhappy (doesn't seem to be flaky, restarting didn't help)

chain/vm/gas_v0.go
arajasek: Yes.

chain/consensus/filcns/upgrades.go
arajasek: This is actually ready, will do.

extern/sector-storage/mock/mock.go
arajasek: Ack

itests/ccupgrade_test.go
arajasek: putting in a note at top of the file

arajasek:
This is now complete, good to merge after review

chain/vm/runtime.go
```diff
@@ -142,7 +142,7 @@ func (rt *Runtime) StorePut(x cbor.Marshaler) cid.Cid {
 
 var _ rt0.Runtime = (*Runtime)(nil)
 var _ rt5.Runtime = (*Runtime)(nil)
-var _ rt6.Runtime = (*Runtime)(nil)
```
ZenGround0, line 145: nit why do you need to keep around assert on rt5 but not rt6?  Can you remove all but rt7? If there's a breaking change to the runtime interface, same method name different parameter types this pattern won't work.

chain/vm/syscalls.go
```diff
@@ -286,13 +288,27 @@ func (ss *syscallShim) VerifyAggregateSeals(aggregate proof5.AggregateSealVerify
...
+
 	if !ok {
 		return fmt.Errorf("invalid aggregate proof")
 	}
 
 	return nil
 }
 
+func (ss *syscallShim) VerifyReplicaUpdate(update proof7.ReplicaUpdateInfo) error {
+	ok, err := ss.verifier.VerifyReplicaUpdate(update)
```
ZenGround0, line 300: With new ffi imported does this PR already give lotus correct snap deals update proof verification?  Or is there another PR to follow landing this all the way?

itests/wdpost_test.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 func TestWindowedPost(t *testing.T) {
-	kit.Expensive(t)
+	//kit.Expensive(t)
```
ZenGround0, line 26: What's this about? Should we delete this line?

ZenGround0 (approved):

itests/wdpost_test.go
arajasek: Oops, that was me debugging test failure thanks for busting me.

chain/actors/builtin/miner/v7.go
```diff
@@ -0,0 +1,570 @@
...
+		SealProof:             v7.SealProof,
+		SealedCID:             v7.SealedCID,
+		DealIDs:               v7.DealIDs,
+		Activation:            v7.Activation,
+		Expiration:            v7.Expiration,
+		DealWeight:            v7.DealWeight,
+		VerifiedDealWeight:    v7.VerifiedDealWeight,
+		InitialPledge:         v7.InitialPledge,
+		ExpectedDayReward:     v7.ExpectedDayReward,
+		ExpectedStoragePledge: v7.ExpectedStoragePledge,
```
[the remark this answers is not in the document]
arajasek: Absolutely yes.

chain/vm/runtime.go
arajasek: Based on a quick offline sync, we think it's best to just to have the assertion for every actors version.

If the breaking change @ZenGround0 describes above does happen, we will have a broader refactor to make here.

chain/vm/syscalls.go
arajasek: gives it (hopefully) correct verification!

PR #7637, opened 2021-11-17
reorder `transfer` checks so as to ensure sending 2B FIL to yourself fails if you don't have that amount

### LINKED ISSUE #7596: Tweak autosend behaviour
We currently short circuit (and always succeed) the balance transfer step if the sender is the same as the receiver. That's fine, but it's slightly nicer to instead fail if the sender doesn't have the appropriate amount of FIL (and do nothing if they do).

laudiacay:
Fixes #7596.

### DIFF
--- a/chain/vm/runtime.go
+++ b/chain/vm/runtime.go
@@ -332,7 +332,7 @@ func (rt *Runtime) DeleteActor(beneficiary address.Address) {
 		}
 
 		// Transfer the executing actor's balance to the beneficiary
-		if err := rt.vm.transfer(rt.Receiver(), beneficiary, act.Balance); err != nil {
+		if err := rt.vm.transfer(rt.Receiver(), beneficiary, act.Balance, rt.vm.ntwkVersion(rt.ctx, rt.vm.blockHeight)); err != nil {
 			panic(aerrors.Fatalf("failed to transfer balance to beneficiary actor: %s", err))
 		}
 	}
--- a/chain/vm/runtime.go
+++ b/chain/vm/runtime.go
@@ -332,7 +332,7 @@ func (rt *Runtime) DeleteActor(beneficiary address.Address) {
 		}
 
 		// Transfer the executing actor's balance to the beneficiary
-		if err := rt.vm.transfer(rt.Receiver(), beneficiary, act.Balance, rt.vm.ntwkVersion(rt.ctx, rt.vm.blockHeight)); err != nil {
+		if err := rt.vm.transfer(rt.Receiver(), beneficiary, act.Balance, rt.NetworkVersion()); err != nil {
 			panic(aerrors.Fatalf("failed to transfer balance to beneficiary actor: %s", err))
 		}
 	}
--- a/chain/vm/vm.go
+++ b/chain/vm/vm.go
@@ -874,7 +874,14 @@ func (vm *VM) transfer(from, to address.Address, amt types.BigInt, networkVersio
 	var fromID, toID address.Address
 	var err error
 	// switching the order around so that transactions for more than the balance sent to self fail
+	fmt.Printf("network version! %s\n", networkVersion)
 	if networkVersion >= network.Version15 {
+		fmt.Println("network version high")
+
+		if amt.LessThan(types.NewInt(0)) {
+			return aerrors.Newf(exitcode.SysErrForbidden, "attempted to transfer negative value: %s", amt)
+		}
+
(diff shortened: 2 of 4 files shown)


jennijuju:
is this rebased onto or based off the right branch?

chain/vm/vm.go
```diff
@@ -869,49 +869,86 @@ func (vm *VM) incrementNonce(addr address.Address) error {
...
+		}
 
-	f, err := vm.cstate.GetActor(fromID)
-	if err != nil {
-		return aerrors.Fatalf("transfer failed when retrieving sender actor: %s", err)
+		if fromID == toID {
+			return nil
+		}
+
+		if amt.LessThan(types.NewInt(0)) {
```
arajasek, line 905: Since we're changing validation order here, I'd move this up to perhaps the very start of this if-block -- it's pretty much the cheapest thing there is to validate.

chain/vm/vm.go
```diff
@@ -869,49 +869,86 @@ func (vm *VM) incrementNonce(addr address.Address) error {
...
-	}
+		if f.Balance.LessThan(amt) {
+			return aerrors.Newf(exitcode.SysErrInsufficientFunds, "transfer failed, insufficient balance in sender actor: %v", f.Balance)
+		}
 
-	if fromID == toID {
-		return nil
-	}
+		if from == to {
+			return nil
```
arajasek, line 893: let's throw in an info log for this case (and the subsequent auto-send cases below)

chain/vm/runtime.go
```diff
@@ -332,7 +332,7 @@ func (rt *Runtime) DeleteActor(beneficiary address.Address) {
 		}
 
 		// Transfer the executing actor's balance to the beneficiary
-		if err := rt.vm.transfer(rt.Receiver(), beneficiary, act.Balance); err != nil {
+		if err := rt.vm.transfer(rt.Receiver(), beneficiary, act.Balance, rt.vm.ntwkVersion(rt.ctx, rt.vm.blockHeight)); err != nil {
```
arajasek, line 335:
```suggestion
		if err := rt.vm.transfer(rt.Receiver(), beneficiary, act.Balance, rt.NetworkVersion()); err != nil {
```

arajasek (commented):
Some nits

itests/self_sent_txn_test.go
```diff
@@ -0,0 +1,58 @@
...
+	kit.QuietMiningLogs()
+
+	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.GenesisNetworkVersion(network.Version15))
+	ens.InterconnectAll().BeginMining(10 * time.Millisecond)
+
+	bal, err := client.WalletBalance(ctx, client.DefaultKey.Address)
+	require.NoError(t, err)
+
+	// send self half of account balance
+	msgExactlyBal := &types.Message{
```
arajasek, line 30: why is this called exactlybal when it's clearly halfbal

itests/self_sent_txn_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/go-state-types/exitcode"
+	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/itests/kit"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSelfSentTxn(t *testing.T) {
```
arajasek, line 18: For completeness, it'd be nice to add the exact same test for v14, and assert that both messages succeed there.

itests/self_sent_txn_test.go
laudiacay: im dum

arajasek (approved):

itests/self_sent_txn_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/go-state-types/exitcode"
+	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/itests/kit"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSelfSentTxnV15(t *testing.T) {
```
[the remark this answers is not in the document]
laudiacay: done

PR #7639, opened 2021-11-17
lotus-shed msg: Decode submessages/msig proposals

magik6k:
Very useful when viewing msig propose messages which call methods on other things


jennijuju:
@dkkapur you will love this

arajasek (approved):

PR #7645, opened 2021-11-18
CLI: Add a lotus multisig cancel command

arajasek:


arajasek:
Manually tested, does the thing

magik6k (approved):

PR #7667, opened 2021-11-23
Shed: Add a util to send a batch of messages

arajasek:
eg (not all messages are valid)

```
Recipient, FIL, Method, Params
t15sk4gsiavj4wxwgdiilzmigddwwded2vo6tt6wq, 1, 0, a1
t1ax46m5gxt47makivrzgnpkxc25nftkd5phpkujy, 1, 0, b1
t1dtvhlbzgqiebfnu4lcejlmldxqm4hq5gquf2nda, 1, 1, nil
t1ezofyps6t4kcoerh2ljemmsl4pyjrryijixvc5a, 1, 0, b1
t1grul3jzzzr5722fmn3macizidayqgswd3qd64hq, 1, 1, b1
t1j7i7qwzvmwrgadslfz5sl26jpym7q72tuc2pipq, 10000, 0, b1
t1kbq37ptnbgmgxscvuaqyq26cpyh2wyqhzkge3xy, 1, 0, b1
t1osv6wuq57c6qf2flbxzzym3hyyj4qdoh2dmbhyq, 1, 0, b1
t1pwgfrr7duqtrkdcnuigccd72p7izqj3dubidgya, 1, 7, nil
t1q7vonc7x47g3addcs2iuestou7uptplfxdg3dqy, 1, 2, b1
t1tn7wuhqyv3xq7as3z4da3pmzwc4myj6tffkj3gq, 1, 0, nil
t1xyx2zgde33piwasw7r4t6tvweda7l4lsc2clr4q, 1, 1, b1
t1yeko3rfkkagtj5pyomkfk7jxthivbaxn7rca5xi, 1, 1, b1
```


cmd/lotus-shed/send-csv.go
```diff
@@ -0,0 +1,104 @@
...
+				From:  sender,
+				Value: abi.TokenAmount(value),
+			}, nil)
+			if err != nil {
+				return err
+			}
+
+			fmt.Printf("sending %s to %s in msg %s\n", value.String(), addr, smsg.Cid())
+
+			if i > 0 && i%100 == 0 {
```
arajasek, line 79: Maybe remove? I'm just vaguely worried about bombarding mempool.

cmd/lotus-shed/send-csv.go
```diff
@@ -0,0 +1,104 @@
...
+
+	lapi "github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/types"
+	lcli "github.com/filecoin-project/lotus/cli"
+)
+
+var sendCsvCmd = &cli.Command{
+	Name:      "send-csv",
+	Usage:     "Utility for sending a batch of balance transfers",
+	ArgsUsage: "[sender] [csvfile]",
```
magik6k, line 25: I'd put the sender in the CSV too

cmd/lotus-shed/send-csv.go
```diff
@@ -0,0 +1,104 @@
...
+
+		defer fileReader.Close() //nolint:errcheck
+		r := csv.NewReader(fileReader)
+		records, err := r.ReadAll()
+		if err != nil {
+			return xerrors.Errorf("read csv: %w", err)
+		}
+
+		var msgCids []cid.Cid
+		for i, e := range records[1:] {
```
magik6k, line 57: We might want to parse/check all records first, before sending anything to the network

magik6k (commented):
It would be really cool if this also optionally supported method/params

cmd/lotus-shed/send-csv.go
```diff
@@ -0,0 +1,104 @@
...
+			if err != nil {
+				return xerrors.Errorf("failed to parse address in row %d: %w", i, err)
+			}
+
+			value, err := types.ParseFIL(strings.TrimSpace(e[1]))
+			if err != nil {
+				return xerrors.Errorf("failed to parse value balance: %w", err)
+			}
+
+			smsg, err := api.MpoolPushMessage(ctx, &types.Message{
```
whyrusleeping, line 68: this should use the batch APIs

cmd/lotus-shed/send-csv.go
whyrusleeping: Do a sanity check that the 'header' row is actually a header row and doesnt contain addresses. That way we dont accidentally skip a row the user expected to be there

cmd/lotus-shed/send-csv.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/filecoin-project/lotus/chain/types"
+	lcli "github.com/filecoin-project/lotus/cli"
+)
+
+var sendCsvCmd = &cli.Command{
+	Name:      "send-csv",
+	Usage:     "Utility for sending a batch of balance transfers",
+	ArgsUsage: "[sender] [csvfile]",
+	Action: func(cctx *cli.Context) error {
+		if cctx.NArg() != 2 {
```
whyrusleeping, line 27: I would personally use the `--from` flag instead of a positional argument, its standard across a lot of other things in our tooling

arajasek: fine, but are you okay with it being mandatory? i hate "default wallet" more with each passing day, especially for something like this.

cmd/lotus-shed/send-csv.go
arajasek: the batch api isn't gonna work well, cuz its inputs need to be signed. that means you're gonna either use the same gas params for all msgs (which probably won't work correctly), or individually estimate (tracking nonce yourself), and then batch, which seems ew.

the batch api isn't doing anything that different than just calling `MpoolPushMessage` individually in a loop, i don't think

arajasek:
> It would be really cool if this also optionally supported method/params

Done

ychiaoli18:
@arajasek it would be great if the CSV also accepts a params field so we can use it to find out if the transaction has already been published to the chain. 

`lotus send --params-hex=hexencoed(params) f14mj3nt7tlgyditvk6h5p6i36vqebwgfqndjeq6a 0.01`

e.g. 
Recipient, FIL, Params
t15sk4gsiavj4wxwgdiilzmigddwwded2vo6tt6wq, 1, params1
t1ax46m5gxt47makivrzgnpkxc25nftkd5phpkujy, 1, params2
t1dtvhlbzgqiebfnu4lcejlmldxqm4hq5gquf2nda, 1, params3

ychiaoli18:
> > It would be really cool if this also optionally supported method/params
> 
> Done

@arajasek could we provide different params for each row? e.g. 

Recipient, FIL, Param
t15sk4gsiavj4wxwgdiilzmigddwwded2vo6tt6wq, 1, Param1              
t1ax46m5gxt47makivrzgnpkxc25nftkd5phpkujy, 1, Param2               
t1dtvhlbzgqiebfnu4lcejlmldxqm4hq5gquf2nda, 1, Param3               
t1ezofyps6t4kcoerh2ljemmsl4pyjrryijixvc5a, 1, Param4               
t1grul3jzzzr5722fmn3macizidayqgswd3qd64hq, 1, Param5             

The current payment tool I'm building generates a batch of lotus send commands (or a CSV after this PR merges) for the billing team to copy and execute in the terminal, but the system needs a way to identify if the transaction has been paid, so we need to provide a unique identifier in the param so the tool can search it on the Filecoin chain to know if it has been paid.

arajasek:
> > > It would be really cool if this also optionally supported method/params
> > 
> > 
> > Done
> 
> @arajasek could we provide different params for each row? e.g.
> 
> Recipient, FIL, Param t15sk4gsiavj4wxwgdiilzmigddwwded2vo6tt6wq, 1, Param1 t1ax46m5gxt47makivrzgnpkxc25nftkd5phpkujy, 1, Param2 t1dtvhlbzgqiebfnu4lcejlmldxqm4hq5gquf2nda, 1, Param3 t1ezofyps6t4kcoerh2ljemmsl4pyjrryijixvc5a, 1, Param4 t1grul3jzzzr5722fmn3macizidayqgswd3qd64hq, 1, Param5
> 
> The current payment tool I'm building generates a batch of lotus send commands (or a CSV after this PR merges) for the billing team to copy and execute in the terminal, but the system needs a way to identify if the transaction has been paid, so each transaction needs to have a unique identifier in the param so the tool can search it on the Filecoin chain to know if it has been paid.

Sure thing, done -- using hex encoding for params, lemme know if you'd prefer something else / want it to be configable.

ychiaoli18:
@arajasek thank you! Would you mind also printing the message ids? I may write a short script to look at the result and send the message ids to my system, so the ideal usage may look like the command below. 

`lotus-shed send-csv data.csv | parse-and-forward-messageid`

arajasek:
@ychiaoli18 The message IDs are being printed out, but in this format

```
fmt.Printf("sending %s to %s in msg %s\n", msg.Value.String(), msg.To, smsg.Cid())
```

I think it's important to do that, so that you know which msg id corresponds to which row of the CSV in case some messages get skipped. I can lower the output to be more script-friendly, though if you'd like (something like just `row no, msgid`).

ychiaoli18:
> @ychiaoli18 The message IDs are being printed out, but in this format
> 
> ```
> fmt.Printf("sending %s to %s in msg %s\n", msg.Value.String(), msg.To, smsg.Cid())
> ```
> 
> I think it's important to do that, so that you know which msg id corresponds to which row of the CSV in case some messages get skipped. I can lower the output to be more script-friendly, though if you'd like (something like just `row no, msgid`).

`row no, msgid` would be awesome, thanks!

arajasek:
@ychiaoli18 done!

cmd/lotus-shed/send-csv.go
magik6k: Probably ok to keep, can change into a flag if needed

magik6k (approved):

PR #7688, opened 2021-11-25
Cleanup partial retrieval codepaths ( zero functional changes )

ribasushi:
- Adjust the ExportRef comment to reflect change from #7673
- Simplify control flow in parseDagSpec() ( read diff under -w )


magik6k (approved):

ribasushi:
@magik6k just a reminder: we should still fix the export-to-car effectively not working as-is

PR #7689, opened 2021-11-25
disable mplex stream muxer

marten-seemann:
Compatibility considerations: The reason we have mplex in IPFS is that js-libp2p doesn't support yamux, so mplex is the fallback stream muxer there. We don't care about js-libp2p in lotus. As far as I know (someone with more ecosystem insight please double check!), all Filecoin implementations support yamux (at least Venus does), so there's no need for a fallback.


vyzo (approved):

vyzo:
needs `go mod tidy`

go.sum
```diff
@@ -388,7 +388,6 @@ github.com/filecoin-project/specs-actors/v3 v3.1.1/go.mod h1:mpynccOLlIRy0QnR008
 github.com/filecoin-project/specs-actors/v4 v4.0.0/go.mod h1:TkHXf/l7Wyw4ZejyXIPS2rK8bBO0rdwhTZyQQgaglng=
 github.com/filecoin-project/specs-actors/v4 v4.0.1 h1:AiWrtvJZ63MHGe6rn7tPu4nSUY8bA1KDNszqJaD5+Fg=
 github.com/filecoin-project/specs-actors/v4 v4.0.1/go.mod h1:TkHXf/l7Wyw4ZejyXIPS2rK8bBO0rdwhTZyQQgaglng=
-github.com/filecoin-project/specs-actors/v5 v5.0.0-20210512015452-4fe3889fff57/go.mod h1:283yBMMUSDB2abcjP/hhrwTkhb9h3sfM6KGrep/ZlBI=
```
magik6k, line 391: Removing this seems to have broken build on CI

Stebalien (approved):
LGTM. I like mplex as a simpler fallback, but that simplicity does have a cost which doesn't seem warranted here.

I expect we'll introduce separate "gateway" nodes eventually that might want to enable that transport, but we can cross that bridge later.

PR #7691, opened 2021-11-25
Add RLE dump code

Kubuxu:


cmd/lotus-shed/sectors.go
```diff
@@ -275,6 +280,113 @@ var terminateSectorPenaltyEstimationCmd = &cli.Command{
...
+				}
+
+				var bf bitfield.BitField
+				err = bf.UnmarshalCBOR(bytes.NewReader(rle))
+				if err != nil {
+					return xerrors.Errorf("decoding bitfield: %w", err)
+				}
+				ri, err := bf.RunIterator()
+				if err != nil {
+					return xerrors.Errorf("creating interator: %w", err)
```
arajasek, line 364:
```suggestion
					return xerrors.Errorf("creating iterator: %w", err)
```

cmd/lotus-shed/sectors.go
```diff
@@ -275,6 +280,113 @@ var terminateSectorPenaltyEstimationCmd = &cli.Command{
...
+	n := sort.Search(len(miners), func(i int) bool {
+		pow := powCache[miners[i]]
+		return pow.IsZero()
+	})
+	return append(miners[0:0:0], miners[:n]...), nil
+}
+
+var dumpRLESectorCmd = &cli.Command{
+	Name:  "dump-rles",
+	Usage: "",
```
arajasek, line 313: could use with input explanation / what happens if no input is provided

arajasek (approved):
Curious what motivated this

Kubuxu:
I was analyzing how good/bad our RLE+ encoding is in practice.
Rough results:
A symbol in our case is a run.
Shannon Entropy of RLEs in AllocatedSectors: Η = 3.468 bits/symbol
RLE+ encoding: ΔH = 0.7687 bits/symbol
RLE+ with small modification: ΔH = 0.203 bits/symbol
RLE+ with Hufman: ΔH = 0.0178 bits/symbol
RLE+ with Asymetric Numeral Systems: ΔH ~= 0.001 bits/symbol

We could save 18% of RLE+ storage but it is such a small fraction that it isn't worth the complexity right now (after collecting data I know that RLEs are only ~10MB of chain state, but the churn is frequent).
The major reason for the high ΔH of our RLE+ coding are 6-bit encodings of symbols 2 and 3.

If I were to design RLE+ today I would have gone with RLE with Huffman coding of small runs and additional data lengths.

PR #7693, opened 2021-11-26
Make small retrieval 200x faster

magik6k:
This PR adds direct random-access capability to the lotus DAGStore mount.

Right now, when any retrieval is performed, DAGStore will read the entire piece and store it in a temporary file for random access. This is acceptable when retrieving the entire piece, but incredibly inefficient when performing retrievals for small % of the original piece.

By avoiding this copy step, small retrievals in many cases become sub-second, and time-to-first-byte in all retrievals is drastically reduced.

Example timings

Running current master, retrieving a few kb from a 16G piece
```
 ./lotus client ls --miner f0... --maxPrice 0 bafy...
Recv 0 B, Paid 0 FIL, Open (New), 0s
Recv 0 B, Paid 0 FIL, DealProposed (WaitForAcceptance), 9ms
Recv 0 B, Paid 0 FIL, DealAccepted (Accepted), 45ms
Recv 0 B, Paid 0 FIL, PaymentChannelSkip (Ongoing), 45ms
Recv 0 B, Paid 0 FIL, Complete (CheckComplete), 1m6.321s
Recv 0 B, Paid 0 FIL, WaitForLastBlocks (WaitingForLastBlocks), 1m6.338s
Recv 79 KiB, Paid 0 FIL, BlocksReceived (WaitingForLastBlocks), 1m6.34s
Recv 79 KiB, Paid 0 FIL, AllBlocksReceived (FinalizingBlockstore), 1m6.354s
Recv 79 KiB, Paid 0 FIL, BlockstoreFinalized (Completed), 1m6.354s
```

Running this PR, also retrieving a few kb from a different 16G piece (different piece to avoid any potential caching)
```
./lotus client ls --miner f02620 --maxPrice 0 QmebwyN5z8CiczuTFsfr3rio3ttHao8CoZwMeFHVu4Xhmd
Recv 0 B, Paid 0 FIL, Open (New), 0s
Recv 0 B, Paid 0 FIL, DealProposed (WaitForAcceptance), 10ms
Recv 0 B, Paid 0 FIL, DealAccepted (Accepted), 67ms
Recv 0 B, Paid 0 FIL, PaymentChannelSkip (Ongoing), 67ms
Recv 0 B, Paid 0 FIL, Complete (CheckComplete), 267ms
Recv 0 B, Paid 0 FIL, WaitForLastBlocks (WaitingForLastBlocks), 280ms
Recv 2.025 KiB, Paid 0 FIL, BlocksReceived (WaitingForLastBlocks), 284ms
Recv 2.025 KiB, Paid 0 FIL, AllBlocksReceived (FinalizingBlockstore), 284ms
Recv 2.025 KiB, Paid 0 FIL, BlockstoreFinalized (Completed), 285ms
```


raulk (commented):
Took a cursory look, and the approach generally LGTM. However, something to call out is that efficiency will be (a) DAG-shape dependent and (b) request-dependent.

Re: (a), a DAG composed of tiny blocks will read to substantial read amplification and roundtripping.

Re: (b), if the caller is fetching the root CID of the deal with the default selector, this equates to fetching the entire deal DAG. At that point it could be more efficient to copy the shard data and serving from local, like we were doing before.

@magik6k -- what do you think about introducing a "planner" component that decides which strategy to use based on (b)?

For (a) we should eventually store metadata about a DAG, something like a `DESCRIBE DAG`, as we receive it, along with some stats (average min/max/avg node size, etc). This would allow us to strategise around (a).

magik6k:
@raulk From my testing, when retrieving the whole piece, reads tend to be sequential because graphsync requests data from the DAGStore in the same order it was written to the CAR file, and the pieceReader is exploiting this property.

For some, bigger partial retrievals most accesses should still be mostly sequential, but possibly with multiple "heads", but here more data/experimentation is needed.

(basically, in my testing, I didn't come up with an example in which copying data would be more efficient)

raulk:
@magik6k Oh yeah, that's a good point. Both the layout of data in disk and the retrieval order are depth-first. That fact, along with the "burn" leniency, should solve both challenges in majority of cases 👍  Maybe we can think of some selectors and DAG combinations whose pattern of access would be different, and test those?

markets/dagstore/piecereader.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (p *pieceReader) ReadAt(b []byte, off int64) (n int, err error) {
+	if err := p.check(); err != nil {
+		return 0, err
+	}
+
+	// get the backing reader into the correct position
+	if p.r == nil {
+		p.rAt = MaxPieceReaderBurnBytes * -2
+	}
```
dirkmc, line 101: I'm not sure I understand what the purpose of this clause is; It seems that if `p.r` is `nil`, `p.rAt` will be set to `MaxPieceReaderBurnBytes * -2` and so the `if` statement on line 105 will always be true.
If so, why not just include `p.r == nil` in the `if` statement on line 105?

dirkmc (approved):
One question but in general LGTM 👍

markets/dagstore/piecereader.go
magik6k: Yeah, no good reason for this to be like that

hunjixin:
is this pr was rollback?

PR #7699, opened 2021-11-26
Add verbose mode to lotus-miner pieces list-cids

magik6k:


cmd/lotus-miner/pieces.go
```diff
@@ -61,9 +68,54 @@ var piecesListCidInfosCmd = &cli.Command{
 			return err
 		}
 
+		w := tablewriter.New(tablewriter.Col("CID"),
```
arajasek, line 71: Isn't this gonna get printed in non verbose mode too

magik6k: Nope, tablewriter doesn't print anything with no entries

arajasek: oh nice

arajasek (approved):

PR #7706, opened 2021-11-29
retrieval: Only output matching nodes, MatchPath dagspec

magik6k:
Depends on https://github.com/ipld/go-ipld-selector-text-lite/pull/5

todo:
- [x] Use go-ipld-selector-text-lite master when merged
- [x] Test non-path-matching retrieval
- [x] Make sure that ls / cat still works


jennijuju:
also need https://github.com/ipld/go-ipld-selector-text-lite/pull/5

api/types.go
```diff
@@ -214,6 +214,10 @@ type DagSpec struct {
 	// - when using textselector, the path specifies subtree
 	// - the matched graph must have a single root
 	DataSelector *Selector
+
+	// MatchPath matches the path traversal when DataSelector is a textselector.
+	// Ignored when DataSelector is a json selector and in non-car retrieval
+	MatchPath bool
```
ribasushi, line 220: That's a crummy name... Perhaps `CarExportFullMerkleProof` ?

magik6k: `ExportMerkleProof`?

node/impl/client/client.go
```diff
@@ -777,13 +778,13 @@ func getDataSelector(dps *api.Selector) (datamodel.Node, error) {
 			ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
 
 			selspec, err := textselector.SelectorSpecFromPath(
-				textselector.Expression(*dps),
+				textselector.Expression(*dps), matchPath,
 
 				// URGH - this is a direct copy from https://github.com/filecoin-project/go-fil-markets/blob/v1.12.0/shared/selectors.go#L10-L16
 				// Unable to use it because we need the SelectorSpec, and markets exposes just a reified node
```
ribasushi, line 784:
```suggestion
```

Comment no longer applicable

api/types.go
```diff
@@ -214,6 +214,10 @@ type DagSpec struct {
 	// - when using textselector, the path specifies subtree
 	// - the matched graph must have a single root
 	DataSelector *Selector
+
+	// ExportMerkleProof matches the path traversal when DataSelector is a textselector.
+	// Ignored when DataSelector is a json selector and in non-car retrieval
```
ribasushi, line 219:
```suggestion
	// ExportMerkleProof is applicable only when exporting to a CAR file via a path textselector
	// When true, in addition to the selection target, the resulting CAR will contain every block along the
	// path back to, and including the original root
	// When false the resulting CAR contains only the blocks of the target subdag 
```

ribasushi (approved):
Great stuff!

2 more suggestions, no functional changes

PR #7707, opened 2021-11-29
Disable appimage

travisperson:
It appears recently something has changed in circleci and we no longer have access to the /dev/snd device which is causing the appimage tests to fail during setup.


magik6k (commented):
Should modify `template.yml`, so that the gen-check doesn't complain.

magik6k (approved):

PR #7709, opened 2021-11-29
fix logic error

coryschwartz:
obviously, we want to run the init if the gate file **doesn't** exist.


magik6k (approved):

travisperson (approved):

coder-chenda:
I use the docker method to start and run the node. I would like to ask if the fix of this bug affects the start and operation of docker? @jennijuju  @coryschwartz

PR #7710, opened 2021-11-30
ffiwrapper: Validate PC2 by calling C1 with random seeds

### LINKED ISSUE #7702: Add checks after computing PC2
After running PC2 we should call Commit1 with random seeds as a check that there were no compute errors in the sealing process.

Without this in case of compute errors we might precommit to bad (and unprovable) CommR, and then get `Invalid vanilla proof generated` in the commit step, at which point the only thing one can do is to remove the sector and lose precommit deposit.

magik6k:
This lets lotus detect compute errors in PC1/PC2 early, before precomitting to sectors which can't be proven (and can't be recovered because the CommR on chain is bad)

Closes #7702

### DIFF
--- a/extern/sector-storage/ffiwrapper/sealer_cgo.go
+++ b/extern/sector-storage/ffiwrapper/sealer_cgo.go
@@ -7,6 +7,9 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/json"
 	"io"
 	"math/bits"
 	"os"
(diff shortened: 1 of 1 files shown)


extern/sector-storage/ffiwrapper/sealer_cgo.go
```diff
@@ -530,9 +533,19 @@ func (sb *Sealer) SealPreCommit1(ctx context.Context, sector storage.SectorRef,
...
 		return nil, xerrors.Errorf("presealing sector %d (%s): %w", sector.ID.Number, paths.Unsealed, err)
 	}
-	return p1o, nil
+
+	p1odec := map[string]interface{}{}
+	if err := json.Unmarshal(p1o, &p1odec); err != nil {
+		return nil, xerrors.Errorf("unmarshaling pc1 output: %w", err)
+	}
+
+	p1odec["_lotus_SealRandomness"] = ticket
```
arajasek, line 542: This should probably be a constant in this file or some such...also maybe rename it to indicate it's not real randomness, just for sanity-checking?

arajasek (approved):

extern/sector-storage/ffiwrapper/sealer_cgo.go
magik6k: Well, it is the real seal randomness (aka the ticket), we use this json thing to transport it into PC2 which doesn't have an easy way to get to it.

PR #7735, opened 2021-12-03
Command to list active sector locks

magik6k:
Needed this to test a theory on why some retrievals might hang, might be useful.

```
$ lotus-miner storage locks
2428    PreCommit1      unsealed(R:1) sealed(W) cache(W)
2429    PreCommit2      sealed(R:1) cache(W)
```


arajasek (approved):
nice!

PR #7739, opened 2021-12-05
first pass at parameters for ingest pubsub subscription

willscott:
cc @gammazero / @masih 

* This uses `/indexer/ingest/<networkname>` for pubsub topic to subscribe to. We should do some sort of different topic for 'real' publishers versus testnets. Think about what you'd like to see on the indexer side in terms of naming. Also note that the convention seems to be that pubsub topics start with an initial '/', which the `indexer/ingest` topic name the indexer is using doesn't.
* The baseline guess at topic weight is based on observed average deal rate for the last month. We'll need to change that weighting as that increases - i picked a factor of 4 without much evidence, happy to think about if there are better numbers to start with, but probably we don't need to block on getting it to reflect node-observed chain activity.

### DIFF
--- a/node/modules/lp2p/pubsub.go
+++ b/node/modules/lp2p/pubsub.go
@@ -114,6 +114,22 @@ func GossipSub(in GossipIn) (service *pubsub.PubSub, err error) {
 		InvalidMessageDeliveriesDecay:  pubsub.ScoreParameterDecay(time.Hour),
 	}
 
+	ingestTopicParams := &pubsub.TopicScoreParams{
+		// expected ~0.5 confirmed deals / min. sampled
+		TopicWeight: 1,
+
+		TimeInMeshWeight:  0.00027, // ~1/3600
+		TimeInMeshQuantum: time.Second,
+		TimeInMeshCap:     1,
+
+		FirstMessageDeliveriesWeight: 5,
+		FirstMessageDeliveriesDecay:  pubsub.ScoreParameterDecay(time.Hour),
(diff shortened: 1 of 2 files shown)


node/modules/lp2p/pubsub.go
```diff
@@ -114,6 +114,22 @@ func GossipSub(in GossipIn) (service *pubsub.PubSub, err error) {
 		InvalidMessageDeliveriesDecay:  pubsub.ScoreParameterDecay(time.Hour),
 	}
 
+	ingestTopicParams := &pubsub.TopicScoreParams{
```
vyzo, line 117: please dont just copy paste these params, you may weaken security properties of the other topics.

Is there proper validation on this topic?
I dont see anything. If not, dont set parameters at all, they are worse than useless.

vyzo (changes requested):
we need some thought/discussion on the topic params.

node/modules/lp2p/pubsub.go
gammazero: @vyzo Do you mean that `topicParams[build.IngestTopic(in.Nn)]` should not be set?  Or, are you saying that it should be set to some default, empty `TopicScoreParams` or `nil`?
```
topicParams[build.IngestTopic(in.Nn)] = &pubsub.TopicScoreParams{}
```

vyzo: unless you have a strong validator (which i dont see anywhere), do *not* set topic parameters.

rjan90:
Closing this PR as outdate/stale.

Issue #7648, opened 2021-11-19
Expose EnableGasTracing as an env_var setting


jennijuju:
EnableGasTracing is super helpful for understanding gas cost analysis, users should be able to get that information for compute state if they want without updating the code. expose this variable as an envvar that user can set, default to be false.

jennijuju:
https://github.com/filecoin-project/lotus/blob/ef0771ee4850c1b7d63bbed0e4d3c909d23012bc/chain/vm/runtime.go#L55

PR #7750, opened 2021-12-10
Expose EnableGasTracing as an env_var

geoff-vball:
Closes issue #7648

### DIFF
--- a/chain/vm/runtime.go
+++ b/chain/vm/runtime.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"os"
 	gruntime "runtime"
 	"time"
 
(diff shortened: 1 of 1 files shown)


chain/vm/runtime.go
```diff
@@ -51,8 +52,7 @@ func (m *Message) ValueReceived() abi.TokenAmount {
 	return m.msg.Value
 }
 
-// EnableGasTracing, if true, outputs gas tracing in execution traces.
-var EnableGasTracing = false
+const EnableGasTracingEnvVar = "LOTUS_VM_ENABLE_GAS_TRACING_VERY_SLOW"
```
magik6k, line 55: I'd do
```suggestion
// EnableGasTracing, if true, outputs gas tracing in execution traces.
var EnableGasTracing = os.Getenv("LOTUS_VM_ENABLE_GAS_TRACING_VERY_SLOW") == "1"
```

So that we don't have to lookup the env var on every gas charge

geoff-vball: Changed, thanks that makes it a lot simpler.

arajasek (approved):

magik6k (approved):

PR #7758, opened 2021-12-10
Relay indexer pubsub messages 

gammazero:
This uses /indexer/ingest/<networkname> for pubsub topic to subscribe to. Test publishers should use a differnt network namd to distinguish them from 'real' publishers.

This replaces PR #7739

### DIFF
--- a/node/modules/services.go
+++ b/node/modules/services.go
@@ -2,6 +2,7 @@ package modules
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strconv"
 	"time"
@@ -15,7 +16,6 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"go.uber.org/fx"
-	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-fil-markets/discovery"
 	discoveryimpl "github.com/filecoin-project/go-fil-markets/discovery/impl"
@@ -58,7 +58,7 @@ func RunHello(mctx helpers.MetricsCtx, lc fx.Lifecycle, h host.Host, svc *hello.
 
 	sub, err := h.EventBus().Subscribe(new(event.EvtPeerIdentificationCompleted), eventbus.BufSize(1024))
 	if err != nil {
-		return xerrors.Errorf("failed to subscribe to event bus: %w", err)
+		return fmt.Errorf("failed to subscribe to event bus: %w", err)
 	}
 
 	ctx := helpers.LifecycleCtx(mctx, lc)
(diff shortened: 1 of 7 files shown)


vyzo (approved):
this is totally fine, but dont you want the  bootstrappers to subscribe as well?

vyzo:
Meaning do an actual subscribe so that they can do PX and fullfill their bootstrapping duties.

gammazero:
@vyzo This side is only publishing on the ingest topic, does it need to subscribe anyway?

gammazero:
Lotus is only every publishing to the ingest topic.  The documentation says:
> To publish a message to some topic, use Publish; you don't need to be subscribed to the topic in order to publish.
But, as a bootstrapping node a subscription to the topic is needed, otherwise they will not pass on publications to other peers?

vyzo:
Publishing does not require a subscription indeed, and this is all fine as is.

Having said that, it would be very helpful for the bootsrappers to actually subscribe so that clients can discover peers through PX.
But this is non normative, really.

node/modules/services.go
```diff
@@ -58,7 +58,7 @@ func RunHello(mctx helpers.MetricsCtx, lc fx.Lifecycle, h host.Host, svc *hello.
 
 	sub, err := h.EventBus().Subscribe(new(event.EvtPeerIdentificationCompleted), eventbus.BufSize(1024))
 	if err != nil {
-		return xerrors.Errorf("failed to subscribe to event bus: %w", err)
+		return fmt.Errorf("failed to subscribe to event bus: %w", err)
```
vyzo, line 61: please dont replace this.  we use xerrors because it has stack traces.

node/modules/services.go
```diff
@@ -198,6 +198,29 @@ func HandleIncomingMessages(mctx helpers.MetricsCtx, lc fx.Lifecycle, ps *pubsub
 	waitForSync(stmgr, pubsubMsgsSyncEpochs, subscribe)
 }
 
+func RelayIndexerMessages(lc fx.Lifecycle, ps *pubsub.PubSub, nn dtypes.NetworkName) error {
+	topicName := build.IngestTopic(nn)
+	topicHandle, err := ps.Join(topicName)
+	if err != nil {
+		return fmt.Errorf("failed to join pubsub topic %s: %w", topicName, err)
```
vyzo, line 205: use xerrors.

vyzo (changes requested):
you need to register a validator, otherwise it will be spamfest.

node/modules/services.go
gammazero: Done

node/modules/services.go
gammazero: Done

gammazero:
I added a validator, but am not sure if it is adequate.  I am not sure is there is anything specific the validator is expected to do, like rate limiting.

gammazero:
Merged into PR #7313 as PR #8026

vyzo:
@gammazero That doesn't change the fact that you need to add a vlidator! The network can be spammed by flooding junk messages to the bootstrappers on the indexer topic, which would then be gossiped to the other nodes in the indexer overlay.

PR #7784, opened 2021-12-13
Merge v15 work into master

arajasek:


arajasek (commented):
Based on a pass, looks like nothing here would explode master (or have any perceptible impact), except for (maybe) the one comment.

@ZenGround0 can you please go over this as well (esp miner changes), just to be suuuuper sure that nothing here will have any impact until the upgrade epoch and CLI / API hooks are introduced?

ZenGround0 (approved):
Good to merge, all the new sector-storage endpoints are unused so there should be no problems from the changes

Issue #4368, opened 2020-10-13
When receiver is an ID address, check the period of and f0 address’s existence in the network before proceed the message 


jennijuju:
only allow send message to an ID address until the t0 address has been in the chain for 900 blocks

This is to avoid the block that the ID was allocated in is reverted due to network reorg

Kubuxu:
Currently, if you send with `From` ID address we will replace it with public key address (type 1 or 3).
Proper solution for that is looking at state of init actor Finality back and checking what is the `nextID` in it to decide if we should use public key address or ID address.

Kubuxu:
The same should be applie to To address, replacing To address to ID address if ID of the `To` public key address is older than Finality and refusing to send with `To` ID address newer than Finality.

arajasek:
Slapping the mpool-management epic label on this, if it isn't covered there, we'll re-triage.

BigLep:
Per 2021-06-14 discussion, this will get closed after Hyperdrive.

Kubuxu:
@BigLep I don't think Hyperdrive solved this. Hyperdrive fixed some minor issues with ID address handling but it doesn't permit you to use ID addresses from Lotus. It also doesn't automatically use ID addresses when possible.

Stebalien:
Was this not fixed in https://github.com/filecoin-project/lotus/pull/6364?

Stebalien:
But we probably need a test.

hunjixin:
is finality(900 height)  a bit big for this issue?

PR #7795, opened 2021-12-15
Checks to-address for finality before adding message to mempool

geoff-vball:
Closes issue #4368

### DIFF
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -670,6 +670,14 @@ func (mp *MessagePool) Push(ctx context.Context, m *types.SignedMessage) (cid.Ci
 		return cid.Undef, err
 	}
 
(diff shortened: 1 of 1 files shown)


arajasek (changes requested):
So this doesn't quite work, unfortunately. `StateAccountKeyAtFinality` will fail if called with something that isn't a secp / bls account -- something like a miner actor doesn't have an "account key", although it does have a robust address.

I think we need to do the following:

- check if `To` is an ID address. If it is,
- check if it existed 900 epochs ago. if it didn't,
- check if the `To` actor is an account. if it is,
- fail: they should be using the pubkey address instead.

Really we should be failing even if the `To` actor isnt' an account, but that's gonna break a looot of tooling, so I'd rather just see this land first.

jennijuju:
feel like we should include this in v1.14.0?

PR #7811, opened 2021-12-17
VM: Circ supply should be constant per epoch

arajasek:
The current behaviour violates the spec (see [here](https://github.com/filecoin-project/specs-actors/blob/master/actors/runtime/runtime.go#L116))


jennijuju (approved):

jennijuju:
why do you piss ci off

Stebalien (approved):

Issue #6017, opened 2021-04-11
[Feature Request] Add a command for query retrieval price of miners


askender:
Clients who will store data in filecoin network need know the retrieval-price of miners before they decide to do so.
But we don't have a command for that.

**Describe the solution you'd like**
We can add the info of retrieval price in   `lotus client query-ask f0xxxx` and `lotus client list-asks`

**Describe alternatives you've considered**
As miners can change the retrieval-price at any time. Clients maybe find that the price is too high for them. We should have a solution for it.

**Additional context**
1. the default retrieve price is high maybe. Client maybe pay a lot if forget to add the `--maxPrice`
```
   lotus client retrieve
   --maxPrice value  maximum price the client is willing to consider (default: 1 FIL)
```
2. the `Unseal Price` should also be open to everyone.

jennijuju:
Complete #7397

dirkmc:
Fixed by https://github.com/filecoin-project/lotus/pull/7814

PR #7814, opened 2021-12-17
feat: #6017 market: retrieval ask CLI command

dirkmc:
## Related Issues
Fixes https://github.com/filecoin-project/lotus/issues/6017
Supersedes https://github.com/filecoin-project/lotus/pull/7397

## Proposed Changes
Add CLI command for retrieval asks

## Example Output
```
$ lotus client query-retrieval-ask --size=134217728 f01000 bafyaeicumxtb5zs3fnli2yndp4iakfit64jdxp7ulajv1lkw2cyiwi3pvm
Ask: f01000
Unseal price: 0.000000001 FIL
Price per byte: 0.000000002 FIL
Payment interval: 1 MiB
Payment interval increase: 2 MiB
Total price for 134217728 bytes: 0.268435457 FIL
```


cli/client.go
```diff
@@ -1030,6 +1031,63 @@ var clientFindCmd = &cli.Command{
 	},
 }
 
+var clientQueryRetrievalAskCmd = &cli.Command{
+	Name:      "query-retrieval-ask",
```
magik6k, line 1035: Maybe just
```suggestion
	Name:      "retrieval-ask",
```

cli/client.go
```diff
@@ -1030,6 +1031,63 @@ var clientFindCmd = &cli.Command{
...
+			return err
+		}
+
+		afmt.Printf("Ask: %s\n", maddr)
+		afmt.Printf("Unseal price: %s\n", types.FIL(ask.UnsealPrice))
+		afmt.Printf("Price per byte: %s\n", types.FIL(ask.PricePerByte))
+		afmt.Printf("Payment interval: %s\n", types.SizeStr(types.NewInt(ask.PaymentInterval)))
+		afmt.Printf("Payment interval increase: %s\n", types.SizeStr(types.NewInt(ask.PaymentIntervalIncrease)))
+
+		size := cctx.Int64("size")
```
magik6k, line 1079: I'm pretty sure that QueryOffer has a Size field (for the whole piece, so doesn't exactly apply to selective retrievals, but imo still useful to print that)

nonsense (approved):
lgtm

cli/client.go
dirkmc: The size is actually passed in as a parameter - if the user passes it, then the "Total Cost" calculation is output - see the Example Output in the PR description.

magik6k: Yes, my point is that the default could be to use the size from QueryOffer if it wasn't specified by the user

dirkmc: Ok I made that change 👍

magik6k (approved):

Issue #7833, opened 2021-12-20
parent-grinding fault overcheck


llifezou:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [X] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
lotus version 1.13.3-dev+mainnet+git.89912598a.dirty
lotus-miner version 1.13.3-dev+mainnet+git.89912598a.dirty
```

### Describe the Bug

Continuously generate blocks, the first block is invalid (including the wrong parent block), causing the second block to fail.

### Logging Information

```text
block 1371234 daemon info:

[...]

PR #7834, opened 2021-12-20
fix: #7833 mine: parent-grinding fault overcheck

llifezou:
## Related Issues
 https://github.com/filecoin-project/lotus/issues/7833

## Additional Information
This pr changes to the api because the slashfilter cannot determine whether the parent block is valid.
This may be the smallest fix, but there are changes to the API interface


ZenGround0:
Hey @llifezou I'm not experienced with this part of the lotus code but its not clear to me that the first block is kept from propagating out over the network.  If it propagates and we bypass slash filtering on the second block the miner will get slashed.  Does this code prevent propagation of the first block?

chain/gen/slashfilter/slashfilter.go
```diff
@@ -48,7 +72,7 @@ func (f *SlashFilter) MinedBlock(bh *types.BlockHeader, parentEpoch abi.ChainEpo
 		}
 	}
 
-	{
+	if checkParentGrinding {
```
ZenGround0, line 75: Ok I'm starting to get it this is not two blocks in the epoch it is parent grinding.  I still need to understand what's going on but my first impression is we either need to change this logic to more closely match the actual slashing logic directly and if we can't there's something unsafe going on.

llifezou: Yes, we should make sure to be safer, because the above is less likely to happen. But what I understand is that in this case, the abnormal parent block is not introduced into the chain, so it will not be punished. I understand that it may be over-checked

ZenGround0: The nuance here is that the bad parent block is spread to the wider network.  I understand that it doesn't make it onto the main chain.  However that is not related to whether the block can cause a slashing event.  As long as the block is sent out over the network anyone else who records that block has the right to slash you if you also mine the second block which ignores the bad parent.

llifezou: Oh, so it is, and if so, isn't it a bit over-punishing?

llifezou: This also means that if miner get the right to manufacture two consecutive blocks, as long as the first block is invalid, the second block will also be lost.

ZenGround0:
Move discussion of the actual problem to [fips discussion](https://github.com/filecoin-project/FIPs/discussions/402).  Closing this down because the path forward won't involve code like this: its either going to be to do nothing, or remove this type of consensus fault from fault checking.

PR #7835, opened 2021-12-20
misc: api: Annotate lotus tests according to listed behaviors

brdji:
## Related Issues
https://github.com/filecoin-project/system-test-matrix/issues/279
https://github.com/filecoin-project/system-test-matrix/issues/226
https://github.com/filecoin-project/system-test-matrix/issues/225

## Proposed Changes
Annotate all tests according to behaviors specified in the system-test-behavior and system-test-matrix. Collect these annotations so that we can gain an overview of lotus test coverage.
The System Test Matrix is a dashboard with a detailed list of FIL features and behaviors and a mapping between those features 
and test suites, systems and subsystems that those features are related to.

Files and tests are annotated using `//stm:` comments.
Every FIL test file is marked with a special annotation at the top of the file (e.g. #unit, #integration, #cli, #api ... )
For example, `//stm: #unit` would mark the file as a unit test file.
The test crawler will gather all test files, including those that are not annotated using a file-level annotation. However, test coverage will be collected only from files bearing a file-level annotation.

Single tests are annotated using `//stm: @` comments, with each behavior preceded by the `@` character.
For example, a test with the following comment
`//stm: @MARKET_RESERVE_FUNDS_001, @MINER_CREATE_BLOCK_001`
is assumed to test: 
- `MARKET_RESERVE_FUNDS_001` behavior `market` system
- `MINER_CREATE_BLOCK_001` behavior of the `miner` system

The system-test-matrix crawler will go through all lotus tests, gather information from all unit and integration tests present,
and document:
- Which behaviors are tested
- Which behaviors are untested
- Which tests cover which behaviors
- Which test files lack annotations
- etc.

This information is then compiled into an output `json` file, and shown in the system-test-matrix dashboard. The dashboard offers visualization and querying of test and behavior coverage of the entire lotus repository.

## Additional Info
See https://github.com/filecoin-project/system-test-matrix for additional information on test annotations.


node/impl/client/client_test.go
```diff
@@ -44,6 +46,7 @@ func TestImportLocal(t *testing.T) {
 	b, err := testdata.ReadFile("testdata/payload.txt")
 	require.NoError(t, err)
 
+	//stm @CLIENT_STORAGE_DEALS_LIST_IMPORTS_001
```
magik6k, line 49: Why do some lines use `//stm:`, and others `//stm`?

node/impl/client/client_test.go
```diff
@@ -1,3 +1,4 @@
+//stm: #unit
```
magik6k, line 1: Can we deduce unit/integration based on directory?

node/impl/client/client_test.go
brdji: The correct form is `//stm:`. The other lines are typos, and will be fixed.

node/impl/client/client_test.go
brdji: I think it's possible to do so (ie. tests in the `itests` folder are integration tests), though I'm not sure if anything more than that is too fragile and inflexible. 
I would suggest something different: the crawler can add an annotation `stm: #TODO` to the top of each test file that does not already have a specified annotation. These TODO's can then be shown as warnings in the dashboard, or in the CI, etc. This way, the developer is forced to fill in the correct type.

magik6k (approved):

PR #7851, opened 2021-12-28
fix: doc generation struct slice example value

zl03jsj:
as you can see, all slice type example value is null in document:
https://github.com/filecoin-project/lotus/blob/921fda94c720fb494d1d33c310404e62e04d9ebe/documentation/en/api-v1-unstable-methods.md?plain=1#L616-L651

fix: ExampleValue for a slice would be nil

add json.RawMessage ExampleValue.

the AlertEvent.Message needs a json.RawMessage example value, or would panic with json.MarshalJSON error.
https://github.com/filecoin-project/lotus/blob/921fda94c720fb494d1d33c310404e62e04d9ebe/journal/alerting/alerts.go#L32-L36


magik6k (commented):
Thanks!

(Needs `make docsgen` to regenerate openrpc blobs)

zl03jsj:
> make docsgen

done

magik6k (approved):
thanks!

PR #7871, opened 2022-01-04
fix: #7869 sealing: Add more deal expiration checks during PRU pipeline

### LINKED ISSUE #7869: Fail early if the deals are expired during PRU handlings
like in `handleReplicaUpdate`, we should add more deal error/expiration handling in `handleProveReplicaUpdate1` and `handleProveReplicaUpdate2`

jennijuju:
## Related Issues

resolves #7869 

## Proposed Changes

add more deal error handling during PRU pipeline so that we can fail early without  wasting too much work 

## Checklist

- [x] The PR title is in the form of of `<PR type>: <#issue number> <area>: <change being made>`
    - example: ` fix: #1234 mempool: Introduce a cache for valid signatures`
- [ ] CI is green

### DIFF
--- a/extern/storage-sealing/checks.go
+++ b/extern/storage-sealing/checks.go
@@ -79,24 +79,6 @@ func checkPieces(ctx context.Context, maddr address.Address, si SectorInfo, api 
 	return nil
 }
 
-func checkDealExpiration(ctx context.Context, sector SectorInfo, api SealingAPI) error {
-	tok, height, err := api.ChainHead(ctx)
-	if err != nil {
-		return &ErrApi{xerrors.Errorf("getting chain head: %w", err)}
-	}
-
-	for i, p := range sector.Pieces {
-		proposal, err := api.StateMarketStorageDealProposal(ctx, p.DealInfo.DealID, tok)
-		if err != nil {
-			return &ErrInvalidDeals{xerrors.Errorf("getting deal %d for piece %d: %w", p.DealInfo.DealID, i, err)}
-		}
-		if height >= proposal.StartEpoch {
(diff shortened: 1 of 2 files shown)


extern/storage-sealing/checks.go
```diff
@@ -80,6 +79,24 @@ func checkPieces(ctx context.Context, maddr address.Address, si SectorInfo, api
 	return nil
 }
 
+func checkDealExpiration(ctx context.Context, sector SectorInfo, api SealingAPI) error {
```
arajasek, line 82: I don't think this is sufficiently different or faster than checkPieces to warrant its own method?

arajasek (commented):
1) i don't know how long `ProveReplicaUpdate1` takes, but it may be rerunning the check before calling `ProveReplicaUpdate2`?
2) what we really want is probably to extend `SectorInfo` to have a `deadline` epoch set to its latest deal activation (can be updated along the way as precommit and so on happens). then, every state transition can check if the sector is doomed before investing any more time in it. 

but that is outside the scope of this pr -- an issue would be nice to have, though.

extern/storage-sealing/checks.go
jennijuju: > 1. i don't know how long `ProveReplicaUpdate1` takes, but it may be rerunning the check before calling `ProveReplicaUpdate2`?

PRU1 computes the vanilia proofs so yah it should take a while, added another check before the PRU2

jennijuju: tho i do think it makes sense to check before PRU2 instead of PRU1 - updated

jennijuju:
> 2. what we really want is probably to extend `SectorInfo` to have a `deadline` epoch set to its latest deal activation (can be updated along the way as precommit and so on happens). then, every state transition can check if the sector is doomed before investing any more time in it.

i understand what you are proposing wrt the adding to sector info, but the earliest deal start epoch should stay the same after the snap deal sector is packed, so it wont be updated once set?

arajasek (approved):

PR #7883, opened 2022-01-04
feat: paychmgr: Support paych funding (a.k.a. fast paid retrieval)

magik6k:
Closes https://github.com/filecoin-project/lotus/issues/7876


paychmgr/simple.go
```diff
@@ -492,6 +567,8 @@ func (ca *channelAccessor) addFunds(ctx context.Context, channelInfo *ChannelInf
 	return &mcid, nil
 }
 
+// TODO func (ca *channelAccessor) freeFunds(ctx context.Context, channelInfo *ChannelInfo, amt, avail types.BigInt) (*cid.Cid, error) {
```
magik6k, line 570: Want before landing, needed for stuff to work well with how we fund partial retrievals (which we fund like a full retrieval)

magik6k: ((apparently we don't have good market hooks for this; can be done later))

markets/retrievaladapter/client.go
```diff
@@ -9,23 +9,54 @@ import (
...
+	"golang.org/x/xerrors"
 
+	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/paych"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/node/impl/full"
 	payapi "github.com/filecoin-project/lotus/node/impl/paych"
 )
 
+func mkPaychReusedCid(addr address.Address) cid.Cid {
```
hannahhoward, line 22: I feel somewhat strongly that rather than do all this hackery with addresses and cids, we should properly handle cid.Undef over in markets. To that end, I've written a PR to do so -- https://github.com/filecoin-project/go-fil-markets/pull/677

api/api_full.go
```diff
@@ -689,7 +689,15 @@ type FullNode interface {
 	// MethodGroup: Paych
 	// The Paych methods are for interacting with and managing payment channels
 
-	PaychGet(ctx context.Context, from, to address.Address, amt types.BigInt) (*ChannelInfo, error)                     //perm:sign
+	// PaychGet gets or creates a payment channel between address pair
```
hannahhoward, line 692: If I am reading the code correctly, we have two bools, but actually one of the combinations is invalid and always errors (reserve = false, off chain = true). I wonder if instead of two bools, this should just be some kind of typed integer to represent the 3 available modes that one can actually use. (on-chain no reserve, on-chain reserve, off-chain reserve)

paychmgr/simple.go
```diff
@@ -452,15 +564,41 @@ func (ca *channelAccessor) waitPaychCreateMsg(ctx context.Context, channelID str
...
+		channelInfo.PendingAvailableAmount = big.NewInt(0)
 		channelInfo.CreateMsg = nil
 	})
 
 	return nil
 }
 
+// completeAvailable fills reserving fund requests using already available funds, without interacting with the chain
+func (ca *channelAccessor) completeAvailable(ctx context.Context, merged *mergedFundsReq, channelInfo *ChannelInfo, amt, av types.BigInt) (*paychFundsRes, types.BigInt) {
+	toReserve := types.BigSub(amt, av)
```
hannahhoward, line 578: as I read this code, if we have available funds already confirmed, but we also have a no-reserve request ready to go that could supply the available funds, we don't use the available funds. I'm curious about the rationale here. Why not just grab the available funds and then let the no-reserve request complete on its own time?

paychmgr/store.go
```diff
@@ -376,14 +379,29 @@ func (ps *Store) GetMessage(ctx context.Context, mcid cid.Cid) (*MsgInfo, error)
...
+func (ps *Store) OutboundActiveByFromTo(ctx context.Context, sma stateManagerAPI, from address.Address, to address.Address) (*ChannelInfo, error) {
 	return ps.findChan(ctx, func(ci *ChannelInfo) bool {
 		if ci.Direction != DirOutbound {
 			return false
 		}
 		if ci.Settling {
 			return false
 		}
+
+		if ci.Channel != nil {
```
hannahhoward, line 391: Is this just an unrelated bug fix?

api/api_full.go
```diff
@@ -850,6 +863,10 @@ type ChannelAvailableFunds struct {
 	ConfirmedAmt types.BigInt
 	// PendingAmt is the amount of funds that are pending confirmation on-chain
 	PendingAmt types.BigInt
+	// AvailableAmt is part of ConfirmedAmt that is available for use (pre-allocated)
+	AvailableAmt types.BigInt
+	// PendingAvailableAmt is the amount of available funds that are pending confirmation on-chain
```
hannahhoward, line 868: To clarify this is a portion of the "PendingAmt" not the "AvailableAmt"

```suggestion
// PendingAvailableAmt is the amount of funds that are pending confirmation on-chain that will become available once confirmed
```

hannahhoward (commented):
This is awesome and much needed.

Various suggestions and changes, though I'm not sure whether any of them are truly blocking.

I do think it would be good possibly to incorporate the markets change.

paychmgr/simple.go
```diff
@@ -135,18 +151,97 @@ func (m *mergedFundsReq) onComplete(res *paychFundsRes) {
...
+		if r.amt.GreaterThan(types.BigSub(avail, used)) {
+			// requests are sorted by amount ascending, so if we hit this, there aren't any more requests we can fill
+
+			if r.opts.OffChain {
+				// can't fill, so OffChain want an error
+				if r.isActive() {
+					failed = types.BigAdd(failed, r.amt)
+					r.onComplete(&paychFundsRes{
+						channel: *channelInfo.Channel,
+						err:     xerrors.Errorf("not enough funds available in the payment channel %s; add funds with 'lotus paych add-funds %s %s %s'", channelInfo.Channel, channelInfo.from(), channelInfo.to(), types.FIL(r.amt).Unitless()),
```
dirkmc, line 191: Suggest simplifying this error message - the message only makes sense if people are calling the API from the command line

paychmgr/simple.go
```diff
@@ -135,18 +151,97 @@ func (m *mergedFundsReq) onComplete(res *paychFundsRes) {
...
+				continue
+			}
+
+			break
+		}
+
+		// don't try to fill inactive requests
+		if !r.isActive() {
+			continue
+		}
```
dirkmc, line 204: Should this check against `r.isActive()` be before the check against `r.amt.GreaterThan(types.BigSub(avail, used))`?

paychmgr/store.go
```diff
@@ -74,6 +73,10 @@ type ChannelInfo struct {
 	// has locally been added to the channel. It should reflect the channel's
 	// Balance on chain as long as all operations occur on the same datastore.
 	Amount types.BigInt
+	// AvailableAmount indicates how much afil is non-reverved
```
dirkmc, line 76:
```suggestion
	// AvailableAmount indicates how much afil is non-reserved
```

api/api_full.go
```diff
@@ -850,6 +863,10 @@ type ChannelAvailableFunds struct {
 	ConfirmedAmt types.BigInt
 	// PendingAmt is the amount of funds that are pending confirmation on-chain
 	PendingAmt types.BigInt
+	// AvailableAmt is part of ConfirmedAmt that is available for use (pre-allocated)
+	AvailableAmt types.BigInt
```
dirkmc, line 867: I suggest re-wording these comments to emphasize the difference between reserved and non-reserved funds. It may be worth using some longer names for the variables themselves to make that clear

Angelo-gh3990:
I get : ERROR: Remote API version didn't match (expected 1.5.0, remote 1.4.0) when I start lotus-miner when I have compiled lotus with this version as well. Even tried compiling Lotus with the master branch. Anyway to get this to work, testing paid retrieval with Elijah. Thanks

paychmgr/store.go
magik6k: Yes, this would break your paychmgr after the channel got settled on-chain

paychmgr/simple.go
magik6k: The call to `completeAmount` below should be doing this.

magik6k:
> Anyway to get this to work, testing paid retrieval with Elijah

Note that this is client-side only change. You don't need to update anything on the miner side to get fast paid retrievals to work.

dirkmc (approved):

hannahhoward (approved):
LGTM -- this hasn't changed much since i last looked but the changes look good

PR #7901, opened 2022-01-07
feat: state: Fast migration for v15

### LINKED ISSUE #7870: v7 migration integration
integrate the actor v7 migration mechansim

arajasek:
## Related Issues

Fixes #7870 

## Checklist

- [x] The PR title is in the form of of `<PR type>: <#issue number> <area>: <change being made>`
    - example: ` fix: #1234 mempool: Introduce a cache for valid signatures`
- [ ] CI is green

### DIFF
--- a/chain/consensus/filcns/upgrades.go
+++ b/chain/consensus/filcns/upgrades.go
@@ -5,6 +5,8 @@ import (
 	"runtime"
 	"time"
 
+	autobatch "github.com/application-research/go-bs-autobatch"
+
 	"github.com/filecoin-project/specs-actors/v6/actors/migration/nv14"
 	"github.com/filecoin-project/specs-actors/v7/actors/migration/nv15"
 
@@ -1245,8 +1247,15 @@ func PreUpgradeActorsV7(ctx context.Context, sm *stmgr.StateManager, cache stmgr
 		workerCount /= 2
 	}
 
-	config := nv15.Config{MaxWorkers: uint(workerCount)}
-	_, err := upgradeActorsV7Common(ctx, sm, cache, root, epoch, ts, config)
+	lbts, lbRoot, err := stmgr.GetLookbackTipSetForRound(ctx, sm, ts, epoch)
+	if err != nil {
+		return xerrors.Errorf("error getting lookback ts for premigration: %w", err)
+	}
+
+	config := nv15.Config{MaxWorkers: uint(workerCount),
+		ProgressLogPeriod: time.Minute * 5}
+
+	_, err = upgradeActorsV7Common(ctx, sm, cache, lbRoot, epoch, lbts, config)
 	return err
 }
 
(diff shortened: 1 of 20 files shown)


arajasek:
Some more experimenting and tweaking is needed, but this gets the job done well. Will share latest numbers soon (roughly 35 mins for premigration, 3 mins for migration on my box).

chain/consensus/filcns/upgrades.go
```diff
@@ -1245,8 +1247,15 @@ func PreUpgradeActorsV7(ctx context.Context, sm *stmgr.StateManager, cache stmgr
 		workerCount /= 2
 	}
 
-	config := nv15.Config{MaxWorkers: uint(workerCount)}
-	_, err := upgradeActorsV7Common(ctx, sm, cache, root, epoch, ts, config)
+	lbts, lbRoot, err := stmgr.GetLookbackTipSetForRound(ctx, sm, ts, epoch)
```
arajasek, line 1250: This migration requires us to be reorg-insensitive, so the premigration has to be based on finality.

chain/consensus/filcns/upgrades.go
```diff
@@ -1255,7 +1264,12 @@ func upgradeActorsV7Common(
 	root cid.Cid, epoch abi.ChainEpoch, ts *types.TipSet,
 	config nv15.Config,
 ) (cid.Cid, error) {
-	buf := blockstore.NewTieredBstore(sm.ChainStore().StateBlockstore(), blockstore.NewMemorySync())
+	writeStore, err := autobatch.NewBlockstore(sm.ChainStore().StateBlockstore(), blockstore.NewMemorySync(), 100_000, 100, true)
```
arajasek, line 1267: 100_000 here wasn't chosen very scientifically...

arajasek:
Update to the tagged actors RC before landing this.

chain/consensus/filcns/upgrades.go
Stebalien: This blockstore:

1. Doesn't guarantee a flush order.
2. Doesn't actually wait in `Flush` for the all data to be written.
3. Definitely wasn't written for this use-case.

3 isn't really an issue, but 1 and 2 are. Ideally the node won't be borked if you crash during a pre-migration.

Stebalien (changes requested):

Stebalien:
> 3 isn't really an issue, but 1 and 2 are. Ideally the node won't be borked if you crash during a pre-migration.

I'd just implement a much simpler version of this. Basically, you need something with 3 stages:

1. Buffered.
2. Flushing.
3. Backing.

Writes go into buffered. When it fills, `put` signals (uses a length-one channel) some background goroutine to flush. That background goroutine will move everything from buffered to flushing, and start writing it to the backing blockstore.

Gets check buffered, then flushing, then backing.

The background goroutine ensures that data is _always_ flushed in-order.

arajasek:
Superseded by #7933

PR #7933, opened 2022-01-11
feat: state: Fast migration for v15

### LINKED ISSUE #7870: v7 migration integration

arajasek:
## Related Issues

fixes #7870 
## Proposed Changes

The chief problem is that we can't hold the entire migrated state in memory (too much changes), but also need it to be fast enough that the premigration doesn't take too long and the migration (ideally) doesn't lead to null blocks.

The newly added autobatch store achieves this with a big caveat: It will take potentially very long to `Get` things that have not already been `Flush`ed. Deletion is not supported at all.
## Checklist

- [ ] CI is green


blockstore/autobatch.go
```diff
@@ -0,0 +1,98 @@
...
+
+// May NOT `Get` blocks that have been `Put` into this store
+// Only guaranteed to fetch those that were already in the backingBs at creation of this store and those at the most recent `Flush`
+func (bs *AutobatchBlockstore) Get(ctx context.Context, c cid.Cid) (block.Block, error) {
+	return bs.backingBs.Get(ctx, c)
+}
+
+func (bs *AutobatchBlockstore) Put(ctx context.Context, blk block.Block) error {
+	bs.bufferedBlksLk.Lock()
+	_, ok := bs.addedCids[blk.Cid()]
```
arajasek, line 75: Idk what the memory impact of this is yet -- I assume it helps perf, though it's fast enough without it.

If we wanted to ameliorate the `Get` may not return blocks that have been newly `Put`ted, we can use this to check _whether_ we should have the block, and then traverse `bs.bufferedBlks`. Atrocious perf, but might as well do it, since we're claiming it should never happen.

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func (bs *AutobatchBlockstore) doFlush(ctx context.Context) error {
+	bs.bufferedLk.Lock()
+	bs.flushingLk.Lock()
+	// We do NOT clear addedCids here, because its purpose is to expedite Puts
+	flushingBlksOrdered := bs.bufferedBlksOrdered
+	bs.flushingBlksMap = bs.bufferedBlksMap
+	bs.bufferedBlksOrdered = []block.Block{}
+	bs.bufferedBlksMap = make(map[cid.Cid]block.Block)
```
Stebalien, line 109: nit: pre-allocate these with the size of the buffer we're replacing. That's a pretty good estimation of how big they'll grow.

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+	return nil, xerrors.New("unsupported")
+
+}
+
+func (bs *AutobatchBlockstore) HashOnRead(enabled bool) {
+	bs.backingBs.HashOnRead(enabled)
+}
+
+func (bs *AutobatchBlockstore) View(ctx context.Context, cid cid.Cid, callback func([]byte) error) error {
+	return xerrors.New("unsupported")
```
Stebalien, line 207: I'd implement this, even if we just call get under the covers.

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+		case <-bs.flushCh:
+			putErr := bs.doFlush(bs.flushCtx)
+			for putErr != nil {
+				select {
+				case <-bs.shutdownCh:
+					bs.flushErr = putErr
+					return
+				default:
+					autolog.Errorf("FLUSH ERRORED: %w, retrying in %v", putErr, bs.flushRetryDelay)
+					time.Sleep(bs.flushRetryDelay)
```
Stebalien, line 92: I'd avoid `time.Sleep` and instead use `time.After` or a timer inside the select.

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+	bs.bufferedBlksOrdered = []block.Block{}
+	bs.bufferedBlksMap = make(map[cid.Cid]block.Block)
+	bs.bufferedLk.Unlock()
+	bs.flushingLk.Unlock()
+	return bs.backingBs.PutMany(ctx, flushingBlksOrdered)
+}
+
+func (bs *AutobatchBlockstore) Shutdown(ctx context.Context) error {
+	// request one last flush of the worker
+	bs.flushCh <- struct{}{}
```
Stebalien, line 117: If we're going to do "one last flush here", there's no need to write to the flush channel.

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+	bs.flushingBlksMap = bs.bufferedBlksMap
+	bs.bufferedBlksOrdered = []block.Block{}
+	bs.bufferedBlksMap = make(map[cid.Cid]block.Block)
+	bs.bufferedLk.Unlock()
+	bs.flushingLk.Unlock()
+	return bs.backingBs.PutMany(ctx, flushingBlksOrdered)
+}
+
+func (bs *AutobatchBlockstore) Shutdown(ctx context.Context) error {
+	// request one last flush of the worker
```
Stebalien, line 116: It's usually nice to make shutdown idempotent. If shutdown is going to do more than just signal then wait, I'd wrap it in a sync.Once.

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+	bs.bufferedLk.Lock()
+	bs.flushingLk.Lock()
+	// We do NOT clear addedCids here, because its purpose is to expedite Puts
+	flushingBlksOrdered := bs.bufferedBlksOrdered
+	bs.flushingBlksMap = bs.bufferedBlksMap
+	bs.bufferedBlksOrdered = []block.Block{}
+	bs.bufferedBlksMap = make(map[cid.Cid]block.Block)
+	bs.bufferedLk.Unlock()
+	bs.flushingLk.Unlock()
+	return bs.backingBs.PutMany(ctx, flushingBlksOrdered)
```
Stebalien, line 112: If this fails, we need to put these blocks back somewhere and need to avoid just trying the next set of blocks.

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+
+type AutobatchBlockstore struct {
+	// TODO: drop if memory consumption is too high
+	addedCids map[cid.Cid]struct{}
+
+	bufferedLk          sync.Mutex
+	bufferedBlksOrdered []block.Block
+	bufferedBlksMap     map[cid.Cid]block.Block
+
+	flushingLk      sync.Mutex
```
Stebalien, line 26: You don't really gain anything by having two locks.

1. Get takes both.
2. doFlush takes both.
3. Put only takes one.

arajasek: Yessir, dropping.

blockstore/autobatch.go
arajasek: Yessir.

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+			putErr := bs.doFlush(bs.flushCtx)
+			for putErr != nil {
+				select {
+				case <-bs.shutdownCh:
+					bs.flushErr = putErr
+					return
+				default:
+					autolog.Errorf("FLUSH ERRORED: %w, retrying in %v", putErr, bs.flushRetryDelay)
+					time.Sleep(bs.flushRetryDelay)
+					putErr = bs.doFlush(bs.flushCtx)
```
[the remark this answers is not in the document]
arajasek: @Stebalien thinks that could happen if the system is under stress -- with some backoff it may succeed?

blockstore/autobatch.go
arajasek: ack

blockstore/autobatch.go
arajasek: I might just drop the methods -- we don't actually need this type to implement Lotus's blockstore interface

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+	go bs.flushWorker()
+
+	return bs
+}
+
+func (bs *AutobatchBlockstore) Put(ctx context.Context, blk block.Block) error {
+	bs.bufferedLk.Lock()
+	_, ok := bs.addedCids[blk.Cid()]
+	if !ok {
+		bs.addedCids[blk.Cid()] = struct{}{}
```
magik6k, line 63: What's the hit rate on this?

blockstore/autobatch.go
```diff
@@ -0,0 +1,208 @@
...
+	// one last flush in case it's needed
+	return bs.doFlush(ctx)
+}
+
+func (bs *AutobatchBlockstore) Get(ctx context.Context, c cid.Cid) (block.Block, error) {
+	// may seem backward to check the backingBs first, but that is the likeliest case
+	blk, err := bs.backingBs.Get(ctx, c)
+	if err == nil {
+		return blk, nil
+	}
```
magik6k, line 134: If we check backingBs isn't this potentially racy with the way we do locks here? If we really want to avoid locking, we probably want to check backingBs once more at the end.

blockstore/autobatch.go
arajasek: Not measured, but I expect it to be quite high -- think of the number of times we'll try to `Put` the empty deadline object.

blockstore/autobatch.go
arajasek: that's a good point -- i'm not super concerned about it because for the migration, we should actually never try to `Get` anything we've `Put`...but a second check of the bs at the end makes sense

blockstore/autobatch.go
magik6k: Yeah, I'm just wondering if it's worth the added memory use (tho if it's insignificant, this should be ok to keep even if it doesn't help that much).

arajasek: Sounds good -- i do have a TODO to drop if the memory use is a problem, we'll see in experiments.

blockstore/autobatch.go
```diff
@@ -0,0 +1,226 @@
...
+		bs.bufferedBatch.blockMap = make(map[cid.Cid]block.Block, len(bs.flushingBatch.blockMap))
+		bs.lock.Unlock()
+	}
+
+	bs.flushErr = bs.backingBs.PutMany(ctx, bs.flushingBatch.blockList)
+	return bs.flushErr
+}
+
+// caller must NOT hold lock
+func (bs *AutobatchBlockstore) Flush(ctx context.Context) error {
```
Stebalien, line 125: This cannot be safely called while the flushWorker is running.

blockstore/autobatch.go
```diff
@@ -0,0 +1,226 @@
...
+	return bs.doFlush(ctx)
+}
+
+func (bs *AutobatchBlockstore) Shutdown(ctx context.Context) error {
+	// shutdown the flush worker
+	bs.shutdownCh <- struct{}{}
```
Stebalien, line 131: Hm. It's unfortunate that this will block indefinitely if we shutdown twice.

blockstore/autobatch.go
```diff
@@ -0,0 +1,226 @@
...
+		return blk, nil
+	}
+
+	if err != ErrNotFound {
+		return blk, err
+	}
+
+	bs.lock.Lock()
+	defer bs.lock.Unlock()
+	v, ok := bs.flushingBatch.blockMap[c]
```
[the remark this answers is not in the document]
arajasek: oop

blockstore/autobatch.go
arajasek: Will try and report back with perf

blockstore/autobatch.go
```diff
@@ -0,0 +1,226 @@
...
+		bs.lock.Lock()
+		// We do NOT clear addedCids here, because its purpose is to expedite Puts
+		bs.flushingBatch = bs.bufferedBatch
+		bs.bufferedBatch.blockList = make([]block.Block, 0, len(bs.flushingBatch.blockList))
+		bs.bufferedBatch.blockMap = make(map[cid.Cid]block.Block, len(bs.flushingBatch.blockMap))
+		bs.lock.Unlock()
+	}
+
+	bs.flushErr = bs.backingBs.PutMany(ctx, bs.flushingBatch.blockList)
+	return bs.flushErr
```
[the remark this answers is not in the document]
arajasek: sure, will do

blockstore/autobatch.go
```diff
@@ -0,0 +1,226 @@
...
+func (bs *AutobatchBlockstore) HashOnRead(enabled bool) {
+	bs.backingBs.HashOnRead(enabled)
+}
+
+func (bs *AutobatchBlockstore) View(ctx context.Context, cid cid.Cid, callback func([]byte) error) error {
+	if err := bs.Flush(ctx); err != nil {
+		return err
+	}
+
+	return bs.backingBs.View(ctx, cid, callback)
```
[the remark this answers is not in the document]
arajasek: Sure

blockstore/autobatch.go
arajasek: Looks like this slowed us down -- trying again without to confirm

arajasek: Confirmed locking in there slows us down from 870 migrations per sec to 350

Stebalien: You kind of need those locks, but they can be rwlocks.

Stebalien: And you mean the locking in _this_ function? Shouldn't we never hit those locks?

Stebalien (approved):
We can iterate on master.

blockstore/autobatch.go
arajasek: if we want to lock to avoid raciness, it has to be at the very top of the method (so always gets hit)

PR #7940, opened 2022-01-12
correctness fixes for the autobatch blockstore

Stebalien:
1. Simplify shutdown and make it idempotent by using a context.
2. Make sure `Flush` actually _fully_ flushes if the previous flush failed.
3. Don't clear the flush batch if flushing fails.

### DIFF
--- a/blockstore/autobatch.go
+++ b/blockstore/autobatch.go
@@ -26,18 +26,17 @@ type AutobatchBlockstore struct {
 	addedCids map[cid.Cid]struct{}
 
 	stateLock     sync.Mutex
-	doFlushLock   sync.Mutex
 	bufferedBatch blockBatch
 
-	flushingBatch   blockBatch
-	flushErr        error
-	flushWorkerDone bool
+	flushingBatch blockBatch
+	flushErr      error
 
 	flushCh chan struct{}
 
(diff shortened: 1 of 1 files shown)


blockstore/autobatch.go
```diff
@@ -88,66 +86,85 @@ func (bs *AutobatchBlockstore) Put(ctx context.Context, blk block.Block) error {
...
-		bs.shutdownCh <- struct{}{}
+	// TODO: Prevent puts after we call this to avoid losing data.
+	bs.shutdown()
+	select {
+	case <-bs.doneCh:
+	case <-ctx.Done():
+		return ctx.Err()
 	}
 
+	bs.doFlushLock.Lock()
```
arajasek, line 165: Is this just to make sure that any earlier claimers of the lock finish before we return?

arajasek (approved):

Stebalien:
(forgot to create the channel)

blockstore/autobatch.go
Stebalien: Concurrent Flush and Shutdown.

jennijuju:
folllowing the PR template - a nice PR title should be created so our change log is pretty

PR #7965, opened 2022-01-17
perf: chain: Disable fsync on the chainstore by default

ribasushi:
Use LOTUS_CHAIN_BADGERSTORE_DISABLE_FSYNC=false to reenable

The amount of PutMany and Put calls caused by IPLD churn is too high for an average
non-nvme SSD to keep up. Since the chainstore is something that does get reset every
so often anyway, it should be acceptable to loosen the crash-resilience in order to
enable more full validating nodes in the wild.

### DIFF
--- a/node/repo/fsrepo.go
+++ b/node/repo/fsrepo.go
@@ -347,10 +347,11 @@ func (fsr *fsLockedRepo) Blockstore(ctx context.Context, domain BlockstoreDomain
 
 		//
 		// Tri-state environment variable LOTUS_CHAIN_BADGERSTORE_DISABLE_FSYNC
-		// - unset == the default (currently fsync enabled)
+		// - unset == the default (currently fsync disabled)
 		// - set with a false-y value == fsync enabled no matter what a future default is
 		// - set with any other value == fsync is disabled ignored defaults (recommended for day-to-day use)
 		//
+		opts.SyncWrites = false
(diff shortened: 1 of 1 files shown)


jennijuju:
There was an opposition to have this exposed as a configurable variable back then https://github.com/filecoin-project/lotus/pull/5050 so im not sure having it as a default is a good idea, can use a re-review here
 and it was also mentioned in https://github.com/filecoin-project/lotus/pull/6817#issuecomment-884784898 that splitstore should cover this, need @vyzo to confirm.

whyrusleeping (approved):
LFG

vyzo:
I think this os a really bad idea, it will be corrupt dbs all over the place.

arajasek:
We can revisit this if it's more urgently needed

BigLep:
Recording some notes from a recent conversation with @ribasushi: "the original concern of 'it may be somewhat unsafe' was valid but this concern goes away when combined with https://github.com/ribasushi/ltsh/commit/744423b5b6cf5"

I'm posting this for if/when this work gets revived.

rvagg:
I'm open to this, but don't feel confident enough to make a call on whether this is a good idea or not without further investigation; is it sufficient to `Flush`? Is doing this via `ValidateChain` and these additional paths sufficient to be "safe"? Can we lock this behaviour in or test it in such a way that if there are future refactors or additional paths are added later on that we ensure we always do this in all the right places into the future?

PR #7890, opened 2022-01-05
Fix: state: circsuypply calc around null blocks

arajasek:
## Related Issues

Bug introduced in #7811 

## Checklist

Before you mark the PR ready for review, please make sure that:
- [x] All commits have a clear commit message.
- [x] The PR title is in the form of of `<PR type>: <#issue number> <area>: <change being made>`
    - example: ` fix: #1234 mempool: Introduce a cache for valid signatures`
    - `PR type`: _fix_, _feat_, _INTERFACE BREAKING CHANGE_, _CONSENSUS BREAKING_, _build_, _chore_, _ci_, _docs_, _misc_,_perf_, _refactor_, _revert_, _style_, _test_
    - `area`: _api_, _chain_, _state_, _vm_, _data transfer_, _market_, _mempool_, _message_, _block production_, _multisig_, _networking_, _paychan_, _proving_, _sealing_, _wallet_
- [ ] This PR has tests for new functionality or change in behaviour
- [ ] If new user-facing features are introduced, clear usage guidelines and / or documentation updates should be included in https://lotus.filecoin.io or [Discussion Tutorials.](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] CI is green

PR #7818, opened 2021-12-18
Refactor: VM: Remove the NetworkVersionGetter

arajasek:
## Related Issues

The primary motivation for this is to support FVM integration, because we don't want to be piping a network version getter across the FFI boundary.

## Proposed Changes

The actual Runtime interface only ever reports the network version at the current epoch to actors. Within the VM too, all but one of the invocations of the version getter is for the current epoch. These can all be satisfied by simply providing the current network version on VM construction.

The exception is the randomness getters, which provide the network version at an arbitrary epoch as input. This PR gets around this by allowing the randomness module to query the network version.

## Checklist

Before you mark the PR ready for review, please make sure that:
- [x] All commits have a clear commit message.
- [x] The PR title is in the form of of `<PR type>: <#issue number> <area>: <change being made>`
    - example: ` fix: #1234 mempool: Introduce a cache for valid signatures`
    - `PR type`: _fix_, _feat_, _INTERFACE BREAKING CHANGE_, _CONSENSUS BREAKING_, _build_, _chore_, _ci_, _docs_, _misc_,_perf_, _refactor_, _revert_, _style_, _test_
    - `area`: _api_, _chain_, _state_, _vm_, _data transfer_, _market_, _mempool_, _message_, _block production_, _multisig_, _networking_, _paychan_, _proving_, _sealing_, _wallet_
- [ ] This PR has tests for new functionality or change in behaviour
- [ ] If new user-facing features are introduced, clear usage guidelines and / or documentation updates should be included in https://lotus.filecoin.io or [Discussion Tutorials.](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [x] CI is green

PR #7966, opened 2022-01-17
Fix: chain: create a new VM for each epoch

arajasek:
## Related Issues

Fixes bug introduced in #7818. The VM's network version was always being determined based on `epoch`, the input to the `ApplyBlocks` method, even though the vm executes (potentially) multiple epochs worth of messages. The result is that if there was a null epoch before the upgrade epoch, the VM used the wrong network version (the new version) for running cron before the migration.

## Proposed Changes

Although this was a fairly specific case, I think this issue shows the risks of trying to hamfistedly reuse the VM for multiple epochs of execution (see #7890 for a similar issue). This PR changes behaviour to make VMs one-per-epoch. We remove the old `SetBlockHeight` method entirely, and always create a new VM when moving between epochs. The result is that fields that need updating between epochs (like network version) are guaranteed to update this way.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/consensus/filcns/compute_state.go
+++ b/chain/consensus/filcns/compute_state.go
@@ -92,16 +92,16 @@ func (t *TipSetExecutor) ApplyBlocks(ctx context.Context, sm *stmgr.StateManager
 		partDone()
 	}()
 
-	makeVmWithBaseState := func(base cid.Cid) (*vm.VM, error) {
+	makeVmWithBaseStateAndEpoch := func(base cid.Cid, e abi.ChainEpoch) (*vm.VM, error) {
 		vmopt := &vm.VMOpts{
 			StateBase:      base,
-			Epoch:          epoch,
+			Epoch:          e,
 			Rand:           r,
 			Bstore:         sm.ChainStore().StateBlockstore(),
 			Actors:         NewActorRegistry(),
 			Syscalls:       sm.Syscalls,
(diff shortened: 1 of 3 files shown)


chain/consensus/filcns/compute_state.go
```diff
@@ -126,59 +121,58 @@ func (t *TipSetExecutor) ApplyBlocks(ctx context.Context, sm *stmgr.StateManager
...
+
 			// run cron for null rounds if any
-			if err := runCron(i); err != nil {
-				return cid.Undef, cid.Undef, err
+			if err = runCron(vmCron, i); err != nil {
+				return cid.Undef, cid.Undef, xerrors.Errorf("running cron: %w", err)
 			}
 
-			pstate, err = vmi.Flush(ctx)
+			pstate, err = vmCron.Flush(ctx)
```
arajasek, line 154: I don't like the repeated use of the same `pstate` and `err` here -- would appreciate suggestions on how to clean this up / derisk this. Winding up with the wrong field in `pstate` would be very bad.

Stebalien: We can always use a temporary then set `pstate = newState`. But this doesn't seem _too_ bad.

Stebalien (approved):
Seems like a reasonable change. I can't find anything wrong with it, at least.

magik6k (approved):
Seems to make sense

PR #7969, opened 2022-01-18
libp2p v0.18.0-rc1 and initial resource manager integration

vyzo:
This upgrades libp2p to v0.18.0-rc1 (and pubsub to v0.6.1 fixing a leak while at it), and provides initial resource manager integration.

~Note: there is a (non catastrophic) bug in our memory accounting somewhere in the stack, which results in log warnings about too much memory released.
This can be ignored for now, but I will try to get it fixed asap.~

Edit: the bug has been fixed in go-yamux v3.0.2

-------------------------------------------------------------------------------------------

## The libp2p Resource Manager

With libp2p-v0.18.0-rc1 we introduce the resource manager, which accounts for and limits and resources used by the libp2p network stack. 
See [the resource manager repo](https://github.com/libp2p/go-libp2p-resource-manager) for design notes and implementation.

This is the initial integration, whereby we initialize a resource manager and provide APIs for inspecting resource usage and setting limits.
In a subsequent pr we will also augment lotus network services to make their resources accountable by the resource manager by organizing them in logical services.

## New APIs/Commands

We add the following APIs/cli commands:
- `lotus net stat <scope>` -- this allows you to inspect current resource usage for a scaope. The scope can be `all`, `system`, `transient`, `svc:<service-name>`, `proto:<protocol-name>`, or `peer:<peer-name>`
- `lotus limit [--set] <scope> [<limit>]` -- this allows you to inspect or set the current limit for a scope. The scope can be as above while the `limit` is a json representation of a limit; see below for the format.

## Limit Configuration

The user can override the default limits by placing a `limits.json` file in `.lotus`.
This file has the following format, with all limits optional and retaining their default value if omitted:
```
{
  "System": <limit>    // system-wide limits; optional
  "Transient": <limit> // transient resource limits

 "ServiceDefault": <limit> // default service limits
 "ServicePeerDefault": <limit> // default per service peer limit
 "Service": { <service>: <limit> ...} // service specific limits
 "ServicePeer": { <service>: <limit> ...} // per service peer limits

 "ProtocolDefault": <limit> // default protocol limits
 "ProtocolPeerDefault": <limit> // default per protocol peer limit
 "Protocol": { <protocol>: <limit> ...} // protocol specific limits
 "ProtocoPeerl": { <protocol>: <limit> ...} // per protocol peer limits

 "PeerDefault": <limit> // default peer limits
 "Peer" : {<peer>: <limit> ...} // peer specific limits

 "Conn": <limit> // connection scope limits; you shouldn't have to modify this.
 "Stream": <limit> // stream scope limits; you shouldn't have to modify this.
}
```

Limit objects specify memory and stream/connection/fd limits. Memory can be either a fixed number (in bytes), a fraction (with min and max in bytes) of total of total memory, or a dynamically computed fraction (with min and max in bytes) of free memory at the time of check.
The object has the following format in json:
```
{
  // memory limit
  "Memory": <int> // fixed memory in bytes; if unset the memory is computed.
  "Dynamic": <bool> // whether the computed memory is static or dynamic
  "MemoryFraction": <float> // fraction of total/free to use 
  "MinMemory": <int> // min memory in bytes
  "MaxMemory": <int> // max memory in bytes

  // network limits
  "Streams": <int> // max total number of streams
  "StreamsInbound": <int> // max inbound streams
  "StreamsOutbound": <int> // max outbound streams
  "Conns": <int> // max total number of connections
  "ConnsInbound": <int> // max inbound connections
  "ConnsOutbound": <int> // max outbound connections
   "FD": <int> // max number of file descriptors
}
```


vyzo:
The bug was in yamux memory accounting, fix is here: https://github.com/libp2p/go-yamux/pull/77

node/modules/lp2p/rcmgr.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+func ResourceManager(lc fx.Lifecycle, repo repo.LockedRepo) (network.ResourceManager, error) {
+	var limiter *rcmgr.BasicLimiter
+	var opts []rcmgr.Option
+
+	repoPath := repo.Path()
+
+	// create limiter -- parse $repo/limits.json if exists
+	limitsFile := filepath.Join(repoPath, "limits.json")
```
magik6k, line 25: Feels like something that should be a part of the libp2p config section, not it's own file.

https://github.com/filecoin-project/lotus/blob/c0f29ea9c6d188f4b4dc997a287bcdb560af7f48/node/config/types.go#L303

cli/net.go
```diff
@@ -606,3 +608,86 @@ var NetBlockListCmd = &cli.Command{
 		return nil
 	},
 }
+
+var NetStatCmd = &cli.Command{
+	Name:      "stat",
+	Usage:     "report resource stat for a scope",
+	ArgsUsage: "scope",
```
magik6k, line 615: Can we add some `Description` of valid scopes?

magik6k (commented):
Looks mostly good. Not sure why linter complains. Also need more `make gen`

node/modules/lp2p/rcmgr.go
vyzo: sure, can do.

cli/net.go
vyzo: sure!

node/modules/lp2p/rcmgr.go
vyzo: oh you want to embed the limits in the config?
I am not sure if it is such a great idea, but i could export the struct used by the parser to embed it.

However, i really prefer to have it separate for easy sharing and modification.

cli/net.go
vyzo: done.

vyzo:
That linter seems psychotic -- it is failing because of a test file in go-fil-markets that has broken because of mocknet api changes in go-libp2p.

magik6k (approved):

PR #7980, opened 2022-01-19
Update to go-data-transfer v1.14 with reverted feature set

hannahhoward:
Reverts the missing blocks feature from go-data-transfer


vyzo (changes requested):
this conflicts with the libp2p update pr and inadvertently pulls it in.

vyzo:
It eithe has to wait until after the libp2p pr or it needs to be pulled in there, so consider retargetting the pr.

vyzo:
this will be picked up in the libp2p pr

PR #7987, opened 2022-01-21
feat: sealer: some replica update improvements

arajasek:
See commit messages


extern/storage-sealing/states_replica_update.go
```diff
@@ -59,10 +59,6 @@ func (m *Sealing) handleSubmitReplicaUpdate(ctx statemachine.Context, sector Sec
 		return nil
 	}
 
-	if err := checkPieces(ctx.Context(), m.maddr, sector, m.Api); err != nil { // Sanity check state
```
Stebalien, line 62: double checking that you intended to remove this because it's not described in the commit

Stebalien (approved):
I'm not sure if I'm fully qualified to review this, but it looks correct.

Is there any way to write a quick test case?

extern/storage-sealing/states_replica_update.go
arajasek: Yes, I did: the subsequent call to checkReplicaUpdate calls checkPieces as the first thing it does

PR #7996, opened 2022-01-24
tvx: supply network version when extracting messages.

raulk:
This PR fixes the `tvx` tool to pass in the relevant network version when extracting test vectors.

### DIFF
--- a/cmd/tvx/extract_message.go
+++ b/cmd/tvx/extract_message.go
@@ -8,12 +8,11 @@ import (
 	"io"
 	"log"
 
-	"github.com/filecoin-project/lotus/api/v0api"
-
 	"github.com/fatih/color"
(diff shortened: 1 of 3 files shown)


cmd/tvx/extract_message.go
```diff
@@ -43,6 +42,15 @@ func doExtractMessage(opts extractOpts) error {
 		return fmt.Errorf("failed to resolve message and tipsets from chain: %w", err)
 	}
 
+	// Assumes that the desired message isn't at the boundary of network versions.
+	// Otherwise this will be inaccurate. But it's such a tiny edge case that
```
arajasek, line 46: These are actually a pretty key edge case that we need to be testing, esp what happens around null tipsets...but that's for later...

arajasek (approved):

PR #7999, opened 2022-01-25
some basic splitstore refactors

vyzo:
Preliminaries for implementing additional functionality on top (sortless compaction, gc, etc), this performs some basic refactoring to avoid polluting ballooning subsequent prs:
- The distinction between MarkSet and MarkSetVisitor is removed.
- SetConcurrent is removed from MarkSet interface; all mark sets must be concurrent, because ...
- Chain walking is now parallelized

### DIFF
--- a/blockstore/splitstore/splitstore_compact.go
+++ b/blockstore/splitstore/splitstore_compact.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"runtime"
 	"sort"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -306,7 +307,7 @@ func (s *SplitStore) doTxnProtect(root cid.Cid, markSet MarkSet) error {
 
 	// Note: cold objects are deleted heaviest first, so the consituents of an object
 	// cannot be deleted before the object itself.
-	return s.walkObjectIncomplete(root, tmpVisitor(),
+	return s.walkObjectIncomplete(root, newTmpVisitor(),
 		func(c cid.Cid) error {
 			if isUnitaryObject(c) {
 				return errStopWalk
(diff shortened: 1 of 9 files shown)


blockstore/splitstore/splitstore_compact.go
```diff
@@ -695,17 +703,42 @@ func (s *SplitStore) walkChain(ts *types.TipSet, inclState, inclMsgs abi.ChainEp
 		// the walk is BFS, so we can reset the walked set in every iteration and avoid building up
 		// a set that contains all blocks (1M epochs -> 5M blocks -> 200MB worth of memory and growing
 		// over time)
-		walked = cid.NewSet()
+		walked = newConcurrentVisitor()
 		walking := toWalk
 		toWalk = nil
```
magik6k, line 708: Technically we don't need this copy; `walking` could be set to nil after populating `workch`

magik6k (approved):
Don't see anything obviously wrong, just one tiny nitpick

blockstore/splitstore/splitstore_compact.go
vyzo: ah yes, and we can reuse the slice; let me fix.

vyzo: hrm, we can't actually reuse the slice, it first comes from the tipset; will just setr to nil after copying.

vyzo: done.

vyzo: you know, we can first make a new one and copy when we initialize and then reuse; that will save somr allocations.
let me do.

vyzo: done done.

magik6k (approved):

Issue #8011, opened 2022-01-30
(SnapDeals) once a sector is marked for snap, new deals will be added to it even the sector is inactive


Reiers:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [X] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon:  1.14.0-rc5+butterflynet+git.40485eb88+api1.5.0
Local: lotus version 1.14.0-rc5+butterflynet+git.40485eb88
```

### Describe the Bug

If you sectors snap-up -> miss wdPost -> get a deal -> that deal will go inside the non-active sector -> and you are left with:

(SEE LOG)

on loop. This has potential to bring the cluster down.
[...]

PR #8014, opened 2022-02-01
fix: sealing: Stop recovery attempts after fault

ZenGround0:
## Related Issues

#8011

## Proposed Changes

This addresses the major concern behind #8011 -- too many resources being used when retrying submission of a replica update message after fault.  

The best solution available now is to check for the fault condition after submit replica update has failed.  

It is understandable to want to stop including faulty sectors at deal inclusion as the name of 8011 suggests.   But that is less good.  Faulting could happen immediately after the last deal was included and the scheduler would be in the same bad state.

The totally correct thing that we should work towards in the future is interrupting the scheduler with an fsm event triggered by listening to the chain for faults (and expirations extensions, etc) happening.  But that is too much refactor for right now.

Fun alternate idea: [this ancient todo](https://github.com/filecoin-project/lotus/blob/master/extern/storage-sealing/states_failed.go#L24) about exponential backoff would address this particular bug too.

## Checklist

- [ ] CI is green

### DIFF
--- a/extern/storage-sealing/upgrade_queue.go
+++ b/extern/storage-sealing/upgrade_queue.go
@@ -3,6 +3,7 @@ package sealing
 import (
 	"context"
 
+	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
 	market7 "github.com/filecoin-project/specs-actors/v7/actors/builtin/market"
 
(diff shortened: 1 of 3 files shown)


jennijuju:
+1 on _Fun alternate idea: this ancient todo about exponential backoff would address this particular bug too._ we really should do a TODO batch sometime soon hahah all our TODOs make so much sense to do!


what's in this PR makes sense to me

tho I think for #8011 we should also at least add the following for v1.14.0 : 
- don't add piece to inactive sector and find a good sector: this can prevent  a good amount of good data transfer work + PSD going into waste
-  do sector activeness check before PRU2 to save gpu work

ZenGround0:
> don't add piece to inactive sector and find a good sector: this can prevent a good amount of good data transfer work + PSD going into waste (since we don't reassign pieces atm)

I'm not enthusiastic about adding another patch to cover this edge case.  It piles on more complexity and an expensive api call to every sector deal match.  Much better would be to focus efforts on reassignment of pieces after the upgrade is aborted or precommit fails.

> do sector activeness check before PRU2 to save gpu work

This I like, I'll add this to the PR.

extern/storage-sealing/upgrade_queue.go
```diff
@@ -110,6 +103,22 @@ func (m *Sealing) MarkForSnapUpgrade(ctx context.Context, id abi.SectorNumber) e
 	return m.sectors.Send(uint64(id), SectorStartCCUpdate{})
 }
 
+func sectorActive(ctx context.Context, api SealingAPI, maddr address.Address, tok TipSetToken, sector abi.SectorNumber) (bool, error) {
+	active, err := api.StateMinerActiveSectors(ctx, maddr, tok)
+	if err != nil {
+		return false, xerrors.Errorf("failed to check active sectors: %w", err)
+	}
+	// Ensure the upgraded sector is active
```
arajasek, line 111:
```suggestion
	// Check if sector is among active sectors
```

arajasek (approved):

PR #8016, opened 2022-02-02
add indexProvider host to markets

nonsense:


documentation/en/default-lotus-miner-config.toml
```diff
@@ -258,38 +258,24 @@
 
 
 [IndexProvider]
-  # The maximum number of multihash chunk links that index provider cache can store before
-  # LRU eviction. If chunks belonging to a single advertisement are larger than the cache can
-  # hold, the cache is resized to be able to hold all links. The actual disk usage depends on
-  # LinkedChunkSize and the length of multihashes. For example, for 128-bit long multihashes
-  # with the default LinkedChunkSize, and LinkCacheSize the cache size can grow to 256MiB. 
-  #
-  # type int
```
masih, line 267: How can I make sure these are added via the auto-generator? would be a shame to lose them

jennijuju: `make gen` Should do it

masih: I think the issue is I "expertly" added these comments manually to the generated files, and running `make gen` removes them :)

jennijuju: Ah I see - I think you need to add it here https://github.com/filecoin-project/lotus/blob/6ef0c382c3445595406b14e22f56d1ec4c4056b3/node/config/types.go#L164, whcih is where  make gen reads from

jennijuju: Default values should be added here for the fields https://github.com/filecoin-project/lotus/blob/6ef0c382c3445595406b14e22f56d1ec4c4056b3/node/config/def.go#L184

masih: The thing that's different about that config struct is that it embeds another struct, and I am not sure if the [generators support embedded](https://github.com/filecoin-project/lotus/pull/7313#discussion_r798537545) structs.

aarshkshah1992:
This looks correct. Need to test it on Sofia.

PR #8017, opened 2022-02-02
test: mempool: Add unit and integration tests

brdji:
## Related Issues 
There were 0 existing messagepool integration tests. Since the pool was only being tested as a side-effect of other features and integration tests, adding tests targeting only the messagepool will allow easier debugging, code safety, and improve coverage.

## Proposed Changes
Add more unit and integration tests covering untested behaviors: 
- Uncovered lines (e.g. `check_test.go`, and `Add` message tests in `messagepool_test.go`)
- Uncovered integration behaviors: `batch push`, `remove`, etc.

## Checklist

- [ ] CI is green


chain/messagepool/check_test.go
```diff
@@ -0,0 +1,224 @@
+//stm: #unit
```
ZenGround0, line 1: What's the meaning and purpose behind these stm comments?

chain/messagepool/messagepool_test.go
```diff
@@ -760,3 +764,298 @@ func TestUpdates(t *testing.T) {
...
+	to := mock.Address(1001)
+
+	{
+		// create a valid messages
+		sm := makeTestMessage(w, from, to, 0, 50_000_000, minimumBaseFee.Uint64())
+		mustAdd(t, mp, sm)
+
+		// try to add it twice
+		err = mp.Add(context.TODO(), sm)
+		assert.Contains(t, err.Error(), "with nonce 0 already in mpool")
```
vyzo, line 934: there has to be a better way than parsing error messages.

chain/messagepool/check_test.go
TheDivic: Those annotations are picked up by a custom test crawler that the Bloxico team implemented as a part of the [System Test Matrix](https://github.com/filecoin-project/system-test-matrix) dashboard. Here's what the [dashboard](https://lotus.systemtestmatrix.com/) looks like. 
It's something we've been working on for the previous couple of months, guided by @raulk and @BlocksOnAChain .
The plan is to keep all the tests annotated from now on, because it gives us a lot more context about the kinds of tests we have and the behaviors that are tested inside of those test than simple code coverage provides.

chain/messagepool/messagepool_test.go
TheDivic: I agree. I think just checking for the existence of the error is good enough, until we figure out a better way.

itests/mempool_test.go
```diff
@@ -0,0 +1,455 @@
...
+	for _, lookMsg := range sms {
+		msgLookup, err := firstNode.StateWaitMsg(ctx, lookMsg.Cid(), 3, api.LookbackNoLimit, true)
+		require.NoError(t, err)
+		require.NotNil(t, msgLookup)
+	}
+}
+
+func TestMemPoolPushTwoNodes(t *testing.T) {
+	//stm: @CHAIN_MEMPOOL_CREATE_MSG_CHAINS_001, @CHAIN_MEMPOOL_SELECT_001
+	//stm: @CHAIN_MEMPOOL_PENDING_001, @CHAIN_STATE_WAIT_MSG_001, @CHAIN_MEMPOOL_CAP_GAS_FEE_001
```
TheMenko, line 93: I think grouping annotations on start of the function is not the best idea.  
While it works with STM, they would better show where the particular test case is in the test function  
if they were placed just above the said test case.

chain/messagepool/check_test.go
ZenGround0: Thanks I knew there'd be a story behind it!

itests/mempool_test.go
ZenGround0: I'll hold off on merge until you guys follow up on this

TheDivic: @TheMenko Generally that's the best approach, but in this particular case those behaviors are happening "under the hood" (deeper in the call chain) so this is kinda the best way to write these annotations.

ZenGround0 (changes requested):
Since your last commit had no logic change I think this mean one of your tests is flaky

ZenGround0:
I have the ability to rerun *failures* but since the test is flaky that won't do much good because a failure always means you need another push, it's rerunning successes that you're interested in.

In general I would run it under right conditions and enough times locally to get a failure reproduction, make the change that removes the flake and try to reproduce again locally if you're unsure.  Usually understanding the root cause of the flake and convincing myself that it's gone is enough for me.

In the rare case local reproduction is just not happening then after making the change I was convinced removed the flake I would make a bunch of noop changes and push enough times to convince myself with high probability the flake was gone.

TheDivic:
@ZenGround0 Thanks for the detailed explanation!  We will find a way to reproduce and fix it locally and we'll get back to you after with an explanation! 🙂

TheDivic:
@ZenGround0 I think this PR is ready to be merged now.
The flake was caused by improper waiting for async operations to complete, using `time.Sleep`. 
I reproduced it locally by running it many times with `go test -count=N`. I used the same approach to validate that it's now fixed.
The solution is explained in detail in my commit messages.
I also extended the `itests/kit` with a useful helper function for debugging and fixing flakes caused by this same reason, as it may be useful in other tests.

itests/kit/circuit.go
```diff
@@ -0,0 +1,34 @@
...
+
+This is how it works:
+	- It runs the `cb` function until it returns true,
+	- waiting for `throttle` duration between each iteration,
+	- or at most `timeout` duration until it breaks test execution.
+
+You can use it if t.Deadline() is not "granular" enough, and you want to know which specific piece of code timed out,
+or you need to set different deadlines in the same test.
+*/
+func CircuitBreaker(t *testing.T, label string, throttle, timeout time.Duration, cb func() bool) {
```
ZenGround0, line 20: Nice!

ZenGround0 (approved):

PR #8029, opened 2022-02-04
splistore cold object reification redux

### LINKED ISSUE #6726: Splitstore: reification of cold objects

vyzo:
Implements cold object reification in the splitstore for VM access, as described in #6726; on top of #8008, cherry picked and adapted from #6914.
Closes #6726

### DIFF
--- a/blockstore/context.go
+++ b/blockstore/context.go
@@ -1,0 +1,21 @@
+package blockstore
+
+import (
+	"context"
+)
+
+type hotViewKey struct{}
+
+var hotView = hotViewKey{}
+
+// WithHotView constructs a new context with an option that provides a hint to the blockstore
+// (e.g. the splitstore) that the object (and its ipld references) should be kept hot.
+func WithHotView(ctx context.Context) context.Context {
+	return context.WithValue(ctx, hotView, struct{}{})
+}
+
+// GetHotView returns true if the hot view option is set in the context
+func GetHotView(ctx context.Context) bool {
+	v := ctx.Value(hotView)
+	return v != nil
+}
--- a/blockstore/context.go
+++ b/blockstore/context.go
@@ -14,8 +14,8 @@ func WithHotView(ctx context.Context) context.Context {
 	return context.WithValue(ctx, hotView, struct{}{})
 }
 
-// GetHotView returns true if the hot view option is set in the context
-func GetHotView(ctx context.Context) bool {
+// IsHotView returns true if the hot view option is set in the context
+func IsHotView(ctx context.Context) bool {
 	v := ctx.Value(hotView)
 	return v != nil
 }
--- a/blockstore/splitstore/splitstore.go
+++ b/blockstore/splitstore/splitstore.go
@@ -161,6 +161,12 @@ type SplitStore struct {
 	txnSyncCond     sync.Cond
 	txnSync         bool
 
+	// background cold object reification
+	reifyMx         sync.Mutex
+	reifyCond       sync.Cond
+	reifyPend       map[cid.Cid]struct{}
+	reifyInProgress map[cid.Cid]struct{}
+
 	// registered protectors
 	protectors []func(func(cid.Cid) error) error
 }
@@ -201,6 +207,10 @@ func Open(path string, ds dstore.Datastore, hot, cold bstore.Blockstore, cfg *Co
 	ss.txnViewsCond.L = &ss.txnViewsMx
 	ss.txnSyncCond.L = &ss.txnSyncMx
 	ss.ctx, ss.cancel = context.WithCancel(context.Background())
+
+	ss.reifyCond.L = &ss.reifyMx
+	ss.reifyPend = make(map[cid.Cid]struct{})
+	ss.reifyInProgress = make(map[cid.Cid]struct{})
 
 	if enableDebugLog {
 		ss.debug, err = openDebugLog(path)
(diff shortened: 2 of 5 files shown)


blockstore/splitstore/splitstore_reify.go
```diff
@@ -0,0 +1,185 @@
...
+		})
+
+	if err != nil {
+		log.Warnf("error walking cold object for reification (cid: %s): %s", c, err)
+		return
+	}
+
+	log.Debugf("reifying %d objects rooted at %s", len(toreify), c)
+
+	batch := make([]blocks.Block, 0, len(toreify))
```
magik6k, line 143: Do we ever worry about this getting too big?

blockstore/splitstore/splitstore_reify.go
```diff
@@ -0,0 +1,185 @@
...
+				return xerrors.Errorf("error checking hotstore: %w", err)
+			}
+
+			if has {
+				if s.txnMarkSet != nil {
+					hasMark, err := s.txnMarkSet.Has(c)
+					if err != nil {
+						log.Warnf("error checking markset: %s", err)
+					} else if hasMark {
+						toforget = append(toforget, c)
```
magik6k, line 123: Isn't `walkObject` parallel? If it is, we'll need locks around those appends.

blockstore/splitstore/splitstore_reify.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+		}
+		if len(totrack) > 0 {
+			if err := s.txnMarkSet.MarkMany(totrack); err != nil {
+				log.Warnf("error marking tracked objects: %s", err)
+			}
+		}
+	} else {
+		if len(toreify) > 0 {
+			s.trackTxnRefMany(toreify)
```
magik6k, line 179: Can `s.txnActive` be false when we are here? If yes, is it ok to ignore those refs? (as trackTxnRefMany will)

blockstore/splitstore/splitstore.go
```diff
@@ -676,6 +706,7 @@ func (s *SplitStore) Close() error {
 		}
 	}
 
+	s.reifyCond.Broadcast()
```
magik6k, line 709: Looks like we're not waiting for reify workers to finish here.. but that's fiine I guess?

blockstore/context.go
```diff
@@ -0,0 +1,21 @@
...
+var hotView = hotViewKey{}
+
+// WithHotView constructs a new context with an option that provides a hint to the blockstore
+// (e.g. the splitstore) that the object (and its ipld references) should be kept hot.
+func WithHotView(ctx context.Context) context.Context {
+	return context.WithValue(ctx, hotView, struct{}{})
+}
+
+// GetHotView returns true if the hot view option is set in the context
+func GetHotView(ctx context.Context) bool {
```
magik6k, line 18: Nit: Maybe `IsHotView` as we're really just checking if we're in one.

magik6k (commented):
Some mostly questions. Would be good to put answers in code comments as well so that future readers will be less confused.

blockstore/splitstore/splitstore_reify.go
vyzo: not really, i havent ever observed reification of more than some 100 objects.

blockstore/splitstore/splitstore_reify.go
vyzo: no, `walkChain` is parallel -- object walkers are serial.

blockstore/splitstore/splitstore_reify.go
vyzo: no, we cant have a txnMarkset if txnActive is false.

vyzo: also we hold the lock and track checks if it is active so we are fine.

vyzo: a comment probably warranted.

blockstore/splitstore/splitstore.go
vyzo: yeah good point.  It is pribably ok, but it will be better to wait for them.

blockstore/context.go
vyzo: sure.

vyzo: renamed to `IsHotView`.

blockstore/splitstore/splitstore.go
vyzo: I added a wait group for the reify workers and wait for them on close.

blockstore/splitstore/splitstore_reify.go
vyzo: added comment.

blockstore/splitstore/splitstore_reify.go
vyzo: added an `isWarm` check in `reifyObject` so that we don't attempt to reify while still warming up.
I will also add a comment here that this is not expected to get big.

PR #8038, opened 2022-02-07
Fix: soup: Downgrade datastore to 0.4.6

arajasek:


jennijuju:
should we ask for a tag for that deps?

PR #8045, opened 2022-02-08
Add indexer pubsub message authentication and rate limiting

gammazero:
This PR authenticates indexer pubsub messages against the blockchain and does rate-limiting of these messages.

## Proposed Changes
- Authenticate that peer that sent the message is the peer that appears in the miner info.
- Rate limit overall messages
- Rate limit repeated announcements

### DIFF
--- a/chain/sub/incoming.go
+++ b/chain/sub/incoming.go
@@ -481,15 +481,15 @@ func (v *IndexerMessageValidator) Validate(ctx context.Context, pid peer.ID, msg
 	// This chain-node should not be publishing its own messages.  These are
 	// relayed from market-nodes.  If a node appears to be local, reject it.
 	if pid == v.self {
-		log.Warnf("refusing to relay indexer message from self")
+		log.Debug("ignoring indexer message from self")
 		stats.Record(ctx, metrics.IndexerMessageValidationFailure.M(1))
-		return pubsub.ValidationReject
+		return pubsub.ValidationIgnore
 	}
 	originPeer := msg.GetFrom()
 	if originPeer == v.self {
-		log.Warnf("refusing to relay indexer message originating from self")
+		log.Debug("ignoring indexer message originating from self")
 		stats.Record(ctx, metrics.IndexerMessageValidationFailure.M(1))
-		return pubsub.ValidationReject
+		return pubsub.ValidationIgnore
 	}
 
 	idxrMsg := dtsync.Message{}
--- a/chain/sub/incoming.go
+++ b/chain/sub/incoming.go
@@ -2,15 +2,17 @@ package sub
 
 import (
 	"context"
-	"fmt"
+	"sync"
 	"time"
 
 	address "github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain"
 	"github.com/filecoin-project/lotus/chain/consensus"
 	"github.com/filecoin-project/lotus/chain/messagepool"
 	"github.com/filecoin-project/lotus/chain/store"
+	"github.com/filecoin-project/lotus/chain/sub/ratelimit"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/metrics"
 	"github.com/filecoin-project/lotus/node/impl/client"
@@ -22,6 +24,7 @@ import (
 	connmgr "github.com/libp2p/go-libp2p-core/connmgr"
 	"github.com/libp2p/go-libp2p-core/peer"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"github.com/multiformats/go-varint"
 	"go.opencensus.io/stats"
 	"go.opencensus.io/tag"
 	"golang.org/x/xerrors"
@@ -168,12 +171,12 @@ func fetchCids(
 	cidIndex := make(map[cid.Cid]int)
 	for i, c := range cids {
 		if c.Prefix() != msgCidPrefix {
-			return fmt.Errorf("invalid msg CID: %s", c)
+			return xerrors.Errorf("invalid msg CID: %s", c)
 		}
 		cidIndex[c] = i
 	}
 	if len(cids) != len(cidIndex) {
-		return fmt.Errorf("duplicate CIDs in fetchCids input")
+		return xerrors.Errorf("duplicate CIDs in fetchCids input")
 	}
 
 	for block := range bserv.GetBlocks(ctx, cids) {
@@ -196,7 +199,7 @@ func fetchCids(
 	if len(cidIndex) > 0 {
 		err := ctx.Err()
 		if err == nil {
-			err = fmt.Errorf("failed to fetch %d messages for unknown reasons", len(cidIndex))
+			err = xerrors.Errorf("failed to fetch %d messages for unknown reasons", len(cidIndex))
 		}
 		return err
 	}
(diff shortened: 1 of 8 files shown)


chain/sub/incoming.go
```diff
@@ -2,18 +2,21 @@ package sub
 
 import (
 	"context"
-	"fmt"
+	"sync"
 	"time"
 
 	address "github.com/filecoin-project/go-address"
+	//"github.com/filecoin-project/lotus/api"
```
vyzo, line 9: ???

chain/sub/incoming.go
```diff
@@ -445,23 +449,202 @@ func recordFailure(ctx context.Context, metric *stats.Int64Measure, failureType
...
+
+	peerCache *lru.TwoQueueCache
+	chainApi  full.ChainModuleAPI
+	stateApi  full.StateModuleAPI
 }
 
-func NewIndexerMessageValidator(self peer.ID) *IndexerMessageValidator {
-	return &IndexerMessageValidator{self: self}
+func NewIndexerMessageValidator(self peer.ID, chainApi full.ChainModuleAPI, stateApi full.StateModuleAPI) *IndexerMessageValidator {
+	peerCache, _ := lru.New2Q(1024)
```
vyzo, line 468: we ahould size this for the expected number of publishers, 8192 seems more reasonable.

chain/sub/incoming.go
```diff
@@ -445,23 +449,202 @@ func recordFailure(ctx context.Context, metric *stats.Int64Measure, failureType
...
+		chainApi:  chainApi,
+		stateApi:  stateApi,
+	}
 }
 
 func (v *IndexerMessageValidator) Validate(ctx context.Context, pid peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
 	// This chain-node should not be publishing its own messages.  These are
-	// relayed from miner-nodes or index publishers.  If a node appears to be
-	// local, reject it.
+	// relayed from market-nodes.  If a node appears to be local, reject it.
```
vyzo, line 480: why? Also see comment below.

chain/sub/incoming.go
```diff
@@ -445,23 +449,202 @@ func recordFailure(ctx context.Context, metric *stats.Int64Measure, failureType
...
 	if pid == v.self {
 		log.Warnf("refusing to relay indexer message from self")
 		stats.Record(ctx, metrics.IndexerMessageValidationFailure.M(1))
 		return pubsub.ValidationReject
 	}
+	originPeer := msg.GetFrom()
+	if originPeer == v.self {
```
vyzo, line 487: this will render your node incapable of publishing, the validator runs for all messages, including yours.

You should check the ReceivedFrom field and if it is locally published, let it through.

chain/sub/incoming.go
```diff
@@ -445,23 +449,202 @@ func recordFailure(ctx context.Context, metric *stats.Int64Measure, failureType
...
+
+	// Decode CID and originator addresses from message.
+	minerID, msgCid, err := decodeIndexerMessage(msg.Data)
+	if err != nil {
+		log.Errorw("Could not decode pubsub message", "err", err)
+		return pubsub.ValidationReject
+	}
+
+	if minerID == "" {
+		log.Warnw("ignoring messsage missing miner id", "peer", originPeer)
```
vyzo, line 501: i would downgrade this log to debug, it will spam the logs once we have more nodes publishing in the indexer network and it is not actionable.

chain/sub/incoming.go
```diff
@@ -445,23 +449,202 @@ func recordFailure(ctx context.Context, metric *stats.Int64Measure, failureType
...
+		log.Warnw("ignoring indexer message", "sender", msgInfo.peerID, "err", err)
+		return true
+	}
+
+	msgInfo.lastCid = msgCid
+
+	return false
+}
+
+func decodeIndexerMessage(data []byte) (string, cid.Cid, error) {
```
vyzo, line 576: why a bespoke data format? Cant you use protobuf or cbor for this?

vyzo (commented):
mostly looks good.

vyzo:
Next step would be to add score params for the topic.

chain/sub/incoming.go
willscott: it's not the lotus daemon that produces these messages, but rather the markets libp2p host.
The markets process in MRA shouldn't have access to the daemon identity key.

The validation structure here is that markets signs it, and lotus daemons relaying the gossip message will look at the lotus miner id that's in the message, look on chain to see the markets libp2p host delegated by that miner, and confirm that the message is signed by that markets libp2p host.

willscott (commented):
> Also what happened to the idea of using the message seqno as a nonce?

This proposes using the message timestamp as the nonce.
We weren't able to see what adding an explicit nonce would provide that timestamp wouldn't - in both cases validators can force monotonically increasing and rate limit based on that field.

vyzo:
> > Also what happened to the idea of using the message seqno as a nonce?
> 
> This proposes using the message timestamp as the nonce.
> We weren't able to see what adding an explicit nonce would provide that timestamp wouldn't - in both cases validators can force monotonically increasing and rate limit based on that field.

Yes, thats exactly what i am asking for, no need for explicit nonce; i dont see it in the diff.

chain/sub/incoming.go
vyzo: thats fine, all i am saying is that you should look at yhe ReceivedFrom and not drop lically published messages.

Isnt the same code used by the miner? How does it enter the overlay?

willscott: it will get sent out on the markets libp2p host which is passed to the index-provider library.

That library will default to only allowing locally generated messages and not relaying any gossip messages.

We're keeping the markets libp2p host as a protected peer to the lotus daemon host, and were worried about cases where a malicious peer was able to connect to the markets process and then send gossip messages that got relayed out from the lotus daemon and affected it's score in the mesh. It seems safer to say the markets libp2p host shouldn't become part of the larger lotus mesh and will only originate it's local gossip messages.

vyzo: so you forward the message to the miner which then publishes to the overlay? Basically some form of delagated publishing.

vyzo: basically i am a little confused as to who publishes (and signs/gets the origin id) here.

vyzo: basically i want to avoid the footgun of accidentally dropping your own messages.

willscott: The markets libp2p host publishes. its only expected connected peer is it's associated lotus daemon which will relay it into the rest of the relay.

The signer/origin will be of the markets libp2p host, not of the lotus daemon itself.

vyzo: does the market process have the miner peer id?

ok, fair enough.

chain/sub/ratelimit/queue.go
```diff
@@ -0,0 +1,89 @@
+package ratelimit
+
+import "errors"
+
+var ErrRate = errors.New("rate exceeded")
```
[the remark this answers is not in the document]
gammazero: Done

chain/sub/incoming.go
gammazero: Done.

chain/sub/incoming.go
gammazero: Done

chain/sub/incoming.go
gammazero: Since this validation only happens for chain nodes (`modules.RelayIndexerMessages` is only a `ChainNode` option) that are relaying a message, and a node that published an indexer message should never relay such a message, this check seems OK.

However, it is really necessary or better to just not worry about checking this?

gammazero:
> Also what happened to the idea of using the message seqno as a nonce?
> 
> Concern is replay attacks that seek to amplify up to whatever the rate limiter allows. I would feel more confident if we were tracking the message nonce, in addition to the rate limiting.

Added replay rejection.

chain/sub/incoming.go
```diff
@@ -445,23 +449,214 @@ func recordFailure(ctx context.Context, metric *stats.Int64Measure, failureType
...
+	}
+
+	// Lock this peer's message info.
+	msgInfo.mutex.Lock()
+	defer msgInfo.mutex.Unlock()
+
+	if ok {
+		// Reject replayed messages.
+		seqno := msg.Message.GetSeqno()
+		if bytes.Equal(msgInfo.lastSeqno, seqno) {
```
vyzo, line 521: compare this for earlier ones; also, ignore, dont reject.

gammazero: done

chain/sub/incoming.go
gammazero: This is now merged, and this PR uses the new encoding as of commit [9481fa0](https://github.com/filecoin-project/lotus/pull/8045/commits/9481fa0a4b5ed18575947fa413d5de936f6437f3)

chain/sub/incoming.go
vyzo: well, lets keep it for now. But i would suggest making it an ignore instead of reject.

vyzo (commented):
I think it looks good now. Lets add some scoring as well and it is ready.

Recommendations for score params: super tiny positive contribution, negative params as in the other topics.

chain/sub/incoming.go
gammazero: Ignored now

vyzo (commented):
make the score contribution from this super tiny, cept for negatives.

node/modules/lp2p/pubsub.go
```diff
@@ -114,6 +114,22 @@ func GossipSub(in GossipIn) (service *pubsub.PubSub, err error) {
...
 
+	ingestTopicParams := &pubsub.TopicScoreParams{
+		// expected ~0.5 confirmed deals / min. sampled
+		TopicWeight: 1,
+
+		TimeInMeshWeight:  0.00027, // ~1/3600
+		TimeInMeshQuantum: time.Second,
+		TimeInMeshCap:     1,
+
+		FirstMessageDeliveriesWeight: 5,
```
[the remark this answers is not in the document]
gammazero: Done.  I really do not have a good sense of how to tune these so I am taking whatever you suggest.

node/modules/lp2p/pubsub.go
```diff
@@ -114,6 +114,22 @@ func GossipSub(in GossipIn) (service *pubsub.PubSub, err error) {
 		InvalidMessageDeliveriesDecay:  pubsub.ScoreParameterDecay(time.Hour),
 	}
 
+	ingestTopicParams := &pubsub.TopicScoreParams{
+		// expected ~0.5 confirmed deals / min. sampled
+		TopicWeight: 1,
```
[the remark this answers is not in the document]
gammazero: Done

vyzo (approved):
I think we are good!

PR #8046, opened 2022-02-08
Upgrade to latest `index-provider` and set miner ID as extra gossip data

masih:
Upgrade to the latest `index-provider` which upgrades the go-legs
protocol to allow the inclusion of extra gossip data that may be used
for gossip validation purposes. In the case of lotus gossip message
validators the miner ID is used to verify the sender's peer ID on chain.

Relates to:
- https://github.com/filecoin-project/lotus/pull/8045

### DIFF
--- a/node/modules/storageminer_idxprov.go
+++ b/node/modules/storageminer_idxprov.go
@@ -19,6 +19,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/peerstore"
 	"go.uber.org/fx"
 
+	"github.com/filecoin-project/go-address"
 	datatransfer "github.com/filecoin-project/go-data-transfer"
 	dtimpl "github.com/filecoin-project/go-data-transfer/impl"
 	dtnet "github.com/filecoin-project/go-data-transfer/network"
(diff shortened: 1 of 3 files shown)


node/modules/storageminer_idxprov.go
```diff
@@ -84,7 +85,15 @@ func IndexProvider(cfg config.IndexProviderConfig) func(params IdxProv, marketHo
...
-		e, err := engine.New(cfg.Ingest, pkey, dt, h, ipds, maddrs)
+		// Get the miner ID and set as extra gossip data.
+		// The extra data is required by the lotus-specific index-provider gossip message validators.
+		address, err := miner.ActorAddress(context.TODO())
+		if err != nil {
+			return nil, xerrors.Errorf("getting miner ID: %w", err)
+		}
+		log.Info("Using extra gossip data in index provider engine: %s", address.String())
+
+		e, err := engine.New(cfg.Ingest, pkey, dt, h, ipds, maddrs, engine.WithExtraGossipData(address.Bytes()))
```
masih, line 96: Cc @gammazero  for visibility

nonsense (approved):

PR #8067, opened 2022-02-10
feat: message: chain nodes relay indexer pubsub messages.

gammazero:
This PR makes lotus chain nodes relay indexer pubsub messages.  Indexer pubsub messages are authenticated against the blockchain and are subject to rate-limiting.

## Related
https://github.com/filecoin-project/lotus/tree/feat/cid-to-piece-idx

## Proposed Changes
- Ignore indexer messages that appear to come from the chain node
- Authenticate message by checking that the sending peer appears in the miner info.
- Rate limit overall messages
- Rate limit repeated announcements
- Ignore replayed messages
- Record messages metrics and topic scoring.


chain/sub/incoming.go
```diff
@@ -444,3 +454,167 @@ func recordFailure(ctx context.Context, metric *stats.Int64Measure, failureType
...
+	if err != nil {
+		log.Errorw("Could not decode indexer pubsub message", "err", err)
+		return pubsub.ValidationReject
+	}
+	if len(idxrMsg.ExtraData) == 0 {
+		log.Debugw("ignoring messsage missing miner id", "peer", originPeer)
+		return pubsub.ValidationIgnore
+	}
+
+	minerID := string(idxrMsg.ExtraData)
```
masih, line 511: @gammazero  the changes from the byte parsing fix needs to be ported here.

See: https://github.com/filecoin-project/lotus/pull/8082

masih (changes requested):

chain/sub/incoming.go
gammazero: Ported, along with the override of "testnetnet" to "mainnet"

masih (approved):

jennijuju:
Hi - are we certain this pr is targeted at v1.13.2? feel like should be a dev branch thing. (master?)

Also - please update the pr title according to instructions described in the pr template if aim to ship in master! thanks

travisperson:
This was an temporary integration required till the feature landed in lotus during the v1.15.x releases

The same change set landed here: https://github.com/filecoin-project/lotus/pull/7313

PR #8078, opened 2022-02-11
Add `lotus net` CLI command to add/remove/list protected peers

masih:
Add CLI commands under `lotus net` to add/ remove/list protected peers.

Relates to: #7313


willscott (approved):
seems useful. i'm in favor of plumbing to the CLI as well.

PR #8082, opened 2022-02-11
Decode gossip extra data as `[]bytes` 

masih:
The type of extra data in go-legs gossip is bytes. But when it is parsed
as miner ID, it is cast to string then parsed. Instead, it should be
decoded from bytes.


willscott:
Thanks for tracking this down, we'll probably want to get it into both this branch, and the back-port for the current prod release

MarcoPolo (commented):
Looks good!

gammazero (approved):

PR #8086, opened 2022-02-14
refactor: convert RepoType from int to interface

nonsense:
## Related Issues
This PR is convering `RepoType` type from `int` to `interface`, so that we can use the `repo` package in other projects and provide a custom Config, that is not referred to in Lotus (such as `config.Boost`).

The new usage can be seen at: https://github.com/filecoin-project/boost/pull/231

---

We want to be using the `filecoin-project/lotus/node/repo` package, due to the fx dependency injection, as we want to integrate all Lotus markets dependencies directly in boost before releasing the alpha version.

## Checklist

- [x] CI is green


dirkmc (approved):

node/repo/memrepo.go
```diff
@@ -106,7 +106,7 @@ func (lmem *lockedMemRepo) Path() string {
 		panic(err) // only used in tests, probably fine
 	}
 
-	if lmem.t == StorageMiner {
+	if lmem.t.Type() == "StorageMiner" || lmem.t.Type() == "Boost" {
```
magik6k, line 109: I'd use trait-interfaces instead of hard-coding (so this would be something like `if lmem.t.(interface{HasStorage()})`, and maybe a separate trait for `deal-staging`)

node/repo/fsrepo.go
```diff
@@ -41,47 +41,83 @@ const (
...
-		"StorageMiner",
-		"Worker",
-		"Wallet",
-		"Markets",
-	}
-	if t < 0 || int(t) > len(s) {
-		return "__invalid__"
-	}
-	return s[t]
+type FullNodeRepoType struct {
```
magik6k, line 64: The `RepoType` suffix feels a bit redundant.

magik6k: Also given that all of those structs are field-less it could be nicer to do

```
type fullNode struct {}
var FullNode fullNode
```

This way we don't need to add `{}` everywhere

node/repo/fsrepo.go
```diff
@@ -41,47 +41,83 @@ const (
...
+
+func (f WalletRepoType) Type() string {
+	return "Wallet"
+}
+
+func (f WalletRepoType) Config() interface{} {
+	return &struct{}{}
 }
 
 func defConfForType(t RepoType) interface{} {
```
magik6k, line 119: Can drop this function now

node/modules/storageminer_dagstore.go
```diff
@@ -23,88 +23,70 @@ const (
...
-		return nil, err
-	}
-
-	// caps the amount of concurrent calls to the storage, so that we don't
-	// spam it during heavy processes like bulk migration.
-	if v, ok := os.LookupEnv("LOTUS_DAGSTORE_MOUNT_CONCURRENCY"); ok {
-		concurrency, err := strconv.Atoi(v)
-		if err == nil {
-			cfg.MaxConcurrencyStorageCalls = concurrency
+func NewMinerAPI(cfg config.DAGStoreConfig) func(fx.Lifecycle, repo.LockedRepo, dtypes.ProviderPieceStore, mdagstore.SectorAccessor) (mdagstore.MinerAPI, error) {
```
magik6k, line 26: Unrelated?

nonsense: This gets rid of the `extractDAGStoreConfig` which casts the `config`, so it is somewhat related and necessary as part this work, but I agree slightly tangential to the rest of the changes.

node/repo/memrepo.go
nonsense: Check last commit, and let me know if this is what you had in mind: https://github.com/filecoin-project/lotus/pull/8086/commits/9fd7e71a07c0772d1a196c4a53fa72140dacf189

nonsense:
Thanks for review @magik6k !!! I think I addressed all your comments.

nonsense:
@magik6k we modified a bit further the config in order to be able to pass in `config.Boost`. I am not happy with how we handle the dynamic configs... any suggestions on how to do that in a better way?

node/config/types.go
```diff
@@ -52,6 +52,32 @@ type StorageMiner struct {
...
+}
+
+func (c *StorageMiner) GetSealingConfig() SealingConfig {
+	return c.Sealing
+}
+
+func (c *StorageMiner) SetSealingConfig(other SealingConfig) {
+	c.Sealing = other
+}
+
```
magik6k, line 80: This file should only contain config structs with docstrings, putting other stuff here will mess with config docsgen

node/repo/fsrepo.go
```diff
@@ -41,47 +41,94 @@ const (
...
+	return "StorageMiner"
+}
+
+func (storageMiner) Config() interface{} {
+	return config.DefaultStorageMiner()
+}
+
+var Markets markets
+
+type markets struct{}
```
magik6k, line 98: Shouldn't this also have `SupportsStagingDeals`?

node/builder.go
```diff
@@ -224,12 +224,16 @@ var LibP2P = Options(
 )
 
 func IsType(t repo.RepoType) func(s *Settings) bool {
-	return func(s *Settings) bool { return s.nodeType == t }
+	return func(s *Settings) bool { return s.nodeType.Type() == t.Type() }
```
magik6k, line 227: Do we actually need the `.Type()` here?

cli/util/api.go
```diff
@@ -32,14 +32,14 @@ const (
 // of the API server (only used by the tests), in the order of precedence they
 // should be applied for the requested kind of node.
 func flagsForAPI(t repo.RepoType) []string {
```
magik6k, line 34: Could we change those repoType->thing functions into methods on RepoType? This way all of those per-type things are in one place.

node/config/types.go
magik6k: (see the comment on top of the file)

node/builder.go
nonsense: Now that we have singleton for `nodeType`, we don't. Removing it.

nonsense:
@magik6k I think I addressed the latest comments.

nonsense:
@magik6k I rebased to master, and I think this is ready for merge. Could you have another look, and let me know if you think we need anything else as part of this PR?

magik6k (approved):

nonsense:
thanks!! :)

PR #8096, opened 2022-02-15
Protect index provider peer ID before connecting to full node

masih:
Add the peer ID of index provider host to the list of protected peers
before connecting to full node. Otherwise, it is possible for the
connection to be reset by full node before we reach the line that adds
the ID to list of protected peers via JsonRPC API.

Relates to:
 - https://github.com/filecoin-project/index-provider/issues/177


willscott:
we should also make sure that just because markets can't dial the daemon libp2p host doesn't mean the full markets process refuses to start up.

nonsense (approved):

PR #8097, opened 2022-02-15
Update to the head of go-fil-markets#673

masih:
Update to the head of the PR that introduces indexing integration in
`go-fil-markets` so that failure to connect to full node is logged only
instead of crashing markets process.

Relates to:
 - https://github.com/filecoin-project/go-fil-markets/pull/673
 - https://github.com/filecoin-project/index-provider/issues/177


willscott (commented):
:+1:

PR #8101, opened 2022-02-16
Fvm: impl VerifyConsensusFault

arajasek:
Needs https://github.com/filecoin-project/ref-fvm/pull/330 and https://github.com/filecoin-project/filecoin-ffi/pull/235

### DIFF
--- a/chain/vm/fvm.go
+++ b/chain/vm/fvm.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 
+	"github.com/filecoin-project/go-state-types/big"
+
 	"github.com/filecoin-project/lotus/chain/state"
 	cbor "github.com/ipfs/go-ipld-cbor"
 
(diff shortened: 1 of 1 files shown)


chain/vm/fvm.go
```diff
@@ -13,20 +19,209 @@ import (
...
 type FvmExtern struct {
 	Rand
 	blockstore.Blockstore
+	epoch   abi.ChainEpoch
+	lbState LookbackStateGetter
+	base    cid.Cid
+}
+
+// Similar to the one in syscalls.go used by the Lotus VM, except it never errors
+// Errors are logged and "no fault" is returned, which is functionally what go-actors does anyway
```
Stebalien, line 40: It would be nice if we could return a better error. E.g., distinguish between invalid arguments and "no fault". But we can do that later.

chain/vm/fvm.go
```diff
@@ -13,20 +19,209 @@ import (
...
+func (x *FvmExtern) VerifyConsensusFault(ctx context.Context, a, b, extra []byte) *ffi_cgo.ConsensusFaultWithGas {
+	ret := &ffi_cgo.ConsensusFaultWithGas{
+		// TODO: is this gonna be a problem on the receiving end? should we return address.NewIDAddress(0) instead?
+		Target: address.Undef,
```
Stebalien, line 44: Probably? I'd consider just returning `nil` in this case, honestly. We can cook up the correct FFI return values in the glue code.

chain/vm/fvm.go
```diff
@@ -13,20 +19,209 @@ import (
...
+		log.Info("invalid consensus fault: cannot decode first block header: %w", decodeErr)
+		return ret
+	}
+
+	if decodeErr := blockB.UnmarshalCBOR(bytes.NewReader(b)); decodeErr != nil {
+		log.Info("invalid consensus fault: cannot decode second block header: %w", decodeErr)
+		return ret
+	}
+
+	// Commented out from the Lotus VM code: FvmExtern only supports v14 and onwards
```
Stebalien, line 67: sg, let's just delete this code.

chain/vm/fvm.go
```diff
@@ -13,20 +19,209 @@ import (
...
+	//
+	//      B
+	//      |
+	//  [A, C]
+	var blockC types.BlockHeader
+	if len(extra) > 0 {
+		if decodeErr := blockC.UnmarshalCBOR(bytes.NewReader(extra)); decodeErr != nil {
+			log.Info("invalid consensus fault: cannot decode extra: %w", decodeErr)
+			// just to match Lotus VM consensus, zero out any already-set faults
+			faultType = ffi_cgo.ConsensusFaultNone
```
Stebalien, line 127: this line does nothing.

Stebalien (approved):

chain/vm/fvm.go
arajasek: Yeah, sounds like a future Steb problem.

chain/vm/fvm.go
arajasek: kk

PR #8105, opened 2022-02-16
Use the same host and datatransfer as markets for index provider

masih:
Remove the bespoke instantiation of libp2p host and datatransfer manager
for index provider and reuse the existing instances used by markets.

The rationale for reuse is the following:
 1. Separation of host introduces a discovery problem, where without
    gossipsub the index provider endpoint will not be discoverable.
    Using the same host as markets would mean the chain can be used to
    discover addresses, putting less empassis on criticality of
    gossipsub considering its set-up cost and lack of message delivery
    guarantees.

 2. Only a single instance of graphsync/datatransfer can be instantiated
    per libp2p host; therefore, if the host is shared, so should be
    datatransfer manager.

 3. it is not clear if the assumptions under which separation was
    decided still hold.

### DIFF
--- a/documentation/en/default-lotus-miner-config.toml
+++ b/documentation/en/default-lotus-miner-config.toml
@@ -273,30 +273,6 @@
   # env var: LOTUS_INDEXPROVIDER_PUBLISHERKIND
   #PublisherKind = "dtsync"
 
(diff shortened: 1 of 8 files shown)


willscott (approved):

dirkmc:
Sorry I'm a little late on the review here, LGTM 👍

PR #8132, opened 2022-02-17
temporarily disable reification

vyzo:
big reifications can use a lot of memory during sync apparently.

Temporarily disabling reification until I can sort out the memory usage during sync.


magik6k (approved):

magik6k:
(backport in case we don't get anything better soon)

PR #8149, opened 2022-02-18
fix: limit reification sizes

vyzo:
It was observed that during sync from snapshot some very large reification walks would trigger, consuming excessive memory.

This introduces a limit to reification mass, and aborts if the visited object count is exceeded.


vyzo:
Note that this also makes the in-memory visitors smarter so as to not waste memory on unitary objects that are not walked anyway.

magik6k (approved):

PR #8158, opened 2022-02-19
one binary, multiple networks

coryschwartz:
## Proposed Changes
Allow the same binary to be used across multiple networks.

Additionally, parameters can be overridden by environment variables.

## Checklist

- [ ] CI is green


coryschwartz:
```LOTUS_NETWORK=butterflynet ./lotus daemon

**Comment by travisperson** [comment]
I think this is a good first pass at this, nothing jump out at me in regard to setting up or managing networks. I think we should discuss this with @arajasek.

**Comment by coryschwartz** [comment]
This allows you to switch networks with the same binary.

There are lots of of reasons it would be good to have a binary that can switch networks, particularly, this makes it easier to operate multiple netowrks with the same packages, docker images, etc.

Additionally, this adds the ability to specify a network definition file aside from the built-in file. Allowing external network definitions will help build new netowrks on the fly with custom parameters.

Currently, this is configured using only environment variables -- I'm not really stuck on this design. I considered re-writing this as a dependency injection, and could be convinced that this would be a better choice.

The current design takes previously constant values and instead exports them as getters only, so they cannot be easily changed at runtime.

**Comment by jennijuju** [comment]
woo this looks cool

File: itests/wdpost_test.go [hunk 1]
```diff
@@ -238,10 +238,6 @@ func TestWindowPostBaseFeeNoBurn(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	och := build.UpgradeClausHeight
-	build.UpgradeClausHeight = 0
-	t.Cleanup(func() { build.UpgradeClausHeight = och })
```
magik6k, line 243: We probably can just drop this entire test

magik6k (commented):
Conflicts/tests need fixing, but this looks really good

travisperson (commented):
This is a pretty big changes and will require communication.

This PR doesn't touch the `Makefile`. However, a lot of the targets are no longer needed and should be removed (all the network builds).

The BuildType + version string should also be addressed.
https://github.com/filecoin-project/lotus/blob/1162b02e609560a6ffcd4d07211100ea18554602/build/version.go#L8-L37

It would probably be good to keep this version reporting, but make it defined by the `LOTUS_NETWORK` / `LOTUS_NETWORK_DEFINITION_FILE`. This will allow people to ask the daemon what network it's talking too, we shouldn't remove this piece of information from users. It's an important bit to gather when users fill out issues.

build/params_debug.go
```diff
@@ -1,11 +0,0 @@
...
-// +build debug
-
-package build
-
-func init() {
-	InsecurePoStValidation = true
-	BuildType |= BuildDebug
-}
-
-// NOTE: Also includes settings from params_2k
```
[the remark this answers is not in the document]
coryschwartz: oo, good point.


I was hoping debug networks could be a runtime rather than build time decision. But there are some things that aren't ready for that, and I don't really want to make this PR any bigger than it already is.

I've added a params_debug back in. It sets the SlashablePowerDelay, Insecure PoSt, and sets the "BuildType" to debug.

coryschwartz:
> This is a pretty big changes and will require communication.
> 
> This PR doesn't touch the `Makefile`. However, a lot of the targets are no longer needed and should be removed (all the network builds).
> 
> The BuildType + version string should also be addressed.
> 
> https://github.com/filecoin-project/lotus/blob/97b92cd529aef61ac7758b5eda3e51640335c5a6/build/version.go#L8-L37
> 
> It would probably be good to keep this version reporting, but make it defined by the `LOTUS_NETWORK` / `LOTUS_NETWORK_DEFINITION_FILE`. This will allow people to ask the daemon what network it's talking too, we shouldn't remove this piece of information from users. It's an important bit to gather when users fill out issues.

Updated the version string so it includes the network name.

It also occurs to me that anyone could create a network that matches a well-known built-in network, so I included a hash of the genesis there as well.

vyzo:
we will need to rethink bundle loading for nv16 with this.

Makefile
```diff
@@ -65,23 +65,11 @@ deps: $(BUILD_DEPS)
...
-2k: build-devnets
-
-calibnet: GOFLAGS+=-tags=calibnet
-calibnet: build-devnets
-
-butterflynet: GOFLAGS+=-tags=butterflynet
-butterflynet: build-devnets
-
-interopnet: GOFLAGS+=-tags=interopnet
-interopnet: build-devnets
```
magik6k, line 83: We may want to keep those for compat for now - just changing the default network in the binary.

magik6k (commented):
I think we can land this if we get the backwards compatibility working

vyzo:
Can we delay until after m1 ships, so that we figure out the bundling strategy?
Right now it would create irreconcilable conflicts with nv16.

ianconsolata:
How is this going to affect https://github.com/filecoin-project/lotus/pull/9625?

Kubuxu:
> How is this going to affect https://github.com/filecoin-project/lotus/pull/9625?

It would no longer be required.

PR #8160, opened 2022-02-21
feat: Make add piece idempotent

aarshkshah1992:
A very common deal making error we encounter is that Boost/Markets makes a call to `AddPiece` and then restarts.
If it restarts before it's able to persist the fact that it's already called AddPiece, it will call `AddPiece` again for the same piece. In this case, the existing API returns an error as it's not idempotent.

This PR allows the user to call `AddPiece` multiple times for the same piece but ensures that it's handed off to the sealing subsystem only once. 

It also makes the method respect the given context.

### DIFF
--- a/extern/storage-sealing/input.go
+++ b/extern/storage-sealing/input.go
@@ -313,15 +313,35 @@ func (m *Sealing) SectorAddPieceToAny(ctx context.Context, size abi.UnpaddedPiec
 	m.inputLk.Lock()
 	if pp, exist := m.pendingPieces[proposalCID(deal)]; exist {
 		m.inputLk.Unlock()
-		select {
-		case <-pp.doneCh:
-			res := pp.resp
+
+		// we already have a pre-existing add piece call for this deal, let's wait for it to finish and see if it's successful
+		for {
+			res, err := waitAddPieceResp(ctx, pp)
+			if err != nil {
+				return api.SectorOffset{}, err
+			}
+			//  there was an error waiting for a pre-existing add piece call, let's retry
+			if res.err != nil {
+				m.inputLk.Lock()
+				pp = m.addPendingPiece(ctx, size, data, deal, sp)
+				m.inputLk.Unlock()
+				continue
+			}
+			// all good, return the response
 			return api.SectorOffset{Sector: res.sn, Offset: res.offset.Padded()}, res.err
-		case <-ctx.Done():
-			return api.SectorOffset{}, ctx.Err()
 		}
 	}
 
+	pp := m.addPendingPiece(ctx, size, data, deal, sp)
+	m.inputLk.Unlock()
+	res, err := waitAddPieceResp(ctx, pp)
+	if err != nil {
+		return api.SectorOffset{}, err
+	}
+	return api.SectorOffset{Sector: res.sn, Offset: res.offset.Padded()}, res.err
+}
+
+func (m *Sealing) addPendingPiece(ctx context.Context, size abi.UnpaddedPieceSize, data storage.Data, deal api.PieceDealInfo, sp abi.RegisteredSealProof) *pendingPiece {
 	doneCh := make(chan struct{})
 	pp := &pendingPiece{
 		doneCh:   doneCh,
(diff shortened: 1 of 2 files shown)


extern/storage-sealing/input.go
```diff
@@ -311,41 +311,45 @@ func (m *Sealing) SectorAddPieceToAny(ctx context.Context, size abi.UnpaddedPiec
...
 	}
 
-	resCh := make(chan struct {
-		sn     abi.SectorNumber
-		offset abi.UnpaddedPieceSize
-		err    error
-	}, 1)
-
-	m.pendingPieces[proposalCID(deal)] = &pendingPiece{
+	doneCh := make(chan struct{}, 1)
```
dirkmc, line 325:
```suggestion
	doneCh := make(chan struct{})
```

extern/storage-sealing/sealing.go
```diff
@@ -140,7 +141,16 @@ func (o *openSector) dealFitsInLifetime(dealEnd abi.ChainEpoch, expF func(sn abi
...
 
+type pieceAcceptResp struct {
+	sn     abi.SectorNumber
+	offset abi.UnpaddedPieceSize
+	err    error
+}
+
 type pendingPiece struct {
+	doneCh chan struct{}
+	resp   atomic.Value
```
dirkmc, line 152: I don't think resp needs to be an atomic value - the doneCh prevents concurrent access

extern/storage-sealing/input.go
```diff
@@ -311,41 +311,45 @@ func (m *Sealing) SectorAddPieceToAny(ctx context.Context, size abi.UnpaddedPiec
...
 
 	m.inputLk.Lock()
-	if _, exist := m.pendingPieces[proposalCID(deal)]; exist {
+	if pp, exist := m.pendingPieces[proposalCID(deal)]; exist {
 		m.inputLk.Unlock()
-		return api.SectorOffset{}, xerrors.Errorf("piece for deal %s already pending", proposalCID(deal))
+		select {
+		case <-pp.doneCh:
+			res := pp.resp.Load().(*pieceAcceptResp)
+			return api.SectorOffset{Sector: res.sn, Offset: res.offset.Padded()}, res.err
```
dirkmc, line 319: Is it safe to call `res.offset.Padded()` if res.err is not nil?

extern/storage-sealing/input.go
aarshkshah1992: Let's keep this for safety, dosen't hurt because the code that writes to this is called sync by the sealer

extern/storage-sealing/sealing.go
aarshkshah1992: good point. fixed.

extern/storage-sealing/input.go
aarshkshah1992: Yup it is, it's an uint.

extern/storage-sealing/input.go
```diff
@@ -311,41 +311,45 @@ func (m *Sealing) SectorAddPieceToAny(ctx context.Context, size abi.UnpaddedPiec
...
+	if pp, exist := m.pendingPieces[proposalCID(deal)]; exist {
 		m.inputLk.Unlock()
-		return api.SectorOffset{}, xerrors.Errorf("piece for deal %s already pending", proposalCID(deal))
+		select {
+		case <-pp.doneCh:
+			res := pp.resp
+			return api.SectorOffset{Sector: res.sn, Offset: res.offset.Padded()}, res.err
+		case <-ctx.Done():
+			return api.SectorOffset{}, ctx.Err()
+		}
```
magik6k, line 322: Could put this into a helper method (e.g. `waitResp`) on `pendingPiece` for readability, and dedupe the same bit of code below.

extern/storage-sealing/input.go
magik6k: We never really write to this channel, just close it, which doesn't block.

extern/storage-sealing/input.go
```diff
@@ -311,41 +311,45 @@ func (m *Sealing) SectorAddPieceToAny(ctx context.Context, size abi.UnpaddedPiec
 	}
 
 	m.inputLk.Lock()
-	if _, exist := m.pendingPieces[proposalCID(deal)]; exist {
+	if pp, exist := m.pendingPieces[proposalCID(deal)]; exist {
```
magik6k, line 314: So there is one somewhat big issue with this - when a market process calls AddPiece, it passes a reader to the call, and that data will be streamed directly into the sector.

If the market process restarts in the middle, or when we already we have a waiting AddPiece call, the stream will be broken, so here we will be waiting for as AddPiece call which is guaranteed to fail - instead, I think we should:
* Wait like we do now;
* If we see an error, try to add the piece to `m.pendingPieces` again  - we know that our reader was not read from, so all the bytes are in correct places, and the stream can be handed off to workers

Tldr I'd make this if into a for loop trying to add the piece to `m.pendingPieces` as long as:
* After waiting we somehow see no error (which is unlikely if we're here, but can happen with weird timing)
* Context is cancelled
* The previous call errored, so we can now add our piece to `m.pendingPieces` again

aarshkshah1992:
@magik6k  Have addressed your review.

jennijuju:
Please update the pr title as instructed in  the pr template before merging

PR #8163, opened 2022-02-21
Fix the epoch used for gas in the message pool & validation

### LINKED ISSUE #6652: PricelistByEpoch Technical Debt

Stebalien:
## Related Issues

Partially fixes #6652.

## Proposed Changes

Bring the gas epoch used in chain validation in-line with the gas epoch used in message execution.

## Additional Info

This change should only affect the one gas repricing we had (Calico). It should only affect consensus if the blocks _at_ the upgrade height included a message without enough gas to cover the message's execution. In the old code:

- The message pool and chain sync logic would have allowed this message.
- The VM itself would have penalized the miner.

Looking at the relevant blocks around 265200 (the Calico upgrade height), I'm not seeing any miner penalties so I have to assume that this didn't happen.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/consensus/filcns/filecoin.go
+++ b/chain/consensus/filcns/filecoin.go
@@ -467,7 +467,7 @@ func (filec *FilecoinEC) checkBlockMessages(ctx context.Context, b *types.FullBl
 	}
 
 	nv := filec.sm.GetNetworkVersion(ctx, b.Header.Height)
-	pl := vm.PricelistByEpoch(baseTs.Height())
+	pl := vm.PricelistByEpoch(b.Header.Height)
 	var sumGasLimit int64
 	checkMsg := func(msg types.ChainMsg) error {
 		m := msg.VMMessage()
--- a/chain/messagepool/check.go
+++ b/chain/messagepool/check.go
@@ -106,7 +106,7 @@ func (mp *MessagePool) checkMessages(ctx context.Context, msgs []*types.Message,
 	curTs := mp.curTs
 	mp.curTsLk.Unlock()
 
-	epoch := curTs.Height()
+	epoch := curTs.Height() + 1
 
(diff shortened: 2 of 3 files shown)


Stebalien:
The right way to test this is to:

1. Create a gas price change that would increase the inclusion price of a message.
2. Create a message with enough gas to cover the inclusion cost before the upgrade, but not after:
    1. The mpool should reject it at the upgrade epoch.
    2. The miner shouldn't include it at the upgrade epoch.
    3. The syncer should not validate the message.

But doing that sounds... painful. Thoughts?

Stebalien:
> Also, just to confirm our understandings align, the reason this was affecting the recent caterpillarnet setups was because the pricelist was changing at the first epoch (even though the Calico upgrade was disabled) because Lotus always starts with the v0 pricelist (for epoch 0) before then switching if appropriate. Is that your take?

Unfortunately, this change isn't actually relevant to that issue. I'm just fixing what I can without breaking anything.

I'll let you merge when/if ready. As you say, writing a test for this isn't high priority.

PR #8168, opened 2022-02-22
feat: market utils: Support unixfsnode in TraverseDag

magik6k:
## Proposed Changes
This adds support for UnixFS ADLs in client parts of lotus.

## Additional Info
Client-side only. Providers already have support for those through Graphsync - https://github.com/ipfs/go-graphsync/blob/8c66ab9e2f562386c59fba2700bb1fcf8c743675/storeutil/storeutil.go#L46

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -100,6 +100,7 @@ require (
 	github.com/ipfs/go-metrics-prometheus v0.0.2
 	github.com/ipfs/go-path v0.0.7
 	github.com/ipfs/go-unixfs v0.3.1
+	github.com/ipfs/go-unixfsnode v1.2.0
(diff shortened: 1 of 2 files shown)


hannahhoward:
@magik6k @willscott https://github.com/filecoin-project/lotus/blob/a05a462c8c9110456c145a69ca202a57fd139940/node/impl/client/client.go#L1012

I'm mildly concerned about this -- we probably explicitly DON'T want unixfs reification (especially HAMT) when writing a CAR.

willscott:
are we going to have a user-configurable selector where we'd have the potential for ADL signalling in that case?

PR #8174, opened 2022-02-23
Use either Lotus VM or FVM consistently

arajasek:
First commit is just a rename, second commit does the thing.


jennijuju:
out of curiosity - why do we need the VMI? how can we use it form network version protected VM selection later?

arajasek:
@jennijuju Not fully figured out yet, but I'm thinking it'll change to take a network version, and create either a lotus vm or FVM based on that (instead of based on an envvar today)

conformance/driver.go
```diff
@@ -160,7 +160,7 @@ func (d *Driver) ExecuteTipset(bs blockstore.Blockstore, ds ds.Batching, params
 			return big.Zero(), nil
 		}
 
-		return vm.NewVM(ctx, vmopt)
```
raulk, line 163: Are you sure this should always use the Lotus VM? I think we need a "nv-based VM selector" that constructs one VM or another based on epoch. But that's getting into the territory of FIP-0031. Just wanted to point out that _this_ here is not something we want in this manner because it prevents us from running test vectors against the FVM entirely.

arajasek: You're right, I just skipped the testing-related changes for now (motivation for this PR is to get caterpillarnet going).

chain/stmgr/call.go
```diff
@@ -219,7 +224,12 @@ func (sm *StateManager) CallWithGas(ctx context.Context, msg *types.Message, pri
 		}
 	}
 
-	fromActor, err := vmi.StateTree().GetActor(msg.From)
+	stTree, err := sm.StateTree(state)
+	if err != nil {
+		return nil, err
+	}
+
+	fromActor, err := stTree.GetActor(msg.From)
```
Stebalien, line 232: So, this is going to fail if the actor was changed by one of the messages in `priorMsgs`. We have a few options:

1. Flush to get the current root CID of the state. Unfortunately, that'll flush things back to badger which we don't want to do.
2. Flush, but wrap the blockstore in a buffered blockstore.
3. Expose a way to call get-actor on the FVM through the FFI.

Stebalien (changes requested):

chain/stmgr/call.go
arajasek: @Stebalien Nice catch! I'm tempted to go with 2, unless you think the FVM being able to support `GetActor` is independently valuable.

Stebalien: I do think being able to access the FVM state while running is useful, but I'd punt on that for now and do the "easy" thing.

chain/stmgr/call.go
```diff
@@ -201,6 +207,11 @@ func (sm *StateManager) CallWithGas(ctx context.Context, msg *types.Message, pri
 		)
 	}
 
+	filVested, err := sm.GetFilVested(ctx, ts.Height()+1)
```
Stebalien, line 210: nit: store the epoch in a variable so we don't accidentally use `ts.Height()` in the somewhere in the future.

chain/stmgr/call.go
Stebalien: But state access will likely involve:

- A `GetActor` call (address -> actor)
- A `ResolveActor` call (address -> id)
- A "blockstore" export. The reverse of the CGO blockstore.

Stebalien (approved):
LGTM

PR #8206, opened 2022-02-28
release worker tracker lock when call cb func

firesWu:
release worker tracker lock when call cb func, so that miner can call the worker api concurrently.


magik6k (approved):
Thanks for the PR!

PR #8208, opened 2022-03-01
fix ReplicaUpdate log

llifezou:
Fix `func (m *Manager) ReplicaUpdate` log level and add sector id to log.


ZenGround0:
thanks missed this and duplicated here: https://github.com/filecoin-project/lotus/pull/8213

Issue #8117, opened 2022-02-16
LOTUS_DISABLE for ntwk v16


jennijuju:
add the flag if we have to rollabck the upgrade immediately after the upgrade since something is really wrong. this should only be used if we wanna rollback within  24 hours after the upgrade epoch, otherwise we should go for #8113  or #8114

jennijuju:
(this is basically a time machine!

jennijuju:
should resolved by #8209

Issue #8061, opened 2022-02-10
Epic: Implement FIP-0031


arajasek:
- [x] #8062 
- [x] #8063
- [x] #8064
- [x] (actors) bake the Miner actor CID into the Power actor CID, and the Paych, msig and miner actor CIDs into the init actor
- [x] #8065 -- we can no longer use it for things like actor CIDs (or method nums)?
- [x] #8066
- [x] #8164

arajasek:
[See FIPs #294](https://github.com/filecoin-project/FIPs/pull/294)

jennijuju:
Done! GO TEAM

Issue #8064, opened 2022-02-10
Add nv16 upgrade logic, switching to the FVM at the appropriate epoch


arajasek:
Note that this means using the FVM _with_ its own gas accounting, as opposed to simply using the FVM to compute the same state as we'd get with Lotus VM

raulk:
Implemented and landed in nv16 branch. Also tested on caterpillarnet.

Issue #8066, opened 2022-02-10
Add a migration that transforms dummy code CIDs to actual CIDs based on the actor code


arajasek:
There's something of an open question about _where_this migration should go. I think it should just be in specs-actors -- the last thing that repo can do is transform the state tree to create the new version for the FVM-based future to use :)

jennijuju:
migration of the v7 -> v8 CID https://github.com/filecoin-project/specs-actors/pull/1576

as a part of the https://github.com/filecoin-project/builtin-actors/issues/4 we will get the new CID into the migration

raulk:
Implemented and landed in nv16 branch.

PR #8209, opened 2022-03-01
Implement FIP-0031

arajasek:
Merge strategy: Land #8293 in master, then land this in the appropriate branch (see @jennijuju note below).

This PR implements [FIP-0031](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0031.md). The key changes are:

- [x] Introducing v8 specs-actors and the necessary migration
- [ ] Integrating v8 builtin actors into the FFI (doesn't need the actors to be code-complete) 
- [ ] Supporting v16 in the FVM (this should just be a matter of adding v16 to the supported versions in the FVM)
- [ ] Using content-addressed CIDs in specs-actors (see https://github.com/filecoin-project/specs-actors/issues/1581, https://github.com/filecoin-project/lotus/issues/8066)

Fixes https://github.com/filecoin-project/lotus/issues/8117.
Fixes https://github.com/filecoin-project/lotus/issues/8061.


jennijuju:
resolves #8061

arajasek:
The test failures are because they try to run with the latest network version (set to v16 in this PR), which is currently unsupported by the FVM and the actors. We could:

- temporarily bump them to using nv15 (this would require undoing some codegen, which is undesirable)
- make getting v8 actors and FVM v16 a blocker for this PR: this probably makes more sense in terms of ordering, but if we decide to do that, I would want to merge `feat/fvm` (the destination branch of this PR) into `master` now (and this can land later).

arajasek:
Closes #8064

jennijuju:
after feat/fvm is merged into master (tmr) then that branch would be closed.
id suggest we then create a nv16 branch, base of release/v1.15.1 and land this PR there

chain/vm/mkactor.go
```diff
@@ -133,6 +134,8 @@ func newAccountActor(ver actors.Version) *types.Actor {
 		code = builtin6.AccountActorCodeID
 	case actors.Version7:
 		code = builtin7.AccountActorCodeID
+	case actors.Version8:
+		code = builtin8.AccountActorCodeID
```
magik6k, line 138: We'll need to set those code CIDs to real code CIDs

magik6k (commented):
Nothing unexpected here, obviously still need real code CIDs

raulk:
Superseded by #8429.

PR #8210, opened 2022-03-01
Open Redirect in node-forge

bhaskarvilles:
parseUrl functionality in node-forge mishandles certain uses of backslash such as https:///\ and interprets the URI as a relative path.

The latest possible version that can be installed is 0.7.5 because of the following conflicting dependency:

react-scripts@3.0.1 requires node-forge@0.7.5 via a transitive dependency on selfsigned@1.10.4

The earliest fixed version is 1.0.0.

### DIFF
--- a/lotuspond/front/package-lock.json
+++ b/lotuspond/front/package-lock.json
@@ -8671,8 +8671,8 @@
       }
     },
     "node-forge": {
-      "version": "0.7.5",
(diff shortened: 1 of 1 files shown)


magik6k:
Thanks for the PR. We're not accepting updates to lotus-pond unless there is malware in current dependencies, see https://github.com/filecoin-project/lotus/pull/8313#issuecomment-1066713752

PR #8265, opened 2022-03-09
feat: shed: blockstore/vlog to car export cmds

magik6k:
Just some tools to get data out of big badger datastores, and into car files.


cmd/lotus-shed/datastore-vlog.go
```diff
@@ -0,0 +1,342 @@
...
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+)
+
+var datastoreVlog2CarCmd = &cli.Command{
+	Name:  "vlog2car",
+	Usage: "convert badger blockstore .vlog to .car",
+	Flags: []cli.Flag{
+		&cli.PathFlag{
+			Name:     "vlog",
```
ZenGround0, line 27: For my own curiosity: what is a badger vlog?  I see a bunch of `.vlog` files under datastore repos.  What is the meaning of each vlog subunit?  Is it a partition of data based on the key for example?

cmd/lotus-shed/datastore-vlog.go
```diff
@@ -0,0 +1,342 @@
...
+	block "github.com/ipfs/go-block-format"
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-base32"
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+)
+
+var datastoreVlog2CarCmd = &cli.Command{
+	Name:  "vlog2car",
+	Usage: "convert badger blockstore .vlog to .car",
```
ZenGround0, line 24: Any warnings needed about the daemon needing to be turned off as in the following utility?

ZenGround0: I'm not seeing any interaction with the repo lock file, any chance for problems if the node is online?

ZenGround0 (approved):
I don't know badger internals so this review is fairly shallow with only some high level fearmongering about repo lock safety.  Even without knowing badger the general idea of car to badger seems *really* nice.

cmd/lotus-shed/datastore-vlog.go
magik6k: vlog files are append-only, so that's (probably) fine

cmd/lotus-shed/datastore-vlog.go
magik6k: It's basically an append-only (afaiu) k-v write log. The `Entry`/`header` structs below tell you more exactly what is in those files

PR #8282, opened 2022-03-10
misc: builder: env var for managing rcmgr

magik6k:
Also disable it by default for now, until we can figure out what's broken (also disabling by default means that people will not set the envvar to disable rcmgr and keep it after we've fixed issues)

### DIFF
--- a/node/builder.go
+++ b/node/builder.go
@@ -3,6 +3,7 @@ package node
 import (
 	"context"
 	"errors"
+	"os"
 	"time"
 
 	metricsi "github.com/ipfs/go-metrics-interface"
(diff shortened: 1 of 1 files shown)


jennijuju (approved):

vyzo (commented):
that wont disable it but rather use the default.

You need to pass network.NullResourceManager to disable it.

jennijuju:
@vyzo lmao request change!!! thank you for reviewing 🟦

ZenGround0 (approved):

vyzo:
#8318 does it properly.

PR #8283, opened 2022-03-10
feat: config: Persistent subsystem log level config

### LINKED ISSUE #5138: Log settings not persistent
**Describe the bug**
When adjusting which events appear in the log (using lotus-miner log set-level --system xxx), the settings only persist until the miner is restarted. If they could be set in config, or persisted in some other way, it would be ideal.

**To Reproduce**
Change log setting. Restart miner. Log settings revert to default.

**Expected behavior**
Settings stick, or another method to set desired log levels.

Daemon:  1.2.2+git.b13226bc2.dirty+api1.0.0
Local: lotus version 1.2.2+git.b13226bc2.dirty

magik6k:
## Related Issues
Closes #5138

### DIFF
--- a/documentation/en/default-lotus-config.toml
+++ b/documentation/en/default-lotus-config.toml
@@ -23,6 +23,12 @@
   #DisableMetadataLog = false
 
 
+[Logging]
+  [Logging.SubsystemLevels]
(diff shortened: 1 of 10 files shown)


ZenGround0 (approved):
LGTM

PR #8293, opened 2022-03-11
Filecoin Virtual Machine integration

raulk:
Related (but not blocked by) https://github.com/filecoin-project/filecoin-ffi/pull/217

This PR integrates the [reference FVM](https://github.com/filecoin-project/ref-fvm) into Lotus. In doing so, it also integrates the new [Rust-based builtin actors ](https://github.com/filecoin-project/builtin-actors) through the FFI.

The salient changes are:

- Abstracting out the Virtual Machine in Lotus with a new Virtual Machine Interface that is implemented by the Lotus VM (suggested by @raulk to be renamed to Legacy VM) and the new FVM struct in Lotus.
- Integrating the FVM and builtin actors into the FFI (see https://github.com/filecoin-project/filecoin-ffi/pull/217). The FFI exposes new endpoints to:
  - Construct FVMs
  - Execute messages (both implicit and explicit)
  - Flush
- Refactor the circulating supply logic to supply a partially-calculated value to the FVM **for v14 (and before) only**. This logic can be dropped if FVM support for v14 is not desired.
- Put usage of the new FVM behind the `LOTUS_USE_FVM_EXPERIMENTAL` flag. If enabled, Lotus will use FVMs (and through it the Rust-based actors) instead of the Lotus VM and Golang-based actors.


go.sum
```diff
@@ -390,8 +389,8 @@ github.com/filecoin-project/specs-actors/v5 v5.0.4/go.mod h1:5BAKRAMsOOlD8+qCw4U
 github.com/filecoin-project/specs-actors/v6 v6.0.0/go.mod h1:V1AYfi5GkHXipx1mnVivoICZh3wtwPxDVuds+fbfQtk=
 github.com/filecoin-project/specs-actors/v6 v6.0.1 h1:laxvHNsvrq83Y9n+W7znVCePi3oLyRf0Rkl4jFO8Wew=
 github.com/filecoin-project/specs-actors/v6 v6.0.1/go.mod h1:V1AYfi5GkHXipx1mnVivoICZh3wtwPxDVuds+fbfQtk=
-github.com/filecoin-project/specs-actors/v7 v7.0.0-20211117170924-fd07a4c7dff9/go.mod h1:p6LIOFezA1rgRLMewbvdi3Pp6SAu+q9FtJ9CAleSjrE=
 github.com/filecoin-project/specs-actors/v7 v7.0.0-20211222192039-c83bea50c402/go.mod h1:p6LIOFezA1rgRLMewbvdi3Pp6SAu+q9FtJ9CAleSjrE=
+github.com/filecoin-project/specs-actors/v7 v7.0.0-rc1.0.20220118005651-2470cb39827e/go.mod h1:TA5FwCna+Yi36POaT7SLKXsgEDvJwc0V/L6ZsO19B9M=
```
jennijuju, line 393: can we haz a [tag](https://github.com/filecoin-project/specs-actors/tags)? whats new in go actors anyways? PRU fix?

go.sum
```diff
@@ -1943,6 +1942,7 @@ github.com/whyrusleeping/cbor-gen v0.0.0-20200826160007-0b9f6c5fb163/go.mod h1:f
 github.com/whyrusleeping/cbor-gen v0.0.0-20210118024343-169e9d70c0c2/go.mod h1:fgkXqYy7bV2cFeIEOkVTZS/WjXARfBqSH6Q2qHL33hQ=
 github.com/whyrusleeping/cbor-gen v0.0.0-20210219115102-f37d292932f2/go.mod h1:fgkXqYy7bV2cFeIEOkVTZS/WjXARfBqSH6Q2qHL33hQ=
 github.com/whyrusleeping/cbor-gen v0.0.0-20210303213153-67a261a1d291/go.mod h1:fgkXqYy7bV2cFeIEOkVTZS/WjXARfBqSH6Q2qHL33hQ=
+github.com/whyrusleeping/cbor-gen v0.0.0-20210713220151-be142a5ae1a8/go.mod h1:fgkXqYy7bV2cFeIEOkVTZS/WjXARfBqSH6Q2qHL33hQ=
```
jennijuju, line 1945: tag plz

jennijuju:
@raulk  IIUC yushi implemented this but you opened pr - if so, you should review this

chain/stmgr/supply.go
```diff
@@ -315,29 +339,12 @@ func (sm *StateManager) GetVMCirculatingSupply(ctx context.Context, height abi.C
...
-	}
-	if sm.postCalicoVesting == nil {
-		err := sm.setupPostCalicoVesting(ctx)
-		if err != nil {
-			return api.CirculatingSupply{}, xerrors.Errorf("failed to setup post-calico vesting schedule: %w", err)
-		}
-	}
+func (sm *StateManager) loadGenesisMsigs(ctx context.Context) error {
+	return nil
+}
```
raulk, line 344: What's the point of this method?

chain/vm/fvm.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	return &ApplyRet{
+		MessageReceipt: types.MessageReceipt{
+			Return:   ret.Return,
+			ExitCode: exitcode.ExitCode(ret.ExitCode),
+			GasUsed:  ret.GasUsed,
+		},
+		GasCosts: nil,
+		// TODO: do these eventually, not consensus critical
```
raulk, line 301:
```suggestion
		// TODO: do these eventually, not consensus critical
		// https://github.com/filecoin-project/ref-fvm/issues/318
```

chain/vm/fvm.go
```diff
@@ -0,0 +1,310 @@
...
+
+	return raddr, gasUsed, nil
+}
+
+type FVM struct {
+	fvm *ffi.FVM
+}
+
+func NewFVM(ctx context.Context, opts *VMOpts) (*FVM, error) {
+	circToReport := opts.FilVested
```
raulk, line 220: Are there no solutions other than snaking through this `FilVested` field? Seems a little bit ad-hoc when combined with the pre-existing `CircSupplyCalc` field. And the fact that it won't be even used nv15 and forward, makes me even uneasier.

An alternative solution would be to expand the signature of `CircSupplyCalc` to take in the network version. Since the value of that is a `StateManager` member, it has all it needs to return `FilVested` if nv < 15.

Then this logic can be entirely independent of the network version, and we'd just pass the value down and let the FVM do the right thing.

chain/vm/vmi.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/ipfs/go-cid"
+)
+
+type VMI interface {
+	ApplyMessage(ctx context.Context, cmsg types.ChainMsg) (*ApplyRet, error)
+	ApplyImplicitMessage(ctx context.Context, msg *types.Message) (*ApplyRet, error)
+	Flush(ctx context.Context) (cid.Cid, error)
+}
```
raulk, line 15: Docs pls.

chain/vm/vmi.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"context"
+	"os"
+
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/ipfs/go-cid"
+)
+
+type VMI interface {
```
raulk, line 11: Ideally change to `Interface` to follow go conventions.

chain/vm/vmi.go
```diff
@@ -0,0 +1,23 @@
...
+	ApplyImplicitMessage(ctx context.Context, msg *types.Message) (*ApplyRet, error)
+	Flush(ctx context.Context) (cid.Cid, error)
+}
+
+func NewVM(ctx context.Context, opts *VMOpts) (VMI, error) {
+	if os.Getenv("LOTUS_USE_FVM_EXPERIMENTAL") == "1" {
+		return NewFVM(ctx, opts)
+	}
+
+	return NewLotusVM(ctx, opts)
```
raulk, line 22: Can we call this `LegacyVM`? It's a bit more specific than `LotusVM`.

go.sum
arajasek: Idk why go sum wants this tbh...but go sum thinks for itself (the version in go.mod is correct)

go.sum
arajasek: no change from master here

chain/consensus/filcns/compute_state.go
```diff
@@ -108,10 +109,20 @@ func (t *TipSetExecutor) ApplyBlocks(ctx context.Context, sm *stmgr.StateManager
...
+
+			vmopt.FilVested = filVested
+			return vm.NewFVM(ctx, vmopt)
+		}
+
 		return sm.VMConstructor()(ctx, vmopt)
 	}
 
-	runCron := func(vmCron *vm.VM, epoch abi.ChainEpoch) error {
+	runCron := func(vmCron vm.VMI, epoch abi.ChainEpoch) error {
```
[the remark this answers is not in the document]
arajasek: Not sure I love that, but I'll follow the convention.

chain/vm/fvm.go
arajasek: @raulk My priority here is not changing the existing logic for the Lotus VM. I would rather have a new field that the LotusVM entirely ignores -- that way if there's anything wrong with this PR, only FVM M0 testing is affected.

If it'll make you feel better, we can drop this entirely from the final released version of Lotus for the v16 upgrade and just always use the calculated CircSupply.

chain/consensus/filcns/compute_state.go
```diff
@@ -108,10 +109,20 @@ func (t *TipSetExecutor) ApplyBlocks(ctx context.Context, sm *stmgr.StateManager
...
+		if os.Getenv("LOTUS_USE_FVM_EXPERIMENTAL") == "1" {
+			filVested, err := sm.GetFilVested(ctx, e)
+			if err != nil {
+				return nil, err
+			}
+
+			vmopt.FilVested = filVested
+			return vm.NewFVM(ctx, vmopt)
+		}
+
```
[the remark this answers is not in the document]
arajasek: Yessir.

chain/stmgr/call.go
```diff
@@ -112,7 +125,12 @@ func (sm *StateManager) Call(ctx context.Context, msg *types.Message, ts *types.
 		)
 	}
 
-	fromActor, err := vmi.StateTree().GetActor(msg.From)
+	stTree, err := sm.StateTree(bstate)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
arajasek: @raulk Well, there's a new potential here, if that's what you mean, I can add a descreptive error message. The cause of the change is just that we don't have (or want) the StateTree method on the VM interface.

chain/stmgr/call.go
```diff
@@ -196,16 +216,23 @@ func (sm *StateManager) CallWithGas(ctx context.Context, msg *types.Message, pri
...
+
+	buffStore := blockstore.NewBuffered(sm.cs.StateBlockstore())
 	vmopt := &vm.VMOpts{
-		StateBase:      state,
-		Epoch:          ts.Height() + 1,
+		StateBase:      stateCid,
+		Epoch:          vmHeight,
 		Rand:           r,
-		Bstore:         sm.cs.StateBlockstore(),
+		Bstore:         buffStore,
```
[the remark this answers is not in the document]
arajasek: Nah, the Lotus VM wraps the store you give it in a buffered store, so this was going into a buffer there (that then got thrown away cuz we never flushed, while still being accessible for determining the state of the sender in between messages).

chain/stmgr/call.go
```diff
@@ -219,7 +246,17 @@ func (sm *StateManager) CallWithGas(ctx context.Context, msg *types.Message, pri
...
-	fromActor, err := vmi.StateTree().GetActor(msg.From)
+	stateCid, err = vmi.Flush(ctx)
+	if err != nil {
+		return nil, xerrors.Errorf("flushing vm: %w", err)
+	}
+
+	stTree, err := state.LoadStateTree(cbor.NewCborStore(buffStore), stateCid)
+	if err != nil {
+		return nil, xerrors.Errorf("loading state tree: %w", err)
+	}
```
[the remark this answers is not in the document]
arajasek: 1. Will do.
2. Eh, I don't think I care enough -- the minor perf hit to simulating messages seems fine. I'd rather not have more `IF FVM` clauses than strictly necessary.

chain/stmgr/supply.go
arajasek: This came from @Stebalien, but looks like unused code accidentally introduced. Confirm pls steb?

chain/vm/vmi.go
arajasek: Sure!

chain/stmgr/supply.go
Stebalien: Yeah, I'm not sure where that came from. I think I was trying to refactor `GetFilVested`? Delete it.

arajasek: Deleting.

chain/vm/fvm.go
```diff
@@ -0,0 +1,312 @@
...
+	Rand
+	blockstore.Blockstore
+	epoch   abi.ChainEpoch
+	lbState LookbackStateGetter
+	base    cid.Cid
+}
+
+// VerifyConsensusFault is similar to the one in syscalls.go used by the LegacyVM, except it never errors
+// Errors are logged and "no fault" is returned, which is functionally what go-actors does anyway
+func (x *FvmExtern) VerifyConsensusFault(ctx context.Context, a, b, extra []byte) (*ffi_cgo.ConsensusFault, int64) {
```
magik6k, line 47: What happens if any of those extern calls panic? Do we crash the node, or do we behave like the current VM and just fail execution?

magik6k (approved):
Looks good, just one question.

chain/consensus/filcns/compute_state.go
```diff
@@ -108,10 +109,23 @@ func (t *TipSetExecutor) ApplyBlocks(ctx context.Context, sm *stmgr.StateManager
 			LookbackState:  stmgr.LookbackStateGetterForTipset(sm, ts),
 		}
 
+		if os.Getenv("LOTUS_USE_FVM_EXPERIMENTAL") == "1" {
+			// This is needed so that the FVM does not have to duplicate the genesis vesting schedule, one
+			// of the components of the circ supply calc.
+			// This field is NOT needed by the LegacyVM, and also NOT needed by the FVM from v15 onwards.
```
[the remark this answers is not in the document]
arajasek: First Protocol Improvement / Bugfix mentioned [here](https://github.com/filecoin-project/tpm/blob/master/Network%20Upgrades/v15.md)

(also adding a link in the comment)

chain/gen/genesis/miners.go
```diff
@@ -95,12 +95,13 @@ func SetupStorageMiners(ctx context.Context, cs *store.ChainStore, sys vm.Syscal
 		Syscalls:       mkFakedSigSyscalls(sys),
 		CircSupplyCalc: csc,
 		NetworkVersion: nv,
-		BaseFee:        types.NewInt(0),
+		BaseFee:        big.Zero(),
+		FilVested:      big.Zero(),
```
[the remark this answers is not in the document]
arajasek: Correct, yeah. Also Vested fil _is_ equal to zero at genesis, so should be consistent.

chain/vm/fvm.go
arajasek: @magik6k Thanks a lot for raising this. This is a somewhat open question about exactly what we wanna do -- currently:

- Rust panics get converted to errors at the FFI layer
- Extern panics such as this one are...unclear? Maybe also get lowered to errors at the FFI laye, maybe crash the node? @Stebalien isn't sure, we can relatively easy answer this by experimenting.
- Any error returned by the FFI (including such downgraded panics) cause the `ApplyMessage` call to fail and the node to freeze. In terms of the `LegacyVM`, this means all such errors are "fatal errors". I'm fairly opposed to this, but a decision will be made later.

raulk (commented):
LGTM, barring @magik6k's very valid concern about panic handling/translation at the FFI layer. I can't approve because GitHub thinks I'm the author of this PR.

chain/vm/fvm.go
raulk: > Extern panics such as this one are...unclear? Maybe also get lowered to errors at the FFI laye, maybe crash the node? @Stebalien isn't sure, we can relatively easy answer this by experimenting.

The Rust => Go call path is brand new, so this definitely deseves a ton of attention.

go.sum
mvdan: yeah just ignore go.sum, it only records all checksums seen as part of resolving deps. you can often end up with several older versions listed in go.sum compared to the one actually being used.

Issue #8307, opened 2022-03-14
infinite loop "state machine error" takes up all the disk space 


Aloxaf:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon:  1.14.4-2+mainnet+git.493b62d45.dirty+api1.4.0
Local: lotus-miner version 1.14.4-2+mainnet+git.493b62d45.dirty
```

### Describe the Bug

This error keep looping, and takes up all my disk space soon.

### Logging Information

```
7920.   2022-03-14 08:51:44 +0800 CST:  [event;sealing.SectorChainPreCommitFailed]      {"User":{}}
        precommit info not found on chain
[...]

PR #8310, opened 2022-03-14
fix handlePreCommitFailed: PreCommitMessage is not nil, but SectorPre…

llifezou:
## Related Issues
https://github.com/filecoin-project/lotus/issues/8307

### DIFF
--- a/extern/storage-sealing/states_failed.go
+++ b/extern/storage-sealing/states_failed.go
@@ -87,6 +87,18 @@ func (m *Sealing) handlePreCommitFailed(ctx statemachine.Context, sector SectorI
 				return err
 			}
 
+			pci, pciErr := m.Api.StateSectorPreCommitInfo(ctx.Context(), m.maddr, sector.SectorNumber, tok)
+			if pciErr != nil {
+				log.Errorf("handlePreCommitFailed: StateSectorPreCommitInfo: api error, not proceeding: %+v", pciErr)
+				return nil
+			}
+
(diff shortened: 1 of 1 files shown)


extern/storage-sealing/states_failed.go
```diff
@@ -87,6 +87,18 @@ func (m *Sealing) handlePreCommitFailed(ctx statemachine.Context, sector SectorI
...
 			}
 
+			pci, pciErr := m.Api.StateSectorPreCommitInfo(ctx.Context(), m.maddr, sector.SectorNumber, tok)
+			if pciErr != nil {
+				log.Errorf("handlePreCommitFailed: StateSectorPreCommitInfo: api error, not proceeding: %+v", pciErr)
+				return nil
+			}
+
+			// PreCommitMessage is not nil, but SectorPreCommitInfo is nil, should try to resubmit
+			if pci == nil {
```
magik6k, line 97: This can also be nil after the message was submitted but wasn't yet executed on-chain (which can take 1-2min)

magik6k (commented):
Thanks

extern/storage-sealing/states_failed.go
llifezou: PreCommitMessage is not nil, it should be possible to determine whether the status is jumped from PreCommitWait or PreCommitBatchWait to PreCommitFailed

llifezou: In `func(m *Sealing) handlePreCommitWait`, I understand that StateWaitMsg should block until query result or return err, is it so?

PR #8313, opened 2022-03-14
Bump Updates to some dependencies

bhaskarvilles:
updates to dependencies


magik6k:
Thanks for the PR!

`lotuspond` is just a piece of developer tooling which we use rather sparsely. It is used localhost-only, and for short sessions. 

Given NPMs not so great history with malware in dependencies, and the fact that this tooling is not exposed to the internet in any way, it's actually safer for us to not update those dependencies. The fact that it's used very sparsely also means that it makes no sense for us to try to review those dependency updates for anything malicious.

PR #8348, opened 2022-03-19
Security Upgrades to Packages

bhaskarvilles:
Bump Dependency Security updates to some packages


magik6k:
As already mentioned in https://github.com/filecoin-project/lotus/pull/8313#issuecomment-1066713752, especially given recent.. news around a certain node package, we're definitely not accepting random updates to npm packages which do not need upgrading.

PR #8429, opened 2022-04-04
FIP0031 refinements

vyzo:
This embeds the bundle for v8; long run we want to autofetch from IPFS, but we discussed this with @raulk and decided it is fine for now, avoiding scope creep.

As it is it will load and embed the bundle from `build/builtin-actors/builtin-actors-v8.car`.

TBD:
- [x] Need to make it download from github CI assets if it is not already there -- blocked on fixing https://github.com/filecoin-project/builtin-actors/issues/187)
  Note: made it upload to estuary and fetch from ipfs instead.
- [X] Update fvm construction to pass the manifest CID -- blocked on https://github.com/filecoin-project/filecoin-ffi/pull/258
- [x] https://github.com/filecoin-project/specs-actors/pull/1587; points to pr branch for now.
- [ ] https://github.com/filecoin-project/ref-fvm/pull/470 and release; uses https://github.com/filecoin-project/filecoin-ffi/pull/262 for now.


node/builder_chain.go
```diff
@@ -48,6 +48,13 @@ var ChainNode = Options(
 	// Full node or lite node
 	// TODO: Fix offline mode
 
+	// FVM: builtin actor bundle loading
+	//      Note: this has to load before the upgrade schedule, so that we can patch in the
+	//      right manifest cid.
+	//      This restriction will be lifted once we have the final actors v8 bundle and we know
+	//      the manifest cid.
+	Override(BuiltinActorsKey, modules.LoadBultinActors),
```
magik6k, line 56: Does this get called? If it's not using an InvokeKey, it's just a Provide, and given that this isn't really a constructor, I wouldn't expect it to get called.

On invoke keys, note that they are called in order the invoke keys are defined, and are essentially the 'roots' from which dependencies get resolved.

Probably the best (a.k.a least bad) way to do that is to make `LoadBultinActors` return a dummy type like `type BuiltinActorsLoaded struct{}` and depend on it in UpgradeSchedule constructor (we do a similar thing when loading genesis into the chainstore)

chain/actors/manifest.go
```diff
@@ -0,0 +1,72 @@
...
+
+var manifests map[Version]*manifest.Manifest
+var actorMeta map[cid.Cid]actorEntry
+
+type actorEntry struct {
+	name    string
+	version Version
+}
+
+func LoadManifests(ctx context.Context, store cbor.IpldStore) error {
```
magik6k, line 27: I don't love globals, but I don't see any good way to get rid of this.

Note that LoadManifests can be called from multiple node constructors in parallel in integration tests, so prooobably want to wrap this in global `sync.Once`..

magik6k (commented):
First limited pass.

You probably want to run full `make gen` to fix codegen issues.

node/builder_chain.go
vyzo: ah thanks, will do the dummy type thing.

chain/actors/manifest.go
vyzo: ok, will wrap it in a Once.

node/builder_chain.go
vyzo: done.

chain/actors/manifest.go
vyzo: done.

arajasek (commented):
Can we add instructions to https://github.com/filecoin-project/lotus/blob/d9100981ada8b3186d906a4f4140b83a819d2299/documentation/misc/actors_version_checklist.md for when we need the next version of actors (v9).

vyzo:
sure!

itests/kit/ensemble.go
```diff
@@ -374,17 +374,20 @@ func (n *Ensemble) Start() *Ensemble {
...
-					SealProofType: proofType,
-					Peer:          abi.PeerID(m.Libp2p.PeerID),
+				// nv, err := m.FullNode.FullNode.StateNetworkVersion(ctx, types.EmptyTSK)
+				// require.NoError(n.t, err)
+
+				// TODO this doesn't currently work with the FVM -- we need to specify policy somehow
+				// proofType, err := miner.WindowPoStProofTypeFromSectorSize(m.options.sectorSize)
+				// require.NoError(n.t, err)
+				// so do this instead, which works:
+				proofType := abi.RegisteredPoStProof_StackedDrgWindow64GiBV1
```
magik6k, line 384: todo

node/builder_chain.go
```diff
@@ -48,9 +48,16 @@ var ChainNode = Options(
...
```
magik6k, line 56: This can now be (even should)
```suggestion
	Override(new(dtypes.BuiltinActorsLoaded), modules.LoadBultinActors),
```

magik6k (commented):
Few comments

build/builtin-actors/fetch-bundles.sh
```diff
@@ -0,0 +1,57 @@
...
+check() {
+    file=$1
+    hash=$2
+    if [ -e "$file" ]; then
+        file_hash=$(sha256sum "$file" | cut -d ' ' -f 1)
+        if [ "$file_hash" == "$hash" ]; then
+            return 0
+        else
+            return 1
+        fi
```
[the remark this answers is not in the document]
vyzo: yeah, i'll simplify a tad.

chain/actors/agen/main.go
```diff
@@ -104,6 +110,7 @@ func generateState(actDir string) error {
 		err := tpl.Execute(&b, map[string]interface{}{
 			"v":      version,
 			"import": getVersionImports()[version],
+			"isv8":   version == 8,
```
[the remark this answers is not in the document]
vyzo: I was struggling to make the template comply using eq, and ended defining it outside coz thats the way imposed by the overlords.

I should make it use >= however, you are right.

chain/consensus/filcns/upgrades.go
```diff
@@ -1356,12 +1420,13 @@ func PreUpgradeActorsV8(ctx context.Context, sm *stmgr.StateManager, cache stmgr
...
 		ProgressLogPeriod: time.Minute * 5}
 
-	_, err = upgradeActorsV8Common(ctx, sm, cache, lbRoot, epoch, lbts, config)
+	_, err = upgradeActorsV8Common(ctx, sm, cache, manifest, lbRoot, epoch, lbts, config)
 	return err
 }
 
 func upgradeActorsV8Common(
 	ctx context.Context, sm *stmgr.StateManager, cache stmgr.MigrationCache,
+	manifest cid.Cid,
```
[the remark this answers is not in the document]
vyzo: hrm, maybe.

itests/kit/ensemble.go
vyzo: will be fixed once we have test bundles to load.

chain/gen/genesis/util.go
```diff
@@ -46,3 +49,36 @@ func doExecValue(ctx context.Context, vm *vm.VM, to, from address.Address, value
...
+		acts = append(acts, a)
+		return nil
+	})
+	if err != nil {
+		return xerrors.Errorf("error collecting actors: %w", err)
+	}
+
+	for _, a := range acts {
+		err = st.MutateActor(a, func(act *types.Actor) error {
+			name := actors.CanonicalName(builtin.ActorNameByCode(act.Code))
```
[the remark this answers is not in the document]
vyzo: well, this happened organically and i ended up with CanonicalName here.
You are probably right and we should use Canonical name in the put/get methods.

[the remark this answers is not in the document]
vyzo: the correct cids... thays a big conversation, but it is not workable to expect specs-actors to have baked in the correct CIDs.

node/builder_chain.go
vyzo: yrah, no need for the special.

vyzo:
So the merge overwrote the specs-actors dep and results in a broken manifest loader; see https://github.com/filecoin-project/specs-actors/pull/1587

vyzo:
fixed, updated https://github.com/filecoin-project/specs-actors/pull/1587

vyzo:
Also, the forks_test ended broken; fixed.

build/builtin-actors/fetch-bundles.sh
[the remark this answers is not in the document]
vyzo: done.

chain/consensus/filcns/upgrades.go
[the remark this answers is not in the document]
vyzo: yeah, you are right -- i will remove it from the schedule and get it here.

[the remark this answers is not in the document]
vyzo: done.

chain/vm/runtime.go
```diff
@@ -222,6 +225,23 @@ func (rt *Runtime) GetActorCodeCID(addr address.Address) (ret cid.Cid, ok bool)
...
+			builder := cid.V1Builder{Codec: cid.Raw, MhType: mh.IDENTITY}
+			synthetic := fmt.Sprintf("fil/%d/%s", av, name)
+			syntheticCid, err := builder.Sum([]byte(synthetic))
+			if err != nil {
+				panic(aerrors.Fatalf("failed to generate synthetic CID: %s", err))
+			}
+
+			return syntheticCid, true
+		}
+	}
```
magik6k, line 243: Really I think we should be trying to get rid of the `id(fil/8/y)` code cids from the codebase, otherwise we'll have to translate them back and forth everywhere, which will be really painful, especially with more network upgrades

node/builder_chain.go
```diff
@@ -48,9 +48,16 @@ var ChainNode = Options(
 	// Full node or lite node
 	// TODO: Fix offline mode
 
+	// FVM: builtin actor bundle loading
+	//      Note: this has to load before the upgrade schedule, so that we can patch in the
+	//      right manifest cid.
+	//      This restriction will be lifted once we have the final actors v8 bundle and we know
+	//      the manifest cid.
+	Override(new(dtypes.BuiltinActorsLoaded), modules.LoadBultinActors),
```
magik6k, line 56: We'll also need to load the manifest on the miner/market nodes as those are using `chain/actors` on top of API-backed blockstores.

This should probably be done as an API helper method, so that other Go programs importing lotus can do that easily too.

magik6k (commented):
Feels close to being landable

chain/gen/genesis/util.go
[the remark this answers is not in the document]
magik6k: The manifest should allow us to remove the baked-in code cids from v8 specs-actors, no?

chain/vm/invoker.go
```diff
@@ -89,22 +89,38 @@ func (ar *ActorRegistry) Invoke(codeCid cid.Cid, rt vmr.Runtime, method abi.Meth
...
+		ai := &actorInfo{
 			methods:   code,
 			vmActor:   a,
 			predicate: pred,
 		}
 
+		ac := a.Code()
+		ar.actors[ac] = ai
+
+		// necessary to make stuff work
```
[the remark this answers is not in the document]
vyzo: agreed.

chain/vm/runtime.go
vyzo: yes, I totally agree, but it's not as simple -- we'll have to do some surgery in specs-actors.

node/builder_chain.go
vyzo: ah, good catch.

vyzo: we'll have to discuss this in the huddle, not sure how to do it.

vyzo:
rebased on nv16 and resolved the conflicts; Summoning @arajasek and @magik6k for final review.

cmd/lotus-miner/init.go
```diff
@@ -214,6 +217,19 @@ var initCmd = &cli.Command{
...
+
+				if err := actors.LoadBundle(context.TODO(), bs, actors.Version8, build.BuiltinActorsV8Bundle()); err != nil {
+					return xerrors.Errorf("error loading actor bundle: %w", err)
+				}
+
+				cborStore := cbor.NewCborStore(bs)
+				if err := actors.LoadManifests(ctx, cborStore); err != nil {
+					return xerrors.Errorf("error loading actor manifests: %w", err)
+				}
+			}
```
magik6k, line 231: Can we put this as a function in chain/actors? Something like 'LoadManifestsFromBuiltin'? We'll probably need to call this in a few more places

chain/vm/runtime.go
magik6k: Won't we remove most Go specs-actors logic, and mostly just keep the data-accessing bits? That should remove almost all dependencies on old code CIDs. (I'm assuming that builtin-actors don't use them - didn't look at them much yet)

chain/gen/genesis/genesis.go
```diff
@@ -580,6 +580,23 @@ func MakeGenesisBlock(ctx context.Context, j journal.Journal, bs bstore.Blocksto
 		return nil, xerrors.Errorf("setup miners failed: %w", err)
 	}
 
+	if template.NetworkVersion >= network.Version16 {
```
magik6k, line 583: Might not be easy, but we should make the genesis logic use the FVM. Definitely not needed now, but this is the only thing which depends on a big chunk of Go specs-actors logic

magik6k: (definitely not something we want to do in this PR / not needed for M1)

chain/gen/genesis/util.go
```diff
@@ -46,3 +49,36 @@ func doExecValue(ctx context.Context, vm *vm.LegacyVM, to, from address.Address,
 
 	return ret.Return, nil
 }
+
+func patchStateTree(st *state.StateTree, nv network.Version) error {
```
magik6k, line 53: Maybe call this `setManifestCodeCids`?

magik6k (commented):
* For whatever reason make-gen complains about stuff in `build/builtin-actors`
* Something about itests / ensemble is very broken - I'm ok with fixing those in feat/nv16 to not block this any more

vyzo:
yrah, i have no idea whats wrong with make gen.

Itests need a policy adjusted bundle before they can be fixed, i have listed it as TODO follow up in the nv16 tracking pr -- basically improving our bundle situation anf fixing itests is next on my pile.

cmd/lotus-miner/init.go
vyzo: sure!

chain/gen/genesis/genesis.go
vyzo: yeah, agreed. The sooner specs-actors retires, the better.

chain/gen/genesis/util.go
vyzo: ok, will rename to something better.

chain/vm/runtime.go
vyzo: if we remove all specs-actors logic we will be fine; really, we should become fvm only, except for state.

cmd/lotus-miner/init.go
vyzo: the other place where we could call it we also have logic for v7 loading, so it doesn't quite map; but this is done, added `actors.LoadManifestFromBundle` which combines `LoadBundle` and `LoadManifests`.

vyzo: see https://github.com/filecoin-project/lotus/pull/8429/commits/beb3369c765e0d813b01fbc71243693380a892fc

chain/gen/genesis/util.go
vyzo: patchManifestCodeCids it is.

magik6k (approved):

chain/gen/genesis/genesis.go
kaka1365: Oki

kaka1365 (commented):
Good

kaka1365:
Good

Issue #6926, opened 2021-07-29
Make "lotus wallet delete" actually delete the key


f8-ptrk:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

if people see "delete" they expect the key to deleted in some way - right now the file holding the key is just renamed 

### Describe the solution you'd like

at least remove the file from the filesystem index. optimally warn if a secure delete operation cannot be perform due to filesystem constraints - if thats not the case delete it, even securely delete it (if thats even possible) or go as far as possible in trying to securely delete it

### Describe alternatives you've considered

recommend burning all disks that ever held an owner key for a miner

### Additional context

https://filecoinproject.slack.com/archives/CEGN061C5/p1627525841079800

dayou5168:
one question: should we keep the rename function which now is called  "delete"

PR #8442, opened 2022-04-06
ux: wallet: update delete usage

rjan90:
Renaming the cmd to soft-delete to better reflect what it does -  and specify that a hard deletion is needed for permanent removal.

Should address #6926 to a degree, becuase it tells the user that extra steps are needed at least.

## Checklist

- [ ] CI is green

### DIFF
--- a/cli/wallet.go
+++ b/cli/wallet.go
@@ -499,7 +499,7 @@ var walletVerify = &cli.Command{
 }
 
 var walletDelete = &cli.Command{
-	Name:      "soft-delete",
+	Name:      "delete",
 	Usage:     "Soft delete an address from the wallet - hard deletion needed for permanent removal",
 	ArgsUsage: "<address> ",
 	Action: func(cctx *cli.Context) error {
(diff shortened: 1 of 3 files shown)


Kubuxu:
We cannot just remove an existing command. We can rename the command but we should at least keep the old command as an alias (and maybe display a deprecation notice, when the alias is used).

Kubuxu (changes requested):
We cannot just rename an existing command, it will break existing scripts.

We could proceed with the rename but we should add an alias to `delete` and maybe a deprecation notice. @jennijuju what do you think?

Kubuxu:
> any preference

no really, can really go either way for me

arajasek:
Updating the usage is sufficient imo.

jennijuju (approved):

magik6k (approved):

momack2:
Is there any instruction we can give folks on HOW to hard delete the file? Seems like an important next step we should make accessible but very clear to avoid mistakes! Is there a “hard-delete” command we could add?

PR #8450, opened 2022-04-07
testkit: give up on waiting for the RPC server to shutdown after 1s

dirkmc:
Some itests fail to complete because the testkit RPC server blocks waiting for all incoming connections to be closed.
This PR changes the shutdown code to wait up to 1s and then give up on shutting down the server.


itests/kit/rpc.go
```diff
@@ -34,6 +37,23 @@ func CreateRPCServer(t *testing.T, handler http.Handler, listener net.Listener)
...
+	}()
+
+	timer := time.NewTimer(waitTime)
+	defer timer.Stop()
+	select {
+	case <-ch:
+	case <-timer.C:
+		fmt.Println(errMsg)
+		return
+	}
```
nonsense, line 54:
```suggestion
	select {
	case <-ch:
	case <-time.After(waitTime):
		fmt.Println(errMsg)
		return
	}
```

nonsense (approved):

magik6k (approved):

PR #8469, opened 2022-04-12
Update libp2p and enable rcmgr by default

vyzo:
- Updates go-libp2p@v0.18.1
- enables rcmgr by default, opt-out with LOTUS_RCMGR=0

### DIFF
--- a/node/modules/lp2p/rcmgr.go
+++ b/node/modules/lp2p/rcmgr.go
@@ -28,8 +28,8 @@ import (
 func ResourceManager(connMgrHi uint) func(lc fx.Lifecycle, repo repo.LockedRepo) (network.ResourceManager, error) {
 	return func(lc fx.Lifecycle, repo repo.LockedRepo) (network.ResourceManager, error) {
 		envvar := os.Getenv("LOTUS_RCMGR")
-		if envvar == "" || envvar == "0" {
-			// TODO opt-in for now -- flip this to enabled by default once we are comfortable with testing
+		if envvar == "0" {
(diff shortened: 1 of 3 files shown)


node/modules/lp2p/rcmgr.go
```diff
@@ -28,8 +28,8 @@ import (
 func ResourceManager(connMgrHi uint) func(lc fx.Lifecycle, repo repo.LockedRepo) (network.ResourceManager, error) {
 	return func(lc fx.Lifecycle, repo repo.LockedRepo) (network.ResourceManager, error) {
 		envvar := os.Getenv("LOTUS_RCMGR")
-		if envvar == "" || envvar == "0" {
-			// TODO opt-in for now -- flip this to enabled by default once we are comfortable with testing
+		if envvar == "0" {
```
magik6k, line 31: We probably want some more testing before defaulting to enabled

magik6k (approved):

node/modules/lp2p/rcmgr.go
vyzo: sure, its master though

jennijuju:
@vyzo isnit possible to have enable by default and the rest in two PRs?
I want to land enable by default in the upcoming release branch only for testing.
Then the rest of dep update will be in both master and release branch.

vyzo:
done.

node/modules/lp2p/rcmgr.go
vyzo: dropped, will make follow up pr wiht it.

vyzo:
@jennijuju #8470

Issue #5371, opened 2021-01-18
unknown actor code bafkqaetgnfwc6mrpon2g64tbm5sw22lomvza


astudnev:
rpc.state_account_key 'f03275'

returns the error:

Error response to Filecoin.StateAccountKey request: 1  failed to get account actor state for f03275: unknown actor code bafkqaetgnfwc6mrpon2g64tbm5sw22lomvza

While reverse works:

rpc.state_lookup_i_d 'f27fxgjifczreumheqdnlhm2esbt2mg5nugcfiyaq', nil

> f03275

It is unclear how to get public key. address for miner?

donnol:
Is there anything update?

eskrano:
Same issue

AllanOricil:
I get this when I run any of the  `lotus filplus` commands

fadyanwar:
Any updates on this?
I get this error during the initial sync using AMIs provided in the docs

 ⨎ lotus sync status
2021-12-30T20:57:41.259Z	WARN	cliutil	util/api.go:162	Couldn't load CLI token, capabilities may be limited: open /var/lib/lotus/token: permission denied
2021-12-30T20:57:41.260Z	WARN	cliutil	util/apiinfo.go:81	API Token not set and requested, capabilities might be limited.
sync status:
worker 4:
	Base:	[bafy2bzacecyu6bhlchulxvfbfrnzgyvxdtmbzgmzwlzkydiu4mq74ch5zcnke bafy2bzacebmo4unhmetnu5q6mhxorve44lifj33r7fppdsiwmz2fr35q5yqzc bafy2bzacea4b73tdwcdog53yrwseo7ubsvm7z7hp24w76gx2itqa75e5zq6ii]
	Target:	[bafy2bzacecqjb6e7mpyr5mxwwlkykv4ebsvykv4hrieutgs5yfw6cguv6mc6m bafy2bzacebyyqd4rfkfqfljucnfbo5qkmy7lrrgwfwnvfrt6e74e5yzeyetf2 bafy2bzacecreb2bswuae24quf4te3idtvizpkh6exurexfuy5hnmxnmsfndjk] (1419683)
	Height diff:	803
	Stage: error
	Height: 0
	Elapsed: 39.416µs
[...]

PR #8486, opened 2022-04-14
Feat: API: Implement StateLookupRobustAddress

geoff-vball:
## Related Issues
Fixes https://github.com/filecoin-project/lotus/issues/5371

## Proposed Changes
added endpoint StateLoockupRobustAddress which will return the robust address corresponding to any ID Address

## Checklist

- [ ] CI is green


api/api_full.go
```diff
@@ -522,8 +522,10 @@ type FullNode interface {
 	StateMarketStorageDeal(context.Context, abi.DealID, types.TipSetKey) (*MarketDeal, error) //perm:read
 	// StateLookupID retrieves the ID address of the given address
 	StateLookupID(context.Context, address.Address, types.TipSetKey) (address.Address, error) //perm:read
-	// StateAccountKey returns the public key address of the given ID address
+	// StateAccountKey returns the public key address of the given ID address for secp and bls accounts
 	StateAccountKey(context.Context, address.Address, types.TipSetKey) (address.Address, error) //perm:read
+	// StateLookupRobustAddress returns the public key address of the given ID address for non-account addresses (multisig, miners etc)
+	StateLookupRobustAddress(context.Context, address.Address, types.TipSetKey) (address.Address, error) //perm:read
```
Kubuxu, line 528: Looking at the implementation this for sure is slow due to data structures. Do we want to allow anyone (including for example browser) to call this?

chain/stmgr/stmgr.go
```diff
@@ -318,6 +321,48 @@ func (sm *StateManager) LookupID(ctx context.Context, addr address.Address, ts *
 	return state.LookupID(addr)
 }
 
+func (sm *StateManager) LookupRobustAddress(ctx context.Context, idAddr address.Address, ts *types.TipSet) (address.Address, error) {
+	idAddrDecoded, err := address.IDFromAddress(idAddr)
+	if err != nil {
+		return address.Undef, xerrors.Errorf("address couldnt decode to id addr: %w", err)
```
arajasek, line 327:
```suggestion
		return address.Undef, xerrors.Errorf("failed to decode provided address as id addr: %w", err)
```

node/impl/full/state.go
```diff
@@ -449,6 +450,21 @@ func (m *StateModule) StateLookupID(ctx context.Context, addr address.Address, t
 	return m.StateManager.LookupID(ctx, addr, ts)
 }
 
+func (m *StateModule) StateLookupRobustAddress(ctx context.Context, addr address.Address, tsk types.TipSetKey) (address.Address, error) {
+	ts, err := m.Chain.GetTipSetFromKey(ctx, tsk)
+	if err != nil {
+		return address.Undef, xerrors.Errorf("loading tipset %s: %w", tsk, err)
+	}
+	if ts.Height() > policy.ChainFinality {
```
arajasek, line 458: So, thinking about this some more, we aren't actually using a finality lookback when resolving `Account`s in `StateAccountKey`. While I think we probably should be, this method should just match behaviour there, so I'd actually drop the finality lookback here (in contrast to what I said we should do in the issue).

itests/self_sent_txn_test.go
```diff
@@ -102,3 +102,23 @@ func TestSelfSentTxnV14(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, exitcode.Ok, mLookup.Receipt.ExitCode)
 }
+
+func TestStateLookupRobustAddress(t *testing.T) {
+	ctx := context.Background()
+	kit.QuietMiningLogs()
+
+	client15, miner, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.GenesisNetworkVersion(network.Version15))
```
arajasek, line 110:
```suggestion
	client, miner, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.GenesisNetworkVersion(network.Version15))
```

Rename this to just `client`? `client15` was relevant to the other test in this file (it was relevant to network v15).

itests/self_sent_txn_test.go
```diff
@@ -102,3 +102,23 @@ func TestSelfSentTxnV14(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, exitcode.Ok, mLookup.Receipt.ExitCode)
 }
+
+func TestStateLookupRobustAddress(t *testing.T) {
```
arajasek, line 106: This test should move into its own file.

api/api_gateway.go
```diff
@@ -54,6 +54,7 @@ type Gateway interface {
 	StateReadState(ctx context.Context, actor address.Address, tsk types.TipSetKey) (*ActorState, error) //perm:read
 	StateListMiners(ctx context.Context, tsk types.TipSetKey) ([]address.Address, error)
 	StateLookupID(ctx context.Context, addr address.Address, tsk types.TipSetKey) (address.Address, error)
+	StateLookupRobustAddress(ctx context.Context, addr address.Address, tsk types.TipSetKey) (address.Address, error)
```
arajasek, line 57: We definitely don't want this in the gateway (sloooow).

chain/stmgr/stmgr.go
```diff
@@ -318,6 +321,48 @@ func (sm *StateManager) LookupID(ctx context.Context, addr address.Address, ts *
...
+	err = initState.ForEachActor(func(id abi.ActorID, addr address.Address) error {
+		if uint64(id) == idAddrDecoded {
+			robustAddr = addr
+			// Hacky way to early return from ForEach
+			return xerrors.New("robust address found")
+		}
+		return nil
+	})
+	if err != nil {
+		if robustAddr == address.Undef {
```
arajasek, line 358: I would move this check to _before_ the error check -- assume that if robustAddr got set, you don't have an error (in contrast to actual correct error handling).

api/api_full.go
Kubuxu: It would be best to bump the StateCompute perm as well (we could add another permission `slow` or somethign).

arajasek: Yeah, that's a good idea, but I think it'll be a future thing.

chain/stmgr/stmgr.go
```diff
@@ -318,6 +321,48 @@ func (sm *StateManager) LookupID(ctx context.Context, addr address.Address, ts *
...
+		if uint64(id) == idAddrDecoded {
+			robustAddr = addr
+			// Hacky way to early return from ForEach
+			return xerrors.New("robust address found")
+		}
+		return nil
+	})
+	if robustAddr == address.Undef {
+		if err == nil {
+			return address.Undef, xerrors.Errorf("Address %s not found", idAddr.String())
```
arajasek, line 359:
```suggestion
			return address.Undef, xerrors.Errorf("Address %d not found", idAddr)
```

arajasek (approved):

PR #8509, opened 2022-04-19
[nv16] actor bundle integration

vyzo:
This ingests the bundles produced at release and selects the appropriate bundle based on build tags.

TBD:
- [x] update ffi
- [x] select bundle by build tags
- [x] DI ingest test bundle by file when instantiating itests


raulk (approved):
I don't really like the per-network `build_*` files, as we are trying to make the network parameterization less static in other efforts. But this works for now.

PR #8517, opened 2022-04-20
feat: Add rate limiting to the lotus gateway

coryschwartz:
Add rate limiting to the lotus gateway.

This allows us to limit the number of requests the lotus gateway will allow.

Each API call can have its own cost depending on how much load is likely to be induced. API calls are permitted to go through, so long as there are enough tokens available in the bucket before the timeout expires.

If there are insufficient tokens, an error is returned and the API call is not performed.


coryschwartz:
Ran a couple of quick tests and dirty tests


Without the rate limiter Queries to get the genesis tipset respond quickly at about 6.7k qps on my test setup.

With the rate limiter, but with `--rate-limit 0`, and there was no impact to performance. 

Actually, the performance improved a little bit... but near enough consider the performance identical

gateway/node.go
```diff
@@ -21,12 +23,18 @@ import (
...
 	_ "github.com/filecoin-project/lotus/lib/sigs/bls"
 	_ "github.com/filecoin-project/lotus/lib/sigs/secp"
+	"github.com/filecoin-project/lotus/metrics"
 	"github.com/filecoin-project/lotus/node/impl/full"
 )
 
 const (
 	DefaultLookbackCap            = time.Hour * 24
 	DefaultStateWaitLookbackLimit = abi.ChainEpoch(20)
+	DefaultRateLimitTimeout       = time.Minute * 10
```
travisperson, line 33: Does this mean a hanging request for 10 minutes?

coryschwartz: We would override this.

travisperson:
Just a note: this appears to be a global rate limit. So while it will protect our backend, it does mean that the system can still be DOS'd very easily just simply making a request to `Filecoin.Version` which will eat 1 token. This can easily exhaust the system of tokens for all other users.

coryschwartz:
> Just a note: this appears to be a global rate limit. So while it will protect our backend, it does mean that the system can still be DOS'd very easily just simply making a request to `Filecoin.Version` which will eat 1 token. This can easily exhaust the system of tokens for all other users.

This is true. This will protect the backend from being overworked so it can continue processing messages under heavy load, but it does not prevent DOS attacks.

There is still some benefit to load shedding at the API layer even if only to allow the backend to not to become overwhelmed.

we can add some additional limits for each connnection and this would partially mitigate the problem. At least then a single individual cannot DOS the service by issuing Filecoin.Version as you suggest. Is that what you are thinking @travisperson ?

DDoS mitigation would likely have to be handled with scalability. As you suggest, this rate limiter would not help in that case except to keep the backend working.

travisperson:
I think it's just something we need to understand and determine if it's acceptable. If one user exhausting all tokens is unacceptable then we need a better solution to limit per connection / ip.

coryschwartz:
I'm moving this back to draft due to concerns about using only global rate limits.

I'll implement a per-connection rate limiting and will produce some performance benchmarks before I put this back out for review.

coryschwartz:
@travisperson can you give this a try? 

There's two new rate limiters in addition to the global one.

One limits the number of HTTP connections you're allowed to open in a single minute.

The other adds limits for RPC call over an established websocket.

gateway/node.go
magik6k: I'd set this to some reasonably high value; 5s is probably good enough

gateway/handler.go
```diff
@@ -44,5 +49,95 @@ func Handler(a api.Gateway, opts ...jsonrpc.ServerOption) (http.Handler, error)
...
+		return
+	}
+	host, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	h.mu.Lock()
+	seen, ok := h.ipmap[host]
```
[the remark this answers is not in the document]
magik6k: My main worry is that with ipv6 you get a /64 subnet, which is a lot of IPs which are tracked separately; but yes, the other limits should help here

magik6k (approved):

PR #8521, opened 2022-04-20
[nv16] fix itests

vyzo:


raulk (approved):

chain/vm/fvm.go
```diff
@@ -257,7 +257,6 @@ type FVM struct {
 }
 
 func NewFVM(ctx context.Context, opts *VMOpts) (*FVM, error) {
-	log.Info("using the FVM, this is experimental!")
```
arajasek, line 260: Let's leave this log in there for now, if only for us to know which VM is being run (we will def take it out before the final release).

arajasek (approved):

chain/vm/fvm.go
arajasek: Merging for now, can add it back later.

vyzo: it's spamming a lot!

PR #8530, opened 2022-04-21
NV16: fix paych

vyzo:


arajasek (approved):

vyzo:
The paych-api itest is failing because it needs a proof, and we don't have our story with the fake proofs figured out yet.

PR #8532, opened 2022-04-21
Interopnet Actual

travisperson:
There seems to be a miss match between `feat/nv16` and what we are running on interopnet.


travisperson:
From the diff, it looks like we reverted a bunch of stuff at some point. 

cc @arajasek

travisperson:
This is when the bootstrap / genesis info was removed https://github.com/filecoin-project/lotus/pull/8374 👅

PR #8534, opened 2022-04-22
Use new go-state-types state accessors

### LINKED ISSUE #8063: Support actor state accessors using canonical actor
I'm not quite sure what the right thing to do here is. Although the _actual_ state objects aren't having their shapes changed, we currently use go-actors code for a bunch of getters (and the occasional setter for test things). We could:

- pull these getters into lotus as some "canonical actor state accessor" suite
- punt the problem: leave them in go-actors and use them -- this isn't _terrible_ because almost all of the accessors are used for non-consensus critical things, but we will eventually have to do the work unless we wanna keep go-actors forever :P 
- do...something else? have some spec-actor-states written in go that always matches the canonical actors?

arajasek:
Resolves #8063 

This PR replaces uses of the specs-actors v8 repo in Lotus and (most of) its deps. The current specs-actors v8 is _not_ the specification of Filecoin actors, and mostly there for prototyping and convenience. This PR integrates actor state types and params from go-state-types.

There are 2 remaining uses of actors v8:

- the v15-v16 migration (I think we're comfortable leaving this there for now)
- the builtin registry uses the exported methods for info about method params / returns. I think this should definitely move to the go-state-types repo as well, but want to descope it from this PR (consider it a future improvement).

Needs 

- [x] https://github.com/filecoin-project/go-state-types/pull/36 (needed by all other PRs)
- [x] https://github.com/filecoin-project/specs-storage/pull/25
- [x] https://github.com/filecoin-project/specs-actors/pull/1590
- [x] https://github.com/filecoin-project/filecoin-ffi/pull/269
- [x] https://github.com/filecoin-project/go-fil-markets/pull/711


api/api_full.go
```diff
@@ -918,6 +919,20 @@ type MinerPower struct {
 	HasMinPower bool
 }
 
+type MinerInfo struct {
```
vyzo, line 922: move this with the other typedefs?

chain/actors/builtin/builtin.go
```diff
@@ -166,8 +93,8 @@ func ActorNameByCode(c cid.Cid) string {
 	case builtin7.IsBuiltinActor(c):
 		return builtin7.ActorNameByCode(c)
 
-	case builtin8.IsBuiltinActor(c):
-		return builtin8.ActorNameByCode(c)
+	case builtin8sa.IsBuiltinActor(c):
```
vyzo, line 96: just builtin8 to be consistent?

chain/actors/builtin/builtin.go.template
```diff
@@ -5,20 +5,22 @@ import (
...
-	{{range .versions}}
-		builtin{{.}} "github.com/filecoin-project/specs-actors{{import .}}actors/builtin"
-		smoothing{{.}} "github.com/filecoin-project/specs-actors{{import .}}actors/util/smoothing"
-	{{end}}
+{{range .versions}}
+    {{if (le . 7)}}
+	    builtin{{.}} "github.com/filecoin-project/specs-actors{{import .}}actors/builtin"
+	{{else}}
+        builtin{{.}} "github.com/filecoin-project/go-state-types/builtin"
+        builtin{{.}}sa "github.com/filecoin-project/specs-actors{{import .}}actors/builtin"
```
vyzo, line 13: drop the `sa` i would say.

chain/actors/builtin/builtin.go.template
```diff
@@ -45,52 +47,31 @@ const (
...
 	switch {
-		{{range .versions}}
-			case builtin{{.}}.IsBuiltinActor(c):
-			return builtin{{.}}.ActorNameByCode(c)
-		{{end}}
+	    {{range .versions}}
+            {{if (le . 7)}}
+                case builtin{{.}}.IsBuiltinActor(c):
+                return builtin{{.}}.ActorNameByCode(c)
+            {{else}}
```
vyzo, line 70: I dont think we either need or want this branch.
After v8 it should always be in the manifest and gst should have no knowledge of this.

vyzo (commented):
looks pretty good, just some small nits.

Also, shouldn't we do something about genesis (for tests)?

api/api_full.go
raulk: The JSON-RPC API struct is different to state struct, I think. I think it's fine here.

vyzo: I meant to `api/types.go`, that's the logical place where it belongs.

arajasek: A bunch of the api types are already in api_full.go, but I think i agree with @vyzo -- moving.

chain/actors/manifest.go
```diff
@@ -24,6 +25,18 @@ var manifestCids map[Version]cid.Cid = map[Version]cid.Cid{
...
+const CronKey = "cron"
+const InitKey = "init"
+const MarketKey = "storagemarket"
+const MinerKey = "storageminer"
+const MultisigKey = "multisig"
+const PaychKey = "paymentchannel"
+const PowerKey = "storagepower"
+const RewardKey = "reward"
+const SystemKey = "system"
+const VerifregKey = "verifiedregistry"
```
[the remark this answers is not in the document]
arajasek: That is nicer!

chain/actors/builtin/account/account.go
```diff
@@ -26,80 +24,22 @@ import (
...
-
-		case actors.Version5:
-			return load5(store, act.Head)
-
-		case actors.Version6:
-			return load6(store, act.Head)
-
-		case actors.Version7:
-			return load7(store, act.Head)
-
```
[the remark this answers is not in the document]
arajasek: No, this is code that was added in [this commit ](https://github.com/filecoin-project/lotus/commit/cdd5420e0c93d60731de2238464bb8d628350256), but I feel like this wasn't the right thing to do -- we shouldn't be relying on the ActorMeta to tell us anything about versions 0-7. The former codeCid-based switch (directly after this) works for those cases.

arajasek:
> Also, shouldn't we do something about genesis (for tests)?

@vyzo This PR makes all genesis happen through the FVM (for actors v8 onwards). Anything else you wanted to see happen for it?

chain/actors/builtin/builtin.go.template
arajasek: I'm not sure I know exactly what you mean here -- what are you proposing dropping?

vyzo:
no, thats what we want!

chain/actors/builtin/builtin.go.template
vyzo: the else branch

raulk (approved):

PR #8557, opened 2022-04-26
sealing: DataCid on workers

magik6k:
Depends on https://github.com/filecoin-project/specs-storage/pull/26

This PR adds a `ComputeDataCid` which will use add-piece workers to compute piece CIDs. This is mostly useful for scaling boost.

Followups:
* lotus-miner command to compute CommP of a file/url
* Integrate into fil-markets (maybe)


extern/sector-storage/ffiwrapper/sealer_cgo.go
```diff
@@ -51,6 +51,128 @@ func (sb *Sealer) NewSector(ctx context.Context, sector storage.SectorRef) error
...
+		}
+		if read == 0 {
+			break
+		}
+
+		done := make(chan struct {
+			cid.Cid
+			error
+		}, 1)
+		pbuf := <-throttle
```
dirkmc, line 103: Can you get the buffer from the throttle channel at the top of this for loop? That way you could read directly into the buffer instead of reading to an intermediate buffer and copying

magik6k: We could, but reading before throttling should be (very slightly) faster / more efficient, as you're not waiting with some threads for the read to complete - the data already sits in a buffer

extern/sector-storage/ffiwrapper/sealer_cgo.go
```diff
@@ -51,6 +51,128 @@ func (sb *Sealer) NewSector(ctx context.Context, sector storage.SectorRef) error
...
+					PieceCID: e.Cid,
+				}, nil
+			case <-ctx.Done():
+				return abi.PieceInfo{}, ctx.Err()
+			}
+		})
+	}
+
+	if len(piecePromises) == 1 {
+		return piecePromises[0]()
```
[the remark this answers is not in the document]
magik6k: Nope, like AddPiece we require that the piece is already power-of-two sized, so this never needs padding

itests/worker_test.go
```diff
@@ -40,6 +41,21 @@ func TestWorkerPledge(t *testing.T) {
 	miner.PledgeSectors(ctx, 1, 0, nil)
 }
 
+func TestWorkerDataCid(t *testing.T) {
```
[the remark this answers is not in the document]
magik6k: Those sizes must be power-of-two, but agreed on the test

magik6k:
So the itest with the bigger buffer is really flaky, with something weird in rpcenc happening:

```
2022-04-27T20:46:10.614+0200	WARN	rpc	go-jsonrpc@v0.1.5/handler.go:279	error in RPC call to 'Filecoin.ComputeDataCid': storage call error 0: %!w(pr read error: http: invalid Read on closed Body [Hostname: ukaszs-MacBook-Pro.local])
    worker_test.go:60: 
        	Error Trace:	worker_test.go:60
        	            	storage call error 0: %!w(pr read error: http: invalid Read on closed Body [Hostname: ukaszs-MacBook-Pro.local])
        	Test:       	TestWorkerDataCid
    blockminer.go:326: shutting down mining
```

Which is weird, because I don't think we've ever saw this in AddPiece (but we've also never tested it in itests with a worker + bigger bit of data); Something about how we handle responses in rpcenc is probably racy.

extern/sector-storage/ffiwrapper/sealer_cgo.go
dirkmc: That makes sense assuming that the io buffer is smaller than chunk.Unpadded(), which I think is correct 👍

dirkmc (approved):
It seems like the rpcenc error is unrelated to this change, so I'm approving 👍 

Seems like it's worth looking into the rpcenc error as a follow-up

dirkmc (approved):

PR #8582, opened 2022-05-02
fix: sealing: Finalize snap sectors before submitting proofs

magik6k:
## Related Issues
Might help with the issue raised in https://filecoinproject.slack.com/archives/C038C47NWAE/p1651270743875089

## Proposed Changes
Move snapdeals sectors to proving storage before submitting the proof to the chain.

### DIFF
--- a/extern/storage-sealing/fsm.go
+++ b/extern/storage-sealing/fsm.go
@@ -166,24 +166,24 @@ var fsmPlanners = map[SectorState]func(events []statemachine.Event, state *Secto
 		on(SectorAbortUpgrade{}, AbortUpgrade),
 	),
 	ProveReplicaUpdate: planOne(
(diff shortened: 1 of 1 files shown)


ZenGround0 (approved):
I understand the change and like that it moves the replica update flow closer to the commit flow but I don't understand how the current state could cause post failures.  Can you describe the mechanism?

magik6k:
Basically we’d submit the proof while the sector is still in sealing storage, and then start to move it - if the sealing pipeline is backlogged, this can take a while, and we may enter a deadline, and you still can’t prove the sector.

The second, possibly more broken, thing that can happen is that we have moved the sealed file, but haven’t yet moved the cache files, we may still try to prove that sector, and there you’ll see missing p_aux

Both of those should be fixed by finalising early.

PR #8593, opened 2022-05-04
[nv16] propagate gas outputs from fvm + integrate safer ffi

vyzo:
See https://github.com/filecoin-project/ref-fvm/issues/523
Depends on https://github.com/filecoin-project/filecoin-ffi/pull/275

Integrates https://github.com/filecoin-project/filecoin-ffi/pull/247.


extern/sector-storage/ffiwrapper/sealer_test.go
```diff
@@ -869,89 +868,6 @@ func TestAddPiece512MPadded(t *testing.T) {
 	require.Equal(t, "baga6ea4seaqonenxyku4o7hr5xkzbqsceipf6xgli3on54beqbk6k246sbooobq", c.PieceCID.String())
 }
 
-func setupLogger(t *testing.T) *bytes.Buffer {
```
raulk, line 872: Could you explain why this needs to go entirely? I understand that `generated` no longer exists, but has all this functionality vanished from the library? Or is it just behind different APIs? If the latter, we should adapt the tests, not drop the coverage.

raulk (commented):
Tests that are failing are already failing.

extern/sector-storage/ffiwrapper/sealer_test.go
vyzo: its gone completely afaict.

raulk: It isn't. AFAIK no functionality has been removed, only APIs have changed. Here's where the log is now:

filecoin-project/filecoin-ffi@bb281b5/cgo/util.go#L11

vyzo: ok feel free to reinstate, it is really ugly however.

raulk: Agreed that the form of the test is terrible, but it is providing certain coverage that we can't drop here.

PR #8606, opened 2022-05-05
Fix: drand: calculation of round from Filecoin epochs

### LINKED ISSUE #2170: Drand rounds are not computed as drand computes them

arajasek:
## Related Issues

Resolves #2170 

## Proposed Changes

This PR was originally opened for the v14 upgrade, but descoped.

## Checklist

- [ ] CI is green


arajasek:
Added test.

chain/beacon/drand/drand.go
```diff
@@ -201,11 +204,32 @@ func (db *DrandBeacon) VerifyEntry(curr types.BeaconEntry, prev types.BeaconEntr
...
+func (db *DrandBeacon) maxBeaconRoundV2(latestTs uint64) uint64 {
+	if latestTs < db.drandGenTime {
+		return 1
+	}
+
+	fromGenesis := latestTs - db.drandGenTime
+	// we take the time from genesis divided by the periods in seconds, that
+	// gives us the number of periods since genesis.  We also add +1 because
+	// round 1 starts at genesis time.
+	return uint64(math.Floor(float64(fromGenesis)/db.interval.Seconds())) + 1
```
Kubuxu, line 232: I feel a bit uneasy using floats for this.
As you only need Floor it should be simple to do using pure ints.

Kubuxu:
Hey, how does it work during the upgrade?
As I recall currently we track the previous randomness beacon and verify the new one based on the previous one. How does this continue to work during the upgrade?

Kubuxu:
Ahh, right in this upgrade we only add on one breacon. Ok, I can see how that can work.

chain/beacon/drand/drand.go
arajasek: I agree, simplifying this.

arajasek:
Yeah, so what should happen is just that the first block after the upgrade kicks in should have 2 beacon entries (exactly as we have after a null tipset).

chain/beacon/drand/drand.go
```diff
@@ -201,11 +203,32 @@ func (db *DrandBeacon) VerifyEntry(curr types.BeaconEntry, prev types.BeaconEntr
...
+}
+
+func (db *DrandBeacon) maxBeaconRoundV1(latestTs uint64) uint64 {
 	dround := (latestTs - db.drandGenTime) / uint64(db.interval.Seconds())
 	return dround
 }
 
+func (db *DrandBeacon) maxBeaconRoundV2(latestTs uint64) uint64 {
+	if latestTs < db.drandGenTime {
+		return 1
```
Stebalien, line 224: not 0?

Stebalien: When `latestTs == db.drandGenTime`, the round will be 1. I don't know if it matters, but it may make more sense to let the round be 0 before that.

Stebalien (approved):

chain/beacon/drand/drand.go
arajasek: Not super important for mainnet, but I think your argument makes sense. Changing.

arajasek: Actually, looking at Drand code, round 0 seems to indicate a special value that says "give me the latest entry".

arajasek: https://github.com/drand/drand/blob/master/client/http/http.go#L311-L315

arajasek: Gonna leave as is, for now.

PR #8617, opened 2022-05-06
Gas estimation: Overestimate before the nv16 upgrade

### LINKED ISSUE #8417: Gas Estimation around NV16
While the final details of the FVM gas model are TBD, it seems likely that the cost of messages will increase after the upgrade to nv16. This poses a challenge for messages that are in the mempool at the time of the upgrade: they will have been estimated with the LegacyVM, and therefore are susceptible to running out of gas when they are executed using the FVM. 

We have a few options here:

1. Do nothing. A few epochs worth of messages will fail with Out of Gas, but all messages estimated after the upgrade will be fine. Most critical messages in Lotus have retry mechanisms on failure, so nothing too bad will happen.
2. Start naively overestimating at some period before the upgrade, by applying (say) a 2x factor. The question would be when to start doing this -- too early and messages that land before the upgrade will pay a lot in overestimation, too late and there will be some messages that were already estimated (and so might fail out of gas).
3. Start estimating using the actual FVM at some period before the upgrade. The same question about when to start would apply. This won't be too hard, but is probably more work than it's worth.

arajasek:
## Related Issues

Resolves #8417 

## Proposed Changes

Opening this using Option 2 in the issue -- we haven't finalized a decision yet, but easy enough to open this PR. 

I'm still not convinced we need to do anything here, and might prefer to go with Option 1 (do nothing). Important messages in Lotus have a retry mechanism that will re-estimate gas (I know PoSt does, would be worth examining other messages around sealing and dealmaking).

Uses a static blowup factor of 2 for now. Exact value is TBD, if we go with this approach.

## Checklist

- [ ] CI is green

### DIFF
--- a/node/impl/full/gas.go
+++ b/node/impl/full/gas.go
@@ -295,31 +295,21 @@ func gasEstimateGasLimit(
 		return -1, xerrors.Errorf("message execution failed: exit %s, reason: %s", res.MsgRct.ExitCode, res.Error)
 	}
 
+	ret := res.MsgRct.GasUsed
+
 	// Special case for PaymentChannel collect, which is deleting actor
+	// We ignore errors in this special case since they CAN occur,
+	// and we just want to detect existing payment channel actors
 	st, err := smgr.ParentState(ts)
-	if err != nil {
-		_ = err
-		// somewhat ignore it as it can happen and we just want to detect
-		// an existing PaymentChannel actor
-		return res.MsgRct.GasUsed, nil
-	}
-	act, err := st.GetActor(msg.To)
-	if err != nil {
-		_ = err
-		// somewhat ignore it as it can happen and we just want to detect
-		// an existing PaymentChannel actor
-		return res.MsgRct.GasUsed, nil
+	if err == nil {
+		act, err := st.GetActor(msg.To)
+		if err == nil && builtin.IsPaymentChannelActor(act.Code) && msgIn.Method == paych.Methods.Collect {
+			// add the refunded gas for DestroyActor back into the gas used
+			ret += 76e3
+		}
 	}
 
-	if !builtin.IsPaymentChannelActor(act.Code) {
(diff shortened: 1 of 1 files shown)


node/impl/full/gas.go
```diff
@@ -295,31 +295,26 @@ func gasEstimateGasLimit(
 		return -1, xerrors.Errorf("message execution failed: exit %s, reason: %s", res.MsgRct.ExitCode, res.Error)
 	}
 
+	ret := res.MsgRct.GasUsed
+
 	// Special case for PaymentChannel collect, which is deleting actor
+	// We ignore errors in this special case since they CAN occur,
+	// and we just want to detect existing payment channel actors
 	st, err := smgr.ParentState(ts)
```
Kubuxu, line 303: This error is being eaten.

node/impl/full/gas.go
```diff
@@ -295,31 +295,26 @@ func gasEstimateGasLimit(
...
 	}
 
-	if !builtin.IsPaymentChannelActor(act.Code) {
-		return res.MsgRct.GasUsed, nil
-	}
-	if msgIn.Method != paych.Methods.Collect {
-		return res.MsgRct.GasUsed, nil
+	// Overestimate gas used around the
+	if ts.Height() <= build.UpgradeFVM1Height && (build.UpgradeFVM1Height-ts.Height() <= 5) {
+		ret *= 2
```
Kubuxu, line 314: The best course of action will be to set the overestimation on a per (actor, method) basis for the top 5 or 7 tuples for 80th percentile and the rest of 80th percentile of all.

arajasek: @Kubuxu I'm happy to go with that. Feel free to take over this PR / close and open a new one, thanks for volunteering to do it :)

Kubuxu:
Superseded by #8620

PR #8620, opened 2022-05-09
Gas estimation: Overestimate based on (actor, method) tuples

### LINKED ISSUE #8417: Gas Estimation around NV16

Kubuxu:
## Related Issues

Resolves #8417

## Proposed Changes

Have few unique multipleirs apart from the base multipleir

### DIFF
--- a/node/impl/full/gas.go
+++ b/node/impl/full/gas.go
@@ -297,6 +297,47 @@ func gasEstimateGasLimit(
 
 	ret := res.MsgRct.GasUsed
 
+	transitionalMulti := 1.0
+	// Overestimate gas around the upgrade
+	if ts.Height() <= build.UpgradeFVM1Height && (build.UpgradeFVM1Height-ts.Height() <= 20) {
+		transitionalMulti = 2.0
+
+		func() {
+			st, err := smgr.ParentState(ts)
+			if err != nil {
+				return
+			}
+			act, err := st.GetActor(msg.To)
+			if err != nil {
+				return
+			}
+
+			if builtin.IsStorageMinerActor(act.Code) {
+				switch msgIn.Method {
+				case 5:
+					transitionalMulti = 3.954
+				case 6:
+					transitionalMulti = 4.095
+				case 7:
+					// skip, stay at 2.0
+					//transitionalMulti = 1.289
+				case 11:
+					transitionalMulti = 17.8758
+				case 16:
+					transitionalMulti = 2.1704
+				case 25:
+					transitionalMulti = 3.1177
+				case 26:
+					transitionalMulti = 2.3322
+				default:
+				}
+			}
+
+			// skip storage market, 80th percentie for everything ~1.9, leave it at 2.0
+		}()
+	}
+	ret = (ret * int64(transitionalMulti*1024)) >> 10
+
 	// Special case for PaymentChannel collect, which is deleting actor
 	// We ignore errors in this special case since they CAN occur,
 	// and we just want to detect existing payment channel actors
@@ -307,11 +348,6 @@ func gasEstimateGasLimit(
 			// add the refunded gas for DestroyActor back into the gas used
 			ret += 76e3
 		}
-	}
-
-	// Overestimate gas used around the
-	if ts.Height() <= build.UpgradeFVM1Height && (build.UpgradeFVM1Height-ts.Height() <= 5) {
-		ret *= 2
 	}
 
 	return ret, nil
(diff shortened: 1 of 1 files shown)


node/impl/full/gas.go
```diff
@@ -297,6 +297,47 @@ func gasEstimateGasLimit(
...
+				return
+			}
+			act, err := st.GetActor(msg.To)
+			if err != nil {
+				return
+			}
+
+			if builtin.IsStorageMinerActor(act.Code) {
+				switch msgIn.Method {
+				case 5:
```
arajasek, line 317: `case miner.Methods.SubmitWindowedPoSt`? (and so on for others)

node/impl/full/gas.go
```diff
@@ -297,6 +297,47 @@ func gasEstimateGasLimit(
 
 	ret := res.MsgRct.GasUsed
 
+	transitionalMulti := 1.0
+	// Overestimate gas around the upgrade
+	if ts.Height() <= build.UpgradeFVM1Height && (build.UpgradeFVM1Height-ts.Height() <= 20) {
+		transitionalMulti = 2.0
+
+		func() {
```
arajasek, line 305: what do we gain by using an anonymous function here?

node/impl/full/gas.go
```diff
@@ -297,6 +297,47 @@ func gasEstimateGasLimit(
 
 	ret := res.MsgRct.GasUsed
 
+	transitionalMulti := 1.0
+	// Overestimate gas around the upgrade
+	if ts.Height() <= build.UpgradeFVM1Height && (build.UpgradeFVM1Height-ts.Height() <= 20) {
```
arajasek, line 302: How did you pick 20 here?

arajasek (approved):
Thanks for the PR! I have some qs / suggestions, but good enough to merge!

node/impl/full/gas.go
Kubuxu: This corresponds to what is reported in collected statistics so it is easier to verify if it stays as numbers, also as this is very temporary code there is little to no harm to having raw numbers.

If you feel strongly about it I can add comments with method names.

node/impl/full/gas.go
Kubuxu: It cleans up the error handling.

node/impl/full/gas.go
Kubuxu: This gives us 10 minutes of overestimation, there are few reasons I went with 20 epochs.
I would be comfortable with 5 minutes, if we didn’t have tipset system, possible off-by-ones somewhere so on, and I feel like increasing it to 10min has a low cost.

node/impl/full/gas.go
arajasek: I don't feel strongly about it, no.

if you want me to double check the values against the collected statistics, i can, but i trust you

node/impl/full/gas.go
arajasek: ah, makes sense

node/impl/full/gas.go
arajasek: Seems reasonable -- i don't think there's a "correct" answer here.

Fatman13:
> Do nothing. A few epochs worth of messages will fail with Out of Gas, but all messages estimated after the upgrade will be fine. Most critical messages in Lotus have retry mechanisms on failure, so nothing too bad will happen.

> Start naively overestimating at some period before the upgrade, by applying (say) a 2x factor. The question would be when to start doing this -- too early and messages that land before the upgrade will pay a lot in overestimation, too late and there will be some messages that were already estimated (and so might fail out of gas).

> Start estimating using the actual FVM at some period before the upgrade. The same question about when to start would apply. This won't be too hard, but is probably more work than it's worth.

Is this PR still going for option 2 like the one before?

PR #8628, opened 2022-05-10
merge v1.15.2 into nv16 staging branch

raulk:


build/version.go
```diff
@@ -37,7 +37,7 @@ func BuildTypeString() string {
 }
 
 // BuildVersion is the local build version
-const BuildVersion = "1.15.2-dev"
+const BuildVersion = "1.15.2"
```
vyzo, line 40: lets add a commit on top, fixing our version?

vyzo (approved):

build/version.go
raulk: I assume this will be 1.16.0-dev, yes, @jennijuju? I'll punch that in. Otherwise we can adjust.

Issue #8625, opened 2022-05-10
 Builtin Actor Bundling Strategy


vyzo:
For NV16 and beyond, where we live in an fvm world, we need an effective mechanism for bundling and loading builtin actors manifests.

## State of Affairs
The current approach, in #8493, is to 
- embed the appropriate v8 bundle at build time, selected by build tags for the various networks
- have the fvm directly embed v7/v6 bundles.

The bundles are downloaded at build time, using the release tag from github.

This is problematic for two reasons:
- it doesn't scale as we move to future network versions, where we'd have to bundle builtin actors for every supported version
- it conflicts with the plan to move to a single binary for all networks (see #8158)
- it requires an unwanted dependency of fvm to actors, just to embed older network version support.
- it requires an ugly environment variable hack to load bundles for tests in DI.

## Dynamic Bundling Proposal
In order to address these issues and sever the dependency from fvm to the older actors bundles, we propose to dynamically fetch and load bundles at start up, using the release tag on github.

More specifically:
- embed the release tags for each builtin actors version (v7, v8, and beyond).
- when the lotus daemon starts, fetch the appropriate bundle (for the selected network) and persist in the `.lotus` repo, and then load into the blockstore once (not each and every time).
- pass the appropriate manifest to the fvm for v7, which is unnecessary for v8 as the manifest is in the system actor state to sever the dependency.

For tests, we propose to fetch the bundle in a fixed location (somewhere in /tmp) so that it can be cached from CI and reused for subsequent test runs.

One minor annoyance/difficulty is that the cli and lotus miner init also need to load the manifests so that they have access to the actor code CIDs; this currently happens using an in memory blockstore; this can continue, but instead of using embedded bundls, use the bundles as found in the repo and fetch dynamically as required.

raulk:
A few notes on requirements:

- There's already a similar precedent of the download-on-startup flow today: the proof parameters. We use https://github.com/filecoin-project/go-paramfetch/, and we should probably unify with that to avoid entropy in the codebase?
- We do need to provide an offline option. Not because daemons won't be having Internet connectivity (we can take that for granted, otherwise they wouldn't be able to participate in the network). However, some nodes in some parts of the world don't have access to GitHub (usually due to censorship). This was a problem with proofs, which have/had to be delivered through region-specific data centers (painful).
	- Should probably make the flow "download if not embedded" (if the VFS returns not found).
- Need to provide an option to test with a local bundle for development.
[...]

PR #8642, opened 2022-05-11
[nv16] dynamic loading of builtin actor bundles

vyzo:
See #8625 

to be resolved: properly getting repo path and network name for loading; see the TODOs; we could do nothing and keep it as is.


build/builtin_actors.go
```diff
@@ -0,0 +1,14 @@
+package build
+
+import (
+	"github.com/filecoin-project/lotus/chain/actors"
+)
+
+// TODO a nicer interface would be to embed (and parse) a toml file containing the releases
+var BuiltinActorReleases map[actors.Version]string
```
raulk, line 8: Yes, this would allow for quicker testing without having to recompile.

chain/actors/bundle.go
```diff
@@ -0,0 +1,151 @@
+package actors
```
raulk, line 1: nit: I'm not convinced that a component that's doing HTTP requests should be in this package. I think this belongs more in the node package.

chain/actors/bundle.go
```diff
@@ -0,0 +1,151 @@
...
+
+	resp, err := http.Get(url) //nolint
+	if err != nil {
+		return xerrors.Errorf("error fetching %s: %w", url, err)
+	}
+	defer resp.Body.Close() //nolint
+
+	if resp.StatusCode != http.StatusOK {
+		return xerrors.Errorf("error fetching %s: http response status is %d", url, resp.StatusCode)
+	}
```
raulk, line 80: Do we provide some retry logic? How would a failure to download be handled here?

chain/actors/bundle.go
```diff
@@ -0,0 +1,151 @@
...
+
+	if err := b.check(bundleBasePath, bundleFile, bundleHash); err != nil {
+		logb.Errorf("error checking bundle %s: %s", bundleName, err)
+		return "", xerrors.Errorf("error checking bundle: %s", err)
+	}
+
+	return bundleFilePath, nil
+}
+
+func (b *BundleFetcher) fetchURL(url, path string) error {
```
raulk, line 69: nit: would it make sense to print a progress bar with loggable format when not TTY? Although defer to Lotus maintainers for hints on preferred UX.

cli/init.go
```diff
@@ -12,11 +11,9 @@ import (
 func init() {
 	// preload manifest so that we have the correct code CID inventory for cli since that doesn't
 	// go through CI
-	if len(build.BuiltinActorsV8Bundle()) > 0 {
-		bs := blockstore.NewMemory()
+	bs := blockstore.NewMemory()
```
raulk, line 14: Is it truly OK to load the bundle into a memory blockstore?

node/modules/builtin_actors.go
```diff
@@ -12,28 +12,60 @@ import (
...
 	"github.com/filecoin-project/lotus/node/modules/helpers"
+	"github.com/filecoin-project/lotus/node/repo"
 
+	cid "github.com/ipfs/go-cid"
+	dstore "github.com/ipfs/go-datastore"
 	cbor "github.com/ipfs/go-ipld-cbor"
 )
 
-func LoadBultinActors(lc fx.Lifecycle, mctx helpers.MetricsCtx, bs dtypes.UniversalBlockstore) (result dtypes.BuiltinActorsLoaded, err error) {
+func LoadBultinActors(lc fx.Lifecycle, mctx helpers.MetricsCtx, r repo.LockedRepo, bs dtypes.UniversalBlockstore, ds dtypes.MetadataDS) (result dtypes.BuiltinActorsLoaded, err error) {
```
raulk, line 22:
```suggestion
func LoadBuiltinActors(lc fx.Lifecycle, mctx helpers.MetricsCtx, r repo.LockedRepo, bs dtypes.UniversalBlockstore, ds dtypes.MetadataDS) (result dtypes.BuiltinActorsLoaded, err error) {
```

raulk (commented):
Light review pass, approach so far LGTM.

build/builtin_actors.go
vyzo: actually, we would embed it so we'd still need a recompile.
It is just slightly nicer.

chain/actors/bundle.go
vyzo: agreed, i'll move it.

chain/actors/bundle.go
vyzo: no retry logic; we should probably retry a few times on tmp errors and 500s.
I will add something.

chain/actors/bundle.go
vyzo: ugh, thats hard to do actually!

cli/init.go
vyzo: here yes, we only need the code CID mapping for the cli process.

node/modules/builtin_actors.go
vyzo: ouch, typo. Cant just apply suggestion though, i also have to fix the override.

chain/actors/bundle.go
vyzo: done.

node/modules/builtin_actors.go
vyzo: fixed.

cli/init.go
raulk: Gotcha. Seems a bit wasteful, though. Can't we just load the root CID and decode the manifest? (Could do in a follow-up PR too)

build/builtin_actors.go
raulk: Duh, true. Would like to have that separate `bundles.spec.toml` file though.

node/bundle/bundle.go
```diff
@@ -49,47 +51,58 @@ func (b *BundleFetcher) Fetch(version Version, release, netw string) (path strin
...
+		defer f.Close() //nolint
 
-	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
-	if err != nil {
-		return xerrors.Errorf("error opening %s for writing: %w", path, err)
-	}
-	defer f.Close() //nolint
+		if _, err := io.Copy(f, resp.Body); err != nil {
+			return xerrors.Errorf("error writing %s: %w", path, err)
+		}
```
raulk, line 100: I wonder if we should add HTTP resumption support. More often than not what will happen is that the download gets interrupted due to a flaky connection. In that case, we don't want to truncate and restart the download from scratch, but rather resume it?

raulk: How does go-paramfetcher do it?

vyzo: matbe, but it shouldnt be much of a proble. It is a single 30M download.

We can consider in a follow up if it turns out to be a problem.

build/builtin_actors.go
vyzo: sure, any syntax suggestions?

cli/init.go
vyzo: the blockstore in thrown away, so it is just transient memory usage; below meh threshold ;)

chain/actors/bundle.go
raulk: https://pkg.go.dev/github.com/gosuri/uiprogress

cli/init.go
raulk: Loading 30MiB into a blockstore is going to add unnecessary latency to _every_, _single_, _CLI_, _command. We don't want that.

vyzo: ok, fine, but how fix that? I don't see any other way.

vyzo: plus we are talking a few milliseconds (maybe a whopping 50 of them), is it really such a big deal?

vyzo: adding a TODO for now.

build/builtin_actors.go
vyzo: ok, looks nice enough.

raulk: In terms of Go structs, I think this would look like (without the toml tags):

```go
type BundleSpec struct {
	Bundles []Bundle
}

type Bundle struct {
	Version actor.Version
	Release string
}
```

cli/init.go
arajasek: I don't see any quick improvements that can be made to avoid this. TODO and move on SGTM.

build/builtin_actors.go
vyzo: done.

build/bundle.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	_ "embed"
+
+	"github.com/filecoin-project/lotus/chain/actors"
+)
+
+var NetworkBundle string
+
+//go:embed bundle.toml
```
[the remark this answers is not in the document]
vyzo: ok

node/modules/builtin_actors.go
```diff
@@ -2,38 +2,69 @@ package modules
...
+
+		// ok, we don't have it -- fetch it and add it to the blockstore
+		mfCid, err := bundle.FetchAndLoadBundle(ctx, r.Path(), bs, av, rel, netw)
+		if err != nil {
 			return result, err
 		}
+
+		if err := ds.Put(ctx, key, mfCid.Bytes()); err != nil {
+			return result, xerrors.Errorf("error storing manifest CID for builtin-actors vrsion %d to the datastore: %w", av, err)
+		}
```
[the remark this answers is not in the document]
vyzo: 1. Not a problem, we check whether the manifest cid is in the blockstore and if not we refetch/reload.

So users are free to delete their blockstore and import a snapshot.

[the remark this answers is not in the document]
vyzo: 2. This is the real blockstore and badger sucks, so we'd keep bloating on every restart.

Granted, these may not be too frequent, but still there is no need to reimport with the logic above.

build/bundle.go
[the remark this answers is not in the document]
vyzo: done.

node/modules/builtin_actors.go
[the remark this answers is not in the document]
vyzo: Basically, for 2 this is exactly what we do: for every relevant bundle, we lookup the manifest cid in the datastore. 
If it is there, we then check the blockstore to make sure it is there.
If not, or if we don't have it in the datastore, then we just go ahead and fetch/load.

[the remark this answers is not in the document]
raulk: Ah got it. Admittedly it's a bit hard to follow what's happening here and why this is needed (some code comments would help). But this answer satisfies me.

I still prefer a version that imports the bundle over a blockstore abstraction that blocks writes for items that already exist. I think that would give you the same effect we're looking for here, with less moving parts, and being more fail-proof.

That said, if @arajasek is happy with this, I have no objections.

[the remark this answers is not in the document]
arajasek: I'm happy with this -- i like the idea of wiring it through a blockstore abstraction, but this gets the job done. Agreed that some comments would help, though, it takes a while to grok this.

arajasek (approved):

node/modules/builtin_actors.go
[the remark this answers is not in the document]
vyzo: added some comments so that the DI loader is easier to read and follow.

raulk:
The final item that was left here was to sever the dependency between filecoin-ffi and the builtin-actors bundle for v7. There are some difficulties in making that happen. @vyzo and I have been discussing offband, and I'm posting my notes here.

- builtin-actors v7 bundles are produced with synthetic code CIDs; that is, with CIDs of the kind `fil/7/storageminer` which do not content-hash over the bytecode.
- That makes the `car#LoadCAR()` content integrity checks fail.
- We could either (a) disarm those checks via an option, or (b) produce the bundle with real CIDs, import it to a memory blockstore in Lotus, and write them for real in the underlying blockstore with the synthetic Code CIDs so that ref-fvm is able to resolve them with the CID they are referenced with from the state tree.
- However, synthetic CIDs use Identity multihashes (i.e. multihashes where the data is inlined in the CID), and Lotus expands such CIDs through the [IDStore](https://github.com/filecoin-project/lotus/blob/dfb65ed89f67c7d4adc2cc1f00a10b2b345d9926/blockstore/idstore.go#L20), without querying the underlying blockstore.
- Therefore, we find that it's ultimately impossible to put the Wasm bytecode under identity multihashes in Lotus' blockstore.
- The potential solutions here are:
	1. Change the IDStore so that it does not expand inline CIDs when the codec is Raw.
	2. Wrap the store in a remapping blockstore that holds rewrite rules such that the synthetic v7 CodeCIDs are transformed to their respective content-addressed CodeCIDs before querying the blockstore.
	3. Do the rewrite fvm-side.

We don't like solution (1) because it introduces a system-wide change to address a very specific problem.
Solutions (2) and (3) require deeper changes and the tradeoff is not worth it (see next paragraph).

The realisation here is that supporting v7 bundles with FVM is only needed **_for testing_**, and only for the short time before nv16 is live. Once nv16 is live with actors v8, there is no reason we will ever want to run actors v7 with FVM. Hence, it does not make sense to implement a definitive solution for a situation that is only temporary.

In conclusion, we will keep the actors v7 dependency in filecoin-ffi and sever it once nv16 is live in mainnet and we no longer need to run canary tests.

raulk:
It is worth noting that the reason why actors v7 works in Rust is because the Rust library for loading CARs is not performing content integrity checks, and the blockstore is not short-circuiting lookups for identity hashes. So this might be a smell that those Rust counterparts are underdeveloped with respect to their Go equivalents, but that's another story.

PR #8653, opened 2022-05-13
[nv16] add support for v7 release bundles

vyzo:
So that we can finally sever the dependency between fvm and builtin-actors.


vyzo:
Closing -- see https://github.com/filecoin-project/lotus/pull/8642#issuecomment-1127546987

PR #8658, opened 2022-05-16
[nv16] niceties for development bundle loading

vyzo:
This adds some niceties for development bundle loading; see the README.

## Related Issues
See #8625

PR #8666, opened 2022-05-17
[nv16] flexible bundle loading

vyzo:
On top of #8658.

Adds flexibility for bundle fetching, which can use an env var or explicit URL to avoid downloading from github.

cc @travisperson

### DIFF
--- a/build/bundle.go
+++ b/build/bundle.go
@@ -21,13 +21,18 @@ type Bundle struct {
 	// Release is the release id
 	Release string
 	// Path is the (optional) bundle path; takes precedence over url
-	Path string
+	Path map[string]string
 	// URL is the (optional) bundle URL; takes precdence over github release
-	URL string
-	// CHecksum is the bundle sha256 checksume in hex, when specifying a URL.
-	Checksum string
+	URL map[string]BundleURL
 	// Devlopment indicates whether this is a development version; when set, in conjunction with path,
 	// it will always load the bundle to the blockstore, without recording the manifest CID in the
 	// datastore.
 	Development bool
+}
+
+type BundleURL struct {
+	// URL is the url of the bundle
+	URL string
+	// Checksum is the sha256 checksum of the bundle
+	Checksum string
 }
--- a/build/bundle.go
+++ b/build/bundle.go
@@ -20,8 +20,14 @@ type Bundle struct {
 	Version actors.Version
 	// Release is the release id
 	Release string
-	// Path is the (optional) bundle path; uses the appropriate release bundle if unset
+	// EnvVar is the (optional) env var specifying the bundle path; takes precdence over path
+	EnvVar string
+	// Path is the (optional) bundle path; takes precedence over url
 	Path string
+	// URL is the (optional) bundle URL; takes precdence over github release
+	URL string
+	// CHecksum is the bundle sha256 checksume in hex, when specifying a URL.
+	Checksum string
 	// Devlopment indicates whether this is a development version; when set, in conjunction with path,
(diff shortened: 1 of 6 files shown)


raulk (changes requested):
Having to modify your `bundle.toml` to specify an envvar which in turn is going to override any other setting seems like a circuitous way of doing things. It also defeats the one of the purposes, which is being able to check out the Lotus repo as-is (or download a pre-built binary) and make it use locally built bundles _instead_ without any changes in sources.

arajasek (approved):

vyzo:
ok fine.

raulk (changes requested):
Having to modify your `bundle.toml` to specify an envvar which in turn is going to override any other setting seems like a circuitous way of doing things. It also defeats the one of the purposes, which is being able to check out the Lotus repo as-is (or download a pre-built binary) and make it use locally built bundles _instead_ without any changes in sources.

node/bundle/manifest.go
```diff
@@ -86,13 +87,11 @@ func FetchAndLoadBundles(ctx context.Context, bs blockstore.Blockstore, bar map[
 	}
 
 	for av, bd := range bar {
+		envvar := fmt.Sprintf("LOGUS_BUILTIN_ACTORS_V%d_BUNDLE", av)
```
travisperson, line 90: Typo `LOGUS` => `LOTUS`

node/modules/builtin_actors.go
```diff
@@ -68,14 +68,11 @@ func LoadBuiltinActors(lc fx.Lifecycle, mctx helpers.MetricsCtx, r repo.LockedRe
 		}
 
 		// we haven't recorded it in the datastore, so we need to load it
+		envvar := fmt.Sprintf("LOGUS_BUILTIN_ACTORS_V%d_BUNDLE", av)
```
travisperson, line 71: Typo `LOGUS` => `LOTUS`

node/bundle/manifest.go
vyzo: aaaaargh, good catch!

vyzo: fixed.

node/modules/builtin_actors.go
vyzo: fixed.

build/bundle.go
```diff
@@ -21,13 +21,18 @@ type Bundle struct {
 	// Release is the release id
 	Release string
 	// Path is the (optional) bundle path; takes precedence over url
-	Path string
+	Path map[string]string
 	// URL is the (optional) bundle URL; takes precdence over github release
-	URL string
-	// CHecksum is the bundle sha256 checksume in hex, when specifying a URL.
-	Checksum string
+	URL map[string]BundleURL
```
raulk, line 26: I'd pluralize these fields.

raulk (approved):
All of this is growing organically at a very rapid rate and I think we're going to have to clean up some of this later, but it's not a big deal because this is all "private" and not exposed over APIs or intended for users to depend on.

build/bundle.go
vyzo: maybe, I kind of prefer singular here; but i don't feel strongly either way.

vyzo:
Yeah, let's get to its final "works really well" phase, and we can do some cleanup.

arajasek:
@raulk Strong agree, there is definitely a round of cleanup that needs to happen -- ultimately these bundles are gonna be "never changing" things (except for devs and network upgrades), so something much more static will work.

PR #8668, opened 2022-05-18
fix wdpost_dispute tests

arajasek:
With both https://github.com/filecoin-project/builtin-actors/pull/372 and https://github.com/filecoin-project/builtin-actors/pull/373, this should make tests green on `feat/nv16`!

### DIFF
--- a/build/bundles.toml
+++ b/build/bundles.toml
@@ -1,3 +1,3 @@
 [[bundles]]
 version = 8
-release = "dev/20220520"
+release = "dev/20220524"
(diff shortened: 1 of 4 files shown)


build/bundles.toml
```diff
@@ -1,3 +1,3 @@
 [[bundles]]
 version = 8
-release = "dev/20220517"
+release = "dev/20220517"
```
vyzo, line 3: this probably needs new release.

vyzo (commented):
test is still red, presumably we need a new actors release.

vyzo (approved):
go go go!

PR #8672, opened 2022-05-18
feat: worker: lotus-worker run --no-default

magik6k:
## Proposed Changes
lotus-worker has gained quite a lot of task-type flags, most of which are opt-out, which is a bit annoying, especially when creating storage-only workers. This PR adds a new flag which makes compute tasks opt-in instead of opt-out:

PreCommit1 worker before:
```
./lotus-worker run --precommit2=false --commit=false --prove-replica-update2=false --replica-update=false
```

PreCommit1 worker after:
```
./lotus-worker run --no-default --precommit1
```

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus-worker/main.go
+++ b/cmd/lotus-worker/main.go
@@ -189,7 +189,7 @@ var runCmd = &cli.Command{
 		},
 		&cli.BoolFlag{
 			Name:  "no-default",
-			Usage: "disable all default tasks, use the worker for storage only",
+			Usage: "disable all default compute tasks, use the worker for storage/fetching only",
 			Value: false,
 		},
 		&cli.IntFlag{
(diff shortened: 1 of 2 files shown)


RobQuistNL:
This is not going to work because if you use dedicated AP machines the PC1 will still need to do a Fetch - this will deadlock the job flow.

I've modified it before to always by default use Fetch (but excluded the other defaults from workers)

magik6k:
You should see fetch on workers with --no-default, we add it here: https://github.com/filecoin-project/lotus/pull/8672/files#diff-f711d6c89c998547913cc59960c1b32609adebd6800ff9b7589d5ac5bf459614R315

RobQuistNL:
Aaah I misread. This just disables all the extra flags after it. I thought you disabled all the other defaults too, my bad :+1:

geoff-vball (approved):

PR #8680, opened 2022-05-18
[nv16] Ensure v8 bundle is loaded during migration

vyzo:
Strictly speaking not necessary, but just to be safe on the defensive programming side.

Also removes a stale TODO comment from the beginning of time.

### DIFF
--- a/chain/consensus/filcns/upgrades.go
+++ b/chain/consensus/filcns/upgrades.go
@@ -44,6 +44,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/vm"
+	"github.com/filecoin-project/lotus/node/bundle"
 )
 
 func DefaultUpgradeSchedule() stmgr.UpgradeSchedule {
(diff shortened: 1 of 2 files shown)


chain/consensus/filcns/upgrades.go
```diff
@@ -1369,6 +1370,13 @@ func upgradeActorsV8Common(
 	buf := blockstore.NewTieredBstore(sm.ChainStore().StateBlockstore(), blockstore.NewMemorySync())
 	store := store.ActorStore(ctx, buf)
 
+	// ensure that the manifet is loaded in the blockstore
+	if err := bundle.FetchAndLoadBundles(ctx, buf, map[actors.Version]build.Bundle{
+		actors.Version8: build.BuiltinActorReleases[actors.Version8],
+	}); err != nil {
+		return cid.Undef, xerrors.Errorf("failed to load manifest bundle: %w", err)
+	}
+
```
raulk, line 1379: Honestly it would be nice to provide a nicer API for the bundle package. The `FetchAndLoadBundles` call here makes me uncomfortable because it can literally go out to the network to fetch a bundle, which will essentially derail the migration (which needs to happen in a few seconds).

Do we have a `bundle.GetManifest(actor.Version) (bundle.Manifest, cid.Cid, error)` call that we can use to get the manifest and its CID, and then call `buf.Has(cid.Cid)` on the manifest CID _and_ its every actor bytecode CID to decide if we need to activate the bundle fetch-and-load machinery?

The way that this `FetchAndLoadBundles` API works

vyzo: wait, the bundles should be there from last restart.
User must be actively stupid to end up with a network call.

vyzo: also, the bundles are fetched in the premigration too.

vyzo: we could add options to the call and instruct it to not make network calls and barf if not there if you think it will changr anything.

raulk: I don't think _ultimately_ it will change much. The reason for that is:

1. During pre-migration, we should be allowed to make network calls -- code today is good.
2. During the migration, we shouldn't be making network calls. However, if we prohibit them and the bundle is not there, the migration will fail entirely. If we allow them, the node will miss the epoch almost certainly, but at least will be able to catch up later. So perhaps it's OK for the migration to be capable to make a network call, but there's something about it that makes me cringe about the design, which I don't think we can resolve without a complete refactor, so not for now.

raulk (approved):

arajasek (approved):

PR #8707, opened 2022-05-23
fix: test: correctly check error in window post test

Stebalien:


Stebalien:
Note: the _other_ test will still fail, because of the bug in the FVM that's currently sitting in master.

PR #8734, opened 2022-05-26
fvm: fail if implicit message fails

### LINKED ISSUE #8463: FVM: Implicit messages should fail on non-zero exit codes
We currently handle FVM implicit messages the same way we do regular messages -- if the FVM returned an error, we do so, otherwise we successfully return (with a non-zero exitcode).

This doesn't match LegacyVM behaviour, which fails if cron / rewards messages fail in _any_ way. We should match that behaviour.

arajasek:
## Related Issues

Resolves #8463

## Proposed Changes

This is not strictly needed, since all existing callsites of `ApplyImplicitMessage` check for a non-zero exit code and fail anyway. But nice to make it _explicit_!

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/vm/fvm.go
+++ b/chain/vm/fvm.go
@@ -3,6 +3,7 @@ package vm
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


chain/vm/fvm.go
```diff
@@ -413,12 +414,18 @@ func (vm *FVM) ApplyImplicitMessage(ctx context.Context, cmsg *types.Message) (*
...
+	applyRet := &ApplyRet{
 		MessageReceipt: receipt,
 		ActorErr:       aerr,
 		ExecutionTrace: et,
 		Duration:       duration,
-	}, nil
+	}
+
+	if ret.ExitCode != 0 {
+		return applyRet, fmt.Errorf("implicit message failed with exit code: %d", ret.ExitCode)
```
magik6k, line 425: Also print `ret.ActorErr`?

magik6k (approved):
1 thing to address

PR #8751, opened 2022-05-30
fix: lotus-bench make command

sunday527:
Solved [Issues #8748 ](https://github.com/filecoin-project/lotus/issues/8748)


magik6k (approved):
Thanks for the PR!

I don't think this changes anything other than the version string, and we will soon move to single binary for multiple networks, but for now this is slightly more correct.

PR #8765, opened 2022-05-31
set the PreCommit1 task num

peter198:
Introduction:
Group task scheduling optimization, adding a limit on the number of PreCommit1 tasks for workers

Statement of needs:
In the original scheduling, there is no limit to the number of workers running PreCommit1 at the same time, and the maximum task configuration size includes the number of CPU cores, available memory, and hard disk space.
There is no reasonable limit. Although workers can execute when they receive too many tasks, they cannot fully exert their computing performance. The parallel PreCommit1 tasks can be reasonably controlled according to the hardware configuration.
Can improve the overall task completion speed.

improve proposals:
When the worker starts, the --parallel-p1-limit parameter sets the maximum PreCommit1 task it can accept. It will be recorded when the miner schedules the PreCommit1 task to the worker
Quantity allocated to it When the number of tasks already allocated reaches the set maximum number of tasks, it will not be allocated to a new task until 1 task is completed, and the allocation will be submitted in advance.

The default value of the --parallel-p1-limit parameter is -1, which means no limit, and if --parallel-p1-limit--parallel-p1-limit--parallel-p1-limit exceeds all the values ​​​​of the hardware configuration of the on-duty worker maximum reception time of
won't work either.
When viewing worker information through lotus-miner sealed workers, the task number information of PreCommit1 will be displayed, as follows:
p1_limit: 15-7, 15 represents the value set by --parallel-p1-limit, and 7 represents the number of PreCommit1 tasks currently running.

### DIFF
--- a/cmd/lotus-miner/run.go
+++ b/cmd/lotus-miner/run.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	_ "net/http/pprof"
 	"os"
+	"strconv"
 
 	"github.com/filecoin-project/lotus/api/v1api"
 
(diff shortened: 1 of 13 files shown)


peter198:
add sealing.go change

PR #8769, opened 2022-05-31
feat: only enable rcmgr by default in full nodes

vyzo:
opt-in with envvar for other nodes, as there are still some issues with markets related protocols

### DIFF
--- a/node/modules/lp2p/rcmgr.go
+++ b/node/modules/lp2p/rcmgr.go
@@ -27,8 +27,10 @@ import (
 
 func ResourceManager(connMgrHi uint) func(lc fx.Lifecycle, repo repo.LockedRepo) (network.ResourceManager, error) {
 	return func(lc fx.Lifecycle, repo repo.LockedRepo) (network.ResourceManager, error) {
+		isFullNode := repo.RepoType().Type() == "FullNode"
 		envvar := os.Getenv("LOTUS_RCMGR")
-		if envvar == "0" {
+		if (isFullNode && envvar == "0") || // enable by default for full nodes
+			(!isFullNode && envvar != "1") { // otherwise, opt-in with envvar
 			// this is enabled by default; specify LOTUS_RCMGR=0 to disable
 			log.Info("libp2p resource manager is disabled")
 			return network.NullResourceManager, nil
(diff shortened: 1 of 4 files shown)


vyzo:
Needs backport to nv16 branch for v1.16.0 release.

node/modules/lp2p/rcmgr.go
```diff
@@ -27,9 +27,10 @@ import (
 
 func ResourceManager(connMgrHi uint) func(lc fx.Lifecycle, repo repo.LockedRepo) (network.ResourceManager, error) {
 	return func(lc fx.Lifecycle, repo repo.LockedRepo) (network.ResourceManager, error) {
+		isFullNode := repo.RepoType().Type() == "FullNode"
 		envvar := os.Getenv("LOTUS_RCMGR")
-		if envvar == "0" {
-			// this is enabled by default; specify LOTUS_RCMGR=0 to disable
+		if (isFullNode && envvar == "0") || // enable by default for full nodes
```
ribasushi, line 32: Logic seems wrong, I think it should be:
```suggestion
		if (isFullNode && envvar != "0") || // enable by default for full nodes
```

vyzo: we might want to fix the comment to avoid tripping casual reading.
Any suggestions?

ribasushi, line 32:
```suggestion
		if (isFullNode && envvar == "0") || // only set NullResourceManager if envvar is explicitly "0"
```

node/modules/lp2p/rcmgr.go
```diff
@@ -27,9 +27,10 @@ import (
...
+			(!isFullNode && envvar != "1") { // otherwise, opt-in with envvar
```
ribasushi, line 33:
```suggestion
			(!isFullNode && envvar != "1") { // set NullResourceManager *unless* envvar is explicitly "1"
```

ribasushi: @vyzo ☝️ maybe

vyzo: thanks, applied.

magik6k (approved):

Issue #8711, opened 2022-05-23
lotus-shed: hello protocol dump


travisperson:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

I would like to be able to query remote peers for their hello protocol data.

https://github.com/filecoin-project/lotus/blob/518dc962ee52a31bce3eecb4f525d4ed91e029e5/node/hello/hello.go#L33-L38
https://github.com/filecoin-project/lotus/blob/v1.15.2/node/hello/hello.go#L33-L38

This will make it easier to probe remote peers and inspect their heaviest tipset, and find what genesis block they are using.

### Describe the solution you'd like

Add a command to `lotus-shed` accepts a multiaddr as an argument and a path to a genesis car file. The command should run the hello protocol against the remote peer, using the genesis data for itself, and then dumps the `HelloMessage` from the peer to stdout as formatted json (such that the output of the command can be used with `jq`.

To the remote peer this should look identical to a new node joining the network with no chain data synced.

This will require setting up a libp2p node, a random PeerID should be fine. Nothing should be written to disk.

rjan90:
Hey! I´m closing this issue now since the desired feature has been implemented and merged 😄 Thanks @swift-mx for making this!

PR #8787, opened 2022-06-02
feat: lotus-shed get remote peer hello message

swift-mx:
## Related Issues
https://github.com/filecoin-project/lotus/issues/8711

## Additional Info

image

## Checklist

- [x] CI is green

### DIFF
--- a/cmd/lotus-shed/hello.go
+++ b/cmd/lotus-shed/hello.go
@@ -31,11 +31,11 @@ var helloCmd = &cli.Command{
 		if err != nil {
 			return err
 		}
+		h.SetStreamHandler(hello.ProtocolID, HandleStream)
 		err = h.Connect(ctx, pis[0])
 		if err != nil {
 			return err
 		}
-		h.SetStreamHandler(hello.ProtocolID, HandleStream)
 		ctx, done := context.WithTimeout(ctx, 5*time.Second)
 		defer done()
 		select {
--- a/cmd/lotus-shed/hello.go
+++ b/cmd/lotus-shed/hello.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"time"
+
 	"github.com/filecoin-project/go-address"
 	cborutil "github.com/filecoin-project/go-cbor-util"
 	"github.com/filecoin-project/lotus/blockstore"
@@ -22,8 +25,6 @@ import (
 	"github.com/multiformats/go-multiaddr"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/xerrors"
-	"os"
-	"time"
 )
 
 var helloCmd = &cli.Command{
@@ -63,7 +64,9 @@ var helloCmd = &cli.Command{
 			return err
 		}
 		h, err := libp2p.New(libp2p.Identity(priv))
-
+		if err != nil {
+			return err
+		}
 		err = h.Connect(ctx, pis[0])
 		if err != nil {
 			return err
--- a/cmd/lotus-shed/hello.go
+++ b/cmd/lotus-shed/hello.go
@@ -4,22 +4,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"os"
 	"time"
 
 	cborutil "github.com/filecoin-project/go-cbor-util"
-	"github.com/filecoin-project/lotus/blockstore"
-	"github.com/filecoin-project/lotus/chain/consensus/filcns"
-	"github.com/filecoin-project/lotus/chain/store"
 	lcli "github.com/filecoin-project/lotus/cli"
 	"github.com/filecoin-project/lotus/node/hello"
-	"github.com/ipfs/go-datastore"
 	"github.com/libp2p/go-libp2p"
-	"github.com/libp2p/go-libp2p-core/host"
 	inet "github.com/libp2p/go-libp2p-core/network"
-	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/urfave/cli/v2"
-	"golang.org/x/xerrors"
 )
 
 var resultCh chan bool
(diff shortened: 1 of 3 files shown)


cmd/lotus-shed/hello.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+		pis, err := addrInfoFromArg(ctx, cctx)
+		if err != nil {
+			return err
+		}
+		priv, _, err := crypto.GenerateKeyPair(crypto.RSA, 2048)
+		if err != nil {
+			return err
+		}
+		h, err := libp2p.New(libp2p.Identity(priv))
```
magik6k, line 66: Shouldn't the default work?
```suggestion
		h, err := libp2p.New()
```

cmd/lotus-shed/hello.go
```diff
@@ -0,0 +1,164 @@
...
+		err = SayHello(ctx, pis[0].ID, h, hmsg)
+		if err != nil {
+			return err
+		}
+		time.Sleep(time.Second * 5)
+		return nil
+	},
+}
+
+func addrInfoFromArg(ctx context.Context, cctx *cli.Context) ([]peer.AddrInfo, error) {
```
magik6k, line 95: We should just make this function public in the CLI package and call it instead of duplicating code - https://github.com/filecoin-project/lotus/blob/master/cli/net.go#L293

cmd/lotus-shed/hello.go
```diff
@@ -0,0 +1,164 @@
...
+		hmsg := &hello.HelloMessage{
+			HeaviestTipSet:       ts.Cids(),
+			HeaviestTipSetHeight: ts.Height(),
+			HeaviestTipSetWeight: weight,
+			GenesisHash:          gen.Cid(),
+		}
+		err = SayHello(ctx, pis[0].ID, h, hmsg)
+		if err != nil {
+			return err
+		}
```
magik6k, line 89: We shouldn't be sending the hello message to the peer we're testing; Sending that message means that we say that we are a filecoin node; It shouldn't break anything, but it's nicer if we don't do that.

The peer will say hello after getting list of our protocols, so this tool just needs to connect and wait.

https://github.com/filecoin-project/lotus/blob/5416ce54431fff8333b6ef80e86ac1194ef23577/node/modules/services.go#L71

cmd/lotus-shed/hello.go
```diff
@@ -0,0 +1,164 @@
...
+		time.Sleep(time.Second * 5)
```
magik6k, line 90: Ideally we'd pass a channel into handleStream, than wait on it here in a select with some context+timeout. This way the command will exit sooner if we get the response early.

Also would be nice to have a flag for this delay, going across some networks may be slower than 5s (but 5s is a good default)

cmd/lotus-shed/hello.go
magik6k: Without that we'll also be able to drop all the genesis loading above

magik6k (commented):
Thanks for the PR!

A few suggestions.

cmd/lotus-shed/hello.go
swift-mx: > We should just make this function public in the CLI package and call it instead of duplicating code - https://github.com/filecoin-project/lotus/blob/master/cli/net.go#L293

I did this at first, but I was worried that I should not modify lotus. I will submitted it soon

cmd/lotus-shed/hello.go
swift-mx: > Shouldn't the default work?

I agree , but the default is the same

cmd/lotus-shed/hello.go
magik6k: You can combine the channel with a timeout context, something like:

```
ctx, done := context.WithTimeout(5 * time.Second)
defer done()

select {
case result := <-resultCh:
 [print result]
case <-ctx.Done():
 [print timeout error]
}

```

cmd/lotus-shed/hello.go
swift-mx: > Without that we'll also be able to drop all the genesis loading above

May be it can get result faster?

cmd/lotus-shed/hello.go
swift-mx: > You can combine the channel with a timeout context, something like:
> 
> ```
> ctx, done := context.WithTimeout(5 * time.Second)
> defer done()
> 
> select {
> case result := <-resultCh:
>  [print result]
> case <-ctx.Done():
>  [print timeout error]
> }
> ```

OK

cmd/lotus-shed/hello.go
magik6k: It shouldn't have any impact, unless maybe libp2p is doing some fancy lazy stream negotiation under the hood.

If it is faster to send something, we should be sending a libp2p ping (like here https://github.com/filecoin-project/lotus/blob/master/node/impl/net/net.go#L184), instead of the hello message.

cmd/lotus-shed/hello.go
```diff
@@ -0,0 +1,120 @@
...
+			HeaviestTipSet:       ts.Cids(),
+			HeaviestTipSetHeight: ts.Height(),
+			HeaviestTipSetWeight: weight,
+			GenesisHash:          gen.Cid(),
+		}
+		err = SayHello(ctx, pis[0].ID, h, hmsg)
+		if err != nil {
+			return err
+		}
+		ctx, done := context.WithTimeout(context.Background(), 5*time.Second)
```
magik6k, line 84: Why not use the cli context as base here?
```suggestion
		ctx, done := context.WithTimeout(ctx, 5*time.Second)
```

cmd/lotus-shed/hello.go
swift-mx: > It shouldn't have any impact, unless maybe libp2p is doing some fancy lazy stream negotiation under the hood.
> 
> If it is faster to send something, we should be sending a libp2p ping (like here https://github.com/filecoin-project/lotus/blob/master/node/impl/net/net.go#L184), instead of the hello message.

It may be have enought time to get result ,I agree with your suggestion

swift-mx:
@magik6k  All suggestions fix , please review again , thks

swift-mx:
new command is : lotus-shed hello [multiaddr]
image

cmd/lotus-shed/hello.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		h, err := libp2p.New()
+		if err != nil {
+			return err
+		}
+		err = h.Connect(ctx, pis[0])
+		if err != nil {
+			return err
+		}
+		h.SetStreamHandler(hello.ProtocolID, HandleStream)
```
magik6k, line 38: We should set the stream handler before connecting, if we don't do that, there's a small race that can occur:
* We connect to the peer
* They get list of our protocols, and don't see hello, so they don't send it
* We register the hello protocol, and wait for it
* It never comes because the peer thinks we don't support hello

magik6k (commented):
One last thing, other than that it looks good.

cmd/lotus-shed/hello.go
swift-mx: You are right , sorry for my carelessness

magik6k (approved):

PR #8838, opened 2022-06-10
feat: refactor: actor bundling system

### LINKED ISSUE #8701: Embed the bundles during make
So that we dont depend on loading bundles on runtime

Stebalien:
## Related Issues

fixes #8701

## Proposed Changes

1. Include the builtin-actors in the lotus source tree.
2. Embed the bundle on build instead of downloading at runtime.
3. Avoid reading the bundle whenever possible by including bundle metadata (the bundle CID, the actor CIDs, etc.).
4. Remove everything related to dependency injection.
    1. We're no longer downloading the bundle, so doing anything ahead of time doesn't really help.
    2. We register the manifests on init because, unfortunately, they're global.
    3. We explicitly load the current actors bundle in the genesis state-tree method.
    4. For testing, we just change the in-use bundle with a bit of a hack. It's not great, but using dependency injection doesn't make any sense either because, again, the manifest information is global.
5. Remove the `bundle.toml` file. Bundles may be overridden by specifying an override path in the parameters file, or an environment variable.

## Checklist

- [x] CI is green

## TODO

- [x] Add a test to make sure that the generated bundle metadata matches the embedded bundle.
- [ ] Test to make sure that loading from paths still works correctly.

Maybes: 

- [ ] Cleanup/improve the download script? Maybe add a make target?
- ~Merge some of the methods in `chain/actors/builtin/builtin.go` with `chain/actors/bundle.go`, because they're basically the same.~


build/params_2k.go
```diff
@@ -53,8 +52,6 @@ var UpgradeChocolateHeight = abi.ChainEpoch(-17)
 var UpgradeOhSnapHeight = abi.ChainEpoch(-18)
 var UpgradeSkyrHeight = abi.ChainEpoch(100)
 
-var ActorsCIDs = map[actors.Version]cid.Cid{}
```
Stebalien, line 56: We're now including this information in the `build/builtin_actors_gen.go` file. Unfortunately, we need the ability to _switch_ bundles at runtime (for testing) so we can't actually guarantee a single CID.

chain/actors/builtin/builtin.go
```diff
@@ -65,9 +67,8 @@ func QAPowerForWeight(size abi.SectorSize, duration abi.ChainEpoch, dealWeight,
 }
 
 func ActorNameByCode(c cid.Cid) string {
```
Stebalien, line 69: We really should be merging this code with the bundle code. Exporting two sets of methods for looking up actor names where one only works after v8 is strange. But I'm planning on leaving that to a different PR.

chain/consensus/filcns/upgrades.go
```diff
@@ -1395,12 +1393,6 @@ func upgradeActorsV8Common(
 		return cid.Undef, xerrors.Errorf("no manifest CID for v8 upgrade")
 	}
 
-	if val, ok := build.ActorsCIDs[actors.Version8]; ok {
```
Stebalien, line 1398: We check that the manifest matches the expected CID when we load the bundle.

cmd/lotus-miner/init.go
```diff
@@ -218,13 +216,6 @@ var initCmd = &cli.Command{
 				return err
 			}
 
-			// load bundles
-			bs := blockstore.NewMemory()
-
-			if err := bundle.FetchAndLoadBundles(ctx, bs, build.BuiltinActorReleases); err != nil {
-				return err
-			}
```
Stebalien, line 226: We used to use this for the side-effect of initializing the registry in `chain/actors`, but that's now always initialized on init.

cmd/lotus-seed/genesis.go
```diff
@@ -580,12 +578,7 @@ var genesisCarCmd = &cli.Command{
 		bstor := blockstore.WrapIDStore(blockstore.NewMemorySync())
 		sbldr := vm.Syscalls(ffiwrapper.ProofVerifier)
 
-		// load appropriate bundles
-		if err := bundle.FetchAndLoadBundles(c.Context, bstor, build.BuiltinActorReleases); err != nil {
```
Stebalien, line 584: `MakeInitialStateTree` now loads the actors bundle, if necessary, for us.

itests/kit/ensemble.go
```diff
@@ -136,6 +136,13 @@ type Ensemble struct {
 
 // NewEnsemble instantiates a new blank Ensemble.
 func NewEnsemble(t *testing.T, opts ...EnsembleOpt) *Ensemble {
+	// Ensure we're using the right actors. This really shouldn't be some global thing, but it's the best we can do for now.
+	if build.InsecurePoStValidation {
+		require.NoError(t, build.UseNetworkBundle("testing-fake-proofs"))
+	} else {
+		require.NoError(t, build.UseNetworkBundle("testing"))
+	}
```
Stebalien, line 144: We used to do this through dependency injection in the test node, but we're setting a global so we might as well do it here.

lotuspond/front/src/chain/methods.json
```diff
@@ -732,21 +732,40 @@
     "RemoveVerifiedClientDataCap"
   ],
   "fil/8/account": [
+    "Send",
+    "Constructor",
+    "PubkeyAddress",
```
Stebalien, line 737: This is because we have two registered copies of the actors. @vyzo why not register the code under one CID here:

https://github.com/filecoin-project/lotus/blob/5bb50289bd3db33e1de9b1474e79ffcdae66dcf7/chain/vm/invoker.go#L167-L170

This isn't very enlightening:

https://github.com/filecoin-project/lotus/blob/5bb50289bd3db33e1de9b1474e79ffcdae66dcf7/chain/vm/invoker.go#L112

node/builder_miner.go
```diff
@@ -51,10 +51,6 @@ var MinerNode = Options(
 
 	// Mining / proving
 	Override(new(*storage.AddressSelector), modules.AddressSelector(nil)),
-
-	// builtin actors manifest
-	Override(new(dtypes.BuiltinActorsLoaded), modules.LoadBuiltinActors),
-	Override(new(dtypes.UniversalBlockstore), modules.MemoryBlockstore),
```
Stebalien, line 57: Didn't this cause problems?

paychmgr/paychget_test.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/filecoin-project/lotus/chain/actors/builtin/paych"
 	paychmock "github.com/filecoin-project/lotus/chain/actors/builtin/paych/mock"
 	"github.com/filecoin-project/lotus/chain/types"
+
```
Stebalien, line 27: Unfortunately, the bundles only get registered if we pull in the build package and the build package already depends on the actors package.

I may be able to refactor some things to avoid this issue, but it's probably best to leave it up to a followup patch.

Stebalien:
The docs-check and gen-check are conflicting and I'm not sure how to get them both happy.

paychmgr/paychget_test.go
Stebalien: Hm. No, that wasn't the issue. Fixed.

lotuspond/front/src/chain/methods.json
Stebalien: Removing the duplicate definitions seems to have worked.

arajasek (commented):
Direction generally LGTM.

build/params_debug.go
```diff
@@ -4,6 +4,9 @@
 package build
 
 func init() {
+	// TODO: This doesn't enable insecure validation in the actors, which is likely a problem.
```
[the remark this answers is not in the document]
arajasek: > Then should we not be using the special FVM bundle that disables window post checking, etc?

So it breaks down into 2 cases: itests, and testnets. 

For itests, we **should** use the special faked FVM bundle if `mockProofs` is set. We are currently using the special bundle for _all_ itests, because we use the special bundle if InsecurePoStValidation is true, and we set InsecurePoStValidation to true for (almost) all itests. This isn't strictly correct, but...fine.

 For testnets, we should **definitely** _not_ use the special bundle if InsecurePoStValidation isn't set (we currently don't), and we should **probably** not use it even if InsecurePoStValidation is set (we currently do). That's cuz the Lotus miner will compute real Window PoSts even if InsecurePoStValidation is set, and Idk what that will do to the faked proofs bundle. The good news is that's what we currently do (we never use the fake-proofs bundle on testnets).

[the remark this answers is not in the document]
Stebalien: Ok, so, this is all broken.

1. We currently globally set `build.InsecurePoStValidation = true` for all tests. We then try to explicitly set it for some tests, but we've already set it (https://github.com/filecoin-project/lotus/blob/54a409777ba0a58d5c1cfa9468fd7ca4a171bc08/itests/kit/init.go#L23) so that doesn't matter.
2. We don't even look at `mockProofs`.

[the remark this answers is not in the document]
Stebalien: Fixed:

1. I'm no longer explicitly setting InsecurePoStValidation inside itests.
2. I'm deciding which post validation to use based on the mockProofs option, not InsecurePoStValidation.

build/builtin_actors.go
```diff
@@ -1,27 +1,263 @@
...
+	"golang.org/x/xerrors"
 )
 
-var BuiltinActorReleases map[actors.Version]Bundle
+//go:embed actors/*.tar.zst
+var embeddedBuiltinActorReleases embed.FS
 
 func init() {
-	BuiltinActorReleases = make(map[actors.Version]Bundle)
+	if BundleOverrides == nil {
```
arajasek, line 30: Just for my understanding, this still runs on every invocation of the Lotus CLI (and shed and others), yes? But is fairly quick if you're just using the embedded bundle?

arajasek (approved):
One question, but this seems like a major improvement!

build/builtin_actors_gen.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"github.com/ipfs/go-cid"
+)
+
+var EmbeddedBuiltinActorsMetadata []*BuiltinActorsMetadata = []*BuiltinActorsMetadata{{
+	Network:     "butterflynet",
+	Version:     8,
+	ManifestCid: MustParseCid("bafy2bzacedktlamzlgecxpnsnods5iyh5uspbg2ukqrrn6h6c2h5iozmtb2d6"),
+	Actors: map[string]cid.Cid{
```
simlecode, line 13: @Stebalien 
I found that the generated actors are different from those described by jennijuju [said](https://github.com/filecoin-project/lotus/discussions/8347#discussioncomment-2853132)
```
Network Version: 16
Actor Version: 8
Manifest CID: bafy2bzacedy4qgxbr6pbyfgcp7s7bdkc2whi5errnw67al5e2tk75j46iucv6

Actor             CID  
account           bafk2bzacebokyrxttmnta57zqv4wmuhitxvois4ms3supyxvebmwnmohcgtxi
cron              bafk2bzacebx4e75g26a74dks5cvoxp26d3ds7ecxhukgrzva6kbfn62fg3whk
init              bafk2bzacecqyh33jcmo6cgfy5ggbg4f4er62cgbowazlicfvxpcw5pisq75pc
storagemarket     bafk2bzacecpjcy5eezq32dcsje5dbiwbbmcxv2cmjhifmkukgzufl62yxkghs
storageminer      bafk2bzaceciht74rpjtr3mdalubcfjzo5n6733i4mmxj2hs5e7caowjdnijiw
multisig          bafk2bzacecg6w7xycxvisxyolz3pcbdhvwp7nxqmvtldncs3vi5fvmgsv2s5u
paymentchannel    bafk2bzacedbg6s7eii6lqqn67za5if6dtebncxanv2ammjd7fb2prf4bbeuxi
storagepower      bafk2bzaceccsmebstaf5mye6ybyssp3svdmvwupk6prjaj2hat2goimmtonnm
reward            bafk2bzacecqrqytydwusmricocq7ix574ujkeqzay5wf2aawmybcufodmyhai
system            bafk2bzacec35iwbebnacuaqszqgdbkkhmniatonaeezkl3kekm57usr7ywbme
verifiedregistry  bafk2bzacecxde6ynf6jslzgfy4q4yvypmzy3bu7vvxgcesqejxmktr5766bza
```

arajasek: @simlecode Yes, that's right, they are slightly newer actors (and not the ones currently on butterflynet). We haven't fully frozen v8 builtin actors yet, so it might change 1-2 more times before the final release.

build/actors/README.md
```diff
@@ -0,0 +1,18 @@
+# Bundles
+
+This directory includes the actors bundles for each release. Each actor bundle is a zstd compressed
+tarfile containing one bundle per network type. These tarfiles are subsequently embedded in the
+lotus binary.
+
+## Updating
+
+To update, run the `./pack.sh` script. For example, the following will pack the builtin actors release `dev/20220602` into the `v8` tarfile.
```
jennijuju, line 9:
```suggestion
To update, run the `./pack.sh` script. For example, the following will pack the [builtin actors release](https://github.com/filecoin-project/builtin-actors/releases) `dev/20220602` into the `v8` tarfile.
```

build/actors/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+```bash
+./pack.sh v8 dev/20220602
+```
+
+This will:
+
+1. Download the actors bundles and pack them into the appropriate tarfile.
+2. Run `make bundle-gen` in the top-level directory to regenerate the bundle metadata file.
```
jennijuju, line 18: can one specify two versions at one time?

build/params_mainnet.go
```diff
@@ -23,6 +21,7 @@ var DrandSchedule = map[abi.ChainEpoch]DrandEnum{
 }
 
 var NetworkBundle = "mainnet"
+var BundleOverrides map[actors.Version]string
```
arajasek, line 24: @jennijuju is concerned about users misusing this and losing consensus -- can we add a comment on this saying "DON'T SET THIS unless you know what you're doing, you will lose consensus with the network!"

build/builtin_actors.go
Stebalien: > Just for my understanding, this still runs on every invocation of the Lotus CLI

Yes.

> But is fairly quick if you're just using the embedded bundle

Unless you specify bundle overrides, it doesn't even load the bundles. In that case, it skips:

https://github.com/filecoin-project/lotus/blob/f01bd41cc811beadd2c6d974d0219ce06307d67c/build/builtin_actors.go#L61-L73

And runs:

https://github.com/filecoin-project/lotus/blob/f01bd41cc811beadd2c6d974d0219ce06307d67c/build/builtin_actors.go#L76-L90

Which uses the pre-extracted bundle metadata without actually loading the bundle itself.

build/actors/README.md
Stebalien: No, you'd have to run the command for each version/release you want to download.

Stebalien: Oh, I see. `make bundle-gen` will generate the metadata for all versions.

PR #8841, opened 2022-06-10
PoC: FVM Debug Dual Execution

vyzo:
See https://github.com/filecoin-project/ref-fvm/issues/592

Depends on:
- [x] https://github.com/filecoin-project/ref-fvm/pull/605
- [x] https://github.com/filecoin-project/filecoin-ffi/pull/288

## How it works
When  `LOTUS_FVM_DEBUG=1`, dual execution is triggered with actor debugging for side effect. If also `LOTUS_FVM_DEBUG_BUNDLE_V8` is also specified, then the bundle is loaded and execution is redirected from the canonical (consensus) actors to the actors in thr bundle during debug execution.

## Some niceties to consider
- Cache at init the debug bundle in an in memory overlay blockstore and reuse it in each debug execution. This will eliminate load iverhead.
- Add an `output` syscall to fvm that captures output during debug execution and is noop otherwise.


vyzo:
CI will have to rerun once the ffi helper publishes the relevant build artifacts (not there yet, have been testing locally with a devnet)

chain/vm/fvm.go
```diff
@@ -308,6 +315,177 @@ func NewFVM(ctx context.Context, opts *VMOpts) (*FVM, error) {
...
+		},
+		Epoch:          opts.Epoch,
+		BaseFee:        opts.BaseFee,
+		BaseCircSupply: circToReport,
+		NetworkVersion: opts.NetworkVersion,
+		StateBase:      opts.StateBase,
+		Tracing:        EnableDetailedTracing,
+	}
+
+	fvmopts.Debug = true
```
arajasek, line 350: This can just go into the declaration above.

chain/vm/fvm.go
```diff
@@ -308,6 +315,177 @@ func NewFVM(ctx context.Context, opts *VMOpts) (*FVM, error) {
...
+
+	// TODO move these two util functions somewhere more general
```
arajasek, line 352: Yes, definitely. If you can, take a look at code blocks [here](https://github.com/filecoin-project/lotus/blob/30981d0fdddac09b7a3ebb7d96e5a911b01eb997/chain/consensus/filcns/upgrades.go#L1483-L1508) and [here](https://github.com/filecoin-project/lotus/blob/30981d0fdddac09b7a3ebb7d96e5a911b01eb997/build/builtin_actors.go#L192-L229). There is likely some de-duping that can happen, I think the right place for this would be `chain/actors/manifest.go`, probably.

chain/vm/fvm.go
```diff
@@ -308,6 +315,177 @@ func NewFVM(ctx context.Context, opts *VMOpts) (*FVM, error) {
...
+		cborStore := cbor.NewCborStore(overlayBstore)
+		mappingCid, err := cborStore.Put(context.TODO(), &mapping)
+		if err != nil {
+			return cid.Undef, err
+		}
+
+		return mappingCid, nil
+	}
+
+	switch opts.NetworkVersion {
```
arajasek, line 410: This is technically wrong -- we should decide based on actors version, not network version. You can use [this method](https://github.com/filecoin-project/lotus/blob/3370847f3e98ed1c4c6c247a01db863b5bddd4a9/chain/actors/version.go#L41).

chain/vm/fvm.go
```diff
@@ -308,6 +315,177 @@ func NewFVM(ctx context.Context, opts *VMOpts) (*FVM, error) {
...
+			if err != nil {
+				return nil, err
+			}
+
+			mf, err := loadManifest(mfCid)
+			if err != nil {
+				return nil, err
+			}
+
+			// create actor redirect mapping from the synthetic Cid to the debug code
```
arajasek, line 423: Let's not rebuild these CIDs here. The easiest way to avoid doing so would be to create a name -> synthetic CID map, and then just use that to build the `actorRedirect` map.

```
	var fromMap = map[string]cid.Cid{
		"init":             builtin7.InitActorCodeID,
		"cron":             builtin7.CronActorCodeID,
		"account":          builtin7.AccountActorCodeID,
		"storagepower":     builtin7.StoragePowerActorCodeID,
		"storageminer":     builtin7.StorageMinerActorCodeID,
		"paymentchannel":   builtin7.PaymentChannelActorCodeID,
		"multisig":         builtin7.MultisigActorCodeID,
		"reward":           builtin7.RewardActorCodeID,
		"verifiedregistry": builtin7.VerifiedRegistryActorCodeID,
	}
```

There's probably a way to do it using the information in the `ActorRegistry` and other places, but I don't immediately see it.

arajasek (commented):
Thank you for taking this on! As a PoC, I think this looks good. I think using envvars to dictate where to look for potential overrides is good enough.

The biggest thing I'd like to see is some way to automatically make this happen for select state computations (eg. when estimating gas, or when `StateReplay` is called). That is not directly relevant to this work, but really needs to happen soon.

chain/vm/fvm.go
vyzo: sure.

chain/vm/fvm.go
vyzo: nitpicker! ok, will fix.

chain/vm/fvm.go
vyzo: ok.

jennijuju:
@arajasek id prefer this PR to be base off master, let me know your thoughts!

chain/vm/fvm.go
vyzo: done.

chain/vm/fvm.go
vyzo: done.

arajasek:
> @arajasek id prefer this PR to be base off master, let me know your thoughts!

I'm not sure -- there are users who want to use them, and ideally they would be able to do so in the minimum release. 

But it is ultimately nice-to-have, and any change late in the release process is undesirable, so happy to go into master instead.

vyzo:
I'll leave the rebase up to you, updating the code now; I just have to move the util functions and it's done.

chain/vm/vmi.go
```diff
@@ -25,11 +24,17 @@ var useFvmForMainnetV15 = os.Getenv("LOTUS_USE_FVM_TO_SYNC_MAINNET_V15") == "1"
...
 	if opts.NetworkVersion >= network.Version16 {
+		if os.Getenv("LOTUS_FVM_DEBUG") == "1" {
+			return NewDualExecutionFVM(ctx, opts)
+		}
 		return NewFVM(ctx, opts)
 	}
 
 	// Remove after v16 upgrade, this is only to support testing and validation of the FVM
 	if useFvmForMainnetV15 && opts.NetworkVersion >= network.Version15 {
+		if os.Getenv("LOTUS_FVM_DEBUG") == "1" {
```
arajasek, line 35: Okay with this duplication cuz it's getting dropped soon

chain/consensus/filcns/upgrades.go
```diff
@@ -1506,12 +1503,8 @@ func LiteMigration(ctx context.Context, bstore blockstore.Blockstore, newActorsM
...
 	migrations := make(map[cid.Cid]cid.Cid)
 
-	for _, entry := range newManifestData.Entries {
-		oldCodeCid, ok := oldManifest.Get(entry.Name)
-		if !ok {
-			return cid.Undef, xerrors.Errorf("code cid for %s actor not found in old manifest", entry.Name)
-		}
-		migrations[oldCodeCid] = entry.Code
+	for i, entry := range newManifestData.Entries {
+		migrations[oldManifestData.Entries[i].Code] = entry.Code
```
Stebalien, line 1507: This assumes that we aren't adding/removing any actors. We need to lookup actors by name.

itests/lite_migration_test.go
```diff
@@ -67,10 +61,8 @@ func TestLiteMigration(t *testing.T) {
 	require.NoError(t, err)
 
 	migrations := make(map[cid.Cid]cid.Cid)
-	for _, entry := range newManifestData.Entries {
-		oldCodeCid, ok := oldManifest.Get(entry.Name)
-		require.True(t, ok)
-		migrations[oldCodeCid] = entry.Code
+	for i, entry := range newManifestData.Entries {
+		migrations[oldManifestData.Entries[i].Code] = entry.Code
```
Stebalien, line 65: ditto

chain/vm/vmi.go
```diff
@@ -25,11 +25,17 @@ var useFvmForMainnetV15 = os.Getenv("LOTUS_USE_FVM_TO_SYNC_MAINNET_V15") == "1"
 
 func NewVM(ctx context.Context, opts *VMOpts) (Interface, error) {
 	if opts.NetworkVersion >= network.Version16 {
+		if os.Getenv("LOTUS_FVM_DEBUG") == "1" {
```
Stebalien, line 28: Put this in a global var so we only check it once?

chain/actors/manifest.go
```diff
@@ -113,6 +113,22 @@ func ReadManifest(ctx context.Context, store cbor.IpldStore, mfCid cid.Cid) (map
...
+	if err := adtStore.Get(ctx, mfCid, &mf); err != nil {
+		return nil, xerrors.Errorf("error reading manifest: %w", err)
+	}
+
+	if err := adtStore.Get(ctx, mf.Data, &mfData); err != nil {
+		return nil, xerrors.Errorf("error fetching data: %w", err)
+	}
+
+	return &mfData, nil
+}
```
Stebalien, line 130: This assumes a single manifest version. Maybe that's fine?

The correct way to do this is:

1. Load the `manifest.Manifest`.
2. Call `mf.Load(ctx, adtStore)`.
3. Return the `manifest.Manifest`.

Stebalien: (and probably add a method to `manifest.Manifest` for enumerating the actors)

(and yeah, I'm not a fan of this API)

chain/vm/fvm.go
```diff
@@ -303,6 +308,111 @@ func NewFVM(ctx context.Context, opts *VMOpts) (*FVM, error) {
...
+	}
+
+	av, err := actors.VersionForNetwork(opts.NetworkVersion)
+	if err != nil {
+		return nil, xerrors.Errorf("error determining actors version for network version %d: %w", opts.NetworkVersion, err)
+	}
+
+	switch av {
+	case actors.Version8:
+		if bundlePath := os.Getenv("LOTUS_FVM_DEBUG_BUNDLE_V8"); bundlePath != "" {
```
Stebalien, line 372: nit: global variable.

chain/vm/fvm.go
```diff
@@ -303,6 +308,111 @@ func NewFVM(ctx context.Context, opts *VMOpts) (*FVM, error) {
...
+
+				actorRedirect[from] = e.Code
+			}
+
+			if len(actorRedirect) > 0 {
+				mappingCid, err := putMapping(actorRedirect)
+				if err != nil {
+					return nil, xerrors.Errorf("error writing redirect mapping: %w", err)
+				}
+				fvmopts.ActorRedirect = mappingCid
```
Stebalien, line 400: nit: can we put this in a separate helper function?

chain/vm/fvm.go
```diff
@@ -303,6 +308,111 @@ func NewFVM(ctx context.Context, opts *VMOpts) (*FVM, error) {
...
+			epoch:      opts.Epoch,
+		},
+		Epoch:          opts.Epoch,
+		BaseFee:        opts.BaseFee,
+		BaseCircSupply: circToReport,
+		NetworkVersion: opts.NetworkVersion,
+		StateBase:      opts.StateBase,
+		Tracing:        EnableDetailedTracing,
+		Debug:          true,
+	}
```
Stebalien, line 343: nit: we should probably have a helper function for this logic, so we don't duplicate it in multiple places. The only difference here is `Debug: true`, and the tiered blockstore. Ideally, we'd have a `DefaultFVMOpts(*VMOpts)` method, or something.

chain/vm/fvm.go
```diff
@@ -427,3 +537,110 @@ func (vm *FVM) ApplyImplicitMessage(ctx context.Context, cmsg *types.Message) (*
...
+		defer wg.Done()
+		ret, err = vm.main.ApplyMessage(ctx, cmsg)
+	}()
+
+	go func() {
+		defer wg.Done()
+		if _, err := vm.debug.ApplyMessage(ctx, cmsg); err != nil {
+			log.Errorf("debug execution failed: %w", err)
+		}
+	}()
```
Stebalien, line 580: So... this doesn't work as the debug messages will have different gas fees. That means:

1. The balances will be different.
2. Some messages may just fail.

This will lead to really annoying and hard to debug issues. We _can_ merge this as a WIP_JUST_TESTING patch, but we need to make that clear.

The right way to do this would be to:

1. Apply in debug mode (disabling gas accounting?). Balances will be correct because we charge for gas by charging for the gas _limit_ up-front, then refunding any leftovers. So the balance will be correct for the duration of the message execution, just not after the message is done executing.
2. Revert.
3. Apply normally.
4. Go back to 1 for the next message.

But:

1. We can't currently perform that revert.
2. That is even more likely to cause problems...

I think the answer here is to leave this as a super experimental feature, but be _very_ clear that this is _just_ for debugging and absolutely not to be relied on.

chain/consensus/filcns/upgrades.go
arajasek: Yeah, you're right, i got carried away with my desperation to simplify things.

chain/vm/vmi.go
arajasek: sure

chain/actors/manifest.go
arajasek: I agree -- I'm gonna drop this method for now, and use the builtin actors keys to ennumerate (and create an issue for the flow you're describing, because I agree it's needed).

chain/vm/fvm.go
arajasek: yuss, good call

chain/vm/fvm.go
arajasek: seems reasonable, thanks!

chain/vm/fvm.go
arajasek: Yeah, I'll throw on a disclaimer.

Honestly, the biggest use I envision getting out of this is actually intentionally failing messages, and using the actor error string to convey debug information (eg. adding `actor_error!("i reached this if branch because the miner power is non-zero");` to the miner_actor).

chain/consensus/filcns/upgrades.go
```diff
@@ -1477,21 +1477,23 @@ func LiteMigration(ctx context.Context, bstore blockstore.Blockstore, newActorsM
...
-	newManifest := manifest.Manifest{}
+	var newManifest manifest.Manifest
 	if err := store.Get(ctx, newActorsManifestCid, &newManifest); err != nil {
+		return cid.Undef, xerrors.Errorf("error getting new manifest: %w", err)
+	}
+
+	// populate the entries field of the manifest
+	if err = newManifest.Load(ctx, store); err != nil {
 		return cid.Undef, xerrors.Errorf("error loading new manifest: %w", err)
 	}
```
Stebalien, line 1494: Can we use the `LoadManifest` helper?

arajasek: yes!

chain/stmgr/utils.go
```diff
@@ -226,30 +226,13 @@ func GetManifest(ctx context.Context, st *state.StateTree) (*manifest.Manifest,
...
-	if err := st.Store.Get(ctx, mf.Data, &manifestData); err != nil {
-		return nil, xerrors.Errorf("filed to load manifest data: %w", err)
+	var mfData manifest.ManifestData
+	if err := wrapStore.Get(ctx, actorsManifestDataCid, &mfData); err != nil {
+		return nil, xerrors.Errorf("error fetching data: %w", err)
 	}
-	return manifestData.Entries, nil
+
+	return &mfData, nil
 }
```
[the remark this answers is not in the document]
arajasek: That won't actually _give_ me a `ManifestData`, which is what I wanna return (I don't actually have a `Manifest` here).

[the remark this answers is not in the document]
Stebalien: But do we need the manifest data?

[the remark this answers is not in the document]
Stebalien: Ah, I see. That makes sense.

Stebalien (approved):

Issue #8793, opened 2022-06-02
Sealing code refators


magik6k:
The sealing code is structured in a very weird way due to historic reasons; We should revisit those decisions, and possibly cleanup some things:

* Move code from `extern/` into directories which make more sense, closer to proving code in `storage/`
* Remove adapter code from `storage/`
* Get rid of the `specs-storage` repo, which as far as I can tell only exists to slow down development on sealing code and be annoying.
* Structure packages in a way which will make future refactors easier

(blocked by nv16 changes landing in master, those refactors would generate a lot of conflicts)

PR #8858, opened 2022-06-14
chore: storage refactors part 1

magik6k:
## Related Issues
Part 1 of https://github.com/filecoin-project/lotus/issues/8793

## Proposed Changes
* [x] Move things in `storage/` into clean sub-packages
* [x] Move non-extern things from `extern/` to packages in `storage/`

For the next PR (didn't want to put logic changes in this PR to avoid mixing them with a big high-noise diff):
* [ ] Remove `storage/adapter_*.go`
* [ ] Drop dependency on specs-storage
* [ ] Add new architecture Readme to `storage/`

Main moves are:
* `storage/wdpost_*.go` -> `storage/wdpost/` - making space for other storage related packages here
* `extern/sector-storage` -> `storage/sealer` - don't pretend like this will live in a separate repo; should be much easier to find
* `extern/sector-storage/stores` -> `storage/paths` - The `stores` package mostly deals with paths with sectors, seems like a much more logical name.
* `extern/storage-sealing` -> `storage/pipeline` - This is the sealing pipeline; didn't want to name it sealer/sealing for the obvious reason that it sounds a bit like `sealer` which is already in use above.

There may be some more PRs moving things around, but this should be the biggest one, addressing the most annoying/overdue things


api/api_storage.go
```diff
@@ -22,9 +22,9 @@ import (
 
 	"github.com/filecoin-project/lotus/chain/actors/builtin"
 	"github.com/filecoin-project/lotus/chain/types"
-	"github.com/filecoin-project/lotus/extern/sector-storage/fsutil"
-	"github.com/filecoin-project/lotus/extern/sector-storage/storiface"
-	"github.com/filecoin-project/lotus/extern/storage-sealing/sealiface"
+	"github.com/filecoin-project/lotus/storage/pipeline/sealiface"
+	"github.com/filecoin-project/lotus/storage/sealer/fsutil"
+	storiface "github.com/filecoin-project/lotus/storage/sealer/storiface"
```
geoff-vball, line 27: Redundant alias (one of many). I noticed some for `storiface` and some `ffiwrapper`.

geoff-vball (commented):
The linter needs to be appeased. 

Github is also complaining about some unsanitized logs, but it looks like most of these are coming from sources with admin perms anyway?

chain/actors/builtin/paych/paych.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
-	"github.com/filecoin-project/go-state-types/big"
+	big "github.com/filecoin-project/go-state-types/big"
```
geoff-vball, line 12: One last redundant alias. This needs to be changed in the template.

PR #8942, opened 2022-06-28
feat: shed: migrate-nv16

arajasek:
Updates the command from nv15

### DIFF
--- a/cmd/lotus-shed/migrations.go
+++ b/cmd/lotus-shed/migrations.go
@@ -21,8 +21,8 @@ import (
 )
 
 var migrationsCmd = &cli.Command{
-	Name:        "migrate-nv15",
-	Description: "Run the specified migration",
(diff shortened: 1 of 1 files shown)


cmd/lotus-shed/migrations.go
```diff
@@ -21,8 +21,8 @@ import (
 )
 
 var migrationsCmd = &cli.Command{
-	Name:        "migrate-nv15",
-	Description: "Run the specified migration",
+	Name:        "migrate-nv16",
```
ZenGround0, line 24: Idea: it might be nice to keep track of the old ones along with brining in the new ones in case someone wants to play with historical migrations.  Probably more work to factor things out right now than its worth 🤷

ZenGround0 (approved):

magik6k (approved):

PR #8978, opened 2022-07-06
fix: Delegate storage auth on market nodes

magik6k:
Currently on market nodes all APIs are behind the market API auth, which can be backed by an API secret that's different from the one on the main lotus-miner storage node.

This is for the most part fine, but breaks if there are storage paths attached to the market node - this is because the storage subsystem assumes that all http auth is backed by the same api secret. Storage on workers just delegates auth to the main storage node to keep the api secret in one place.

This PR implements similar mechanism for lotus market nodes.

(boost may need to backport this tho I'm not sure if they support attaching storage paths)


node/impl/storminer.go
```diff
@@ -97,6 +98,9 @@ type StorageMinerAPI struct {
 	Epp gen.WinningPoStProver `optional:"true"`
 	DS  dtypes.MetadataDS
 
+	// StorageService is populated when we're not the main storage node (e.g. we're a markets node)
+	StorageService modules.MinerStorageService `optional:"true"`
```
ZenGround0, line 102: How does this get set.  I'm guessing there's DI magic but I don't see the field being set or an explicit config to initialize this which I kind of expect since its optional.

ZenGround0 (approved):

node/impl/storminer.go
magik6k: The StorageMinerAPI struct is filled through fx.Extract, with each field being field individually thanks to the `fx.In` "trait" on the struct. `optional` tells Fx to not fail if a constructor for `modules.MinerStorageService` doesn't exist in the DI container - instead it leaves it not set (nil).

PR #8986, opened 2022-07-07
feat: recovery: Config for maximum partition count per message

arajasek:
## Related Issues

It is possible that users might have DeclareFaultsRecovered Messages with gasused approaching the block gas limit. See msg bafy2bzacebgkmblyhe5jptkdyiaq5euk6o4w5pdgdhxfeegt4thyp4jf2svea.

## Proposed Changes

#8982 , but for DeclareFaultsRecovered

	// Maximum number of partitions to declare in a single DeclareFaultsRecovered message. 0 = no limit.

	// In some cases when submitting DeclareFaultsRecovered messages,
	// there may be too many recoveries to fit in a BlockGasLimit.
	// In those cases it may be necessary to set this value to something low (eg 1);
	// Note that setting this value lower may result in less efficient gas use - more messages will be sent than needed,
	// resulting in more total gas use (but each message will have lower gas limit)

We should probably just recommend users set the value to 1 if they run into this issue.

## Checklist

- [ ] CI is green

### DIFF
--- a/storage/wdpost/wdpost_run_faults.go
+++ b/storage/wdpost/wdpost_run_faults.go
@@ -36,14 +36,15 @@ import (
 // TODO: the waiting should happen in the background. Right now this
 //  is blocking/delaying the actual generation and submission of WindowPoSts in
 //  this deadline!
-func (s *WindowPoStScheduler) declareRecoveries(ctx context.Context, dlIdx uint64, partitions []api.Partition, tsk types.TipSetKey) ([]miner.RecoveryDeclaration, *types.SignedMessage, error) {
+func (s *WindowPoStScheduler) declareRecoveries(ctx context.Context, dlIdx uint64, partitions []api.Partition, tsk types.TipSetKey) ([][]miner.RecoveryDeclaration, []*types.SignedMessage, error) {
 	ctx, span := trace.StartSpan(ctx, "storage.declareRecoveries")
 	defer span.End()
 
 	faulty := uint64(0)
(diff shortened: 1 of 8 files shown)


magik6k (commented):
Would be good to somehow test this

storage/wdpost/wdpost_run_faults.go
```diff
@@ -36,14 +36,14 @@ import (
...
-func (s *WindowPoStScheduler) declareRecoveries(ctx context.Context, dlIdx uint64, partitions []api.Partition, tsk types.TipSetKey) ([]miner.RecoveryDeclaration, *types.SignedMessage, error) {
+func (s *WindowPoStScheduler) declareRecoveries(ctx context.Context, dlIdx uint64, partitions []api.Partition, tsk types.TipSetKey) ([]*[]miner.RecoveryDeclaration, []*types.SignedMessage, error) {
 	ctx, span := trace.StartSpan(ctx, "storage.declareRecoveries")
 	defer span.End()
 
 	faulty := uint64(0)
-	params := &miner.DeclareFaultsRecoveredParams{
-		Recoveries: []miner.RecoveryDeclaration{},
-	}
+	currRecoveries := &[]miner.RecoveryDeclaration{}
```
magik6k, line 44: Pointers to arrays feel rather hacky. Can we just operate on the last item in `batchedRecoveryDecls`, instead of this separate `currRecoveries`?

magik6k (approved):

Issue #8697, opened 2022-05-23
Updating a sector state does not work


RobQuistNL:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [X] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
1.15.3-rc1
```

### Describe the Bug

```
$ lotus-miner sectors list
481  Committing  NO       NO      n/a                           CC
```

```
$ lotus-miner sectors update-state --really-do-it 481 ComputeProofFailed

```

```
$ lotus-miner sectors list
481  Committing  NO       NO      n/a                           CC
```

```
$ lotus-miner sectors update-state --really-do-it 481 SubmitCommitAggregate

```

```
$ lotus-miner sectors list
481  Committing  NO       NO      n/a                           CC
```

### Logging Information

```text
-
```

### Repo Steps

Run update-state command
[...]

PR #8991, opened 2022-07-07
fix: gas: estimate gas with a zero base-fee

Stebalien:
## Related Issues

https://github.com/filecoin-project/lotus/issues/8697

## Proposed Changes

Estimate gas with a zero base-fee. Otherwise, an account will need funds to estimate the max possible gas a message could take (which is usually the block gas limit).

This does mean gas estimation no longer checks if the sending account has enough funds to cover the message cost, but MpoolPush will now do this. Furthermore, subsystems like the window post worker will carefully pick an account with funds to cover the message. Previously, the fact that gas estimation would fail up-front caused this code to be less useful.
## Checklist

- [x] CI is green

### DIFF
--- a/chain/stmgr/call.go
+++ b/chain/stmgr/call.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
 	"github.com/filecoin-project/go-state-types/crypto"
 
 	"github.com/filecoin-project/lotus/api"
@@ -156,6 +157,10 @@ func (sm *StateManager) CallWithGas(ctx context.Context, msg *types.Message, pri
 	ctx, span := trace.StartSpan(ctx, "statemanager.CallWithGas")
 	defer span.End()
 
+	// Copy the message as we'll be modifying the nonce.
+	msgCopy := *msg
+	msg = &msgCopy
+
 	if ts == nil {
 		ts = sm.cs.GetHeaviestTipSet()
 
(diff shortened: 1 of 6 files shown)


itests/deals_retry_deal_no_funds_test.go
```diff
@@ -178,69 +186,3 @@ func TestDealsRetryLackOfFunds_blockInPublishDeal(t *testing.T) {
 	case <-time.After(time.Second * 15):
 	}
 }
-
-func TestDealsRetryLackOfFunds_belowLimit(t *testing.T) {
```
Stebalien, line 182: As far as I can tell, this was testing a bug? We should always allow retries, no matter how low our account balance is.

chain/stmgr/call.go
```diff
@@ -156,6 +157,10 @@ func (sm *StateManager) CallWithGas(ctx context.Context, msg *types.Message, pri
 	ctx, span := trace.StartSpan(ctx, "statemanager.CallWithGas")
 	defer span.End()
 
+	// Copy the message as we'll be modifying the nonce.
```
nonsense, line 160: I don't understand why we need this copy now, and not before?

Stebalien: We needed it before as well, technically. In practice, this doesn't matter because there's exactly one caller of this method (gasEstimateGasLimit) and that caller _also_ copies the message.

itests/gas_estimation_test.go
```diff
@@ -0,0 +1,54 @@
...
+		Value: big.Zero(),
+	}
+
+	sm, err := client.MpoolPushMessage(ctx, msg, nil)
+	require.NoError(t, err)
+
+	_, err = client.StateWaitMsg(ctx, sm.Cid(), 3, api.LookbackNoLimit, true)
+	require.NoError(t, err)
+
+	// Create that address.
```
arajasek, line 43: Wrong comment.

itests/deals_retry_deal_no_funds_test.go
arajasek: I don't think that's true -- I believe we explicitly _don't_ auto-retry (can be manually triggered, or happens on restart, per discussion [here](https://github.com/filecoin-project/lotus/pull/7454#issue-1018338971)).

Is this test failing as a result of your change? It may just need to have its funds lowered to, like, 1 attoFIL.

arajasek (commented):
Thank you for the PR, looks pretty landable, but a couple things.

itests/deals_retry_deal_no_funds_test.go
Stebalien: But yes, we never auto-retry. But that wasn't what this was testing. This was testing whether or not the deal _failed_ if we were below the limit.

arajasek: Yeah, you seem to be right...

arajasek (approved):

PR #8994, opened 2022-07-07
_ci_: update snapcraft and release flow logic

ianconsolata:
## Related Issues
https://app.circleci.com/pipelines/github/filecoin-project/lotus/21961/workflows/3bb820f2-4f7a-43d9-880f-7fe4dfecdc72/jobs/513420
Snap builds were previously failing due to an authentication issue. I think it was likely caused by outdated credentials, but snapcraft has also changed their auth method to use an env variable recently and deprecated the old value. So I set that new env var in CircleCI project settings, meaning we should be able to remove this section of CI where we write the old auth file.

I also tweaked the jobs for snapcraft and dockerhub to do trial-runs on all release branches, but still only publish on release tags. While it wouldn't have caught this issue, since it was publish related, I think it's still good practice to add this as a small sanity check.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,7 +115,7 @@ jobs:
           command: |
             git --no-pager diff go.mod go.sum
             git --no-pager diff --quiet go.mod go.sum
-  build-all:
+  build-linux:
     executor: golang
     steps:
       - install-deps
@@ -518,6 +518,19 @@ jobs:
   publish:
     description: publish binary artifacts
     executor: ubuntu
+    parameters:
+      linux:
+        default: false
+        description: publish linux binaries?
+        type: boolean
+      darwin:
+        default: false
+        description: publish darwin binaries?
+        type: boolean
+      appimage:
+        default: false
+        description: publish appimage binaries?
+        type: boolean
     steps:
       - run:
           name: Install git jq curl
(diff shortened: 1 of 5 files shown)


ianconsolata:
I opted to make modified copies of `scripts/publish-release` and `scripts/build-bundle` because I wasn't sure where else those scripts might be used, and if it was safe to modify them. If they are not used anywhere else, I can modify this PR to edit those scripts in place instead.

.circleci/config.yml
```diff
@@ -536,6 +541,19 @@ jobs:
   publish:
     description: publish binary artifacts
     executor: ubuntu
+    parameters:
+      linux:
+        default: false
+        description: publish linux binaries?
```
ZenGround0, line 547: Why the question mark?

.circleci/config.yml
```diff
@@ -713,14 +741,19 @@ jobs:
           at: "."
       - packer_build:
           template: tools/packer/lotus-snap.pkr.hcl
-  publish-dockerhub:
+
+  build-dockerhub:
```
ZenGround0, line 745: Why change name?  The old name matches description better.

ZenGround0 (approved):
This is well outside my area of expertise but looks ok

.circleci/config.yml
```diff
@@ -115,7 +119,8 @@ jobs:
           command: |
             git --no-pager diff go.mod go.sum
             git --no-pager diff --quiet go.mod go.sum
-  build-all:
+
+  build-linux:
```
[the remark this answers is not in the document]
ianconsolata: It only builds linux binaries, there are separate build tasks for other target platforms. I think the all was supposed to mean it's building all lotus commands, but I found it very confusing. `build-linux` matches pattern used elsewhere better (i.e. `build-macos`)

.circleci/config.yml
ianconsolata: I can make the comment clearer, but the parameter is to indicate whether the linux binaries should be published or not. I usually use ? to indicate true/false things.

coryschwartz:
@ianconsolata we need  to make the gen-check test pass as well.... This check actually re-generates the circleci config from template and checks for differences.

scripts/publish-release.sh
```diff
@@ -78,6 +78,7 @@ do
   echo "Uploading ${RELEASE_FILE}..."
```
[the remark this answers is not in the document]
ianconsolata: I don't think so, I wasn't sure if it was referenced anywhere outside of the circleci file. I removed it for now.

.circleci/config.yml
ianconsolata: I removed this for now though. It was a minor change, and not really that important.

PR #9056, opened 2022-07-19
feat:chain:splitstore chain prune

ZenGround0:
## Related Issues
#6577 

## Proposed Changes
Because many changes from the existing coldstore GC PR (#6728) have made it into lotus master and have been heavily modified with the last year of changes this PR is meant to replace #6728.  

The scope of this PR is adding correct splitstore coldstore GC, integration tests of the splitstore, and the `chain prune` command to trigger this manually.  Auto cold GC will happen in another PR

## Additional Info
There are a few tricky changes to the splitstore itself
- [x] splitstore correctly serves transaction requests with respect to the right store during the critical section.  compactType field and splitstore.go functions are modified to cover this.
- [x] purge is generalized to delete from the correct store (again compactType achieves this)
- [x] markset now generally contains both cold and hot store live cids in order to handle transaction protection uniformly.  I've optimized puts to not be protected during cold compaction. Reification still needs work and I might do a similar optimization but cold reified objects might show up in hotstore compaction markset to keep things simple.
- [x] new paths and startup code so coldstore compaction can benefit from checkpoints and recovery

Testing also has some nuanced challenges.  The splitstore / itest combined complex relies on the coldstore being the universal blockstore for genesis or any other setup state to be loaded.  Coldstore can only be gced if it is backed by a badger blockstore as map blockstores don't implement the BlockstoreGC interface.  This means that we need to add an option to itest setup to instantiate a fs repo to get a universal badger blockstore.  We don't actually need this if we don't test coldstore gc.

Testing
- [x] framework writing small amount of hash linked state and then allowing to overwrite it to test whether hot and cold store moved around or deleted that state
- [x] One hotstore discard happy path test
- [x] One badger coldstore and chain prune happy path test

## Checklist

- [ ] CI is green


itests/kit/blockminer.go
```diff
@@ -255,6 +258,15 @@ func (bm *BlockMiner) MineBlocks(ctx context.Context, blocktime time.Duration) {
 		defer bm.wg.Done()
 
 		for {
+			if paused := atomic.LoadInt64(&bm.pause); paused == 1 {
```
ZenGround0, line 261: I'm an atomic n00b so I might be getting the usage quite wrong

itests/kit/blockminer.go
```diff
@@ -283,6 +295,18 @@ func (bm *BlockMiner) InjectNulls(rounds abi.ChainEpoch) {
 	atomic.AddInt64(&bm.nextNulls, int64(rounds))
 }
 
+// Pause compels the miner to wait for a signal to restart
+func (bm *BlockMiner) Pause() {
```
ZenGround0, line 299: These helped me avoid races in test and seem generally useful.  Worth checking carefully to make sure I'm not opening up deadlocks.

itests/kit/ensemble.go
```diff
@@ -335,7 +336,49 @@ func (n *Ensemble) Start() *Ensemble {
...
 	for i, full := range n.inactive.fullnodes {
-		r := repo.NewMemory(nil)
+
+		var r repo.Repo
+		if !full.options.fsrepo {
+			rmem := repo.NewMemory(nil)
+			n.t.Cleanup(rmem.Cleanup)
+			r = rmem
+		} else {
+			repoPath, err := ioutil.TempDir(os.TempDir(), "lotus-itest-fsrepo-temp-")
```
ZenGround0, line 346: Coldstore prune won't work unless we have a badger BS so I'm introducing this.  Please double check I'm not leaving any stray testing files behind

blockstore/idstore.go
```diff
@@ -102,6 +102,14 @@ func (b *idstore) Put(ctx context.Context, blk blocks.Block) error {
 	return b.bs.Put(ctx, blk)
 }
 
+func (b *idstore) ForEachKey(f func(cid.Cid) error) error {
```
ZenGround0, line 105: Needed so that badge backed cold store satisfies the interface needed for pruning

blockstore/mem.go
```diff
@@ -14,35 +14,36 @@ func NewMemory() MemBlockstore {
 }
 
 // MemBlockstore is a terminal blockstore that keeps blocks in memory.
-type MemBlockstore map[cid.Cid]blocks.Block
+// To match behavior of badger blockstore we index by multihash only.
+type MemBlockstore map[string]blocks.Block
```
ZenGround0, line 18: Something I learned during this work is that the badger backed blockstores store mh only as keys to blocks. Until now the MemBlockstore has had different behavior, only matching when codec also matches. When using a coldstore backed by this MemBlockstore this led to a byzantine failure case.  I think this is more correct across the board.

ZenGround0:
I'll add a `chain prune` discussion tutorial before merging

blockstore/splitstore/splitstore_prune.go
```diff
@@ -0,0 +1,592 @@
...
+	}
+
+	// ensure that we have compacted at least once
+	if s.compactionIndex == 0 {
+		atomic.StoreInt32(&s.compacting, 0)
+		return xerrors.Errorf("splitstore has not compacted yet")
+	}
+
+	// get the current tipset
+	curTs := s.chain.GetHeaviestTipSet()
```
magik6k, line 129: Don't we technically want the last tipset we've seen by the head change handler, or does that not really matter?

blockstore/splitstore/splitstore_prune.go
```diff
@@ -0,0 +1,592 @@
...
+		s.prune(curTs, retainStateP, doGC)
+
+		log.Infow("prune done", "took", time.Since(start))
+	}()
+
+	return nil
+}
+
+func (s *SplitStore) prune(curTs *types.TipSet, retainStateP func(int64) bool, doGC func() error) {
+	log.Info("waiting for active views to complete")
```
magik6k, line 150: This and the log below can probably be demoted to debug

blockstore/splitstore/splitstore_prune.go
```diff
@@ -0,0 +1,592 @@
...
+	default:
+	}
+
+	log.Infow("chain walk done", "walked", walkCnt)
+
+	return nil
+}
+
+// like walkObject but treats missing references laxly; faster version of walkObjectIncomplete
+// without an occurs check.
```
magik6k, line 550: Is skipping the check actually faster here? I'd guess yes for receipts/messages, but not for stateroots which are mostly the same between epochs

blockstore/splitstore/splitstore_prune.go
```diff
@@ -0,0 +1,592 @@
...
+	deadw, err := NewColdSetWriter(s.deadSetPath())
+	if err != nil {
+		return xerrors.Errorf("error creating coldset: %w", err)
+	}
+	defer deadw.Close() //nolint:errcheck
+
+	// some stats for logging
+	var liveCnt, deadCnt int
+
+	err = s.cold.(bstore.BlockstoreIterator).ForEachKey(func(c cid.Cid) error {
```
magik6k, line 242: This feels like it will be very very slow on larger nodes (which is maybe fine, just don't do coldstore gc on archival nodes)

magik6k (commented):
First pass, couple of (mostly nitpicky) comments. I feel like I have a decent idea of what's going on here, and the core logic makes sense.

Would be nice to setup some chaos-style tests for both this and hotstore compaction, crashing the process in the middle of compacting, and seeing if we resume correctly (not as a part of this PR, but before we call splitstore stable)

api/api_full.go
```diff
@@ -169,6 +169,10 @@ type FullNode interface {
 	// If oldmsgskip is set, messages from before the requested roots are also not included.
 	ChainExport(ctx context.Context, nroots abi.ChainEpoch, oldmsgskip bool, tsk types.TipSetKey) (<-chan []byte, error) //perm:read
 
+	// ChainPrune prunes the stored chain state and garbage collects; only supported if you
+	// are using the splitstore
+	ChainPrune(ctx context.Context, opts map[string]interface{}) error //perm:admin
```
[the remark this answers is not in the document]
ZenGround0: Yeah that would almost be too nice...

itests/kit/ensemble_opts.go
```diff
@@ -65,3 +65,10 @@ func Account(key *key.Key, balance abi.TokenAmount) EnsembleOpt {
 		return nil
 	}
 }
+
+func Time(offset time.Duration) EnsembleOpt {
```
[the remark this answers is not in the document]
ZenGround0: I'm actually going to delete this because I ended up hacking around itest timing in a different way.  Using this option makes tests incredibly slow so its kind of a footgun

blockstore/splitstore/splitstore.go
```diff
@@ -260,8 +284,14 @@ func (s *SplitStore) Has(ctx context.Context, cid cid.Cid) (bool, error) {
 		if has {
 			return s.has(cid)
 		}
-
-		return s.cold.Has(ctx, cid)
+		switch s.compactType {
```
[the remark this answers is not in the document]
ZenGround0: We are in the critical section of (either hot or cold) compaction.  This means that we need to treat the markset as the source of truth about which blocks are actually in the stores.  Above we handle the case where the markset protects this block.  Now we are in the case where the markset does not protect which means the block is not in the store being compacted.  We still want to check cold if its not in the hotstore and similarly we want to check hot if its not in the coldstore.

The last case is weird and will probably never happen but its more correct I think.  Someone might be able to trigger it by adding state to a fork, watching that get pruned out of the cold store and then having access to that state show up in the head of the chain, we'd want to make sure we can get it from the hotstore in that case.

blockstore/splitstore/splitstore.go
```diff
@@ -260,8 +284,14 @@ func (s *SplitStore) Has(ctx context.Context, cid cid.Cid) (bool, error) {
...
+		case hot:
+			return s.cold.Has(ctx, cid)
+		case cold:
+			return s.hot.Has(ctx, cid)
```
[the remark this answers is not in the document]
ZenGround0: I don't think so.  Tracking references for protection won't have any effect because we are already in the critical section which begins by taking all protected refs and adding to the markset.

There is no correctness issue because we respect the markset's view of the compacting store.  For example if it is the hot store that is compacting and the markset doesn't have this cid then we only return what the cold store has and so don't leak an incorrect transient has.

blockstore/splitstore/splitstore_compact.go
```diff
@@ -1268,6 +1300,7 @@ func (s *SplitStore) completeCompaction() error {
 	defer markSet.Close() //nolint:errcheck
 
 	// PURGE
+	s.compactType = hot
```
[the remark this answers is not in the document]
ZenGround0: We do but that is handled in `completePrune`.  The compaction being used for the general GC thing and as one of compaction vs prune is unfortunate for confusion like this.

blockstore/splitstore/splitstore_prune.go
ZenGround0: Do you think this applies to nodes running with historical message data?  I've been hoping to make the lotus default do cold store pruning indefinitely and keep growing message datastore size.  Maybe I need to rethink this?

blockstore/splitstore/splitstore_prune.go
ZenGround0: This is actually code I copied wholesale from @vyzo's original PR and do not understand.  It looks like this is ok as the problem is just that it might not be as fast as it claims it is.  But let me know if you think there is a problem here and I can explore

itests/kit/blockminer.go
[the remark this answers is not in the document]
ZenGround0: Huh this is odd.  I don't know how blocks are getting mined in the test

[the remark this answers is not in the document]
ZenGround0: Oh no it makes sense unless I'm missing something.  The ending bracket doesn't match the loop it just matches the if statement.

itests/kit/blockminer.go
```diff
@@ -283,6 +295,18 @@ func (bm *BlockMiner) InjectNulls(rounds abi.ChainEpoch) {
...
+	atomic.SwapInt64(&bm.pause, 1)
+}
+
+// Restart continues mining after a pause
+func (bm *BlockMiner) Restart() {
+	paused := atomic.LoadInt64(&bm.pause)
```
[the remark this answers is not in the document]
ZenGround0: Nice and then I think I can remove the Store after unpause

blockstore/splitstore/splitstore_prune.go
ZenGround0: I don't think it matters much.  This code path is hit through a direct call to `chain prune` so I think users expect to prune against whatever the latest head is.

blockstore/splitstore/splitstore.go
```diff
@@ -408,12 +455,12 @@ func (s *SplitStore) Put(ctx context.Context, blk blocks.Block) error {
 	s.debug.LogWrite(blk)
 
 	// critical section
-	if s.txnMarkSet != nil {
+	if s.txnMarkSet != nil && s.compactType == hot { // puts only touch hot store
```
magik6k, line 458: Do we also care about warmup here?

blockstore/splitstore/splitstore_prune.go
magik6k: Yeah, we probably want some metrics around this.

Also doing marking/gc in batches instead of having to mark all dead objects first could help a lot - now you can at least restart your node wihdout losing that much progress, and things just get lighter at their own pace.

Not sure if we want this in this PR, possibly just saying that splitstore may not work that well with really big coldstores is good enough.

magik6k (approved):
Nothing serious to complain about.

Definitely needs lots of testing, and maybe measuring perf in a bunch of scenarios, but that seems to be planned already.

magik6k:
> Nothing serious to complain about.

Well, other than a bunch of itests hanging for whatever reason

itests/kit/blockminer.go
```diff
@@ -335,4 +359,8 @@ func (bm *BlockMiner) Stop() {
 	bm.t.Log("shutting down mining")
 	bm.cancel()
 	bm.wg.Wait()
+	if bm.unpause != nil {
+		close(bm.unpause)
+		bm.unpause = nil
```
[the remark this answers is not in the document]
ZenGround0: The problem is that we call Stop twice in lots of tests so we get a double close.  This causes a bunch of test faiures in CI.  I think selecting on the nil channel should be ok as long as its done with ctx.Done() in the select as well because we cancel the context before setting nil

blockstore/splitstore/splitstore_prune.go
ZenGround0: I kicked off some brainstorming here: https://github.com/filecoin-project/lotus/discussions/9128. We have a few good options if this starts to look like a problem experimentally.

blockstore/splitstore/splitstore.go
ZenGround0: We don't because warmup does not have a critical section

PR #9061, opened 2022-07-19
feat: API:  support typed errors over RPC

### LINKED ISSUE #9041: Make batch message more clever
### Checklist

### Lotus component

- [ ] lotus client
- [ ] lotus JSON-RPC API
- [X] Other

### Improvement Suggestion

as stated in PRs like https://github.com/filecoin-project/lotus/pull/8986

batch messages may hit block limit - lotus should be clever enough know when to chunk them and how.

arajasek:
## Related Issues

This is work towards, but does not finish #9041 

## Proposed Changes

- Updates go-jsonrpc.
- Introduces API errors, with only one case right now -- OutOfGas.
- Demo test.

## Additional Info

- [x] Needs https://github.com/filecoin-project/go-jsonrpc/pull/36 & tag
- [ ] The chief limitation here is that _wrapped_ errors do not get picked up by the RPC code, which relies on reflection. This is a non-trivial issue, although it is not a blocker. I'd like to discuss this with @magik6k, but it doesn't preclude landing this PR.

## Checklist

- [ ] CI is green


magik6k:
> The chief limitation here is that wrapped errors do not get picked up by the RPC code, which relies on reflection.

We could make go-jsonrpc recursively [errors.Unwrap](https://pkg.go.dev/errors#Unwrap) the error until it hits a registered error (with some depth limit just in case), then have it return a "wrapped" version of that error:
* What the error code would be in that case is a good question, we can go one of a few routes:
  * Pre-register the 'wrapped' error type, which would marshal into the `{errorMsg, subErrCode, subErr}` - probably closest how errors are wrapped in Go, but weird in non-Go apps - you can't tell the error type just by looking at the error code in the jsonrpc response
  * Register wrapped error types for each `wrapable` error type - you can tell the error just by looking at the error code, but now there are 2 error codes per error type.
  * Mix of the above - when registering an error specify if it's wrapable - if it is, always marshal it as wrapped json, if it is not, keep current behavior.

cmd/lotus/daemon.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/filecoin-project/go-paramfetch"
 
 	"github.com/filecoin-project/lotus/api"
+	lapi "github.com/filecoin-project/lotus/api"
```
magik6k, line 34: duplicate import

arajasek: fixed

magik6k:
Tests seem to be broken because of this really robust error check: https://github.com/filecoin-project/lotus/blob/master/itests/kit/blockminer.go#L194 (which we can now make actually robust with this PR)

lib/retry/retry_test.go
```diff
@@ -7,19 +7,21 @@ import (
...
 	errorsToRetry := []error{&jsonrpc.RPCConnectionError{}}
-	require.False(t, ErrorIsIn(xerrors.Errorf("random error"), errorsToRetry))
+	require.False(t, api.ErrorIsIn(xerrors.Errorf("random error"), errorsToRetry))
 }
 
 func TestRetryWrappedErrorIsInTrue(t *testing.T) {
 	errorsToRetry := []error{&jsonrpc.RPCConnectionError{}}
-	require.True(t, ErrorIsIn(xerrors.Errorf("wrapped: %w", &jsonrpc.RPCConnectionError{}), errorsToRetry))
+	require.True(t, api.ErrorIsIn(xerrors.Errorf("wrapped: %w", &jsonrpc.RPCConnectionError{}), errorsToRetry))
 }
```
magik6k, line 27: Technically those tests should also be moved to `api/`

itests/kit/blockminer.go
```diff
@@ -191,7 +192,7 @@ func (bm *BlockMiner) MineBlocksMustPost(ctx context.Context, blocktime time.Dur
 			reportSuccessFn := func(success bool, epoch abi.ChainEpoch, err error) {
 				// if api shuts down before mining, we may get an error which we should probably just ignore
 				// (fixing it will require rewriting most of the mining loop)
-				if err != nil && !strings.Contains(err.Error(), "websocket connection closed") {
+				if err != nil && !strings.Contains(err.Error(), "websocket connection closed") && !api.ErrorIsIn(err, []error{&jsonrpc.RPCConnectionError{}}) {
```
magik6k, line 195: supernit:
```suggestion
				if err != nil && !strings.Contains(err.Error(), "websocket connection closed") && !api.ErrorIsIn(err, []error{new(jsonrpc.RPCConnectionError)}) {
```

magik6k (approved):

dirkmc:
Another use case where this would be helpful:
https://github.com/filecoin-project/boost/blob/a58ea5a92a4237a018f9e7dc700936a5cebb0688/cmd/booster-bitswap/remoteblockstore/remoteblockstore.go#L97-L102

hunjixin:
sorry for these questions after the code merge

Recently venus intended to submit some of venus changes and bug fixes for rpc to the upstream, but when pushing it, it found many conflicts with TypeError

In PR, I think the current implementation of TypeError is a bit complicated.

I'm not sure if users really need the MarshalJson/UnMarshalJson functionality, but I'm looking at lotus/boost usage and it's enough to determine the type of error, and in venus requirements it's enough to determine the type of error.
in addition, I have the impression that adding meta fields may cause errors in the jsonrpc library in rust, which does not allow additional fields. meta is not a in standard jsonrpc protocol.

ErrorCode is a good idea, but maybe more better to define ErrorCode as true error type like below code

```go
type ErrorCode int

func (e ErrorCode) Error() string {
    return fmt.Sprintf("error code %d", e)
}
```

for rpc server , user fmt.Error to return specific error
```go
if res.MsgRct.ExitCode == exitcode.SysErrOutOfGas {
        return -1, rpc.ErrorCode(exitcode.SysErrOutOfGas)
}
```

or  convert all exitcode to rpc error

```go
if res.MsgRct.ExitCode != exitcode.Ok {
        return -1, xerrors.Errorf("message execution failed: exit %w, reason: %s", rpc.ErrorCode(res.MsgRct.ExitCode), res.Error)
}
```

for rpc client, use this code to check error type
```go
    require.True(t, xerrors.Is(err, rpc.EOutOfGas))
```

If  write it this way, not need to add the Register function in Rpc, and the Rpc implementation can be simpler without adding the meta field in Respnse.

Issue #6599, opened 2021-06-25
[Mining Issue] The sector is not finalized after adding a storage path if the previous store was full


Shekelme:
> Note: For security-related bugs/issues, please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).

**Describe the problem**
If the existing storage was full at the time of sector finalization, then when adding a new storage path, the finalization process does not resume (does not restart) when a new storage path is added. Operator action is required (eg restarting the miner).
After restart, sectors are successfully finalized.

```
2021-06-25T19:45:30.780+0300	DEBUG	advmgr	sector-storage/worker_local.go:129	acquired sector {{79817 488} 8} (e:4; a:0): {{0 0}   /media/filecoin/sealing_space/cache/s-t079817-488}
2021-06-25T19:45:30.780 INFO filecoin_proofs::api::post_util > clear_cache:start
2021-06-25T19:45:30.781 INFO filecoin_proofs::api::post_util > clear_cache:finish
2021-06-25T19:45:30.781+0300	DEBUG	advmgr	sector-storage/sched.go:354	SCHED 2 queued; 2 open windows
2021-06-25T19:45:30.781+0300	DEBUG	stores	stores/index.go:400	not allocating on 923441f2-af26-4e1a-8f98-8c0887a478f3, out of space (available: 488792653824, need: 518832049356)
2021-06-25T19:45:30.781+0300	ERROR	advmgr	sector-storage/sched.go:404	trySched(1) req.sel.Ok error: finding best alloc storage:
    github.com/filecoin-project/lotus/extern/sector-storage.(*allocSelector).Ok
        /home/don_genaro/lotus/extern/sector-storage/selector_alloc.go:55
  - no good path found:
    github.com/filecoin-project/lotus/extern/sector-storage/stores.(*Index).StorageBestAlloc
        /home/don_genaro/lotus/extern/sector-storage/stores/index.go:418
2021-06-25T19:45:30.781+0300	DEBUG	stores	stores/index.go:400	not allocating on 923441f2-af26-4e1a-8f98-8c0887a478f3, out of space (available: 488792653824, need: 518832049356)
2021-06-25T19:45:30.781+0300	DEBUG	stores	stores/index.go:400	not allocating on 923441f2-af26-4e1a-8f98-8c0887a478f3, out of space (available: 488792653824, need: 518832049356)
[...]

PR #9064, opened 2022-07-20
feat: Sealing: SchedulerRemoveRequest

LexLuthr:
## Related Issues

https://github.com/filecoin-project/lotus/issues/6599

## Proposed Changes

Right now there is now way to remove a request from the scheduler. We need to wait for scheduler to assign the request to a worker and then the worker job can be aborted. This creates a problem, when sector that the task is supposed to be carried out does not exist. The request stays there forever and spamming the logs.

## Additional Info

A workaround seems to be effective. User can create false sector files and let the job get assigned to a worker and abort it.
But a proper method would be better to fix this as incorrect CID in dagstore can trigger UNS jobs that will stay stuck in this state. I don't think removing the CID from dagstore will work as there is no context carried over AFAIK. I haven't create a command yet as I don't want to keep it under lotus-miner to avoid misuse and wrecking a running scheduler. 

## Checklist

- [x] CI is green


api/api_storage.go
```diff
@@ -145,6 +145,8 @@ type StorageMiner interface {
 	// SealingSchedDiag dumps internal sealing scheduler state
 	SealingSchedDiag(ctx context.Context, doSched bool) (interface{}, error) //perm:admin
 	SealingAbort(ctx context.Context, call storiface.CallID) error           //perm:admin
+	//SealingSchedRemove removes a request from sealing pipeline
+	SealingRemoveRequest(ctx context.Context, SchedId uuid.UUID) error //perm:admin
```
magik6k, line 149: by convention param names usually start lower-case

```suggestion
	SealingRemoveRequest(ctx context.Context, schedId uuid.UUID) error //perm:admin
```

cmd/lotus-miner/sealing.go
```diff
@@ -365,6 +365,12 @@ var sealingAbortCmd = &cli.Command{
 	Name:      "abort",
 	Usage:     "Abort a running job",
 	ArgsUsage: "[callid]",
+	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "requestId",
```
magik6k, line 370: nit: cli flags usually don't use camelCase.

I'd call this flag `sched`, indicating that we want to cancel a scheduler request. `request-id` doesn't really sound like something related to the scheduler more to a request on a worker.

node/impl/storminer.go
```diff
@@ -459,6 +459,10 @@ func (sm *StorageMinerAPI) SealingAbort(ctx context.Context, call storiface.Call
 	return sm.StorageMgr.Abort(ctx, call)
 }
 
+func (sm *StorageMinerAPI) SealingRemoveRequest(ctx context.Context, SchedId uuid.UUID) error {
+	return sm.StorageMgr.RemoveSchedRequest(ctx, SchedId)
```
magik6k, line 463:
```suggestion
func (sm *StorageMinerAPI) SealingRemoveRequest(ctx context.Context, schedId uuid.UUID) error {
	return sm.StorageMgr.RemoveSchedRequest(ctx, schedId)
```

storage/sealer/manager.go
```diff
@@ -1168,6 +1168,12 @@ func (m *Manager) SchedDiag(ctx context.Context, doSched bool) (interface{}, err
 	return i, nil
 }
 
+func (m *Manager) RemoveSchedRequest(ctx context.Context, SchedId uuid.UUID) error {
+	m.workLk.Lock()
+	defer m.workLk.Unlock()
```
magik6k, line 1173: I don't think we need this lock.

storage/sealer/sched.go
```diff
@@ -139,6 +141,13 @@ type workerResponse struct {
 	err error
 }
 
+type rmRequest struct {
+	id     uuid.UUID
+	rmresE chan error
+	rmresC chan struct{}
```
magik6k, line 147: * Can be a single channel, `nil` error indicating success.
* The `rm` prefix is redundant, it's already in the struct name.
```suggestion
	res chan error
```

storage/sealer/sched.go
```diff
@@ -168,7 +177,8 @@ func newScheduler(assigner string) (*Scheduler, error) {
 			prepared: map[uuid.UUID]trackedWork{},
 		},
 
-		info: make(chan func(interface{})),
+		info:      make(chan func(interface{})),
+		rmRequest: make(chan *rmRequest, 1),
```
magik6k, line 181: This channel doesn't really need to be buffered, it doesn't change how much waiting we're doing, and it may even lead to a case where we put the request here, then cancel the RemoveRequest context, and the response has nowhere to go.

storage/sealer/sched.go
```diff
@@ -381,6 +396,51 @@ func (sh *Scheduler) Info(ctx context.Context) (interface{}, error) {
...
+		if r.SchedId == rmrequest.id {
+			queue.Remove(i)
+			rmrequest.rmresC <- struct{}{}
+		}
+	}
+	rmrequest.rmresE <- xerrors.New("No request with provided details found")
+}
+
+func (sh *Scheduler) RemoveRequest(ctx context.Context, schedId uuid.UUID) error {
+	retE := make(chan error)
```
magik6k, line 416: This channel must be buffered to guarantee that responding doesn't block, if it's not this may happen:
* RemoveRequest is called
* The request is sent, and removeRequest has started
* The context in RemoveRequest is cancelled and the method quits
* removeRequest / runSched goroutine hangs on `rmrequest.res <- ..`

```suggestion
	retE := make(chan error, 1)
```

storage/sealer/sched.go
```diff
@@ -381,6 +396,51 @@ func (sh *Scheduler) Info(ctx context.Context) (interface{}, error) {
...
+
+	if sh.SchedQueue.Len() < 0 {
+		rmrequest.rmresE <- xerrors.New("No requests in the scheduler")
+	}
+
+	queue := sh.SchedQueue
+	for i, r := range *queue {
+		if r.SchedId == rmrequest.id {
+			queue.Remove(i)
+			rmrequest.rmresC <- struct{}{}
```
magik6k, line 409: I'd return after responding to guarantee that only one response can be sent (also avoid doing redundant work).

```suggestion
			rmrequest.rmresC <- struct{}{}
			return
```

storage/sealer/sched.go
```diff
@@ -381,6 +396,51 @@ func (sh *Scheduler) Info(ctx context.Context) (interface{}, error) {
 	}
 }
 
+func (sh *Scheduler) removeRequest(rmrequest *rmRequest) {
+
+	if sh.SchedQueue.Len() < 0 {
+		rmrequest.rmresE <- xerrors.New("No requests in the scheduler")
```
magik6k, line 402:
```suggestion
		rmrequest.rmresE <- xerrors.New("No requests in the scheduler")
		return
```

storage/sealer/sched.go
```diff
@@ -139,6 +141,13 @@ type workerResponse struct {
 	err error
 }
 
+type rmRequest struct {
+	id     uuid.UUID
+	rmresE chan error
+	rmresC chan struct{}
+	Ctx    context.Context
```
magik6k, line 148: This isn't used. Also note that in Go it's contexts should only be passed through func params, saving them in structs is not the best practice (although there are some cases where that's necessary)

```suggestion
```

storage/sealer/manager.go
```diff
@@ -1168,6 +1168,12 @@ func (m *Manager) SchedDiag(ctx context.Context, doSched bool) (interface{}, err
 	return i, nil
 }
 
+func (m *Manager) RemoveSchedRequest(ctx context.Context, SchedId uuid.UUID) error {
```
magik6k, line 1171:
```suggestion
func (m *Manager) RemoveSchedRequest(ctx context.Context, schedId uuid.UUID) error {
```

storage/sealer/sched.go
```diff
@@ -381,6 +381,22 @@ func (sh *Scheduler) Info(ctx context.Context) (interface{}, error) {
...
+func (sh *Scheduler) RemoveRequest(ctx context.Context, sectorID abi.SectorID, tasktype sealtasks.TaskType, priority int) error {
+	if sh.SchedQueue.Len() < 0 {
+		return xerrors.Errorf("No requests in the scheduler")
+	}
+	sh.workersLk.Lock()
+	defer sh.workersLk.Unlock()
+	queue := sh.SchedQueue
+	for i, r := range *queue {
+		if r.Sector.ID == sectorID && r.Priority == priority && r.TaskType == tasktype { // TODO: Add check to ensure request in not scheduled
+			queue.Remove(i)
```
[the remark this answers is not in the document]
LexLuthr: Yeah, I missed that one. Added it before the return.

itests/worker_test.go
```diff
@@ -401,3 +404,75 @@ func TestWindowPostWorkerManualPoSt(t *testing.T) {
...
+	}
+
+	go miner.PledgeSectors(ctx, 1, 0, nil)
+	// Keep checking till sector state is PC2, the request should get stuck as worker cannot process PC2
+	for {
+		st, err := miner.SectorsStatus(ctx, abi.SectorNumber(1), false)
+		require.NoError(t, err)
+		if st.State == api.SectorState(sealing.PreCommit2) {
+			break
+		}
```
magik6k, line 444: Probably want to add some sleep here
```suggestion
		}
		time.Sleep(time.Second)
```

itests/worker_test.go
```diff
@@ -401,3 +404,75 @@ func TestWindowPostWorkerManualPoSt(t *testing.T) {
...
+		ReturnedWork []string
+		Waiting      []string
+		CallToWork   map[string]string
+		EarlyRet     []string
+	}
+
+	go miner.PledgeSectors(ctx, 1, 0, nil)
+	// Keep checking till sector state is PC2, the request should get stuck as worker cannot process PC2
+	for {
+		st, err := miner.SectorsStatus(ctx, abi.SectorNumber(1), false)
```
magik6k, line 440: You likely want to call miner.SectorListNonGenesis() to get the correct sector number. Sectors 0/1 are going to be presealed genesis sectors

LexLuthr: @magik6k I actually updated the code a little bit. I am using StartPledge() as it gives back a map of pledged sector and thus no more guessing game. 
I am stuck with casting the interface{} to struct to extract the data of SealingSchedDiag to get UUID. Can you help with that? 
I will add the sleep once I am able to figure out this data extraction.

LexLuthr: Test fixed

magik6k (approved):

PR #9072, opened 2022-07-21
fix: cli: `./lotus-miner actor control list`, if the `owner` is not account

zl03jsj:
fix, if a miner's owner is a multisig,
command `./lotus-miner actor control list` get failed.

### DIFF
--- a/cmd/lotus-miner/actor.go
+++ b/cmd/lotus-miner/actor.go
@@ -538,20 +538,14 @@ var actorControlList = &cli.Command{
 			}
 			b := actor.Balance
 
-			var k address.Address
+			var k = a
 			// 'a' maybe a 'robust', in that case, 'StateAccountKey' returns an error.
 			if builtin2.IsAccountActor(actor.Code) {
 				if k, err = api.StateAccountKey(ctx, a, types.EmptyTSK); err != nil {
 					fmt.Printf("%s\t%s: error getting account key: %s\n", name, a, err)
 					return
 				}
-			} else { // if builtin2.IsMultisigActor(actor.Code), Are there any other cases exist?
-				if k, err = api.StateLookupRobustAddress(ctx, a, types.EmptyTSK); err != nil {
-					fmt.Printf("%s\t%s: error getting robust address: %s\n", name, a, err)
-					return
-				}
 			}
-
 			kstr := k.String()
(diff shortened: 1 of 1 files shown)


cmd/lotus-miner/actor.go
```diff
@@ -530,16 +531,25 @@ var actorControlList = &cli.Command{
...
+
+			var k address.Address
+			// 'a' maybe a 'robust', in that case, 'StateAccountKey' returns an error.
+			if builtin2.IsAccountActor(actor.Code) {
+				if k, err = api.StateAccountKey(ctx, a, types.EmptyTSK); err != nil {
+					fmt.Printf("%s\t%s: error getting account key: %s\n", name, a, err)
+					return
+				}
+			} else { // if builtin2.IsMultisigActor(actor.Code), Are there any other cases exist?
+				if k, err = api.StateLookupRobustAddress(ctx, a, types.EmptyTSK); err != nil {
```
magik6k, line 549: `StateLookupRobustAddress` is really slow, I'd say it's better to just do `k = a` here.

cmd/lotus-miner/actor.go
```diff
@@ -530,16 +531,25 @@ var actorControlList = &cli.Command{
...
+			b := actor.Balance
+
+			var k address.Address
+			// 'a' maybe a 'robust', in that case, 'StateAccountKey' returns an error.
+			if builtin2.IsAccountActor(actor.Code) {
+				if k, err = api.StateAccountKey(ctx, a, types.EmptyTSK); err != nil {
+					fmt.Printf("%s\t%s: error getting account key: %s\n", name, a, err)
+					return
+				}
+			} else { // if builtin2.IsMultisigActor(actor.Code), Are there any other cases exist?
```
magik6k, line 548: With FVM you'll be able to use custom actors to proxy/filter calls to the miner actor, so in not-so-distant future we'll see non-account/non-multisig actors here

magik6k (commented):
Thanks for the PR.

Just one comment to address.

cmd/lotus-miner/actor.go
zl03jsj: since show a actor id or address has little influence, just use `k = a` is a better way.

magik6k (approved):

PR #9082, opened 2022-07-25
Path Traversal Bug

bhaskarvilles:
Unsanitized input from os.Args[2] argument flows into io.ioutil.ReadFile, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to read arbitrary files.

Signed-off-by: Bhaskar <ram@hacker.ind.in>

### DIFF
--- a/gen/inline-gen/main.go
+++ b/gen/inline-gen/main.go
@@ -19,7 +19,7 @@ const (
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


magik6k (commented):
This is a dev-only binary, called as part of codegen.

PR #9013, opened 2022-07-11
feat: storage: Path type filters

magik6k:
## Related Issues
Implements https://github.com/filecoin-project/lotus/issues/8794

## Proposed Changes
Add two new fields to `[storagepath]/sectorstore.json`:
```yaml
// AllowTypes lists sector file types which are allowed to be put into this
// path. If empty, all file types are allowed.
//
// Valid values
// - "unsealed"
// - "sealed"
// - "cache"
// - "update"
// - "update-cache"
// Any other value will generate a warning and be ignored.

"AllowTypes": [
  "sealed", "unsealed", "update",
],



// DenyTypes lists sector file types which aren't allowed to be put into this
// path.
//
// Valid values
// - "unsealed"
// - "sealed"
// - "cache"
// - "update"
// - "update-cache"
// Any other value will generate a warning and be ignored.

"DenyTypes": [
  "cache", "update-cache"
]
```

After a path with this config is attached, lotus-miner will stop putting new files of disallowed types into that path.

If there are existing files with disallowed types in a path, those files will remain readable for PoSt/Retrieval, so the worst that can happen in case of misconfiguration is that sealing tasks will get stuck waiting for storage to become available.

The most basic setup to separate sealed data from unsealed copies would be to:
* Not have any filters on sealing paths (no change)
* Add `"DenyTypes": ["unsealed"]` to long-term sealed storage paths
* Add `"AllowTypes": ["unsealed"]` to long-term unsealed storage paths

Here's how storage list looks with filters set:
2022-07-12-131643_816x445_scrot

## Checklist

Before you mark the PR ready for review, please make sure that:
- [x] All commits have a clear commit message.
- [x] The PR title is in the form of of `<PR type>: <area>: <change being made>`
    - example: ` fix: mempool: Introduce a cache for valid signatures`
    - `PR type`: _fix_, _feat_, _INTERFACE BREAKING CHANGE_, _CONSENSUS BREAKING_, _build_, _chore_, _ci_, _docs_,_perf_, _refactor_, _revert_, _style_, _test_
    - `area`: _api_, _chain_, _state_, _vm_, _data transfer_, _market_, _mempool_, _message_, _block production_, _multisig_, _networking_, _paychan_, _proving_, _sealing_, _wallet_, _deps_
- [x] This PR has tests for new functionality or change in behaviour
[...]

PR #9089, opened 2022-07-27
fix: fsutil: account for subvolumes under storage path

clinta:
## Related Issues
Filesystems like zfs can contain sub-volumes which are distinct datasets
but share free space. There are reasons an operator may want the sealed
or cache directory to be on distinct volumes, which makes the accounting
for the drive capacity inaccurate.

## Proposed Changes
Enumerate all mounts under the storage path, and accumulate
the used space to accurately determine filesystem capacity.

## Checklist

- [x] CI is green

### DIFF
--- a/storage/sealer/fsutil/statfs_unix.go
+++ b/storage/sealer/fsutil/statfs_unix.go
@@ -4,6 +4,11 @@
 package fsutil
 
 import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"os"
+	"strings"
 	"syscall"
 
 	"golang.org/x/xerrors"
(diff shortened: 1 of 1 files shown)


clinta:
This has been tested on a calibnet miner and works as expected, now accurately showing disk usage that includes subvolumes.

I'm interested in any advise on writing tests for this. I think testing behavior will require creating filesystems which will require tests to be run as root, which I do not think is desirable.

storage/sealer/fsutil/statfs_unix.go
```diff
@@ -15,12 +20,113 @@ func Statfs(path string) (FsStat, error) {
...
+		log.Warnf("could not get all mounts under %s: %e", path, err)
+	}
+
+	for _, mount := range mounts {
+		var mountStat syscall.Statfs_t
+		if err := syscall.Statfs(mount, &mountStat); err != nil {
+			log.Warnf("could not get stats for mount %s: %e", mount, err)
+			continue
+		}
+		used += int64(mountStat.Blocks-mountStat.Bavail) * int64(mountStat.Bsize)
```
magik6k, line 39: I feel like we'll need to make this configurable somehow, or at the very least a bit smarter.

* What if there are separate filesystems here - in that case we'd want to add those to `available` as well
* Does this work with btrfs subvolumes the same way? (I'm forgetting how those work now, but would be nice if they also worked with this)

storage/sealer/fsutil/statfs_unix.go
```diff
@@ -15,12 +20,113 @@ func Statfs(path string) (FsStat, error) {
...
+	}
+
 	// force int64 to handle platform specific differences
 	//nolint:unconvert
 	return FsStat{
-		Capacity: int64(stat.Blocks) * int64(stat.Bsize),
+		Capacity: used + available,
```
magik6k, line 45: I'm actually not sure if this correct, the capacity we care about here is really "how much data can I put in this path in total", so it maybe makes sense that it drops if it's used in another subvolume.

Afaict we're only using the `Capacity` field in the CLI, which I guess can break the display in some weird ways?

storage/sealer/fsutil/statfs_unix.go
```diff
@@ -15,12 +20,113 @@ func Statfs(path string) (FsStat, error) {
...
-		Available:   int64(stat.Bavail) * int64(stat.Bsize),
-		FSAvailable: int64(stat.Bavail) * int64(stat.Bsize),
+		Available:   available,
+		FSAvailable: available,
 	}, nil
 }
+
+func getMountsUnder(path string) ([]string, error) {
+	mounts := []string{}
+	f, err := os.Open("/proc/self/mountinfo")
```
magik6k, line 54: Does that work on non-linux unixes?

magik6k (commented):
Given that https://github.com/filecoin-project/lotus/pull/9013 now lets you specify what you want stored in each storage path, I'd say that maybe it's better to require that there are no sub-mounts in storage paths.

Supporting multiple mounts in each path generates way too many weird edge-cases to support correctly - e.g. we check avail space for cache files in `storagepath/`, and see that it has enough space, but it turns out that `storagepath/cache` is a different mount with not enogh space - so we'll fail to put the file there.

storage/sealer/fsutil/statfs_unix.go
clinta: You're right that this would not be accurate if the filesystems do not share available space. I'm not sure how that can be accurately accounted for. I don't know of any way to reliably determine if two mounts share a pool of available space. You could guess that they do if the available space is the same. But with reservations and quotas you could have a situation where the available space is different, but still shared.

My thinking was this wouldn't be any worse than it works today, and would improve things for one specific use case, even if it doesn't fix every possible situation.

I don't have any experience with btrfs, so I can't comment on that.

storage/sealer/fsutil/statfs_unix.go
clinta: It is primarily an interface issue I was aiming to fix. As is, when the sealed subvolume grows, instead of showing more drive space used, the total size of the drive gradually goes down.

clinta:
How would #9013 handle the situation where two storage paths share a pool of available space?

If a storage path for sealed data, and a storage path for unsealed data share a pool of free space, and there is enough space to store one, but not both of the incoming items, will it fill the drives?

I wasn't aware of #9013 when I wrote this, but now that I am I'm thinking that there needs to be some way to inform lotus that two storage paths share free space, so that space reservations are counted against all paths that share free space.

rjan90:

rjan90:
> If there is no response within 7 days, the PR will be closed.

Closing

Issue #5227, opened 2020-12-19
lotus-worker has no "stop" command


f8-ptrk:
## Edit (@jennijuju )
Add a `lotus-worker stop` to properly terminate worker process.

## Orignal 

the title says it all

1.2.3,1.3.0...you name it

see the second part of this slack thread:

https://filecoinproject.slack.com/archives/CPFTWMY7N/p1608251903163700

@dineshshenoy 

sorry for the low quality issue - in a hurry - i will be on it Monday again and make it beautiful!

arajasek:
Doesn't seem like the highest priority

f8-ptrk:
the problem is that PC2/C2 workers do corrupt their *.log files when killing+restarting them quite often. one needs to delete the files by hand in order to make the workers start up again or, if upgrades/restarts are performed automatically, purging the whole worker folder

William8Work:
ran into this problem after upgrading v1.5.3 (I guess it's been there for a while). Had to manually located the corrupted log files and delete them in order to get the worker started.

arajasek:
Yeah, we should get around to this -- see the `lotus-miner stop` [command](https://github.com/filecoin-project/lotus/blob/e0522a6b1fa2a27e7347c3b16575f1d261a40652/cmd/lotus-miner/stop.go#L12-L27) for inspiration

rjan90:
Closing this feature request now, since the feature has be merged!

PR #9101, opened 2022-07-29
feat: worker: Add stop cmd for lotus worker

shrenujbansal:
## Related Issues

https://github.com/filecoin-project/lotus/issues/5227

## Proposed Changes

Adding cmd to stop lotus worker similar to existing processes

## Checklist

- [ ] CI is green


cmd/lotus-worker/main.go
```diff
@@ -623,6 +645,15 @@ var runCmd = &cli.Command{
 			}
 		}()
 
+		go func() {
+			<-workerApi.Done()
+			log.Warn("Shutting down...")
```
magik6k, line 650: We should let lotus-miner know that the worker is shutting down here before stopping worker API server - this can be one in one of two ways:
* Wait for 20-30 seconds here - workers are pinged every 10s (`paths.HeartbeatInterval`), and if a ping tells the lotus-miner that the worker is shutting down, it will get fully unregistered instead of being marked as disabled (see https://github.com/filecoin-project/lotus/blob/master/storage/sealer/sched_worker.go#L120-L121)
* We could add a new API method to `lotus-miner` which would tell a `schedWorker.handleWorker` goroutine to check the worker without waiting - this is more complicated, but wouldn't require waiting as much
  * Or even add a method to explicitly disconnect / drop the worker

For now just waiting is ok, we can improve this later.

magik6k (commented):
Thanks for the PR!

Needs running `make gen` to generate some missing docs.

cmd/lotus-worker/main.go
shrenujbansal: Added the wait

magik6k (approved):
Thanks for the PR!

PR #9108, opened 2022-08-01
fix: sector counter for recovery

arajasek:
I don't think #9106 is quite right.


storage/wdpost/wdpost_run_faults.go
```diff
@@ -104,22 +104,25 @@ func (s *WindowPoStScheduler) declareRecoveries(ctx context.Context, dlIdx uint6
 			Sectors:   recovered,
 		})
 
-		if recoveringSectorLimit > 0 && int64(totalRecoveries) >= recoveringSectorLimit {
-			log.Errorw("reached recovering sector limit, not all sectors will be marked as recovered")
+		totalSectorsToRecover += recoveredCount
```
arajasek, line 107: this is the fix, was just being ++ed before, which is

magik6k (approved):

PR #9126, opened 2022-08-04
refactor: chain/actors/builtin: update actor code gen for consumption by lily

frrist:
## Related Issues
This PR implements the required actor state logic for use by Lily. This change will allow lily to remove its ported actor accessor code and instead depend on lotus - making network upgrades easier for the lily development team.

## Proposed Changes
In `chain/actors/builtin` the init, market, miner, multisig, power, reward, and verifreg packages receive the following new methods:
- `AllCodes() []cid.Cid`: returns a list of actor CID's for the corresponding actor type.
- `VersionCodes() map[actors.Version]cid.Cid`: returns a map of actor versions to their corresponding CIDs.

Additionally, the `State` interfaces in the packages mentioned above receive the following new methods:
- `Code() cid.Cid`: returns the CID of the actor.
- `ActorVersion() actors.Version`: returns the version of the actor.
- `ActorKey() string`: returns the ActorKey (string) of the actor (used for getting its CID via `GetActorCodeID`.

Lastly, the following changes were made to the `State` interface for each package mentioned above. These changes allow lily to perform diff operations against `ADTs` efficiently:
- init
    - `AddressMap() (adt.Map, error)`: returns the address map as an `adt.Map`
    - `AddressMapBitWidth() int`: returns the bitwidth used by the address map
    - `AddressMapHashFunction() func(input []byte) []byte)`: returns the hash function used by the address map
- market
    - `DealProposals.ProposalArray()`: renames and exports the method (logic unchanged)
    - `DealProposals.Decode()`: exports the method (logic unchanged)
    - `DealProposalsAmtBitWidth() int`: returns the bitwidth used by the Proposals Array
    - `DealStatesAmtBitWidth() int`: returns the bitwidth used by the States Array
- miner
    - `SectorsArray() (adt.Array, error)`: exports the method (logic unchanged)
    - `SectorsAmtBitWidth() int`: returns the bitwidth used by the sectors array
    - `DecodeSectorOnChainInfo(*cbg.Deferred) (SectorOnChainInfo, error)`: exports the method (logic unchanged)
    - `PreCommitsMap() (adt.Map, error)`: exports the method (logic unchanged)
    - `PreCommitsMapBitWidth() int`: returns the bitwidth used by the PreCommits map
    - `PreCommitsMapHashFunction() func(input []byte) []byte`: returns the hash funciton used by the PreCommits map
    - `DecodeSectorPreCommitOnChainInfo(*cbg.Deferred) (miner8.SectorPreCommitOnChainInfo, error)`: exported (logic unchanged)
- power
    - I'm tired of typing this out, code changes similar to description above, export maps and their bitwidth and hash functions
- reward
    - ditto
- verifreg
    - ditto

## Additional Info
Might we consider lifting this entire package out of lotus? As a follow-on I'd like to add mocks for the actor states to make testing code that uses them easier.

## Checklist

- [ ] CI is green


frrist:
I'd like to have this in with or before v1.18.0 is cut, think that's doable?

chain/actors/builtin/init/actor.go.template
```diff
@@ -87,6 +87,28 @@ type State interface {
 	// Sets the address map for the init actor. This should only be used for testing.
 	SetAddressMap(mcid cid.Cid) error
 
-	AddressMap() (adt.Map, error)
 	GetState() interface{}
+
+	Code() cid.Cid
```
ZenGround0, line 92: Please add these for all the sad unwanted actors that you haven't touched.

chain/actors/builtin/miner/actor.go.template
```diff
@@ -102,12 +103,21 @@ type State interface {
...
-	// Diff helpers. Used by Diff* functions internally.
-	sectors() (adt.Array, error)
-	decodeSectorOnChainInfo(*cbg.Deferred) (SectorOnChainInfo, error)
-	precommits() (adt.Map, error)
-	decodeSectorPreCommitOnChainInfo(*cbg.Deferred) (miner{{.latestVersion}}.SectorPreCommitOnChainInfo, error)
 	GetState() interface{}
+
+	SectorsArray() (adt.Array, error)
+	SectorsAmtBitWidth() int
+	DecodeSectorOnChainInfo(*cbg.Deferred) (SectorOnChainInfo, error)
```
ZenGround0, line 110: Discussed offline I want all the things that need a decode method like this to just become top level interface types (like Deadline Partition below) that have the needed accessor interface to meet diffing needs.

ZenGround0: @magik6k do you think this is necessary?  How do you feel about merging without changing from this public Decode method?

chain/actors/builtin/init/state.go.template
```diff
@@ -119,10 +128,49 @@ func (s *state{{.v}}) SetAddressMap(mcid cid.Cid) error {
...
+        return func(input []byte) []byte {
+            res := sha256simd.Sum256(input)
+            return res[:]
+        }
+    {{- else}}
+        return func(input []byte) []byte {
+            res := sha256.Sum256(input)
+            return res[:]
+        }
+    {{- end}}
```
magik6k, line 175: are those somehow giving different hashes?

chain/actors/builtin/miner/actor.go.template
magik6k: I don't love this, but I see why we need this; Ideally we'd have a better/more consistent way to go through state at any version.

chain/actors/builtin/init/state.go.template
frrist: I would hope not - implemented this way since different actor versions used different packages

chain/actors/builtin/init/actor.go.template
frrist: done via: [a78b642](https://github.com/filecoin-project/lotus/pull/9126/commits/a78b64202ecb359d9144deb7a310cfdf3ec13277), [6d04195](https://github.com/filecoin-project/lotus/pull/9126/commits/6d041958106af625a8edb4ec9fcacbc11b1afa02), [9d0a2e6](https://github.com/filecoin-project/lotus/pull/9126/commits/9d0a2e673e7d8f2160bb38827fb21152f324c328), [9b7af83](https://github.com/filecoin-project/lotus/pull/9126/commits/9b7af83fa909a444d923814ce9abf6e44c7bffff)

chain/actors/builtin/miner/actor.go.template
frrist: Would you be willing to punt addressing this to a different PR? 😁 🙏 
 I don't love it either, and wrapping it with an interface as Zen suggested would certainly be better, but the work to do so is nontrivial.

chain/actors/builtin/init/state.go.template
```diff
@@ -119,10 +128,49 @@ func (s *state{{.v}}) SetAddressMap(mcid cid.Cid) error {
...
+
 func (s *state{{.v}}) AddressMap() (adt.Map, error) {
 	return adt{{.v}}.AsMap(s.store, s.State.AddressMap{{if (ge .v 3)}}, builtin{{.v}}.DefaultHamtBitwidth{{end}})
 }
 
-func (s *state{{.v}}) GetState() interface{} {
-	return &s.State
-}
\ No newline at end of file
+func (s *state{{.v}}) AddressMapBitWidth() int {
```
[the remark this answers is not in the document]
frrist: Ideas on how to make the implementation more compact?
in `chain/actors/adt/adt.go` I've defined:
```go
type MapHashFunc func([]byte) []byte

type MapOpts interface {
	BitWidth() int
	HashFunc() MapHashFunc
}
```
implementing the interface then looks something like:
```go
type mapOpts struct {
	hashFunc adt.MapHashFunc
	bitWidth func() int
}

func (m mapOpts) HashFunc() adt.MapHashFunc { return m.hashFunc }
func (m mapOpts) BitWidth() int             { return m.bitWidth() }

func (s *state2) AddressMap() (adt.Map, adt.MapOpts, error) {
	m, err := adt2.AsMap(s.store, s.State.AddressMap)
	if err != nil {
		return nil, nil, err
	}
	t := struct{ mapOpts }{}
	t.bitWidth = func() int { return 5 }
	t.hashFunc = func(input []byte) []byte {
		res := sha256.Sum256(input)
		return res[:]
	}
	return m, t, nil
}
```

I'd like to avoid redefining `mapOpts` everywhere it's needed to satisfy the `adt.MapOpts` interface; couple ideas:
1) make MapOpts a structure and drop the whole interface thing, may be annoying to work with later if we suspect there to be more options
2) keep the MapOpts interface and move the `mapOpts` structure somewhere in the `chain/actors` package
3) don't return a structure or an interface and instead define `AddressMap() (adt.Map, int, ad.MapHashFunc, error)`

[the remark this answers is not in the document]
ZenGround0: I didn't look too deeply at this but making MapOpts a struct seems great.  We can always come back and generalize and extend later but for now lets keep it simple

chain/actors/builtin/miner/actor.go.template
arajasek: I...would prefer not to punt this because I'm worried it'll never happen, and this is fairly ugly. I could help with the wrangling if that would make it easier.

arajasek (commented):
Thanks for the PR, and sorry about the 22 days it took me to get to reviewing it!

> Additionally, the State interfaces in the packages mentioned above receive the following new methods:
> Code() cid.Cid: returns the CID of the actor.
> ActorVersion() actors.Version: returns the version of the actor.
> ActorKey() string: returns the ActorKey (string) of the actor (used for getting its CID via GetActorCodeID.

I don't like this change, it smells wrong to me. A State interface / object doesn't (and shouldn't) know anything about the actor code / version / key. Anyone creating or using such a State must already have that information, so this feels backward to me.

Can we walk through how these methods are used (and whether we can avoid having them)?

> AddressMap() (adt.Map, error): returns the address map as an adt.Map

What is the change here? Does it not already do this?

> I'd like to have this in with or before v1.18.0 is cut, think that's doable?

Yes, I think so! Indeed, I think it'd be good to get this in before the meat of the work for v1.18.0 begins.

chain/actors/builtin/init/state.go.template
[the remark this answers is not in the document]
frrist: Great, I will leave this as is then.

chain/actors/builtin/miner/actor.go.template
frrist: I'd love help wrangling this as I am not sure what shape it should be in, would you mind doing that bit?

frrist:
> Thanks for the PR, and sorry about the 22 days it took me to get to reviewing it!

Well, it took me about the same amount of time to get back to this, so all good!

> I don't like this change, it smells wrong to me. A State interface / object doesn't (and shouldn't) know anything about the actor code / version / key. Anyone creating or using such a State must already have that information, so this feels backward to me.

It's required (in Lily) by methods that diff actor state internals. These methods typically take two State interfaces and performe a diff of their internal structures (like their HAMTs). For upgrade boundaries its important to know if the actors are the same version when performing the diff since their bitwidths are not constant. It's also used to define the new methods `AllCodes()` and `VersionCodes()`, which I would very much like to keep around as there isn't a canonical way (that I am aware of) to get a list of all the different actor CIDs. 

If we are to strictly follow the principle of: "Anyone creating or using such a State must already have that information" then the methods which expose the miners SectorBitWidth, for example, don't have a place here either. 

Overall how big of a blocker is adding these methods? I'd really like to keep them

>> AddressMap() (adt.Map, error): returns the address map as an adt.Map

> What is the change here? Does it not already do this?

The method is now exported, small change

> Yes, I think so! Indeed, I think it'd be good to get this in before the meat of the work for v1.18.0 begins.

Hope I am not to late for this 🙏

chain/actors/builtin/miner/actor.go.template
jennijuju: @frrist lotus team wont have bandwidth to do this until we ship v1.18.0 - so if this is something you need in this release, I’m afraid you have to take this on yourself 🙏

arajasek:
> Overall how big of a blocker is adding these methods? I'd really like to keep them

Yeah, we can keep them. My argument is based on distaste (and not "this will break things"), so it's not too important.

> > Yes, I think so! Indeed, I think it'd be good to get this in before the meat of the work for v1.18.0 begins.
> 
> Hope I am not to late for this pray

You're late in that the meat of the v1.18.0 work has begun (and landed)...But that just means you have a painful rebase to do. Let's get this on top of the `release/v1.18.0` branch, because I think that's where it'll get merged into.

jennijuju:
> Let's get this on top of the release/v1.18.0 branch, because I think that's where it'll get merged into.

We have discussed about it offline and have agreed this should land in master for a future feature release:
- `release/v1.18.0` has created with the right base, and this is NOT a consensus breaking change or a must have lotus refactor for nv17. Thus according to our mandatory release strategy and testing flow for a upgrade release, this should be out of scope for v1.18.0.
- this may land and target for the November feature release - v1.19.0. Code freeze targeted on Nov 9th, 2022.

frrist:
closed in favor of https://github.com/filecoin-project/lotus/pull/9667

Issue #7748, opened 2021-12-10
Expose mechanism to DeclareFaultsRecovered manually 


jennijuju:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

after faulting a sector, most of the time it takes SP less than 24 hours to recover the sector. As soon as they've done that, they should be able to manually declare the sector to be recovered .

### Describe the solution you'd like

expose an api to call [DeclareFaultsRecovered](https://github.com/filecoin-project/specs-actors/blob/4d19cab4cc26e65e97bdaecd3f8c781e52b586ca/actors/builtin/miner/miner_actor.go#L1662)

have a cli like `lotus-miner proving declare-recovery [sector id1........sector idN]`

jennijuju:
add api to lotus example https://github.com/filecoin-project/lotus/pull/6642/files

after adding this, run `make gen` will generate the api docs

Aloxaf:
Is there any advantage to declare recovered as soon as sectors are recovered? As I understand, WindowPost will not be submitted and SP's power won't be recovered, until the deadline is reached.

PR #9144, opened 2022-08-09
feat: proving: Introduce manual sector fault recovery

LexLuthr:
## Related Issues

https://github.com/filecoin-project/lotus/issues/7748

## Proposed Changes

Allow users to declare fault recovery manually rather than waiting for windowPost.

## Checklist

- [x] CI is green


jennijuju:
can we have a test?

LexLuthr:
> can we have a test?

Jen, a real world test or an itest? Because itest is already there.

storage/wdpost/wdpost_run_faults.go
```diff
@@ -343,3 +344,142 @@ func (s *WindowPoStScheduler) asyncFaultRecover(di dline.Info, ts *types.TipSet)
...
+	if err := s.prepareMessage(ctx, msg, spec); err != nil {
+		return nil, err
+	}
+	sm, err := s.api.MpoolPushMessage(ctx, msg, &api.MessageSendSpec{MaxFee: abi.TokenAmount(s.feeCfg.MaxWindowPoStGasFee)})
+	if err != nil {
+		return nil, xerrors.Errorf("pushing message to mpool: %w", err)
+	}
+
+	mcids = append(mcids, sm.Cid())
+	return mcids, nil
```
geoff-vball, line 484: This is mostly duplicated from lines 418-447. We should extract this core logic and just call it from both places.

node/impl/storminer.go
```diff
@@ -1299,6 +1299,25 @@ func (sm *StorageMinerAPI) ComputeProof(ctx context.Context, ssi []builtin.Exten
...
+func (sm *StorageMinerAPI) RecoverFault(ctx context.Context, sectors []abi.SectorNumber) ([]cid.Cid, error) {
+	allsectors, err := sm.Miner.ListSectors()
+	if err != nil {
+		return nil, xerrors.Errorf("could not get a list of all sectors from the miner: %w", err)
+	}
+	found := false
+	for _, v := range sectors {
+		for _, s := range allsectors {
+			if v == s.SectorNumber {
+				found = true
```
geoff-vball, line 1311: We should break out of the inner for-loop once we've found the sector.

cmd/lotus-miner/proving.go
```diff
@@ -644,3 +647,75 @@ It will not send any messages to the chain.`,
...
+				}
+
+				if wait.Receipt.ExitCode != 0 {
+					results <- xerrors.Errorf("Failed to execute message %s: %w", wait.Message, wait.Receipt.ExitCode.Error())
+				}
+				results <- nil
+			}(msg)
+		}
+
+		time.Sleep(time.Duration((cctx.Int("confidence")*30)+30) * time.Second)
```
geoff-vball, line 712: There should be a way to programmatically wait for all of our goroutines to finish instead of sleeping like this.

cmd/lotus-miner/proving.go
```diff
@@ -644,3 +647,75 @@ It will not send any messages to the chain.`,
...
+		if err != nil {
+			return err
+		}
+		defer closer()
+
+		api, acloser, err := lcli.GetFullNodeAPI(cctx)
+		if err != nil {
+			return err
+		}
+		defer acloser()
```
geoff-vball, line 687: nit: I would get this after the call to `RecoverFault`, so it is called right before it is needed

itests/wdpost_config_test.go
```diff
@@ -300,3 +300,118 @@ func TestWindowPostMaxSectorsRecoveryConfig(t *testing.T) {
...
+		}, false)
+	})
+	require.NoError(t, err)
+
+	// Try to manually recover the sector
+	t.Log("Send recovery message")
+	_, err = miner.RecoverFault(ctx, failedSectors)
+	require.NoError(t, err)
+
+	time.Sleep(30 * time.Second)
```
geoff-vball, line 408: This seems like an awfully long time to be sleeping in a test. We should probably be using `WaitTillChain` here as well.

itests/wdpost_config_test.go
```diff
@@ -300,3 +300,118 @@ func TestWindowPostMaxSectorsRecoveryConfig(t *testing.T) {
...
+
+	p, err = client.StateMinerPower(ctx, maddr, types.EmptyTSK)
+	require.NoError(t, err)
+
+	require.Equal(t, p.MinerPower, p.TotalPower)
+
+	sectors = p.MinerPower.RawBytePower.Uint64() / uint64(ssz)
+	require.Equal(t, nSectors+kit.DefaultPresealsPerBootstrapMiner, int(sectors))
```
geoff-vball, line 416: I know this is a huge nit but I think is important. We should avoid division when possible. Get the RBP and then compare is against (nSectors+kit.DefaultPresealsPerBootstrapMiner) * uint64(ssz)

geoff-vball (commented):
This is looking great!! Just a few comments. Let me know if you have any questions :)

cmd/lotus-miner/proving.go
LexLuthr: I have switched to a WaitGroup here. It would make sense to remove confidence flag if we are using a WaitGroup.
Dropping confidence.

itests/wdpost_config_test.go
LexLuthr: I have removed this check as part of rewriting the itest

itests/wdpost_config_test.go
LexLuthr: I have made changes to the itest to compare the correct values. Thanks for pointing out the 30 second delay as wdPost was happening in these 30 secs.
This was leading to power getting restored. New test compares the fault sector count and recovery sector counts.

storage/wdpost/wdpost_run_faults.go
```diff
@@ -343,3 +344,121 @@ func (s *WindowPoStScheduler) asyncFaultRecover(di dline.Info, ts *types.TipSet)
...
+			Partition: i.partition,
+			Sectors:   sectorinbit,
+		})
+	}
+
+	// Batch if maxPartitionsPerRecoveryMessage is set
+	if s.maxPartitionsPerRecoveryMessage > 0 {
+		tmpRecoveryDecls := RecoveryDecls
+		var RecoveryBatches [][]miner.RecoveryDeclaration
+		I := 0
```
geoff-vball, line 403: Do we use `I` somewhere? I might just be missing it

storage/wdpost/wdpost_run_faults.go
```diff
@@ -343,3 +344,121 @@ func (s *WindowPoStScheduler) asyncFaultRecover(di dline.Info, ts *types.TipSet)
...
+		for _, Batch := range RecoveryBatches {
+			msg, err := s.manualRecoveryMsg(ctx, Batch)
+			if err != nil {
+				return nil, err
+			}
+
+			mcids = append(mcids, msg)
+		}
+
+		return mcids, nil
```
geoff-vball, line 425: I think we should pull this out of the `if` statement it's in and only have one return location for this function. Both code paths should populate `RecoveryBatches` and then call this for loop. Feel free to reach out if this doesn't make sense!

storage/wdpost/wdpost_run_faults.go
```diff
@@ -343,3 +344,112 @@ func (s *WindowPoStScheduler) asyncFaultRecover(di dline.Info, ts *types.TipSet)
...
+			Batch := tmpRecoveryDecls[len(tmpRecoveryDecls)-s.maxPartitionsPerRecoveryMessage:]
+			tmpRecoveryDecls = tmpRecoveryDecls[:len(tmpRecoveryDecls)-s.maxPartitionsPerPostMessage]
+			RecoveryBatches = append(RecoveryBatches, Batch)
+		}
+
+		// Add remaining as new batch
+		RecoveryBatches = append(RecoveryBatches, tmpRecoveryDecls)
+	}
+
+	RecoveryBatches = append(RecoveryBatches, RecoveryDecls)
```
geoff-vball, line 415: I think this needs to go in an `else` statement with the `if` above. We don't want to execute this after we've done all the batches.

storage/wdpost/wdpost_run_faults.go
```diff
@@ -343,3 +344,112 @@ func (s *WindowPoStScheduler) asyncFaultRecover(di dline.Info, ts *types.TipSet)
...
+		RecoveryDecls = append(RecoveryDecls, miner.RecoveryDeclaration{
+			Deadline:  i.deadline,
+			Partition: i.partition,
+			Sectors:   sectorinbit,
+		})
+	}
+
+	// Batch if maxPartitionsPerRecoveryMessage is set
+	if s.maxPartitionsPerRecoveryMessage > 0 {
+		tmpRecoveryDecls := RecoveryDecls
```
geoff-vball, line 402: This operation doesn't actually copy the slice. This will just give you a pointer to the same slice. That being said, with how we've changed the code now, you don't need a temporary array at all, you can just manipulate `RecoveryDecls` directly in this block.

geoff-vball (commented):
Thanks so much for all the quick fixes! Just a few more small changes needed and we can get this merged!

geoff-vball:
Just to update the status of this ticket. @Reiers did some testing, found some small issues. @LexLuthr has fixed them, but we're waiting for @Reiers to come back from OOO to retest.

LexLuthr:
@Reiers How long does the command stays stuck if you don't kill it?

geoff-vball:
@Reiers I did some quick testing and it looks like closing the channel (current version) works for me. Can you give it another try when you get a chance?

Reiers:
Sure! I will test first thing tmr.

Reiers:
Closing the channel did the trick - no issues.

geoff-vball (approved):

PR #9168, opened 2022-08-14
bump eventsource from 1.0.7 to 1.1.2 in /lotuspond/front

bhaskarvilles:
An Upgrade to the eventsource in the /lotuspod/front

Signed-off-by: Bhaskar Ram <bhaskarvilles@duck.com>


magik6k:
As already mentioned in https://github.com/filecoin-project/lotus/pull/8313#issuecomment-1066713752, we're not accepting random updates to npm packages which do not need upgrading.

PR #9208, opened 2022-08-24
api: ethrpc: implement a few ethereum JSON-RPC API methods

ychiaoli18:
## Related Issues
https://github.com/filecoin-project/FIPs/discussions/422

Partly solves https://github.com/filecoin-project/ref-fvm/issues/854.

## Proposed Changes
implemented the following methods
* eth_blockNumber
* eth_accounts
* eth_getBlockTransactionCountByNumber
* eth_getBlockTransactionCountByHash
* eth_getBalance
* net_listening
* net_version

## Checklist

- [ ] CI is green


node/rpc.go
```diff
@@ -73,6 +73,10 @@ func FullNodeHandler(a v1api.FullNode, permissioned bool, opts ...jsonrpc.Server
 		rpcServer.Register("Filecoin", hnd)
 		rpcServer.AliasMethod("rpc.discover", "Filecoin.Discover")
 
+		rpcServer.AliasMethod("eth_accounts", "Filecoin.EthAccounts")
+		rpcServer.AliasMethod("eth_blockNumber", "Filecoin.EthBlockNumber")
+		rpcServer.AliasMethod("eth_getBlockTransactionCountByNumber", "Filecoin.EthGetBlockTransactionCountByNumber")
```
magik6k, line 78: This aliasing is fine, but could be made a little easier (to maintain/document) by:
* Adding 'alias' tags in `api/` - for example `EthAccounts(context.Context) ([]string, error) //perm:read alias:eth_accounts`
* Allowing the 'alias' tag in https://github.com/filecoin-project/lotus/blob/master/gen/api/proxygen.go#L262
* Using reflect to go through fields of the generated [api proxy struct](https://github.com/filecoin-project/lotus/blob/master/api/proxy_gen.go#L104-L109) to do the aliasing here automatically

The few reasons to do that would be:
* Adding eth_ methods doesn't require going through multiple files; Aliases are in the same file as the API interface
* It's more self-documenting, and easier to add support to api-doc generation (api/docgen)
* Arguably easier to maintain (everything is in one place)

(this is just a suggestion, I'm fine with doing things the current way as well)

node/impl/full/eth.go
```diff
@@ -0,0 +1,76 @@
...
+func (a *EthModule) EthAccounts(context.Context) ([]string, error) {
+	// The lotus node is not expected to hold manage accounts, so we'll always return an empty array
+	return []string{}, nil
+}
+
+func (a *EthModule) EthGetBlockTransactionCountByNumber(ctx context.Context, blkNumHex string) (string, error) {
+	blkNum, err := strconv.ParseInt(strings.Replace(blkNumHex, "0x", "", -1) , 16, 64)
+	if err != nil {
+		return "0x0", xerrors.Errorf("invalid block number %s: %w", blkNumHex, err)
+	}
```
magik6k, line 59: You can make go-jsonrpc handle eths 0x param stuff for you:
* Define a type like `type ethInt int64`
* Implement `json.[Un]Marshaler` on that type
* Use that type in method params instead of `string`

documentation/en/api-v0-methods-miner.md
```diff
@@ -1087,7 +1087,7 @@ Response:
         }
       },
       "Diagnostics": [
-        "string value"
```
raulk, line 1090: We'll need to introduce the Ethereum address type even if it's just an alias for now, to avoid example mangling.

node/rpc.go
raulk: Agree!

@ychiaoli18 For now I'd suggest we focus on buildling out the functionality, and later we can remove all the aliases from here and make them annotation-driven, as @magik6k proposes.

ychiaoli18: Agree! I'll mark these as TODO so we can revisit after completing all the functionalities in the next milestones!

documentation/en/api-v0-methods-miner.md
ychiaoli18: introduced a EthAddress type

node/impl/full/eth.go
ychiaoli18: introduced EthInt

node/impl/full/eth.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (a *EthModule) EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNumHex string, txIndexHex string) (api.EthTx, error) {
+	return api.EthTx{}, nil
+}
+
+// EthGetCode returns string value of the compiled bytecode
+func (a *EthModule) EthGetCode(ctx context.Context, address string) (string, error) {
+	return "", nil
+}
```
raulk, line 150: Easy way to get this is to perform the equivalent of an eth_call to the GetBytecode method upcoming here: https://github.com/filecoin-project/ref-fvm/issues/838

node/impl/full/eth.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (a *EthModule) EthGetStorageAt(ctx context.Context, address string, positionHex string, blkParam string) (string, error) {
+	return "", nil
+}
```
raulk, line 154: Will need to delve into the state object, interpret it, and extract the storage key. Alternatively, we could add a `GetStorageAt` method to the EVM actor, but repeated calls to `EthGetStorageAt` would end up instantiating a Wasm container every time just to extract the storage value.

node/impl/full/eth.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return api.EthBigInt{}, xerrors.Errorf("cannot parse address: %w", err)
+	}
+
+	filAddr, err := addr.ToFilecoinAddress()
+	if err != nil {
+		return api.EthBigInt{}, err
+	}
+	balance, err := a.WalletAPI.WalletBalance(ctx, filAddr)
+	return api.EthBigInt{balance.Int}, nil
```
raulk, line 167: This wouldn't be from the wallet, but rather from the state tree.

node/impl/full/eth.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (a *EthModule) EthChainId(ctx context.Context) (api.EthInt, error) {
+	return api.EthInt(0), nil
+}
```
raulk, line 172: https://github.com/filecoin-project/ref-fvm/issues/681

node/impl/full/eth.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (a *EthModule) NetVersion(ctx context.Context) (string, error) {
+	// Note that networkId is not encoded in hex
+	return "1", nil
+}
```
raulk, line 177: Let's return the Filecoin network version?

raulk (commented):
This is a pretty good starting point. Just two blocking things:

1. We need to straighten out the gomock situation here (I don't think these files need to be committed).
2. EthGetBalance => need to return the balance from the state tree.

Other than that, I'm happy to merge this into the branch and iterate on it.

api/api_full.go
```diff
@@ -738,6 +738,37 @@ type FullNode interface {
...
+	EthBlockNumber(ctx context.Context) (EthInt, error) //perm:read
+	// EthGetBlockTransactionCountByNumber returns the number of messages in the TipSet
+	EthGetBlockTransactionCountByNumber(context.Context, string) (EthInt, error) //perm:read
+	// EthGetBlockTransactionCountByHash returns the number of messages in the TipSet
+	EthGetBlockTransactionCountByHash(context.Context, string) (EthInt, error) //perm:read
+
+	EthGetBlockByHash(ctx context.Context, blkHash string, fullTxInfo bool) (EthBlock, error)                       //perm:read
+	EthGetBlockByNumber(ctx context.Context, blkNumHex string, fullTxInfo bool) (EthBlock, error)                   //perm:read
+	EthGetTransactionByHash(ctx context.Context, txHash string) (EthTx, error)                                      //perm:read
+	EthGetTransactionCount(ctx context.Context, sender string, blkOpt string) (EthInt, error)                       //perm:read
```
magik6k, line 756: If you also implement [UnmarshalJSON](https://pkg.go.dev/encoding/json#Unmarshaler) on Eth* types you should be able to use them for parameters as well

ychiaoli18: addressed comment

raulk (approved):
gencheck and tests are failing, but those fail on the base branch right now and we're choosing to ignore them temporarily.

PR #9212, opened 2022-08-24
feat: shed: add a --max-size flag to vlog2car

magik6k:
Moderately useful

### DIFF
--- a/cmd/lotus-shed/datastore-vlog.go
+++ b/cmd/lotus-shed/datastore-vlog.go
@@ -34,12 +34,13 @@ var datastoreVlog2CarCmd = &cli.Command{
 			Usage: "datastore prefix",
 			Value: "/blocks/",
 		},
+		&cli.Uint64Flag{Name: "max-size", Value: 32000},
 	},
 	ArgsUsage: "[vlog...]",
 	Action: func(cctx *cli.Context) error {
 		ctx := cctx.Context
 
-		maxSz := uint64(31 << 30)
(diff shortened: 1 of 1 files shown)


cmd/lotus-shed/datastore-vlog.go
```diff
@@ -34,12 +34,13 @@ var datastoreVlog2CarCmd = &cli.Command{
 			Usage: "datastore prefix",
 			Value: "/blocks/",
 		},
+		&cli.Uint64Flag{Name: "max-size", Value: 32000},
```
geoff-vball, line 37: We should add a `usage` that says `max-size` is in MiB

cmd/lotus-shed/datastore-vlog.go
```diff
@@ -34,12 +34,13 @@ var datastoreVlog2CarCmd = &cli.Command{
...
 	},
 	ArgsUsage: "[vlog...]",
 	Action: func(cctx *cli.Context) error {
 		ctx := cctx.Context
 
-		maxSz := uint64(31 << 30)
+		maxSz := uint64(cctx.Uint64("max-size") << 20)
```
geoff-vball, line 43: Are we able convert a blockstore bigger than 31 << 30? If not, we should add a check here to make sure the input is less than 31 << 10

magik6k: This can can do that, it's just usually not useful as in most cases you'll want your cars to be 32/64ish GiB

PR #9259, opened 2022-09-02
fix: ipld: restore e2e online deal and retrievals w/ identity CIDs

rvagg:
## Related Issues

Depends on:
* https://github.com/ipld/go-car/pull/332
* https://github.com/filecoin-project/go-fil-markets/pull/749

Related:
* https://github.com/filecoin-project/boost/pull/715
* https://github.com/filecoin-project/go-fil-markets/pull/747

## Proposed Changes

This used to work prior to the CARv2 blockstore interface was introduced to back
Graphsync retrievals and client retrieval CAR export. When that was done, the
StoreIdentityCIDs option was omitted in some key places, and the CARv2
blockstore interface always returned true from Has() for identity CIDs which
prevents Graphsync from storing them.

Since launch, car.SelectiveCar has been the canonical way to produce CARs for
online deals to calculate CommP. This is still being done on the client side
in Lotus for online deals, but it's no longer done on the server side, and
neither is it done for retrieval export on the client side.

## Additional Info

As per discussion in https://github.com/ipld/go-car/pull/332, we do have the
option of changing online deals to _not_ include identity CIDs, therefore
ditching use of car.SelectiveCar (replacing with a CARv2 alternative); but that
would be a breaking change—like the blockstore changes were a breaking change,
but that was accidental. There's also other places where car.SelectiveCar
is simulated, e.g. node/impl/client/client.go#outputCAR.

In addition, I think we should probably be documenting how identity CIDs are handled throughout Lotus - for the online deal flow as well as for retrievals (https://github.com/filecoin-project/go-fil-markets/pull/747 is another example of problems with identity CIDs that is only indirectly related to this PR).

## Checklist

- [ ] CI is green


rvagg:
Sorry, I forgot to mention the second commit in here is related but not directly—undoing the identity CID option for unixfs creation on `lotus client import` since it's one of the major sources of our problems—i.e. we wouldn't have to deal with any of this if there weren't identity CIDs being generated and stored in Filecoin.

go.mod
```diff
@@ -37,7 +37,7 @@ require (
 	github.com/filecoin-project/go-data-transfer v1.15.2
 	github.com/filecoin-project/go-fil-commcid v0.1.0
 	github.com/filecoin-project/go-fil-commp-hashhash v0.1.0
-	github.com/filecoin-project/go-fil-markets v1.24.0
+	github.com/filecoin-project/go-fil-markets v1.24.1-0.20220902112703-c0372f17c119
```
magik6k, line 40: Needs to be tagged before we merge to master (+the related PR needs merging)

lib/unixfs/filestore.go
```diff
@@ -34,11 +33,7 @@ func CidBuilder() (cid.Builder, error) {
 		return nil, fmt.Errorf("failed to initialize UnixFS CID Builder: %w", err)
 	}
 	prefix.MhType = DefaultHashFunction
-	b := cidutil.InlineBuilder{
-		Builder: prefix,
-		Limit:   126,
-	}
-	return b, nil
+	return prefix, nil
```
magik6k, line 36: Not sure if we should include this change in this PR; Yes, there were bugs in ID cid handling, but that doesn't negate the fact that it's an useful feature.

rvagg: I'm proposing that we don't do this in Lotus and don't really encourage it for Filecoin, especially for the online deal flow. For two reasons:

1. In Filecoin, there's a relationship between a block and the piece it's stored in. This is complicated for identity CIDs and is a problem when we have retrieval modes where you're able to ask for an arbitrary (not necessarily deal root) "payload CID" and have to figure out what piece it's in so in order to make decisions about availability and price. Illustrated by the difficulty in dealing with identity CIDs as roots in https://github.com/filecoin-project/go-fil-markets/pull/747 (this is not really a "fix" for that problem, just a patch to make it workable, it'd not be a problem if we didn't have identity CIDs that can fit CIDs) but the same applies with non-root identity CIDs unless we make sure to store them as blocks and index them as blocks, thereby reducing their utility and increasing the complexity on all of the layers we manage. This gets even more complicated when we index the content in a piece and send that to an indexer. Should the indexer ignore them? Does the CID:Piece relationship matter at that level?
2. The break-even cost of storing inline CIDs is pretty low with the way we're having to store them. For an online deal, we end up at only 32-bytes break-even and that's without counting the extra CARv2 index that's generated (I think the break-even for SPs with the CARv2 index is just ~20-bytes, far from the 126 max default here).

Screenshot 2022-09-05 at 8 54 58 pm

(I wrote about this in more detail [over here](https://github.com/multiformats/multihash/issues/130))

So, maybe we just avoid it entirely? At least in lotus—we can't stop them entirely but we can set best practice.

magik6k: I see. Yeah avoiding those is probably reasonable.

I wonder if we could allow them, but only for leaf nodes, where we wouldn't need to worry about traversal edge-cases.

rvagg: Maybe, but in the current form `CidBuilder` only gets the bytes so it'd need to decode them to look for links to determine that: https://github.com/ipfs/go-cidutil/blob/ac549eaa2ca06b13a1e5b37ee2696cd72b6481a7/inline.go#L21-L26, that's probably not very efficient.

Alternatively we could introduce an optional `LeafCidBuilder` into `DagBuilderHelper` and special-case use it when it's set for leaves: https://github.com/ipfs/go-unixfs/blob/2c23c3ea6fae3ef1b487cfc0c606a4ffc7893676/importer/helpers/dagbuilder.go#L154

rvagg:
@magik6k @jennijuju I know we're dealing with some deprecated components here, but boost is still relying on some parts of lotus this touches, particularly the itest kit, and I'm running into this problem over there! Which is what reminded me I still had this open.

This is updated now to latest `master` and should be safe to merge. The change set is much more minimal now that `master` has evolved past some of the original dependencies I had to update for this. It would be great to have this merged so we can finally close this PR.

PR #9260, opened 2022-09-02
M2: Update FVM

Stebalien:
Update the FVM in experimental/m2:

- Update the FVM.
- Add gas tracing support (cherry-picked from #9134).
- Add support for configuring debugging through build params. Previously, this was hard-coded in the experimental/m2 FFI.

This also switches the FFI branch to feat/m2 from experimental/m2. cc @vyzo could you take a look at the diff?


build/params_2k.go
```diff
@@ -22,6 +22,7 @@ const GenesisFile = ""
 // VERY experimental! Should not be merged into master / any release branch yet!
 var NetworkBundle = "devnet-wasm"
 var BundleOverrides map[actors.Version]string
+var ActorDebugging = false
```
vyzo, line 25: we probably want this to be true so that peoplr can get logs and backtraces.

build/params_wallaby.go
```diff
@@ -17,6 +17,7 @@ import (
 
 var NetworkBundle = "wallaby"
 var BundleOverrides map[actors.Version]string
+var ActorDebugging = true
```
vyzo, line 20: Should probably call out that this is changing from true to false.

vyzo (approved):
Lets keep the 2k actor debugging default enabled so that people can debug their actors.

Other than that, LGTM.

build/params_wallaby.go
raulk: Worth indicating in the handoff to Wallaby that we're enabling this feature now so that node operators expect the extra output in logs?

raulk (approved):

Issue #8664, opened 2022-05-17
lotus-shed terminate sectors fails with Multisig Owner


marshyonline:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### Lotus Version

```text
lotus-shed version 1.15.3-rc1
```

### Describe the Bug

When a miner is using a MultiSig Owner - The `lotus-shed terminate sectors` command will fail with

```/lotus# ./lotus-shed sectors terminate --really-do-it 79563
2022-05-17T18:37:03.800+1000    WARN    lotus-shed      lotus-shed/main.go:106  mpool push message:
    main.glob..func92
        /root/workspace/lotus/cmd/lotus-shed/sectors.go:150
[...]

PR #9291, opened 2022-09-12
feat: cli: Add option to terminate sectors from worker address

geoff-vball:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/8664

## Proposed Changes
Adds a flag to allow either owner address or worker address to send terminate sectors message.

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus-shed/sectors.go
+++ b/cmd/lotus-shed/sectors.go
@@ -58,6 +58,10 @@ var terminateSectorCmd = &cli.Command{
 			Name:  "really-do-it",
 			Usage: "pass this flag if you know what you are doing",
 		},
+		&cli.StringFlag{
+			Name:  "from",
+			Usage: "specify the address to send the terminate message from",
+		},
 	},
 	Action: func(cctx *cli.Context) error {
 		if cctx.Args().Len() < 1 {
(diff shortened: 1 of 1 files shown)


arajasek:
Was this change motivated by someone wanting it?

Regardless, I think this would still be incomplete, because you can also use control addresses to send this message. I'd suggest just changing it to an optional `from` flag, and using that address if set. If unset, I'd probably default to using the worker.

geoff-vball:
@arajasek added the related ticket. What's a control address? Changes have been made

arajasek (approved):

arajasek:
@geoff-vball "Control addresses" are a set of addresses that can control a miner. They are significantly less powerful than the owner, and have a similar privilege level to the worker with some key differences:
- you can have a set of control addresses, but only one worker
- they can be either secp or bls (and maybe even, like, multisigs), while the worker must be BLS
- only the worker can sign block headers.

The idea is just to have more than one address that can do things like submit posts, seal sectors, etc...so if one key is inaccessible / out of funds, a fallback control address can Do the Thing.

See the miner actor source code for more.

Issue #9130, opened 2022-08-05
Support multiple chain nodes per miner node


magik6k:
Currently each lotus-miner instance can only be backed by a single lotus chain node. This is obviously not ideal as there is no redundancy for chain operations - if the chain node dies, the miner node will become unusable until the chain node is fixed or switched.

This can be done in two ways:
* Make it possible to attach multiple chain nodes to a single miner, balance api calls between them
  * This needs https://github.com/filecoin-project/lotus/issues/9129 to work well
* Same as above, but use one node, and only switch when that node dies
  * Even in this case we most likely want https://github.com/filecoin-project/lotus/issues/9129 to avoid weird races on nonces

The need for https://github.com/filecoin-project/lotus/issues/9129 could be mitigated by requiring use of separate control addresses on each node, but that makes setup harder, and requires a lot of on-chain accounts, which is not ideal.

magik6k:
Whichever way we go, we likely want to start by writing a helper which takes multiple API connections - basically a method accepting multiple api interfaces and outputting one [api proxy](https://github.com/filecoin-project/lotus/blob/master/api/proxy_gen.go).

We already have some code doing similar transforms on the API interfaces:
* For auth - https://github.com/filecoin-project/go-jsonrpc/blob/master/auth/auth.go#L34; https://github.com/filecoin-project/lotus/blob/master/api/permissioned.go#L22
* For metrics - https://github.com/filecoin-project/lotus/blob/edcb28737c4f2478676bf2dbb065dd7e9876cdc0/metrics/proxy/proxy.go#L13

Then we just need to make it possible to connect to multiple chain nodes in https://github.com/filecoin-project/lotus/blob/master/cmd/lotus-miner/run.go#L77 using that helper

Fatman13:
Venus team have a proposal to approach this problem too. Comments and questions are welcomed!

Proposal:
- [ ] https://github.com/filecoin-project/devgrants/issues/770

PR #9294, opened 2022-09-12
WIP: Raft consensus for lotus nodes in a cluster

shrenujbansal:
## Related Issues

https://github.com/filecoin-project/lotus/issues/9130

## Proposed Changes

Adding raft consensus to lotus nodes to maintain consistent state for nonces and messages being published to the lotus node

## Checklist

- [ ] CI is green


itests/kit/node_full.go
```diff
@@ -29,9 +37,21 @@ type TestFullNode struct {
 	ListenAddr multiaddr.Multiaddr
 	DefaultKey *key.Key
 
+	//Libp2p struct {
```
shrenujbansal, line 40: Remove this

itests/raft_messagesigner_test.go
```diff
@@ -0,0 +1,238 @@
...
+	require.NoError(t, err)
+	mLookup, err := node0.StateWaitMsg(ctx, smHalfBal.Cid(), 3, api.LookbackNoLimit, true)
+	require.NoError(t, err)
+	require.Equal(t, exitcode.Ok, mLookup.Receipt.ExitCode)
+
+	rstate0 := getRaftState(ctx, t, &node0)
+	rstate1 := getRaftState(ctx, t, &node1)
+	rstate2 := getRaftState(ctx, t, &node2)
+
+	require.True(t, reflect.DeepEqual(rstate0, rstate1))
```
magik6k, line 143: I believe you can use `require.EqualValues(t, rstate0, rstate1)`, which uses reflect.DeepEqual under the hood (but also prints a nice diff when values don't match)

itests/raft_messagesigner_test.go
```diff
@@ -0,0 +1,238 @@
...
+	newLeader := leader
+	for _, n := range nodes {
+		if n != leaderNode {
+			newLeader, err = n.RaftLeader(ctx)
+			require.NoError(t, err)
+			require.NotEqual(t, newLeader, leader)
+		}
+	}
+
+	require.NotEqual(t, newLeader, leader)
```
magik6k, line 212: Would be better to use `require.Eventually` which in most cases will be faster, and gets better error messages when it fails

itests/self_sent_txn_test.go
```diff
@@ -26,7 +26,7 @@ func TestSelfSentTxnV15(t *testing.T) {
 
 	kit.QuietMiningLogs()
 
-	client15, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.GenesisNetworkVersion(network.Version15))
+	client15, _, ens := kit.EnsembleMinimal(t, kit.MockProofs()) //, kit.GenesisNetworkVersion(network.Version15))
```
magik6k, line 29: Likely want to undo this change.

lib/consensus/raft/config.go
```diff
@@ -0,0 +1,319 @@
...
+	DefaultDataSubFolder        = "raft"
+	DefaultWaitForLeaderTimeout = 15 * time.Second
+	DefaultCommitRetries        = 1
+	DefaultNetworkTimeout       = 100 * time.Second
+	DefaultCommitRetryDelay     = 200 * time.Millisecond
+	DefaultBackupsRotate        = 6
+	DefaultDatastoreNamespace   = "/r" // from "/raft"
+)
+
+// Config allows to configure the Raft Consensus component for ipfs-cluster.
```
magik6k, line 28: We'll want to remove all this commented-out code before merging

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,563 @@
...
+
+func newRaftState() RaftState {
+	return RaftState{NonceMap: make(map[addr.Address]uint64),
+		MsgUuids: make(map[uuid.UUID]*types.SignedMessage)}
+}
+
+type ConsensusOp struct {
+	Nonce     uint64               `codec:"nonce,omitempty"`
+	Uuid      uuid.UUID            `codec:"uuid,omitempty"`
+	Addr      addr.Address         `codec:"addr,omitempty"`
```
magik6k, line 43: Don't we have the address/nonce in the SignedMsg?

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,563 @@
...
+	libp2praft "github.com/libp2p/go-libp2p-raft"
+	host "github.com/libp2p/go-libp2p/core/host"
+	peer "github.com/libp2p/go-libp2p/core/peer"
+)
+
+var logger = logging.Logger("raft")
+
+type RaftState struct {
+	NonceMap map[addr.Address]uint64
+	MsgUuids map[uuid.UUID]*types.SignedMessage
```
magik6k, line 32: Did we decide if we want to use persistent raft state? If yes this will need to live in the datastore - possibly using https://github.com/filecoin-project/go-statestore (as an aside, we probably can make statestore a lot nicer to use with generics)

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,563 @@
...
+	cc.cancel()
+	close(cc.rpcReady)
+	return nil
+}
+
+// SetClient makes the component ready to perform RPC requets
+func (cc *Consensus) SetClient(c *rpc.Client) {
+	cc.rpcClient = c
+	cc.rpcReady <- struct{}{}
+}
```
magik6k, line 248: We should be able to remove this method, and provide the `rpc.Client` directly to `NewConsensus` (which would also get rid of `NewConsensusWithRPCClient` and the `rpcReady` channel)

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,563 @@
...
+// IsTrustedPeer returns true. In Raft we trust all peers.
+func (cc *Consensus) IsTrustedPeer(ctx context.Context, p peer.ID) bool {
+	return true
+}
+
+// Trust is a no-Op.
+func (cc *Consensus) Trust(ctx context.Context, pid peer.ID) error { return nil }
+
+// Distrust is a no-Op.
+func (cc *Consensus) Distrust(ctx context.Context, pid peer.ID) error { return nil }
```
magik6k, line 268: Can we remove this?

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,563 @@
...
+	}
+	return finalErr
+}
+
+// RaftState retrieves the current consensus RaftState. It may error if no RaftState has
+// been agreed upon or the state is not consistent. The returned RaftState is the
+// last agreed-upon RaftState known by this node. No writes are allowed, as all
+// writes to the shared state should happen through the Consensus component
+// methods.
+func (cc *Consensus) State(ctx context.Context) (consensus.State, error) {
```
magik6k, line 449: Don't we want to return `RaftState` instead of `consensus.State` here?

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,563 @@
...
+	//_, span := trace.StartSpan(ctx, "consensus/RaftState")
+	//defer span.End()
+
+	st, err := cc.consensus.GetLogHead()
+	if err == libp2praft.ErrNoState {
+		fmt.Println("Err no state return")
```
magik6k, line 455:
```suggestion
```

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,563 @@
...
+//	if !snapExists {
+//		return st, nil
+//	}
+//
+//	err = st.Unmarshal(r)
+//	if err != nil {
+//		return nil, err
+//	}
+//	return st, nil
+//}
```
magik6k, line 563:
```suggestion
```

lib/consensus/raft/interfaces.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// Component represents a piece of ipfscluster. Cluster components
+// usually run their own goroutines (a http server for example). They
+// communicate with the main Cluster component and other components
+// (both local and remote), using an instance of rpc.Client.
+type Component interface {
+	SetClient(*rpc.Client)
+	Shutdown(context.Context) error
+}
```
magik6k, line 18: We probably don't need this to be a separate interface

lib/consensus/raft/raft.go
```diff
@@ -0,0 +1,700 @@
...
+func (rw *raftWrapper) makeTransport() (err error) {
+	logger.Debug("creating libp2p Raft transport")
+	rw.transport, err = p2praft.NewLibp2pTransport(
+		rw.host,
+		rw.config.NetworkTimeout,
+	)
+	return err
+}
+
+func (rw *raftWrapper) makeStores() error {
```
magik6k, line 128: TBD if we want to use persistent stores here.

node/builder.go
```diff
@@ -121,6 +121,7 @@ const (
 	SettlePaymentChannelsKey
 	RunPeerTaggerKey
 	SetupFallbackBlockstoresKey
+	RPCServer
```
magik6k, line 124: I'd be specific so that this isn't confused as having anything to do with JsonRPC

```suggestion
	GoRPCServer
```

node/builder_chain.go
```diff
@@ -236,6 +239,15 @@ func ConfigFullNode(c interface{}) Option {
 			Unset(new(*wallet.LocalWallet)),
 			Override(new(wallet.Default), wallet.NilDefault),
 		),
+		// Chain node cluster enabled
+		If(cfg.Raft.ClusterModeEnabled,
+			Override(new(*gorpc.Client), modules.NewRPCClient),
+			Override(new(*consensus.Consensus), consensus2.NewConsensusWithRPCClient(false)),
+			Override(new(*messagesigner.MessageSignerConsensus), messagesigner.NewMessageSignerConsensus),
+			Override(new(messagesigner.MsgSigner), func(ms *messagesigner.MessageSignerConsensus) *messagesigner.MessageSignerConsensus { return ms }),
```
magik6k, line 247: Can use the `From` helper

```suggestion
			Override(new(messagesigner.MsgSigner), From(new(*messagesigner.MessageSignerConsensus))),
```

node/config/def.go
```diff
@@ -274,3 +278,35 @@ func (dur Duration) MarshalText() ([]byte, error) {
...
 }
+
+var (
+	DefaultDataSubFolder        = "raft"
+	DefaultWaitForLeaderTimeout = 15 * time.Second
+	DefaultCommitRetries        = 1
+	DefaultNetworkTimeout       = 100 * time.Second
+	DefaultCommitRetryDelay     = 200 * time.Millisecond
+	DefaultBackupsRotate        = 6
+	DefaultDatastoreNamespace   = "/r" // from "/raft"
```
magik6k, line 289: I'd use full `/raft`
```suggestion
	DefaultDatastoreNamespace   = "/raft"
```

node/config/types.go
```diff
@@ -590,3 +595,39 @@ type Wallet struct {
...
+	// failing an operation.
+	WaitForLeaderTimeout time.Duration
+	// NetworkTimeout specifies how long before a Raft network
+	// operation is timed out
+	NetworkTimeout time.Duration
+	// CommitRetries specifies how many times we retry a failed commit until
+	// we give up.
+	CommitRetries int
+	// How long to wait between retries
+	CommitRetryDelay time.Duration
```
magik6k, line 621: There's a special `Duration` type in the config package which serializes to toml better
```suggestion
	WaitForLeaderTimeout Duration
	// NetworkTimeout specifies how long before a Raft network
	// operation is timed out
	NetworkTimeout Duration
	// CommitRetries specifies how many times we retry a failed commit until
	// we give up.
	CommitRetries int
	// How long to wait between retries
	CommitRetryDelay Duration
```

node/config/types.go
```diff
@@ -590,3 +595,39 @@ type Wallet struct {
...
+	// BackupsRotate specifies the maximum number of Raft's DataFolder
+	// copies that we keep as backups (renaming) after cleanup.
+	BackupsRotate int
+	// Namespace to use when writing keys to the datastore
+	DatastoreNamespace string
+
+	// A Hashicorp Raft's configuration object.
+	RaftConfig *hraft.Config
```
magik6k, line 629: We likely want to "unpack" this config here

node/modules/rpc.go
```diff
@@ -0,0 +1,66 @@
...
+
+type RPCHandler struct {
+	mpoolAPI full.MpoolAPI
+	cons     *consensus.Consensus
+}
+
+//type ConsensusRPCAPI struct {
+//	cons       *consensus.Consensus
+//	rpcHandler *RPCHandler
+//}
```
magik6k, line 25:
```suggestion
```

node/modules/rpc.go
```diff
@@ -0,0 +1,66 @@
...
+	protocolID := protocol.ID("/p2p/rpc/ping")
+	return rpc.NewClient(host, protocolID)
+}
+
+func NewRPCServer(host host.Host, rpcHandler *RPCHandler) error {
+	protocolID := protocol.ID("/p2p/rpc/ping")
+	rpcServer := rpc.NewServer(host, protocolID)
+	return rpcServer.RegisterName("Consensus", rpcHandler)
+	//return err
+}
```
magik6k, line 56: Protocol ID needs to be set to something better.

```suggestion
func NewRPCClient(host host.Host) *rpc.Client {
	protocolID := protocol.ID("/rpc/lotus-chain/v0")
	return rpc.NewClient(host, protocolID)
}

func NewRPCServer(host host.Host, rpcHandler *RPCHandler) error {
	protocolID := protocol.ID("/rpc/lotus-chain/v0")
	rpcServer := rpc.NewServer(host, protocolID)
	return rpcServer.RegisterName("Consensus", rpcHandler)
}
```

node/modules/rpc.go
```diff
@@ -0,0 +1,66 @@
...
+
+func NewRPCClient(host host.Host) *rpc.Client {
+	protocolID := protocol.ID("/p2p/rpc/ping")
+	return rpc.NewClient(host, protocolID)
+}
+
+func NewRPCServer(host host.Host, rpcHandler *RPCHandler) error {
+	protocolID := protocol.ID("/p2p/rpc/ping")
+	rpcServer := rpc.NewServer(host, protocolID)
+	return rpcServer.RegisterName("Consensus", rpcHandler)
```
magik6k, line 54: We must set some ACLs up for the RPC server, otherwise random people on the network will be able to call those and sign messages with our keys.

magik6k (commented):
First pass, doesn't feel like there's that much left to do, mostly cleaning up / making stuff more robust.

* Looks like a devnet genesis (`devgen.car`) got committed by accident; Given that it's a little bit we'll need to make sure it's removed from all commits in the PR to not carry it with git history after merging

lib/consensus/raft/consensus.go
hsanjuan: Raft stores the serialization of the state in the snapshot store regularly, and the last index applied to it. So as long as you use a non-in-memory store (i.e Bolt), then you're good.

It is recommended that the RaftState implements Marshal/Unmarshal: https://github.com/libp2p/go-libp2p-raft/blob/master/consensus.go#L20-L36, although I'm guessing in this case the Msgpack default encoding should work.

chain/messagesigner/messagesigner.go
```diff
@@ -29,6 +31,19 @@ type MpoolNonceAPI interface {
...
+type MsgSigner interface {
+	SignMessage(ctx context.Context, msg *types.Message, spec *api.MessageSendSpec, cb func(*types.SignedMessage) error) (*types.SignedMessage, error)
+	GetSignedMessage(ctx context.Context, uuid uuid.UUID) (*types.SignedMessage, error)
+	StoreSignedMessage(ctx context.Context, uuid uuid.UUID, message *types.SignedMessage) error
+	NextNonce(ctx context.Context, addr address.Address) (uint64, error)
+	SaveNonce(ctx context.Context, addr address.Address, nonce uint64) error
+	DstoreKey(addr address.Address) datastore.Key
+	IsLeader(ctx context.Context) bool
+	RaftLeader(ctx context.Context) (peer.ID, error)
+	RedirectToLeader(ctx context.Context, method string, arg interface{}, ret interface{}) (bool, error)
```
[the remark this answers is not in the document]
shrenujbansal: Yes. This is what I brought up earlier on. Not sure how much of a choice we have other than to redo what we decide consensus should be part of

chain/messagesigner/messagesigner.go
```diff
@@ -29,6 +31,19 @@ type MpoolNonceAPI interface {
...
+	GetRaftState(ctx context.Context) (consensus.State, error)
```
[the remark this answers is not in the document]
shrenujbansal: Same reason as above. Everything consensus related has to be exposed via the MessageSigner
It needs to be on the interface because the FullNodeAPI needs to have this method also

node/config/types.go
shrenujbansal: It maybe be better to start with the default config always and then unpack this after. I feel usage might be a little bit more overwhelming if people have to start understanding the intricacies of raft immediately?

chain/messagesigner/messagesigner.go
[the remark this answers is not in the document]
magik6k: We could have an optional dependency on the Consensus struct for those APIs; The MsgSigner interface should only be used to abstract out message-signing related things:

image

node/config/types.go
magik6k: That sounds reasonable

chain/messagesigner/messagesigner.go
```diff
@@ -29,6 +31,19 @@ type MpoolNonceAPI interface {
 	GetActor(context.Context, address.Address, types.TipSetKey) (*types.Actor, error)
 }
 
+type MsgSigner interface {
+	SignMessage(ctx context.Context, msg *types.Message, spec *api.MessageSendSpec, cb func(*types.SignedMessage) error) (*types.SignedMessage, error)
+	GetSignedMessage(ctx context.Context, uuid uuid.UUID) (*types.SignedMessage, error)
+	StoreSignedMessage(ctx context.Context, uuid uuid.UUID, message *types.SignedMessage) error
+	NextNonce(ctx context.Context, addr address.Address) (uint64, error)
+	SaveNonce(ctx context.Context, addr address.Address, nonce uint64) error
+	DstoreKey(addr address.Address) datastore.Key
```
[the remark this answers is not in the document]
shrenujbansal: done

node/config/types.go
shrenujbansal: done

node/modules/rpc.go
shrenujbansal: Will remove all the commented out stuff once we get to a good final state

node/modules/rpc.go
shrenujbansal: done

node/modules/rpc.go
shrenujbansal: done

chain/messagesigner/messagesigner_consensus.go
```diff
@@ -0,0 +1,99 @@
...
+
+	signedMsg, err := ms.MsgSigner.SignMessage(ctx, msg, spec, cb)
+	if err != nil {
+		return nil, err
+	}
+
+	// We can't have an empty/default uuid as part of the consensus state so generate a new uuid if spec is empty
+	u := uuid.New()
+	if spec != nil {
+		u = spec.MsgUuid
```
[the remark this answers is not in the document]
shrenujbansal: This needs to be set here because the consensus state needs a non empty uuid. And the API today allows you to push a message with an empty uuid
Are you saying you want to set a new value elsewhere instead?

itests/self_sent_txn_test.go
shrenujbansal: good spot. got swept in accidentally

itests/raft_messagesigner_test.go
shrenujbansal: Not sure I understand. This is making sure the raft leader has changed after the original leader went down
require.Eventually waits for the condition to be equal?

lib/consensus/raft/config.go
shrenujbansal: yes will do a big cleanup soon

lib/consensus/raft/consensus.go
shrenujbansal: We do. I figured it'd be nice to have separately since those are the core parts of the state/op but I'm ok just leaving it in SignedMessage if you think its a concern

lib/consensus/raft/consensus.go
shrenujbansal: consensus.State is an interface implemented by RaftState

itests/raft_messagesigner_test.go
shrenujbansal: done

lib/consensus/raft/consensus.go
shrenujbansal: Will make this part of the big cleanup

node/builder.go
shrenujbansal: done

lib/consensus/raft/consensus.go
shrenujbansal: could probably then also move finishBootstrap into newConsensus instead of having it as a goroutine

chain/messagesigner/messagesigner.go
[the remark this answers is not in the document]
shrenujbansal: Moved this to RaftAPI

chain/messagesigner/messagesigner.go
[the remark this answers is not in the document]
shrenujbansal: Changed dependencies to the above diagram

cli/util/api.go
```diff
@@ -185,6 +230,77 @@ func GetFullNodeAPI(ctx *cli.Context) (v0api.FullNode, jsonrpc.ClientCloser, err
...
+}
+
+func FullNodeProxy[T api.FullNode](ins []T, outstr *api.FullNodeStruct) {
+	outs := api.GetInternalStructs(outstr)
+
+	var rins []reflect.Value
+	//peertoNode := make(map[peer.ID]reflect.Value)
+	for _, in := range ins {
+		rin := reflect.ValueOf(in)
+		rins = append(rins, rin)
```
magik6k, line 245: Nit: I'd collapse this into a single line

```suggestion
		rins = append(rins, reflect.ValueOf(in))
```

cli/util/api.go
```diff
@@ -185,6 +230,77 @@ func GetFullNodeAPI(ctx *cli.Context) (v0api.FullNode, jsonrpc.ClientCloser, err
...
+		//peertoNode[ins] = rin
+	}
+
+	for _, out := range outs {
+		rint := reflect.ValueOf(out).Elem()
```
magik6k, line 250: I realize this is coming from the code I wrote - can we rename `rint` to something like `rProxyInternal` or any other more descriptive name?

cli/util/api.go
```diff
@@ -185,6 +230,77 @@ func GetFullNodeAPI(ctx *cli.Context) (v0api.FullNode, jsonrpc.ClientCloser, err
...
+			rint.Field(f).Set(reflect.MakeFunc(field.Type, func(args []reflect.Value) (results []reflect.Value) {
+				errorsToRetry := []error{&jsonrpc.RPCConnectionError{}, &jsonrpc.ErrClient{}}
+				initialBackoff, err := time.ParseDuration("1s")
+				if err != nil {
+					return nil
+				}
+
+				ctx := args[0].Interface().(context.Context)
+
+				curr := -1
```
magik6k, line 269: Can we add a comment explaining this -1?

cli/util/apiinfo.go
```diff
@@ -36,6 +37,31 @@ func ParseApiInfo(s string) APIInfo {
 	}
 }
 
+func ParseApiInfoMulti(s string) []APIInfo {
+	var apiInfos []APIInfo
+
+	if infoWithToken.Match([]byte(s)) {
```
magik6k, line 43: How does this work if we didn't update the `infoWithToken` regex above?

documentation/en/default-lotus-config.toml
```diff
@@ -226,3 +226,77 @@
     #ColdStoreRetention = 0
 
 
+[Raft]
```
magik6k, line 229: Maybe call this group "Cluster"? Raft is something most people will need to look up to know what it's doing.

node/modules/storageminer.go
```diff
@@ -89,12 +87,14 @@ func (a *UuidWrapper) MpoolPushMessage(ctx context.Context, msg *types.Message,
...
-	}
-	return retry.Retry(5, initialBackoff, errorsToRetry, func() (*types.SignedMessage, error) { return a.FullNode.MpoolPushMessage(ctx, msg, spec) })
+	return a.FullNode.MpoolPushMessage(ctx, msg, spec)
+
+	//errorsToRetry := []error{&jsonrpc.RPCConnectionError{}, &jsonrpc.ErrClient{}}
+	//initialBackoff, err := time.ParseDuration("1s")
+	//if err != nil {
+	//	return nil, err
+	//}
+	//return retry.Retry(5, initialBackoff, errorsToRetry, func() (*types.SignedMessage, error) { return a.FullNode.MpoolPushMessage(ctx, msg, spec) })
```
magik6k, line 97:
```suggestion
```

chain/messagesigner/messagesigner_consensus.go
```diff
@@ -0,0 +1,104 @@
...
+	return ms.consensus.IsLeader(ctx)
+}
+
+func (ms *MessageSignerConsensus) RedirectToLeader(ctx context.Context, method string, arg interface{}, ret interface{}) (bool, error) {
+	ok, err := ms.consensus.RedirectToLeader(method, arg, ret.(*types.SignedMessage))
+	if err != nil {
+		return ok, err
+	}
+	return ok, nil
+}
```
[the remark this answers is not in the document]
shrenujbansal: No, this is needed to be called in MpoolPushMessage

chain/messagesigner/messagesigner_consensus.go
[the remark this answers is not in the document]
shrenujbansal: It can't be a 0 value because then we would reject any messages that don't have it originally set

cli/util/apiinfo.go
shrenujbansal: I'm not sure. Do you expect it to fail?

lib/consensus/raft/raft.go
shrenujbansal: Using persistent stores here

cli/util/api.go
```diff
@@ -76,65 +79,107 @@ func GetAPIInfo(ctx *cli.Context, t repo.RepoType) (APIInfo, error) {
...
-func GetRawAPI(ctx *cli.Context, t repo.RepoType, version string) (string, http.Header, error) {
-	ainfo, err := GetAPIInfo(ctx, t)
+func GetAPIInfo(ctx *cli.Context, t repo.RepoType) (APIInfo, error) {
+	ainfos, err := GetAPIInfoMulti(ctx, t)
 	if err != nil {
-		return "", nil, xerrors.Errorf("could not get API info for %s: %w", t.Type(), err)
+		return APIInfo{}, err
+	}
+
+	if len(ainfos) > 1 {
```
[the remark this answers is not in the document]
shrenujbansal: Possibly. Accounted for this use case

cli/util/api.go
shrenujbansal: This logic needs to be made better. Right now, we always call node 0 first. Ideally we want to call the leader node first but getting that info in here is a challenge

chain/messagesigner/messagesigner_consensus.go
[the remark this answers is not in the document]
magik6k: But isn't this creating a regression for the fix in https://github.com/filecoin-project/lotus/pull/9303

cli/util/apiinfo.go
magik6k: Yeah, I'm a little confused why this method works with multiple API infos. I'm assuming it's exercised in your manual testing.

shrenujbansal: yep. I set the FULLNODE_API_INFO env var for the miner

lib/consensus/raft/config.go
```diff
@@ -0,0 +1,143 @@
...
+	// CommitRetries specifies how many times we retry a failed commit until
+	// we give up.
+	CommitRetries int
+	// How long to wait between retries
+	CommitRetryDelay time.Duration
+	// BackupsRotate specifies the maximum number of Raft's DataFolder
+	// copies that we keep as backups (renaming) after cleanup.
+	BackupsRotate int
+	// Namespace to use when writing keys to the datastore
+	DatastoreNamespace string
```
magik6k, line 53: I don't think we use this namespace anywhere

node/config/def.go
```diff
@@ -287,3 +288,27 @@ const (
...
+func DefaultUserRaftConfig() *UserRaftConfig {
+	var cfg UserRaftConfig
+	cfg.DataFolder = "" // empty so it gets omitted
+	cfg.InitPeersetMultiAddr = []string{}
+	cfg.WaitForLeaderTimeout = Duration(DefaultWaitForLeaderTimeout)
+	cfg.NetworkTimeout = Duration(DefaultNetworkTimeout)
+	cfg.CommitRetries = DefaultCommitRetries
+	cfg.CommitRetryDelay = Duration(DefaultCommitRetryDelay)
+	cfg.BackupsRotate = DefaultBackupsRotate
+	cfg.DatastoreNamespace = DefaultDatastoreNamespace
```
magik6k, line 311: Datastore namespaces shouldn't be configurable in the user config

node/impl/full/raft.go
```diff
@@ -0,0 +1,50 @@
...
+	raftState, err := r.MessageSigner.GetRaftState(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &api.RaftStateData{NonceMap: raftState.NonceMap, MsgUuids: raftState.MsgUuids}, nil
+}
+
+func (r *RaftAPI) Leader(ctx context.Context) (peer.ID, error) {
+	if r.MessageSigner == nil {
+		return "", xerrors.Errorf("Raft consensus not enabled. Please check your configuration")
```
magik6k, line 33:
```suggestion
		return "", xerrors.Errorf("raft consensus not enabled. Please check your configuration")
```

node/impl/full/raft.go
```diff
@@ -0,0 +1,50 @@
...
+func (r *RaftAPI) IsLeader(ctx context.Context) bool {
+	if r.MessageSigner == nil {
+		return true
+	}
+	return r.MessageSigner.IsLeader(ctx)
+}
+
+func (r *RaftAPI) RedirectToLeader(ctx context.Context, method string, arg interface{}, ret interface{}) (bool, error) {
+	if r.MessageSigner == nil {
+		return false, xerrors.Errorf("Raft consensus not enabled. Please check your configuration")
```
magik6k, line 47:
```suggestion
		return false, xerrors.Errorf("raft consensus not enabled. Please check your configuration")
```

cli/util/api.go
```diff
@@ -214,6 +319,58 @@ func GetFullNodeAPIV1(ctx *cli.Context) (v1api.FullNode, jsonrpc.ClientCloser, e
 	return v1API, closer, nil
 }
 
+func GetFullNodeAPIV1New(ctx *cli.Context) (v1api.FullNode, jsonrpc.ClientCloser, error) {
```
magik6k, line 322: I think it's fine to call this `GetFullNodeAPI`, and the other one `GetFullNodeAPISingle`, and have everything go through `GetFullNodeAPI`, so everything gets multi-node support

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,618 @@
...
+	cc.cancel()
+	close(cc.rpcReady)
+	return nil
+}
+
+// Ready returns a channel which is signaled when the Consensus
+// algorithm has finished bootstrapping and is ready to use
+func (cc *Consensus) Ready(ctx context.Context) <-chan struct{} {
+	//_, span := trace.StartSpan(ctx, "consensus/Ready")
+	//defer span.End()
```
magik6k, line 313: I'd either drop tracing or uncomment tracing. (we use `"go.opencensus.io/trace"` in lotus, with span names separated by `.`s. Also the namespace should be `raft` as I'm pretty sure we already generate consensus spans, so this would be `trace.StartSpan(ctx, "raft.Ready")`)

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,618 @@
...
+		}
+
+		// try to send it to the leader
+		// RedirectToLeader has it's own retry loop. If this fails
+		// we're done here.
+
+		//ok, err := cc.RedirectToLeader(rpcOp, redirectArg, struct{}{})
+		//if err != nil || ok {
+		//	return err
+		//}
```
magik6k, line 421:
```suggestion
```

lib/consensus/raft/raft.go
```diff
@@ -0,0 +1,718 @@
...
+	}
+	return err
+}
+
+// Shutdown shutdown Raft and closes the BoltDB.
+func (rw *raftWrapper) Shutdown(ctx context.Context) error {
+	//_, span := trace.StartSpan(ctx, "consensus/raft/Shutdown")
+	//defer span.End()
+
+	errMsgs := ""
```
magik6k, line 428: I'd use `go.uber.org/multierr.Append / Combine` for this

https://pkg.go.dev/go.uber.org/multierr#hdr-Overview

lib/consensus/raft/raft.go
```diff
@@ -0,0 +1,718 @@
...
+
+	if len(peers) == 1 && peers[0] == peer {
+		return errors.New("cannot remove ourselves from a 1-peer cluster")
+	}
+
+	rmFuture := rw.raft.RemoveServer(
+		hraft.ServerID(peer),
+		0,
+		0,
+	) // TODO: Extra cfg value?
```
magik6k, line 513: Do we know what this todo is about?

magik6k (commented):
Looks like there is a commit updating `extern/filecoin-ffi` by accident

node/config/def.go
```diff
@@ -287,3 +288,27 @@ const (
...
+	DefaultDataSubFolder        = "raft"
+	DefaultWaitForLeaderTimeout = 15 * time.Second
+	DefaultCommitRetries        = 1
+	DefaultNetworkTimeout       = 100 * time.Second
+	DefaultCommitRetryDelay     = 200 * time.Millisecond
+	DefaultBackupsRotate        = 6
+	DefaultDatastoreNamespace   = "/raft"
+)
+
+func DefaultUserRaftConfig() *UserRaftConfig {
```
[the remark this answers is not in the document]
shrenujbansal: exactly why? The other default configs all return a pointer

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,618 @@
...
+
+	logger.Info("stopping Consensus component")
+
+	// Raft Shutdown
+	err := cc.raft.Shutdown(ctx)
+	if err != nil {
+		logger.Error(err)
+	}
+
+	if cc.config.HostShutdown {
```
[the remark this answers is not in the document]
shrenujbansal: I imagine to close the libp2p host? Do we already do this elsewhere?

lib/consensus/raft/config.go
shrenujbansal: seems like its used the in OfflineState function

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,618 @@
...
+//	if !snapExists {
+//		return st, nil
+//	}
+//
+//	err = st.Unmarshal(r)
+//	if err != nil {
+//		return nil, err
+//	}
+//	return st, nil
+//}
```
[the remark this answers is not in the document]
shrenujbansal: Should we get rid of this or keep it around if we need to use it down the line?

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,618 @@
...
+		readyCh:   make(chan struct{}, 1),
+		repo:      repo,
+	}
+
+	go cc.finishBootstrap()
+	return cc, nil
+
+}
+
+func NewConsensusWithRPCClient(staging bool) func(host host.Host,
```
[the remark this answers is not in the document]
shrenujbansal: I'll add a TODO

lib/consensus/raft/raft.go
shrenujbansal: no idea. This was copy pasta. I'll remove it

shrenujbansal: No idea. This was copy pasta. Removed

cli/util/api.go
shrenujbansal: There's already something called GetFullNodeAPI which I believe points to the v0 API?
Changed the name to the default one

lib/consensus/raft/raft.go
hsanjuan: I also have no idea :man_shrugging:

lib/consensus/raft/raft.go
shrenujbansal: done

lib/consensus/raft/consensus.go
shrenujbansal: removed tracing

documentation/en/default-lotus-config.toml
```diff
@@ -202,3 +202,77 @@
     #HotStoreFullGCFrequency = 20
 
 
+[Cluster]
+  # config to enabled node cluster with raft consensus
```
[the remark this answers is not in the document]
shrenujbansal: done

chain/messagepool/provider.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/network"
 
-	"github.com/filecoin-project/lotus/chain/messagesigner"
+	//"github.com/filecoin-project/lotus/chain/messagesigner"
```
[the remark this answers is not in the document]
shrenujbansal: done

lib/consensus/raft/consensus.go
```diff
@@ -0,0 +1,618 @@
...
+	//ctx, span := trace.StartSpan(ctx, "consensus/Shutdown")
+	//defer span.End()
+
+	//cc.shutdownLock.Lock()
+	//defer cc.shutdownLock.Unlock()
+
+	//if cc.shutdown {
+	//	logger.Debug("already shutdown")
+	//	return nil
+	//}
```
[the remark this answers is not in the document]
hsanjuan: you can probably drop if you don't expect your Shutdown method to be called multiple times. Cluster can be used as a library, and people can potentially call Shutdown multiple times and we added these guards in general to all Shutdown methods.

node/config/types.go
```diff
@@ -600,3 +601,38 @@ type Wallet struct {
 type FeeConfig struct {
 	DefaultMaxFee types.FIL
 }
+
+type UserRaftConfig struct {
+	// config to enabled node cluster with raft consensus
+	ClusterModeEnabled bool
+	// will shutdown libp2p host on shutdown. Useful for testing
+	HostShutdown bool
+	// A folder to store Raft's data.
```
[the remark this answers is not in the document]
shrenujbansal: are you referring to HostShutdown or DataFolder? Also why not?

chain/messagesigner/messagesigner.go
```diff
@@ -24,21 +25,31 @@ const dsKeyMsgUUIDSet = "MsgUuidSet"
...
-type MpoolNonceAPI interface {
-	GetNonce(context.Context, address.Address, types.TipSetKey) (uint64, error)
-	GetActor(context.Context, address.Address, types.TipSetKey) (*types.Actor, error)
+type MsgSigner interface {
+	SignMessage(ctx context.Context, msg *types.Message, spec *api.MessageSendSpec, cb func(*types.SignedMessage) error) (*types.SignedMessage, error)
+	GetSignedMessage(ctx context.Context, uuid uuid.UUID) (*types.SignedMessage, error)
+	StoreSignedMessage(ctx context.Context, uuid uuid.UUID, message *types.SignedMessage) error
+	NextNonce(ctx context.Context, addr address.Address) (uint64, error)
+	SaveNonce(ctx context.Context, addr address.Address, nonce uint64) error
+	dstoreKey(addr address.Address) datastore.Key
```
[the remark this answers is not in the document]
shrenujbansal: done

lib/consensus/raft/consensus.go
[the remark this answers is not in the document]
magik6k: Yeah, that happens in the libp2p constructor is what is supposed to manage libp2p lifecycle

lib/consensus/raft/consensus.go
magik6k: I'd drop those fields, as it's just more state that needs to be managed

node/config/types.go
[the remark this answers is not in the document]
magik6k: HostShutdown; it probably doesn't need to exist at all (the entire HostShutdown code path)

lib/consensus/raft/consensus.go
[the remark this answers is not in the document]
shrenujbansal: removed

node/config/def.go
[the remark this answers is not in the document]
magik6k: Oh, I guess they do

lib/consensus/raft/consensus.go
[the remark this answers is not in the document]
magik6k: Generally we never keep commented out code, this is what Git is for.

lib/consensus/raft/config.go
magik6k: Yeah, but that's commented out, and even if it wasn't datastore namespaces aren't something that should be exposed in the user config

cli/util/api.go
magik6k: Sorry meant GetFullNodeAPI**V1**

lib/consensus/raft/consensus.go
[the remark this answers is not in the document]
shrenujbansal: Removed

chain/messagepool/messagepool.go
```diff
@@ -667,7 +668,7 @@ func (mp *MessagePool) verifyMsgBeforeAdd(ctx context.Context, m *types.SignedMe
 	return publish, nil
 }
 
-func (mp *MessagePool) Push(ctx context.Context, m *types.SignedMessage) (cid.Cid, error) {
+func (mp *MessagePool) Push(ctx context.Context, m *types.SignedMessage, publish bool) (cid.Cid, error) {
```
geoff-vball, line 671: I don't see a similar change for `PushUntrusted`. Is that intentional?

geoff-vball: Also, could you leave a clear comment on what exactly this bool does?

shrenujbansal: We shouldn't need it for PushUntrusted since its not used in syncing the message pool. I think this API should really be folded back into Push itself and add a untrusted param to it

magik6k:
Dependency diff if anyone else wants to review: https://gist.github.com/magik6k/ada79bebc4f5ccf3078d72f6513a49e2

geoff-vball (approved):

chain/messagepool/messagepool.go
geoff-vball: Yes I definitely agree :)

Issue #9257, opened 2022-09-02
Integrate the new WithdrawBalance to support caller as owner or beneficairy 


jennijuju:
Two options I can think of
-  move [ActorWithdrawBalance](https://github.com/filecoin-project/lotus/blob/45d1bd61cecd8e723f2239979377b8ac92c62362/node/impl/storminer.go#L1330) out of Storage Miner API to full node API.
   - take the caller's address as a param
   - make sure the caller is either the beneficiary or the owner 
   - then proceed
   - we should rename the API to `MinerWithdrawBalace` probably
- or we can keep ActorWithdrawBalance as it is today, and add a `BeneficiaryWithdrawBalance` 

I prefer 1 personally but that might be considered as a breaking api change. if so we might wanna do 2.  @magik6k @geoff-vball  - thoughts?

jennijuju:
discussed in standup - we will go with @geoff-vball 's reco Option 2

TippyFlitsUK:
From an SP's perspective, I agree that option 2 would be the preferred option.

Issue #9258, opened 2022-09-02
add caller for actorWithdrawCmd


jennijuju:
`actorWithdrawCmd` should now take a from address

`lotus-miner actor withdraw <actor>  <caller address>`

the cmd should only proceed if the caller is the owner or beneficiary of the actor

geoff-vball:
Implemented in https://github.com/filecoin-project/lotus/pull/9296

actorWithdrawCmd now takes an optional --beneficiary flag to send the message from the beneficiary address. No need for a `--from` flag here because there are only two possible senders.

I will add a `--actor` flag to this command so we can specify which miner to send the message to.

jennijuju:
> just to get that right - if a beneficiary address is set and i send the message from the owner - the owner gets the funds?

Nope. According to the fip it’s always to beneficiary.

PR #9296, opened 2022-09-12
feat: api/cli: beneficiary withdraw api and cli

geoff-vball:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/9257 and https://github.com/filecoin-project/lotus/issues/9258

## Proposed Changes
Adds an api call and cli command to withdraw funds from a miner as the beneficiary.

## Checklist

- [ ] CI is green


arajasek (changes requested):
I don't think we want the new API method in the FullNodeAPI here, just the Storage API. See `ActorWithdrawBalance` for an equivalent method to use as a blueprint.

geoff-vball:
@arajasek Does the storage API get run on the `lotus` node or through `lotus-miner`? I think I feel pretty strongly that this is an action that shouldn't require running `lotus-miner` to achieve.

arajasek:
@geoff-vball Yes, api_storage is lotus-miner only. I would match what we currently have for `(Owner)WithdrawBalance` for now (which is to only have it on the miner API), and create a ticket to support both `(Owner)WithdrawBalance` and `BeneficiaryWithdrawBalance` over the node API. I'd be...skeptical about whether these are actions we want to be able to take remotely over an API, but we can descope that discussion.

Additionally, I would move the CLI commands to the following two locations, you can match existing behaviour there: https://github.com/filecoin-project/lotus/blob/e65fae28de2a44947dd24af8c7dafcade29af1a4/cmd/lotus-shed/actor.go#L39, https://github.com/filecoin-project/lotus/blob/15ea3953f09311d1587d1c3291013771ca7ba3e7/cmd/lotus-miner/actor.go#L233. The first can be called by anyone who has built lotus-shed, the second requires the miner running in the background.

TL;DR: let's match what we have for the existing withdrawBalance method and take it from there.

cmd/lotus-shed/actor.go
```diff
@@ -159,6 +160,78 @@ var actorWithdrawCmd = &cli.Command{
...
+		minerAPI, closer, err := lcli.GetStorageMinerAPI(cctx)
+		if err != nil {
+			return err
+		}
+		defer closer()
+
+		msgCid, err := minerAPI.BeneficiaryWithdrawBalance(ctx, maddr, amount)
+		if err != nil {
+			return err
+		}
```
geoff-vball, line 207: @arajasek I imagine that I might have to used `MpoolPushMessage` here instead of relying on the miner-api that might not be present?

cmd/lotus-miner/actor.go
```diff
@@ -40,6 +40,7 @@ var actorCmd = &cli.Command{
 	Subcommands: []*cli.Command{
 		actorSetAddrsCmd,
 		actorWithdrawCmd,
```
arajasek, line 42: Can we rename this to ownerWithdrawCmd, rename the command itself to `owner-withdraw`, but leave `withdraw` as an alias?

cmd/lotus-miner/actor.go
```diff
@@ -305,6 +306,78 @@ var actorWithdrawCmd = &cli.Command{
 	},
 }
 
+var beneficiaryWithdrawCmd = &cli.Command{
+	Name:      "beneficiary-withdraw",
+	Usage:     "withdraw available balance to beneficiary address. Will not work from non-beneficiary owner addresses. Use lotus-miner for non-beneficiary owner withdrawals",
```
arajasek, line 311: This helptext  is confusing me -- what are you trying to say here? I think what you need to tell the user is that this command will try to _send_ the message from the beneficiary address. If one isn't set, or if they just want to use the owner address, they should instead use `lotus-miner withdraw`?

cmd/lotus-shed/actor.go
arajasek: Yes, see the command directly above this one. Honestly, you could fold this command into that one -- just add a param for the user to indicate whether to send from owner or beneficiary. They're otherwise identical.

node/impl/storminer.go
```diff
@@ -1387,3 +1387,43 @@ func (sm *StorageMinerAPI) ActorWithdrawBalance(ctx context.Context, amount abi.
 
 	return smsg.Cid(), nil
 }
+
+func (sm *StorageMinerAPI) BeneficiaryWithdrawBalance(ctx context.Context, minerAddress address.Address, amount abi.TokenAmount) (cid.Cid, error) {
```
arajasek: That's a lot of duplicated code -- let's extract all of this into a helper that takes a `useOwner` boolean param indicating who the sender should be?

cmd/lotus-miner/actor.go
```diff
@@ -232,14 +233,18 @@ var actorSetPeeridCmd = &cli.Command{
...
 	ArgsUsage: "[amount (FIL)]",
 	Flags: []cli.Flag{
 		&cli.IntFlag{
 			Name:  "confidence",
 			Usage: "number of block confirmations to wait for",
 			Value: int(build.MessageConfidence),
 		},
+		&cli.BoolFlag{
+			Name:  "beneficiary",
+			Usage: "send withdraw command using beneficiary key",
```
arajasek, line 246:
```suggestion
			Usage: "send withdraw message from the beneficiary address",
```

cmd/lotus-miner/actor.go
geoff-vball: Refactored everything. Now the cli commands just need a flag to know to send from beneficiary.

cmd/lotus-shed/actor.go
```diff
@@ -50,6 +50,10 @@ var actorWithdrawCmd = &cli.Command{
 			Usage: "number of block confirmations to wait for",
 			Value: int(build.MessageConfidence),
 		},
+		&cli.BoolFlag{
+			Name:  "beneficiary",
+			Usage: "send withdraw command using beneficiary key",
```
arajasek, line 55:
```suggestion
			Usage: "send withdraw message from the beneficiary address",
```

PR #9267, opened 2022-09-05
feat: network: add nv17 and integrate the corresponding go state type

jennijuju:
## Related Issues

Towards #9246, #9248, #9249

## Proposed Changes

- Upgrades go-state-types to include v9 types and migration
- Adds network version 17 and the upgrade height
- Adds the v17 migration
- Updates to FVM release/v2 (through FFI)
  - This includes reviving https://github.com/filecoin-project/lotus/pull/9134/files, which is incorporated into this PR
- Adds the v9 actors bundles (using https://github.com/filecoin-project/builtin-actors/releases/tag/dev%2F20220907-v9 for now)
- Refactors the Actor Registry creation process
  - This is something we cut from scope in v16 by instead using the v8 specs-actors, and monkey-patching the synthetic -> real CIDs
  - With this PR, we get the method signatures from go-state-types (see https://github.com/filecoin-project/go-state-types/pull/66)
  - We use that information to create new `RegistryEntry`s that satisfy the same interface
  - There is a downgrade in UX experience here: we no longer have the method names in the registry
  - I don't know how to include this info, suggestions welcome. For now, we're including the MethodNum instead.
- Handles most of the integration work needed for FIP-0034
  - We use the new v9 miner types wherever possible
  - This mostly requires handling the fact that `SectorPreCommitInfo` and `PreCommitSectorParams` are no longer the same type
  - We simply create `PreCommitSectorParams` from `SectorPreCommitInfo` wherever necessary
  - We also need to recalibrate the calculation of necessary Pre-Commit Deposit when estimating collateral needed for miners
- Handles most of the integration work needed for FIP-0041 (by skipping it for now)
  - We do NOT start calling the new `PreCommitSectorBatch2` and `ProveReplicaUpdates2` methods. It isn't necessary to do so as part of this PR, and I would propose delaying calling these methods until after the nv17 upgrade.
  - See https://github.com/filecoin-project/lotus/issues/9250#issuecomment-1242766564.
- Moves things around in the state interfaces to allow for better versioning of the go-state-types objects. This is expected to be a one-time refactor, not needed every time.
## Additional Info

- [x] Needs https://github.com/filecoin-project/go-state-types/pull/68
- [x] Needs go-state-types tag
- [x] Needs https://github.com/filecoin-project/ref-fvm/pull/861 (and maybe an FVM pre-release)
- [ ] Needs https://github.com/filecoin-project/filecoin-ffi/pull/320 AFTER https://github.com/filecoin-project/ref-fvm/pull/861

[...]

PR #9319, opened 2022-09-15
fix: dont change the latest network version 

jennijuju:
dont think newest network version  should be nv17 in v1.17.2 

(shall only be updated close to v1.18.0 -rc

### DIFF
--- a/build/params_shared_vals.go
+++ b/build/params_shared_vals.go
@@ -33,7 +33,7 @@ const NewestNetworkVersion = network.Version{{.latestNetworkVersion}}
 
 /* inline-gen start */
 
-const NewestNetworkVersion = network.Version17
+const NewestNetworkVersion = network.Version16
 
 /* inline-gen end */
 


arajasek:
Lol, we had this discussion when you were live :cattyping: on #9267 and you wanted to update it!

I'm not sure whether I agree with this change. The biggest reason to keep it is that it makes (most) itests run with the "latest" code. This forces us to iron out all bugs when we integrate a new network version.

This param is otherwise almost exclusively used for devnets and tests. What's your motivation to _not_ update it?

jennijuju:
I remember I asked if we should - and the answer was yes. But I regret it lolol.
- adds confusion to user who reads the code
- I believe devnet is a bit messed up with this, devnet starts at nv16 but there is a couple things we use latest network version

geoff-vball:
@jennijuju Then I think we need to fix the things that are using the wrong network version. We shouldn't have to slow down development because of that

jennijuju:
How does that slow down the development? 
And key reason, the newest network version should stay v16 until v1.18.0

geoff-vball:
@jennijuju Does anyone actually look at the code besides you? :)

The line you changed is autogenerated from the list of network versions in `go-state-types`. We would have to cut a new release of `go-state-types` to change it back. We can if you really want us to, but I strongly believe it's not worth our time.

arajasek:
> I believe devnet is a bit messed up with this, devnet starts at nv16 but there is a couple things we use latest network version

@jennijuju That just sounds like an issue we should fix -- can you share what seems bork? Should be a quick fix.

arajasek:
> The line you changed is autogenerated from the list of network versions in go-state-types. We would have to cut a new release of go-state-types to change it back

Nah, we can just stop autogenerating that line

arajasek:
Killing this in favour of #9351

PR #9397, opened 2022-09-29
Eth JSON-RPC API: implement eth_getCode and eth_getStorageAt

raulk:
eth_getStorageAt comes with a major hack for now because of a serde snafu in the EVM runtime in builtin-actors. We are not updating the bundle for Iron, so we patch it here. The type we used in the EVM runtime for the storage slot serializes to a hex-encoded int instead of to purely an big-endian array of bytes. We'll fix this at origin in builtin-actors.

Another caveat is that retrieving a storage key that doesn't exist returns an error. Instead, it should return a 0x0.


node/impl/full/eth.go
```diff
@@ -207,21 +210,155 @@ func (a *EthModule) EthGetTransactionReceipt(ctx context.Context, txHash api.Eth
...
+
+	if err != nil {
+		return nil, xerrors.Errorf("Call failed: %w", err)
+	}
+
+	if res.MsgRct == nil {
+		return nil, fmt.Errorf("no message receipt")
+	}
+
+	if res.MsgRct.ExitCode != exitcode.Ok {
```
ychiaoli18, line 267: Should we use `res.MsgRct.ExitCode.IsSuccess()`

node/impl/full/eth.go
```diff
@@ -207,21 +210,155 @@ func (a *EthModule) EthGetTransactionReceipt(ctx context.Context, txHash api.Eth
...
+		return nil, xerrors.Errorf("message execution failed: exit %s, reason: %s", res.MsgRct.ExitCode, res.Error)
+	}
+
+	fmt.Printf("%x\n", res.MsgRct.Return)
```
ychiaoli18, line 271: should we remove this print?

ychiaoli18 (approved):
LGTM

node/impl/full/eth.go
raulk: Oh yes, debugging leftovers.

node/impl/full/eth.go
raulk: Done.

PR #9405, opened 2022-10-03
Abstract common consensus functions and consensus interface

adlrocha:
## Related Issues
#1

## Proposed Changes
This PR extends the consensus interface in order to prepare it for the implementation of other consensus algorithms. It also extracts from `FilecoinEC` common logic that will be shared by all consensus implementations.
- Extracted block checks and validation to common functions.
- Extracted `ValidatedPubSub` as a common function.
- Abstracted block signature and verification, and the validation of blocks and their headers. 
- Extracted `compute_state` as common code for all consensus algorithms.
- Abstracted the application of rewards.

## Integration of the interface
To illustrate how this interface can be integrated for the implementation of alternative consensus algorithm, I will share how I imagine it to work and be implemented for Mir:
- `CreateBlock` runs the block proposal logic for Mir in Lotus. In `CreateBlock` Lotus validators pull unverified messages from the mempool, and send them to Mir for ordering. Once the ordered batch is output by Mir, a new Filecoin block is assembled and broadcast through GossipSub to the network.
- Validators are not subscribed to the block gossipsub topic, as they already receive all the information about new blocks through Mir.
- Other Lotus nodes in the network will receive updates about new blocks being proposed by Mir through the Gossipsub `block` topic. When a new block is received, GossipSub triggers `ValidateBlockHeader` before delivering the block to the node. If the validation succeeds, the block is passed to the syncer which runs `ValidateBlock` to perform a full block validation to determine if the block is valid and can be executed. As part of these verifications, `VerifyBlockSignature` may be called. 
- Finally, the execution of the block triggers the consensus-specific `RewardFunc` that parametrized how should block rewards be distributed among validators.


chain/consensus/compute_state.go
```diff
@@ -222,29 +224,9 @@ func (t *TipSetExecutor) ApplyBlocks(ctx context.Context,
...
-		}
-		if em != nil {
-			if err := em.MessageApplied(ctx, ts, rwMsg.Cid(), rwMsg, ret, true); err != nil {
-				return cid.Undef, cid.Undef, xerrors.Errorf("callback failed on reward message: %w", err)
-			}
-		}
-
-		if ret.ExitCode != 0 {
-			return cid.Undef, cid.Undef, xerrors.Errorf("reward application message failed (exit %d): %s", ret.ExitCode, ret.ActorErr)
+		rErr := t.reward(ctx, vmi, em, epoch, ts, params)
```
magik6k, line 227: I'd pass non-serialized params into this function - passing cbor-ized params into a function which in meant to be customizable feels a bit odd

magik6k (commented):
First really quick review pass, will do one more careful one tomorrow;

Generally everything looks good, should be easy to land this.

chain/consensus/compute_state.go
adlrocha: Yep, I was having a hard time deciding the best type to use as params. What would you recommend, a `cbor.Marshaler` or passing directly `reward.BlockRewardParams` and figure out if we need to generalize it more in the future? (for now, `BlockRewardParams` has all the info we need for our use cases).

magik6k: BlockRewardParams for now, can be customized later

adlrocha:
Closing in favor of https://github.com/filecoin-project/lotus/pull/9481 that already targets `master` as `experimental/fvm-m2` will soon be deprecated.

Issue #9276, opened 2022-09-08
Integrate FIP0045 - deal packing 


jennijuju:
FIP0045 specifies, 

```
The miner actor requires that the sector into which a claimed piece is committed must have a
**remaining lifespan that is sufficient to fulfil the minimum term**.
Thus, when a sector terminates at its scheduled expiration epoch, the claim's term has been met.

Similarly, the miner actor requires that sector into which the piece is committed must have a
**remaining lifespan no longer than that which would reach the maximum term**. 
That is, a sector's commitment expiration epoch must fall between the minimum and maximum terms
of any verified claim it holds.

```
[source ](https://github.com/filecoin-project/FIPs/blob/44a623449b8a3ea48d9502dfdc38443303ac8226/FIPS/fip-0045.md?plain=1#L427-L434)

so that means when packing deals/datacap to sectors, we need to make sure

all deal/datacap term min ≤ sector end epoch ≤ all dealdatacap term max

### scenario A:  committing one  deal / sector
only simply need to make sure 
sector_duration ~== term_max

### scenario B:  two or more  deals 

*adding on to the existing packing logic

sector_start == term_start_A == term_start_B…..

sector_duration = MIN(term_max_A, term_max_B...)

When a new comes in
if is_regular && size_fit{
      if deal_duration > sector_duration
                → new sector
     else
                → put 
}
else { //is_verified && size_fit 
      if term_min > sector_duration
             → new sector
     else { //term_min ≤ sector_duration
              if term_max > sector_duration
                 → new sector
             else  //
                  → put
     }
}

jennijuju:
potentially  update `func (m *Sealing) updateInput` ?

PR #9412, opened 2022-10-04
feat: sealing pipeline: Prepare deal assigning logic for FIP-45

magik6k:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/9276

## What this does
1. Add some checks which prevent deals from being assigned to sectors with expirations above termMax
2. Make cc-upgrade sector selection logic aware of termMax, prefer sectors which can fit most queued deals unless there are multiple sectors which can accommodate the same amount of deal data capped at sector size.


storage/pipeline/input.go
```diff
@@ -341,14 +350,25 @@ func (m *Sealing) SectorAddPieceToAny(ctx context.Context, size abi.UnpaddedPiec
 	return api.SectorOffset{Sector: res.sn, Offset: res.offset.Padded()}, res.err
 }
 
+func (m *Sealing) getClaimTerms(ctx context.Context, deal api.PieceDealInfo) (pieceClaimBounds, error) {
+	// TODO: TODO! TODO! get the real claim bounds here
+
+	return pieceClaimBounds{
+		claimTermEnd: deal.DealProposal.EndEpoch + policy.GetSectorMaxLifetime(abi.RegisteredSealProof_StackedDrg32GiBV1_1, network.Version17),
+	}, nil
+}
```
magik6k, line 359: Only todo, needs some API plumbing, cc @geoff-vball

storage/pipeline/input.go
```diff
@@ -521,80 +562,104 @@ func (m *Sealing) updateInput(ctx context.Context, sp abi.RegisteredSealProof) e
...
+// pendingPieceIndex is an index in the Sealing.pendingPieces map
+type pendingPieceIndex cid.Cid
 
-	for _, piece := range m.pendingPieces {
-		if piece.assigned {
-			continue // already assigned to a sector, skip
-		}
-		candidates = append(candidates, piece)
-	}
+type pieceBound struct {
```
magik6k, line 568: This type makes the assumption that:
* TermMin doesn't matter because DealEnd>=TermMin
  * So if we check that sector expiration is after DealEnd it's also after TermMin
* TermMax is always higher than DealEnd

jennijuju: isnt term min and term max duration?

magik6k: They are, but as far as I understand, as soon as the deal is published, they get locked in to `dealStart+dealDuration` / `dealStart+termMax`, and this logic uses those absolute epochs

storage/pipeline/input.go
geoff-vball: API is in the latest version of `gstuart/integrate-verifreg-changes-fip45`. The api you'll want is `StateGetAllocationForPendingDeal`

storage/pipeline/input.go
```diff
@@ -341,14 +350,25 @@ func (m *Sealing) SectorAddPieceToAny(ctx context.Context, size abi.UnpaddedPiec
 	return api.SectorOffset{Sector: res.sn, Offset: res.offset.Padded()}, res.err
 }
 
+func (m *Sealing) getClaimTerms(ctx context.Context, deal api.PieceDealInfo) (pieceClaimBounds, error) {
+	// TODO: TODO! TODO! get the real claim bounds here
```
ZenGround0, line 354: If you want to defer the work of doing this "the right way" then you can compute these just like f05: https://github.com/filecoin-project/builtin-actors/blob/master/actors/market/src/lib.rs#L1143
https://github.com/filecoin-project/builtin-actors/blob/master/runtime/src/runtime/policy.rs#L408
Its not really any more brittle than "the right way" since the next change to the market <> miner interface is probably going to break both.

Like you probably know already "the right way" is 1. go to market to read map of pending deals to allocation ids 2. go to verifreg and read the allocation table with provider / allocation id

storage/pipeline/input.go
```diff
@@ -611,19 +676,43 @@ func (m *Sealing) maybeUpgradeSector(ctx context.Context, sp abi.RegisteredSealP
...
+			}
+		}
+
 		// if best is below target, we want larger expirations
 		// if best is above target, we want lower pledge, but only if still above target
 
+		// todo: after nv17 "target expiration" doesn't really make that much sense
+		//  (tho to be fair it doesn't make too much sense now either)
+		//  we probably want the lowest expiration that's still above the configured
+		//  minimum, and has can fit most candidate deals
```
ZenGround0, line 709: nit: can fit

ZenGround0 (commented):
Looks ok, I'll review again once todos are gone

storage/pipeline/input.go
geoff-vball: I've already implemented the API to do this "the right way" so I think it's better to do that.

api/api_full.go
```diff
@@ -528,7 +528,8 @@ type FullNode interface {
 	StateMarketDeals(context.Context, types.TipSetKey) (map[string]*MarketDeal, error) //perm:read
 	// StateMarketStorageDeal returns information about the indicated deal
 	StateMarketStorageDeal(context.Context, abi.DealID, types.TipSetKey) (*MarketDeal, error) //perm:read
-	// StateGetAllocationForPendingDeal returns the allocation for a given deal ID of a pending deal.
+	// StateGetAllocationForPendingDeal returns the allocation for a given deal ID of a pending deal. Returns nil if
+	// pending allocation is not found.
```
magik6k, line 532: Q: Can an allocation be not found even when there's a pending allocation in the market actor? This is an assumption I'm making in the impl below.

ZenGround0: Unless we've made a programmer error then no, there will always be an allocation there.

chain/actors/builtin/market/market.go
```diff
@@ -127,7 +127,7 @@ type State interface {
 	) (weight, verifiedWeight abi.DealWeight, err error)
 	NextID() (abi.DealID, error)
 	GetState() interface{}
-	GetAllocationIdForPendingDeal(dealId abi.DealID) (verifregtypes.AllocationId, error)
+	GetAllocationIdForPendingDeal(dealId abi.DealID) (*verifregtypes.AllocationId, error)
```
ZenGround0, line 130: Instead of using nil you shoulld use the zero value which stands for invalid allocation.  Updated go-state-types to help out: https://github.com/filecoin-project/go-state-types/pull/93

ZenGround0 (approved):

ZenGround0:
FYI @jennijuju this is the kind of thing we definitely want an integration test for after code freeze

jennijuju:
> FYI @jennijuju this is the kind of thing we definitely want an integration test for after code freeze

noted! ideally we shall have an integration test WITH the pr.. (but given we are cut close to the ddl I will create an follow up issue! thanks for the ping.

ZenGround0 (approved):

storage/pipeline/input.go
```diff
@@ -28,8 +29,13 @@ import (
 
 func (m *Sealing) handleWaitDeals(ctx statemachine.Context, sector SectorInfo) error {
 	var used abi.UnpaddedPieceSize
+	var lastDealEnd abi.ChainEpoch
 	for _, piece := range sector.Pieces {
 		used += piece.Piece.Size.Unpadded()
+
+		if piece.DealInfo.DealProposal.EndEpoch > lastDealEnd {
```
zl03jsj, line 36: Because piece.DealInfo is a pointer, should check if it is nil first?

zl03jsj, line 36: Because piece.DealInfo is a pointer, should check if it is nil first?

arajasek: Thank you for the review!

arajasek (approved):

PR #9458, opened 2022-10-09
Fix(security) : Path Traversal

bhaskarvilles:
Unsanitized input from a CLI argument flows into io.ioutil.ReadFile, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to read arbitrary files.

Data is 'tainted' if it comes from an insecure source '(os.Args[2])' such as a file, the network, or the user.


magik6k:
inline-gen is a codegen script which is only called as a build step, it doesn't need to check it's inputs.

PR #9465, opened 2022-10-11
feat: build: Embed some more build information in binaries

magik6k:
Also make the `version` command work in offline mode.


rjan90:
@magik6k Want to keep this around/finish it soon?

Issue #9494, opened 2022-10-14
Introduce GO_VERSION file as single source of truth for external and internal build automation


ianconsolata:

smagdali:
This is great, but to clarify- Is this advising people that they should only build with this exact version, or is it a minimum? If exact, what advice should we give people using Go for other things on their systems that might require later versions?

I'll update the documentation when I do my next pass.

ianconsolata:
I think for build automation, it should be exact. This is mostly designed for build automation. I think for folks building manually, they should only use the version listed in the go mod file (i.e. anything `1.18`). The go mod docs don't let you specify a max version, and make it sound like you should always be able to use a later version to run older stuff -- but clearly that's not the case with our project.

Go makes it pretty easy to install multiple versions without clobbering the main version: https://go.dev/doc/manage-install. If we don't want folks to do that manually (which is kind of a hassle), I could tweak the makefile to download the specific version of go mentioned in the GO_VERSION file as a dependency, and then use that binary for building instead of the default `go` binary.

ianconsolata:
Oh, there is also a Go version check in the makefile that checks to make sure the version is above the minimum version, we could also just modify that so it errors if the version is too high as well.

ianconsolata:
We could also start testing higher go versions to make sure the min is actually a min (i.e. use both 1.18 and 1.19) in tests, but that is a much bigger burden on the team, so I wouldn't want to do that unless that's the direction @jennijuju wants to go.

smagdali:
Based on your comments, I think we're probably fine as is - installing a specific version of go for us sounds like bad practice and additional team burdens are bad. I think a min test is fine.

smagdali:
can you clarify/expand? which platforms? i was testing this a couple of weeks ago - don't have access to my notes right now

i can retest on monday

ianconsolata:
Renamed to GO_VERSION_MIN and merged with https://github.com/filecoin-project/lotus/pull/9496

ianconsolata:
There is no max version. All go versions above the min version should be
supported. There’s currently a bug with lotus-fountain as you shared, but
the rest builds fine. If you want to use a specific version for you build
automation, you should use GO_VERSION_MIN.

PR #9496, opened 2022-10-14
_ci_ Appimage go1.18.1 fix

ianconsolata:
## Related Issues
The original PR (https://github.com/filecoin-project/lotus/pull/9389) to address this issue didn't remove the old version of golang from the machine image first, which causes errors (https://app.circleci.com/pipelines/github/filecoin-project/lotus/23476/workflows/9956a546-066d-4d95-861e-7a50237a6762/jobs/609677)

## Proposed Changes
This PR fixes that by removing the old version of go. 

It also introduces a new GO_VERSION file that defines the target go version in a single place that external and internal build automation processes can use to determine which version of go should be used when building the project: https://github.com/filecoin-project/lotus/issues/9494

## Checklist

- [X] CI is green


geoff-vball:
@ianconsolata Based on the conversation [here](https://github.com/filecoin-project/lotus/issues/9494#issuecomment-1279103907), I think the file as-is will cause a bit of confusion. I suggest either renaming the file to `GO_VERSION_MIN` or moving it to the `.circleci` folder as-is.

ianconsolata:
Why do we use the minimum supported version to build the project? Will that always be the case? Why not use the latest 1.18 version to build, instead of explicitly using the min version? (Happy to name it whatever, just trying to understand the scope of how the team treats go versions)

geoff-vball:
@ianconsolata All I know is what @magik6k mentioned in slack the other day:

> So far we've been sticking to
Minimum 1 minor version behind
So that users have good amount of time to update and new features have time to stabilize, or are discovered to be buggy in some way
Support latest as soon as it's out
Many people are using Arch Linux, including many devs, and Arch gets new Go releases very quickly

We could definitely use a version higher than our minimum to build. We have a bug right now stopping use from using 1.19.

ianconsolata:
Ok, anywhere else this should be integrated into the build tooling?

geoff-vball:
@ianconsolata Looks good to me!

geoff-vball (approved):

PR #9511, opened 2022-10-18
fix(api): eth/id address conversion

Stebalien:
Specifically, check if we have the expected prefix and return a sane error.


raulk (approved):
Not a fan of the error saying "id-in-eth" (we've been using the term "masked ID address"), but will revisit anyway with f4 addressing.

PR #9525, opened 2022-10-20
feat: state: add a second premigration for nv17

arajasek:
## Related Issues

More work towards optimizing the migration

## Proposed Changes

We move the first premigration back to be 2 hours before the upgrade.
- Motivation: This is the part we _really_ care about. We do not want any users entering the upgrade without at least one premigration, if possible.

We add a second premigration 7.5 minutes before the upgrade.
- Motivation: This covers any state that's changed since the first premigration, and seems to give a 20% speed up to the actual migration time. If this doesn't complete on time, it isn't important.

This PR also adds more functionality to the lotus-shed tools around migrations and state diffs.

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus-shed/migrations.go
+++ b/cmd/lotus-shed/migrations.go
@@ -176,7 +176,7 @@ var migrationsCmd = &cli.Command{
 		fmt.Println("completed round actual (without cache), took ", uncachedMigrationTime)
 
 		if cctx.Bool("check-invariants") {
-			err = checkStateInvariants(ctx, blk.ParentStateRoot, newCid1, bs, blk.Height-1)
+			err = checkMigrationInvariants(ctx, blk.ParentStateRoot, newCid1, bs, blk.Height-1)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


cmd/lotus-shed/migrations.go
```diff
@@ -127,41 +130,65 @@ var migrationsCmd = &cli.Command{
...
+			}
 		}
 
 		return nil
 	},
 }
 
-func checkStateInvariants(ctx context.Context, v8StateRoot cid.Cid, v9StateRoot cid.Cid, bs blockstore.Blockstore) error {
-	actorStore := store.ActorStore(ctx, blockstore.NewTieredBstore(bs, blockstore.NewMemorySync()))
+func checkStateInvariants(ctx context.Context, v8StateRoot cid.Cid, v9StateRoot cid.Cid, bs blockstore.Blockstore, epoch abi.ChainEpoch) error {
```
ZenGround0, line 189: For clarity all instances of `checkStateInvariants` should change to i.e. `checkMigrationInvariants`.  This is the historical name for the pure state invariants check called as a subset of all migration checks in the commented out code below. Calls to check only the v9 pure state invariants should exist as a standalone command.

ZenGround0 (approved):

cmd/lotus-shed/migrations.go
arajasek: Good feedback, thank you!

PR #9532, opened 2022-10-21
feat: use f4 eth addresses wherever possible

Stebalien:
Previously, we'd use embedded ID addresses.

### DIFF
--- a/api/eth_types.go
+++ b/api/eth_types.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/ipfs/go-cid"
 	"github.com/multiformats/go-multihash"
+	"github.com/multiformats/go-varint"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-address"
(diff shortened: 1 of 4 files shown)


Stebalien:
Builds on #9531, merge that first.

api/eth_types.go
```diff
@@ -289,15 +290,41 @@ func (ea EthAddress) ToFilecoinAddress() (address.Address, error) {
 	return addr, nil
 }
 
-func EthAddressFromFilecoinIDAddress(addr address.Address) (EthAddress, error) {
-	id, err := address.IDFromAddress(addr)
-	if err != nil {
-		return EthAddress{}, err
+func TryEthAddressFromFilecoinAddress(addr address.Address, allowId bool) (EthAddress, bool, error) {
```
vyzo, line 293: when would we want to not allow ID?

vyzo (approved):

api/eth_types.go
Stebalien: When we want to check if we _have_ an eth address (take a look at how we use this)

raulk (approved):
This will definitely make addresses feel less foreign in tooling compared to their masked ID addresses.

PR #9613, opened 2022-11-09
feat: storage: Force exit GenerateSingleVanillaProof on cancelled context

magik6k:
## Proposed Changes
* Add some more logs to wdpost cycle logic
* Respect context in local GenerateSingleVanillaProof
  * before it could block forever if underlying storage reads blocked, e.g. with disconnected NFS, which would lead to the entire PoSt failing
  * Now we'll return context error, which should result in only the slow-to-read sectors being skipped; There will also be a big, detailed log message printed.

## Checklist

- [x] CI is green

### DIFF
--- a/storage/paths/local.go
+++ b/storage/paths/local.go
@@ -806,7 +806,7 @@ func (st *Local) GenerateSingleVanillaProof(ctx context.Context, minerID abi.Act
 		log.Errorw("failed to generate valilla PoSt proof before context cancellation", "err", ctx.Err(), "duration", time.Now().Sub(start), "cache-id", cacheID, "sealed-id", sealedID, "cache", cache, "sealed", sealed)
 
 		// this will leave the GenerateSingleVanillaProof goroutine hanging, but that's still less bad than failing PoSt
-		return nil, xerrors.Errorf("failed to generate valilla proof before context cancellation: %w", ctx.Err())
+		return nil, xerrors.Errorf("failed to generate vanilla proof before context cancellation: %w", ctx.Err())
 	}
 }
 
--- a/storage/paths/local.go
+++ b/storage/paths/local.go
@@ -759,20 +759,22 @@ func (st *Local) GenerateSingleVanillaProof(ctx context.Context, minerID abi.Act
 		ProofType: si.SealProof,
 	}
 
-	var cache string
-	var sealed string
+	var cache, sealed, cacheID, sealedID string
+
 	if si.Update {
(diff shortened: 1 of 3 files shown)


storage/paths/local.go
```diff
@@ -789,7 +792,22 @@ func (st *Local) GenerateSingleVanillaProof(ctx context.Context, minerID abi.Act
...
+	}()
+
+	select {
+	case r := <-resCh:
+		return r.Unwrap()
+	case <-ctx.Done():
+		log.Errorw("failed to generate valilla PoSt proof before context cancellation", "err", ctx.Err(), "duration", time.Now().Sub(start), "cache-id", cacheID, "sealed-id", sealedID, "cache", cache, "sealed", sealed)
+
+		// this will leave the GenerateSingleVanillaProof goroutine hanging, but that's still less bad than failing PoSt
+		return nil, xerrors.Errorf("failed to generate valilla proof before context cancellation: %w", ctx.Err())
```
piotr-bochenek, line 809:
```suggestion
		return nil, xerrors.Errorf("failed to generate vanilla proof before context cancellation: %w", ctx.Err())
```

storage/paths/local.go
```diff
@@ -789,7 +792,22 @@ func (st *Local) GenerateSingleVanillaProof(ctx context.Context, minerID abi.Act
...
+
+	select {
+	case r := <-resCh:
+		return r.Unwrap()
+	case <-ctx.Done():
+		log.Errorw("failed to generate valilla PoSt proof before context cancellation", "err", ctx.Err(), "duration", time.Now().Sub(start), "cache-id", cacheID, "sealed-id", sealedID, "cache", cache, "sealed", sealed)
+
+		// this will leave the GenerateSingleVanillaProof goroutine hanging, but that's still less bad than failing PoSt
+		return nil, xerrors.Errorf("failed to generate vanilla proof before context cancellation: %w", ctx.Err())
+	}
```
geoff-vball, line 810: Is there a way to structure this to not leave the goroutine hanging in the error case?

magik6k: It does, but that's better than failing PoSt

magik6k: (in case of context error the channel is buffered, so the goroutine will exit as soon as it writes the result; the channel with an unread result can be GCed)

geoff-vball: Ah okay, I thought we were going to have goroutines piling up. That's totally fine

geoff-vball (approved):

Issue #9396, opened 2022-09-29
Snap Deals always creates an unsealed copy even though "AlwaysKeepUnsealedCopy = false"


jamerduhgamer:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [X] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
lotus version
Daemon:  1.17.0+mainnet+git.2830429ad+api1.5.0
Local: lotus version 1.17.1+mainnet+git.8db6a939c

lotus-miner version
Daemon:  1.17.1+mainnet+git.8db6a939c+api1.5.0
Local: lotus-miner version 1.17.1+mainnet+git.8db6a939c
```

### Describe the Bug

When performing snap deals, an unsealed copy will always be created even though the "AlwaysKeepUnsealedCopy = false".

### Logging Information

```text
[...]

PR #9614, opened 2022-11-09
refactor: sealing: minor refactor of FinalizeReplicaUpdate

arajasek:
## Related Issues

I was investigating #9396, and ran into this confusing piece of code. This PR doesn't actually solve 9396, though.

## Proposed Changes

State the condition when we _want_ to move unsealed files more clearly.

I'm still a little unsure of this method. We might want to refresh the value of `moveUnsealed` after we've called into `FinalizeReplicaUpdate`, since that call might delete the unsealed files. But it should only do that when `keepUnsealed == 0`, so I guess it's fine.

## Checklist

- [ ] CI is green


magik6k (approved):
This looks at least less confusing

PR #8803, opened 2022-06-03
M2.1 (FEVM) incremental delivery branch

arajasek:
This is the M2 development branch, used for incremental development and delivery.

***NOTE*** DO NOT PUSH DIRECTLY IN THIS BRANCH; always use a pr.

Incremental delivery checklist:
- [ ] create release staging branch and pr.
- [ ] update code for changes shipping in the next release in staging branch.
- [ ] update actors bundle
- [ ] run smoke test; simplecoin for now, suite tbd.
- [ ] get tick from raulk or vyzo
- [ ] merge staging branch here
- [ ] create handoff issue in wallaby
- [ ] update wallaby.car

‐--

Reopened #8474 against `release/v1.16.0`.

# DO NOT MERGE

See https://github.com/filecoin-project/ref-fvm/issues/444

Dependencies:
- https://github.com/filecoin-project/builtin-actors/pull/216
- https://github.com/filecoin-project/ref-fvm/pull/485
- https://github.com/filecoin-project/filecoin-ffi/pull/264
- ~https://github.com/filecoin-project/specs-actors/pull/1588~
- https://github.com/filecoin-project/go-state-types/pull/39

PR #9617, opened 2022-11-09
NV18: FEVM implementation

vyzo:
Cherrypicks the `experimental/fvm-m2` (see #8803) branch to bring all FEVM related functionality into nv18.


Stebalien (commented):
First pass.

Stebalien:
From the first pass:

1. Let's pull out the "tipsetcid" logic into a new PR so we can merge that as its own thing.
2. Let's pull out the MinTimestamp bug fixes, and also merge that as a separate PR.
3. Let's pull all the "delegated signature" logic into a separate commit, so we can _avoid_ merging that. We don't have nearly enough checks to feel comfortable merging that to master, and it'll be replaced by account abstraction.

vyzo:
>     1. Let's pull out the "tipsetcid" logic into a new PR so we can merge that as its own thing.
> 

Presumably in the base pr? Ok, I am find with this, even though I'd like to avoid rebases.

>     2. Let's pull out the MinTimestamp bug fixes, and also merge that as a separate PR.
 
Again presumably in the base pr; fine by this, but it should also pick up the tipset to fvm opts threading, which triggers the bug(s).


>     3. Let's pull all the "delegated signature" logic into a separate commit, so we can _avoid_ merging that. We don't have nearly enough checks to feel comfortable merging that to master, and it'll be replaced by account abstraction.

Uh oh, this is easier said than done. It is not cleanly separated in one commit, it's a mess all over the pace with the lisbon rush hacks. So we'll have to be careful how we handle this.

api/eth_types.go
```diff
@@ -0,0 +1,442 @@
...
+	TransactionsRoot EthHash    `json:"transactionsRoot"`
+	ReceiptsRoot     EthHash    `json:"receiptsRoot"`
+	// TODO: include LogsBloom
+	Difficulty    EthUint64 `json:"difficulty"`
+	Number        EthUint64 `json:"number"`
+	GasLimit      EthUint64 `json:"gasLimit"`
+	GasUsed       EthUint64 `json:"gasUsed"`
+	Timestamp     EthUint64 `json:"timestamp"`
+	Extradata     []byte    `json:"extraData"`
+	MixHash       EthHash   `json:"mixHash"`
```
Stebalien, line 127: Shouldn't this just be `hash`?

Stebalien: https://openethereum.github.io/JSONRPC-eth-module#eth_getblockbynumber

api/eth_types.go
```diff
@@ -0,0 +1,442 @@
...
+
+type EthBlock struct {
+	ParentHash       EthHash    `json:"parentHash"`
+	Sha3Uncles       EthHash    `json:"sha3Uncles"`
+	Miner            EthAddress `json:"miner"`
+	StateRoot        EthHash    `json:"stateRoot"`
+	TransactionsRoot EthHash    `json:"transactionsRoot"`
+	ReceiptsRoot     EthHash    `json:"receiptsRoot"`
+	// TODO: include LogsBloom
+	Difficulty    EthUint64 `json:"difficulty"`
```
Stebalien, line 121: Do we need totalDifficulty? Even if zero?

api/eth_types.go
```diff
@@ -0,0 +1,442 @@
...
+	From             EthAddress  `json:"from"`
+	To               *EthAddress `json:"to"`
+	// Logs
+	// LogsBloom
+	StateRoot         EthHash     `json:"root"`
+	Status            EthUint64   `json:"status"`
+	ContractAddress   *EthAddress `json:"contractAddress"`
+	CumulativeGasUsed EthUint64   `json:"cumulativeGasUsed"`
+	GasUsed           EthUint64   `json:"gasUsed"`
+	EffectiveGasPrice EthBigInt   `json:"effectiveGasPrice"`
```
Stebalien, line 192: Is this a thing? I don't see it in the docs: https://openethereum.github.io/JSONRPC.html#Receipts

api/eth_types.go
```diff
@@ -0,0 +1,442 @@
...
+	decoded, err := hex.DecodeString(s)
+	if err != nil {
+		return err
+	}
+
+	*e = decoded
+	return nil
+}
+
+type EthBlock struct {
```
Stebalien, line 113: Missing `type`?

api/eth_transactions.go
```diff
@@ -0,0 +1,508 @@
...
+	BlockHash            EthHash     `json:"blockHash"`
+	BlockNumber          EthUint64   `json:"blockNumber"`
+	TransactionIndex     EthUint64   `json:"transacionIndex"`
+	From                 EthAddress  `json:"from"`
+	To                   *EthAddress `json:"to"`
+	Value                EthBigInt   `json:"value"`
+	Type                 EthUint64   `json:"type"`
+	Input                EthBytes    `json:"input"`
+	Gas                  EthUint64   `json:"gas"`
+	GasLimit             *EthUint64  `json:"gasLimit,omitempty"`
```
Stebalien, line 40: I can't find this in the docs.

api/eth_transactions.go
```diff
@@ -0,0 +1,508 @@
...
+	MaxFeePerGas         EthBigInt   `json:"maxFeePerGas"`
+	MaxPriorityFeePerGas EthBigInt   `json:"maxPriorityFeePerGas"`
+	V                    EthBytes    `json:"v"`
+	R                    EthBytes    `json:"r"`
+	S                    EthBytes    `json:"s"`
+}
```
Stebalien, line 46: missing: creates, condition, publicKey, raw,

node/impl/full/eth.go
```diff
@@ -0,0 +1,828 @@
...
+	EthAccounts(ctx context.Context) ([]api.EthAddress, error)
+	EthGetBlockTransactionCountByNumber(ctx context.Context, blkNum api.EthUint64) (api.EthUint64, error)
+	EthGetBlockTransactionCountByHash(ctx context.Context, blkHash api.EthHash) (api.EthUint64, error)
+	EthGetBlockByHash(ctx context.Context, blkHash api.EthHash, fullTxInfo bool) (api.EthBlock, error)
+	EthGetBlockByNumber(ctx context.Context, blkNum string, fullTxInfo bool) (api.EthBlock, error)
+	EthGetTransactionByHash(ctx context.Context, txHash *api.EthHash) (*api.EthTx, error)
+	EthGetTransactionCount(ctx context.Context, sender api.EthAddress, blkOpt string) (api.EthUint64, error)
+	EthGetTransactionReceipt(ctx context.Context, txHash api.EthHash) (*api.EthTxReceipt, error)
+	EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash api.EthHash, txIndex api.EthUint64) (api.EthTx, error)
+	EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNum api.EthUint64, txIndex api.EthUint64) (api.EthTx, error)
```
Stebalien, line 43: This response is wrong. It should be a block-ish thing: https://openethereum.github.io/JSONRPC-eth-module#eth_gettransactionbyblockhashandindex

api/eth_transactions.go
Stebalien: gasPrice is also missing.

api/eth_types.go
```diff
@@ -0,0 +1,442 @@
...
+}
+
+type EthCall struct {
+	From     *EthAddress `json:"from"`
+	To       *EthAddress `json:"to"`
+	Gas      EthUint64   `json:"gas"`
+	GasPrice EthBigInt   `json:"gasPrice"`
+	Value    EthBigInt   `json:"value"`
+	Data     EthBytes    `json:"data"`
+}
```
Stebalien, line 165: - nonce
- condition?

https://openethereum.github.io/JSONRPC.html#transactions

vyzo:
Ok, eth apis need some fixing; I'll do a fixup pass.

api/eth_transactions.go
raulk: @ychiaoli18 – do you recall why this was necessary? I wasn't able to find it in the docs either.

api/eth_transactions.go
raulk: > missing: creates, condition, publicKey, raw,

These are non-standard fields. I was able to find them in the OpenEthereum docs, but OpenEthereum is Parity's former Eth1 client, which they abandoned. (Yes, it's a misleading name)

A good reference are the official [Eth docs](https://ethereum.org/es/developers/docs/apis/json-rpc/#eth_gettransactionbyhash) (theory), constrating them with the [Infura docs](https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_gettransactionbyhash) (a more practical view).

raulk: > gasPrice is also missing.

This is an omission, which hasn't bit us yet but it will. So yes, should add. cc @ychiaoli18

api/eth_transactions.go
```diff
@@ -0,0 +1,508 @@
...
+	var to address.Address
+	var params []byte
+
+	if tx.To == nil && tx.Input == nil {
+		return nil, fmt.Errorf("to and input cannot both be empty")
+	}
+
+	var method abi.MethodNum
+	if tx.To == nil {
+		// TODO https://github.com/filecoin-project/ref-fvm/issues/992
```
raulk, line 129:
```suggestion
```

api/eth_types.go
raulk: Yep: https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_getblockbynumber

api/eth_types.go
raulk: I believe so: https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_getblockbynumber.

api/eth_types.go
raulk: Nope, those are OpenEthereum-specific. See https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_call

However, we may need to allow these optional fields:
- maxPriorityFeePerGas
- maxFeePerGas

Not sure why they are accepted as part of eth_call or eth_estimateGas.

api/eth_types.go
raulk: Those docs are not updated because the OpenEthereum client was deprecated before EIP-1559 kicked in. But this field was introduced then: https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_gettransactionreceipt

api/eth_types.go
```diff
@@ -0,0 +1,442 @@
...
+	if lookup.Receipt.ExitCode.IsSuccess() {
+		receipt.Status = 1
+	}
+	if lookup.Receipt.ExitCode.IsError() {
+		receipt.Status = 0
+	}
+
+	receipt.GasUsed = EthUint64(lookup.Receipt.GasUsed)
+
+	// TODO: handle CumulativeGasUsed
```
raulk, line 229: Need to track a TODO for this. cc @ychiaoli18

chain/consensus/filcns/filecoin.go
```diff
@@ -433,6 +436,19 @@ func (filec *FilecoinEC) VerifyWinningPoStProof(ctx context.Context, nv network.
...
+		return true
+	}
+
+	// HACK: Allow Eth embryos to send messages
+	if !builtin.IsEmbryo(act.Code) || act.Address == nil || act.Address.Protocol() != address.Delegated {
+		return false
+	}
+	id, _, err := varint.FromUvarint(act.Address.Payload())
+	return err == nil && id == builtintypes.EthereumAddressManagerActorID
+}
```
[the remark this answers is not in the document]
raulk: Yep, we'll need to keep working off the staging branch until we have account abstraction.

chain/consensus/filcns/upgrades.go
```diff
@@ -1557,8 +1557,8 @@ func UpgradeActorsV10(ctx context.Context, sm *stmgr.StateManager, cache stmgr.M
 	root cid.Cid, epoch abi.ChainEpoch, ts *types.TipSet) (cid.Cid, error) {
 
 	// TODO migration
-	// - the init actor state to include the (empty) installed actors field
-	// - state tree migration to v5
+	// - state tree migration to v5; must set predictable addresses for all account actors
+	// - create EAM actor
```
raulk, line 1561: We want to create Eth address `0x00..00` too as part of the migration. This address is defined by spec. cc @Stebalien

chain/vm/vm.go
```diff
@@ -56,12 +56,19 @@ func ResolveToKeyAddr(state types.StateTree, cst cbor.IpldStore, addr address.Ad
 		return address.Undef, xerrors.Errorf("failed to find actor: %s", addr)
 	}
 
+	if state.Version() >= types.StateTreeVersion5 {
+		if act.Address == nil {
+			return address.Undef, xerrors.Errorf("actor doesn't have expected address: %s", addr)
```
raulk, line 61: The error refers to lacking a _predictable_ address, right? Also switching the order for a clearer message.

```suggestion
			return address.Undef, xerrors.Errorf("actor at %s doesn't have a predictable address", addr)
```

go.mod
```diff
@@ -133,7 +133,7 @@ require (
 	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	github.com/urfave/cli/v2 v2.8.1
 	github.com/whyrusleeping/bencher v0.0.0-20190829221104-bb6607aa8bba
-	github.com/whyrusleeping/cbor-gen v0.0.0-20220514204315-f29c37e9c44c
+	github.com/whyrusleeping/cbor-gen v0.0.0-20221021053955-c138aae13722
```
raulk, line 136: We now have a proper release for this, will push a commit shortly.

node/impl/full/eth.go
```diff
@@ -0,0 +1,828 @@
...
+		return nil, fmt.Errorf("failed to construct system sender address: %w", err)
+	}
+
+	// TODO super duper hack (raulk). The EVM runtime actor uses the U256 parameter type in
+	//  GetStorageAtParams, which serializes as a hex-encoded string. It should serialize
+	//  as bytes. We didn't get to fix in time for Iron, so for now we just pass
+	//  through the hex-encoded value passed through the Eth JSON-RPC API, by remarshalling it.
+	//  We don't fix this at origin (builtin-actors) because we are not updating the bundle
+	//  for Iron.
+	tmp, err := position.MarshalJSON()
```
raulk, line 312: Ideally we'd fix https://github.com/filecoin-project/ref-fvm/issues/912.

cli/chain.go
```diff
@@ -1509,3 +1515,256 @@ var ChainPruneCmd = &cli.Command{
...
+			})
+			if err != nil {
+				return fmt.Errorf("failed to serialize Create params: %w", err)
+			}
+			method = builtintypes.MethodsEAM.Create
+		} else {
+			var salt [32]byte
+			binary.BigEndian.PutUint64(salt[:], nonce)
+
+			params, err = actors.SerializeParams(&eam.Create2Params{
```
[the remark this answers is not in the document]
vyzo: I'll add a TODO

api/eth_transactions.go
```diff
@@ -0,0 +1,508 @@
...
+func (tx *EthTxArgs) ToSignedMessage() (*types.SignedMessage, error) {
+	from, err := tx.Sender()
+	if err != nil {
+		return nil, err
+	}
+
+	var to address.Address
+	var params []byte
+
+	if tx.To == nil && tx.Input == nil {
```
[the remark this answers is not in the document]
vyzo: done.

api/eth_transactions.go
```diff
@@ -0,0 +1,508 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return append([]byte{0x02}, encoded...), nil
+}
+
+func (tx *EthTxArgs) Signature() (*typescrypto.Signature, error) {
+	if tx.V == nil || tx.R == nil || tx.S == nil {
+		return nil, fmt.Errorf("one of V, R, or S is nil")
+	}
```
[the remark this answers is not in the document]
vyzo: removed.

api/eth_transactions.go
```diff
@@ -0,0 +1,508 @@
...
+		Value:                value,
+		Input:                input,
+		R:                    padLeadingZeros(R, 32),
+		S:                    padLeadingZeros(S, 32),
+		V:                    V,
+	}
+	return &args, nil
+}
+
+func ParseEthTxArgs(data []byte) (*EthTxArgs, error) {
```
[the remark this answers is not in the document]
vyzo: done.

api/eth_transactions.go
```diff
@@ -0,0 +1,508 @@
...
+	}
+	return val, nil
+}
+
+func parseEthAddr(v interface{}) (*EthAddress, error) {
+	b, err := parseBytes(v)
+	if err != nil {
+		return nil, err
+	}
+	if b == nil || len(b) == 0 {
```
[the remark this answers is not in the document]
vyzo: removed.

api/rlp.go
```diff
@@ -0,0 +1,176 @@
...
+			break
+		}
+	}
+
+	res := buf.Bytes()[firstNonZeroIndex:]
+	return res, nil
+}
+
+func encodeRLP(val interface{}) ([]byte, error) {
+	if data, ok := val.([]byte); ok {
```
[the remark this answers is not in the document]
vyzo: done.

api/rlp.go
```diff
@@ -0,0 +1,176 @@
...
+		return 0, xerrors.Errorf("invalid rlp data: out of bound while parsing list")
+	}
+	return int(decodedLength), nil
+}
+
+func decodeListElems(data []byte, length int) (res []interface{}, err error) {
+	totalConsumed := 0
+	result := []interface{}{}
+
+	// set a limit to make sure it doesn't loop infinitely
```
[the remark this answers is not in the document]
vyzo: done.

build/params_wallaby.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+var NetworkBundle = "wallaby"
+var BundleOverrides map[actors.Version]string
+var ActorDebugging = true
+
+const BootstrappersFile = "wallabynet.pi"
+const GenesisFile = "wallabynet.car"
+
+const GenesisNetworkVersion = network.Version16
```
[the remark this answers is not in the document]
vyzo: fixed.

chain/gen/genesis/miners.go
```diff
@@ -459,7 +459,6 @@ func SetupStorageMiners(ctx context.Context, cs *store.ChainStore, sys vm.Syscal
 
 					paramBytes = mustEnc(confirmParams)
 				}
-
 				_, err = doExecValue(ctx, genesisVm, minerInfos[i].maddr, power.Address, big.Zero(), builtintypes.MethodsMiner.ConfirmSectorProofsValid, paramBytes)
```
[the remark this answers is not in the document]
vyzo: done.

chain/store/store.go
```diff
@@ -649,6 +650,12 @@ func (cs *ChainStore) takeHeaviestTipSet(ctx context.Context, ts *types.TipSet)
 		return nil
 	}
 
+	tskBlk, err := ts.Key().ToStorageBlock()
+	if err != nil {
+		log.Errorf("failed to create a block from tsk: %s", ts.Key())
+	}
+	_ = cs.chainLocalBlockstore.Put(ctx, tskBlk)
```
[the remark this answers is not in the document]
vyzo: done, we also have to return early in the failure above.

chain/store/store.go
```diff
@@ -649,6 +650,12 @@ func (cs *ChainStore) takeHeaviestTipSet(ctx context.Context, ts *types.TipSet)
 		return nil
 	}
 
+	tskBlk, err := ts.Key().ToStorageBlock()
+	if err != nil {
+		log.Errorf("failed to create a block from tsk: %s", ts.Key())
```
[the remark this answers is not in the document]
vyzo: yup.

cli/chain.go
```diff
@@ -1509,3 +1515,256 @@ var ChainPruneCmd = &cli.Command{
 		return api.ChainPrune(ctx, opts)
 	},
 }
+
+// TODO: Find a home for this.
+func isNativeEthereumAddress(addr address.Address) bool {
```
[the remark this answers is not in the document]
vyzo: done.

lib/sigs/secp/init.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/minio/blake2b-simd"
 
 	"github.com/filecoin-project/go-address"
-	"github.com/filecoin-project/go-crypto"
+	gocrypto "github.com/filecoin-project/go-crypto"
```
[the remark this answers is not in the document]
vyzo: done.

chain/consensus/filcns/upgrades.go
vyzo: will add a comment here.

chain/vm/vm.go
vyzo: yar, will fix.

chain/consensus/filcns/filecoin.go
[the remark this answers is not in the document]
vyzo: yeah, and that's why I was so adamant on review here. I didn't know what the status of delegated signing was (answer: transient, will go away once we have account abstraction proper).

chain/store/store.go
```diff
@@ -1165,6 +1172,24 @@ func (cs *ChainStore) GetTipsetByHeight(ctx context.Context, h abi.ChainEpoch, t
 	return cs.LoadTipSet(ctx, lbts.Parents())
 }
 
+func (cs *ChainStore) GetTipSetByCid(ctx context.Context, c cid.Cid) (*types.TipSet, error) {
```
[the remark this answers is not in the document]
vyzo: I kind of agree with @raulk here, as it would be dead code if we lifted it.
But ultimately it is @arajasek 's decision; master yushie your input is requestd here.

cli/chain.go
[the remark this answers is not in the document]
vyzo: has TODO now, will fix it when we pack a new bundle.

api/rlp_test.go
```diff
@@ -0,0 +1,190 @@
...
+	require.NoError(t, err)
+	orig, err := decoded.OriginalRlpMsg()
+	fmt.Println(hex.EncodeToString(orig))
+
+	// correct f4 addr: f410fkkld55ioe7qg24wvt7fu6pbknb56ht7pt4zamxa
+
+	fmt.Println(decoded.Sender())
+
+	fmt.Printf("%+v\n", decoded)
+}
```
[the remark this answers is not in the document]
vyzo: yar.

chain/actors/builtin/embryo.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+func IsEmbryo(c cid.Cid) bool {
+	name, _, ok := actors.GetActorMetaByCode(c)
+	if ok {
+		return name == "embryo"
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
vyzo: lol.

chain/vm/vm.go
vyzo: done.

chain/consensus/filcns/upgrades.go
vyzo: done.

api/eth_transactions.go
vyzo: done.

chain/actors/builtin/embryo.go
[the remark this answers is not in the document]
vyzo: fixed.

chain/store/store.go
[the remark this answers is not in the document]
Stebalien: Basically, I'm trying to reduce our final patch to "FEVM only" changes to make it easier to review.

chain/types/tipset.go
```diff
@@ -196,8 +196,19 @@ func (ts *TipSet) MinTicket() *Ticket {
 }
 
 func (ts *TipSet) MinTimestamp() uint64 {
```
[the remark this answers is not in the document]
Stebalien: My point is: we might as well land bug fixes separately. We really _don't_ want to revert something and lose a bug fix in the process.

api/eth_types.go
Stebalien: /ihateeverything

api/eth_types.go
ychiaoli18: I believe [this](https://ethereum.github.io/execution-apis/api-documentation/) is the official doc and is more up-to-date. It does have a field called `mixHash`, which is different from `hash` (block hash). It seems that `hash` is missing and I'll add it to the struct. 

Same for the [go-ethereum implementation](https://github.com/ethereum/go-ethereum/blob/c4a662176ec11b9d5718904ccefee753637ab377/core/types/block.go#L84)

chain/types/tipset.go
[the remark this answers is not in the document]
vyzo: yes, sure, i have no problem lifting this to the base pr, its commits are clean.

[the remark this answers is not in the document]
vyzo: also the bug only manifests itself with the new code (it was there all along though, just waiting to be triggered)

api/eth_types.go
ychiaoli18: added

api/eth_types.go
ychiaoli18: @Stebalien not sure what's missing?

api/eth_transactions.go
ychiaoli18: It seems that this was included accidentally. I'll remove it.

node/impl/full/eth.go
Stebalien: Done.

api/eth_transactions.go
ychiaoli18: @raulk It seems that `gasPrice` is only included in the legacy and 2930 transactions but not 1559 transactions ([doc](https://ethereum.github.io/execution-apis/api-documentation/)). Therefore we probably don't need to add this field?

node/impl/full/eth.go
ychiaoli18: It seems that it should return a tx object in the doc?

api/eth_types.go
ychiaoli18: I don't think they are needed here, because they'll be discarded anyway when we turn this EthCall into a Filecoin message. It won't cause any compatibility issues because users can specify those additional parameters if they want, it's just that they won't be used.

node/impl/full/eth.go
Stebalien: Sorry, linked to the wrong docs: https://openethereum.github.io/JSONRPC-eth-module#eth_getTransactionByBlockNumberAndIndex

api/eth_types.go
Stebalien: I'm not sure what I was looking at here either (probably the wrong type).

build/bootstrap/wallabynet.pi
```diff
@@ -0,0 +1,4 @@
+/dns4/de0.bootstrap.wallaby.network/tcp/1337/p2p/12D3KooWHAvUVk5XuxSwi2dNLWbTDDRSGeHxMuWdQ3SQpRuNHbLz
```
jennijuju, line 1: Do not merge

build/params_wallaby.go
```diff
@@ -0,0 +1,93 @@
+//go:build wallabynet
```
jennijuju, line 1: Do not merge https://filecoinproject.slack.com/archives/CP50PPW2X/p1668792668315419

jennijuju (dismissed):

chain/vm/vm.go
```diff
@@ -47,7 +47,7 @@ var (
 
 // ResolveToKeyAddr returns the public key type of address (`BLS`/`SECP256K1`) of an account actor identified by `addr`.
```
arajasek, line 48: fix comment

arajasek: and name, as steb says below

chain/vm/invoker.go
```diff
@@ -284,7 +284,8 @@ func DecodeParams(b []byte, out interface{}) error {
 }
 
 func DumpActorState(i *ActorRegistry, act *types.Actor, b []byte) (interface{}, error) {
-	if builtin.IsAccountActor(act.Code) { // Account code special case
+	// Account & Embryo code special case
+	if builtin.IsAccountActor(act.Code) || builtin.IsEmbryoActor(act.Code) {
```
arajasek, line 288: This shouldn't be necessary, even for the account case.

cli/chain.go
```diff
@@ -1504,3 +1511,298 @@ var ChainPruneCmd = &cli.Command{
 		return api.ChainPrune(ctx, opts)
 	},
 }
+
+// TODO: Find a home for this.
```
arajasek, line 1515: Yeah, we have this in 7 different places, we should resolve this before merging.

cli/chain.go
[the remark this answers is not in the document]
arajasek: Needs resolution.

cli/chain.go
```diff
@@ -63,6 +68,8 @@ var ChainCmd = &cli.Command{
 		ChainEncodeCmd,
 		ChainDisputeSetCmd,
 		ChainPruneCmd,
+		ChainExecEVMCmd,
```
arajasek, line 71: `chain` is not the right place for this, can they go into the new `eth` commands?

arajasek: And rename execevm to create-evm or some such.

cli/state.go
```diff
@@ -776,6 +776,7 @@ var StateGetActorCmd = &cli.Command{
 		fmt.Printf("Nonce:\t\t%d\n", a.Nonce)
 		fmt.Printf("Code:\t\t%s (%s)\n", a.Code, strtype)
 		fmt.Printf("Head:\t\t%s\n", a.Head)
+		fmt.Printf("Predictable address:\t\t%s\n", a.Address)
```
arajasek, line 779: I'm told @Stebalien will cry if anyone says the phrase "predictable address".

cmd/lotus-miner/sectors.go
```diff
@@ -1119,31 +1119,21 @@ var sectorsExtendCmd = &cli.Command{
 	ArgsUsage: "<sectorNumbers...>",
 	Flags: []cli.Flag{
 		&cli.Int64Flag{
-			Name:     "new-expiration",
-			Usage:    "new expiration epoch",
-			Required: false,
+			Name:  "new-expiration",
+			Usage: "new expiration epoch",
```
arajasek, line 1123: Sholud move into `master` directly.

cmd/lotus-miner/sectors.go
```diff
@@ -1119,31 +1119,21 @@ var sectorsExtendCmd = &cli.Command{
...
 		},
 		&cli.BoolFlag{
-			Name:     "v1-sectors",
-			Usage:    "renews all v1 sectors up to the maximum possible lifetime",
-			Required: false,
-		},
-		&cli.Int64Flag{
```
arajasek, line 1131: I don't think we should have dropped this?

gen/main.go
```diff
@@ -29,12 +29,14 @@ func main() {
 		types.MsgMeta{},
 		types.ActorV4{},
 		types.ActorV5{},
-		types.MessageReceipt{},
+		// types.MessageReceipt{}, // Custom serde to deal with versioning.
```
arajasek, line 32: Just remove this entirely?

go.mod
arajasek: Still TODO

itests/eth_deploy_test.go
```diff
@@ -0,0 +1,164 @@
...
+	fmt.Println(hash)
+
+	mpoolTx, err := client.EthGetTransactionByHash(ctx, &hash)
+	require.NoError(t, err)
+
+	// require that the hashes are identical
+	require.Equal(t, hash, mpoolTx.Hash)
+
+	// these fields should be nil because the tx hasn't landed on chain.
+	require.Nil(t, mpoolTx.BlockNumber)
```
arajasek, line 106: This can't land as-is. We can either skip the assertion if the msg has already landed (see [here](https://github.com/filecoin-project/lotus/blob/f04aca767f5362868ea3d680c6a3538b4d1babcd/itests/migration_nv17_test.go#L241-L243) for an example), or pause mining between the embryo creation and this assertion (see [here](https://github.com/filecoin-project/lotus/pull/9933#discussion_r1056871841) for more discussion).

itests/eth_deploy_test.go
```diff
@@ -0,0 +1,164 @@
...
+	require.Nil(t, mpoolTx.BlockHash)
+	require.Nil(t, mpoolTx.TransactionIndex)
+
+	changes, err := client.EthGetFilterChanges(ctx, pendingFilter)
+	require.NoError(t, err)
+	require.Len(t, changes.Results, 1)
+	require.Equal(t, hash.String(), changes.Results[0])
+
+	var receipt *api.EthTxReceipt
+	for i := 0; i < 20; i++ {
```
arajasek, line 116: The right time to exit this loop isn't after 20 iterations, but when `StateWaitMsg` returns -- let's wait on that event while continuing to make this assertion (more [here](https://github.com/filecoin-project/lotus/pull/9933#discussion_r1056479329)).

itests/eth_filter_test.go
```diff
@@ -0,0 +1,802 @@
...
+			From:  client.DefaultKey.Address,
+			To:    addr,
+			Value: each,
+		}
+
+		sm, err := client.MpoolPushMessage(ctx, msg, nil)
+		require.NoError(t, err)
+		require.EqualValues(t, i, sm.Message.Nonce)
+
+		// FIXME this was here and unused. Use or remove.
```
arajasek, line 91: FIXME, it says.

node/impl/full/chain.go
```diff
@@ -654,6 +656,33 @@ func (a *ChainAPI) ChainBlockstoreInfo(ctx context.Context) (map[string]interfac
 	return info.Info(), nil
 }
 
+// ChainGetEvents returns the events under an event AMT root CID.
+//
+// TODO (raulk) make copies of this logic elsewhere use this (e.g. itests, CLI, events filter).
```
arajasek, line 661: Yes please.

itests/kit/state.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/filecoin-project/go-address"
+	actorstypes "github.com/filecoin-project/go-state-types/actors"
+
+	"github.com/filecoin-project/lotus/chain/actors"
+	"github.com/filecoin-project/lotus/chain/types"
+)
+
+// AssertActorType verifies that the supplied address is an actor of the
+// specified type (as per its manifest key).
+func (f *TestFullNode) AssertActorType(ctx context.Context, addr address.Address, actorType string) {
```
arajasek, line 17: <3

lib/sigs/delegated/eth.go
```diff
@@ -0,0 +1,26 @@
...
+	"fmt"
+
+	"golang.org/x/crypto/sha3"
+)
+
+// EthAddressFromPubKey returns the Ethereum address corresponding to an
+// uncompressed secp256k1 public key.
+//
+// TODO move somewhere else, this likely doesn't belong here.
+func EthAddressFromPubKey(pubk []byte) ([]byte, error) {
```
arajasek, line 13: Assuming it doesn't cause import cycles, the eth types place might be a good home.

arajasek: but honestly i'm...fine with it being here.

node/impl/full/state.go
```diff
@@ -617,14 +618,22 @@ func (m *StateModule) StateWaitMsg(ctx context.Context, msg cid.Cid, confidence
 
 		t, err := stmgr.GetReturnType(ctx, m.StateManager, vmsg.To, vmsg.Method, ts)
 		if err != nil {
-			return nil, xerrors.Errorf("failed to get return type: %w", err)
-		}
-
-		if err := t.UnmarshalCBOR(bytes.NewReader(recpt.Return)); err != nil {
-			return nil, err
+			if errors.Is(err, stmgr.ErrMetadataNotFound) {
+				// This is not nececessary an error -- EVM methods (and in the future native actors) may
```
arajasek, line 622:
```suggestion
				// This is not necessarily an error -- EVM methods (and in the future native actors) may
```

arajasek (commented):
So, this is in pretty good shape. I didn't cover all the new logic in the greatest possible detail (esp the event indexing logic), but I think the consensus-critical path is mostly good to go (and everything else can be improved). There are some blockers, though:

- [ ] PR description (please use the PR template)
- [ ] Change destination branch to `release/v1.20.0`
- [ ] Decision: These hacks have either been dropped or patched, so wallaby can land in this PR
- Decide on whether wallaby lands in the destination branch: 
  - Figure out what wallaby hacks live in the ntwk/wallaby branch, and whether they can be dropped / merged by this point
  - I would prefer to have wallaby land, but don't want that if we'll still need hacks in a different branch to sync anyway
- [ ] NV checks on validity of senders and siggys
- [ ] Open issues for event logic optimizations
- [ ] https://github.com/filecoin-project/ref-fvm/issues/1345 (unbork tests)
- [ ] https://github.com/filecoin-project/ref-fvm/issues/1173
- [ ] Extract EthTxDigest calculations (filcns, messagepool, and maybe other places)
    - [ ] Should the digest be a hash of the RLP encoding? 
- [ ] Stmgr call f099 hack should be dropped
- [ ] Extract the repeated `isNativeEthereumAddress`ing.
- [ ] CLI should use Create, not Create2.
- [ ] CLI eth commands should move out of `chain`
- [ ] Potential test flakiness nononoooo

itests/eth_deploy_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/chain/types/ethtypes"
+	"github.com/filecoin-project/lotus/itests/kit"
+	"github.com/filecoin-project/lotus/node/config"
+)
+
+// TestDeployment smoke tests the deployment of a contract via the
+// Ethereum JSON-RPC endpoint, from an EEOA.
+func TestDeployment(t *testing.T) {
+	// TODO the contract installation and invocation can be lifted into utility methods
```
arajasek, line 27: Either implement or drop this TODO.
Also uncomment the quieting of the mining logs?

itests/eth_deploy_test.go
```diff
@@ -0,0 +1,164 @@
...
+		S:                    big.Zero(),
+	}
+
+	client.EVM().SignTransaction(&tx, key.PrivateKey)
+
+	pendingFilter, err := client.EthNewPendingTransactionFilter(ctx)
+	require.NoError(t, err)
+
+	hash := client.EVM().SubmitTransaction(ctx, &tx)
+	fmt.Println(hash)
```
arajasek, line 97: Remove?

chain/consensus/filcns/filecoin.go
```diff
@@ -433,6 +436,19 @@ func (filec *FilecoinEC) VerifyWinningPoStProof(ctx context.Context, nv network.
 	return nil
 }
 
+func IsValidForSending(act *types.Actor) bool {
```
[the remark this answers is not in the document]
raulk: I went carefully through this method and I don't think this needs a network version gate, because:
1. Placeholder actors cannot be created before nv18.
2. EthAccount actors cannot be created before nv18.
3. Delegated addresses cannot be created before nv18.

Note: if we sprinkle network version gates in random places, we absolutely cannot start Hyperspace with nv1. Also note that this pattern won't interact well with diverging network lineages across networks, without also including the network name in the gate.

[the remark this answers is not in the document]
raulk: I will still add it, but it's entirely superfluous.

api/api_full.go
```diff
@@ -27,10 +27,11 @@ import (
 	abinetwork "github.com/filecoin-project/go-state-types/network"
 
 	apitypes "github.com/filecoin-project/lotus/api/types"
-	"github.com/filecoin-project/lotus/chain/actors/builtin"
+	builtinactors "github.com/filecoin-project/lotus/chain/actors/builtin"
```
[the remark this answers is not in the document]
raulk: Not entirely sure, probably some IDE playing tricks.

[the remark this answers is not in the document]
raulk: Probably to disambiguate this well-known import: pkg.go.dev/builtin

chain/consensus/filcns/filecoin.go
```diff
@@ -543,7 +559,7 @@ func (filec *FilecoinEC) checkBlockMessages(ctx context.Context, b *types.FullBl
 	smArr := blockadt.MakeEmptyArray(tmpstore)
 	for i, m := range b.SecpkMessages {
 		if filec.sm.GetNetworkVersion(ctx, b.Header.Height) >= network.Version14 {
-			if m.Signature.Type != crypto.SigTypeSecp256k1 {
+			if m.Signature.Type != crypto.SigTypeSecp256k1 && m.Signature.Type != crypto.SigTypeDelegated {
```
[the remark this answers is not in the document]
raulk: Done, called `ValidateSecpkSigType`

chain/consensus/filcns/filecoin.go
```diff
@@ -559,7 +575,20 @@ func (filec *FilecoinEC) checkBlockMessages(ctx context.Context, b *types.FullBl
 			return xerrors.Errorf("failed to resolve key addr: %w", err)
 		}
 
-		if err := sigs.Verify(&m.Signature, kaddr, m.Message.Cid().Bytes()); err != nil {
+		digest := m.Message.Cid().Bytes()
+		if m.Signature.Type == crypto.SigTypeDelegated {
```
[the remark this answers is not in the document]
raulk: And, in fact, we are taking the simple route. I've extracted signature verification to a utility method.

api/api_full.go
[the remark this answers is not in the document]
raulk: Fixed in https://github.com/filecoin-project/lotus/pull/9984.

chain/messagepool/messagepool.go
```diff
@@ -795,7 +795,19 @@ func (mp *MessagePool) VerifyMsgSig(m *types.SignedMessage) error {
 		return nil
 	}
 
-	if err := sigs.Verify(&m.Signature, m.Message.From, m.Message.Cid().Bytes()); err != nil {
+	if m.Signature.Type == crypto.SigTypeDelegated {
+		txArgs, err := ethtypes.NewEthTxArgsFromMessage(&m.Message)
```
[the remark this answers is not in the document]
raulk: Extracted in https://github.com/filecoin-project/lotus/pull/9984.

chain/vm/invoker.go
raulk: This is gone now.

cli/chain.go
raulk: Pretty sure that there's a bug here anyway.

cmd/lotus-miner/sectors.go
raulk: I don't think this is coming from this PR?

raulk: It's also fixed in the branch apparently.

raulk: Probably because the branch was rebased.

cmd/lotus-miner/sectors.go
raulk: See below.

itests/eth_filter_test.go
raulk: This was tagged as FIXME by @geoff-vball, but I believe the original intention was to assert that the pending transaction hashes received match the message CIDs of the pushed messages.

raulk: Fixed by strengthening the assertions and actually using this. https://github.com/filecoin-project/lotus/pull/9984

node/impl/full/state.go
raulk: Fixed in https://github.com/filecoin-project/lotus/pull/9984.

node/impl/full/chain.go
raulk: All of the events logic is about to change with https://github.com/filecoin-project/ref-fvm/issues/1376, so this will have to get refactored. For context, ref-fvm no longer writes events to the store and instead just returns them in ApplyRet.

gen/main.go
raulk: If you really want to. I think it's a useful strategic marker, as this case is exceptional due to versioning.

raulk: I'll keep it. If you feel strongly about it, let me know.

chain/types/message_receipt.go
```diff
@@ -3,15 +3,57 @@ package types
...
 type MessageReceipt struct {
-	ExitCode exitcode.ExitCode
-	Return   []byte
-	GasUsed  int64
+	version MessageReceiptVersion
+
+	ExitCode   exitcode.ExitCode
+	Return     []byte
+	GasUsed    int64
+	EventsRoot *cid.Cid // Root of Event AMT
```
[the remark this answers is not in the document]
raulk: Done.

api/api_full.go
```diff
@@ -1252,3 +1327,23 @@ type PruneOpts struct {
...
+
+type EthTxReceipt struct {
+	TransactionHash  ethtypes.EthHash     `json:"transactionHash"`
+	TransactionIndex ethtypes.EthUint64   `json:"transactionIndex"`
+	BlockHash        ethtypes.EthHash     `json:"blockHash"`
+	BlockNumber      ethtypes.EthUint64   `json:"blockNumber"`
+	From             ethtypes.EthAddress  `json:"from"`
+	To               *ethtypes.EthAddress `json:"to"`
+	// Logs
+	// LogsBloom
```
magik6k, line 1339: Is this todo?

chain/gen/gen.go
```diff
@@ -468,10 +468,6 @@ func (cg *ChainGen) NextTipSetFromMinersWithMessagesAndNulls(base *types.TipSet,
 					return nil, xerrors.Errorf("making a block for next tipset failed: %w", err)
 				}
 
-				if err := cg.cs.PersistBlockHeaders(context.TODO(), fblk.Header); err != nil {
-					return nil, xerrors.Errorf("chainstore AddBlock: %w", err)
-				}
-
```
magik6k, line 474: What's the reason for dropping this?

chain/gen/genesis/fevm.go
```diff
@@ -0,0 +1,113 @@
...
+		return cid.Undef, fmt.Errorf("failed to parse ETH0 init code during genesis: %w", err)
+	}
+
+	ctorParams := &evm10.ConstructorParams{
+		Creator: make([]byte, 20), // self!
+		// TODO we have a bunch of bugs in the evm constructor around empty contracts
+		// - empty init code is not allowed
+		// - returning an empty contract is not allowed
+		// So this uses code that constructs a just return contract until that can be fixed
+		// and we can pass an empty byte array
```
magik6k, line 92: Have those bugs been fixed since this todo was written?

chain/types/ethtypes/eth_transactions.go
```diff
@@ -0,0 +1,584 @@
...
+	if err != nil {
+		return nil, err
+	}
+	decoded, ok := d.([]interface{})
+	if !ok {
+		return nil, fmt.Errorf("not an EIP-1559 transaction: decoded data is not a list")
+	}
+
+	if len(decoded) != 9 && len(decoded) != 12 {
+		return nil, fmt.Errorf("not an EIP-1559 transaction: should have 6 or 9 elements in the list")
```
magik6k, line 386: Given that the code below always accesses 12 elements of `decoded` shouldn't this be always checking for 12? Also the error message seems to have different numbers

chain/types/ethtypes/rlp.go
```diff
@@ -0,0 +1,180 @@
...
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+
+	"golang.org/x/xerrors"
+)
+
+// set a limit to make sure it doesn't loop infinitely
+const maxListElements = 500
```
magik6k, line 12: How was this number selected?

chain/types/tipset.go
```diff
@@ -196,8 +196,19 @@ func (ts *TipSet) MinTicket() *Ticket {
...
-	minTs := ts.Blocks()[0].Timestamp
-	for _, bh := range ts.Blocks()[1:] {
+	if ts == nil {
+		return 0
+	}
+
+	blks := ts.Blocks()
+
+	if len(blks) == 0 {
```
magik6k, line 205: How can you ever have an instance of TIpSet with no blocks? I wouldn't expect that to be possible

chain/vm/invoker.go
```diff
@@ -284,16 +284,16 @@ func DecodeParams(b []byte, out interface{}) error {
...
-	}
-
 	actInfo, ok := i.actors[act.Code]
 	if !ok {
 		return nil, xerrors.Errorf("state type for actor %s not found", act.Code)
 	}
 
 	um := actInfo.vmActor.State()
+	if um == nil {
+		// TODO: I would like to assert that we have the empty object here
```
magik6k, line 294: todo

chain/wallet/key/key.go
```diff
@@ -50,6 +52,17 @@ func NewKey(keyinfo types.KeyInfo) (*Key, error) {
 		if err != nil {
 			return nil, xerrors.Errorf("converting Secp256k1 to address: %w", err)
 		}
+	case types.KTDelegated:
+		// Assume eth for now
```
magik6k, line 56: Can we add some standard NOTE comment or something like that to places where we are making assumptions like this?

documentation/en/api-v1-unstable-methods.md
```diff
@@ -1899,232 +1990,1184 @@ Response:
...
+  204,
+  38,
+  79,
+  36,
+  11,
+  176,
+  102,
+  64,
+  49
+]
```
magik6k, line 2937: Some of those responses don't look quite right, are we sure this isn't what's sent over the wire?

documentation/en/cli-lotus.md
```diff
@@ -2089,6 +2090,8 @@ COMMANDS:
      encode                            encode various types
      disputer                          interact with the window post disputer
      prune                             prune the stored chain state and perform garbage collection
+     create-evm-actor                  Create an new EVM actor via the init actor and return its address
+     invoke-evm-actor                  Invoke a contract entry point in an EVM actor
```
magik6k, line 2094: Those command probably belong in the `evm` group. `chain` commands are mostly meant for interacting with the "low-level" chain (think block-headers), then `state` commands are for native filecoin state

documentation/en/default-lotus-config.toml
```diff
@@ -293,3 +293,52 @@
...
+  #MaxFilterResults = 10000
+
+  # MaxFilterHeightRange specifies the maximum range of heights that can be used in a filter (to avoid querying
+  # the entire chain)
+  #
+  # type: uint64
+  # env var: LOTUS_ACTOREVENT_MAXFILTERHEIGHTRANGE
+  #MaxFilterHeightRange = 2880
+
+  # EventHistoryDatabasePath is the full path to a sqlite database that will be used to index actor events to
```
magik6k, line 336: Do we require that this path is absolute? If not, can we specify what is it relative to?

itests/contracts/SimpleCoin.bin
```diff
@@ -0,0 +1 @@
...82169050919050565b60006102d7826102ac565b9050919050565b6102e7816102cc565b81146102f257600080fd5b50565b600081359050610304816102de565b92915050565b6000602082840312156103205761031f6102a7565b5b600061032e848285016102f5565b91505092915050565b6000819050919050565b61034a81610337565b82525050565b60006020820190506103656000830184610341565b92915050565b61037481610337565b811461037f57600080fd5b50565b6000813590506103918161036b565b92915050565b600080604083850312156103ae576103ad6102a7565b5b60006103bc858286016102f5565b92505060206103cd85828601610382565b9150509250929050565b60008115159050919050565b6103ec816103d7565b82525050565b600060208201905061040760008301846103e3565b92915050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601160045260246000fd5b600061044782610337565b915061045283610337565b925082820261046081610337565b915082820484148315176104775761047661040d565b5b5092915050565b600061048982610337565b915061049483610337565b92508282039050818111156104ac576104ab61040d565b5b92915050565b60006104bd82610337565b91506104c883610337565b92508282019050808211156104e0576104df61040d565b5b9291505056fea26469706673582212205ede41ff9072784ccc19ac18de0781558d305a8139361fa85dc51a8614e47d8c64736f6c63430008110033
```
magik6k, line 1: _technically_ this file should be `.hex`, practically this comment can be ignored

.circleci/config.yml
```diff
@@ -726,6 +726,46 @@ workflows:
           suite: itest-dup_mpool_messages
           target: "./itests/dup_mpool_messages_test.go"
       
+      - test:
+          name: test-itest-eth_account_abstraction
+          suite: itest-eth_account_abstraction
+          target: "./itests/eth_account_abstraction_test.go"
+
```
magik6k, line 733: Most of those tests are very short, and instantiating separate runner for each is probably very overkill (and expensive in terms of CI credits). Easiest fix would be to just throw those into a single file, tho maybe not what we want to do

(this can be addressed later)

lib/sigs/delegated/eth.go
magik6k, line 13: Don't we want to also assert the length here?

node/impl/full/state.go
```diff
@@ -617,14 +618,22 @@ func (m *StateModule) StateWaitMsg(ctx context.Context, msg cid.Cid, confidence
 
 		t, err := stmgr.GetReturnType(ctx, m.StateManager, vmsg.To, vmsg.Method, ts)
 		if err != nil {
-			return nil, xerrors.Errorf("failed to get return type: %w", err)
-		}
-
-		if err := t.UnmarshalCBOR(bytes.NewReader(recpt.Return)); err != nil {
-			return nil, err
+			if errors.Is(err, stmgr.ErrMetadataNotFound) {
```
magik6k, line 621: Nit: could `switch` on err instead of doing nested ifs

node/modules/actorevent.go
```diff
@@ -0,0 +1,138 @@
...
+			Chain:    cs,
+			StateAPI: stateapi,
+			ChainAPI: chainapi,
+		}
+		ee.FilterStore = filter.NewMemFilterStore(cfg.MaxFilters)
+
+		// Start garbage collection for filters
+		lc.Append(fx.Hook{
+			OnStart: func(ctx context.Context) error {
+				go ee.GC(ctx, time.Duration(cfg.FilterTTL))
```
magik6k, line 57: The context in `OnStart` is only alive during node startup (for ~5 seconds), we probably want the context we get from `ctx := helpers.LifecycleCtx(mctx, lc)` below

chain/gen/genesis/fevm.go
raulk: Not able to find any of these TODOs in the latest version of the code. Were you reviewing an old commit? (Maybe I could find out by clicking thirough the UI, but the UI is really sluggish on such a big PR) What review strategy are you using here to dig up old TODOs?

chain/types/ethtypes/eth_transactions.go
raulk: Yeah this seems off.

chain/types/ethtypes/rlp.go
raulk: I believe this is an arbitrary backstop number. Given that we use the RLP encoder/decoder for transactions only, we could lower this to something a lot more conservative, like 16 or 32.

chain/types/tipset.go
raulk: @vyzo -- could you explain this?

chain/wallet/key/key.go
raulk: Can do, referencing the transitory Delegated signature verification as defined in FIP-0055.

documentation/en/api-v1-unstable-methods.md
raulk: We are missing a doc example for the EthFilterID type.

documentation/en/default-lotus-config.toml
raulk: Can check.

.circleci/config.yml
raulk: The granularity is too fine, but we're letting these tests develop independently. There are several folks working on them and this strategy reduces conflicts. We can then go and consolidate them pretty easily.

chain/types/tipset.go
vyzo: conf tests bomb wihout this iirc.

vyzo: also, null rounds?

magik6k: Null rounds shouldn't ever be represented as tipsets with no blocks; Null-round generally means that the tipset at that epoch doesn't exist - and the next tipset that does exist links straight to first epoch with blocks

chain/types/ethtypes/eth_transactions.go
raulk: Fixed.

chain/types/ethtypes/rlp.go
raulk: Made is stricter: 12.

documentation/en/cli-lotus.md
raulk: Moved to the `evm` command group.

documentation/en/default-lotus-config.toml
raulk: When a relative path is supplied, it's relative to CWD.

raulk: Documented (and fixed the field name in the comment).

lib/sigs/delegated/eth.go
raulk: This was implemented in recent commits.

node/impl/full/state.go
raulk: Implemented.

raulk:
TODO:
- [ ] https://github.com/filecoin-project/lotus/pull/9617#discussion_r1067262736
- [ ] https://github.com/filecoin-project/lotus/pull/9617#discussion_r1067256318
- [ ] https://github.com/filecoin-project/lotus/pull/9617#discussion_r1067254414
- [ ] https://github.com/filecoin-project/lotus/pull/9617#discussion_r1067212637
- [ ] Fix unit tests

chain/vm/vm.go
raulk: Changed to ResolveToDeterministicAddr. If we don't feel good with this name, please open an issue to discuss and rename.

go.mod
raulk: I believe this was supposed to be pointing to go-jsonrpc and not to cbor-gen. That repo does not do releases.

itests/eth_deploy_test.go
raulk: Let's address at a later stage on master.

chain/gen/gen.go
raulk: @arajasek -- I suspect this change is yours. I am going to add a call to PersistTipset here; I'm assuming that's what you wanted.

raulk: Ah, it's already below. @magik6k -- we now have a PersistTipset method. This collects block headers into a tipset inside the loop, and then persists the tipset outside the loop.

chain/gen/genesis/fevm.go
```diff
@@ -0,0 +1,113 @@
...
+		return cid.Undef, fmt.Errorf("creating vm: %w", err)
+	}
+
+	// The ETH0 address is occupied by an empty contract EVM actor
+	evmCodeCid, ok := actors.GetActorCodeID(av, manifest.EvmKey)
+	if !ok {
+		return cid.Undef, fmt.Errorf("failed to get CodeCID for EVM during genesis")
+	}
+
+	// initcode:
```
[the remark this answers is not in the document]
raulk: This is now gone, but we cannot invoke thke init actor because EthAccounts are not constructable that way.

chain/messagepool/selection_test.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/types/mock"
 	"github.com/filecoin-project/lotus/chain/wallet"
 	_ "github.com/filecoin-project/lotus/lib/sigs/bls"
+	_ "github.com/filecoin-project/lotus/lib/sigs/delegated"
```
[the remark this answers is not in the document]
raulk: Well, it is going into master now.

cli/chain.go
raulk: They're now in the `evm` command group.

cli/chain.go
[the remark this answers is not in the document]
raulk: Done.

itests/eth_deploy_test.go
raulk: Tracking as TODO::FVM

PR #9363, opened 2022-09-23
fix: docker: make compatible with arm platform

airenas:
## Related Issues
Resolves #9362 

## Proposed Changes
1. update rust installation
2. add possibility to set FFI_BUILD_FROM_SOURCE
3. fix hardcoded paths: x86_64-linux-gnu

## Checklist

Before you mark the PR ready for review, please make sure that:
- [x] All commits have a clear commit message.
- [x] The PR title is in the form of of `<PR type>: <area>: <change being made>`
    - example: ` fix: mempool: Introduce a cache for valid signatures`
    - `PR type`: _fix_, _feat_, _INTERFACE BREAKING CHANGE_, _CONSENSUS BREAKING_, _build_, _chore_, _ci_, _docs_,_perf_, _refactor_, _revert_, _style_, _test_
    - `area`: _api_, _chain_, _state_, _vm_, _data transfer_, _market_, _mempool_, _message_, _block production_, _multisig_, _networking_, _paychan_, _proving_, _sealing_, _wallet_, _deps_
- [ ] This PR has tests for new functionality or change in behaviour
- [ ] If new user-facing features are introduced, clear usage guidelines and / or documentation updates should be included in https://lotus.filecoin.io or [Discussion Tutorials.](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] CI is green

PR #9625, opened 2022-11-10
_ci_ Build docker containers automatically for butterflynet, calibnet, and debug

ianconsolata:
## Related Issues
[](https://github.com/filecoin-project/lotus/pull/9363)

## Proposed Changes
This is a major refactor of our dockerfile to support the following
- The lotus image will remain as is.
- The lotus-test image will be deprecated.
- The lotus-all-in-one image will also ship with the lotus-seed and lotus-fountain binaries, which it currently does not.
- The lotus-all-in-one image will be built in debug, calibnet, and butterflynet modes in addition to the (current) mainnet mode.
- The lotus-all-in-one image will now be published regularly using the following tags:
  - 1.18.0-rc1 , 1.18.0-rc1-debug, 1.18.0-rc1-calibnet, 1.18.0-rc1-butterflynet . This pattern will be used for all lotus releases, including RC releases.
  - nightly, nightly-debug, nightly-calibnet, nightly-butterflynet
  - stable, stable-debug, stable-calibnet, stable-butterflynet
  
I left the existing `Dockerfile.lotus` unchanged for backwards compatibility, but we can probably safely remove it after this is merged and released.

This also removes:
- old publish to dockerhub job
- publishing to ECR (a private amazon container registry like dockerhub that no one seems to be using anymore).

## Checklist

- [X] CI is green


geoff-vball (approved):
@ianconsolata Looks good, just needs a `make gen`

Dockerfile
```diff
@@ -0,0 +1,133 @@
...
+FROM lotus-base AS lotus
+MAINTAINER Lotus Development Team
+
+COPY --from=lotus-builder /opt/filecoin/lotus /usr/local/bin/
+COPY --from=lotus-builder /opt/filecoin/lotus-shed /usr/local/bin/
+COPY scripts/docker-lotus-entrypoint.sh /
+
+ENV FILECOIN_PARAMETER_CACHE /var/tmp/filecoin-proof-parameters
+ENV LOTUS_PATH /var/lib/lotus
+ENV DOCKER_LOTUS_IMPORT_SNAPSHOT https://fil-chain-snapshots-fallback.s3.amazonaws.com/mainnet/minimal_finality_stateroots_latest.car
```
magik6k, line 76: We should swap this to the new snapshots - `https://snapshots.${network}.filops.net/minimal/latest`, e.g. https://snapshots.mainnet.filops.net/minimal/latest

(not sure how it's used under the hood, if the url needs to be resolved first, that can be done with - ` curl -ILs -w %{url_effective} -o /dev/null "https://snapshots.mainnet.filops.net/minimal/latest"`)

The new snapshots have a slightly more robust setup, and come from cloudflare R2, instead of S3, and are usually much faster to fetch (I can fetch them at 2Gbps in EU, which is all my internet can do, fetching from S3 is 10x slower)

Dockerfile
```diff
@@ -0,0 +1,133 @@
+#####################################
+FROM golang:1.18.1-buster AS lotus-builder
```
magik6k, line 2: Probably want a newer Go version

magik6k (commented):
2 Nits, but generally looks good.

(also pushed a make-gen fix)

Dockerfile
ianconsolata: 1.18.1 is what we currently build all our builds with (see https://github.com/filecoin-project/lotus/blob/master/.circleci/config.yml#L9 and https://github.com/filecoin-project/lotus/blob/master/GO_VERSION_MIN).

 If we want a newer version, we should update that across the board, and probably in a separate PR.

Dockerfile
ianconsolata: Ok, right now we don't really support building the `lotus` image with anything but mainnet, so I will make this default to the mainnet string, and then whoever is running the container can supply a different string if they built it for that network.

I could also extend this image to support supplying a network parameter, instead of the currently method which relies on the docker build command to supply GOFLAGS to build a current image, but that's more work and I think this method is just as configurable.

Dockerfile
arajasek: Discussed in standup: @ianconsolata can you please update to 1.18.8, and enforce it in the Makefile (in a separate PR)?

ianconsolata: Once that one is merged, I will update this PR to include the changes.

ianconsolata:
https://github.com/filecoin-project/lotus/pull/9712 is merged, and I brought those changes over to this PR, so I think this is good to go too now?

geoff-vball (approved):

Issue #9626, opened 2022-11-10
Remove exposing full node libp2p address as a requirement for index-provider


LexLuthr:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated improvement to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [X] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Improvement Suggestion

Currently, index-provider requires full node to be publicly exposed. AFAIK, no other lotus-miner module has thing requirement and this becomes another security concern.

If the libp2p address is not reachable we get the below error. This behaviour is consistent on Boost as well. 

```
failed to connect index provider host with the full node: failed to connect index provider host with the full node: failed to dial 12D3KooWJgxyYahRiJ6rHUJPeigWoXJ8SwFKNKHSxXje1D1BvGVR:
  * [/ip4/76.219.232.45/tcp/24002] dial tcp4 76.219.232.45:24002: connect: connection refused
waiting for legacy storage provider 'ready' event 
```

We need to change the direction of libp2p connection init. Rather than dialling from market to fullNode, we should dial from fullNode to market. Thus removing the requirement to have libp2p exposed.

PR #9627, opened 2022-11-10
fix: index-provider: change connection initiation direction between daemon and market

LexLuthr:
## Related Issues

https://github.com/filecoin-project/lotus/issues/9626

## Proposed Changes

We need to change the direction of libp2p connection init. Rather than dialling from market to fullNode, we should dial from fullNode to market. Thus removing the requirement to have libp2p exposed.

## Checklist

- [x] CI is green

### DIFF
--- a/markets/idxprov/mesh.go
+++ b/markets/idxprov/mesh.go
@@ -38,10 +38,13 @@ func (mc Libp2pMeshCreator) Connect(ctx context.Context) error {
 		return fmt.Errorf("failed to fetch full node listen addrs, err: %w", err)
 	}
 
-	// Connect to the full node, ask it to protect the connection and protect the connection on
-	// markets end too.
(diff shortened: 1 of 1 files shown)


magik6k (commented):
This makes sense.

We now don't need to gen `faddrs` above, also should fix the log below.

LexLuthr:
> This makes sense.
> 
> We now don't need to gen `faddrs` above, also should fix the log below.

@magik6k As per my understanding, we still need to protect the connection from both sides. `faddrs` is being used for that. 
Can you confirm which log do I need to change?

magik6k:
* Log - the `	log.Debugw("successfully connected to full node and asked it protect indexer provider peer conn", "fullNodeInfo", faddrs.String(), "peerId", marketsPeerID)` one - tho reading this code today it probably doesn't make sense to change it.
* Oh didn't notice the conn protection part

magik6k (approved):

PR #9635, opened 2022-11-14
NV18: FEVM: Basic smoke test

vyzo:
The basic FEVM smoke test with SimpleCoin.


node/impl/full/state.go
```diff
@@ -617,14 +617,15 @@ func (m *StateModule) StateWaitMsg(ctx context.Context, msg cid.Cid, confidence
...
 		t, err := stmgr.GetReturnType(ctx, m.StateManager, vmsg.To, vmsg.Method, ts)
 		if err != nil {
-			return nil, xerrors.Errorf("failed to get return type: %w", err)
-		}
-
-		if err := t.UnmarshalCBOR(bytes.NewReader(recpt.Return)); err != nil {
-			return nil, err
+			// This is not nececessary an error -- EVM methods (and in the future native actors) may
+			//  return just bytes.
+			log.Debugf("failed to get return type: %s", err)
```
maciejwitowski, line 622: Besides the unknown return type, after this change, we will also ignore errors indicating that loading actor failed or the method is unknown. Are we fine with ignoring them?

vyzo: that's a question for ayush, but it is kind if inevitable when native actors land.

raulk: Assuming that Wasm actors will exchange CBOR (at least for M2.2, I doubt we’ll want to introduce multicodec shenanigans then), you could make `GetReturnType` return `cbor.Value` when the method cannot be identified. Then the serde should work just fine and we don't need to uglily patch things here.

maciejwitowski: cc @arajasek

vyzo: There is a potentially better solution: introduce a non decoding variant of StateWaitMsg, so that we dont even try to deserialize in there and we use that onr.
Extant code can keep using the strict version.

raulk: I don't think Lotus needs more API entropy.

vyzo: the current API aint right however, we do want to wait for messages without caring about decoding the return value.

raulk: Yep, agree. Could you do a downstream usage analysis of the decoded field? I see it's not even being used at the CLI level.

vyzo: Sure, i can grep, but iys a public API so who really knows what users do...

vyzo: so nothing in cli touches ReturnDec; grepping the rest of lotus yields similar results.

vyzo: So to summarize:
- nothing in the entire lotus code base uses `ReturnDec`.
- we can just drop the field if we want and nothing would break.
- current patch (don't cry if it is missing) seems correct and future proof.

arajasek: - We definitely need to keep the existing functionality, there are users depending on it (eg. figuring out the `txn_id` of a `Propose`d multisig message).
- Do not want new method pls.
- I would propose adding an `ErrNotFound` that `GetReturnType` and `GetParamType` can return, and simply returning the raw bytes in that case. I'm not sure if that has a similar effect to @raulk's `cbor.Value` proposition (not sure what `cbor.Value` does).

vyzo: https://github.com/filecoin-project/lotus/pull/9635/commits/e019b206b521166ef053ab20845a10183732ed7b

arajasek (commented):
Oh no, this test is far too CLI-y and string-matchy for my preference. This is problematic for 2 reasons:

- Such tests are notoriously brittle and lead to future :crying_cat_face: 
- Such tests are less useful as "demonstrators" of how things work (what messages are sent, what's returned, etc.)

I think it  would be much nicer to actually read the contract bytes, formulate the messages as appropriate, and parse returns with assertions. This will definitely be a more bloated test (much more code), but overall better.

(If we want this as a separate CLI test, just to cover CLI functionality, I'm less opposed to that).

vyzo:
Basically you want me to copy the cli code ;)

I guess I could do that.

vyzo:
But yes, we do want to cover cli functionality too.
We have been using this as a smoke test for a long time.

Stebalien:
> But yes, we do want to cover cli functionality too.

That would go in `cli/chain_test.go`. We usually test the API and CLI separately.

arajasek:
> It is an itest however, and there plenty of cli tests in the itest bundle.
> 
> I propose we keep this here and add a second test that does everything by hand.

Yeah, I don't think there should be, and don't really want more. [Here's a good test](https://github.com/filecoin-project/lotus/pull/9549/files#diff-b2804ffb3f0ff97ed87f569da4890f5697b7bdf1db1f0b786d20ff2ccb9ca6bd) to use as a model -- it's basically laid out as:
- setup
- push message, assertion
- push message, assertion
- migration (not relevant to this PR), assertion
- push moar messages, moar assertions.

vyzo:
ok, I rewrote the thing as requested and removed the cli test; I will squash on merge and there will be no sign of it....

itests/fevm_test.go
```diff
@@ -0,0 +1,170 @@
...
+	builtintypes "github.com/filecoin-project/go-state-types/builtin"
+	"github.com/filecoin-project/go-state-types/builtin/v10/eam"
+	"github.com/filecoin-project/lotus/chain/actors"
+	"github.com/filecoin-project/lotus/chain/types"
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+// TestFEVMBasic does a basic fevm contract installation and invocation
+func TestFEVMBasic(t *testing.T) {
+	_ = os.Setenv("BELLMAN_NO_GPU", "1")
```
arajasek, line 28: Why do we need to do this?

itests/fevm_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	t.Log("sending create message")
+	smsg, err := client.MpoolPushMessage(ctx, msg, nil)
+	require.NoError(t, err)
+
+	t.Log("waiting for message to execute")
+	wait, err := client.StateWaitMsg(ctx, smsg.Cid(), 0, 0, false)
+	require.NoError(t, err)
+
+	require.Equal(t, int(wait.Receipt.ExitCode), 0, "contract installation failed")
```
arajasek, line 79:
```suggestion
	require.True(t, wait.Receipt.ExitCode.IsSuccess(), "contract installation failed")
```

itests/fevm_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+		t.Log("sending invoke message")
+		smsg, err := client.MpoolPushMessage(ctx, msg, nil)
+		require.NoError(t, err)
+
+		t.Log("waiting for message to execute")
+		wait, err := client.StateWaitMsg(ctx, smsg.Cid(), 0, 0, false)
+		require.NoError(t, err)
+
+		require.Equal(t, int(wait.Receipt.ExitCode), 0, "contract execution failed")
```
arajasek, line 120: same as above

itests/fevm_test.go
vyzo: I dont know, its from other tests to not try to use gpu for crypto. Maybe it is not necessary.

itests/fevm_test.go
vyzo: ok

itests/fevm_test.go
vyzo: ack.

itests/fevm_test.go
vyzo: done.

itests/fevm_test.go
vyzo: done.

itests/fevm_test.go
vyzo: I removed it, doesnt seem to be doing anything anyway. Let's get rid of all this cargo cult stuff :)

PR #9636, opened 2022-11-14
feat: events: version the MessageReceipt structure.

raulk:
This is implemented as an internal version discriminator. An alternative would've been to turn MessageReceipt into an interface with two backing structs for v0 and v1. But that approach would've broken all usage sites.

CBOR serde for MessageReceipt is no longer automatically generated. We use custom logic to deal with versioning in both directions. This logic uses the cborgen serde code as a basis.

## Checklist

- [ ] CI is green


chain/types/message_receipt_cbor.go
```diff
@@ -0,0 +1,220 @@
...
+
+	"github.com/filecoin-project/go-state-types/exitcode"
+	cbg "github.com/whyrusleeping/cbor-gen"
+	"golang.org/x/xerrors"
+)
+
+// This file contains custom CBOR serde logic to deal with the new versioned
+// MessageReceipt resulting from the introduction of actor events (FIP-0049).
+
+var lengthBufMessageReceipt = []byte{132}
```
vyzo, line 15: thats probably not correct, the header includes number of fields.

chain/types/message_receipt_cbor.go
```diff
@@ -0,0 +1,220 @@
...
+		return err
+	}
+
+	// t.ExitCode (exitcode.ExitCode) (int64)
+	if mr.ExitCode >= 0 {
+		if err := cw.WriteMajorTypeHeader(cbg.MajUnsignedInt, uint64(mr.ExitCode)); err != nil {
+			return err
+		}
+	} else {
+		if err := cw.WriteMajorTypeHeader(cbg.MajNegativeInt, uint64(-mr.ExitCode-1)); err != nil {
```
vyzo, line 37: woah, we need to be able to decode this in other languages too you know...

vyzo (commented):
custom cbor looks incorrect and there is the issue of codecs in other languages, rust at minimum.

vyzo:
Cant we do something similar to the state tree?

chain/types/message_receipt_cbor.go
raulk: This was generated by cborgen -- what is the concrete problem you see?

chain/types/message_receipt_cbor.go
raulk: Fixed.

vyzo (approved):
lools ok and we have test for the serializer, so green tick it is!

vyzo (commented):
You could avoid the pointer wrapping and make the version specific serde free standing functions, but up to you.

vyzo:
Note test failure.

raulk:
> You could avoid the pointer wrapping and make the version specific serde free standing functions, but up to you.

Not worth it. I expect that to be optimized away by newtype compiler optimizations anyway.

PR #9639, opened 2022-11-14
Ethereum actor events API fixes

iand:
Some fixes from original event api PR (#9623) that was merged earlier. 

 - Rename modules.EthEvent function to EthEventAPI
 - Change event emitter type to abi.ActorID
 - Ensure event matching is performed on f4 addresses by converting actor id to f4 at match time


iand:
Fixing a bug in tests, don't merge yet

iand:
Fixed the test bug, this is ready for review

iand:
@raulk PTAL

node/modules/actorevent.go
```diff
@@ -50,7 +53,25 @@ func EthEvent(cfg config.ActorEventConfig) func(helpers.MetricsCtx, fx.Lifecycle
 
 		if cfg.EnableRealTimeFilterAPI {
 			ee.EventFilterManager = &filter.EventFilterManager{
-				ChainStore:       cs,
+				ChainStore: cs,
+				AddressResolver: func(ctx context.Context, emitter abi.ActorID, ts *types.TipSet) (address.Address, bool) {
```
raulk, line 57: Let's keep a TODO somewhere that we expect to focus optimization efforts here.

node/modules/actorevent.go
```diff
@@ -50,7 +53,25 @@ func EthEvent(cfg config.ActorEventConfig) func(helpers.MetricsCtx, fx.Lifecycle
...
+					if err != nil {
+						return address.Undef, false
+					}
+					addr, err := sm.LookupRobustAddress(ctx, idAddr, ts)
+					if err != nil {
+						return address.Undef, false
+					}
+
+					// if robust address is not f4 then we won't match against it so bail early
+					if addr.Protocol() != address.Delegated {
```
raulk, line 69: Also need to check that the namespace of the delegated address is the EAM.

raulk (approved):
Is this simpler/cleaner than having a predicate? I'm happy either way, but would love to know the reasoning.

Issue #8491, opened 2022-04-14
lotus-miner fails to remove files from workers (Both Snap and Regular)


Reiers:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### Lotus Version

```text
Daemon:  1.15.2-rc1+mainnet+git.dcf6f6414+api1.5.0
Local: lotus version 1.15.2-rc1+mainnet+git.dcf6f6414
```

## Hi Team

First, I was tracking that `lotus-miner sectors abort-upgrade <sectornumb>` did not clear or remove Unsealed, Sealed,  Caches on the workers. 
It only removes the content in update and update-cache.

But after some more tests - I can see that regular sealing sectors gets left behind on the workers as well after,
- `lotus-miner sealing abort <sectornumb>`
[...]

Issue #9140, opened 2022-08-09
Stuck in FinalizeSector


long568:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon:  1.17.0+mainnet+git.2830429ad+api1.5.0
Local: lotus version 1.17.0+mainnet+git.49a964e77.dirty
Daemon:  1.17.0+mainnet+git.49a964e77.dirty+api1.5.0
Local: lotus-miner version 1.17.0+mainnet+git.49a964e77.dirty
```

### Describe the Bug

Stuck in FinalizeSector
I use BOOSTD for markets.
I ONLY modified the RUST code.
Config:
`[Subsystems]
  EnableMining = true
  EnableSealing = true
  EnableSectorStorage = true
  EnableMarkets = false
[...]

Issue #9033, opened 2022-07-14
Finalizing Sectors from Workers


piknikSteven2021:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
lotus-miner version 1.17.0-rc2+mainnet+git.6aaacf8ef

lotus-miner version 1.16.0+mainnet+git.01254ab32.dirty
```

### Describe the Bug

We've been having the issue of the workers not sending sectors to long-term storage despite storage being declared on the worker - this has happened on v1.15.2, v1.16.0 and v1.17-rc2.
DisableRemoteFinalize turned on to stop the miner from pulling sectors and congesting its NFS connection to the storage. (v1.17-rc2)
The miner is tasking the worker to finalize but there is nothing happening on the worker.
[...]

Issue #8783, opened 2022-06-01
Bug in EarlyFinalize occurs when the ProveCommit can't be pushed on chain


RobQuistNL:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
1.15.2-rc2
```

### Describe the Bug

WHen I run lotus I see the miner log filled with;
```
not selecting on <storagelocation2>, not in allowed group, allow map[]; path has []
not selecting on <storagelocation5>, not in allowed group, allow map[]; path has []
not selecting on <storagelocation8>, not in allowed group, allow map[]; path has []
not selecting on <storagelocation4>, not in allowed group, allow map[]; path has []
```
Etc.

I do not have any groups defined anywhere

### Logging Information

```text
-
```

[...]

PR #9648, opened 2022-11-15
fix: sealing: More complete snapdeals abort cleanup 

magik6k:
## Related Issues
Will fix https://github.com/filecoin-project/lotus/issues/8491

Might fix:
* https://github.com/filecoin-project/lotus/issues/9140
* https://github.com/filecoin-project/lotus/issues/9033
* https://github.com/filecoin-project/lotus/issues/8783 (maybe?)

## Proposed Changes
* Cleanup all things which need cleaning up in AbortUpgrade handler
* Fix FinalizeSector
  * Move all logic around managing unsealed data to ReleaseUnsealed / extend ReleaseUnseaed to be able to partially free the unsealed file
  * Simplify Finalize* functions
    * Should fix a lot of issues with finalize where it can't be scheduled, or is scheduled on the wrong machine

## Additional Info
This PR changes some worker APIs, so it bumps the worker API version - **IT REQUIRES UPDATING BOTH `lotus-miner` AND `lotus-worker`**

## Checklist

- [x] CI is green


storage/pipeline/states_failed.go
```diff
@@ -425,16 +426,19 @@ func (m *Sealing) handleAbortUpgrade(ctx statemachine.Context, sector SectorInfo
...
-		return xerrors.Errorf("getting sealing config: %w", err)
+	// This removes the unsealed file from all storage
+	// note: we're not keeping anything unsealed because we're reverting to CC
+	if err := m.sealer.ReleaseUnsealed(ctx.Context(), m.minerSector(sector.SectorType, sector.SectorNumber), []storiface.Range{}); err != nil {
+		log.Error(err)
 	}
 
-	if err := m.sealer.ReleaseUnsealed(ctx.Context(), m.minerSector(sector.SectorType, sector.SectorNumber), sector.keepUnsealedRanges(sector.CCPieces, true, cfg.AlwaysKeepUnsealedCopy)); err != nil {
+	// and makes sure sealed/cache files only exist in long-term-storage
+	if err := m.sealer.FinalizeSector(ctx.Context(), m.minerSector(sector.SectorType, sector.SectorNumber)); err != nil {
```
magik6k, line 441: This is the main fix - we weren't cleaning up sealed/cache files at all, and we were passing ranges to keep to ReleaseUnsealed.

itests/kit/node_miner.go
```diff
@@ -224,3 +224,36 @@ func (tm *TestMiner) SectorsListNonGenesis(ctx context.Context) ([]abi.SectorNum
 
 	return l[tm.PresealSectors:], nil
 }
+
+type SchedInfo struct {
+	CallToWork   struct{}
+	EarlyRet     interface{}
+	ReturnedWork interface{}
+	SchedInfo    struct {
```
shrenujbansal, line 232: Is this intentional? Might wanna have a different name for this struct object than the higher level type?

lib/consensus/raft/config.go
```diff
@@ -97,8 +97,8 @@ func NewClusterRaftConfig(userRaftConfig *config.UserRaftConfig) *ClusterRaftCon
 
 }
 
-//// Validate checks that this configuration has working values,
-//// at least in appearance.
+// // Validate checks that this configuration has working values,
```
shrenujbansal, line 100: This is from my PR but could you remove the 2nd set of // if changing this

storage/sealer/manager.go
```diff
@@ -608,52 +608,66 @@ func (m *Manager) SealCommit2(ctx context.Context, sector storiface.SectorRef, p
...
+	if err != nil {
+		return false, "", xerrors.Errorf("finding sector: %w", err)
+	}
+	if len(stores) == 0 {
+		return false, "", nil
+	}
+
+	for _, store := range stores {
+		if store.CanSeal {
+			return true, storiface.PathSealing, nil
```
shrenujbansal, line 624: so the type is PathSealing if even a single store has CanSeal set to be true? Could you explain why that is?

storage/sealer/manager.go
```diff
@@ -755,9 +761,15 @@ func (m *Manager) FinalizeReplicaUpdate(ctx context.Context, sector storiface.Se
 
 	err = multierr.Append(move(storiface.FTUpdate|storiface.FTUpdateCache), move(storiface.FTCache))
 	err = multierr.Append(err, move(storiface.FTSealed)) // Sealed separate from cache just in case ReleaseSectorKey was already called
-	// if we found unsealed files, AND have been asked to keep at least one, move unsealed
-	if moveUnsealed != storiface.FTNone && len(keepUnsealed) != 0 {
-		err = multierr.Append(err, move(moveUnsealed))
+
+	{
```
shrenujbansal, line 765: why is there a code block here?

storage/sealer/sealtasks/task.go
```diff
@@ -67,7 +69,8 @@ var shortNames = map[TaskType]string{
 	TTCommit1:    "C1",
 	TTCommit2:    "C2",
 
-	TTFinalize: "FIN",
+	TTFinalize:        "FIN",
+	TTReleaseUnsealed: "FUS",
```
shrenujbansal, line 73: RUS?

itests/kit/node_miner.go
magik6k: This is a horrible hack extracted from a different test; Technically the test can depend on the real struct, but that's a tiny bit tricky now

lib/consensus/raft/config.go
magik6k: Yeah, some linter started having opinions about this recently, we will probably be seeing a bunch of diffs like this for a while.

storage/sealer/sealtasks/task.go
magik6k: I went with FUS (to mean FinalizeUnsealed), to be consistent with other finalization-related short names; Tho I guess the task should be called FinalizeUnsealed in that case..

storage/sealer/manager.go
magik6k: This is at the sealing stage where we only expect a single copy of sector data in paths, so if we see more than one sector, something is quite broken. Inverting this shouldn't change anything.

storage/sealer/manager.go
magik6k: Just enclosing sealedStores in a tighter scope so it's easier to tell that it's not used anywhere below; basically just a code style thing

itests/kit/node_miner.go
magik6k: Oh it's not easy to reuse that struct..

See https://github.com/filecoin-project/lotus/blob/master/storage/sealer/manager.go#L1225-L1239

This code was written in the days where we had big plans to make scheduling / sealing code easily swappable, but that has resulted in this vague loose diagnostic interface which tries hard to not commit to a schema. Probably should be fixed one day, but maybe out of scope for this PR.

storage/sealer/sealtasks/task.go
magik6k: Renamed but kept some other ReleaseUnsealed in less visible places to avoid breaking the API too much; This code will realistically all get replaced when the new scheduler comes, so just a non-confusing UX is probably good enough.

itests/worker_upgrade_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+		for id, decls := range paths {
+			pinfo, err := miner.StorageInfo(ctx, id)
+			require.NoError(t, err)
+
+			switch {
+			case id == wpaths[0].ID: // worker path
+				if workers != storiface.FTNone {
+					require.Len(t, decls, 1)
+					require.EqualValues(t, decls[0].SectorFileType.Strings(), workers.Strings())
```
shrenujbansal, line 95: Expected and actual args are reversed here

storage/sealer/manager.go
shrenujbansal: Why is it important that those are not used below?

itests/kit/node_miner.go
shrenujbansal: Can you add a comment referencing the place in code this is being replicated from? In the event there are changes made to the struct in manager.go, things will need to change here too?
Also probably a TODO to use a single definition

storage/sealer/sealtasks/task.go
shrenujbansal: I think ideally we should keep the names of the API and tasks consistent considering thats the pattern followed for others?

storage/sealer/manager.go
magik6k: It's not important outside making it easier to see where variables are used

storage/sealer/sealtasks/task.go
magik6k: I didn't want to change the already existing method on the miner api - ReturnReleaseUnsealed was already there; This has the subtle side effect of making lotus-miner technically forward-compatible with workers (e.g. if you downgrade lotus-miner when you have newer versions of workers still running, and trying to return ReleaseUnsealed those results will get recorded correctly)

tbh it doesn't matter that much, just felt like the right approach.

itests/worker_upgrade_test.go
magik6k: Now that's a nitpick :p

shrenujbansal (approved):

RobQuistNL:
Good stuff! Will let you know if we run into it again with this fix.

PR #9650, opened 2022-11-15
chore: vm: more logs to VerifyConsensusFault

arajasek:
## Related Issues

Ellmatic from the Chainsafe team requested more info about what happened in a VCF message. Lotus as is makes it hard to answer this question.

## Proposed Changes

Log errors in this method so they're visible in lotus.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/vm/syscalls.go
+++ b/chain/vm/syscalls.go
@@ -97,50 +97,69 @@ func (ss *syscallShim) VerifyConsensusFault(a, b, extra []byte) (*runtime7.Conse
 	// for that reason when checking block parent relationships, rather than instantiating a Tipset to do so
 	// (which runs a syntactic check), we do it directly on the CIDs.
 
+	var err error
 	// (0) cheap preliminary checks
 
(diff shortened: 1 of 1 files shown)


geoff-vball:
This `VerifyConsensusFault()` method is only called in one non-spec-actors place - `chain/vm/gas.go`. Can we just grab the error in that function and log it instead of adding all this extra code?

arajasek:
@geoff-vball Not quite -- this method is invoked by the FVM during message execution through the FFI: https://github.com/filecoin-project/filecoin-ffi/blob/master/cgo/interface.go#L33.

geoff-vball:
Oh that's interesting. I didn't know we did that. It's probably still better to wrap this function then, no?

geoff-vball:
@arajasek Isn't the function that's called by the FVM this one? https://github.com/filecoin-project/lotus/blob/04f0af586521f024c70a71ce06ef5ad6cd6391b5/chain/vm/fvm.go#L104

PR #9664, opened 2022-11-16
NV18: FEVM: setup FEVM during genesis: create ETH0 actor

vyzo:
we need this for wallaby and devnets, will be created by migration normally.

### DIFF
--- a/chain/gen/genesis/fevm.go
+++ b/chain/gen/genesis/fevm.go
@@ -2,11 +2,11 @@ package genesis
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
 
 	"github.com/ipfs/go-cid"
 
-	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	actorstypes "github.com/filecoin-project/go-state-types/actors"
 	"github.com/filecoin-project/go-state-types/big"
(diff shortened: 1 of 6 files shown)


vyzo:
CI is decisively not happy; fixing.

arajasek (approved):
LGTM barring sad CI -- feel free to land after fixing that, or get an FVM-y person to also :eye:

vyzo:
main issue was from,to inversion in the implicit message, fixed that, but we now have an annoying issue in the fvm kernel rejecting our message because origin doesn't have a predictable address... SIGH!!!!!

chain/gen/genesis/fevm.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/filecoin-project/go-state-types/network"
+
+	"github.com/filecoin-project/lotus/chain/actors"
+	"github.com/filecoin-project/lotus/chain/consensus/filcns"
+	"github.com/filecoin-project/lotus/chain/state"
+	"github.com/filecoin-project/lotus/chain/store"
+	"github.com/filecoin-project/lotus/chain/vm"
+)
+
+func SetupFEVM(ctx context.Context, cs *store.ChainStore, sys vm.SyscallBuilder, sroot cid.Cid, nv network.Version) (cid.Cid, error) {
```
Stebalien, line 25: nit: I'd expect this to also handle setting up the _EAM_, but we can rework that later.

chain/gen/genesis/fevm.go
```diff
@@ -0,0 +1,114 @@
...
+	// return
+	// code_begin:
+	// push1 0x00
+	// push1 0x00
+	// return
+	// code_end:
+	initcode, err := hex.DecodeString("600580600b6000396000f360006000f3")
+	if err != nil {
+		return cid.Undef, fmt.Errorf("failed to parse ETH0 init code during genesis: %w", err)
+	}
```
Stebalien, line 83: for future readers, this will go away very soon.

chain/gen/genesis/genesis.go
```diff
@@ -694,6 +701,7 @@ func SetupEAM(_ context.Context, nst *state.StateTree, nv network.Version) error
 		Code:    codecid,
 		Head:    vm.EmptyObjectCid,
 		Balance: big.Zero(),
+		Address: &builtintypes.EthereumAddressManagerActorAddr, // so that it can create ETH0
```
Stebalien, line 704: This is very wrong and could break _many_ things.

Stebalien: We assume that these addresses are always f1/f3/f4, never f0.

Stebalien: I'm going to see what I can do to fix this.

Stebalien:
@vyzo yes, I know. I retconned everything and did it the boring way. Unfortunately, this was the easiest way to avoid setting the predictable address.

Stebalien:
Ok, I undid my changes because genesis has some issues. We're going with this approach, because whatever.

Stebalien (approved):
Ok, a day of messing with this, LGTM for now.

PR #9687, opened 2022-11-18
feat: fvm: add support for looking up past tipset CIDs

Stebalien:
We do this by adding yet another "getter" to the VM that resolves an epoch into a TipSetKey.

Alternatively, we could stick this functionality on `Rand`, but then it wouldn't really be "rand" but something else...

Depends on https://github.com/filecoin-project/filecoin-ffi/pull/339

Merge plan:

1. Merge https://github.com/filecoin-project/filecoin-ffi/pull/339.
2. Rebase this PR on _master_.
3. Merge this PR to master.
4. Port this change to the nv18 branches.

Unfortunately, it needs to land on master because it bubbles a breaking change in the FFI (where the FVM now needs the ability to lookup tipset CIDs).

Alternatively, we could make this _optional_ with a runtime type assertion. That is, if we find that we're using "v3" inside the FFI, we could type-assert that the "externs" supports the `TipsetCid` method. That would let us merge the FFI changes _without_ making any changes to the interface.

### DIFF
--- a/chain/types/tipset_key.go
+++ b/chain/types/tipset_key.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"strings"
 
+	block "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	typegen "github.com/whyrusleeping/cbor-gen"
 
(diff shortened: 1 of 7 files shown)


arajasek (approved):
Yeah, this is how I would do it too.

Stebalien:
@arajasek I've rebased on master, merge when ready.

Note: this updates the FFI. _Technically_, we can get away with not doing that on master if you're so inclined.

Stebalien:
Actually, I'm just going to revert that part.

chain/types/tipset_key.go
```diff
@@ -98,6 +99,28 @@ func (k *TipSetKey) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+func (k TipSetKey) Cid() (cid.Cid, error) {
```
Stebalien, line 102: @arajasek cherry-picked from feat/nv18-fevm. Please review.

Issue #9159, opened 2022-08-11
Allow lotus-worker to unseal sector even if does not store unsealed data long-term


geoff-vball:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated improvement to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [X] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Improvement Suggestion

Currently if a lotus-miner sends an unseal job to a lotus-worker, the job will fail unless the storage of the worker has the `store` flag for unsealed data set. This is because the worker tries to store the unsealed data long-term before sending it back to the miner, but this fails and the miner never gets the unsealed data. 

There are reasons the worker might not want to store unsealed data long-term, so we should allow the worker to store the data temporarily. The worker should be able to unseal the data, store it temporarily, send it back to the miner, then erase the unsealed data. 

In addition to this, when the worker returns the unsealed data to the miner, the miner should be able to choose whether to keep a copy of the unsealed data, or erase it after it is served to the client. We can do this with an env var.

Relevant slack discussion: https://filecoinproject.slack.com/archives/CP50PPW2X/p1660224454108749

stuberman:
`the job will fail unless the storage of the worker has the store flag for unsealed data set`

I did try to set the sealing worker to 'store' in addition to 'seal' and then the long term sectors were being stored on the sealing worker and not in the lotus-miner long term storage repo.

donkabat:
[...]

PR #9694, opened 2022-11-21
fix: sealing: Set all path types for Unseal pipeline to sealing storage

shrenujbansal:
## Related Issues

https://github.com/filecoin-project/lotus/issues/9159

## Proposed Changes

Currently when unsealing a sector, its expected that the unsealing worker has access to long term storage and stores the unsealed data there. We need to be able to unseal data and deliver it to the miner without having access to long term storage for retrievals.
Instead of using long term storage, use sealing storage to in the unsealing pipeline

## Checklist

- [x] CI is green

### DIFF
--- a/itests/sector_unseal_test.go
+++ b/itests/sector_unseal_test.go
@@ -1,0 +1,87 @@
+package itests
+
+import (
+	"context"
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/lotus/itests/kit"
+	"github.com/filecoin-project/lotus/node/config"
+	"github.com/filecoin-project/lotus/storage/sealer/sealtasks"
+	"github.com/filecoin-project/lotus/storage/sealer/storiface"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func TestUnsealPiece(t *testing.T) {
+	ctx := context.Background()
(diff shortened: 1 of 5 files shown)


itests/sector_unseal_test.go
```diff
@@ -0,0 +1,90 @@
...
+	require.NoError(t, err)
+
+	sectorRef := storiface.SectorRef{
+		ID:        abi.SectorID{Miner: abi.ActorID(minerId), Number: sector},
+		ProofType: sinfo.SealProof,
+	}
+
+	err = miner.SectorsUnsealPiece(ctx, sectorRef, 0, 0, sinfo.Ticket.Value, sinfo.CommD)
+	require.NoError(t, err)
+}
```
magik6k, line 90: Can we add some asserts that:
* The Unsealed file is in long-term storage
* Sealed/Cache still are in long-term storage
* There is nothing left in sealing storage

magik6k (commented):
I'd add some asserts to the test to make sure this does what we expect

itests/sector_unseal_test.go
shrenujbansal: done

hyunmoon:
I can confirm that this issue is not fixed in v1.19.0+mainnet

hyunmoon:
But I'm glad to know that I can work around the issue by letting the workers have access to the long term storage.

PR #9712, opened 2022-11-23
_build_: Update to Go Version 1.18.8

ianconsolata:
## Checklist

- [ ] CI is green


geoff-vball:
Could you please update the README to reflect this change?

There are two more places that use go `1.18.1` that we might want to change. 
- `.github/workflows/codeql-analysis.yml`
- The Dockerfiles inside of `testplans/docker-images`

I don't have any context on these files and whether we should be updating them. @arajasek @magik6k can you lend a hand here?

ianconsolata:
The linter job is also failing: https://circleci.com/gh/filecoin-project/lotus/688918 -- not sure why that would be the case for a minor version upgrade like this?

jennijuju (approved):

cli/util/api.go
```diff
@@ -274,7 +274,7 @@ func FullNodeProxy[T api.FullNode](ins []T, outstr *api.FullNodeStruct) {
 				}
 
 				total := len(rins)
-				result, err := retry.Retry(ctx, 5, initialBackoff, errorsToRetry, func() (results []reflect.Value, err2 error) {
+				result, _ := retry.Retry(ctx, 5, initialBackoff, errorsToRetry, func() (results []reflect.Value, err2 error) {
```
geoff-vball, line 277: I feel like we should bubble up an error here if we get one? @arajasek thoughts?

cmd/lotus-shed/migrations.go
```diff
@@ -240,7 +240,7 @@ func checkMigrationInvariants(ctx context.Context, v8StateRootCid cid.Cid, v9Sta
 		return err
 	}
 
-	v9actorTree, err := builtin.LoadTree(actorStore, v9stateRoot.Actors)
+	v9actorTree, _ := builtin.LoadTree(actorStore, v9stateRoot.Actors)
```
geoff-vball, line 243: We should definitely check the error here.

lib/consensus/raft/consensus.go
```diff
@@ -151,7 +151,7 @@ func NewConsensus(host host.Host, cfg *ClusterRaftConfig, mpool *messagepool.Mes
 	consensus.SetActor(actor)
 
 	peers := []peer.ID{}
-	addrInfos, err := addrutil.ParseAddresses(ctx, cfg.InitPeerset)
+	addrInfos, _ := addrutil.ParseAddresses(ctx, cfg.InitPeerset)
```
geoff-vball, line 154: We should check the error here too

geoff-vball:
Thanks for working on these!

cmd/lotus-shed/migrations.go
ianconsolata: And do what? Panic?

lib/consensus/raft/consensus.go
ianconsolata: Same here — what should I do with it?

cmd/lotus-shed/migrations.go
geoff-vball: Just return it

`	if err != nil {
		return err
	}`

lib/consensus/raft/consensus.go
geoff-vball: `	if err != nil {
		return nil, err
	}`

geoff-vball:
Other than the few that I commented on, we should probably leave the rest of the linter error for @arajasek or @magik6k to take a look at.

ianconsolata:
Agreed, the others seemed like they had more complex solutions, and I didn't want to change any http servers without checking in with folks who know the codebase better first. Any thoughts on why these linter checks only got caught after the update to 1.18.8? They seem like they should have been caught before this PR as well.

cli/util/api.go
geoff-vball: I investigated this a bit and think it might be the cause of a flaky test. Followup here https://github.com/filecoin-project/lotus/issues/9735

geoff-vball:
@ianconsolata Looks like the golang update woke up the linter? :confused:

cli/util/api.go
```diff
@@ -274,7 +274,7 @@ func FullNodeProxy[T api.FullNode](ins []T, outstr *api.FullNodeStruct) {
 				}
 
 				total := len(rins)
-				result, err := retry.Retry(ctx, 5, initialBackoff, errorsToRetry, func() (results []reflect.Value, err2 error) {
+				result, _ := retry.Retry(ctx, 5, initialBackoff, errorsToRetry, func() ([]reflect.Value, error) {
```
geoff-vball, line 277: This is just a temporary ignore for the linter... This gas should probably be propagated https://github.com/filecoin-project/lotus/issues/9735

itests/raft_messagesigner_test.go
```diff
@@ -79,7 +79,6 @@ func setup(ctx context.Context, t *testing.T, node0 *kit.TestFullNode, node1 *ki
 		node.Override(node.GoRPCServer, modules.NewRPCServer),
 	)
 	//raftOps := kit.ConstructorOpts()
-	kit.ThroughRPC()
```
geoff-vball, line 82: This is a pure function. It wasn't doing anything here.

geoff-vball:
Linter was complaining about http server timeouts because of slow loris attacks. I'm not sure exactly what values we might want here, but I just put in a default of 3 so you can see where we should put them.

node/rpc.go
```diff
@@ -51,7 +52,8 @@ func ServeRPC(h http.Handler, id string, addr multiaddr.Multiaddr) (StopFunc, er
 
 	// Instantiate the server and start listening.
 	srv := &http.Server{
-		Handler: h,
+		Handler:           h,
+		ReadHeaderTimeout: 3 * time.Second,
```
magik6k, line 56: How did we get this number? What was the previous default?

I imagine under high load setting this to just 3 seconds may cause issues

geoff-vball: I commented above, I just chose this number at random. I'm happy for thoughts on what number to use here and for the other timeouts. Before we didn't have any timeout at all, which is probably bad and why the linter was mad.

magik6k: Let's set this to 30 seconds for now, 3 seconds is almost definitely too low

ianconsolata:
@geoff-vball I think now that all the linter errors addressed, we can merge this?

magik6k (approved):

PR #9599, opened 2022-11-07
NV18 Staging Branch

vyzo:
This is the actual branch to deliver M2.1 as NV18 to the network.
It is intended to integrate all M2.1 changes for merge to master as soon as possible.

Workplan:
- [X] actors version 10 scaffolding
- [x] m2 bundle for version 10
- [x] state tree version 5
- [ ] network version 18 migration and upgrade scaffolding -- can happen in follow up pr.
- [ ] GREEN CI
- [x] Merge AFTER `release/v1.19.0` is cut (Tuesday the 15th).

PR #9716, opened 2022-11-23
NV18:  update go-state-types

vyzo:
Ok, so we changed the aliases to copies and everything should be fine, right? Nope.

This so far fixes the chain compilation errors, but all hell breaks loose with hardcoded references to v9 scattered around the codebase.

Notes:
Do make gen first (or chain/actors/agen) and see for yourself.
@arajasek I am ready to give up here, all this because we don't have nice aliases any longer.


vyzo:
Closing and reopening to get github to see the gst update as well.

vyzo:
Ok, the go-state-types update commit is here.

arajasek:
Done in #9599

PR #9746, opened 2022-11-29
feat: mpool/wdpost: Maximize feecap config

magik6k:
## Related Issues
I imagine there is something open about messages getting stuck around upgrades, but can't find an issue.

## Proposed Changes
We've seen messages get stuck in mpool when basefee is rising rapidly, caused by underestimating FeeCap.

This PR adds a new field to `MessageSendSpec`:
```go
// MaximizeFeeCap makes message FeeCap be based entirely on MaxFee
MaximizeFeeCap bool
```

Because of the basefee mechanism there isn't munch downside to setting FeeCap as high as we're willing to pay for the message

Currently that field can be configured only by windowPoSt logic through SP node configuration:
```toml
  # type: bool
  # env var: LOTUS_FEES_MAXIMIZEWINDOWPOSTFEECAP
  MaximizeWindowPoStFeeCap = false
```

TODO:
* [x] Decide if we want to default this to true?
  * At api level? Only for PoSt?
* [ ] <s>Add configuration for other subsystems</s> can be done later
* [x] Test in a network

## Checklist

- [x] CI is green


rjan90:
> Decide if we want to default this to true?

I would slightly lean to setting it to true by default. Reasoning being:
- Users already seem to think that they will always be able to publish their windowPoSt message as long as their `MaxWindowPoStGasFee`is set high enough.
- The cost of fees paid for `DeclareFaultsRecovered` and non-optimistic `SubmitWindowedPoStmessages` is probably higher then if there is a short lived basefee spike.

Reiers:
Message sent on mainnet, confirming that the message fee-cap is more than 10 times the current network base fee.
https://filfox.info/en/message/bafy2bzacebs77wmcwzdebplxr4ixqrmifw2o6gxnghsayvuawt2s5wdxdwzqc?t=4

Before:
https://filfox.info/en/message/bafy2bzacectvuk4s72hdknppwh3e5aa4smcfrbancfcdb6r5aclqtc3lpabjs?t=4

arajasek (approved):
LGTM, though @magik6k should :eyes: the commit i pushed and revert it if it's wrong for some reason.

chain/messagepool/messagepool.go
```diff
@@ -222,15 +224,12 @@ func CapGasFee(mff dtypes.DefaultMaxFeeFunc, msg *types.Message, sendSpec *api.M
...
 
-	gl := types.NewInt(uint64(msg.GasLimit))
-	totalFee := types.BigMul(msg.GasFeeCap, gl)
-
-	if totalFee.LessThanEqual(maxFee) {
-		msg.GasPremium = big.Min(msg.GasFeeCap, msg.GasPremium) // cap premium at FeeCap
-		return
+	gaslimit := types.NewInt(uint64(msg.GasLimit))
+	totalFee := types.BigMul(msg.GasFeeCap, gaslimit)
+	if !maximizeFeeCap && totalFee.GreaterThan(maxFee) {
```
arajasek, line 229:
```suggestion
	if maximizeFeeCap || totalFee.GreaterThan(maxFee) {
```

beck-8:
Sorry, I am just posting a situation here, not saying this is a problem, I just want everyone to understand.
After many miners upgraded, I saw that they could not submit wdpost messages, all because of this parameter. They were small miners, so they kept the balance of the control wallet only 0.5-5FiL, so they had problems, and they were not clear about this part of the code change.
So I think it is possible to post this part of the changes clearly in the version changelog so that everyone can understand it?

Issue #9735, opened 2022-11-25
TestGoRPCAuth is flaky


geoff-vball:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### Lotus Version

```text
Master
```

### Describe the Bug

TestGoRPCAuth fails with the following error: 
`        	Error Trace:	/home/circleci/project/itests/ensemble.go:673
        	            				/home/circleci/project/itests/raft_messagesigner_test.go:554
        	Error:      	Received unexpected error:
        	            	failed to sign message: key not found, addr=f3vrsoe2mloxxj22x277fcizcctqc5i6ay6z37mz6ewy5tisf5rxohfulsy7gtur54umumsd3bzagaay3pklta: key not found`

### Logging Information

```text
.
```

### Repo Steps

1. Run '...'
[...]

PR #9763, opened 2022-12-01
fix: test: Deflake raft test

geoff-vball:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/9735

## Proposed Changes
This test has 3 nodes in a cluster, and one on its own. The one on its own was accidentally being passed `initPeerSet` which would cause it to connect to the cluster, and occasionally be the leader, which caused the test to flake.

I also added some small changes to make my linter happy while I was here.

## Additional Info
Ran this test x100 on my machine with no failures. It was failing 25% of the time before.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/consensus/filcns/filecoin.go
+++ b/chain/consensus/filcns/filecoin.go
@@ -122,7 +122,7 @@ func (filec *FilecoinEC) ValidateBlock(ctx context.Context, b *types.FullBlock) 
 		return xerrors.Errorf("block was from the future (now=%d, blk=%d): %w", now, h.Timestamp, consensus.ErrTemporal)
 	}
 	if h.Timestamp > now {
-		log.Warn("Got block from the future, but within threshold ", h.Timestamp, build.Clock.Now().Unix())
+		log.Warn("Got block from the future, but within threshold", h.Timestamp, build.Clock.Now().Unix())
 	}
 
 	msgsCheck := async.Err(func() error {
--- a/chain/consensus/filcns/filecoin.go
+++ b/chain/consensus/filcns/filecoin.go
@@ -122,7 +122,7 @@ func (filec *FilecoinEC) ValidateBlock(ctx context.Context, b *types.FullBlock) 
 		return xerrors.Errorf("block was from the future (now=%d, blk=%d): %w", now, h.Timestamp, consensus.ErrTemporal)
 	}
 	if h.Timestamp > now {
-		log.Warn("Got block from the future, but within threshold", h.Timestamp, build.Clock.Now().Unix())
+		log.Warn("Got block from the future, but within threshold ", h.Timestamp, build.Clock.Now().Unix())
 	}
 
 	msgsCheck := async.Err(func() error {
--- a/itests/kit/ensemble.go
+++ b/itests/kit/ensemble.go
@@ -194,7 +194,7 @@ func (n *Ensemble) FullNode(full *TestFullNode, opts ...NodeOpt) *Ensemble {
 		require.NoError(n.t, err)
 	}
 
-	key, err := key.GenerateKey(types.KTBLS)
+	k, err := key.GenerateKey(types.KTBLS)
 	require.NoError(n.t, err)
 
 	if !n.bootstrapped && !options.balance.IsZero() {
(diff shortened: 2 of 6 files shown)


itests/kit/ensemble.go
```diff
@@ -194,7 +194,7 @@ func (n *Ensemble) FullNode(full *TestFullNode, opts ...NodeOpt) *Ensemble {
 		require.NoError(n.t, err)
 	}
 
-	key, err := key.GenerateKey(types.KTBLS)
+	k, err := key.GenerateKey(types.KTBLS)
```
shrenujbansal, line 197: Any reason for the var name change?

itests/kit/ensemble.go
```diff
@@ -679,9 +679,9 @@ func (n *Ensemble) Start() *Ensemble {
 
 		var mineBlock = make(chan lotusminer.MineReq)
 
-		copy := *m.FullNode
```
shrenujbansal, line 682: Same here. Why the var name change?

itests/raft_messagesigner_test.go
```diff
@@ -536,7 +542,7 @@ func TestGoRPCAuth(t *testing.T) {
 	wrappedFullNode := kit.MergeFullNodes(nodes)
 
 	ens.MinerEnroll(&miner, wrappedFullNode, kit.WithAllSubsystems(), kit.ThroughRPC())
```
shrenujbansal, line 544: If the below changes work well, could you also change this in the setup function?

itests/kit/ensemble.go
geoff-vball: linter doesn't like shadowed names. `key` is the name of a package, so we should use something else for the variable.

itests/kit/ensemble.go
geoff-vball: This one shadows a golang builtin

itests/raft_messagesigner_test.go
```diff
@@ -487,12 +479,18 @@ func TestChainStoreSync(t *testing.T) {
 	}
 }
 
+func TestGoRPCAuthMultiple(t *testing.T) {
```
shrenujbansal, line 482: Remove this. We can use count for running this several times

geoff-vball: oopsie

itests/raft_messagesigner_test.go
geoff-vball: Done!

itests/raft_messagesigner_test.go
```diff
@@ -507,27 +499,33 @@ func TestGoRPCAuth(t *testing.T) {
...
+			cfg := consensus.DefaultClusterRaftConfig()
+			cfg.InitPeerset = initPeerSet
+			return cfg
+		}))
+
+	soloClusterOps := kit.ConstructorOpts(
+		node.Override(new(*consensus.ClusterRaftConfig), consensus.DefaultClusterRaftConfig()),
+	)
+
+	ens := kit.NewEnsemble(t).FullNode(&node0, raftOps, connectedClusterOps, kit.ThroughRPC()).FullNode(&node1, raftOps, connectedClusterOps, kit.ThroughRPC()).FullNode(&node2, raftOps, connectedClusterOps, kit.ThroughRPC()).FullNode(&node3, raftOps, soloClusterOps)
```
magik6k, line 527: That’s a long line

chain/consensus/filcns/filecoin.go
```diff
@@ -122,7 +122,7 @@ func (filec *FilecoinEC) ValidateBlock(ctx context.Context, b *types.FullBlock)
 		return xerrors.Errorf("block was from the future (now=%d, blk=%d): %w", now, h.Timestamp, consensus.ErrTemporal)
 	}
 	if h.Timestamp > now {
-		log.Warn("Got block from the future, but within threshold", h.Timestamp, build.Clock.Now().Unix())
+		log.Warn("Got block from the future, but within threshold ", h.Timestamp, build.Clock.Now().Unix())
```
magik6k, line 125: Stray space

itests/raft_messagesigner_test.go
```diff
@@ -507,27 +499,33 @@ func TestGoRPCAuth(t *testing.T) {
 	pkey2, _ := generatePrivKey()
 
 	pkeys := []*kit.Libp2p{pkey0, pkey1, pkey2}
```
magik6k, line 501: How was the third node able to become a leader when it wasn't part of the init peerset?

magik6k: If a node isn't part of a cluster it shouldn't be able to connect to it, and definitely shouldn't be able to become a leader - arguably this test was failing correctly, showing that it's possible for external nodes to "break" into a cluster

magik6k: (note that raft/gorpc are on the public libp2p interface, and being in a cluster is the only thing preventing calls to MpoolPushMessage by libp2p peers; If that check can be defeated that would be extremely bad)

geoff-vball: Okay, let's parking lot this for standup

Issue #9769, opened 2022-12-02
1.18.0, the miner process crashes during execution of the AddPiece phase


Shekelme:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [X] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
lotus --version
lotus version 1.18.0+mainnet+git.bd10bdf99
lotus-miner --version
lotus-miner version 1.18.0+mainnet+git.bd10bdf99
```

### Describe the Bug

From time to time, the miner process crashes during (or when trying to) execute the AddPiece phase. In the time that has passed since the network upgrade, this has already happened twice.

### Logging Information

```text
Last lines of miner.log:

[...]

PR #9822, opened 2022-12-09
fix: sealing: Avoid nil dereference in debug log

magik6k:
## Related Issues
Fixes https://github.com/filecoin-project/lotus/issues/9769

## Proposed Changes
Don't panic, also improve assigner debug logs

## Additional Info
Before the Shark upgrade `checkDealAssignable` would only ever return `false` if the sector was an upgrade sector.

When that happened we'd try to get sector expiration for the `CC update sector %d cannot fit deal...` debug log, which would panic when we try that on non-upgrade sectors - which can now happen due to the new TermMax handling logic in `checkDealAssignable` when the sector has some deals assigned, and we have an incoming deal which can't fit into the sector.

## Checklist

- [x] CI is green

### DIFF
--- a/storage/pipeline/input.go
+++ b/storage/pipeline/input.go
@@ -449,6 +449,9 @@ func (m *Sealing) updateInput(ctx context.Context, sp abi.RegisteredSealProof) e
 		if err != nil {
 			return 0, big.Zero(), err
 		}
+		if onChainInfo == nil {
+			return 0, big.Zero(), xerrors.Errorf("sector info for sector %d not found", sn)
+		}
 		memo[sn] = struct {
 			e abi.ChainEpoch
 			p abi.TokenAmount
--- a/storage/pipeline/input.go
+++ b/storage/pipeline/input.go
@@ -494,10 +494,6 @@ func (m *Sealing) updateInput(ctx context.Context, sp abi.RegisteredSealProof) e
 				continue
 			}
 			if !ok {
-				exp, _, _ := getExpirationCached(sector.number)
-
-				// todo move this log into checkDealAssignable, make more detailed about the reason
-				log.Debugf("CC update sector %d cannot fit deal, expiration %d before deal end epoch %d", id, exp, piece.deal.DealProposal.EndEpoch)
 				continue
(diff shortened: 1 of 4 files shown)


RobQuistNL:
I'm trying to manually apply this patch on our production setups as we speak

RobQuistNL:
Tested on prod machines & fixes issue when applied to 1.18.0

jennijuju:
thanks @RobQuistNL ! Magik has done. the same and we will try to get this reviewed and do a patch asap!

storage/pipeline/input.go
```diff
@@ -449,6 +449,9 @@ func (m *Sealing) updateInput(ctx context.Context, sp abi.RegisteredSealProof) e
 		if err != nil {
 			return 0, big.Zero(), err
 		}
+		if onChainInfo == nil {
+			return 0, big.Zero(), xerrors.Errorf("sector info for sector %d not found", sn)
+		}
```
geoff-vball, line 454: Instead of fixing this here, I think I would rather fix `GetSector()` in `chain/actors/builtin/miner/vX.go` (these are generated files). I think we should return an error from `GetSector()` if we fail to find it, instead of the `nil, nil` that we're currently returning. It's not idiomatic and is what caused this issue in the first place.

geoff-vball (changes requested):

storage/pipeline/input.go
magik6k: This is documented behaviour, we want some way there to indicate “sector not found” that works over jsonrpc - Im pręty sure other parts of the codebase depend on this behaviour

geoff-vball: Did Aayush get the typed json-rpc errors PR done? There are other parts of our code that aren't checking for `nil, nil`, such as `warmpup.go` that we should also fix if we're going to keep the behaviour the same. We should also throw comments in a couple more places just to make sure as well.

arajasek (approved):
This looks good to me (i would've preferred to only include the panic fix for the patch release, but I think this is correct).

PR #9828, opened 2022-12-10
Messagepool: cache delegated siggys same as secp

arajasek:
## Related Issues

We currently use a key for delegated signatures based on the Ethereum encoding. While correct, this is:

- less extensible to other delegated siggys
- more work than necessary.

## Proposed Changes

Just use the message CID, same as we do for secp messages. This _should_ be completely safe, since it includes the (unchangeable) message itself and its accompanying signature.

## Checklist

- [ ] CI is green


ychiaoli18 (approved):
This makes sense and is more efficient than the original approach!

raulk:
This optimization already landed as part of another cleanup batch:

https://github.com/filecoin-project/lotus/blob/cdf3812e4081d35692b5ab81c573ab82d0df69cf/chain/messagepool/messagepool.go#L778-L779

PR #9831, opened 2022-12-11
Account abstraction

arajasek:
Allows sends from Embryos if and only if their delegated addresses are valid Ethereum Account addresses.


chain/consensus/filcns/filecoin.go
```diff
@@ -436,15 +436,15 @@ func (filec *FilecoinEC) VerifyWinningPoStProof(ctx context.Context, nv network.
...
 
-func isValidForSending(act *types.Actor) bool {
+func IsValidForSending(act *types.Actor) bool {
 	if builtin.IsAccountActor(act.Code) {
 		return true
 	}
 
-	// HACK: Allow Eth embryos to send messages
-	if !builtin.IsEmbryo(act.Code) || act.Address == nil || act.Address.Protocol() != address.Delegated {
+	if !builtin.IsEmbryo(act.Code) || act.Address == nil || act.Address.Protocol() != address.Delegated || act.Nonce != 0 {
```
Stebalien, line 444: Also ethereum accounts? Or are we including that in `IsAccountActor`?

chain/consensus/filcns/filecoin.go
```diff
@@ -521,7 +521,7 @@ func (filec *FilecoinEC) checkBlockMessages(ctx context.Context, b *types.FullBl
 				return xerrors.Errorf("failed to get actor: %w", err)
 			}
 
-			if !isValidForSending(act) {
+			if !IsValidForSending(act) {
```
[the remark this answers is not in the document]
arajasek: Yeah, we don't need to, but why not...i'll throw it in.

chain/consensus/filcns/filecoin.go
arajasek: Eth Accounts didn't exist yet -- adding now.

Stebalien (approved):

Issue #9820, opened 2022-12-08
EthBlockFromFilecoinTipSet bugfixes & improvements


arajasek:
The `newEthBlockFromFilecoinTipSet` method currently:

- gets the list of messages to be executed in the tipset, using `BlockMsgsForTipset`, and flattening the result
- gets the `MsgLookup` of each message by calling `StateSearchMsg` without specifying a starting tipset for the searchback, and with `allowReplaced = true`
- calls `newEthTxFromFilecoinMessageLookup` to find the appropriate index for the message to include in the return

## Suggested improvements
There are a few minor improvements that should be made to this approach:

- `MessagesForTipset` is a better method to retrieve executed messages, it does the flattening of the messages for you
- If we are going to use `StateSearchMsg`, we should not set `allowReplaced` to true -- this will return the _latest_ message with the matching nonce, which may not actually be the message in this tipset (eg. if it incurred a penalty and aborted without actually incrementing the nonce)
  - We would have to slightly change behaviour here to return a typed error for "replaced", and not-fail in that case
- We don't need to lookup the index for each message in `newEthTxFromFilecoinMessageLookup`, we should already have this information

## Alternative Approach
However, I would propose a different approach based on `StateCompute` instead. This method can be called with `height` set to `-1`, and with no `msgs` as input. Doing so will simply redo the computation, and return every message and its corresponding receipt _in order of execution_. I think this is going to be faster and correcter (avoids the bugs mentioned above), while still including all the necessary information.

If any information is missing, we can look into including it in the output of `StateCompute` if necessary.

We should likely replace the use of repeated `StateSearchMsg`s with a call to `StateCompute` instead.

PR #9837, opened 2022-12-11
eth: rpc: various Ethereum JSON-RPC API fixes

ychiaoli18:
## Related Issues
Fixes https://github.com/filecoin-project/ref-fvm/issues/1016
Fixes https://github.com/filecoin-project/ref-fvm/issues/1158
Fixes https://github.com/filecoin-project/ref-fvm/issues/1196
Fixes https://github.com/filecoin-project/ref-fvm/issues/1269
Fixes https://github.com/filecoin-project/lotus/issues/9820

## Checklist

- [ ] CI is green


node/impl/full/eth.go
```diff
@@ -754,6 +805,69 @@ func (a *EthModule) lookupEthAddress(ctx context.Context, addr address.Address)
...
+	if toAddr != nil {
+		if decodedParams, err := cbg.ReadByteArray(bytes.NewReader(smsg.Message.Params), uint64(len(smsg.Message.Params))); err == nil {
+			input = decodedParams
+		}
+	}
+
+	r, s, v, err := api.RecoverSignature(smsg.Signature)
+	if err != nil {
+		return api.EthTx{}, err
+	}
```
raulk, line 851: This will error out when dealing with non-Ethereum messages. Need to return 0 placeholders there.

api/eth_types.go
```diff
@@ -83,6 +83,9 @@ func (e EthBytes) MarshalJSON() ([]byte, error) {
 	if len(e) == 0 {
 		return json.Marshal("0x")
 	}
+	if len(e) == 1 && e[0] == 0 {
+		return json.Marshal("0x0")
+	}
 	s := hex.EncodeToString(e)
 	if len(s)%2 == 1 {
 		s = "0" + s
```
raulk, line 91: This is probably superfluous. My tests show that `hex.EncodeToString` always returns 2 * n characters, and I expect the behaviour to be consistent with [`EncodeLen`](https://pkg.go.dev/encoding/hex#EncodedLen), which also returns 2 * n as per asserted behavior in docs.

api/eth_types.go
```diff
@@ -83,6 +83,9 @@ func (e EthBytes) MarshalJSON() ([]byte, error) {
 	if len(e) == 0 {
 		return json.Marshal("0x")
 	}
+	if len(e) == 1 && e[0] == 0 {
```
raulk, line 86: What case does this deal with? Where do we return a non-nil byte string with a single 0 byte (which is different to an empty byte string).

raulk (commented):
We desperately need integration tests for all of this.

api/eth_types.go
ychiaoli18: According to [go-ethereum implementation](https://github.com/ethereum/go-ethereum/blob/687e4dc8555b3b53fd8d4a4974b8f8f4b594b552/common/hexutil/json.go#L147-L148), the bigint's zero value should marshal to 0x0. I'll change it accordingly.

ychiaoli18: This should be removed, and the type of R, S, V should be big.Int instead of bytes. The empty value of big.Int should be 0x0 and the empty value of bytes should be 0x

chain/types/ethtypes/eth_transactions.go
```diff
@@ -292,6 +300,33 @@ func (tx *EthTxArgs) Sender() (address.Address, error) {
 	return address.NewDelegatedAddress(builtintypes.EthereumAddressManagerActorID, ethAddr)
 }
 
+func RecoverSignature(sig typescrypto.Signature) (r, s, v EthBigInt, err error) {
+	if sig.Type != typescrypto.SigTypeDelegated {
+		return EthBigIntZero, EthBigIntZero, EthBigIntZero, fmt.Errorf("RecoverSignature only supports Delegated signature")
+	}
+
+	if len(sig.Data) != 65 {
+		return EthBigIntZero, EthBigIntZero, EthBigIntZero, fmt.Errorf("signature should be 65 bytes long, but get %v", len(sig.Data))
```
raulk, line 309:
```suggestion
		return EthBigIntZero, EthBigIntZero, EthBigIntZero, fmt.Errorf("signature should be 65 bytes long, but got %d bytes", len(sig.Data))
```

chain/types/ethtypes/eth_transactions_test.go
```diff
@@ -63,6 +63,52 @@ func TestTxArgs(t *testing.T) {
 	}
 }
 
+func TestSignatures(t *testing.T) {
```
raulk, line 66: Should add bad cases too.

PR #9858, opened 2022-12-13
feat:networking:  (Synchronous) Consistent Broadcast for Filecoin EC

adlrocha:
## Related Issues
Implementation of https://github.com/filecoin-project/FIPs/pull/527

## Checklist

- [ ] CI is green


jennijuju:
Oh whoa I didn’t even notice this until now! Thank you Alfonso for getting this in! @filecoin-project/lotus-tse can **one** of you throw this branch on your miner and confirm win posts stability? (AFTER WINTER BREAK)

chain/sub/incoming.go
```diff
@@ -102,6 +107,20 @@ func HandleIncomingBlocks(ctx context.Context, bsub *pubsub.Subscription, s *cha
 				log.Warnw("received block with large delay from miner", "block", blk.Cid(), "delay", delay, "miner", blk.Header.Miner)
 			}
 
+			// When we propose a new block ourselves, the proposed block also gets here through SyncSubmitBlock.
+			// If we are the block proposers we don't need to wait for delivery, we know the blocks are
+			// honest.
+			if src != self {
+				log.Infof("Waiting for consistent broadcast of block in height: %v", blk.Header.Height)
```
magik6k, line 114: Isn't this (and the one below) log a bit spammy?

itests/kit/init.go
```diff
@@ -40,6 +41,12 @@ func init() {
 
 	build.InsecurePoStValidation = true
 
+	// NOTE: If we want the payment channel itests to pass that use a
+	// block time of 5*millisecond, we need to set the consistent broadcast
+	// delay to something lower than that block time.
+	// todo: configure such a low delay only for paych tests.
+	build.CBDeliveryDelay = 2 * time.Millisecond
```
magik6k, line 48: Can we fully disable this for most tests? wdpost tests for example use 2ms block time, and it would be good to not slow them down even more.

magik6k (approved):

magik6k (commented):
(meant to comment, not approve)

itests/kit/init.go
adlrocha: I completely disabled consistent broadcast for itests. That being said, it would be great if we can at least enable it on a smoke consensus or deals itests with a decent block time (if there is any) just as a sanity-check that this doesn't break anything in the future.

chain/sub/incoming.go
```diff
@@ -102,6 +107,20 @@ func HandleIncomingBlocks(ctx context.Context, bsub *pubsub.Subscription, s *cha
 				log.Warnw("received block with large delay from miner", "block", blk.Cid(), "delay", delay, "miner", blk.Header.Miner)
 			}
 
+			// When we propose a new block ourselves, the proposed block also gets here through SyncSubmitBlock.
+			// If we are the block proposers we don't need to wait for delivery, we know the blocks are
+			// honest.
+			if src != self {
+				log.Debugf("Waiting for consistent broadcast of block in height: %v", blk.Header.Height)
+				if err := cb.WaitForDelivery(blk.Header); err != nil {
+					log.Errorf("couldn't deliver block to syncer over pubsub: %s; source: %s", err, src)
```
arajasek, line 116: i'd slightly tweak the error message here -- it currently sounds like something failed with the delivery effort, when in reality it's just because we don't _want_ to deliver this block to pubsub (for the reasons contained in `err`)

chain/sub/bcast/consistent.go
```diff
@@ -0,0 +1,180 @@
...
+	GcSanityCheck = 5
+	// GcLookback determines the number of epochs kept in the consistent
+	// broadcast cache.
+	GcLookback = 2
+)
+
+type blksInfo struct {
+	ctx    context.Context
+	cancel context.CancelFunc
+	blks   []cid.Cid
```
arajasek, line 33: I'm unsure about the thread-safety of this `blks` slice. We seem to be reading & modifying it somewhat unsafely, and while it's probably the case that the system won't ever race here, I'm...not sure.

Can we reason through this a bit?

chain/sub/bcast/consistent.go
```diff
@@ -0,0 +1,180 @@
...
+		return fmt.Errorf("equivocation detected for epoch %d. Two blocks being broadcast with same VRFProof", bh.Height)
+	}
+	return nil
+}
+
+func (cb *ConsistentBCast) GarbageCollect(currEpoch abi.ChainEpoch) {
+	cb.lk.Lock()
+	defer cb.lk.Unlock()
+
+	// keep currEpoch-2 and delete a few more in the past
```
arajasek, line 169: I'd consider doing a "fuller" (eg. lookback the past 24 hours) GC every (say) 1000 epochs or so. It should be pretty fast, and might be necessary to catch any escapees ;)

arajasek (commented):
Broadly LGTM, a few questions / suggestions.

chain/sub/bcast/consistent.go
```diff
@@ -0,0 +1,180 @@
...
+	delay time.Duration
+	// FIXME: Make this a slice??? Less storage but needs indexing logic.
+	m map[abi.ChainEpoch]*bcastDict
+}
+
+func newBcastDict() *bcastDict {
+	return &bcastDict{new(sync.Map)}
+}
+
+// TODO: the VRFProof may already be small enough so we may not need to use a hash here.
```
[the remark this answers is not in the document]
magik6k: Second that, turning it into a multihash is almost definitely slower, and also makes this code confusing to follow

chain/sub/bcast/consistent.go
```diff
@@ -0,0 +1,180 @@
...
+	return fmt.Errorf("different blocks with the same ticket already seen")
+}
+
+func (cb *ConsistentBCast) Len() int {
+	cb.lk.RLock()
+	defer cb.lk.RUnlock()
+	return len(cb.m)
+}
+
+func (cb *ConsistentBCast) RcvBlock(ctx context.Context, blk *types.BlockMsg) {
```
magik6k, line 111: Can we add doc comments to RcvBlock / WaitForDelivery which explain in some more depth what those methods are doing?

chain/sub/bcast/consistent.go
```diff
@@ -0,0 +1,180 @@
...
+	cb.lk.RLock()
+	bcastDict := cb.m[bh.Height]
+	cb.lk.RUnlock()
+	key, err := BCastKey(bh)
+	if err != nil {
+		return err
+	}
+	bInfo, ok := bcastDict.load(key)
+	if !ok {
+		return fmt.Errorf("something went wrong, unknown block with Epoch + VRFProof (cid=%s) in consistent broadcast storage", key)
```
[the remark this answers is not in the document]
arajasek: Seconded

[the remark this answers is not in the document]
adlrocha: Yep, my bad. I wrote this when we were having an active discussion about the "deprecation" of `xerrors` and forgot to roll back. Fixed.

chain/sub/bcast/consistent.go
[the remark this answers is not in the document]
adlrocha: Good points. Fixed.

build/params_mainnet.go
```diff
@@ -126,3 +127,7 @@ const BootstrapPeerThreshold = 4
 
 // we skip checks on message validity in this block to sidestep the zero-bls signature
 var WhitelistedBlock = MustParseCid("bafy2bzaceapyg2uyzk7vueh3xccxkuwbz3nxewjyguoxvhx77malc2lzn2ybi")
+
+// CBDeliveryDelay is the delay before deliver in the synchronous consistent broadcast.
+// This determines the wait time for the detection of potential equivocations.
+var CBDeliveryDelay = 6 * time.Second
```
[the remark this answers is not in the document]
adlrocha: > Does this impact block propagation? I guess no since pubsub checks don't touch this logic.

It shouldn't, right? The only thing we are delaying is the delivery of the block to the syncer (i.e. the call to [`InformNewBlock`](https://github.com/filecoin-project/lotus/pull/9858/files#diff-2b53d4ebd3dc2ee0318d15b3d746a581c19c2d376cf1bea25e309dc61214cfafR124)), but this is done after the Gossipsub validation so the propagation of the block should follow its course seamlessly. 

> Can we set this to a lower value while still getting the expected security benefits?

This is a great point, and this could definitely be a problem. These 6 seconds is the theoretical time that it takes for a block to be propagated to all of the network by Gossipsub. The assumption is that after this time there is high probability that any equivocation has been propagated enough to be caught by a majority of miners. Lowering this value will impact the effectiveness of the scheme, but I don't know to what extent or what is the minimum threshold. I know @jsoares and @guy.goren were working on getting some real data from mainnet with ProbeLab to see empirically what is the optimal delay. I'll let them chime in in case they have more info.

chain/sync_test.go
```diff
@@ -45,6 +45,11 @@ func init() {
 	policy.SetSupportedProofTypes(abi.RegisteredSealProof_StackedDrg2KiBV1)
 	policy.SetConsensusMinerMinPower(abi.NewStoragePower(2048))
 	policy.SetMinVerifiedDealSize(abi.NewStoragePower(256))
+
+	// these tests assume really fast block times. disabling
+	// the consistent broadcast delay to avoid them from adding
+	// an unnecessary overhead.
+	build.CBDeliveryDelay = 2 * time.Millisecond
```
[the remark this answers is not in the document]
adlrocha: You mean deferring setting back this to the original value at the end of the last test (assuming that tests run sequentially), or at the end of every test in case they are run randomly?

build/params_mainnet.go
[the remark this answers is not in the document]
jsoares: > Does this impact block propagation? I guess no since pubsub checks don't touch this logic.

As @adlrocha said, propagation shouldn't be affected. Anything else would negate the benefits of the patch: we want the guard time to be larger than the propagation time, which is easier the faster propagation is.

> Can we set this to a lower value while still getting the expected security benefits?

The value can be set lower. In fact, there is no "cliff" behaviour and the security gains can be approximated (if not formally, at least intuitively) as a logistic function centred on the average propagation delay: while you don't necessarily _need_ to cover the 99th-percentile, too low thresholds would catch few attacks and you'd always want to at least cover the average case.

Alas, our only real-world data was collected from the bootstrapper nodes (via Sentinel). [Our analysis of that data](https://gist.github.com/jsoares/9ce4c0ba6ebcfd2afa8f8993890e2d98) shows that the vast majority of blocks reach the bootstrappers in < 1s. However, there are a few weaknesses here, primarily as (i) we're trusting source timestamps; (ii) bootstrapper nodes are not necessarily representative of the average node in terms of connectivity (both link & centrality, even mesh configuration). 

In the meantime, ProbeLab is working on a network-wide GossipSub study that will yield a higher-confidence distribution for block propagation. Alas, that work is taking longer than we expected and there is still no time frame for the results. The plan was, therefore, to start with the 6s used as the assumption in the GossipSub security analysis, keep it user-configurable, and push better defaults once we have the data. I expect we'll land in the 1 to 2 second range, but don't hold me to that.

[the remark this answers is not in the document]
magik6k: Note that here I'm more concerned about a block arriving late making SPs be late with their own block.

Propagation [delay is 10s](https://github.com/filecoin-project/lotus/blob/master/build/params_mainnet.go#L98), state compute today takes up to 14s due to cron issues. With this patch worst case if some set of SPs choose to send blocks late, we'd have essentially no time spare for WindowPoSt, essentially cutting into propagation delay for the next epoch - which afaict would be either comparable or worse than epoch boundary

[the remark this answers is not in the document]
jsoares: Right, that's a valid concern given we're right at the threshold. How high would you be comfortable going? Given the context above, I think we could conceivably adjust down a second or two to leave some buffer time -- and still reassess when the study results are available.

arajasek:
So I think I share @magik6k's concern here -- this is unfortunate, but the good news is that we're expecting things to get better fairy soon.

What would we think about shipping this with an artificially low value (2 seconds?), and running that until nv19. That gives time for any CRITICAL issues with the code to emerge (not that I expect any). Upon nv19, we can set this to the desired 6 seconds.

We also _probably_ want to lower the default propagation delay when this is set to 6 seconds, I think?

jsoares:
> So I think I share @magik6k's concern here -- this is unfortunate, but the good news is that we're expecting things to get better fairy soon.
> 
> What would we think about shipping this with an artificially low value (2 seconds?), and running that until nv19. That gives time for any CRITICAL issues with the code to emerge (not that I expect any). Upon nv19, we can set this to the desired 6 seconds.
> 
> We also _probably_ want to lower the default propagation delay when this is set to 6 seconds, I think?

(note: I'm responding to this but I'm not the author of the FIP...)

That seems reasonable to me. I'm not sure it will be that far off the final numbers. 

As for adjusting the cutoff, I think the propagation study will definitely provide good data to inform this. That said:
1. The two things are somewhat independent, as long as they fit within the time available. `PropagationDelaySecs` sets the cutoff for _accepting_ blocks, whereas `CBDeliveryDelay` defines how long we'll wait to potentially _reject_ a block received by the cutoff. It does **not** extend the acceptance window. (also, both of these things have pretty bad names, as they describe bounds rather than delays)
2.  Setting the cutoff too low can have a noticeable negative impact, whereas setting the guard time too low will just decrease its effectiveness. So we probably want to be more conservative with respect to the former.

magik6k:
> It does not extend the acceptance window

That's correct, but in the current implementation it extends the latest time at which block validation begins. Ideally we'd start validating those blocks "optimistically", and somehow retro-reject them if we see another within `CBDeliveryDelay`

jsoares:
> Ideally we'd start validating those blocks "optimistically", and somehow retro-reject them if we see another within `CBDeliveryDelay`

Indeed, and that's suggested in the FIP as a potential optimisation. It's unclear to me how much of a gain that represents in practice, and it may add a bit of complexity, but it's a possible direction if we are indeed time-constrained. My expectation is still that the final parameters will be very comfortable wrt to the available window.

chain/sub/bcast/consistent.go
arajasek: Actually, looking at the code closer, we don't actually ever modify it. So we're probably okay here.

arajasek: @adlrocha I don't think that's sufficient. I think it needs to be held until no more writing happens, which is the end of the method (as currently written). Consider the following situation:

- 2 blocks A and B come in with the same VRFProof. 
- A gets the lock first, creates an empty BCastDict (line 119), doesn't find a match for the common `key` on line 125, drops the lock
- B gets the lock, loads the empty BCastDict created by A, also doesn't find a match for the common `key` on line 125, drops the lock
- A `store`s `blkACid` as the only entry corresponding to `key` on line 141, returns
- B _overwrites_ that, storing `blkBCid` as the _only_ entry corresponding to `key` on line 141, returns
- Both A and B succeed their respective calls to `WaitForDelivery`

adlrocha:
@arajasek, all comments addressed. I changed the locking strategy to make it less fine-grained but easier to reason about (as discussed above). We could recover the fine-grained locking and make it more understandable by using https://github.com/gammazero/keymutex, but I didn't want to introduce any additional dependency.

(For context, we used this dependency extensively for storetheindex, and is written and maintained by Andrew).

build/params_mainnet.go
```diff
@@ -126,3 +127,7 @@ const BootstrapPeerThreshold = 4
 
 // we skip checks on message validity in this block to sidestep the zero-bls signature
 var WhitelistedBlock = MustParseCid("bafy2bzaceapyg2uyzk7vueh3xccxkuwbz3nxewjyguoxvhx77malc2lzn2ybi")
+
+// CBDeliveryDelay is the delay before deliver in the synchronous consistent broadcast.
+// This determines the wait time for the detection of potential equivocations.
+var CBDeliveryDelay = 2 * time.Second
```
arajasek, line 133: Do we not have to add this variable to the other params files too? Are you able to (for example) `make calibnet` successfully?

Also why is it a var and not a const here?

adlrocha: Good catch, it should be a `const` and it should be an all testnet builds for them to compile. Fixed. Thanks!

Issue #9816, opened 2022-12-08
Add eth_ operations to the Lotus Gateway


maciejwitowski:
This is a tracking issue for https://github.com/filecoin-project/ref-fvm/issues/1214

Quoting the original one for completeness:

Hello devs,

Glif nodes provider is here.
Right now we have a doubt how to give users access to the wss endpoint. **The main pain is `eth_subscribe` operation - it is require write permissions.** 
For now we have 2 models of access to the wss:
1. Lotus Gateway, but it is limited only for Filecoin operations
2. Access with JWT token with `read, write` permissions - but this isn't weapon of choice - `write` permissions are too wide.

I believe the easiest way will be to add eth operations to the Lotus Gateway.

The place where the Lotus Gateway code is placed: https://github.com/filecoin-project/lotus/blob/release/fevm-wallaby/api/api_gateway.go

PR #9863, opened 2022-12-13
feat: eth gateway api

magik6k:
## Related Issues

resolves https://github.com/filecoin-project/lotus/issues/9816

## Proposed Changes
Add eth_ rpc support to gateway

## Additional Info
The actual limiter code in gateway still todo, this mostly does the plumbing for now.

## Checklist

- [ ] CI is green


magik6k:
Current state:
* All eth_ calls are now exposed by lotus-gateway
* EthAccounts returns empty list
* All calls use the default 3 tokens for now
* EthFeeHistory lookback is limited to 128 epochs by default
* By default each connection can install up to 16 filters
* subscribe is not supported
  * The only real way to do this correctly requires a bunch of changes is go-jsonrpc.
    * I'm surprised that what we have today for eth_subscribe works at all
      * It doesn't return the subscription id correctly
      * It doesn't invoke the correct notification on the client
      * The params in the notifications are in the wrong format (array with [id, value] vs object with those)
      * eth_unsubscribe id has nothing to do with the id used by go-jsonrpc when handling the channel
    * To fix this we'll need to add proper support for reverse calls in go-jsonrpc, and add support for object-based parameters
      * Or allow hooking into the raw rpc stream, but that has performance implications
      * Or add support for custom channel logic, but that's about the same amout of work needed for reverse calling, and less flexible
      * Implementing reverse calls is probably 1-2 days, but is that worth it now for proper eth_subscribe support?

magik6k:
Current state:
* Like previously eth_subscribe is not supported yet, waiting for https://github.com/filecoin-project/go-jsonrpc/pull/88, but that shouldn't prevent this PR from being testable/useful
* All other eth_ calls are supported in gateway with limits similar to other filecoin-native RPCs
* Lite nodes should now route all eth_ calls to the backing gateway node, so eth lite-nodes are a thing now.

travisperson (approved):
Just 1 typo I believe, otherwise LGTM

Issue #9825, opened 2022-12-09
Refactor: Move "core" Eth types from api to types


arajasek:
It's inappropriate for the consensus components in Lotus to be depending on the API. With https://github.com/filecoin-project/ref-fvm/issues/1204, things like `EthTxArgs` become part of the Filecoin protocol, and so should move to chain/types.

arajasek:
We should add a `ToDigest` method on Ethereum Txs for siggy validation too.

PR #9870, opened 2022-12-14
chore: remove dependencies on api

geoff-vball:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/9825

## Proposed Changes
Moves a bunch of stuff around so that packages that shouldn't depend on `api` don't have to anymore.

## Checklist

- [ ] CI is green


arajasek:
@geoff-vball Thanks for this! I need to think a bit more about whether we want to move things like `HeadChange` etc. into the core consensus types. It's a tradeoff between severing odd dependencies on API types vs. bloating the core consensus types.

How doable would it be to land _just_ the last commit? That's the one I'm keenest on, because the only change urgently needed (blocking v18) IMO is having the core chain validation (state computation) path run through API types.

geoff-vball:
Might get to these changes later. For now, just making minimal changes in https://github.com/filecoin-project/lotus/pull/9872

Issue #9778, opened 2022-12-02
NV18: EVM actor: Add tooling to inspect state


arajasek:
We'll probably want CLI (and perhaps API) tooling to query an EVM contract's `contract_state` and other fields.

jennijuju:
- get contract params
- get contract address (eth and filecoin)
- eth contract storage location map
- offline mode of `invoke-evm-actor`
- translate between eth tx id to filecoin message cid

Stebalien:
> `eth contract stat <filecoin-or-eth-address>`

Returns:

- Eth address
- Fil address
- Code hash

> `eth contract storage get <filecoin-or-eth-address> <key>`

Returns the value of the storage slot.

> `eth contract call [--to <filecoin-or-eth-address>] [--from ...] <params>`

Call `EthCall` to test a contract execution. I believe this already mostly exists in `invoke-evm-actor`, we just need a version that only runs it locally.


> `eth contract storage list <filecoin-or-eth-address>`

Returns a list of storage keys.

> `eth contract storage list --include-values <filecoin-or-eth-address>`

Returns a list of storage key/value tuples.

Stebalien:
- And some way to convert eth addresses to fil addresses and vice-versa.
- And some way to do the same for eth txn ids and message CIDs.

Stebalien:
Note: storage listing will be hard right now because we don't expose this from the actors.

jennijuju:
@shrenujbansal I found this can be useful for eth tx hash -> filecoin cid https://github.com/filecoin-project/lotus/blob/f681a350fcc9c39a3022a6708fa5da777dd83efa/api/eth_types.go#L415

shrenujbansal:
Revised scope for now:

eth contract stat <filecoin-or-eth-address>
eth contract call [--to <filecoin-or-eth-address>] [--from ...] <params>
<ways to use convert filecoin/eth addrs back and forth>

Postponing:
storage slot tools - needs updated go libs for KAMT parsing
eth tx id -> filecoin cid

PR #9897, opened 2022-12-16
feat: cli: Add cli to inspect Eth addresses and simulate calls

shrenujbansal:
## Related Issues

https://github.com/filecoin-project/lotus/issues/9778

## Proposed Changes

> ➜  lotus git:(sbansal/eth-state-cli) ✗ ./lotus eth stat --filaddr t410fiux3jxq5bcdb3ctzgghvxt25ss3zgos2ifm3qki
> Filecoin address:  f410fiux3jxq5bcdb3ctzgghvxt25ss3zgos2ifm3qki
> Eth address:  0x452fb4de1d08861d8a79318f5bcf5d94b7933a5a
> Code cid:  bafk2bzacednmsvivt6tszxgnn3xy3iuwqqlhn5uif6mf4gj2du6ec4m6emble

## Checklist

- [x] CI is green

### DIFF
--- a/cli/eth.go
+++ b/cli/eth.go
@@ -4,12 +4,15 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+
 	"github.com/filecoin-project/go-address"
+
 	"github.com/filecoin-project/lotus/api/v0api"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
-	"github.com/urfave/cli/v2"
-	"golang.org/x/xerrors"
 )
 
 var EthCmd = &cli.Command{
@@ -28,6 +31,9 @@ func ethAddrFromFilecoinAddress(ctx context.Context, addr address.Address, fnapi
 	switch addr.Protocol() {
 	case address.BLS, address.SECP256K1:
 		faddr, err = fnapi.StateLookupID(ctx, addr, types.EmptyTSK)
+		if err != nil {
+			return ethtypes.EthAddress{}, addr, err
+		}
 	case address.Actor, address.ID:
 		f0addr, err := fnapi.StateLookupID(ctx, addr, types.EmptyTSK)
 		if err != nil {
@@ -91,6 +97,9 @@ var EthGetAddressCmd = &cli.Command{
 				return err
 			}
 			eaddr, faddr, err = ethAddrFromFilecoinAddress(ctx, addr, api)
+			if err != nil {
+				return err
+			}
 		} else if ethaddr != "" {
 			addr, err := hex.DecodeString(ethaddr)
 			if err != nil {
--- a/cli/eth.go
+++ b/cli/eth.go
@@ -24,62 +24,23 @@ var EthCmd = &cli.Command{
 	},
 }
 
-func ethAddrFromFilecoinAddress(ctx context.Context, addr address.Address, fnapi v0api.FullNode) (ethtypes.EthAddress, address.Address, error) {
-	var faddr address.Address
-	var err error
-
-	switch addr.Protocol() {
-	case address.BLS, address.SECP256K1:
-		faddr, err = fnapi.StateLookupID(ctx, addr, types.EmptyTSK)
-		if err != nil {
-			return ethtypes.EthAddress{}, addr, err
-		}
-	case address.Actor, address.ID:
-		faddr, err = fnapi.StateLookupID(ctx, addr, types.EmptyTSK)
-		if err != nil {
-			return ethtypes.EthAddress{}, addr, err
-		}
-		fAct, err := fnapi.StateGetActor(ctx, faddr, types.EmptyTSK)
-		if err != nil {
-			return ethtypes.EthAddress{}, addr, err
-		}
-		if fAct.Address != nil && (*fAct.Address).Protocol() == address.Delegated {
-			faddr = *fAct.Address
-		}
-	case address.Delegated:
-		faddr = addr
-	default:
-		return ethtypes.EthAddress{}, addr, xerrors.Errorf("Filecoin address doesn't match known protocols")
-	}
-
-	ethAddr, err := ethtypes.EthAddressFromFilecoinAddress(faddr)
-	if err != nil {
-		return ethtypes.EthAddress{}, addr, err
-	}
-
-	return ethAddr, faddr, nil
-}
-
 var EthGetAddressCmd = &cli.Command{
-	Name:      "stat",
-	Usage:     "Print eth/filecoin addrs and code cid",
-	ArgsUsage: "",
+	Name:  "stat",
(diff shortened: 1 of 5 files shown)


cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+var EthCmd = &cli.Command{
+	Name:  "eth",
+	Usage: "Query eth contract state",
+	Subcommands: []*cli.Command{
+		EthGetAddressCmd,
+		EthCallSimulateCmd,
+	},
+}
+
+func ethAddrFromFilecoinAddress(ctx context.Context, addr address.Address, fnapi v0api.FullNode) (ethtypes.EthAddress, address.Address, error) {
```
arajasek, line 27: Let's move the helper to the end of the file?

cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+var EthGetAddressCmd = &cli.Command{
+	Name:      "stat",
+	Usage:     "Print eth/filecoin addrs and code cid",
+	ArgsUsage: "",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "filaddr",
+			Value: "",
```
arajasek, line 70:
```suggestion
```

cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+			Usage: "Filecoin address",
+		},
+		&cli.StringFlag{
+			Name:  "ethaddr",
+			Value: "",
```
arajasek, line 75:
```suggestion
```

cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+	Usage:     "Print eth/filecoin addrs and code cid",
+	ArgsUsage: "",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "filaddr",
+			Value: "",
+			Usage: "Filecoin address",
+		},
+		&cli.StringFlag{
+			Name:  "ethaddr",
```
arajasek, line 74:
```suggestion
			Name:  "ethAddr",
```

cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+	return ethAddr, faddr, nil
+}
+
+var EthGetAddressCmd = &cli.Command{
+	Name:      "stat",
+	Usage:     "Print eth/filecoin addrs and code cid",
+	ArgsUsage: "",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "filaddr",
```
arajasek, line 69:
```suggestion
			Name:  "filAddr",
```

cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+	ethAddr, err := ethtypes.EthAddressFromFilecoinAddress(faddr)
+	if err != nil {
+		return ethtypes.EthAddress{}, addr, err
+	}
+
+	return ethAddr, faddr, nil
+}
+
+var EthGetAddressCmd = &cli.Command{
+	Name:      "stat",
```
arajasek, line 64: Odd name for the command?

arajasek (commented):
Thanks for this PR! It's really nice to see this stuff coming along.

I think there's a non-trivial product question around what  we want the "get address" method to return. At its simplest, we could do the following:

- given a filecoin address, only convert f410 addresses to Eth addresses (and fail everything else)
- given an eth address, simply invoke `ToFilecoinAddress`

We _could_ do more. We could convert any secp address into the corresponding f4 adderss in the EAM's namespace. We could be as aggressive as the `lookupEthAddress` method (see [here](https://github.com/filecoin-project/lotus/blob/d71f5a7d12ef56270c43c7837ff6b3f30f50b009/node/impl/full/eth.go#L723)).

I think we need to loop in the FVM's product team for design feedback here. In the meantime, I would propose implementing the simpler approach for filecoin -> eth and landing that:

- given a filecoin address, if f410 convert it
- if not f410, lookup the actor, lookup its `Address`, and if it's f410, convert it

cli/eth.go
shrenujbansal: This was specified by steb. I don't have a strong opinion. Do you have a preferred name?

cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+	Usage:     "Simulate an eth contract call",
+	ArgsUsage: "[from] [to] [params]",
+	Action: func(cctx *cli.Context) error {
+
+		var fromEthAddr ethtypes.EthAddress
+		fromAddr, err := hex.DecodeString(cctx.Args().Get(0))
+		if err != nil {
+			return err
+		}
+		copy(fromEthAddr[:], fromAddr)
```
geoff-vball, line 106: We should use EthAddressFromHex() here.

cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+
+		var toEthAddr ethtypes.EthAddress
+		toAddr, err := hex.DecodeString(cctx.Args().Get(1))
+		if err != nil {
+			return err
+		}
+		copy(toEthAddr[:], toAddr)
```
geoff-vball, line 113: EthAddressFromHex() here as well

cli/eth.go
geoff-vball: Better actually to use it on line 102. That way we only have 1 function call and don't have to pre-declare `fromEthAddr`

cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+	ethAddr, err := ethtypes.EthAddressFromFilecoinAddress(faddr)
+	if err != nil {
+		return ethtypes.EthAddress{}, addr, err
+	}
+
+	return ethAddr, faddr, nil
+}
+
+var EthGetAddressCmd = &cli.Command{
+	Name:      "stat",
```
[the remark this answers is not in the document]
shrenujbansal: Could you refer to what the 2 cmd should be?
The idea of this cmd is "give me the fil/eth addr and ill give you all relevant info about this actor/contract"
I suppose I should change the name from EthGetAddressCmd to EthGetInfo

cli/eth.go
shrenujbansal: sigh. I looked for a helper for a bit but couldn't find it. Thanks for pointing it out

cli/eth.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+		} else if ethAddr != "" {
+			if ethAddr[:2] == "0x" {
+				ethAddr = ethAddr[2:]
+			}
+			addr, err := hex.DecodeString(ethAddr)
+			if err != nil {
+				return err
+			}
+			copy(eaddr[:], addr)
```
geoff-vball, line 72: Oh, same thing with ethtypes.EthAddressFromHex() here

cli/eth.go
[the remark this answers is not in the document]
geoff-vball: `EthGetInfoFromFilecoinAddr` and `EthGetInfoFromEthAddr` maybe? I guess it's fine as is. Just seems like we're mostly doing two separate thing, even though the result comes from the same api call.

geoff-vball (approved):

PR #9899, opened 2022-12-16
_ci_ Add CircleCI Dependency caching

ianconsolata:
## Related Issues
None, but this will speed up most jobs by about 60 seconds

## Proposed Changes
Cache our go dependencies

## Additional Info
Cache keys are prefix matched, so if the checksum of the `go.sum` or `go.mod` fails to match, it will load the most recent dependencies for the current GO_VERSION_MIN and call make deps to update them, saving them under the full key.

See full documentation to understand how the cache keys work:
  https://circleci.com/docs/caching/
  
## Checklist

- [X] CI is green

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,6 +109,23 @@ commands:
           name: fetch all tags
           command: |
             git fetch --all
+  install_go_dependencies:
+    steps:
+      - restore_cache:
+          keys:
+            - v1-go-mod-{{ arch }}-{{ checksum "GO_VERSION_MIN" }}-{{ checksum "go.mod" }}-{{ checksum "go.sum" }}
+            - v1-go-mod-{{ arch }}-{{ checksum "GO_VERSION_MIN" }}-{{ checksum "go.mod" }}
+            - v1-go-mod-{{ arch }}-{{ checksum "GO_VERSION_MIN" }}
+            - v1-go-mod-{{ arch }}
+            - v1-go-mod-
+      - run:
+          command: make deps
+          no_output_timeout: 30m
+      - save_cache:
+          key: v1-go-mod-{{ arch }}-{{ checksum "GO_VERSION_MIN" }}-{{ checksum "go.mod" }}-{{ checksum "go.sum" }}
+          paths:
+            - "/go/pkg/mod"
+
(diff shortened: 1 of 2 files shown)


ianconsolata:
Wow, @geoff-vball just checked out my build and on one of the test jobs where it loaded the cache, the make deps step took only 3 seconds! Usually it takes us about a minute, so this should shave about a minute off every ci run, and many many more minutes off our bill based on how much we parallelize things.

As a step two, I think we could cache the FFI build / download as well, though that requires more careful thinking about cache keys. Perhaps just checksum the `extern/filecoin-ffi/` folder? Or use the subproject commit somehow?

geoff-vball:
@ianconsolata I think we might not have the right folder yet... `make deps` is fast, but we split `make deps lotus` into `make deps` and `make lotus`, with `make lotus` still taking about a minute, so we haven't saved much yet. It's still downloading all of the dependencies, so I think we have the wrong folder.

geoff-vball:
As for `/extern/filecoin-ffi`, `git submodule status` will return the commit hash of the all the submodules (along with some other info). We could cache off of the commit hash

ianconsolata:
Unfortunately we can't run arbitrary commands in the cache keys: https://circleci.com/docs/caching/#using-keys-and-templates

geoff-vball:
`/home/circleci/go` is the right path

ianconsolata:
We could also try to build a custom build env container or something to cache the apt dependencies, but loading them only takes 3 seconds or so, so I think the maintenance overhead of that is not worth it.

ianconsolata:
I set the GOPATH to `/go` instead, I think the default home directory is in a different place on MacOS. This should make the GOPATH consistent across architectures.

geoff-vball:
Looks like it doesn't like you touching the root dir

ianconsolata:
OK, finally got it working and it basically doesn't seem to save us any time at all. In fact, loading and saving the cache may actually take more time that letting the build process download the go dependencies while building. Getting the FFI builds also only takes about 3 seconds, and the apt deps also only take a few seconds. Overall, even if we do invest the time in more aggressive caching, it won't save us much (if any) time.

Issue #9846, opened 2022-12-12
bug: splitstore: deadlock during compaction (setup?)


magik6k:
Goroutines: https://gist.github.com/magik6k/4e13c2d53a5ba491d35b16d7a24c1b4a

Got what looks a deadlock in compaction a few momets after startup:

Splitstore compact goroutine, holds `Splitstore.mx.Lock()`, stuck on `Mpool.lk.Lock`
```
goroutine 24285 [semacquire, 27 minutes]:
sync.runtime_SemacquireMutex(0xc041b135c0?, 0xe0?, 0x4611e7c?)
	/usr/lib/go/src/runtime/sema.go:77 +0x25
sync.(*Mutex).lockSlow(0xc0239ca680)
	/usr/lib/go/src/sync/mutex.go:171 +0x165
sync.(*Mutex).Lock(...)
	/usr/lib/go/src/sync/mutex.go:90
github.com/filecoin-project/lotus/chain/messagepool.(*MessagePool).ForEachPendingMessage(0x18?, 0xc05ec065e8)
	/home/magik6k/github.com/filecoin-project/go-lotus/chain/messagepool/messagepool.go:462 +0x71
github.com/filecoin-project/lotus/blockstore/splitstore.(*SplitStore).applyProtectors(0xc002ecc280)
	/home/magik6k/github.com/filecoin-project/go-lotus/blockstore/splitstore/splitstore_compact.go:460 +0x185
github.com/filecoin-project/lotus/blockstore/splitstore.(*SplitStore).doCompact(0xc002ecc280, 0xc0532271c0)
	/home/magik6k/github.com/filecoin-project/go-lotus/blockstore/splitstore/splitstore_compact.go:544 +0x59d
github.com/filecoin-project/lotus/blockstore/splitstore.(*SplitStore).compact(0xc002ecc280, 0x0?)
	/home/magik6k/github.com/filecoin-project/go-lotus/blockstore/splitstore/splitstore_compact.go:497 +0x198
github.com/filecoin-project/lotus/blockstore/splitstore.(*SplitStore).HeadChange.func1()
	/home/magik6k/github.com/filecoin-project/go-lotus/blockstore/splitstore/splitstore_compact.go:131 +0x133
created by github.com/filecoin-project/lotus/blockstore/splitstore.(*SplitStore).HeadChange
	/home/magik6k/github.com/filecoin-project/go-lotus/blockstore/splitstore/splitstore_compact.go:124 +0x305
```

[...]

PR #9903, opened 2022-12-17
fix: splitstore: Don't deadlock in mpool protector

magik6k:
## Related Issues
Addresses https://github.com/filecoin-project/lotus/issues/9846

## Proposed Changes
Don't deadlock in splitstore compaction when applying mpool protector; Bail from compaction when mpool is busy.

A better fix would be to maybe drop the splitstore lock while applying protectors, and re-get the lock in protector callback.

## Additional Info
My node has hit this 4 times in a row now, and it's very annoying, so I have fixed it.

## Checklist

- [ ] CI is green


arajasek (approved):
Acceptable quickfix, I think.

PR #9910, opened 2022-12-19
feat: ci: make ci more efficient

geoff-vball:
## Proposed Changes

- Creates an initial build step that all commands can share so they don't all have to build themselves.
- Uses medium+ sized containers for most things. Some jobs are kept on 2xl executors because of parallelism/RAM requirements. Medium+ uses 15 credits/min compared to 80/min for 2xl, and barely causes a slowdown in many of our tests. [More info on executors here](https://circleci.com/product/features/resource-classes/#docker).
- Did some general cleanup and refactoring.

## Additional Info
You can look at the stats for a run of master as it is currently [here](https://app.circleci.com/insights/github/filecoin-project/lotus/workflows/ci/overview?branch=gstuart%2Fci-bench-old).
You can look at the stats for this brach [here](https://app.circleci.com/insights/github/filecoin-project/lotus/workflows/ci/overview?branch=gstuart%2Fci-medium%2B).

Based on these results, we should see a decrease in credit usage of almost 75% on each successful run. In addition, I've decreased the timeout for tests from 30m to 20m, and rerunning failed tests will no longer have to rebuild lotus. This should get us some additional savings dealing with our flaky tests (they're next on my list to tackle though).

These changes only lead to a ci slowdown from about 9min to 11min, but will speed up re-running flaky tests. We can probably target our longer tests with more powerful executors if we want to get this time back down, but I think it's still completely in the realm of acceptable.

There is a chance that running on the less powerful executors might increase test flakiness. We can tune this on a test-by-test basis if we want. I'm hoping to tackle some of the flakiness directly so it's not even a consideration.

## Checklist

- [ ] CI is green

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ executors:
     docker:
       # Must match GO_VERSION_MIN in project root
       - image: cimg/go:1.18.8
-    resource_class: large
+    resource_class: medium+
   golang-2xl:
     docker:
       # Must match GO_VERSION_MIN in project root
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -186,9 +186,8 @@ jobs:
               --format standard-verbose \
               --junitfile /tmp/test-reports/<< parameters.suite >>/junit.xml \
               --jsonfile /tmp/test-artifacts/<< parameters.suite >>.json \
-              -- \
-              << parameters.go-test-flags >> \
-              << parameters.target >>
+              --packages="<< parameters.target >>" \
+              -- << parameters.go-test-flags >>
           no_output_timeout: 30m
       - store_test_results:
           path: /tmp/test-reports
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,19 +36,13 @@ commands:
     steps:
       - checkout
       - git_fetch_all_tags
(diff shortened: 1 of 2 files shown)


magik6k (commented):
Config changes look good.

However I'm not sure if the cost savings are worth slowing each CI iteration by 2min / 20%.

I'm not sure how much each run is in $ terms, but I'd be surprised if it's over single-digit-usd. Dev time is imo worth far more than that (math can/should be done for a more solid argument here).

magik6k:
(do note that you / others on the team will be staring at CI jobs hundreds of times in the next year, waiting to see if the test they wrote works; Minutes here turn into days/weeks of dev time)

geoff-vball:
@magik6k The pricing information I could find is [here](https://circleci.com/docs/faq/#what-do-i-pay-for). The current CI uses about ~20k credits per run, which puts the cost per run around 12 USD. With our average usage over the last 90 days, we've been using about $6K/month (and most people were gone for 2 weeks). 

Also note that in the happy path, this slows us down a bit, but it will actually speed up rerunning flaky tests because they won't have to build again.

magik6k:
Oooh, the faster re-run makes this objectively better.

magik6k (approved):

PR #9933, opened 2022-12-23
fix: eth_getTransactionByHash + eth_getTransactionReceipt fixes

raulk:
## Related Issues

- Hash was 0, thus leading to issues with Hardhat and other tools.
- BlockHash, BlockNumber, and TransactionIndex must be null (not zero) when tx is returned from mpool.
- Transaction receipt with no logs should return an empty logs array (instead of null).
- For correctness, a transaction receipt with no logs should return a 256 bytes sized bloom.
- Return nonce.

## Proposed Changes

Fix these issues and add coverage in FEVM integration tests.

## Checklist

- [ ] CI is green


itests/eth_deploy_test.go
```diff
@@ -94,17 +96,25 @@ func TestDeployment(t *testing.T) {
 	hash := client.EVM().SubmitTransaction(ctx, &tx)
 	fmt.Println(hash)
 
+	mpoolTx, err := client.EthGetTransactionByHash(ctx, &hash)
+	require.NoError(t, err)
+
+	// require that the hashes are identical
+	require.Equal(t, hash, mpoolTx.Hash)
+
+	// these fields should be nil because the tx hasn't landed on chain.
```
arajasek, line 105: Yeah, let's not play this game, we're inviting flakiness here.

Can we just try to lookup the receipt before making these assertions? If found, let's simply skip them (with a log saying that we're doing so).

itests/eth_deploy_test.go
```diff
@@ -28,7 +28,9 @@ func TestDeployment(t *testing.T) {
 	// He who writes the second test, shall do that.
 	// kit.QuietMiningLogs()
 
-	blockTime := 100 * time.Millisecond
+	// reasonable blocktime so that the tx sits in the mpool for a bit during the test.
+	// although this is non-deterministic...
+	blockTime := 1 * time.Second
```
arajasek, line 33: I'd consider increasing this -- this test isn't really constrained by epochs (needs ~3 epochs), so even if the blocktime were like 5 seconds, it would still finish pretty quickly

itests/eth_deploy_test.go
```diff
@@ -94,17 +96,25 @@ func TestDeployment(t *testing.T) {
...
 	changes, err := client.EthGetFilterChanges(ctx, pendingFilter)
 	require.NoError(t, err)
 	require.Len(t, changes.Results, 1)
 	require.Equal(t, hash.String(), changes.Results[0])
 
-	time.Sleep(5 * time.Second)
-
 	var receipt *api.EthTxReceipt
-	for i := 0; i < 10000000000; i++ {
+	for i := 0; i < 20; i++ {
```
arajasek, line 116: Let's just actually `StateWaitMsg` here instead of this funkiness -- if we don't have the tooling to easily get the msgCid, slap a TODO on it for now (I think the work @geoff-vball  is doing will make this easy to do).

arajasek (approved):
Approved if we want a quick merge, but would like to see the suggested changes to the test implemented (can wait until we do the feat/nv18-fevm -> master merge if necessary).

itests/eth_deploy_test.go
raulk: This is actually a pretty important assertion (we had bugs here before, so want to lock these for regression testing). I'm onboard with deflaking, as long as we don't withdraw coverage we need.

itests/eth_deploy_test.go
raulk: Plus the boot up time for the chain (around 10 epochs).

itests/eth_deploy_test.go
raulk: We do need to check that the receipt is null while the transaction hasn't landed on chain -- it's part of the test's coverage.

raulk:
In case you're wondering if I'm making up the how this API should behave, I am not. Sources include:

- https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_gettransactionbyhash
- https://ethereum.org/es/developers/docs/apis/json-rpc/
- Alchemy docs

itests/eth_deploy_test.go
geoff-vball: The Miners in the itest framework have `pause` and `unpause` functions exposed, but we should implement a function for the test kit that allows us to pause and unpause all mining.

itests/eth_deploy_test.go
arajasek: Sure, keep doing that every 500ms, but the cue to stop is StateWaitMsg returning (not "i waited 10 seconds").

raulk: We still need a timeout for the test (should not let it run indefinitely). We could have a timed context with StateWaitMsg, but at that point we're achieving the same effect as here with more complicated code, so it's a matter of choice.

PR #9961, opened 2023-01-04
Eth JSON-RPC: support passing uint64 in JSON-RPC arguments for EthUint64

ychiaoli18:
## Related Issues
Fixes https://github.com/filecoin-project/ref-fvm/issues/1365

## Checklist

- [ ] CI is green


chain/types/ethtypes/eth_types.go
```diff
@@ -40,18 +40,24 @@ func (e EthUint64) MarshalJSON() ([]byte, error) {
...
+		*e = eint
+		return nil
+	} else if eint, err := strconv.ParseUint(string(b), 10, 64); err == nil {
+		*e = EthUint64(eint)
+		return nil
 	}
-	eint := EthUint64(parsedInt)
-	*e = eint
-	return nil
+	return fmt.Errorf("cannot parse %s into EthUint64", string(b))
```
raulk, line 60: This message is externally visible, and users don't understand internal types of Lotus.

```suggestion
	return fmt.Errorf("cannot interpret %s as a hex-encoded uint64, or a number", string(b))
```

raulk (commented):
Could you please add a test for `eth_feeHistory` here? We should be testing all styles of arguments.

Plus points if you investigate exactly what MetaMask, ethers.js, web3.js, and other tools use and add test cases for those, since we know this method is a bit of a snowflake.

node/impl/full/eth.go
```diff
@@ -633,19 +637,23 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, blkCount ethtypes.EthUint
...
+	ret := ethtypes.EthFeeHistory{
 		OldestBlock:   ethtypes.EthUint64(oldestBlkHeight),
 		BaseFeePerGas: baseFeeArray,
 		GasUsedRatio:  gasUsedRatioArray,
-	}, nil
+	}
+	if params.RewardPercentiles != nil {
+		reward := make([][]ethtypes.EthBigInt, 0)
+		ret.Reward = &reward
+	}
```
magik6k, line 655: Is this a todo, or is there actually a valid reason why we always populate ret.Reward with an empty array?

magik6k (approved):
Looks good, just one question + conflicts need resolving

node/impl/full/eth.go
pl-kevin: This is a todo that we should probably implement in the future.

ychiaoli18: This is a todo that we should probably implement in the future.

raulk: So this should return:

> An array of effective priority fee per gas data points from a single block.

We need to calculate the requested percentiles of _effective_ gas premium based on the newest block (I presume it's the newest, need to dig in). _Effective_ means we're clamped at the gas_fee_cap - base_fee.

raulk: https://github.com/filecoin-project/lotus/issues/10236

magik6k (approved):

Issue #9839, opened 2022-12-07
Transaction Hash Mismatch when deploying contract using Open Zeppelin Upgrades


scotthconner:
Edited by @jennijuju 

To mitigate this issue, we will expand the existing scope of the sqlite DB that is introduced for events indexing to cover the etherum tx id <> filecoin message CID mapping. We should introduce a configurable variable for node operators to choose if they wanna store this mapping or not, it should be disabled by default.

For the mapping, when we get an ethereum transaction, we reformat it into a ["filecoin" message and get a filecoin CID](https://github.com/filecoin-project/lotus/blob/f681a350fcc9c39a3022a6708fa5da777dd83efa/node/impl/full/eth.go#L516).  We need to keep that tx id <> Filecoin CID mapping.

---
I am trying to deploy an OpenZeppelin ERC1155 contract to FEVM wallaby using the UUPSUpgradeable proxy pattern, and Hardhat upgrades plugin. The code looks as follows:

```js
const deployment = await upgrades.deployProxy(contract, preparedArguments);
        await deployment.deployed();`
```

I've had to override the provider to manually set the gas prices to prevent a legacy transaction error:

```js
if (chainId.toString() === '31415') {
      // Wrap the provider so we can override fee data.
      const provider = new ethers.providers.FallbackProvider([ethers.provider], 1);
      provider.getFeeData = async () => FEE_DATA;
      // override the signer, connected to the provider with hardcoded fee data
      owner = new ethers.Wallet(process.env.MY_PRIVATE_KEY, provider);
    }
```

However, upon deployment I get an error around mismatch expected transaction hash. The hash that comes back is different than the one expected. I'm not certain if this is a post-deployment issue, and the contract is fine, or if this prevented the deployment altogether. It's worth noting that I am using a private key from ethereum, and not certain if that would cause a hash difference. In either case, this is currently blocking FEVM users from using deployable contracts like they would on ethereum, as far as I can tell,

```bash
$ npx hardhat deploy --network wallaby --contract KeyVault

==== GENIE, DEPLOY! ====

{
  "contract": "KeyVault",
  "force": false,
  "upgrade": false
}

=== SIGNER INFO ===

 Signer Network Chain ID: 31415
 Signer Wallet Address: 0x4E5d95F1D3d1b1FB4a169554A6bff1fD164ACa2c
 Signer Balance: 4998.9

=== CONTRACT INFO ===

 Input alias: KeyVault
 Factory Signer Chain ID: 31415
[...]

PR #9965, opened 2023-01-04
feat: Store mapping of eth transaction hashes to message cids

geoff-vball:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/9839

## Proposed Changes
Adds a database to keep a mapping of Eth Transaction Hash to Filecoin Message CID. Originally we were just using reversible transformation of the Message CID for the hash, but Eth tooling expects a specific method of generation, so we have changed it to match. This is only true for Ethereum messages, all other messages will use the old reversible transformation.

I've created a separate database from the events database just for ease of use. If we really want them to be in the same database, we should refactor the database to be more globally accessible. 

@arajasek is worried that saving this mapping for all mpool messages is a possible attack vector to spam the node.

I do want to add options for being able to clear out old entries to the database. I'm welcome to suggestions on how this might be implemented.

## Checklist

- [ ] CI is green


chain/types/ethtypes/eth_types.go
```diff
@@ -392,10 +392,21 @@ func ParseEthHash(s string) (EthHash, error) {
...
+}
+
 func (h EthHash) String() string {
 	return "0x" + hex.EncodeToString(h[:])
 }
 
+// Should ONLY be used for blocks and Filecoin messages. Eth transactions expect a different hashing scheme.
 func (h EthHash) ToCid() cid.Cid {
 	// err is always nil
 	mh, _ := multihash.EncodeName(h[:], "blake2b-256")
```
magik6k, line 412: Could we somehow type-alias EthHash to at least have some reasonable indication of what hash function we're expecting it to carry?

documentation/en/default-lotus-config.toml
```diff
@@ -342,3 +342,11 @@
   #ActorEventDatabasePath = ""
 
 
+[EthTxHashConfig]
```
magik6k, line 345: * We're in the config, calling things Config here is a bit redundant
* Group name ideally shouldn't be that narrow.
* I'd call this group `FEVM`/`EVM` or something broad like that given that we're passing this config into evm module.

node/modules/ethmodule.go
```diff
@@ -0,0 +1,84 @@
...
+			// mapping functionality disabled. Nothing to do here
+			return em, nil
+		}
+
+		dbPath, err := r.SqlitePath()
+		if err != nil {
+			return nil, err
+		}
+
+		transactionHashLookup, err := chain.NewTransactionHashLookup(filepath.Join(dbPath + "txHash.db"))
```
magik6k, line 41: I'd keep the filename all-lowercase.. because filesystems.
```suggestion
		transactionHashLookup, err := chain.NewTransactionHashLookup(filepath.Join(dbPath + "txhash.db"))
```

node/config/def.go
```diff
@@ -107,6 +107,9 @@ func DefaultFullNode() *FullNode {
 			MaxFilterResults:        10000,
 			MaxFilterHeightRange:    2880, // conservative limit of one day
 		},
+		EthTxHashConfig: EthTxHashConfig{
+			EnableEthHashToFilecoinCidMapping: true,
```
magik6k, line 111: I'd prefer this be disabled by default.

Initially majority of nodes won't care about eth transactions, and this is another thing in the node repo which right now will grow in an unbounded way.

chain/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,162 @@
+package chain
```
magik6k, line 1: This belongs in its own package (or at the very least doesn't belong in the root chain package)

chain/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,162 @@
...
+	"PRAGMA journal_mode = WAL",
+	"PRAGMA read_uncommitted = ON",
+}
+
+var ddls = []string{
+	`CREATE TABLE IF NOT EXISTS tx_hash_lookup (
+		hash TEXT PRIMARY KEY,
+		cid TEXT NOT NULL,
+		epoch INT NOT NULL
+	)`,
```
magik6k, line 32: Don't we want some indexes here?
* On hash for lockups by hash
* On CID by lockups by cid
* Sorted index on epoch for efficient range delete

Something like
```sql
create table if not exists tx_hash_lookup
(
    hash  text    not null
        primary key
        unique,
    cid   text    not null
        unique,
    epoch integer not null
);

create unique index if not exists tx_hash_lookup_cid_uindex
    on tx_hash_lookup (cid);

create index if not exists tx_hash_lookup_epoch_index
    on tx_hash_lookup (epoch);

create unique index if not exists tx_hash_lookup_hash_uindex
    on tx_hash_lookup (hash);
```

node/config/def.go
geoff-vball: Yeah, I'll change this. Just had it on by default for ease of testing.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+	hashEntry, err := ei.db.Prepare(insertTxHash)
+	if err != nil {
+		return xerrors.Errorf("prepare insert event: %w", err)
+	}
+
+	_, err = hashEntry.Exec(txHash.String(), c.String(), epoch)
+	return err
+}
+
+func (ei *TransactionHashLookup) LookupCidFromTxHash(txHash ethtypes.EthHash) (cid.Cid, error) {
```
raulk, line 72: You can drop the parameter type from the name, as it's implied by the signature.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+		return cid.Undef, xerrors.Errorf("transaction hash %s not found", txHash.String())
+	}
+	err = q.Scan(&c)
+	if err != nil {
+		return cid.Undef, err
+	}
+	return cid.Decode(c)
+}
+
+func (ei *TransactionHashLookup) LookupTxHashFromCid(c cid.Cid) (ethtypes.EthHash, error) {
```
raulk, line 89: You can drop the parameter type from the name, as it's implied by the signature.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+func (ei *TransactionHashLookup) LookupCidFromTxHash(txHash ethtypes.EthHash) (cid.Cid, error) {
+	q, err := ei.db.Query("SELECT cid FROM tx_hash_lookup WHERE hash = :hash;", sql.Named("hash", txHash.String()))
+	if err != nil {
+		return cid.Undef, err
+	}
+
+	var c string
+	if !q.Next() {
+		return cid.Undef, xerrors.Errorf("transaction hash %s not found", txHash.String())
+	}
```
raulk, line 81: We should either use a sentinel error here `ErrNotFound` or the ok out parameter pattern. Otherwise it's not easy to distinguish "not found" from an internal error (which might be important for the case where an EthTx does not have a corresponding MessageCID because it's the multihash-encoded EthTx -- non-Eth messages).

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return xerrors.Errorf("prepare insert event: %w", err)
+	}
+
+	_, err = hashEntry.Exec(txHash.String(), c.String(), epoch)
+	return err
+}
+
+func (ei *TransactionHashLookup) LookupCidFromTxHash(txHash ethtypes.EthHash) (cid.Cid, error) {
+	q, err := ei.db.Query("SELECT cid FROM tx_hash_lookup WHERE hash = :hash;", sql.Named("hash", txHash.String()))
```
raulk, line 73: You should use QueryRow, as you only ever expect a single row here (given the unique constraint). That will save some code and overhead.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	err = q.Scan(&c)
+	if err != nil {
+		return cid.Undef, err
+	}
+	return cid.Decode(c)
+}
+
+func (ei *TransactionHashLookup) LookupTxHashFromCid(c cid.Cid) (ethtypes.EthHash, error) {
+	q, err := ei.db.Query("SELECT hash FROM tx_hash_lookup WHERE cid = :cid;", sql.Named("cid", c.String()))
```
raulk, line 90: We should have all these statements as prepared statements instead of ad-hoc statements. Prepared statements perform much better.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	err = q.Scan(&hashString)
+	if err != nil {
+		return ethtypes.EmptyEthHash, err
+	}
+	return ethtypes.ParseEthHash(hashString)
+}
+
+func (ei *TransactionHashLookup) RemoveEntriesOlderThan(epoch abi.ChainEpoch) (int64, error) {
+	res, err := ei.db.Exec("DELETE FROM tx_hash_lookup WHERE epoch < :epoch;", sql.Named("epoch", epoch))
```
raulk, line 107: Prepared statement.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+
+	q, err := db.Query("SELECT name FROM sqlite_master WHERE type='table' AND name='_meta';")
+	if err == sql.ErrNoRows || !q.Next() {
+		// empty database, create the schema
+		for _, ddl := range ddls {
+			if _, err := db.Exec(ddl); err != nil {
+				_ = db.Close()
+				return nil, xerrors.Errorf("exec ddl %q: %w", ddl, err)
+			}
+		}
```
raulk, line 136: I would prefer if you assert against the number of tables you expect and their respective names. That way, if something gets corrupted (e.g. and just one table remains, e.g. if someone deleted the hash lookup table but not meta), you would be resilient enough to reconstruct the schema. You already do `CREATE IF NOT EXISTS`, so no risk of erroring if a table already exists.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+	} else if err != nil {
+		_ = db.Close()
+		return nil, xerrors.Errorf("looking for _meta table: %w", err)
+	} else {
+		// Ensure we don't open a database from a different schema version
+
```
raulk, line 142: nit: unnecessary blank line

chain/ethhashlookup/eth_transaction_hash_lookup.go
raulk: Same below.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+	"PRAGMA auto_vacuum = NONE",
+	"PRAGMA automatic_index = OFF",
+	"PRAGMA journal_mode = WAL",
+	"PRAGMA read_uncommitted = ON",
+}
+
+var ddls = []string{
+	`CREATE TABLE IF NOT EXISTS tx_hash_lookup (
+		hash TEXT PRIMARY KEY UNIQUE,
+		cid TEXT NOT NULL UNIQUE,
```
raulk, line 30: Any reason we store strings and not blobs that we can just cast away as CIDs and EthHashes instead of parsing?

documentation/en/default-lotus-config.toml
```diff
@@ -343,3 +343,11 @@
   #ActorEventDatabasePath = ""
 
 
+[Fevm]
+  # EnableEthHashToFilecoinCidMapping enables storing a mapping of eth transaction hashes to filecoin message Cids
+  #
+  # type: bool
+  # env var: LOTUS_FEVM_ENABLEETHHASHTOFILECOINCIDMAPPING
+  #EnableEthHashToFilecoinCidMapping = false
```
raulk, line 351: Do we want to add the lookback period for garbage collection here?

chain/types/ethtypes/eth_types.go
raulk: Not sure I follow this comment @magik6k. You want better type safety to prevent accidental ToCid calls?

chain/types/ethtypes/eth_types.go
```diff
@@ -556,7 +567,7 @@ type EthLog struct {
 	// The index corresponds to the sequence of messages produced by ChainGetParentMessages
 	TransactionIndex EthUint64 `json:"transactionIndex"`
 
-	// TransactionHash is the cid of the message that produced the event log.
+	// TransactionHash is the hash of the RLP message that produced the event log.
 	TransactionHash EthHash `json:"transactionHash"`
```
raulk, line 571: Eventually we might want to return events of built-in actors transactions too, but that gets a bit trickier anyway, so this is fine.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
+package ethhashlookup
```
raulk, line 1: We can probably just name this file `table.go`, no need to repeat the package name.

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+
+const (
+	insertTxHash = `INSERT INTO tx_hash_lookup
+	(hash, cid, epoch)
+	VALUES(?, ?, ?)
+	ON CONFLICT (hash) DO UPDATE SET epoch = EXCLUDED.epoch
+	WHERE epoch > EXCLUDED.epoch`
+)
+
+type TransactionHashLookup struct {
```
raulk, line 58: General go idiom is to leverage the package name to convey information, instead of duplicating it in type names, e.g. `ethhashlookup.Table` or `ethhash.LookupTable` (latter probably probably better since it defines an area and not just one mechanism).

node/config/types.go
```diff
@@ -28,6 +28,7 @@ type FullNode struct {
 	Chainstore Chainstore
 	Cluster    UserRaftConfig
 	ActorEvent ActorEventConfig
+	Fevm       FevmConfig
```
raulk, line 31: Can we call this FilecoinEVM? `FEVM` is a cute pet name we've been using, but the clearer the better in config files.

node/impl/full/eth.go
```diff
@@ -254,10 +256,21 @@ func (a *EthModule) EthGetTransactionByHash(ctx context.Context, txHash *ethtype
...
 	}
 
-	cid := txHash.ToCid()
+	c := cid.Undef
+	if a.EthTxHashManager != nil {
+		var err error
+		c, err = a.EthTxHashManager.TransactionHashLookup.LookupCidFromTxHash(*txHash)
+		if err != nil {
+			log.Debug("could not find transaction hash %s in lookup table", txHash.String())
+		}
```
raulk, line 265: This dismisses internal errors and falls back to taking the CID, which is definitely a no-no, since this would could return two different values for the same transaction depending on the internal state of the node.

node/impl/full/eth.go
```diff
@@ -305,19 +323,30 @@ func (a *EthModule) EthGetTransactionCount(ctx context.Context, sender ethtypes.
...
 
 func (a *EthModule) EthGetTransactionReceipt(ctx context.Context, txHash ethtypes.EthHash) (*api.EthTxReceipt, error) {
-	cid := txHash.ToCid()
+	c := cid.Undef
+	if a.EthTxHashManager != nil {
+		var err error
+		c, err = a.EthTxHashManager.TransactionHashLookup.LookupCidFromTxHash(txHash)
+		if err != nil {
+			log.Debug("could not find transaction hash %s in lookup table", txHash.String())
+		}
```
raulk, line 332: Same as above.

node/impl/full/eth.go
```diff
@@ -286,6 +299,11 @@ func (a *EthModule) EthGetTransactionByHash(ctx context.Context, txHash *ethtype
 	return nil, nil
 }
 
+func (a *EthModule) EthGetTransactionHashByCid(ctx context.Context, cid cid.Cid) (*ethtypes.EthHash, error) {
+	hash, err := EthTxHashFromFilecoinMessageCid(ctx, cid, a.StateAPI)
+	return &hash, err
```
raulk, line 304: Right now a miss would translate to an error. We should return nil. But we can fix that once we actually identify "not found" errors.

node/impl/full/eth.go
```diff
@@ -305,19 +323,30 @@ func (a *EthModule) EthGetTransactionCount(ctx context.Context, sender ethtypes.
...
+	}
+	// This isn't an eth transaction we have the mapping for, so let's look it up as a filecoin message
+	if c == cid.Undef {
+		c = txHash.ToCid()
+	}
 
-	msgLookup, err := a.StateAPI.StateSearchMsg(ctx, types.EmptyTSK, cid, api.LookbackNoLimit, true)
-	if err != nil {
+	msgLookup, err := a.StateAPI.StateSearchMsg(ctx, types.EmptyTSK, c, api.LookbackNoLimit, true)
+	if err != nil || msgLookup == nil {
```
raulk, line 340: I believe that a nil MsgLookup is handled inside newEthTxFromFilecoinMessageLookup. We should clean that up if we don't want it there, and are instead going to assume a non-nil input in that function.

node/impl/full/eth.go
```diff
@@ -305,19 +323,30 @@ func (a *EthModule) EthGetTransactionCount(ctx context.Context, sender ethtypes.
...
 		return nil, nil
 	}
 
 	tx, err := newEthTxFromFilecoinMessageLookup(ctx, msgLookup, -1, a.Chain, a.StateAPI)
-	if err != nil {
+	if err != nil && tx.Hash == ethtypes.EmptyEthHash {
```
raulk, line 345: Why this?

node/impl/full/eth.go
```diff
@@ -640,11 +669,12 @@ func (a *EthModule) EthSendRawTransaction(ctx context.Context, rawTx ethtypes.Et
 		smsg.Message.Method = builtinactors.MethodSend
 	}
 
-	cid, err := a.MpoolAPI.MpoolPush(ctx, smsg)
+	_, err = a.MpoolAPI.MpoolPush(ctx, smsg)
```
raulk, line 672: There's a race condition here. Let's add the entry to the lookup table here synchronously, since the client could ask for its transaction through another JSON-RPC call before the mpool handler has processed the entry.

BTW -- what are we doing to clean up the mapping table from transactions that never get included?

node/impl/full/eth.go
```diff
@@ -1398,18 +1427,15 @@ func newEthBlockFromFilecoinTipSet(ctx context.Context, ts *types.TipSet, fullTx
...
-				return ethtypes.EthBlock{}, nil
-			}
 			block.Transactions = append(block.Transactions, tx)
 		} else {
-			hash, err := ethtypes.EthHashFromCid(msg.Cid())
-			if err != nil {
-				return ethtypes.EthBlock{}, err
-			}
-			block.Transactions = append(block.Transactions, hash.String())
+			block.Transactions = append(block.Transactions, tx.Hash.String())
```
raulk, line 1438: If the client does not ask for full tx info, it's very wasteful to load the full tx from the chain. Let's keep the previous code design here, and simply perform a lookup in the table in the `else` block.

node/modules/ethmodule.go
```diff
@@ -0,0 +1,84 @@
...
+			// mapping functionality disabled. Nothing to do here
+			return em, nil
+		}
+
+		dbPath, err := r.SqlitePath()
+		if err != nil {
+			return nil, err
+		}
+
+		transactionHashLookup, err := ethhashlookup.NewTransactionHashLookup(filepath.Join(dbPath, "txhash.db"))
```
raulk, line 41:
```suggestion
		transactionHashLookup, err := ethhashlookup.NewTransactionHashLookup(filepath.Join(dbPath, "eth_txhash.db"))
```

node/repo/interface.go
```diff
@@ -69,6 +69,9 @@ type LockedRepo interface {
 	// SplitstorePath returns the path for the SplitStore
 	SplitstorePath() (string, error)
 
+	// SqlitePath returns the path for the Sqlite database
+	SqlitePath() (string, error)
```
raulk, line 73: I think this is a bit ambiguous, since there's already a sqlite for events. Either we add that path here too, or we qualify this path to say that it's for the EthTxHash mapping.

node/impl/full/eth.go
```diff
@@ -1703,6 +1762,68 @@ func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLook
...
+}
+
+type EthTxHashManager struct {
+	StateAPI              StateAPI
+	TransactionHashLookup *ethhashlookup.TransactionHashLookup
+}
+
+func (m EthTxHashManager) Revert(ctx context.Context, from, to *types.TipSet) error {
+	return nil
+}
```
raulk, line 1799: The edge case where a reorg happens and messages get included in a future epoch is handled by the ON CONFLICT clause, which is great. However, if a message is added in tipset 1A, then we reorg to tipset 1B which does not include the message, and it only lands in tipset 4A, we would've spent 3 tipsets thinking that the message exists and it landed on epoch 1. Is that correct?

Similarly, there's the edge case where the chain reorgs and completely removes a message such that it never again lands on chain.

I'm not sure how important consistency in such scenarios is -- it might be acceptable to have wrong data in the index if the epoch is only used for garbage collection. But we should _never_ return data over the interface that is not truthful!

node/impl/full/eth.go
```diff
@@ -1703,6 +1762,68 @@ func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLook
...
+
+	return nil
+}
+
+type EthTxHashManager struct {
+	StateAPI              StateAPI
+	TransactionHashLookup *ethhashlookup.TransactionHashLookup
+}
+
+func (m EthTxHashManager) Revert(ctx context.Context, from, to *types.TipSet) error {
```
raulk, line 1797: You 100% want pointer receivers here.

node/impl/full/eth.go
```diff
@@ -1703,6 +1762,68 @@ func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLook
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+type EthTxHashManager struct {
```
raulk, line 1792: What about garbage collection? Do we use the epoch somewhere?

node/impl/full/eth.go
```diff
@@ -1703,6 +1762,68 @@ func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLook
 	return receipt, nil
 }
 
+func (m EthTxHashManager) Apply(ctx context.Context, from, to *types.TipSet) error {
```
raulk, line 1765: You 100% want pointer receivers here.

raulk (changes requested):
Great work here! Feel free to ping me if you have any doubts about these comments.

chain/ethhash/eth_transaction_hash_lookup.go
```diff
@@ -0,0 +1,166 @@
...
+var ddls = []string{
+	`CREATE TABLE IF NOT EXISTS tx_hash_lookup (
+		hash TEXT PRIMARY KEY UNIQUE,
+		cid TEXT NOT NULL UNIQUE,
+		epoch INT NOT NULL
+	)`,
+
+	`CREATE UNIQUE INDEX IF NOT EXISTS tx_hash_lookup_hash_uindex ON tx_hash_lookup (hash)`,
+
+	`CREATE INDEX IF NOT EXISTS tx_hash_lookup_epoch_index ON tx_hash_lookup (epoch)`,
```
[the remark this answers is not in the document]
geoff-vball: I don't know if there are any good uses for storing that index. If we have blocks with transactions, we'll have to calculate the hashes from the messages anyway in the case that they're not in the index. If we do end up storing some mapping of `eth hash` -> `something else`, then it could be useful, but I don't think we have such a mapping right now?

node/repo/interface.go
geoff-vball: I do think we should map the events db to the same path. It gives an easy place to find the databases that we don't need to configure. 

It also allows us to keep the database inside the `LOTUS_PATH`, which I wasn't able to do with the setup from the events db (though that might have been my error with the way our dependency injection works).

node/impl/full/eth.go
geoff-vball: I think I want to use just raw timestamps instead of epochs here. The column should just be for optionally periodically clearing out the db so users can manage the size of their state. The problem with the current setup is that we don't know when to clear out mpool mappings if they're removed. We don't need the epoch information in this table, we can look it up once we get the cid. 

If we're not storing the epoch, I don't think there are any inconsistency issues.

node/impl/full/eth.go
geoff-vball: I'll add tomorrow

api/api_full.go
```diff
@@ -778,6 +778,7 @@ type FullNode interface {
 	EthGetBlockByHash(ctx context.Context, blkHash ethtypes.EthHash, fullTxInfo bool) (ethtypes.EthBlock, error)                               //perm:read
 	EthGetBlockByNumber(ctx context.Context, blkNum string, fullTxInfo bool) (ethtypes.EthBlock, error)                                        //perm:read
 	EthGetTransactionByHash(ctx context.Context, txHash *ethtypes.EthHash) (*ethtypes.EthTx, error)                                            //perm:read
+	EthGetTransactionHashByCid(ctx context.Context, cid cid.Cid) (*ethtypes.EthHash, error)                                                    //perm:read
```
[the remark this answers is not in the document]
geoff-vball: Could you explain exactly what you mean? Do you want another API in `EthAPI` on the same level as `EthModuleAPI` and `EthEventAPI`?

chain/eth_transaction_hash_lookup.go
geoff-vball: Unique colums (including primary key) get indexed automatically. I've added an index for the `insertion_time`, which has replaced the epoch for garbage collection purposes.

node/impl/full/eth.go
geoff-vball: Added

node/impl/full/eth.go
raulk: AFAIK once a transaction is successfully added in the mpool (has passed validation), I believe that four logical outcomes are possible:
1. It can land on chain, and is eventually finalized.
2. It can land on chain and then be reverted due to a reorg, "returning" to the mpool.
3. It can be replaced.
4. It can drop.

Are we handling all of these? Wanna make sure that there's no possible issue here with stale or bad data in the cache?

geoff-vball: We don't explicitly remove mappings for messages in scenarios 2-4, so we might end up with a few mappings in the database for transactions that never fully land on chain. I don't think this is an issue. I've removed the epoch which could've led to inconsistencies. There's no harm in returning a cid for a message that is no longer relevant - the caller will look up the cid and will be unable to find it in their store.

scotthconner: @geoff-vball are you saying that we have covered the proper use cases? To your point, stale or useless messages are likely harmless to leave around outside of the space it takes up. When it comes to getting a canonical version of the data, I imagine that comes from a snapshot?

Either way, being able to garbage collect that information seems orthogonal to the criticality of how re-orgs are handled.

Are you saying that we are safe against the 4 considered cases? Is there any other blocker?

geoff-vball: Yeah, there are no situations where you'd be missing something from your DB, unless it has been garbage collected by the retention policy, or you had the DB turned off when you synced.

geoff-vball: Just to summarize how it works: 
Adds mapping when:
- Tx added to mpool
- Tx processed in block

Removes mappings when:
- Checks every hour for mappings that have been in the database longer than the retention policy.

raulk (approved):
Great work here. Some loose ends we need to tie up in subsequent PRs:
1. Populating the index on snapshot import.
2. Mpool index management (see threads above).

node/impl/full/eth.go
scotthconner: What is the default retention policy? Curious.

mur-me:
Howdy! 
Glif is here.

Our RPC endpoints is still having issues with `EthGetBlockByHash` operation on the Metamask -> RPC.

I've go deeper and it looks like that something goes wrong on the lotus side, because it putting from ETH Hash into the `txhash.db`

Example:
`0xbc51be0da51511bcf2f9b7fa4915b7fee936d3cf34e4d6ac811f1aefb283fc5d|bafy2bzaceb7qsy5gvx7kygtol5sw7yypomslher73vbp2ssb6rnzrwo6rykku|2023-01-18 13:24:00` - line from db

Just look to the glif explorer hashes, e.x bafy2bzaceb7qsy5gvx7kygtol5sw7yypomslher73vbp2ssb6rnzrwo6rykku - https://explorer.glif.io/tx/bafy2bzaceb7qsy5gvx7kygtol5sw7yypomslher73vbp2ssb6rnzrwo6rykku/?network=hyperspace

You can see that there is no 0xbc51be0da51511bcf2f9b7fa4915b7fee936d3cf34e4d6ac811f1aefb283fc5d like it stored in the db, but there is 0x7f0963a6adfeac1a6e5f656fe30f7324b3923fdd42fd4a41f45b98d9de8e14aa

node/impl/full/eth.go
geoff-vball: @scotthconner Default sets the database to not garbage collect at all. If they notice their db getting too large they can change the retention policy. 

Thinking about this some more, I think the database should mirror the retention policy of messages in the blockstore. The mappings become (relatively) useless if we prune a message from our store. You'll be able to find the cid, but you won't be able to use to cid to look up the message. 

You could theoretically use that cid to look up the transaction on filfox or filscan. Do we know if they're planning on implementing lookups by Transaction Hashes?

Issue #9849, opened 2022-12-13
Eth JSON RPC API integration tests


jennijuju:
## Fetching information

- [x]  eth_getBalance @shrenujbansal https://github.com/filecoin-project/lotus/pull/9966#issue-1519451852
	- [x]  for existent f4 address
	- [x]  for inexistent f4 address (returns 0x0)
	- [x]  for existent ID masked address
	- [x]  for inexistent ID masked address (returns 0x0)
- [ ]  eth_getBlockByHash 
	- [ ]  with true/false in the second parameter (to return transactions or not) -- check it does the right thing
	- [ ]  tipset with multiple blocks, verify txs deduplicated
	- [ ]  inexistent block hash
	- [ ]  invalid block hash
- [ ]  eth_getBlockByNumber 
	- [ ]  same as above, but with inexistent block number
	- [ ]  null round
- [ ]  eth_getTransactionByBlockHashAndIndex 
    - [ ]  in the same test case as above, verify that we're able to retrieve a transaction by its index within a tipset
- [ ]  eth_getTransactionByBlockNumberAndIndex
	- [ ]  same as above
- [ ]  eth_protocolVersion returns network version

## Logs @iand  

- [x] eth_getLogs (see https://github.com/filecoin-project/lotus/pull/10052)
  - [x] Validate AND filters
  - [x] Validate OR filters
  - [x] Validate block range queries
  - [x] Validate address filter
- [ ] eth_getTransactionReceipt
  - [ ] Validate logs correctly populated
- [x] Validate logs correctness with respect to a Solidity contract ABI. Assert correct event signature, values, and data. Can use a no table ERC-20 token contact. (part of #10052)

## Submitting transactions @shrenujbansal 

**Contract invocations**
https://github.com/filecoin-project/lotus/pull/9966
- [x] Submit a correctly formed EIP-1559 transaction with valid signature => returns tx id
- [x] Submit a correctly formed EIP-1559 transaction with invalid signature => returns error
- [ ] Submit a mangled EIP-1559 transaction with valid signature => returns error
- [x] Submit a well-formed Legacy transaction with valid signature => returns error
- [x] Submit a transaction with invalid V in signature => returns error (might not be easy to produce)
	- https://github.com/filecoin-project/lotus/pull/9970/files#r1068006314

**Value transfers**

All of the above for value transfers.

**Contract deployments**

All of the above with contract creations (nil To).

## Filters @iand 

[...]

PR #9966, opened 2023-01-04
itests: Add tests for eth get balance and transaction submits

shrenujbansal:
## Related Issues

https://github.com/filecoin-project/lotus/issues/9849

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/eth_balance_test.go
+++ b/itests/eth_balance_test.go
@@ -1,0 +1,71 @@
+package itests
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/chain/types/ethtypes"
+	"github.com/filecoin-project/lotus/itests/kit"
+)
+
+func TestEthGetBalanceExistingF4address(t *testing.T) {
+	blockTime := 100 * time.Millisecond
+	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
+	ens.InterconnectAll().BeginMining(blockTime)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	_, ethAddr, deployer := client.EVM().NewAccount()
+
+	fundAmount := types.FromFil(10)
+	// send some funds to the f410 address
+	kit.SendFunds(ctx, t, client, deployer, fundAmount)
+
+	balance, err := client.EthGetBalance(ctx, ethAddr, "latest")
+	require.NoError(t, err)
+	require.Equal(t, balance, ethtypes.EthBigInt{fundAmount.Int})
+}
+
+func TestEthGetBalanceNonExistentF4address(t *testing.T) {
+	blockTime := 100 * time.Millisecond
+	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
+	ens.InterconnectAll().BeginMining(blockTime)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	_, ethAddr, _ := client.EVM().NewAccount()
+
+	balance, err := client.EthGetBalance(ctx, ethAddr, "latest")
+	require.NoError(t, err)
+	require.Equal(t, balance, ethtypes.EthBigIntZero)
+}
+
+func TestEthGetBalanceExistentIDMaskedAddr(t *testing.T) {
+	blockTime := 100 * time.Millisecond
+	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
+	ens.InterconnectAll().BeginMining(blockTime)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	faddr, err := client.WalletDefaultAddress(ctx)
+	require.NoError(t, err)
+	fid, err := client.StateLookupID(ctx, faddr, types.EmptyTSK)
+	require.NoError(t, err)
+
+	ethAddr, err := ethtypes.EthAddressFromFilecoinAddress(fid)
+	require.NoError(t, err)
+
+	balance, err := client.WalletBalance(ctx, fid)
+	require.NoError(t, err)
+
+	ebal, err := client.EthGetBalance(ctx, ethAddr, "latest")
+	require.NoError(t, err)
+	require.Equal(t, ebal, ethtypes.EthBigInt{balance.Int})
+}
--- a/itests/eth_balance_test.go
+++ b/itests/eth_balance_test.go
@@ -2,6 +2,8 @@ package itests
 
 import (
 	"context"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
 	"testing"
 	"time"
 
@@ -84,10 +86,12 @@ func TestEthGetBalanceBuiltinActor(t *testing.T) {
 	fid, err := address.NewFromString("f05")
 	require.NoError(t, err)
 
+	kit.SendFunds(ctx, t, client, fid, abi.TokenAmount{big.NewInt(10).Int})
+
 	ethAddr, err := ethtypes.EthAddressFromFilecoinAddress(fid)
 	require.NoError(t, err)
 
 	ebal, err := client.EthGetBalance(ctx, ethAddr, "latest")
 	require.NoError(t, err)
-	require.Equal(t, ethtypes.EthBigIntZero, ebal)
+	require.Equal(t, ethtypes.EthBigInt{big.NewInt(10).Int}, ebal)
 }
--- a/itests/eth_balance_test.go
+++ b/itests/eth_balance_test.go
@@ -7,6 +7,8 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/filecoin-project/go-address"
+
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
 	"github.com/filecoin-project/lotus/itests/kit"
@@ -22,7 +24,7 @@ func TestEthGetBalanceExistingF4address(t *testing.T) {
 
 	_, ethAddr, deployer := client.EVM().NewAccount()
 
-	fundAmount := types.FromFil(10)
+	fundAmount := types.FromFil(0)
 	// send some funds to the f410 address
 	kit.SendFunds(ctx, t, client, deployer, fundAmount)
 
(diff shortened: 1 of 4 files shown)


itests/eth_balance_test.go
```diff
@@ -0,0 +1,71 @@
...
+	blockTime := 100 * time.Millisecond
+	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
+	ens.InterconnectAll().BeginMining(blockTime)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	_, ethAddr, _ := client.EVM().NewAccount()
+
+	balance, err := client.EthGetBalance(ctx, ethAddr, "latest")
```
geoff-vball, line 44: Grr, strings bad. (not your fault, but feel free to add an enum if you're inspired)

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,200 @@
...
+	unsigned, err := tx.ToRlpUnsignedMsg()
+	require.NoError(t, err)
+	// Submit transaction without signing message
+	_, err = client.EVM().EthSendRawTransaction(ctx, unsigned)
+	require.Error(t, err)
+
+	client.EVM().SignTransaction(&tx, key.PrivateKey)
+
+	hash := client.EVM().SubmitTransaction(ctx, &tx)
+	fmt.Println(hash)
```
geoff-vball, line 76: Remove this if it was for debugging or make the message more clear if you want it printed out.

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+	var receipt *api.EthTxReceipt
+	for i := 0; i < 10000000000; i++ {
+		receipt, err = client.EthGetTransactionReceipt(ctx, hash)
+		fmt.Println(receipt, err)
```
geoff-vball, line 81: Same here

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,200 @@
...
+		if err != nil || receipt == nil {
+			time.Sleep(500 * time.Millisecond)
+			continue
+		}
+		break
+	}
```
geoff-vball, line 87: I think it would make more sense if this loop was just infinite if we're gonna rely on the timeout?

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,200 @@
...
+func TestLegacyTransaction(t *testing.T) {
+	blockTime := 100 * time.Millisecond
+	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
+
+	ens.InterconnectAll().BeginMining(blockTime)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	txBytes, err := hex.DecodeString("f86f830131cf8504a817c800825208942cf1e5a8250ded8835694ebeb90cfa0237fcb9b1882ec4a5251d1100008026a0f5f8d2244d619e211eeb634acd1bea0762b7b4c97bba9f01287c82bfab73f911a015be7982898aa7cc6c6f27ff33e999e4119d6cd51330353474b98067ff56d930")
```
geoff-vball, line 104: Just leave a comment where you got this string from.

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// create a new Ethereum account
+	key, ethAddr, deployer := client.EVM().NewAccount()
+
+	// send some funds to the f410 address
+	kit.SendFunds(ctx, t, client, deployer, types.FromFil(10))
+
+	// verify balances.
+	//bal := client.EVM().AssertAddressBalanceConsistent(ctx, deployer)
+	//require.Equal(t, types.FromFil(10), bal)
```
geoff-vball, line 135: uncomment or remove

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,200 @@
...
+	var receipt *api.EthTxReceipt
+	for i := 0; i < 10000000000; i++ {
+		receipt, err = client.EthGetTransactionReceipt(ctx, hash)
+		fmt.Println(receipt, err)
+		if err != nil || receipt == nil {
+			time.Sleep(500 * time.Millisecond)
+			continue
+		}
+		break
+	}
```
geoff-vball, line 183: same as above

itests/eth_balance_test.go
shrenujbansal: This usage is fine. Its how the ethereum API is defined
https://docs.alchemy.com/reference/eth-getbalance

itests/eth_transactions_test.go
shrenujbansal: This was copy pasta from another test. There is no timeout in this. I could change this to have a timeout instead but seems unnecessary for a test

itests/eth_balance_test.go
arajasek: Yeah, this is unfortunately literally the Eth API spec.

itests/eth_balance_test.go
```diff
@@ -0,0 +1,71 @@
...
+	blockTime := 100 * time.Millisecond
+	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
+	ens.InterconnectAll().BeginMining(blockTime)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	_, ethAddr, deployer := client.EVM().NewAccount()
+
+	fundAmount := types.FromFil(10)
```
arajasek, line 25: I believe it is valid to send 0 FIL to an account, which should create it (with 0 balance). That might be an interesting quick case to add to this test.

itests/eth_balance_test.go
```diff
@@ -0,0 +1,71 @@
...
+	defer cancel()
+
+	_, ethAddr, _ := client.EVM().NewAccount()
+
+	balance, err := client.EthGetBalance(ctx, ethAddr, "latest")
+	require.NoError(t, err)
+	require.Equal(t, balance, ethtypes.EthBigIntZero)
+}
+
+func TestEthGetBalanceExistentIDMaskedAddr(t *testing.T) {
```
arajasek, line 49: Might be interesting to add a second case to this test that gets the balance of a non-account (eg. a miner, or even a singleton actor like the Power / Market / Reward actor). My understanding is that that should work.

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,299 @@
...
+		MaxPriorityFeePerGas: big.Int(maxPriorityFeePerGas),
+		GasLimit:             int(gaslimit),
+		V:                    big.Zero(),
+		R:                    big.Zero(),
+		S:                    big.Zero(),
+	}
+
+	unsigned, err := tx.ToRlpUnsignedMsg()
+	require.NoError(t, err)
+	// Submit transaction without signing message
```
arajasek, line 69: I would consider:

- Signing the tx first
- Flip a bit in the siggy, send that raw tx, assert error
- Unflip the bit, send that raw tx, assert good

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	// This is a legacy style transaction obtained from etherscan
+	// Tx details: https://etherscan.io/getRawTx?tx=0x0763262208d89efeeb50c8bb05b50c537903fe9d7bdef3b223fd1f5f69f69b32
+	txBytes, err := hex.DecodeString("f86f830131cf8504a817c800825208942cf1e5a8250ded8835694ebeb90cfa0237fcb9b1882ec4a5251d1100008026a0f5f8d2244d619e211eeb634acd1bea0762b7b4c97bba9f01287c82bfab73f911a015be7982898aa7cc6c6f27ff33e999e4119d6cd51330353474b98067ff56d930")
+	require.NoError(t, err)
+	_, err = client.EVM().EthSendRawTransaction(ctx, txBytes)
+	require.Errorf(t, err, "legacy transaction is not supported")
```
arajasek, line 99: Can we do a `require.ErrorContains`, with something about the error string here?

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,299 @@
...
+		GasLimit:             int(gaslimit),
+		Input:                contract,
+		V:                    big.Zero(),
+		R:                    big.Zero(),
+		S:                    big.Zero(),
+	}
+
+	unsigned, err := tx.ToRlpUnsignedMsg()
+	require.NoError(t, err)
+	// Submit transaction without signing message
```
arajasek, line 153: Similar suggestion as above.

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,299 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, receipt)
+	require.EqualValues(t, ethtypes.EthUint64(0x1), receipt.Status)
+
+	// Get contract address.
+	contractAddr := client.EVM().ComputeContractAddress(ethAddr, 0)
+
+	// INVOKE CONTRACT
+
+	params, err := hex.DecodeString("f8b2cb4f000000000000000000000000ff00000000000000000000000000000000000064")
```
arajasek, line 241: Comment describing what's happening here (which method's being invoked, what the params are, and maybe a link to the Solidity ABI)

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,299 @@
...
+		GasLimit:             int(gaslimit),
+		Input:                params,
+		V:                    big.Zero(),
+		R:                    big.Zero(),
+		S:                    big.Zero(),
+	}
+
+	unsigned, err := tx.ToRlpUnsignedMsg()
+	require.NoError(t, err)
+	// Submit transaction without signing message
```
arajasek, line 270: Similar suggestion as above.

itests/eth_transactions_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+	contract, err := hex.DecodeString(string(contractHex))
+	require.NoError(t, err)
+
+	// create a new Ethereum account
+	key, ethAddr, deployer := client.EVM().NewAccount()
+	// send some funds to the f410 address
+	kit.SendFunds(ctx, t, client, deployer, types.FromFil(10))
+
+	// DEPLOY CONTRACT
```
arajasek, line 201: I was gonna suggest extracting this into a helper, but it looks like there already is one -- `client.EVM().DeployContract()` should be able to do this entire step (and might be able to replace other examples of this boilerplate in this PR).

shrenujbansal: This is what I was trying to use earlier but it doesn't work because this function signs the transaction with the client keys but we need to sign the transaction with the eth account keys in this case
As we spoke, a client which is not an eth account will not be able to send eth txs today
I can extract this into a different helper

itests/eth_balance_test.go
shrenujbansal: done

itests/eth_transactions_test.go
shrenujbansal: done

itests/eth_transactions_test.go
shrenujbansal: done

itests/eth_transactions_test.go
shrenujbansal: done

itests/eth_transactions_test.go
shrenujbansal: done

itests/eth_transactions_test.go
shrenujbansal: done

itests/eth_balance_test.go
shrenujbansal: done

itests/eth_balance_test.go
```diff
@@ -69,3 +71,23 @@ func TestEthGetBalanceExistentIDMaskedAddr(t *testing.T) {
...
+	// Address for market actor
+	fid, err := address.NewFromString("f05")
+	require.NoError(t, err)
+
+	ethAddr, err := ethtypes.EthAddressFromFilecoinAddress(fid)
+	require.NoError(t, err)
+
+	ebal, err := client.EthGetBalance(ctx, ethAddr, "latest")
+	require.NoError(t, err)
+	require.Equal(t, ethtypes.EthBigIntZero, ebal)
```
arajasek, line 92: Hmm, I'd rather this be non-zero -- can we fund the market actor (just send it 10 FIL), and assert that's what the call returns?

arajasek (approved):
One nit

itests/eth_balance_test.go
arajasek: Nope!

PR #9970, opened 2023-01-06
Eth JSON-RPC: use ToFilecoinAddress() to get f4 addr and validate v sig + cleanups

ychiaoli18:
## Related Issues
Closes https://github.com/filecoin-project/ref-fvm/issues/1030.
Closes https://github.com/filecoin-project/ref-fvm/issues/1335.

## Checklist

- [ ] CI is green


chain/types/ethtypes/eth_types.go
```diff
@@ -228,6 +228,57 @@ func (n *EthNonce) UnmarshalJSON(b []byte) error {
 
 type EthAddress [EthAddressLength]byte
 
+// NewEthAddressFromPubKey returns the Ethereum address corresponding to an
+// uncompressed secp256k1 public key.
+func NewEthAddressFromPubKey(pubk []byte) ([]byte, error) {
+	// if we get an uncompressed public key (that's what we get from the library,
+	// but putting this check here for defensiveness), strip the prefix
+	if pubk[0] != 0x04 {
```
Stebalien, line 236: empty check?

chain/types/ethtypes/eth_types.go
```diff
@@ -228,6 +228,57 @@ func (n *EthNonce) UnmarshalJSON(b []byte) error {
...
+		return nil, fmt.Errorf("expected first byte of secp256k1 to be 0x04 (uncompressed)")
+	}
+	pubk = pubk[1:]
```
Stebalien, line 239: length check? Can we check the expected length of the public key?

chain/types/ethtypes/eth_transactions.go
```diff
@@ -446,6 +450,13 @@ func parseEip1559Tx(data []byte) (*EthTxArgs, error) {
 		return nil, err
 	}
 
+	// EIP-1559 and EIP-2930 transactions only support 0 or 1 for v
+	// Legacy and EIP-155 transactions support other values
+	// https://github.com/ethers-io/ethers.js/blob/56fabe987bb8c1e4891fdf1e5d3fe8a4c0471751/packages/transactions/src.ts/index.ts#L333
+	if !v.Equals(big.NewInt(0)) && !v.Equals(big.NewInt(1)) {
+		return nil, fmt.Errorf("EIP-1559 transactions only support 0 or 1 for v")
+	}
```
raulk, line 458: Would like to add an integration test. Recorded in #9849.

chain/types/ethtypes/eth_types.go
```diff
@@ -228,6 +228,61 @@ func (n *EthNonce) UnmarshalJSON(b []byte) error {
...
 type EthAddress [EthAddressLength]byte
 
+// NewEthAddressFromPubKey returns the Ethereum address corresponding to an
+// uncompressed secp256k1 public key.
+func NewEthAddressFromPubKey(pubk []byte) ([]byte, error) {
+	// if we get an uncompressed public key (that's what we get from the library,
+	// but putting this check here for defensiveness), strip the prefix
+	const pubKeyLen = 65
+	if len(pubk) != pubKeyLen {
+		return nil, fmt.Errorf("public key should have %d in length, but gets %d", pubKeyLen, len(pubk))
```
raulk, line 238: nit

```suggestion
		return nil, fmt.Errorf("public key should have %d in length, but got %d", pubKeyLen, len(pubk))
```

chain/types/ethtypes/eth_types.go
```diff
@@ -228,6 +228,61 @@ func (n *EthNonce) UnmarshalJSON(b []byte) error {
...
+
+	// Calculate the Ethereum address based on the keccak hash of the pubkey.
+	hasher := sha3.NewLegacyKeccak256()
+	hasher.Write(pubk)
+	ethAddr := hasher.Sum(nil)[12:]
+	return ethAddr, nil
+}
+
+func NewEthAddressFromFilecoinAddress(addr address.Address) (EthAddress, error) {
+	ethAddr, ok, err := TryEthAddressFromFilecoinAddress(addr, true)
```
raulk, line 254: I know this is not touched here, but the interface of this function is very tricky, and that trickiness leaks here. It only returns ok=false, err=nil in one case: when we are not allowing IDs and the supplied address is an ID address. I would simplify this to return the EthAddress and an error only, with a predefined error when the supplied address is an ID address.

If we do that, we can get rid of `NewEthAddressFromFilecoinAddress` and use `TryEthAddressFromFilecoinAddress` directly everywhere, which is also a better name because this is really a conversion, not a construction.

lib/sigs/delegated/eth.go
```diff
@@ -1,26 +0,0 @@
...
-		return nil, fmt.Errorf("expected first byte of secp256k1 to be 0x04 (uncompressed)")
-	}
-	pubk = pubk[1:]
-
-	// Calculate the Ethereum address based on the keccak hash of the pubkey.
-	hasher := sha3.NewLegacyKeccak256()
-	hasher.Write(pubk)
-	ethAddr := hasher.Sum(nil)[12:]
-	return ethAddr, nil
-}
```
raulk, line 26: This was already covered in #9984, which landed in the branch. Will resolve the conflict.

Issue #9891, opened 2022-12-16
Lotus can seal CC sectors but is not able to start AP for sectors with deals.


donkabat:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [X] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon:  1.19.0+mainnet+git.64059ca87+api1.5.0
Local: lotus-miner version 1.19.0+mainnet+git.64059ca87
lotus-worker version 1.18.2+mainnet+git.9416ceb7d
boostd version 1.5.1-rc2+git.b5768c8
```

### Describe the Bug

I moved miner and boost and joined them on one server.
Miner's config dir copied manually between severs on version v1.18.0 and next upgraded to v1.19.0.
Boostd built and recovered from backup.

Lotus can seal CC sectors but is not able to start AP for sectors with deals.

```
[...]

PR #9979, opened 2023-01-10
fix: sealing: disregard MaxWaitDealsSectors if full-size piece is waiting

jcace:
## Related Issues
https://github.com/filecoin-project/lotus/issues/9891

## Proposed Changes
`tryGetDealSector` will now check to see if there's a _full-size_ (aka Piece Size ==  Sector Size) sector waiting. If there is, it will disregard the `MaxWaitDealsSectors` config and try to create/upgrade a new sector for the pending piece.

## Additional Info
This is to avoid a situation where Lotus waits for smaller deals to fill up the sectors in `WaitDeals`, even though there are full-size Pieces that could start sealing. I think it makes sense to always create/upgrade a new sector for those, as there's no point to have a full (maximum QAP) sector be blocked by smaller ones. 

## Checklist

- [ ] CI is green

### DIFF
--- a/storage/pipeline/input.go
+++ b/storage/pipeline/input.go
@@ -797,6 +797,7 @@ func (m *Sealing) tryGetDealSector(ctx context.Context, sp abi.RegisteredSealPro
 		// Check if any pending pieces are full size
 		if piece.size == abi.PaddedPieceSize(ssize).Unpadded() {
 			fullSizePieceIsPending = true
+			break
 		}
 	}
 
--- a/storage/pipeline/input.go
+++ b/storage/pipeline/input.go
@@ -578,7 +578,7 @@ func (m *Sealing) updateInput(ctx context.Context, sp abi.RegisteredSealProof) e
 
 	if len(toAssign) > 0 {
 		log.Errorf("we are trying to create a new sector with open sectors %v", m.openSectors)
-		if err := m.tryGetDealSector(ctx, sp, getExpirationCached); err != nil {
+		if err := m.tryGetDealSector(ctx, sp, ssize, getExpirationCached); err != nil {
 			log.Errorw("Failed to create a new sector for deals", "error", err)
 		}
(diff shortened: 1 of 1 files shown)


storage/pipeline/input.go
```diff
@@ -792,8 +792,16 @@ func (m *Sealing) tryGetDealSector(ctx context.Context, sp abi.RegisteredSealPro
 		return xerrors.Errorf("getting storage config: %w", err)
 	}
 
-	// if we're above WaitDeals limit, we don't want to add more staging sectors
-	if cfg.MaxWaitDealsSectors > 0 && m.stats.curStaging() >= cfg.MaxWaitDealsSectors {
+	fullSizePieceIsPending := false
+	for _, piece := range m.pendingPieces {
+		// Check if any pending pieces are full size
+		if piece.size == abi.PaddedPieceSize(ssize).Unpadded() {
```
jcace, line 798: I'm a little fuzzy on the math here. If we have a `32GiB` sector pending and the sector size is `32GiB`, will be padding work out and this will be true?

magik6k: Yep, this is correct

magik6k (commented):
This has a bug where a failing AddPiece may generate unwanted sectors in WaitDeals state - this is also the case today, but you won't get more than MaxWaitDealsSectors number of sectors in that state today - with this patch you may see many more sectors enter WaitDeals in case of failed AddPiece operations.

An alternative to this patch could be being more aggressive in `maybeStartSealing` on deciding that the sector should go to sealing when we see that it is "full enough" and blocking other deals from being packed into fresh sectors.

rjan90:

rjan90:
> Lotus maintainers need more of your input before merging it, please address the suggested changes or reply to the comments or this PR will be closed in 72 hours.

Closing since 72hours has passed

PR #9998, opened 2023-01-13
NV18: Filecoin EVM runtime + Actor Events + EthAccount + EAM + f4 addressing

raulk:
Final landable version of https://github.com/filecoin-project/lotus/pull/9617 with review comments addressed.

---

## Related Issues

This mega-PR is the result of several months of development on the FVM Milestone 2.1 (Filecoin EVM compatibility). It includes the Lotus-side implementation of the following FIPs:

- [FIP-0048](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0048.md) (f4 address class)
- [FIP-0049](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0049.md) (Actor Events)
- [FIP-0054](https://github.com/filecoin-project/FIPs/pull/569) (Filecoin EVM Runtime)
- [FIP-0055](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0055.md) (Supporting Ethereum Accounts, Addresses, and Transactions)

## Proposed Changes

This 14,000 LoC PR has a ton of changes, but they broadly fall into these headlines:

1. The f4 address class is introduced with support for delegated addresses.
2. Uses a new bundle that includes the brand new EVM, EAM, and EthAccount actors.
4. An implementation of the Ethereum JSON-RPC API is included, mapping over to Filecoin specifics.
5. Added support for Actor Events (although needs an update to align with the final draft of FIP-0049 (tracked in https://github.com/filecoin-project/FIPs/pull/581). This includes event processing, indexing, and subscription mechanisms.
6. Adds support for the Wallaby testnet, although the intention is to move such support to a separate branch.
7. The migration for nv18 is included.
8. Genesis changes to instantiate the Eth Null Address at 0x0 and the EAM are also included.
9. An `eth` CLI has been implemented for diagnostics and quick experimentation.
10. TipsetCID generation and persistence is also implemented as per FIP-0054.
11. FVM v3 is integrated via filecoin-ffi.

## Additional Info

- There are some TODOs left over from the original review, marked with TODO::FVM. Fixing them is tracked in https://github.com/filecoin-project/ref-fvm/issues/1446.
- Should probably merge by squashing and making sure that the collective authorship is retained (I can take care of this once approved).

## Checklist

	- In progress
- [ ] CI is green


chain/consensus/filcns/upgrades.go
```diff
@@ -1580,11 +1592,110 @@ func upgradeActorsV9Common(
...
+}
+
+func PreUpgradeActorsV10(ctx context.Context, sm *stmgr.StateManager, cache stmgr.MigrationCache, root cid.Cid, epoch abi.ChainEpoch, ts *types.TipSet) error {
+	// Use half the CPUs for pre-migration, but leave at least 3.
+	workerCount := MigrationMaxWorkerCount
+	if workerCount <= 4 {
+		workerCount = 1
+	} else {
+		workerCount /= 2
+	}
```
magik6k, line 1623: Not too sure about the math here, doesn't look like we always have at least 3 workers

chain/events/filter/event.go
```diff
@@ -0,0 +1,474 @@
...
+	return true
+}
+
+func (f *EventFilter) matchAddress(o address.Address) bool {
+	if len(f.addresses) == 0 {
+		return true
+	}
+
+	// Assume short lists of addresses
+	// TODO: binary search for longer lists or restrict list length
```
magik6k, line 182: Really easy with https://pkg.go.dev/sort#Search

chain/events/filter/index.go
```diff
@@ -0,0 +1,434 @@
...
+	"PRAGMA temp_store = memory",
+	"PRAGMA mmap_size = 30000000000",
+	"PRAGMA page_size = 32768",
+	"PRAGMA auto_vacuum = NONE",
+	"PRAGMA automatic_index = OFF",
+	"PRAGMA journal_mode = WAL",
+	"PRAGMA read_uncommitted = ON",
+}
+
+var ddls = []string{
```
magik6k, line 34: Don't we want to add at least some indexes here?

chain/messagepool/messagepool.go
```diff
@@ -853,18 +852,32 @@ func (mp *MessagePool) addTs(ctx context.Context, m *types.SignedMessage, curTs
...
+	if err != nil {
+		return false, xerrors.Errorf("failed to get sender actor: %w", err)
+	}
+
+	// This message can only be included in the _next_ epoch and beyond, hence the +1.
+	epoch := curTs.Height() + 1
+	nv := mp.api.StateNetworkVersion(ctx, epoch)
+
+	// TODO: I'm not thrilled about depending on filcns here, but I prefer this to duplicating logic
+	if !filcns.IsValidForSending(nv, senderAct) {
```
magik6k, line 865: Pretty sure one of the recently merged IPC PRs has moved this to a better place

chain/messagepool/selection.go
```diff
@@ -97,7 +97,7 @@ func (sm *selectedMessages) tryToAdd(mc *msgChain) bool {
 		sm.msgs = append(sm.msgs, mc.msgs...)
 		sm.blsLimit -= l
 		sm.gasLimit -= mc.gasLimit
-	} else if mc.sigType == crypto.SigTypeSecp256k1 {
+	} else if mc.sigType == crypto.SigTypeSecp256k1 || mc.sigType == crypto.SigTypeDelegated {
```
magik6k, line 100: Not related to this particular line, but I feel like after the upgrade we should rename places which handle both delegated and secp messages to `signed` or something like that (including the field in blockheader, which we can rename because it's tuple-serialized).

chain/signatures.go
```diff
@@ -0,0 +1,53 @@
+package chain
```
magik6k, line 1: Not a fan of putting things in the chain package, can easily generate dependency loops

chain/stmgr/call.go
```diff
@@ -217,6 +218,24 @@ func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, pr
...
+			msgApply = &types.SignedMessage{
+				Message: *msg,
+				Signature: crypto.Signature{
+					Type: crypto.SigTypeDelegated,
+					Data: make([]byte, 65),
+				},
+			}
+		default:
+			// XXX: Hack to make sending from f099 (and others) "just work".
+			// REMOVE ME.
```
magik6k, line 231: REMOVE ME

chain/stmgr/stmgr.go
```diff
@@ -42,7 +42,7 @@ type StateManagerAPI interface {
 	GetPaychState(ctx context.Context, addr address.Address, ts *types.TipSet) (*types.Actor, paych.State, error)
 	LoadActorTsk(ctx context.Context, addr address.Address, tsk types.TipSetKey) (*types.Actor, error)
 	LookupID(ctx context.Context, addr address.Address, ts *types.TipSet) (address.Address, error)
-	ResolveToKeyAddress(ctx context.Context, addr address.Address, ts *types.TipSet) (address.Address, error)
+	ResolveToDeterministicAddress(ctx context.Context, addr address.Address, ts *types.TipSet) (address.Address, error)
```
magik6k, line 45: Pretty sure this rename breaks lite mode as StateManagerAPI is also supplied from `api.Gateway`

chain/events/filter/index.go
raulk: Yep, we do. @iand wanted to get a good understanding of performance before indices are added, but I am fully onboard with adding them already since it's a no-brainer.

chain/messagepool/messagepool.go
raulk: Probably on master, not on release/v1.20.0; we'll need to reconcile once we merge release/v1.20.0 into master.

chain/consensus/filcns/upgrades.go
raulk: @arajasek ?

chain/messagepool/selection.go
raulk: Yep, Signature-bearing or something rather than Signed. (BLS message are also signed, we just aggregate their signatures when we pack them into a block).

chain/signatures.go
raulk: Ack, if that's the case we can move out. This used to be in a signatures package, but @arajasek asked to move it out. We can move it somewhere else easily.

chain/stmgr/call.go
raulk: Fixed.

chain/consensus/filcns/upgrades.go
arajasek: It's at least 3 workers _if possible_ -- that is, never take more than 1, if that would doing so would leave regular processing with fewer than 3. But do take at least one always.

Issue #10013, opened 2023-01-14
Eth: newEthTxFromFilecoinMessage: unify with eth_transactions.go


arajasek:
If we can't unify it, we should drop support for Create and Create2, only Create External is actually supported anymore.

Stebalien:
Done

Stebalien:
Er, not quite. Ideally we'd change [ethCallToFilecoinMessage](https://github.com/filecoin-project/lotus/blob/2dd2edc1d3b3a14a9421985a96fcd4d77b52f99e/node/impl/full/eth.go#L777-L834) and have it:

1. Create an EthTxArgs.
2. Call `tx.ToUnsignedMessage`.

PR #10005, opened 2023-01-13
update evm actor creation for eam refactor

vyzo:
Updates for changes in https://github.com/filecoin-project/builtin-actors/pull/1032
See https://github.com/filecoin-project/ref-fvm/issues/1434

Depends:
- [x] https://github.com/filecoin-project/go-state-types/pull/149, https://github.com/filecoin-project/go-state-types/pull/150, and by extension https://github.com/filecoin-project/go-state-types/pull/147 
- [x] Actor Bundle including changes from https://github.com/filecoin-project/builtin-actors/pull/1032

PR #10017, opened 2023-01-14
Refactor: Unify EthTx to FilecoinMessage methods

arajasek:
## Related Issues

#10013 

## Proposed Changes

Unifies the common message decoding logic, making `newEthTxFromFilecoinMessage` a much simpler method.

## Checklist

- [ ] CI is green

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -1455,77 +1455,23 @@ func lookupEthAddress(ctx context.Context, addr address.Address, sa StateAPI) (e
 }
 
 func newEthTxFromFilecoinMessage(ctx context.Context, smsg *types.SignedMessage, sa StateAPI) (ethtypes.EthTx, error) {
+	tx, err := ethtypes.EthTxFromSignedMessage(smsg)
+	if err != nil {
+		return ethtypes.EthTx{}, xerrors.Errorf("failed to convert from signed message: %w", err)
+	}
+
 	fromEthAddr, err := lookupEthAddress(ctx, smsg.Message.From, sa)
 	if err != nil {
-		return ethtypes.EthTx{}, err
+		return ethtypes.EthTx{}, xerrors.Errorf("failed to lookup from ethaddress: %w", err)
 	}
 
 	toEthAddr, err := lookupEthAddress(ctx, smsg.Message.To, sa)
 	if err != nil {
-		return ethtypes.EthTx{}, err
+		return ethtypes.EthTx{}, xerrors.Errorf("failed to lookup to ethaddress: %w", err)
 	}
 
-	toAddr := &toEthAddr
-	input := smsg.Message.Params
-	// Check to see if we need to decode as contract deployment.
-	// We don't need to resolve the to address, because there's only one form (an ID).
-	if smsg.Message.To == builtintypes.EthereumAddressManagerActorAddr {
-		switch smsg.Message.Method {
-		case builtintypes.MethodsEAM.Create:
-			toAddr = nil
-			var params eam.CreateParams
-			err = params.UnmarshalCBOR(bytes.NewReader(smsg.Message.Params))
-			input = params.Initcode
-		case builtintypes.MethodsEAM.Create2:
-			toAddr = nil
-			var params eam.Create2Params
-			err = params.UnmarshalCBOR(bytes.NewReader(smsg.Message.Params))
-			input = params.Initcode
-		case builtintypes.MethodsEAM.CreateExternal:
-			toAddr = nil
-			var params abi.CborBytes
-			err = params.UnmarshalCBOR(bytes.NewReader(smsg.Message.Params))
-			input = []byte(params)
-		}
-		if err != nil {
-			return ethtypes.EthTx{}, err
-		}
-	}
-	// Otherwise, try to decode as a cbor byte array.
-	// TODO: Actually check if this is an ethereum call. This code will work for demo purposes, but is not correct.
-	if toAddr != nil {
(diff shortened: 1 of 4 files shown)


arajasek:
Test failure is known, and will be fixed in #10005 by @jennijuju by updating bundle

geoff-vball:
Hey, as-is this doesn’t work yet. (But is much much cleaner thank you thank you).

I have two tests written that you can copy into your PR and test this with (just comment out the one line that will show up as red). I ran them and they’re both broken.

 https://github.com/filecoin-project/lotus/blob/c5c19dbf1d7d1cd0a28307ee56a257e7cd6c7bc5/itests/eth_hash_lookup_test.go#L196

https://github.com/filecoin-project/lotus/blob/c5c19dbf1d7d1cd0a28307ee56a257e7cd6c7bc5/itests/eth_hash_lookup_test.go#L266

Note that BLS messages have always been broken, but secp was working before your refactor. If you want to try to wrangle these you can, otherwise I’ll rebase by PR onto yours and try to make it work, probably in the wee hours of the morning tomorrow.

node/impl/full/eth.go
```diff
@@ -1455,77 +1455,23 @@ func lookupEthAddress(ctx context.Context, addr address.Address, sa StateAPI) (e
...
 	if err != nil {
-		return ethtypes.EthTx{}, err
+		return ethtypes.EthTx{}, xerrors.Errorf("failed to convert from signed message: %w", err)
 	}
 
-	toAddr := &toEthAddr
-	input := smsg.Message.Params
-	// Check to see if we need to decode as contract deployment.
-	// We don't need to resolve the to address, because there's only one form (an ID).
-	if smsg.Message.To == builtintypes.EthereumAddressManagerActorAddr {
```
Stebalien, line 1472: So, the downside of unifying this is that we'll no longer be able to interpret "native" messages as ethereum transactions. On the other hand... that's probably not desirable anyways.

node/impl/full/eth.go
```diff
@@ -1455,77 +1455,23 @@ func lookupEthAddress(ctx context.Context, addr address.Address, sa StateAPI) (e
...
-	// Otherwise, try to decode as a cbor byte array.
-	// TODO: Actually check if this is an ethereum call. This code will work for demo purposes, but is not correct.
-	if toAddr != nil {
-		if decodedParams, err := cbg.ReadByteArray(bytes.NewReader(smsg.Message.Params), uint64(len(smsg.Message.Params))); err == nil {
-			input = decodedParams
-		}
-	}
-
-	r, s, v, err := ethtypes.RecoverSignature(smsg.Signature)
+	fromEthAddr, err := lookupEthAddress(ctx, smsg.Message.From, sa)
```
Stebalien, line 1463: Given my comment below, I don't think we need this anymore. We _should_ be able to just use the `From` field from the message as-is, as it _must_ be an f410 address for the transaction to be "valid".

node/impl/full/eth.go
```diff
@@ -1455,77 +1455,23 @@ func lookupEthAddress(ctx context.Context, addr address.Address, sa StateAPI) (e
...
 	if err != nil {
-		// we don't want to return error if the message is not an Eth tx
-		r, s, v = ethtypes.EthBigIntZero, ethtypes.EthBigIntZero, ethtypes.EthBigIntZero
+		return ethtypes.EthTx{}, xerrors.Errorf("failed to lookup from ethaddress: %w", err)
 	}
 
-	hash, err := ethtypes.EthHashFromCid(smsg.Cid())
+	toEthAddr, err := lookupEthAddress(ctx, smsg.Message.To, sa)
```
Stebalien, line 1468: If this is a valid eth transaction, `to` should either be an ID address, or an f410 address. We shouldn't need to resolve this if we're going to be limiting support to _real_ eth transactions.

Stebalien (commented):
In-line with @geoff-vball's comments, the code here used to work with native accounts and now it doesn't. However, to be honest, we might not want it to work with native accounts anyways? I'm not sure.

node/impl/full/eth.go
raulk, line 1472: Doesn't this change lose the ability to reverse engineer EVM contract creations performed from EEOAs? What's the motivation here?

raulk (changes requested):
Let's document all the difficulties in returning native messages, e.g.:

- Inability to deal with non-existent f1 and f3 recipient addresses. Not a problem for executed messages (who have acquired an ID address), but a problem for `pending` messages.
- Inability to extract BLS signatures. This is fine, just set these fields to the zero value.
- Cannot convey the method number as a dedicated field. Can instead prepend it to the input params, Solidity-style by convention (need to document that convention).
- No possibility to communicate a message version number. We haven't ever moved from 0, so don't expect that to be a problem. Any future message schemas would need to map to the Eth schema anyway.

We have two options here:
1. Solve for these difficulties, and retain the ability to return non-Eth messages.
2. Drop this functionality.

A grey area solution that returns inaccurate data or incomplete data is not safe.

Reasons we want to support it:
1. Doesn't require Eth clients to learn a new API (Filecoin API, which doesn't have js libraries as mature as Eth).
2. Doesn't cause fragmentation in the API (some APIs deal with one part of the chain, other APIs with other parts of the chain), which is consistent with our interoperability principle and avoidance of hyperspecialization.

arajasek:
We're gonna block this until https://github.com/filecoin-project/lotus/pull/9965 lands, then land into https://github.com/filecoin-project/lotus/pull/10026 at which point it'll be good for review.

arajasek:
Superseded by #10095

Issue #9977, opened 2023-01-09
FEVM - add tipset CID to block store when importing snapshot


jennijuju:
To resolve https://github.com/filecoin-project/ref-fvm/issues/1053

we can simply calculate and persist these CIDs on snapshot import. (refer to https://github.com/filecoin-project/lotus/blob/402173ea16800b93f1d51f3178838c9b2283afa5/chain/store/store.go#L960 on how)

shrenujbansal:
@jennijuju I tried looking at the ref-fvm issue but not sure I understand. Is this issue referring to the single tipset which is set as the head of the node during a snapshot? Or is it referring to all of the tipsets we download as part of the snapshot?

PR #10042, opened 2023-01-17
feat: snapshot: Store tipset key cids in chain store during snapshot import

shrenujbansal:
## Related Issues

https://github.com/filecoin-project/lotus/issues/9977

## Proposed Changes

On snapshot import, persist the tipset key cids in the chain blockstore

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/store/snapshot.go
+++ b/chain/store/snapshot.go
@@ -22,7 +22,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 )
 
-const MAX_TIPSETS_FROM_SNAPSHOT = 2000
+const TIPSETKEY_BACKFILL_RANGE = 2 * build.Finality
 
 func (cs *ChainStore) UnionStore() bstore.Blockstore {
 	return bstore.Union(cs.stateBlockstore, cs.chainBlockstore)
@@ -116,7 +116,7 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (*types.TipSet, e
 	}
 
 	ts := root
-	for i := 0; i < MAX_TIPSETS_FROM_SNAPSHOT; i++ {
+	for i := 0; i < int(TIPSETKEY_BACKFILL_RANGE); i++ {
 		err = cs.PersistTipset(ctx, ts)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


chain/store/snapshot.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 )
 
+const MAX_TIPSETS_FROM_SNAPSHOT = 2000
```
shrenujbansal, line 25: Not sure if this is the best place to put this or if we already have this constant available somewhere. Happy to take suggestions

chain/store/store_test.go
```diff
@@ -124,6 +125,51 @@ func TestChainExportImport(t *testing.T) {
...
+	}
+
+	nbs := blockstore.NewMemorySync()
+	cs := store.NewChainStore(nbs, nbs, datastore.NewMapDatastore(), filcns.Weight, nil)
+	defer cs.Close() //nolint:errcheck
+
+	root, err := cs.Import(ctx, buf)
+	require.NoError(t, err)
+
+	err = cs.SetHead(ctx, last)
```
arajasek, line 156: Nit: I would do the sethead _after_ having made the root = last assertion.

chain/store/snapshot.go
arajasek: I would define this as 2 finalities instead of hardcoding -- I think that's a reasonable description of "minimum meaningful snapshot".

And I would rename it something specific to the TSK calculation (eg. TIPSETKEY_BACKFILL_RANGE).

And if you felt inspired, you could make it controllable by an envvar, but I honestly don't think there'll be much use for that functionality (can always do later).

arajasek:
Basically looks good to land, thank you!

chain/store/snapshot.go
shrenujbansal: What do you mean by define as "2 finalities"? Are you referring to a constant for a finality defined as number of tipsets?

chain/store/store_test.go
shrenujbansal: done

chain/store/snapshot.go
shrenujbansal: done

arajasek (approved):

PR #10052, opened 2023-01-18
itests: Add tests for EthGetLogs filter combinations

iand:
## Related Issues
Part of https://github.com/filecoin-project/lotus/issues/9849

## Proposed Changes
 - Refactors EthFilter itests to simplify and remove repeated code
 - Adds a new solidity contract (EventMatrix) for testing all the filter combinations
 - Add `TestEthGetLogs `with subtests exercising filter combinations:
   - one address, two alternate addresses 
   - one topic, one topic with alternate values (OR)
   - two topics with single and alternate values  (AND then OR)
 - Add `TestEthGetLogsWithBlockRanges` with subtests testing various block ranges
   - before a block height
   - after block height
   - within block height range
   - outside range
   - latest/earliest keywords 
 - Fixes a double decoding of cbor (but https://github.com/filecoin-project/ref-fvm/issues/1345 will remove the original decoding) - currently we are consistent, but we will move the decoding boundary so that cbor encoded values become the internal format to be decoded when emitted at the api level (currently they are decoded when being ingested)
 - Fixes a parsing error of topic values from JSON that prevented OR conditions 

## Checklist

- [ ] CI is green


itests/eth_filter_test.go
```diff
@@ -463,353 +1908,145 @@ func ParseEthLog(in map[string]interface{}) (*ethtypes.EthLog, error) {
...
-	time.Sleep(blockTime * 6)
-
-	for i := 0; i < iterations; i++ {
-		// log a four topic event with data
-		ret := client.EVM().InvokeSolidity(ctx, fromAddr, idAddr, []byte{0x00, 0x00, 0x00, 0x02}, nil)
-		require.True(ret.Receipt.ExitCode.IsSuccess(), "contract execution failed")
+func decodeLogBytes(orig []byte) []byte {
+	if orig == nil {
+		return orig
 	}
```
raulk, line 1914: `len(orig) == 0` is safer.

chain/events/filter/event.go
```diff
@@ -100,9 +100,18 @@ func (f *EventFilter) CollectEvents(ctx context.Context, te *TipSetEvents, rever
...
 			}
 
+			decodedEntries := make([]types.EventEntry, len(ev.Entries))
+			for i, entry := range ev.Entries {
+				decodedEntries[i] = types.EventEntry{
+					Flags: entry.Flags,
+					Key:   entry.Key,
+					Value: decodeLogBytes(entry.Value),
+				}
+			}
```
raulk, line 110: Doesn't this go in the opposite direction than Steb pointed out in https://github.com/filecoin-project/ref-fvm/issues/1345 (1)?

raulk (approved):

PR #10063, opened 2023-01-19
feat: cli: get tipset keys by height

jennijuju:
./lotus chain gettipset --height 9316
: [bafy2bzacecwefa7e7ryue6awxppqetr566katdb7zdu75t65dlr3nutf4zqf4]

### DIFF
--- a/cli/chain.go
+++ b/cli/chain.go
@@ -63,6 +63,7 @@ var ChainCmd = &cli.Command{
 		ChainEncodeCmd,
 		ChainDisputeSetCmd,
 		ChainPruneCmd,
+		ChainGetTipSetByHeightCmd,
 	},
 }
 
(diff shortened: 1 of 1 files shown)


cli/chain.go
```diff
@@ -1504,3 +1505,38 @@ var ChainPruneCmd = &cli.Command{
 		return api.ChainPrune(ctx, opts)
 	},
 }
+
+var ChainGetTipSetByHeightCmd = &cli.Command{
+	Name:    "gettipset",
```
arajasek, line 1510:
```suggestion
	Name:    "get-tipset",
```

cli/chain.go
```diff
@@ -1504,3 +1505,38 @@ var ChainPruneCmd = &cli.Command{
...
+	Aliases: []string{"love"},
```
arajasek, line 1511: ?

arajasek, line 1511:
```suggestion
	Aliases:   []string{"getblock"},
```

cli/chain.go
```diff
@@ -1504,3 +1505,38 @@ var ChainPruneCmd = &cli.Command{
...
+	Usage:   "View a segment of the chain",
+	Flags: []cli.Flag{
+		&cli.Uint64Flag{Name: "height", DefaultText: "current head"},
```
arajasek, line 1514: I would suggest making this a mandatory parameter. We already have `lotus chain head` to print the head.

cli/chain.go
```diff
@@ -1504,3 +1505,38 @@ var ChainPruneCmd = &cli.Command{
...
+		if cctx.IsSet("height") {
+			head, err = api.ChainGetTipSetByHeight(ctx, abi.ChainEpoch(cctx.Uint64("height")), types.EmptyTSK)
+		} else {
+			head, err = api.ChainHead(ctx)
+		}
+		if err != nil {
+			return err
+		}
+
+		head, err = api.ChainGetTipSet(ctx, head.Parents())
```
arajasek, line 1535: Hmm, I'm not sure I agree with printing info about the parents here. That might match some of our Ethereum-modelling behaviour, but it is just wrong to say the tipset at this height is the parent here.

cli/chain.go
```diff
@@ -1504,3 +1505,38 @@ var ChainPruneCmd = &cli.Command{
...
+		if err != nil {
+			return err
+		}
+		fmt.Printf(": %s\n", head.Key().Cids())
```
raulk, line 1539: Can we print out the whole tipset in addition to the key?

cli/chain.go
```diff
@@ -1504,3 +1505,38 @@ var ChainPruneCmd = &cli.Command{
 		return api.ChainPrune(ctx, opts)
 	},
 }
+
+var ChainGetTipSetByHeightCmd = &cli.Command{
+	Name:    "gettipset",
+	Aliases: []string{"love"},
+	Usage:   "View a segment of the chain",
```
arajasek, line 1512:
```suggestion
	Usage:   "Get a tipset at the specified height, and print the CIDs of the block headers"
```

PR #10077, opened 2023-01-19
feat: Events API enabled by default if EnableEthRPC is true

geoff-vball:
## Proposed Changes
Sets a default path for the Events Database to be in the same directory as the tx hash database.
Enables all event API functionality if EnableEthRPC is set to true, otherwise it is always disabled.
Changes the Event Config to allow opt-out of these event APIs

## Checklist

- [ ] CI is green

### DIFF
--- a/node/config/def.go
+++ b/node/config/def.go
@@ -102,7 +102,7 @@ func DefaultFullNode() *FullNode {
 		Fevm: FevmConfig{
 			EnableEthRPC:                 false,
 			EthTxHashMappingLifetimeDays: 0,
-			Events: EventsConfig{
+			Events: Events{
 				DisableRealTimeFilterAPI: false,
 				DisableHistoricFilterAPI: false,
 				FilterTTL:                Duration(time.Hour * 24),
--- a/node/config/def.go
+++ b/node/config/def.go
@@ -99,17 +99,17 @@ func DefaultFullNode() *FullNode {
 			},
 		},
 		Cluster: *DefaultUserRaftConfig(),
-		ActorEvent: ActorEventConfig{
-			EnableRealTimeFilterAPI: false,
-			EnableHistoricFilterAPI: false,
-			FilterTTL:               Duration(time.Hour * 24),
-			MaxFilters:              100,
-			MaxFilterResults:        10000,
-			MaxFilterHeightRange:    2880, // conservative limit of one day
-		},
 		Fevm: FevmConfig{
 			EnableEthRPC:                 false,
 			EthTxHashMappingLifetimeDays: 0,
+			Events: EventsConfig{
+				DisableRealTimeFilterAPI: false,
+				DisableHistoricFilterAPI: false,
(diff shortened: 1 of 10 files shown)


node/config/types.go
```diff
@@ -660,13 +660,15 @@ type UserRaftConfig struct {
 	Tracing bool
 }
 
-type ActorEventConfig struct {
-	// EnableRealTimeFilterAPI enables APIs that can create and query filters for actor events as they are emitted.
-	EnableRealTimeFilterAPI bool
+type EventsConfig struct {
+	// EnableEthRPC enables APIs that can create and query filters for actor events as they are emitted.
+	// DisableRealTimeFilterAPI will disable those APIs if you want to opt-out
```
raulk, line 665: docs

node/modules/actorevent.go
```diff
@@ -63,9 +65,20 @@ func EthEventAPI(cfg config.ActorEventConfig) func(helpers.MetricsCtx, fx.Lifecy
 
 		// Enable indexing of actor events
 		var eventIndex *filter.EventIndex
-		if cfg.EnableHistoricFilterAPI {
+		if !cfg.DisableHistoricFilterAPI {
+			var dbPath string
+			if cfg.ActorEventDatabasePath == "" {
```
raulk, line 70: I wonder if we can simplify this to DatabasePath, or do we need to qualify?

raulk (commented):
Direction LGTM.

node/modules/actorevent.go
geoff-vball: Yep, we can do that!

travisperson (approved):
LGTM just need to update the comments on the `EventConfig` to correctly document the new Disable* values.

node/config/def.go
```diff
@@ -99,13 +99,13 @@ func DefaultFullNode() *FullNode {
...
 		},
 		Cluster: *DefaultUserRaftConfig(),
-		ActorEvent: ActorEventConfig{
-			EnableRealTimeFilterAPI: false,
-			EnableHistoricFilterAPI: false,
-			FilterTTL:               Duration(time.Hour * 24),
-			MaxFilters:              100,
-			MaxFilterResults:        10000,
-			MaxFilterHeightRange:    2880, // conservative limit of one day
+		Events: EventsConfig{
```
jennijuju, line 102: hrm @raulk  -  I thought we agreed to move this section under `FEVM` ? then the rename make sense cuz its FEVM.Events?

raulk: Yes, I believe this makes sense for now. While the Events subsystem is technically independent of FEVM, this is a user-facing configuration file and I think it leads to better product usability for node operators if all the new options they care about are nested within the Fevm group. We might need to rejigger this when builtin-actors are emitting events or with Wasm actors, but we can cross that bridge when we get to it, because I expect changes in the event subsystem anyway.

PR #10084, opened 2023-01-20
feat: minor improvements to Ethereum delegated siggys

arajasek:
## Related Issues

Minor improvements noticed while writing this logic into FIP-0055.

## Proposed Changes

- Uniformly use cborgen's `ReadByteArray` / `WriteByteArray`
- Check parameter length against expectations based on MethodNum at the end

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/types/ethtypes/eth_transactions.go
+++ b/chain/types/ethtypes/eth_transactions.go
@@ -12,13 +12,11 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	gocrypto "github.com/filecoin-project/go-crypto"
-	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
 	builtintypes "github.com/filecoin-project/go-state-types/builtin"
 	typescrypto "github.com/filecoin-project/go-state-types/crypto"
 
 	"github.com/filecoin-project/lotus/build"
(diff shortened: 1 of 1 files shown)


chain/types/ethtypes/eth_transactions.go
```diff
@@ -103,18 +101,18 @@ func EthTxArgsFromMessage(msg *types.Message) (EthTxArgs, error) {
...
-				return EthTxArgs{}, xerrors.Errorf("cannot invoke contracts with empty parameters from an eth-account")
-			}
 		}
 	}
 
 	if paramsReader.Len() != 0 {
 		return EthTxArgs{}, xerrors.Errorf("extra data found in params")
 	}
 
+	if len(params) == 0 && msg.Method != builtintypes.MethodSend {
```
arajasek, line 111: @Stebalien I _think_ we need this to cover a case where the EthTx calls the EAM's CreateExternal method with the CBOR-encoding of the empty byte array, which would fail to roundtrip.

arajasek: Feels cleaner regardless.

Stebalien: Yes, but isn't that case already covered? I.e., we should fail to read a bytes array from the params in that case. But either way works.

Stebalien (approved):

PR #10085, opened 2023-01-20
fix: event filters use cbor encoding internally

iand:
## Related Issues
Fixes https://github.com/filecoin-project/ref-fvm/issues/1345

## Proposed Changes
All cbor decoding/encoding of event log values by filters is now handled in the Eth API. Internally all values are kept as cbor encoded byte arrays.

*NOTE: this invalidates previous data stored in the event log sqlite database so events indexed before this change will not be discoverable. Advise removing the database to start indexing to a new one.*

## Additional Info
This is currently based on https://github.com/filecoin-project/lotus/pull/10083 which contains comprehensive itests for this area.

## Checklist

- [ ] CI is green

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -980,17 +980,9 @@ func (e *EthEvent) installEthFilterSpec(ctx context.Context, filterSpec *ethtype
 		addresses = append(addresses, a)
 	}
 
-	for idx, vals := range filterSpec.Topics {
-		if len(vals) == 0 {
-			continue
-		}
-		// Ethereum topics are emitted using `LOG{0..4}` opcodes resulting in topics1..4
-		key := fmt.Sprintf("topic%d", idx+1)
-		for _, v := range vals {
-			buf := make([]byte, len(v[:]))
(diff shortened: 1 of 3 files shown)


node/impl/full/eth.go
```diff
@@ -1905,3 +1891,51 @@ func leftpad32(orig []byte) []byte {
...
+}
+
+func cborEncodeTopicValue(orig []byte) ([]byte, error) {
+	var buf bytes.Buffer
+	err := cbg.WriteByteArray(&buf, trimLeadingZeros(orig))
+	if err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
```
raulk, line 1911: I'm starting to doubt whether trimming leading zeroes when emitting topic values in the EVM actor is the right choice. It clearly makes logic more complicated/cumbersome in places like this, at the benefit of a minimal gas savings for the developer. I think we should back out from this decision in the upcoming FIP-0049 alignment PRs. @Stebalien -- what do you think?

raulk (approved):

PR #10095, opened 2023-01-21
Refactor: Unify EthTx to FilecoinMessage methods v2

geoff-vball:
## Related Issues
Rebases https://github.com/filecoin-project/lotus/pull/10017 and https://github.com/filecoin-project/lotus/pull/10026 onto `release/v1.20.0`

## Proposed Changes
Mostly a refactor

When converting a non-eth transaction into an `EthTx`, it will now convert the to/from into eth addresses if possible.

You can still look up a BLS message by either the cid of the unsigned or signed (if you have it) message. We just now always use the unsigned cid for the conversion to a transaction hash. Then looking up by transaction hash, it will return the cid of the unsigned BLS message.

## Checklist

- [ ] CI is green


node/impl/full/eth.go
```diff
@@ -1550,80 +1550,34 @@ func EthTxHashFromSignedFilecoinMessage(ctx context.Context, smsg *types.SignedM
...
 
-	// This is an eth tx
-	if smsg.Signature.Type == crypto.SigTypeDelegated {
 		tx.Hash, err = tx.TxHash()
 		if err != nil {
 			return tx, err
 		}
-	} else if smsg.Signature.Type == crypto.SigTypeUnknown { // BLS Filecoin message
+	} else if smsg.Signature.Type == crypto.SigTypeUnknown { // Executed BLS Filecoin message
+		tx = ethtypes.EthTxFromNativeMessage(smsg.VMMessage())
```
arajasek, line 1574: Hmm, this means there isn't a canonical EthTx representation of BLS messages -- depending on status, the hash may or may not include the BLS signature.

1) Is this the status quo today (irrespective of this PR)?
2) Is this desirable? Would it be better to say BLS messages never include the siggy in their EthTx hash?

node/impl/full/eth.go
```diff
@@ -1741,7 +1695,8 @@ func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLook
 		LogsBloom:        ethtypes.EmptyEthBloom[:],
 	}
 
-	if receipt.To == nil && lookup.Receipt.ExitCode.IsSuccess() {
+	// V.Int will be nil for non-eth transactions, so we want to skip unmarshalling the create return.
```
arajasek, line 1698: I'm not sure I understand what's going on here -- why is V.Int relevant to the unmarshaling of the create return? Regardless of whether this was an eth tx or not, the return should be unmarshalable if the message was successful?

node/impl/full/eth.go
geoff-vball: That was the status quo, but I do think it is bad. I've changed the flow to always use the unsigned message for BLS messages to calculate it's EthTx hash.

node/impl/full/eth.go
geoff-vball: The issue was that I didn't want the unmarshal to fail and return an empty struct. I've refactored this function to do all the general stuff first, and then return early if this isn't an eth tx.

node/impl/full/eth.go
arajasek: I think that's an improvement, but we should get product insight here 

@aashidham @raulk thoughts?

geoff-vball: FYI, you can still look up a BLS message by either the cid of the unsigned or signed (if you have it) message. We just now always use the unsigned cid for the conversion to a transaction hash.  Then looking up by transaction hash, it will return the cid of the unsigned BLS message.

node/impl/full/eth.go
```diff
@@ -1633,10 +1589,32 @@ func NewEthTxFromFilecoinMessage(ctx context.Context, smsg *types.SignedMessage,
...
+		MaxFeePerGas:         ethtypes.EthBigInt(msg.GasFeeCap),
+		MaxPriorityFeePerGas: ethtypes.EthBigInt(msg.GasPremium),
+	}
+}
+
+// newEthTxFromMessageLookup creates an ethereum transaction from filecoin message lookup. If a negative txIdx is passed
+// into the function, it looks up the transaction index of the message in the tipset, otherwise it uses the txIdx passed into the
 // function
-func newEthTxFromFilecoinMessageLookup(ctx context.Context, msgLookup *api.MsgLookup, txIdx int, cs *store.ChainStore, sa StateAPI) (ethtypes.EthTx, error) {
+func newEthTxFromMessageLookup(ctx context.Context, msgLookup *api.MsgLookup, txIdx int, cs *store.ChainStore, sa StateAPI) (ethtypes.EthTx, error) {
```
raulk, line 1617: Could consider adding this as a member of the EthModule struct so we don't have to pass down the chainstore and the StateAPI. Unless it's called from elsewhere?

node/impl/full/eth.go
```diff
@@ -1633,10 +1589,32 @@ func NewEthTxFromFilecoinMessage(ctx context.Context, smsg *types.SignedMessage,
...
 	if msgLookup == nil {
 		return ethtypes.EthTx{}, fmt.Errorf("msg does not exist")
 	}
```
raulk, line 1620: I have a local patch that hosts this nil check to the callers. Nil-guards at the top of a function tends to be an anti-pattern. I'd support if you want to do it here!

node/impl/full/eth.go
```diff
@@ -1538,93 +1538,49 @@ func EthTxHashFromFilecoinMessageCid(ctx context.Context, c cid.Cid, sa StateAPI
...
-		MaxFeePerGas:         ethtypes.EthBigInt(smsg.Message.GasFeeCap),
-		MaxPriorityFeePerGas: ethtypes.EthBigInt(smsg.Message.GasPremium),
-		V:                    v,
-		R:                    r,
-		S:                    s,
-	}
+		tx.From, err = lookupEthAddress(ctx, smsg.Message.From, sa)
+		if err != nil {
+			return ethtypes.EthTx{}, xerrors.Errorf("failed to lookup from ethaddress: %w", err)
+		}
```
raulk, line 1569: Where does the To address get converted now?

chain/types/ethtypes/eth_transactions.go
```diff
@@ -58,7 +58,44 @@ type EthTxArgs struct {
...
 }
 
-func EthTxArgsFromMessage(msg *types.Message) (EthTxArgs, error) {
+// EthTxFromSignedEthMessage does NOT populate:
+// - BlockHash
+// - BlockNumber
+// - TransactionIndex
+// - From
+// - Hash
+func EthTxFromSignedEthMessage(smsg *types.SignedMessage) (EthTx, error) {
```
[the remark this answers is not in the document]
geoff-vball: We kind of need some way of differentiating between `types.Message`/`types.SignedMessage` that can be converted into a legitimate `EthTx` or not. It makes sense to me to use `Transaction`/`Tx` for the Ethereum shape, and `Message` for the Filecoin shape, but I know other people might have differing thoughts about those words.

I just want to avoid the confusion of someone thinking they can use this function for any signed message.

node/impl/full/eth.go
geoff-vball: `ethSubscription.start` needs `newEthBlockFromFilecoinTipSet` which needs `newEthTxFromMessageLookup`. So we'd need to give the `EthSubscriptionManager`/`ethSubscription` a pointer to the `EthModule` in order to do that.

node/impl/full/eth.go
geoff-vball: There was only one caller missing the check, so I've added it there and removed this.

node/impl/full/eth.go
geoff-vball: The `To` address get populated in `EthTxArgsFromUnsignedEthMessage`. Any messages sent this way have to be for f0 or f4, which means we don't need to looks anything up in state, we just convert them using `EthAddressFromFilecoinAddress`.

geoff-vball: I might be missing an edge case here that I haven't thought of though. What are your thoughts?

geoff-vball: Come to think of it, we should be able to do the same for `From` in this case, as it should always be an f4? It's only for Messages sent in native form that we might have to look up, say, the ID address of an f1 and then transform it into our eth-equivalent.

node/impl/full/eth.go
raulk: Gotcha.

raulk (approved):
This looks good. Gets rid of some cruft related to old EAM method that were invokable externally and are no longer so, superseded by CreateExternal. I think there's a lot more consolidation/cleanup that we can still do.

For example, I would like to see the `EthTxArgs` type be renamed to `EthRlpTx` and functions  associated with that type as members of the new type.

And there's more cleanup possible within the various conversion functions.

raulk:
The failing test is a flaky test.

Issue #9883, opened 2022-12-15
TestPaymentChannelsAPI is flaky


geoff-vball:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
master
```

### Describe the Bug

Test sometimes fails

### Logging Information

https://app.circleci.com/pipelines/github/filecoin-project/lotus/25015/workflows/d4f9890d-ca91-4a01-8ae3-c4354fc57066/jobs/719963/tests#failed-test-0

https://app.circleci.com/pipelines/github/filecoin-project/lotus/25014/workflows/06172fc3-c528-4e3e-a47e-86d721561c24/jobs/719831/tests#failed-test-0

[...]

PR #10100, opened 2023-01-23
fix: itests: Fix flaky paych test

magik6k:
## Related Issues
Did we really not have an issue for this?

## Proposed Changes
Fix the test

## Additional Info
The test was failing likely due to it running on two nodes, and in some cases when a message creating a channel landed on one node, we'd query state from the other node - which isn't the block miner, and could fall behind in sync (block time was 5ms, so that could happen quite easily). The fix is to wait for the nodes to be in sync at certain points.

## Checklist

- [x] CI is green

### DIFF
--- a/itests/paych_api_test.go
+++ b/itests/paych_api_test.go
@@ -38,7 +38,7 @@ func TestPaymentChannelsAPI(t *testing.T) {
 	kit.QuietMiningLogs()
 
 	ctx := context.Background()
-	blockTime := 5 * time.Millisecond
+	blockTime := 10 * time.Millisecond
 
 	var (
 		paymentCreator  kit.TestFullNode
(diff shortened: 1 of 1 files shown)


arajasek:
> ## Related Issues
> Did we really not have an issue for this?
> 

https://github.com/filecoin-project/lotus/issues/9883?

itests/paych_api_test.go
```diff
@@ -55,6 +55,15 @@ func TestPaymentChannelsAPI(t *testing.T) {
 	bms := ens.BeginMiningMustPost(blockTime)
 	bm := bms[0]
 
+	waitRecvInSync := func() {
+		// paymentCreator is the block miner, in some cases paymentReceiver may fall behind, so we wait for it to catch up
+
+		head, err := paymentReceiver.ChainHead(ctx)
+		require.NoError(t, err)
+
+		paymentReceiver.WaitTillChain(ctx, kit.HeightAtLeast(head.Height()))
```
arajasek, line 64: I'd even suggest doing head + 5 here, just in case of minor reorgs.

arajasek (approved):
I think this makes sense, though I don't know for certain this was the cause of (some of)  the flakiness. I suspect a similar pattern might cause flakiness in other tests too.

itests/paych_api_test.go
magik6k: There's one miner in that test, can't have reorgs

PR #10114, opened 2023-01-25
chore: cli: cleanup cli 

rjan90:
## Related Issues
Standarize cli/code functions similar to: https://github.com/filecoin-project/lotus/pull/9317

## Proposed Changes
- Change to cctx.NArg() instead of cctx.Args().xxx
- Add ArgsUsage to commands that was missing it
- Add check for args and print help on functions that did not have it

## Checklist

- [ ] CI is green

### DIFF
--- a/cli/chain.go
+++ b/cli/chain.go
@@ -112,8 +112,8 @@ var ChainGetBlock = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
-		if !cctx.Args().Present() {
-			return fmt.Errorf("must pass cid of block to print")
+		if cctx.NArg() != 1 {
+			return IncorrectNumArgs(cctx)
 		}
 
 		bcid, err := cid.Decode(cctx.Args().First())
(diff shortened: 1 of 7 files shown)


geoff-vball:
Thanks, this looks great! Just needs a `make docsgen-cli`

geoff-vball:
@rjan90 I've rerun the flaky tests a few times. It looks like the remaining failures are real. Looks like strictly checking the number of params in a few cases caused problems. Could be the test that's wrong

rjan90:
@geoff-vball Yeah, looks like there is issues with the `lotus wallet markets add` test, I will check it up tomorrow

rjan90:
Looks like all checks have passed now. I ended up having a not so strict ArgCheck for the `lotus wallet markets add` cmd. The failure seemed to come from a discrepency in the CLI and test for [MarketAddBalance](https://pkg.go.dev/github.com/filecoin-project/lotus/api/v0api#FullNode.MarketAddBalance).

While MarketAddBalance needs `wallet`, `addr` and `amt`, the cli-function only asks for amt and defaults to the default address for `wallet` and `addr` if not anything else is specified through the options `--address` and `--from`. So there is potentially an edge case failure in the cli, where it will fail if lotus has not an default adress set and the wallet and addr is not set through the options.

geoff-vball (approved):

geoff-vball:
Awesome, that's totally fine by me.

Issue #10115, opened 2023-01-25
rpcenc: Reader redirection may get stuck if the reader request is cancelled


magik6k:
This probably locked up AddPiece operations on filcollins

```
goroutine 579256 [chan send, 958 minutes]:
github.com/filecoin-project/lotus/lib/rpcenc.(*RpcReader).redirect.func1.1()
	/home/nonsense/lotus/lib/rpcenc/reader.go:320 +0x5e
sync.(*Once).doSlow(0xc00fcd8c70?, 0x8b16f1?)
	/usr/local/go/src/sync/once.go:68 +0xc2
sync.(*Once).Do(...)
	/usr/local/go/src/sync/once.go:59
github.com/filecoin-project/lotus/lib/rpcenc.(*RpcReader).redirect.func1()
	/home/nonsense/lotus/lib/rpcenc/reader.go:319 +0x6d
sync.(*Once).doSlow(0x1?, 0xc0240c1fc0?)
	/usr/local/go/src/sync/once.go:68 +0xc2
sync.(*Once).Do(...)
	/usr/local/go/src/sync/once.go:59
github.com/filecoin-project/lotus/lib/rpcenc.(*RpcReader).redirect(0xc010b48cf0?, {0xc0198decd0?, 0x24?})
	/home/nonsense/lotus/lib/rpcenc/reader.go:318 +0x76
github.com/filecoin-project/lotus/lib/rpcenc.ReaderParamEncoder.func1({0x41e8940?, 0xc019b971f0?, 0xc00fcd8fa8?})
	/home/nonsense/lotus/lib/rpcenc/reader.go:124 +0x3b9
github.com/filecoin-project/go-jsonrpc.(*rpcFunc).handleRpcCall(0xc013b8dc70, {0xc00f862100, 0x5, 0x5})
	/home/nonsense/go/pkg/mod/github.com/filecoin-project/go-jsonrpc@v0.1.8/client.go:478 +0xe63
github.com/filecoin-project/lotus/api.(*WorkerStruct).AddPiece(0x203000?, {0x5042198?, 0xc019c38e00?}, {{0xc0226a8990?, 0x722f5a?}, 0xc0226a8b20?}, {0xc0178fa9d0?, 0xc0226a88b8?, 0xc0226a89ec?}, 0x3f8000000, ...)
	/home/nonsense/lotus/api/proxy_gen.go:5641 +0x9b
[...]

PR #10116, opened 2023-01-25
fix: rpcenc: Don't hang when source dies

magik6k:
## Related Issues
Couldn't find anything about "AddPiece stuck" where it wasn't showing up in either `sealig sched-diag` or `sealing jobs`, but maybe there is an issue about that

Fixes https://github.com/filecoin-project/lotus/issues/10115

## Checklist

- [ ] CI is green

### DIFF
--- a/lib/rpcenc/reader.go
+++ b/lib/rpcenc/reader.go
@@ -282,7 +282,7 @@ func (w *RpcReader) Read(p []byte) (int, error) {
 	}
 
 	if w.postBody == nil {
-		return 0, xerrors.Errorf("reader already closed or redirected")
+		return 0, xerrors.Errorf("reader already closed, redirected or cancelled")
 	}
 
 	n, err := w.postBody.Read(p)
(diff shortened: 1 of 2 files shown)


lib/rpcenc/reader.go
```diff
@@ -406,6 +406,29 @@ func ReaderParamDecoder() (http.HandlerFunc, jsonrpc.ServerOption) {
...
+					case r, ok := <-wr.res:
+						if !ok {
+							return
+						}
+						log.Errorw("discarding read res", "type", r.rt, "meta", r.meta)
+					case <-closed:
+						return
+					}
+				}
+			}()
```
travisperson, line 424: What does this extra `closed` channel do when we are immediately closing it? Isn't this going to race between the two `select` cases and possibly immediately returning without reading anything off of `wr.res`?

lib/rpcenc/reader_test.go
```diff
@@ -31,6 +36,24 @@ func (h *ReaderHandler) ReadAllApi(ctx context.Context, r io.Reader, mustRedir b
...
+	h.cont <- struct{}{}
+	<-h.cont
+
+	var m []byte
+	m, h.subErr = h.readApi(ctx, r)
+
+	h.cont <- struct{}{}
+
+	return m, h.subErr
+}
```
travisperson, line 55: This method seems to only be for testing, should it not be a public method on `ReadHandler`?

travisperson (commented):
I kind of understand what this test is doing, and I think it looks good, but probably worth having another set of eyes on it.

lib/rpcenc/reader.go
magik6k: It cleans up this goroutine.

`wr.res` can only be written in `wr.beginOnce`, and is only written to once, so by calling it below, we know that `wr.res` must have been written to already. 

This bit of code spins up a go-routine which will read pending sends into `wr.res`, then will make sure that `wr.beginOnce` was done (so `wr.res` can't be written to anymore), then it will shutdown the draining goroutine by closing the `closed` channel.

lib/rpcenc/reader_test.go
magik6k: `ReadHandler` is also private to the test. The method must be public for go-jsonrpc to pick it up when creating test rpc server.

travisperson: Oh dump me, it's defined in this file.

arajasek (approved):
LGTM

PR #10117, opened 2023-01-25
feat: vm: Assert empty object CID when dumping state

arajasek:
## Related Issues

Resolves a TODO in code.

## Proposed Changes

We have certain actors that have no state. We register these actors as having `nil` states in the actor registry. When dumping an actor's state, we currently just skip these actors (return nil).

In Filecoin today, however, the Head for these actors must always be the EmptyObjectCID. This PR changes DumpActorState behaviour to actually assert that and fail if the Head is something else -- this is only a UX change, but is still correcter.

## Checklist

- [ ] CI is green


maciejwitowski:
Fixes https://github.com/filecoin-project/ref-fvm/issues/1504

geoff-vball (approved):

PR #10125, opened 2023-01-26
_ci_ / _build_ : Fix dirty git state when building docker images

ianconsolata:
## Related Issues
@geoff-vball This fixes some tricky issues with our docker build process that have been around for a long time, and result in the git state being dirty every time we build docker containers. That means our published containers contain lotus binaries that return `.dirty` as part of `lotus --version`

This was happening because the `.dockerignore`, which is a symlinked copy of the `.gitignore`, was ignore files that were checked into git state, and removing them from the docker build context.

## Proposed Changes
The original build had dirty state becuase of the AppDir folder (which can safely be removed now that AppImage is deprecated), and because of a file in extern/filecoin-ffi which matched a regexp. This PR removes the AppDir, and adds the extern/filecoin-ffi file to the include list in the `.dockerignore`/`.gitignore`. It also adds a test in the Dockerfile that will break the docker build if this ever happens again, so we can make sure we are always releasing images built from clean git state.

## Checklist

- [X] CI is green


geoff-vball (approved):
Nice! Thanks for the fix :)

GlacierWalrus (approved):

magik6k (approved):

PR #10132, opened 2023-01-26
evm: feat: always call contracts on InvokeContract

Stebalien:
This:

1. Always uses InvokeContract, never reducing to a method-0 value transfer.
2. Always encodes the parameters as a cbor byte array for simplicity.

This removes _many_ edge cases and makes it possible to run contract
logic with no parameters.

Depends on https://github.com/filecoin-project/ref-fvm/issues/1567

Fixes the issues outlined in https://github.com/filecoin-project/ref-fvm/issues/1302

NOTE: This change doesn't include any network version checks. We'll have to implement hyperspace specific ones in a separate PR.

### DIFF
--- a/chain/types/ethtypes/eth_transactions.go
+++ b/chain/types/ethtypes/eth_transactions.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	gocrypto "github.com/filecoin-project/go-crypto"
+	"github.com/filecoin-project/go-state-types/abi"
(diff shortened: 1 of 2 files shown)


Stebalien:
We need to release/bubble the FVM first.

Stebalien:
FIP changes: https://github.com/filecoin-project/FIPs/pull/606

Stebalien:
Fixed in https://github.com/filecoin-project/lotus/pull/10239/files.

PR #10143, opened 2023-01-30
fix:must close when sqlite return sql.Rows

imxyb:
## Related Issues
After the sqlite query returns `sql.Rows`, which should need to be closed, otherwise it will result in too many open files

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [x] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [x] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #10144, opened 2023-01-30
fix: eth: plug sqlite leak

geoff-vball:
## Related Issues
Alternative implementation of https://github.com/filecoin-project/lotus/pull/10143

## Proposed Changes
Calling `db.Query()` gives back a file-descriptor that has to be closed, this was causing fd leaks. Since we're querying unique values, we can just call `db.QueryRow()` which is more succinct and doesn't cause leaks.

## Checklist

- [ ] CI is green


raulk (approved):
I think we had identified this in the original review -- glad we're fixing it!

arajasek (approved):

PR #9192, opened 2022-08-20
Chain ranged export

frrist:
## Checklist

Before you mark the PR ready for review, please make sure that:
- [ ] All commits have a clear commit message.
- [ ] The PR title is in the form of of `<PR type>: <area>: <change being made>`
    - example: ` fix: mempool: Introduce a cache for valid signatures`
    - `PR type`: _fix_, _feat_, _INTERFACE BREAKING CHANGE_, _CONSENSUS BREAKING_, _build_, _chore_, _ci_, _docs_,_perf_, _refactor_, _revert_, _style_, _test_
    - `area`: _api_, _chain_, _state_, _vm_, _data transfer_, _market_, _mempool_, _message_, _block production_, _multisig_, _networking_, _paychan_, _proving_, _sealing_, _wallet_, _deps_
- [ ] This PR has tests for new functionality or change in behaviour
- [ ] If new user-facing features are introduced, clear usage guidelines and / or documentation updates should be included in https://lotus.filecoin.io or [Discussion Tutorials.](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] CI is green

PR #10145, opened 2023-01-30
perf: chain: export-range

hsanjuan:
Taking over from #9192 .

## Related Issues
https://github.com/filecoin-project/lotus/pull/9192

## Proposed Changes
A new `chain export-range` command that can create archival-grade ranged exports of the chain as quickly as possible (letting Lotus write to disk directly).

## Checklist

- [ ] CI is green

### DIFF
--- a/api/api_full.go
+++ b/api/api_full.go
@@ -172,6 +172,10 @@ type FullNode interface {
 	// If oldmsgskip is set, messages from before the requested roots are also not included.
 	ChainExport(ctx context.Context, nroots abi.ChainEpoch, oldmsgskip bool, tsk types.TipSetKey) (<-chan []byte, error) //perm:read
 
+	ChainExportRange(ctx context.Context, head, tail types.TipSetKey, cfg *ChainExportConfig) (<-chan []byte, error) //perm:read
+
+	ChainExportRangeInternal(ctx context.Context, head, tail types.TipSetKey, cfg *ChainExportConfig) error //perm:read
+
 	// ChainPrune prunes the stored chain state and garbage collects; only supported if you
 	// are using the splitstore
 	ChainPrune(ctx context.Context, opts PruneOpts) error //perm:admin
--- a/api/v0api/full.go
+++ b/api/v0api/full.go
@@ -161,6 +161,10 @@ type FullNode interface {
 	// If oldmsgskip is set, messages from before the requested roots are also not included.
 	ChainExport(ctx context.Context, nroots abi.ChainEpoch, oldmsgskip bool, tsk types.TipSetKey) (<-chan []byte, error) //perm:read
 
+	ChainExportRange(ctx context.Context, head, tail types.TipSetKey, cfg *api.ChainExportConfig) (<-chan []byte, error) //perm:read
+
+	ChainExportRangeInternal(ctx context.Context, head, tail types.TipSetKey, cfg *api.ChainExportConfig) error //perm:read
+
 	// MethodGroup: Beacon
 	// The Beacon method group contains methods for interacting with the random beacon (DRAND)
 
--- a/node/impl/full/chain.go
+++ b/node/impl/full/chain.go
@@ -5,11 +5,14 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"math"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/ipfs/go-blockservice"
 	"github.com/ipfs/go-cid"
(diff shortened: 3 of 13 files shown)


hsanjuan:
The code from my side is mostly finished. I have this running in "production" already, as we rely on it for archival snapshots. Testing has involved making sure that it exports the exact same set of blocks as the previous code (#9192). We have used snapshots generated by the previous code and this code to process the full Filecoin chain with Lily, so we are reasonably sure that they include what they should.

In terms of performance, this version of the code seems to be a bit faster (~10%), mostly due to bugs fixed and improved blockstore access. Memory profiles are the same, essentially a large amount of memory is required to keep the "processed CID set" and the task queue (we cannot control its growth since a blockstore block can have an arbitrary number of children that need to be queued). That said memory usage is consistent. We had episodes of OOM/leaks with the previous code, but those have not happened so far with this one. The addressing of some deadlock-ing code makes me feel better about them though.

In the previous PR (#9192 ), there was a comment from @magik6k about back-pressure and improved RPC handling when doing a normal export (without `--internal`).

I am of the opinion that we should remove that possibility all together and that export-range should only export to local disk. This is way faster. Currently snapshots are very large and take about 1h this way. I don't find it very useful to be able to ship them remotely via the RPC API (even more given that they are not compressed).

In any case, I would implement that later if there is a better way that the one currently used, rather than letting that delay having this functionality upstreamed.

--

I don't know what is with CI. I'm happy updating documentation when we have finished discussion here.

api/api_full.go
```diff
@@ -172,6 +172,10 @@ type FullNode interface {
 	// If oldmsgskip is set, messages from before the requested roots are also not included.
 	ChainExport(ctx context.Context, nroots abi.ChainEpoch, oldmsgskip bool, tsk types.TipSetKey) (<-chan []byte, error) //perm:read
 
+	ChainExportRange(ctx context.Context, head, tail types.TipSetKey, cfg ChainExportConfig) (<-chan []byte, error) //perm:read
+
+	ChainExportRangeInternal(ctx context.Context, head, tail types.TipSetKey, cfg ChainExportConfig) error //perm:read
```
magik6k, line 177: This one writes a lot of data locally, definitely higher perm that read; I'd say it should be `admin` because it's easy enough to abuse to bring the node down

api/v0api/full.go
```diff
@@ -161,6 +161,10 @@ type FullNode interface {
 	// If oldmsgskip is set, messages from before the requested roots are also not included.
 	ChainExport(ctx context.Context, nroots abi.ChainEpoch, oldmsgskip bool, tsk types.TipSetKey) (<-chan []byte, error) //perm:read
 
+	ChainExportRange(ctx context.Context, head, tail types.TipSetKey, cfg api.ChainExportConfig) (<-chan []byte, error) //perm:read
+
+	ChainExportRangeInternal(ctx context.Context, head, tail types.TipSetKey, cfg api.ChainExportConfig) error //perm:read
+
```
magik6k, line 167: Do we really need in on v0 api?

node/impl/full/chain.go
```diff
@@ -589,6 +593,111 @@ func (m *ChainModule) ChainGetMessage(ctx context.Context, mc cid.Cid) (*types.M
...
+func (a ChainAPI) ChainExportRangeInternal(ctx context.Context, head, tail types.TipSetKey, cfg api.ChainExportConfig) error {
+	headTs, err := a.Chain.GetTipSetFromKey(ctx, head)
+	if err != nil {
+		return xerrors.Errorf("loading tipset %s: %w", head, err)
+	}
+	tailTs, err := a.Chain.GetTipSetFromKey(ctx, tail)
+	if err != nil {
+		return xerrors.Errorf("loading tipset %s: %w", tail, err)
+	}
+	fileName := fmt.Sprintf("./snapshot_%d_%d_%d.car", tailTs.Height(), headTs.Height(), time.Now().Unix())
```
magik6k, line 605: This arguably should be written to the repo path, which you can get by adding `repo.LockedRepo` to the ChainAPI struct, and calling `.Path` on it.

node/impl/full/chain.go
```diff
@@ -589,6 +593,111 @@ func (m *ChainModule) ChainGetMessage(ctx context.Context, mc cid.Cid) (*types.M
...
 
+func (a ChainAPI) ChainExportRangeInternal(ctx context.Context, head, tail types.TipSetKey, cfg api.ChainExportConfig) error {
+	headTs, err := a.Chain.GetTipSetFromKey(ctx, head)
+	if err != nil {
+		return xerrors.Errorf("loading tipset %s: %w", head, err)
+	}
+	tailTs, err := a.Chain.GetTipSetFromKey(ctx, tail)
+	if err != nil {
+		return xerrors.Errorf("loading tipset %s: %w", tail, err)
+	}
```
magik6k, line 604: Nit: should check that head.Height is > tail

api/v0api/full.go
hsanjuan: I don't know, is that like the older version kept for compatibility? Then I'd say no...

node/impl/full/chain.go
hsanjuan: https://github.com/filecoin-project/lotus/pull/10145/commits/160c2dbf9d1eef6fe77677cdd85b2e386ea3de33

api/v0api/full.go
hsanjuan: https://github.com/filecoin-project/lotus/pull/10145/commits/71b478e38fb835ef4572b0157feb0e8b4927559d

api/api_full.go
hsanjuan: https://github.com/filecoin-project/lotus/pull/10145/commits/71b478e38fb835ef4572b0157feb0e8b4927559d

node/impl/full/chain.go
hsanjuan: `ERROR: initializing node: applying node options failed: repo is already locked (lotus daemon already running)`

hsanjuan:
```diff
diff --git a/node/impl/full/chain.go b/node/impl/full/chain.go
index 5dedc6991..6923e311c 100644
--- a/node/impl/full/chain.go
+++ b/node/impl/full/chain.go
@@ -42,6 +42,7 @@ import (
        "github.com/filecoin-project/lotus/lib/oldpath"
        "github.com/filecoin-project/lotus/lib/oldpath/oldresolver"
        "github.com/filecoin-project/lotus/node/modules/dtypes"
+       "github.com/filecoin-project/lotus/node/repo"
 )

 var log = logging.Logger("fullnode")
@@ -93,6 +94,8 @@ type ChainAPI struct {

        // BaseBlockstore is the underlying blockstore
        BaseBlockstore dtypes.BaseBlockstore
+
+       Repo repo.LockedRepo
 }

 func (m *ChainModule) ChainNotify(ctx context.Context) (<-chan []*api.HeadChange, error) {
@@ -606,7 +609,7 @@ func (a ChainAPI) ChainExportRangeInternal(ctx context.Context, head, tail types
                return xerrors.Errorf("Height of head-tipset (%d) must be greater or equal to the height of the tail-tipset (%d)", headTs.Height(), tailTs.Height())
        }

-       fileName := fmt.Sprintf("./snapshot_%d_%d_%d.car", tailTs.Height(), headTs.Height(), time.Now().Unix())
+       fileName := filepath.Join(a.Repo.Path(), fmt.Sprintf("snapshot_%d_%d_%d.car", tailTs.Height(), headTs.Height(), time.Now().Unix()))
        absFileName, err := filepath.Abs(fileName)
        if err != nil {
                return err
```

travisperson:
> I am of the opinion that we should remove that possibility all together and that export-range should only export to local disk. This is way faster.

Removing this will mean that we will not be able to use this new export code with the existing snapshot service and will have to continue to use the `ChainExport` method until API support is added or a new design for snapshots is created to handle local disks.

cc @ognots

hsanjuan:
> > I am of the opinion that we should remove that possibility all together and that export-range should only export to local disk. This is way faster.
> 
> Removing this will mean that we will not be able to use this new export code with the existing snapshot service and will have to continue to use the `ChainExport` method until API support is added or a new design for snapshots is created to handle local disks.
> 
> cc @ognots

It is not going to be much faster than what you already have when pushing the snapshot through the API endpoint.

hsanjuan:
ok, I think I am good wrt to tests. The things that are failing or stuck don't seem related to this.

It is a bit of a pain to rebase on master given the conflicts on generated-code. What is left to do? I can write a tutorial, but hopefully can do post-merge?

magik6k (approved):

Issue #1614, opened 2020-04-27
block service always get failed at  "stream reset" or "i/o deadline reached"


taoshengshi:
**Describe the bug**
lotus sync block failed and raised error as the below picture speaks.

**To Reproduce**
1. Run './lotus daemon' to join the interopnet
2. Run './lotus sync status'
3. Wait awhile and See error

**Expected behavior**
![image](https://user-images.githubusercontent.com/33315004/80378090-b9332a00-88ce-11ea-9b57-52053e3e2835.png)

**Screenshots**
as above.

**Version (run `lotus --version`):**
network : interopnet
commit : 1bfa2311d6931dc0c961d42f561ba0268dbd21c3

**Additional context**
not sure '5 * time.Second' is enough and how to sync recurrently when meet 'stream reset'.
lotus/chain/blocksync/blocksync_client.go:294
s.SetWriteDeadline(time.Now().Add(5 * time.Second))

phil-andrews:
this might have to do with your open file limit - https://docs.lotu.sh/en+setup-troubleshooting

`ulimit -n 10000`

magik6k:
We've rewritten big chocks of this sync code, I'm pretty sure it now handles timeouts correctly

PR #10151, opened 2023-01-31
Eth JSON-RPC: from in eth_getTransactionByHash is not correctly popul…

maciejwitowski:
…ated #1614

## Related Issues
Closes https://github.com/filecoin-project/ref-fvm/issues/1614

- [ ] CI is green

### DIFF
--- a/chain/types/ethtypes/eth_transactions.go
+++ b/chain/types/ethtypes/eth_transactions.go
@@ -77,16 +77,10 @@ func EthTxFromSignedEthMessage(smsg *types.SignedMessage) (EthTx, error) {
 		return EthTx{}, xerrors.Errorf("failed to recover signature: %w", err)
 	}
 
-	fromAddr, err := EthAddressFromFilecoinAddress(smsg.Message.From)
-	if err != nil {
-		return EthTx{}, xerrors.Errorf("failed to get from: %w", err)
-	}
-
 	return EthTx{
 		Nonce:                EthUint64(txArgs.Nonce),
 		ChainID:              EthUint64(txArgs.ChainID),
 		To:                   txArgs.To,
-		From:                 fromAddr,
 		Value:                EthBigInt(txArgs.Value),
 		Type:                 Eip1559TxType,
 		Gas:                  EthUint64(txArgs.GasLimit),
--- a/chain/types/ethtypes/eth_transactions.go
+++ b/chain/types/ethtypes/eth_transactions.go
@@ -77,10 +77,16 @@ func EthTxFromSignedEthMessage(smsg *types.SignedMessage) (EthTx, error) {
 		return EthTx{}, xerrors.Errorf("failed to recover signature: %w", err)
 	}
 
+	fromAddr, err := EthAddressFromFilecoinAddress(smsg.Message.From)
+	if err != nil {
+		return EthTx{}, xerrors.Errorf("failed to get from: %w", err)
+	}
+
 	return EthTx{
 		Nonce:                EthUint64(txArgs.Nonce),
 		ChainID:              EthUint64(txArgs.ChainID),
 		To:                   txArgs.To,
+		From:                 fromAddr,
 		Value:                EthBigInt(txArgs.Value),
 		Type:                 Eip1559TxType,
 		Gas:                  EthUint64(txArgs.GasLimit),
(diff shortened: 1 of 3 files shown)


chain/types/ethtypes/eth_transactions.go
```diff
@@ -46,6 +46,7 @@ type EthTxArgs struct {
 	ChainID              int         `json:"chainId"`
 	Nonce                int         `json:"nonce"`
 	To                   *EthAddress `json:"to"`
+	From                 *EthAddress `json:"from"`
```
arajasek, line 49: I don't think we want to add this to `EthTxArgs` --the transaction arguments are independent of any single sender -- it should be the case that whoever signs them has a valid tx ready to go.

api/api_full.go
```diff
@@ -1345,7 +1345,7 @@ type EthTxReceipt struct {
 	TransactionIndex  ethtypes.EthUint64   `json:"transactionIndex"`
 	BlockHash         ethtypes.EthHash     `json:"blockHash"`
 	BlockNumber       ethtypes.EthUint64   `json:"blockNumber"`
-	From              ethtypes.EthAddress  `json:"from"`
+	From              *ethtypes.EthAddress `json:"from"`
```
arajasek, line 1348: What's motivating this change?

chain/types/ethtypes/eth_transactions.go
```diff
@@ -122,6 +125,12 @@ func EthTxArgsFromUnsignedEthMessage(msg *types.Message) (EthTxArgs, error) {
 		}
 		to = &addr
 
+		fromAddr, err := EthAddressFromFilecoinAddress(msg.From)
```
arajasek, line 128: There's a slight issue with this -- we might not be returning the "best" possible EthAddress here. In particular, we'd like to resolve to _actual_ eth addresses if the `msg.From` is an ID address (by looking up the corresponding f4 address). See `node/impl/full/eth.go`'s `lookupEthAddress` for the best-effort implementation of this.

I think we want to make this change at the API level when forming an `EthTx`. I think `newEthTxFromSignedMessage` is likely the best place for this.

arajasek:
Not sure what's wrong with CI, but it's definitely not caused by this PR :)

chain/types/ethtypes/eth_transactions.go
```diff
@@ -77,10 +77,16 @@ func EthTxFromSignedEthMessage(smsg *types.SignedMessage) (EthTx, error) {
 		return EthTx{}, xerrors.Errorf("failed to recover signature: %w", err)
 	}
 
+	fromAddr, err := EthAddressFromFilecoinAddress(smsg.Message.From)
```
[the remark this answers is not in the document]
maciejwitowski: Pushing this 1 level up makes sense 👍

node/impl/full/eth.go
```diff
@@ -1571,6 +1571,13 @@ func newEthTxFromSignedMessage(ctx context.Context, smsg *types.SignedMessage, s
 		if err != nil {
 			return ethtypes.EthTx{}, err
```
arajasek, line 1572: Nit: Can you please also include an error message here? It's _totally_ not related to this PR, but it's been frustrating not knowing where errors have been originating from in the Eth API (so I'm trying to slowly add them).
```suggestion
			return ethtypes.EthTx{}, xerrors.Errorf("failed to calculate hash for ethTx: %w", err)

```

arajasek (commented):
Change looks good, but let's expand one of the tests to cover this case (clearly needed!)

node/impl/full/eth.go
maciejwitowski: Absolutely ❤️

arajasek (approved):
LGTM!

itests/eth_transactions_test.go
```diff
@@ -83,6 +83,11 @@ func TestValueTransferValidSignature(t *testing.T) {
 
 	// Success.
 	require.EqualValues(t, ethtypes.EthUint64(0x1), receipt.Status)
+
+	ethTx, err := client.EthGetTransactionByHash(ctx, &hash)
+	require.Nil(t, err)
+	require.EqualValues(t, ethAddr, ethTx.From)
```
raulk, line 89: Let's also assert against all `tx` fields? That way if something else fails to populate/breaks, we'll catch it too.

raulk (approved):
Can strengthen the test.

itests/eth_transactions_test.go
maciejwitowski: Won't be able to do it today, but happy to land this later this week 🙂

PR #10168, opened 2023-02-01
Test: assert all fields returned from EthGetTransactionByHash have expected values

maciejwitowski:


geoff-vball (approved):
Just fill out a brief description, but otherwise looks good to go :)

Issue #9776, opened 2022-12-02
NV18: Confirm tooling that assumes knowledge of method params / returns don't break


arajasek:
We have tooling that can be used to encode / decode the params & return value of messages. This is important to support for builtin-actors, but cannot be supported for user-defined actors. We need to make sure that such tooling gracefully fails in this case.

jennijuju:
can be a thing happen during calibration

scotthconner:
Is this client side tooling, or libraries/functions inside of lotus itself? Trying to understand this item.

jennijuju:
@scotthconner in lotus cli/api

arajasek:
The best way to confirm this is to call `lotus chain decode params` with the params supplied to messages that deploy and invoke contracts on a devnet / testnet. We expect the "outer" params to decode correctly, but not those that are actually passed to the contract.

More importantly, no panicking / ugly failing.

PR #10173, opened 2023-02-01
chore: cli: Confirm tooling that assumes knowledge of method params / returns don't break

geoff-vball:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/9776#event-82953d91-7041-554f-b87e-d92435a345c1

## Proposed Changes
Tests the decoding the params/returns for InvokeContract and CreateExternal. These were the only new function calls I could find on hyperspace through filfox.

## Checklist

- [ ] CI is green


arajasek:
Thank you for this, it's good to see! Can we improve the tests to include actual assertions? I think we should be able to roundtrip the params in question, and use that for assertions. This might look something like:

- Defining a returnStruct
```
var dummyReturn Return {
	ActorID: 6491,
	RobustAddress: f275fpmqmhgj3zsfytktprbnftx7bojqyga3wyu5a (decoded from address)
	EthAddress    []byte{...the bytes}
}
``` 
- Marshaling that into a bytestr, which should give you the `8319195b5502ff4af64187327799171354df10b4b3bfc2e4c30654922d6956c99e12dfeb3224dea977d0939758a1fe` string
- Then going through the process this test currently does, and comparing the return against `dummyReturn`

arajasek (approved):

PR #10193, opened 2023-02-06
itest: fix FEVM tests for upstream changes

Stebalien:


itests/fevm_test.go
```diff
@@ -151,19 +151,19 @@ func TestFEVMRecursive2(t *testing.T) {
 }
 
 // TestFEVMBasic does a basic fevm contract installation and invocation
-// recursive delegate call succeeds up to 238 times
+// recursive delegate call succeeds up to 228 times
 func TestFEVMRecursiveDelegatecall(t *testing.T) {
```
Stebalien, line 155: This test isn't making a lot of sense to me:

1. `recursiveDelegatecallFail` has always expected 228 successes, not 238.
2. If I change the "failure" loop to start at 229, the calls _succeed_ and the test fails. But if I change the success loop to end at 238, the success loop fails.

So there appears to be a gulf between success/failure. This is not good but I'm not sure what's going on.

arajasek: also nit that the comment on top of the test is wrong

Stebalien: (fixed)

magik6k: The 30 increment in the loop makes it hard to tell what the exact threshold is

magik6k (approved):

itests/fevm_test.go
snissn: @magik6k @Stebalien thanks 100% agree - i reorganized the test https://github.com/filecoin-project/lotus/pull/10195

Issue #10185, opened 2023-02-03
build: include bundled actors tag somewhere


arajasek:
We're facing a lot of confusion not knowing what actors bundle release is being run. This information is _technically_ deducable, since the builtin_actors_gen.go file contains the manifest CIDs, which can be cross-compared against the release artifacts in builtin-actors, but that's a lot.

We should modify pack.sh to include this information...somewhere (perhaps builtin_actors_gen.go). Note that this information will not (necessarily) be correct -- no part of the system will be verifying the tag on startup, but that's okay -- this is just for our own sanity.

snissn:
Hey @arajasek - can you provide some extra guidance here? I found the CID associated with the ManifestCid - and I find when I grep my repo it finds that CID in the lotus binary, but beyond that I'm not sure what next steps to take. I also tried to fetch the CID using ipfs but it wasn't downloading.

arajasek:
Yeah, so what I think we should do is incorporate the actual git tag of the bundle being used.

Lotus currently updates / adds bundles with the `pack.sh` script, which takes a git tag as input, and fetches the bundle artifact directly from GitHub. I think it'd be very useful if as part of this process, we _recorded_ the tag somewhere in Lotus itself. This would serve as an easy reference to the Git commit you're running (for example if you wanted to examine the actors source code).

I'm not fussed about _where_ we put this info. The pack.sh script invokes the generator in gen/bundle/bundle.go, which generates builtin_actors_gen.go. It might be possible to include the git tag as input to that generator. Or, if it's easier, just generate a different file with this info.

jennijuju:
so the way we run the ./pack.sh is `./pack.sh <version> <bundle tag>` -  as the script is run, if we can just append the version + bundle into the builtin_actors_gen.go (first line is also okay).

snissn:
Does it make sense to add the git tag to BuiltinActorsMetadata or would the git info be better as a comment in builtin_actors_gen.go?

jennijuju:
> Does it make sense to add the git tag to BuiltinActorsMetadata or would the git info be better as a comment in builtin_actors_gen.go?

Either is fine imho!

snissn:
Great! This PR adds the git tag to BuiltinActorsMetadata  https://github.com/filecoin-project/lotus/pull/10200

I'm trying to test the command but am hitting a snag 
`panic: bundle bundle '"mainnet.tar.zst"' doesn't start with a 'v'`

[...]

PR #10200, opened 2023-02-07
feat: pack: add git tag to builtin_actors_gen

snissn:
## Related Issues
 https://github.com/filecoin-project/lotus/issues/10185

## Proposed Changes
Pass git tag from pack.sh to builtin_actors through the makefile and add BundleGitTag to builtin_actors_gen.go

## Checklist

- [ ] CI is green

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -298,7 +298,7 @@ actors-gen: actors-code-gen fiximports
 .PHONY: actors-gen
 
 bundle-gen:
-	$(GOCC) run ./gen/bundle
+	$(GOCC) run ./gen/bundle $(RELEASE)
 	$(GOCC) fmt ./build/...
 .PHONY: bundle-gen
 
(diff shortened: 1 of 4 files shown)


arajasek (approved):
Thanks for the PR, this works well!

The only catch here is that invoking `bundle-gen` without the new `RELEASE`  input will blow away the BundleGitTag previously set.

I think this is fine -- this value will only get set through the `pack.sh` script. The only caveat is that `bundle-gen` is currently invoked by `make gen`, a command we regularly invoke to generate other auto-generated matter (like documentation). I think the correct think to do is to just drop `bundle-gen` from being part of `gen` entirely -- it's not actually useful today anyway.

(doing so will also fix the CI failure)

snissn:
Thanks @arajasek I pushed a commit that removes bundle-gen from make gen!

PR #10205, opened 2023-02-07
fvm: chore: update FVM

Stebalien:
This:

1. Updates the builtin actors bundle (for actors v10).
2. Updates the event entry type to include the codec.
3. Removes the cbor encoding and zero trimming from event data.

I've chose to:

1. _Not_ add codec handling to the event filtering system for now.
2. _Skip_ events with unexpected codecs.

We don't actually _allow_ these events in the FVM right now, and it simplifies the implementation.

However, I _am_ recording the codecs in the database so we don't have to migrate it later.


chain/events/filter/event.go
```diff
@@ -105,18 +105,9 @@ func (f *EventFilter) CollectEvents(ctx context.Context, te *TipSetEvents, rever
...
 			}
 
-			entries := make([]types.EventEntry, len(ev.Entries))
-			for i, entry := range ev.Entries {
-				entries[i] = types.EventEntry{
-					Flags: entry.Flags,
-					Key:   entry.Key,
-					Value: entry.Value,
-				}
-			}
```
Stebalien, line 115: I'm not actually sure what this code was supposed to be doing. I assume we didn't need to copy?

chain/events/filter/index.go
```diff
@@ -270,6 +272,7 @@ func (ei *EventIndex) PrefillFilter(ctx context.Context, f *EventFilter) error {
 			event.reverted,
 			event_entry.flags,
 			event_entry.key,
+			event_entry.codec,
```
Stebalien, line 275: We should be matching on codec + value, but that's not critical right now because we only support CBOR.

Stebalien:
Depends on https://github.com/filecoin-project/filecoin-ffi/pull/360

itests/eth_filter_test.go
```diff
@@ -291,7 +290,7 @@ func TestEthNewFilterDefaultSpec(t *testing.T) {
 				paddedEthBytes([]byte{0x33, 0x33}),
 				paddedEthBytes([]byte{0x44, 0x44}),
 			},
-			Data: paddedEthBytes([]byte{0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88}),
+			Data: []byte{0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88},
```
Stebalien, line 293: This expectation was wrong, as far as I can tell. The _topics_ are always 32 bytes but the _data_ is arbitrary.

build/builtin_actors_gen.go
```diff
@@ -29,9 +30,10 @@ var EmbeddedBuiltinActorsMetadata []*BuiltinActorsMetadata = []*BuiltinActorsMet
 		"verifiedregistry": MustParseCid("bafk2bzacebu4joy25gneu2qv3qfm3ktakzalndjrbhekeqrqk3zhotv6nyy2g"),
 	},
 }, {
-	Network:     "butterflynet",
-	Version:     9,
-	ManifestCid: MustParseCid("bafy2bzaceba5qgs4z3imhlxwds5vamahngatvuuglbv5yl3ftfiosj6ud5chs"),
+	Network:      "butterflynet",
+	Version:      9,
+	BundleGitTag: "dev/20230207-pre-rc.7",
```
jennijuju, line 35: hrmm for butterfly we likely want to start with the mainnet v9 actor and migrate to this new v10 bundle for testing purpose

jennijuju: oh no i think its just the wrong tag - i will create an issue

chain/events/filter/event.go
iand: I think originally we did some decoding here, which was removed

chain/events/filter/index.go
raulk: I think you mean Raw?

raulk: I think you mean Raw?

arajasek (approved):

PR #10214, opened 2023-02-08
feat: eth: harden event processing

Stebalien:
This commit hardens event processing:

1. Handles out of order topics.
2. Ensures that topics are 32 bytes.
3. Uses the correct type for topics.
4. Detects duplicate topics/values.


Stebalien:
Uh. Ok github, not sure what you were thinking there.

Stebalien:
https://github.com/filecoin-project/lotus/pull/10215

PR #10206, opened 2023-02-07
test: itest: fevm block.difficulty

snissn:
## Related Issues
 https://github.com/filecoin-project/ref-fvm/issues/1648

## Proposed Changes
add a test and solidity file for block.difficulty 

## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #10216, opened 2023-02-08
fix: stmgr: make the tipset and height agree when estimating gas

Stebalien:
## Related Issues

Should fix https://github.com/filecoin-project/ref-fvm/issues/1648.

## Proposed Changes

Specifically re-execute all messages in the current tipset, tacking the new message onto the end. That way, the epoch is the epoch of the current tipset.

We could try to "make" a fake block and use that, but that's unlikely to work well.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/call.go
+++ b/chain/stmgr/call.go
@@ -107,22 +107,16 @@ func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, pr
 		}
 	}
 
-	var vmHeight abi.ChainEpoch
-	if checkGas {
-		// Since we're simulating a future message, pretend we're applying it in the "next" tipset
-		vmHeight = ts.Height() + 1
-		if stateCid == cid.Undef {
-			stateCid, _, err = sm.TipSetState(ctx, ts)
-			if err != nil {
-				return nil, xerrors.Errorf("computing tipset state: %w", err)
-			}
+	// Unless executing on a specific state cid, apply all the messages from the current tipset
+	// first. Unfortunately, we can't just execute the tipset, because that will run cron. We
+	// don't want to apply miner messages after cron runs in a given epoch.
+	if stateCid == cid.Undef {
+		stateCid = ts.ParentState()
+		tsMsgs, err := sm.cs.MessagesForTipset(ctx, ts)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to lookup messages for parent tipset: %w", err)
 		}
-	} else {
-		// If we're not checking gas, we don't want to have to execute the tipset like above. This saves a lot of computation time
-		vmHeight = pts.Height() + 1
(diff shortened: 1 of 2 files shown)


maciejwitowski:
Tested https://github.com/filecoin-project/lotus/pull/10216, it works.

snissn:
This test also confirms this PR works https://github.com/filecoin-project/lotus/pull/10206

chain/stmgr/call.go
```diff
@@ -142,14 +136,14 @@ func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, pr
 	buffStore := blockstore.NewTieredBstore(sm.cs.StateBlockstore(), blockstore.NewMemorySync())
 	vmopt := &vm.VMOpts{
 		StateBase:      stateCid,
-		Epoch:          vmHeight,
+		Epoch:          ts.Height(),
```
snissn, line 139: we're changing the value of Epoch here from vmHeight (ts.Height()+1) to ts.Height() is that off by 1 change ok?

chain/stmgr/call.go
```diff
@@ -107,22 +107,16 @@ func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, pr
...
+	// Unless executing on a specific state cid, apply all the messages from the current tipset
+	// first. Unfortunately, we can't just execute the tipset, because that will run cron. We
+	// don't want to apply miner messages after cron runs in a given epoch.
+	if stateCid == cid.Undef {
+		stateCid = ts.ParentState()
+		tsMsgs, err := sm.cs.MessagesForTipset(ctx, ts)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to lookup messages for parent tipset: %w", err)
 		}
+		priorMsgs = append(tsMsgs, priorMsgs...)
```
magik6k, line 119: Are we sure this isn't breaking StateCall / `lotus state compute-state`?

Exchanges rely on this to track transactions.

chain/stmgr/call.go
Stebalien: Yes, that's the point.

chain/stmgr/call.go
Stebalien: That uses https://github.com/filecoin-project/lotus/blob/7aa6663127982aa404e68bdefaadb3f041eb5120/chain/stmgr/utils.go#L68

Stebalien: Which is also incorrect? Well, it's correct if you don't pass any messages.

Stebalien: Hm. It does affect `StateCall`. I'm going to rework this a bit.

Stebalien: Done. I now only do this for `CallWithGas`.

arajasek (approved):

Issue #10106, opened 2023-01-24
fevm: EthTx serialization does not match openrpc spec


iand:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [x] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
releases/v1.20.0
```

### Describe the Bug

The [openrpc conformance tests](https://github.com/filecoin-project/lotus/pull/10102) fail when checking the schema of Ethereum transactions returned from the eth json-rpc api.

```
    eth_conformance_test.go:306: - transactions: Must validate at least one schema (anyOf)
    eth_conformance_test.go:306: - transactions.0: Must validate one and only one schema (oneOf)
    eth_conformance_test.go:306: - transactions.0: accessList is required
    eth_conformance_test.go:306: - transactions.0: yParity is required
```

Affected methods:
[...]

PR #10223, opened 2023-02-09
fix: ethrpc: Add missing fields to EthTx

magik6k:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/10106

## Checklist

- [ ] CI is green


chain/types/ethtypes/eth_transactions.go
```diff
@@ -37,7 +37,8 @@ type EthTx struct {
 	Gas                  EthUint64   `json:"gas"`
 	MaxFeePerGas         EthBigInt   `json:"maxFeePerGas"`
 	MaxPriorityFeePerGas EthBigInt   `json:"maxPriorityFeePerGas"`
-	V                    EthBigInt   `json:"v"`
+	AccessList           []EthHash   `json:"accessList"`
+	V                    EthBigInt   `json:"yParity"`
```
magik6k, line 41: Afaict what we put here is in fact YParity according to https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2098.md; V is only used for legacy transactions

This value seems to be set to 0/1; not 27/28 which are V values

Not sure if it makes sense to rename the V field to `YParity` as well, no strong opinions.

chain/types/ethtypes/eth_transactions.go
```diff
@@ -37,7 +37,8 @@ type EthTx struct {
 	Gas                  EthUint64   `json:"gas"`
 	MaxFeePerGas         EthBigInt   `json:"maxFeePerGas"`
 	MaxPriorityFeePerGas EthBigInt   `json:"maxPriorityFeePerGas"`
-	V                    EthBigInt   `json:"v"`
+	AccessList           []EthHash   `json:"accessList"`
```
magik6k, line 40: Those come from https://eips.ethereum.org/EIPS/eip-2930, which I don't think are supported in FEVM

Given that those aren't possible to specify (?), and optional, I just add the field and keep it empty, which is enough to keep openrpc happy.

arajasek (approved):
Changes themselves look good to me, but review from @ychiaoli18 / @raulk / Aadi would be good

chain/types/ethtypes/eth_transactions.go
ychiaoli18: I'd prefer to rename it to yParity as well to keep them consistent.

ychiaoli18 (approved):

PR #10227, opened 2023-02-09
fix: Don't call WalletExport in msg signing flows

magik6k:
## Proposed Changes
Don't call WalletExport just to get key type

## Additional Info
Calls to WalletExport breaks remote wallets, looks really bad in interactive wallets, and probably also breaks ledger wallet.

## Checklist

- [ ] CI is green


arajasek (approved):
This doesn't work if the `msg.From` is an ID address, but looking at the code that case doesn't work anyway...

magik6k:
I'm msg.From is resolved before we call into the wallet

Issue #10226, opened 2023-02-09
Lotus: Fix inclusion fee under new gas model


Stebalien:
We need to account for the state update fee for the account. Otherwise:

1. A message will fail because its inclusion fee will be too small.
2. All subsequent messages will fail because their nonces won't match.

PR #10228, opened 2023-02-09
fix: gas: update ffi & correct the message inclusion cost in nv18

Stebalien:
## Related Issues

https://github.com/filecoin-project/lotus/issues/10226

## Proposed Changes

This fixes two issues:

1. We need to charge an additional 475,000 gas to update the sending actor.
2. We needed to update the FVM as it was miscalculating the inclusion cost.

I've also removed all the other fees from the price list as they don't make sense anymore.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/vm/gas.go
+++ b/chain/vm/gas.go
@@ -213,6 +213,16 @@ var Prices = map[abi.ChainEpoch]Pricelist{
 
 		verifyReplicaUpdate: 36316136,
 	},
+	build.UpgradeHyggeHeight: &pricelistV0{
+		computeGasMulti: 1,
+		storageGasMulti: 1300, // only applies to messages/return values.
+
+		onChainMessageComputeBase:    38863 + 475000, // includes the actor update cost
+		onChainMessageStorageBase:    36,
+		onChainMessageStoragePerByte: 1,
+
+		onChainReturnValuePerByte: 1,
+	},
 }
 
 // PricelistByEpoch finds the latest prices for the given epoch
--- a/itests/fevm_test.go
+++ b/itests/fevm_test.go
@@ -58,7 +58,7 @@ func buildInputFromuint64(number uint64) []byte {
 // recursive delegate calls that fail due to gas limits are currently getting to 229 iterations
 // before running out of gas
 func recursiveDelegatecallFail(ctx context.Context, t *testing.T, client *kit.TestFullNode, filename string, count uint64) {
-	expectedIterationsBeforeFailing := int(229)
+	expectedIterationsBeforeFailing := int(220)
 	fromAddr, idAddr := client.EVM().DeployContractFromFilename(ctx, filename)
 	t.Log("recursion count - ", count)
 	inputData := buildInputFromuint64(count)
@@ -158,7 +158,7 @@ func TestFEVMRecursiveDelegatecallCount(t *testing.T) {
 	ctx, cancel, client := kit.SetupFEVMTest(t)
 	defer cancel()
 
-	highestSuccessCount := uint64(235)
+	highestSuccessCount := uint64(225)
 
 	filename := "contracts/RecursiveDelegeatecall.hex"
 	recursiveDelegatecallSuccess(ctx, t, client, filename, uint64(1))
--- a/itests/gas_estimation_test.go
+++ b/itests/gas_estimation_test.go
@@ -2,16 +2,19 @@ package itests
 
 import (
 	"context"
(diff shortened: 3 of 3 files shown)


Stebalien:
Eh, steb broke this.

arajasek (approved):
The source code change looks good, but:

- Update PR title / desc to mention you're bumping the FFI
- I really do not like the noop test. It _should_ be possible to get the detailed exec trace, and pull out the `OnChainMessageInclulsion` cost, I think?
- If not, comments explaining its magic, please

Stebalien:
I'm trying to make sure that the inclusion cost is correct. `OnChainMessage` is used inside lotus to decide whether or not a message was valid for inclusion, so I want to make sure that the gas returned from that method is _actually_ sufficient.

I've extended the test to also test that:

1. If we have exactly enough gas for inclusion but try to do anything, we run out of gas.
2. If we're one gas too short for inclusion, we get rejected by lotus.

Stebalien:
Updated title and description.

chain/vm/gas.go
```diff
@@ -213,6 +213,16 @@ var Prices = map[abi.ChainEpoch]Pricelist{
...
+	build.UpgradeHyggeHeight: &pricelistV0{
+		computeGasMulti: 1,
+		storageGasMulti: 1300, // only applies to messages/return values.
+
+		onChainMessageComputeBase:    38863 + 475000, // includes the actor update cost
+		onChainMessageStorageBase:    36,
+		onChainMessageStoragePerByte: 1,
+
+		onChainReturnValuePerByte: 1,
+	},
```
raulk, line 225: I'm checking where this pricelist gets used in Lotus, because we may need to patch something for the hyperspace upgrade too (or maybe not worth it, depending on how and where it gets used).

itests/fevm_test.go
```diff
@@ -58,7 +58,7 @@ func buildInputFromuint64(number uint64) []byte {
 // recursive delegate calls that fail due to gas limits are currently getting to 229 iterations
 // before running out of gas
 func recursiveDelegatecallFail(ctx context.Context, t *testing.T, client *kit.TestFullNode, filename string, count uint64) {
-	expectedIterationsBeforeFailing := int(229)
+	expectedIterationsBeforeFailing := int(220)
```
raulk, line 61: Mild worry that this test as written is highly sensitive to gas changes. They are in flux these days as we close down the FEVM release (and will stabilise from now on), however, we will need to keep adjusting this test with further gas changes down the line.

raulk (approved):

raulk:
The test-conformance-bleeding-edge job has been removed in https://github.com/filecoin-project/lotus/pull/10189 as part of cleanup.

raulk:
Rebasing this PR on top of master, and also fixing the lint issue.

itests/gas_estimation_test.go
```diff
@@ -52,3 +60,110 @@ func TestEstimateGasNoFunds(t *testing.T) {
...
+		var err error
+		smsg, err = client.WalletSignMessage(ctx, client.DefaultKey.Address, msg)
+		require.NoError(t, err)
+		estimatedGas := vm.PricelistByEpoch(math.MaxInt).OnChainMessage(smsg.ChainLength()).Total()
+		if estimatedGas == msg.GasLimit {
+			break
+		}
+		// Try 10 times to get the right gas value.
+		require.Less(t, i, 10, "unable to estimate gas: %s != %s", estimatedGas, msg.GasLimit)
+		msg.GasLimit = estimatedGas
```
magik6k, line 113: Do we loop because setting the gas limit makes the message longer, so we need to re-re-compute the limit to get correct one?

If yes, that probably deserves a code comment.

magik6k (approved):
Changes look good, feel free to merge whenever tests pass

itests/fevm_test.go
Stebalien: > Mild worry that this test as written is highly sensitive to gas changes

It's annoying, but I'd like to keep it as it tells us when we're changing things.

> In the future, we should be OK with ratching up this number, but we absolutely cannot decrease it as it may break existing contracts.

This is the same as any other gas related thing. If you're contract makes recursive calls >200 deep, don't.

In general, changing gas values can break any contract.

itests/gas_estimation_test.go
Stebalien: Yes.

Issue #10170, opened 2023-02-01
Ethereum JSON-RPC responses should not include leading zero in hex


iand:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
release/v1.20.0
```

### Describe the Bug

According to the json-rpc schema, many places that represent values as hex must not contain a leading zero (after the `0x`). For example `0x098a0c6ef2224342b2a8727ae5111fd000000000000000000000000000000000` must be encoded as `0x98a0c6ef2224342b2a8727ae5111fd000000000000000000000000000000000`. The Go hex encoder pads to an even number of characters by including a leading zero if necessary.

The confirmance tests in https://github.com/filecoin-project/lotus/pull/10102 enforce this and I'm seeing intermittent failures

[...]

PR #10234, opened 2023-02-10
feath: eth json rpc: enforce valid hex spec

snissn:
ethereum json rpc api spec in EthUint64, EthBigInt, and EthNonce that hex data  must begin with "0x" and leading zeros ie 0x0001 are invalid hex

## Related Issues
https://github.com/filecoin-project/lotus/issues/10170

## Proposed Changes
enforce hex formatting requirements in constructors and unmarshalls for eth numbers

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/types/ethtypes/eth_types.go
+++ b/chain/types/ethtypes/eth_types.go
@@ -45,6 +45,13 @@ func (e *EthUint64) UnmarshalJSON(b []byte) error {
 	if err := json.Unmarshal(b, &s); err != nil {
 		return err
 	}
+	if !startsWith0x(s) {
+		return fmt.Errorf("EthNonce UnmarshalJSON on %s failed, must start with 0x", s)
+	}
+	if hasLeadingZeroes(s) {
+		return fmt.Errorf("EthUint64 UnmarshalJSON on %s failed, must not have a leading 0x0 before the value", s)
+	}
+
 	parsedInt, err := strconv.ParseUint(strings.Replace(s, "0x", "", -1), 16, 64)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


chain/types/ethtypes/eth_types.go
```diff
@@ -45,6 +45,13 @@ func (e *EthUint64) UnmarshalJSON(b []byte) error {
 	if err := json.Unmarshal(b, &s); err != nil {
 		return err
 	}
+	if !startsWith0x(s) {
+		return fmt.Errorf("EthNonce UnmarshalJSON on %s failed, must start with 0x", s)
+	}
+	if hasLeadingZeroes(s) {
+		return fmt.Errorf("EthUint64 UnmarshalJSON on %s failed, must not have a leading 0x0 before the value", s)
+	}
```
magik6k, line 53: I don't think it's worth enforcing, inevitably some random important library will do this wrong.

magik6k (commented):
We still need to handle the Marshal side, which is arguably more important here.

chain/types/ethtypes/eth_types.go
snissn: it feels icky at this point to merge -- but i'm following through with the ask from @jennijuju and making the PR

raulk: Agree with the assessment here. I'd rather us be more lenient on the _input_ side of things. I am checking with other clients to confirm if they apply such strictness here though.

raulk: So other popular clients are truly enforcing the 0x prefix for incoming Bigints and Uint64, and only allow the first character to be a zero when string length = 1 (i.e. no zero prefixes, indeed).

chain/types/ethtypes/eth_types_test.go
```diff
@@ -45,13 +47,29 @@ func TestEthIntUnmarshalJSON(t *testing.T) {
 		require.Equal(t, i, tc.Output)
 	}
 }
+func TestEthIntUnmarshalJSONFails(t *testing.T) {
+	testcases := []TestCase{
+		{[]byte("\"0x\""), EthUint64(0)},
+		{[]byte("\"0xabcdefghijklmnop\""), EthUint64(65)},
+		{[]byte("\"0x0400\""), EthUint64(1024)},
+		{[]byte("\"0x00000000000\""), EthUint64(1024)},
```
raulk, line 55: What are the values on `Output`? This can do without using `TestCase`?

jennijuju:
chatted with @raulk  - we agreed it seems like there is no security concern here, yet getting things more stricter now might break tooling, so  for now, i will convert this to draft PR and we can revisit next week if necessary.

PR #10239, opened 2023-02-11
eth: FIP-0055: implement final version of transitory delegated signature.

raulk:
We now only accept two methods on delegated signatures:

- `EAM#CreateExternal` on contract deployment.
- `EVM#InvokeContract` otherwise.

Support for bare sends (method 0) via Eth messages has been dropped, so as to always trigger smart contract logic via `EVM#InvokeContract`.

This PR also adds a test to verify that's the case.

Next week, we should look at cleaning up and consolidating EVM tests. They have grown organically, many private utility functions have sprouted here and there that belong in the itests kit, and I'm not happy with how we're using time to poll for transaction receipts (even though that's how Ethereum tools do it, so in theory it's the most realistic way of testing).

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/eth_account_abstraction_test.go
+++ b/itests/eth_account_abstraction_test.go
@@ -137,7 +137,7 @@ func TestPlaceholderPromotion(t *testing.T) {
 }
 
 // Tests that an placeholder turns into an EthAccout even if the message fails
-func TestEthAccountAbstractionFailure(t *testing.T) {
+func TestPlaceholderPromotionFailure(t *testing.T) {
 	kit.QuietMiningLogs()
 
 	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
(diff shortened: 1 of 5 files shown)


itests/eth_account_abstraction_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/filecoin-project/lotus/itests/kit"
 )
 
-// TestEthAccountAbstraction goes over the account abstraction workflow:
+// TestEthAccountAbstraction goes over the placeholder creation and promotion workflow:
 // - an placeholder is created when it receives a message
 // - the placeholder turns into an EOA when it sends a message
 func TestEthAccountAbstraction(t *testing.T) {
```
raulk, line 30: These tests shouldn't really be called "account abstraction" because we do not have account abstraction yet. We should rename them in a separate cleanup patch.

itests/fevm_test.go
```diff
@@ -786,3 +788,57 @@ func TestFEVMDestroyCreate2(t *testing.T) {
...
+
+	gaslimit, err := client.EthEstimateGas(ctx, ethtypes.EthCall{
+		From:  &accntEth,
+		To:    &contractEth,
+		Value: ethtypes.EthBigInt(types.FromFil(1)),
+	})
+	require.NoError(t, err)
+
+	maxPriorityFeePerGas, err := client.EthMaxPriorityFeePerGas(ctx)
+	require.NoError(t, err)
```
raulk: This logic is repeated in various places and it's worth extracting to a helper in the itests kit, which takes an *EthTxArgs, forms the EthCall, estimates gas, gets the priority fee, and fills those details in the Eth tx.

arajasek (approved):

PR #10241, opened 2023-02-11
Hyperspace nv20: create nv and add backwards compatibility with nv19

raulk:
This PR adds support for nv20 by upgrading the FVM and actors, and introduces backwards compatibility with nv19 during the pre-upgrade period to nv20. See commit log for details.


chain/types/legacy/event.go
```diff
@@ -0,0 +1,90 @@
+package legacy
```
raulk, line 1: Legacy types cordoned off in a subpackage for easier identification and maintenance.

PR #10240, opened 2023-02-11
Hyperspace nv20 staging branch

raulk:

PR #10243, opened 2023-02-12
Hyperspace nv20: upgrade and migration.

raulk:
This PR adds support for Hyperspace's network version 20.

**Depedencies-wise:**

- Upgrades FFI to its Hyperspace branch: https://github.com/filecoin-project/filecoin-ffi/tree/ntwk/hyperspace
	- Integrates the following ref-fvm branch: https://github.com/filecoin-project/ref-fvm/tree/release/hyperspace-nv19
- Adds actor bundles for nv20: https://github.com/filecoin-project/builtin-actors/releases/tag/dev%2F20230210-pre-rc.8
- Upgrades to this branch of go-state-types: https://github.com/filecoin-project/go-state-types/pull/159

**Functionality-wise:**

- Re-adds support for the legacy event type (containing no codec field).
- Adds code to adapt old EVM events to new EVM events, so that the rest of the system only needs to know about the latter.
- Maintains consensus on events in nv19 and below.
- Migrates the event index by upgrading old events to new events, backfilling the codec, removing the CBOR framing from values, rewriting topic keys, and padding topic values to 32 bytes.
- Makes all parts of the system use `ChainGetEvents` to load events. This makes it easy to centralise the "upgrade" logic when reading old events from the store.
- Retains compatibility with Delegated signatures carrying MethodSend in network versions below 20.
- On network versions below 20, still generates Delegated messages with MethodSend if no parameters are set.

**Test plan:**

See https://github.com/filecoin-project/lotus/pull/10240.


chain/events/filter/index.go
```diff
@@ -404,3 +412,118 @@ func (ei *EventIndex) PrefillFilter(ctx context.Context, f *EventFilter) error {
 
 	return nil
 }
+
+func (ei *EventIndex) RunMigration(from, to int) error {
```
raulk, line 416: I tested the migration on my local database successfully. Was able to verify that we kept returning data correctly through the Eth JSON-RPC. Considered adding a test, but because this is very temporary code I felt it wasn't worth it. Can reconsider if folks feel strongly about this.

chain/consensus/filcns/upgrades.go
```diff
@@ -1723,6 +1727,27 @@ func UpgradeActorsV11(ctx context.Context, sm *stmgr.StateManager, _ stmgr.Migra
 	return LiteMigration(ctx, bstore, newActorsManifestCid, root, oldAv, newAv, oldStateTreeVersion, newStateTreeVersion)
 }
 
+func UpgradeActorsV12(ctx context.Context, sm *stmgr.StateManager, _ stmgr.MigrationCache, _ stmgr.ExecMonitor, root cid.Cid, _ abi.ChainEpoch, _ *types.TipSet) (cid.Cid, error) {
+	oldAv := actorstypes.Version11
+	newAv := actorstypes.Version12
+	// This may change for upgrade
```
arajasek, line 1733: Will it?

arajasek (approved):
Broadly looks good to me -- didn't look too closely at the events changes, but everything else looks like this will be a successful upgrade :)

chain/messagepool/messagepool.go
```diff
@@ -794,7 +794,12 @@ func (mp *MessagePool) VerifyMsgSig(m *types.SignedMessage) error {
 		return nil
 	}
 
-	if err := chain.AuthenticateMessage(m, m.Message.From); err != nil {
+	nv, err := mp.getNtwkVersion(mp.curTs.Height())
```
Stebalien, line 797: I think we need to migrate the message pool as well. Otherwise, we'll pack invalid messages and end up with a bunch of invalid blocks.

chain/vm/fvm.go
```diff
@@ -707,3 +715,28 @@ func (r *xRedirect) MarshalCBOR(w io.Writer) error {
 
 	return nil
 }
+
+func DecodeEvents(input []byte, legacy bool) ([]cbg.CBORMarshaler, error) {
```
Stebalien, line 719: I'd suggest just passing in the network version here instead of a boolean.

Stebalien: Note: you could also just check the codec (which will be 0 for legacy events).

itests/migration_nv20_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/filecoin-project/lotus/chain/state"
+	"github.com/filecoin-project/lotus/chain/stmgr"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/chain/types/ethtypes"
+	"github.com/filecoin-project/lotus/chain/vm"
+	"github.com/filecoin-project/lotus/itests/kit"
+	"github.com/filecoin-project/lotus/node/impl"
+)
+
+func TestMigrationNV20(t *testing.T) {
```
Stebalien, line 30: We should run a test migration where:

1. A message from an ethereum account is "stuck" in the mpool (e.g., nonce issue).
2. We run the migration.
3. We unstick the message.

Right now, this will lead to a bad block including the bad message.

node/impl/full/chain.go
```diff
@@ -665,18 +665,41 @@ func (a *ChainAPI) ChainGetEvents(ctx context.Context, root cid.Cid) ([]types.Ev
 	if err != nil {
 		return nil, xerrors.Errorf("load events amt: %w", err)
 	}
+	if evtArr.Len() == 0 {
+		return nil, nil
+	}
+
+	// Hyperspace nv20: we peek into the events array to figure out if these
```
Stebalien, line 672: This is highlighting an issue with this API. I'm not sure what the right fix is, but this is going to be painful going forward if/when we change the events structure.

chain/messagepool/messagepool.go
Stebalien: Hm. I'm not sure if we can really "migrate" the message pool, as we need to handle reverts across the upgrade epoch. Thoughts @arajasek?

Stebalien: One solution would be to migrate (well, filter) the mpool when we mint a block.... but then we'd need to detect if/when that's necessary.

Stebalien: Wait, no, nvm. Re-orgs don't matter here because an older block will never be _produced_ after the migration epoch. So yes, we can filter the mpool at the migration height.

chain/consensus/filcns/upgrades.go
raulk: Probably copy pasta 🤷

chain/messagepool/messagepool.go
raulk: The activity in Hyperspace has decreased drastically. Most tipsets are packing 0-2 messages. I would support a simpler solution, like clearing the mempool and relying on wallets to resubmit messages. Ethereum wallets do this.

chain/vm/fvm.go
raulk: Checking the codec should work (and it's simpler).

chain/messagepool/messagepool.go
arajasek: Yeah, I thought about this, but I think "doing nothing" is the right thing here. IIUC, this will only affect the following messages:

1. Bare sends (method 0) from t410 addresses
1. Contract deployments with empty bytecode.

Eyeballing Filscan, there's ~1 of the first kind of message every hour. I'm not sure how many of the second kind we have, but my guess is virtually none.

raulk: What would happen in the event of a bad block in a network dominated by a single miner?

arajasek: @raulk That miner would have to manually disable its slash block filter in order to produce a new block. There's also a risk that the miner could be reported for a consensus fault, rendering it unable to mine for a period.

Stebalien: Yep. We've discussed it.

- This'll get caught when the miner tries to sync their own block.
- They can then just clear their mpool.

Stebalien (approved):

PR #10244, opened 2023-02-13
chore: market: disable lotus market by default as the service is EOL by Jan 31, 2023.

jennijuju:
As announced in 2022 November, we are deprecating lotus market support by Jan 31, 2023, and users are encouraged to migrate to lotus miner compatible deal-making focused software like boost.


jennijuju:
@magik6k will the deal-making configs impact retrievals?

PR #10257, opened 2023-02-13
fix: eth: correctly convert filecoin message <-> eth txn

Stebalien:
## Proposed Changes

1. We do allow deploying with empty initcode.
2. Make sure that the encoded "code" is non-empty, if specified.

Basically, this makes everything consistent (and it's how I specified it in the FIP).

## Checklist

- [ ] CI is green


arajasek (approved):

raulk (approved):
Can you merge this into https://github.com/filecoin-project/lotus/pull/10243 as well? Should not be hard since the only path that is changing here is the contract deployment one.

Stebalien:
Merged.

PR #10258, opened 2023-02-13
fix: stmgr: copy the message before modifying it

Stebalien:


Stebalien:
I can also make this against master as it's not critical.

magik6k (approved):

PR #10259, opened 2023-02-13
fix: stmgr: check message validity before invoking vm

Stebalien:
## Proposed Changes

Otherwise we may, e.g., try to estimate gas on a message to an f4 address before the nv18 migration.

I'm _not_ checking the "prior messages" here as this is just a sanity check.

## Checklist

- [ ] CI is green


arajasek (approved):
Seems kinda unnecessary, but no opposition to this.

PR #10264, opened 2023-02-13
fix: eth: rename yParity to v in transaction return values

Stebalien:
## Related Issues

fixes https://github.com/filecoin-project/ref-fvm/issues/1662

## Proposed Changes

The openethereum docs say it should be `yParity`, but Geth uses `v` and tools rely on `v` so we should use `v`. I've taken the liberty of fixing the conformance tests at the same time.

## Checklist

- [ ] CI is green


arajasek (approved):
LGTM, though I didn't independently confirm that `Geth uses v and tools rely on v so we should use v`

raulk (approved):
Cross-checked and confirmed to be correct.

PR #10271, opened 2023-02-14
fix: tracer: emit raw peer ids for compatibility with libp2p tracer

iand:
## Related Issues
The pubsub tracer emits traces in JSON format or posts JSON documents to an opensearch endpoint. The traces are a combination of libp2p-pubsub traces and a custom LotusTraceEvent, intermixed in the JSON stream. The libp2p traces use `[]byte` as the type for peer IDs which, in JSON, is base64 encoded. 

The LotusTraceEvent type uses a `string` for peer IDs (prettified) which is emitted as-is in the JSON.

However clients have no sensible way to distinguish the two encodings, the stream consists of a mix of the two representations:

```
{"type":11,"peerID":"ACQIARIgTM4a29NqIGdalfumxVWLOB6hZ4xm0ybAT4giXOe8/G4=","timestamp":1676302327478329014,"graft": ... 
{"type":100,"peerID":"12D3KooWEzBS8UbFGKaHGcksuteyq363zdfHCNB9ZrEzuHbWTyQy","timestamp":1676302331378019221,"peerScore ...
```

If a client attempts to decode into a struct using a `[]byte` field they will get the base64 decoded bytes of the libp2p peer IDs and the garbled base64 decoding of the pretty peer ID emitted in LotusTraceEvent

If they use a string field, they get the pretty peer ID from LotusTraceEvent and a base64 encoded string for the libp2p traces. 

Since one supported client is ElasticSearch, there is not much scope for the user to decode the peer IDs properly.

Fix this by using `[]byte` and raw peer IDs in the LotusTraceEvent to retain compatibility the [pre-existing libp2p trace field type](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/trace.pb.go#L100) 

## Proposed Changes
Change `LotusTraceEvent.PeerID` and `TraceEventPeerScore.PeerID` types to `[]byte`

## Checklist

- [ x PR title is in the form of of `<PR type>: <area>: <change being made>`
- [x] CI is green

### DIFF
--- a/node/modules/tracer/tracer.go
+++ b/node/modules/tracer/tracer.go
@@ -31,7 +31,7 @@ const (
 
 type LotusTraceEvent struct {
 	Type       pubsub_pb.TraceEvent_Type `json:"type,omitempty"`
-	PeerID     string                    `json:"peerID,omitempty"`
+	PeerID     []byte                    `json:"peerID,omitempty"`
 	Timestamp  *int64                    `json:"timestamp,omitempty"`
 	PeerScore  TraceEventPeerScore       `json:"peerScore,omitempty"`
 	SourceAuth string                    `json:"sourceAuth,omitempty"`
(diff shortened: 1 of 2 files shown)


iand:
The itest failures seem unrelated

magik6k (approved):

PR #10281, opened 2023-02-15
test: eth: make sure we can deploy a new placeholder on transfer

Stebalien:
## Related Issues

fixes https://github.com/filecoin-project/ref-fvm/issues/1670

## Proposed Changes

We have a test that triggers smart-contract logic on "transfers", but nothing that tries to create a new actor as a side-effect of a transfer.

### DIFF
--- a/itests/fevm_test.go
+++ b/itests/fevm_test.go
@@ -844,6 +844,39 @@ func TestFEVMBareTransferTriggersSmartContractLogic(t *testing.T) {
 	require.Len(t, receipt.Logs, 1)
 }
 
+// This test ensures that we can deploy new contracts from a solidity call to `transfer` without
+// exceeding the 10M gas limit.
+func TestFEVMTestDeployOnTransfer(t *testing.T) {
+	ctx, cancel, client := kit.SetupFEVMTest(t)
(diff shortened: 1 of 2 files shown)


itests/fevm_test.go
```diff
@@ -844,6 +844,36 @@ func TestFEVMBareTransferTriggersSmartContractLogic(t *testing.T) {
...
+	value := big.NewInt(100)
+	entryPoint := kit.CalcFuncSignature("sendEthToB(address)")
+	require.NoError(t, err)
+	ret, err := client.EVM().InvokeSolidityWithValue(ctx, fromAddr, contractAddr, entryPoint, params[:], value)
+	require.NoError(t, err)
+	require.True(t, ret.Receipt.ExitCode.IsSuccess())
+
+	balance, err := client.EVM().EthGetBalance(ctx, randomAddr, "latest")
+	require.NoError(t, err)
+	require.Equal(t, value.Int, balance.Int)
```
arajasek, line 873: I'd assert that the actor at randomAddr is a placeholder too.

arajasek (approved):

Issue #10048, opened 2023-01-18
Migration optimization 


jennijuju:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated improvement to propose.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Improvement Suggestion

during shark upgrade, full nodes / nodes has more chainstate with less ram on machine had trouble with migration. 

there was a couple optimization being proposed by pl netops team
- implement memory maps
- be able to opt out the (pre)migration cache
- opt in to keep the cache without batch flush

ask:
🤔 about an optimization that works best and implement it.

travisperson:
I went back and listen to our discussion to flesh out the proposed optimizations above and I think there is really only one concrete suggestion.

- implement memory maps
- opt-in to keep the cache without batch flush

These two points kind of boil down to the same thing as I was able to derive from the video notes.

We basically want to cache the intermediate statetree root after the migration completes such that on future calls to `HandleStateForks` we can look up the migrated statetree root and immediately return with the migrated statetree root. It would probably be best to avoid using the migration cache itself as that is a very large cache and ideally would be cleaned up fairly quickly after a successful migration to reduce memory usage.

Right now we could fairly easily store the migrated stateroot root on the migration structure directly. However, I think we should also think about persisting this value in the chain store as well to avoid having to redo the migration work after a restart occurs.
[...]

PR #10283, opened 2023-02-15
feat: stmgr: add env to disable premigrations

travisperson:
Setting the environment variable `LOTUS_DISABLE_PRE_MIGRATIONS=1` will discard premigrations for all upgrade.

## Related Issues
#10048

## Proposed Changes
- Add environment variable `LOTUS_DISABLE_PRE_MIGRATIONS` which when set to `1` stops pre-migrations from being executed.
- Migrations are still scheduled and part of the statement managers' migration table.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/forks.go
+++ b/chain/stmgr/forks.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"os"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -25,6 +27,9 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/vm"
 )
+
+// EnvDisablePreMigrations when set to '1' stops pre-migrations from running
+const EnvDisablePreMigrations = "LOTUS_DISABLE_PRE_MIGRATIONS"
 
 // MigrationCache can be used to cache information used by a migration. This is primarily useful to
 // "pre-compute" some migration state ahead of time, and make it accessible in the migration itself.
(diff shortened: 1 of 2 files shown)


chain/stmgr/forks.go
```diff
@@ -218,6 +223,11 @@ func runPreMigration(ctx context.Context, sm *StateManager, fn PreMigrationFunc,
 	height := ts.Height()
 	parent := ts.ParentState()
 
+	if disabled := os.Getenv(EnvDisablePreMigrations); strings.TrimSpace(disabled) == "1" {
+		log.Warn("SKIPPING pre-migration", "height", height)
+		return
+	}
+
```
travisperson, line 230: My original thought was to just set the `PreMigration` array for each migration to empty during the construction of the state manager when this environment var is set. This would lead to nothing getting scheduled.

If we had a way to print the current schedule / next scheduled task I think this would still be the correct way to go. However, I don't believe we have that functionality. So instead I'm opting to just exit pre-migrations early.

This will ensure that a log message is created around the time the pre-migration was supposed to run stating that it was skipped. I think this will make debugging things a bit easier

travisperson:
Is there any documentation in this lotus here that also should be updated as part of this PR? Otherwise I think the migration documentation I want to write will be the location of this new environment variable.

chain/stmgr/forks.go
arajasek: I think this is a pretty good way to go about it, honestly.

arajasek (approved):
LGTM, thank you!

This would be relatively easy to test -- we can spin up custom upgrade schedules in itests (eg. https://github.com/filecoin-project/lotus/blob/cdf3812e4081d35692b5ab81c573ab82d0df69cf/itests/migration_nv18_test.go#L42). I think we could have a test that simply panics on pre-migration, and succeeds on the actual migration -- we succeed if we get to the next network version.

I don't think it's critical to have this test, but worth doing.

PR #10284, opened 2023-02-15
fix: eth: correctly decode EthGetStorageAt output

Stebalien:
## Related Issues

fixes https://github.com/filecoin-project/ref-fvm/issues/1621

(probably)

## Proposed Changes

Correctly decode actor output in EthGetStorageAt, and test it.

Also:

1. Actually use the passed block param.
2. Check if the target actor is an EVM actor to avoid nonsense outputs.

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -506,13 +506,18 @@ func (a *EthModule) EthGetCode(ctx context.Context, ethAddr ethtypes.EthAddress,
 }
 
 func (a *EthModule) EthGetStorageAt(ctx context.Context, ethAddr ethtypes.EthAddress, position ethtypes.EthBytes, blkParam string) (ethtypes.EthBytes, error) {
+	ts, err := a.parseBlkParam(ctx, blkParam)
+	if err != nil {
+		return nil, xerrors.Errorf("cannot parse block param: %s", blkParam)
+	}
+
 	l := len(position)
 	if l > 32 {
 		return nil, fmt.Errorf("supplied storage key is too long")
 	}
 
 	// pad with zero bytes if smaller than 32 bytes
-	position = append(make([]byte, 32-l, 32-l), position...)
+	position = append(make([]byte, 32-l), position...)
 
 	to, err := ethAddr.ToFilecoinAddress()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


node/impl/full/eth.go
```diff
@@ -525,6 +530,20 @@ func (a *EthModule) EthGetStorageAt(ctx context.Context, ethAddr ethtypes.EthAdd
...
+			return nil, nil
+		}
+		return nil, xerrors.Errorf("failed to lookup contract %s: %w", ethAddr, err)
+	}
+
+	// Not a contract. Technically, we should return an empty slot... but an error is more user
+	// friendly, IMO.
+	if !builtinactors.IsEvmActor(actor.Code) {
+		return nil, xerrors.Errorf("actor is not an ethereum contract")
+	}
```
raulk, line 545: This will break libraries. Eth libraries expect a zero answer for anything that is not a smart contract OR for storage keys that don't exist.

Stebalien: Got it.

node/impl/full/eth.go
```diff
@@ -506,13 +506,18 @@ func (a *EthModule) EthGetCode(ctx context.Context, ethAddr ethtypes.EthAddress,
...
+
 	l := len(position)
 	if l > 32 {
 		return nil, fmt.Errorf("supplied storage key is too long")
 	}
 
 	// pad with zero bytes if smaller than 32 bytes
-	position = append(make([]byte, 32-l, 32-l), position...)
+	position = append(make([]byte, 32-l), position...)
```
raulk, line 520: supernit: Could move this below to where it's used.

raulk (commented):
A dedicated test would be nice that goes over the different cases here:

- Actor is an evm smart contract and has the storage slot filled => return value.
- Actor is an evm smart contract and does not have the storage slot filled => return 32-byte padded 0.
- Actor is not an evm smart contract => ^^
- Actor does not exist => ^^

(I volunteer to write it tomorrow if you'd like me to.)

raulk (approved):
Other than the above, LGTM now.

Stebalien:
Test issue: https://github.com/filecoin-project/ref-fvm/issues/1671

node/impl/full/eth.go
Stebalien: I have that there to put it next to the length check. I kept the length check at the top to fail early when passed an invalid parameter.

jennijuju (approved):

PR #10286, opened 2023-02-16
feat: EthAPI: Add EthAddressToFilecoinAddress

### LINKED ISSUE #10279: Expose API to convert between eth address and filecoin address
### Checklist

### Lotus component

- [ ] lotus client
- [X] lotus JSON-RPC API
- [ ] Other

folks like exchanges might need an easy api to call get the f410 addresss of a 0x(ETH Address)

add an api takes an 0x ETH Address and return f410 -see `./lotus evm stat`
(may consider the other way around too tho can't think of a use case yet)

_No response_

### Additional context

_No response_

arajasek:
## Related Issues

Fixes #10279 

## Proposed Changes

Simple API that users can call to go from EthAddresses to Filecoin Addresses

## Checklist

- [ ] CI is green


itests/eth_api_test.go
```diff
@@ -0,0 +1,35 @@
+package itests
+
```
shrenujbansal, line 2: Don't have a strong opinion on this but maybe move this to fevm_address_test.go and rename that file to eth_address_test.go since that name does cover this domain

itests/eth_api_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/filecoin-project/lotus/itests/kit"
+)
+
+func TestEthAddressToFilecoinAddress(t *testing.T) {
+	client, _, _ := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	secpKey, err := key.GenerateKey(types.KTDelegated)
```
shrenujbansal, line 22: Is this still a secpKey since the type here is delegated?

itests/eth_api_test.go
```diff
@@ -0,0 +1,35 @@
...
+	filecoinAddr, err := client.WalletImport(ctx, &secpKey.KeyInfo)
+	require.NoError(t, err)
+
+	ethAddr, err := ethtypes.EthAddressFromFilecoinAddress(filecoinAddr)
+	require.NoError(t, err)
+
+	apiFilAddr, err := client.EthAddressToFilecoinAddress(ctx, ethAddr)
+	require.NoError(t, err)
+
+	require.Equal(t, filecoinAddr, apiFilAddr)
```
shrenujbansal, line 34: Could you enhance this test or add another test, to test the conversion for non delegated addresses?

itests/eth_api_test.go
arajasek: I think we need to have a dedicated test suite for the EthAPI, my hope is that this will be the start of that.

itests/eth_api_test.go
arajasek: Yes, "Delegated" signatures are solely secp.

itests/eth_api_test.go
shrenujbansal: Fair point but the Eth API seems very large to stick in a single file?

itests/eth_api_test.go
shrenujbansal: For my own knowledge,
We also have a secp address type which I presume have secp keys
What is the mapping of address types to key types? Are those orthogonal concepts?

shrenujbansal (approved):

itests/eth_api_test.go
arajasek: So this is a little gnarly. Secp addresses correspond to secp keys. Delegated addresses _today_ solely correspond to secp keys. They will eventually support other kinds of signatures, such as "account abstraction".

So why do we have the KTDelegated sigtype today? It's because the Lotus keystore couples the stored key with its address representation. So when working with keys, we need to treat delegated sigtypes differently from secp sigtypes (eg. [here](https://github.com/filecoin-project/lotus/blob/cdf3812e4081d35692b5ab81c573ab82d0df69cf/chain/wallet/key/key.go#L55-L69)).

PR #10294, opened 2023-02-16
fix: pack: support network name overrides in bundle git tags

snissn:
## Related Issues
https://github.com/filecoin-project/lotus/issues/10185#issuecomment-1422864836

## Proposed Changes
add support for overrides 

## Checklist

- [ ] CI is green

### DIFF
--- a/gen/bundle/bundle.go
+++ b/gen/bundle/bundle.go
@@ -46,9 +46,9 @@ func main() {
 	}
 
 	var metadataSpecificVersion []*build.BuiltinActorsMetadata
-
 	// see ./build/actors/pack.sh
-	// expected args are git bundle tag then number of per network overrides
+  // expected args are:
+	// $(GOCC) run ./gen/bundle $(VERSION) $(RELEASE) $(RELEASE_OVERRIDES)
 	// overrides are in the format network_name=override
 
 	if len(os.Args) > 1 {
@@ -67,7 +67,6 @@ func main() {
 				} else {
 					m.BundleGitTag = os.Args[2]
 				}
-				fmt.Println("hi")
 				metadataSpecificVersion = append(metadataSpecificVersion, m)
 			}
 		}
--- a/gen/bundle/bundle.go
+++ b/gen/bundle/bundle.go
@@ -47,7 +47,7 @@ func main() {
 
 	var metadataSpecificVersion []*build.BuiltinActorsMetadata
 	// see ./build/actors/pack.sh
-  // expected args are:
+	// expected args are:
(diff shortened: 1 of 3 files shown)


arajasek (commented):
I think there's still one piece of information that needs to be passed through, which is the actors version being updated. As currently written, we'll respect the network overrides, but still set _all actor versions_ to the git tag, not just the one being packed.

As an example, `./pack.sh v9 dev/20230114-pre-rc.2 mainnet=v9.0.3 calibrationnet=v9.0.0 devnet=v9.0.3 devnet-wasm=v9.0.3 testing=v9.0.3 testing-fake-proofs=v9.0.3 butterflynet=v9.0.2` should only update v9 actors.

gen/bundle/bundle.go
```diff
@@ -32,19 +34,46 @@ var EmbeddedBuiltinActorsMetadata []*BuiltinActorsMetadata = []*BuiltinActorsMet
...
-	// TODO: Re-enable this when we can set the tag for ONLY the appropriate version
-	// https://github.com/filecoin-project/lotus/issues/10185#issuecomment-1422864836
-	//if len(os.Args) > 1 {
-	//	for _, m := range metadata {
-	//		m.BundleGitTag = os.Args[1]
-	//	}
-	//}
+	var metadataSpecificVersion []*build.BuiltinActorsMetadata
+
+	// see ./build/actors/pack.sh
```
arajasek, line 50: comment needs updating to include the actors version

gen/bundle/bundle.go
```diff
@@ -32,19 +34,46 @@ var EmbeddedBuiltinActorsMetadata []*BuiltinActorsMetadata = []*BuiltinActorsMet
...
+		for _, m := range metadata {
+			if strings.HasPrefix(version, fmt.Sprintf("v%d", m.Version)) {
+				// correct version
+				override, ok := overrides[m.Network]
+				if ok {
+					m.BundleGitTag = override
+				} else {
+					m.BundleGitTag = os.Args[2]
+				}
+				fmt.Println("hi")
```
arajasek, line 70: hi

arajasek (commented):
Only issue I see with this is that it's gonna skip the metadata for those versions we _aren't_ setting, leading to them getting overwritten.

I think we can just drop the `metadataSpecificVersion`, and mutate the `metadata` object in place. That will modify it as needed, while leaving everything else untouched.

gen/bundle/bundle.go
snissn: hi 😆 🙊

arajasek (approved):

PR #10295, opened 2023-02-16
feat: eth: parse revert data

Stebalien:
## Proposed Changes

Parse the revert data and include it in the error message of EthCall to improve degradability. We don't really want to do this in the FVM because it's Ethereum (solidity) specific, but this makes sense to do in the Ethereum API. It might also make sense to do in the EVM itself, but it's too late to consider that right now.

This change only applies to `EthCall`.  The next step is to improve the error message output from `EthEstimateGas`.

## Additional Info

See https://docs.soliditylang.org/en/latest/control-structures.html#panic-via-assert-and-error-via-require

## Checklist

- [x] CI is green


node/impl/full/eth.go
```diff
@@ -363,6 +363,19 @@ func (a *EthModule) EthGetTransactionCount(ctx context.Context, sender ethtypes.
...
+	if err != nil {
+		if xerrors.Is(err, types.ErrActorNotFound) {
+			return 0, nil
+		}
+		return 0, xerrors.Errorf("failed to lookup contract %s: %w", sender, err)
+	}
+
+	switch {
+	case builtinactors.IsAccountActor(actor.Code), builtinactors.IsEthAccountActor(actor.Code):
+
```
raulk, line 376: Looks like leftover code?

Stebalien: Uh, yes.

node/impl/full/eth.go
```diff
@@ -2180,6 +2181,92 @@ func parseEthTopics(topics ethtypes.EthTopicSpec) (map[string][][]byte, error) {
...
+// See https://docs.soliditylang.org/en/latest/control-structures.html#panic-via-assert-and-error-via-require
+func parseEthRevert(ret []byte) string {
+	if len(ret) == 0 {
+		return "none"
+	}
+	var cbytes abi.CborBytes
+	if err := cbytes.UnmarshalCBOR(bytes.NewReader(ret)); err != nil {
+		return "ERROR: revert reason is not cbor encoded bytes"
+	}
+	if len(cbytes) == 0 {
```
shrenujbansal, line 2199: In what scenario would you run into this and not 2192?

node/impl/full/eth.go
```diff
@@ -2180,6 +2181,92 @@ func parseEthTopics(topics ethtypes.EthTopicSpec) (map[string][][]byte, error) {
...
+			codeInt := big.PositiveFromUnsignedBytes(cbytes)
+			return fmt.Sprintf("Panic(%s)", ethtypes.EthBigInt(codeInt).String())
+		}
+		switch code {
+		case 0x00:
+			return "Panic()"
+		case 0x01:
+			return "Assert()"
+		case 0x11:
+			return "ArithmaticOverflow()"
```
shrenujbansal, line 2222: *Arithmetic

node/impl/full/eth.go
Stebalien: 2192 means that the contract returned nothing. In this case, the contract returned an empty cbor byte string.

Really, we should _only_ see the second case (right now). We should optimize contracts to return "nothing" in this case, but we haven't.

Stebalien: I.e. `void != []`

node/impl/full/eth.go
```diff
@@ -2180,6 +2181,92 @@ func parseEthTopics(topics ethtypes.EthTopicSpec) (map[string][][]byte, error) {
...
+	case panicFunctionSelector:
+		cbytes := cbytes[4 : 4+32]
+		// Read the and check the code.
+		code, err := ethtypes.EthUint64FromBytes(cbytes)
+		if err != nil {
+			// If it's too big, just return the raw value.
+			codeInt := big.PositiveFromUnsignedBytes(cbytes)
+			return fmt.Sprintf("Panic(%s)", ethtypes.EthBigInt(codeInt).String())
+		}
+		switch code {
```
shrenujbansal, line 2216: Where are the below codes defined?
Stick these in a const map and return that?

Stebalien: Link above, but you're right. We can just use a map. I'll make one.

node/impl/full/eth.go
shrenujbansal: ok, so something like
`revert() vs revert("")`

itests/fevm_test.go
```diff
@@ -868,3 +868,32 @@ func TestFEVMGetBlockDifficulty(t *testing.T) {
...
+
+	_, contractAddr := e.DeployContractFromFilename(ctx, "contracts/Errors.hex")
+	contractAddrEth, err := ethtypes.EthAddressFromFilecoinAddress(contractAddr)
+	require.NoError(t, err)
+	for sig, expected := range map[string]string{
+		"failRevertEmpty()":  "none",
+		"failRevertReason()": "Error(my reason)",
+		"failAssert()":       "Assert()",
+		"failDivZero()":      "DivideByZero()",
+		"failCustom()":       "0x09caebf3",
```
shrenujbansal, line 886: How do you get "0x09caebf3"?
maybe add a comment on the side to explain

shrenujbansal (approved):

itests/fevm_test.go
Stebalien: I'm now calculating it inline.

PR #10298, opened 2023-02-17
feat: eth: return revert data on failed gas estimation

Stebalien:
## Proposed Changes

Return revert data on failed gas estimation.

Unfortunately, we need to execute the message twice to get this (unless we want to change some APIs). But it's unlikely to be a performance issue and will definitely help people debug failures.

## Checklist

- [x] CI is green

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -888,7 +888,7 @@ func (a *EthModule) EthEstimateGas(ctx context.Context, tx ethtypes.EthCall) (et
 	msg.GasLimit = 0
 
 	ts := a.Chain.GetHeaviestTipSet()
-	msg, err = a.GasAPI.GasEstimateMessageGas(ctx, msg, nil, ts.Key())
+	gassedMsg, err := a.GasAPI.GasEstimateMessageGas(ctx, msg, nil, ts.Key())
 	if err != nil {
 		// On failure, GasEstimateMessageGas doesn't actually return the invocation result,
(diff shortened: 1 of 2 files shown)


node/impl/full/eth.go
```diff
@@ -890,6 +890,15 @@ func (a *EthModule) EthEstimateGas(ctx context.Context, tx ethtypes.EthCall) (et
 	ts := a.Chain.GetHeaviestTipSet()
 	msg, err = a.GasAPI.GasEstimateMessageGas(ctx, msg, nil, ts.Key())
 	if err != nil {
+		// On failure, GasEstimateMessageGas doesn't actually return the invocation result,
+		// it just returns an error. That means we can't get the revert reason.
+		//
+		// So we re-execute the message with EthCall (well, applyMessage which contains the
+		// guts of EthCall). This will give us an ethereum specific error with revert
+		// information.
+		if _, err2 := a.applyMessage(ctx, msg, ts.Key()); err2 != nil {
```
shrenujbansal, line 899: Can we get a test for this?

Stebalien: The main logic should already be covered by the test in the previous patch. Although I guess I can just extend that test to call this as well?

Stebalien: Oh, well, of course. It panics. Because I didn't write a test.

Stebalien: Always the simple patches that can't possibly be wrong.

shrenujbansal (approved):

Issue #10255, opened 2023-02-13
ETH API: EthGetTransactionCount incorrect for non accounts


Stebalien:
EthGetTransactionCount needs to do a bit more special casing:

1. For accounts, it correctly uses the account's nonce.
2. For EVM contracts, it needs to ask the EVM contract for its _internal_ nonce.
3. For other native actors, it should always return 1.

jennijuju:
maybe rc2 (confirm with yushi)

PR #10300, opened 2023-02-17
fix: eth: return the correct nonce from EthGetTransactionCount

Stebalien:
## Related Issues

fixes https://github.com/filecoin-project/lotus/issues/10255

## Proposed Changes

In EthGetTransactionCount, get the "correct" nonce for EVM contracts in case someone is using this method to, e.g., predict CREATE addresses.

EVM contracts track this number internally.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/actors/builtin/evm/evm.go
+++ b/chain/actors/builtin/evm/evm.go
@@ -48,6 +48,7 @@ type State interface {
 	cbor.Marshaler
 
 	Nonce() (uint64, error)
+	IsAlive() (bool, error)
 	GetState() interface{}
 
 	GetBytecode() ([]byte, error)
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -31,6 +31,7 @@ import (
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain/actors"
 	builtinactors "github.com/filecoin-project/lotus/chain/actors/builtin"
+	builtinevm "github.com/filecoin-project/lotus/chain/actors/builtin/evm"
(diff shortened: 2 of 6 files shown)


chain/actors/builtin/evm/evm.go
```diff
@@ -48,6 +48,7 @@ type State interface {
 	cbor.Marshaler
 
 	Nonce() (uint64, error)
+	IsAlive() (bool, error)
```
shrenujbansal, line 51: I don't see an implementation anywhere? Is this the right PR?

node/impl/full/eth.go
```diff
@@ -367,6 +368,29 @@ func (a *EthModule) EthGetTransactionCount(ctx context.Context, sender ethtypes.
 		return ethtypes.EthUint64(0), xerrors.Errorf("cannot parse block param: %s", blkParam)
 	}
 
+	// First, handle the case where the "sender" is an EVM actor.
```
shrenujbansal, line 371: this comment should be on line 380?

chain/actors/builtin/evm/evm.go
shrenujbansal: nvm

node/impl/full/eth.go
Stebalien: It's the entire section. I've put it into a block to make that clear.

shrenujbansal (approved):

node/impl/full/eth.go
Stebalien: Eh, I switched it to an if chain.

arajasek (approved):

PR #10309, opened 2023-02-17
fix: eth: eth_subscribe tag

Stebalien:
This fixes some errors caught by `go vet`.

1. Fixes some lints that were previously being ignored (because `go vet` doesn't actually have a built-in way to ignore errors).
2. Fixes the formatting of multiple JSON-RPC API struct tags.

However, I'm not sure why this matters and/or why we even have this tag.

### DIFF
--- a/api/proxy_gen.go
+++ b/api/proxy_gen.go
@@ -117,7 +117,7 @@ type EthSubscriberStruct struct {
 }
 
 type EthSubscriberMethods struct {
-	EthSubscription func(p0 context.Context, p1 jsonrpc.RawParams) error `notify:"true"rpc_method:"eth_subscription"`
+	EthSubscription func(p0 context.Context, p1 jsonrpc.RawParams) error `notify:"true" rpc_method:"eth_subscription"`
 }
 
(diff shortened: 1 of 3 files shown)


magik6k:
> I'm not sure why this matters and/or why we even have this tag.

It's probably complaining about the formatting for style reasons - those tags actually parse fine without the space. Also we need them on the proxy struct as it's where the jsonrpc client logic will pull them from. Notif makes the call be a notification, and rpc_method sets a custom rpc method name, which is important for subscription notification

magik6k (approved):

Stebalien:
> rpc_method sets a custom rpc method name, which is important for subscription notification

Except it's the _only_ user of `rpc_method`?

magik6k:
We could, and probably should rpc_method on other eth methods, but it is technically only necessary on eth subscription callback, as that is a call made by lotus back to the client, so that one must have the correct method name.

Other eth methods don’t really need to be called as eth_ if they are called on lotus rpc, but it would be technically more correct if lotus used eth_

PR #10344, opened 2023-02-24
fix: cli: send with InvokeEVM when sending from an eth account

Stebalien:
## Proposed Changes

In the `lotus send` cli, replace method 0 with InvokeEVM when the sender is an ethereum-style account.

This will make `lotus send` "work", mostly. However, there are some
caveats:

1. Sending to, e.g., a storage provider will still fail (doesn't implement InvokeEVM).
2. We're changing the method number specified by the user.
3. This will end up running code.

## Checklist

- [x] CI is green

### DIFF
--- a/cli/send_test.go
+++ b/cli/send_test.go
@@ -7,14 +7,16 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	ucli "github.com/urfave/cli/v2"
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/builtin"
 
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/chain/types/ethtypes"
 )
 
 func mustAddr(a address.Address, err error) address.Address {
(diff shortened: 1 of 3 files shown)


arajasek (commented):
No, I don't like this. I think we want to:

- fail if the user supplies any methodNum when sending from f4
- sub in the `InvokeContract` MethodNum only for gas estimation
- everything else should "just work"

Stebalien:
(still need to write some tests)

cli/send_test.go
```diff
@@ -65,7 +67,51 @@ func TestSendCLI(t *testing.T) {
...
+				To:     mustAddr(address.NewIDAddress(1)),
+				Val:    oneFil,
+				Method: builtin.MethodsEVM.InvokeContract,
+				Params: paramsBuf.Bytes(),
+			}).Return(arbtProto, nil),
+			mockSrvcs.EXPECT().PublishMessage(gomock.Any(), arbtProto, false).
+				Return(sigMsg, nil, nil),
+			mockSrvcs.EXPECT().Close(),
+		)
+		err = app.Run([]string{"lotus", "send", "--from-eth-addr", testEthAddr.String(), "--params-hex", "01020304", "f01", "1"})
```
Stebalien, line 113: Note: this _won't_ work if the the default address is an eth address, because the command won't see that. Is that something we want to support?

jennijuju: (nope? lotus doesnt support import eth address to wallet yet, would be nice to have later for us to test but out of scope of this pr imho

arajasek (approved):

Issue #9407, opened 2022-10-03
GET_xx_MAX_CONCURRENT env variable is not being enforced


rjan90:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon:  1.17.2-rc2+mainnet+git.3d23a54fe+api1.5.0
Local: lotus version 1.17.2-rc2+mainnet+git.3d23a54fe
```

### Describe the Bug

The `GET_XX_MAX_CONCURRENT=` enviorment variable is not being enforced between a AP-worker and a PC1-worker. I also tried exporting all the different GET_size_MAX_CONCURRENT enviroments, just to rule out if it was similar [to this issue](https://github.com/filecoin-project/lotus/issues/9213) where the sealer doesn't set the sector seal proof type on the scheduler request.

[...]

Issue #8566, opened 2022-04-28
Scheduler does not honour available disk space on long term storage


RobQuistNL:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
1.15.2-rc2
```

### Describe the Bug

When running a lotus-worker with all flags on false - so we just have these tasks;

```golang
taskTypes = append(
  taskTypes, sealtasks.TTFetch, 
  sealtasks.TTCommit1, 
  sealtasks.TTProveReplicaUpdate1, 
  sealtasks.TTFinalize, 
  sealtasks.TTFinalizeReplicaUpdate
)
```
and just have long-term storage attached - get mass assigned GETs, even though there are free workers with more free diskspace.

### Logging Information

```text
-
```

### Repo Steps

1. Run worker on storage machine
[...]

PR #10356, opened 2023-02-27
feat: sched: Assigner experiments

magik6k:
## Proposed Changes
Currently lotus-miner can be configured with one of two assigners:
* `utilization` (default) - For each task pick a worker with a lowest 'utilization factor', which is based on ratios of compute resources
* `spread` - In each assign loop try to assign to a worker with least tasks assigned so far

This PR adds a few experimental assigners:
* `experiment-spread-qcount` - Like spread, but also takes into account task counts which are running/preparing/queued
* `experiment-spread-tasks` -  Like spread, but counts running tasks on a per-task-type basis
* `experiment-spread-tasks-qcount` - The two above combined - count running tasks grouped by task type, also takes into account tasks which are running/preparing/queued
* `experiment-random` - In each schedule loop figure a set of all workers which can handle the task.. then pick a random one

## Additional Info
* [x] Would be good to try all of those and see how they behave
* [x] Interested to hear if `experiment-random` / `experiment-spread-tasks-qcount` work better for `--no-default` storage workers
* [x] Interested to hear additional ideas

## Checklist

- [x] CI is green

### DIFF
--- a/storage/sealer/manager.go
+++ b/storage/sealer/manager.go
@@ -289,14 +289,20 @@ func (m *Manager) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	m.remoteHnd.ServeHTTP(w, r)
 }
 
-func schedNop(context.Context, Worker) error {
-	return nil
+var schedNop = PrepareAction{
+	Action: func(ctx context.Context, w Worker) error {
+		return nil
+	},
+	PrepType: sealtasks.TTNoop,
 }
 
-func (m *Manager) schedFetch(sector storiface.SectorRef, ft storiface.SectorFileType, ptype storiface.PathType, am storiface.AcquireMode) func(context.Context, Worker) error {
-	return func(ctx context.Context, worker Worker) error {
-		_, err := m.waitSimpleCall(ctx)(worker.Fetch(ctx, sector, ft, ptype, am))
-		return err
+func (m *Manager) schedFetch(sector storiface.SectorRef, ft storiface.SectorFileType, ptype storiface.PathType, am storiface.AcquireMode) PrepareAction {
+	return PrepareAction{
+		Action: func(ctx context.Context, worker Worker) error {
+			_, err := m.waitSimpleCall(ctx)(worker.Fetch(ctx, sector, ft, ptype, am))
+			return err
+		},
+		PrepType: sealtasks.TTFetch,
 	}
 }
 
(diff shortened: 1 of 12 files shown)


rjan90:
> Interested to hear if experiment-random / experiment-spread-tasks-qcount work better for --no-default storage workers

Ran som tests for these [here](https://github.com/filecoin-project/lotus/issues/8566#issuecomment-1446978856). And these definetly make things way better for --no-default workers (not every sector assign to a single storage worker).

> Interested to hear additional ideas

Only other idea I hear been floating around is a round robin throught the list of workers, but unsure if that would be possible with the current limitations of the built-in scheduler. I do however think that the random assigner with a possible fix for https://github.com/filecoin-project/lotus/issues/9407#issuecomment-1446945462, will be sufficient for spreading these sectors around.

magik6k:
Pushed a commit which implements the fix for https://github.com/filecoin-project/lotus/issues/9407#issuecomment-1446945462.

RobQuistNL:
The one and only scheduler implementation I'd love to have would be "round-robin". For example, say our setup has this layout:

```text
MachineA
       |- AP_1
       |- AP_2
       \- PC1_1

MachineB
       |- AP_3
       |- AP_4
       \- PC1_2
       
MachineC
       |- PC2_1
       \- C2_1

MachineD
       |- PC2_2
       \- C2_2
```

When we have the current scheduler, it would assign AP to the AP workers, decently spread. AP workers take work as they can. When we say the PC1 worker prefers work from their local disk, you end up with PC1_1 getting a lot of work. If we pledge every 4 minutes;

- AP_1 gets an AP job (Sector0), finishes it in 1 minute.
- PC1_1 gets Sector0 because its file is local

- AP_1 gets an AP job (Sector1), finishes it in 1 minute.
- PC1_1 gets Sector1 because its file is local

etc..

Only once both AP_1 and AP_2 are doing things, will PC1_2 receive jobs, because then the AP will be happening on MachineB.

Eventually:
PC1_1 is finally full with the amount of PC1 jobs that it can do;
AP_1 still gets the jobs.
PC1_2 (MachineB) will have to fetch data from AP_1 (MachineA)

This generates a lot of network strain.

Best solution: round-robin the workers.

Keep a list of workers, keep track of "assigned job count"

```
AP_1 | 0
AP_2 | 0
PC1_1 | 0
PC1_2 | 0
```

Sort the list by descending job count, assign a job, increment the job counter.

You will end up with (no matter the time / spread) all AP_1 machines getting an even amount of jobs, thus the PC1 workers getting an even amount of jobs, as well as the PC2 / C2 workers. 

This will hold up if you pledge every 10 seconds, of every 50 minutes.
This will also prevent any worker from being overloaded, and it allows for PC1 to prefer local AP work.

In the case a machine has been selected by round-robin but it doesn't have the resources (this should never happen in case of similar hardware / times), you can always skip over it and take the next one in the list. It will end up with a skewed list where the "overloaded" worker always comes first due to a low job count, but this shouldn't matter in my opinion.

ZenGround0 (approved):
I skimmed this though this part of the system is probably the most foreign to me so not more than that.  I can go over this with you in more depth later this week if you want a more extensive review.  Approving now in the interest of getting in by code freeze.

hyunmoon:
lotus-miner started crashing (twice so far, once a day) since I strated using the code on master, and using experiment-spread-tasks-qcount.
The scheduler is was working as intended which is excellent, and I can't say for sure what caused the crash but I've decided to step back to a released version for now.

PR #10382, opened 2023-03-03
Update to go-data-transfer v2 and libp2p, still wip

hannahhoward:
# Goals

Update to the new version of go-data-transfer to unblock other deps updates

# Status

- needs libp2p vet
- stuck on update go-index-provider


node/modules/lp2p/rcmgr.go
```diff
@@ -111,10 +113,12 @@ func ResourceManager(connMgrHi uint) func(lc fx.Lifecycle, repo repo.LockedRepo)
 		if err != nil {
 			return nil, fmt.Errorf("error creating resource manager stats reporter: %w", err)
 		}
-		err = view.Register(obs.DefaultViews...)
-		if err != nil {
-			return nil, fmt.Errorf("error registering rcmgr metrics: %w", err)
-		}
+
+		// TODO: Libp2p converted to native prometheus metrics -- how do we register them
```
MarcoPolo, line 117: Looks like you're using the [default prometheus register](https://github.com/filecoin-project/lotus/blob/4588523de72452355fb6338176dbab6e69c5853a/cmd/lotus-stats/main.go#L159). You should be able to just do: https://github.com/libp2p/go-libp2p/blob/master/dashboards/resource-manager/README.md

e.g. `rcmgrObs.MustRegisterWith(prometheus.DefaultRegisterer)`

node/modules/lp2p/rcmgr.go
```diff
@@ -52,37 +50,41 @@ func ResourceManager(connMgrHi uint) func(lc fx.Lifecycle, repo repo.LockedRepo)
 		// For every extra 1GB of memory we have available, increase our limit by 1GiB
 		defaultLimits.SystemLimitIncrease.Memory = 1 << 30
 		defaultLimitConfig := defaultLimits.AutoScale()
-		if defaultLimitConfig.System.Memory > 4<<30 {
+
+		changes := rcmgr.PartialLimitConfig{}
```
MarcoPolo, line 54: The diff in this file looks good to me for the rcmgr type changes. Sorry for the work, but hopefully it removes a foot gun.

chain/beacon/drand/drand_test.go
```diff
@@ -24,7 +24,7 @@ func TestPrintGroupInfo(t *testing.T) {
 	})
 	chain, err := cg.FetchChainInfo(nil)
 	assert.NoError(t, err)
-	err = chain.ToJSON(os.Stdout)
+	err = chain.ToJSON(os.Stdout, nil)
```
marten-seemann, line 27: Why use semver and then make breaking changes in a minor release? 🤬

node/modules/lp2p/rcmgr.go
marten-seemann: Done.

marten-seemann: Actually, that doesn't work. It makes the tests panic, presumable when multiple hosts are started: https://app.circleci.com/pipelines/github/filecoin-project/lotus/27088/workflows/e4e8800d-6d83-4e2f-950f-3c619de1866a/jobs/874717/steps

MarcoPolo: ah right during a test you may spin up multiple of these.

marten-seemann: I added a `sync.Once`. Maybe not the nicest solution, but this should help us unblock this PR.

magik6k (approved):

Issue #10386, opened 2023-03-03
Lotus Hotstore Badger GC 


ZenGround0:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

While investigating splitstore its become clear that lots of hotstore bloat comes in the form of non-GCed dead badger state.  Moving GC will handle this but at large disk overhead.  During normal operation it will likely make sense to run online GC of the hotstore at low thresholds to catch up on large amounts of state bloat.  

### Describe the solution you'd like

Cmd line tool for running badger gc 

### Additional context

Ideally this solution will allow for incremental GCOnce badger GC or periodic run through vlog until no rewrites GC as happens today after online compactions

PR #10353, opened 2023-02-27
Remove flatten

ZenGround0:
## Proposed Changes

Experimental patch to reduce online GC costs of splitstore by removing Flatten before doing vlog GC.

## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #10387, opened 2023-03-03
feat:splitstore:Badger GC of hotstore command

ZenGround0:
## Related Issues

#10386

## Proposed Changes

`chain prune hot` implements badger online GC of hotstore in user driven way

## Additional Info

In addition to making the new feature this PR:
* Limits total number of flatten workers to 7 as the conclusion of investigation started in #10353 
* Respects context during garbage collection to allow for better shutdown behavior for very low GC thresholds
* Rename existing `chain prune` to `chain prune compact-cold` to make name more descriptive 

Unsure about
* I bumped lotus api version number, not sure if that's correct

## Checklist

- [ ] CI is green


api/version.go
```diff
@@ -55,7 +55,7 @@ func VersionForType(nodeType NodeType) (Version, error) {
 // semver versions of the rpc api exposed
 var (
 	FullAPIVersion0 = newVer(1, 5, 0)
-	FullAPIVersion1 = newVer(2, 3, 0)
+	FullAPIVersion1 = newVer(2, 4, 0)
```
magik6k, line 58: The api version shouldn't be changed unless there are actual breaking changes which may break existing things which depend on the API.

Bumping this number means that everyone needs to also update lotus-miner / boost instances, which is really annoying (especially boost, which needs to ship a version supporting the new api)

magik6k (commented):
Splistore-related changes look good.

API version really shouldn't be bumped unless really necessary.

magik6k (approved):

Issue #8561, opened 2022-04-28
lotus-miner proving compute window-post shows weird results


RobQuistNL:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
1.15.2-rc1
```

### Describe the Bug

Running `lotus-miner proving compute window-post 0` yields weird results. I assumed "skipped" would contain skipped sector numbers. This bunch of random numbers makes no sense to me. Can't be sure its not sector numbers because of https://github.com/filecoin-project/lotus/issues/8560 though.

### Logging Information

```text
$ lotus-miner proving compute window-post 0
Took 3m35.236895707s
[...]

PR #10390, opened 2023-03-04
chore: make debugging windowPoSt-failures human readable

beck-8:
## Related Issues

https://github.com/filecoin-project/lotus/issues/8561
## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus-miner/proving.go
+++ b/cmd/lotus-miner/proving.go
@@ -19,6 +19,7 @@ import (
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/proof"
 
 	"github.com/filecoin-project/lotus/blockstore"
 	"github.com/filecoin-project/lotus/build"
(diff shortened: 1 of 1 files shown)


cmd/lotus-miner/proving.go
```diff
@@ -640,7 +641,42 @@ It will not send any messages to the chain.`,
...
+				}
+				postParam.Partitions[id].Skipped = sectors
+			}
+			postParam.Proofs = i.Proofs
+			postParam.ChainCommitEpoch = i.ChainCommitEpoch
+			postParam.ChainCommitRand = i.ChainCommitRand
+			postParams = append(postParams, postParam)
+		}
+
+		jr, err := json.Marshal(postParams)
```
magik6k, line 679: I would just output in a human readable format. This was outputting json because that was the simplest way to have some useful debug output.

beck-8: Sorry, I didn't understand what you meant. You mean it's not easy enough to use is it now?

beck-8: I think this is enough, if there is a more elegant way, please suggest it and I will learn.
```
lotus-miner proving compute window-post 0|grep -v Took|jq .
or
lotus-miner proving compute window-post 0|grep -v Took|jq .[].Partitions[].Skipped
```

magik6k: It is easy now, but I'm not sure if the output should be json as most commands output data in more human-readable form (and some commands have --json for json output)

e.g.
```
$ lotus-miner sectors status 1715                                                                                                                                                                                                                                          SectorID:	1715
Status:		Proving
CIDcommD:	baga6ea4seaqao7s73y24kcutaosvacpdjgfe5pw76ooefnyqw4ynr3d2y6x2mpq
[...]
```

I guess at least this json could be output with json.MarshalIndent.

Note that this is purely for consistency with other commands, there is nothing technically wrong with this approach.

beck-8: ok,I agree with standardization the output.

magik6k (approved):

Issue #10388, opened 2023-03-03
Splitstore: proposal for MaxHotBytesTarget and space adaptive GC


ZenGround0:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated improvement to propose.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Improvement Suggestion

# Breakdown of splitstore problems and needs

Here is a useful breakdown from @f8-ptrk categorizing splitstore problems
> three things are important for us:
> a) we know it's upper boundaries in resource usage
> b) we can rely on it not causing getting out of sync
> c) we can rely on it to have everything available for  sealing/deals

Point c) is about soundness of compaction.  There is no evidence of problems with this today.

An attempt at fixing issues with point b) is being addressed in #10387 by limiting worker goroutines running flatten operations.  Its possible (likely?) there are still problems here.  To the extent this is still a problem the bottleneck is probably contention between GC and block sync reads / writes at the badger level.  We need to keep monitoring this issue and if it persists we can revisit designing something that prioritizes block sync over GC.  The easier solutions involve coarsely shutting off badger GC / compaction entirely after somehow detecting out of sync.  Harder solutions might involve somehow structurally doing compaction access of badger at a lower priority than chain sync access.  It's unclear if this is possible.

[...]

PR #10391, opened 2023-03-05
feat:splitstore:Configure max space used by hotstore and GC makes best effort to respect

ZenGround0:
## Related Issues

#10388 

## Proposed Changes
* Add size reporting to chain walking subfunctions
* Record sizes of the different DAGs walked during compaction
* New config value for targeting max total space used by hotstore
* GC logic to trigger moving GC adaptively based on hotstore size 
* Most importantly: splitstore refuses to do moving GC if disk usage would overflow the targetted max

## Checklist

- [ ] CI is green

### DIFF
--- a/blockstore/splitstore/splitstore_compact.go
+++ b/blockstore/splitstore/splitstore_compact.go
@@ -199,6 +199,8 @@ func (s *SplitStore) markLiveRefs(cids []cid.Cid) {
 	log.Debugf("marking %d live refs", len(cids))
 	startMark := time.Now()
 
+	szMarked := new(int64)
+
 	count := new(int32)
 	visitor := newConcurrentVisitor()
 	walkObject := func(c cid.Cid) (int, error) {
@@ -228,10 +230,12 @@ func (s *SplitStore) markLiveRefs(cids []cid.Cid) {
 
 	// optimize the common case of single put
 	if len(cids) == 1 {
-		if _, err := walkObject(cids[0]); err != nil {
+		sz, err := walkObject(cids[0])
+		if err != nil {
 			log.Errorf("error marking tipset refs: %s", err)
 		}
 		log.Debugw("marking live refs done", "took", time.Since(startMark), "marked", *count)
+		atomic.AddInt64(szMarked, int64(sz))
 		return
 	}
 
@@ -243,9 +247,11 @@ func (s *SplitStore) markLiveRefs(cids []cid.Cid) {
 
 	worker := func() error {
 		for c := range workch {
-			if _, err := walkObject(c); err != nil {
+			sz, err := walkObject(c)
+			if err != nil {
 				return err
 			}
+			atomic.AddInt64(szMarked, int64(sz))
 		}
 
 		return nil
@@ -268,7 +274,8 @@ func (s *SplitStore) markLiveRefs(cids []cid.Cid) {
 		log.Errorf("error marking tipset refs: %s", err)
 	}
 
-	log.Debugw("marking live refs done", "took", time.Since(startMark), "marked", *count)
+	log.Debugw("marking live refs done", "took", time.Since(startMark), "marked", *count, "size marked", *szMarked)
+	s.szMarkedLiveRefs += *szMarked
 }
 
 // transactionally protect a view
@@ -600,7 +607,6 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 	}
 
 	err = s.walkChain(curTs, boundaryEpoch, inclMsgsEpoch, &noopVisitor{}, fHot, fCold)
-
 	if err != nil {
 		return xerrors.Errorf("error marking: %w", err)
 	}
@@ -640,7 +646,7 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 	defer purgew.Close() //nolint:errcheck
 
 	// some stats for logging
-	var hotCnt, coldCnt, purgeCnt int
+	var hotCnt, coldCnt, purgeCnt, szPurge int
 	err = s.hot.ForEachKey(func(c cid.Cid) error {
 		// was it marked?
 		mark, err := markSet.Has(c)
@@ -651,6 +657,16 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 		if mark {
 			hotCnt++
 			return nil
+		}
+		if sz, err := s.hot.GetSize(s.ctx, c); err != nil {
+			if ipld.IsNotFound(err) {
+				log.Warnf("hotstore missing expected block %s", c)
+				return nil
+			}
+
+			return xerrors.Errorf("error retrieving block %s from hotstore: %w", c, err)
+		} else {
+			szPurge += sz
 		}
 
 		// it needs to be removed from hot store, mark it as candidate for purge
@@ -691,7 +707,8 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 
 	log.Infow("cold collection done", "took", time.Since(startCollect))
 
-	log.Infow("compaction stats", "hot", hotCnt, "cold", coldCnt, "purge", purgeCnt)
+	log.Infow("compaction stats", "hot", hotCnt, "cold", coldCnt, "purge", purgeCnt, "purge size", szPurge)
+	s.szToPurge = int64(szPurge)
 	stats.Record(s.ctx, metrics.SplitstoreCompactionHot.M(int64(hotCnt)))
 	stats.Record(s.ctx, metrics.SplitstoreCompactionCold.M(int64(coldCnt)))
 
@@ -775,8 +792,8 @@ func (s *SplitStore) doCompact(curTs *types.TipSet) error {
 		return xerrors.Errorf("error purging cold objects: %w", err)
 	}
 	log.Infow("purging cold objects from hotstore done", "took", time.Since(startPurge))
-
 	s.endCriticalSection()
+	log.Infow("total protected size", s.szProtectedTxns, "total marked live size", s.szMarkedLiveRefs)
 
 	if err := checkpoint.Close(); err != nil {
 		log.Warnf("error closing checkpoint: %s", err)
@@ -1069,7 +1086,7 @@ func (s *SplitStore) walkChain(ts *types.TipSet, inclState, inclMsgs abi.ChainEp
 	}
 
 	log.Infow("chain walk done", "walked", *walkCnt, "scanned", *scanCnt, "walk size", szWalk)
-
+	s.szWalk = *szWalk
 	return nil
 }
 
--- a/blockstore/splitstore/splitstore_compact.go
+++ b/blockstore/splitstore/splitstore_compact.go
@@ -201,7 +201,7 @@ func (s *SplitStore) markLiveRefs(cids []cid.Cid) {
 
 	count := new(int32)
 	visitor := newConcurrentVisitor()
-	walkObject := func(c cid.Cid) error {
+	walkObject := func(c cid.Cid) (int, error) {
 		return s.walkObjectIncomplete(c, visitor,
 			func(c cid.Cid) error {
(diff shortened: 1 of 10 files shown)


vyzo (commented):
hmmmm, ok, looks sane.
 
Potential concern: can we end up continuously moving gc in every compaction?

documentation/en/default-lotus-config.toml
```diff
@@ -230,6 +230,17 @@
...
+    # HotStoreMaxSpaceTarget sets a target max disk size for the hotstore. Splitstore GC
+    # will run moving GC if disk utilization gets within a threshold (150 GB) of the target.
+    # Splitstore GC will NOT run moving GC if the total size of the move would get
+    # within 50 GB of the target, and instead will run a more aggressive online GC.
+    # If both HotStoreFullGCFrequency and HotStoreMaxSpaceTarget are set then splitstore
+    # GC will trigger moving GC if either configuration condition is met.
+    #
+    # type: uint64
+    # env var: LOTUS_CHAINSTORE_SPLITSTORE_HOTSTOREMAXSPACETARGET
+    #HotStoreMaxSpaceTarget = 0
```
magik6k, line 242: Can we give some minimum recommendation, e.g. in terms of a multiplier of the "lightweight" (currently 123GiB) snapshot size.

Something like 4-5x the snapshot size should work (would be good to test)

blockstore/splitstore/splitstore_gc.go
```diff
@@ -7,10 +7,56 @@ import (
...
+	var err error
+	sizer, ok := s.hot.(bstore.BlockstoreSize)
+	if ok {
+		hotSize, err = sizer.Size()
+		if err != nil {
+			log.Warnf("error getting hotstore size: %s, estimating empty hot store for targeting", err)
+			hotSize = 0
+		}
+	} else {
+		hotSize = 0
```
magik6k, line 41: This probably shouldn't happen, so print a loud log?

blockstore/splitstore/splitstore_compact.go
```diff
@@ -1047,123 +1088,129 @@ func (s *SplitStore) walkChain(ts *types.TipSet, inclState, inclMsgs abi.ChainEp
...
 	}
 
-	return nil
+	return sz, nil
 }
 
 // like walkObject, but the object may be potentially incomplete (references missing)
-func (s *SplitStore) walkObjectIncomplete(c cid.Cid, visitor ObjectVisitor, f, missing func(cid.Cid) error) error {
+func (s *SplitStore) walkObjectIncomplete(c cid.Cid, visitor ObjectVisitor, f, missing func(cid.Cid) error) (int, error) {
+	sz := 0
```
magik6k, line 1149: Reasonably confident that a 32 bit int will overflow on a bunch of actors

blockstore/splitstore/splitstore_compact.go
```diff
@@ -1047,123 +1088,129 @@ func (s *SplitStore) walkChain(ts *types.TipSet, inclState, inclMsgs abi.ChainEp
...
 
 	// check this before recursing
 	if err := s.checkClosing(); err != nil {
-		return err
+		return sz, err
 	}
 
 	var links []cid.Cid
 	err = s.view(c, func(data []byte) error {
+		sz += len(data)
```
magik6k, line 1195: The cid/entry also has size, I'd say at least 100bytes of overhead per key, and that may be optimistic (badger structures aren't all very light, keys are both in vlogs / SSTs, etc.).

blockstore/splitstore/splitstore_gc.go
```diff
@@ -7,10 +7,56 @@ import (
...
+	}
+
+	copySizeApprox := s.szKeys + s.szMarkedLiveRefs + s.szProtectedTxns + s.szWalk
+	shouldTarget := s.cfg.HotstoreMaxSpaceTarget > 0 && hotSize+copySizeApprox > int64(s.cfg.HotstoreMaxSpaceTarget)-targetThreshold
+	shouldFreq := s.cfg.HotStoreFullGCFrequency > 0 && s.compactionIndex%int64(s.cfg.HotStoreFullGCFrequency) == 0
+	shouldDoFull := shouldTarget || shouldFreq
+	canDoFull := s.cfg.HotstoreMaxSpaceTarget == 0 || hotSize+copySizeApprox < int64(s.cfg.HotstoreMaxSpaceTarget)-targetBuffer
+	log.Infof("measured hot store size: %d, approximate new size: %d, should do full %t, can do full %t", hotSize, copySizeApprox, shouldDoFull, canDoFull)
```
magik6k, line 49: Would also log actual size after moving GC so that we can tell how close/far the estimate is

blockstore/splitstore/splitstore_gc.go
```diff
@@ -7,10 +7,56 @@ import (
...
+
 	var opts []bstore.BlockstoreGCOption
-	if s.cfg.HotStoreFullGCFrequency > 0 && s.compactionIndex%int64(s.cfg.HotStoreFullGCFrequency) == 0 {
+	if shouldDoFull && canDoFull {
 		opts = append(opts, bstore.WithFullGC(true))
+	} else if shouldDoFull && !canDoFull {
+		log.Warnf("Attention! Estimated moving GC size %d is not within safety buffer %d of target max %d, performing aggressive online GC to attempt to bring hotstore size down safely", copySizeApprox, targetBuffer, s.cfg.HotstoreMaxSpaceTarget)
+		log.Warn("If problem continues you can 1) temporarily allocate more disk space to hotstore and 2) reflect in HotstoreMaxSpaceTarget OR trigger manual move with `lotus chain prune hot-moving`")
+		log.Warn("If problem continues and you do not have any more disk space you can run continue to manually trigger online GC at aggressive thresholds (< 0.01) with `lotus chain prune hot`")
```
magik6k, line 57: Probably worth using the alert system to raise a (generally much more visible / loud) alert.

See e.g. https://github.com/filecoin-project/lotus/blob/master/node/modules/alerts.go

(alerts are for things which tell the user that something either is, or is about to be severely broken, and they need to take action)

magik6k (commented):
General idea makes a ton of sense. Couple of comments

Would be great to hook-up those sizes / size estimates to the metrics endpoint somehow - https://github.com/filecoin-project/lotus/blob/master/metrics/metrics.go#L160

ZenGround0:
> Potential concern: can we end up continuously moving gc in every compaction?

Reasonable concern, this is definitely possible.  In my opinion this is more of a feature than a bug however.  Misconfigurations are certainly possible but I would say we don't expose any worse footgun than allowing FullGCFreq = 1 with current config.  

What I've come to believe is there are only two real problems users face with splitstore:  disk overflow and chain out of sync.  The only risk with too much moving GC when it is done without overflowing disk is that it gets the chain out of sync.  But I think we can address chain out of sync additionally in #10392 or an extension after fixing a solution to disk overflow.

ribasushi:
> FullGCFreq = 1 with current config.

This is how I configure my experimental splitstore nodes, anything besides "do moving GC every time" makes no sense from an operational perspective: you *want* your dataset to always be as small as possible.

ZenGround0:
> anything besides "do moving GC every time" makes no sense from an operational perspective

While I agree that "do moving GC every time" *often* makes sense from an operational perspective I think this is overstated.  Moving GC is objectively more expensive than online in terms of memory, compute and disk io.  Paying the cost on every compaction to keep datastore down is likely *often* a good idea but keeping more disk around and then doing this in batches is a reasonable point in the tradeoff space.

documentation/en/default-lotus-config.toml
ZenGround0: From lots of recent operational experience we know that fully clean hotstore size is somewhere between 220 and 250 G.  So double this at 500GB is the absolute minimum, call it 550 GB with the safety buffer.  This turns out to be ~5x snapshot size so your intuition is good.

ZenGround0: The correspondence between snapshot size and hotstore size is somewhat confounded since snapshot = state + 3*(churn + messages) and hotstore = state + [between 4 and 5]*(churn + messages)

blockstore/splitstore/splitstore_gc.go
ZenGround0: Yes awesome, after online GC too.

ZenGround0:
@magik6k ready for another round.  I ran out of time so filing this follow up to get alerting and metrics in place after the merge: #10428.

node/config/types.go
```diff
@@ -601,6 +601,16 @@ type Splitstore struct {
 	// A value of 0 disables, while a value 1 will do full GC in every compaction.
 	// Default is 20 (about once a week).
 	HotStoreFullGCFrequency uint64
+	// HotStoreMaxSpaceTarget sets a target max disk size for the hotstore. Splitstore GC
+	// will run moving GC if disk utilization gets within a threshold (150 GB) of the target.
+	// Splitstore GC will NOT run moving GC if the total size of the move would get
+	// within 50 GB of the target, and instead will run a more aggressive online GC.
```
magik6k, line 607: Isn't the move basically always going to be larger than 100GiB tho? In which case it would never run because we'd always already be withing 50G of the target

blockstore/splitstore/splitstore_compact.go
```diff
@@ -1047,123 +1077,129 @@ func (s *SplitStore) walkChain(ts *types.TipSet, inclState, inclMsgs abi.ChainEp
 		}
 	}
 
-	log.Infow("chain walk done", "walked", *walkCnt, "scanned", *scanCnt)
-
+	log.Infow("chain walk done", "walked", *walkCnt, "scanned", *scanCnt, "walk size", szWalk)
+	s.szWalk = *szWalk
```
magik6k, line 1081: So while there is a 99% chance that this is correct, I'd still use atomic.LoadInt to get the value, just out of abundance of caution around thread-shared variables

blockstore/splitstore/splitstore_gc.go
```diff
@@ -7,15 +7,65 @@ import (
...
 )
 
+const (
+	// When < 150 GB of space would remain during moving GC, trigger moving GC
+	targetThreshold = 150_000_000_000
+	// Don't attempt moving GC with 50 GB or less would remain during moving GC
+	targetBuffer = 50_000_000_000
+	// Fraction of garbage in badger vlog for online GC traversal to collect garbage
+	aggressiveOnlineGCThreshold = 0.0001
+)
```
magik6k, line 17: Can we make those uppercase / var so that it's possible to set with compiler flags / from other packages?

magik6k (commented):
One question, 2 small nits, but other than that looks good.

If it's not perfect bugs can be fixed in the RC process

node/config/types.go
ZenGround0: Move is always going to be larger than 100GiB but it should still run, and I observe it running on my mainnet node with the current code.  The target is for the entire hotstore size.  So if the Newly Moved + Current sizes are too close (within 50GB) to the target then move won't run.  But if Newly Moved + Current are within 150 GB of the target but not 50 GB to the target than move runs.

Example 1 you have a hotstore of size 300G and a target of 800G, compaction computes a totally GCed size of 250G.  Then 300 + 250 G = 550 < 800 - 150 = 650, so move won't run.

Example 2 same sizes but target is 650, then 650 - 150 = 500 < 550 so move will try to run.  Move is able to run within safety buffer because 650 - 50 = 600 > 550.

Example 3 target is 599.  Move now can't run because total moving size 550 > 559 - 50.  So we will run aggressive online GC in hopes of getting hotstore size down back into a safe region.

magik6k: I see..

This should probably work, I guess we'll see in real world deployments.

Actually it would be really good to make it possible to configure targetThreshold / targetBuffer, ideally through config, but envvars would also work for now

magik6k (approved):

ZenGround0:
Thanks for the make gen assist @magik6k !!

Issue #10393, opened 2023-03-06
Solve Codefi gas oracle returning 0 priority fee


raulk:
## Situation

The Codefi `suggestedGasFees` oracle keeps returning `latestPriorityFeeRange` and `historicalPriorityFeeRange` as 0 on Hyperspace, even when messages have been included in the last 1-2 tipsets with non-zero priority fee.

```sh
curl 'https://gas-api.metaswap.codefi.network/networks/3141/suggestedGasFees'
```

```json
{
  "low": {
    "suggestedMaxPriorityFeePerGas": "0",
    "suggestedMaxFeePerGas": "0.0000001",
    "minWaitTimeEstimate": 15000,
    "maxWaitTimeEstimate": 30000
  },
  "medium": {
    "suggestedMaxPriorityFeePerGas": "0",
    "suggestedMaxFeePerGas": "0.000000135",
    "minWaitTimeEstimate": 15000,
    "maxWaitTimeEstimate": 45000
  },
  "high": {
    "suggestedMaxPriorityFeePerGas": "0",
    "suggestedMaxFeePerGas": "0.00000017",
    "minWaitTimeEstimate": 15000,
    "maxWaitTimeEstimate": 60000
  },
  "estimatedBaseFee": "0.0000001",
  "networkCongestion": 0,
  "latestPriorityFeeRange": [
    "0",
    "0"
  ],
  "historicalPriorityFeeRange": [
    "0",
    "0"
  ],
  "historicalBaseFeeRange": [
    "0.0000001",
    "0.0000001"
  ],
  "priorityFeeTrend": "level",
  "baseFeeTrend": "up"
}
```

## Impact

This makes MetaMask default to a priority fee of 0. If this situation persists in mainnet, MetaMask senders will have problems with message inclusion.

## Notes

- Codefi's gas oracle uses `eth_feeHistory` under the hood, so it's possible that a bug in `eth_feeHistory` is reporting a zero priority fee.
- Perhaps `eth_feeHistory` should be clamping to [`MinGasPremium` (0.1 nanoFIL) as a bare minimum](https://github.com/filecoin-project/lotus/blob/03b419e3d47f6a1f99045a39eb2fdd5e24d3ad59/node/impl/full/gas.go#L217). 
- `eth_maxPriorityFeePerGas` seems ok:
	```
	⟩ curl "https://api.hyperspace.node.glif.io/rpc/v1" \
	          -X POST \
[...]

PR #10413, opened 2023-03-07
fix: EthAPI: use StateCompute for feeHistory; apply minimum gas premium

ychiaoli18:
## Related Issues
https://github.com/filecoin-project/lotus/issues/10393
https://github.com/filecoin-project/lotus/issues/10414

## Proposed Changes
1. use StateCompute instead of StateSearchMsg for feeHistory
2. return MinGasPremium when there's no tx in the block

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/eth_fee_history_test.go
+++ b/itests/eth_fee_history_test.go
@@ -3,12 +3,14 @@ package itests
 import (
 	"context"
 	"encoding/json"
+	"sort"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-jsonrpc"
+	"github.com/filecoin-project/go-state-types/abi"
 
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
@@ -16,6 +18,21 @@ import (
 	"github.com/filecoin-project/lotus/lib/result"
 	"github.com/filecoin-project/lotus/node/impl/full"
 )
+
+func getAnswer(tsHeights []int, requestAmount, startHeight int) (count, oldestHeight int) {
+	idx := sort.SearchInts(tsHeights, startHeight)
+	// handle null rounds
+	for tsHeights[idx] > startHeight {
+		idx--
+	}
+	cnt := requestAmount
+	oldestIdx := idx - requestAmount + 1
+	if idx < cnt {
+		cnt = idx + 1
+		oldestIdx = 0
+	}
+	return cnt, tsHeights[oldestIdx]
+}
 
 func TestEthFeeHistory(t *testing.T) {
 	require := require.New(t)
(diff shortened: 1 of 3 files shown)


node/impl/full/eth.go
```diff
@@ -696,28 +696,30 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
 	rewardsArray := make([][]ethtypes.EthBigInt, 0)
 
 	for ts.Height() >= abi.ChainEpoch(oldestBlkHeight) {
-		// Unfortunately we need to rebuild the full message view so we can
-		// totalize gas used in the tipset.
-		msgs, err := a.Chain.MessagesForTipset(ctx, ts)
+		compOutput, err := a.StateCompute(ctx, ts.Height(), nil, ts.Key())
```
raulk, line 699: How does this deal with null rounds?

magik6k (approved):
Looks like this works. Also should be faster, I think.

raulk (approved):

itests/eth_fee_history_test.go
```diff
@@ -22,13 +24,16 @@ func TestEthFeeHistory(t *testing.T) {
...
 	blockTime := 100 * time.Millisecond
 	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
-	ens.InterconnectAll().BeginMining(blockTime)
+	miner := ens.InterconnectAll().BeginMining(blockTime)
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
 	// Wait for the network to create 20 blocks
 	client.WaitTillChain(ctx, kit.HeightAtLeast(20))
```
arajasek, line 33: What happens if some of these rounds are null blocks? I can't tell whether the test assertions are still correct.

node/impl/full/eth.go
```diff
@@ -695,29 +691,31 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
 	gasUsedRatioArray := []float64{}
 	rewardsArray := make([][]ethtypes.EthBigInt, 0)
 
-	for ts.Height() >= abi.ChainEpoch(oldestBlkHeight) {
-		// Unfortunately we need to rebuild the full message view so we can
-		// totalize gas used in the tipset.
-		msgs, err := a.Chain.MessagesForTipset(ctx, ts)
+	for len(gasUsedRatioArray) < int(params.BlkCount) && ts.Height() > 0 {
+		compOutput, err := a.StateCompute(ctx, ts.Height(), nil, ts.Key())
```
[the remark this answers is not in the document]
raulk: I'm doing a sanity check on the logic here, it's ultimately in `parseBlkParam`.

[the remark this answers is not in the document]
raulk: https://github.com/filecoin-project/lotus/issues/10426 -- will be fixed separately

itests/eth_fee_history_test.go
```diff
@@ -29,95 +46,109 @@ func TestEthFeeHistory(t *testing.T) {
...
+			for _, c := range chg {
+				tsHeights = append(tsHeights, int(c.Val.Height()))
+			}
+		}
+	}()
+
+	miner[0].InjectNulls(abi.ChainEpoch(1))
+	<-time.After(5 * blockTime)
+	miner[0].InjectNulls(abi.ChainEpoch(3))
+	<-time.After(2 * blockTime)
```
raulk, line 65: Depending on time is usually flaky. Consider waiting for N number of tipsets produced again. See https://github.com/filecoin-project/lotus/pull/10427/files#diff-57c54de2d685e694dee103dc15a360776a5b30b5ec7fc77b5150c3756525d683R322-R328 for some hints.

itests/eth_fee_history_test.go
```diff
@@ -29,95 +46,109 @@ func TestEthFeeHistory(t *testing.T) {
...
+	miner[0].InjectNulls(abi.ChainEpoch(1))
+
+	// Wait for the network to create at least 20 blocks
```
raulk, line 68: Tipsets, not blocks.

itests/eth_fee_history_test.go
```diff
@@ -29,95 +46,109 @@ func TestEthFeeHistory(t *testing.T) {
...
 	client.WaitTillChain(ctx, kit.HeightAtLeast(20))
 	for _, m := range miner {
 		m.Pause()
 	}
+	<-time.After(5 * blockTime)
```
raulk, line 73: This is racy. It's possible that the miners may have not pausd yet. Safer to do a ChainNotify with a 5 second timeout and assert that it errors with context deadline exceeded.

itests/eth_fee_history_test.go
```diff
@@ -29,95 +46,109 @@ func TestEthFeeHistory(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
-	// Wait for the network to create 20 blocks
+	heads, err := client.ChainNotify(ctx)
+	require.NoError(err)
+
+	// Save the full view of the tipsets to calculate the answer when there are null rounds
+	tsHeights := []int{1}
```
raulk, line 53: Why does this need to be preloaded with 1?

itests/eth_fee_history_test.go
```diff
@@ -29,95 +46,109 @@ func TestEthFeeHistory(t *testing.T) {
...
+	go func() {
+		for chg := range heads {
+			for _, c := range chg {
+				tsHeights = append(tsHeights, int(c.Val.Height()))
```
raulk, line 57: Note that the first notification you receive will be the current tipset, i.e. c.Type will be store.HCCurrent.

itests/eth_fee_history_test.go
```diff
@@ -29,95 +46,109 @@ func TestEthFeeHistory(t *testing.T) {
...
+	// Wait for the network to create at least 20 blocks
 	client.WaitTillChain(ctx, kit.HeightAtLeast(20))
 	for _, m := range miner {
 		m.Pause()
 	}
+	<-time.After(5 * blockTime)
+
+	sort.Ints(tsHeights)
+
+	latestBlk := ethtypes.EthUint64(tsHeights[len(tsHeights)-2])
```
raulk, line 77: Can you add a comment indicating that this is the last executed tipset?

itests/eth_fee_history_test.go
```diff
@@ -17,6 +19,21 @@ import (
 	"github.com/filecoin-project/lotus/node/impl/full"
 )
 
+func getAnswer(tsHeights []int, requestAmount, startHeight int) (count, oldestHeight int) {
```
raulk, line 22: This needs docs and comments. Can we rename to `calculateExpectations`?

node/impl/full/eth.go
```diff
@@ -252,7 +252,7 @@ func (a *EthModule) parseBlkParam(ctx context.Context, blkParam string) (tipset
 		if err != nil {
 			return nil, fmt.Errorf("cannot parse block number: %v", err)
 		}
-		ts, err := a.Chain.GetTipsetByHeight(ctx, abi.ChainEpoch(num), nil, false)
+		ts, err := a.Chain.GetTipsetByHeight(ctx, abi.ChainEpoch(num), nil, true)
```
raulk, line 255: Can you revert this change, since it's already in https://github.com/filecoin-project/lotus/pull/10427?

node/impl/full/eth.go
```diff
@@ -691,7 +691,8 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
 	gasUsedRatioArray := []float64{}
 	rewardsArray := make([][]ethtypes.EthBigInt, 0)
 
-	for len(gasUsedRatioArray) < int(params.BlkCount) && ts.Height() > 0 {
+	blockCount := 0
```
raulk, line 694: Let's call this `blocksIncluded`.

itests/eth_fee_history_test.go
```diff
@@ -17,6 +19,21 @@ import (
...
+	idx := sort.SearchInts(tsHeights, startHeight)
```
raulk, line 23: latestIdx

itests/eth_fee_history_test.go
```diff
@@ -17,6 +19,21 @@ import (
...
+	// handle null rounds
+	for tsHeights[idx] > startHeight {
+		idx--
+	}
+	cnt := requestAmount
+	oldestIdx := idx - requestAmount + 1
+	if idx < cnt {
+		cnt = idx + 1
```
raulk, line 31: Not sure I follow this.

If you have `tsHeights = [1, 2, 3, 4, 6, 8]`, and you request `startHeight = 5`, you will have `idx = 4` (current value 6). If you request 10 heights, you will enter this if, and will end up returning cnt = 5, even though you would only process 4 heights (1..4)?

itests/eth_fee_history_test.go
ychiaoli18: It seems that ChainNotify doesn't send the tipset with height = 1

itests/eth_fee_history_test.go
ychiaoli18: In your example, idx would be 3 (idx would be the index of the largest number that's smaller than the target), so it would return 4 entries. I can rewrite it to make it clearer.

node/impl/full/eth.go
ychiaoli18: The e2e test would fail if this line is not included. I can rebase if #10427 is merged.

itests/eth_fee_history_test.go
raulk: Hm, not according to the Godocs and this playground: https://play.golang.com/p/bQyjTqj9Gxb

ychiaoli18: There's a [for loop](https://github.com/filecoin-project/lotus/pull/10413/files/1747893ead537a24b39082c2f7000ceb742b5e13..61d2d461e1e4dc29b360f095d070b48377922c7d#diff-4ec74518af0f29b47821833fc7b7ff99d23cb2b21ce3bfe7deeb6c9994f64646R25-R27) that decreases the idx

itests/eth_fee_history_test.go
raulk: Hm, that won't work, but we should wait for 5 seconds of ChainNotify not sending any notifications.

raulk: Honestly I think this is fine, just keep it as it is. Pause is synchronous, so we're just waiting on draining all pending chain notifications, and 5 seconds is of plenty time for that.

itests/eth_fee_history_test.go
raulk: Hm, that suggests raciness. The ChainNotify is probably starting on tipset 2. Note that ChainNotify sends a "current" notification on subscribe, and the subscription might be starting on height 2. Do we absolutely need to use ChainNotify here? Alternatively you can walk the chain and populate tsHeights after the fact, i.e. after pausing the miner?

node/impl/full/eth.go
raulk: Gotcha, thanks.

itests/eth_fee_history_test.go
raulk: ChainNotify being called before the miner starts resolves my raciness concerns. Starting at 2 is probably an artifact from the testkit, and the way we set up miners.

raulk (approved):

PR #10416, opened 2023-03-08
feat: mempool: Reduce minimum replace fee from 1.25x to 1.1x

### LINKED ISSUE #10415: GasPremium: Change ReplaceByFeeRatio to 1.1
It (`ReplaceByFeeRatio`) is currently 1.25x but Ethereum uses 1.1x. The actual value here isn't that important, but choosing a number _higher_ than Ethereum causes issues with Ethereum tooling like metamask when they try to cancel/replace messages.

The alternative would be to try to get upstream tools (metamask, etc.) to make this configurable, but that'll likely be an up-hill battle where changing it in lotus should be pretty easy.

**Impact:**

- When replacing a message, the new gas premium will be 1.1x the old gas premium, instead of 1.25x.
- This is not expected to have a noticeable impact on storage provider revenue as it only affects replaced messages (which are not common) and only affects the gas premium (which is usually very low anyways).
- This is not expected to have any impact on network security. It'll make it slightly cheaper to spam pubsub, but 10% versus 25% doesn't really make a difference.
- This will not affect pubsub scores.

**Implementation:**

There are two sides to this:

1. The default fee bump applied by the user when they replace a message.
2. The minimum fee bump allowed by the message pool.

For now, we should change the second one but _not_ the first one. Once enough of the network has updated, we can change the first one as well.

Stebalien:
However, we're leaving the default at 1.25x for backwards compatibility, for now.

Also:

1. Actually use the configured replace fee ratio.
2. Store said ratios as percentages instead of floats. 1.25, or `1+1/(2^2)`, can be represented as a float. 1.1, or `1 + 1/(2 * 5)`, cannot.

## Related Issues

fixes #10415

## Proposed Changes

Reduce minimum replace fee from 1.25x to 1.1x. However, we're leaving the default at 1.25x for backwards compatibility, for now.

Also:

1. Actually use the configured replace fee ratio.
2. Store said ratios as percentages instead of floats. 1.25, or 1+1/(2^2), can be represented as a float. 1.1, or 1 + 1/(2 * 5), cannot.

## Checklist

- [ ] CI is green


chain/types/percent.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"golang.org/x/xerrors"
+)
+
+// Percent stores a signed percentage as an int64. When converted to a string (or json), it's stored
+// as a decimal with two places (e.g., 100% -> 1.00).
+type Percent int64
+
+func (p Percent) String() string {
+	return fmt.Sprintf(`%d.%d`, p/100, p%100)
```
magik6k, line 16: Nit: will not be correct for negative numbers

magik6k (approved):
One non-blocking nit, looks good otherwise

chain/types/percent.go
Stebalien: Done. Thanks.

arajasek (approved):

Issue #10403, opened 2023-03-07
`transactionIndex` differs between `eth_getBlockByNumber` and `eth_getTransactionReceipt`


raulk:
The `transactionIndex` returns different results in the `eth_getBlockByNumber` and `eth_getTransactionReceipt` interfaces. The reason is that `txIdx` did not ignore system messages, and this is not as expected.

Example in hyperspace:
```
 curl -s --location 'http://localhost:8839/rpc/v1' \
--header 'Content-Type: application/json' \
--data '{
    "jsonrpc":"2.0",
    "method":"eth_getTransactionReceipt",
    "params":["0xcc502ba55e9d312349c647783964c7edb00c3890b96842bd53e718475dc8c71e"],
    "id":1
}' |gojq
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "blockHash": "0x52206698ebadc31c0e3e211a9cd3f02dc81603f74e8a604dc460fa3cab1f597b",
    "blockNumber": "0x1f302",
    "contractAddress": null,
    "cumulativeGasUsed": "0x0",
    "effectiveGasPrice": "0x48724a74",
    "from": "0xff000000000000000000000000000000000005e4",
    "gasUsed": "0x135482a",
    "logs": [],
    "logsBloom": "0x09000c0000085003840012000100102008204100010000002105018000100000040000400024000100800000003000000010001002001110500000020016a0048880004040800081488100088e80400208048011010410000120000180002086200000280a022004000000812068490000002004000400000a01001000420800010000440400000002410100000004045800200500b000a0800140024001000060081000008109000000001800020800000420008083800060888044000040000200840a00002010600000000010d0800400000000080048081408000800600028400800048000400000400000a1000108000020001003440504088000026000",
    "root": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "status": "0x1",
    "to": "0xff0000000000000000000000000000000000088a",
    "transactionHash": "0xcc502ba55e9d312349c647783964c7edb00c3890b96842bd53e718475dc8c71e",
    "transactionIndex": "0x0",
    "type": "0x2"
  }
}
```

```
curl -s --location 'http://localhost:8839/rpc/v1' \
--header 'Content-Type: application/json' \
--data '{
"jsonrpc": "2.0",
"method": "eth_getBlockByNumber",
"id": 1,
"params": ["0x1f302", true]
}
' |gojq
{
  "id": 1,
  "jsonrpc": "2.0",
[...]

PR #10419, opened 2023-03-08
fix: eth API: return correct txIdx around null blocks

arajasek:
## Related Issues

#10403 , thanks to @wanyvic for the report!

## Proposed Changes

We're skipping system messages from the list of transactions returned, but not adjusting the txIdx accordingly. This is only noticed when the previous round was a null tipset.

Adjust for skipped system messages.

## Additional Info

Bug introduced in #10380.

## Checklist

- [ ] CI is green

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -1796,11 +1796,15 @@ func newEthBlockFromFilecoinTipSet(ctx context.Context, ts *types.TipSet, fullTx
 		return ethtypes.EthBlock{}, xerrors.Errorf("failed to compute state: %w", err)
 	}
 
-	for txIdx, msg := range compOutput.Trace {
+	txIdx := 0
+	for _, msg := range compOutput.Trace {
 		// skip system messages like reward application and cron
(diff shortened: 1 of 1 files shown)


node/impl/full/eth.go
```diff
@@ -1796,9 +1796,11 @@ func newEthBlockFromFilecoinTipSet(ctx context.Context, ts *types.TipSet, fullTx
 		return ethtypes.EthBlock{}, xerrors.Errorf("failed to compute state: %w", err)
 	}
 
+	skippedTxs := 0
 	for txIdx, msg := range compOutput.Trace {
```
Stebalien, line 1800:
```suggestion
	txIdx := 0
	for _, msg := range compOutput.Trace {
```

?

Then just increment the index for each non-skipped message.

arajasek: oh yeah that's better

raulk (approved):

raulk (approved):
A test to verify consistency would be nice.

PR #10422, opened 2023-03-08
docs: API Gateway: patch documentation note about make gen command

ArseniiPetrovich:
## Proposed Changes

Simple code comment update around gateway patch

## Additional Info

`make gen` didn't work with me. Then @jennijuju proposed to use `make clean && make all` first, which helped. I think it would be great to have full command here

## Checklist

- [ ] CI is green


magik6k:
`make deps` instead of `make all` should also work, but be way faster

ArseniiPetrovich:
Yes, worked for me :)

magik6k (approved):

PR #10423, opened 2023-03-08
feat: chain: make chain tipset fetching 1000x faster

arajasek:
## Related Issues

@jennijuju   flagged that we need faster fetching of tipsets by height for Eth tooling support.

## Proposed Changes

- Increase the skipcache size to 2^20 entries
- Drop the tipset itself from the skipcache entries. This is a lot of information to hold in memory that we don't actually need.
- Drop the parentHeight from the skipcache entries. This is just 8 bytes, but I don't think we need it.
- With that, we can easily rely on the skip cache to load 20 million epochs (so good until 2040).
- Each cache entry itself is ~308 bytes (an int64 + a TSK, using ~5 CIDs per TSK, which is aggressive), which means the total cache size when we get to 20M epochs will be ~0.3 GiB.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/store/index.go
+++ b/chain/store/index.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"os"
 	"strconv"
+	"sync"
 
-	lru "github.com/hashicorp/golang-lru"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-state-types/abi"
@@ -13,7 +13,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 )
 
-var DefaultChainIndexCacheSize = 32 << 10
+var DefaultChainIndexCacheSize = 32 << 15
 
 func init() {
 	if s := os.Getenv("LOTUS_CHAIN_INDEX_CACHE"); s != "" {
@@ -27,7 +27,8 @@ func init() {
 }
 
 type ChainIndex struct {
-	skipCache *lru.ARCCache
+	indexCacheLk sync.Mutex
+	indexCache   map[types.TipSetKey]*lbEntry
 
 	loadTipSet loadTipSetFunc
 
@@ -36,17 +37,14 @@ type ChainIndex struct {
 type loadTipSetFunc func(context.Context, types.TipSetKey) (*types.TipSet, error)
 
 func NewChainIndex(lts loadTipSetFunc) *ChainIndex {
-	sc, _ := lru.NewARC(DefaultChainIndexCacheSize)
 	return &ChainIndex{
-		skipCache:  sc,
+		indexCache: make(map[types.TipSetKey]*lbEntry, DefaultChainIndexCacheSize),
 		loadTipSet: lts,
 		skipLength: 20,
 	}
 }
 
 type lbEntry struct {
-	ts           *types.TipSet
-	parentHeight abi.ChainEpoch
 	targetHeight abi.ChainEpoch
 	target       types.TipSetKey
 }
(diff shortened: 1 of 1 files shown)


arajasek:
Timing results on mainnet (with some diagnostic logging, will update without that, will be faster):

```
aayush@aayush-desktop ~/p/lotus (master)> time ./lotus chain list --count=3 --height=200
198: (Aug 24 19:39:00) [ bafy2bzacear4yqnylpchlufj5nccvdv56zi7655jtich2chbk2w7oci4fv4b6: f01002, ]
199: (Aug 24 19:39:30) [ bafy2bzacecrlj6vujyqwm5daygo27ksaibr37wyklcby6ihb4kiejlfludzyi: f01002, ]
200: (Aug 24 19:40:00) [ bafy2bzacedgt32ubhpx4hu3hrbu663djcykuplt4jc5mnyii2wlzx5ovsj7cs: f01000,bafy2bzaceb6ksh75v3m6dydsutfrcaoe542d44rotznniq5poi2bfirthl6se: f01002, ]

Executed in  398.06 secs      fish           external
   usr time  736.38 millis  373.00 micros  736.01 millis
   sys time   70.71 millis  125.00 micros   70.58 millis

aayush@aayush-desktop ~/p/lotus (asr/fast-index)> time ./lotus chain list --count=3 --height=200
198: (Aug 24 19:39:00) [ bafy2bzacear4yqnylpchlufj5nccvdv56zi7655jtich2chbk2w7oci4fv4b6: f01002, ]
199: (Aug 24 19:39:30) [ bafy2bzacecrlj6vujyqwm5daygo27ksaibr37wyklcby6ihb4kiejlfludzyi: f01002, ]
200: (Aug 24 19:40:00) [ bafy2bzacedgt32ubhpx4hu3hrbu663djcykuplt4jc5mnyii2wlzx5ovsj7cs: f01000,bafy2bzaceb6ksh75v3m6dydsutfrcaoe542d44rotznniq5poi2bfirthl6se: f01002, ]

Executed in  957.65 millis    fish           external
   usr time  439.20 millis  383.00 micros  438.81 millis
   sys time   87.91 millis  129.00 micros   87.78 millis
```

Warming the cache takes ~6 minutes, subsequent queries are less than a second.

Stebalien (approved):
LGTM. A couple of comments:

1. I'd just drop the ARC cache and just use a map. We're likely wasting a _lot_ of time updating the "recency" metrics on that cache every time we use it (and the memory locality is likely really bad).
2. I'm somewhat concerned that a single query might take 6 minutes (will timeout on most services). We might want some form of background job to warm the cache. But (a) that can happen later and (b) service providers can likely work around this issue themselves.

Stebalien:
If this isn't enough, we can consider implementing a real skiplist (with multiple levels).

arajasek:
@Stebalien Thanks for the review and the suggestions!

I switched to using a map. It appears to have made the warmup _slower_, which is somewhat unexpected, but subsequent lookups faster. I think that's still a win, in addition to the overall simplicity, so I'm gonna push that up.

chain/store/index.go
```diff
@@ -36,17 +35,14 @@ type ChainIndex struct {
 type loadTipSetFunc func(context.Context, types.TipSetKey) (*types.TipSet, error)
 
 func NewChainIndex(lts loadTipSetFunc) *ChainIndex {
-	sc, _ := lru.NewARC(DefaultChainIndexCacheSize)
 	return &ChainIndex{
-		skipCache:  sc,
+		indexCache: make(map[types.TipSetKey]*lbEntry),
```
Stebalien, line 39: It's slower because you're not initializing the map to DefaultChainIndexCacheSize.

Stebalien (approved):

chain/store/index.go
Stebalien: Also note: I _think_ you can just use `lbEntry` by value to save some allocations which should speed things up even more.

PR #10433, opened 2023-03-09
feat: worker: Ensure tempdir exists on startup

magik6k:
## Related Issues
This is something I occasionally run into in my setup when starting workers with custom TMPDIR, which on some of my systems disappears after reboot.

If not caught this would cause tasks like AddPiece to fail with a somewhat crypic error like
```
Caused by:
    0: failed to create data store
    1: No such file or directory (os error 2)
``` 

## Proposed Changes
Make UX slightly better when the tempdir doesn't exist.

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus-worker/main.go
+++ b/cmd/lotus-worker/main.go
@@ -296,6 +296,13 @@ var runCmd = &cli.Command{
 			}
 		}
 
+		// ensure tmpdir exists
+		td := os.TempDir()
+		if err := os.MkdirAll(td, 0755); err != nil {
+			return xerrors.Errorf("ensuring temp dir %s exists: %w", td, err)
+		}
+
+		// Check file descriptor limit
 		limit, _, err := ulimit.GetLimit()
 		switch {
 		case err == ulimit.ErrUnsupported:


cmd/lotus-worker/main.go
```diff
@@ -296,6 +296,13 @@ var runCmd = &cli.Command{
 			}
 		}
 
+		// ensure tmpdir exists
+		td := os.TempDir()
+		if err := os.MkdirAll(td, 0755); err != nil {
```
AnomalRoil, line 301: If the directory exists but your don't have perms that wouldn't detect it I think.

You might want to use `MkdirTemp("", "lotus")` instead maybe since it's going to try and create a new directory in the TempDir.

magik6k: Perms are basically never an issue, but non-existent temp directories tend to be - this is because in many setups you'll set `TMPDIR=/tmp/something`, and just forget to create those directories. It's also worth noting that people tend to run workers as root, because that's easier than setting up all perms needed for everything to work (cuda, memory locking, thread pinning, etc.)

shrenujbansal:
Where do we have cleanup for this temp dir?

fridrik01 (approved):
LGTM

shrenujbansal (approved):

PR #10435, opened 2023-03-09
chore: ffi: update the FFI for release

Stebalien:
Specifically, this updates the proofs libraries and enables CUDA by default.


arajasek (approved):
Also needed in `master`?

Stebalien:
> Also needed in master?

Yes, but I figured we'd just merge this branch into master?

arajasek:
> > Also needed in master?
> 
> Yes, but I figured we'd just merge this branch into master?

That only happens at the _end_ of the release process (so 3 weeks from now after testing). So I think it needs to happen now.

PR #10439, opened 2023-03-10
fix: cherrypick/conflict wrongfully resolved

jennijuju:
seems like the right code should be https://github.com/filecoin-project/lotus/blob/58900a70333a11a903cf9fe3f29e6a5c309cb000/node/impl/full/eth.go#L251-L268

but https://github.com/filecoin-project/lotus/blob/99749e9b9bbe7d2ec89807d342fa63e3f1a4f2d8/node/impl/full/eth.go#L251-L265  was outdated


Stebalien (approved):
Would explain the failing test.

Stebalien:
Looks like something is still wrong with the go.mod file.

jennijuju:
> Would explain the failing test.

I thought was flakey..... my bad...

Stebalien:
> I thought was flakey..... my bad...

So did I.

PR #10445, opened 2023-03-10
feat: state: skip tipset execution when possible

Stebalien:
## Proposed Changes

Instead of always executing the tipset inside `StateManager.TipSetState`, try to find a successor tipset and return the parent state & receipts as recorded.

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [x] New features have usage guidelines and / or documentation updates in
  - [x] [Lotus Documentation](https://lotus.filecoin.io)
  - [x] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [x] CI is green

Issue #10288, opened 2023-02-16
ETH API: Remove message search from EthFeeHistory


Stebalien:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated improvement to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Improvement Suggestion

EthFeeHistory currently:

1. Loads all messages in a tipset.
2. Searches for the message receipts using `StateSearchMsg`.
3. Computes the "gas reward" from the message + receipt.

It also looks like it starts search at the current head.

This has a runtime of `(height_scanned*messages_per_tipset)^2`.

Given that we're walking backwards, we can instead:

1. Load the "parent receipts" from the current tipset.
2. Walk back one tipset.
3. Load the messages.
4. Use the receipts from step 1 and messages from step 3.
5. Repeat.

PR #10446, opened 2023-03-11
feat: eth: optimize eth block loading + eth_feeHistory

Stebalien:
## Related Issues

The current implementations of eth_getBlock* and eth_feeHistory have proven computationally expensive due to their usage of full tipset computations (`StateCompute`). This PR replaces the expensive logic with receipt lookups.

## Proposed Changes

For eth_getBlock* and eth_feeHistory, instead of replaying the tipset:

1. Lookup the tipset receipts.
2. Load the receipts.
3. Calculate everything from there (we have the gas used, etc.)

This also removes support for the `pending` block parameter as it's impossible to handle it without forcing state computes, which are prohibitively expensive. The Eth community is evaluating a proposal to also remove it in their next fork, given its limited applications due to MEV and other reasons.

## Additional Info

We've gone down from 10 seconds to various minutes (depending on the method and amount of data request) to almost instantaneous responses.

### After

```
⟩ time curl --silent --url http://localhost:1234/rpc/v1 \
      -X POST \
      -H "Content-Type: application/json" \
      -d '{"jsonrpc":"2.0","method":"eth_feeHistory","params":[ "0x08", "latest", [10, 50, 90] ],"id":1}'
{"jsonrpc":"2.0","result":{"oldestBlock":"0x28d39e","baseFeePerGas":["0x56ceb9b","0x56ceb9b","0x56ceb9b","0x56ceb9b","0x56ceb9b","0x56ceb9b","0x56ceb9b","0x56ceb9b","0x56ceb9b"],"gasUsedRatio":[1.1866695961,2.0196181532,2.5688713134,1.6794295302,1.7004244153,1.2444834388,1.736932316,2.2669179384],"reward":[["0x2de55df","0x2e29630","0x2e8e694"],["0x2d92b3b","0x2e665f8","0x5a874a8"],["0x294d17d","0x2e5547e","0x2eb2dbb"],["0x11c1e0d","0x2e35135","0x38a8b8f"],["0x11b8225","0x2e4b13e","0x39fb27c"],["0x2e18bb6","0x2e7c695","0x5d51c6a"],["0x2e0836d","0x2e70add","0x2ec2445"],["0x2e18763","0x2e429c9","0x2ebcc8a"]]},"id":1}

Executed in  345.14 millis    fish           external
   usr time    3.56 millis    0.12 millis    3.44 millis
   sys time    7.72 millis    3.36 millis    4.36 millis

~/W/pl/lotus · (steb/optimize-eth-block±)
⟩ time curl --silent --url http://localhost:1234/rpc/v1 \
      -X POST \
      -H "Content-Type: application/json" \
      -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":[ "latest", false ],"id":1}'
...

Executed in  366.47 millis    fish           external
   usr time    4.64 millis    0.12 millis    4.52 millis
   sys time   11.66 millis    4.58 millis    7.08 millis
```

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/eth_block_hash_test.go
+++ b/itests/eth_block_hash_test.go
@@ -49,9 +49,6 @@ func TestEthBlockHashesCorrect_MultiBlockTipset(t *testing.T) {
 	// let the chain run a little bit longer to minimise the chance of reorgs
 	n2.WaitTillChain(ctx, kit.HeightAtLeast(head.Height()+50))
 
-	head, err = n2.ChainHead(context.Background())
-	require.NoError(t, err)
-
 	for i := 1; i <= int(head.Height()); i++ {
 		hex := fmt.Sprintf("0x%x", i)
 
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -1797,30 +1797,42 @@ func newEthBlockFromFilecoinTipSet(ctx context.Context, ts *types.TipSet, fullTx
 		return ethtypes.EthBlock{}, xerrors.Errorf("error loading messages for tipset: %v: %w", ts, err)
 	}
 
-	block := ethtypes.NewEthBlock(len(msgs) > 0)
-
-	gasUsed := int64(0)
-	compOutput, err := sa.StateCompute(ctx, ts.Height(), nil, ts.Key())
+	_, rcptRoot, err := sa.StateManager.TipSetState(ctx, ts)
 	if err != nil {
 		return ethtypes.EthBlock{}, xerrors.Errorf("failed to compute state: %w", err)
 	}
 
-	txIdx := 0
-	for _, msg := range compOutput.Trace {
-		// skip system messages like reward application and cron
-		if msg.Msg.From == builtintypes.SystemActorAddr {
-			continue
-		}
+	rcpts, err := cs.ReadReceipts(ctx, rcptRoot)
+	if err != nil {
+		return ethtypes.EthBlock{}, xerrors.Errorf("error loading receipts for tipset: %v: %w", ts, err)
+	}
 
-		ti := ethtypes.EthUint64(txIdx)
-		txIdx++
+	if len(msgs) != len(rcpts) {
+		return ethtypes.EthBlock{}, xerrors.Errorf("receipts and message array lengths didn't match for tipset: %v: %w", ts, err)
+	}
 
-		gasUsed += msg.MsgRct.GasUsed
-		smsgCid, err := getSignedMessage(ctx, cs, msg.MsgCid)
-		if err != nil {
-			return ethtypes.EthBlock{}, xerrors.Errorf("failed to get signed msg %s: %w", msg.MsgCid, err)
+	block := ethtypes.NewEthBlock(len(msgs) > 0)
+
+	gasUsed := int64(0)
+	for i, msg := range msgs {
+		rcpt := rcpts[i]
+		ti := ethtypes.EthUint64(i)
+		gasUsed += rcpt.GasUsed
+		var smsg *types.SignedMessage
+		switch msg := msg.(type) {
+		case *types.SignedMessage:
+			smsg = msg
+		case *types.Message:
+			smsg = &types.SignedMessage{
+				Message: *msg,
+				Signature: crypto.Signature{
+					Type: crypto.SigTypeBLS,
+				},
+			}
+		default:
+			return ethtypes.EthBlock{}, xerrors.Errorf("failed to get signed msg %s: %w", msg.Cid(), err)
 		}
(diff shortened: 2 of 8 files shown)


Stebalien:
Blocked on #10445.

itests/eth_block_hash_test.go
```diff
@@ -49,9 +49,6 @@ func TestEthBlockHashesCorrect_MultiBlockTipset(t *testing.T) {
 	// let the chain run a little bit longer to minimise the chance of reorgs
 	n2.WaitTillChain(ctx, kit.HeightAtLeast(head.Height()+50))
 
-	head, err = n2.ChainHead(context.Background())
-	require.NoError(t, err)
```
raulk, line 53: Any particular reason to remove this?

raulk (approved):
This looks good, about to test it.

itests/eth_block_hash_test.go
Stebalien: It defeats the point of waiting 50 blocks for the chain to stabilize.

raulk: Not sure I follow the thinking? This is just a sanity check. Might be worthless but what point does it defeat exactly?

Stebalien: Unless we wait a bit after getting the head, there's no guarantee that `EthGetBlockByNumber` and `EthGetBlockByHash` will return the same value at the same height due to reorgs.

I _thought_ that that was the entire point of waiting an extra 50 blocks. If that wasn't, the comment:

> let the chain run a little bit longer to minimise the chance of reorgs

Is wrong.

raulk: I see, all cleared up. `head` is being used below to traverse the chain up until that point. We do not need to traverse the extra 50 tipsets, those are there for stability. This is fine.

raulk (approved):
I migrated `eth_feeHistory` to reading receipts, and dropped support for the `pending` block parameter. Fleshed out the description and added numbers. This is ready to go.

raulk:
@Stebalien can you review my changes?

node/impl/full/eth.go
```diff
@@ -689,49 +689,43 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
...
 		if err != nil {
-			return ethtypes.EthFeeHistory{}, xerrors.Errorf("cannot lookup the status of tipset: %v: %w", ts, err)
+			return ethtypes.EthFeeHistory{}, xerrors.Errorf("failed to retrieve messages and receipts for height %d: %w", ts.Height(), err)
 		}
 
 		txGasRewards := gasRewardSorter{}
-		for _, msg := range compOutput.Trace {
-			if msg.Msg.From == builtintypes.SystemActorAddr {
+		for i, msg := range msgs {
+			if msg.VMMessage().From == builtintypes.SystemActorAddr {
```
Stebalien, line 713: This isn't necessary anymore. The messages in the tipset will never be from the system.

node/impl/full/eth.go
```diff
@@ -689,49 +689,43 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
...
+		baseFeeArray      = []ethtypes.EthBigInt{ethtypes.EthBigInt(basefee)}
+		rewardsArray      = make([][]ethtypes.EthBigInt, 0)
+		gasUsedRatioArray = []float64{}
+		blocksIncluded    int
+	)
 
-	blocksIncluded := 0
 	for blocksIncluded < int(params.BlkCount) && ts.Height() > 0 {
-		compOutput, err := a.StateCompute(ctx, ts.Height(), nil, ts.Key())
+		msgs, rcpts, err := messagesAndReceipts(ctx, ts, a.Chain, a.StateAPI)
```
Stebalien, line 706: This works but it isn't great. Ideally we'd just walk the chain backwards as described in #10288.

Stebalien (commented):
I can't approve but the new changes look correct.

Stebalien:
It looks like we need to fix the conformance tests to drop the "pending" one.

node/impl/full/eth.go
raulk: Aren't the mechanics the same?

icodeface:
How to get correct nonce with pending txs after "pending" param is dropped ?

go-lifei:
> How to get correct nonce with pending txs after "pending" param is dropped ?

        let nonce = await ethers.provider.getTransactionCount(
            addr,
            'latest'
        )

icodeface:
> getTransactionCount

This doesn't take into transactions in the mempool, which could cause nonce collisions.

PR #10451, opened 2023-03-11
gateway: fix: drop overzealous guard on MsigGetVested

arajasek:
## Related Issues

Folks were trying to use the chain.love gateway to get info about a msig's vesting funds, starting from genesis. This didn't work, because we guard on the start tipset (making sure it's within the last 24 hours).

## Proposed Changes

There's no actual reason to guard on the start tipset, though -- we're not actually loading its state, just its height. We could honestly change this API to take a start height, it'd be better.

But the quick solution is just to drop the guard on the start tipset for the gateway.

## Checklist

- [ ] CI is green


raulk (approved):
Just learnt something new about how this method works (went digging to understand better...)

Issue #10326, opened 2023-02-22
Better Lotus indices


raulk:
## Context

Lotus’ chain and state storage revolves around a monolithic blockstore that contains raw chain and state data, referred to by CID. There exists a skiplist of size 10 for chain lookbacks (which was known to be flaky in the past, probably fixed now? would help to have utilisation stats to confirm).

However, there are no indices to facilitate common chain queries. For example, loading a tipset by height, finding the tipset/blocks where a message was included, finding messages involving specific actors within a particular time range, etc. are costly operations. Most of these queries involve walking chain history and/or performing a number of sequential queries through traversals.

## Goal

Lotus optimizes common data access patterns to reduce latency, disk I/O and computational cost, for a better developer and node operator experience.

## Change motivation

With FEVM about to be launch, we expect the programmatic usage of Lotus' API is going to increase. Wallets, dapps, libraries, and utilities will all start consuming Lotus' API through both the Eth JSON-RPC API, and its Filecoin-specific API.

## Asks

Introduce indices and support data structures to enable these access patterns.

**High-prio**

1. Randomly seek to a chain height -- make O(1) instead of O(n).
2. Identify which tipset and blocks a message (identified by CID) landed in. -- make O(1) instead of O(n^m).
3. Load the receipt for a given message (identified by CID). -- make O(1) instead of O(n^m+1).

**Medium-prio**

1. Ability to get gas used of a tipset instantaneously.
2. Ability to cache the canonical order of messages in a tipset.
3. Ability to quickly know the message count in a tipset.

## Considerations

- Need to deal with reorgs.
- Need to deal with splistore (whatever archiving policy is chosen may need to extend to indices).
- Consider garbage collection for indices.
- Figure out the right sync/async mix to ensure consistency yet not block chain sync pathways.
- Make sure indices are consistent across load balanced Lotus (maybe nothing to do here).

arajasek:
We (@raulk @jennijuju @vyzo @Stebalien @magik6k @Kubuxu ) had an extended discussion on this today. We emerged with the following chief themes:

- Solving the reorg problem is going to be _hard_, any indices should be untouchable by reorgs (past finality only)
- The primary index needed is a mapping from message CID to tipset in which it was included. Everything else is either quickly computable once we have that, or already fast enough today.
- The index is only as useful as the amount of chain history (including state) that one has.

Given that, the concrete proposal is to:

[...]

PR #10452, opened 2023-03-11
[WIP] Message Index

vyzo:
See #10326 

## Design Overview
This departs from the discussion in #10326 by implementing a simpler solution that has no memory component and stores everything in sqlite.
On head change notifications, we first delete data from reverted tipsets and then insert data for applied tipsets (in a transaction).
This allows us to trivially deal with finality commitments.

### DIFF
--- a/chain/index/interface.go
+++ b/chain/index/interface.go
@@ -1,0 +1,31 @@
+package index
+
+import (
+	"context"
+
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/ipfs/go-cid"
+)
+
+// MsgInfo is the Message metadata the index tracks.
+type MsgInfo struct {
+	// the message this record refers to
+	Message cid.Cid
+	// the epoch whre this message was executed
+	Epoch abi.ChainEpoch
+	// the tipset where this messages executed
+	Tipset types.TipSetKey
+	// the first block in the tipset where the message was executed
+	Block cid.Cid
+	// the index of the message in the block
+	Index int
+}
+
+// MsgIndex is the interface to the message index
+type MsgIndex interface {
+	// GetMsgInfo retrieves the message metadata through the index.
+	GetMsgInfo(ctx context.Context, m cid.Cid) (MsgInfo, error)
+	// Close closes the index
+	Close() error
+}
--- a/chain/index/msgindex.go
+++ b/chain/index/msgindex.go
@@ -1,0 +1,141 @@
+package index
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"io/fs"
+	"os"
+	"path"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
+	_ "github.com/mattn/go-sqlite3"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/lotus/chain/store"
+	"github.com/filecoin-project/lotus/chain/types"
+)
+
+type msgIndex struct {
+	cs *store.ChainStore
+
+	db *sql.DB
+}
+
+var _ MsgIndex = (*msgIndex)(nil)
+
+var log = logging.Logger("chain/index")
+
+var (
+	dbName = "msgindex.db"
+
+	coalesceMinDelay      = 100 * time.Millisecond
+	coalesceMaxDelay      = time.Second
+	coalesceMergeInterval = 100 * time.Millisecond
+)
+
+func NewMsgIndex(basePath string, cs *store.ChainStore) (MsgIndex, error) {
+	var (
+		mkdb   bool
+		dbPath string
+		err    error
+	)
+
+	if basePath == ":memory:" {
+		// for testing
+		mkdb = true
+		dbPath = basePath
+		goto opendb
+	}
+
+	err = os.MkdirAll(basePath, 0755)
+	if err != nil {
+		return nil, xerrors.Errorf("error creating msgindex base directory: %w", err)
+	}
+
+	dbPath = path.Join(basePath, dbName)
+	_, err = os.Stat(dbPath)
+	switch {
+	case errors.Is(err, fs.ErrNotExist):
+		mkdb = true
+
+	case err != nil:
+		return nil, xerrors.Errorf("error stating msgindex database: %w", err)
+	}
+
+opendb:
+	db, err := sql.Open("sqlite3", dbPath)
+	if err != nil {
+		// TODO [nice to have]: automaticaly delete corrupt databases
+		//      but for now we can just error and let the operator delete.
+		return nil, xerrors.Errorf("error opening msgindex database: %w", err)
+	}
+
+	if mkdb {
+		err = createTables(db)
+		if err != nil {
+			return nil, xerrors.Errorf("error creating msgindex database: %w", err)
+		}
+	} else {
+		err = reconcileIndex(db, cs)
+		if err != nil {
+			return nil, xerrors.Errorf("error reconciling msgindex database: %w", err)
+		}
+	}
+
+	msgIndex := &msgIndex{db: db, cs: cs}
+	err = msgIndex.prepareStatements()
+	if err != nil {
+		err2 := db.Close()
+		if err2 != nil {
+			log.Errorf("error closing msgindex database: %s", err2)
+		}
+
+		return nil, xerrors.Errorf("error preparing msgindex database statements: %w", err)
+	}
+
+	rnf := store.WrapHeadChangeCoalescer(
+		msgIndex.onHeadChange,
+		coalesceMinDelay,
+		coalesceMaxDelay,
+		coalesceMergeInterval,
+	)
+	cs.SubscribeHeadChanges(rnf)
+
+	return msgIndex, nil
+}
+
+// init utilities
+func createTables(db *sql.DB) error {
+	// TODO
+	return errors.New("TODO: index.createTables")
+}
+
+func reconcileIndex(db *sql.DB, cs *store.ChainStore) error {
+	// TODO
+	return errors.New("TODO: index.reconcileIndex")
+}
+
+func (x *msgIndex) prepareStatements() error {
+	// TODO
+	return errors.New("TODO: msgIndex.prepareStatements")
+}
+
+// head change notifee
+func (x *msgIndex) onHeadChange(rev, app []*types.TipSet) error {
+	// TODO
+	return errors.New("TODO: msgIndex.onHeadChange")
+}
+
+// interface
+func (x *msgIndex) GetMsgInfo(ctx context.Context, m cid.Cid) (MsgInfo, error) {
+	// TODO
+	return MsgInfo{}, errors.New("TODO: msgIndex.GetMsgInfo")
+}
+
+func (x *msgIndex) Close() error {
+	// TODO
+	return errors.New("TODO: msgIndex.Close")
+}


chain/index/interface.go
```diff
@@ -0,0 +1,31 @@
...
+	// the message this record refers to
+	Message cid.Cid
+	// the epoch whre this message was executed
+	Epoch abi.ChainEpoch
+	// the tipset where this messages executed
+	Tipset types.TipSetKey
+	// the first block in the tipset where the message was executed
+	Block cid.Cid
+	// the index of the message in the block
+	Index int
```
raulk, line 22: I think you can remove the Block and instead have the canonicalized execution index (which is the index at which the receipt will be stored in the AMT).

vyzo: ok, I will see whether I can get rid of it when I get to implementation details.

chain/index/interface.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// MsgInfo is the Message metadata the index tracks.
+type MsgInfo struct {
+	// the message this record refers to
+	Message cid.Cid
+	// the epoch whre this message was executed
+	Epoch abi.ChainEpoch
+	// the tipset where this messages executed
+	Tipset types.TipSetKey
```
raulk: Let's use a TipsetCID instead. The TipSetKey is wasteful. It will imply an extra lookup to go from TipsetCid to TipsetKey, but it's worth it to save an average of 4x times the amount of bytes that would otherwise go here, on every record. Discarding a second table + foreign key + joins because that's too much complexity and antithetical to the main goal of speeding things up.

vyzo: done.

chain/index/interface.go
vyzo: Yeah, we don't need it; dropped it from the interface.

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+		// TODO [nice to have]: automaticaly delete corrupt databases
+		//      but for now we can just error and let the operator delete.
+		return nil, xerrors.Errorf("error opening msgindex database: %w", err)
+	}
+
+	if mkdb {
+		err = createTables(db)
+		if err != nil {
+			return nil, xerrors.Errorf("error creating msgindex database: %w", err)
+		}
```
raulk, line 103: Can save all this dance by using `CREATE TABLE IF NOT EXISTS`. See https://github.com/filecoin-project/lotus/blob/master/chain/events/filter/index.go#L32-L61.

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+
+		// TODO we may consider populating the index in this case.
+	} else {
+		err = reconcileIndex(db, cs)
+		if err != nil {
+			return nil, xerrors.Errorf("error reconciling msgindex database: %w", err)
+		}
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
raulk, line 113: This isn't the right context to use. Let's make the constructor accept a context, like the [Mpool](https://github.com/filecoin-project/lotus/blob/58900a70333a11a903cf9fe3f29e6a5c309cb000/chain/messagepool/messagepool.go#L371), and feed the lifecycle context during dependency injection.

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+		db:     db,
+		cs:     cs,
+		sema:   make(chan struct{}, 1),
+		cancel: cancel,
+	}
+
+	err = msgIndex.prepareStatements()
+	if err != nil {
+		err2 := db.Close()
+		if err2 != nil {
```
raulk, line 125:
```suggestion
		if err := db.Close(); err != nil {
```

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+	msgIndex.workers.Add(1)
+	go msgIndex.background(ctx)
+
+	return msgIndex, nil
+}
+
+// init utilities
+func createTables(db *sql.DB) error {
+	// Just a single table for now; ghetto, but this an index so we denormalize to avoid joins.
+	if _, err := db.Exec("CREATE TABLE Messages (cid VARCHAR(80) PRIMARY KEY, tipset VARCHAR(80), xepoch INTEGER, xindex INTEGER)"); err != nil {
```
raulk, line 149: Can we please use the same pattern for initializing tables as the event index? https://github.com/filecoin-project/lotus/blob/master/chain/events/filter/index.go#L32-L61

- Have the DDLs hoisted as a global var for better visiblity.
- Use CREATE IF NOT EXISTS
- Add a version table to facilitate future migrations.
- Lower case table name
- `cid` should be of BLOB type?
- same for `tipset`
- `exec_epoch` and `exec_index` please -- costs nothing and it's immediately obvious.

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+		return err
+	}
+
+	// TODO Should we add an index for tipset to speed up deletion on revert?
```
raulk, line 153: Yes, otherwise the cost of reverts will scale with the size of the DB, and that will be problematic.

raulk (commented):
Going through it, expect more comments shortly.

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+	return nil
+}
+
+func reconcileIndex(db *sql.DB, cs ChainStore) error {
+	// Invariant: after reconciliation, every tipset in the index is in the current chain; ie either
+	//  the chain head or reachable by walking the chain.
+	// Algorithm:
+	//  1. Count mesages in index; if none, trivially reconciled.
```
raulk, line 161:
```suggestion
	//  1. Count messages in index; if none, trivially reconciled.
```

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+	if err != nil {
+		return xerrors.Errorf("prepare insertMsgStmt: %w", err)
+	}
+	x.insertMsgStmt = stmt
+
+	stmt, err = x.db.Prepare("DELETE FROM Messages WHERE tipset = ?")
+	if err != nil {
+		return xerrors.Errorf("prepare deleteTipSetStmt: %w", err)
+	}
+	x.deleteTipSetStmt = stmt
```
raulk, line 244: Let's hoist these as consts, makes them more visible.

See https://github.com/filecoin-project/lotus/blob/master/chain/events/filter/index.go#L65-L73.

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+			return err
+		}
+		seen[key] = struct{}{}
+		xindex++
+
+		return nil
+	}
+
+	for _, blk := range ts.Blocks() {
+		bmsgs, smsgs, err := x.cs.MessagesForBlock(ctx, blk)
```
raulk, line 357: This should use `MessagesForTipset` instead.

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+}
+
+func (x *msgIndex) doApply(ctx context.Context, tx *sql.Tx, ts *types.TipSet) error {
+	tscid, err := ts.Key().Cid()
+	if err != nil {
+		return xerrors.Errorf("error computing tipset cid: %w", err)
+	}
+
+	tskey := tscid.String()
+	xepoch := int64(ts.Height())
```
raulk, line 338: This is not the execution epoch, but the _inclusion_ epoch.

chain/index/msgindex.go
vyzo: we can, but we still need to dance for reconciliation.

chain/index/msgindex.go
vyzo: I disagree in this. This is a background component, it is certainly the right context.

Passing a lifecycle context is an antipattern, and we are doing it wrong.
I am sure @stebalien has strong opinions here.

chain/index/msgindex.go
vyzo: sure.

chain/index/msgindex.go
vyzo: Ok.

Cid shouldnt be blib if we want to primary key it imho.

vyzo: Same for tipset if we use an index, which we will.

chain/index/msgindex.go
vyzo: yeah, makes sense, will add.

chain/index/msgindex.go
```diff
@@ -0,0 +1,433 @@
...
+	return nil
+}
+
+func (x *msgIndex) background(ctx context.Context) {
+	defer x.workers.Done()
+
+	for {
+		select {
+		case <-x.sema:
+			err := x.processHeadChanges(ctx)
```
raulk, line 278: Unfortunately this logic is not as straightforward as it appears here. This index only cares about executed tipsets, which means that we need to consume the parent of each head, but only when the head has actually been executed itself. Sadly this logic consumes speculative tipsets while they are being formed, which is not what we want.

In other words, we should track tipsets with a **confidence of 2**, i.e. once the **execution** tipset has been committed to chain. Then we can add the messages from the execution tipset, which is head.height - 2 at minimum, or more if there are null rounds.

```
inclusion ts <--- execution ts <--- head
```

---

Also, we should extract and reuse the [event observer](https://github.com/filecoin-project/lotus/blob/e65fae28de2a44947dd24af8c7dafcade29af1a4/chain/events/observer.go#L19) whcih already tracks tipsets, performs reverts, and does reconciliation by replaying a stream of reverts and applies for a particular chain path.

That component is actually not interdependent with the event index, so it can be extracted easily!

raulk (commented):
More comments, let's reuse the event observer logic by extracting it first to a common component.

chain/index/msgindex.go
raulk: Aren't index initialization and reconciliation independent of one another?

chain/index/msgindex.go
raulk: We shouldn't be using contexts for lifecycle management, agree, but that's Lotus' reality today and we should just adhere to it. If we want to move away from that pattern, it should be done as part of a larger refactor, not in a piecemeal way.

chain/stmgr/searchwait.go
```diff
@@ -159,6 +173,21 @@ func (sm *StateManager) SearchForMessage(ctx context.Context, head *types.TipSet
...
+	}
+
+	ts, err := sm.cs.GetTipSetByCid(ctx, minfo.TipSet)
+	if err != nil {
+		return nil, nil, cid.Undef, err
+	}
+
+	r, foundMsg, err := sm.tipsetExecutedMessage(ctx, ts, mcid, m.VMMessage(), false)
+	return ts, r, foundMsg, err
+}
```
raulk, line 189: We should use the execution tipset and the execution index to just rescue the receipt and call it a day here, instead of handing off to `tipsetExecutedMessage`. However, @arajasek wanted to contain the changeset leaving out the execution index from the index, in which case this logic can stay.

chain/index/msgindex.go
vyzo: ack, will fix.

vyzo:
We discussed sync, and we think that:
1. It is ok to use the inclusion index.
2. We can drop the execution index.

In this light, we probably need to do very little. 

We can even keep the coaleacer instead of the event observer. Also note that confidence can be implemented at read time as well, by comparing inclusion height with the current head.

raulk (commented):
Just had a chat with @vyzo. While having the execution tipset and the execution index would be ideal, it'll be easier to get this merged if we stick to what the discussion in #10326 ends up suggesting. So:

1. Let's store the inclusion tipset, which is what this logic is actually storing now (even though the field is incorrectly named as _execution_ tipset -- fix that).
2. Let's remove the execution index.
3. Reuse the events observer, extracting it from the `chain/events` package since it's not coupled to events.
4. Set a confidence of 1 (see how [the `headEventsObserver` does it](https://github.com/filecoin-project/lotus/blob/e65fae28de2a44947dd24af8c7dafcade29af1a4/chain/events/events_height.go#L157)). Not 2 since we're tracking the inclusion tipset and not the execution tipset now.
5. Align the code with the existing event index to avoid code entropy.

## TODOs in addition to the above

(perhaps incomplete, @vyzo please add if I omitted something)

- [ ] Add itests.
- [ ] Add a backfill command in lotus-shed.
- [ ] Make sure we're using the right sqlite pragmas to: (1) fsync on transaction commit, and (2) allow to open the sqlite db from multiple processes (so that the backfill script can operate while the node is live).

vyzo:
Checklist looks good, but we might want to punt on it a bit if it is tricky.

It looks like the observer might be trivial to integrate to use it instead of the coalescer. 

A couple if additional considerations:
- backfill at startup the first time we create the index, for some epoch range. Not sure how good an idea it is, compared to shed, but thats why i wanr to punt on it in this pr.
- we may want to disable the fallback to prevent dos vectors. At the very least provide an option to limit the range may br necessary.

raulk (commented):
Another thing, and this one may lead to more discussion.

**We should not fall back to a full search on miss.**

These guardrails are _not_ safe to have unless we bound the lookup, which we can't do because AFAIK Eth tools expect to be able to access historical receipts.

Note that Eth clients can serve receipts with a single DB lookup, and node operators like Infura, etc. offer instantaneous access to historical receipts.

The receipt for [this transaction](https://etherscan.io/tx/0xfaa6150811c3e9f377fb6872a8cd5a92daa38a98ec7a40f3c848e4c76c460c89) from 7+ years ago was loaded in sub-second response times:

```
⟩ time curl --url https://mainnet.infura.io/v3/<key> \
      -X POST \
      -H "Content-Type: application/json" \
      -d '{"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["0xfaa6150811c3e9f377fb6872a8cd5a92daa38a98ec7a40f3c848e4c76c460c89"],"id":1}'
{"jsonrpc":"2.0","id":1,"result":{"blockHash":"0xe33a89a33196a9c6f11a51940823711b18ce10800a5c49fe65f473c75b430965","blockNumber":"0x7a126","contractAddress":null,"cumulativeGasUsed":"0x5208","effectiveGasPrice":"0xba43b7400","from":"0x935e1a56bf95c223e93a41e8bdf63959e99ae080","gasUsed":"0x5208","logs":[],"logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","root":"0xc1d169faae783f0a94f4edc1f50f44ab3179a053949eede9d13db674e678c819","to":"0x8263b7757b139904c7cd8511b3d0a6d1479848ba","transactionHash":"0xfaa6150811c3e9f377fb6872a8cd5a92daa38a98ec7a40f3c848e4c76c460c89","transactionIndex":"0x0","type":"0x0"}}
Executed in  422.21 millis    fish           external
   usr time   14.40 millis    0.07 millis   14.34 millis
   sys time   11.65 millis    3.83 millis    7.82 millis
```

raulk (commented):
One more thing: we need to populate the index on snapshot import (added to the TODO list above)

chain/index/msgindex.go
vyzo: yeah, all we need to do is stat to know if it already exists.
Removed the unnecessary parts of the dance.

chain/index/msgindex.go
vyzo: I will invoke the @Stebalien shell here. Codewise it is easy to hang this context from the lifecycle context.

chain/index/msgindex.go
vyzo: Done:
- hoisted DDLs and stmts on top to vars and consts
- used CREATE IF NOT EXISTS
- added version table
- lower cased table name
- renamed tipset to tipset_cid, xepoch to just epoch, and dropped xindex.

See above for why I am keeping VARCHAR(80) for the cid types:
- we need primary key for cid
- we need index for tipset_cid

chain/index/msgindex.go
vyzo: done.

chain/index/msgindex.go
vyzo: fixed.

chain/index/msgindex.go
vyzo: For readers of this pr: see comments below, we are going inclusion tipset.

chain/stmgr/searchwait.go
vyzo: we dropped the index from the index.

chain/index/msgindex.go
vyzo: done.

vyzo:
Update: points (1), (2), and (5) addressed; will reify and use the events observer next instead of a head change coalescer.

chain/index/msgindex.go
magik6k: The lifecycle context should be fine to use. If it is created before the OnClose hook is set, it is guaranteed to be cancelled after Close is called - which is the whole point of the lifecycle system. Lifecycle contexts are somewhat of a hack, but in most cases they just make cleanup of things slightly more robust.

The other thing that context is used for is metrics / tracing, and in this specific case I believe there actually are metrics which will be dropped due to this background context not being wired to the main MetricsCtx

vyzo: well ok, i will wire it then.

vyzo: done.

magik6k (commented):
Looks reasonable, couple of comments.

I don't have enough context to have strong opinions on exact semantics, so I'll leave that to @raulk

chain/index/msgindex.go
```diff
@@ -0,0 +1,438 @@
...
+    	version UINT64 NOT NULL UNIQUE
+	)`,
+	`INSERT OR IGNORE INTO _meta (version) VALUES (1)`,
+}
+var dbPragmas = []string{}
+
+const (
+	// prepared stmts
+	dbqGetMessageInfo       = "SELECT tipset_cid, epoch FROM messages WHERE cid = ?"
+	dbqInsertMessage        = "INSERT INTO messages VALUES (?, ?, ?)"
```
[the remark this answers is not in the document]
vyzo: there should be no conflict, really.

But maybe we do want some policy.

chain/index/msgindex.go
```diff
@@ -0,0 +1,438 @@
...
+
+	tx, err := x.db.Begin()
+	if err != nil {
+		return xerrors.Errorf("error creating transaction: %w", err)
+	}
+
+	for _, hc := range pend {
+		for _, ts := range hc.rev {
+			if err := x.doRevert(ctx, tx, ts); err != nil {
+				tx.Rollback() //nolint
```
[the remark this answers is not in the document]
vyzo: ack, will do.

node/modules/msgindex.go
```diff
@@ -0,0 +1,36 @@
...
+	lc.Append(fx.Hook{
+		OnStop: func(_ context.Context) error {
+			return msgIndex.Close()
+		},
+	})
+
+	return msgIndex, nil
+}
+
+func DummyMsgIndex() (index.MsgIndex, error) {
```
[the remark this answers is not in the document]
vyzo: ok, will just drop the error :)

node/builder_chain.go
```diff
@@ -79,6 +80,7 @@ var ChainNode = Options(
 	Override(new(stmgr.Executor), consensus.NewTipSetExecutor(filcns.RewardFunc)),
 	Override(new(consensus.Consensus), filcns.NewFilecoinExpectedConsensus),
 	Override(new(*store.ChainStore), modules.ChainStore),
+	Override(new(index.MsgIndex), modules.MsgIndex),
```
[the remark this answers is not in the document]
vyzo: yeah, good point.
Should I add a section in the config with an enabled bool? We might want to add more knobs.

vyzo:
@raulk the events observer is not as straightforward to extract, as (1) it has a sprawl of dependencies on other things in the package and (2) it has no tests!

I think we should do this extraction and replacement of the coalescer in a follow up pr.
For now, we can add a check on read and wait until the height exceeds the record height before returning it, effectively implementing confidence.

Note: I do agree that the events observer is more efficient, as we do (potentially a lot) less db i/o. But I don't think it is prudent to do it now.

Opinions? @magik6k @arajasek yours matters too.

chain/index/interface.go
```diff
@@ -0,0 +1,43 @@
...
+	// the tipset where this messages was included
+	TipSet cid.Cid
+	// the epoch whre this message was included
+	Epoch abi.ChainEpoch
+}
+
+// MsgIndex is the interface to the message index
+type MsgIndex interface {
+	// GetMsgInfo retrieves the message metadata through the index.
+	GetMsgInfo(ctx context.Context, m cid.Cid) (MsgInfo, error)
```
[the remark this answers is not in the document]
vyzo: it's the chain message cid. 

So it is signed for SECP messages, and (if I understand the whole bls thingie correctly) unsigned for BLS? Sigh, this thing drives me nuts.

vyzo:
Actually we don't even need to add the check in the implementation, the client can implement its own confidence by comparing with the head -- this in stmgr.

Also, we can increase the coalescer delay all the way to 30s probably (and override in the unit tests so that they don't last forever.

chain/index/msgindex.go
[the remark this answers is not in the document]
vyzo: done.

node/modules/msgindex.go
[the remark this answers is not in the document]
vyzo: done.

vyzo:
Re: fsync: It looks like we don't have to do anything to get synchronous writes for transactions.
See https://www.sqlite.org/pragma.html#pragma_synchronous

chain/index/interface.go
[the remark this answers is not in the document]
magik6k: Yeah, bls is unsigned on chain.

vyzo:
Re: multiprocess access to db: https://www.sqlite.org/pragma.html#pragma_locking_mode

Here again it looks like we don't have to do anything.

node/builder_chain.go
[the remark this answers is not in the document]
vyzo: https://github.com/filecoin-project/lotus/pull/10452/commits/db8b593c9bb751128893fa2809ea2c0720bbba64

chain/index/interface.go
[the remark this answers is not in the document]
vyzo: https://github.com/filecoin-project/lotus/pull/10452/commits/5e011d536b0ff2af2cc2958b9643dc420876e118

vyzo:
hrm, CI is doing something weird with my itest:
```
# command-line-arguments [command-line-arguments.test]
itests/msgindex_test.go:88:11: undefined: apiSuite
WARN invalid TestEvent: FAIL	command-line-arguments [build failed]
bad output from test2json: FAIL	command-line-arguments [build failed]

=== Errors
itests/msgindex_test.go:88:11: undefined: apiSuite

DONE 0 tests, 1 error in 108.253s

Exited with code exit status 2

```

needless to say, it works fine locally.
@magik6k any ideas here?

cmd/lotus-shed/msgindex.go
```diff
@@ -0,0 +1,206 @@
...
+			return err
+		}
+
+		startHeight := int64(cctx.Int("from"))
+		if startHeight == 0 {
+			startHeight = int64(curTs.Height()) - 1
+		}
+		epochs := cctx.Int("epochs")
+
+		dbPath := path.Join(cctx.String("repo"), "sqlite", "msgindex.db")
```
magik6k, line 67: Is `~/.lotus` actually expanded by sql.Open?

magik6k (approved):
LGTM, tho @raulk should also have a chance to review before merging.

chain/index/interface.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+var ErrNotFound = errors.New("message not found")
+var ErrClosed = errors.New("index closed")
+
+// MsgInfo is the Message metadata the index tracks.
+type MsgInfo struct {
+	// the message this record refers to
+	Message cid.Cid
+	// the tipset where this messages was included
```
arajasek, line 19:
```suggestion
	// the tipset where this message was included
```

chain/stmgr/searchwait.go
```diff
@@ -159,6 +173,41 @@ func (sm *StateManager) SearchForMessage(ctx context.Context, head *types.TipSet
...
+	}
+
+	// now get the execution tipset
+	// TODO optimization: the index should have it implicitly so we can return it in the msginfo.
+	xts, err := sm.cs.GetTipsetByHeight(ctx, minfo.Epoch+1, curTs, false)
+	if err != nil {
+		return nil, nil, cid.Undef, err
+	}
+
+	// check that it is indeed the parent of the inclusion tipset
```
arajasek, line 196: This comment is inverted -- the inclusion tipset is the parent of xts, not the other way around.

chain/stmgr/searchwait.go
```diff
@@ -145,7 +146,20 @@ func (sm *StateManager) SearchForMessage(ctx context.Context, head *types.TipSet
 		return head, r, foundMsg, nil
 	}
 
-	fts, r, foundMsg, err := sm.searchBackForMsg(ctx, head, msg, lookbackLimit, allowReplaced)
+	fts, r, foundMsg, err := sm.searchForIndexedMsg(ctx, mcid, msg)
```
arajasek, line 149: As currently written, this will not benefit calls to `WaitForMessage` (the API's `StateWaitMsg`). You can either perform a quick refactor here to speed up that method too, or just open an enhancement issue?

chain/stmgr/searchwait.go
```diff
@@ -159,6 +173,41 @@ func (sm *StateManager) SearchForMessage(ctx context.Context, head *types.TipSet
...
+	if err != nil {
+		return nil, nil, cid.Undef, xerrors.Errorf("error computing tipset cid: %w", err)
+	}
+
+	if !parentCid.Equals(minfo.TipSet) {
+		return nil, nil, cid.Undef, xerrors.Errorf("inclusion tipset mismatch: have %s, expected %s", parentCid, minfo.TipSet)
+	}
+
+	r, foundMsg, err := sm.tipsetExecutedMessage(ctx, xts, mcid, m.VMMessage(), false)
+	return xts, r, foundMsg, err
```
arajasek, line 208: This is risky -- it's possible that `tipsetExecutedMessage` "succeeds" (doesn't error), [without actually finding the message](https://github.com/filecoin-project/lotus/blob/47646cbd6966ee440da5ffe7933e73ec67853f39/chain/stmgr/searchwait.go#L333). As written, we `SearchForMessage` will successfully return incomplete / wrong information in that case (see line 153).

We should check that `r` and `foundMsg` are non-nil / non-undef either here, or in callers of this (or both).

arajasek (commented):
Thanks for the PR! This generally looks good. I have a few open questions:

- Did you try / can we try having the ChainStore own this index? I think it belongs there more, "searching back for a message" is very much a chain operation (even if its caller is currently the stmgr).
- As written, we can only backfill manually. That means if your node is down for 30 mins and then rebooted, you'll be missing index entries for that period. Are we okay with that?
- We currently only reconcile on startup. Did you consider triggering the reconciliation process internally if the indexer ends up in some wonky state (eg. errors while processing head change). It also might be useful for callers of the index to be able to "tell" the indexer that it's inconsistent, and trigger a reconciliation (eg. if for some reason we didn't _actually_ find the message in the returned tipset)
- On the flip-side, as written with the current user of this index, we _only_ need the Tipset CID in the MsgInfo struct for reconciliation, a process we don't expect to need -- we don't actually need it for the lookback, because we load the "execution" tipset and pull the info out of it anyway. This means that we could drop the size of this db by ~half (one of the 2 CIDs), and have this just me a `(msgCid, height)` mapping. In essence we can get a much smaller database footprint (50%) at the cost of much slower / worse reconciliation (either manually inspecting every record, or just dropping any records that callers report as "bad").
- I have pretty high confidence in the correctness of this thanks to how simple it is. Even so, could we consider adding a "paranoid" mode, that uses the cache, then does the lookback anyway, and screams if results don't match? This can be run by us, as well as any integration partners with high security standards for this portion of the codebase (we have some such partners).
- I think it'll be the case that if this is slow, it'll slow down overall sync. Is that so, and if so can we maybe add some timing metrics?
- I'm strongly in favour of having a single table with the minor duplication here :+1:

chain/index/msgindex.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+
+	boundaryEpoch := abi.ChainEpoch(result)
+
+	countMsgsStmt, err := db.Prepare(dbqCountTipsetMessages)
+	if err != nil {
+		return xerrors.Errorf("error preparing statement: %w", err)
+	}
+
+	curTs := cs.GetHeaviestTipSet()
```
Stebalien, line 224: This is racy unless we can guarantee that the chainstore hasn't been "started" yet. Is that the case @arajasek?

Otherwise, we need a way to atomically get the head and subscribe at the same time. Alternatively, we can defer this reconcile step until we get the first event from the chainstore and reconcile then.

Stebalien: Actually, that second part is kind of nice because it avoids blocking on start.

chain/index/msgindex.go
```diff
@@ -0,0 +1,458 @@
...
+			return nil, xerrors.Errorf("error reconciling msgindex database: %w", err)
+		}
+	}
+
+	ctx, cancel := context.WithCancel(lctx)
+
+	msgIndex := &msgIndex{
+		db:     db,
+		cs:     cs,
+		sema:   make(chan struct{}, 1),
```
Stebalien, line 145: ultra-nit: this is a flag, not a semaphore.

vyzo:
@arajasek thanks for the review, these are some great questions.

> Did you try / can we try having the ChainStore own this index? I think it belongs  there more, "searching back for a message" is very much a chain operation (even if its caller is currently the stmgr).

I have considered this and I think it has _great_ potential. 
If we had the index owned by the ChainStore a whole lot of things could be significantly improved in terms of performance.

Some very basic things like `GetTipsetByHeight`, `GetMessagesForTipset`, etc can go through the index.

The main reason  I didn't do it for the MVP is that it is a much more invasive change in the most consensus critical part of the system. So a different standard of applicability and testing must be adhered there, and I'd rather do it in a follow up.

The main question is whether it needs to become synchronous in terms of head change notification processing in order to make ChainStore keep the authoritative source of truth invariant. We discussed this with @Stebalien and we think it doesn't have to, we can check if the current tipset (or the one we are based on in general) is in the index, and in that case proceed and use it -- otherwise fallback to what we already have. We should probably tweak down the coalescer delay a bit too, and generally experiment with it. Note that the reason for not wanting synchronous operation is that we will thrash the database with writes/deletes while the tipset is building and this is something we want to avoid for obvious reasons (perf, durability of SSDs, and so on).

At any rate, I think that having the index owned by the ChainStore is a natural progression from the MVP and a very important follow up for subsequent prs.

> As written, we can only backfill manually. That means if your node is down for 30 mins and then rebooted, you'll be missing index entries for that period. Are we okay with that?

I think we are ok for MVP, but we do need to do something about it. The first thing to do (as discussed with @raulk) is to fill the index on snapshot import, and that should probably cover us. We can also consider backfilling when first creating the index as well, but I think snapshot import is a better place to handle this. 

At any rate, part of the follow up plan to address this.

> We currently only reconcile on startup. Did you consider triggering the reconciliation process internally if the indexer ends up in some wonky state (eg. errors while processing head change). It also might be useful for callers of the index to be able to "tell" the indexer that it's inconsistent, and trigger a reconciliation (eg. if for some reason we didn't actually find the message in the returned tipset)

Yeah, that's a good point. The problem is the synchronous nature of it, hence startup seemed like the logical place to do it. We should definitely consider expanding the adaptive capabilities in follow up, but I am not convinced it is necessary.

> On the flip-side, as written with the current user of this index, we only need the Tipset CID in the MsgInfo struct for reconciliation, a process we don't expect to need -- we don't actually need it for the lookback, because we load the "execution" tipset and pull the info out of it anyway. This means that we could drop the size of this db by ~half (one of the 2 CIDs), and have this just me a (msgCid, height) mapping. In essence we can get a much smaller database footprint (50%) at the cost of much slower / worse reconciliation (either manually inspecting every record, or just dropping any records that callers report as "bad").

I don't think we should do that, having the tipset cid in the index opens up great opportunities. See the notes above about ChainStore ownership, but at the very least we can return the execution tipset this way in addition to the inclsion tipset on queries if we have it. I left a TODO for an optimization follow up in the search implementation.

> I have pretty high confidence in the correctness of this thanks to how simple it is. Even so, could we consider adding a "paranoid" mode, that uses the cache, then does the lookback anyway, and screams if results don't match? This can be run by us, as well as any integration partners with high security standards for this portion of the codebase (we have some such partners).

That's probably a good idea, and it will definitely helps us build confidence harder.

> I think it'll be the case that if this is slow, it'll slow down overall sync. Is that so, and if so can we maybe add some timing metrics?

I don't think it will be slow (sqlite is an amazing piece of software), but you are absolutely right that we should have some timing measurements. I will add some logging to that regard.

> I'm strongly in favour of having a single table with the minor duplication here +1

yeah, it makes things so much simpler! And it is not that big of a waste, that 2x is not that bad.

chain/stmgr/searchwait.go
```diff
@@ -159,6 +173,41 @@ func (sm *StateManager) SearchForMessage(ctx context.Context, head *types.TipSet
 	return fts, r, foundMsg, nil
 }
 
+func (sm *StateManager) searchForIndexedMsg(ctx context.Context, mcid cid.Cid, m types.ChainMsg) (*types.TipSet, *types.MessageReceipt, cid.Cid, error) {
+	minfo, err := sm.msgIndex.GetMsgInfo(ctx, mcid)
+	if err != nil {
+		return nil, nil, cid.Undef, err
```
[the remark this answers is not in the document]
vyzo: ack, will do.

chain/index/msgindex.go
```diff
@@ -0,0 +1,458 @@
...
+
+	case errors.Is(err, fs.ErrNotExist):
+
+	case err != nil:
+		return nil, xerrors.Errorf("error stating msgindex database: %w", err)
+	}
+
+	db, err := sql.Open("sqlite3", dbPath)
+	if err != nil {
+		// TODO [nice to have]: automaticaly delete corrupt databases
```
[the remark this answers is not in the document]
vyzo: yes, of course. 

Part of making the TODO was to discuss them, see what we need to do in the initial MVP, and the rest can go into a follow up issue.

chain/index/msgindex.go
```diff
@@ -0,0 +1,458 @@
...
+	defer x.closeLk.RUnlock()
+
+	if x.closed {
+		return nil
+	}
+
+	// do it in the background to avoid blocking head change processing
+	x.mx.Lock()
+	x.pend = append(x.pend, headChange{rev: rev, app: app})
+	// TODO log loudly if this is building backlog (it shouldn't but better be safe on this)
```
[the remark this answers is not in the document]
vyzo: ok.

chain/index/msgindex.go
```diff
@@ -0,0 +1,458 @@
...
+
+func (x *msgIndex) background(ctx context.Context) {
+	defer x.workers.Done()
+
+	for {
+		select {
+		case <-x.sema:
+			err := x.processHeadChanges(ctx)
+			if err != nil {
+				// TODO should we shut down the index altogether? we just log for now.
```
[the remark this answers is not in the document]
vyzo: ok, it is a reasonable thing to do.

chain/index/msgindex.go
```diff
@@ -0,0 +1,458 @@
...
+
+	tx, err := x.db.Begin()
+	if err != nil {
+		return xerrors.Errorf("error creating transaction: %w", err)
+	}
+
+	for _, hc := range pend {
+		for _, ts := range hc.rev {
+			if err := x.doRevert(ctx, tx, ts); err != nil {
+				if err := tx.Rollback(); err != nil {
```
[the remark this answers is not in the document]
vyzo: ha, ok.

chain/index/msgindex.go
```diff
@@ -0,0 +1,458 @@
...
+}
+
+func reconcileIndex(db *sql.DB, cs ChainStore) error {
+	// Invariant: after reconciliation, every tipset in the index is in the current chain; ie either
+	//  the chain head or reachable by walking the chain.
+	// Algorithm:
+	//  1. Count messages in index; if none, trivially reconciled.
+	//     TODO we may consider populating the index in that case
+	//  2. Find the minimum tipset in the index; this will mark the end of the reconciliation walk
+	//  3. Walk from current tipset until we find a tipset in the index.
```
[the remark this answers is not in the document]
vyzo: Do the backfill? Probably not that much slower, but it is something we can measure and implement in the follow up.

I went for KISS here, that's the only reason i didn't backfill while reconciling.

chain/stmgr/searchwait.go
vyzo: it does read funny, indeed; will fix.

chain/stmgr/searchwait.go
vyzo: let's TODO it for the follow up.

chain/stmgr/searchwait.go
vyzo: ok

chain/index/msgindex.go
vyzo: Yeah, but I am not convinced it is a problem -- let's wait for @arajasek to chime in here.

chain/index/msgindex.go
vyzo: ha! I can rename if you insist, you are technically right.

chain/index/msgindex.go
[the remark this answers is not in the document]
vyzo: ok, the shed tool might manual insert stuff over existing thinges, so we probably want an ON CONFLICT REPLACE in the table definition.

cmd/lotus-shed/msgindex.go
vyzo: https://github.com/filecoin-project/lotus/pull/10452/commits/1676d5148436136c74afc9dcc050106750e5aa42

chain/index/msgindex.go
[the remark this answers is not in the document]
vyzo: https://github.com/filecoin-project/lotus/pull/10452/commits/1676d5148436136c74afc9dcc050106750e5aa42

chain/stmgr/searchwait.go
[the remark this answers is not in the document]
vyzo: done.

chain/index/interface.go
```diff
@@ -0,0 +1,45 @@
...
+	TipSet cid.Cid
+	// the epoch whre this message was included
```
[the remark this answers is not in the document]
vyzo: done.

chain/index/msgindex.go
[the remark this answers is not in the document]
vyzo: done.

chain/stmgr/searchwait.go
vyzo: added TODO.

chain/stmgr/searchwait.go
vyzo: added check in call site.

vyzo:
Updates:
- lotus-shed now expands the repo dir to avoid the tilde footgun
- added an ON CONFLICT clause in the table for the message cid, to avoid conflicts from stuff inserted manually with shed
- addressed @arajasek review comments.

chain/index/msgindex.go
arajasek: Yeah, I don't think this is a problem -- the chainstore shouldn't have "started", and even if it had, I don't think I see a problem if this races?

I do agree with @Stebalien that reconciling on first event would be nicer, though, but it's non-blocking IMO.

arajasek (approved):

vyzo:
yessir, i shall do.

jennijuju:
merge block: can we have some user documentation is the PR description? i.e: what kind of the configuration exists? any tooling to manage the db?

vyzo:
> merge block: can we have some user documentation is the PR description? i.e: what kind of the configuration exists? any tooling to manage the db? 

@jennijuju done, updated the pr description.

vyzo:
@arajasek follow up issue: https://github.com/filecoin-project/lotus/issues/10536

PR #10457, opened 2023-03-12
feat: discard state writes in StateCompute and StateReplay.

raulk:
## Related Issues

This originally started because Glif and other operators begun having blockstore ballooning issues starting with v1.20.3. The issue was the usage of `StateCompute` in the Eth API. That method finishes by flushing the new state to the blockstore. Badger is an LSM + value log store, and it happily accepts every write.

image

## Proposed Changes

The fix is to add a flag to skip writing the state tree when executing a tipset. By default we write, except when the caller requests us not to. Instead of polluting APIs with yet another option, I decided to add an option struct (`ExecutorOpts`). Unfortunately, `TipSetState` is used a many places, so to avoid breakage I introduced a `TipSetWithOpts` method. We _could_ also use the functional options pattern, but those are generally used in more public/user-facing APIs.

This change is no longer necessary for the Eth API, since we no longer need to compute state after https://github.com/filecoin-project/lotus/pull/10446. However, they will be very beneficial to users of `StateCompute` (e.g. exchanges) and `StateReplay`.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/execute.go
+++ b/chain/stmgr/execute.go
@@ -13,6 +13,15 @@ import (
 )
 
 func (sm *StateManager) TipSetState(ctx context.Context, ts *types.TipSet) (st cid.Cid, rec cid.Cid, err error) {
+	return sm.TipSetStateWithOpts(ctx, ts, &TipSetStateOpts{})
+}
+
+type TipSetStateOpts struct {
+	// DiscardState: see ExecutorOpts.DiscardState.
+	DiscardState bool
+}
+
+func (sm *StateManager) TipSetStateWithOpts(ctx context.Context, ts *types.TipSet, opts *TipSetStateOpts) (st cid.Cid, rec cid.Cid, err error) {
 	ctx, span := trace.StartSpan(ctx, "tipSetState")
 	defer span.End()
 	if span.IsRecordingEvents() {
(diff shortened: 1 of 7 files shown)


chain/stmgr/execute.go
```diff
@@ -13,6 +13,15 @@ import (
 )
 
 func (sm *StateManager) TipSetState(ctx context.Context, ts *types.TipSet) (st cid.Cid, rec cid.Cid, err error) {
+	return sm.TipSetStateWithOpts(ctx, ts, &TipSetStateOpts{})
+}
+
+type TipSetStateOpts struct {
```
arajasek, line 19: I'm not convinced this is necessary? With the existing tipsetstate cache, and the newly added fast lookup in https://github.com/filecoin-project/lotus/pull/10445, I think we can drop this. We'll only wind up executing the tipset (making the write) if:

- we're executing a tipset that isn't in our main chain
- we're missing the actual state / receipts root.

arajasek (commented):
Broadly looks good to me, and a nice improvement. One note.

Stebalien:
> However, they will be very beneficial to users of StateCompute (e.g. exchanges) and StateReplay.

I'm not so sure. Those users likely _need_ the state.

Is the issue here that we don't check if the blockstore has it before writing?

raulk:
Superseded by #10680, which is simplier and less invasive.

eshon:
FYI - Glif Nodes just reported that archival nodes now seem to be growing at about 19 GiB / day, down from about 50 GiB / day before. (An FEVM archival node is ~3.6 TiB currently.)

PR #10461, opened 2023-03-13
fix: itest: avoid failing the test when we race the miner

Stebalien:
I've been seeing quite a few tests failing here, so I'm hoping this will deflake them a bit.


Stebalien:
Hypothesis: When we ended the test, the miner was still running and attempted to create one more block. This failed because the test was done and the lotus node shut down.

Ideally all those "rpc error" checks should catch this, but apparently those aren't sufficient. Checking the context is a better catch-all solution.

arajasek (approved):

PR #10463, opened 2023-03-13
refactor: update cache to the new generic version

Stebalien:
- Adds type safety.
- Reduces allocations.
- ~Fixes the drand cache (was storing by value, but retrieving by pointer)~ (steven can't read)

### DIFF
--- a/chain/beacon/drand/drand.go
+++ b/chain/beacon/drand/drand.go
@@ -12,7 +12,7 @@ import (
 	gclient "github.com/drand/drand/lp2p/client"
 	"github.com/drand/kyber"
 	kzap "github.com/go-kit/kit/log/zap"
-	lru "github.com/hashicorp/golang-lru"
+	lru "github.com/hashicorp/golang-lru/v2"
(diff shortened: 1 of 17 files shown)


chain/beacon/drand/drand.go
```diff
@@ -160,16 +160,12 @@ func (db *DrandBeacon) Entry(ctx context.Context, round uint64) <-chan beacon.Re
 	return out
 }
 func (db *DrandBeacon) cacheValue(e types.BeaconEntry) {
-	db.localCache.Add(e.Round, e)
+	db.localCache.Add(e.Round, &e)
```
Stebalien, line 163: the bug

Stebalien:
(why now? I want to use an LRU cache in the Eth API, but want to be able to use the generic version)

chain/beacon/drand/drand.go
Stebalien: Ok, not a bug. Steven's just stupid.

arajasek (approved):
LGTM.

Issue #10484, opened 2023-03-15
Contract Deal Enabled Boost is Unable to Sart Due To EthRPC Subscription Lock


TippyFlitsUK:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [X] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon:  1.21.0-dev+mainnet+git.58900a703+api1.5.0
Local: lotus-miner version 1.21.0-dev+mainnet+git.58900a703
boostd version 1.6.0+git.08d959a

Also reported by users running lotus v.1.20.0
```

### Repro Steps

1. Run '...'
2. Do '...'
3. See error '...'
...
1. Start Boost with contract deals enabled.
2. Stop Boost
3. Start Boost
4 Boost freezes at `contract deal monitor starting`

### Describe the Bug

A number of SPs who are actively participating in the FVM deal contract testing experienced a failure to fully restart a contract deals enabled Boost.

The initial start of Boost with contract deals enabled did not present any issues however a simple Boost restart with contract deals enabled caused Boost to freeze at `contract deal monitor starting`. Basic dealmaking and Web UI were both broken.

A simple workaround for testing SPs is:
1. Stop boostd
2. Restart daemon and miner
3. Start Boost again

@magik6k also suspects that this may be related to a separate EthRPC discussion here - https://filecoinproject.slack.com/archives/CPFTWMY7N/p1678880057771859

[lex.goroutines.txt](https://github.com/filecoin-project/lotus/files/10980124/lex.goroutines.txt)

### Logging Information

```text
Goroutine logs attached above.
```

PR #10485, opened 2023-03-15
fix: ethrpc: Don't lock up when eth subscriber goes away

magik6k:
## Related Issues
Fixes https://github.com/filecoin-project/lotus/issues/10484
Brings in https://github.com/filecoin-project/go-jsonrpc/pull/98

## Proposed Changes
* Bring a fix for jsonrpc reverse client hang on closed connections
* Buffer eth_subscribe messages if the subscriber is slow
* Always uninstall sub-related filters when the sub dies, not only when it's explicitly cancelled

## Checklist

- [ ] CI is green

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -304,6 +304,7 @@ require (
 	github.com/rivo/uniseg v0.1.0 // indirect
 	github.com/rs/cors v1.7.0 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
+	github.com/segmentio/fasthash v1.0.3 // indirect
 	github.com/shirou/gopsutil v2.18.12+incompatible // indirect
 	github.com/sirupsen/logrus v1.9.0 // indirect
 	github.com/spaolacci/murmur3 v1.1.0 // indirect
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -319,6 +320,7 @@ require (
 	github.com/xrash/smetrics v0.0.0-20201216005158-039620a65673 // indirect
 	github.com/zondax/hid v0.9.1 // indirect
 	github.com/zondax/ledger-go v0.12.1 // indirect
+	github.com/zyedidia/generic v1.2.1 // indirect
```
magik6k, line 323: This is a very small library with a bunch of very nice generic helpers. Wanted to use it since forever, finally got a quality excuse.

node/impl/full/eth.go
```diff
@@ -1743,11 +1792,23 @@ func (e *ethSubscription) start(ctx context.Context) {
...
+	}
 
 	if e.quit != nil {
 		e.quit()
 		e.quit = nil
+		e.mu.Unlock()
+
+		for _, f := range e.filters {
+			// note: the context in actually unused in uninstallFilter
+			if err := e.uninstallFilter(context.TODO(), f); err != nil {
```
ZenGround0, line 1807: double checking: these don't take long and won't hang node shutdown right?

ZenGround0 (approved):

node/impl/full/eth.go
magik6k: Correct

TippyFlitsUK:
Looking good Magik!! I've tested across three stack restarts and 8 Boost restarts and have been able to start every time!!

PR #10488, opened 2023-03-15
feat:config:force existing users to opt into new defaults

ZenGround0:
## Related Issues

This is the change to 1) force users to explicitly set splitstore config by uncommenting the default, 2) for new nodes do the uncommenting automatically so they will opt into the default automatically 

## Proposed Changes

- Add a notion of uncommenting filters to config update.  Then apply a filter on EnableSplitstore in the default full node config
- Add a notion of loadability checks to the lotus config.  Then apply a check ensuring `#EnableSplitstore =` is not in the config to ensure that users with old configs will reset their config explicity.

## Additional Info

- The patterns introduced here can hopefully be used in other sensative default change situations
- We can probably revert the splitstore specific filters and loadability checks in a few months when the community is adjusted to the new defaults 

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus/config.go
+++ b/cmd/lotus/config.go
@@ -31,7 +31,7 @@ var configDefaultCmd = &cli.Command{
 	Action: func(cctx *cli.Context) error {
 		c := config.DefaultFullNode()
 
-		cb, err := config.ConfigUpdate(c, nil, !cctx.Bool("no-comment"), config.DefaultFullNodeCommentFilter())
+		cb, err := config.ConfigUpdate(c, nil, !cctx.Bool("no-comment"), config.KeepUncommented(config.MatchEnableSplitstoreField))
 		if err != nil {
 			return err
 		}
@@ -83,7 +83,7 @@ var configUpdateCmd = &cli.Command{
 
 		cfgDef := config.DefaultFullNode()
 
-		updated, err := config.ConfigUpdate(cfgNode, cfgDef, !cctx.Bool("no-comment"), config.DefaultFullNodeCommentFilter())
+		updated, err := config.ConfigUpdate(cfgNode, cfgDef, !cctx.Bool("no-comment"), config.KeepUncommented(config.MatchEnableSplitstoreField))
 		if err != nil {
 			return err
 		}
--- a/cmd/lotus/config.go
+++ b/cmd/lotus/config.go
@@ -31,7 +31,7 @@ var configDefaultCmd = &cli.Command{
 	Action: func(cctx *cli.Context) error {
 		c := config.DefaultFullNode()
 
-		cb, err := config.ConfigUpdate(c, nil, !cctx.Bool("no-comment"), config.KeepUncommented(config.MatchEnableSplitstoreField))
+		cb, err := config.ConfigUpdate(c, nil, config.Commented(!cctx.Bool("no-comment")), config.DefaultKeepUncommented())
 		if err != nil {
 			return err
 		}
@@ -83,7 +83,7 @@ var configUpdateCmd = &cli.Command{
 
 		cfgDef := config.DefaultFullNode()
 
-		updated, err := config.ConfigUpdate(cfgNode, cfgDef, !cctx.Bool("no-comment"), config.KeepUncommented(config.MatchEnableSplitstoreField))
+		updated, err := config.ConfigUpdate(cfgNode, cfgDef, config.Commented(!cctx.Bool("no-comment")), config.DefaultKeepUncommented())
 		if err != nil {
 			return err
 		}
--- a/cmd/lotus/config.go
+++ b/cmd/lotus/config.go
@@ -31,7 +31,7 @@ var configDefaultCmd = &cli.Command{
 	Action: func(cctx *cli.Context) error {
 		c := config.DefaultFullNode()
 
-		cb, err := config.ConfigUpdate(c, nil, !cctx.Bool("no-comment"))
+		cb, err := config.ConfigUpdate(c, nil, !cctx.Bool("no-comment"), config.CommentFilterDefault())
 		if err != nil {
 			return err
 		}
@@ -83,7 +83,7 @@ var configUpdateCmd = &cli.Command{
 
 		cfgDef := config.DefaultFullNode()
 
-		updated, err := config.ConfigUpdate(cfgNode, cfgDef, !cctx.Bool("no-comment"))
+		updated, err := config.ConfigUpdate(cfgNode, cfgDef, !cctx.Bool("no-comment"), config.CommentFilterDefault())
 		if err != nil {
 			return err
 		}
--- a/cmd/lotus/config.go
+++ b/cmd/lotus/config.go
@@ -31,7 +31,7 @@ var configDefaultCmd = &cli.Command{
 	Action: func(cctx *cli.Context) error {
 		c := config.DefaultFullNode()
 
-		cb, err := config.ConfigUpdate(c, nil, !cctx.Bool("no-comment"), config.CommentFilterDefault())
+		cb, err := config.ConfigUpdate(c, nil, !cctx.Bool("no-comment"), config.DefaultFullNodeCommentFilter())
 		if err != nil {
 			return err
 		}
@@ -83,7 +83,7 @@ var configUpdateCmd = &cli.Command{
 
 		cfgDef := config.DefaultFullNode()
 
-		updated, err := config.ConfigUpdate(cfgNode, cfgDef, !cctx.Bool("no-comment"), config.CommentFilterDefault())
+		updated, err := config.ConfigUpdate(cfgNode, cfgDef, !cctx.Bool("no-comment"), config.DefaultFullNodeCommentFilter())
 		if err != nil {
 			return err
 		}
--- a/node/config/load.go
+++ b/node/config/load.go
@@ -46,7 +46,21 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
 	return cfg, nil
 }
 
-func ConfigUpdate(cfgCur, cfgDef interface{}, comment bool) ([]byte, error) {
+type keepUncommented func(string) bool
+
+func CommentFilterDefault() keepUncommented {
+	return func(s string) bool {
+		return strings.Contains(s, "EnableSplitstore")
+	}
+}
+
+func CommentFilterNone() keepUncommented {
+	return func(string) bool {
+		return false
+	}
+}
+
+func ConfigUpdate(cfgCur, cfgDef interface{}, comment bool, filter keepUncommented) ([]byte, error) {
 	var nodeStr, defStr string
 	if cfgDef != nil {
 		buf := new(bytes.Buffer)
@@ -131,7 +145,7 @@ func ConfigUpdate(cfgCur, cfgDef interface{}, comment bool) ([]byte, error) {
 			}
 
 			// if there is the same line in the default config, comment it out it output
-			if _, found := defaults[strings.TrimSpace(nodeLines[i])]; (cfgDef == nil || found) && len(line) > 0 {
+			if _, found := defaults[strings.TrimSpace(nodeLines[i])]; (cfgDef == nil || found) && len(line) > 0 && !filter(line) {
 				line = pad + "#" + line[len(pad):]
 			}
 			outLines = append(outLines, line)
@@ -159,5 +173,5 @@ func ConfigUpdate(cfgCur, cfgDef interface{}, comment bool) ([]byte, error) {
 }
 
 func ConfigComment(t interface{}) ([]byte, error) {
-	return ConfigUpdate(t, nil, true)
+	return ConfigUpdate(t, nil, true, CommentFilterDefault())
 }
--- a/node/config/load.go
+++ b/node/config/load.go
@@ -80,9 +80,9 @@ type loadabilityCheck struct {
 
 func DefaultFullNodeLoadabilityCheck() loadabilityCheck {
 	configOk := func(cfgRaw string) error {
-		poisonStr := "#EnableSplitstore =" // splitstore must be set explicitly
-		if strings.Contains(cfgRaw, poisonStr) {
-			return xerrors.Errorf("Config contains disallowed string %s, refusing to load. Please explicitly set EnableSplitstore. Set it to false to keep chain management unchanged from March 2023 set it to true if you want to discard old state by default", poisonStr)
+		explicitSplitstoreRx := regexp.MustCompile("\n[\t\n\f\r ]*EnableSplitstore =")
+		if match := explicitSplitstoreRx.FindString(cfgRaw); match == "" {
+			return xerrors.Errorf("Config does not contain explicit set of EnableSplitstore field, refusing to load. Please explicitly set EnableSplitstore. Set it to false to keep chain management unchanged from March 2023 set it to true if you want to discard old state by default")
 		}
 		return nil
 	}
--- a/node/config/load.go
+++ b/node/config/load.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"reflect"
 	"regexp"
@@ -17,15 +18,40 @@ import (
 
 // FromFile loads config from a specified file overriding defaults specified in
 // the def parameter. If file does not exist or is empty defaults are assumed.
-func FromFile(path string, def interface{}) (interface{}, error) {
+func FromFile(path string, def interface{}, lc loadabilityCheck) (interface{}, error) {
+	// check for loadability
 	file, err := os.Open(path)
 	switch {
 	case os.IsNotExist(err):
+		if err := lc.configNotFound(); err != nil {
+			return nil, err
+		}
 		return def, nil
 	case err != nil:
 		return nil, err
 	}
+	cfgBs, err := ioutil.ReadAll(file)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to read config for loadability checks %w", err)
+	}
+	s := string(cfgBs)
+	if err := lc.configOk(s); err != nil {
+		return nil, xerrors.Errorf("config failed loadability check: %w", err)
+	}
+	if err := file.Close(); err != nil {
+		return nil, err
+	}
 
+	file, err = os.Open(path)
+	switch {
+	case os.IsNotExist(err):
+		if err := lc.configNotFound(); err != nil {
+			return nil, err
+		}
+		return def, nil
+	case err != nil:
+		return nil, err
+	}
 	defer file.Close() //nolint:errcheck // The file is RO
 	return FromReader(file, def)
 }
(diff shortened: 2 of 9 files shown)


node/config/load.go
```diff
@@ -17,15 +18,40 @@ import (
 
 // FromFile loads config from a specified file overriding defaults specified in
 // the def parameter. If file does not exist or is empty defaults are assumed.
-func FromFile(path string, def interface{}) (interface{}, error) {
+func FromFile(path string, def interface{}, lc loadabilityCheck) (interface{}, error) {
```
Stebalien, line 21: This isn't a "loadability" check, it's a config loading config. I'd:

1. Use the functional options pattern, or maybe a struct. Definitely not a private interface.
2. Call these config loading options, or something like that.

node/config/load.go
```diff
@@ -17,15 +18,40 @@ import (
...
+	// check for loadability
 	file, err := os.Open(path)
 	switch {
 	case os.IsNotExist(err):
+		if err := lc.configNotFound(); err != nil {
+			return nil, err
+		}
 		return def, nil
```
Stebalien, line 29: There shouldn't be a separate "default". Instead, just have one `default() (interface{}, error)` function.

Stebalien: (possibly attached to the config loading options)

node/config/load.go
```diff
@@ -17,15 +18,40 @@ import (
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to read config for loadability checks %w", err)
+	}
+	s := string(cfgBs)
+	if err := lc.configOk(s); err != nil {
+		return nil, xerrors.Errorf("config failed loadability check: %w", err)
+	}
+	if err := file.Close(); err != nil {
+		return nil, err
+	}
```
Stebalien, line 43: Please close the file first.

Stebalien: Or defer closing it.

Stebalien: At the moment, we leak it on error.

node/config/load.go
```diff
@@ -17,15 +18,40 @@ import (
...
 
+	file, err = os.Open(path)
```
Stebalien, line 45: Why are we doing this twice?

node/config/load.go
```diff
@@ -46,7 +72,51 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
...
+
+func DefaultFullNodeLoadabilityCheck() loadabilityCheck {
+	configOk := func(cfgRaw string) error {
+		poisonStr := "#EnableSplitstore =" // splitstore must be set explicitly
+		if strings.Contains(cfgRaw, poisonStr) {
+			return xerrors.Errorf("Config contains disallowed string %s, refusing to load. Please explicitly set EnableSplitstore. Set it to false to keep chain management unchanged from March 2023 set it to true if you want to discard old state by default", poisonStr)
+		}
+		return nil
+	}
+	configNotFound := func() error {
```
Stebalien, line 89: Having a specific comment about the splitstore here isn't going to age well.

Stebalien: And shouldn't the default config handle this for us?

Stebalien: If the user reads this error message, they're going to come to slack and say "wtf, help!".

node/config/load.go
```diff
@@ -46,7 +72,51 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
...
+		configOk:       func(string) error { return nil },
+		configNotFound: func() error { return nil },
+	}
+}
+
+type keepUncommented func(string) bool
+
+func DefaultFullNodeCommentFilter() keepUncommented {
+	return func(s string) bool {
+		return strings.Contains(s, "EnableSplitstore")
```
Stebalien, line 109: This is really sloppy. I assume you're saying "keep the enable splitstore line uncommented"?

Ideally, we'd actually follow the structure of the config. At a _minimum_, we should be checking that this string appears at the beginning of the line (i.e., `^EnableSplitstore =`).

But there also needs to be a lot of comments/documentation explaining why we're doing something like this. It's really non-obvious.

node/config/load.go
```diff
@@ -46,7 +72,51 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
...
+	}
+}
+
+func CommentFilterNone() keepUncommented {
+	return func(string) bool {
+		return false
+	}
+}
+
+func ConfigUpdate(cfgCur, cfgDef interface{}, comment bool, filter keepUncommented) ([]byte, error) {
```
Stebalien, line 119: Docs. Also, please use something like the functional options pattern.

Stebalien (changes requested):

node/config/load.go
```diff
@@ -46,7 +72,51 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
...
+type loadabilityCheck struct {
+	configOk       func(string) error
+	configNotFound func() error
+}
+
+func DefaultFullNodeLoadabilityCheck() loadabilityCheck {
+	configOk := func(cfgRaw string) error {
+		poisonStr := "#EnableSplitstore =" // splitstore must be set explicitly
+		if strings.Contains(cfgRaw, poisonStr) {
+			return xerrors.Errorf("Config contains disallowed string %s, refusing to load. Please explicitly set EnableSplitstore. Set it to false to keep chain management unchanged from March 2023 set it to true if you want to discard old state by default", poisonStr)
```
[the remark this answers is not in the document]
ZenGround0: > What if I just want the default?
> won't this cause every single node to fail to start on upgrade?

The point of this change (idea originating [here](https://filecoinproject.slack.com/archives/CP50PPW2X/p1678803533926769?thread_ts=1678803343.792909&cid=CP50PPW2X) is:
1) If you want the default and you have an existing, unset config: too bad.  You have to set it explicitly or your node won't start. This way users won't accidentally overwrite data.
2) If you already set your existing config explicitly then your config values won't be changing so that's ok and your node starts.
3) If you start a new node then the default will be set explicitly for you as part of this PR's change to the default config and your node will start.

> What if I comment out one EnableSplitstore line but keep another?

This is a good point I hadn't thought of.  What I would like to be doing instead is parsing the TOML and making sure that EnableSplitstore field is set there. I'll go back and take a look.

node/config/load.go
ZenGround0: Mostly just me not knowing how computers work, I'll figure out how to reuse to make this nicer.

node/config/load.go
ZenGround0: > If the user reads this error message, they're going to come to slack and say "wtf, help!".

I kind of wrote this expecting that to be fine.  I thought of this as going out of our way to really make sure people don't lose data with unexpected discard mode even if they *delete their config file* which I didn't think was common at all.

That said I can add in instructions to regenerate a default config file and warn that splitstore will be set in that config.

node/config/load.go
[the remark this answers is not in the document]
Stebalien: > The point of this change (idea originating [here](https://filecoinproject.slack.com/archives/CP50PPW2X/p1678803533926769?thread_ts=1678803343.792909&cid=CP50PPW2X) is:

Can we just warn loudly and let the user fix their config if they want? If they don't, isn't that their problem? It's not like non-splitstore configs aren't supported...

node/config/load.go
ZenGround0: Its a little sloppy and I agree `^EnableSplitstore =` is better but this *does* follow the structure of the config.  When we traverse ConfigUpdate this gets applied to config field lines, not sections, not whitespace, not comments.

node/config/load.go
ZenGround0: I'll take a look and do it if its a small change but I'd like to limit the scope of the PR to avoid improving the existing lotus config code.

node/config/load.go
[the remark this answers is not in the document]
ZenGround0: 🤷 this honestly isn't something I thought of or particularly care about as a node operator myself.  Other node operators [appear to a lot](https://filecoinproject.slack.com/archives/CP50PPW2X/p1678916423822579?thread_ts=1678916327.487719&cid=CP50PPW2X) and data loss is scary so it makes some sense.  @jennijuju as the progenitor of this change could you weigh in?

node/config/load.go
Stebalien: We definitely don't want to be editing/changing defaults. But yeah, a quick "run this command to get unstuck" would be good.

node/config/load.go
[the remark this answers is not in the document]
ZenGround0: I guess going back to the original message it looks like @jennijuju is listening to the advice of glif on this one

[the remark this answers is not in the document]
ZenGround0: Supporting a safety measure they want seems reasonable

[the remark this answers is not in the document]
ribasushi: Basically if one runs the new default on a fullnode - it will be effectively destroyed within a ~day.

Because of how expensive catch up is - restoring from backup is hardly an option.

@ZenGround0 perhaps the right direction instead of the config dance is to just check how big the badgerstore is and refuse to fuck with it over a certain size?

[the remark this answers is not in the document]
ZenGround0: Im pretty sure what actually happens is that the universal store will just stay static with all the old state ( the same way the ~130 Gigs of snapshot stay around forever) but the new state will begin to be trimmed. That's still bad but its a much easier problem to recover from. I think I'll test this and if communicate it which will hopefully lead to some more chill.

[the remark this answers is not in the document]
ZenGround0: Keeping this thread alive but the original concern is addressed in the latest commit.  We now use regex to ensure that an uncommented `EnableSplitstore =` preceded only by whitespace exists in the config which should be the correct check.

node/config/load.go
ZenGround0: ok I've added beginning of line and whitespace aware regexp imo this is good enough

node/config/load.go
ZenGround0: Ok added `lotus config default` instruction.  Reiterating that we don't want the default config handling this edge case for us while transition from splitstore and that its ok to reference splitstore specifically in the error.  Its not going to age well but were also going to remove this error and that message in a few months when everyone has updated their node and dealt with this.

node/config/load.go
ZenGround0: Added a small amount of docs and one functional option related to what I'm doing.

node/config/load.go
```diff
@@ -17,17 +18,40 @@ import (
...
+	var loadOpts cfgLoadOpts
+	for _, opt := range opts {
+		if err := opt(&loadOpts); err != nil {
+			return nil, xerrors.Errorf("failed to apply load cfg option: %w", err)
+		}
+	}
+
+	// check for loadability
 	file, err := os.Open(path)
+	defer file.Close() //nolint:errcheck // The file is RO
```
Stebalien: No, we need to check the error first. Or this can panic.

node/config/load.go
```diff
@@ -17,17 +18,40 @@ import (
...
+	tee := io.TeeReader(file, &buf)
+	cfgBs, err := ioutil.ReadAll(tee)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to read config for loadability checks %w", err)
+	}
+	s := string(cfgBs)
+	if err := loadOpts.validate(s); err != nil {
+		return nil, xerrors.Errorf("config failed loadability check: %w", err)
+	}
+	return FromReader(&buf, def)
```
Stebalien, line 54: Checkout `buf.String`, you don't need the tee reader.

node/config/load.go
```diff
@@ -17,17 +18,40 @@ import (
...
 	file, err := os.Open(path)
+	defer file.Close() //nolint:errcheck // The file is RO
 	switch {
 	case os.IsNotExist(err):
+		if loadOpts.canFallbackOnDefault != nil {
+			if err := loadOpts.canFallbackOnDefault(); err != nil {
+				return nil, err
+			}
+		}
 		return def, nil
```
Stebalien, line 39: Ah, sorry, I wasn't clear. We have two options to control "defaults", we should have a single option. E.g., something like:

```
func DefaultConfig(func() (interface{}, error))
```

This code here would then read:

```go
if loadOps.defaultConfig != nil {
    return loadOps.defaultConfig()
}
return nil, xerrors.Errorf("no config found")
```

node/config/load.go
```diff
@@ -46,7 +70,93 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
 	return cfg, nil
 }
 
-func ConfigUpdate(cfgCur, cfgDef interface{}, comment bool) ([]byte, error) {
+// function from raw config string to loadability
+type loadabilityCheck struct {
```
Stebalien: (can we kill the term "loadability check")

Stebalien (changes requested):

cmd/lotus/config.go
```diff
@@ -31,7 +31,7 @@ var configDefaultCmd = &cli.Command{
 	Action: func(cctx *cli.Context) error {
 		c := config.DefaultFullNode()
 
-		cb, err := config.ConfigUpdate(c, nil, !cctx.Bool("no-comment"))
+		cb, err := config.ConfigUpdate(c, nil, !cctx.Bool("no-comment"), config.KeepUncommented(config.MatchEnableSplitstoreField))
```
Stebalien, line 34: Let's provide a default. We shouldn't have things mentioning the splitstore all over lotus.

node/config/load.go
```diff
@@ -46,7 +70,70 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
...
+
+// Match the EnableSplitstore field
+func MatchEnableSplitstoreField(s string) bool {
+	enableSplitstoreRx := regexp.MustCompile("^[\t\n\f\r ]*EnableSplitstore =")
+	match := enableSplitstoreRx.FindString(s)
+	return match != ""
+}
+
+// ConfigUpdate takes in a config and a default config and optionally comments out default values
+func ConfigUpdate(cfgCur, cfgDef interface{}, comment bool, opts ...UpdateCfgOpt) ([]byte, error) {
```
Stebalien, line 130: If we're going with functional options, I'd use them everywhere (e.g., `comment`).

Also, I apologize. I asked you to use functional options because the whole "private config interface" thing was not great. But a simple struct would likely work as well if you think that would be better.

node/config/load.go
```diff
@@ -46,7 +70,70 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
...
+		return nil
+	}
+}
+
+func NoDefaultForSplitstoreTransition() error {
+	return xerrors.Errorf("FullNode config not found and fallback to default disallowed while we transition to splitstore discard default.  Use `lotus config default` to set this repo up with a default config.  Be sure to set `EnableSplitstore` to `false` if you are running a full archive node")
+}
+
+func ValidateSplitstoreSet(cfgRaw string) error {
+	explicitSplitstoreRx := regexp.MustCompile("\n[\t\n\f\r ]*EnableSplitstore =")
```
Stebalien, line 99: `"(?m)^\s*EnableSplitstore\s*="`

- multiline (`^` matches beginning of line)
- match any number of non-line whitespace
- the string
- non-line whitespace
- equals

node/config/load.go
```diff
@@ -46,7 +70,70 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
...
+	if match := explicitSplitstoreRx.FindString(cfgRaw); match == "" {
```
Stebalien, line 100:
```suggestion
	if !explicitSplitstoreRx.MatchString(cfgRaw) {
```

Stebalien: Same below.

node/config/load.go
Stebalien: Basically, we're doing two things here:

1. We're validating the config.
2. We're providing a default config, if applicable.

I'm assuming that "loadability check" is "can we load this config", but it's just really confusing.

node/config/load.go
ZenGround0: That makes sense and I tried to do it.  The problem is we have two different uses of the default.  One to fallback on if the config file is not found and one for us to use to set the defaults when reading the existing config from disk.  This change wants two different configuration options: a default config to pass in to the function populating the config and a way to error if there is no config specified on disk

ZenGround0: I'll add the default value as a separate functional option though

node/config/load.go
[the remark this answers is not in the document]
ZenGround0: @ribasushi I've confirmed that `datastore/chain` will stay untouched with default discard splitstore so all of the old historical data will stick around even if we don't get config dance right

node/config/load.go
```diff
@@ -46,7 +79,89 @@ func FromReader(reader io.Reader, def interface{}) (interface{}, error) {
...
+
+func DefaultKeepUncommented() UpdateCfgOpt {
+	return KeepUncommented(MatchEnableSplitstoreField)
+}
+
+// Match the EnableSplitstore field
+func MatchEnableSplitstoreField(s string) bool {
+	enableSplitstoreRx := regexp.MustCompile(`(?m)^\s*EnableSplitstore\s*=`)
+	return enableSplitstoreRx.MatchString(s)
+}
```
magik6k, line 155: Nit: would move next to the other Splitstore functions, so this hack is contained in one code block

magik6k (commented):
Definitely needs some testing

I don't love that we're forcing users to take action on this, tho on the other hand it seems like without splitstore most non-fullnodes require regular maintenance anyways, and splitstore nodes will have this explicitly set already.

There will be some disruption to user flow, but I don't think it will be too bad - and it can likely be mitigated by clear comms in changelog / announcements.

node/config/load.go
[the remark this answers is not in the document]
Stebalien: I'm suggesting the other way around. Warn loudly if the user hasn't configured anything, but leave the default the way it is for existing nodes. But I guess we'll need some form of hack either way.

node/config/load_test.go
```diff
@@ -58,9 +60,82 @@ func TestParitalConfig(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.True(t, MatchEnableSplitstoreField(string(cfgStrFromDef)))
+}
+
+// Loading a config with commented EnableSplitstore fails when setting validator
+func TestValidateConfigSetsEnableSplitstore(t *testing.T) {
+	cfgCommentedOutEnableSS, err := ConfigUpdate(DefaultFullNode(), DefaultFullNode(), Commented(true))
+	assert.NoError(t, err)
+	// this comments out EnableSplitstore
+	assert.False(t, MatchEnableSplitstoreField(string(cfgCommentedOutEnableSS)))
```
magik6k, line 115: This comment seems misplaced

magik6k (approved):
Looks good, feel free to merge whenever.

Ideally we'd give this a try on a real node too.

ZenGround0:
Thanks

> Ideally we'd give this a try on a real node too.

My mainnet node just crashed so this is a good opportunity.

PR #10489, opened 2023-03-16
shed: get balances of evm accounts 

jennijuju:
`./lotus-shed evm-balance` gets total balances of existing eth account actors, evm actors and placeholder actors.

thanks @ZenGround0, for figuring out the datastore reads!


arajasek (approved):

arajasek (approved):
Gets the job done!

Issue #9774, opened 2022-12-02
NV18: Investigate contract logging configuration


arajasek:
We need to make sure that it's easy for users to opt out of storing any contract logs locally, and that users who _do_ store such logs can configure how much is stored.

arajasek:
Note: there may be nothing to do here beyond checking the current system.

jennijuju:
we also need to think about the expected splitstore behaviour upon node configuration.

Things to discover potentially for node operators
-  by default do not store logs, export LOTUS_STORE_EVENTS to opt in?  
    - one can still get events by `StateCompute` or `StateReplay` the message
- for node operators who want to store logs, should we add filter support and discard the non-subscribed events? what kind of the filter support should be exposed if so, i.e:  all contracts under the same `f0410` fam? one single contract? sender? 
- how should splitstore handles GC when the node operator ops in to store all logs? expose a splitstore config val `StoreLogs = false`, when its false GC logs in non-discard(`universal` and `message`) mode; when its true save logs in coldstore with the rest of the state in `universal` and `message` mode?
- if we support filters & subscription mentioned in point II - should splitstore (hot and cold) GC handles subscribed logs and unsubscribed logs differently?
   

I personally feel like for v1.12.0 with the time we have, we may consider 
- by default do not store logs, export LOTUS_STORE_ALL_EVENTS for node operators to opt-in. 
- adding filter support upon node operators feature request.

jennijuju:
we can also consider for nv18 release, we just dont save events/logs in coldstore, 
and based on node operators needs, we can add more advanced node configuration later on.

magik6k:
Config for event index: https://github.com/filecoin-project/lotus/blob/6da93cf921af29e1c6a5e5b701170ef83a3e93e7/documentation/en/default-lotus-config.toml#L267-L314

jennijuju:
@magik6k how are events and logs handled with splitstore?

magik6k:
They should be with receipts

raulk:
I have a concrete proposal which I'll try to write up today.

jennijuju:
> I have a concrete proposal which I'll try to write up today.

@raulk do you still remember what we are talking about here?

jennijuju:
> Document the existing configuration

@Reiers can you take this over? I'd add a session under https://lotus.filecoin.io/lotus/configure/defaults/ as `Eth Service`

jennijuju:
@Reiers update plz?

[...]

PR #10494, opened 2023-03-16
feat: events: make event saving configurable independent of eth api

Stebalien:
## Related Issues

Related to https://github.com/filecoin-project/lotus/issues/9774

## Proposed Changes

We can now instruct lotus to store events, even if the Eth JSON-RPC API is disabled.

## Additional Info

Bikeshed on name?

## Checklist

- [ ] CI is green

### DIFF
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -693,7 +693,10 @@ type FevmConfig struct {
 }
 
 type Events struct {
-	// EnableEthRPC enables APIs that
+	// Save specifies whether or not to save and index emitted events, or discard them. Events
+	// are discarded by default unless the Eth RPC is also enabled by setting EnableEthRPC.
+	Save bool
+
 	// DisableRealTimeFilterAPI will disable the RealTimeFilterAPI that can create and query filters for actor events as they are emitted.
 	// The API is enabled when EnableEthRPC is true, but can be disabled selectively with this flag.
 	DisableRealTimeFilterAPI bool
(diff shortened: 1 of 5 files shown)


Stebalien:
(I'll run `make gen` and add a test once we settle on the name and whether or not we actually want this)

node/config/types.go
```diff
@@ -691,6 +691,10 @@ type FevmConfig struct {
 }
 
 type Events struct {
+	// Save specifies whether or not to save emitted events, or discard them. Events are
+	// discarded by default.
+	Save bool
```
magik6k, line 696: Name sounds good

We should make the comment say that it is enabled implicitly when EnableEthRPC is set - or we can make the node when save isn't set when EnableEthRPC is (can be done easily with the `Error` node builder option)

magik6k:
> I'll run make gen

Note that you can also copy the output from the gen-check CI diff into `git apply`

arajasek:
@Stebalien I have no problem with this, but would lean towards not merging this unless you can point me to someone that wants it? Lotus has a lot of knobs already, and FEVM added yet more (all necessary). I'm a little disinclined to add another finer-tuned one.

Stebalien:
IIRC, @hsanjuan and friends wanted this for archival. Although I think they ended up just enabling the API. So yeah, I don't think it's critical.

Stebalien:
@ribasushi want to submit a rebased PR with docs and tests?

PR #10514, opened 2023-03-20
fix: explicit error in StateSearchMsg when search stops due to lookback.

raulk:
## Related Issues

Currently, `StateSearchMsg` will behave identically (externally) when the message is not found at all and when the search is is halted due to the lookback limit. This prevents users from reasoning about why a method that returned a value correctly at time X, is no longer capable of returning the value at X+limit+1.

User report on Slack: https://filecoinproject.slack.com/archives/CRK2LKYHW/p1679225798632029

## Proposed Changes

Return an explicit error when the above circumstance occurs.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/searchwait.go
+++ b/chain/stmgr/searchwait.go
@@ -185,9 +185,15 @@ func (sm *StateManager) searchBackForMsg(ctx context.Context, from *types.TipSet
 	for {
 		// If we've reached the genesis block, or we've reached the limit of
 		// how far back to look
-		if cur.Height() == 0 || !noLimit && cur.Height() <= limitHeight {
+		if cur.Height() == 0 {
 			// it ain't here!
 			return nil, nil, cid.Undef, nil
+		}
+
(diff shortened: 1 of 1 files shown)


magik6k (approved):
So technically this is a slightly breaking change - nil result with nil error indicates that the message wasn't found "successfully", but this impl is arguably more correct - now nil/nil result means that the message is really not there.

Looking around the codebase I don't think this will break anything, but we should test how boost/dealmaking tools behave.

If you want to be extra fancy, you can define a typed error, and add it to https://github.com/filecoin-project/lotus/blob/master/api/api_errors.go#L39, which will make go-jsonrpc set the error code + recover the error type on the other side.

jennijuju:
Please don’t merge before confirming breakage! (The ask is to  add a comment in this PR description to document any potential impact so we can capture in change log later on 

need to confirm whether exchange users this method

rjan90:
Hey! 👋

rjan90:

Closing

Issue #10504, opened 2023-03-17
Cache execution traces (State Compute Cache)


Stebalien:
Some node operators, especially exchanges and chain explorers, heavily rely on execution traces. Ideally said operators would call `StateCompute` once, but this isn't always the case.

We can optimize this by caching the result of:

https://github.com/filecoin-project/lotus/blob/97a3b6e06df365d7af8970f93bad17f4862e7395/chain/stmgr/execute.go#L130-L137

Specifically, we should introduce a simple ARC cache here. It can be small (16-32), but any amount of caching here will likely improve performance. Ideally we'd have a per-tipset lock to avoid duplicating work, but that can be done in a followup if it ends up being expensive.

PR #10517, opened 2023-03-20
feat: Add small cache to execution traces

fridrik01:
## Related Issues
https://github.com/filecoin-project/lotus/issues/10504

## Proposed Changes
This PR adds a small cache to calls to `ExecutionTrace` which helps improve performance for node operators like exchanges and block explorers.

If items is in cache calls to this function will be 2-3x faster. 

## Additional Info
I ran a local test for 30minutes where I called `lotus state compute-state --json` in a loop from two threads. 
- Before this cache, each query took close to 8-10 seconds to run, but with this change it goes down to under 3 seconds.
- Cache hit rate was 82% (671/816)

NOTE: Stress testing this over an hour resulted in Lotus process residual memory taking almost 30GB, I am kind concerned that it grew this large and GC didn't kick in, so will continue to stress test this for longer. 

EDIT: Ran longer stress test, memory stayed flat so should be good

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/execute.go
+++ b/chain/stmgr/execute.go
@@ -128,10 +128,17 @@ func (sm *StateManager) ExecutionTraceWithMonitor(ctx context.Context, ts *types
 }
 
 func (sm *StateManager) ExecutionTrace(ctx context.Context, ts *types.TipSet) (cid.Cid, []*api.InvocResult, error) {
+	if entry, ok := sm.execTraceCache.Get(ts); ok {
+		return entry.cid, entry.invocTrace, nil
+	}
+
 	var invocTrace []*api.InvocResult
 	st, err := sm.ExecutionTraceWithMonitor(ctx, ts, &InvocationTracer{trace: &invocTrace})
 	if err != nil {
 		return cid.Undef, nil, err
 	}
+
+	sm.execTraceCache.Add(ts, tipSetCacheEntry{st, invocTrace})
+
 	return st, invocTrace, nil
 }
--- a/chain/stmgr/execute.go
+++ b/chain/stmgr/execute.go
@@ -130,15 +130,16 @@ func (sm *StateManager) ExecutionTraceWithMonitor(ctx context.Context, ts *types
 func (sm *StateManager) ExecutionTrace(ctx context.Context, ts *types.TipSet) (cid.Cid, []*api.InvocResult, error) {
 	tsKey := ts.Key()
 
-	{
-		// check if we have the trace for this tipset in the cache
-		sm.execTraceCacheLock.Lock()
-		defer sm.execTraceCacheLock.Unlock()
-		if entry, ok := sm.execTraceCache.Get(tsKey); ok {
-			// we have to make a deep copy since caller can modify the invocTrace
-			return entry.postStateRoot, makeDeepCopy(entry.invocTrace), nil
-		}
+	// check if we have the trace for this tipset in the cache
+	sm.execTraceCacheLock.Lock()
+	if entry, ok := sm.execTraceCache.Get(tsKey); ok {
+		// we have to make a deep copy since caller can modify the invocTrace
+		// and we don't want that to change what we store in cache
+		invocTraceCopy := makeDeepCopy(entry.invocTrace)
+		sm.execTraceCacheLock.Unlock()
+		return entry.postStateRoot, invocTraceCopy, nil
 	}
+	sm.execTraceCacheLock.Unlock()
 
 	var invocTrace []*api.InvocResult
 	st, err := sm.ExecutionTraceWithMonitor(ctx, ts, &InvocationTracer{trace: &invocTrace})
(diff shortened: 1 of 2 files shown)


chain/stmgr/stmgr.go
```diff
@@ -136,6 +139,10 @@ type StateManager struct {
 	tsExec        Executor
 	tsExecMonitor ExecMonitor
 	beacon        beacon.Schedule
+
+	// We keep a small cache for calls to ExecutionTrace which helps improve
+	// performance for node operators like exchanges and block explorers
+	execTraceCache *lru.ARCCache[*types.TipSet, tipSetCacheEntry]
```
raulk, line 145: A pointer type like this won't work for a key (this will compare pointers, which will most certainly be different at insertion and access times, even if they point to values that are deeply equal). Let's use the TipsetCid instead.

chain/stmgr/stmgr.go
```diff
@@ -144,6 +151,11 @@ type treeCache struct {
 	tree *state.StateTree
 }
 
+type tipSetCacheEntry struct {
+	cid        cid.Cid
```
raulk, line 155: Let's call this postStateRoot?

chain/stmgr/execute.go
```diff
@@ -128,10 +128,17 @@ func (sm *StateManager) ExecutionTraceWithMonitor(ctx context.Context, ts *types
 }
 
 func (sm *StateManager) ExecutionTrace(ctx context.Context, ts *types.TipSet) (cid.Cid, []*api.InvocResult, error) {
+	if entry, ok := sm.execTraceCache.Get(ts); ok {
+		return entry.cid, entry.invocTrace, nil
```
raulk, line 132: For better safety, you will (unfortunately) want to copy the invocation result objects. If you hand out a slice of pointers, the caller can mutate your objects.

raulk (changes requested):

chain/stmgr/execute.go
fridrik01: Ah, ok so on cache hit we need to do a deep copy of the cached `invocTrace` and return the copy

fridrik01: Also, we also need to make a deep copy when we add to the cache since the caller can mutate that object as well. I assume we also have to deal with thread safety here as well so we need to manage a lock to make sure we have consistent view while making the deep copy.

fridrik01:
Updated the PR, waiting on my lotus to sync so I can test this

chain/stmgr/execute.go
```diff
@@ -128,10 +128,38 @@ func (sm *StateManager) ExecutionTraceWithMonitor(ctx context.Context, ts *types
 }
 
 func (sm *StateManager) ExecutionTrace(ctx context.Context, ts *types.TipSet) (cid.Cid, []*api.InvocResult, error) {
+	tsKey := ts.Key()
+
+	{
+		// check if we have the trace for this tipset in the cache
+		sm.execTraceCacheLock.Lock()
```
arajasek, line 135: Do we need a lock? I thought the ARCCache is thread-safe.

fridrik01: For example, if two threads (T1,T2) enter this function with the same TipSet key and its not in the cache. Let say T1 finishes and stores in the cache. Then T3 enters with the same TipSet key as T1 and T2. For T3 it gets a cache hit. While its making the deep copy, then lets say T2 finishes and updates the cache. This means that when making the deepcopy on a cache hit other threads may update its value. This is therefore a critical section and needs to be guarded.

arajasek: Yeah, I _think_ you're right, and obviously better safe than sorry.

I am, however, concerned about slowing down this workflow by effectively forcing only one ExecutionTrace to be computed (or loaded-and-copied) at a time.

What do we think about dropping the lock before beginning an _actual_ computation of the ExecutionTrace (so line 142), making the deep copy, and then retaking it before writing it to the cache.

fridrik01: @arajasek Ah, I thought the scope I introduced in  (in line 133-141) would run the defer when it exits the scope, however that is probably not the case and how defer works in Go (it only does that on function exit). Good catch!

arajasek (approved):
LGTM

fridrik01:
@raulk could you take another look at this PR so I can get it landed?

chain/stmgr/stmgr.go
```diff
@@ -39,6 +40,8 @@ import (
 const LookbackNoLimit = api.LookbackNoLimit
 const ReceiptAmtBitwidth = 3
 
+const execTraceCacheSize = 16
```
raulk, line 43: We probably want to make this configurable because SPs may want to disable it while exchanges may want to crank it up? We also are going with intuition for this value, so having ability to change it without a new build would help.

raulk (approved):
LGTM but would add an env variable to adjust the cache size. (Probably not a config value, unless we want to expose with a deep knob so openly)

chain/stmgr/stmgr.go
fridrik01: Will make a followup PR which makes this tuneable via env variable so we can get this landed

Issue #10519, opened 2023-03-20
Add tipset -> Message lookup cache


fridrik01:
Right now, resolving the messages in a tipset requires quite a few state loads. We can add a cache in the [ChainStore](https://github.com/filecoin-project/lotus/blob/162e07fc95060399a03dd38104dee92746e4eef0/chain/store/messages.go#L106)’s `BlockMsgsForTipset`

 method. This is probably the biggest “win” as the cache should be trivial and should speed up quite a few lotus APIs.

This will optimize:
- The Eth API block accesses.
-  Fee history.
- Tipset message counts
- StateListMessages
-  (and likely more)

arajasek:
Is this really worth doing? Just based on some quick metrics, this method is currently taking 25ms on average (on my desktop). With [this optimization](https://github.com/filecoin-project/lotus/pull/10552), it falls to 3-5 ms.

PR #10528, opened 2023-03-21
feat: Add tipset -> Message lookup cache

fridrik01:
## Related Issues
See https://github.com/filecoin-project/lotus/issues/10519

## Proposed Changes
This commit adds a cache for looking up block messages in a tipset. This should improve performance for quite a few lotus APIs, including:

- The Eth API block accesses.
- Fee history.
- Tipset message counts
- StateListMessages
- (and likely more)

## Additional Info
I made the cache size configurable through an env variable LOTUS_CHAIN_TIPSET_MESSAGE, so we need to update the Lotus doc accortingly (https://lotus.filecoin.io/lotus/configure/defaults/#environment-variables)

## Checklist

- [x ] Commits have a clear commit message.
- [x ] PR title is in the form of of `<PR type>: <area>: <change being made>`
- [ ] CI is green

### DIFF
--- a/chain/store/messages.go
+++ b/chain/store/messages.go
@@ -105,8 +105,8 @@ type BlockMessages struct {
 
 func (cs *ChainStore) BlockMsgsForTipset(ctx context.Context, ts *types.TipSet) ([]BlockMessages, error) {
 	// returned BlockMessages match block order in tipset
-
-	if entry, ok := cs.tsMsgCace.Get(ts); ok {
+	tsKey := ts.Key()
+	if entry, ok := cs.tsMsgCache.Get(tsKey); ok {
 		return entry, nil
 	}
 
@@ -182,7 +182,7 @@ func (cs *ChainStore) BlockMsgsForTipset(ctx context.Context, ts *types.TipSet) 
 		out = append(out, bm)
 	}
 
-	cs.tsMsgCace.Add(ts, out)
+	cs.tsMsgCache.Add(tsKey, out)
 
 	return out, nil
 }
--- a/chain/store/messages.go
+++ b/chain/store/messages.go
@@ -106,6 +106,10 @@ type BlockMessages struct {
 func (cs *ChainStore) BlockMsgsForTipset(ctx context.Context, ts *types.TipSet) ([]BlockMessages, error) {
 	// returned BlockMessages match block order in tipset
 
+	if entry, ok := cs.tsMsgCace.Get(ts); ok {
+		return entry, nil
+	}
+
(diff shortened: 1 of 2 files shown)


chain/store/messages.go
```diff
@@ -106,6 +106,10 @@ type BlockMessages struct {
 func (cs *ChainStore) BlockMsgsForTipset(ctx context.Context, ts *types.TipSet) ([]BlockMessages, error) {
 	// returned BlockMessages match block order in tipset
 
+	if entry, ok := cs.tsMsgCace.Get(ts); ok {
```
raulk, line 109: Typo?

```suggestion
	if entry, ok := cs.tsMsgCache.Get(ts); ok {
```

raulk: This won't work with a tipset pointer as a key. See the other PR. We'll probably want the TIpsetKey.

chain/store/store.go
```diff
@@ -47,6 +47,7 @@ var (
 
 var DefaultTipSetCacheSize = 8192
 var DefaultMsgMetaCacheSize = 2048
+var DefaultTipSetMessageCacheSize = 8192
```
raulk, line 50: This seems too aggressive. How did we arrive to this number? Assuming an average of 150 messages per tipset, this could be holding a total of 1228800 messages. And if we store this, do we even need the above message meta cache any longer?

chain/store/store.go
```diff
@@ -66,6 +67,14 @@ func init() {
 		}
 		DefaultMsgMetaCacheSize = mmcs
 	}
+
+	if s := os.Getenv("LOTUS_CHAIN_TIPSET_MESSAGE"); s != "" {
```
raulk, line 71:
```suggestion
	if s := os.Getenv("LOTUS_CHAIN_TIPSET_MESSAGE_CACHE"); s != "" {
```

raulk (changes requested):

chain/store/store.go
fridrik01: I just put some number there when I was working on the PR, it was still in draft. I am not sure what number to put here, how large are each message on average? Anyway, I lowered it to 1024 but that may still be too high?

arajasek (commented):
I'm not very convinced this is necessary, to be honest -- see https://github.com/filecoin-project/lotus/issues/10519#issuecomment-1481461487.

fridrik01:
Closing for now, we can review again if we think this is needed once https://github.com/filecoin-project/lotus/pull/10552 lands

Issue #10492, opened 2023-03-16
Post-upgrade: export range fails due to CID not found in the blockstore


hsanjuan:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to lotus.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon:  1.21.0-dev+mainnet+git.5707732+api1.5.0
Local: lotus version 1.21.0-dev+mainnet+git.5707732
```

### Repro Steps

```
lotus chain export-range --internal --messages --receipts --stateroots --workers 50 --head "2684160" --tail "2681280" --write-buffer=5000000 export.car
```

### Describe the Bug

Running Lotus on the commit where the 1.20 branch was merged to master, as the `export-range` functionality becomes available then.

After many days running fine, on the day next to that of the network upgrade (2023-03-14), performing archival snapshots started failing with unretrievable blocks (that hasn't happened in the history of the chain).

```
  | ERROR: exporting chain range: writing object to car, bs.Get: ipld: could not find bafy2bzacecdlvh4udaxpx3motumzj46nms32atnmuxuhusfxcaqzazdrskoso
-- | --
```

The export process traverses and tries to save all blocks and every DagCBOR Cid referenced from those block.Messages, block.ParentMessageReceipts and blocks.ParentStateRoots (when the blocks are among the specified heights).

Apparently `bafy2bzacecdlvh4udaxpx3motumzj46nms32atnmuxuhusfxcaqzazdrskoso` that is nowhere to be found in the Lotus datastore, despite being referenced from somewhere.

Can someone shed some light as to what that might be?

### Logging Information

[...]

PR #10535, opened 2023-03-22
Fix: export-range: Ignore ipld Blocks not found in Receipts.

hsanjuan:
## Related Issues

#10492 

## Proposed Changes

These commits improve error reporting during export-range so that users can easily pinpoint in which epoch and in which part of the export an error happens (particularly getting an error when reading a block to export).

Additionally, given there are references to never-written CIDs in Receipts (see linked issue), rather than aborting the export, it will just print a warning when these are found.  This is less strict than decoding receipts and only skipping EventRoot CIDs, but possibly achieves the same result without the complexity.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/store/snapshot.go
+++ b/chain/store/snapshot.go
@@ -425,7 +425,7 @@ func (s *walkScheduler) processTask(t walkTask, workerN int) error {
 
 	blk, err := s.store.Get(s.ctx, t.c)
 	if errors.Is(err, format.ErrNotFound{}) && t.topLevelTaskType == receiptTask {
-		log.Warnw("ignoring not-found block in Receipts",
+		log.Debugw("ignoring not-found block in Receipts",
 			"block", t.blockCid,
 			"epoch", t.epoch,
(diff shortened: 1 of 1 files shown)


chain/store/snapshot.go
```diff
@@ -416,8 +424,17 @@ func (s *walkScheduler) processTask(t walkTask, workerN int) error {
 	}
 
 	blk, err := s.store.Get(s.ctx, t.c)
+	if errors.Is(err, format.ErrNotFound{}) && t.topLevelTaskType == receiptTask {
+		log.Warnw("ignoring not-found block in Receipts",
+			"block", t.blockCid,
+			"epoch", t.epoch,
+			"cid", t.c)
```
magik6k, line 431: Isn't this really spammy?

hsanjuan: hmm it can be. Will fix

arajasek (approved):

Issue #7002, opened 2021-08-07
ProveCommit aggregate SysErrOutOfGas when publishing and no aggreate


f8-ptrk:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

lotus miner/worker - sealing

### Lotus Version

```text
Daemon:  1.11.1-rc2+mainnet+git.40449f1cc+api1.2.0
Local: lotus-miner version 1.11.1-rc2+mainnet+git.40449f1cc
```

### Describe the Bug

when aggregating provecommits is turned on and the miner manually publishes them via `lotus-miner sectors batching commit --publish-now` and the base Fee is under the threshold for aggregating some of the prove commit messages send to the chain are running into a  SysErrOutOfGas error

https://filfox.info/en/address/f0418632

see messages send at height 999926. 

```

lotus-miner sectors batching commit --publish-now
Batch 0:
	Message: bafy2bzacecbvw2lcwrdmi55capbjzlciit3xkqxx5ixvp7irees4ygi262jx4
	Sectors:
		10587	OK
Batch 1:
	Message: bafy2bzaceausw5bjdki7vudz4b75bqtqxwcy4peakmgachhcvroi2kwqmychk
	Sectors:
		10605	OK
Batch 2:
[...]

PR #10543, opened 2023-03-22
Add feature to stagger sector prove commit submission

shrenujbansal:
## Related Issues

https://github.com/filecoin-project/lotus/issues/7002

## Proposed Changes

Add a config to allow SPs to stagger sector prove commit submissions
This is done because gas estimates for ProveCommits are non deterministic and increasing as a large number of sectors get committed within the same epoch resulting in occasionally failed msgs. Submitting a smaller number of prove commits per epoch would reduce the possibility of failed msgs 

## Checklist

- [ ] CI is green


storage/pipeline/commit_batch.go
```diff
@@ -429,6 +438,8 @@ func (b *CommitBatcher) processIndividually(cfg sealiface.Config) ([]sealiface.C
 		}
 
 		res = append(res, r)
+
+		sectorsProcessed += 1
```
magik6k, line 442: Lint probably wants this to be ++

storage/pipeline/commit_batch.go
```diff
@@ -414,12 +415,20 @@ func (b *CommitBatcher) processIndividually(cfg sealiface.Config) ([]sealiface.C
...
 	for sn, info := range b.todo {
 		r := sealiface.CommitBatchRes{
 			Sectors:       []abi.SectorNumber{sn},
 			FailedSectors: map[abi.SectorNumber]string{},
 		}
 
+		if cfg.MaxSectorProveCommitsSubmittedPerEpoch > 0 &&
+			uint64(sectorsProcessed) >= cfg.MaxSectorProveCommitsSubmittedPerEpoch {
+			time.Sleep(builtin.EpochDurationSeconds * time.Second)
+			sectorsProcessed = 0
```
magik6k, line 429: I'd try to avoid the sleep in any way possible.

Polling api.ChainHead every 2-3s, and looking for height change is probably good enough

node/config/types.go
```diff
@@ -426,6 +426,13 @@ type SealingConfig struct {
 	TerminateBatchMin  uint64
 	TerminateBatchWait Duration
 
+	// When submitting several sector prove commits message simultaneously, this option allows you to
```
magik6k, line 429:
```suggestion
	// When submitting several sector prove commit messages simultaneously, this option allows you to
```

node/config/types.go
```diff
@@ -426,6 +426,13 @@ type SealingConfig struct {
...
+	// stagger the number of prove commits submitted per epoch
+	// This is done because gas estimates for ProveCommits are non deterministic and increasing as a large
+	// number of sectors get committed within the same epoch resulting in occasionally failed msgs.
+	// Submitting a smaller number of prove commits per epoch would reduce the possibility of failed msgs
+	MaxSectorProveCommitsSubmittedPerEpoch uint64
```
magik6k, line 434: Can we move this above `Terminate*` stuff, so all commit things are in one block?

api/api_full.go
```diff
@@ -955,7 +955,7 @@ type MsgLookup struct {
 
 type MsgGasCost struct {
 	Message            cid.Cid // Can be different than requested, in case it was replaced, but only gas values changed
-	GasUsed            abi.TokenAmount
```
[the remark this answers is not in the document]
shrenujbansal: which APIs would this break? I didn't see this part of any of the node RPC APIs
only saw its usage here, which seems like its a util anyway?
`func MakeMsgGasCost(msg *types.Message, ret *vm.ApplyRet) api.MsgGasCost {
`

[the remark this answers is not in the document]
magik6k: Hmmm, it is returned through InvocResult from StateCompute, but looking for uses in code actually shows nothing.

There is small chance that some integrations do use that, but there's also message receipt which seems like the more convenient place to pull this number from. So changing this is probably fine.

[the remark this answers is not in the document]
magik6k: Er, except unmarshaling of the StateCompute result will break, so we probably also want to rename the field to avoid breaking that

storage/pipeline/commit_batch.go
```diff
@@ -414,12 +415,30 @@ func (b *CommitBatcher) processIndividually(cfg sealiface.Config) ([]sealiface.C
...
+			for tmp.Height() <= ts.Height() {
+				tmp, err = b.api.ChainHead(b.mctx)
+				if err != nil {
+					log.Errorf("getting chain head: %+v", err)
+					return nil, err
+				}
+				time.Sleep(3 * time.Second)
+			}
+
+			sectorsProcessed = 0
```
magik6k, line 439: Save tmp to ts here?
```suggestion
			sectorsProcessed = 0
			ts = tmp
```

api/api_full.go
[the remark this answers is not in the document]
shrenujbansal: Ok, I'll leave this unchanged for now

magik6k (approved):
Looks good, would be great to test on a real setup

PR #10548, opened 2023-03-23
hack: don't load prior messages when estimating gas

arajasek:
## Related Issues

Glif nodes are experiencing performance issues. @snissn reported that contention over the message pool locks are causing these RPC calls to be slow.

## Proposed Changes

 **This patch is "wrong", and shouldn't be applied to master, but will likely cause no issues to Glif users.**

This patch changes gas estimation to NOT load pending messages from the mempool, eliminating the slow lock contention observed by Snissn. This will make gas used estimation slightly less accurate when there are pending messages from a sender. This is an issue for SPs, and other users sending large numbers of more "complex" messages (not bare transfers), but likely will not affect most Glif users.

This is one of a few simple hacky-patches we can apply to improve overall performance.

## Checklist

- [ ] CI is green

### DIFF
--- a/node/impl/full/gas.go
+++ b/node/impl/full/gas.go
@@ -262,24 +262,26 @@ func gasEstimateCallWithGas(
 	currTs *types.TipSet,
 ) (*api.InvocResult, []types.ChainMsg, *types.TipSet, error) {
 	msg := *msgIn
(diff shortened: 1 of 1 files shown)


magik6k (commented):
This hack is really hacky.

Almost guaranteed to break things, definitely better to ship https://github.com/filecoin-project/lotus/pull/10545

PR #10552, opened 2023-03-23
feat: chainstore: optimize BlockMsgsForTipset

arajasek:
## Related Issues

@raulk observed some slowness due to the repeated resolution of senders in this method.

It's necessary to resolve senders if AND ONLY IF we have a message sent by an ID address. If we don't (which we almost _never_ do), we can just use robust addresses.

## Proposed Changes

Use robust addresses, instead of ID addresses, to check nonces when selecting BlockMsgsForTipset, BUT check if we have any messages sent from ID addresses.

If we do find any messages sent from ID addresses, switch to using ID addresses for the resolution.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/store/messages.go
+++ b/chain/store/messages.go
@@ -114,12 +114,35 @@ func (cs *ChainStore) BlockMsgsForTipset(ctx context.Context, ts *types.TipSet) 
 		return nil, xerrors.Errorf("failed to load state tree at tipset %s: %w", ts, err)
 	}
 
+	useIds := false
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
This assumes that accounts will only ever have an f0 plus an f1/f3/f4 address, never more. That's currently true and likely to continue to be true? Maybe? With account abstraction, that may go away.

But I guess this should be OK for now.

PR #10553, opened 2023-03-23
feat: stmgr: speed up calculation of genesis circ supply

arajasek:
## Related Issues

@jennijuju's goroutines reported that when trying to send many messages (or just create a lot of FVMs), this lock becomes contended for.

## Proposed Changes

- Drop the check on `genesisMarketFunds`, which is zero on mainnet (confirmed)
- Only take the lock over the genesis multisigs when actually calculating vesting schedules, which _should_ happen once, but _was_ happening every time because `genesisMarketFunds` was equal to 0

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/supply.go
+++ b/chain/stmgr/supply.go
@@ -56,6 +56,8 @@ func (sm *StateManager) setupGenesisVestingSchedule(ctx context.Context) error {
 		return xerrors.Errorf("setting up genesis pledge: %w", err)
 	}
 
+	sm.genesisMsigLk.Lock()
+	defer sm.genesisMsigLk.Unlock()
 	sm.genesisPledge = gp
 
 	totalsByEpoch := make(map[abi.ChainEpoch]abi.TokenAmount)
@@ -122,6 +124,8 @@ func (sm *StateManager) setupPostIgnitionVesting(ctx context.Context) error {
 	totalsByEpoch[sixYears] = big.NewInt(100_000_000)
 	totalsByEpoch[sixYears] = big.Add(totalsByEpoch[sixYears], big.NewInt(300_000_000))
 
+	sm.genesisMsigLk.Lock()
+	defer sm.genesisMsigLk.Unlock()
 	sm.postIgnitionVesting = make([]msig0.State, 0, len(totalsByEpoch))
 	for k, v := range totalsByEpoch {
 		ns := msig0.State{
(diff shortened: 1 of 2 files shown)


chain/stmgr/supply.go
```diff
@@ -198,27 +192,32 @@ func (sm *StateManager) setupPostCalicoVesting(ctx context.Context) error {
 func (sm *StateManager) GetFilVested(ctx context.Context, height abi.ChainEpoch) (abi.TokenAmount, error) {
 	vf := big.Zero()
 
-	sm.genesisMsigLk.Lock()
-	defer sm.genesisMsigLk.Unlock()
-
 	// TODO: combine all this?
-	if sm.preIgnitionVesting == nil || sm.genesisPledge.IsZero() || sm.genesisMarketFunds.IsZero() {
+	if sm.preIgnitionVesting == nil || sm.genesisPledge.IsZero() {
+		sm.genesisMsigLk.Lock()
```
shrenujbansal, line 197: Move lock to surround these 3 if statements instead of 3 different lock statements?

chain/stmgr/supply.go
```diff
@@ -198,27 +192,32 @@ func (sm *StateManager) setupPostCalicoVesting(ctx context.Context) error {
...
 		err := sm.setupGenesisVestingSchedule(ctx)
 		if err != nil {
 			return vf, xerrors.Errorf("failed to setup pre-ignition vesting schedule: %w", err)
 		}
+		sm.genesisMsigLk.Unlock()
```
shrenujbansal, line 202: Each of these unlocks don't happen if we err out above

arajasek: Quite right, I'll move it up above the err return.

chain/stmgr/supply.go
```diff
@@ -198,24 +192,29 @@ func (sm *StateManager) setupPostCalicoVesting(ctx context.Context) error {
 func (sm *StateManager) GetFilVested(ctx context.Context, height abi.ChainEpoch) (abi.TokenAmount, error) {
 	vf := big.Zero()
 
-	sm.genesisMsigLk.Lock()
-	defer sm.genesisMsigLk.Unlock()
-
 	// TODO: combine all this?
-	if sm.preIgnitionVesting == nil || sm.genesisPledge.IsZero() || sm.genesisMarketFunds.IsZero() {
+	if sm.preIgnitionVesting == nil || sm.genesisPledge.IsZero() {
+		sm.genesisMsigLk.Lock()
```
shrenujbansal, line 197: I think it might make more sense to move this lock into each of the vesting functions as they are also owned by the state manager and we are trying to make this lock more granular

arajasek: Sure, I'm happy to do that. Note that we _should_ only be calling each of these methods once, but I think that makes sense.

shrenujbansal (approved):

Issue #10538, opened 2023-03-22
Eth RPC: Address performance of EthGetTransactionCount


maciejwitowski:
We identified `EthGetTransactionCount` as one of the hotspots on Glif production nodes. The goal of this issue is to investigate the reason and fix it.

![image](https://user-images.githubusercontent.com/17002645/226991823-8252cceb-82bb-4933-925a-a259735269a2.png)

jennijuju:
@snissn i think we should  look into  MpoolGetNonce - which exchanges have been reporting a 6-8 sec slow response. If we can narrow down the issue there - it should improve eth get tx count too

snissn:
i'm having trouble reproducing on master -- i'm running a lotus node and pummeling the node with the bash script below the response time is between 20 and 50 ms. Could there just be other slow endpoints causing their performance issues but eth_getTransactionCount is fast? 

```
while true; do
        {
                time curl  http://localhost:1234/rpc/v1  -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionCount","params":["0x7C2540eE20773995a7304A2ab6fd785cBBef8354", "latest"],"id":1,"jsonrpc":"2.0"}'
        } 2>&1 |grep real|cut -f2 -dm |cut -f1 -ds
sleep 0.1
done
```

jennijuju:
We might need to spin up a local net, then send like 300 messages and having them in mpool, and then call the api to repro

snissn:
this is interesting - an LRU cache for nonce with only 256 slots in it  https://github.com/filecoin-project/lotus/blob/master/chain/messagepool/messagepool.go#L374

Stebalien:
So, the issue is that https://github.com/filecoin-project/lotus/blob/3f74840b67d7e84949939bf281c7427f4500875e/chain/messagepool/messagepool.go#L1081 computes the tipset under a lock.


The answer is to not call `GetActorAfter`. Instead, we need to look at the tipset iself:

1. Check if the nonce is cached.
2. If not, load the tipset messages, look for the relevent message, and fill the cache with all nonces found.
3. Check the state _before_ the tipset (i.e., `GetActor`, not `GetActorAfter`).
4. Update the cache.

At the same time, we'll _definitely_ want to increase the size of this cache. Also, for the current head, we could pre-cache in `HeadChange`. We _already_ process all messages to update the message pool, so this operation should be _basically_ free.

maciejwitowski:
[...]

PR #10561, opened 2023-03-25
perf: message pool: change locks to RWMutexes for performance

snissn:
## Related Issues
should effect https://github.com/filecoin-project/lotus/issues/10538 because of message pool locks

## Proposed Changes
message pool has two locks - change mp.lk and mp.curTsLk to RW locks for API efficiency 

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/messagepool/check.go
+++ b/chain/messagepool/check.go
@@ -32,14 +32,14 @@ func (mp *MessagePool) CheckMessages(ctx context.Context, protos []*api.MessageP
 // CheckPendingMessages performs a set of logical sets for all messages pending from a given actor
 func (mp *MessagePool) CheckPendingMessages(ctx context.Context, from address.Address) ([][]api.MessageCheckStatus, error) {
 	var msgs []*types.Message
-	mp.lk.Lock()
+	mp.lk.RLock()
 	mset, ok := mp.pending[from]
 	if ok {
 		for _, sm := range mset.msgs {
 			msgs = append(msgs, &sm.Message)
 		}
 	}
-	mp.lk.Unlock()
+	mp.lk.RUnlock()
 
 	if len(msgs) == 0 {
 		return nil, nil
@@ -58,7 +58,7 @@ func (mp *MessagePool) CheckReplaceMessages(ctx context.Context, replace []*type
 	msgMap := make(map[address.Address]map[uint64]*types.Message)
 	count := 0
 
-	mp.lk.Lock()
+	mp.lk.RLock()
 	for _, m := range replace {
 		mmap, ok := msgMap[m.From]
 		if !ok {
@@ -76,7 +76,7 @@ func (mp *MessagePool) CheckReplaceMessages(ctx context.Context, replace []*type
 		}
 		mmap[m.Nonce] = m
 	}
-	mp.lk.Unlock()
+	mp.lk.RUnlock()
 
 	msgs := make([]*types.Message, 0, count)
 	start := 0
@@ -103,9 +103,9 @@ func (mp *MessagePool) checkMessages(ctx context.Context, msgs []*types.Message,
 	if mp.api.IsLite() {
 		return nil, nil
 	}
-	mp.curTsLk.Lock()
+	mp.curTsLk.RLock()
 	curTs := mp.curTs
-	mp.curTsLk.Unlock()
+	mp.curTsLk.RUnlock()
 
 	epoch := curTs.Height() + 1
 
@@ -143,7 +143,7 @@ func (mp *MessagePool) checkMessages(ctx context.Context, msgs []*types.Message,
 
 		st, ok := state[m.From]
 		if !ok {
-			mp.lk.Lock()
+			mp.lk.RLock()
 			mset, ok := mp.pending[m.From]
 			if ok && !interned {
 				st = &actorState{nextNonce: mset.nextNonce, requiredFunds: mset.requiredFunds}
@@ -151,14 +151,14 @@ func (mp *MessagePool) checkMessages(ctx context.Context, msgs []*types.Message,
 					st.requiredFunds = new(stdbig.Int).Add(st.requiredFunds, m.Message.Value.Int)
 				}
 				state[m.From] = st
-				mp.lk.Unlock()
+				mp.lk.RUnlock()
 
 				check.OK = true
 				check.Hint = map[string]interface{}{
 					"nonce": st.nextNonce,
 				}
 			} else {
-				mp.lk.Unlock()
+				mp.lk.RUnlock()
 
 				stateNonce, err := mp.getStateNonce(ctx, m.From, curTs)
 				if err != nil {
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -118,7 +118,7 @@ func init() {
 }
 
 type MessagePool struct {
-	lk sync.Mutex
+	lk sync.RWMutex
 
 	ds dtypes.MetadataDS
 
@@ -139,7 +139,7 @@ type MessagePool struct {
 
 	keyCache map[address.Address]address.Address
 
-	curTsLk sync.Mutex // DO NOT LOCK INSIDE lk
+	curTsLk sync.RWMutex // DO NOT LOCK INSIDE lk
 	curTs   *types.TipSet
 
 	cfgLk sync.RWMutex
@@ -1001,19 +1001,19 @@ func (mp *MessagePool) addLocked(ctx context.Context, m *types.SignedMessage, st
 }
 
 func (mp *MessagePool) GetNonce(ctx context.Context, addr address.Address, _ types.TipSetKey) (uint64, error) {
-	mp.curTsLk.Lock()
-	defer mp.curTsLk.Unlock()
+	mp.curTsLk.RLock()
+	defer mp.curTsLk.RUnlock()
 
-	mp.lk.Lock()
-	defer mp.lk.Unlock()
+	mp.lk.RLock()
+	defer mp.lk.RUnlock()
 
 	return mp.getNonceLocked(ctx, addr, mp.curTs)
 }
 
 // GetActor should not be used. It is only here to satisfy interface mess caused by lite node handling
 func (mp *MessagePool) GetActor(_ context.Context, addr address.Address, _ types.TipSetKey) (*types.Actor, error) {
-	mp.curTsLk.Lock()
-	defer mp.curTsLk.Unlock()
+	mp.curTsLk.RLock()
+	defer mp.curTsLk.RUnlock()
 	return mp.api.GetActorAfter(addr, mp.curTs)
 }
 
@@ -1164,11 +1164,11 @@ func (mp *MessagePool) remove(ctx context.Context, from address.Address, nonce u
 }
 
 func (mp *MessagePool) Pending(ctx context.Context) ([]*types.SignedMessage, *types.TipSet) {
-	mp.curTsLk.Lock()
-	defer mp.curTsLk.Unlock()
+	mp.curTsLk.RLock()
+	defer mp.curTsLk.RUnlock()
 
-	mp.lk.Lock()
-	defer mp.lk.Unlock()
+	mp.lk.RLock()
+	defer mp.lk.RUnlock()
 
 	return mp.allPending(ctx)
 }
@@ -1184,11 +1184,11 @@ func (mp *MessagePool) allPending(ctx context.Context) ([]*types.SignedMessage, 
 }
 
 func (mp *MessagePool) PendingFor(ctx context.Context, a address.Address) ([]*types.SignedMessage, *types.TipSet) {
-	mp.curTsLk.Lock()
-	defer mp.curTsLk.Unlock()
+	mp.curTsLk.RLock()
+	defer mp.curTsLk.RUnlock()
 
-	mp.lk.Lock()
-	defer mp.lk.Unlock()
+	mp.lk.RLock()
+	defer mp.lk.RUnlock()
 	return mp.pendingFor(ctx, a), mp.curTs
 }
 
@@ -1237,9 +1237,9 @@ func (mp *MessagePool) HeadChange(ctx context.Context, revert []*types.TipSet, a
 
 	maybeRepub := func(cid cid.Cid) {
 		if !repubTrigger {
-			mp.lk.Lock()
+			mp.lk.RLock()
 			_, republished := mp.republished[cid]
-			mp.lk.Unlock()
+			mp.lk.RUnlock()
 			if republished {
 				repubTrigger = true
 			}
@@ -1310,9 +1310,9 @@ func (mp *MessagePool) HeadChange(ctx context.Context, revert []*types.TipSet, a
 	}
 
 	if len(revert) > 0 && futureDebug {
-		mp.lk.Lock()
+		mp.lk.RLock()
 		msgs, ts := mp.allPending(ctx)
-		mp.lk.Unlock()
+		mp.lk.RUnlock()
 
 		buckets := map[address.Address]*statBucket{}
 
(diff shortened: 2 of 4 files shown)


chain/messagepool/check.go
```diff
@@ -32,14 +32,14 @@ func (mp *MessagePool) CheckMessages(ctx context.Context, protos []*api.MessageP
 // CheckPendingMessages performs a set of logical sets for all messages pending from a given actor
 func (mp *MessagePool) CheckPendingMessages(ctx context.Context, from address.Address) ([][]api.MessageCheckStatus, error) {
 	var msgs []*types.Message
-	mp.lk.Lock()
```
Jorropo, line 35: ~~`RLock` is more complex and can be much slower in some condition, I would not trust this without a benchmark.~~

Actually I see this is already an `sync.RWMutex`, I wouldn't trust this but this is not like this changing much compared to the current state so that probably fine.

chain/messagepool/check.go
```diff
@@ -32,14 +32,14 @@ func (mp *MessagePool) CheckMessages(ctx context.Context, protos []*api.MessageP
...
+	mp.lk.RLock()
 	mset, ok := mp.pending[from]
 	if ok {
 		for _, sm := range mset.msgs {
```
Jorropo, line 38: That will undoubtly makes thoses critical sections faster (replace the `allocate, copy, grow -> repeat` pattern of append with a single allocation).
```suggestion
		msgs = make([]*types.Message, len(mset.msgs))
		for _, sm := range mset.msgs {
```

Jorropo (changes requested):

chain/messagepool/selection.go
```diff
@@ -43,6 +43,7 @@ func (mp *MessagePool) SelectMessages(ctx context.Context, ts *types.TipSet, tq
 	mp.curTsLk.Lock()
 	defer mp.curTsLk.Unlock()
 
+	//TODO confirm if we can switch to RLock here for performance
```
magik6k, line 46: Yes, that should be safe

magik6k (commented):
Generally looks good, we can switch to RLock in message selection as well.

snissn:
closed #10545 for this -- 

The lock contention seems gone from the eth rpc path now! 100 requests per second compared with 8 seconds per request :) :) 

```
$ cat bench.sh
ab -p post_data.json -T application/json  -c 10 -n 20000 http://localhost:1234/rpc/v1
ubuntu@ip-10-0-4-29:~$ bash bench.sh |tee bench-PR.out
This is ApacheBench, Version 2.3 <$Revision: 1843412 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 2000 requests
Completed 4000 requests
Completed 6000 requests
Completed 8000 requests
Completed 10000 requests
Completed 12000 requests
Completed 14000 requests
Completed 16000 requests
Completed 18000 requests
Completed 20000 requests
Finished 20000 requests


Server Software:
Server Hostname:        localhost
Server Port:            1234

Document Path:          /rpc/v1
Document Length:        45 bytes

Concurrency Level:      10
Time taken for tests:   189.902 seconds
Complete requests:      20000
Failed requests:        0
Total transferred:      3240000 bytes
Total body sent:        6700000
HTML transferred:       900000 bytes
Requests per second:    105.32 [#/sec] (mean)
Time per request:       94.951 [ms] (mean)
Time per request:       9.495 [ms] (mean, across all concurrent requests)
Transfer rate:          16.66 [Kbytes/sec] received
                        34.45 kb/s sent
                        51.12 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    29   95  59.0     83     843
Waiting:       29   95  59.0     83     843
Total:         29   95  59.0     83     843

Percentage of the requests served within a certain time (ms)
  50%     83
  66%     93
  75%    112
  80%    124
  90%    163
  95%    210
  98%    268
  99%    316
 100%    843 (longest request)
```


Open for reviews and nits!

snissn:
I'm looking into the failing tests not sure if actual or flaky

snissn:
hmm definitely still seeing problems with this - not ready just yet

snissn:
Ok great! I think the last issue was [repub needs Lock not RLock](https://github.com/filecoin-project/lotus/pull/10561/commits/831f8a499d33aab69e9859f1dd4baecff4888e70)
[831f8a4](https://github.com/filecoin-project/lotus/pull/10561/commits/831f8a499d33aab69e9859f1dd4baecff4888e70)

tests are passing now, ready for review

chain/messagepool/messagepool.go
```diff
@@ -763,7 +763,30 @@ func (mp *MessagePool) Add(ctx context.Context, m *types.SignedMessage) error {
...
+	mp.curTsLk.RLock()
+	tmpCurTs := mp.curTs
+	mp.curTsLk.RUnlock()
+	_, _ = mp.api.GetActorAfter(m.Message.From, tmpCurTs)
+	_, _ = mp.getStateNonce(ctx, m.Message.From, tmpCurTs)
+
+	cacheSecondTime := true
+	//if the newly acquired Ts is not the one we just cached, let go of the lock, cache it and open the lock again and repeat....
+	for cacheSecondTime {
+		mp.curTsLk.Lock()
```
arajasek, line 776: I think we just want an `RLock` here

chain/messagepool/messagepool.go
```diff
@@ -763,7 +763,30 @@ func (mp *MessagePool) Add(ctx context.Context, m *types.SignedMessage) error {
...
+		writeCurTs := mp.curTs
+
+		if writeCurTs == tmpCurTs {
+			break // we have this cached we can skip
+		}
+		mp.curTsLk.Unlock()
+		tmpCurTs = writeCurTs
+		_, _ = mp.api.GetActorAfter(m.Message.From, tmpCurTs)
+		_, _ = mp.getStateNonce(ctx, m.Message.From, tmpCurTs)
+		cacheSecondTime = false
```
arajasek, line 786: Are we sure we wanna set this to false? Do we not wanna loop again, so as to check whether curTs has changed?

I think we only ever exit this loop at the break on line 780, at which point we can be confident we still hold the curTsLk, but I'm not completely sure.

chain/messagepool/messagepool.go
```diff
@@ -852,14 +875,14 @@ func (mp *MessagePool) addTs(ctx context.Context, m *types.SignedMessage, curTs
...
 
-	mp.lk.Lock()
-	defer mp.lk.Unlock()
-
 	senderAct, err := mp.api.GetActorAfter(m.Message.From, curTs)
 	if err != nil {
 		return false, xerrors.Errorf("failed to get sender actor: %w", err)
 	}
 
+	mp.lk.Lock()
```
arajasek, line 883: I think this can go a little further down (before the call to `verifyMsgBeforeAdd`, at minimum?)

chain/messagepool/repub.go
```diff
@@ -20,17 +20,18 @@ const repubMsgLimit = 30
...
 	baseFee, err := mp.api.ChainComputeBaseFee(context.TODO(), ts)
+	mp.curTsLk.RUnlock()
 	if err != nil {
-		mp.curTsLk.Unlock()
 		return xerrors.Errorf("computing basefee: %w", err)
 	}
 	baseFeeLowerBound := getBaseFeeLowerBound(baseFee, baseFeeLowerBoundFactor)
 
 	pending := make(map[address.Address]map[uint64]*types.SignedMessage)
+	mp.curTsLk.Lock()
```
arajasek, line 34: Also RLock, I think?

chain/messagepool/messagepool.go
```diff
@@ -763,7 +763,28 @@ func (mp *MessagePool) Add(ctx context.Context, m *types.SignedMessage) error {
...
+
 	mp.curTsLk.Lock()
+	if tmpCurTs == mp.curTs {
+		//with the lock enabled, mp.curTs is the same Ts as we just had, so we know that our computations are cached
+	} else {
+		//curTs has been updated so we want to cache the new one:
+		tmpCurTs = mp.curTs
+		//we want to release the lock, cache the computations then grab it again
+		mp.curTsLk.Unlock()
+		_, _ = mp.api.GetActorAfter(m.Message.From, tmpCurTs)
```
Stebalien, line 782: We either need to keep the lock, or do this in a loop.

chain/messagepool/repub.go
Stebalien: Not sure if there's any point releasing the lock here either.

chain/messagepool/messagepool.go
Stebalien: Or, if we're just doing it for the side-effect, we can just drop this check entirely. It'll be slow occasionally, but not normally.

Stebalien: Nevermind, you explained it below.

chain/messagepool/messagepool.go
```diff
@@ -1001,19 +1022,19 @@ func (mp *MessagePool) addLocked(ctx context.Context, m *types.SignedMessage, st
...
-	defer mp.curTsLk.Unlock()
+	mp.curTsLk.RLock()
+	defer mp.curTsLk.RUnlock()
 
-	mp.lk.Lock()
-	defer mp.lk.Unlock()
+	mp.lk.RLock()
+	defer mp.lk.RUnlock()
 
 	return mp.getNonceLocked(ctx, addr, mp.curTs)
```
Stebalien, line 1031: This currently needs a write lock for:

https://github.com/filecoin-project/lotus/blob/841b7b873005c6ddfb536125ca9a90b92b23d8c9/chain/messagepool/messagepool.go#L489-L490

We should consider replacing the `keyCache` with an LRU anyways? That locks internally so we wouldn't need to worry about this.

chain/messagepool/messagepool.go
```diff
@@ -1184,11 +1205,11 @@ func (mp *MessagePool) allPending(ctx context.Context) ([]*types.SignedMessage,
...
-	mp.curTsLk.Lock()
-	defer mp.curTsLk.Unlock()
+	mp.curTsLk.RLock()
+	defer mp.curTsLk.RUnlock()
 
-	mp.lk.Lock()
-	defer mp.lk.Unlock()
+	mp.lk.RLock()
+	defer mp.lk.RUnlock()
 	return mp.pendingFor(ctx, a), mp.curTs
```
Stebalien, line 1213: This calls `getPendignMset` which calls `resolveToKey` which requires a write lock, at the moment (see above).

chain/messagepool/check.go
arajasek: You're right, but this method isn't on the critical path for mpool functionality. Still worth doing, perhaps as a separate patch.

chain/messagepool/messagepool.go
```diff
@@ -474,8 +475,8 @@ func (mp *MessagePool) TryForEachPendingMessage(f func(cid.Cid) error) error {
 
 func (mp *MessagePool) resolveToKey(ctx context.Context, addr address.Address) (address.Address, error) {
 	// check the cache
```
arajasek, line 477: Can we just check whether the sender is a robust address (anything except an ID address is robust) here, and return if so? That will:

- be faster
- reduce the size of this cache by half, because we can drop `mp.keyCache.Add(ka, ka)` on line 491.

chain/messagepool/messagepool.go
```diff
@@ -763,7 +764,28 @@ func (mp *MessagePool) Add(ctx context.Context, m *types.SignedMessage) error {
 		<-mp.addSema
 	}()
 
+	mp.curTsLk.RLock()
+	tmpCurTs := mp.curTs
+	mp.curTsLk.RUnlock()
+
+	//ensures computations are cached without holding lack
```
arajasek, line 771:
```suggestion
	//ensures computations are cached without holding lock
```

chain/messagepool/messagepool.go
Stebalien: (seems like something that can be done in a future patch)

chain/messagepool/messagepool.go
snissn: resolved by making keyCache an LRU cache

chain/messagepool/selection.go
snissn: thank you! Logged for this enhancement issue here: https://github.com/filecoin-project/lotus/issues/10597

chain/messagepool/repub.go
```diff
@@ -20,18 +20,18 @@ const repubMsgLimit = 30
...
-		mp.curTsLk.Unlock()
+		mp.curTsLk.RUnlock()
 		return xerrors.Errorf("computing basefee: %w", err)
 	}
 	baseFeeLowerBound := getBaseFeeLowerBound(baseFee, baseFeeLowerBoundFactor)
 
 	pending := make(map[address.Address]map[uint64]*types.SignedMessage)
-	mp.lk.Lock()
+	mp.lk.RLock()
 	mp.republished = nil // clear this to avoid races triggering an early republish
```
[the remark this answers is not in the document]
snissn: logged here to confirm whether we need a Lock or an Rlock in enhancement issue: https://github.com/filecoin-project/lotus/issues/10597

chain/messagepool/repub.go
snissn: Thanks - I have logged to investigate and optimize this in the third item of this enhancement issue: https://github.com/filecoin-project/lotus/issues/10597

chain/messagepool/messagepool.go
snissn: Thanks - I have logged this as the second item in this enhancement issue: https://github.com/filecoin-project/lotus/issues/10597

chain/messagepool/check.go
snissn: Thank you @Jorropo I have logged your change in the first item of this enhancement issue: https://github.com/filecoin-project/lotus/issues/10597

chain/messagepool/messagepool.go
snissn: Hey! I rewrote the change and commited it here: 2fa95a09be11deeeadb7fdd5401814178c6d9307 Steb reviewed approves

chain/messagepool/repub.go
```diff
@@ -20,17 +20,18 @@ const repubMsgLimit = 30
 var RepublishBatchDelay = 100 * time.Millisecond
 
 func (mp *MessagePool) republishPendingMessages(ctx context.Context) error {
-	mp.curTsLk.Lock()
+	mp.curTsLk.RLock()
 	ts := mp.curTs
 
 	baseFee, err := mp.api.ChainComputeBaseFee(context.TODO(), ts)
+	mp.curTsLk.RUnlock()
```
arajasek, line 27: Move this up one more line? We don't need to hold the curTsLk while base fee is computed over the API.

arajasek (approved):

chain/messagepool/repub.go
snissn: Added this note to #3 on the follow on PR - https://github.com/filecoin-project/lotus/issues/10597

chain/messagepool/check.go
```diff
@@ -32,14 +32,14 @@ func (mp *MessagePool) CheckMessages(ctx context.Context, protos []*api.MessageP
 // CheckPendingMessages performs a set of logical sets for all messages pending from a given actor
 func (mp *MessagePool) CheckPendingMessages(ctx context.Context, from address.Address) ([][]api.MessageCheckStatus, error) {
 	var msgs []*types.Message
-	mp.lk.Lock()
+	mp.lk.RLock()
 	mset, ok := mp.pending[from]
```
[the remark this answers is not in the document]
snissn: thanks! added to follow on PR as #5 and #6

PR #10572, opened 2023-03-27
fix: itests: Don't call t.Error in MineBlocks goroutine

magik6k:
## Related Issues
This is a probable cause for <s>a lot</s> maybe some of flakiness in all itests - https://github.com/filecoin-project/lotus/issues?q=is%3Aissue+is%3Aopen+label%3Aimpact%2Ftest-flakiness

## Proposed Changes
When `miner.MineOne` returns an error, which can occasionally happen on normal test shutdown with unlucky timing (e.g. chain blockstore closed during a call to MinerGetBaseInfo while it is computing state), just log the error and quit.

## Additional Info
Previously the call to t.Error could make the failure look very random as it is called in a goroutine, and may be called after the original calling (sub)test has already finished.

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/batch_deal_test.go
+++ b/itests/batch_deal_test.go
@@ -37,6 +37,8 @@ func TestBatchDealInput(t *testing.T) {
 
 	run := func(piece, deals, expectSectors int) func(t *testing.T) {
 		return func(t *testing.T) {
+			t.Logf("batchtest start")
+
 			ctx := context.Background()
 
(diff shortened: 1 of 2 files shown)


arajasek (approved):
I think this makes sense, thanks.

I'm not sure whether you intend to merge the additional logging to batch_deal_test, but I would be in favour of doing so -- this test basically fails 100% of the time, so more info good.

Issue #10584, opened 2023-03-29
Make execution trace cache size configurable via env variable


fridrik01:
## Related issue 

https://github.com/filecoin-project/lotus/issues/10504

## Proposed change

We want to make the execution trace cache size configurable SPs may want to disable it while exchanges may want to crank it up. We were also are going with intuition for this value, so having ability to change it without a new build would help.

PR #10585, opened 2023-03-29
feat: Make execution trace configurable via env variable

fridrik01:
## Related Issues
https://github.com/filecoin-project/lotus/issues/10584
https://github.com/filecoin-project/lotus/issues/10504

## Proposed Changes
We want to make the execution trace cache size configurable as SPs may want to disable it while exchanges may want to crank it up.

We were also are going with intuition for this value, so having ability to change it without a new build would help.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/stmgr.go
+++ b/chain/stmgr/stmgr.go
@@ -80,9 +80,9 @@ func init() {
 		letc, err := strconv.Atoi(s)
 		if err != nil {
 			log.Errorf("failed to parse 'LOTUS_EXEC_TRACE_CACHE' env var: %s", err)
+		} else {
+			defaultExecTraceCacheSize = letc
 		}
-
-		defaultExecTraceCacheSize = letc
 	}
 }
 
--- a/chain/stmgr/stmgr.go
+++ b/chain/stmgr/stmgr.go
@@ -3,6 +3,8 @@ package stmgr
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 	"sync"
 
 	lru "github.com/hashicorp/golang-lru/v2"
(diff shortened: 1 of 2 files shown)


chain/stmgr/stmgr.go
```diff
@@ -74,6 +75,17 @@ func (m *migrationResultCache) keyForMigration(root cid.Cid) dstore.Key {
...
 }
 
+func init() {
+	if s := os.Getenv("LOTUS_EXEC_TRACE_CACHE"); s != "" {
+		letc, err := strconv.Atoi(s)
+		if err != nil {
+			log.Errorf("failed to parse 'LOTUS_EXEC_TRACE_CACHE' env var: %s", err)
+		}
+
+		defaultExecTraceCacheSize = letc
```
vyzo, line 85: this will zero if it fails to parse, is that what we want?

vyzo (commented):
looks reasonable, just a comment on initialkzation.

chain/stmgr/stmgr.go
```diff
@@ -74,6 +75,17 @@ func (m *migrationResultCache) keyForMigration(root cid.Cid) dstore.Key {
 	return dstore.NewKey(kStr)
 }
 
+func init() {
+	if s := os.Getenv("LOTUS_EXEC_TRACE_CACHE"); s != "" {
```
arajasek, line 79: Rename to LOTUS_EXEC_TRACE_CACHE_SIZE

chain/stmgr/stmgr.go
```diff
@@ -40,8 +42,7 @@ import (
 const LookbackNoLimit = api.LookbackNoLimit
 const ReceiptAmtBitwidth = 3
 
-const execTraceCacheSize = 16
-
+var defaultExecTraceCacheSize = 16
```
arajasek, line 45: Keep the old name? Because if the envvar is set, we set this field, which means it's no longer a default size?

arajasek (approved):

Issue #9073, opened 2022-07-21
`GET/Fetch` clutter the logs with ERROR messages looking for sectors that are not created yet


Reiers:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### Lotus Version

```text
Daemon:  1.17.0+mainnet+git.2830429ad+api1.5.0
Local: lotus version 1.17.0+mainnet+git.2830429ad
```

### Describe the Bug

During `FinalizeReplicaUpdate` the logs get filled up with these ERROR messages. 
**It doesn't not break anything,** but the logs get unreadable with 50+ SnapDeals in `FinalizeReplicaUpdate`.

### Logging Information

```text
[...]

PR #10588, opened 2023-03-29
fix: log: Stop logging `file does not exists`

rjan90:
## Related Issues
fixes: #9073 

## Proposed Changes
Do not log 'file does not exist' errors when retrieving disk usage information. These errors are expected, as storage can be reserved before any files are written, and they were cluttering the logs unnecessarily.

## Checklist

- [ ] CI is green

### DIFF
--- a/storage/paths/localstorage_cached.go
+++ b/storage/paths/localstorage_cached.go
@@ -1,6 +1,7 @@
 package paths
 
 import (
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


storage/paths/localstorage_cached.go
```diff
@@ -103,7 +104,9 @@ func (c *cachedLocalStorage) DiskUsage(path string) (int64, error) {
 		go func() {
 			du, err := c.base.DiskUsage(path)
 			if err != nil {
-				log.Errorw("error getting disk usage", "path", path, "error", err)
+				if !os.IsNotExist(err) {
```
arajasek, line 107: Maybe still debug-log the IsNotExist error?

arajasek (approved):

storage/paths/localstorage_cached.go
rjan90: Based on the this code it seems like we actually do not care: https://github.com/filecoin-project/lotus/blob/a99946937c3ae82c9b3b89baeb5f883abad92eae/storage/paths/local.go#L83-L91

I think the reason for it, is that it can take quite a bit of time between the sector being reserved, and the file being transferred over.

I can add the same comment to this part of the code as well?

rjan90: It also seems from my testing that the defualt logging setting for `stores` is DEBUG, which makes this logging a bit unbearable. I will look into tuning that down in another PR.

PR #10392, opened 2023-03-05
Spike: Pause compaction when out of sync

ZenGround0:
## Related Issues

I ended up playing with some of the chain sync contention protection ideas from #10388 

## Proposed Changes

Use head change out of sync detection to signal compaction to pause.  Signal continue when chain is back in sync

## Additional Info

@vyzo if you can take a look at this for feasibility it would be a big help.

## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #10621, opened 2023-04-04
feat:splitstore:limit moving gc threads

ZenGround0:
## Related Issues

After recent fix #10392 *should* be fine but we just had a bug report that wasn't caught during test so let's let that change sit a little before incorporating in a release. 

However we do want to add configuration that makes moving GC more common to the release for better disk management.  This PR is a small subset of #10392 which should make moving gc less likely to cause out of sync that is really safe without more testing time.

## Checklist

- [ ] CI is green

### DIFF
--- a/blockstore/badger/blockstore.go
+++ b/blockstore/badger/blockstore.go
@@ -396,6 +396,9 @@ func (b *Blockstore) doCopy(from, to *badger.DB) error {
 	if workers < 2 {
 		workers = 2
 	}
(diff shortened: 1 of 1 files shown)


arajasek (approved):

arajasek:
LGTM -- so we wanna land this in `master`, backport it onto the release, and then rebase #10392 on top of latest master (which will have this change)?

PR #10633, opened 2023-04-06
Fix 1.21 regression: GET_32G_MAX_CONCURRENT + mixed prepared/executing leads to stuck scheduler

steffengy:
If you have 12 GET tasks and GET_32G_MAX_CONCURRENT=1, sealing jobs will only show assigned tasks for GET of the miner and is stuck.
I believe this to be a regression of 1.21 unifying the counters, in the case of GETs where PrepType and TaskType both being seal/v0/fetch leading to a state where tasks are blocked since already counted towards the limit.

So while #9407 is now enforced, overall broken state.

More like a draft PR I've confirmed to seem to work, if you have a better way to fix, let me know.
This works by not counting the same SchedId twice towards the limit.
@magik6k @rjan90 

While as seen in below discussion the patch isnt exactly 100% right yet, its closer to previous behavior and atleast doesnt lead to not working at all / stuck scheduler.

### DIFF
--- a/storage/sealer/sched_resources.go
+++ b/storage/sealer/sched_resources.go
@@ -3,6 +3,8 @@ package sealer
 import (
 	"sync"
 
+	"github.com/google/uuid"
+
 	"github.com/filecoin-project/lotus/storage/sealer/sealtasks"
 	"github.com/filecoin-project/lotus/storage/sealer/storiface"
 )
@@ -20,7 +22,7 @@ type ActiveResources struct {
 }
 
 type taskCounter struct {
-	taskCounters map[sealtasks.SealTaskType]int
+	taskCounters map[sealtasks.SealTaskType]map[uuid.UUID]bool
 
 	// this lock is technically redundant, as ActiveResources is always accessed
 	// with the worker lock, but let's not panic if we ever change that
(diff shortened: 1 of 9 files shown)


storage/sealer/sched_resources.go
```diff
@@ -29,43 +31,52 @@ type taskCounter struct {
...
-	tc.taskCounters[tt]++
+	tc.getUnlocked(tt)[schedID] = true
 }
 
-func (tc *taskCounter) Free(tt sealtasks.SealTaskType) {
+func (tc *taskCounter) Free(tt sealtasks.SealTaskType, schedID uuid.UUID) {
 	tc.lk.Lock()
 	defer tc.lk.Unlock()
-	tc.taskCounters[tt]--
+	delete(tc.getUnlocked(tt), schedID)
```
magik6k, line 47: So I believe this has a bug which makes it not work as we'd expect:
* First we call .Add for preparing
* Then we call .Add on active in .withResources
* In .withResources we call .Free on preparing
* Now the task runs - and we don't count the resources
* Then we call .Free on preparing

This can be fixed by swapping `map[sealtasks.SealTaskType]map[uuid.UUID]bool` for `map[sealtasks.SealTaskType]map[uuid.UUID]int` and count how many times Add/Free was called.

magik6k (commented):
Great catch!

I believe this is not counting quite correctly, but the fix looks good otherwise.

storage/sealer/sched_resources.go
steffengy: Yeah its allowing more than it should as-is.
Trying that and counting +1/-1 for each Add/Free invocation leads to wrong stuck behavior again:
```
ID        Sector  Worker    Hostname    Task  State        Time
00000000  23      4154622e  x  GET   assigned(1)  21.2s
00000000  24      4154622e  x  GET   assigned(2)  21.2s
```
Thats interesting, since only reasonable idea I have is that there's an invocation missing somewhere.

steffengy: @magik6k 
Apparently the issue is the last Free for a SchedId happens after the last `WINDOW assignPreparingWork` 
(which in turn says ` not scheduling on worker  for startPreparing`).
A quick test of returning `update=true` in `schedWorker::waitForUpdates` to force a regular invocation of that leads to the "GET backlog" getting fully resolved over time as well.

steffengy: @magik6k Any update on this? _This is now broken in 1.21, 1.22 and 1.23_.

storage/sealer/sched_worker.go
```diff
@@ -416,7 +416,7 @@ assignLoop:
 				}
 
 				needRes := worker.Info.Resources.ResourceSpec(todo.Sector.ProofType, todo.TaskType)
-				if worker.active.CanHandleRequest(todo.SealTask(), needRes, sw.wid, "startPreparing", worker.Info) {
+				if worker.active.CanHandleRequest(todo.SchedId, todo.SealTask(), needRes, sw.wid, "startPreparing", worker.Info) {
```
magik6k, line 419: Not related to this PR, but should be
```suggestion
				if worker.active.CanHandleRequest(todo.SchedId, todo.SealTask(), needRes, sw.wid, "startReady", worker.Info) {
```

storage/sealer/sched_resources.go
magik6k: > only reasonable idea I have is that there's an invocation missing somewhere.

How did you implement `Free` with the counter? Did you `delete` from the map when the counter reached zero? If not the `tasks := a.taskCounters.Get(tt) [...] if len(tasks) >= needRes.MaxConcurrent` below would not work correctly as it still counted the zero enties.

steffengy: @magik6k Yeah I've accounted for that:
https://github.com/steffengy/lotus/commit/a3b7ec2447af0849b10a3e65fe67120a3f8781ff
(I think else it also wouldnt work with making "scheduler entry" more often by returning `update=true` in `schedWorker::waitForUpdates` as that just evaluates the same at a later time; so its a timing/ordering issue last Free happening after the next scheduler entry for startPreparing and stuck there until next scheduling happening)

TippyFlitsUK:
The Lotus Team is actively working on this issue and will provide an update ASAP! 🙏

storage/sealer/sched_resources.go
magik6k: I've spent some time playing with it (in https://github.com/filecoin-project/lotus/pull/10850), and I think I found the issue with integer counters - https://github.com/filecoin-project/lotus/pull/10850/commits/240fc6d3b905f69ec61bebcfbe18c3f1cd7686a6

Would be great if you could see if it works for you

steffengy: Great! Yeah in a quick test that looks working and fix-wise it seems plausible that that was the issue.
Closing this PR in favor of yours.

PR #10637, opened 2023-04-07
fix: cap the message gas limit at the block gas limit

Stebalien:
## Related Issues

May fix @shrenujbansal's issue.

## Proposed Changes

Technically, if a message is near the block gas limit, this method could over-estimate past the block gas limit. Instead, cap at the block gas limit.


magik6k (approved):
Oh that's a bug

PR #10646, opened 2023-04-10
fix: check for nil bcastDict

Stebalien:
Also hold the lock when checking the length of the blocks in the bcastDict.


ZenGround0 (approved):
LGTM

PR #10648, opened 2023-04-11
fix:mpool: prune excess messages before selection

vyzo:
We have observed some very high buildup in the mpool, which cannot be cleared fast enough due to pruning cooldown.
This apparently results in messages selection taking too long (10-20s).

This attempts to resolve the issue by eagerly pruning (regardless of cooldown) before selecting, when we have excess messages.


vyzo:
yes  we have a separate fix for that.

This is to ensure you can't ever select in an inordinately large set.

PR #10651, opened 2023-04-11
feat: shed: Add --below-basefee to mpool clear

magik6k:
## Additional Info
Not tested at all

## Checklist

- [ ] CI is green


chain/messagepool/messagepool.go
```diff
@@ -448,6 +448,36 @@ func New(ctx context.Context, api Provider, ds dtypes.MetadataDS, us stmgr.Upgra
...
 
+func (ms *msgSet) removeMessagesBelowBaseFee(baseFee types.BigInt) bool {
+	noncesToRemove := make([]uint64, 0)
+
+	// Find nonces of messages below the base fee and mark the first nonce to remove
+	var firstNonceToRemove uint64 = math.MaxUint64
+	for nonce, msg := range ms.msgs {
+		if types.BigCmp(msg.Message.GasFeeCap, baseFee) < 0 {
+			noncesToRemove = append(noncesToRemove, nonce)
+			if nonce > ms.nextNonce && firstNonceToRemove == math.MaxUint64 {
```
arajasek, line 459: I'm not sure I understand the `nonce > ms.nextNonce` check. Why do we only set `firstNonceToRemove` in this case?

Depending on what the intended logic we want for this, we _probably_ just want to find the lowest nonce that has `GasFeeCap < baseFee`, and then drop all subsequent messages?

chain/messagepool/messagepool.go
```diff
@@ -1584,6 +1614,11 @@ func (mp *MessagePool) Clear(ctx context.Context, local bool) {
 
 			if ok {
 				for _, m := range mset.msgs {
+					if types.BigCmp(co.MinBaseFee, big.Zero()) > 0 &&
+						types.BigCmp(m.Message.GasFeeCap, co.MinBaseFee) < 0 {
```
arajasek, line 1618: We _don't_ delete messages with `GasFeeCap < MinBaseFee`? Because?

arajasek (commented):
I'm not sure I understand the core logic here, simple as it is. There's arguably a few different ways we can implement this, and we need to worry about "holes" in our mSets (I think). It looks like you're trying to handle that, but I'm not sure I understand how.

chain/messagepool/messagepool.go
magik6k: This.. maay be in the wrong order

magik6k: That is we should remove messages below the specified minimum basefee

chain/messagepool/messagepool.go
magik6k: That would be the much less dumb way to do that, assuming that message sets are sorted we could just truncate.

arajasek: They're not, they're a map.

chain/messagepool/messagepool.go
arajasek: That makes more sense.

arajasek (commented):
I'm gonna say let's not merge this unless a pressing need emerges? The mpool really should be doing this kinda thinking itself when necessary.

(Closing for now, but feel free to yell)

PR #10652, opened 2023-04-11
feat: pubsub: treat ErrGasFeeCapTooLow as ignore, not reject

arajasek:
## Related Issues

Currently, ErrGasFeeCapTooLow results in a ValidationReject, that lowers peer scores. I _think_ this is too extreme, as it can happen for somewhat benign reasons such as the base fee changing, a node briefly falling out of sync etc. 

## Proposed Changes

I'm proposing treating this as an Ignore, similar to how we treat errors around message nonces, not-enough-funds, etc. 

I _think_ we should do this, because otherwise we risk dropping good peers when _we_ are out of sync. However, this might require some careful thought, so please feel free to disagree.

## Checklist

- [ ] CI is green


ZenGround0 (approved):
I don't know this part of the system particularly well but this change and the reasoning behind it SGTM

PR #10653, opened 2023-04-11
Wire in profile logging for MpoolSelect

vyzo:
adds a fancy context option to track and log timing information for profiling things, specifically mpool select.

We can add more spans that are profiled, but this should be a good enough start.

### DIFF
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -39,6 +39,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/vm"
 	"github.com/filecoin-project/lotus/journal"
+	"github.com/filecoin-project/lotus/lib/prof"
 	"github.com/filecoin-project/lotus/metrics"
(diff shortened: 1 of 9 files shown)


magik6k (commented):
Shouldn’t we use the opencensus tracer like everywhere else? (that allows you to connect jaeger tracer and passes traces over jsonrpc)

vyzo:
We can, but do we care?

magik6k:
I believe it's better to have one profiling framework instead of two. Usage is really easy:
* In code just put span on context: https://github.com/filecoin-project/lotus/blob/a503a0edaa66aa05e344749e0e6b1f43c80d172f/chain/stmgr/execute.go#L16-L17 (this connects to MetricsCtx from DI)
* Use with the quite nice jaeger tracer: https://github.com/filecoin-project/lotus/blob/master/documentation/en/jaeger-tracing.md

The main advantage is that with that you will be able to dive deeper into calls made by mpool, because e.g. VM calls also generate those traces

arajasek:
Reverting to draft until @vyzo can action on @magik6k's suggestions

PR #10654, opened 2023-04-11
feat:metrics:mempool - Add MessagePending metric

GlacierWalrus:
## Proposed Changes

Add a new MessagePending metric to allow monitoring Pending Messages
## Checklist

- [ ] CI is green

### DIFF
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math"
 	stdbig "math/big"
+	"os"
 	"sort"
 	"sync"
 	"time"
@@ -312,7 +313,7 @@ func (ms *msgSet) add(ctx context.Context, m *types.SignedMessage, mp *MessagePo
 	ms.msgs[m.Message.Nonce] = m
 	ms.requiredFunds.Add(ms.requiredFunds, m.Message.RequiredFunds().Int)
 	// ms.requiredFunds.Add(ms.requiredFunds, m.Message.Value.Int)
-	if mp.cfg.OptionalMetrics.MpoolSize {
+	if mp.mpoolMonitoringEnabled() {
 		stats.Record(ctx, metrics.MessagePending.M(int64(int64(mp.getTotalPendingMessages()))))
 	}
 
@@ -378,6 +379,10 @@ func (ms *msgSet) toSlice() []*types.SignedMessage {
 	})
 
 	return set
+}
+
+func (mp *MessagePool) mpoolMonitoringEnabled() bool {
+	return mp.getConfig().OptionalMetrics.MpoolSize || os.Getenv("LOTUS_MPOOL_DETAILED_METRICS_ENABLED") == "true"
 }
 
 func New(ctx context.Context, api Provider, ds dtypes.MetadataDS, us stmgr.UpgradeSchedule, netName dtypes.NetworkName, j journal.Journal) (*MessagePool, error) {
@@ -526,7 +531,7 @@ func (mp *MessagePool) setPendingMset(ctx context.Context, addr address.Address,
 	}
 
 	mp.pending[ra] = ms
-	if mp.cfg.OptionalMetrics.MpoolSize {
+	if mp.mpoolMonitoringEnabled() {
 		stats.Record(ctx, metrics.MessagePending.M(int64(int64(mp.getTotalPendingMessages()))))
 	}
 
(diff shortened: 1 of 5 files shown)


ZenGround0:
The CI test failure `FAIL: chain/index TestBasicMsgIndex` appears to be an actual failure and not a flake.  Either the test needs to be updated or an underlying bug needs fixing.  Didn't investigate enough to be able to say which.

arajasek:
@ZenGround0 Rebased onto master to fix test issues (you and I fixed this a while ago).

chain/messagepool/messagepool.go
```diff
@@ -303,6 +304,7 @@ func (ms *msgSet) add(m *types.SignedMessage, mp *MessagePool, strict, untrusted
 	ms.msgs[m.Message.Nonce] = m
 	ms.requiredFunds.Add(ms.requiredFunds, m.Message.RequiredFunds().Int)
 	// ms.requiredFunds.Add(ms.requiredFunds, m.Message.Value.Int)
+	stats.Record(ctx, metrics.MessagePending.M(int64(len(mp.pending))))
```
arajasek, line 307: This is gonna record the number of entries in `mp.pending`, which is the total number of senders that currently have messages in our mpool. 

Is that what we want to record here, or do we want the total number of messages in our mpool?

arajasek (commented):
Nothing that should break things here, but I'm unsure whether this is actually the metric we want.

chain/messagepool/messagepool.go
GlacierWalrus: Total number of messages! What do I need to record that instead?

arajasek: @GlacierWalrus Poking around the code, it looks like `allPending()` seems to do what we want here. 

This is NOT gonna be a fast method, though, so we either want to make this metric opt-in (if it isn't already), or only capture it infrequently.

GlacierWalrus: @arajasek I've made updates as you suggested, but I'm having trouble working out how to configure mpool. I would be grateful if you could advise here. Configuring seems to work with `lotus mpool config '{...}'` but it should be configurable from the lotus config file/env vars.

GlacierWalrus:
> An alternative (perhaps better) way to do this would be to spin up a statsWorker on mpool startup (assuming the config is set), and have that record this value every 5 minutes or so.

I'm going to have to disagree with this approach as I believe it's likely to obscure issues rather than expose them.

5 minutes will provide a very low data resolution, I'm guessing the MessagePending could in theory increase significantly in size, an then reduce back to a normal value multiple times with a 5 minute period, which might be completely unreported.

I also believe it's bad practice as metrics endpoints are considered to be contemporary- when you scrape it, you should be seeing up to date values.

Thanks for your suggestion about configuring the mpool directly from an env var

rjan90:
Was the wanted metric here to have `Number of messages in the mpool`? If so, this PR can probably be closed. As we have that metric in master now:

```
# HELP lotus_mpool_message_count Number of messages in the mpool
# TYPE lotus_mpool_message_count gauge
lotus_mpool_message_count 180
```

GlacierWalrus:
@rjan90 Yes indeed it was about counting pending messages in the mpool.

I had a quick look at master, and there does seem to be some metrics being recorded, though it doesn't match me expectation of where/how it's recorded, and doesn't seem to have a config control which @arajasek requested in this thread.

Feel free to do as you wish with this PR, @jennijuju requested we lend a hand with this but the PR just stalled and there is _an_ implementation of metrics in place so just make sure the cat lady is happy?

rjan90:
> I had a quick look at master, and there does seem to be some metrics being recorded, though it doesn't match me expectation of where/how it's recorded, and doesn't seem to have a config control which @arajasek requested in this thread.

Yeah, there is no config control as requested in this PR. Instead its recording the `currentSize` that is being incremented in addLocked: https://github.com/filecoin-project/lotus/blob/1ca071ef7773db72e10db2fc6b6c38c5b8dbe725/chain/messagepool/messagepool.go#L1002-L1026
And decremented in the remove function:
https://github.com/filecoin-project/lotus/blob/1ca071ef7773db72e10db2fc6b6c38c5b8dbe725/chain/messagepool/messagepool.go#L1195-L1222

I will confirm with Aayush and Jen, but think we have the metric wanted with the above recording.

GlacierWalrus:
I don't want to spend a lot of time looking into this as it's somewhat beyond my concerns these days, but the reason that the code in master didn't match my expectations is because I remember there were quite a few points in the code where the mempool could be updated (which led to the request of having it config controlled), and [as I mentioned earlier in the thread](https://github.com/filecoin-project/lotus/pull/10654#issuecomment-1547866014), I was reluctant to implement simpler changes which caused the metrics to misrepresent the state of the mempool.

Hopefully it's due to my lack of deep-knowledge of the codebase, or due to the code changing such that the stats recording is fully correct.

Anyway, I don't think I'm willing to spend time on rebasing this anyway, so let's go ahead and close this, if you're happy, I'm happy :)

PR #10666, opened 2023-04-13
chore:releasepolish v1.22 release

ZenGround0:
## Additional Info

RC3
- Register v11 actors so state read-state commands will work among other things
- Update calibration to happen 4-20 16:30 UTC
- Add shed inspection commands of SI and market cron used to confirm butterfly state and useful for calibration 

## Checklist

- [ ] CI is green

### DIFF
--- a/build/bootstrap/butterflynet.pi
+++ b/build/bootstrap/butterflynet.pi
@@ -1,2 +1,2 @@
-/dns4/bootstrap-0.butterfly.fildev.network/tcp/1347/p2p/12D3KooWF31bhPqjeAxNSCvWTLGq84iuMhKvbgK31rgTXfRjgByn
-/dns4/bootstrap-1.butterfly.fildev.network/tcp/1347/p2p/12D3KooWEY2LteCs2tYLmAWZ6U9eHzgnvmVsd5pSrKLg8nn3A6nh
+/dns4/bootstrap-0.butterfly.fildev.network/tcp/1347/p2p/12D3KooWEwwFTMAi8rMnVX4fzih9pW5xZgTMbcacAFPinaT1g4wo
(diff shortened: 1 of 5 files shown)


shrenujbansal (commented):
Changes look good for the most part. Could you fix gen and lint checks?
Also could you post an example output of the lotus-shed cmds you just added?

ZenGround0:
## cron state command

Assuming at tipset height 300 there are three deals, dealid 1 scheduled for cron process at epoch 1000, dealid 2 and 3 scheduled for 1001 we get:
```
lotus-shed market cron-state --tipset=@300
{"1000":[1], "1001":[2,3]}
``` 
tipset flag is optional defaulting to head

## sector info command 
```
./lotus-shed miner sectorinfo --tipset=@296 t01012 64                                       
8f184007d82a5829000182e20381e80220f484322806367e97188763a81fda5618675f3bdd47fa5ce7a97c065a3c8f175e801901201a0017a4fc404048003782dac0000000480023d24224dd85674800552adf417114230040f6f5
```

I went with raw cbor so that this will work the same across all actors versions + using cbor.me is pretty easy for the simple struct.  It is also nice to work with raw bytes because we immediately get size of sector info.

shrenujbansal:
> I went with raw cbor so that this will work the same across all actors versions + using cbor.me is pretty easy for the simple struct. It is also nice to work with raw bytes because we immediately get size of sector info.

I would actually convert this into something consumable by human eyes. Cbor output is not very useful to look at without having to insert into another tool

ZenGround0:
> I would actually convert this into something consumable by human eyes. Cbor output is not very useful to look at without having to insert into another tool

I wouldn't land it like this into `lotus` proper but this is lotus-shed so think functional tools tools for getting work done that are rough around the edges.  Using raw bytes makes handling all past and future network versions trivially easy.  This way I was able to use the tool to inspect different sector info schemas at nv18 and nv19 to complete butterfly testing.  The alternative is a lot of error handling and updating on format upgrade which will realistically always go stale and make this break and unused.

shrenujbansal (approved):

PR #10671, opened 2023-04-14
chore: update go-libp2p to v0.27.3

marten-seemann:
## Checklist

- [ ] CI is green


cli/net.go
```diff
@@ -445,9 +445,6 @@ var NetReachability = &cli.Command{
 		}
 
 		fmt.Println("AutoNAT status: ", i.Reachability.String())
-		if i.PublicAddr != "" {
-			fmt.Println("Public address: ", i.PublicAddr)
-		}
```
marten-seemann, line 450: I removed this log (?) output. It was misleading to begin with, a libp2p node doesn't only have a single public address. The correct way to obtain the addresses is to use the `Addrs` method on the host.

node/impl/net/net.go
```diff
@@ -143,19 +143,7 @@ func (a *NetAPI) NetAutoNatStatus(ctx context.Context) (i api.NatInfo, err error
 		}, nil
 	}
 
-	var maddr string
-	if autonat.Status() == network.ReachabilityPublic {
-		pa, err := autonat.PublicAddr()
```
marten-seemann, line 148: We removed the `PublicAddr` method from AutoNAT.

api/api_net.go
```diff
@@ -73,5 +73,4 @@ type CommonNet interface {
 
 type NatInfo struct {
 	Reachability network.Reachability
-	PublicAddr   string
```
magik6k, line 76: Can we change this to an array, and populate this from Addrs? We have a bunch of tools / docs which want this API / CLI (mostly there some docs that refer to the Public address output from cli)

jacobheun:
What's the state of this? I was hoping to get go-fil-markets updated but it needs this update for libp2p.

magik6k (approved):

PR #10674, opened 2023-04-16
feat: market: Register a custom shard indexer for the dagstore

willscott:
This change will result in lotus parsing deals for local indexing according to https://github.com/filecoin-project/FIPs/discussions/512

## Related Issues
This is a continuation of https://github.com/filecoin-project/dagstore/pull/154 - This piece of the logic was seen as filecoin-specific rather than part of common dagstore functionality. A parallel to this logic is expected to be ported also to boost, but there remain confguration in the use of the dagstore where this is the common locations where this indexing functionality can be registered.

## Proposed Changes
* Register a `ShardIndexer` for the dagstore - such that when a shard is indexed it is considered both for indexing with a data segment index and the existing  direct-car indexing.

## Additional Info
cc @Kubuxu @dirkmc 

## Checklist

- [x] CI is green


willscott:
This is the current location where the dagstore / metadata occurs. If we're supporting legacy configurations, this is the 'compatible' location for such code to live.

currently we're only targeting newer boost installations, so the active branch is https://github.com/filecoin-project/boost/pull/1495

rjan90:

PR #10680, opened 2023-04-17
feat: badger: add a has check before writing to reduce duplicates

Stebalien:
## Related Issues

fixes #10457

## Proposed Changes

Check if we have blocks before writing them (in badger, only). By default, badger just appends the block which will grow the size of the datastore until garbage collected. Worse, badger garbage collection is expensive and not fully effective.

## Additional Info

Downsides:

1. This code is racy. I.e., if we write the same block multiple times in parallel, we'll end up with multiple versions. We could prevent this with transactions, but that has other performance implications.
2. This adds a serial read before every write, reducing performance.

However, the average block sync time (on my machine) is 10.5s before and after this patch, so it shouldn't matter.

## Checklist

- [x] CI is green

### DIFF
--- a/blockstore/badger/blockstore.go
+++ b/blockstore/badger/blockstore.go
@@ -732,6 +732,20 @@ func (b *Blockstore) Put(ctx context.Context, block blocks.Block) error {
 	}
 
 	put := func(db *badger.DB) error {
+		// Check if we have it before writing it.
+		switch err := db.View(func(txn *badger.Txn) error {
+			_, err := txn.Get(k)
+			return err
+		}); err {
+		case badger.ErrKeyNotFound:
+		case nil:
+			// Already exists, skip the put.
+			return nil
+		default:
+			return err
+		}
+
+		// Then write it.
 		err := db.Update(func(txn *badger.Txn) error {
 			return txn.Set(k, block.RawData())
 		})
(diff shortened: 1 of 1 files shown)


Stebalien:
Looks good. Average 10.5 blocktimes both ways.

Stebalien:
Really, I'm seeing 10.2s block times with this patch, so it may even be speeding things up a bit by not writing duplicate state.

vyzo (commented):
Can we update and return a sentinel error if exists?
It might be more efficient, as it saves a round trip in badger.

vyzo:
Not an error, I was suggesting using `badger.Update` and doing the occurs check in there, doing bothing and possibly returning a sentinel error if it exists.


Basically move the occurs check into the `Update` that follows this code.

blockstore/badger/blockstore.go
```diff
@@ -732,6 +732,20 @@ func (b *Blockstore) Put(ctx context.Context, block blocks.Block) error {
...
+		}); err {
+		case badger.ErrKeyNotFound:
+		case nil:
+			// Already exists, skip the put.
+			return nil
+		default:
+			return err
+		}
+
+		// Then write it.
```
raulk, line 748: Should be relatively more efficient to do this check inside `db.Update` below and discarding the transaction if the key exists, instead of starting a new transaction. Although in practice it probably doesn't matter much.

raulk (approved):

Stebalien:
I've been running this for several days now and haven't had any issues.

magik6k:
* why is this check needed, isn’t vm Flush code only saving new cids? (I guess it’s dag diff, not blockstore based?)
* Can we cache the has checks?
* Can we test this on a large full node, where the index doesn’t fit in memory?

ankit-gautam23 (approved):
LGTM

magik6k (approved):
Probably fine to merge and see how it performs.

Ideally we'd check on a full-node first (can give you access to one if you don't have already)

Stebalien:
> Ideally we'd check on a full-node first (can give you access to one if you don't have already)

I've tested it on a snapshot node, but not a node with significant historical state. I don't _think_ this will slow that node down, but I'm happy to test if you give me access.

Issue #10672, opened 2023-04-14
Chain `ExportRangeInternal` returns an EOF error when exporting stateroots


ognots:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to lotus.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon:  1.21.0-dev+mainnet+git.3f74840b6+api1.5.0
Local: lotus version 1.21.0-dev+mainnet+git.3f74840b6
```

### Repro Steps

```lotus chain export-range --head @2770000 --tail @2768000  --messages=true --receipts=true --stateroots=true```

### Describe the Bug

When attempting to call [ExportRangeInternal](https://github.com/filecoin-project/lotus/blob/master/node/impl/full/chain.go#L588-L634) with messages, receipts and stateroots, an ambiguous `EOF` is returned.
1. I would expect the error message to be helpful
2. I would expect the snapshot to complete on a lotus node containing chain data for the tipset range

The error appears to be in the walker, judging from log output.

This error does not occur if one sets `--stateroots=false`

### Logging Information

```text
{"level":"error","ts":"2023-04-14T13:11:35.729-0400","logger":"chainstore","caller":"store/snapshot.go:343","msg":"export workers finished with error","error":"EOF"}
Apr 14 13:11:35 stongops-worker-2 lotus[2335532]: {"level":"error","ts":"2023-04-14T13:11:35.817-0400","logger":"chainstore","caller":"store/snapshot.go:582","msg":"walker scheduler","error":"EOF"}
[...]

PR #10684, opened 2023-04-18
chainstore: Fix raw blocks getting scanned for links during snapshots

hsanjuan:
We have to save raw blocks to the snapshot, but we should not be scanning them for additional links as if they were CBOR blocks.

This cleans the logic a bit (we were checking that the parent was a CBOR block before queueing up the children, but then scanning the children... it was weird).

Additionally, more verbose logging is added for the next time ScanForLinks fails (currently very little info was given).

Our ScanForLinks callback should only enqueue CBOR for further processing.

## Related Issues

#10672 

## Proposed Changes

Fixes raw blocks being scanned for links as if they were CBOR.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/store/snapshot.go
+++ b/chain/store/snapshot.go
@@ -371,6 +371,9 @@ func (s *walkScheduler) enqueueIfNew(task walkTask) {
 		//log.Infow("ignored", "cid", todo.c.String())
 		return
 	}
+
+	// This lets through RAW and CBOR blocks, the only two types that we
+	// end up writing to the exported CAR.
 	if task.c.Prefix().Codec != cid.Raw && task.c.Prefix().Codec != cid.DagCBOR {
 		//log.Infow("ignored", "cid", todo.c.String())
 		return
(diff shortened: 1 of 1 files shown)


hsanjuan:
@ognots do you have a way to test this on the problematic epoch? (2764810)

ognots:
> @ognots do you have a way to test this on the problematic epoch? (2764810)

yeah, I can find an older snapshot to restore from, and I'll test it.
so far testing has been :+1: on a recent epoch

chain/store/snapshot.go
```diff
@@ -442,10 +445,19 @@ func (s *walkScheduler) processTask(t walkTask, workerN int) error {
 		b: blk,
 	}
 
+	// We exported the ipld block. If it wasn't a CBOR block, there's nothing
+	// else to do and we can bail out early as it won't have any links
+	// etc.
+	if t.c.Prefix().Codec != cid.DagCBOR || t.c.Prefix().MhType == mh.IDENTITY {
```
Stebalien, line 451: While it likely doesn't matter (at the moment, we _should_ scan for links in "identity" blocks. In theory, I can inline dag-cbor into an identity hash.

Stebalien (commented):
👍

chain/store/snapshot.go
hsanjuan: @Stebalien I never knew the reasons to skip identity hashes, it was something inherited from @frrist code. I can't say if there's an instance of an identity hash in the whole chain history (? genesis or something?)

Stebalien: Hm. So, we shouldn't include blocks for identity hashes, but we should technically be traversing them and including their children, if any.

On the other hand, I can guarantee that this won't make a difference in practice (at the moment, at least). So we should be good to merge this as-os.

Stebalien (approved):

chain/store/snapshot.go
hsanjuan: ok, happy to send follow-up in a few days.

frrist: fwiw I inherited it from: https://github.com/filecoin-project/lotus/blob/e945c0d6f25f490323ec542d535f584022a20a36/chain/store/snapshot.go#L668

PR #10690, opened 2023-04-18
fix: remove pointless panic

Stebalien:
Technically, the block validator caught this panic. But it's pointless because we have a _real_ mechanism to return the validation reason, which we should have been using.

In general, panicing like this is a very bad idea because it's non-obvious and, in this case, completely undocumented.


vyzo (approved):
huh, that looks pointless indeed.

Maybe log something?

magik6k (approved):

PR #10691, opened 2023-04-18
feat: sync: validate (early) that blocks fall within range

Stebalien:
## Proposed Changes

This will reject blocks in pubsub validation if they're either:

1. Too far into the future (5 blocks beyond the expected head).
2. Too far into the past (before finality).

Specifically:

1. We were previously rejecting future blocks in the sync logic, but not in pubsub itself.
2. We never used to check if a block was too _old_.

Motivation: Blocks that are too new/too old can cause us to perform quite a bit of unnecessary work.

We should carefully consider that second check as we may need to tweak it (e.g., to handle network issues?). We may want to base this check on the latest synced head, not the _expected_ epoch.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/consensus/iface.go
+++ b/chain/consensus/iface.go
@@ -32,9 +32,10 @@ type Consensus interface {
 	// the block (signature verifications, VRF checks, message validity, etc.)
 	ValidateBlock(ctx context.Context, b *types.FullBlock) (err error)
 
-	// IsEpochBeyondCurrMax is used to configure the fork rules for longest-chain
-	// consensus protocols.
-	IsEpochBeyondCurrMax(epoch abi.ChainEpoch) bool
+	// IsEpochInConsensusRange returns true if the epoch is "in range" for consensus. That is:
+	// - It's not before finality.
+	// - It's not too far in the future.
+	IsEpochInConsensusRange(epoch abi.ChainEpoch) bool
 
 	// CreateBlock implements all the logic required to propose and assemble a new Filecoin block.
 	//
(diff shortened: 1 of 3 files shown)


Stebalien:
Well, this definitely broke something.

Stebalien:
> We should carefully consider that second check as we may need to tweak it (e.g., to handle network issues?). We may want to base this check on the latest synced head, not the expected epoch.

Ok, I needed this, but I'm concerned I'm misunderstanding something here. As far as I know, the "catch up" code path shouldn't end up calling `IsEpochInConsensusRange`.

chain/consensus/iface.go
```diff
@@ -71,6 +72,11 @@ func ValidateBlockPubsub(ctx context.Context, cns Consensus, self bool, msg *pub
 		return pubsub.ValidationReject, what
 	}
 
+	if !cns.IsEpochInConsensusRange(blk.Header.Height) {
+		log.Warnf("received block outside of consensus range (%d)", blk.Header.Height)
+		return pubsub.ValidationReject, "invalid_block_height"
```
vyzo, line 77: maybe ValidationIgnore?

Can this happen by accident or normal operation? If thats the case, it should br ValidationIgnore.

magik6k (approved):
Not entirely sure about Ignore/Reject, I guess if we want to be conservative, ignore is better, but I don't have a strong opinion.

Other parts LGTM.

vyzo:
I think it is better to be conservative and  Ignore here, otherwise it might have adverse effects with the score penalties.

Basically we should only do Reject for obviously bad behaviours, not things that can happen naturally.

Stebalien:
Ok, this should be good to go. It ignores blocks before finality with respect to our current head and 5 blocks after the expected head.

vyzo (approved):

Issue #10597, opened 2023-03-29
Messagepool lower priority optimizations


snissn:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [X] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

related to issue #10561 `perf: message pool: change locks to RWMutexes for performance #10561` are the following issues that were identified that are to be added in a future pull request. The motivation for the two separate PRs is to isolate the threading changes as much as possible for correctness reasons.

1. https://github.com/filecoin-project/lotus/pull/10561/files#r1148322934 in `chain/messagepool/check.go` pre-allocate the size of the array msgs array using `msgs = make([]*types.Message, 0, len(mset.msgs))`
2. https://github.com/filecoin-project/lotus/pull/10561/files#r1152024072 remove the requirement of `mp.keyCache.Add(ka, ka)` by checking that the address is a robust address
3. https://github.com/filecoin-project/lotus/pull/10561/files#r1150875872 chain/messagepool/repub.go - optimize the Lock timing and confirm if all tests pass with a read lock
4. https://github.com/filecoin-project/lotus/pull/10561/files#r1150137317 chain/messagepool/selection.go change to read lock

### Describe the solution you'd like

related to issue #10561 `perf: message pool: change locks to RWMutexes for performance #10561` are the following issues that were identified that are to be added in a future pull request. The motivation for the two separate PRs is to isolate the threading changes as much as possible for correctness reasons.

1. https://github.com/filecoin-project/lotus/pull/10561/files#r1148322934 in `chain/messagepool/check.go` pre-allocate the size of the array msgs array using `msgs = make([]*types.Message, 0, len(mset.msgs))`
2. https://github.com/filecoin-project/lotus/pull/10561/files#r1152024072 remove the requirement of `mp.keyCache.Add(ka, ka)` by checking that the address is a robust address
[...]

PR #10693, opened 2023-04-19
perf: mempool: lower priority optimizations

snissn:
## Related Issues
https://github.com/filecoin-project/lotus/issues/10597

## Proposed Changes

 1. https://github.com/filecoin-project/lotus/pull/10561/files#r1148322934 in chain/messagepool/check.go pre-allocate the size of the array msgs array using msgs = make([]*types.Message, 0, len(mset.msgs))
 2. https://github.com/filecoin-project/lotus/pull/10561/files#r1152024072 remove the requirement of mp.keyCache.Add(ka, ka) by checking that the address is a robust address
 3. https://github.com/filecoin-project/lotus/pull/10561/files#r1150875872 chain/messagepool/repub.go - optimize the Lock timing and confirm if all tests pass with a read lock -- also ensure that mp.curTsLk.RLock() is not held during mp.api.ChainComputeBaseFee
 4. https://github.com/filecoin-project/lotus/pull/10561/files#r1150137317 chain/messagepool/selection.go change to read lock
 5. in [chain/messagepool/check.go](https://github.com/filecoin-project/lotus/pull/10561/files/41ea5a6f638ca5108a51f70a4cef85d9b98802b7#diff-ed714997f1033a2434cc1cdfd1f231667eb107c47a885b4b8761619445ea218c) pending has a giant warning saying do NOT use it directly -- we should probably be calling getPendingMset here for address resolution?

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/messagepool/check.go
+++ b/chain/messagepool/check.go
@@ -35,6 +35,7 @@ func (mp *MessagePool) CheckPendingMessages(ctx context.Context, from address.Ad
 	mp.lk.RLock()
 	mset, ok := mp.pending[from]
 	if ok {
+		msgs = make([]*types.Message, 0, len(mset.msgs))
 		for _, sm := range mset.msgs {
 			msgs = append(msgs, &sm.Message)
 		}
--- a/chain/messagepool/check.go
+++ b/chain/messagepool/check.go
@@ -33,7 +33,11 @@ func (mp *MessagePool) CheckMessages(ctx context.Context, protos []*api.MessageP
 func (mp *MessagePool) CheckPendingMessages(ctx context.Context, from address.Address) ([][]api.MessageCheckStatus, error) {
 	var msgs []*types.Message
 	mp.lk.RLock()
-	mset, ok := mp.pending[from]
+	mset, ok, err := mp.getPendingMset(ctx, from)
+	if err != nil {
+		log.Warnf("errored while getting pending mset: %w", err)
+		return nil, err
+	}
 	if ok {
 		msgs = make([]*types.Message, 0, len(mset.msgs))
 		for _, sm := range mset.msgs {
@@ -65,7 +69,11 @@ func (mp *MessagePool) CheckReplaceMessages(ctx context.Context, replace []*type
 		if !ok {
 			mmap = make(map[uint64]*types.Message)
 			msgMap[m.From] = mmap
-			mset, ok := mp.pending[m.From]
+			mset, ok, err := mp.getPendingMset(ctx, m.From)
+			if err != nil {
+				log.Warnf("errored while getting pending mset: %w", err)
+				return nil, err
+			}
 			if ok {
 				count += len(mset.msgs)
 				for _, sm := range mset.msgs {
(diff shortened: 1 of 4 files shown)


chain/messagepool/messagepool.go
```diff
@@ -474,6 +474,15 @@ func (mp *MessagePool) TryForEachPendingMessage(f func(cid.Cid) error) error {
...
 
 func (mp *MessagePool) resolveToKey(ctx context.Context, addr address.Address) (address.Address, error) {
+	//if addr is not an ID addr, then it is already resolved to a key
+	if addr.Protocol() != address.ID {
+		return addr, nil
+	}
+	return mp.resolveToKeyFromID(ctx, addr)
+}
+
+func (mp *MessagePool) resolveToKeyFromID(ctx context.Context, addr address.Address) (address.Address, error) {
```
arajasek, line 484: I don't think we need a new method here, just add the "is not ID" check to `resolveToKey` and leave it as before?

arajasek (commented):
Will do a second pass tomorrow just to squint at the locks, but LGTM.

chain/messagepool/messagepool.go
snissn: I prefer the way I have it. I think it's easier to read! Definitely happy to defer to your suggestion if you have a preference

chain/messagepool/check.go
```diff
@@ -33,8 +33,13 @@ func (mp *MessagePool) CheckMessages(ctx context.Context, protos []*api.MessageP
 func (mp *MessagePool) CheckPendingMessages(ctx context.Context, from address.Address) ([][]api.MessageCheckStatus, error) {
 	var msgs []*types.Message
 	mp.lk.RLock()
-	mset, ok := mp.pending[from]
+	mset, ok, err := mp.getPendingMset(ctx, from)
+	if err != nil {
+		log.Warnf("errored while getting pending mset: %w", err)
+		return nil, err
```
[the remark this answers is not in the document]
arajasek: (it makes debugging a lot easier when things go wrong)

chain/messagepool/messagepool.go
arajasek: No strong preference!

arajasek (approved):

chain/messagepool/messagepool.go
arajasek, line 484: @Stebalien In what cases can an actor have both an f4 and an f2 address? Is that just EVM actors (not including EthAccounts)?

I wanna make sure you can never have an actor have both an f4 and an f2 address AND serving as a top-level sender, because that would make things funky here.

(Not really related to changes in this PR)

Stebalien: Just EVM actors, at least for now. Account abstraction may someday change this, but that's a whole can of worms.

Stebalien (approved):
This looks correct. I'm slightly concerned about the dance with `mp.republished` (we're assuming that nothing changes while we drop the lock), but that's existing behavior.

PR #10694, opened 2023-04-19
fix: chain: record heaviest tipset before notifying

Stebalien:
Clearly this hasn't caused any issues, but I'm pretty sure we should be updating the current head _before_ notifying about it.


magik6k (approved):
IDK if this change makes any practical difference, but I can see how it can feel slightly more correct.

Stebalien:
Yeah, it shouldn't make any practical difference. It just makes me nervous.

PR #10695, opened 2023-04-19
multi minning

siriusyim:
## Related Issues
none

## Proposed Changes
In each epoch, start multiple threads for mining. Whenever a fork occurs, start a mining thread. When submitting a block, select the heaviest block in this epoch for submission

## Additional Info
none

## Checklist

- [ ] CI is green

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -7,6 +7,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"os"
+	"sort"
 	"sync"
 	"time"
 
@@ -29,6 +30,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/gen"
 	"github.com/filecoin-project/lotus/chain/gen/slashfilter"
 	lrand "github.com/filecoin-project/lotus/chain/rand"
+	"github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/chain/types"
 	cliutil "github.com/filecoin-project/lotus/cli/util"
 	"github.com/filecoin-project/lotus/journal"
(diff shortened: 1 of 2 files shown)


magik6k (commented):
Thanks for looking into this!

I believe this can be optimized quite a bit more
* The winning ticket / winningPoSt is based on drand randomness, which shouldn't change when changing the mining base
* Message selection probably shouldn't also change
* Reading this code it looks like we will do those things multiple times, and we shouldn't have to

siriusyim:
> Thanks for looking into this!
> 
> I believe this can be optimized quite a bit more
> 
> * The winning ticket / winningPoSt is based on drand randomness, which shouldn't change when changing the mining base
> * Message selection probably shouldn't also change
> * Reading this code it looks like we will do those things multiple times, and we shouldn't have to

Good idea, it can be optimized here

Message selection may change when changing the base , because of different bases, their block messages may be different

siriusyim:
> > Thanks for looking into this!
> > I believe this can be optimized quite a bit more
> > 
> > * The winning ticket / winningPoSt is based on drand randomness, which shouldn't change when changing the mining base
> > * Message selection probably shouldn't also change
> > * Reading this code it looks like we will do those things multiple times, and we shouldn't have to
> 
> Good idea, it can be optimized here
> 
> Message selection may change when changing the base , because of different bases, their block messages may be different

Hi,@magik6k ，According to your suggestion, I improved the code, and the test-itest-batch_deal test failed, but I did not find the reason

ZenGround0:
> test-itest-batch_deal test failed,

This is a known flaky test so it should be unrelated to your change

PR #10700, opened 2023-04-19
perf: Address performance of EthGetTransactionCount 

fridrik01:
Fixes: https://github.com/filecoin-project/lotus/issues/10538

**Context**
We have observed that `EthGetTransactionCount` is one of the hotspots on Glif production notes, and we are seeing regular 10-20 second latencies when calling this rpc method. We have also been able to replicate this issue on local nodes on our devboxes on mainnet when calling this method repeatedly.

**The fix**
I tracked the high latency spikes and they were correlated when we were running ExecuteTipSet while following the chain.

To address this, we should not rely on tipset computation to get nounce and instead look at the parent tipset and then count the messages sent from the 'addr' as suggested by steb.

**Test plan**
I started and synced my lotus node on mainnet and started calling  `EthGetTransactionCount` rpc method in a loop.

The latency spikes seemed to be all gone, which I then confirmed by looking at the `http://localhost:1234/debug/metrics` for the `getStateNonce` metric (getnonce_ms)

image

Of almost 35k calls, all were within 8ms (before this fix, there were often outliers in the 5+ second range)

### DIFF
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -371,7 +371,7 @@ func (ms *msgSet) toSlice() []*types.SignedMessage {
 func New(ctx context.Context, api Provider, ds dtypes.MetadataDS, us stmgr.UpgradeSchedule, netName dtypes.NetworkName, j journal.Journal) (*MessagePool, error) {
 	cache, _ := lru.New2Q[cid.Cid, crypto.Signature](build.BlsSignatureCacheSize)
 	verifcache, _ := lru.New2Q[string, struct{}](build.VerifSigCacheSize)
-	stateNonceCache, _ := lru.New[stateNonceCacheKey, uint64](256)
+	stateNonceCache, _ := lru.New[stateNonceCacheKey, uint64](32768) // 32k * ~200 bytes = 6MB
 	keycache, _ := lru.New[address.Address, address.Address](1_000_000)
 
 	cfg, err := loadConfig(ctx, ds)
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -1078,17 +1078,17 @@ func (mp *MessagePool) getStateNonce(ctx context.Context, addr address.Address, 
 		return n, nil
 	}
 
-	raddr, err := mp.resolveToKey(ctx, addr)
-	if err != nil {
-		return 0, err
-	}
-
 	// get the nonce from the actor before ts
 	actor, err := mp.api.GetActorBefore(addr, ts)
 	if err != nil {
 		return 0, err
 	}
 	nextNonce := actor.Nonce
+
+	raddr, err := mp.resolveToKey(ctx, addr)
+	if err != nil {
+		return 0, err
+	}
 
 	// loop over all messages sent by 'addr' and find the highest nonce
 	messages, err := mp.api.MessagesForTipset(ctx, ts)
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -169,13 +169,13 @@ type MessagePool struct {
 
 	sigValCache *lru.TwoQueueCache[string, struct{}]
 
-	nonceCache *lru.Cache[nonceCacheKey, uint64]
+	stateNonceCache *lru.Cache[stateNonceCacheKey, uint64]
 
 	evtTypes [3]journal.EventType
 	journal  journal.Journal
 }
 
-type nonceCacheKey struct {
+type stateNonceCacheKey struct {
 	tsk  types.TipSetKey
 	addr address.Address
 }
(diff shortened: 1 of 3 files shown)


chain/messagepool/provider.go
```diff
@@ -77,6 +78,10 @@ func (mpp *mpoolProvider) PubSubPublish(k string, v []byte) error {
 	return mpp.ps.Publish(k, v) // nolint
 }
 
+func (mpp *mpoolProvider) GetActorBefore(addr address.Address, ts *types.TipSet) (*types.Actor, error) {
+	return mpp.sm.LoadActor(context.TODO(), addr, ts)
```
fridrik01, line 82: TODO: Support IsLite

Stebalien:
I can confirm that this fixes the performance issue (or, at least, we don't get 10s pauses).

chain/messagepool/messagepool.go
```diff
@@ -1078,14 +1078,42 @@ func (mp *MessagePool) getStateNonce(ctx context.Context, addr address.Address,
...
+	if err != nil {
+		return 0, err
+	}
+
+	// get the nonce from the actor before ts
+	actor, err := mp.api.GetActorBefore(addr, ts)
+	if err != nil {
+		return 0, err
+	}
+	highestNounce := actor.Nonce
```
raulk, line 1091:
```suggestion
	highestNonce := actor.Nonce
```

chain/messagepool/messagepool.go
```diff
@@ -1078,14 +1078,42 @@ func (mp *MessagePool) getStateNonce(ctx context.Context, addr address.Address,
...
+
+		if maddr == raddr {
+			if msg.Nonce+1 > highestNounce {
+				highestNounce = msg.Nonce + 1
+			}
+		}
+	}
 
-	mp.nonceCache.Add(nk, act.Nonce)
+	mp.nonceCache.Add(nk, highestNounce)
```
raulk, line 1114: Would prefer to call it nextNonce.

chain/messagepool/messagepool.go
```diff
@@ -1078,14 +1078,42 @@ func (mp *MessagePool) getStateNonce(ctx context.Context, addr address.Address,
...
+
+		maddr, err := mp.resolveToKey(ctx, msg.From)
+		if err != nil {
+			log.Warnf("failed to resolve message from address: %s", err)
+			continue
+		}
+
+		if maddr == raddr {
+			if msg.Nonce+1 > highestNounce {
+				highestNounce = msg.Nonce + 1
```
raulk, line 1109:
```suggestion
			if n := msg.Nonce+1; n > highestNounce {
				highestNounce = n
```

chain/messagepool/messagepool.go
raulk, line 1114: Would suggest to clean up here and rename this cache to stateNonceCache.

chain/messagepool/messagepool.go
```diff
@@ -371,7 +371,7 @@ func (ms *msgSet) toSlice() []*types.SignedMessage {
 func New(ctx context.Context, api Provider, ds dtypes.MetadataDS, us stmgr.UpgradeSchedule, netName dtypes.NetworkName, j journal.Journal) (*MessagePool, error) {
 	cache, _ := lru.New2Q[cid.Cid, crypto.Signature](build.BlsSignatureCacheSize)
 	verifcache, _ := lru.New2Q[string, struct{}](build.VerifSigCacheSize)
-	noncecache, _ := lru.New[nonceCacheKey, uint64](256)
+	noncecache, _ := lru.New[nonceCacheKey, uint64](32768) // 32k * ~200 bytes = 6MB
```
raulk, line 374: Does it make sense to impose this memory tax on every node, when only hosted RPC nodes would truly benefit as they are public deployments? Besides that, I'm not convinced this increase is useful since the cache will be practically stale every 30 seconds? (The most popular usage pattern is getting the nonce of the pending tipset)

chain/messagepool/messagepool.go
```diff
@@ -1078,14 +1078,42 @@ func (mp *MessagePool) getStateNonce(ctx context.Context, addr address.Address,
 		return n, nil
 	}
 
-	act, err := mp.api.GetActorAfter(addr, ts)
+	raddr, err := mp.resolveToKey(ctx, addr)
```
maciejwitowski, line 1081: `raddr` is only used much later in the `for` loop. It was suspicious for me that we assigned `raddr` but then kept on using `addr` in `GetActorBefore`. Wdyt about moving this line just before `for`?

chain/messagepool/provider.go
```diff
@@ -58,6 +60,23 @@ func (mpp *mpoolProvider) IsLite() bool {
 	return mpp.lite != nil
 }
 
+func (mpp *mpoolProvider) getActorLite(addr address.Address, ts *types.TipSet) (*types.Actor, error) {
```
maciejwitowski, line 63: It's a bit easier to follow for me if refactoring like this (this only extracts the function without any functional changes, right?) is done in a separate commit.

maciejwitowski (approved):
One more hotspot down! 💪

Stebalien (approved):
I agree with https://github.com/filecoin-project/lotus/pull/10700/files#r1173643973, but it's not a high priority.

We can definitely optimize this more as discussed (e.g., iterating over messages once per tipset and caching the results) but this fixes the main performance issue.

fridrik01:
@raulk Updated test plan with results from concurrent stress test, overall looks good!

PR #10713, opened 2023-04-20
fix: events: don't set GC confidence to 1

### LINKED ISSUE #10706: FEVM: Event observers: Use correct confidence
The event observers used by the EthAPI and Eth Events systems are created with a "confidence" of 1. I'm not sure how this was chosen, but this means in practice that any reorg is treated as a failure, since it is a "revert past finality" (see [here](https://github.com/filecoin-project/lotus/blob/e65fae28de2a44947dd24af8c7dafcade29af1a4/chain/events/observer.go#L214)).

This is gonna make whatever's listening to these events fairly buggy. We should _definitely_ set a larger value, and _probably_ just use the default, but that depends on how correctly these listeners handle revert events.

Stebalien:
## Proposed Changes

- Make `NewEventsWithConfidence` private, and rename to `newEventsWithGCConfidence`.
- Use the default GC confidence everywhere.

The function/parameter were poorly named and should never have been exposed. "GC" confidence should always be the same, this parameter doesn't let us actually set the _confidence_, just the point before which we no longer support reverts.

## Additional Info

fixes #10706

## Checklist

- [ x PR title is in the form of of `<PR type>: <area>: <change being made>`
- [x] CI is green

### DIFF
--- a/chain/events/events.go
+++ b/chain/events/events.go
@@ -47,7 +47,7 @@ type Events struct {
 	*hcEvents
 }
 
-func NewEventsWithConfidence(ctx context.Context, api EventAPI, gcConfidence abi.ChainEpoch) (*Events, error) {
+func newEventsWithGCConfidence(ctx context.Context, api EventAPI, gcConfidence abi.ChainEpoch) (*Events, error) {
 	cache := newCache(api, gcConfidence)
 
 	ob := newObserver(cache, gcConfidence)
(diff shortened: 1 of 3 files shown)


magik6k:
TestTxReceiptBloom appears to be sad, not sure if I saw it being flaky before

Stebalien:
Yeah, definitely failing.

arajasek (approved):

Issue #10622, opened 2023-04-05
Lotus to support eth_syncing RPC method


honghaoq:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [X] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Filecoin nodes don’t support the eth_syncing RPC [method](https://www.quicknode.com/docs/ethereum/eth_syncing). Some of the bridge validators were using that in their monitoring tools. It would be nice to have that supported.

### Describe the solution you'd like

Lotus to support eth_syncing RPC method

honghaoq:
@maciejwitowski FYI, can we get this into FVM eng planning?

maciejwitowski:
>  It would be nice to have that supported.

@honghaoq is it nice-to-have for them or a blocker? Also, any deadlines from their side?

honghaoq:
Nice to have, not blocking and no deadline from their side - they are still planning to landing on mainnet this week.

PR #4599, opened 2020-10-26
rewrite sync manager

vyzo:
~Note: hasn't been yet tested in the live network.~
It is currently being tested in the live network.

PR #10719, opened 2023-04-21
feat: Add eth_syncing RPC method

fridrik01:
Fixes: #10622 

This PR adds `eth_syncing` as a new RPC method to Lotus. This method returns an object with data about the sync status or false.

NOTE: I added `SyncApi` as dependency to the `EthModule` in order to implement this functionality, not sure if there was any way around that...

NOTE: If calling `eth_syncing` as soon as lotus daemon starts (within a <5 seconds) then it will result in an error (see test plan) as the syncing hasn't started properly, but I am  to making it instead wait.

**Test plan**

Started my lotus daemon on mainnet from snapshot and calling this method produces the following output depending on the sync status:

Before starting sync:
```
curl http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"eth_syncing","params":[],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","id":1,"error":{"code":1,"message":"no active syncs, try again"}}
```

During sync:
```
curl http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"eth_syncing","params":[],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","result":{"startingblock":"0x2a9e8e","currentblock":"0x2a9e95","highestblock":"0x2a9ea8"},"id":1}
```

After sync is done (according to spec `"result"` should be set to `"false"` in this case):
```
url http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"eth_syncing","params":[],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","result":"false","id":1}
```


node/impl/full/eth.go
```diff
@@ -672,6 +674,42 @@ func (a *EthModule) EthChainId(ctx context.Context) (ethtypes.EthUint64, error)
...
+	if len(state.ActiveSyncs) == 0 {
+		return ethtypes.EthSyncingResult{}, errors.New("no active syncs, try again")
+	}
+
+	working := -1
+	for i, ss := range state.ActiveSyncs {
+		if ss.Stage == api.StageIdle {
+			continue
+		}
+		working = i
```
maciejwitowski, line 692: Can you explain this? Why are looking for the index of the last non-idle one from ActiveSyncs and eg not the 1st one?

fridrik01: I wanted to have the same logic as when calling `lotus sync wait` which is done like this (see https://github.com/filecoin-project/lotus/blob/master/cli/sync.go#L286)

There is also [this commit](https://github.com/filecoin-project/lotus/pull/4599/commits/d6e2c80608179fa7abc488889fee1478fc0d1493) which picks the last worker in case all are idle, but it does not go in details as to why

PR #10732, opened 2023-04-22
itests: Test PoSt V1_1 on workers

magik6k:
## Proposed Changes
Just wanted to make sure this is covered, sector skipping logic may change after nv20 (tho the current one should still work, but better to have that covered explicitly)

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/worker_test.go
+++ b/itests/worker_test.go
@@ -1,6 +1,7 @@
 package itests
 
 import (
+	"bytes"
 	"context"
 	"strings"
 	"sync/atomic"
@@ -13,13 +14,17 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
+	miner11 "github.com/filecoin-project/go-state-types/builtin/v11/miner"
+	"github.com/filecoin-project/go-state-types/network"
 
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/itests/kit"
 	"github.com/filecoin-project/lotus/node"
+	"github.com/filecoin-project/lotus/node/config"
 	"github.com/filecoin-project/lotus/node/impl"
+	"github.com/filecoin-project/lotus/node/modules"
 	"github.com/filecoin-project/lotus/node/repo"
 	"github.com/filecoin-project/lotus/storage/paths"
 	sealing "github.com/filecoin-project/lotus/storage/pipeline"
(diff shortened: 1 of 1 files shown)


itests/worker_test.go
```diff
@@ -500,3 +505,228 @@ func TestWorkerName(t *testing.T) {
...
+	}, 30*time.Second, 100*time.Millisecond)
+
+	// Wait until just before a deadline opens
+	{
+		di, err := client.StateMinerProvingDeadline(ctx, maddr, types.EmptyTSK)
+		require.NoError(t, err)
+
+		di = di.NextNotElapsed()
+
+		t.Log("Running one proving period")
```
ZenGround0, line 577: It could be the case we wait a whole proving period (2880 epochs) if block mining surpasses deadline end between the previous code block and this one, but I think the happy path is ~ a proving deadline  aka a challenge window

`di` will be the same as in the previous block assuming mining doesn't pass the whole deadline between this and previous code block.  In that case waitUntil will be ~ current epoch + 39 (60 - 20 - 1).

ZenGround0: I think there's a race where block mining passes the end of this deadline before we wait for the epoch and stop mining which will cause post for this deadline to be missed and violation of assumptions in the next code block (`require.Greater(t, len(parts), 0)` will fail)

itests/worker_test.go
```diff
@@ -500,3 +505,228 @@ func TestWorkerName(t *testing.T) {
...
+		parts, err := client.StateMinerPartitions(ctx, maddr, di.Index+1, types.EmptyTSK)
+		require.NoError(t, err)
+		require.Greater(t, len(parts), 0)
+
+		secs := parts[0].AllSectors
+		n, err := secs.Count()
+		require.NoError(t, err)
+		require.Equal(t, uint64(2), n)
+
+		// Drop the sector in second partition
```
ZenGround0, line 599: nit: first partition

itests/worker_test.go
```diff
@@ -500,3 +505,228 @@ func TestWorkerName(t *testing.T) {
...
+		sid, err := secs.First()
+		require.NoError(t, err)
+
+		t.Logf("Drop sector %d; dl %d part %d", sid, di.Index+1, 0)
```
ZenGround0, line 603: nit: why +1 on deadline index? Shouldn't it just be di.Index?

itests/worker_test.go
```diff
@@ -500,3 +505,228 @@ func TestWorkerName(t *testing.T) {
...
+	require.Equal(t, network.Version20, nv)
+
+	require.True(t, pmr.Receipt.ExitCode.IsSuccess())
+
+	slmsg, err := client.ChainGetMessage(ctx, slm[0])
+	require.NoError(t, err)
+
+	var params miner11.SubmitWindowedPoStParams
+	require.NoError(t, params.UnmarshalCBOR(bytes.NewBuffer(slmsg.Params)))
+	require.Equal(t, abi.RegisteredPoStProof_StackedDrgWindow2KiBV1_1, params.Proofs[0].PoStProof)
```
ZenGround0, line 649: 👍

itests/worker_test.go
```diff
@@ -500,3 +505,228 @@ func TestWorkerName(t *testing.T) {
...
+	// Wait until just before a deadline opens
+	{
+		di, err := client.StateMinerProvingDeadline(ctx, maddr, types.EmptyTSK)
+		require.NoError(t, err)
+
+		di = di.NextNotElapsed()
+
+		t.Log("Running one proving period")
+		waitUntil := di.Open + di.WPoStChallengeWindow - di.WPoStChallengeLookback - 1
+		client.WaitTillChain(ctx, kit.HeightAtLeast(waitUntil))
```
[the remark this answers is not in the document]
magik6k: DisableBuiltinWindowPoSt doesn't influence this, unfortunately we don't have a "mine blocks until height" helper, and that would be non trivial to implement.

There is a slim chance that this may be flaky, but given that every windowpost test is doing the same thing, and I don't recall ever seeing this actually being flaky I'd argue this is good enough for now.

itests/worker_test.go
magik6k: Copy-paste artifact

arajasek (approved):

PR #10747, opened 2023-04-24
feat: nv19: Make using old proofs a configurable option

arajasek:
## Related Issues

We extended the period between nv19 and nv20 to 3 weeks so as to allow for dependence on old (v1) proofs in the unlikely event that there are issues with new proofs. We also want to make it easy for users to opt to use old proofs during the rollover period, just in case.

## Proposed Changes

This PR makes the proof type selection configurable behind the `LOTUS_NV19_OLD_PROOFS` envvar. If set to 1, the lotus-miner will generate "old" proofs during nv19. The default is still "new" proofs for nv19, and only new proofs will be generated for nv20.

## Checklist

- [ ] CI is green

### DIFF
--- a/storage/wdpost/wdpost_run.go
+++ b/storage/wdpost/wdpost_run.go
@@ -3,6 +3,7 @@ package wdpost
 import (
 	"bytes"
 	"context"
+	"os"
 	"time"
 
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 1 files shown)


storage/wdpost/wdpost_run.go
```diff
@@ -426,6 +427,14 @@ func (s *WindowPoStScheduler) runPoStCycle(ctx context.Context, manual bool, di
 				return nil, xerrors.Errorf("failed to get window post type: %w", err)
 			}
 
+			if nv == network.Version19 && os.Getenv("LOTUS_NV19_OLD_PROOFS") == "1" {
```
arajasek, line 430: Note that I'm NOT switching to old proofs in the `checkProvable` step, because that is unnecessary.

arajasek:
Tested this works by setting the envvar in TestWindowPostV1P1NV19

ZenGround0 (approved):

arajasek:
Not gonna merge unless this is actually needed, so as to avoid dead code in the future.

PR #10756, opened 2023-04-25
feat: sync: harden chain sync

Stebalien:
## Proposed Changes

This makes two changes:

1. Write headers in chain order, not reverse order. This shouldn't matter, but it may help if badger crashes.
2. Force `collectChain` to collect the correct chain. This shouldn't matter, but we might as well (as long as it's safe).

WRT 1, I'm worried about the following:

1. Attempt to sync a chain from epochs 3-7.
2. Write tipsets 7, 6, 5.
3. Badger realizes that the current vlog is "full", so it creates a new one.
4. Write tipsets 4, 3.

My concern is that, while badger will write _synchronously_, it doesn't call `fsync `on the file itself, and the directory containing the file. That means, if the machine crashes, the new vlog might just disappear.

Of course, the _right_ solution would be to fix badger to sync the file/directory metadata, but that's not likely to happen, ever.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/sync.go
+++ b/chain/sync.go
@@ -534,17 +534,6 @@ func (syncer *Syncer) Sync(ctx context.Context, maybeHead *types.TipSet) error {
 		return xerrors.Errorf("collectChain failed: %w", err)
 	}
 
-	// At this point we have accepted and synced to the new `maybeHead`
-	// (`StageSyncComplete`).
-	if err := syncer.store.PutTipSet(ctx, maybeHead); err != nil {
-		span.AddAttributes(trace.StringAttribute("put_error", err.Error()))
-		span.SetStatus(trace.Status{
-			Code:    13,
-			Message: err.Error(),
-		})
-		return xerrors.Errorf("failed to put synced tipset to chainstore: %w", err)
-	}
-
 	peers := syncer.receiptTracker.GetPeers(maybeHead)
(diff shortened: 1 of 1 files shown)


arajasek (commented):
@Stebalien I think the 2 changes you describe in the PR description make sense and are good to go. I don't understand why we're no longer putting the `maybeHead`, though.

Stebalien:
> I don't understand why we're no longer putting the maybeHead, though.

We should already be doing that here:

https://github.com/filecoin-project/lotus/blob/3f03e63fd1bcebc37fc582bf84d86808fcb6f850/chain/sync.go#L1191-L1198

We check that `headers[0]` is the correct head here: https://github.com/filecoin-project/lotus/blob/3f03e63fd1bcebc37fc582bf84d86808fcb6f850/chain/sync.go#L1184-L1186.

arajasek:
@Stebalien I do not think you're right here -- this is `PutTipset`ting, which is tremendously unfortunately named, but is what actually:

- expands tipsets where possible
- actually updates the chain head (if necessary).

I think dropping that code will break everything, though I'm admittedly not 100% sure.

Stebalien:
Oh! Yes, you're right. I saw `P...TipSet` and didn't even register that they were different functions.

arajasek (approved):

PR #10766, opened 2023-04-26
fix: stmgr: optimize StateLookupRobustAddress

### LINKED ISSUE #10763: 1.23.0 performance regression
An integ partner reported that `StateLookupRobustAddress` is significantly slower on 1.23.0 than 1.22.1. We should profile this, confirm the report, and fix if confirmed.

The root cause is likely to be at a layer lower than `StateLookupRobustAddress`, so it might affect multiple methods.

Stebalien:
## Related Issues

related to #10763

## Proposed Changes

Optimize `StateLookupRobustAddress`.

We might as well try to lookup the "key" address first. Note: this will also prefer f4 addresses over f2 addresses, which may not be what we want (they don't have the same guarantees).

**NOTE**: This will currently return prefer delegated addresses, when available. Previously, we'd non-deterministically pick either the delegated address or f2 address. We could, alternatively, prefer f2 addresses with a performance penalty.

**NOTE**: This will actually reduce the performance (slightly) when looking up miner actor's address. So we might not want it.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/stmgr/stmgr.go
+++ b/chain/stmgr/stmgr.go
@@ -403,9 +403,17 @@ func (sm *StateManager) LookupRobustAddress(ctx context.Context, idAddr address.
 	}
 
 	// Try to resolve directly.
-	resolved, err := vm.ResolveToDeterministicAddr(stateTree, cst, idAddr)
+	deterministicAddr, err := vm.ResolveToDeterministicAddr(stateTree, cst, idAddr)
 	if err == nil {
-		return resolved, nil
+		// Return immediately if we have a key address.
+		switch deterministicAddr.Protocol() {
+		case address.BLS, address.SECP256K1:
+			return deterministicAddr, nil
+		case address.Delegated:
+		default:
+			// Shouldn't be possible. Log and continue.
+			log.Errorf("unexpected deterministic address %s", deterministicAddr)
+		}
 	}
 
 	// Otherwise, search the init actor for an appropriate f2 address.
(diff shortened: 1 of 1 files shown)


arajasek:
- This will certainly make the case of resolving accounts faster, and I do think that's a useful case to cover (based on what I know of the users calling this).
- I do think this method should prefer f2 addresses where possible. Can you remind me in which cases an actor can have both an f2 and f4 address?
- While this looks like a nice improvement, I do not think it "fixes" #10763, which is more interested in trying to repro and confirm a regression between 1.22.1 and 1.23.0, so let's unlink this PR from that?

Stebalien:
> I do think this method should prefer f2 addresses where possible. Can you remind me in which cases an actor can have both an f2 and f4 address?

Everything created through the init actor has an f2 address. So EVM actors will have both an f2/f4.

> While this looks like a nice improvement, I do not think it "fixes" https://github.com/filecoin-project/lotus/issues/10763, which is more interested in trying to repro and confirm a regression between 1.22.1 and 1.23.0, so let's unlink this PR from that?

Fair enough... but can't find any even remotely relevant code unless we're just building with different go versions...

Stebalien:
Note: This PR is now messy enough that I'm not _entirely_ sure if we want to merge it. It should work, it's just not quite as trivial of an optimization.

arajasek:
Yeah, I'm not rushing to merge this either, but it _is_both a perf improvement in some cases (accounts), and (IMO) a correctness improvement in other cases (having both f4 and f2 addresses).

So...yes?

PR #10769, opened 2023-04-26
test: eth: deflake multiblock lookup test

Stebalien:
- Increase epoch times to give the miners a chance to see each other's blocks.
- Wait longer for a multi-block tipset.
- Reduce the initial wait (we're increasing the block times and I don't really feel like waiting around).

This isn't exactly a fix, but it should (hopefully) reduce the chances of a flake. If it doesn't.... well, we'll have to try something else.


arajasek (approved):
All sound reasonable to me.

PR #10780, opened 2023-04-28
Validate that FromBlock/ToBlock epoch is indeed a hex value

fridrik01:
**Context**
When calling ETH rpc method `eth_getLogs` and giving it a `fromBlock`/`toBlock` as epoch then return an error if its not in a hexadecimal format (instead of silently passing through and returning nothing or even worse potentially giving incorrect results)

**Test plan**

Verify that giving epoch ("2812200") as decimal returns a nice error:
```
curl -s http://localhost:1234/rpc/v1  -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"fromBlock": "2812200"}],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","id":1,"error":{"code":1,"message":"FromBlock is not a hex"}}
```

And that giving the same epoch in hex ("0x2AE928") works:
```
curl -s http://localhost:1234/rpc/v1  -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"fromBlock": "0x2AE928"}],"id":1,"jsonrpc":"2.0"}'|jq|head -n 15
{
  "jsonrpc": "2.0",
  "result": [
    {
      "address": "0xa35bf7ef23b09fb099e6e75a078727dff14b7576",
      "data": "0x0000000000000000000000000000000000000000000000000429d069189e0000",
      "topics": [
        "0xe1fffcc4923d04b559f4d29a8bfc6cda04eb5b0d3c460751c2402c5c5cc9109c",
        "0x0000000000000000000000006c29c4366a2f9ac5f083ef3394d9849b02a690c0"
      ],
      "removed": false,
      "logIndex": "0x0",
      "transactionIndex": "0x3f",
      "transactionHash": "0xce8a0e185d8e64ea5a9aeae51fde7f46aaa57239d1e3ce351759ccd9fb604c69",
      "blockHash": "0x3aa8d8603c95936cc964c64951bc75af118c5f79a8e81c8130281ec62ef98130",
```

### DIFF
--- a/chain/types/ethtypes/eth_types.go
+++ b/chain/types/ethtypes/eth_types.go
@@ -548,12 +548,12 @@ func (h EthSubscriptionID) String() string {
 }
 
 type EthFilterSpec struct {
-	// Interpreted as an epoch or one of "latest" for last mined block, "earliest" for first,
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
 	// "pending" for not yet committed messages.
 	// Optional, default: "latest".
 	FromBlock *string `json:"fromBlock,omitempty"`
 
-	// Interpreted as an epoch or one of "latest" for last mined block, "earliest" for first,
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
 	// "pending" for not yet committed messages.
 	// Optional, default: "latest".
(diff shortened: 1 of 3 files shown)


arajasek (commented):
LGTM, but let's add a very simple test to eth_filter_test.go for this.

Stebalien (approved):

Issue #10781, opened 2023-04-28
The ETH Rpc eth_getLogs is missing supports for blockHash param


fridrik01:
When calling the `eth_getLogs` rpc method one of the optional parameters is `blockhash` which if specified, looks up the tipset with that blockhash and returns all logs from that tipset (same as giving it the fromBlock == toBlock from the height at that tipset:

From the [eth docs](https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_getlogs): 

![image](https://user-images.githubusercontent.com/124135/235121087-70f88de8-3eb8-44e3-b168-9bc685e722cf.png)

PR #10782, opened 2023-04-28
feat: eth: Add support for blockHash param in eth_getLogs

fridrik01:
Fixes: #10781 

**Context**
When working on https://github.com/filecoin-project/lotus/pull/10780 I noticed we currently did not support the `blockHash` param for the `eth_getLogs` rpc method. If it was specified by user he got confusing results (fromBlock/toBlock were both 0). This PR therefore adds this feature as described in the [eth docs](https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_getlogs)

**Test plan**

Lets consider the tipset 2848101 (hex: 0x2B7565) and get the logs for only that tipset (using fromBlock/toBlock). Since the log output it long lets just count the lines and compute the sha1sum:
```
curl -s http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"fromBlock": "0x2B7565", "toBlock": "0x2B7565"}],"id":1,"jsonrpc":"2.0"}'|jq|wc -l
231

curl -s http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"fromBlock": "0x2B7565", "toBlock": "0x2B7565"}],"id":1,"jsonrpc":"2.0"}'|jq|sha1sum
50cbd75fd8e9059441d91173c53372237f5f3600  -
```

Now lets get the same tipset by specifying its blockHash and observe that we get the same results (since using blockHash is the same as specifying fromBlock=toBlock of the tipset):

```
curl -s http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"blockHash": "0xe8179500429f4c485015a9253e9ee7270a1d3f773fc33ea21d2f007e14623980"}],"id":1,"jsonrpc":"2.0"}'|jq|wc -l
231

curl -s http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"blockHash": "0xe8179500429f4c485015a9253e9ee7270a1d3f773fc33ea21d2f007e14623980"}],"id":1,"jsonrpc":"2.0"}'|jq|sha1sum 
50cbd75fd8e9059441d91173c53372237f5f3600  -
```

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -1209,7 +1209,7 @@ func (e *EthEvent) installEthFilterSpec(ctx context.Context, filterSpec *ethtype
 			return nil, xerrors.Errorf("must not specify block hash and from/to block")
 		}
 
-		// TODO: derive a tipset hash from eth hash - might need to push this down into the EventFilterManager
+		tipsetCid = filterSpec.BlockHash.ToCid()
 	} else {
 		if filterSpec.FromBlock == nil || *filterSpec.FromBlock == "latest" {
 			ts := e.Chain.GetHeaviestTipSet()


Stebalien:
I think we need to handle this in

https://github.com/filecoin-project/lotus/blob/8695fe342f44e809f507ae3d7fdd717383ce2985/node/impl/full/eth.go#L1173

(we may not have the block yet, or it may be on another fork)

I think we need to set `tipsetCid`:

https://github.com/filecoin-project/lotus/blob/8695fe342f44e809f507ae3d7fdd717383ce2985/node/impl/full/eth.go#L1163

Which will be matched here:

https://github.com/filecoin-project/lotus/blob/8695fe342f44e809f507ae3d7fdd717383ce2985/chain/events/filter/event.go#L164-L170

arajasek (approved):

PR #10783, opened 2023-04-28
(DO NOT MERGE) Remove semaphore that serializes access to Add/Push/PushUntrusted in MessagePool

fridrik01:
PROPOSAL ONLY - Do not merge

This PR removes a semaphore that actually serialized all access to the `Add()`, `Push()` and `PushUntrusted()` methods in `chain/messagepool` so that only one goroutine thread can run it at once.

This semaphore may have be contributing to the high latency spikes we see on the glif nodes

These methods already have a mutex guarding critical sections, but the question is whether its needed when publishing message to topics (which I guess can take a while since it calls out to libp2p and other peers).

### DIFF
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -122,8 +122,6 @@ type MessagePool struct {
 
 	ds dtypes.MetadataDS
 
-	addSema chan struct{}
-
 	closer chan struct{}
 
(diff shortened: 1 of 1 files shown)


fridrik01:
Talked about this with Steb, and this semaphore may be required to prevent exploitation of overloading the messagequeue, however the comment in the code for preventing lock contention should be changed

Issue #10560, opened 2023-03-24
lotus-fountain: make compatible with 0x addresses


ognots:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

make calibrationnet and other devnets FVM friendly, by making `lotus-fountain` compatible with 0x addresses.
this is necessary for the [quick start tutorial](https://docs.filecoin.io/smart-contracts/fundamentals/erc-20-quickstart/) to work as published

### Describe the solution you'd like

Call `EthAddressToFilecoinAddress` when an 0x type address is detected in lotus-fountain's address field in the SENDING FUNDS section

ognots:
Relevant code https://github.com/filecoin-project/lotus/tree/master/cmd/lotus-fountain

maciejwitowski:
Took a stab at implementing it https://github.com/filecoin-project/lotus/pull/10784

ognots:
Thanks @maciejwitowski this looks great. we'll still need a review from the lotus maintainers team. Let me know if you want me to escalate the priority of the review, this is a very important feature for the community

PR #10784, opened 2023-04-28
lotus-fountain: make compatible with 0x addresses #10560

maciejwitowski:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/10560

## Proposed Changes
Check if address starts with `Ox`. If so, parse it as Eth address. Otherwise, use the previous logic.

### DIFF
--- a/cmd/lotus-fountain/main.go
+++ b/cmd/lotus-fountain/main.go
@@ -195,9 +195,11 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var to address.Address
-
 	addressInput := r.FormValue("address")
+
+	var filecoinAddress address.Address
+	var decodeError error
+
 	if strings.HasPrefix(addressInput, "0x") {
 		ethAddress, err := ethtypes.ParseEthAddress(addressInput)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


ognots (approved):

cmd/lotus-fountain/main.go
```diff
@@ -193,14 +195,34 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		filecoinAddress, err := ethAddress.ToFilecoinAddress()
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		to = filecoinAddress
+	} else {
+		filecoinAddress, err := address.NewFromString(r.FormValue("address"))
```
fridrik01, line 215:
```suggestion
		filecoinAddress, err := address.NewFromString(addressInput)
```

cmd/lotus-fountain/main.go
```diff
@@ -193,14 +195,34 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		filecoinAddress, err := ethAddress.ToFilecoinAddress()
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		to = filecoinAddress
```
fridrik01, line 213: Is it ok to skip the check for empty address with checking filecoinAddress == address.Undef like when its not a 0x address? Can we make the checks for both scenarios the same and maybe refactor the code a bit to use the same error checking code?

cmd/lotus-fountain/main.go
maciejwitowski: Good spot, will merge into the commit.

cmd/lotus-fountain/main.go
maciejwitowski: `ToFilecoinAddress` can return `address.Undef` but in such case it also returns `error` so it's not strictly needed because the error is checked here. The point about having the same error handling is legit, will do.

maciejwitowski: @fridrik01 Is this the way to go in Go https://github.com/filecoin-project/lotus/pull/10784/commits/6c25f2172e41fddb775cfb911f1681ca7b2507db ? 🙂

ZenGround0 (approved):

cmd/lotus-fountain/main.go
```diff
@@ -193,12 +195,30 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	to, err := address.NewFromString(r.FormValue("address"))
-	if err != nil {
+	var to address.Address
```
magik6k, line 198: We're not setting this anywhere, and using for the send destination, probably want to remove this var, and swap uses of `to` to `filecoinAddress`

magik6k (changes requested):

fridrik01 (approved):
LGTM

fridrik01 (approved):
LGTM

PR #10800, opened 2023-05-02
feat: chainstore: batch writes of tipsets

arajasek:
## Related Issues

Helps with #10788, but does NOT close it.

## Proposed Changes

- Refactor the chainstore to persist tipsets in a batch
- The batching logic already exists, so this is mostly about providing input in a batch to this method

## Checklist

- [ ] CI is green


chain/store/snapshot.go
```diff
@@ -123,11 +123,9 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (*types.TipSet, e
 	}
 
 	ts := root
+	tssToPersist := make([]*types.TipSet, 0, uint64(TipsetkeyBackfillRange)*build.BlocksPerEpoch)
```
Stebalien, line 126:
```suggestion
	tssToPersist := make([]*types.TipSet, 0, uint64(TipsetkeyBackfillRange))
```

ZenGround0, line 126: nit: drop *build.BlocksPerEpoch since one slice item per tipset

chain/store/snapshot.go
```diff
@@ -123,11 +123,9 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (*types.TipSet, e
...
 	for i := 0; i < int(TipsetkeyBackfillRange); i++ {
-		err = cs.PersistTipset(ctx, ts)
-		if err != nil {
-			return nil, err
-		}
+		tssToPersist = append(tssToPersist, ts)
 		parentTsKey := ts.Parents()
 		ts, err = cs.LoadTipSet(ctx, parentTsKey)
```
ZenGround0, line 130: Sanity checking since I don't know this code: is this actually going to work now that we are not persisting tipsets one by one in this loop before the check?  We're putting these tipsets to cs somewhere else first right?

ZenGround0 (approved):

chain/store/snapshot.go
arajasek: My understanding is yes, because we've already `Put` all of these objects into our store above. This loop is about:
- sanity checking that we have them
- writing the TsKeys into the store for Eth stuff

It's probably not actually any faster to batch this, but I figured we might as well for cleanness.

PR #10805, opened 2023-05-02
feat: wallet security solution

cdcdx:
## Related Issues
  - Some time ago, many miners’ wallet private keys were stolen, and the owner was tampered with. Now the wallet encryption part is sorted out and open sourced. Welcome to integrate;

## Security Advice
  - This solution uses AES encryption, which guarantees the security of funds and private keys to a certain extent, but there is still a risk of leakage, please be aware;
  - While encrypting the private key of the wallet, it is necessary to strengthen internal management and control device access rights;
  - Remember to modify `walletSaltPwd` variable before compiling;

## Proposed Changes
  - [x] compatible adaptations:
    - [x] Compatible with the official lotus node api;
    - [x] Adapt to lotus and lotus-wallet;
  - [x] Wallet token management:
    - [x] Add, delete, clear wallet flags
  - [x] Wallet password management:
    - [x] Add, reset, clear wallet password
    - [x] Normal wallet and crypto wallet swap;
  - [x] Wallet message management:
    - [x] Guarantee the security of funds: password verification is required for encrypted wallet transfer operations;
    - [x] Protect the security of the private key: encrypt the wallet to export the private key and delete the password that needs to be verified;
    - [x] Sealed messages are not affected: Encrypted wallets seal messages without verifying passwords;

## Additional Info
  - Wallet mark management:
```shell
    lotus wallet mark add    ##Add/update wallet mark
    lotus wallet mark del    ##Delete wallet mark
    lotus wallet mark clear  ##clear wallet mark
```
  - Wallet password management:
```shell
    lotus wallet passwd add    ##Add password
    lotus wallet passwd reset  ##Change password
    lotus wallet passwd clear  ##Clear password
    lotus wallet encrypt <f1xxx/f3xxx/all>  ##Ordinary wallet->encrypted wallet
    lotus wallet decrypt <f1xxx/f3xxx/all>  ##encryption wallet->ordinary wallet
```


Stebalien:
Why are you not using the out-of-process wallet? Anything in-process is going to be strictly less secure.

cdcdx:
> Why are you not using the out-of-process wallet? Anything in-process is going to be strictly less secure.

Yes, this solution can ensure the security of worker-address and post-address .

Stebalien:
Hm. It seems like the "remote" wallet type should fit your needs. Especially because it:

- Allows you to use hardware wallets.
- Can prevent keys from being dumped from memory.

We can't accept security critical code like this without a careful audit and consideration as people will rely on it. Unfortunately, your solution does not appear to be secure; any secure solution would at _least_ require some form of password when starting lotus.

PR #10812, opened 2023-05-03
Change args check

rjan90:
## Proposed Changes
Change args check in `lotus chain set` cmd to make it possible to actually use the --epoch option and --genesis options. Regression from  #10114.

## Checklist

- [ ] CI is green

### DIFF
--- a/cli/chain.go
+++ b/cli/chain.go
@@ -388,6 +388,10 @@ var ChainSetHeadCmd = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
+		if cctx.Args().Present() {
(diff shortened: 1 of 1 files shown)


cli/chain.go
```diff
@@ -388,10 +388,6 @@ var ChainSetHeadCmd = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
-		if cctx.NArg() != 1 {
```
arajasek, line 391: Consider just changing to `if cctx.Args().Present()`, and failing in that case?

arajasek (approved):

Issue #10814, opened 2023-05-03
EIP-1898 support needed for The Graph compatibility


fridrik01:
See https://github.com/filecoin-project/fvm-pm/issues/389

PR #10815, opened 2023-05-03
Add EIP-1898 support needed for The Graph compatibility

fridrik01:
Fixes: https://github.com/filecoin-project/lotus/issues/10814
See also: https://github.com/filecoin-project/fvm-pm/issues/389

This PR updates the following RPC methods according to [EIP-1898](https://eips.ethereum.org/EIPS/eip-1898) specs.

The following RPC methods are affected:
- `eth_getBalance`
- `eth_getStorageAt`
- `eth_getTransactionCount`
- `eth_getCode`
- `eth_call`

Note that `eth_getBlockByNumber` was not included in this list in the spec although it seems it should be affected also?

Currently these methods all accept a `blkParam` string which can be one of "latest", "earliest", "pending", or a block number (decimal or hex). The spec enables caller to additionally specify a json hash which can include the following fields: 

- `blockNumber EthUint64`: A block number (decimal or hex) which is similar to the original use of the `blkParam` string
- `blockHash EthHash`: The block hash
-  `requireCanonical bool`) If true we should make sure the block is in the canonical chain 

Since the `blkParam` needs to support both being a number/string and a json hash then this to properly work we need to introduce a new struct with pointer fields to check if they exist.  This is done in the `EthBlockParamByNumberOrHash` struct which first tries to unmarshal as a json hash (according to eip-1898) and then fallback to unmarshal as string/number. 

**Test plan**

Check that specifying by predefined blocks works:

```
curl  http://localhost:1234/rpc/v1  -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionCount","params":["0xd4c70007F3F502f212c7e6794b94C06F36173B36", "latest"],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","result":"0x2b","id":1}
```

Check that specifying by number works:

```
curl -s http://localhost:1234/rpc/v1  -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionCount","params":["0xd4c70007F3F502f212c7e6794b94C06F36173B36", "0x2B229E"],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","result":"0x2b","id":1}
```

Check that specifying by blockNumber works:
```
curl -s http://localhost:1234/rpc/v1  -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionCount","params":["0xd4c70007F3F502f212c7e6794b94C06F36173B36", {"blockNumber": "0x2B229E"}],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","result":"0x2b","id":1}
```

Check that specifying by blockHash works:
```
curl -s http://localhost:1234/rpc/v1  -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionCount","params":["0xd4c70007F3F502f212c7e6794b94C06F36173B36", {"blockHash": "0x7f695f396c2782d772a46af38428ef8f27918ba8158b36854d547ff8e3b994b9"}],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","result":"0x2b","id":1}
```

Check that specifying by blockHash works and when requireCanonical is set:
```
curl -s http://localhost:1234/rpc/v1  -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionCount","params":["0xd4c70007F3F502f212c7e6794b94C06F36173B36", {"blockHash": "0x7f695f396c2782d772a46af38428ef8f27918ba8158b36854d547ff8e3b994b9", "requireCanonical": true}],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","result":"0x2b","id":1}
```

Check that specifying both blockNumber and blochHash should fail:

```
curl -s http://localhost:1234/rpc/v1  -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionCount","params":["0xd4c70007F3F502f212c7e6794b94C06F36173B36", {"blockNumber": "0x2B229E", "blockHash": "0x7f695f396c2782d772a46af38428ef8f27918ba8158b36854d547ff8e3b994b9", "requireCanonical": true}],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32700,"message":"unmarshaling params for 'eth_getTransactionCount' (param: *ethtypes.EthBlockParamByNumberOrHash): cannot specify both blockNumber and blockHash"}}
```

Check specifying by blockHash with requireCanonical is set where blockHash is not in the canonical chain:
```
TODO: Still trying to find a way to get a orphaned blockHash
```


chain/types/ethtypes/eth_types.go
```diff
@@ -815,3 +815,82 @@ func (e EthFeeHistoryParams) MarshalJSON() ([]byte, error) {
 	}
 	return json.Marshal([]interface{}{e.BlkCount, e.NewestBlkNum})
 }
+
+type EthBlockParamByNumberOrHash struct {
+	PredefinedBlock *string
```
Stebalien, line 820: I'd consider using negative constants to represent "latest", etc. Storing them in the `BlockNumber` field. I.e., we'd have three fields only:

```go
const (
    EthBlockLatest EthUint64 = -iota
    EthBlockPending
    EthBlockEarliest
    // ...
)

type EthBlockParam struct {
    BlockNumber      *EthUint64
    BlockHash        *EthHash
    RequireCanonical bool
}
```

fridrik01: Would love feedback from @filecoin-project/lotus-maintainers team on Stebs comment here.

chain/types/ethtypes/eth_types.go
```diff
@@ -815,3 +815,82 @@ func (e EthFeeHistoryParams) MarshalJSON() ([]byte, error) {
...
+	Number          *EthUint64
```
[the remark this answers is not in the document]
fridrik01: @Stebalien I looked better into merging the `Number`/`BlockNumber` into one and I think it may not be such a great idea.

For context, the issue is that currently the `blkParam` can be one of "latest", "earliest", "pending", or a block number (decimal or hex) but with this PR we additionally need to support that `blkParam` to be able to be a json hash witch a certain named fields (blockNumber,blockHash,requireCanonical).  To implement this, I introduced this `EthBlockParamByNumberOrHash` struct which has fields for both cases (normal as before + support for json hashes) and handles the al the construction/json deserialization. 

It would be possible to merge the `Number` and `BlockNumber` in `EthBlockParamByNumberOrHash`into one, but I would think that makes the code less readable since right now its easy to see from reading the code whether we are talking about a normal block number or a one we need to support for the EIP-1891. Also, it would make error handling cumbersome (like [here](https://github.com/filecoin-project/lotus/pull/10815/files#diff-b4b04abadd60724f3b7710cced91031f331c2edacb30a1c1bb7d5aeb67fc414cR103-R106)).

chain/types/ethtypes/eth_types.go
```diff
@@ -839,3 +839,80 @@ func (e EthFeeHistoryParams) MarshalJSON() ([]byte, error) {
 	}
 	return json.Marshal([]interface{}{e.BlkCount, e.NewestBlkNum})
 }
+
+type EthBlockNumberOrHash struct {
+	PredefinedBlock  *string
```
Stebalien, line 844: This field shouldn't be serialized to json.

node/impl/full/eth.go
```diff
@@ -241,7 +241,68 @@ func (a *EthModule) EthGetBlockByHash(ctx context.Context, blkHash ethtypes.EthH
...
+					return ts, nil
+				} else if currTs.Height() < ts.Height() {
+					return nil, fmt.Errorf("could not find block hash %s in canonical chain", blkParam.BlockHash.ToCid())
+				}
+
+				currTs, err = a.Chain.LoadTipSet(ctx, currTs.Parents())
+				if err != nil {
+					return nil, fmt.Errorf("failed to load tipset: %v", err)
+				}
+			}
```
Stebalien, line 296: Call `ChainGetTipSetByHeight` and let that walk for you. That'll use the index/skip-list.

Stebalien: Then, when you have the tipset at the target height:

1. Make sure it's actually at the target height (nulls, etc.). Just in case.
2. Make sure it equals the expected tipset.

Stebalien (commented):
Two small changes but otherwise lgtm.

node/impl/full/eth.go
```diff
@@ -241,7 +241,65 @@ func (a *EthModule) EthGetBlockByHash(ctx context.Context, blkHash ethtypes.EthH
 	return newEthBlockFromFilecoinTipSet(ctx, ts, fullTxInfo, a.Chain, a.StateAPI)
 }
 
-func (a *EthModule) parseBlkParam(ctx context.Context, blkParam string, strict bool) (tipset *types.TipSet, err error) {
+func (a *EthModule) getTipsetByEthBlockNumberOrHash(ctx context.Context, blkParam ethtypes.EthBlockNumberOrHash) (*types.TipSet, error) {
+	head := a.Chain.GetHeaviestTipSet()
+
+	predefined := blkParam.PredefinedBlock
+	if predefined != nil {
+		if *predefined == "earliest" {
```
Stebalien, line 249: This should probably be a switch, but it's not critical.

Stebalien (approved):

Stebalien:
Looks like the final issue is the linter not being happy about some error formatting.

fridrik01:
> Looks like the final issue is the linter not being happy about some error formatting.

I need to look better into it tomorrow, but I think its because of the change of ignoring `PredefinedBlock` when marshaling json.

Stebalien:
Well, the linter is mad about an `Errorf`.

fridrik01:
@Stebalien There were several failed tests that were due to the itest/testkit creating a client and setting the `ThroughRPC()` option. This causes the struct to always be serialized to JSON and in case of testing the predefined blocks (like [here](https://github.com/filecoin-project/lotus/blob/10814-eip-1891/itests/eth_balance_test.go#L95)) then when we try to unmarshal the struct, we get an empty hash `{}` which results in a failure. 

To address this we can either:
1. Serialize this field as json
2. Disable RPC in our tests
3. Merge predefined block field with the BlockNumber but that would  require creating a new `EthInt64` type which would require a big refactor in the codebase as this is used everywhere.

I updated the PR with 1) in mind, but let me know if you would prefer other options

Stebalien:
I believe the correct thing to do is to marshal it into the `blockNumber` field.

Stebalien:
Well, it's unclear. That's what go-ethereum would do, but it's probably not correct.

IMO, the "most correct" thing to do would be to implement `MarshalJSON` and:

1. If a predefined block is specified, marshal to just that as a string.
2. Otherwise, marshal as an object.

Stebalien:
My concern with the current approach is that we're now exposing `predefiendBlock` as a field on the API, when it's _supposed_ to just be an implementation detail.

fridrik01:
> Well, it's unclear. That's what go-ethereum would do, but it's probably not correct.
> 
> IMO, the "most correct" thing to do would be to implement `MarshalJSON` and:
> 
> 1. If a predefined block is specified, marshal to just that as a string.
> 2. Otherwise, marshal as an object.

That should work, just create a tmp struct like in Unmarshal, let me do that!

Stebalien (approved):

PR #10816, opened 2023-05-03
feat: deflake TestEthFeeHistory

arajasek:
## Related Issues

This test has been observed to flake, eg. [here](https://app.circleci.com/pipelines/github/filecoin-project/lotus/28509/workflows/82949f35-6daa-479a-98d7-46223cf253b7/jobs/957928).

This particular failure arose because the test as currently written always expects the oldest non-genesis height to be 1. It IS possible for the first round to be null, though, which is what happened in the failure linked above.

## Proposed Changes

Construct the heights based on the actual chain, not by hard-coding a first height of 1 and then observing chain head changes.

Happily, I was able to reproduce the flake by forcing the first tipset to be null, and confirmed that the change made here fixes the failure.

## Additional Info

It would not surprise me if this test continues to flake for other reasons, but I am confident in this improvement.

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/eth_fee_history_test.go
+++ b/itests/eth_fee_history_test.go
@@ -50,19 +50,6 @@ func TestEthFeeHistory(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
-	heads, err := client.ChainNotify(ctx)
-	require.NoError(err)
-
-	// Save the full view of the tipsets to calculate the answer when there are null rounds
-	tsHeights := []int{1}
-	go func() {
-		for chg := range heads {
-			for _, c := range chg {
-				tsHeights = append(tsHeights, int(c.Val.Height()))
-			}
-		}
-	}()
-
 	miner := ens.InterconnectAll().BeginMining(blockTime)
 
 	client.WaitTillChain(ctx, kit.HeightAtLeast(7))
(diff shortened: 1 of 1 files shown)


itests/eth_fee_history_test.go
```diff
@@ -89,7 +76,15 @@ func TestEthFeeHistory(t *testing.T) {
 		}
 	}()
 
-	sort.Ints(tsHeights)
+	currTs, err := client.ChainHead(ctx)
+	require.NoError(err)
+
+	var tsHeights []int
+	for currTs.Height() != 0 {
+		tsHeights = append([]int{int(currTs.Height())}, tsHeights...)
```
fridrik01, line 84: Not sure of the intent here, but if if you want reverse order, would it not be better/readable to call append like normally and then sort afterwards?

itests/eth_fee_history_test.go
```diff
@@ -50,19 +50,6 @@ func TestEthFeeHistory(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
-	heads, err := client.ChainNotify(ctx)
-	require.NoError(err)
-
-	// Save the full view of the tipsets to calculate the answer when there are null rounds
-	tsHeights := []int{1}
```
arajasek, line 57: Note: removing this line is the actual deflaking fix

raulk (approved):
SGTM on populating our understanding of the chain after the fact.

Issue #10820, opened 2023-05-04
Follower node Metadata error


snissn:
Running this branch https://github.com/filecoin-project/lotus/tree/mikers/feat/cassandra-store when I run a follower node connected to yugadb through cassandra and run this rpc command:

```
 curl http://localhost:4234/rpc/v1 -X POST -H "Content-Type: application/json" --dat
a '{"method":"Filecoin.EthEstimateGas","params":[{"from":"0x08ca94cdD897F4382cb4E83Cb58e976ce06a7E68","to":"0x08ca94cdD897F4382cb4
E83Cb58e976ce06a7E68","value":"0x186a0"}],"id":1,"jsonrpc":"2.0"}'
```
I get this error:

```

2023-05-03T14:29:44.995-1000    WARN    rpc     go-jsonrpc@v0.2.3/handler.go:406        error in RPC call to 'Filecoin.EthEstimateGas': failed to estimate gas:
    github.com/filecoin-project/lotus/node/impl/full.(*EthModule).EthEstimateGas
        /home/mikers/dev/filecoin-project/lotus/node/impl/full/eth.go:941
  - CallWithGas failed:
    github.com/filecoin-project/lotus/node/impl/full.(*EthModule).applyMessage
        /home/mikers/dev/filecoin-project/lotus/node/impl/full/eth.go:909
  - call raw get actor:
    github.com/filecoin-project/lotus/chain/stmgr.(*StateManager).callInternal
        /home/mikers/dev/filecoin-project/lotus/chain/stmgr/call.go:193
  - resolution lookup failed (t410fbdfjjtoys72dqlfu5a6llduxntqgu7tizvs7jga):
    github.com/filecoin-project/lotus/chain/state.(*StateTree).GetActor
        /home/mikers/dev/filecoin-project/lotus/chain/state/statetree.go:372
  - resolve address t410fbdfjjtoys72dqlfu5a6llduxntqgu7tizvs7jga:
    github.com/filecoin-project/lotus/chain/state.(*StateTree).lookupIDinternal
        /home/mikers/dev/filecoin-project/lotus/chain/state/statetree.go:337
  - actor not found
 ```

the error is resolved when i copy over the ~150k metadata storage db into the follower node lotus working directory.

snissn:
I did some digging, and identified that the issue is that the genesis block is not correctly set up to include the correct fvm version number.

[...]

Issue #10630, opened 2023-04-05
Scalable Lotus Node Infrastructure with YugabyteDB


snissn:
# Title: Scalable Lotus Node Infrastructure with YugabyteDB

## Summary

This proposal aims to create a scalable Lotus infrastructure leveraging YugabyteDB as the backend datastore, supporting multiple RPC nodes behind a load balancer, and supporting linearly being able to increase read IOPs through horizontally scaling and adding more nodes to the Yugabyte cluster. Because of the Yugabyte cluster, all of the RPC nodes will see identical and synchronized chain state.

## Proposal

1. Integrate YugabyteDB as the backend datastore for Lotus nodes.
   - [Link to GitHub Pull Request](https://github.com/filecoin-project/lotus/pull/10624)
   
2. Implement a new node mode called "follower."
   - Add a new node "mode" called "follower," where the node doesn't sync but follows the heads of other nodes.
       - Modify `node/builder.go#L241-L246` to add `isFollowerNode`.
   - Don't create syncer / syncer services:
       - Modify `node/builder_chain.go#L91-L92` and `node/builder_chain.go#L174` to wrap the initialization in an if statement that does not initialize if the node is a follower node.
   - Don't subscribe to blocks/messages.
   - Connect to a shared chain blockstore with #10624 (YugabyteDB).
   - Write a small service replacing the syncer, get heads with ChainNotify from a "full" node RPC, and pass them to `chainstore.PutTipSet`:
       - Modify `chain/store/store.go#L380`.

3. Support multiple RPC nodes behind a load balancer.
   - Set up multiple Lotus nodes in "follower" mode.
   - Use a load balancer to distribute incoming requests.

4. Benchmark the new infrastructure to ensure performance improvements.
   - Test the new infrastructure setup with community RPC node providers to validate its effectiveness.

5. Create documentation for setting up the proposed infrastructure.
   - Provide clear and concise documentation for the community to adopt and implement this scalable Lotus infrastructure.

## Infrastructure Diagram

```

                    ┌───────────────┐
                    │ Load Balancer │
                    └──────┬────────┘
                           │
              ┌────────────┴────────────┐
              │                         │
┌─────────────┴───────┐      ┌────────────┴────────┐     ...   ┌─────────────┴───────┐
│RPC Node (Follower) 1│      │RPC Node (Follower) 2│           │RPC Node (Follower) N│
└─────┬───────────────┘      └─────┬───────────────┘           └─────┬───────────────┘
[...]

PR #10821, opened 2023-05-04
feat: chain/state: yugadb store

snissn:
# Title: Scalable Lotus Node Infrastructure with YugabyteDB

## Summary

This proposal aims to create a scalable Lotus infrastructure leveraging YugabyteDB as the backend datastore, supporting multiple RPC nodes behind a load balancer, and supporting linearly being able to increase read IOPs through horizontally scaling and adding more nodes to the Yugabyte cluster. Because of the Yugabyte cluster, all of the RPC nodes will see identical and synchronized chain state.

## Infrastructure Diagram

 
image (4)

# Local Test Server Setup Instructions

The instructions below guide you in setting up a local test server. This setup aims for scalability, robustness, and distribution. It can be easily extended to a larger, scalable infrastructure.

## Prerequisites

1. Set up standard lotus node prerequisites. Detailed instructions can be found [here](https://lotus.filecoin.io/lotus/install/linux/).

## Setup Instructions

1. **Set up YugabyteDB:**
   - You can download YugabyteDB from [this link](https://download.yugabyte.com/#linux).
   - Execute the following commands:
     ```
     wget https://downloads.yugabyte.com/releases/2.18.0.0/yugabyte-2.18.0.0-b65-linux-x86_64.tar.gz
     tar xvfz yugabyte-2.18.0.0-b65-linux-x86_64.tar.gz && cd yugabyte-2.18.0.0/
     ./bin/post_install.sh
     ./bin/yugabyted start
     ```

2. **Download Mainnet snapshot:**
   - You can download it from [this link](https://lotus.filecoin.io/lotus/manage/chain-management/).
   - Execute the following command:
     ```
     aria2c -x5 https://snapshots.mainnet.filops.net/minimal/latest.zst
     ```

3. **Sync chain:**

   - Create default config:
     ```
     mkdir -p ~/.lotus
     ./lotus config default > ~/.lotus/config.toml
     ```

   - Disable splitstore:
     ```
     EnableSplitstore = false
     ```

   - Import mainnet snapshot data. This process should take approximately 20 minutes:
     ```
     LOTUS_PATH=~/.lotus LOTUS_CASSANDRA_UNIVERSAL_STORE=127.0.1.1:9042 ./lotus daemon --halt-after-import --import-snapshot latest.car.zst 
     ```

   - Start node daemon:
     ```
     LOTUS_PATH=~/.lotus LOTUS_CASSANDRA_UNIVERSAL_STORE=127.0.1.1:9042 ./lotus daemon
     ```

4. **Configure Follower node:**

   - Create the follower node configuration:
     ```
     mkdir -p ~/.lotusfollower
     ./lotus config default > ~/.lotusfollower/config.toml
     ```

   - Update follower config:
     - Disable splitstore
     - Enable eth rpc:
       ```
       EnableEthRPC = true
       ```
     - Change port to 4234 (to not conflict with syncer daemon):
       ```
       ListenAddress = "/ip4/127.0.0.1/tcp/4234/http"
       ```

   - Start follower node:
     ```
     LOTUS_PATH=~/.lotusfollower/ FULLNODE_API_INFO=ws://localhost:1234 LOTUS_CASSANDRA_UNIVERSAL_STORE=127.0.1.1:9042 ~/filecoin-project/lotus/lotus daemon --follower
     ```

   - Test that the follower node responds to eth rpc queries:
     ```
     curl http://localhost:4234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"Filecoin.EthEstimateGas","params":[{"from":"0x08ca94cdD897F4382cb4E83Cb58e976ce06a7E68","to":"0x08ca94cdD897F4382cb4E83Cb58e976ce06a7E68","value":"0x186a0"}],"id":1,"jsonrpc":"2.0"}'
     ```

## Suggested Hardware for Single Node Testing

- 2TB 4.0 NVMe SSD: [980 PRO w/ Heatsink PCIe® 4.0 NVMe™ SSD 2TB](https://www.samsung.com/us/computing/memory-storage/solid-state-drives/980-pro-w-heatsink-pcie-4-0-nvme-ssd-2tb-mz-v8p2t0cw/) ($170)
- 32GB Ram: 4 x [DDR4 8GB 3200MHz](https://www.amazon.com/Patriot-Signature-Line-3200MHz-Single/dp/B0887VLQPB?th=1)  ($76)
- 6 core Intel chip: [Intel Core i5-11400F @ 2.60GHz](https://www.cpubenchmark.net/cpu.php?id=4226&cpu=Intel+Core+i5-11400F+%40+2.60GHz) ($133)

## Related Issues
Overview of the rational behind this PR, with some of this issues content repeated here: https://github.com/filecoin-project/lotus/issues/10630

issue related to metadata shared between syncer and follower node that was resolved in this pr: https://github.com/filecoin-project/lotus/issues/10820

## Checklist

- [ ] CI is green


snissn:
For dev testing:

Yugabyte DB setup:

```
docker pull yugabytedb/yugabyte:2.17.2.0-b216
docker run -d --name yugabyte  -p7000:7000 -p9000:9000 -p5433:5433 -p9042:9042 yugabytedb/yugabyte:2.17.2.0-b216 bin/yugabyted start --daemon=false --ui=false
```

Chain Sync node start up command:
```
LOTUS_PATH=/mnt/e/yuga   LOTUS_CASSANDRA_UNIVERSAL_STORE=localhost:9042 ~/dev/filecoin-project/lotus/lotus daemon --import-snapshot /mnt/e/calibnet/527280_2023_05_03T20_13_00Z.car
```
Follower node start up command:

```
LOTUS_PATH=/mnt/e/yugafollower/  FULLNODE_API_INFO=http://localhost:1234 LOTUS_CASSANDRA_UNIVERSAL_STORE=localhost:9042 ~/dev/filecoin-project/lotus/lotus daemon   --follower
```

chain/sync.go
```diff
@@ -160,6 +160,37 @@ func NewSyncer(ds dtypes.MetadataDS,
...
+	sm *stmgr.StateManager,
+	exchange exchange.Client,
+	syncMgrCtor SyncManagerCtor,
+	connmgr connmgr.ConnManager,
+	self peer.ID,
+	beacon beacon.Schedule,
+	gent Genesis,
+	consensus consensus.Consensus) (*Syncer, error) {
+
+	s := &Syncer{
```
magik6k, line 174: We shouldn't run the chain syncer on follower nodes - the follower node should be following ChainHead from syncer node(s).

snissn: If i do not start up the Chain syncer, I get an error in the Hello module, line 85:



```
if hmsg.GenesisHash != hs.syncer.Genesis.Cids()[0] {
                log.Debugf("other peer has different genesis! (%s)", hmsg.GenesisHash)
                _ = s.Conn().Close()
                return
}
```

I'm investigating disabling the Hello module in Follower mode as well

snissn: That seems to work now!

snissn: #e8e96a267

fridrik01:
@snissn could you add some details in the PR description, I think it would make reviewing easier and quicker for such a large PR/LOC

cmd/lotus/daemon.go
```diff
@@ -515,23 +537,19 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
...
-		return xerrors.Errorf("flushing validation cache failed: %w", err)
-	}
-
-	gb, err := cst.GetTipsetByHeight(ctx, 0, ts, true)
-	if err != nil {
-		return err
-	}
+	//if err := cst.FlushValidationCache(ctx); err != nil {
+	//return xerrors.Errorf("flushing validation cache failed: %w", err)
+	//}
```
magik6k, line 549: Pretty sure we want this

```suggestion
	if err := cst.FlushValidationCache(ctx); err != nil {
		return xerrors.Errorf("flushing validation cache failed: %w", err)
	}
```

(If we want to make it faster, this technically could run in parallel with the import)

cmd/lotus/daemon.go
```diff
@@ -127,6 +130,10 @@ var DaemonCmd = &cli.Command{
 			Name:  "lite",
 			Usage: "start lotus in lite mode",
 		},
+		&cli.BoolFlag{
+			Name:  "follower",
+			Usage: "start lotus in follower mode",
```
magik6k, line 135:
```suggestion
			Usage: "start lotus in experimental follower mode",
```

node/builder_chain.go
```diff
@@ -79,17 +79,36 @@ var ChainNode = Options(
...
 	// It will be called implicitly by the Syncer constructor.
 	Override(new(chain.SyncManagerCtor), func() chain.SyncManagerCtor { return chain.NewSyncManager }),
-	Override(new(*chain.Syncer), modules.NewSyncer),
+
+	ApplyIf(isNotFollower,
+		Override(new(*chain.Syncer), modules.NewSyncer),
+	),
+
+	ApplyIf(isFollower,
+		Override(new(*chain.Syncer), modules.NewSyncerFollower),
```
magik6k, line 109: In the follower mode we shouldn't need any of:
* `chain.SyncManagerCtor`
* `*chain.Syncer`
* `*hello.Service` / `RunHelloKey` / `HandleIncomingBlocksKey`
* `*peermgr.PeerMgr`
* `exchange.Server`
* Possibly should also somehow disable pubsub subscriptions

Techincally this is not critical for landing this PR in a very experimental form, but not running those services would be much more correct

node/config/def.go
```diff
@@ -117,6 +117,45 @@ func DefaultFullNode() *FullNode {
 	}
 }
 
+// DefaultFullNode returns the default config
+func DefaultFollowerNode() *Follower {
```
magik6k, line 121: Is this used? Follower node is still a "FullNode" node type, it shouldn't be a different type.

(FullNode should really be called ChainNode, but renaming that now is not really relevant to this PR)

magik6k (commented):
First pass.

I don't see the code which would subscribe to head changes on the node we're following, and tell the follower node ChainStore about new heads (by calling `ChainStore.PutTipSet`) - that bit of code definitely needs to exist, otherwise followers won't see the chain advance.

The follower nodes also shouldn't instantiate the default syncer - having it be there, even not started may mess with the node state - and we want to rely on the "leader" node for the sync state instead of doing the sync ourselves.

Stebalien:
Once we merge https://github.com/filecoin-project/filecoin-ffi/pull/408, I want to test again without a write buffer to see if we what the performance looks like. That's the one part I'm not super comfortable about.

blockstore/readonly.go
```diff
@@ -0,0 +1,62 @@
...
+
+func WithReadonly(base Blockstore) *ReadonlyBlockstore {
+	bs := &ReadonlyBlockstore{
+		base: base,
+	}
+	return bs
+}
+
+func ReadonlyError(description string) error {
+	return errors.New("Write protected access attempted on Readonly Blockstore from method " + description)
```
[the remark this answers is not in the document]
snissn: fixed 319fba4a8

blockstore/cassbs/cassandra_readonly_ds.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/query"
+)
+
+func ReadonlyError(description string) error {
+	return errors.New("Write protected access attempted on Readonly Blockstore from method " + description)
```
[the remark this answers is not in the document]
snissn: fixed cfa0d3df2

chain/sync.go
```diff
@@ -160,6 +160,37 @@ func NewSyncer(ds dtypes.MetadataDS,
 	return s, nil
 }
 
+// NewSyncerFollower creates a new Syncer object that the follower node uses
+func NewSyncerFollower(ds dtypes.MetadataDS,
```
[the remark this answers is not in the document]
snissn: This code never runs! Thanks! Deleted it 420c271f23673a29a93577ec344b82d50ed73fb8

cmd/lotus/daemon.go
```diff
@@ -300,7 +309,7 @@ var DaemonCmd = &cli.Command{
 		// If the daemon is started in "lite mode", provide a  Gateway
 		// for RPC calls
 		liteModeDeps := node.Options()
-		if isLite {
+		if isLite || isFollower {
```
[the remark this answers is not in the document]
snissn: eb8947c6b added a check here

cmd/lotus/daemon.go
snissn: Good catch, thanks

cmd/lotus/daemon.go
snissn: great idea

node/builder_chain.go
snissn: resolved in node/builder_chain.go

node/config/def.go
snissn: not used, thanks, deleted it aa059e64c

snissn:
> I don't see the code which would subscribe to head changes on the node we're following, and tell the follower node ChainStore about new heads (by calling `ChainStore.PutTipSet`) - that bit of code definitely needs to exist, otherwise followers won't see the chain advance.
> 
> The follower nodes also shouldn't instantiate the default syncer - having it be there, even not started may mess with the node state - and we want to rely on the "leader" node for the sync state instead of doing the sync ourselves.

I think that this is resolved by using yugabyte db for storing the metadata, chain and state

blockstore/cached.go
```diff
@@ -0,0 +1,172 @@
...
+		bs.writeLk.Lock()
+		for _, blk := range toPut {
+			delete(bs.pendingWrites, blk.Cid())
+		}
+		bs.writeLk.Unlock()
+
+		if err != nil {
+			log.Error("failed to write to disk", err)
+		}
+	}()
```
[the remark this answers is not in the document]
snissn: yes i agree! I found that while this can work when the database is sufficiently fast enough, when the servers are underpowered, it can be tweaked. I'll add this to a future work list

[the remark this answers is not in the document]
snissn: added here, thanks! ec1b99306

blockstore/cassbs/cassandra_ds.go
```diff
@@ -0,0 +1,253 @@
...
+	batch     *gocql.Batch
+	wg        sync.WaitGroup
+	nameSpace string
+}
+
+func (c *cassandraBatch) Put(ctx context.Context, key datastore.Key, value []byte) error {
+	statement := fmt.Sprintf("UPDATE %s.%s SET value = ? WHERE key = ?", c.nameSpace, c.nameSpace)
+	c.batch.Query(statement, value, toCasKey(key))
+
+	if c.batch.Size() >= 64 {
```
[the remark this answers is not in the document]
snissn: While the size here seemed to be working well enough for me, it is definitely worth looking at later.

blockstore/readonly.go
```diff
@@ -0,0 +1,62 @@
...
+	return bs.base.Get(ctx, c)
+}
+func (bs *ReadonlyBlockstore) GetSize(ctx context.Context, c cid.Cid) (int, error) {
+	return bs.base.GetSize(ctx, c)
+}
+func (bs *ReadonlyBlockstore) Has(ctx context.Context, c cid.Cid) (bool, error) {
+	return bs.base.Has(ctx, c)
+}
+func (bs *ReadonlyBlockstore) HashOnRead(hor bool) {
+	log.Warnf("called HashOnRead on Readonly blockstore; function not supported; ignoring")
```
magik6k, line 61: Also ok to just leave this method empty, this is a trusted datastore
```suggestion
```

cmd/lotus/daemon.go
```diff
@@ -187,6 +194,15 @@ var DaemonCmd = &cli.Command{
 	Action: func(cctx *cli.Context) error {
 		isLite := cctx.Bool("lite")
 
+		isFollower := cctx.Bool("follower")
+		if isFollower {
+			if os.Getenv("LOTUS_CASSANDRA_UNIVERSAL_STORE") == "" {
+				return xerrors.Errorf("Follower nodes must have LOTUS_CASSANDRA_UNIVERSAL_STORE set to connect to Cassandra db campatible database.")
```
magik6k, line 200:
```suggestion
				return xerrors.Errorf("follower nodes must have LOTUS_CASSANDRA_UNIVERSAL_STORE set to connect to Cassandra db campatible database")
```

node/config/types.go
```diff
@@ -31,6 +31,18 @@ type FullNode struct {
...
+	Common
+	Client     Client
+	Wallet     Wallet
+	Fees       FeeConfig
+	Chainstore Chainstore
+	Cluster    UserRaftConfig
+	Fevm       FevmConfig
+	Index      IndexConfig
+}
+
```
magik6k, line 45: This shouldn't be needed, we're still starting a chain node (which in the code is called a "full node", but that's really any chain node)

magik6k:
> I think that this is resolved by using yugabyte db for storing the metadata, chain and state

No, it's not - the chainstore has a number of in-memory references, that aren't updated to match the on-disk datastore.

Even the basic RPCs, like `ChainHead` read from an in-memory reference stored in the chainstore: https://github.com/filecoin-project/lotus/blob/b0d1385273557e1842474adc445a253ece959286/chain/store/store.go#L932-L936

The main "source of truth" for every chain node on what the current chain head is, is the [`heaviest` field in the ChainStore](https://github.com/filecoin-project/lotus/blob/b0d1385273557e1842474adc445a253ece959286/chain/store/store.go#L109). That field in loaded from the datastore (which in the follower node scenario comes from Cassandra), the loading happens in [`ChainStore.loadHead`](https://github.com/filecoin-project/lotus/blob/b0d1385273557e1842474adc445a253ece959286/chain/store/store.go#L219) - but that only happens once on startup. During node runtime the head in updated in [`ChainStore.takeHeaviestTipSet`](https://github.com/filecoin-project/lotus/blob/b0d1385273557e1842474adc445a253ece959286/chain/store/store.go#L644), which is called in [`MaybeTakeHeavierTipSet`](https://github.com/filecoin-project/lotus/blob/b0d1385273557e1842474adc445a253ece959286/chain/store/store.go#L412)

If you don't call MaybeTakeHeavierTipSet on the follower nodes with new head updates from the leader node, the chain view of the follower nodes will never advance, calls like `lotus chain list` will only update after restarting the follower node, and internal subsystems on the follower node will not be getting notifications about new tipsets (tho in the case of follower nodes this "only" impacts ChainNotify + some ETH APIs)

node/config/types.go
snissn: done

magik6k:
> implementing a publish/subscribe api between the leader node and the follower node such that when the leader updates the chain head, it publishes a message that the followers are subscribed to who then call [ChainStore.loadHead](https://github.com/filecoin-project/lotus/blob/b0d1385273557e1842474adc445a253ece959286/chain/store/store.go#L219) to update things

You don't need any new API, and you don't need to call loadHead.

All that needs to happen is in `modules.NewSyncerFollower`:
* You want to get the instance of leader RPC API - you do that by simply adding a `api.Gateway` parameter to that function
  * api.Gateway is injected into the DI container [here](https://github.com/filecoin-project/lotus/blob/c59cbd5a9c6aafb48dcb064ac91446e22d0ffbd3/cmd/lotus/daemon.go#L349) in the daemon command when in lite/follower modes
* Then you start a goroutine which subscribes to head changes on the leader with `ChainNotify`
* And for every head change you simply call [MaybeTakeHeavierTipSet](https://github.com/filecoin-project/lotus/blob/b0d1385273557e1842474adc445a253ece959286/chain/store/store.go#L412) on the chanistore
  * There should be no way those head changes would get rejected, if the leader accepted the head using this logic, followers will also take that head
  * This way all internal reorg-related logic will run correctly. Simply calling `loadHead` wouldn't do that, which would lead to things being very broken

node/modules/chain.go
```diff
@@ -175,6 +231,68 @@ func NewSyncer(params SyncerParams) (*chain.Syncer, error) {
...
 }
 
+func NewSyncerFollower(params SyncerFollowerParams) (*chain.Syncer, error) {
+	var (
+		g  = params.GatewayAPI
+		lc = params.Lifecycle
+		sm = params.StateManager
+	)
+	lc.Append(fx.Hook{
+		OnStart: func(ctx context.Context) error {
```
magik6k, line 241: The OnStart context is cancelled after the startup sequence, so you really don't want to use that one.

The correct way to get a context in constructors is with `ctx := helpers.LifececleCtx(mctx, lc)`, where mctx is iirc `dtypes.MetrictCtx` (metricsCtx is a special root context that never gets cancelled, but carries values used for collecting opencensus metrics/tracing, LifececleCtx connects it to a lifecycle which ensures correct order for stopping components - so the context you get from there will be cancelled after everything else that depends on the syncer has stopped)

node/modules/chain.go
```diff
@@ -175,6 +231,68 @@ func NewSyncer(params SyncerParams) (*chain.Syncer, error) {
...
+					head, ok := <-headChanges
+					if !ok {
+
+						//todo can we auto reconnect to the gateway?
+						//if the gateway disconnects is there a way to retry?
+						//panic is ok response because infrastructure should reboot rpc node upon error
+						// this will happen when the leader is restarted
+
+						log.Error("Notify stream was invalid")
+						panic("Notify stream was invalid")
```
magik6k, line 259: Yes, just call `g.ChainNotify(ctx)` in a loop with some sleep. go-jsonrpc will handle reconnecting under the hood, but the channel gets closed when the websocket jsonrpc connection drops.

Here's a good example - https://github.com/filecoin-project/lotus/blob/0374c5072804f5d8c79236944ef468af75f9627c/node/health.go#L48-L73

blockstore/cassbs/cassandra_ds.go
```diff
@@ -0,0 +1,258 @@
...
+func (c *cassandraBatch) Put(ctx context.Context, key datastore.Key, value []byte) error {
+	statement := fmt.Sprintf("UPDATE %s.%s SET value = ? WHERE key = ?", c.nameSpace, c.nameSpace)
+	c.batch.Query(statement, value, toCasKey(key))
+
+	// todo setting a batch size of 1 violates the whole point of the batch writes
+	// however with batch writes, there is some sort of race condition where the leader
+	// will update the chain head and notify the followers before the data has all been commited to the db
+	// this means that the followers will be behind because they do not see the latest chain head when the
+	// hello service notifies them of a new block
+	if c.batch.Size() >= 1 {
```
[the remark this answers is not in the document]
magik6k: The Commit method below should be called from the Flush call in block state compute, probably good first step is to trace the flush call in a debugger to see if it actually calls Commit

(This Flush - https://github.com/filecoin-project/lotus/blob/0374c5072804f5d8c79236944ef468af75f9627c/chain/consensus/compute_state.go#L295)

rjan90:
Hey! 👋

rjan90:

Closing

PR #10832, opened 2023-05-04
chore: deps: update go-jsonrpc for batch operations

Stebalien:
## Related Issues

fixes https://github.com/filecoin-project/ref-fvm/issues/1047

## Proposed Changes

This updates to the latest go-jsonrpc release that supports batch operations. Basically, in addition to individual requests, a client can send an array of requests and receive back an array of responses.

This should have no additional security implications as:

1. Any timeouts apply to the entire batch.
2. The max request size is unchanged.

The FVM team would like this as some Ethereum tools expect bulk json-rpc support.


magik6k:
Pulled in https://github.com/filecoin-project/lotus/pull/10845

PR #10843, opened 2023-05-08
feat: deflake paych_api_test

arajasek:
## Related Issues

This test has been observed to flake a lot, example https://app.circleci.com/pipelines/github/filecoin-project/lotus/28643/workflows/32a2f145-5cac-492b-a00e-3ce566d50691/jobs/965295.

The problem is that we try to inject a lot of nulls in order to fast-forward the Paych Settle Delay (which is 12 hours on mainnet). This is nice to speed up the test, but just leads to missed PoSts, especially because we fast-forward entire deadlines at a time.

## Proposed Changes

Just wait for the PaychSettleDelay to "normally" pass -- this is only 14 seconds, we don't need to risk flakiness. We could consider skipping fewer epochs at a time, but I honestly don't think it's worth it.

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/paych_api_test.go
+++ b/itests/paych_api_test.go
@@ -10,7 +10,6 @@ import (
 	cbor "github.com/ipfs/go-ipld-cbor"
 	"github.com/stretchr/testify/require"
 
-	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
 	paychtypes "github.com/filecoin-project/go-state-types/builtin/v8/paych"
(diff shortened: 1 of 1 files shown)


itests/paych_api_test.go
```diff
@@ -197,7 +194,11 @@ func TestPaymentChannelsAPI(t *testing.T) {
 	require.Errorf(t, err, "Expected shortfall error of %d", excessAmt)
 
 	// wait for the settlement period to pass before collecting
-	waitForBlocks(ctx, t, bm, paymentReceiver, receiverAddr, policy.PaychSettleDelay)
+	head, err := paymentReceiver.ChainHead(ctx)
+	require.NoError(t, err)
+
+	// it doesn't matter which node we "wait" on
+	paymentReceiver.WaitTillChain(ctx, kit.HeightAtLeast(head.Height()+policy.PaychSettleDelay+5))
```
ZenGround0, line 201: what's up with `5`?

ZenGround0 (approved):

itests/paych_api_test.go
magik6k: Making sure that all nodes are at that height? Probably would be best to do as a helper, `kit.WaitTillChainAll..` or something like that

arajasek: @ZenGround0 Just a little extra buffer, not needed
@magik6k Yeah, good suggestion, will do

PR #10854, opened 2023-05-10
Revert "Revert "libp2p: add QUIC and WebTransport to default listen addresses""

magik6k:
Reverts filecoin-project/lotus#10852, fixes tests

### DIFF
--- a/documentation/en/default-lotus-config.toml
+++ b/documentation/en/default-lotus-config.toml
@@ -38,7 +38,7 @@
   #
   # type: []string
   # env var: LOTUS_LIBP2P_LISTENADDRESSES
-  #ListenAddresses = ["/ip4/0.0.0.0/tcp/0", "/ip6/::/tcp/0"]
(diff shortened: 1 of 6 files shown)


marten-seemann:
The error comes from quic-go: https://github.com/quic-go/quic-go/blob/v0.34.0/send_stream.go#L404-L407
It happens when you first reset the stream, then call close. In most circumstances, it can be ignored. Is there any reason you're surfacing this kind of error?

arajasek:
> The error comes from quic-go: https://github.com/quic-go/quic-go/blob/v0.34.0/send_stream.go#L404-L407 It happens when you first reset the stream, then call close. In most circumstances, it can be ignored. Is there any reason you're surfacing this kind of error?

It looks this error makes the deal pipeline think the deal has failed. I'd have to dig to see where that happens (and how we could ignore it, assuming we should be doing so).

arajasek (approved):

marten-seemann:
> > The error comes from quic-go: https://github.com/quic-go/quic-go/blob/v0.34.0/send_stream.go#L404-L407 It happens when you first reset the stream, then call close. In most circumstances, it can be ignored. Is there any reason you're surfacing this kind of error?
> 
> It looks this error makes the deal pipeline think the deal has failed. I'd have to dig to see where that happens (and how we could ignore it, assuming we should be doing so).

@arajasek What did you digging reveal?

Why was this change silently reverted?

PR #10855, opened 2023-05-10
fix: splitstore: remove deadlock around waiting for sync

arajasek:
## Related Issues

#10788?

#10641 introduced the possibility of a deadlock, because `Put` might sleep waiting on sync to catch up while holding the all-important `txnLk`. This is definitely wrong, we also just don't want `Put`s to be yielding and sleeping.

## Proposed Changes

Add a non-yielding version of `walkObjectIncomplete` that gets called by `Put` and `PutMany`.

## Additional Info

I checked thoroughly, but am still not _confident_ that other methods more directly involved in compaction don't have the same problem. This entire locking mechanism is...tricky.

## Checklist

- [ ] CI is green

### DIFF
--- a/blockstore/splitstore/splitstore_compact.go
+++ b/blockstore/splitstore/splitstore_compact.go
@@ -254,7 +254,7 @@ func (s *SplitStore) markLiveRefs(cids []cid.Cid) {
 			func(missing cid.Cid) error {
 				log.Warnf("missing object reference %s in %s", missing, c)
 				return errStopWalk
-			})
+			}, false)
 	}
 
 	// optimize the common case of single put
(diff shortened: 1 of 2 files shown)


blockstore/splitstore/splitstore_compact.go
```diff
@@ -1222,8 +1222,10 @@ func (s *SplitStore) walkObjectIncomplete(c cid.Cid, visitor ObjectVisitor, f, m
 	}
 
 	// check this before recursing
-	if err := s.checkYield(); err != nil {
-		return sz, err
+	if yielding {
+		if err := s.checkYield(); err != nil {
```
ZenGround0, line 1226: Another approach that is a bit cleaner is to change this (and maybe `walkObject`) to always call `checkClosing`, then do a `checkYield` in every call to [`walkBlock`](https://github.com/filecoin-project/lotus/pull/10855/files#diff-f39742667127f7440576464ad7286c341abd00ec405df1612a8bcceb6b8881abR976).  Its probably fine for sync protection to just do yielding once for every block walked during compaction.

ZenGround0 (approved):

PR #10858, opened 2023-05-11
feat: deflake sector_import_simple

arajasek:
## Related Issues

This test has been observed to flake, eg. [here](https://app.circleci.com/pipelines/github/filecoin-project/lotus/28749/workflows/32662994-991f-47d6-b7bc-6c3af046edbb/jobs/971223), due to missed PoSts.

## Proposed Changes

Use 2 miners instead of one for this test. We made the identical change to sector_import_full_test.go in #10764, and it hasn't flaked once since then (afaic tell).

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/sector_import_simple_test.go
+++ b/itests/sector_import_simple_test.go
@@ -39,7 +39,9 @@ func TestSectorImportAfterPC2(t *testing.T) {
 	////////
 	// Start a miner node
 
-	client, miner, ens := kit.EnsembleMinimal(t, kit.ThroughRPC())
+	// We use two miners so that in case the actively tested miner misses PoSt, we still have a blockchain
+	client, miner, _, ens := kit.EnsembleOneTwo(t, kit.ThroughRPC())
+
 	ens.InterconnectAll().BeginMining(blockTime)
 
 	ctx := context.Background()


ZenGround0 (approved):

itests/sector_import_simple_test.go
```diff
@@ -39,7 +39,9 @@ func TestSectorImportAfterPC2(t *testing.T) {
 	////////
 	// Start a miner node
 
-	client, miner, ens := kit.EnsembleMinimal(t, kit.ThroughRPC())
+	// We use two miners so that in case the actively tested miner misses PoSt, we still have a blockchain
+	client, miner, _, ens := kit.EnsembleOneTwo(t, kit.ThroughRPC())
+
 	ens.InterconnectAll().BeginMining(blockTime)
```
magik6k, line 45: Actually in this case BeginMiningMustPoSt should help. What I think may be happening here is that we are trying to produce the post snark, but that is blocked while the sector is Committing - in which case EnsembleOneTwo may still not be robust enough, because Committing even for 2k sectors takes a few seconds, and takes the global gpu lock, which will prevent any post from being created

arajasek: We've had _pretty_ good luck since we turned the other test to use 2 miners. I'm gonna merge this for now, but will definitely try `MustPost`ing if the issue recurs.

PR #10833, opened 2023-05-05
perf: mpool: Implement transactionLk to ensure consistent state access in MessagePool

snissn:
This commit introduces a new lock transactionLk to the MessagePool struct in order to provide a consistent view of the state and protect from state changes. 

The primary objective of transactionLk is to ensure there is at most a single concurrent writer of message pool state. By holding this lock, necessary computations can be performed before making the actual state changes under the existing curTsLk. Meanwhile, other methods like GetStateNonce can still acquire read locks on curTsLk, ensuring a consistent view of the state. This approach allows writers, such as the addTs function, to rely on the transactionLk to maintain the exclusivity of a single concurrent writer during their execution.

It is worth determining if mp.transactionLk  makes mp.Lk redundant and can be removed. However we may need both as mp.Lk has a comment indicating it needs to be inside curTsLk and transactionLk has to be used outside of curTsLk

Key changes include:

- Add transactionLk to the MessagePool struct.
- Acquire and release transactionLk in functions like addTs, HeadChange, pruneExcessMessages, republishPendingMessages, and SelectMessages.

## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #10865, opened 2023-05-12
perf: mpool: Implement transactionLk to ensure consistent state access in MessagePool

snissn:
Title: Refine MessagePool Locking Mechanism to Optimize RPC Calls Performance

Description:
This commit introduces significant refinements to the locking mechanism in the MessagePool. The primary objective is to reduce the duration that locks impede RPC calls from accessing data, thereby enhancing the overall performance of various MessagePool operations.

Key Changes include:

1. Introduction of a new lock, transactionLk. This lock plays a crucial role in improving concurrency control in the MessagePool. It works in conjunction with the existing lock (previously known as curTsLk) to ensure proper synchronization when accessing both pending messages and current tipset related data.

2. The existing lock, curTsLk, has been renamed to stateLk, to better reflect its purpose. This lock's usage has been refined to optimize its performance.

3. Locking and unlocking procedures have been updated in several functions, including but not limited to CheckMessages, CheckPendingMessages, CheckReplaceMessages, checkMessages, New, TryForEachPendingMessage, addTs, addLoaded, addSkipChecks, GetNonce, Pending, PendingFor, HeadChange, Clear, pruneExcessMessages, and republishPendingMessages.

The integration of these changes aims to improve the efficiency of the MessagePool, particularly in terms of handling RPC calls, and to boost the overall performance of the system.

## Related Issues
squashed from https://github.com/filecoin-project/lotus/pull/10833

## Checklist

- [ ] CI is green


chain/messagepool/check.go
```diff
@@ -32,7 +32,10 @@ func (mp *MessagePool) CheckMessages(ctx context.Context, protos []*api.MessageP
 // CheckPendingMessages performs a set of logical sets for all messages pending from a given actor
 func (mp *MessagePool) CheckPendingMessages(ctx context.Context, from address.Address) ([][]api.MessageCheckStatus, error) {
 	var msgs []*types.Message
-	mp.lk.RLock()
```
fridrik01, line 35: Can we remove the `lk` lock from the MessagePool struct, or is it still used somewhere?

chain/messagepool/messagepool.go
```diff
@@ -139,8 +139,24 @@ type MessagePool struct {
...
 
-	curTsLk sync.RWMutex // DO NOT LOCK INSIDE lk
-	curTs   *types.TipSet
+	// curTs stands for 'current TipSet'
+	curTs *types.TipSet
+
+	// The transactionLk lock is used to ensure consistency when transacting between the TipSet (TS) and the Message Pool.
+	// Any modification to curTs must be wrapped in a transactionLk.
+	// This lock is not needed if you're only interested in a consistent view of the state without changing it.
+	transactionLk sync.RWMutex
```
fridrik01, line 148: I believe this lock can be changed in to a `sync.Mutex` since the `RLock()` is never used?

chain/messagepool/messagepool.go
```diff
@@ -139,8 +139,24 @@ type MessagePool struct {
...
+	// The stateLk lock is used when you need to interact with the state within the Message Pool.
+	// If the state is being modified, use the Write form of this mutex.
+	// Ensure that the transactionLk lock is acquired before the stateLk lock to maintain consistency.
+	// Example usage:
+	// mp.transactionLk.Lock()
+	// defer mp.transactionLk.Unlock()
+	//
+	// mp.stateLk.Lock()
+	// defer mp.stateLk.Unlock()
+	stateLk sync.RWMutex
```
fridrik01, line 159: It would simplify the code a lot IMO if we could merge the transactionLk/stateLk into one lock or lock object, is that something that you considered?

chain/messagepool/messagepool.go
```diff
@@ -755,42 +770,13 @@ func (mp *MessagePool) Add(ctx context.Context, m *types.SignedMessage) error {
...
-		mp.curTsLk.Unlock()
-		_, _ = mp.api.GetActorAfter(m.Message.From, tmpCurTs)
-		_, _ = mp.getStateNonce(ctx, m.Message.From, tmpCurTs)
-		mp.curTsLk.Lock()
-		//now that we have the lock, we continue, we could do this as a loop forever, but that's bad to loop forever, and this was added as an optimization and it seems once is enough because the computation < block time
-	}
-
-	defer mp.curTsLk.Unlock()
-
-	_, err = mp.addTs(ctx, m, mp.curTs, false, false)
```
fridrik01, line 793: Very nice

chain/messagepool/messagepool.go
```diff
@@ -863,7 +849,21 @@ func (mp *MessagePool) checkBalance(ctx context.Context, m *types.SignedMessage,
 	return nil
 }
 
-func (mp *MessagePool) addTs(ctx context.Context, m *types.SignedMessage, curTs *types.TipSet, local, untrusted bool) (bool, error) {
+func (mp *MessagePool) addTs(ctx context.Context, m *types.SignedMessage, local, untrusted bool) (bool, error) {
+
```
fridrik01, line 853:
```suggestion
```

chain/messagepool/messagepool.go
snissn: thank you! resolved and deleted

chain/messagepool/check.go
snissn: Yes!! We can! Deleted lk in b0f3311d8!

chain/messagepool/messagepool.go
arajasek: @snissn What was the resolution to this?

chain/messagepool/messagepool.go
```diff
@@ -449,8 +470,8 @@ func New(ctx context.Context, api Provider, ds dtypes.MetadataDS, us stmgr.Upgra
 }
 
 func (mp *MessagePool) ForEachPendingMessage(f func(cid.Cid) error) error {
-	mp.lk.Lock()
-	defer mp.lk.Unlock()
+	mp.transactionLk.Lock()
```
arajasek, line 473: I think it's sufficient to Lock mp.stateLk here, since we just need `pending` to not change under our feet?

chain/messagepool/messagepool.go
```diff
@@ -863,7 +847,20 @@ func (mp *MessagePool) checkBalance(ctx context.Context, m *types.SignedMessage,
...
 }
 
-func (mp *MessagePool) addTs(ctx context.Context, m *types.SignedMessage, curTs *types.TipSet, local, untrusted bool) (bool, error) {
+func (mp *MessagePool) addTs(ctx context.Context, m *types.SignedMessage, local, untrusted bool) (bool, error) {
+	//ensures that we have a consistent view of the state
+	mp.transactionLk.Lock()
+	defer mp.transactionLk.Unlock()
+
+	mp.stateLk.RLock()
+	err := mp.checkMessage(ctx, m)
```
arajasek, line 856: I think we can refactor `checkMessage` to not need to reference `curTs` at all -- I can push up a commit that does that, and I think in that case we can change this to only take the `stateLk` _after_ having called `checkMessage`.

chain/messagepool/messagepool.go
```diff
@@ -863,7 +847,20 @@ func (mp *MessagePool) checkBalance(ctx context.Context, m *types.SignedMessage,
 	return nil
 }
 
-func (mp *MessagePool) addTs(ctx context.Context, m *types.SignedMessage, curTs *types.TipSet, local, untrusted bool) (bool, error) {
+func (mp *MessagePool) addTs(ctx context.Context, m *types.SignedMessage, local, untrusted bool) (bool, error) {
+	//ensures that we have a consistent view of the state
+	mp.transactionLk.Lock()
```
arajasek, line 852: It's not immediately clear to me why we need the `transactionLk` here, and in some of the subsequent methods, since we aren't modifying `curTs`. I can totally believe it's necessary, but can you please talk me through it?

chain/messagepool/messagepool.go
```diff
@@ -1256,9 +1235,7 @@ func (mp *MessagePool) pendingFor(ctx context.Context, a address.Address) []*typ
 	return mset.toSlice()
 }
 
-func (mp *MessagePool) HeadChange(ctx context.Context, revert []*types.TipSet, apply []*types.TipSet) error {
-	mp.curTsLk.Lock()
-	defer mp.curTsLk.Unlock()
+func (mp *MessagePool) headChange(ctx context.Context, revert []*types.TipSet, apply []*types.TipSet) error {
```
arajasek, line 1238: Please make a note that caller must hold `transactionLk` and `stateLk`

chain/messagepool/pruning.go
```diff
@@ -14,12 +14,13 @@ import (
 )
 
 func (mp *MessagePool) pruneExcessMessages() error {
-	mp.curTsLk.Lock()
-	ts := mp.curTs
-	mp.curTsLk.Unlock()
+	mp.transactionLk.Lock()
```
arajasek, line 17: Also not clear to me that we need `transactionLk` here, but unsure

arajasek (commented):
Few more questions after earlier review in #10833

chain/messagepool/messagepool.go
snissn: I think this was commented on before @fridrik01 got further into reading the PR, and there was a subsequent comment from him clarifying. It is necessary to have two locks for this PR.

chain/messagepool/messagepool.go
snissn: That sounds great to me! Have you gotten to that yet? Should we set up a new ticket for that?

chain/messagepool/messagepool.go
snissn: Hey! That's a good question! When the tip set changes, the message pool is pruned to remove any messages that were in the blocks:

```
// caller must hold transactionLk and stateLk
func (mp *MessagePool) headChange(ctx context.Context, revert []*types.TipSet, apply []*types.TipSet) error {

  repubTrigger := false
  rmsgs := make(map[address.Address]map[uint64]*types.SignedMessage)
  add := func(m *types.SignedMessage) {
    s, ok := rmsgs[m.Message.From]
    if !ok {
      s = make(map[uint64]*types.SignedMessage)
      rmsgs[m.Message.From] = s
    }
    s[m.Message.Nonce] = m
  }
  rm := func(from address.Address, nonce uint64) {
    s, ok := rmsgs[from]
    if !ok {
      mp.remove(ctx, from, nonce, true)
      return
    }

    if _, ok := s[nonce]; ok {
      delete(s, nonce)
      return
    }

    mp.remove(ctx, from, nonce, true)
  }
...
  for _, ts := range apply {
    mp.curTs = ts

    for _, b := range ts.Blocks() {
      bmsgs, smsgs, err := mp.api.MessagesForBlock(ctx, b)
      if err != nil {
        xerr := xerrors.Errorf("failed to get messages for apply block %s(height %d) (msgroot = %s): %w", b.Cid(), b.Height, b.Messages, err)
        log.Errorf("error retrieving messages for block: %s", xerr)
        merr = multierror.Append(merr, xerr)
        continue
      }

      for _, msg := range smsgs {
        rm(msg.Message.From, msg.Message.Nonce)
        maybeRepub(msg.Cid())
      }

      for _, msg := range bmsgs {
        rm(msg.From, msg.Nonce)
        maybeRepub(msg.Cid())
      }
    }
  }
```
we want to wrap changes to tipset and/or messages in a pair of transaction and state locks because changing the message pool requires a coordination lock on the tipset because the tipset changes the message pool

chain/messagepool/pruning.go
snissn: same as above - we are changing the message pool so we want to use the transaction and state locks because we are changing either the tipset or the messages

chain/messagepool/messagepool.go
snissn: Transaction lock needs to be used to safely make changes to state. So changes to the state must be wrapped in a transaction lock. Hope that makes sense! Also the protector is who calls this function so I am pretty sure it wants write access not read access, so i used a write lock here

chain/messagepool/check.go
```diff
@@ -32,10 +32,12 @@ func (mp *MessagePool) CheckMessages(ctx context.Context, protos []*api.MessageP
...
+
+	mp.stateLk.RLock()
+	defer mp.stateLk.RUnlock()
```
Stebalien, line 37: This locks a much _wider_ range. Are we sure that's a good idea?

Stebalien (commented):
While this does simplify things, it also significantly increases the granularity of some of the locks. Are we sure it improves performance?

rjan90:
Hey! 👋

rjan90:

Closing

PR #10866, opened 2023-05-12
feat: chainstore: remove locking in index cache

arajasek:
## Related Issues

#10788?

Our friends working with Lotus shared 8 goroutine dumps from a node that was stuck syncing. Contention over the index cache's lock was a common theme.

## Proposed Changes

Use a threadsafe cache instead of a mutex-map here.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/store/index.go
+++ b/chain/store/index.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"os"
 	"strconv"
-	"sync"
 
+	lru "github.com/hashicorp/golang-lru/v2"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-state-types/abi"
@@ -27,8 +27,7 @@ func init() {
 }
 
 type ChainIndex struct {
-	indexCacheLk sync.Mutex
-	indexCache   map[types.TipSetKey]*lbEntry
+	indexCache *lru.ARCCache[types.TipSetKey, *lbEntry]
 
 	loadTipSet loadTipSetFunc
 
@@ -37,8 +36,9 @@ type ChainIndex struct {
 type loadTipSetFunc func(context.Context, types.TipSetKey) (*types.TipSet, error)
 
 func NewChainIndex(lts loadTipSetFunc) *ChainIndex {
+	sc, _ := lru.NewARC[types.TipSetKey, *lbEntry](DefaultChainIndexCacheSize)
 	return &ChainIndex{
-		indexCache: make(map[types.TipSetKey]*lbEntry, DefaultChainIndexCacheSize),
+		indexCache: sc,
 		loadTipSet: lts,
 		skipLength: 20,
 	}
@@ -59,11 +59,9 @@ func (ci *ChainIndex) GetTipsetByHeight(ctx context.Context, from *types.TipSet,
 		return nil, xerrors.Errorf("failed to round down: %w", err)
 	}
 
-	ci.indexCacheLk.Lock()
-	defer ci.indexCacheLk.Unlock()
 	cur := rounded.Key()
 	for {
-		lbe, ok := ci.indexCache[cur]
+		lbe, ok := ci.indexCache.Get(cur)
 		if !ok {
 			fc, err := ci.fillCache(ctx, cur)
 			if err != nil {
@@ -137,7 +135,7 @@ func (ci *ChainIndex) fillCache(ctx context.Context, tsk types.TipSetKey) (*lbEn
 		targetHeight: skipTarget.Height(),
 		target:       skipTarget.Key(),
 	}
-	ci.indexCache[tsk] = lbe
+	ci.indexCache.Add(tsk, lbe)
 
 	return lbe, nil
 }


fridrik01 (approved):
LGTM

chain/store/index.go
```diff
@@ -59,11 +59,9 @@ func (ci *ChainIndex) GetTipsetByHeight(ctx context.Context, from *types.TipSet,
 		return nil, xerrors.Errorf("failed to round down: %w", err)
 	}
 
-	ci.indexCacheLk.Lock()
```
snissn, line 62: do we need to have a transactional / consistent view of indexCache during this loop? Can there be other actors changing indexCache while this loop is running that will create unintended behavior?

fillCache has a comment that we want to be holding indexCacheLk during the call -- https://github.com/filecoin-project/lotus/blob/master/chain/store/index.go#L99 

There's a lot going on in these loops. I think we just need to make sure that there's always a consistent view of the data

snissn:
if we want to merge this, we should remove the comment here:

https://github.com/filecoin-project/lotus/pull/10866/files#diff-806866d49e06f388f52056a4cce962c76af2eb7218913aa473da3c15fcd65e18R97 
```
// Caller must hold indexCacheLk
func (ci *ChainIndex) fillCache(ctx context.Context, tsk types.TipSetKey) (*lbEntry, error) {
```

also we need to make sure that this change is safe based on the expectations of transactional consistency that was previously in 

GetTipsetByHeight

we may be needed to have a lk to ensure we need a transactional view of the state, and that lock may be named indexCacheLk but it may want to be renamed to indicate the transaction needed. 

Also I haven't fully read the code yet so lmk if anything here can be clarified on!

chain/store/index.go
```diff
@@ -27,8 +27,7 @@ func init() {
 }
 
 type ChainIndex struct {
-	indexCacheLk sync.Mutex
-	indexCache   map[types.TipSetKey]*lbEntry
+	indexCache *lru.ARCCache[types.TipSetKey, *lbEntry]
```
shrenujbansal, line 30: We initially had an ARCCache here which was dropped in favor of a simple map with a mutex for the reasons pointed here: https://github.com/filecoin-project/lotus/pull/10423#pullrequestreview-1331509633
I think we can keep the mutex but reduce the size of critical section covered by the mutex

shrenujbansal:
So this PR is essentially reducing the critical section of the chain index lock from https://github.com/filecoin-project/lotus/blob/90c89288b62c3343e4885ac1dfb8498fbd863fea/chain/store/index.go#L64-L93 to the individual cache operations.

However, with this change, in the event that several processes call GetTipsetByHeight simultaneously, all these processes will run the fillCache function https://github.com/filecoin-project/lotus/blob/90c89288b62c3343e4885ac1dfb8498fbd863fea/chain/store/index.go#L100  which will be wasteful.
I'm not sure if this such a case happens often, but if it does, it makes sense to keep the lock coverage until the end of fillCache

@magik6k Thoughts? Are the operations performed in fillCache relatively expensive?

snissn:
> However, with this change, in the event that several processes call GetTipsetByHeight simultaneously, all these processes will run the fillCache function

We could add a lock on each tipset and ensure that only one caller of fillCache is happening at a time by holding that lock

chain/store/index.go
Kubuxu: It could be replaced by a simple LRU cache if that is a concern.

Kubuxu:
I think the more correct but still not too complex thing to do here is to use a sharded mutex based on the tipset during fillCache and access to the cache.

Sharded Mutex is an array of predetermined size containing mutexes. When trying to lock the mutex by key, the key (tipset key in this case) is hashed using a [fast non-cryptographic hash function ](https://pkg.go.dev/hash/maphash#String) and the mutex at `digest % N_shards` is taken.

hash/maphash.Hash uses runtime seed, this seed should be the same across invocations. [maphash.String](https://pkg.go.dev/hash/maphash#String) is preferred due to optimizations. The seed can be generated using [maphash.MakeSeed](https://pkg.go.dev/hash/maphash#MakeSeed).

Kubuxu:
@arajasek can you share these goroutine dumps with me?

rjan90:
> @arajasek can you share these goroutine dumps with me?

@Kubuxu Here are the goroutine dumps:

[lpgr_node1.txt](https://github.com/filecoin-project/lotus/files/11488326/lpgr_node1.txt)
[lpgr_node2.txt](https://github.com/filecoin-project/lotus/files/11488327/lpgr_node2.txt)
[lpgr_node3.txt](https://github.com/filecoin-project/lotus/files/11488329/lpgr_node3.txt)
[lpgr_node4.txt](https://github.com/filecoin-project/lotus/files/11488330/lpgr_node4.txt)
[lpgr_node5.txt](https://github.com/filecoin-project/lotus/files/11488332/lpgr_node5.txt)
[lpgr_node6.txt](https://github.com/filecoin-project/lotus/files/11488333/lpgr_node6.txt)
[lpgr_node7.txt](https://github.com/filecoin-project/lotus/files/11488334/lpgr_node7.txt)
[lpgr_node8.txt](https://github.com/filecoin-project/lotus/files/11488335/lpgr_node8.txt)

jennijuju:
> @arajasek can you share these goroutine dumps with me?

I will on slack

snissn:
@Kubuxu I think you are describing a mutex like a "striped mutex" https://pkg.go.dev/github.com/nmvalera/striped-mutex not used this library before but worth trying before writing our own

Kubuxu:
Yeah, it is what I'm explaining. I haven't bothered to look.

Kubuxu:
For the infura goroutine dumps, we have discovered that there was a request made to g et block at height -76, and it was looking like the ChainIndex was making some progress.

shrenujbansal:
@Kubuxu Wouldn't a sharded mutex potentially result in the same problem as this?
For instance, you could have 2 threads which try to get tipsets 0 (all the way back) and 1,400,0000 (half way back).
In this case, both threads would end up doing the long walkback which we'd probably want to avoid

I would think an adhoc process which regularly populates the cache with the updated epochs might be more suitable in this case. Not sure how we'd implement it within the process though

Kubuxu:
> For instance, you could have 2 threads which try to get tipsets 0 (all the way back) and 1,400,0000 (half way back).
In this case, both threads would end up doing the long walkback which we'd probably want to avoid.

What we want to avoid is walking the 20 blocks needlessly. Only one of the threads can take a shard of a mutex for walking back the 20 blocks to build an index entry for the given tipset.

Let's say the first tipset not covered by the index is TipsetA. Both of them reach that point concurrently, they check the xsync/Map, no entry for TipsetA. Now they will attempt to lock shard number `hash(TipsetA) % N_shards`, one of them will win, this thread will start walking the 20 blocks to get from TipsetA to TipsetB at height `TipsetA.height - 20`. It will find TipsetB, put in in the cache, then release the lock and move on to checking the cache for TipsetB. This is when the second thread will acquire the lock, it will recheck the map, and notice the entry for TipsetA pointing towards TipsetB.
Both threads will check the cache for TipsetB and the process repeats.

Kubuxu:
> I would think an adhoc process which regularly populates the cache with the updated epochs might be more suitable in this case. Not sure how we'd implement it within the process though

The issue isn't recent epochs but far history.

arajasek:
Thanks for the discussion & work here everyone!

PR #10896, opened 2023-05-19
feat: chainstore: sharded mutex for filling chain height index

Kubuxu:
This PR introduces as sharded mutex within the ChainIndex#GetTipsetByHeight.
It also replaces a go map with xsync.Map which doesn't require locking.

The lock is taken when it appears that ChainIndex filling work should be
started. After claiming the lock, the status of the cache is rechecked,
if the entry is still missing, the fillCache is started.

Thanks to @snissn and @arajasek for debugging and taking initial stabs at this.

Supersedes #10866 and 10885

### DIFF
--- a/chain/store/index.go
+++ b/chain/store/index.go
@@ -2,18 +2,21 @@ package store
 
 import (
 	"context"
+	"hash/maphash"
 	"os"
 	"strconv"
-	"sync"
 
+	"github.com/puzpuzpuz/xsync/v2"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-state-types/abi"
 
 	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/lib/shardedmutex"
 )
 
-var DefaultChainIndexCacheSize = 32 << 15
+// DefaultChainIndexCacheSize no longer sets the maximum size, just the inital size of the map.
+var DefaultChainIndexCacheSize = 1 << 15
 
 func init() {
 	if s := os.Getenv("LOTUS_CHAIN_INDEX_CACHE"); s != "" {
@@ -27,8 +30,9 @@ func init() {
 }
 
 type ChainIndex struct {
-	indexCacheLk sync.Mutex
-	indexCache   map[types.TipSetKey]*lbEntry
+	indexCache *xsync.MapOf[types.TipSetKey, *lbEntry]
+
+	fillCacheLock shardedmutex.ShardedMutexFor[types.TipSetKey]
 
 	loadTipSet loadTipSetFunc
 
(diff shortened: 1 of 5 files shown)


snissn:
> // DefaultChainIndexCacheSize no longer sets the maximum size, just the initial size of the map.

are we ok with changing how the cache works? In this new PR we are letting the indexCache grow unbounded

Kubuxu:
It already became unloaded with the switch from LRU cache to map some time ago. I just documented it.

chain/store/index.go
```diff
@@ -59,17 +68,23 @@ func (ci *ChainIndex) GetTipsetByHeight(ctx context.Context, from *types.TipSet,
...
 	for {
-		lbe, ok := ci.indexCache[cur]
+		lbe, ok := ci.indexCache.Load(cur) // check the cache
 		if !ok {
-			fc, err := ci.fillCache(ctx, cur)
-			if err != nil {
-				return nil, xerrors.Errorf("failed to fill cache: %w", err)
+			lk := ci.fillCacheLock.GetLock(cur)
+			lk.Lock()                         // if entry is missing, take the lock
+			lbe, ok = ci.indexCache.Load(cur) // check if someone else added it while we waited for lock
```
magik6k, line 77: Why would we ever need more than one fillCacheLock? Shouldn't something like this be basically the same:

```
lbe, ok := indexCache.Load(cur)
if !ok {
  ci.fillCacheLk.Lock()
  lbe, ok = indexCache.Load(cur)
  if !ok {
    ...
  }

  ci.fillCacheLk.Unlock()
```

Kubuxu: The lock is shared, so filling the cache at epoch `current-1000000` (when the work started at `current - 40`, is not consistently blocking the progress of loading the cache at `current-20`.

I agree that with the lock re-shuffle and using `xsync.Map`, we could get away with one standard lock, but it also can create issues even with lock fairness guarantees. For exampleif 100threads try to fill the `current-1000000` and 1 thread trying to do `current -20` it will cause the `current-20` to progress very slowly due to the 100:1 lock fairness.

By using sharded lock there is only 1/32 chance that the near history will hit the old history lock. And even if it does, it will be for one 20-block walk cycle, on the next walk they have 1/1024 chance to collide again.

snissn:
> In this commit [90c8928](https://github.com/filecoin-project/lotus/commit/90c89288b62c3343e4885ac1dfb8498fbd863fea)

yes you are right! thank you for clarifying

chain/store/index.go
Kubuxu: indexCacheLk is not needed given sync safe map

> that will block if other cache fill is active, but that's fine

This is where we disagree. Only one active walk is a risky and suboptimal solution. We can get parallel walks with no drawback with minimal increase in complexity.

Let's take an Infura node as an example. I'm assuming they expose the API to the public.
As a malicious actor, I invoke `GetTipsetByHeight(x)` for `x=1..1000`. This means that 1000 goroutines are trying to walk the chain down, and all are attempting to lock `fillCacheLock`.
If, as part of the sync process, there is a need to take the `fillCacheLock` (and there is), then the sync goroutine has to compete with 1000 other goroutines attempting to take this lock.

These goroutines will take the lock, either fill the cache or check the entry and find it, release the lock and attempt to retake it.
On expectation, the sync goroutine will have to wait for 1000 `fillCache` cycles before being allowed to take the lock (assuming the lock is perfectly fair).
`Mean@GeometricDistribution[1/1001] = 1000`

With the Sharded Locking scheme, assuming in case of a collision that the sync goroutine always loses in lock fairness (way more pessimistic assumption than above, the sync goroutine has a 96% chance of progressing without waiting at all and a 99.99% chance to progress after waiting for just two `fillCache` cycles.
As you can see this completely eliminates

magik6k: Ok, I think I see the point - basically if we have a bunch of parallel walks all loading the chain, in case of a miss we'd potentially need to wait for the fill lock for a really long time as other goroutines will also want to access it - in which case we'd see a larger latency.

I was thinking in terms of only a few parallel walks, where that wouldn't be an issue, but having many, many walks this could be a problem.

magik6k (approved):
This is much better than what we have right now

PR #10897, opened 2023-05-19
[WIP] fix: dedup event logs at the API level

fridrik01:
TODO: Test if works and then optimzie

### DIFF
--- a/chain/events/filter/event.go
+++ b/chain/events/filter/event.go
@@ -55,6 +55,53 @@ type CollectedEvent struct {
 	MsgCid      cid.Cid         // cid of message that produced event
 }
 
+// EqualsExceptRevert returns true if e and other are equal except for the Reverted field
+func (e *CollectedEvent) EqualsExceptRevert(other *CollectedEvent) bool {
+	if e == nil && other == nil {
+		return true
+	}
+
+	if e == nil || other == nil {
+		return false
+	}
+
+	if e.EventIdx != other.EventIdx {
+		return false
+	}
+
+	if e.Height != other.Height {
+		return false
+	}
+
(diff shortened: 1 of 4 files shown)


chain/events/filter/event.go
```diff
@@ -55,6 +55,53 @@ type CollectedEvent struct {
...
+	if e.EmitterAddr != other.EmitterAddr {
+		return false
+	}
+
+	if len(e.Entries) != len(other.Entries) {
+		return false
+	}
+
+	for i, e1 := range e.Entries {
+		if e1.Equals(&other.Entries[i]) {
```
maciejwitowski, line 97:
```suggestion
		if !e1.Equals(&other.Entries[i]) {
```

Stebalien:
This works but it's a bit messy. When we talked about it in person, I was hoping we could just remove the "reverted" flag from events in the database entirely, but, given the way this patch works, we actually need those events to cancel reverted events.

fridrik01:
> This works but it's a bit messy. When we talked about it in person, I was hoping we could just remove the "reverted" flag from events in the database entirely, but, given the way this patch works, we actually need those events to cancel reverted events.

Right now the events table is just used for append-only logs for all events and keeping that as-is we need the reverted flag to handle de-dups. Besides changing the events db schema and changing this from an append-only log to support deletes/updates at the db level, do you have suggestions to make this somewhat less hacky?

Stebalien:
Not really, unfortunately.

Also note, we don't need to change the database schema to fix the duplication. The current schema was designed to be updated, not appended to, we just messed it up.

fridrik01:
Closing this PR and will look into addressing this on the db level instead

Issue #10532, opened 2023-03-22
Re-tune piece-reader to allow for random parallel access of data


rjan90:
Re-tune the piece reader code to allow for random, parallel access to unblock retrieval bottlenecks.

Quote from Magik:

> from what I’ve observed in booster-bitswap (which most of the time is serving ~150 blocks/sec), the bottleneck comes from the piece-reader code, which was optimized for sequential full-piece retrievals. Given that now we need more random-access reads of data in sectors, that bit of code needs to be re-tuned to allow random, parallel access (lack of which seems to be a major bottleneck in traces I’ve observed).
> 
> Just tuning piece-reader should be easy ~10x, for more we may need to educate SPs on how to store unsealed sectors (e.g. when using RAID configure large stripe size, so that only one/fewer drive heads need to move to read each block)

Estimated time effort: 1-2 days

magik6k:
Some more context:
* PieceReader code lives in https://github.com/filecoin-project/lotus/blob/master/storage/sealer/piece_reader.go
* It is essentially just a simple wrapper around a `func(offset int64) io.ReadCloser` (which is basically just a getter for http-request-reader for unsealed sector data) into a `io.ReadSeeker` + a thread-safe `io.ReaderAt`.
* Today it is entirely optimized for a single, sequential, forward reader, which can occasionally skip forwards - which is roughly the workload we see from graphsync
* The reader setup function it operates on is not the cheapest thing to call - it's starting an http request, which may imply a new tcp stream, so each call is likely a couple ms.
  * So to mitigate that cost, for sequential reads, PieceReader may discard some data in case of a forward seek
  * Again, this only really makes sense in case of a single Graphsync retrieval
  * And the way the logic is implemented means that we can only have a single parallel stream reading from a PieceReader, which is a major bottleneck in some cases - e.g. random retrievals with bitswap

* The ask in this issue is that we need to rewrite this logic to support random, parallel ReadAt calls with better performance (e.g. actually support reading from the sector with multiple streams)
  * Bonus points for not sacrificing sequential retrieval efficiency too much, e.g. by having a small pool of readers, and being smart about which one we restart on a random read / reusing existing readers for sequential reads

shrenujbansal:
@magik6k @rjan90 This domain/logic/code is completely foreign to me. Could you provide some basic instructions I can perform to witness the core issue locally?
So I'm guessing something that performs a retrieval on an unsealed sector and takes long

davidd8:
[...]

PR #10913, opened 2023-05-23
feat: piecereader: Allow parallel access

magik6k:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/10532
Replaces https://github.com/filecoin-project/lotus/pull/10895

## Proposed Changes
* Allow pieceReader reader-getters to define exact read size
* Rewrite pieceReader.ReadAt to be called in parallel, while still caching small amounts ahead for parallel sequential reads
  * This is necessary, because carv1 reads byte-by-byte for the header/cid varints, saves ~9 requests per block read
* Kept the old more aggressive prefetching mechanism for ReadSeeker, which is only used by the indexing scan, and makes it wayy faster

## Additional Info
* [ ] Make sure tests pass
* [ ] Test on a real setup, best in booster-bitswap
  * [ ] With a race detector

This PR is based on v1.23.0 so that it's possible to test with boost

## Checklist

- [ ] CI is green

### DIFF
--- a/storage/sealer/piece_provider.go
+++ b/storage/sealer/piece_provider.go
@@ -3,6 +3,7 @@ package sealer
 import (
 	"bufio"
 	"context"
+	pool "github.com/libp2p/go-buffer-pool"
 	"io"
 
 	"github.com/ipfs/go-cid"
@@ -93,8 +94,6 @@ func (p *pieceProvider) tryReadUnsealedPiece(ctx context.Context, pc cid.Cid, se
 		return nil, nil
 	}
 
-	buf := make([]byte, fr32.BufSize(pieceSize.Padded()))
-
 	pr, err := (&pieceReader{
 		ctx: ctx,
 		getReader: func(ctx context.Context, startOffset, readSize uint64) (io.ReadCloser, error) {
@@ -108,6 +107,8 @@ func (p *pieceProvider) tryReadUnsealedPiece(ctx context.Context, pc cid.Cid, se
 			if err != nil {
 				return nil, xerrors.Errorf("getting reader at +%d: %w", startOffsetAligned, err)
 			}
+
+			buf := pool.Get(fr32.BufSize(pieceSize.Padded()))
 
 			upr, err := fr32.NewUnpadReaderBuf(r, pieceSize.Padded(), buf)
 			if err != nil {
@@ -129,6 +130,7 @@ func (p *pieceProvider) tryReadUnsealedPiece(ctx context.Context, pc cid.Cid, se
 			}{
 				Reader: bir,
 				Closer: funcCloser(func() error {
+					pool.Put(buf)
 					return r.Close()
 				}),
 			}, nil
--- a/storage/sealer/piece_provider.go
+++ b/storage/sealer/piece_provider.go
@@ -3,10 +3,10 @@ package sealer
 import (
 	"bufio"
 	"context"
-	pool "github.com/libp2p/go-buffer-pool"
 	"io"
 
 	"github.com/ipfs/go-cid"
+	pool "github.com/libp2p/go-buffer-pool"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/dagstore/mount"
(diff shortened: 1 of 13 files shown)


storage/sealer/piece_provider.go
```diff
@@ -93,43 +95,54 @@ func (p *pieceProvider) tryReadUnsealedPiece(ctx context.Context, pc cid.Cid, se
...
-
 	pr, err := (&pieceReader{
 		ctx: ctx,
-		getReader: func(ctx context.Context, startOffset uint64) (io.ReadCloser, error) {
+		getReader: func(ctx context.Context, startOffset, readSize uint64) (io.ReadCloser, error) {
 			startOffsetAligned := storiface.UnpaddedByteIndex(startOffset / 127 * 127) // floor to multiple of 127
+			startOffsetDiff := int(startOffset - uint64(startOffsetAligned))
+
+			endOffset := startOffset + readSize
+			endOffsetAligned := storiface.UnpaddedByteIndex((endOffset + 126) / 127 * 127) // ceil to multiple of 127
```
snadrus, line 105: Could these "magic numbers" be calculated in a constant from other constants or documented values?

storage/sealer/piece_provider.go
```diff
@@ -93,43 +95,54 @@ func (p *pieceProvider) tryReadUnsealedPiece(ctx context.Context, pc cid.Cid, se
...
 
-			r, err := rg(startOffsetAligned.Padded())
+			r, err := rg(startOffsetAligned.Padded(), endOffsetAligned.Padded())
```
snadrus, line 107: Could rg have a more descriptive name for what it does?

storage/sealer/piece_reader.go
```diff
@@ -30,20 +34,38 @@ type pieceReader struct {
...
+	remReads *lru.Cache[int64, []byte] // data start offset -> data
+	// todo try carrying a "bytes read sequentially so far" counter with those
+	//  cacahed byte buffers, increase buffer sizes when we see that we're doing
+	//  a long sequential read
 }
 
 func (p *pieceReader) init() (_ *pieceReader, err error) {
 	stats.Record(p.ctx, metrics.DagStorePRInitCount.M(1))
 
+	p.seqMCtx, _ = tag.New(p.ctx, tag.Upsert(metrics.PRReadType, "seq"))
```
snadrus, line 59: context.Context is never supposed to be stored in a struct, so:
can we have init(ctx context.Context) ?
Here and the next line, can tag.New take that Context and return it as well (since it should be non-destructive).

storage/sealer/piece_provider.go
```diff
@@ -93,43 +95,54 @@ func (p *pieceProvider) tryReadUnsealedPiece(ctx context.Context, pc cid.Cid, se
...
 
+			var closeOnce sync.Once
+
 			return struct {
 				io.Reader
 				io.Closer
 			}{
 				Reader: bir,
 				Closer: funcCloser(func() error {
+					// this close can be called more than once - double close signals to the paths.Reader that we are done with the piece
```
snadrus, line 136: Is there a chance that the second "closer" could read or otherwise interact with the object?
If it's in the pool, r.Close() could close the subsequent use of this object. 
If we cannot control the number of handles held for the object, its not mem-pool capable.

storage/sealer/piece_reader.go
```diff
@@ -30,20 +34,38 @@ type pieceReader struct {
 	len       abi.UnpaddedPieceSize
 	onClose   context.CancelFunc
 
+	seqMCtx context.Context
```
snadrus, line 37: same as above

snadrus (changes requested):

storage/sealer/piece_reader.go
magik6k: Yeahh, moved the ctx into init params, but the `MCtx` ones will probably have to stay in that struct - opencensus sadly needs a context to record stats on (technically we could hack into it's internal packages, but that's even more gnarly)

snadrus:
Looks good. Let me know when you're ready for me to approve.

magik6k:
Merged master into here so that the previous commits can still be depended on in boost

Should be mergeable now.

snadrus (approved):

Issue #10871, opened 2023-05-15
Implement a tooling for slasher 


jennijuju:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Implement a tooling for reporting ConsensusFault 

### Describe the solution you'd like

it can be a separate process like https://github.com/filecoin-project/lotus/blob/bc6017194d6b96beb281f34bb1bd422e861c30e2/cli/disputer.go#L26

### Additional context

(notes from @Kubuxu how he used to implemented one: 
```there is a RPC for incoming blocks, I passed them through modified version of the slashing filter such that it supports multiple miners and outputs all offending blocks.```

jennijuju:
[marco-storswift](https://github.com/marco-storswift) will take a stab at this!

Kubuxu:
I have found the code in my backups: https://github.com/filecoin-project/lotus/compare/feat/old/consensus-faulter

marco-storswift:
@Kubuxu, Thanks very mush  , i meet the case when I do this in rcp call.
 2023-05-23T14:56:06.666+0800    WARN    rpc     go-jsonrpc@v0.3.1/websocket.go:330      failed to send request  {"method": "xrpc.cancel", "id": 2, "error": "websocket: close sent"}
  so i want to impl in lotus  daemon --slash-consensus,  use SlashFilter.MinedBlock to check consensus faulter.
  i will commit this PR today

PR #10917, opened 2023-05-24
Implement a tooling for slasher

swift-mx:
## Related Issues

https://github.com/filecoin-project/lotus/issues/10871

## Checklist

- [ ] CI is green


jennijuju:
Thanks for the contribution!

cmd/lotus/daemon.go
```diff
@@ -574,3 +600,88 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
...
+
+	for block := range blocks {
+		log.Infof("deal with block: %d, %v, %s", block.Height, block.Miner, block.Cid())
+		if otherBlock, err := a.SlashFilterMinedBlock(ctx, block); err != nil {
+			if otherBlock == nil {
+				continue
+			}
+			log.Errorf("<!!> SLASH FILTER ERROR: %s", err)
+			var fromAddr address.Address
+			if from == "" {
```
arajasek, line 620: Let's set the sender address at the start of this method (outside of the for loop).

cmd/lotus/daemon.go
```diff
@@ -160,6 +166,19 @@ var DaemonCmd = &cli.Command{
 			Name:  "restore-config",
 			Usage: "config file to use when restoring from backup",
 		},
+		&cli.BoolFlag{
+			Name:  "slash-consensus",
+			Usage: "Report consensus fault",
+			Value: false,
+		},
+		&cli.StringFlag{
+			Name:  "from",
```
arajasek, line 175: Rename this flag to --slasher-sender?

cmd/lotus/daemon.go
```diff
@@ -160,6 +166,19 @@ var DaemonCmd = &cli.Command{
...
+			Usage: "optionally specify the account to report consensus from",
+		},
+		&cli.StringFlag{
+			Name:  "extra",
```
arajasek, line 179: I don't understand how we could meaningfully use this `extra` param?

cmd/lotus/daemon.go
```diff
@@ -574,3 +600,88 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
...
+			}
+			message, err := a.MpoolPushMessage(ctx, &types.Message{
+				To:     block.Miner,
+				From:   fromAddr,
+				Value:  types.NewInt(0),
+				Method: builtin.MethodsMiner.ReportConsensusFault,
+				Params: enc,
+			}, nil)
+			if err != nil {
+				continue
```
arajasek, line 680: Log an error here, please

cmd/lotus/daemon.go
```diff
@@ -574,3 +600,88 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
 
 	return nil
 }
+
+func slashConsensus(a lapi.FullNode, from string, extra string) error {
+	ctx := context.Background()
+
+	blocks, err := a.SyncIncomingBlocks(ctx)
+	if err != nil {
+		return err
```
arajasek, line 609: Please chain this and all other errors for clarity (that is, `return xerrors.Errorf("sync incoming blocks failed: %w", err)`.

node/impl/full/sync.go
```diff
@@ -105,6 +106,26 @@ func (a *SyncAPI) SyncIncomingBlocks(ctx context.Context) (<-chan *types.BlockHe
 	return a.Syncer.IncomingBlocks(ctx)
 }
 
+func (a *SyncAPI) SlashFilterMinedBlock(ctx context.Context, bh *types.BlockHeader) (*types.BlockHeader, error) {
+	parent, err := a.Syncer.ChainStore().GetBlock(ctx, bh.Parents[0])
+	if err != nil {
+		return nil, nil
+	}
+	err = a.SlashFilter.MinedBlock(ctx, bh, parent.Height)
+	if err != nil && strings.Contains(err.Error(), "mining faults") {
```
arajasek, line 115: I think we should refactor `MinedBlock` to return the `others`, instead of parsing the error message here.

node/impl/full/sync.go
```diff
@@ -105,6 +106,26 @@ func (a *SyncAPI) SyncIncomingBlocks(ctx context.Context) (<-chan *types.BlockHe
 	return a.Syncer.IncomingBlocks(ctx)
 }
 
+func (a *SyncAPI) SlashFilterMinedBlock(ctx context.Context, bh *types.BlockHeader) (*types.BlockHeader, error) {
+	parent, err := a.Syncer.ChainStore().GetBlock(ctx, bh.Parents[0])
+	if err != nil {
+		return nil, nil
+	}
+	err = a.SlashFilter.MinedBlock(ctx, bh, parent.Height)
```
arajasek, line 114: Is the SlashFilter always initiated, or could it be `nil`?

node/impl/full/sync.go
```diff
@@ -105,6 +106,26 @@ func (a *SyncAPI) SyncIncomingBlocks(ctx context.Context) (<-chan *types.BlockHe
 	return a.Syncer.IncomingBlocks(ctx)
 }
 
+func (a *SyncAPI) SlashFilterMinedBlock(ctx context.Context, bh *types.BlockHeader) (*types.BlockHeader, error) {
```
arajasek, line 109: Instead of adding this as an API endpoint, I think we can just create a separate SlashFilter in its own directory for the purpose of monitoring the entire network.

arajasek (commented):
Thanks for this, it'll be really good to have this in Lotus and easy to run!

I think the high-level approach is on the right track, but this unfortunately doesn't work as intended yet -- the SlashFilter is currently not keyed by miner. It assumes we are only interested in one miner. 

I think in order to make this work, we will need to expand it to be keyed by both the miner AND by epoch / parents. That way we can correctly monitor all miners for consensus faults, without any false positives.

We will likely also want to Garbage Collect these entries periodically -- anything older than (say) 3 days, can be dropped.

cmd/lotus/daemon.go
swift-mx: good idea

cmd/lotus/daemon.go
swift-mx: I have referred to the official implementation and now removed this parameter

node/impl/full/sync.go
swift-mx: It shouldn't be possible for nil

swift-mx:
reopen here https://github.com/filecoin-project/lotus/pull/10928

Issue #10663, opened 2023-04-13
Investigate using the message index in the ChainStore


fridrik01:
See Epic: Message Index https://github.com/filecoin-project/lotus/issues/10536 for background

In PR https://github.com/filecoin-project/lotus/pull/10452 we introduced a Message Index to Lotus which is backed by a sqlite backend and allows us to speed up certain calls like `StateSearchMessage`. 

We may be able to use this index in the ChainStore, but it's unclear. We should investigate (see https://github.com/filecoin-project/lotus/pull/10452#issuecomment-1467322283 for context).

maciejwitowski:
Lowering the priority based on info from @fridrik01 

> It turns out that that task was more an "investigation" rather than actually refactor since it was not clear how much that change will improve performance

Stebalien:
Benchmarking determined that this isn't required.

PR #10922, opened 2023-05-25
perf: use sqlite/msgindex.db as `epoch=>tipsetkey` index to speed up `eth_getBlockByNumber` queries

i-norden:
## Related Issues

Builds on-top of https://github.com/filecoin-project/lotus/pull/10632
https://github.com/filecoin-project/lotus/issues/10663
https://github.com/vulcanize/filecoin-indexing/issues/20

## Proposed Changes

in sqlite/msgindex.db we have a table which maps message_cids to epochs and tipset cids for the tipset in which these messages appear. We could also leverage this table as an index for looking up the tipset corresponding to a specific epoch when proceeding down the `eth_getBlockByNumber` search path described in https://github.com/vulcanize/filecoin-indexing/issues/20#issue-1724539676 (assuming my understanding there is correct).

## Additional Info

Opening this as a draft hoping to spur some discussion/feedback, I am not entirely confident in my line of thinking for this. Additionally, if the reasoning is sound, I would appreciate maintainer's opinion on how to best access this index from the `ChainStore`/`ChainIndex` (i.e. https://github.com/filecoin-project/lotus/pull/10922/files#diff-6032ffd931dc723fcaa5d9b7926dae0b7fa6bea74763d8e0a0693d2b45992976R164).

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/index/interface.go
+++ b/chain/index/interface.go
@@ -33,6 +33,14 @@ type MsgIndex interface {
 	Close() error
 }
 
+// TipsetIndex is the interface to the tipset index
+type TipsetIndex interface {
+	// GetTipsetCID returns the tipset cid for the given epoch
+	GetTipsetCID(ctx context.Context, epoch abi.ChainEpoch) (*cid.Cid, error)
+	// Close closes the index
+	Close() error
+}
+
 type dummyMsgIndex struct{}
 
 func (dummyMsgIndex) GetMsgInfo(ctx context.Context, mCid cid.Cid) (MsgInfo, cid.Cid, error) {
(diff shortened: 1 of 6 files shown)


fridrik01:
@i-norden is this not a duplicate of https://github.com/filecoin-project/lotus/pull/10632?

i-norden:
@fridrik01 there is duplicate code here, I didn't see that PR- will rebase this on-top of that! The difference here is we also expose a `GetTipsetByEpoch` method on-top of the newly created `tipset_epochs ` index to go from `epoch => tipsetkey` directly and then use that method in `ChainIndex.GetTipsetByHeight`.

i-norden:
Hey @fridrik01 wanted to check if my response makes sense, and if so if this would be a desirable update. I believe such a `GetTipsetByEpoch` method could also replace the the tipset skiplist as it is used in the `eth_getBlockByNumber` code path as well.

fridrik01:
I don't think this is required as the skip list is very fast, I tried it on a FEVM archival node and IIRC it was in the order of 10ms so this won't help much. See also #10528 which was not merged due to the complexity not being worth it

i-norden:
Thanks @fridrik01, that makes sense.

PR #10928, opened 2023-05-30
Implement a tooling for slasher 

swift-mx:
## Related Issues

this is reopen for https://github.com/filecoin-project/lotus/pull/10917
## Checklist

- [ ] CI is green


chain/gen/slashfilter/slashfilter.go
```diff
@@ -78,45 +78,50 @@ func (f *SlashFilter) MinedBlock(ctx context.Context, bh *types.BlockHeader, par
...
 	if err := f.byEpoch.Put(ctx, epochKey, bh.Cid().Bytes()); err != nil {
-		return xerrors.Errorf("putting byEpoch entry: %w", err)
+		return cid.Undef, xerrors.Errorf("putting byEpoch entry: %w", err)
 	}
 
-	return nil
+	return cid.Undef, nil
+}
+
+func (f *SlashFilter) MinedBlock(ctx context.Context, bh *types.BlockHeader, parentEpoch abi.ChainEpoch) error {
```
arajasek, line 97: I don't think we need a new method here -- we can just refactor the callers of `MinedBlock` to expect 2 returns.

chain/gen/slashfilter/slashfilter.go
```diff
@@ -78,45 +78,50 @@ func (f *SlashFilter) MinedBlock(ctx context.Context, bh *types.BlockHeader, par
...
+	_, err := f.CheckBlock(ctx, bh, parentEpoch)
+	return err
 }
 
-func checkFault(ctx context.Context, t ds.Datastore, key ds.Key, bh *types.BlockHeader, faultType string) error {
+func checkFault(ctx context.Context, t ds.Datastore, key ds.Key, bh *types.BlockHeader, faultType string) (cid.Cid, error) {
 	fault, err := t.Has(ctx, key)
 	if err != nil {
-		return err
+		return cid.Undef, err
```
arajasek, line 105: Please chain the error here and in other returns (`return cid.Undef, xerrors.Errorf("failed to read from datastore: %w", err))

chain/gen/slashfilter/slashfilter.go
```diff
@@ -78,45 +78,50 @@ func (f *SlashFilter) MinedBlock(ctx context.Context, bh *types.BlockHeader, par
 			}
 
 			if !found {
-				return xerrors.Errorf("produced block would trigger 'parent-grinding fault' consensus fault; miner: %s; bh: %s, expected parent: %s", bh.Miner, bh.Cid(), parent)
+				return cid.Undef, xerrors.Errorf("produced block would trigger 'parent-grinding fault' consensus fault; miner: %s; bh: %s, expected parent: %s", bh.Miner, bh.Cid(), parent)
```
arajasek, line 81: We should still return `parent` as the witness here -- we will just have to special-case handle it in the caller.

cmd/lotus/daemon.go
```diff
@@ -574,3 +603,99 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
...
+			}
+			message, err := a.MpoolPushMessage(ctx, &types.Message{
+				To:     block.Miner,
+				From:   fromAddr,
+				Value:  types.NewInt(0),
+				Method: builtin.MethodsMiner.ReportConsensusFault,
+				Params: enc,
+			}, nil)
+			if err != nil {
+				log.Errorf("ReportConsensusFault to messagepool error:%s", err)
```
arajasek, line 677:
```suggestion
				log.Errorf("ReportConsensusFault to messagepool error:%w", err)
```

cmd/lotus/daemon.go
```diff
@@ -574,3 +603,99 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
...
+func slashFilterMinedBlock(ctx context.Context, sf *slashfilter.SlashFilter, a lapi.FullNode, bh *types.BlockHeader) (*types.BlockHeader, error) {
+	parent, err := a.ChainGetBlock(ctx, bh.Parents[0])
+	if err != nil {
+		return nil, xerrors.Errorf("chain get block error:%s", err)
+	}
+	otherCid, err := sf.CheckBlock(ctx, bh, parent.Height)
+	if err != nil {
+		return nil, xerrors.Errorf("slash filter check block error:%s", err)
+	}
+	if otherCid != cid.Undef {
```
arajasek, line 696: We can handle the parent-grinding mining fault here by returning a slice of `BlockHeader`s as a result.

We can determine whether this is a valid parent-grinding mining fault by checking whether `bh`s "grandparents" are the same as `otherHeader`'s parents. In this case, we can return any block from `bh`'s parent tipset as the "witness" needed. See [here](https://github.com/filecoin-project/lotus/blob/dbbcf4b2ee9626796e23a096c66e67ff350810e4/chain/vm/fvm.go#L125-L144) for how this is used.

cmd/lotus/daemon.go
```diff
@@ -160,6 +169,19 @@ var DaemonCmd = &cli.Command{
 			Name:  "restore-config",
 			Usage: "config file to use when restoring from backup",
 		},
+		&cli.BoolFlag{
```
arajasek, line 172: Let's not have this in lotus daemon -- instead, I think this should be a standalone process that connects to a lotus node (exactly the way [the miner does this](https://github.com/filecoin-project/lotus/blob/b7227d3a6b2b9adfde8ab6f2ae04e2cf2bffa481/cmd/lotus-miner/run.go#L77-L81)). 

Do you think that's reasonable? If so, can you please move this into a new lotus-slasher process? If not, happy to discuss more!

arajasek (commented):
This looks pretty close to landable, thank you! We'll also definitely want to test this out (perhaps by connecting to calibrationnet and intentionally creating some consensus faults there).

arajasek (commented):
This looks pretty close to landable, thank you! We'll also definitely want to test this out (perhaps by connecting to calibrationnet and intentionally creating some consensus faults there).

cmd/lotus/daemon.go
swift-mx: I have tried using lotus-slasher before, But encountered the problem of rpc blocking(too many rpc request)

swift-mx: If there is a good idea to solve this problem, I would be happy to add an independent process

cmd/lotus/daemon.go
swift-mx: Thank you for your suggestion. I will update it later

cmd/lotus/daemon.go
arajasek: Hmm, I'm a little surprised to hear that -- @magik6k do you have any suggestions here?

It's fine to leave it as a goroutine for now, though.

chain/gen/slashfilter/slashfilter.go
swift-mx: This is done in order to reduce modifications, and this method can be removed if necessary.

arajasek: Do the modifications get big? I think it should just be a couple other lines in a couple other files.

chain/gen/slashfilter/slashfilter.go
arajasek: I think we resolved this without actually making the change?

cmd/lotus/daemon.go
```diff
@@ -574,3 +603,114 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
...
+	// Here extra is the "witness", a third block that shows the connection between A and B as
+	// A's sibling and B's parent.
+	// Specifically, since A is of lower height, it must be that B was mined omitting A from its tipset
+	//
+	//      B
+	//      |
+	//  [A, C]
+	if types.CidArrsEqual(blockA.Parents, blockC.Parents) && blockA.Height == blockC.Height &&
+		types.CidArrsContains(blockB.Parents, blockC.Cid()) && !types.CidArrsContains(blockB.Parents, blockA.Cid()) {
+		return blockC, xerrors.Errorf("chain get other block error:%s", err)
```
arajasek, line 712: We need to return both blockA and blockC here, so that blockC can be supplied in the `Extra` field of `ReportConsensusFaultParams`

arajasek (approved):
This LGTM! Thank you very much!

chain/gen/slashfilter/slashfilter.go
```diff
@@ -26,20 +26,20 @@ func New(dstore ds.Batching) *SlashFilter {
...
 
-func (f *SlashFilter) MinedBlock(ctx context.Context, bh *types.BlockHeader, parentEpoch abi.ChainEpoch) error {
+func (f *SlashFilter) MinedBlock(ctx context.Context, bh *types.BlockHeader, parentEpoch abi.ChainEpoch) (cid.Cid, error) {
 	epochKey := ds.NewKey(fmt.Sprintf("/%s/%d", bh.Miner, bh.Height))
 	{
 		// double-fork mining (2 blocks at one epoch)
-		if err := checkFault(ctx, f.byEpoch, epochKey, bh, "double-fork mining faults"); err != nil {
-			return err
+		if witness, err := checkFault(ctx, f.byEpoch, epochKey, bh, "double-fork mining faults"); err != nil {
+			return witness, err
```
arajasek, line 34: Chain error here, and in cases below, please.

PR #10955, opened 2023-06-05
Prefill GetTipsetByHeight skiplist cache on lotus startup

fridrik01:
**Context**
The `GetTipsetByHeight` method internally maintains a skiplist to more efficiently lookup a tipsets at a given height. 

When users call a RPC method that call this method (for example `eth_getBlockByNumber`) then if this skiplist is not already cached then the user needs to wait until this skiplist is populated in the background which can take 10 minutes or longer.

Instead, lets just always populate and fully prefill this skiplist in the background as soon as lotus starts up (only when EnableEthRPC is set). This will not fix the issue, but it will be more likely to give better user experience.

**Test plan**

Compiled and restarted Lotus and observed the skiplist was generated on startup

```
2023-06-05T17:43:59.757-0500	INFO	modules	modules/ethmodule.go:61	Start prefilling GetTipsetByHeight cache
...
2023-06-05T17:44:41.722-0500	INFO	modules	modules/ethmodule.go:66	Prefilling GetTipsetByHeight done in 41.965361958s
```

### DIFF
--- a/node/modules/ethmodule.go
+++ b/node/modules/ethmodule.go
@@ -4,8 +4,11 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"time"
 
 	"go.uber.org/fx"
+
+	"github.com/filecoin-project/go-state-types/abi"
 
 	"github.com/filecoin-project/lotus/chain/ethhashlookup"
 	"github.com/filecoin-project/lotus/chain/events"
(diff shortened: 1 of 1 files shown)


arajasek (approved):

arajasek:
./scripts/fiximports should make CI green

Issue #10906, opened 2023-05-22
Random Sync Failure Reported by Multiple Users - Continues the 10788 Discussion.


TippyFlitsUK:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Various - Please see discussion links in the description
```

### Repro Steps

1. Run lotus daemon
2. Lotus loses and regains sync in a cyclical manner

### Describe the Bug

This new issue aims to collate and organise multiple instances of SP feedback from the existing https://github.com/filecoin-project/lotus/issues/10788 issue and across other messaging platforms such as Slack.

A new discussion thread has been created for each contributor and existing feedback has been pre-filled with the details that have already been provided.

The Lotus Team would be very grateful if returning contributors could add additional feedback and logs to their own dedicated discussion thread as and when it becomes available.

If you are experiencing this issue for the first time, please feel free to add your feedback and logs below this thread and the team will distribute it accordingly.

Many thanks all!! 🙏

- #### @RobQuistNL - Logs & Feedback - https://github.com/filecoin-project/lotus/discussions/10907#discussioncomment-5970070
- #### @marshyonline - Logs & Feedback - https://github.com/filecoin-project/lotus/discussions/10907#discussioncomment-5970251
- #### @scaseye - Logs & Feedback - https://github.com/filecoin-project/lotus/discussions/10907#discussioncomment-5970251
- #### @marco-storswift - Logs & Feedback - https://github.com/filecoin-project/lotus/discussions/10907#discussioncomment-5971005
[...]

PR #10973, opened 2023-06-12
fix: pubsub: do not treat ErrExistingNonce as Reject

arajasek:
## Related Issues

This might help with #10906.

The `ErrExistingNonce` error is thrown when [we already have the exact message we're trying to add ](https://github.com/filecoin-project/lotus/blob/fb228ebbff5b2e3bbb6a51f2b2cfe232f2f16f94/chain/messagepool/messagepool.go#L284-L285)in our pending mset. It is an error to try and add this message, but it is not wrong for us to _receive_ a message multiple times, and try to add it. We should definitely not penalize peers who inform us of messages we've already added.

What's not clear to me is whether something earlier in the pipeline should be realizing that we've already processed this message. Regardless, however, I think we should be treating this case as an Ignore, not a Reject. 

## Proposed Changes

Lower the pubsub handling of this particular failure to Ignore.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/sub/incoming.go
+++ b/chain/sub/incoming.go
@@ -361,6 +361,8 @@ func (mv *MessageValidator) Validate(ctx context.Context, pid peer.ID, msg *pubs
 		case xerrors.Is(err, messagepool.ErrGasFeeCapTooLow):
 			fallthrough
 		case xerrors.Is(err, messagepool.ErrNonceTooLow):
+			fallthrough
+		case xerrors.Is(err, messagepool.ErrExistingNonce):
 			return pubsub.ValidationIgnore
 		default:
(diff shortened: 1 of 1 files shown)


chain/sub/incoming.go
```diff
@@ -362,6 +362,8 @@ func (mv *MessageValidator) Validate(ctx context.Context, pid peer.ID, msg *pubs
 			fallthrough
 		case xerrors.Is(err, messagepool.ErrNonceTooLow):
 			return pubsub.ValidationIgnore
```
Stebalien, line 364: fallthrough?

Stebalien (approved):

PR #11003, opened 2023-06-24
chore: chain: add env var to disable migration

kasteph:
## Related Issues
We are running Lily, which is dependent on Lotus, for historical backfills. We use historical snapshots with the full state tree of a given range of epochs so we shouldn't need to do to the migration job I think.

## Proposed Changes
Adding an environmental variable, that when set to 1, stops the migration function from running.

## Additional Info
I am not sure whether this warrants documentation as the other similar env var `LOTUS_DISABLE_PRE_MIGRATIONS` does not.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/stmgr/forks.go
+++ b/chain/stmgr/forks.go
@@ -31,6 +31,11 @@ import (
 // EnvDisablePreMigrations when set to '1' stops pre-migrations from running
 const EnvDisablePreMigrations = "LOTUS_DISABLE_PRE_MIGRATIONS"
 
+// EnvDisableMigrations when set to '1' stops migrations from running
+// Useful when running a non-bootstrapped node initialized with
+// a chain snapshot containing the full state tree
+const EnvDisableMigrations = "LOTUS_DISABLE_MIGRATIONS"
+
 // MigrationCache can be used to cache information used by a migration. This is primarily useful to
 // "pre-compute" some migration state ahead of time, and make it accessible in the migration itself.
 type MigrationCache interface {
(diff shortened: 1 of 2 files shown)


kasteph:
@birdychang

kasteph:
Hi @arajasek,

Thanks for the feedback! I'm still quite new to the internals so I appreciate the attention to the feedback.

> Can we talk through the motivation for this PR a bit more? 

We are performing historical backfills of the Filecoin chain data on BigQuery using Lily with an archival snapshot of the chain containing receipts. However, it can sometimes happen that the migration step and the `ChainGetTipset` falls out of order (for e.g. the migration step finishes after) and so the job fails.


> Do you know what specific step is triggering the migration in your operation?

The migration step is getting triggered when we perform the Lily walk, which is running the [`StateManager`](https://github.com/filecoin-project/lily/blob/master/lens/lily/modules/statemanager.go#L25-L35) under the hood.

> The faster tipset state lookup in the stmgr should be skipping the actual state computation entirely if you've already imported the chain (not sure whether or not you have).

We have imported the chain before we run the Lotus daemon. Then we attempt the Lily walk of the chain on a non-bootstrapped node.

> Returning root, nil in HandleStateForks is wrong (if there's a migration), so whatever is performing this operation is going to be getting an incorrect result -- that may be an issue for your operation

Oh I see, let me revisit that and see if I can figure out what the correct way to do it is.

Thanks again!

arajasek:
No worries, thanks a lot for engaging! I'm not too familiar with how Lily works, so we'll meet in the middle somewhere :D 

I think the thing to figure out is exactly where the call to `HandleStateForks` is coming from in Lily, and whether that's essential. I'm worried that it might be for the correct functioning of the Lily walk (though I'm not too sure what that entails). 

If that is the case, I think what we might do is include the "hard-cached" results of the migration in Lotus binaries. When running the fork, we can check if we have the result pre-computed, and use that if so. This will allow everything else to function as-is, and should be fully correct.

rjan90:
Hey! 👋

rjan90:

Closing

PR #11006, opened 2023-06-26
chore: not display privatekey

beck-8:
## Related Issues

Let the user's private key no longer be displayed in the terminal to prevent malicious monitoring of the screen and stealing the private key.
## Proposed Changes

```bash
$ ./lotus wallet import
Enter private key(not display in the terminal): 
imported key f1xxxxxxx successfully!

$ ./lotus wallet import
Enter private key(not display in the terminal): 
ERROR: saving to keystore: checking key before put 'wallet-f1xxxxxxx': key already exists

$ echo 7bxxxxxxx | ./lotus wallet import
imported key f1xxxxxxx successfully!

$ echo 7bxxxxxxx | ./lotus wallet import
ERROR: saving to keystore: checking key before put 'wallet-f1xxxxxxx': key already exists
```
## Checklist

- [ ] CI is green

### DIFF
--- a/cli/wallet.go
+++ b/cli/wallet.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/urfave/cli/v2"
+	"golang.org/x/term"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-address"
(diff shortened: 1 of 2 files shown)


Stebalien:
Seems reasonable to me but could you switch to https://pkg.go.dev/golang.org/x/term? The library you're using is deprecated.

Stebalien:
Ah, wait. So, it looks like this code is _designed_ for reading the key from stdin, not from a user typing it.

Stebalien:
Hm... although it still prompts.... Ok, I think the right answer is:

1. Check `IsTerminal`.
2. If it's a terminal, read it as a password (letting the user type it).
3. Otherwise, just read it from stdin.

beck-8:
> Seems reasonable to me but could you switch to https://pkg.go.dev/golang.org/x/term? The library you're using is deprecated.

this is ok

Stebalien:
Are you sure this works when stdin is a pipe, not a terminal? That's why I suggested calling `IsTerminal` first.

beck-8:
> Are you sure this works when stdin is a pipe, not a terminal? That's why I suggested calling `IsTerminal` first.

It is not supported, nor does it want to support it, and the original logic does not support it. ReadPassword will return the reason for not supporting it.
You mean you want me to support pipes? Or let me handle errors gracefully (I think ReadPassword will return an error if there is a problem, so there is no need to handle it)?
Please correct me if I understand something wrong, or if the information is wrong.
image

beck-8:
> Are you sure this works when stdin is a pipe, not a terminal? That's why I suggested calling `IsTerminal` first.

I see what you mean, thank you very much.

beck-8:
forgive me for using force-push, I just want to make the commit cleaner

Stebalien (approved):
LGTM!

Issue #10884, opened 2023-05-18
Filecoin state profiling 


jennijuju:
edited by @ZenGround0 
## User Story

I am a user who develops the filecoin protocol or software involved in running or using the protocol.

1) I want to understand what is happening with filecoin state.  I want to see the current breakdown of byte usage by the protocol to understand if there are dangerous trends that need to be addressed and get ideas for how to fix them.

2) I also want to be able to go back and analyze how previous changes have impacted snapshot size / protocol data usage.

3) I can visualize how new changes I have developed impact snapshot size / protocol data usage.
## Acceptance Criteria

I can run a command or a small set of commands that gives me two things provided with a filecoin snapshot:

1) data on the breakdown of state usage of a chain snapshot broken down into a) state tree b) message / headers c) state churn.  Further breakdown of state and churn by i) actor type ii) actor type field and further breakdown of messages by message type

2) a visualization of this data for quick inspection.  This visualization will be a [DiskInventoryX](http://derlien.com/) style treemap breakdown of the above data.
![image](https://github.com/filecoin-project/lotus/assets/5515260/9a9f9f44-fe1b-421c-ada9-235411e89b30)

```[tasklist]
#### Deliverables
- [ ] These outputs will be judged against a first motivating task: we will use them to explain exactly how the nv19 protocol upgrade reduced the snapshot size by ~50%. https://github.com/filecoin-project/lotus/issues/11035
- [ ] run this process in our own infra on a regular basis to monitor the network's condition. (https://github.com/filecoin-project/lotus/issues/11022)
- [x] Document how to use this tool (shipped at https://github.com/filecoin-project/lotus/discussions/11037)
```

## Techincal Breakdowns

```[tasklist]
- [x] Import snapshot into something that allows for analaysis (current plan is to use lotus)
- [x] Debug and repurpose existing `lotus-shed stateroot stat` as needed
- [x] Develop similar functionalities for whole snapshot churn / messages / headers
- [x] Experiment with and decide on level of granularity, likely breaking actor types down into top level fields
- [x] Use a data output format that is good for people and plotting programs (probably json)
- [x] Plotting program: probably using [this tool]*(https://plotly.com/python/treemaps/)
```
Simplifying assumptions (we can revisit these) 
* Ignore overlap among actors to start
* Don't look at individual actors to start, just actor types
* All operations are over a single snapshot, forget about measuring longer time scales

[Bonus Cake](https://www.youtube.com/watch?v=-aSQleq_we8) 
[...]

PR #11012, opened 2023-06-28
feat:profiling:state summary and visualization

ZenGround0:
## Related Issues

#10884 

## Proposed Changes

* One lotus-shed command that takes a non-running repo and generates a summary of data usage by protocol area
* A python script for visualizing the data 

### Testplan

I've done many manual runs on calibration and mainnet repos and both lotus-shed and python script generate plausible output.  Ive cross checked output data and it appears legit matching lotus chain store size + providing an explanation for pre nv19 snapshot bloat.  I think this is good enough for running a monitoring service.  If we find problems we can fix them as they show up.

### Speed

While reviewing please keep an eye out for optimization opportunities in the shed command, today it takes about 5 hours to run so a 20% improvement is an hour saved.

### Halp

@jennijuju where should documentation for this go?  I have lotus-shed command help but probably we want something a little more thorough.  Is https://github.com/filecoin-project/lotus/discussions/categories/tutorials enough?

### Extension ideas

I am hoping to tweak the python script to take in visualization chart type information so we can visualize same data in a few different ways.

State summary shed command is readily extensible by using a path identifier for protocol areas.  We could extend this to breakdown message usage by message type for example.

We could use similar visualization on gas traces.  In particular cron gas traces would be useful to keep track of in the same way.

## Checklist

- [x] Commits have a clear commit message. (they will after squash)
- [ ] CI is green


cmd/lotus-shed/state-stats.go
```diff
@@ -166,49 +186,393 @@ var statObjCmd = &cli.Command{
...
+					return
 				}
 			}
 		}()
+		visit := func(c cid.Cid) bool {
+			ch := make(chan bool)
+			cidCh <- cidCall{c: c, resp: ch}
+			out := <-ch
+			return out
+		}
```
magik6k, line 400: I'd just use a mutex, should me much faster

cmd/lotus-shed/state-stats.go
```diff
@@ -166,49 +186,393 @@ var statObjCmd = &cli.Command{
...
+}
+
+type StoreHandle struct {
+	bs     blockstore.Blockstore
+	cs     *store.ChainStore
+	sm     *stmgr.StateManager
+	closer func()
+}
+
+func loadChainStore(ctx context.Context, repoPath string) (*StoreHandle, error) {
```
magik6k, line 216: Probably could have a mode where this loads a snapshot to a memory-map blockstore, would need a decent amount of RAM, but should be much faster

snissn:
Having this in two branches is causing an issue on the deployment infra. I think this code is good to merge, and all of the comments above are good suggestions that we should open a second issue for.

ZenGround0:
> I think this code is good to merge

yup, I just need a green check mark, can you approve @snissn ?

cmd/lotus-shed/state-stats.go
ZenGround0: I tried it out, was minorly faster but this is apparently not a bottleneck

cmd/lotus-shed/state-stats.go
```diff
@@ -427,6 +748,93 @@ to reduce the number of decode operations performed by caching the decoded objec
...
+		results = append(results, res)
+	}
+
+	// now walk the top level object of actor state
+	dsc := &dagStatCollector{
+		ds:   dag,
+		walk: carWalkFunc,
+	}
+
+	if err := merkledag.Walk(ctx, dsc.walkLinks, in.c, visit, merkledag.Concurrent()); err != nil {
```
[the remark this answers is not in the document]
snissn: this PR has been working well in testing, @ZenGround0 if you can investigate this, but i don't think we're bottle necked here

snissn (approved):
looks good to me, has been working and generated expected output for us!

arajasek (commented):
Sorry, I never submitted this (partial) review!

PR #11015, opened 2023-06-28
feat: Make sure we don't store duplidate actor events caused to reorgs in events.db

fridrik01:
Fixes: https://github.com/filecoin-project/ref-fvm/issues/1350

## Context
When lotus is configured to enable indexing of actor events (when EnableEthRPC=true and DisableHistoricFilterAPI=false) then Lotus will observe for new tipset and store all the actor events in a sqlite table stored at sqlite/events.db.

However, users have been reporting that these events are duplicated and hard to understand (see https://github.com/filecoin-project/ref-fvm/issues/1350).

To address this, I originally implemented a quick fix (https://github.com/filecoin-project/lotus/pull/10897) which addressed this in the API level but the team decided that we instead want a less hacky fix where we do not store duplicate events in the database to start with. 

This PR implements this on the DB level where depending on whether we are applying messages to revert or not we:

```
if revert: 
  mark all events events in this tipset as reverted
else
  for each event 
    if event already exists
       update event and set revert to false
    else 
       insert new event   
```

I introduced a new events migration which adds an index on (height,tipset_key) which I confirmed is used by all relevant sql statements:

```
EXPLAIN QUERY PLAN SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?
- SEARCH event USING INDEX height_tipset_key (height=? AND tipset_key=?)
EXPLAIN QUERY PLAN UPDATE event SET reverted=true WHERE height=? AND tipset_key=?
- SEARCH event USING INDEX height_tipset_key (height=? AND tipset_key=?)
EXPLAIN QUERY PLAN UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?
- SEARCH event USING INDEX height_tipset_key (height=? AND tipset_key=?)
```

Furthermore, this PR also include the migration of existing actor events to not contain duplicate events

## Test plan

I started a node with the changes in this PR and let it run for 40min (epoch 2988646-2988730). Then I started another node from master branch and then compared the generated actor events within those two epoch ranges between the two instances:

```sql
SELECT * FROM (
  SELECT e.height, e.tipset_key, e.emitter_addr, e.event_index, e.message_cid, e.message_index, e.reverted, ee.indexed, ee.flags, ee.key, ee.codec, ee.value 
  FROM event e, event_entry ee
  WHERE ee.event_id = e.id AND e.height > 2988646 AND e.height < 2988730
  EXCEPT
  SELECT e.height, e.tipset_key, e.emitter_addr, e.event_index, e.message_cid, e.message_index, e.reverted, ee.indexed, ee.flags, ee.key, ee.codec, ee.value 
  FROM events2.event e, events2.event_entry ee
  WHERE ee.event_id = e.id AND e.height > 2988646 AND e.height < 2988730
) WHERE reverted = 0

Result: 0 rows returned in 29ms
```
As expected there were no differences.

### DIFF
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -7,14 +7,17 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
 	_ "github.com/mattn/go-sqlite3"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 
+	"github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/chain/types"
 )
 
@@ -42,6 +45,8 @@ var ddls = []string{
 		reverted INTEGER NOT NULL
 	)`,
 
+	`CREATE INDEX IF NOT EXISTS height_tipset_key ON event (height,tipset_key)`,
+
 	`CREATE TABLE IF NOT EXISTS event_entry (
 		event_id INTEGER,
 		indexed INTEGER NOT NULL,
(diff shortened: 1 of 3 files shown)


chain/events/filter/index.go
```diff
@@ -56,27 +61,203 @@ var ddls = []string{
     	version UINT64 NOT NULL UNIQUE
 	)`,
 
-	// version 1.
 	`INSERT OR IGNORE INTO _meta (version) VALUES (1)`,
+	`INSERT OR IGNORE INTO _meta (version) VALUES (2)`,
```
Stebalien, line 65: Shouldn't we get rid of the previous line?

chain/events/filter/index.go
```diff
@@ -56,27 +61,203 @@ var ddls = []string{
...
+func (ei *EventIndex) migrateToVersion2(ctx context.Context, chainStore *store.ChainStore) error {
+	if chainStore == nil {
+		return errors.New("chainStore is nil")
+	}
+
+	now := time.Now()
+
+	var err error
+	var tx *sql.Tx
+	tx, err = ei.db.Begin()
```
Stebalien, line 130:
```suggestion
	tx, err := ei.db.Begin()
```

chain/events/filter/index.go
```diff
@@ -56,27 +61,203 @@ var ddls = []string{
...
+		return xerrors.Errorf("query min height: %w", err)
+	}
+
+	currTs := chainStore.GetHeaviestTipSet()
+	log.Infof("Migrating events from head to %d", minHeight.Int64)
+
+	for {
+		if int64(currTs.Height()) < minHeight.Int64 {
+			break
+		}
```
Stebalien, line 179:
```suggestion
	for int64(currTs.Height()) >= minHeight.Int64 {
```

chain/events/filter/index.go
```diff
@@ -56,27 +61,203 @@ var ddls = []string{
...
+
+		tsKey := currTs.Parents()
+		var ts *types.TipSet
+		ts, err = chainStore.GetTipSetFromKey(ctx, tsKey)
```
Stebalien, line 183:
```suggestion
		currTs, err = chainStore.GetTipSetFromKey(ctx, tsKey)
```

This lets us avoid re-assigning currTs below. Unless we want to process the first tipset (we aren't currently doing that anyways).

chain/events/filter/index.go
```diff
@@ -56,27 +61,203 @@ var ddls = []string{
...
+		}
+		log.Debugf("deleted %d events from tipset %s\n", nrRowsAffected, tsKeyCid.String())
+
+		currTs = ts
+	}
+
+	// delete all entries that have an event_id that doesn't exist (since we don't have a foreign
+	// key constraint that gives us cascading deletes)
+	var res sql.Result
+	res, err = tx.Exec("DELETE FROM event_entry WHERE event_id NOT IN (SELECT id FROM event)")
```
Stebalien, line 238: In general, avoid `var` unless it actually simplifies things. Here, we can just shadow `err`.

chain/events/filter/index_test.go
```diff
@@ -74,7 +74,7 @@ func TestEventIndexPrefillFilter(t *testing.T) {
 
 	dbPath := filepath.Join(workDir, "actorevents.db")
 
-	ei, err := NewEventIndex(dbPath)
+	ei, err := NewEventIndex(context.TODO(), dbPath, nil)
```
Stebalien, line 77: For tests, just use `context.Background()`. Use `context.TODO()` for code where we actually plan on replacing it with a real context.

Stebalien (commented):
First pass.

chain/events/filter/index.go
fridrik01: I am doing this intentionally as to not shadow `err` so the defer block that runs in the end will detect whether it needs to rollback the transaction or not. If I shadow it, then it it won't work. Explicitly rollbacking the transaction on error is a good practice to reduce locks held, minimize the time transaction is open, etc.

chain/events/filter/index.go
fridrik01: We could but I kept it so we have an explicit trace of all versions that we have used, but yes it is not necessary.

Stebalien: Fair enough.

chain/events/filter/index.go
Stebalien: Ah, ok. So, for the defer block, I'd:

1. Name the returned error something like `_reterr`.
2. Check `_reterr` in the defer block.

Stebalien: E.g., see https://github.com/filecoin-project/lotus/blob/8b849e6ad983c4020b94e582ca645d90c4874830/cmd/lotus-sim/simulation/simulation.go#L184-L199

fridrik01: Oh, cool, I didnt know that was possible

fridrik01: Alternatively we can just call `defer tx.Rollback()` as soon as we start the transaction. If we committed the transaction, then that calling Rollback is a no-op, and otherwise we rollback. Then we don't need to have named returns

chain/events/filter/index.go
```diff
@@ -56,27 +61,177 @@ var ddls = []string{
...
+		return xerrors.Errorf("prepare stmtRestoreEvent: %w", err)
+	}
+
+	return nil
+}
+
+func (ei *EventIndex) migrateToVersion2(ctx context.Context, chainStore *store.ChainStore) error {
+	if chainStore == nil {
+		return errors.New("chainStore is nil")
+	}
```
Stebalien, line 124: Unless we pass a nil chainstore on a happy path, I wouldn't bother with this check. From a code-style standpoint, I see this and start wondering if there might be reasonable cases where this is null.

chain/events/filter/index.go
```diff
@@ -170,42 +361,86 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
...
+						entry.Codec,                 // codec
+						entry.Value,                 // value
+					)
+					if err != nil {
+						return xerrors.Errorf("exec insert entry: %w", err)
+					}
+				}
+			} else {
+				// event already exists, lets mark it as not reverted
+				res, err := tx.Stmt(ei.stmtRestoreEvent).Exec(
```
Stebalien, line 416: _Technically_, due to the transaction semantics, if any events are recorded, all should be recorded. So we should be able to check:

1. Are any events recorded for this tipset. If so, unmark them as reverted and return.
2. Otherwise, insert them.

But, we can also keep the current logic unless it becomes a performance issue.

Stebalien (approved):
Couple of small things but they're up to you. Otherwise, LGTM!

chain/events/filter/index.go
Stebalien: Yeah, that's definitely the best approach. Named returns are... best avoided unless you really need them.

Stebalien:
You'll have to tell the linter to "go away" by appending `//nolint:errcheck` to the deferred reverts.

arajasek:
@fridrik01 @Stebalien So, the migration in this PR is...slow. On @TippyFlitsUK's machine, it seemed to be happening at a pace of less than 1 per second, which would take (at least) 400k seconds, or 5 days, for anyone with history going back to FEVM launch. Further, it's blocking the daemon's API coming alive, which is an issue for SPs, RPC service providers, etc. Can we:

- Speed up the migration? :D 
- Run the migration in the background, without blocking the daemon's? That way you have the node up and running, even if the EventsAPI subsystem is down
- Run the migration entirely offline?

Stebalien:
One second per epoch is really strange. It should be milliseconds (we have a constant number of sql queries per epoch). I've filed an issue and will investigate: https://github.com/filecoin-project/lotus/issues/11056.

Parallelism should be the easiest way to speed this up, but even then, I'm not sure how much parallelism we can get out of sqlite here.

PR #11016, opened 2023-06-28
refactor: Make all validation error actions explicit

shrenujbansal:
## Proposed Changes

- Make all validation error actions explicit in the switch check
- Move ErrNotEnoughFunds to return pubsub IGNORE since this can be possible from an innocent actor if a set of transactions gets executed out of order leaving no funds for one of them
- Add a comment when adding new errors to also include the error in the switch check

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/sub/incoming.go
+++ b/chain/sub/incoming.go
@@ -350,22 +350,16 @@ func (mv *MessageValidator) Validate(ctx context.Context, pid peer.ID, msg *pubs
 		)
 		recordFailure(ctx, metrics.MessageValidationFailure, "add")
 		switch {
-		case xerrors.Is(err, messagepool.ErrSoftValidationFailure):
+		case xerrors.Is(err, messagepool.ErrMessageTooBig):
 			fallthrough
-		case xerrors.Is(err, messagepool.ErrRBFTooLowPremium):
+		case xerrors.Is(err, messagepool.ErrMessageValueTooHigh):
 			fallthrough
-		case xerrors.Is(err, messagepool.ErrTooManyPendingMessages):
+		case xerrors.Is(err, messagepool.ErrNotEnoughFunds):
 			fallthrough
-		case xerrors.Is(err, messagepool.ErrNonceGap):
-			fallthrough
-		case xerrors.Is(err, messagepool.ErrGasFeeCapTooLow):
-			fallthrough
-		case xerrors.Is(err, messagepool.ErrNonceTooLow):
-			fallthrough
-		case xerrors.Is(err, messagepool.ErrExistingNonce):
-			return pubsub.ValidationIgnore
-		default:
+		case xerrors.Is(err, messagepool.ErrInvalidToAddr):
 			return pubsub.ValidationReject
+		default:
+			return pubsub.ValidationIgnore
 		}
 	}
 
--- a/chain/sub/incoming.go
+++ b/chain/sub/incoming.go
@@ -350,16 +350,32 @@ func (mv *MessageValidator) Validate(ctx context.Context, pid peer.ID, msg *pubs
 		)
 		recordFailure(ctx, metrics.MessageValidationFailure, "add")
 		switch {
+
+		case xerrors.Is(err, messagepool.ErrSoftValidationFailure):
+			fallthrough
+		case xerrors.Is(err, messagepool.ErrRBFTooLowPremium):
+			fallthrough
+		case xerrors.Is(err, messagepool.ErrTooManyPendingMessages):
+			fallthrough
+		case xerrors.Is(err, messagepool.ErrNonceGap):
+			fallthrough
+		case xerrors.Is(err, messagepool.ErrGasFeeCapTooLow):
+			fallthrough
+		case xerrors.Is(err, messagepool.ErrNonceTooLow):
+			fallthrough
+		case xerrors.Is(err, messagepool.ErrNotEnoughFunds):
+			fallthrough
+		case xerrors.Is(err, messagepool.ErrExistingNonce):
+			return pubsub.ValidationIgnore
+
 		case xerrors.Is(err, messagepool.ErrMessageTooBig):
 			fallthrough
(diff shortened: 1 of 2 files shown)


chain/sub/incoming.go
```diff
@@ -350,22 +350,16 @@ func (mv *MessageValidator) Validate(ctx context.Context, pid peer.ID, msg *pubs
...
-		case xerrors.Is(err, messagepool.ErrNonceGap):
-			fallthrough
-		case xerrors.Is(err, messagepool.ErrGasFeeCapTooLow):
-			fallthrough
-		case xerrors.Is(err, messagepool.ErrNonceTooLow):
-			fallthrough
-		case xerrors.Is(err, messagepool.ErrExistingNonce):
-			return pubsub.ValidationIgnore
-		default:
+		case xerrors.Is(err, messagepool.ErrInvalidToAddr):
```
Stebalien, line 359: I'm not sure if any of these should be "reject" as they can happen during forks/syncing.

shrenujbansal: To be clear, this is exactly the current behavior. Below are the errors we mark as reject:
ErrMessageTooBig
ErrMessageValueTooHigh
ErrNotEnoughFunds
ErrInvalidToAddr

Are you saying none of these should be marked as reject? @Stebalien

Stebalien: Yeah, I know. I'm saying that they only "too big" should likely be an error. But we should check this with @arajasek first.

arajasek: I think they're all fine to treat as REJECT, except `ErrNotEnoughFunds`.

`ErrMessageValueTooHigh` is trying to send more than 2B FIL, while `ErrInvalidToAddr` is just checking for non-nil recipient, both of which seem fine to REJECT.

Agreed that `ErrNotEnoughFunds` should be an IGNORE, though.

Stebalien: Ah, I see. Sg.

shrenujbansal: Curious, why would ErrNotEnoughFunds be an IGNORE? Wouldn't you expect the sender to have done some basic validation before sending out the msg?

Stebalien: Whether or not there are enough available funds depends on the tipset against which the message is checked, other messages in the message pool, etc.

shrenujbansal: I see, so you could have funds available when the txn was created but subsequently created txns which also used up the funds could have been executed first leaving insufficient funds for the first one. Is that right?

magik6k: Pretty sure that we want ErrNotEnoughFunds to be reject, otherwise I can just spam the network with RBF messages from an address with barely any fil

arajasek: @magik6k But none of those messages would actually get added to anyone's mempools (assuming you don't have the FIL for it). You'll just spam and get ignored?

Sounds like you're maybe saying `ErrRBFTooLowPremium` should be a reject? Which I...might agree with...maybe?

shrenujbansal:
> Unfortunately, a lowered peer score is the _smaller_ penalty here. What you really don't want is to be _not_ punishing peers that spam you invalid messages (eg. aren't validly signed, aren't actually messages but garbage data, etc.)

Could you elaborate on this? Why is a lowered peer score the smaller penalty when we've noticed especially in this recent case, that it has drastic implications for a peer not actually doing anything wrong? 
If we decide to ignore a newly added error, it would mean minimal change to the peer score and not adding that msg to the mpool, so why would that be considered a bigger penalty?

Bringing the erroneous cases to be stated explicitly has benefits as can be witnessed in this very PR discussion. Previously, most people weren't even aware of whats being rejected
If however, there are security implications, I would say a good middle ground would be to define a static map where each error has the validation action (ignore/reject) is stated explicitly
Thoughts? @arajasek

chain/sub/incoming.go
magik6k: Hmm, I guess if we ignore, those messages won't get very far anyways, so it either doesn't really matter, or ignore is leaning on a slightly safer side when it comes to pubsub scoring.

arajasek:
> Could you elaborate on this? Why is a lowered peer score the smaller penalty when we've noticed especially in this recent case, that it has drastic implications for a peer not actually doing anything wrong? If we decide to ignore a newly added error, it would mean minimal change to the peer score and not adding that msg to the mpool, so why would that be considered a bigger penalty?

Because otherwise I (an attacker) can send you an endless stream of bad messages, and have you spend all your time validating (invalid) signatures / reading state / generally wasting time. I think you need to be built to respond to bad actors, assume errors are malicious, and only carve out very specific exceptions.

> 
> Bringing the erroneous cases to be stated explicitly has benefits as can be witnessed in this very PR discussion. Previously, most people weren't even aware of whats being rejected If however, there are security implications, I would say a good middle ground would be to define a static map where each error has the validation action (ignore/reject) stated explicitly Thoughts? @arajasek

I'm not sure how much a static map achieves, because you'll still have some "default" action you take for errors that aren't defined in that map. I do think this winds up being a binary decision -- either REJECT all errors, with a carveout for specific typed IGNOREs, or IGNORE all errors, with a carveout for specific typed REJECTs.

shrenujbansal:
> Because otherwise I (an attacker) can send you an endless stream of bad messages, and have you spend all your time validating (invalid) signatures / reading state / generally wasting time. I think you need to be built to respond to bad actors, assume errors are malicious, and only carve out very specific exceptions

This makes sense. However, here https://github.com/filecoin-project/lotus/pull/11016#discussion_r1246724524 you mentioned that for an ErrNotEnoughFunds error we could set it as IGNORE because even if someone spams you it doesn't add to the mpool
Couldn't an attacker send you an endless stream of msgs where the account doesn't have enough funds and you'd spend all your time validating bad msgs then?
Also, what's to preclude this from happening with the other types of errors we mark as IGNORE today?

jennijuju:
No tests?

arajasek (approved):
LGTM

Issue #10958, opened 2023-06-06
flag to remove existing repo data on snapshot import


magik6k:
Something so that when reimporting a snapshot users wouldn't have to manually have to remove chain data directories.

PR #11032, opened 2023-06-30
feat: snapshot: remove existing chain

rjan90:
## Related Issues
Closes: https://github.com/filecoin-project/lotus/issues/10958

## Proposed Changes
Add a flag (`remove-existing-chain`) to remove the existing chain data on import of a snapshot, reducing manual operation needed when node operators want to prune their current chain.

## Additional Info
**Some open questions:**
- Do we only want the removal of the chain data to be possible when importing a snapshot?
   - Do we want it to default to true when importing a snapshot?
- Do we want to clear the chain data dirs, or delete it?
   - Are there any edge-cases where deleting the chain data dir is bad before importing a snapshot? 

**Testing:**
- [x] Tested on a node without SplitStore
- [x] Tested on a node with SplitStore

**Docs (todo):**
- [ ] Update this section: https://lotus.filecoin.io/lotus/manage/chain-management/#compacting-the-chain-data

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus/daemon.go
+++ b/cmd/lotus/daemon.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime/pprof"
 	"strings"
 
@@ -27,6 +28,7 @@ import (
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
+	"go.uber.org/multierr"
 	"golang.org/x/xerrors"
 	"gopkg.in/cheggaaa/pb.v1"
 
@@ -127,6 +129,10 @@ var DaemonCmd = &cli.Command{
 		&cli.StringFlag{
 			Name:  "import-snapshot",
 			Usage: "import chain state from a given chain export file or url",
+		},
+		&cli.BoolFlag{
+			Name:  "remove-existing-chain",
+			Usage: "remove existing chain and splitstore data on a snapshot-import",
 		},
 		&cli.BoolFlag{
 			Name:  "halt-after-import",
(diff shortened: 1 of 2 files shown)


rjan90:
> Do we want it to default to true when importing a snapshot?

With regards to this, I defaulting it to true would help a lot of new users which may not be familiar with the need to remove the existing chain data before importing a Snapshot. But if we want to go in this direction, I think we need to only make it possible when only when importing a snapshot.

ZenGround0 (commented):
Overall looks good, has it been tested?

rjan90:
> Overall looks good, has it been tested?

It has been tested on a node running with SplitStore, and on a node running without SplitStore and it cleared both successfully before starting the import of the SnapShot.

Two question I have for you @ZenGround0 is:
- What is the difference between `clearSplitstoreDir` and `deleteSplitstoreKeys` functions in the `lotus-shedsplitstore clear` command you created are (I borrowed code from that)? I currently use `clearSplitstoreDir` as it seemed most applicable for what we wanted to achieve here.
- Are there any edge-cases where deleting the SplitStore dir is wrong vs just its content. Afaik the folder is created if its not there.

rjan90:
Updated the PR with some changes based on feedback/reviews:
- Removed unused clearSplitStoreDir function ()
- Originally I had this: https://github.com/filecoin-project/lotus/blob/5201744199777772208d83ea5938d55970cd27e2/cmd/lotus/daemon.go#L313-L317
But Andy shared that `If .Config() panics,  close won't happen here`. So I extracted the removeExistingChain function, and used defer to ensure that the lockedRepo is closed, even if a panic occurs. (https://github.com/filecoin-project/lotus/pull/11032/commits/dedfa143ce18186f296d0ec98c67817ab3f6b817)
- Fixed errCheck

rjan90:
**Testing on a SplitStore-node:**
```
[Chainstore]
  # type: bool
  # env var: LOTUS_CHAINSTORE_ENABLESPLITSTORE
  EnableSplitstore = true
```

See that there is data in the chain-folder:
```
/mnt/lotuschain/.lotus/datastore/chain$ ls
000000.vlog  000010.vlog  000020.vlog  000030.vlog  000040.vlog  000050.vlog  000060.vlog  000070.vlog  000080.vlog  000090.vlog  000320.sst  000453.sst  000522.sst  000555.sst  000605.sst  000615.sst  000629.sst  000639.sst
-----
```

See that there is data in the splitstore-folder:
```
/mnt/lotuschain/.lotus/datastore/splitstore$ ls
hot.badger  markset.badger
```

Run `lotus daemon --import-snapshot /home/3005040_2023_07_04T08_00_00Z.car.zst --remove-existing-chain=true --halt-after-import`

````
....
2023-07-04T15:24:02.887+0200	INFO	main	lotus/daemon.go:774	removing splitstore directory...
2023-07-04T15:24:03.709+0200	INFO	main	lotus/daemon.go:787	removing chain directory:/mnt/lotuschain/.lotus/datastore/chain
2023-07-04T15:24:12.949+0200	INFO	main	lotus/daemon.go:794	chain and splitstore data have been removed
......
2023-07-04T15:24:13.058+0200	INFO	main	lotus/daemon.go:543	importing chain from /home/3005040_2023_07_04T08_00_00Z.car.zst...
 382.48 MiB / 49.02 GiB [=>--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------]   0.76% 21.15 MiB/s 39m10s
````

Check that existing chain data was removed before the import-snapshot started:
```
/mnt/lotuschain/.lotus/datastore$ ls
chain  client  metadata  staging
```

And
```
/mnt/lotuschain/.lotus/datastore/chain$ ls
000000.vlog  KEYREGISTRY  LOCK  MANIFEST
```

ZenGround0 (approved):
Thanks for bearing with the late review here

> What is the difference between clearSplitstoreDir and deleteSplitstoreKeys 

I forget what I was doing, the `clear..` function doesn't delete the top level dir, you used the simpler / better one.  One thing however: I think we should confirm that if you delete the splitstore data and run the node with splitstore then it correctly recreates the splitstore dir on startup.  I think you might have already tested this, just double checking.

> Docs (todo):

Since we're not making this the default I think we're good to merge without changing chain management docs.

PR #11060, opened 2023-07-08
feat: slasher: improve UX

arajasek:
Minor improvement to the UX of the slasher to fail if  a user attempts to run the slasher, but doesn't supply a path for the database.


fridrik01 (approved):
👍

PR #11062, opened 2023-07-11
Sturdydb

snadrus:
## Proposed Changes
Introduces SturdyDB, a DB adapter that 
  A. handles multiple Yugabyte instances, 
  B. prevents SQL injection, 
  C. automatically manages version despite a non-linear commit history, 
  D. allows parallel integration test runs
  E. Keeps precise Prometheus metrics, and 
  F. has powerful helpers (QueryRow, Select, BeginTransaction) with examples.


node/config/types.go
```diff
@@ -61,6 +61,8 @@ type StorageMiner struct {
 	Fees          MinerFeeConfig
 	Addresses     MinerAddressConfig
 	DAGStore      DAGStoreConfig
+
+	SturdyDB
```
magik6k, line 65: This should be a named field, otherwise the fields will not be in their own toml section
```suggestion
	SturdyDB SturdyDB
```

Also for the main config I'd avoid codenames, those are usually going to confuse new users.

Could call this field `ClusterDB` / `ClusterCoordinator` / something that makes it clear that this is configuring a database that can be shared between many lotus-miner instances and is not an optional feature

node/config/types.go
```diff
@@ -732,3 +734,25 @@ type IndexConfig struct {
...
+
+	// The database (logical partition) within Yugabyte. Blank for default.
+	Database string
+
+	// The port to find Yugabyte. Blank for default.
+	Port string
+
+	// ITest is for optimized integration testing and not
+	// for production. Blank for default production configuration.
+	ITest string
```
magik6k, line 757: Shouldn't need to be here

node/impl/storminer.go
```diff
@@ -122,6 +123,8 @@ type StorageMinerAPI struct {
 	GetSealingConfigFunc                        dtypes.GetSealingConfigFunc                        `optional:"true"`
 	GetExpectedSealDurationFunc                 dtypes.GetExpectedSealDurationFunc                 `optional:"true"`
 	SetExpectedSealDurationFunc                 dtypes.SetExpectedSealDurationFunc                 `optional:"true"`
+
+	SturdyDB *sturdydb.DB
```
magik6k, line 127: Probably shouldn't need to be here unless we want to add RPCs which call this

node/builder_miner.go
```diff
@@ -231,6 +232,7 @@ func ConfigStorageMiner(c interface{}) Option {
 		Override(new(config.SealerConfig), cfg.Storage),
 		Override(new(config.ProvingConfig), cfg.Proving),
 		Override(new(*ctladdr.AddressSelector), modules.AddressSelector(&cfg.Addresses)),
+		Override(new(*sturdydb.DB), sturdydb.NewFromConfig),
```
magik6k, line 235: NewFromConfig should be accepting the config as a parameter and it should return the constructor with the config
```suggestion
		Override(new(*sturdydb.DB), sturdydb.NewFromConfig(cfg.SturdyDB, false)),
```

The second `false` param can be for the ITest flag, which we can just set in the itest ensemble

extern/go-state-types
magik6k, line 1: Committed by accident?

extern/filecoin-ffi
magik6k, line 1: Committed by accident?

node/config/types.go
```diff
@@ -732,3 +734,25 @@ type IndexConfig struct {
 	// EnableMsgIndex enables indexing of messages on chain.
 	EnableMsgIndex bool
 }
+
```
shrenujbansal, line 737: Please run make gen and make docsgen-cli to generate the updated config files

documentation/en/default-lotus-miner-config.toml
```diff
@@ -1,3 +1,22 @@
+# env var: LOTUS__HOSTS
```
shrenujbansal, line 1: This should be autogenerated by "make gen"

node/config/types.go
snadrus: It's a schema picker so we can parallelize integration tests...
1. simply for single-binary tests.
2. for multi-binary tests using configuration.

PR #11068, opened 2023-07-11
Sturdydb

snadrus:
Proposed Changes
Introduces SturdyDB, a DB adapter that

1. handles multiple Yugabyte instances,
2. prevents SQL injection,
3. automatically manages version despite a non-linear commit history,
4. allows parallel integration test runs
5. Keeps precise Prometheus metrics, and
6. has powerful helpers (QueryRow, Select, BeginTransaction) with examples.

Concerns

1. Configuration is not right yet.
2. Once consumed, integration tests will need a database for this.


snadrus:
I see requests for updates in the draft PR. I will bring those over here this afternoon once everyone has their turn on commenting there.

PR #11069, opened 2023-07-12
feat/clusterdb

snadrus:
**Proposed Changes**
Introduces HarmonyDB, a DB adapter that

1. handles multiple Yugabyte instances,
2. prevents SQL injection,
3. automatically manages version despite a non-linear commit history,
4. allows parallel integration test runs
5. keeps precise Prometheus metrics, and
6. has powerful helpers (QueryRow, Select, BeginTransaction) with examples.
7. Includes changes to integration test framework so testing is now trivial both on circleci and for locally-installed default yugabyte servers.


node/config/types.go
```diff
@@ -732,3 +734,25 @@ type IndexConfig struct {
...
+	Username string
+
+	// The password for the related username. Blank for default.
+	Password string
+
+	// The database (logical partition) within Yugabyte. Blank for default.
+	Database string
+
+	// The port to find Yugabyte. Blank for default.
+	Port string
```
shrenujbansal, line 753: Why is Port a string?

snadrus: named ports are a thing

extern/filecoin-ffi
magik6k, line 1: looks accidentally committed

node/config/types.go
```diff
@@ -732,3 +734,25 @@ type IndexConfig struct {
...
+
+	// ITest is for optimized integration testing and not
+	// for production. Blank for default production configuration.
+	ITest string
```
magik6k, line 757: This shouldn't be needed

lib/harmony/harmonydb/harmonydb.go
```diff
@@ -0,0 +1,266 @@
...
+	"fmt"
+	"math/rand"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/jackc/pgx/v5"
```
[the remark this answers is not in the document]
snadrus: Since we have pgx v5 and they forked v4, I'm not sure what we are missing. 
But between already supporting fallback and allowing SPs to set the initial server, I don't feel a big need to move to this, especially since only 1 db instance will be common at the beginning.

Issue #11035, opened 2023-06-30
Confirming and Quantifying AMT churn problems


ZenGround0:
## User Story

Because of work on #10884 we believe we understand what is going on with the unexpected filecoin state drop.  The problem looks like the AMT.  Our hypothesis is that its overhead nodes are a large part of its structure and a large portion of them are being rewritten very often with market cron updates.  We need to confirm this hypothesis to make sure there is not some other bug in the market actor or AMT that we are missing.

We have a proposed FIP to improve the market cron [in the longer term](https://github.com/filecoin-project/FIPs/discussions/638#discussion-4887671).  If our hypothesis is confirmed then it will confirm that the proposal will work and further support its utility by demonstrating that it will save 4 G a day on chain store / snapshot size.

## Acceptance criteria

- [x] Confirm that this level of massive write amplification comes from overhead nodes as expected and is and reproducible from rewriting AMT values.
- [x] If hypothesis correct document support for long-term market cron fix in [the existing FIP discussion](https://github.com/filecoin-project/FIPs/discussions/638#discussion-4887671).  If not, reevaluate the long term market cron fix
- [x] If hypothesis correct document this datastructure behavior so that the community can be aware of the problem and make better data structure choices in future protocol changes.  Capture in a new FIP discussion about datastructures and the AMT github repo

## Technical Breakdown

1. Investigate AMT code and confirm that this checks out from the inspection
2. Write a basic tool that does market cron style processing on an AMT of similar size to States AMT, confirm that the expected amount of  data is rewritten.

## Background

With help from #10884 we have learned that the nv19 snapshot size reduction is explained by a massive reduction (~55 GB / 2000 epochs) in the market actor's States AMT.  The market actor's States AMT is < 500 MB. In less than a day of syncing all lotus nodes essentially rewrote the whole AMT's size 100 times.  However the protocol logic in cron processing dictates a rewrite of the AMT data once a day.  Our datastruture choice is ~100x less efficient than it should be.

Today's situation is 10x better (now the market actor's state AMT is ~4GB) because we've slowed market cron processing by 30x.  However it is still noticeably inefficient compared to its actual size - 500MB.  Every 2000 epochs AMT churn generates > 4 GB, which is still ~10x the size of the States AMT and by far the single largest contributor to churn.

jennijuju:
https://github.com/filecoin-project/FIPs/discussions/638#discussioncomment-6440054

ZenGround0:
Trail of outputs
[...]

PR #11075, opened 2023-07-13
feat:lotus-bench:AMT benchmarking

ZenGround0:
## Related Issues

#11035 

## Proposed Changes

`lotus-bench amt` creates an in memory amt and measures a breakdown of its size and node count between link and value nodes.  It further measures the size of churn when overwriting a portion of the amt.

## Additional Info

This doesn't need any additional testing, I've used the tool enough to see that it generally works and its results make sense in the context of the problem I used it to solve.  

If anyone cared to do so they could generalize the amt setup for different bitwidths and sizes.  This was not valuable for the problem I was trying to solve so I have deferred doing this.

## Checklist

  - [x] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials): https://github.com/filecoin-project/lotus/discussions/11074
- [ ] CI is green


cmd/lotus-bench/amt_internal.go
```diff
@@ -0,0 +1,309 @@
+package main
+
```
ZenGround0, line 2: This is just copy pasted from AMT package. Couldnt import because it is internal over there.

snissn: Can you explain/link directly to the AMT package?

ZenGround0: And this is just code for reading/writing from go memory structures to cbor bytes

snissn: I think the rest of the code looks great, just this question: Does it make sense to add a comment to the top of the file with this link and maybe an explanation of why you are adding it to this repo instead of linking/import it?

snissn: Also there's a lint issue

snissn (approved):
:+1:

PR #11081, opened 2023-07-17
feat: vm: allow raw "cbor" in state and use the new go-multicodec

Stebalien:
## Proposed Changes

1. Switch to go-multicodec as the source of multicodec code information. This gives us a central, generated source of multicodec codes.
2. Use this library inside the VM and shapshot logic to consistently allow CBOR, in addition to DagCBOR.
3. Remove the hard-coded CBOR constant.

This doesn't add any functionality in practice, it's just a post FEVM cleanup.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/store/snapshot.go
+++ b/chain/store/snapshot.go
@@ -15,7 +15,7 @@ import (
 	"github.com/ipld/go-car"
 	carutil "github.com/ipld/go-car/util"
 	carv2 "github.com/ipld/go-car/v2"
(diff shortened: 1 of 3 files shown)


Stebalien:
I've tested this on calibnet and snapshots appear to round-trip. I.e.:

1. Import snapshot.
2. Export snapshot at the same height.
3. Check hashes and they matched.

PR #11100, opened 2023-07-25
Add new tracing API

fridrik01:
## Related Issues
Related: https://github.com/filecoin-project/ref-fvm/issues/1793 and https://github.com/filecoin-project/fvm-pm/issues/613

Relevant PR: https://github.com/filecoin-project/ref-fvm/pull/1823 and https://github.com/filecoin-project/filecoin-ffi/pull/411

## Proposed Changes
This PR adds a new tracing API to lotus RPC which should unblock Blockscout/ngram which rely on this feature.  

The following two RPC methods will be exported alongside other eth rpc methods:
- [trace_block](https://docs.alchemy.com/reference/trace-block)
- [trace_replayBlockTransactions](https://docs.alchemy.com/reference/trace-replayblocktransactions)

I used the OpenEthereum/Nethermind/Erigon client version since I liked that API more than what Geth implements and implemented the two required trace methods [as listed](https://docs.blockscout.com/for-developers/information-and-settings/node-tracing-json-rpc-requirements#enable-tracing-to-fetch-internal-transactions) by Blockscout.

## Additional Info
Note that I thought about adding a new lotus config that enables/disables the tracing API, but for now just made it be set any time `EnableEthRPC` is set. 

### Test plan

**Testing trace_block**

<details>
<summary>
curl -s http://localhost:1234/rpc/v1   -X POST   -H "Content-Type: application/json"   --data '{"method":"trace_block", "params":["3166855"],"id":1,"jsonrpc":"2.0"}'|jq
</summary>

```json
{
  "jsonrpc": "2.0",
  "result": [
    {
      "action": {
        "callType": "call",
        "from": "0xff0000000000000000000000000000000023f918",
        "to": "0xff000000000000000000000000000000001ced62",
        "gas": "0x1741f00",
        "input": "0x868e10c400000000000000000000000000000000000000000000000000000000000000050000000000000000000000000000000000000000000000000000000000000051000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000f285168182004081820d58c0841978f977b7804532f774a39ae2292b76808a8186f6947fd1db444069d8892b19aa56908ef0ec7b40c585ebb96f354bb27b7f9e7b7b04db7a841a7b61e746e52f1f80a5680e39a831e6d361b507df8d2119c98f364b1d62e1e9d76adbe923d50cf913d1e176c0acaf1448a84e74ccaf5c49e0ca886877538d55cbf63fdf84374b57b4ccb9d853e2d366bbb3d51d0d2198688d43d0c134ec3b64252e58d525dc883b5856d9cc18564d008be045be415cd8efb182dee4b65ba5523142ba506b121a0030527b5820fb2947b9765ce0872aaa93d286aa9f188fecf69e70be84f22be9d2bdd5c8ec760000000000000000000000000000",
        "value": "0x0"
      },
      "result": {
        "gasUsed": "0x150bad3",
        "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
      },
      "subtraces": 0,
      "traceAddress": [],
      "Type": "call",
      "blockHash": "0x0117cdaea1750423788078050cadfcfcd368c3f39e8359d2e2b3a1ce9049ce73",
      "blockNumber": 3166855,
      "transactionHash": "0xb2407c04381dedb286426d121aff82643a6b6f623f7d39111a6856bf8d2291fd",
      "transactionPosition": 1
    },
    {
      "action": {
        "callType": "call",
        "from": "0xff0000000000000000000000000000000023d0e7",
        "to": "0xff000000000000000000000000000000001d6702",
        "gas": "0x1882fda",
        "input": "0x868e10c400000000000000000000000000000000000000000000000000000000000000050000000000000000000000000000000000000000000000000000000000000051000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000f38518288182004081820d58c08984dd25b82e696b09fbfbe0bfe1b24abb87bed1291024f8fcff5ba6861bfcd36024d541128704bdeadb6390d673709592b7a7fbbd439faa76ae4010397723dca8b71123c3dc7cfd3116b94671a78de14684f502ede8bcd4fa7a1df09d2da5a70c1f503769a68e1e5790200c7a148b797455dfce3f70bb797405dbe346a8570d98edbd9752be5fdc2bbbdce03c7edffaa190a395e2c832549ed1583f5e4c50e1e940faa9756a3e570d7c75903997d2f0b30df58487d17e4b03f5f764a02d07de1a0030527b5820fb2947b9765ce0872aaa93d286aa9f188fecf69e70be84f22be9d2bdd5c8ec7600000000000000000000000000",
        "value": "0x0"
      },
      "result": {
        "gasUsed": "0x1623212",
        "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
      },
      "subtraces": 0,
      "traceAddress": [],
      "Type": "call",
      "blockHash": "0x0117cdaea1750423788078050cadfcfcd368c3f39e8359d2e2b3a1ce9049ce73",
      "blockNumber": 3166855,
      "transactionHash": "0x39c961b9d5a2d8c2b37fb7a939eae302d59ac845a710bda099dcd42ae2d52a54",
      "transactionPosition": 2
    },
    {
      "action": {
        "callType": "call",
        "from": "0xff0000000000000000000000000000000000f356",
        "to": "0xff0000000000000000000000000000000000cd1a",
        "gas": "0xbbe9a",
        "input": "0x868e10c40000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000",
        "value": "0x29a2241af62c0000"
      },
      "result": {
        "gasUsed": "0x12045b",
        "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
      },
      "subtraces": 0,
      "traceAddress": [],
      "Type": "call",
      "blockHash": "0x0117cdaea1750423788078050cadfcfcd368c3f39e8359d2e2b3a1ce9049ce73",
      "blockNumber": 3166855,
      "transactionHash": "0xc8c227a3a581949db0ba5b89a036f2341342b30af328a07218edb6385cd667d4",
      "transactionPosition": 3
    },
    {
      "action": {
        "callType": "call",
        "from": "0xff0000000000000000000000000000000000f356",
        "to": "0xff00000000000000000000000000000000010874",
        "gas": "0xbbe9a",
        "input": "0x868e10c40000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000",
        "value": "0x29a2241af62c0000"
      },
      "result": {
        "gasUsed": "0x12045b",
        "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
      },
      "subtraces": 0,
      "traceAddress": [],
      "Type": "call",
      "blockHash": "0x0117cdaea1750423788078050cadfcfcd368c3f39e8359d2e2b3a1ce9049ce73",
      "blockNumber": 3166855,
      "transactionHash": "0xba3db829b9e17b2e0d25c239e00305d3a7ba6881851b5cbe8791f4f20d289a35",
      "transactionPosition": 4
    },
    {
      "action": {
        "callType": "call",
        "from": "0xff0000000000000000000000000000000000f356",
        "to": "0xff00000000000000000000000000000000010874",
        "gas": "0xbbe9a",
        "input": "0x868e10c40000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000",
        "value": "0x29a2241af62c0000"
      },
      "result": {
        "gasUsed": "0x12045b",
        "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
      },
      "subtraces": 0,
      "traceAddress": [],
      "Type": "call",
      "blockHash": "0x0117cdaea1750423788078050cadfcfcd368c3f39e8359d2e2b3a1ce9049ce73",
      "blockNumber": 3166855,
      "transactionHash": "0xe4e433cb335fb1290d1ace74d42a5f225d9e7f9ddb1e47e799e8b1f9e7d12582",
      "transactionPosition": 5
    },

   ...
```
</details>

**Testing trace_replayBlockTransactions**

<details>
<summary>
curl -s http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"trace_replayBlockTransactions", "params":["3166855", ["trace"]],"id":1,"jsonrpc":"2.0"}'|jq 
</summary>

```json
{
  "jsonrpc": "2.0",
  "result": [
    {
      "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
      "stateDiff": null,
      "trace": [
        {
          "action": {
            "callType": "call",
            "from": "0xff0000000000000000000000000000000023f918",
            "to": "0xff000000000000000000000000000000001ced62",
            "gas": "0x1741f00",
            "input": "0x868e10c400000000000000000000000000000000000000000000000000000000000000050000000000000000000000000000000000000000000000000000000000000051000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000f285168182004081820d58c0841978f977b7804532f774a39ae2292b76808a8186f6947fd1db444069d8892b19aa56908ef0ec7b40c585ebb96f354bb27b7f9e7b7b04db7a841a7b61e746e52f1f80a5680e39a831e6d361b507df8d2119c98f364b1d62e1e9d76adbe923d50cf913d1e176c0acaf1448a84e74ccaf5c49e0ca886877538d55cbf63fdf84374b57b4ccb9d853e2d366bbb3d51d0d2198688d43d0c134ec3b64252e58d525dc883b5856d9cc18564d008be045be415cd8efb182dee4b65ba5523142ba506b121a0030527b5820fb2947b9765ce0872aaa93d286aa9f188fecf69e70be84f22be9d2bdd5c8ec760000000000000000000000000000",
            "value": "0x0"
          },
          "result": {
            "gasUsed": "0x150bad3",
            "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
          },
          "subtraces": 0,
          "traceAddress": [],
          "Type": "call"
        }
      ],
      "transactionHash": "0xb2407c04381dedb286426d121aff82643a6b6f623f7d39111a6856bf8d2291fd",
      "vmTrace": null
    },
    {
      "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
      "stateDiff": null,
      "trace": [
        {
          "action": {
            "callType": "call",
            "from": "0xff0000000000000000000000000000000023d0e7",
            "to": "0xff000000000000000000000000000000001d6702",
            "gas": "0x1882fda",
            "input": "0x868e10c400000000000000000000000000000000000000000000000000000000000000050000000000000000000000000000000000000000000000000000000000000051000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000f38518288182004081820d58c08984dd25b82e696b09fbfbe0bfe1b24abb87bed1291024f8fcff5ba6861bfcd36024d541128704bdeadb6390d673709592b7a7fbbd439faa76ae4010397723dca8b71123c3dc7cfd3116b94671a78de14684f502ede8bcd4fa7a1df09d2da5a70c1f503769a68e1e5790200c7a148b797455dfce3f70bb797405dbe346a8570d98edbd9752be5fdc2bbbdce03c7edffaa190a395e2c832549ed1583f5e4c50e1e940faa9756a3e570d7c75903997d2f0b30df58487d17e4b03f5f764a02d07de1a0030527b5820fb2947b9765ce0872aaa93d286aa9f188fecf69e70be84f22be9d2bdd5c8ec7600000000000000000000000000",
            "value": "0x0"
          },
          "result": {
            "gasUsed": "0x1623212",
            "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
          },
          "subtraces": 0,
          "traceAddress": [],
          "Type": "call"
        }
      ],
      "transactionHash": "0x39c961b9d5a2d8c2b37fb7a939eae302d59ac845a710bda099dcd42ae2d52a54",
      "vmTrace": null
    },
    {
      "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
      "stateDiff": null,
      "trace": [
        {
          "action": {
            "callType": "call",
            "from": "0xff0000000000000000000000000000000000f356",
            "to": "0xff0000000000000000000000000000000000cd1a",
            "gas": "0xbbe9a",
            "input": "0x868e10c40000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000",
            "value": "0x29a2241af62c0000"
          },
          "result": {
            "gasUsed": "0x12045b",
            "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
          },
          "subtraces": 0,
          "traceAddress": [],
          "Type": "call"
        }
      ],
      "transactionHash": "0xc8c227a3a581949db0ba5b89a036f2341342b30af328a07218edb6385cd667d4",
      "vmTrace": null
    },
    {
      "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
      "stateDiff": null,
      "trace": [
        {
          "action": {
            "callType": "call",
            "from": "0xff0000000000000000000000000000000000f356",
            "to": "0xff00000000000000000000000000000000010874",
            "gas": "0xbbe9a",
            "input": "0x868e10c40000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000",
            "value": "0x29a2241af62c0000"
          },
          "result": {
            "gasUsed": "0x12045b",
            "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
          },
          "subtraces": 0,
          "traceAddress": [],
          "Type": "call"
        }
      ],
      "transactionHash": "0xba3db829b9e17b2e0d25c239e00305d3a7ba6881851b5cbe8791f4f20d289a35",
      "vmTrace": null
    },
    {
      "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
      "stateDiff": null,
      "trace": [
        {
          "action": {
            "callType": "call",
            "from": "0xff0000000000000000000000000000000000f356",
            "to": "0xff00000000000000000000000000000000010874",
            "gas": "0xbbe9a",
            "input": "0x868e10c40000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000",
            "value": "0x29a2241af62c0000"
          },
          "result": {
            "gasUsed": "0x12045b",
            "output": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
          },
          "subtraces": 0,
          "traceAddress": [],
          "Type": "call"
        }
      ],
      "transactionHash": "0xe4e433cb335fb1290d1ace74d42a5f225d9e7f9ddb1e47e799e8b1f9e7d12582",
      "vmTrace": null
    },
    ...
```
</details>

## Checklist

- [ ] CI is green


node/impl/full/trace.go
```diff
@@ -0,0 +1,263 @@
...
+type EthTraceAPI interface {
+	TraceBlock(ctx context.Context, blkNum string) (interface{}, error)
+	TraceReplayBlockTransactions(ctx context.Context, blkNum string, traceTypes []string) (interface{}, error)
+}
+
+var (
+	_ EthTraceAPI = *new(api.FullNode)
+)
+
+type EthTrace struct {
```
magik6k, line 28: This should use `fx.In`, that would make it not need a constructor (and be consistent with other API impl structs)

Stebalien:
From our discussion, we have a few edge cases we need to handle... quite a few. Ideally, we'd handle _all_ of these cases inside the ethereum json-rpc API module in lotus without touching the FVM (except to, maybe, add additional tracing information where necessary). The FVM is otherwise supposed to have no idea that the EVM exists (except for a few hacks that will hopefully go away in the near future).

I hope I included everything... but, to be honest, we'll probably find more edge cases and I likely underspecified a few things.

# Native calls

(both from EVM actors _and_ from native actors)

In general (exceptions noted in the EVM section), we need to format native calls in a way that makes sense to Ethereum tooling. That means, if we see a call _not_ on with InvokeEVM method number (use [the constant](https://github.com/filecoin-project/go-state-types/blob/ff2ed169ff566458f2acd8b135d62e8ca27e7d0c/builtin/methods.go#L367), we need to convert the inputs & outputs into solidity "ABI".

Luckily, we already have a mechanism for this. Specifically, when an EVM actor is invoked with a method number above 1023 that's _not_ `frc42(InvokeEVM)` (3844450837), the EVM actor hits [this](https://github.com/filecoin-project/builtin-actors/blob/a77674d65d42dc3547ed2753d5e8c729971920ef/actors/evm/src/lib.rs#L420) fallback method, translates the [inputs into Solidity ABI](https://github.com/filecoin-project/builtin-actors/blob/a77674d65d42dc3547ed2753d5e8c729971920ef/actors/evm/src/lib.rs#L331-L349), then [translates the outputs](https://github.com/filecoin-project/builtin-actors/blob/a77674d65d42dc3547ed2753d5e8c729971920ef/actors/evm/src/lib.rs#L358-L403) from Solidity ABI back into a "native" return value.

Here, we need to:

1. Translate the call inputs into Solidity ABI, as above.
2. Translate the outputs _into_ Solidity ABI (reverse of what we're doing in the EVM.

Basically, we want to make native calls look like calls to a contract like our [FilecoinFallback](https://github.com/filecoin-project/builtin-actors/blob/a77674d65d42dc3547ed2753d5e8c729971920ef/actors/evm/tests/contracts/FilecoinFallback.sol#L4-L18) test.

# Native actor creation

In Filecoin, actor creation happens by sending messages (calling actors):

1. Actor A calls to the init actor on method 2 (Exec, bad name).
2. The init actor creates the target actor B then calls it on method 1 (the constructor).

Unfortunately, it's going to be difficult to make this look like an EVM contract creation but we can hack it...

Personally, I'd:

1. Collapse these two calls into a single "create".
2. Pretend that it's a `CREATE` instruction.
3.  Use `[0xFE]` as the initcode (the parameters). This bytecode is invalid so it'll never be mistaken for an EVM contract.
4. Drop any return data.

# EVM contract creation

With EVM contracts, there's an extra step

1. EVM contract A calls the EAM (Ethereum Address Manager) on method 2 (create) or 3 (create2).
5. The EAM calls the init actor on method 3 (Exec4).
6. The init actor creates the target actor B then calls it on method 1.

In the EVM trace, we need to make these look like `CREATE` and `CREATE2` calls, skipping over the EAM and Init actor.

# EVM call special casing

As noted above, there are some exceptions to the rule where we convert "native" calls into "solidity ABI" calls. Specifically, any outbound call from an EVM actor on methods 1-1023. These calls are side-effects from EVM instructions and should be dropped from the trace.

# EVM -> EVM calls

As we discussed, we need to handle EVM -> EVM calls:

1. Normal  EVM->EVM calls will look like a call to InvokeEVM (method 3844450837). We just need to decode the parameters/return value. We also need to look at the readonly flag.
2. The tricky case is DelegateCall. In that case, we need to:
    1. Look for from an EVM actor to itself on [InvokeContractDelegate](https://github.com/filecoin-project/builtin-actors/blob/a77674d65d42dc3547ed2753d5e8c729971920ef/actors/evm/src/lib.rs#L67), method 6.
    2. Search _backwards_ in the trace for a call to another actor (A) on method 3 (https://github.com/filecoin-project/builtin-actors/blob/a77674d65d42dc3547ed2753d5e8c729971920ef/actors/evm/src/lib.rs#L64).
    3. Treat this as a delegate call to actor A.

fridrik01:
@Stebalien for the EVM -> EVM calls specifically for 2) regarding DelegateCall, when you mention that we need to "search backwards in the trace for a call to another actor (A) on method 3" does that mean we need to look up the call hierarchy or  could should this be in a previous subcall at the same level (probably the latter since that is the only thing I see in the traces but just wanted to confirm) ?

Stebalien:
Same level and it _should_ be the adjacent call. Really, I'd memoize the last call on method 3 when handling calls from the EVM.

arajasek:
@fridrik01 Did you consider simply adding this into the existing "EthAPI"? This seems to fundamentally be a restatement of existing Lotus APIs like `StateReplay` etc., except Eth-friendly, and we've put those kinds of methods in the EthAPI so far.

I might be missing something.

fridrik01:
> @fridrik01 Did you consider simply adding this into the existing "EthAPI"? This seems to fundamentally be a restatement of existing Lotus APIs like `StateReplay` etc., except Eth-friendly, and we've put those kinds of methods in the EthAPI so far.
> 
> I might be missing something.

Yes, I think I started actually with the implementation first in the EthAPI but soon refactored it to keep them separate as its very tracing specific and introduces multiple types which would add complexity to an already huge file (it would blow up to 3K+ LOC with this PR).

arajasek:
> > @fridrik01 Did you consider simply adding this into the existing "EthAPI"? This seems to fundamentally be a restatement of existing Lotus APIs like `StateReplay` etc., except Eth-friendly, and we've put those kinds of methods in the EthAPI so far.
> > I might be missing something.
> 
> Yes, I think I started actually with the implementation first in the EthAPI but soon refactored it to keep them separate as its very tracing specific and introduces multiple types which would add complexity to an already huge file (it would blow up to 3K+ LOC with this PR).

Hmm, so I  definitely agree that that file is quite large already. Would you be open to doing the following:

- A prefactor that introduces a new `eth_util.go` file that just contains all the helpers. We can also move any types in the `eth.go` file to `eth_types.go` in this refactor.
- Introducing these methods in the `EthAPI`, with the large helpers put in `eth_util.go` (or even some `eth_trace.go` file?

I do think we should avoid adding a new module if we can.

fridrik01:
@arajasek I removed the module and refactored the `eth.go` so that now it almost exclusively has only the interface methods. I moved the trace related code to `eth_trace.go`, event related code to `eth_event.go` and helper/util functions to `eth_utils.go`.  This does make this PR way large, but the relevant code for review is in the two Trace methods in `eth`.go and then all of `eth_trace.go`.

api/api_full.go
```diff
@@ -863,6 +863,13 @@ type FullNode interface {
 	// Returns the client version
 	Web3ClientVersion(ctx context.Context) (string, error) //perm:read
 
+	// TraceAPI related methods
+	//
+	// Returns traces created at given block
+	TraceBlock(ctx context.Context, blkNum string) (interface{}, error) //perm:read
```
arajasek, line 869: Rename to `EthTraceBlock`?

node/impl/full/eth.go
```diff
@@ -922,62 +825,135 @@ func (a *EthModule) Web3ClientVersion(ctx context.Context) (string, error) {
...
 }
 
-func (a *EthModule) ethCallToFilecoinMessage(ctx context.Context, tx ethtypes.EthCall) (*types.Message, error) {
-	var from address.Address
-	if tx.From == nil || *tx.From == (ethtypes.EthAddress{}) {
-		// Send from the filecoin "system" address.
-		var err error
-		from, err = (ethtypes.EthAddress{}).ToFilecoinAddress()
+func (a *EthModule) TraceBlock(ctx context.Context, blkNum string) (interface{}, error) {
+	ts, err := getTipsetByBlockNr(ctx, a.Chain, blkNum, false)
```
arajasek, line 829: nit: i personally dislike "abbreviated" names like this -- would prefer `getTipsetByBlockNumber`, but feel free to ignore.

node/impl/full/eth.go
```diff
@@ -922,62 +825,135 @@ func (a *EthModule) Web3ClientVersion(ctx context.Context) (string, error) {
...
+	if err != nil {
+		return nil, err
```
arajasek, line 831:
```suggestion
		return nil, xerrors.Errorf("failed to get tipset: %w", err)
```

(we've had a LOT of problems debugging over the years when errors haven't been chained)

api/api_full.go
arajasek: And is there a reason we're returning interfaces here? Can we not say we return `[]*TraceBlock`?

node/impl/full/eth.go
```diff
@@ -922,62 +825,135 @@ func (a *EthModule) Web3ClientVersion(ctx context.Context) (string, error) {
...
+
+	blkHash, err := ethtypes.EthHashFromCid(cid)
+	if err != nil {
+		return nil, err
+	}
+
+	allTraces := make([]*TraceBlock, 0, len(trace))
+	for _, ir := range trace {
+		// ignore messages from f00
+		if ir.Msg.From.String() == "f00" {
```
arajasek, line 862:
```suggestion
		if ir.Msg.From == builtinactors.SystemActorAddr {
```

node/impl/full/eth.go
arajasek: few more places to chain errors below

node/impl/full/eth.go
```diff
@@ -922,62 +825,135 @@ func (a *EthModule) Web3ClientVersion(ctx context.Context) (string, error) {
...
+			continue
+		}
+
+		idx := -1
+		for msgIdx, msg := range msgs {
```
arajasek, line 867: While this is correct, you _should_ be able to trust the order of the ExecutionTrace. Simply start a counter at 0, and increment it for all messages not sent by the system actor.

arajasek (commented):
Partial review

node/impl/full/trace.go
```diff
@@ -0,0 +1,515 @@
...
+
+	blkHash, err := ethtypes.EthHashFromCid(cid)
+	if err != nil {
+		return nil, err
+	}
+
+	allTraces := make([]*TraceBlock, 0, len(trace))
+	for _, ir := range trace {
+		// ignore messages from f00
+		if ir.Msg.From.String() == "f00" {
```
Stebalien: Compare against https://github.com/filecoin-project/lotus/blob/cbbb76b2f095782e5c4d717b5c2603ccd41d21ff/chain/actors/builtin/builtin.go#L26

node/impl/full/trace.go
```diff
@@ -0,0 +1,515 @@
...
+			continue
+		}
+
+		idx := -1
+		for msgIdx, msg := range msgs {
+			if ir.Msg.From == msg.Message.From {
```
Stebalien: We can have multiple messages from the same sender. We need to instead use the message CIDs. I'd:

1. Build a map of message CID to index.
2. Do the lookup here.

I.e., `ir.MsgCid`, `msg.Cid`.

Stebalien: Actually, I don't _think_ this is even necessary? The index in the trace should match the index in the message list as we execute (trace) messages in-order.

node/impl/full/trace.go
```diff
@@ -0,0 +1,515 @@
...
+			continue
+		}
+
+		traces := []*Trace{}
+		err = buildTraces(&traces, nil, []int{}, ir.ExecutionTrace, int64(ts.Height()))
+		if err != nil {
+			return nil, xerrors.Errorf("failed when building traces: %w", err)
+		}
+
+		traceBlocks := make([]*TraceBlock, 0, len(trace))
```
Stebalien:
```suggestion
		traceBlocks := make([]*TraceBlock, 0, len(traces))
```

node/impl/full/trace.go
```diff
@@ -0,0 +1,515 @@
...
+
+	_, trace, err := e.StateManager.ExecutionTrace(ctx, ts)
+	if err != nil {
+		return nil, xerrors.Errorf("failed when calling ExecutionTrace: %w", err)
+	}
+
+	allTraces := make([]*TraceReplayBlockTransaction, 0, len(trace))
+	for _, ir := range trace {
+		// ignore messages from f00
+		if ir.Msg.From.String() == "f00" {
```
Stebalien: same as above.

node/impl/full/trace.go
```diff
@@ -0,0 +1,515 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if txHash == nil {
+			log.Warnf("cannot find transaction hash for cid %s", ir.MsgCid)
+			continue
+		}
+
+		t := TraceReplayBlockTransaction{
+			Output:          hex.EncodeToString(ir.MsgRct.Return),
```
Stebalien: We need to decode this return value, unfortunately:

1. If empty, treat it as empty bytes.
2. If non-empty, decode it as a cbor byte array, then hex-encode it.

Take a look at:

https://github.com/filecoin-project/lotus/blob/cbbb76b2f095782e5c4d717b5c2603ccd41d21ff/node/impl/full/eth.go#L695-L698

node/impl/full/trace.go
```diff
@@ -0,0 +1,515 @@
...
+
+	return nil
+}
+
+// buildTraces recursively builds the traces for a given ExecutionTrace by walking the subcalls
+func buildTraces(traces *[]*Trace, parent *Trace, addr []int, et types.ExecutionTrace, height int64) error {
+	trace := &Trace{
+		Action: Action{
+			From:    et.Msg.From.String(),
+			To:      et.Msg.To.String(),
```
Stebalien: We should be converting these to Eth addresses. I.e. https://github.com/filecoin-project/lotus/blob/cbbb76b2f095782e5c4d717b5c2603ccd41d21ff/chain/types/ethtypes/eth_types.go#L339

Stebalien: Hm. Also, this is going to be tricky.

1. For sub-calls, the "from" address will always be an ID address. But this is going to be _very_ confusing as users will expect f410f addresses (I.e., the "true" eth address of a contract).
2. These addresses may not be convertible into Eth-style addresses. I.e., the function I linked above may fail. In that case, we'd need to lookup the associated ID address of the actor, then convert _that_ into an eth-style address (which will always work).

There are several subtitles here. I'll think on this and get back to you tomorrow.

node/impl/full/trace.go
```diff
@@ -0,0 +1,515 @@
...
+			Gas:     ethtypes.EthUint64(et.Msg.GasLimit),
+			Input:   hex.EncodeToString(et.Msg.Params),
+			Value:   ethtypes.EthBigInt(et.Msg.Value),
+			method:  et.Msg.Method,
+			codeCid: et.Msg.CodeCid,
+		},
+		Result: Result{
+			GasUsed: ethtypes.EthUint64(et.SumGas().TotalGas),
+			Output:  hex.EncodeToString(et.MsgRct.Return),
```
Stebalien: ditto decoding.

node/impl/full/trace.go
```diff
@@ -0,0 +1,515 @@
...
+}
+
+// buildTraces recursively builds the traces for a given ExecutionTrace by walking the subcalls
+func buildTraces(traces *[]*Trace, parent *Trace, addr []int, et types.ExecutionTrace, height int64) error {
+	trace := &Trace{
+		Action: Action{
+			From:    et.Msg.From.String(),
+			To:      et.Msg.To.String(),
+			Gas:     ethtypes.EthUint64(et.Msg.GasLimit),
+			Input:   hex.EncodeToString(et.Msg.Params),
```
Stebalien: Ditto decoding.

node/impl/full/trace.go
Stebalien: https://github.com/filecoin-project/lotus/blob/cbbb76b2f095782e5c4d717b5c2603ccd41d21ff/node/impl/full/eth.go#L2047 will be helpful.

node/impl/full/trace.go
Stebalien: Well.... ok, it's worse:

1. We need to do this _if_ the method number is "InvokeEVM" (InvokeContract).
2. If we're invoking one of the "create" methods on the EAM.

Otherwise... we need to do the whole "make it look like solidity ABI" dance.

E.g., take a look at

https://github.com/filecoin-project/lotus/blob/cbbb76b2f095782e5c4d717b5c2603ccd41d21ff/node/impl/full/eth.go#L2108

https://github.com/filecoin-project/lotus/blob/cbbb76b2f095782e5c4d717b5c2603ccd41d21ff/chain/types/ethtypes/eth_transactions.go#L67

Stebalien (commented):
Initial pass. I'll continue tomorrow.

api/api_full.go
fridrik01: > Rename to `EthTraceBlock`?
I considered doing that, but skipped the Eth prefix since the method is called "trace_block" and not "eth_traceBlock", but otherwise do not strongly prefer one or the other. Let me refactor the interface as per you latter comment and if it makes sense to rename I will do that

> And is there a reason we're returning interfaces here? Can we not say we return []*TraceBlock?
Definitely, good point!

node/impl/full/eth.go
fridrik01: will change as other methods use Number instead of Nr

node/impl/full/trace.go
fridrik01: good catch!

node/impl/full/trace.go
fridrik01: > Otherwise... we need to do the whole "make it look like solidity ABI" dance.

@Stebalien could to clarify what you mean by that, its a bit unclear how the (`newEthTxFromSignedMessage`,`EthTxFromSignedEthMessage`) would apply in this case

Stebalien: I mean, take a look at how those functions interpret message parameters and return values. I.e.:

- We decode parameters and return values as "cbor bytes" (usually).
- We treat messages to the CreateExternal (4) method on the EAM specially. They are contract creations, not normal messages. I assume that the EVM traces handle this specially as well.

node/impl/full/trace.go
Stebalien: From what I can tell, yes.

Unless the contract in question somehow was created and deleted in the same epoch. Which isn't currently possible (as far as I know) but.... eh, it's probably fine to fail in that case and we can handle it later.

Stebalien: > Unless the contract in question somehow was created and deleted in the same epoch. Which isn't currently possible (as far as I know) but.... eh, it's probably fine to fail in that case and we can handle it later.

Ah... ok, so, this can happen:

1. A message creates a contract.
2. Stuff happens.
3. The top-level message gets reverted due to some failure.

In this case, the contract won't exist in state so `lookupEthAddress` may fail.

So................

First, I think we'll need to add some form of "actor resolved" event to the trace. E.g., replace `Invoke(Cid)` with `ActorResolved(ActorID, CodeCID)`. That ensures that we _always_ have enough information to fill out the trace.

Then:

1. When parsing the trace, remember the address on which the _caller_ was called and always _try_ use that as the "from" address (using `EthAddressFromFilecoinAddress`).
2. If that doesn't work, convert the ID from the trace.
3. For the top-level sender, we'll have to use `lookupEthAddress`.

Stebalien:
Writing down what we discussed in the sync: We need to make everything "look" like an Ethereum call/transaction. This means:

1. Finding everything that _is_ an Ethereum call/transaction/contract-creation, and correctly decoding parameters/return values.
2. Finding everything that _isn't_ an Ethereum call/transaction/contract-creation (i.e., all "native Filecoin" calls), and encoding the parameters & return value as a "solidity ABI" call.

Concretely:

1. If it's a call to the InvokeEVM method (on any actor), decode the input/output as bytes. To do this, you'll need to look at the _codec_ in the trace:
    1. If the codec is 0, treat it as empty bytes.
    2. If the codec is multicodec.DagCbor or multicodec.Cbor, decode it as a cbor byte array.
    3. If the codec is multicodec.Raw, treat it as raw bytes.
    4. Otherwise.... error. We don't support anything else right now.
2. If it's a call to CreateExternal on the EAM, decode the input (params) as described above in (InvokeEVM) and treat it as a contract creation (CREATE, not CREATE2). However, you'll need to decode the _return value_ specially:
    1. If the call is _successful_ (exit code 0), the return value is the address of the created actor, plus some other stuff. I _think_ we can just ignore that return value, but I'm not sure what other EVM implementations use. They _may_ treat the new actor's EVM address as the return value?
    2. If it's a _failure_ , the return value is bytes (decoded as in InvokeEVM). This return value is an error message and, again, we'll need to look at other Ethereum clients to figure out how to handle it.
3. Otherwise, handle calls as described in https://github.com/filecoin-project/lotus/pull/11100#issuecomment-1657638885 (it looks like you're already doing this, but I'll do an in-depth review pass on that now).

node/impl/full/trace.go
fridrik01: now using `lookupEthAddress`, but left a TODO for supporting the error case, for now, if an error occurs when calling `lookupEthAddress` then we fail that trace only but otherwise continue

node/impl/full/trace.go
fridrik01: @Stebalien for the ir.MsgRct.Return there is no codec included in the receipt and when I use UnmarshalCBOR something like this:

```go
var output abi.CborBytes
if len(ir.MsgRct.Return) > 0 && ir.MsgRct.ExitCode.IsSuccess() {
    if err := output.UnmarshalCBOR(bytes.NewReader(ir.MsgRct.Return)); err != nil {
	log.Warnf("failed to unmarshal cbor return value '%s': %w", string(ir.MsgRct.Return), err)
	//return nil, xerrors.Errorf("failed to unmarshal cbor return value '%s': %w", string(ir.MsgRct.Return), err)
    }
    // pad with zero bytes if smaller than 32 bytes
    output = append(make([]byte, 32-len(output), 32), output...)
}
```

I get an error `expected byte array`, similarly it does not work to use `cbg.ReadByteArray` either

Stebalien: For the codec: So, I think you can use the top-level "return" in the trace. That should be equivalent to the receipt, and should actually include the codec.

However, I _think_ the issue here is that you're decoding non-EVM return values? I.e., it's a return value from a call on a method that's not `InvokeEVM`, but some other thing. In that case, instead of _decoding_ it as cbor-bytes, we need to _encode_ it as solidity-ABI.

Or am I missing something.

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,333 @@
...
+		}
+	}
+
+	// TODO: is it OK to check this here or is this only specific to certain edge case (evm to evm)?
+	if et.Msg.ReadOnly {
+		trace.SetCallType("staticcall")
+	}
+
+	// there are several edge cases thar require special handling when displaying the traces
+	if parent != nil {
```
Stebalien, line 132: I'm curious why you're looking backwards instead of forwards. I.e., the other way to write this is:

1. Detect calls to the `Exec` method on the init actor.
2. Then run that sub-trace through a special method to find the call to the constructor.

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,333 @@
...
+	if parent != nil && builtinactors.IsEvmActor(parent.Action.CodeCid) && et.Msg.Method == builtin.MethodsEVM.InvokeContract {
+		log.Debugf("found InvokeContract call at height: %d", height)
+		trace.Action.Input, err = decodePayload(et.Msg.Params, et.Msg.ParamsCodec)
+		if err != nil {
+			return xerrors.Errorf("buildTraces: %w", err)
+		}
+		trace.Result.Output, err = decodePayload(et.MsgRct.Return, et.MsgRct.ReturnCodec)
+		if err != nil {
+			return xerrors.Errorf("buildTraces: %w", err)
+		}
```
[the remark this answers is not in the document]
fridrik01: Makes sense, but will leave as is (and add a TODO) since I expect that to be super rare (?).  I want to catch decode issues as real errors for now because that is more likely a bug in this code than a gibberish return value from actors

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,333 @@
...
+	if et.Msg.ReadOnly {
+		trace.SetCallType("staticcall")
+	}
+
+	// there are several edge cases thar require special handling when displaying the traces
+	if parent != nil {
+		// Handle Native actor creation
+		//
+		// Actor A calls to the init actor on method 2 and The init actor creates the target actor B then calls it on method 1
+		if parent.Action.To == builtin.InitActorAddr.String() &&
```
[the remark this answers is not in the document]
fridrik01: Yes, you are correct, after I changed the From/To to an eth address this won't work anymore. Will add the original address to the `EthTraceAction` making it easy to check

node/impl/full/trace.go
fridrik01: Ah, using the top level return in the trace works

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/builtin"
+	"github.com/filecoin-project/go-state-types/exitcode"
+
+	builtinactors "github.com/filecoin-project/lotus/chain/actors/builtin"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/chain/types/ethtypes"
+)
+
+func decodePayload(payload []byte, codec uint64) (ethtypes.EthBytes, error) {
```
Stebalien, line 23: doc comment

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+	to, err := lookupEthAddress(ctx, et.Msg.To, sa)
+	if err != nil {
+		log.Warnf("buildTraces: failed to lookup to address %s: %w", et.Msg.To, err)
+		return nil
+	}
+
+	trace := &ethtypes.EthTrace{
+		Action: ethtypes.EthTraceAction{
+			From:  from.String(),
+			To:    to.String(),
```
Stebalien, line 65: EthAddress if possible?

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+			FilecoinFrom:    et.Msg.From,
+			FilecoinTo:      et.Msg.To,
+			FilecoinMethod:  et.Msg.Method,
+			FilecoinCodeCid: et.Msg.CodeCid,
+		},
+		Result: ethtypes.EthTraceResult{
+			GasUsed: ethtypes.EthUint64(et.SumGas().TotalGas),
+			Output:  nil,
+		},
+		Subtraces:    len(et.Subcalls),
```
Stebalien, line 79: What about calls to special methods from the evm that we skip? I think we may need to set this _after_ the fact.

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+			trace.Result.Output, err = decodePayload(et.MsgRct.Return, et.MsgRct.ReturnCodec)
+			if err != nil {
+				return xerrors.Errorf("buildTraces: %w", err)
+			}
+		}
+
+		// treat this as a contract creation
+		trace.SetCallType("create")
+	} else {
+		trace.Action.Input, err = handleFilecoinMethodInput(et.Msg.Method, et.Msg.ParamsCodec, et.Msg.Params)
```
Stebalien, line 122: Add a comment explaining that we're going to assume a native method, but may edit/retcon it later.

node/impl/full/eth_trace.go
Stebalien: I'd at least explicitly state that we're "editing" the trace backwards.

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+		// Handle Native actor creation
+		//
+		// Actor A calls to the init actor on method 2 and The init actor creates the target actor B then calls it on method 1
+		if parent.Action.FilecoinTo == builtin.InitActorAddr &&
+			parent.Action.FilecoinMethod == builtin.MethodsInit.Exec &&
+			et.Msg.Method == builtin.MethodConstructor {
+			log.Debugf("Native actor creation! method:%d, code:%s, height:%d", et.Msg.Method, et.Msg.CodeCid.String(), height)
+			parent.SetCallType("create")
+			parent.Action.To = et.Msg.To.String()
+			parent.Action.Input = []byte{0x0, 0x0, 0x0, 0xFE}
```
Stebalien, line 148: This reads as STOP STOP STOP INVALID. I don't think we need the padding, but I'm not sure.

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+		//
+		// 1) EVM contract A calls the EAM (Ethereum Address Manager) on method 2 (create) or 3 (create2).
+		// 2) The EAM calls the init actor on method 3 (Exec4).
+		// 3) The init actor creates the target actor B then calls it on method 1.
+		if parent.Parent != nil {
+			calledCreateOnEAM := parent.Parent.Action.FilecoinTo == builtin.EthereumAddressManagerActorAddr &&
+				(parent.Parent.Action.FilecoinMethod == builtin.MethodsEAM.Create || parent.Parent.Action.FilecoinMethod == builtin.MethodsEAM.Create2)
+			eamCalledInitOnExec4 := parent.Action.FilecoinTo == builtin.InitActorAddr &&
+				parent.Action.FilecoinMethod == builtin.MethodsInit.Exec4
+			initCreatedActor := trace.Action.FilecoinMethod == builtin.MethodConstructor
```
Stebalien, line 167: switch?

Stebalien: Or maybe some more variables?

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+
+			if calledCreateOnEAM && eamCalledInitOnExec4 && initCreatedActor {
```
Stebalien, line 169: What if one of these steps fail? We should still treat it as an attempted creation.

Stebalien: We have a few cases:

1. Failures. I.e., we may fail halfway through this path for some reason and need to treat it as a failed to creation.
2. Resurrection! EAM "create" and/or "create2" may not call the init actor but may, instead call "resurrect" on an existing but dead EVM actor. Let's handle that later, but it's an important part of "EVM-style" code upgrades (which are _not_ the same as FVM code upgrades).

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+
+		// Handle EVM call special casing
+		//
+		// Any outbound call from an EVM actor on methods 1-1023 are side-effects from EVM instructions
+		// and should be dropped from the trace.
+		if builtinactors.IsEvmActor(parent.Action.FilecoinCodeCid) &&
+			et.Msg.Method > 0 &&
+			et.Msg.Method <= 1023 {
+			log.Debugf("found outbound call from an EVM actor on method 1-1023 method:%d, code:%s, height:%d", et.Msg.Method, parent.Action.FilecoinCodeCid.String(), height)
+			// TODO: if I handle this case and drop this call from the trace then I am not able to detect delegate calls
```
Stebalien, line 197: At a minimum, drop this call.

Stebalien: It could be GetBytecode, GetBytecodeHash, etc. I.e., something internal.

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+		}
+
+		// EVM -> EVM calls
+		//
+		// Check for normal EVM to EVM calls and decode the params and return values
+		if builtinactors.IsEvmActor(parent.Action.FilecoinCodeCid) &&
```
Stebalien, line 203: I think we can drop this.

node/impl/full/eth_trace.go
Stebalien: Some field in the parent for "last getbytecode"?

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,339 @@
...
+		// Handle delegate calls
+		//
+		// 1) Look for trace from an EVM actor to itself on InvokeContractDelegate, method 6.
+		// 2) Check that the previous trace calls another actor on method 3 (GetByteCode) and they are at the same level (same parent)
+		// 3) Treat this as a delegate call to actor A.
+		if trace.Action.From == trace.Action.To &&
+			trace.Action.FilecoinMethod == builtin.MethodsEVM.InvokeContractDelegate &&
+			len(*traces) > 0 {
+			log.Debugf("found delegate call, height: %d", height)
+			prev := (*traces)[len(*traces)-1]
```
Stebalien, line 228: We'd look at `parent.LastGetBytecode` instead of looking at the trace.

node/impl/full/eth_trace.go
Stebalien: I.e., need to drop all these calls _but_ we also need to be able to lookup the "last" GetBytecode when we do the delegate call.

Stebalien: I.e., we need to record `CalledGetBytecodeOn`

Stebalien: And remember to skip all the children of this call as well. I.e., we'll _always_ early-return after handling this call.

node/impl/full/eth_trace.go
fridrik01: Ah, I probably have confused my byte-2-byte testing which I did against the builtinactors version of handleFilecoinMethodInput/handle_filecoin_method_input to this Go versions I have here, and not specifically to what to return here!

node/impl/full/eth_trace.go
fridrik01: Added a TODO for this

node/impl/full/eth_trace.go
fridrik01: Done!

fridrik01:
@Stebalien Updated the PR with changes per your feedback, thanks!

chain/types/ethtypes/eth_types.go
```diff
@@ -938,7 +938,8 @@ type EthTrace struct {
 	TraceAddress []int          `json:"traceAddress"`
 	Type         string         `json:"Type"`
 
-	Parent *EthTrace `json:"-"`
+	Parent       *EthTrace `json:"-"`
+	LastByteCode *EthTrace `json:"-"`
```
Stebalien, line 942: Let's just add a comment here to explain WTF this is and why we have it.

node/impl/full/eth_trace.go
```diff
@@ -186,53 +192,44 @@ func buildTraces(ctx context.Context, traces *[]*ethtypes.EthTrace, parent *etht
...
-			len(*traces) > 0 {
-			log.Debugf("found delegate call, height: %d", height)
-			prev := (*traces)[len(*traces)-1]
+		if parent.LastByteCode != nil && trace.Action.From == trace.Action.To &&
+			trace.Action.FilecoinMethod == builtin.MethodsEVM.InvokeContractDelegate {
+			log.Debugf("COND7 found delegate call, height: %d", height)
+			prev := parent.LastByteCode
 			if prev.Action.From == trace.Action.From && prev.Action.FilecoinMethod == builtin.MethodsEVM.GetBytecode && prev.Parent == trace.Parent {
 				trace.SetCallType("delegatecall")
 				trace.Action.To = prev.Action.To
```
Stebalien, line 206: Hm. We need to parse the input and return value here:

1. Input is https://github.com/filecoin-project/builtin-actors/blob/fe72aa8e14bc566d661d47625dcdbdd960bf4525/actors/evm/src/types.rs#L46-L55. All we care about is the `input` field of that struct.
2. The output is just cbor-encoded bytes (we need to decode it the same way we decode the output of InvokeEVM).

node/impl/full/eth_trace.go
```diff
@@ -186,53 +192,44 @@ func buildTraces(ctx context.Context, traces *[]*ethtypes.EthTrace, parent *etht
...
 			}
+		} else {
+			// Handle EVM call special casing
+			//
+			// Any outbound call from an EVM actor on methods 1-1023 are side-effects from EVM instructions
+			// and should be dropped from the trace.
+			if builtinactors.IsEvmActor(parent.Action.FilecoinCodeCid) &&
```
Stebalien, line 213: Both this case and the `InvokeContractDelegate` are EVM special cases. We should put this check on the outside.

Stebalien (commented):
Ok, I think those are the final comments.

node/impl/full/eth_trace.go
fridrik01: Done!

fridrik01:
@Stebalien Cool, thanks for the quick review! I updated the PR per your most recent comments!

With the change of storing pointer to previous calls to GetBytecode I now can properly detect InvokeContractDelegate calls which didn't work before. Also tested that unmarshalling DelegateCallParams works as expected.

node/impl/full/eth_trace.go
```diff
@@ -0,0 +1,351 @@
...
+			parent.Action.FilecoinMethod == builtin.MethodsInit.Exec &&
+			et.Msg.Method == builtin.MethodConstructor {
+			log.Debugf("COND3 Native actor creation! method:%d, code:%s, height:%d", et.Msg.Method, et.Msg.CodeCid.String(), height)
+			parent.SetCallType("create")
+			parent.Action.To = to
+			parent.Action.Input = []byte{0xFE}
+			parent.Result.Output = nil
+
+			// there should never be any subcalls when creating a native actor
+			return nil
```
Stebalien, line 158: This isn't correct but no native actor _currently_ calls another in the constructor. Can we change this comment? Or just change the code, supporting this use-case shouldn't be complicated unless I'm mistaken.

Stebalien (approved):
Modulo one comment, this looks mergable (as long as we make it clear that it's still experimental and has some missing features).

Stebalien:
@fridrik01 could you file issues for the remaining parts:

1. Support failures while creating contracts.
2. Support sub-calls from "native" contract constructors.
3. Support tracing messages that create native actors (e.g., f1 accounts) when they're reverted at the top level. This case will require adding IDs to traces.

fridrik01:
Updated the CHANGELOG and added this as a highly experimental feature

fridrik01:
Created followup tasks:
- #11217
- #11218
- #11219

BlocksOnAChain:
@fridrik01 can we merge this one now, since we created follow-up tasks from our TO-DOs?

arajasek (approved):
Okay this LGTM

Issue #10793, opened 2023-05-01
FRC0051 Implementation Proposal


Stebalien:
### Added by @arajasek 

Implement https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0051.md

## Technical Breakdown
```[tasklist]
### Development
- [x] Chainstore: drop all blocks from the same miner in the same tipset
- [x] Open design question: Decide how to catch equivocations in the miner (see discussion below)
- [x] Implement equivocation-checking in the miner loop
- [ ] Add metrics to the miner loop to track how long mining takes
```

```[tasklist]
### Testing
- [ ] Unit test: Chainstore rejects all blocks from the same miner when forming or expanding tipsets
- [ ] Functional test: Confirm that miners still produce blocks
- [ ] Performance test: Confirm that the time to produce a block does not increase by more than 2 seconds with this change
```

---
created by @Stebalien 
[FRC0051 ](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0051.md )proposes rejecting equivocations. I.e., multiple blocks at the same height with the same ticket.

This was previously implemented in the sync engine but reverted in #10777 due to DoS and other issues.

Here, I propose that re-implementing this FRC, but inside the chainstore, re-using all the existing tipset tracking logic. Specifically, when we expand a tipset, we currently [drop](https://github.com/filecoin-project/lotus/blob/727a71186bbeb2405c489e3467c129a5a31f5a9a/chain/store/store.go#L1042-L1045) duplicate blocks from the same miner. Instead, we should drop _all_ blocks at a given height from a miner if more than one block is produced (or drop all blocks if we see multiple blocks with the same ticket? I think these are equivalent).

However, to actually catch equivocations, we need a delay after we receive the last block in a tipset but before we create the block. Otherwise, an equivocating miner could send out multiple blocks to different miners at the last possible moment, right before the end of the propagation delay period.

Unfortunately, it looks like any reasonable solution _will_ require modifying the block miner. But it sould be a simple modification. Basically:

1. As described above, drop blocks from miners that equivocate (inside the chainstore).
2. Inside the miner, after the block propagation delay (10s) take the current tipset.
3. Wait 2s (the equivocation delay).
4. Ask the client for the latest tipset, again.
5. Treat the _intersection_ between these two tipsets as the base tipset for mining.

[...]

PR #11104, opened 2023-07-27
feat: chainstore: FRC-0051: Remove all equivocated blocks from tipsets

arajasek:
## Related Issues

#10793 

## Proposed Changes

As discussed in #10793, we now reject ALL blocks with the same miner and same height (irrespective of other details like common parents, common ticket, etc.).  Further, when asked to consider taking a new tipset as head, we first refresh our `heaviest` tipset in order to check whether it's weight has DECREASED as a result of equivocation.

This PR also refactors / simplifies much of the "taking heavier tipset" logic. We now simply:
- `PersistTipsets` as we receive them, which also writes the tipset keys to the blockstore
- `AddToTipsetTracker` as we validate tipsets (blocks)
- When a new "head" is synced to / new block comes in, we simply notify the chainstore of its existence by calling `RefreshHeaviestTipSet` with the height of the new tipset.
- `RefreshHeaviestTipSet` works by:
  - Forming the best tipset at its CURRENT head's height -- if this best tipset is LIGHTER, we immediately take it as our new head, since that implies equivocation
  - Forming the best tipset at the input height (assuming different from height in previous step)
  - taking the heaviest of these tipsets as new head

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/gen/gen.go
+++ b/chain/gen/gen.go
@@ -449,18 +449,19 @@ func (cg *ChainGen) NextTipSetFromMiners(base *types.TipSet, miners []address.Ad
 }
 
 func (cg *ChainGen) NextTipSetFromMinersWithMessagesAndNulls(base *types.TipSet, miners []address.Address, msgs [][]*types.SignedMessage, nulls abi.ChainEpoch) (*store.FullTipSet, error) {
+	ctx := context.TODO()
 	var blks []*types.FullBlock
 
 	for round := base.Height() + nulls + 1; len(blks) == 0; round++ {
 		for mi, m := range miners {
-			bvals, et, ticket, err := cg.nextBlockProof(context.TODO(), base, m, round)
+			bvals, et, ticket, err := cg.nextBlockProof(ctx, base, m, round)
 			if err != nil {
 				return nil, xerrors.Errorf("next block proof: %w", err)
 			}
 
 			if et != nil {
 				// TODO: maybe think about passing in more real parameters to this?
-				wpost, err := cg.eppProvs[m].ComputeProof(context.TODO(), nil, nil, round, network.Version0)
+				wpost, err := cg.eppProvs[m].ComputeProof(ctx, nil, nil, round, network.Version0)
 				if err != nil {
 					return nil, err
 				}
@@ -476,8 +477,18 @@ func (cg *ChainGen) NextTipSetFromMinersWithMessagesAndNulls(base *types.TipSet,
 	}
 
 	fts := store.NewFullTipSet(blks)
-	if err := cg.cs.PutTipSet(context.TODO(), fts.TipSet()); err != nil {
-		return nil, err
+	if err := cg.cs.PersistTipsets(ctx, []*types.TipSet{fts.TipSet()}); err != nil {
+		return nil, xerrors.Errorf("failed to persist tipset: %w", err)
+	}
+
+	for _, blk := range blks {
+		if err := cg.cs.AddToTipSetTracker(ctx, blk.Header); err != nil {
+			return nil, xerrors.Errorf("failed to add to tipset tracker: %w", err)
+		}
+	}
+
+	if err := cg.cs.RefreshHeaviestTipSet(ctx, fts.TipSet().Height()); err != nil {
+		return nil, xerrors.Errorf("failed to put tipset: %w", err)
 	}
 
 	cg.CurTipset = fts
--- a/chain/store/store.go
+++ b/chain/store/store.go
@@ -367,49 +367,32 @@ func (cs *ChainStore) UnmarkBlockAsValidated(ctx context.Context, blkid cid.Cid)
 func (cs *ChainStore) SetGenesis(ctx context.Context, b *types.BlockHeader) error {
 	ts, err := types.NewTipSet([]*types.BlockHeader{b})
 	if err != nil {
-		return err
+		return xerrors.Errorf("failed to construct genesis tipset: %w", err)
 	}
 
-	if err := cs.PutTipSet(ctx, ts); err != nil {
-		return err
+	if err := cs.PersistTipsets(ctx, []*types.TipSet{ts}); err != nil {
+		return xerrors.Errorf("failed to persist genesis tipset: %w", err)
+	}
+
+	if err := cs.AddToTipSetTracker(ctx, b); err != nil {
+		return xerrors.Errorf("failed to add genesis tipset to tracker: %w", err)
+	}
+
+	if err := cs.RefreshHeaviestTipSet(ctx, ts.Height()); err != nil {
+		return xerrors.Errorf("failed to put genesis tipset: %w", err)
 	}
 
 	return cs.metadataDs.Put(ctx, dstore.NewKey("0"), b.Cid().Bytes())
 }
 
-func (cs *ChainStore) PutTipSet(ctx context.Context, ts *types.TipSet) error {
-	if err := cs.PersistTipsets(ctx, []*types.TipSet{ts}); err != nil {
-		return xerrors.Errorf("failed to persist tipset: %w", err)
-	}
-
-	expanded, err := cs.expandTipset(ctx, ts.Blocks()[0])
-	if err != nil {
-		return xerrors.Errorf("errored while expanding tipset: %w", err)
-	}
-
-	if expanded.Key() != ts.Key() {
-		log.Debugf("expanded %s into %s\n", ts.Cids(), expanded.Cids())
-
-		tsBlk, err := expanded.Key().ToStorageBlock()
-		if err != nil {
-			return xerrors.Errorf("failed to get tipset key block: %w", err)
-		}
-
-		if err = cs.chainLocalBlockstore.Put(ctx, tsBlk); err != nil {
-			return xerrors.Errorf("failed to put tipset key block: %w", err)
-		}
-	}
-
-	if err := cs.MaybeTakeHeavierTipSet(ctx, expanded); err != nil {
-		return xerrors.Errorf("MaybeTakeHeavierTipSet failed in PutTipSet: %w", err)
-	}
-	return nil
-}
-
-// MaybeTakeHeavierTipSet evaluates the incoming tipset and locks it in our
-// internal state as our new head, if and only if it is heavier than the current
-// head and does not exceed the maximum fork length.
-func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipSet) error {
+// RefreshHeaviestTipSet receives a newTsHeight at which a new tipset might exist. It then:
+// - "refreshes" the heaviest tipset that can be formed at its current heaviest height
+//   - if equivocation is detected among the miners of the current heaviest tipset, the head is immediately updated to the heaviest tipset that can be formed in a range of 5 epochs
+//
+// - forms the best tipset that can be formed at the _input_ height
+// - compares the three tipset weights: "current" heaviest tipset, "refreshed" tipset, and best tipset at newTsHeight
+// - updates "current" heaviest to the heaviest of those 3 tipsets (if an update is needed), assuming it doesn't violate the maximum fork rule
+func (cs *ChainStore) RefreshHeaviestTipSet(ctx context.Context, newTsHeight abi.ChainEpoch) error {
 	for {
 		cs.heaviestLk.Lock()
 		if len(cs.reorgCh) < reorgChBuf/2 {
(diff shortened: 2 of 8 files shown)


chain/store/store.go
```diff
@@ -1027,7 +1027,7 @@ func (cs *ChainStore) persistBlockHeaders(ctx context.Context, b ...*types.Block
 	return err
 }
 
-func (cs *ChainStore) expandTipset(ctx context.Context, b *types.BlockHeader) (*types.TipSet, error) {
+func (cs *ChainStore) ExpandTipset(ctx context.Context, b *types.BlockHeader) (*types.TipSet, error) {
```
arajasek, line 1030: Exported for testing.

chain/store/store.go
```diff
@@ -1052,7 +1052,15 @@ func (cs *ChainStore) expandTipset(ctx context.Context, b *types.BlockHeader) (*
 		}
 
 		if cid, found := inclMiners[h.Miner]; found {
-			log.Warnf("Have multiple blocks from miner %s at height %d in our tipset cache %s-%s", h.Miner, h.Height, h.Cid(), cid)
+			log.Warnf("Have multiple blocks from miner %s at height %d in our tipset cache %s-%s, purging them all", h.Miner, h.Height, h.Cid(), cid)
+			newAll := make([]*types.BlockHeader, 0, len(all))
```
arajasek, line 1056: We could ALSO drop the blocks from `cs.tipsets` entirely if we wanted to. I'm not sure this is the best thing to do, though, since that tracker is currently split between the "writer" method `AddToTipsetTracker` and the "reader" method `ExpandTipset`.

Stebalien: I'd leave them so we can detect new ones.

chain/store/store.go
```diff
@@ -1052,7 +1052,15 @@ func (cs *ChainStore) expandTipset(ctx context.Context, b *types.BlockHeader) (*
...
+			for _, blk := range all {
+				if blk.Miner != h.Miner {
+					newAll = append(newAll, blk)
```
Stebalien, line 1059: nit: I'd just collect all these blocks into a miner -> block map, then put them into an array at the very end.

Stebalien (approved):

chain/gen/gen.go
```diff
@@ -476,8 +476,18 @@ func (cg *ChainGen) NextTipSetFromMinersWithMessagesAndNulls(base *types.TipSet,
 	}
 
 	fts := store.NewFullTipSet(blks)
+	if err := cg.cs.PersistTipsets(context.TODO(), []*types.TipSet{fts.TipSet()}); err != nil {
+		return nil, xerrors.Errorf("failed to persist tipset: %w", err)
+	}
+
+	for _, blk := range fts.TipSet().Blocks() {
```
arajasek, line 483: Not strictly necessary since we do this in `PutTipSet` too, but doesn't seem harmful to do for tests.

chain/store/store.go
```diff
@@ -378,19 +378,30 @@ func (cs *ChainStore) SetGenesis(ctx context.Context, b *types.BlockHeader) erro
...
+			return xerrors.Errorf("failed to add block to ts tracker: %w", err)
+		}
+	}
+
 	if err := cs.PersistTipsets(ctx, []*types.TipSet{ts}); err != nil {
 		return xerrors.Errorf("failed to persist tipset: %w", err)
 	}
 
-	expanded, err := cs.expandTipset(ctx, ts.Blocks()[0])
+	expanded, err := cs.ExpandTipsetByHeightAndParent(ctx, ts.Height(), ts.Parents().Cids())
```
arajasek, line 391: I'm tempted to just use the new `FormBestTipSetForHeight` method here as well, and drop `ExpandTipsetByHeightAndParent` entirely.

chain/store/store.go
```diff
@@ -1027,44 +1087,111 @@ func (cs *ChainStore) persistBlockHeaders(ctx context.Context, b ...*types.Block
...
+	}
+
+	if len(ret) == 0 {
+		return types.EmptyTSK, nil
+	}
+
+	return types.NewTipSetKey(ret...), nil
+}
+
+func (cs *ChainStore) FormBestTipSetForHeight(ctx context.Context, height abi.ChainEpoch) (*types.TipSet, types.BigInt, error) {
```
Stebalien, line 1137: This should not be exported.

chain/store/store.go
```diff
@@ -426,39 +437,88 @@ func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipS
...
+	heaviestWeight, err := cs.weight(ctx, cs.StateBlockstore(), cs.heaviest)
+	if err != nil {
+		return xerrors.Errorf("failed to calculate currentHeaviest's weight: %w", err)
+	}
+
+	heaviestHeight := abi.ChainEpoch(0)
+	if cs.heaviest != nil {
+		heaviestHeight = cs.heaviest.Height()
+	}
+	newHeaviest, newHeaviestWeight, err := cs.FormBestTipSetForHeight(ctx, heaviestHeight)
```
Stebalien, line 449: nit: FormHeaviestTipSetForHeight?

chain/store/store.go
```diff
@@ -409,7 +420,7 @@ func (cs *ChainStore) PutTipSet(ctx context.Context, ts *types.TipSet) error {
 // MaybeTakeHeavierTipSet evaluates the incoming tipset and locks it in our
 // internal state as our new head, if and only if it is heavier than the current
 // head and does not exceed the maximum fork length.
-func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipSet) error {
+func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, newTsk types.TipSetKey) error {
```
Stebalien, line 423: can we unexport this method?

Stebalien: and may be call it `informNewTipSet`?

chain/store/store.go
```diff
@@ -426,39 +437,88 @@ func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipS
...
+		heaviestHeight = cs.heaviest.Height()
+	}
+	newHeaviest, newHeaviestWeight, err := cs.FormBestTipSetForHeight(ctx, heaviestHeight)
+	if err != nil {
+		return xerrors.Errorf("failed to reform head at same height: %w", err)
+	}
+
+	if newHeaviest == nil || newHeaviestWeight.LessThan(heaviestWeight) {
+		log.Warnf("chainstore heaviest tipset's weight SHRANK from %d (%s) to %d (%s) due to equivocation", heaviestWeight, cs.heaviest, newHeaviestWeight, newHeaviest)
+		// refresh heaviestWeight 10 times moving up and down
```
Stebalien, line 456: Needs lots of comments.

chain/store/store.go
Stebalien: Basically, make it clear that we're telling the system "hey, we have this new tipset", maybe take it, maybe take something entirely different!

chain/store/store.go
```diff
@@ -426,39 +437,88 @@ func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipS
...
+
+		for i := heaviestHeight + 5; i > heaviestHeight-5; i-- {
```
Stebalien, line 458: Remind me why we're not handling this case in `AddToTipSetTracker`? Especially because I expect we'll only enter this code path from an explicit sync, not from random blocks we receive over pubsub?

Stebalien (commented):
Some initial comments but I want to walk through this together tomorrow.

chain/store/store.go
arajasek: It's a bit slower, but I think I'd prefer doing it anyway

chain/store/store.go
```diff
@@ -378,19 +378,30 @@ func (cs *ChainStore) SetGenesis(ctx context.Context, b *types.BlockHeader) erro
 }
 
 func (cs *ChainStore) PutTipSet(ctx context.Context, ts *types.TipSet) error {
+	for _, blk := range ts.Blocks() {
+		if err := cs.AddToTipSetTracker(ctx, blk); err != nil {
+			return xerrors.Errorf("failed to add block to ts tracker: %w", err)
+		}
+	}
+
 	if err := cs.PersistTipsets(ctx, []*types.TipSet{ts}); err != nil {
```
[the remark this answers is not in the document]
arajasek: I think the right sequence of steps is:
- Persist, when you receive a block of interest
- Track, when you actually validate it
- Assume in "PutTipSet" that Persisting and Tracking has already happened

I'm gonna implement that and state the assumption for now.

chain/store/store.go
arajasek: Only exported for tests, which need to be in a separate package (since they depend on the chain generator, which depends on the chainstore...)

Stebalien: Ugh.... ok.

chain/sync.go
```diff
@@ -247,6 +247,7 @@ func (syncer *Syncer) InformNewHead(from peer.ID, fts *store.FullTipSet) bool {
 		return false
 	}
 
+	// TODO: this method name is a lie
```
Stebalien, line 250: 1. Can we document why it's a lie?
2. Can we add a 🍰 emoji?

chain/store/store.go
```diff
@@ -367,49 +367,32 @@ func (cs *ChainStore) UnmarkBlockAsValidated(ctx context.Context, blkid cid.Cid)
...
-func (cs *ChainStore) PutTipSet(ctx context.Context, ts *types.TipSet) error {
 	if err := cs.PersistTipsets(ctx, []*types.TipSet{ts}); err != nil {
-		return xerrors.Errorf("failed to persist tipset: %w", err)
+		return xerrors.Errorf("failed to persist genesis tipset: %w", err)
 	}
 
-	expanded, err := cs.expandTipset(ctx, ts.Blocks()[0])
-	if err != nil {
-		return xerrors.Errorf("errored while expanding tipset: %w", err)
+	if err := cs.AddToTipSetTracker(ctx, b); err != nil {
```
Stebalien, line 377: nit: it would be nice if this just handled writing blocks internally, but I guess that's probably not critical right now?

chain/store/store.go
```diff
@@ -367,49 +367,32 @@ func (cs *ChainStore) UnmarkBlockAsValidated(ctx context.Context, blkid cid.Cid)
...
-
-		tsBlk, err := expanded.Key().ToStorageBlock()
-		if err != nil {
-			return xerrors.Errorf("failed to get tipset key block: %w", err)
-		}
-
-		if err = cs.chainLocalBlockstore.Put(ctx, tsBlk); err != nil {
-			return xerrors.Errorf("failed to put tipset key block: %w", err)
-		}
+	if err := cs.RefreshHeaviestTipSet(ctx, ts.Height()); err != nil {
```
Stebalien, line 381: Hm... for genesis, are we sure any of this is right? Shouldn't we _force_ replace everything? Like:

1. Discard all tipsets.
2. Replace the first tipset.
3. Move on?

Otherwise, e.g., the `metadataDs.Put` line below could do the wrong thing.

(not like this is likely to cause issues, but it's still a bit funky)

chain/store/store.go
Stebalien: Still needs some comments explaining why.

chain/store/store.go
```diff
@@ -426,39 +409,85 @@ func (cs *ChainStore) MaybeTakeHeavierTipSet(ctx context.Context, ts *types.TipS
...
 
-		exceeds, err := cs.exceedsForkLength(ctx, cs.heaviest, ts)
+		errTake := cs.takeHeaviestTipSet(ctx, newHeaviest)
+		if errTake != nil {
+			return xerrors.Errorf("failed to take newHeaviest tipset as head: %w", err)
+		}
+	}
+
+	// if the new height we were notified about isn't what we just refreshed at, see if we have a heavier tipset there
+	if newTsHeight != newHeaviest.Height() {
```
Stebalien, line 456: I'm not sure how this is possible? If we refresh at a different height, shouldn't we have picked the heaviest tipset at that height?

chain/store/store.go
```diff
@@ -1027,44 +1066,72 @@ func (cs *ChainStore) persistBlockHeaders(ctx context.Context, b ...*types.Block
 	return err
 }
 
-func (cs *ChainStore) expandTipset(ctx context.Context, b *types.BlockHeader) (*types.TipSet, error) {
-	// Hold lock for the whole function for now, if it becomes a problem we can
-	// fix pretty easily
+// FormHeaviestTipSetForHeight looks up all valid blocks at a given height, and returns the heaviest tipset that can be made at that height
+// It does not consider ANY blocks from miners that have "equivocated" (produced 2 blocks at the same height)
+func (cs *ChainStore) FormHeaviestTipSetForHeight(ctx context.Context, height abi.ChainEpoch) (*types.TipSet, types.BigInt, error) {
```
Stebalien, line 1071: This looks correct and probably "fast enough". But I'd like to poke at making it faster in the future by pre-grouping by height, etc.

Stebalien (approved):
nits, but this looks correct

chain/store/store.go
arajasek: Yeah, not critical I think. Also not super easy to do, because we want to be batching those writes, and this method tracks things one block at a time.

chain/store/store.go
arajasek: I think you're right, but I was going for matching existing behaviour here, which this should do?

Stebalien: Yeah, you're right.

Stebalien: Let's just keep a list of TODOs, merge this, then we can make followup PRs (if/when we get time).

chain/store/store.go
arajasek: No, not necessarily. We'll only refresh at different heights if there was equivocation. Basically, assume current height `heaviestHeight` is 10, and we were informed about `newTsHeight` of 11.

- We refresh by finding the heaviest tipset at 10
- If the "refreshed" tipset at 10 is not lighter than our current heaviest tipset at 10, we'll just move on
- At that point, we hit this line, and decide we should form the heaviest tipset at height 11, and compare the two.

(If the "refreshed" tipset WAS lighter, though, then yes we would likely form a heavier tipset at 11, and immediately take that as our head, in which case we'd short-circuit here)

Stebalien: I think I understand?

Stebalien (approved):

Stebalien:
Ok, actually. I'm sorry, but I'm going to need to do this: is there any way to write a simple equivocation test?

arajasek:
> Ok, actually. I'm sorry, but I'm going to need to do this: is there any way to write a simple equivocation test?

That's at least kiiinda what [this test](https://github.com/filecoin-project/lotus/blob/261a398f36d43f468d95ce5901f68ba697465974/chain/store/store_test.go#L243) is intended to be. Definitely others that could be written, though...

chain/store/store.go
arajasek: Let's talk through tomorrow, but I'm pretty sure it's correct.

Stebalien:
(but it would be nice to get one that tests switching away from a tipset)

arajasek:
@Stebalien Added more in[ latest commit](https://github.com/filecoin-project/lotus/pull/11104/commits/2cbdedc209832a1462f3ec9cb1a377a391d6122e), please suggest more if you think of them -- we need to get this _right_.

Stebalien:
Well, the tests I'd _like_ are:

1. We switch tipsets at the current height (needs multiple tipsets with different parents).
2. We switch to a different height because we lose all the blocks in the tipset.
3. And maybe some kind of validation that we don't crash with too many null blocks in case 2?

arajasek:
@Stebalien Thanks, all good ideas. I've added all 3 cases in latest commit, please take a look.

Stebalien (approved):
LGTM!

PR #11108, opened 2023-07-27
Harmonytask

snadrus:
## Related Issues
HarmonyTask implementation

## Proposed Changes
HarmonyTasks package and Resources package to manage tasks. 

## Additional Info
Tasks are listed in Yugabyte and competed-for through a greedy (but not too greedy) algorithm that enables tasks to hop from creator-machine to consumer machine or even from task-completion-machine to following-task-defined-elsewhere-machine. 
See doc.go 
Overview in https://docs.google.com/document/d/1lPe-6HdVe_oU0mD29f9Zi4233_xhSbyiXxM_6zXG4hY/edit#heading=h.k4h9btwn9p1y


magik6k:
I see commits from https://github.com/filecoin-project/lotus/pull/11069, should probably rebase on feat/sturdypost to clean that up a bit

lib/harmony/harmonytask/harmonytask.go
```diff
@@ -0,0 +1,462 @@
...
+	//		})
+	//	  }
+	// }
+	Adder(AddTaskFunc)
+}
+
+type AddTaskFunc func(extraInfo func(TaskID, *harmonydb.Tx) bool)
+
+type TaskEngine struct {
+	ctx            context.Context
```
magik6k, line 94: Can we avoid the context in the struct?

magik6k: Oh, I see, this is used for graceful exit.

Typically we use two `chan struct{}` for that, much more robust, and plays nicely with uber/fx.Lifecycle, which makes sure that everything actually stops cleanly, and in the correct order.

Basically you have 2 channels:
* `stopping`, which is closed when we want to stop
* `stopped`, which is closed in the goroutine that we want to exit
* Then you have a `Stop(shutdownCtx)` method, which closes `stopping`, and waits for `stopped` (or `shutdownCtx`, which is a context provided by fx.Lifecycle, and has a 15s timeout iirc)

lib/harmony/harmonytask/harmonytask.go
```diff
@@ -0,0 +1,462 @@
...
+
+	// TODO block bumps & follows by unreg from DBs.
+	_, err := e.db.Exec(context.Background(), `DELETE FROM harmony_task_impl WHERE owner_id=$1`, e.ownerID)
+	if err != nil {
+		logger.Warn("Could not clean-up impl table: %w", err)
+	}
+	_, err = e.db.Exec(context.Background(), `DELETE FROM harmony_task_follow WHERE owner_id=$1`, e.ownerID)
+	if err != nil {
+		logger.Warn("Could not clean-up impl table: %w", err)
+	}
```
magik6k, line 217: Any side effects if one passes and the second fails? Would we want this to be a transaction or is it better if at least one passes?

lib/harmony/harmonydb/sql/20230719.sql
```diff
@@ -0,0 +1,51 @@
...
+    name VARCHAR(8) NOT NULL,
+    posted TIMESTAMP NOT NULL, 
+    work_start TIMESTAMP NOT NULL, 
+    work_end TIMESTAMP NOT NULL, 
+    result BOOLEAN NOT NULL, 
+    err varchar
+);
+COMMENT ON COLUMN harmony_task_history.result IS 'Use to detemine if this was a successful run.';
+
+CREATE TABLE harmony_task_follow (
```
magik6k, line 40: Nit: Would be nice to have a comment briefly explaining the follow / impl mechanics (or at least pointing to somewhere with the explanation)

lib/harmony/harmonytask/doc.go
```diff
@@ -0,0 +1,47 @@
+/*
+	Package harmomnytask implements a pure (no task logic), distributed
+	task manager. This clean interface allows a task implementer to completely
+
+avoid being concerned with task scheduling and management.
+It's based on the idea of tasks as small units of work broken from other
+work by hardware, parallelizabilty, reliability, or any other reason.
+Workers will be Greedy: vaccuuming up their favorite jobs from a list.
```
magik6k, line 8: > Workers will be Greedy

This is not always the best strategy, tho it should result in a kinda-round-robin, so it's probably fine for now.

magik6k: Probably not worth doing in the first iteration, but we could build some more elaborate task claiming process where workers can say how much they want a task, and upsert their proposals for some short period.

lib/harmony/harmonytask/harmonytask.go
```diff
@@ -0,0 +1,462 @@
...
+		return
+	busy:
+		ch := make(chan bool)
+		var once sync.Once
+		e.workAdderMutex.UnlockNotify = func() {
+			once.Do(func() {
+				close(ch)
+			})
+		}
+		e.workAdderMutex.Unlock()
```
magik6k, line 235: Pretty sure this Unlock will immediately call UnlockNotify

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,274 @@
...
+	logging "github.com/ipfs/go-log/v2"
+)
+
+var logger = logging.Logger("harmonytask")
+
+type taskTypeHandler struct {
+	TaskInterface
+	TaskTypeDetails
+	TaskEngine *TaskEngine
+	Count      int /// locked by TaskEngine's mutex
```
magik6k, line 22: Any reason to not use atomic.Int for this?

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,274 @@
...
+	"net/http"
+	"strconv"
+	"sync/atomic"
+	"time"
+
+	"github.com/filecoin-project/lotus/lib/harmony/harmonydb"
+	logging "github.com/ipfs/go-log/v2"
+)
+
+var logger = logging.Logger("harmonytask")
```
magik6k, line 16: Nit: We usually call the logger instance `log`

```suggestion
var log = logging.Logger("harmonytask")
```

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,274 @@
...
+	if len(ids) == 0 {
+		return true // stop looking for takers
+	}
+
+	h.TaskEngine.workAdderMutex.Lock()
+	defer h.TaskEngine.workAdderMutex.Unlock()
+
+	// 1. Can we do any more of this task type?
+	if h.Max > -1 && h.Count == h.Max {
+		logger.Info("Did not accept " + h.Name + " task: at max already.")
```
magik6k, line 65: Would be great to have these use the structured-logging calls, e.g.
```suggestion
		logger.Infow("did not accept task", "type", h.Name, "reason", "at max already.")
```

Most deployments will dump their logs at things like logstash, and structured logs are vastly easier to query and make pretty Grafana dashboards from

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,274 @@
...
+		return errors.New("Did not accept " + h.Name + " task: out of cpu")
+	}
+	if h.Cost.Ram > r.Ram {
+		return errors.New("Did not accept " + h.Name + " task: out of RAM")
+	}
+	if r.Gpu-h.Cost.Gpu < 0 {
+		return errors.New("Did not accept " + h.Name + " task: out of available GPU")
+	}
+	if h.Cost.Scratch > r.Scratch {
+		return errors.New("Did not accept " + h.Name + " task: out of scratch space")
```
magik6k, line 213: Scratch space is a bit more tricky, because:
* One worker may have multiple separate scratch space disks operating with independent partitions / mounts.
* Multiple workers may share one scratch space path.

The sector-storage subsystem has a mechanism to reserve storage when paths are shared between machines, that works through the sector index.

Here we could have a mechanism, where:
* We would check if we can claim some work
* If we can, we'd set up scratch space reservation, which for harmonytask would probably be a per-task-type callback like `ReserveResources(some-task-details-because-storage-reservations-need-sector-id) (free func(), err error)`
* If that doesn't succeed, we don't claim the task, and maybe set some backoff on trying to claim a task again
* After the work is done, we'd free the reservation.

lib/harmony/resources/memsys.go
```diff
@@ -0,0 +1,22 @@
...
+	s, err := syscall.Sysctl(name)
+	if err != nil {
+		return 0, err
+	}
+	// hack because the string conversion above drops a \0
+	b := []byte(s)
+	if len(b) < 8 {
+		b = append(b, 0)
+	}
+	return *(*uint64)(unsafe.Pointer(&b[0])), nil
```
magik6k, line 21: Could use binary.LittleEndian.Uint64(b), should compile to roughly the same thing, except without any exciting unsafe

lib/harmony/resources/resources.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+	}
+
+	{ // GPU boolean
+		gpus, err := ffi.GetGPUDevices()
+		if err != nil {
+			logger.Errorf("getting gpu devices failed: %+v", err)
+		}
+		all := strings.ToLower(strings.Join(gpus, ","))
+		if len(gpus) > 1 || strings.Contains(all, "ati") || strings.Contains(all, "nvidia") {
```
magik6k, line 134: Why not simply set `res.Gpu = 1` when we have at least one gpu?

lib/harmony/resources/resources.go
```diff
@@ -0,0 +1,181 @@
...
+		d, err := p.GetDevices(cl.DeviceTypeAll)
+		if err != nil {
+			logger.Error(err)
+			return 0
+		}
+		return lo.SumBy(d, func(d *cl.Device) int64 { return d.GlobalMemSize() })
+	}))
+}
+
+func DiskFree(path string) (uint64, error) {
```
magik6k, line 159: I would abstract disk space away, it's a slightly more complicated resource than what we would want to handle here imo

magik6k (commented):
First pass, overall looks good

lib/harmony/harmonytask/harmonytask.go
snadrus: It best to clean up these tables as much as possible since they become serial HTTP requests later on. It's a best-effort design, so excess entries don't break things too much: we just fall back to the slow way if it takes over 1 minute.

lib/harmony/resources/resources.go
snadrus: I'm trying to get an exact count. 
But more generally there's the problem of resource consumption. I'm wondering if we should just model GpuRAM and ignore count entirely. Then we can consume it like any regular (CPU, RAM) resource.

lib/harmony/harmonytask/harmonytask.go
snadrus: The advantage of using context is libraries like PGX respond to this and will actually cancel DB operations in progress in order to return. This will cause API call receivers during this period to error at the DB calls so that no new work could get picked-up.
I like hooking it up to fx.Lifecycle so only 1 call is needed to close down everything
	lc.Append(fx.Hook{
		OnStop: func(ctx context.Context) error {
		       e.grace()
                        e.gracefulShutdown()
		},
	})
, but I see a lot of value in keeping the context too (for the DB benefit). WDYT?

lib/harmony/harmonydb/sql/20230719.sql
snadrus: added to doc.go

lib/harmony/harmonytask/doc.go
snadrus: We do round robin after every accept. 
I propose that (for non-urgent work) we refuse acceping work in CanAccept() for a few polling periods if it's relatively undesirable (such as not having the storage locally yet it's local to another).

lib/harmony/harmonytask/harmonytask.go
magik6k: > will actually cancel DB operations in progress in order to return

I don't think we have any long-running DB operations that we'd want to cancel in-flight, I'd expect harmonytask to stop cleanly first, and only after that the PG connection to close

Issue #11061, opened 2023-07-10
Failed Disputer Attempts Create Massive Daemon Log Entries


TippyFlitsUK:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### Lotus Version

```text
Daemon:  1.23.4-dev+mainnet+git.4bc032430+api1.5.0
Local: lotus version 1.23.4-dev+mainnet+git.4bc032430

Also confirmed in 1.23.2 & 1.23.1
```

### Repro Steps

1. Run disputer
2. Browse lotus daemon logs

### Describe the Bug

I am running the disputer using the following command:
```
lotus chain disputer --max-fee 0.5 --from f3r7wzp27frp4huhcrk2k2hfumwqmcclwmhzz6ieg2imjs5au6jgyf4p5kohqox7nwfsjpwo6yvilgxepqeciq start |& tee -a lotus-disputer.log
```
The disputer appears to be starting and challenging proofs without issue:
```
2023-07-10T15:39:02.173+0100	INFO	disputer	cli/disputer.go:268	checked proofs	{"count": 25, "duration": 2.346893413}
2023-07-10T15:39:29.757+0100	INFO	disputer	cli/disputer.go:226	last checked epoch	{"epoch": "3023119"}
2023-07-10T15:39:32.518+0100	INFO	disputer	cli/disputer.go:268	checked proofs	{"count": 32, "duration": 2.760902041}
2023-07-10T15:40:00.223+0100	INFO	disputer	cli/disputer.go:226	last checked epoch	{"epoch": "3023120"}
[...]

PR #11120, opened 2023-07-31
feat: FVM: do not error on unsuccessful implicit messages

arajasek:
## Related Issues

#11061

The FVM currently errors on unsuccessful `Implicit` messages -- that is, a failure exit code is treated as an error. This essentially assumes that all callers of `ApplyImplicitMessage` wants to treat a non-zero code as an error. While this is generally true, some callers (eg. when simulating messages) might not want to handle non-zero exit codes as errors.

## Proposed Changes

- Change the FVM to NOT treat non-zero exit codes as errors
- Change the core state computation pipeline to explicitly check exit codes for Cron and Reward actors, and fail if unsuccessful

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/consensus/compute_state.go
+++ b/chain/consensus/compute_state.go
@@ -135,6 +135,10 @@ func (t *TipSetExecutor) ApplyBlocks(ctx context.Context,
 			return xerrors.Errorf("running cron: %w", err)
 		}
 
+		if !ret.ExitCode.IsSuccess() {
+			return xerrors.Errorf("cron failed with exit code %d: %w", ret.ExitCode, ret.ActorErr)
+		}
+
 		cronGas += ret.GasUsed
 
 		if em != nil {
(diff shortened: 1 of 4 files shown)


chain/consensus/compute_state.go
```diff
@@ -135,6 +135,10 @@ func (t *TipSetExecutor) ApplyBlocks(ctx context.Context,
 			return xerrors.Errorf("running cron: %w", err)
 		}
 
+		if !ret.ExitCode.IsSuccess() {
+			return xerrors.Errorf("cron failed with exit code %d: %s", ret.ExitCode, ret.ActorErr.Error())
```
ZenGround0, line 139: nit for here and reward error handling: to maintain existing behavior you could end err string with %w and wrap the error instead of printing as string

ZenGround0 (approved):

PR #11135, opened 2023-08-04
WIP: sector index yugabyte impl

shrenujbansal:
## Checklist

- [ ] CI is green


itests/harmonydb_test.go
```diff
@@ -42,7 +42,8 @@ func TestCrud(t *testing.T) {
 			Animal      string `db:"content"`
 			Unpopulated int
 		}
-		err = cdb.Select(ctx, &ints, "SELECT content, some_int FROM itest_scratch")
+		contentfilter := []string{"cows", "cats"}
```
shrenujbansal, line 45: Remove this change

storage/paths/index.go
```diff
@@ -73,6 +157,36 @@ type Index struct {
 	stores  map[storiface.ID]*storageEntry
 }
 
+type DBIndex struct {
```
shrenujbansal, line 160: Move DBIndex and related funcs into its own file(?)

lib/harmony/harmonydb/sql/20230712.sql
```diff
@@ -0,0 +1,37 @@
+create table SectorLocation
+(
+    "miner_id"        int8,
+    "sector_num"   int8,
```
magik6k, line 4: Why int8, and not just int (/maybe bigint which is more standard)?

lib/harmony/harmonydb/sql/20230712.sql
```diff
@@ -0,0 +1,37 @@
...
+    "storage_id"  varchar not null
+        constraint "StorageLocation_pkey"
+            primary key,
+    "urls"       varchar,
+    "weight"     int8,
+    "max_storage" int8,
+    "can_seal"    bool,
+    "can_store"   bool,
+    "groups"     varchar,
+    "allow_to"    varchar,
```
magik6k, line 24: Wouldn't making those into [arrays](https://www.postgresql.org/docs/current/arrays.html) make some queries easier?

storage/paths/index.go
```diff
@@ -116,6 +283,156 @@ func (i *Index) StorageList(ctx context.Context) (map[storiface.ID][]storiface.D
...
+		})
+	}
+
+	for _, u := range si.URLs {
+		if _, err := url.Parse(u); err != nil {
+			return xerrors.Errorf("failed to parse url %s: %w", si.URLs, err)
+		}
+	}
+
+	// TODO: make below part of a single transaction
```
magik6k, line 373: Could probably be done with a single `insert into .. on conflict do update set ...` query. See e.g. https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-upsert/

magik6k: This is what gpt4 came up with when I asked it to rewrite this with a single query:

```go
var urls sql.NullString
var storageId sql.NullString

// Prepare the URL list for the new storage location
currUrls := strings.Join(union(strings.Split(urls.String, ","), si.URLs), ",")

// Prepare the INSERT ... ON CONFLICT ... DO UPDATE query
_, err = dbi.harmonyDB.Exec(ctx,
	`INSERT INTO StorageLocation (
		storage_id, 
		urls, 
		weight, 
		max_storage, 
		can_seal, 
		can_store, 
		groups, 
		allow_to, 
		allow_types, 
		deny_types,
		capacity,
		available,
		fs_available,
		reserved,
		used,
		created_at
	) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16)
	ON CONFLICT (storage_id) 
	DO UPDATE SET 
		urls = EXCLUDED.urls,
		weight = EXCLUDED.weight,
		max_storage = EXCLUDED.max_storage,
		can_seal = EXCLUDED.can_seal,
		can_store = EXCLUDED.can_store,
		groups = EXCLUDED.groups,
		allow_to = EXCLUDED.allow_to,
		allow_types = EXCLUDED.allow_types,
		deny_types = EXCLUDED.deny_types`,
	si.ID,
	currUrls,
	si.Weight,
	si.MaxStorage,
	si.CanSeal,
	si.CanStore,
	strings.Join(si.Groups, ","),
	strings.Join(si.AllowTo, ","),
	strings.Join(si.AllowTypes, ","),
	strings.Join(si.DenyTypes, ","),
	st.Capacity,
	st.Available,
	st.FSAvailable,
	st.Reserved,
	st.Used,
	time.Now())

if err != nil {
	log.Errorf("StorageAttach insert/update fails: ", err)
	return xerrors.Errorf("StorageAttach insert/update fails: ", err)
}
return nil
```

storage/paths/index.go
```diff
@@ -61,6 +64,87 @@ type storageEntry struct {
 	heartbeatErr  error
 }
 
+type IndexProxy struct {
```
magik6k, line 67: Would be nice to move the proxy / DB impl to separate files before this is merged

magik6k: For testing could also ask chatgpt to write a version of the proxy which would call both backends and compare results.

lib/harmony/harmonydb/sql/20230712.sql
shrenujbansal: bigint is the same as int8. I created the columns with bigint and the ddl generated showed int8

lib/harmony/harmonydb/sql/20230712.sql
shrenujbansal: I originally defined these as arrays but that seemed to cause more problems than it solved. Sticking to more fundamental types seems to be the better course of action

storage/paths/index.go
shrenujbansal: > Would be nice to move the proxy / DB impl to separate files before this is merged
Yep, thats part of my TODOs

> For testing could also ask chatgpt to write a version of the proxy which would call both backends and compare results
The problem here is how do we get a bool to decide whats the primary path without actually modifying the API

storage/paths/index.go
shrenujbansal: Ah this is neat Postgres feature! Most other langs make you use temp tables which is a pain

documentation/en/default-lotus-miner-config.toml
```diff
@@ -145,6 +145,10 @@
   # env var: LOTUS_SUBSYSTEMS_ENABLEMARKETS
   #EnableMarkets = false
 
+  # type: bool
+  # env var: LOTUS_SUBSYSTEMS_ENABLESECTORINDEXDB
+  #EnableSectorIndexDB = false
```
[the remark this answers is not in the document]
shrenujbansal: I don't think I agree with this. The HarmonyDB section is for configuring database specific configs which are part of the DB abstraction layer
The Sector Index is a component that simply uses HarmonyDB and should not be tied to the same config

storage/paths/index.go
shrenujbansal: Unfortunately, you still need the first select since you need to create the updated url list so still need a transaction but this does consolidate the upsert operation

storage/paths/index.go
```diff
@@ -116,6 +283,156 @@ func (i *Index) StorageList(ctx context.Context) (map[storiface.ID][]storiface.D
 	return out, nil
 }
 
+func union(a, b []string) []string {
```
snadrus, line 286: lo.Union() appears to be the same

storage/paths/index.go
```diff
@@ -215,6 +532,63 @@ func (i *Index) StorageAttach(ctx context.Context, si storiface.StorageInfo, st
...
+	return nil
+}
+
+func (dbi *DBIndex) StorageDetach(ctx context.Context, id storiface.ID, url string) error {
+
+	// If url not in path urls, error out
+	// if this is only path url for this storage path, drop storage path and sector decls which have this as a storage path
+
+	var qUrls sql.NullString
+	err := dbi.harmonyDB.QueryRow(ctx, "select urls from storagelocation where storage_id=$1", string(id)).Scan(&qUrls)
```
snadrus, line 545: optional:  (since empty string is handled correctly)
qUrls := ""
err := dbi.harmonyDB.QueryRow(ctx, "select COALESCE(urls,'') from storagelocation where storage_id=$1", string(id)).Scan(&qUrls)

storage/paths/index.go
```diff
@@ -415,6 +907,193 @@ func (i *Index) StorageDropSector(ctx context.Context, storageID storiface.ID, s
...
+			Weight      uint64
+			Can_seal    bool
+			Can_store   bool
+			Groups      string
+			Allow_types string
+			Deny_types  string
+		}
+		err = dbi.harmonyDB.Select(ctx, &rows,
+			`select storage_id, 
+					  	urls,
```
snadrus, line 1025: please use spaces instead of tabs for this as it formats weird for different editors. Also seen above.

storage/paths/index.go
```diff
@@ -575,6 +1290,81 @@ func (i *Index) StorageInfo(ctx context.Context, id storiface.ID) (storiface.Sto
...
+		spaceReq, err = allocate.StoreSpaceUse(ssize)
+	default:
+		return nil, xerrors.Errorf("Unexpected path type")
+	}
+	if err != nil {
+		return nil, xerrors.Errorf("estimating required space: %w", err)
+	}
+
+	var checkString string
+	if pathType == storiface.PathSealing {
```
snadrus, line 1309: This kinda stuff is supposed to be blocked by the db system. 
Consider instead:
     ptBool := pathType = storiface.PathSealing
     in SQL:
     WHERE ($1 AND can_seal=TRUE OR NOT $1 AND can_store=TRUE)
     (having $1 be ptBool)

storage/paths/index.go
shrenujbansal: Is this a different lib?

cmd/lotus-miner/init.go
```diff
@@ -463,7 +465,13 @@ func storageMinerInit(ctx context.Context, cctx *cli.Context, api v1api.FullNode
 			wsts := statestore.New(namespace.Wrap(mds, modules.WorkerCallsPrefix))
 			smsts := statestore.New(namespace.Wrap(mds, modules.ManagerWorkPrefix))
 
-			si := paths.NewIndex(nil)
+			harmonyDB, err := harmonydb.New([]string{"127.0.0.1"}, "yugabyte", "yugabyte", "yugabyte", "5433", "",
```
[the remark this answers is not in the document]
shrenujbansal: This will need to be input via cmdline or ENV
The config doesn't exist at this stage

lib/harmony/harmonydb/sql/20230712.sql
```diff
@@ -0,0 +1,37 @@
+create table SectorLocation
+(
+    "miner_id"        int8,
+    "sector_num"   int8,
+    "sector_filetype" int,
+    "storage_id"      varchar,
+    "is_primary"      bool,
+    constraint SectorLocation_pk
+        primary key ("miner_id", "sector_num", "sector_filetype", "storage_id")
```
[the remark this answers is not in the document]
shrenujbansal: I don't think that's needed. Its very common to have composite keys
Creating another "id" will also mean we'll have add management for that id and currently no one will be querying for that field either

storage/paths/index.go
shrenujbansal: Do you have a preferred whitespace formatter we can run on the code?
Ideally, this would be done as part of make gen

storage/paths/index.go
shrenujbansal: > This kinda stuff is supposed to be blocked by the db system.

Why is that?

The suggested query wouldn't have the same result as paths can have both CanSeal and CanStore set to true

storage/paths/index.go
```diff
@@ -87,6 +201,59 @@ func NewIndex(al *alerting.Alerting) *Index {
...
+	var sectorEntries []struct {
+		Storage_id      string
+		Miner_id        sql.NullInt64
+		Sector_num      sql.NullInt64
+		Sector_filetype sql.NullInt32
+		Is_primary      sql.NullBool
+	}
+
+	err := dbi.harmonyDB.Select(ctx, &sectorEntries,
+		"select stor.storage_id, miner_id, sector_num, sector_filetype, is_primary from storagelocation stor left join sectorlocation sec on stor.storage_id=sec.storage_id")
```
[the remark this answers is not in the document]
shrenujbansal: done

storage/paths/index.go
shrenujbansal: done

lib/harmony/harmonydb/sql/20230712.sql
```diff
@@ -1,12 +1,14 @@
...
-    "sector_num"   int8,
+    "miner_id"        bigint,
+    "sector_num"   bigint,
     "sector_filetype" int,
     "storage_id"      varchar,
     "is_primary"      bool,
     constraint SectorLocation_pk
         primary key ("miner_id", "sector_num", "sector_filetype", "storage_id")
+
+    -- TODO: Maybe add index on above PK fields
```
snadrus, line 11: a PK is enforced by an internal unique index

storage/paths/db_index.go
```diff
@@ -0,0 +1,882 @@
...
+				SectorID:       sectorID,
+				SectorFileType: fileType,
+			})
+		}
+	}
+
+	return out, nil
+}
+
+func union(a, b []string) []string {
```
snadrus, line 92: lo.Union() ?

storage/paths/db_index.go
```diff
@@ -0,0 +1,882 @@
...
+								groups, 
+								allow_to, 
+								allow_types, 
+								deny_types 
+						 FROM storagelocation 
+						 WHERE %s and available >= $1
+						 and NOW()-last_heartbeat < $2 
+						 and heartbeat_err is null
+						order by (available::numeric * weight) desc`, checkString)
+	err = dbi.harmonyDB.Select(ctx, &rows,
```
snadrus, line 703: no need to build a query. Just pass in (pathType==storiface.PathSealing) as $4
and ($4 AND can_seal OR NOT $4 AND can_store)

storage/paths/db_index.go
```diff
@@ -0,0 +1,882 @@
...
+		WriteTs        sql.NullTime
+	}
+	_, err := dbi.harmonyDB.BeginTransaction(ctx, func(tx *harmonydb.Tx) (commit bool, err error) {
+
+		fts := (read | write).AllSet()
+		err = dbi.harmonyDB.Select(ctx, &rows,
+			`				SELECT sector_filetype, read_ts, read_refs, write_ts 
+								 FROM sectorlocation 
+								 WHERE miner_id=$1 
+								   AND sector_num=$2
```
snadrus, line 750: weird huge gap

snadrus (approved):

storage/paths/db_index.go
shrenujbansal: Whats the downside of building a query?

> Just pass in (pathType==storiface.PathSealing) as $4
and ($4 AND can_seal OR NOT $4 AND can_store)

This assumes that either CanSeal or CanStore will be set when actually both of them could be set

snadrus: Query building leads to SQL injections if done incorrectly (this one looks fine).
Honestly I thought I had put things in place to block query building (harmonydb.rawString). 

I attempted to replicate the above logic with: NOT $4
       The point is to say that the logic could live in the SQL.

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+	ftValid := false
+	for _, fileType := range storiface.PathTypes {
+		if fileType&ft == 0 {
+			ftValid = true
+			break
+		}
+	}
+	if !ftValid {
+		return xerrors.Errorf("Invalid filetype")
+	}
```
magik6k, line 365: * Would be nice to dedupe this with the same bit of code below
* Could call `ft.AllSet() -> []SectorFileType` and iterate over that
  * Or do asserts on that, at least reduces the direct bit fiddling here

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+			CanStore:   row.CanStore,
+			Primary:    row.IsPrimary,
+			AllowTypes: splitString(row.AllowTypes),
+			DenyTypes:  splitString(row.DenyTypes),
+		})
+
+		storageWithSector[row.StorageId] = true
+
+		allowTo := splitString(row.AllowTo)
+		if allowList != nil && len(allowTo) > 0 {
```
magik6k, line 508: Shouldn't this be
```suggestion
		if allowTo != nil && len(allowTo) > 0 {
```

The current way won't ever set the allowList afaict

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+			`SELECT storage_id, 
+					  	urls, 
+					  	weight,
+					  	can_seal,
+					  	can_store,
+					  	groups, 
+					  	allow_types, 
+					  	deny_types
+				FROM storagelocation 
+				WHERE can_seal=true
```
magik6k, line 553: We shouldn't be checking for can_seal here, StorageFindSector may also be used (not entirely sure if it actually is) by sector-finalization code, which wants to move the sector into a `can_store` path (which in almost all cases won't be `can_seal`)

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+func (dbi *DBIndex) StorageBestAlloc(ctx context.Context, allocate storiface.SectorFileType, ssize abi.SectorSize, pathType storiface.PathType) ([]storiface.StorageInfo, error) {
+	var err error
+	var spaceReq uint64
+	switch pathType {
+	case storiface.PathSealing:
+		spaceReq, err = allocate.SealSpaceUse(ssize)
+	case storiface.PathStorage:
+		spaceReq, err = allocate.StoreSpaceUse(ssize)
+	default:
+		return nil, xerrors.Errorf("Unexpected path type")
```
magik6k, line 662:
```suggestion
		return nil, xerrors.Errorf("unexpected path type")
```

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+
+	ftValid := false
+	for _, fileType := range storiface.PathTypes {
+		if fileType&ft == 0 {
+			ftValid = true
+			break
+		}
+	}
+	if !ftValid {
+		return xerrors.Errorf("Invalid filetype")
```
magik6k, line 364:
```suggestion
		return xerrors.Errorf("invalid filetype")
```

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+			lockMap[row.SectorFileType] = locks{
+				readTs:   row.ReadTs,
+				readRefs: row.ReadRefs,
+				writeTs:  row.WriteTs,
+			}
+		}
+
+		// Check if we can acquire write locks
+		// Conditions: No write lock or write lock is stale, No read lock or read lock is stale
+		for wft := range write.AllSet() {
```
magik6k, line 776: Pretty sure this should be
```suggestion
		for _, wft := range write.AllSet() {
```

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+				lockMap[storiface.SectorFileType(wft)].writeTs.Time.After(time.Now().Add(LockTimeOut))) ||
+				(lockMap[storiface.SectorFileType(wft)].readTs.Valid &&
+					lockMap[storiface.SectorFileType(wft)].writeTs.Time.After(time.Now().Add(LockTimeOut))) {
+				return false, xerrors.Errorf("Cannot acquire writelock for sector %v filetype %d already locked", sector, wft)
+			}
+		}
+
+		// Check if we can acquire read locks
+		// Conditions: No write lock or write lock is stale
+		for rft := range read.AllSet() {
```
magik6k, line 787:
```suggestion
		for _, rft := range read.AllSet() {
```

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+				readRefs: row.ReadRefs,
+				writeTs:  row.WriteTs,
+			}
+		}
+
+		// Check if we can acquire write locks
+		// Conditions: No write lock or write lock is stale, No read lock or read lock is stale
+		for wft := range write.AllSet() {
+			if (lockMap[storiface.SectorFileType(wft)].writeTs.Valid &&
+				lockMap[storiface.SectorFileType(wft)].writeTs.Time.After(time.Now().Add(LockTimeOut))) ||
```
magik6k, line 778: Could write a tiny helper like `checkLockTimeout(sql.NullTime) -> bool`, would make this much more readable

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+				(lockMap[storiface.SectorFileType(wft)].readTs.Valid &&
+					lockMap[storiface.SectorFileType(wft)].writeTs.Time.After(time.Now().Add(LockTimeOut))) {
+				return false, xerrors.Errorf("Cannot acquire writelock for sector %v filetype %d already locked", sector, wft)
```
magik6k, line 781: Nit: All errors should be lower-cased
```suggestion
				return false, xerrors.Errorf("cannot acquire writelock for sector %v filetype %d already locked", sector, wft)
```

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+
+	return true, nil
+}
+
+func (dbi *DBIndex) StorageLock(ctx context.Context, sector abi.SectorID, read storiface.SectorFileType, write storiface.SectorFileType) error {
+
+	_, err := dbi.lock(ctx, sector, read, write)
+	if err != nil {
+		return err
+	}
```
magik6k, line 832: This probably should retry / backoff / block when we don't get the lock immediately

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+	var result storiface.SectorLocks
+	for sector, locks := range sectorLocks {
+		var lock storiface.SectorLock
+		lock.Sector = sector
+		lock.Read[locks.sectorFileType] = locks.readRefs
+		if locks.writeLk {
+			lock.Write[locks.sectorFileType] = 1
+		} else {
+			lock.Write[locks.sectorFileType] = 0
+		}
```
magik6k, line 885: * We don't assign to the result
* Should probably filter out timed-out locks here

magik6k (commented):
First pass, this looks like a real thing now, just a few comments

storage/paths/db_index.go
```diff
@@ -0,0 +1,891 @@
...
+		// Single transaction to drop storage path and sector decls which have this as a storage path
+		_, err := dbi.harmonyDB.BeginTransaction(ctx, func(tx *harmonydb.Tx) (commit bool, err error) {
+			// Drop storage path completely
+			_, err = dbi.harmonyDB.Exec(ctx, "DELETE FROM storagelocation WHERE storage_id=$1", id)
+			if err != nil {
+				return false, err
+			}
+
+			// Drop all sectors entries which use this storage path
+			_, err = dbi.harmonyDB.Exec(ctx, "DELETE FROM sectorlocation WHERE storage_id=$1", id)
```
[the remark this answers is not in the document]
shrenujbansal: These are 2 different tables. StorageLocation and SectorLocation

storage/paths/db_index.go
shrenujbansal: This is exactly the logic in the index today: https://github.com/filecoin-project/lotus/blob/77594442471f6512d5657c5be9ca4511eea4f60b/storage/paths/index.go#L426

Keep in mind I've changed the name from allowTo (from the old index) to allowList( in the new index) to disambiguate allowTo from the DB

storage/paths/db_index.go
shrenujbansal: The current in memory index has this constraint here: https://github.com/filecoin-project/lotus/blob/77594442471f6512d5657c5be9ca4511eea4f60b/storage/paths/index.go#L491

Let me know if you think thats a bug in the current implementation as well

node/config/types.go
```diff
@@ -109,6 +109,7 @@ type MinerSubsystemConfig struct {
 	EnableSealing       bool
 	EnableSectorStorage bool
 	EnableMarkets       bool
+	EnableSectorIndexDB bool
```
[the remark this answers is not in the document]
shrenujbansal: done

storage/paths/db_index.go
shrenujbansal: done

storage/paths/db_index.go
shrenujbansal: Added. Could you review the parameters for the retries and wait time and see if they make sense? @magik6k

storage/paths/db_index.go
shrenujbansal: done

lib/harmony/harmonydb/sql/20230712.sql
```diff
@@ -0,0 +1,41 @@
+create table sectorlocation
```
magik6k, line 1: Can we either make this `SectorLocation`, or the one below `storagelocation`, so that we're consistent with text case?

magik6k: May also want to pick slightly different names, `storagelocation` / `sectorlocation` look very similar visually, maybe `storage_bucket` / `storage_objects` (or `path`/`files`, just something that looks visually distinct)

storage/paths/db_index.go
```diff
@@ -0,0 +1,986 @@
...
+	}
+
+	return true, nil
+
+}
+
+func (dbi *DBIndex) StorageLock(ctx context.Context, sector abi.SectorID, read storiface.SectorFileType, write storiface.SectorFileType) error {
+
+	retries := 5
+	waitTime := 1
```
magik6k, line 880: We want to block in StorageLock until we either succeed at getting the lock, or until the context is cancelled
* There should be no retry limit
* Ideally we'd have some backoff on initial retries, starting at let's say retry after 1s, and increase to 1min, then just try every 1min

storage/paths/db_index.go
```diff
@@ -0,0 +1,986 @@
...
+	// generate uuid for this lock owner
+	lockUuid := uuid.New()
+
+	var err error
+	for i := 0; i < retries; i++ {
+		_, err := dbi.lock(ctx, sector, read, write, lockUuid)
+		if err == nil {
+			break
+		}
+		time.Sleep(time.Duration(waitTime) * time.Second)
```
magik6k, line 891: Ideally would be time.After in a select which would quit early when the context gets cancelled

storage/paths/db_index.go
```diff
@@ -0,0 +1,986 @@
...
+	if read|write == 0 {
+		return false, nil
+	}
+
+	if read|write > (1<<storiface.FileTypes)-1 {
+		return false, xerrors.Errorf("unknown file types specified")
+	}
+
+	// Relinquish write locks
+	_, err := dbi.harmonyDB.Exec(ctx,
```
magik6k, line 842: This ctx is guaranteed to be cancelled, we probably want to use a `.Background / .TODO` here or in `StorageLock` where we call this `unlock`

magik6k (commented):
This is getting really close,

Few comments on the locking side of things, and it would be nice to avoid passing a query built through Sprintf in StorageBestAlloc. Other than that this looks shippable

magik6k:
(also CI seems sad, but not sure if this is related to this PR or feat/sturdypost)

storage/paths/db_index.go
shrenujbansal: Sorry, I don't quite understand the above statement. Could you provide an example of what you're suggesting?

storage/paths/db_index.go
shrenujbansal: Does an increasing backoff make sense here? Most of the time you just want to try to achieve the lock after the previous user is finished with it. If you keep on increasing the backoff, it just increases the chances of you not getting the lock the next time because a different user waiting for a shorter period will get to it first

shrenujbansal: Also, I don't think we should try to block indefinitely. There needs to be some threshold after which you just return an API error.

For ex. if the database goes down, we don't want API callers to be stuck forever waiting for a lock their not gonna get

storage/paths/db_index.go
shrenujbansal: done

storage/paths/db_index.go
shrenujbansal: done

storage/paths/db_index.go
magik6k:
```
select {
case <-time.After(time.Duration(waitTime) * time.Second):
case <-ctx.Done():
  return ctx.Err()
}
```

storage/paths/db_index.go
magik6k: But we want to block on trying to get the lock in case someone else is holding it, and the way I've read this code is that we will only attempt to get the lock for 5 seconds, which is not nearly long enough if there's a long-running operation holding the lock (which can be hours in case of PC1)

shrenujbansal: Right. The retry parameters need to be tuned as I mentioned in a different comment but does the general concept of having a regular backoff + limit on the amount of times you retry make sense?

I could see an increasing backoff make sense when there are different task profiles with touching the same file with large variances in hold time.
I suppose you can not have a limit and rely on a timeout limit of the original RPC call (if one existst) which cancels the ctx. But I don't know if those 2 things are the case atm. Happy to reconsider if they are

storage/paths/db_index.go
shrenujbansal: done

storage/paths/db_index.go
```diff
@@ -0,0 +1,999 @@
...
+
+	waitTime := 1
+	// generate uuid for this lock owner
+	lockUuid := uuid.New()
+
+	// retry with exponential backoff and block until lock is acquired
+	for {
+		locked, err := dbi.lock(ctx, sector, read, write, lockUuid)
+		// if err is not nil and is not because we cannot acquire lock, retry
+		if err != nil && !strings.Contains(err.Error(), "cannot acquire") {
```
magik6k, line 889: Can we check for some defined error? Define `var errFailedToAcquireLock = errors.New("...")` somewhere above, and return `xerrors.Errorf("acquiring write locks for sector..: %w", ..., errFailedToAcquireLock)`, then check here with `errors.Is` / `errors.As`.

magik6k (commented):
Mostly small nitpicky stuff, everything looks pretty solid now, next review should be shipping the code

node/builder_miner.go
```diff
@@ -126,10 +126,14 @@ func ConfigStorageMiner(c interface{}) Option {
...
 
+		Override(new(*harmonydb.DB), func(cfg config.HarmonyDB, id harmonydb.ITestID) (*harmonydb.DB, error) {
+			return harmonydb.NewFromConfigWithITestID(cfg)(id)
+		}),
+
 		If(cfg.Subsystems.EnableSectorStorage,
 			// Sector storage
-			Override(new(*paths.Index), paths.NewIndex),
-			Override(new(paths.SectorIndex), From(new(*paths.Index))),
+			Override(new(*paths.IndexProxy), paths.NewIndexProxyHelper(true)),
```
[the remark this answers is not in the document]
shrenujbansal: Ah yes nice catch

storage/paths/db_index.go
shrenujbansal: done

PR #11147, opened 2023-08-08
fix: CI: make test-unit-rest actually be the rest of the tests

arajasek:
## Related Issues

unit-rest is supposed to be a catch-all suite for all remaining tests. We currently run all the tests in unit-node, unit-storage, and unit-cli a second time in unit-rest, because we populate the exclusion map with absolute paths, but `WalkDir` checks against relative paths.

## Proposed Changes

Populate the map with both absolute and relative paths, thus avoiding all this duplicated work

## Checklist

- [ ] CI is green

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1043,7 +1043,7 @@ workflows:
           requires:
             - build
           suite: utest-unit-rest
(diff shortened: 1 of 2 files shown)


Stebalien (approved):
This makes sense, and I guess we likely don't want to test extern? Or do we?

jennijuju (approved):

arajasek:
> This makes sense, and I guess we likely don't want to test extern? Or do we?

I don't think we do -- those tests are independently covered.

PR #11157, opened 2023-08-10
feat: miner: implement FRC-0051

arajasek:
## Related Issues

#11104 

## Proposed Changes

- Add a (currently unconfigurable) EquivocationDelaySecs build parameter set to 2 seconds for all networks besides devnets
- In the `mineOne` process, at the _very end_, refresh the base we're mining on
- If the base has changed, we "refresh" our base to be the _intersection_ of the old base and new base. The only effect this should have is to _remove_ any blocks from miners that have equivocated.
  - If we do this, we need to re-select messages (since some might no longer be valid), and re-compute ticket if the MinTicket has changed
- If the base hasn't changed, proceed as always

## Checklist

- [ ] CI is green

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -579,6 +579,7 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 	// If the base has changed, we take the _intersection_ of our old base and new base,
 	// thus ejecting blocks from any equivocating miners, without taking any new blocks.
 	if newBase.TipSet.Height() == base.TipSet.Height() && !newBase.TipSet.Equals(base.TipSet) {
+		log.Warnf("base changed from %s to %s, taking intersection", base.TipSet.Key(), newBase.TipSet.Key())
 		newBaseMap := map[cid.Cid]struct{}{}
 		for _, newBaseBlk := range newBase.TipSet.Cids() {
 			newBaseMap[newBaseBlk] = struct{}{}
@@ -599,6 +600,8 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 			}
 
 			if !base.TipSet.MinTicket().Equals(refreshedBase.MinTicket()) {
+				log.Warn("recomputing ticket due to base refresh")
+
 				ticket, err = m.computeTicket(ctx, &rbase, round, refreshedBase.MinTicket(), mbi)
 				if err != nil {
 					err = xerrors.Errorf("failed to refresh ticket: %w", err)
@@ -606,6 +609,7 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 				}
 			}
 
+			log.Warn("re-selecting messages due to base refresh")
 			// refresh messages, as the selected messages may no longer be valid
 			msgs, err = m.api.MpoolSelect(ctx, refreshedBase.Key(), ticket.Quality())
 			if err != nil {
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -555,7 +555,7 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 	tProof := build.Clock.Now()
 
 	// get pending messages early,
-	msgs, err := m.api.MpoolSelect(context.TODO(), base.TipSet.Key(), ticket.Quality())
+	msgs, err := m.api.MpoolSelect(ctx, base.TipSet.Key(), ticket.Quality())
 	if err != nil {
 		err = xerrors.Errorf("failed to select messages for block: %w", err)
 		return nil, err
@@ -563,6 +563,11 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 
 	tEquivocateWait := build.Clock.Now()
 
+	// This next block exists to "catch" equivocating miners,
+	// who submit 2 blocks at the same height at different times in order to split the network.
+	// To safeguard against this, we make sure it's been EquivocationDelaySecs since our base was calculated,
+	// then re-calculate it.
+	// If the daemon detected equivocated blocks, those blocks will no longer be in the new base.
 	// TODO: make param
 	m.niceSleep(time.Until(base.ComputeTime.Add(2 * time.Second)))
 	newBase, err := m.GetBestMiningCandidate(ctx)
@@ -571,24 +576,38 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 		return nil, err
 	}
 
-	// Only factor in equivocated blocks if doing so will not risk us missing a block
-	if newBase.TipSet.Height() == base.TipSet.Height() && newBase.TipSet.MinTicket().Equals(base.TipSet.MinTicket()) {
+	// If the MinTicket is still the same, we take the _intersection_ of our old base and new base,
+	// thus ejecting blocks from any equivocating miners, without taking any new blocks.
+	// If the MinTicket is not the same, then the work we've done so far is no longer valid.
+	// Instead of choosing to miss a block, we submit our best-effort block anyway.
+	if !newBase.TipSet.Equals(base.TipSet) && newBase.TipSet.MinTicket().Equals(base.TipSet.MinTicket()) {
 		newBaseMap := map[cid.Cid]struct{}{}
 		for _, newBaseBlk := range newBase.TipSet.Cids() {
 			newBaseMap[newBaseBlk] = struct{}{}
 		}
 
-		refreshedBase := make([]*types.BlockHeader, 0, len(base.TipSet.Cids()))
+		refreshedBaseBlocks := make([]*types.BlockHeader, 0, len(base.TipSet.Cids()))
 		for _, baseBlk := range base.TipSet.Blocks() {
 			if _, ok := newBaseMap[baseBlk.Cid()]; ok {
-				refreshedBase = append(refreshedBase, baseBlk)
+				refreshedBaseBlocks = append(refreshedBaseBlocks, baseBlk)
 			}
 		}
 
-		base.TipSet, err = types.NewTipSet(refreshedBase)
-		if err != nil {
-			err = xerrors.Errorf("failed to create new tipset when refreshing: %w", err)
-			return nil, err
+		if len(refreshedBaseBlocks) != len(base.TipSet.Blocks()) {
+			refreshedBase, err := types.NewTipSet(refreshedBaseBlocks)
+			if err != nil {
+				err = xerrors.Errorf("failed to create new tipset when refreshing: %w", err)
+				return nil, err
+			}
+
+			base.TipSet = refreshedBase
+
+			// refresh messages, as the selected messages may no longer be valid
+			msgs, err = m.api.MpoolSelect(ctx, base.TipSet.Key(), ticket.Quality())
+			if err != nil {
+				err = xerrors.Errorf("failed to re-select messages for block: %w", err)
+				return nil, err
+			}
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


miner/miner.go
```diff
@@ -507,12 +510,14 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 		rbase = bvals[len(bvals)-1]
 	}
 
+	// TODO: refactor: only pass minTicket (and anything else needed) here
```
arajasek, line 513: Will do in either this or separate prelim PR

miner/miner.go
```diff
@@ -507,12 +510,14 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
...
 	ticket, err := m.computeTicket(ctx, &rbase, base, mbi)
 	if err != nil {
 		err = xerrors.Errorf("scratching ticket failed: %w", err)
 		return nil, err
 	}
 
+	// TODO: refactor: drop base.Tipset here entirely
```
arajasek, line 520: Will do in either this or separate prelim PR

miner/miner.go
```diff
@@ -560,6 +565,37 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 		return nil, err
 	}
 
+	tEquivocateWait := build.Clock.Now()
+
+	// TODO: make param
```
arajasek, line 570: Will do tomorrow

miner/miner.go
```diff
@@ -560,6 +565,37 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
...
+	m.niceSleep(time.Until(base.ComputeTime.Add(2 * time.Second)))
+	newBase, err := m.GetBestMiningCandidate(ctx)
+	if err != nil {
+		err = xerrors.Errorf("failed to refresh best mining candidate: %w", err)
+		return nil, err
+	}
+
+	// Only factor in equivocated blocks if doing so will not risk us missing a block
+	if newBase.TipSet.Height() == base.TipSet.Height() && newBase.TipSet.MinTicket().Equals(base.TipSet.MinTicket()) {
```
Stebalien, line 579: Shouldn't this be checking if the tipsets are not equal?

Stebalien (commented):
Looks like the right approach.

miner/miner.go
```diff
@@ -560,6 +565,37 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
...
+
+	// TODO: make param
+	m.niceSleep(time.Until(base.ComputeTime.Add(2 * time.Second)))
+	newBase, err := m.GetBestMiningCandidate(ctx)
+	if err != nil {
+		err = xerrors.Errorf("failed to refresh best mining candidate: %w", err)
+		return nil, err
+	}
+
+	// Only factor in equivocated blocks if doing so will not risk us missing a block
```
Stebalien, line 578: Needs a comment describing what we're doing (i.e., keeping the intersection between the two tipsets).

miner/miner.go
Stebalien: Wait, why are we looking at the min ticket at all?

Also, I realize there's a case we didn't consider: we switch to an entirely different tipset with a different set of parents. In that case, we can't look at the intersection... I'm not sure what we should do (ideally we'd be able to ask for the "best tipset with the current parents"...)

arajasek: Sorry, I should've explained my thinking here a lot more. My goal is to protect miner's chances of winning a block as much as possible, even if that allows for _some_ equivocation from other miners. Specifically, when only bother modifying our base tipset if all the work we've done so far (ticket scratching, etc.) is still valid. If the refreshed head invalidates any of our work, we just say "I'm choosing to ignore that", and submit the block anyway.

As a result, I think the only thing we need to look at here is the MinTicket -- if that's the same, all our work so far is still valid*, we can intersect our tipsets and submit our block on the intersection. If it's different, regardless of whether that's because we just have a new head, or because equivocation occurred in a way that caused the MinTicket to change, we submit our block as prepared.

This isn't necessarily the right thing to protect for, but I really don't want to increase the chance of missed blocks. Happy to hear arguments otherwise.

*all our work so far, with one exception that I hadn't though about: the messages we've selected might need to change. We can either just move message selection to after this check, or re-select messages in the unlikely case that we end up here. Thoughts?

arajasek: (But, yes, we should be short-circuiting if the tipsets are equal)

Stebalien: That mostly makes sense but I have a few questions:

1. What if the new min ticket is actually better?
2. What if switching the base tipset doesn't matter?

I'm also concerned that this makes things easier to attack. E.g., if I think I'm winning, I can equivocate because I'll be the min ticket and nobody will switch away from that min ticket.

Stebalien (commented):
The code looks correct. But we do need a test of some form.

miner/miner.go
```diff
@@ -554,12 +556,67 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
...
 
+	tEquivocateWait := build.Clock.Now()
+
+	// This next block exists to "catch" equivocating miners,
+	// who submit 2 blocks at the same height at different times in order to split the network.
+	// To safeguard against this, we make sure it's been EquivocationDelaySecs since our base was calculated,
+	// then re-calculate it.
+	// If the daemon detected equivocated blocks, those blocks will no longer be in the new base.
+	m.niceSleep(time.Until(base.ComputeTime.Add(time.Duration(build.EquivocationDelaySecs) * time.Second)))
+	newBase, err := m.GetBestMiningCandidate(ctx)
```
shrenujbansal, line 573: Does GetBestMiningCandidate already get rid of blocks from the equivocating miners?

miner/miner.go
```diff
@@ -554,12 +556,67 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
...
+		}
+
+		if len(refreshedBaseBlocks) != len(base.TipSet.Blocks()) {
+			refreshedBase, err := types.NewTipSet(refreshedBaseBlocks)
+			if err != nil {
+				err = xerrors.Errorf("failed to create new tipset when refreshing: %w", err)
+				return nil, err
+			}
+
+			if !base.TipSet.MinTicket().Equals(refreshedBase.MinTicket()) {
```
shrenujbansal, line 601: Just curious, if the refreshedBase is already different from the og Base, why could the ticket be the same?

miner/miner.go
arajasek: Yes, now that #11104 has landed. Really all that method does is ask for the node's `ChainHead`, which can rely on the new chainstore logic to have purged equivocation-tainted blocks.

miner/miner.go
arajasek: The `MinTicket` is simply the smallest (lexicographically earliest) ticket of all the blocks in a tipset. The tipset being different doesn't _necessarily_ mean that this value has changed. If the tipset changes from `[A, B, C, D]` to `[A, D]`, but `A` had the smallest ticket, the `MinTicket` would be unchanged.

miner/miner.go
```diff
@@ -554,12 +556,67 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
...
+	m.niceSleep(time.Until(base.ComputeTime.Add(time.Duration(build.EquivocationDelaySecs) * time.Second)))
+	newBase, err := m.GetBestMiningCandidate(ctx)
+	if err != nil {
+		err = xerrors.Errorf("failed to refresh best mining candidate: %w", err)
+		return nil, err
+	}
+
+	// If the base has changed, we take the _intersection_ of our old base and new base,
+	// thus ejecting blocks from any equivocating miners, without taking any new blocks.
+	if !newBase.TipSet.Equals(base.TipSet) {
```
magik6k, line 581: Don't we also want to check that the new base height matches? Or does GetBestMiningCandidate somehow always return the same height?

miner/miner.go
```diff
@@ -554,12 +556,67 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
...
+
+		if len(refreshedBaseBlocks) != len(base.TipSet.Blocks()) {
+			refreshedBase, err := types.NewTipSet(refreshedBaseBlocks)
+			if err != nil {
+				err = xerrors.Errorf("failed to create new tipset when refreshing: %w", err)
+				return nil, err
+			}
+
+			if !base.TipSet.MinTicket().Equals(refreshedBase.MinTicket()) {
+				ticket, err = m.computeTicket(ctx, &rbase, round, refreshedBase.MinTicket(), mbi)
```
magik6k, line 602: Isn't the ticket computed from drand and the power table at finality? If that's the case, how could the ticket ever change (assuming that the new base is always at the same round)

magik6k (commented):
Makes sense, 2 questions before I approve

miner/miner.go
arajasek: We seem to also incorporate chain randomness (from the parent) into it: https://github.com/filecoin-project/lotus/blob/master/miner/miner.go#L599

Reiers:
Been smooth sailing the whole week. Nothing alarming in the logs. No missed blocks. 

```
reiers@filecoin-amd-miner:~$ ./checkMinedToday_super.sh /home/reiers/yellifbreak_miner2.logGOOD! All blocks were mined!
Number of mined block(s):   3
Number of winning ticket(s): 3
Number of elections ticket(s): 3049
Number of times tookMilliseconds is over 10: 134
Number of times baseDelta has been over 10 seconds: 58
Number of SLASHFILTER logs: 0
Number of logs added in this commit: 0
Number of CAUTION block delay logs: 0
Number of completed mineOne logs: 3052
reiers@filecoin-amd-miner:~$ ./checkMinedToday_super.sh /home/reiers/yellifbreak_miner.log

GOOD! All blocks were mined!
Number of mined block(s):   24
Number of winning ticket(s): 24
Number of elections ticket(s): 7557
Number of times tookMilliseconds is over 10: 1814
Number of times baseDelta has been over 10 seconds: 84
Number of SLASHFILTER logs: 0
Number of logs added in this commit: 0
Number of CAUTION block delay logs: 0
Number of completed mineOne logs: 7581
```

miner/miner.go
```diff
@@ -599,13 +600,18 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 			}
 
 			if !base.TipSet.MinTicket().Equals(refreshedBase.MinTicket()) {
+				// TODO: Drop to Warn after initial testing
+				log.Error("recomputing ticket due to base refresh")
```
arajasek, line 604: Drop / lower me before merge

miner/miner.go
```diff
@@ -599,13 +600,18 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
...
+
 				ticket, err = m.computeTicket(ctx, &rbase, round, refreshedBase.MinTicket(), mbi)
 				if err != nil {
 					err = xerrors.Errorf("failed to refresh ticket: %w", err)
 					return nil, err
 				}
 			}
 
+			// TODO: Drop to Warn after initial testing
```
arajasek, line 613: Drop / lower me before merge

shrenujbansal (approved):

marco-storswift:
https://github.com/filecoin-project/lotus/blob/master/miner/miner.go#L647。  I see tPending took 38 second in logfile

arajasek:
@marco-storswift Thank you for the report! Can you please share a bit more of the logs around there? I'd like to take a look.

Your report also made me realize that the logs aren't quite correct here. I've opened #11241 to fix this up.

marco-storswift:
lotus-miner  version 1.23.2

marco-storswift:
> @marco-storswift Thanks for the response. What's the lotus daemon version? Also 1.23.2?

yes， the same version 1.23.2

PR #11159, opened 2023-08-11
fix: chainstore: do not get stuck in unhappy equivocation cases

arajasek:
## Related Issues

I was doing some manual testing / code inspection of #11104, and realized there's an unfortunate case here where we can get stuck, especially immediately after startup when the `tipsets` cache isn't populated

## Proposed Changes

When refreshing our current head, try to find something within 5 epochs. If we find NOTHING (due to equivocation, or due to the `tipsets` cache being unpopulated), just stick with your equivocated head.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/store/store.go
+++ b/chain/store/store.go
@@ -446,13 +446,15 @@ func (cs *ChainStore) RefreshHeaviestTipSet(ctx context.Context, newTsHeight abi
 			}
 		}
 
-		if newHeaviest == nil {
-			return xerrors.Errorf("failed to refresh to a new valid tipset")
-		}
-
-		errTake := cs.takeHeaviestTipSet(ctx, newHeaviest)
-		if errTake != nil {
-			return xerrors.Errorf("failed to take newHeaviest tipset as head: %w", err)
+		// if we've found something, we know it's the heaviest equivocation-free head, take it IMMEDIATELY
+		if newHeaviest != nil {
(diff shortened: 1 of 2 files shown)


chain/store/store.go
```diff
@@ -446,13 +446,15 @@ func (cs *ChainStore) RefreshHeaviestTipSet(ctx context.Context, newTsHeight abi
...
-			return xerrors.Errorf("failed to take newHeaviest tipset as head: %w", err)
+		// if we've found something, we know it's the heaviest equivocation-free head, take it IMMEDIATELY
+		if newHeaviest != nil {
+			errTake := cs.takeHeaviestTipSet(ctx, newHeaviest)
+			if errTake != nil {
+				return xerrors.Errorf("failed to take newHeaviest tipset as head: %w", err)
+			}
+		} else {
+			// if we haven't found something, just stay with our equivocation-y head
+			newHeaviest = cs.heaviest
```
Stebalien, line 457: Is there any reason not to just return nil here and move on?

arajasek: Yes -- we still want to try and form a tipset at the new height we've been notified about, and take that if it's better?

Stebalien: I think I'm convinced.

Stebalien (approved):

PR #11165, opened 2023-08-14
feat:miner:harmonytask

snadrus:
## Related Issues
HarmonyTask implementation (PR attempt 2)

## Proposed Changes
HarmonyTasks package and Resources package to manage tasks. 

## Additional Info
Tasks are listed in Yugabyte and competed-for through a greedy (but not too greedy) algorithm that enables tasks to hop from creator-machine to consumer machine or even from task-completion-machine to following-task-defined-elsewhere-machine. 
See doc.go 
Overview in https://docs.google.com/document/d/1lPe-6HdVe_oU0mD29f9Zi4233_xhSbyiXxM_6zXG4hY/edit#heading=h.k4h9btwn9p1y


magik6k:
The other PR: https://github.com/filecoin-project/lotus/pull/11108

snadrus:
@magik6k  Please review this one. The other somehow pulled-in feat/sturdypost so was abandoned.

lib/harmony/harmonytask/harmonytask.go
```diff
@@ -0,0 +1,386 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"sync/atomic"
+	"time"
+
+	"github.com/filecoin-project/lotus/lib/harmony/resources"
+	"github.com/gin-gonic/gin"
```
magik6k, line 11: In most other places we tend to use github.com/gorilla/mux where the default Golang mux/router isn't enough. No strong opinions, but would be slightly more consistent

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,284 @@
...
+	if !did {
+		return
+	}
+
+	if !h.considerWork([]TaskID{tID}) {
+		h.TaskEngine.bump(h.Name) // We can't do it. How about someone else.
+	}
+}
+
+func (h *taskTypeHandler) considerWork(ids []TaskID) (workAccepted bool) {
```
magik6k, line 54: Could add logs in places calling this to have some way to see what mechanisms are actually triggering work scheduling in production systems.

Would be either logs next to calls to considerWork, or some `reason string` param here that would get logged here

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,284 @@
...
+	logging "github.com/ipfs/go-log/v2"
+)
+
+var log = logging.Logger("harmonytask")
+
+type taskTypeHandler struct {
+	TaskInterface
+	TaskTypeDetails
+	TaskEngine *TaskEngine
+	Count      atomic.Int32 /// locked by TaskEngine's mutex
```
magik6k, line 22: Atomic locked?

lib/harmony/harmonytask/harmonytask.go
```diff
@@ -0,0 +1,386 @@
...
+	Adder(AddTaskFunc)
+}
+
+type AddTaskFunc func(extraInfo func(TaskID, *harmonydb.Tx) bool)
+
+type TaskEngine struct {
+	ctx            context.Context
+	handlers       []*taskTypeHandler
+	db             *harmonydb.DB
+	workAdderMutex *notifyingMx
```
magik6k, line 97: I don't see UnlockNotify be used anywhere

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,284 @@
...
+		log.Error(err)
+		return false
+	}
+	if tID == nil {
+		log.Infow("did not accept task", "task_id", ids[0], "reason", "CanAccept() refused")
+		return false
+	}
+
+	// 4. Can we claim the work for our hostname?
+	ct, err := h.TaskEngine.db.Exec(h.TaskEngine.ctx, "UPDATE harmony_task SET owner_id=$1 WHERE id=$2 AND owner_id IS NULL", h.TaskEngine.ownerID, *tID)
```
magik6k, line 88: Not a big issue, just a thought, but it feels a bit weird that we're accessing a private field in of a public field

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,284 @@
...
+			err := h.TaskEngine.db.QueryRow(context.Background(),
+				`SELECT owner_id FROM harmony_task WHERE id=$1`, *tID).Scan(&owner)
+			if err != nil {
+				log.Error("Cannot determine ownership: ", err)
+				return false
+			}
+			return owner == h.TaskEngine.ownerID
+		})
+		if doErr != nil {
+			log.Error("Do("+h.Name+", taskID="+strconv.Itoa(int(*tID))+") returned error: ", doErr)
```
magik6k, line 139: Structured Errorw for this would be nice

lib/harmony/resources/resources.go
```diff
@@ -0,0 +1,180 @@
...
+		found := 0
+		for _, b := range bytes.Split(b, []byte("\n")) {
+			if lotusRE.Match(b) {
+				found++
+			}
+		}
+		if found > 1 {
+			logger.Error("This Lotus process should run alone on a machine. Use CGroup.")
+		}
+	}
```
magik6k, line 119: Not entirely sure if we want to enforce this separation.

Conceptually I get why we want it, but due to how the scheduling of GPU-related work works on the rust side, it is very common to run a process per-GPU, perhaps with numactl/taskset assignments to separate cores, where I believe ps -ef will still "see" other processes. Using cgroups for separation (unless that's easy with systemd services and I don't know about that) will make creating that kind of setup quite a bit more involved.

lib/harmony/resources/resources.go
```diff
@@ -0,0 +1,180 @@
...
+	MachineID int
+}
+type Reg struct {
+	Resources
+	shutdown atomic.Bool
+}
+
+var logger = logging.Logger("harmonytask")
+
+var lotusRE = regexp.MustCompile("lotus-worker|lotus-harmony|yugabyted")
```
magik6k, line 41: Is yugabyted the single-node yugabyte binary?

* Do we actually want to disallow sharing one host between YB/lotus?
  * I get that one can run into resources of the other
  * But I'm not sure if it's worth complicating the setup for smaller SPs to the extent where you can't put YB next to one of lotus-providers
    * I can also see why we may want to push people to not de that
    * Maybe raise an Alert? (this would require setting up the Alerting system in lotus-provider, so.. idk, let's keep this constraint for now, will be easier to remove / loosen later than tighten it)
* In multi-node clusters the processes are called `yb-master` and `yb-tserver`

magik6k (commented):
This is really starting to look like code that is going to be actually in-use in production in near-ish future. Couple of comments, and definitely can't wait to see stuff implemented on top

lib/harmony/harmonytask/harmonytask.go
snadrus: For consistency. Later, a modern replacement might be nice:
Gin:     
   f.GET("/"+name+"/:tID", func(c *gin.Context) {
      tID := c.Params("tID")
Gorilla: 
   f.Path("/"+name+"/{tID}").Methods("GET").HandlerFunc(func(w http.ResponseWriter, r *http.Request){
      tID := mux.Vars(r)["tID"]

lib/harmony/harmonytask/taskTypeHandler.go
snadrus: ownerID? 
I didn't want anyone to change it from code. 
Although once in the database, security is a bit different.

lib/harmony/harmonytask/taskTypeHandler.go
snadrus: Thanks! Added line 107. I think that's what you're looking for.

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		log.Error("Could not record transaction: ", err)
+		return
+	}
+	if !cm {
+		log.Error("Committing the task records failed")
+	}
+}
+
+func (h *taskTypeHandler) AssertMachineHasCapacity() error {
```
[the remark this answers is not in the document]
snadrus: That's what TaskTypeDetails.Max already does. It'll need to be wired-up to a config system. 
I figure the limits of the hardware should be considered higher than the user-specified max, so there's no need to ignore the hardware limiter.

lib/harmony/resources/resources.go
snadrus: They could use task maximums too. I added that to the warning message.

lib/harmony/resources/resources.go
snadrus: It's less a constraint than a log complaint. 
I doubt that a small provider will overwhelm 1 reasonable machine with YB+Window+Winning, but I'd rather be extra chatty about what we aren't managing.

lib/harmony/harmonytask/doc.go
```diff
@@ -0,0 +1,79 @@
+/*
+	Package harmomnytask implements a pure (no task logic), distributed
```
shrenujbansal, line 2: *harmonytask

lib/harmony/harmonytask/doc.go
```diff
@@ -0,0 +1,79 @@
...
+*
+As we are not expecting DBAs in this database, it's important to know
+what grows uncontrolled. The only harmony_* table is _task_history
+(somewhat quickly) and harmony_machines (slowly). These will need a
+clean-up for after the task data could never be acted upon.
+but the design **requires** extraInfo tables to grow until the task's
+info could not possibly be used by a following task, including slow
+release rollout. This would normally be in the order of months old.
+*
+Other possible enhancements include more collaboative coordination
```
shrenujbansal, line 46: *collaborative

shrenujbansal:
Looking at the wdpost task generation logic, I'm a little concerned that the Task Interface framework doesn't quite fit the bill and there has to be a non significant amount of work done to make it fit

shrenujbansal:
How was the case with concurrent tasks inserting and taking a task tested?

shrenujbansal:
I'm happy to just get this merged in our branch since this is additive code and iterate further on it as we see fit

lib/harmony/resources/miniopencl/miniopencl.go
magik6k, line 1: If this code was borrowed from some other lib we should at least add a comment referencing it

lib/harmony/resources/resources.go
```diff
@@ -0,0 +1,180 @@
...
+	err := unix.Sysinfo(&in)
+	if err != nil {
+		return 0
+	}
+	// If this is a 32-bit system, then these fields are
+	// uint32 instead of uint64.
+	// So we always convert to uint64 to match signature.
+	return uint64(in.Freeram) * uint64(in.Unit)
+}
+*/
```
magik6k, line 180: Looks unused

```suggestion
```

lib/harmony/resources/resources.go
```diff
@@ -0,0 +1,180 @@
...
+		} else {
+			reg.MachineID = ownerID[0]
+			_, err := db.Exec(ctx, `UPDATE harmony_machines SET
+		   cpu=$1, ram=$2, gpu=$3, gpuram=$4 WHERE id=$6`,
+				reg.Cpu, reg.Ram, reg.Gpu, reg.GpuRam, reg.Resources.MachineID)
+			if err != nil {
+				return nil, err
+			}
+		}
+		CleanupMachines(context.Background(), db)
```
magik6k, line 75: This returns rows affected? If so, we should probably log that when it's more than 0

lib/harmony/harmonytask/harmonytask.go
```diff
@@ -0,0 +1,396 @@
...
+	// ONLY be called by harmonytask.
+	// Indicate if the task no-longer needs scheduling with done=true including
+	// cases where it's past the deadline.
+	Do(taskID TaskID, stillOwned func() bool) (done bool, err error)
+
+	// CanAccept should return if the task can run on this machine. It should
+	// return null if the task type is not allowed on this machine.
+	// It should select the task it most wants to accomplish.
+	// It is also responsible for determining disk space (including scratch).
+	CanAccept([]TaskID) (*TaskID, error)
```
magik6k, line 60: Should it also reserve disk space for the task?

lib/harmony/harmonytask/harmonytask.go
```diff
@@ -0,0 +1,396 @@
...
+
+	// block bumps & follows by unreg from DBs.
+	_, err := e.db.Exec(context.TODO(), `DELETE FROM harmony_task_impl WHERE owner_id=$1`, e.ownerID)
+	if err != nil {
+		log.Warn("Could not clean-up impl table: %w", err)
+	}
+	_, err = e.db.Exec(context.Background(), `DELETE FROM harmony_task_follow WHERE owner_id=$1`, e.ownerID)
+	if err != nil {
+		log.Warn("Could not clean-up impl table: %w", err)
+	}
```
magik6k, line 217: Define one ctx?

lib/harmony/harmonytask/harmonytask.go
```diff
@@ -0,0 +1,396 @@
...
+		if err != nil {
+			return nil, err
+		}
+		for _, w := range taskRet {
+			// edge-case: if old assignments are not available tasks, unlock them.
+			h := e.taskMap[w.Name]
+			if h == nil {
+				_, err := db.Exec(e.ctx, `UPDATE harmony_task SET owner=NULL WHERE id=$1`, w.ID)
+				if err != nil {
+					log.Error("Cannot remove self from owner field: ", err)
```
magik6k, line 184: For errors like this Errorw is better, generally easier to query / track / alert

lib/harmony/harmonytask/harmonytask.go
```diff
@@ -0,0 +1,396 @@
...
+					log.Error("Could not query harmony_task: ", err)
+					return // not recoverable here
+				}
+				if ct > 0 {
+					continue
+				}
+				// we need to create this task
+				if !src.h.Follows[fromName](TaskID(workAlreadyDone), src.h.AddTask) {
+					// But someone may have beaten us to it.
+					log.Infof("Unable to add task %s following Task(%d, %s)", src.h.Name, workAlreadyDone, fromName)
```
magik6k, line 274: It that's not uncommon we may want to set this log to debug

lib/harmony/harmonytask/taskTypeHandler.go
```diff
@@ -0,0 +1,285 @@
...
+		defer func() {
+			if r := recover(); r != nil {
+				log.Error("Recovered from a serious error "+
+					"while processing "+h.Name+" task "+strconv.Itoa(int(*tID))+": ", r)
+			}
+			h.Count.Add(-1)
+
+			h.recordCompletion(*tID, workStart, done, doErr)
+			if done {
+				h.triggerCompletionListeners(*tID)
```
magik6k, line 122: So if I'm reading this correctly, the case where the process dies between `recordCompletion` and `triggerCompletionListeners` is handled by other processes looking at harmony_task_history and figuring out what follows to trigger right?

magik6k (commented):
First pass, I feel like I'm starting to get a pretty good intuition of how this works.

There will probably be a number of edge-cases that we will discover as we exercise this logic, but to do that we need to start actually building on top of it.

lib/harmony/harmonytask/harmonytask.go
snadrus: yes, thanks.

lib/harmony/harmonytask/taskTypeHandler.go
snadrus: Correct. There's always the poller in the ugliest of cases.

lib/harmony/resources/resources.go
snadrus: yes, but it I will leave DiskFree() for easier CanAccept()

PR #11182, opened 2023-08-18
Take Download out of Sealer time

snadrus:
Downloading a 1GB file during a unit test is bad enough, but doing it in serial with a 600mb file followed by a lot of work is too much. 

A patch sent to Orjan downloads those in parallel. 

This takes the download out of the 30s timer and into a 1m30m (all test targets) window.
This will buy more time for the test to complete no matter what the delay is (if the problem is a timeout).

IF this fixes CI, we should discuss why we are having repeated, enormous downloads from filecoin websites as part of unit testing.

### DIFF
--- a/storage/sealer/ffiwrapper/sealer_test.go
+++ b/storage/sealer/ffiwrapper/sealer_test.go
@@ -412,6 +412,16 @@ func TestSealPoStNoCommit(t *testing.T) {
 	fmt.Printf("EPoSt: %s\n", epost.Sub(precommit).String())
 }
 
+func TestMain(m *testing.M) {
+	//setup()
+	// Here it no-longer is bound to 30s but has 1m30s for the whole suite.
+	getGrothParamFileAndVerifyingKeys(sectorSize)
+
+	code := m.Run()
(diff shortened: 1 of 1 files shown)


storage/sealer/ffiwrapper/sealer_test.go
```diff
@@ -412,6 +412,16 @@ func TestSealPoStNoCommit(t *testing.T) {
 	fmt.Printf("EPoSt: %s\n", epost.Sub(precommit).String())
 }
 
+func TestMain(m *testing.M) {
+	//setup()
+	// Here it no-longer is bound to 30s but has 1m30s for the whole suite.
```
magik6k, line 417: What 30s timer? Is there some test time limit I'm not aware of? Where is that defined?

snadrus: Go tests fail after 30 seconds by default. When implementing TestMain, you get 1m30s for the whole setup + test, and the 30s test-body window still applies.

magik6k (approved):

PR #11197, opened 2023-08-22
feat: shed: fix blockstore prune

Stebalien:
## Proposed Changes

1. Implement a pass-through for garbage collection on the idstore.
2. Fix the `lotus-shed state-prune` command.

NOTE: the new performance of running a full prune will be significantly less as this version doesn't batch. However, it should now actually _work_ and most users will be using the splitstore anyways.

## Checklist

- x ] PR title is in the form of of `<PR type>: <area>: <change being made>`
- [ ] CI is green

### DIFF
--- a/cmd/lotus-shed/pruning.go
+++ b/cmd/lotus-shed/pruning.go
@@ -12,7 +12,7 @@ import (
 
 	"github.com/filecoin-project/go-state-types/abi"
 
-	badgerbs "github.com/filecoin-project/lotus/blockstore/badger"
+	"github.com/filecoin-project/lotus/blockstore"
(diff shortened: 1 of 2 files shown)


cmd/lotus-shed/pruning.go
```diff
@@ -225,12 +214,12 @@ var stateTreePruneCmd = &cli.Command{
 		var deleteCount int
 		var goodHits int
 		for k := range keys {
-			if goodSet.HasRaw(k.Bytes()) {
+			if goodSet.Has(k) {
```
Stebalien, line 217: I'm not sure why we did used `HasRaw` before as this way should bee cheaper.

magik6k (approved):

PR #11199, opened 2023-08-23
Remove dead bootstrap nodes from mainnet.pi

GlacierWalrus:


GlacierWalrus:
ghcli screwup, stay tuned for the more exciting sequel PR

PR #11200, opened 2023-08-23
Remove dead bootstrap nodes from mainnet.pi

GlacierWalrus:
These nodes have been offline for many moons.


magik6k (commented):
This was already asked on Slack, but I'll also ask here for visibility: Was it confirmed that those nodes are actually dead, or are they just hard to reach outside of China?

GlacierWalrus:
Thanks for following up.

Not confirmed yet, I need to figure out the right test setup to confirm this (and then subsequently acquire nodes to test with).

GlacierWalrus:
ipfsmain nodes confirmed to be shut down, working on the rest :)

Stebalien:
@GlacierWalrus any updates? Alternatively, could you slim this down to a list of "confirmed dead" nodes?

GlacierWalrus:
@Stebalien thanks for staying on top of this.

Yep let's remove the ipfsmain boostrap nodes for now, and when this resurfaces as my highest priority I'll chase up the others.

Stebalien:
@GlacierWalrus thanks!

magik6k (approved):

PR #11202, opened 2023-08-23
fix: chain: Do not update message info cache until after message validation

gammazero:
## Related Issues
The sequence number used for replay detection was being updated before message validation confirmed that the message originated from the correct host. This would allow one host A to create a message with the ID of another host B that could then update the cached sequence number for B. While the message from A would fail validation and be ignored, the cached sequence number for B would get updated. This would lead to a temporary DoS for host B as its messages were incorrectly rejected as replays.

Host A would need to be authorized to send messages on the network, so this issue could only be caused by an authorized host.

## Proposed Changes
This fixes the issue by setting the cached sequence number after message validation.

A unit test is included that demonstrates the issue is fixed. It fails with the previous code and works with the fixed code.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/sub/incoming.go
+++ b/chain/sub/incoming.go
@@ -530,8 +530,8 @@ func (v *IndexerMessageValidator) Validate(ctx context.Context, pid peer.ID, msg
 	msgCid := idxrMsg.Cid
 
 	var msgInfo *peerMsgInfo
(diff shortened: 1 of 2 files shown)


arajasek (approved):
LGTM, thank you!

Issue #11106, opened 2023-07-27
Lotus-Provider Binary


snadrus:
## User Story
Lotus-Provider will be the new, YugabyteDB-tasked binaries that will eventually replace all the miner/worker tasks (while mostly leaving alone the old binaries in-case they have considerable efforts necessary to migrate to the new setup). 

## Acceptance Criteria
lotus-provider binary that builds with the normal commands, receives config, expresses it to the database.

```[tasklist]
### Techincal tasks:
- [ ] Create lotus-provider and wire it up
- [ ] It should use the Miner's config system and HTTP system to future be able to support Sealing APIs
- [ ] It should support SectorIndex APIs -- Note this has a dependency on Shrenuj
- [ ] It should run a harmony-task instance
```

```[tasklist]
### Techincal decesion to keep in mind:
- [ ] Stage it to resemble lotus-miner internally to be easy to later handle (the remaining) the Miner public APIs.
- [ ] It should use the HarmonyTask library
```

```[tasklist]
### Testing
- [ ] Can I build & run lotus-provider and have it implement file retrieval APIs and established HTTP APIs for monitoring?
```

rjan90:
Closed by: https://github.com/filecoin-project/lotus/pull/11203

PR #11203, opened 2023-08-24
Feat/lotus provider

snadrus:
Lotus-provider binary as defined in https://github.com/filecoin-project/lotus/issues/11106

On hold for harmonytask integration


Makefile
```diff
@@ -115,7 +119,7 @@ lotus-gateway: $(BUILD_DEPS)
 .PHONY: lotus-gateway
 BINS+=lotus-gateway
 
-build: lotus lotus-miner lotus-worker
+build: lotus lotus-miner lotus-worker lotus-provider
```
magik6k, line 122: Wouldn't include it in build just yet

node/builder.go
```diff
@@ -341,6 +341,38 @@ func Repo(r repo.Repo) Option {
...
+		return Options(
+			func(s *Settings) error { s.Base = true; return nil }, // mark Base as applied
+			ApplyIf(func(s *Settings) bool { return s.Config },
+				Error(errors.New("the Base() option must be set before Config option")),
+			),
+			Override(new(repo.LockedRepo), modules.LockedRepo(lr)), // module handles closing
+
+			Override(new(ci.PrivKey), lp2p.PrivKey),
+			Override(new(ci.PubKey), ci.PrivKey.GetPublic),
+			Override(new(peer.ID), peer.IDFromPublicKey),
```
magik6k, line 364: Shouldn't need this, it's just libp2p stuff, which we don't want on lotus-provider

lib/harmony/harmonydb/harmonydb.go
```diff
@@ -16,6 +17,7 @@ import (
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/kr/pretty"
```
magik6k, line 20: Can we avoid pulling in a random new dependency just to print a network address?

cmd/lotus-provider/main.go
```diff
@@ -0,0 +1,117 @@
...
+				Name:    "panic-reports",
+				EnvVars: []string{"LOTUS_PANIC_REPORT_PATH"},
+				Hidden:  true,
+				Value:   "~/.lotusprovider", // should follow --repo default
+			},
+			&cli.StringFlag{
+				Name:    "repo",
+				EnvVars: []string{"LOTUS_PATH"},
+				Hidden:  true,
+				Value:   "~/.lotus", // TODO: Consider XDG_DATA_HOME
```
magik6k, line 93: We can drop this TODO, the ship of moving .lotus somewhere else has realistically sailed at this point

cmd/lotus-provider/run.go
```diff
@@ -0,0 +1,284 @@
...
+		&cli.StringFlag{
+			Name:    "db_password",
+			EnvVars: []string{"LOTUS_DB_PASSWORD"},
+			Value:   "yugabyte",
+		},
+		&cli.StringFlag{
+			Name:    "db_port",
+			EnvVars: []string{"LOTUS_DB_PORT"},
+			Hidden:  true,
+			Value:   "5433",
```
magik6k, line 84: Would use kebab-case for flags, `--db-name` imo looks better than `--db_name`, is easier to type and consistent with other flags
```suggestion
			Name:    "db-name",
			EnvVars: []string{"LOTUS_DB_NAME"},
			Value:   "yugabyte",
		},
		&cli.StringFlag{
			Name:    "db-user",
			EnvVars: []string{"LOTUS_DB_USER"},
			Value:   "yugabyte",
		},
		&cli.StringFlag{
			Name:    "db-password",
			EnvVars: []string{"LOTUS_DB_PASSWORD"},
			Value:   "yugabyte",
		},
		&cli.StringFlag{
			Name:    "db-port",
			EnvVars: []string{"LOTUS_DB_PORT"},
			Hidden:  true,
			Value:   "5433",
```

cmd/lotus-provider/run.go
```diff
@@ -0,0 +1,284 @@
...
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "provider-api",
+			Usage: "2345",
+		},
+		&cli.BoolFlag{
+			Name:  "enable-gpu-proving",
+			Usage: "enable use of GPU for mining operations",
+			Value: true,
+		},
```
magik6k, line 49: Would drop this for now, I don't think anyone uses cpu for proving these days

cmd/lotus-provider/run.go
```diff
@@ -0,0 +1,284 @@
...
+			}
+
+			lr, err := r.Lock(repo.Provider)
+			if err != nil {
+				return err
+			}
+
+			var localPaths []storiface.LocalPath
+
+			if !cctx.Bool("no-local-storage") {
```
magik6k, line 141: Can we default to no local storage? Maybe ask TSE for opinion on this, but it arguably would be nice if the setup required explicitly setting paths for sector storage (calling `lotus-provider storage attach --some --stuff` after init)

cmd/lotus-provider/run.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/filecoin-project/lotus/node/modules/dtypes"
+	"github.com/filecoin-project/lotus/node/repo"
+	"github.com/filecoin-project/lotus/storage/paths"
+	"github.com/filecoin-project/lotus/storage/sealer/storiface"
+)
+
+var runCmd = &cli.Command{
+	Name:  "run",
+	Usage: "Start a lotus provider process",
+	Flags: []cli.Flag{
```
magik6k, line 40: From scriptable infra standpoint it might be nice to have a `--halt-after-init` flag which would only run init

cmd/lotus-provider/run.go
```diff
@@ -0,0 +1,284 @@
...
+			node.ShutdownHandler{Component: "provider", StopFunc: stop},
+		)
+
+		<-finishCh
+		return nil
+	},
+}
+
+func extractRoutableIP(timeout time.Duration) (string, error) {
+	conn, err := net.DialTimeout("udp", "8.8.8.8:80", timeout)
```
magik6k, line 278: This won't work in firewalled environments or in places on earth where 8.8.8.8 isn't easily reachable.

We will need a connection to some chain node rpc from lotus-provider in all(?) cases, so it should be reasonable to repurpose the method from lotus-worker to use the chain node connection.

(I also saw a helper to get this from the DB connection, that also seems reasonable)

cmd/lotus-provider/run.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/filecoin-project/lotus/storage/sealer/storiface"
+)
+
+var runCmd = &cli.Command{
+	Name:  "run",
+	Usage: "Start a lotus provider process",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "provider-api",
+			Usage: "2345",
```
magik6k, line 43: Can we set this to a different port (this one is the same as lotus-worker)

Something like `12300` seems reasonable, doesn't appear to be used, and the `00` part can be easily incremented in case people really want to run multiple endpoints on a single IP

scripts/lotus-provider.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Lotus Provider
+After=network.target
+After=lotus-daemon.service
+Wants=lotus-daemon.service
```
magik6k, line 5: Not sure if we want to require this, it reads as if we'd want lotus-deamon on the same machine, which in most production cases probably won't be the case

scripts/lotus-provider.service
```diff
@@ -0,0 +1,13 @@
...
+
+[Service]
+ExecStart=/usr/local/bin/lotus-provider run
+Environment=GOLOG_FILE="/var/log/lotus/provider.log"
+Environment=GOLOG_LOG_FMT="json"
```
magik6k, line 10: Almost definitely want to increase the default FD limit

```suggestion
Environment=GOLOG_LOG_FMT="json"
LimitNOFILE=1000000
```

magik6k (commented):
Looks good, couple of, mostly UX-ish, comments

cmd/lotus-provider/run.go
rjan90: I agree to default to no-local-storage. It makes the users more aware of where they are setting up their storage/sealing paths, and we already recommend users to use that flag in the docs: https://lotus.filecoin.io/storage-providers/setup/initialize/#initialization

cmd/lotus-provider/run.go
snadrus: The DB helper is the way to go

shrenujbansal (approved):
Just need to get CI to be green and then we can get this merged

PR #11211, opened 2023-08-26
feat: Add ETA to lotus sync wait

fridrik01:
## Related Issues
none

## Proposed Changes
This PR adds a ETA indicator to `lotus sync wait` command which I felt was missing when waiting for my node to finish syncing

## Additional Info
image

## Checklist

- [ ] CI is green

### DIFF
--- a/cli/util/eta.go
+++ b/cli/util/eta.go
@@ -5,13 +5,13 @@ import (
 	"time"
 )
 
-// ETA implements a very simple eta estimate based on the number of remaining items. It does not
+// ETA implements a very simple eta calculator based on the number of remaining items. It does not
 // require knowing the work size in advance and is therefore suitable for streaming workloads and
 // also does not require that consecutive updates have a monotonically decreasing remaining value.
 type ETA struct {
 	// max number of items to keep in memory
 	maxItems int
-	// a list of most recently updated items
+	// a queue of most recently updated items
 	items []item
 	// we store the last calculated ETA which we reuse if there was not change in remaining items
 	cachedETA string
@@ -47,14 +47,13 @@ func (e *ETA) Update(remaining int64) string {
 		return e.cachedETA
 	}
 
-	// remove oldest item from buffer if its full
+	// append the item to the queue and remove the oldest item if needed
 	if len(e.items) >= e.maxItems {
 		e.items = e.items[1:]
 	}
-
 	e.items = append(e.items, item)
 
-	// calculate the average processing time per item in the buffer
+	// calculate the average processing time per item in the queue
 	diffMs := e.items[len(e.items)-1].timestamp.Sub(e.items[0].timestamp).Milliseconds()
 	avg := diffMs / int64(len(e.items))
 
@@ -68,9 +67,8 @@ func (e *ETA) Update(remaining int64) string {
 
 func msToETA(ms int64) string {
 	seconds := ms / 1000
-	sec := seconds % 60
 	minutes := seconds / 60
-	min := minutes % 60
-	hour := minutes / 60
-	return fmt.Sprintf("%02dh:%02dm:%02ds", hour, min, sec)
+	hours := minutes / 60
+
+	return fmt.Sprintf("%02dh:%02dm:%02ds", hours, minutes%60, seconds%60)
 }
--- a/cli/util/eta.go
+++ b/cli/util/eta.go
@@ -12,57 +12,53 @@ type ETA struct {
 	// max number of items to keep in memory
 	maxItems int
 	// a queue of most recently updated items
-	items []item
+	items []time.Time
 	// we store the last calculated ETA which we reuse if there was not change in remaining items
-	cachedETA string
-}
-
-type item struct {
-	timestamp time.Time
-	remaining int64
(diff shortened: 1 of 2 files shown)


cli/util/eta.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// use that average processing time to estimate how long the remaining items will take
+	// and cache that ETA so we don't have to recalculate it on every update unless the
+	// remaining value changes
+	e.cachedETA = msToETA(avg * remaining)
+
+	return e.cachedETA
+}
+
+func msToETA(ms int64) string {
```
raulk, line 67: You should be able to construct a time.Duration here and then use its strong representation (unless you're wanting fancier formatting).

fridrik01: its a minor thing but using `time.Duration` does not render as nicely when redrawing the terminal line since its not fixed width (%02) so it keeps shifting

cli/util/eta.go
```diff
@@ -0,0 +1,74 @@
+package cliutil
+
+import (
+	"fmt"
+	"time"
+)
+
+// ETA implements a very simple eta calculator based on the number of remaining items. It does not
+// require knowing the work size in advance and is therefore suitable for streaming workloads and
+// also does not require that consecutive updates have a monotonically decreasing remaining value.
```
Stebalien, line 10: Off topic: I'm curious why you chose sliding window over exponentially weighted moving average. Sliding window is probably _better_, I'm just usually too lazy for that and was wondering what your reasoning was.

cli/util/eta.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// append the item to the queue and remove the oldest item if needed
+	if len(e.items) >= e.maxItems {
+		e.items = e.items[1:]
+	}
+	e.items = append(e.items, item)
+
+	// calculate the average processing time per item in the queue
+	diffMs := e.items[len(e.items)-1].timestamp.Sub(e.items[0].timestamp).Milliseconds()
+	avg := diffMs / int64(len(e.items))
```
Stebalien, line 58: Can we guarantee that we have one update per epoch? If not, I assume we need to store the number of epochs processed (or heights, I guess) as well.

Of course, heights can go backwards if we revert...

cli/util/eta.go
```diff
@@ -0,0 +1,74 @@
...
+		maxItems: maxItems,
+		items:    make([]item, 0),
+	}
+}
+
+// Update updates the ETA calculator with the remaining number of items and returns the ETA
+func (e *ETA) Update(remaining int64) string {
+	item := item{
+		timestamp: time.Now(),
+		remaining: remaining,
```
Stebalien, line 37: From the current code, we only need to track the _last_ value of `remaining`, right?

cli/util/eta.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	if len(e.items) == 0 {
+		e.items = append(e.items, item)
+		return ""
```
Stebalien, line 42: nit: "unknown"

cli/util/eta.go
fridrik01: I used it while debugging, but we can remove remaining from the struct and just store the last one.

cli/util/eta.go
Stebalien: I think this kind of works _if and only if_ we process blocks slower than once per second. In that case, each item will correspond to one block.

Otherwise, `avg` will be 1s. Or 10s/10. I.e., the time between the first item and the last (10s because we update once per second) and the number of items (10).

fridrik01: @Stebalien note that we only add items if the `remaining` changes so I don't think that's an issue, or am I misunderstanding?

Stebalien: Let's say remaining reduces by 5 each time. We should estimate that we're processing 5 blocks per second. However, instead, we'll estimate that we're processing 1 item per second and multiply `1 * remaining`.

fridrik01: Ah now I see what you mean, and yes you are indeed correct here, will update the PR, good catch!

cli/util/eta.go
```diff
@@ -12,46 +12,58 @@ type ETA struct {
...
 	}
 
-	// we ignore updates with the same remaining value and just return the previous ETA
-	if e.lastRemaining == remaining {
+	if e.items[len(e.items)-1].remaining == remaining {
+		// we ignore updates with the same remaining value and just return the previous ETA
 		return e.lastETA
+	} else if e.items[len(e.items)-1].remaining < remaining {
+		// in case remaining went up from previous updates then we don't know how many items
+		// were actually updated, lets assume 1 and update the remaining value of all items
```
Stebalien, line 50: This works but it does introduce an annoying "blip". Can we extrapolate from the current rate instead? That'll keep the rate the same.

Stebalien (commented):
Looks reasonable, although see my comment.

Stebalien (approved):

Stebalien:
"It works". It could probably use some more smoothing (e.g., a longer history), but it works and we can improve it later if we get complaints.

PR #11214, opened 2023-08-29
fix(client): single-root error message

rvagg:
Drive-by error message improvement having just received this error message many layers removed from this source.

I also wonder what it would take to remove this constraint finally? We've been relaxing the len(Roots)==1 constraint everywhere else.


magik6k (approved):

magik6k:
The market subsystem is deprecated on the lotus side, but I totally agree that we should allow an arbitrary number of roots in filecoin / go-car, anywhere from zero to N

PR #11228, opened 2023-08-31
feat: miner: log detailed timing breakdown when mining takes longer than the block's timestamp

arajasek:
## Related Issues

We're looking into reports of missed blocks due to mining taking too long. It's currently hard to analyze what the cause might be, because we only log the detailed breakdown of timing if the entire process was longer than 30s. However, we also want this information if by the time we're done producing the block, we've already reached the block's timestamp

## Proposed Changes

Log the detailed breakdown more liberally

## Checklist

- [ ] CI is green


fridrik01 (approved):
LGTM

PR #11235, opened 2023-09-02
add go linter - "unused"

snissn:
## Related Issues
when we upgraded go versions, it was pointed out to me that "unused" is a recommended linter. Adding unused triggered a few errors regarding unused close and I clean them up here

## Checklist

- [ ] CI is green


storage/wdpost/wdpost_run.go
```diff
@@ -32,6 +32,8 @@ import (
 )
 
 // recordPoStFailure records a failure in the journal.
+//
+//nolint:unused
```
fridrik01, line 36: are these false positives?

fridrik01 (approved):
Nice!

storage/wdpost/wdpost_run.go
arajasek: Yes, they're called over an API.

chain/sub/ratelimit/queue_test.go
```diff
@@ -0,0 +1,61 @@
+package ratelimit
+
+import (
+	"testing"
+)
+
+func TestQueue(t *testing.T) {
```
arajasek, line 7: No objection to this test, obviously, but I'm surprised it was needed to make the queue lok "used" -- it should already look used because of its use in rate limiting.

Do you understand why this was the case?

arajasek (commented):
Thank you for doing this!

chain/sub/ratelimit/queue_test.go
snissn: Yes! I sort of understand, for some reason pop is unused:

```
chain/sub/ratelimit/queue.go:38:17: func `(*queue).pop` is unused (unused)
func (q *queue) pop() int64 {
```

it's pretty common to write a little datastructure library, write out a bunch of api/funcs for it, and then not use a few when you have your final implementation. I think there's some combination of peeking and truncating tha thappens in ./chain/sub/ratelimit/window.go instead, but i haven't done a thorough dive into it. Given my expectaiton of this pattern of building a general tool and then just using some of it, the test made more sense than deleting pop in case in future we wanted to reorganize this code

arajasek: Got it, thanks for the explanation! Makes total sense.

Issue #11251, opened 2023-09-06
mining instability issue tracker


jennijuju:
> Note: Currently, this happens occasionally and we are not seeing a consistent network wide issue yet. The team monitoring closely. If you are impacted by this issue constantly, please let us know in #fil-lotus-help. I'd recommend those who observes an duration increase in their block generation to lower `PROPAGATION_DELAY_SECS` to 6-8 sec.

on Aug 28th, @beck-8 reported that the network is less stable with small forks, and storage providers are  observing delay in [block generation and an increase in missing winning blocks.](https://filecoinproject.slack.com/archives/CPFTWMY7N/p1693192780857229).

Unfortunately, we didn't get a chance to get to #10888 (other than message execution profiling #10892 ) work yet, in which aims  to develop better monitoring, tooling and metrics to help mining node operators to collect more detail and gain more visibility into the mining executions when an issue occur. Thus, currently, while we have some guess into what could be the cause, we are not so sure.

We are creating this issue to track the investigation and development effort. @fridrik01 will add on more details plans/update later. For now, the info we have is
- it seems like a `MpoolSelect` is taking longer when mpool is packed which impacts winningPoSt. @rjan90 suspected that we might have some high contention/waiting locks in mpool logics.
![image](https://github.com/filecoin-project/lotus/assets/42981373/32673881-eabd-47a0-ae22-9098b2eca6d5)
- @rjan90 was able to produce the issue with `lotus-shed mpool miner-select-messages ` , see findings here.
- @marco-storswift is also observing some `tPendings` could take >30s.

Potential Todos:
- [ ] https://github.com/filecoin-project/lotus/issues/11233
- [ ] review and land #10787 
- [ ] maybe #10692 ?
- [ ] maybe #10597 ?
- [ ] lotus miner team should prioritize #10754 (cc @snadrus )

fridrik01:
TL;DR My current theory on this, is that its related to splitstore pruning logic which seems to correlate exactly with the high latency of `MpoolSelect`. 

Here are updates after past 2 days debugging this issue:
1. I setup grafana/prometheus and installed the node_exporter and ran this to collect system metrics.
2. I added more logging in the `TipSetExecutor::ApplyBlocks` method too see where inside that method time was spent.
3. I created a new simple benchmark tool for calling binaries and analyzing their latencies/distribution  (similar to `lotus-bench rpc` expect for binaries).
[...]

PR #11269, opened 2023-09-14
fix: Add time slicing to splitstore purging to reduce lock congestion

fridrik01:
## Related Issues
See https://github.com/filecoin-project/lotus/issues/11251

## Context
When splitstore is enabled we perform compaction every 7h30m. During this compaction (which takes several minutes on my fast SSD) there is  relatively small purging step which takes approx 15sec. The purging step chunks the cids which should be purged and operates on them in blocks. In each block it acquires an exclusive lock which it shares with other I/O operation (for example FVM when applying messages). 

So when purging happens, other operations like MpoolSelect block for the whole 15sec duration when it needs to compute tipset state (TipSetState).

## Proposed Changes
I added a simple time slicing to the purging step which only tries to purge for 1 sec, while pausing for 1 sec until it is done. This effectively makes the purge 2x slower, but I don't think we care so much as it only happens every 7h30m and purging is already relatively fast compare the the whole splitstore compaction.

## Test plan
Using the same testing method as described in this [comment](https://github.com/filecoin-project/lotus/issues/11251#issuecomment-1717753704) I observed with this fix that during splitstore compaction the whole TipSetState went down from 15sec to ~2sec.

## Additional Info
See https://github.com/filecoin-project/lotus/issues/11251#issuecomment-1717753704

## Checklist

- [ ] CI is green

### DIFF
--- a/blockstore/splitstore/splitstore_compact.go
+++ b/blockstore/splitstore/splitstore_compact.go
@@ -1377,7 +1377,7 @@ func (s *SplitStore) purge(coldr *ColdSetReader, checkpoint *Checkpoint, markSet
 	err := coldr.ForEach(func(c cid.Cid) error {
 		batch = append(batch, c)
 		if len(batch) == batchSize {
-			// add some throttling to the purge as this a very disk I/O heavy operation that
+			// add some time slicing to the purge as this a very disk I/O heavy operation that
 			// requires write access to txnLk that may starve other operations that require
 			// access to the blockstore.
 			if time.Since(now) > time.Second {
--- a/blockstore/splitstore/splitstore_compact.go
+++ b/blockstore/splitstore/splitstore_compact.go
@@ -1372,9 +1372,19 @@ func (s *SplitStore) purge(coldr *ColdSetReader, checkpoint *Checkpoint, markSet
 		return err
 	}
 
+	now := time.Now()
+
 	err := coldr.ForEach(func(c cid.Cid) error {
 		batch = append(batch, c)
 		if len(batch) == batchSize {
+			// add some throttling to the purge as this a very disk I/O heavy operation that
+			// requires write access to txnLk that may starve other operations that require
+			// access to the blockstore.
+			if time.Since(now) > time.Second {
+				time.Sleep(time.Second)
+				now = time.Now()
+			}
+
 			return deleteBatch()
 		}
 
--- a/blockstore/splitstore/splitstore_compact.go
+++ b/blockstore/splitstore/splitstore_compact.go
@@ -66,8 +66,9 @@ var (
 )
 
 const (
-	batchSize  = 16384
-	cidKeySize = 128
+	batchSize              = 16384
+	cidKeySize             = 128
+	purgeWorkSliceDuration = time.Second
 )
 
 func (s *SplitStore) HeadChange(_, apply []*types.TipSet) error {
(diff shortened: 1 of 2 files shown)


Stebalien:
We likely either need to:

1. Take some kind of lock when computing tipsets (and/or mining blocks). 1s isn't quite enough to compute a tipset so we'll end up pausing multiple times.
2. Pause longer. IMO, there's no rush here unless I'm missing something.

That's not to say that the current patch won't help. Even if we only pause for 1s, that means we'll be able to spend 50% of the time computing the tipset which means it'll take ~6s on average instead of 3. But I'd still consider pausing for longer.

Stebalien:
> That's not to say that the current patch won't help. Even if we only pause for 1s, that means we'll be able to spend 50% of the time computing the tipset which means it'll take ~6s on average instead of 3. But I'd still consider pausing for longer.

I.e., I'd consider pausing for 4s every 1s, or something like that. That gives us 20% utilization, which means it should only make blocks take 20% longer.

_Ideally_ we'd look at datastore metrics and determine that the datastore is "busy". But that's more complicated.

Stebalien (approved):

Stebalien:
LGTM but this one needs an approval from @arajasek.

blockstore/splitstore/splitstore_compact.go
```diff
@@ -1372,9 +1372,21 @@ func (s *SplitStore) purge(coldr *ColdSetReader, checkpoint *Checkpoint, markSet
...
 	err := coldr.ForEach(func(c cid.Cid) error {
 		batch = append(batch, c)
 		if len(batch) == batchSize {
+			// add some time slicing to the purge as this a very disk I/O heavy operation that
+			// requires write access to txnLk that may starve other operations that require
+			// access to the blockstore.
+			elapsed := time.Since(now)
+			if elapsed > time.Second {
+				// work 1 second, sleep 4, or 20% utilization
+				time.Sleep(4 * elapsed)
```
arajasek, line 1386: `4 * elapsed`? Do we not want `4 * time.Second` here?

arajasek:
I don't see any harm in doing this, but I'm not sure I fully believe the logic here. 

> When splitstore is enabled we perform compaction every 7h30m. During this compaction (which takes several minutes on my fast SSD) there is relatively small purging step which takes approx 15sec. The purging step chunks the cids which should be purged and operates on them in blocks. In each block it acquires an exclusive lock which it shares with other I/O operation (for example FVM when applying messages).

We're saying that the purging step takes 15 seconds, but operates in batches. The batch size is 16k, which seems relatively small to me. We take the lock for each batch.

> So when purging happens, other operations like MpoolSelect block for the whole 15sec duration when it needs to compute tipset state (TipSetState).

In that case, we shouldn't be hogging the lock for the entire 15s, no? We're aggressively trying to grab it between batches, but we should still be releasing the lock for other operations in between batches. Is it the case that a singly batch is taking 15s?

arajasek (commented):
See comment about the premise itself. If we're landing, let's make the "work" time and "sleep" time constants (no need to be configurable yet).

blockstore/splitstore/splitstore_compact.go
fridrik01: if we schedule slightly over second, this means that we sleep a bit longer guaranteeing a 20% utilization

arajasek: Yup, that makes sense.

arajasek:
@fridrik01 Aaah, the lock contention is for every single _message_ application. Got it, that makes complete sense. Thank you!

arajasek (approved):
Make `time.Second` a constant, but LGTM!

PR #11270, opened 2023-09-14
feat: Add lotus-bench cli option to stress test any binary

fridrik01:
## Related Issues
Written to help with solving https://github.com/filecoin-project/lotus/issues/11251
Extension to the RPC benchmark introduced in https://github.com/filecoin-project/lotus/pull/10761

## Proposed Changes
This PR extends the RPC benchmark to support running any binary and analyze their latency and histogram distribution, track most common errors, perform stress testing under different concurrency levels and see how it works under different QPS. 

## Test plan

Lets say you want to understand how `lotus-shed mpool miner-select-messages` behaves over time. 

You can use this new `lotus-bench cli` tool to achieve this, for example the following command runs the command using a single thread (concurrency=1) at 1 qps and provides a realtime updated latency/histogram UI and it runs for 24 hours. 

`lotus-bench cli  --cmd='lotus-shed mpool miner-select-messages' --concurrency=1 --qps=1 --watch=5s --duration=24h`

It will produce something like:
image

## Additional Info
This works exactly the same as the rpc benchmark, which has a detailed summary (see [PR](https://github.com/filecoin-project/lotus/pull/10761))

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus-bench/cli.go
+++ b/cmd/lotus-bench/cli.go
@@ -74,7 +74,7 @@ Here are some real examples:
 
 		var cmds []*CMD
 		for _, str := range cctx.StringSlice("cmd") {
-			entries := strings.SplitN(str, ":", 4)
+			entries := strings.SplitN(str, ":", 3)
 			if len(entries) == 0 {
 				return errors.New("invalid cmd format")
 			}
@@ -153,7 +153,7 @@ Here are some real examples:
 				ticker := time.NewTicker(cctx.Duration("watch"))
 				for {
 					clearAndPrintReport := func() {
-						// clear the screen move the curser to the top left
+						// clear the screen move the cursor to the top left
 						fmt.Print("\033[2J")
 						fmt.Printf("\033[%d;%dH", 1, 1)
 						for i, cmd := range cmds {
--- a/cmd/lotus-bench/cli.go
+++ b/cmd/lotus-bench/cli.go
@@ -1,0 +1,312 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"os/signal"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/urfave/cli/v2"
+)
+
+var cliCmd = &cli.Command{
+	Name:  "cli",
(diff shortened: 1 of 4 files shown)


cmd/lotus-bench/cli.go
```diff
@@ -0,0 +1,312 @@
...
+* Gives a nice reporting summary of the stress testing of each command (including latency distribution, histogram and more)
+* Easy to use
+
+To use this benchmark you must specify the commands you want to test using the --cmd options, the format of it is:
+
+  --cmd=CMD[:CONCURRENCY][:QPS] where only NAME is required.
+
+Here are some real examples:
+  lotus-bench cli --cmd='lotus-shed mpool miner-select-messages' // runs the command with default concurrency and qps
+  lotus-bench cli --cmd='lotus-shed mpool miner-select-messages:3'  // override concurrency to 3
```
Stebalien, line 34: Using `:` could cause issues down the road (for any command including a literal `:`. Why not just rely on the flags?

Also, for the command, I'd suggest using positional parameters. I.e., "the rest of the args".

Stebalien (commented):
This seems generally useful but an odd addition to `lotus-bench` given that it's general-purpose.

It looks like you'd be better off with something like https://github.com/sharkdp/hyperfine (available in pretty much every package manager)

fridrik01:
> This seems generally useful but an odd addition to `lotus-bench` given that it's general-purpose.
> 
> It looks like you'd be better off with something like https://github.com/sharkdp/hyperfine (available in pretty much every package manager)

Yes, this could be an independent general purpose tool but I put it here for a lack of a better place to put it as it already shares many of the same code when stress testing our RPC endpoints, and this would enable us to do the same by stress testing lotus backend through our different cli commands.

I am not familiar with hyperfine but this does seem different enough as its based on generating specific workloads for a  certain duration where you can tune the concurrency/qps for each independently and then get simple histogram/latency distribution.

cmd/lotus-bench/cli.go
arajasek: +1 here.

arajasek (approved):
This LGTM, and I'm personally happy to land as-is. @magik6k do you wanna weigh in too?

magik6k:
> @magik6k do you wanna weigh in too?

Didn't review, but I'm fine with this landing

PR #11277, opened 2023-09-19
Feat: Lotus Daemon CLI: Auto remove existing chain if importing chain file or snapshot

mb1896:
## Related Issues
Based on the discussion https://filecoinproject.slack.com/archives/CP50PPW2X/p1695108301930859

## Proposed Changes
When user wants to import chain file or snapshot, we will by default remove the local chain data, unless the user says otherwise.

## Additional Info
None.

## Checklist

- [ ] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ This feature release requires a **minimum Go version of v1.19.12 or higher to su
   - feat: Lotus Gateway: add MpoolPending, ChainGetBlock and MinerGetBaseInfo ([filecoin-project/lotus#10929](https://github.com/filecoin-project/lotus/pull/10929))
 
 ## Improvements
+  - chore: Auto remove local chain data when importing chain file or snapshot ([filecoin-project/lotus#11277](https://github.com/filecoin-project/lotus/pull/11277))
   - chore: update ffi & fvm ([filecoin-project/lotus#11040](https://github.com/filecoin-project/lotus/pull/11040))
   - feat: Make sure we don't store duplidate actor events caused to reorgs in events.db ([filecoin-project/lotus#11015](https://github.com/filecoin-project/lotus/pull/11015))
   - sealing: Use only non-assigned deals when selecting snap sectors ([filecoin-project/lotus#11002](https://github.com/filecoin-project/lotus/pull/11002))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Lotus changelog
 
 # UNRELEASED
+- chore: Auto remove local chain data when importing chain file or snapshot ([filecoin-project/lotus#11277](https://github.com/filecoin-project/lotus/pull/11277))
 
 ## New features
 - feat: Added new tracing API (**HIGHLY EXPERIMENTAL**) supporting two RPC methods: `trace_block` and `trace_replayBlockTransactions` ([filecoin-project/lotus#11100](https://github.com/filecoin-project/lotus/pull/11100))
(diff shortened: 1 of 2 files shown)


jennijuju:
please do follow the pr template and fill the relevant part in!

cmd/lotus/daemon.go
```diff
@@ -269,7 +269,17 @@ var DaemonCmd = &cli.Command{
...
+		snapshot := cctx.String("import-snapshot")
+		willImportChain := false
+		if chainfile != "" || snapshot != "" {
+			if chainfile != "" && snapshot != "" {
+				return fmt.Errorf("cannot specify both 'import-snapshot' and 'import-chain'")
+			}
+			willImportChain = true
+		}
+
+		if cctx.Bool("remove-existing-chain") || willImportChain {
```
jennijuju, line 282: should we specify the behaviour expected if `remove-existing-chain` is set to be false yet `willImportChain` is true? warning & remove, or error out? or follow `remove-existing-chain`to not remove and continue with import?

mb1896: Good point, maybe we should just error out. Otherwise we are somehow creating undefined behaviors?

arajasek: @mb1896 What I would suggest is:

- if `remove-existing-chain`, just remove the chain
- if it's not specified, prompt the user with a y/n question asking whether they want the chain to be deleted -- if yes, do so, if not proceed without doing so

mb1896: Thansk for the comments! Fixed

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ This feature release requires a **minimum Go version of v1.19.12 or higher to su
   - feat: Lotus Gateway: add MpoolPending, ChainGetBlock and MinerGetBaseInfo ([filecoin-project/lotus#10929](https://github.com/filecoin-project/lotus/pull/10929))
 
 ## Improvements
+  - chore: Auto remove local chain data when importing chain file or snapshot ([filecoin-project/lotus#11277](https://github.com/filecoin-project/lotus/pull/11277))
```
arajasek, line 37: Thanks for including this! It needs to be moved up to the `UNRELEASED` section, though

cmd/lotus/daemon.go
```diff
@@ -269,7 +269,37 @@ var DaemonCmd = &cli.Command{
...
+				return fmt.Errorf("cannot specify both 'import-snapshot' and 'import-chain'")
+			}
+			willImportChain = true
+		}
+
+		willRemoveChain := cctx.Bool("remove-existing-chain")
+		if willImportChain && !willRemoveChain {
+			// Confirm with the user about the intention to remove chain data.
+			reader := bufio.NewReader(os.Stdin)
+			fmt.Print("Importing chain or snapshot will by default delete existing local chain data. Do you want to proceed and delete? (yes/no): ")
```
arajasek, line 286:
```suggestion
			fmt.Print("Do you want to delete existing local chain data as part of the import process? (yes/no): ")
```

arajasek: (just a bit more clear, there isn't really a default since we're forcing users to specify)

cmd/lotus/daemon.go
```diff
@@ -269,7 +269,37 @@ var DaemonCmd = &cli.Command{
...
+			userInput, err := reader.ReadString('\n')
+			if err != nil {
+				return xerrors.Errorf("reading user input: %w", err)
+			}
+			userInput = strings.ToLower(strings.TrimSpace(userInput))
+
+			if userInput == "yes" {
```
arajasek, line 293: Probably want to handle case-sensitivity here

arajasek (approved):
LGTM, feel free to merge after addressing minor comments.

CHANGELOG.md
mb1896: Thanks. Fixed

cmd/lotus/daemon.go
mb1896: You are right, however we couldn't differentiate between
- user not specifying a flag value
- user specifically specifying a flag to false

That's why we have to ask for the user's input...

cmd/lotus/daemon.go
mb1896: The `ToLower()` function handled it.

PR #11278, opened 2023-09-20
LotusProvider database-oriented config system

snadrus:
## Related Issues
Lotus-Provider config

## Proposed Changes
Config in the DB

## Checklist

- [ ] CI is green


cmd/lotus-provider/config.go
```diff
@@ -1,61 +1,173 @@
...
+		}
+
+		fn := args.First()
+		bytes, err := os.ReadFile(fn)
+		if err != nil {
+			return fmt.Errorf("cannot read file %w", err)
+		}
+
+		lp := config.DefaultLotusProvider() // ensure it's toml
+		toml.Decode(string(bytes), lp)
```
magik6k, line 74: Any error to check here?

jennijuju:
can we have better PR title and pr description plz? while I understand this is just merging to a feature branch, however, I doubt we will have a thorough PR description to cover all the smallest pieces when its time to merge to master. as it is today -
- its hard for devs who just wanna get quick context on the PR to know what this is about
- impossible for user to tell what this means to them
- hard for @filecoin-project/lotus-tse  to then test and create user docs (we can let this point go if you plan to do the first pass on user docs yourself)

lib/harmony/harmonydb/sql/20230919.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE harmony_config (
+    id SERIAL PRIMARY KEY NOT NULL,
+    title VARCHAR(300) UNIQUE NOT NULL,
```
magik6k, line 3: Probably being nitpicky here:
* Why use varchar instead of simple text?
* Why have unique title and id? Can we just have the title?
* Do we want a separate `type` so that we can support other node types can also use this?
* super-nit: maybe use `name` instead of `titile`?

cmd/lotus-provider/config.go
```diff
@@ -1,61 +1,180 @@
...
+		if args.Len() != 1 {
+			return errors.New("must have exactly 1 arg for the file name")
+		}
+		db, err := makeDB(cctx)
+		if err != nil {
+			return err
+		}
+
+		fn := args.First()
+		bytes, err := os.ReadFile(fn)
```
magik6k, line 69: Can we make the arg optional, and use stdin when no args are set?

node/config/types.go
```diff
@@ -65,6 +65,15 @@ type StorageMiner struct {
 	HarmonyDB HarmonyDB
 }
 
+type LotusProviderConfig struct {
+	EnableWindowPost  bool
+	EnableWinningPost bool
```
magik6k, line 70: Maybe put this in a non-top-level ProviderSubsystemsConfig struct?

lib/harmony/harmonydb/sql/20230919.sql
snadrus: Titles should have reasonable length limits. 
id is far better for if we ever need to do joins. 
type: A single text blob covering all products makes better sense to me so they don't need to duplicate anything. 
name is a SQL reserved word: it's doable but harder to work with: `name`

node/config/types.go
snadrus: Then what would we munge into the layer's name? (today myLayer.toml becomes myLayer)

cmd/lotus-provider/config.go
snadrus: Then what would we munge into the layer's name? (today myLayer.toml becomes myLayer)

snadrus:
test-unit-storage is a problem with pulling in nv21's go-state-types

magik6k:
`make gen` is angry

```
diff --git a/node/config/doc_gen.go b/node/config/doc_gen.go
index c3af2740c..e2e47afc7 100644
--- a/node/config/doc_gen.go
+++ b/node/config/doc_gen.go
@@ -703,14 +703,8 @@ over the worker address if this flag is set.`,
        },
        "LotusProviderConfig": {
                {
-                       Name: "EnableWindowPost",
-                       Type: "bool",
-
-                       Comment: ``,
-               },
-               {
-                       Name: "EnableWinningPost",
-                       Type: "bool",
+                       Name: "Subsystems",
+                       Type: "ProviderSubsystemsConfig",
 
                        Comment: ``,
                },
@@ -921,6 +915,20 @@ This is useful to allow workers to bypass the lotus miner to access sector infor
                        Comment: ``,
                },
        },
+       "ProviderSubsystemsConfig": {
+               {
+                       Name: "EnableWindowPost",
+                       Type: "bool",
+
+                       Comment: ``,
+               },
+               {
+                       Name: "EnableWinningPost",
+                       Type: "bool",
+
+                       Comment: ``,
+               },
+       },
        "ProvingConfig": {
                {
                        Name: "ParallelCheckLimit",
```

(pretty sure that you can paste that into `git apply` instad of running `make docsgen-cli`)

rjan90:
IIRC, we would also need the `StartEpochSealingBuffer` config, as its one of the configs that is currently in the `Dealmaking` section, but are used outside of Markets, and moving that into the SealingConfig would make sense. cc: @magik6k to confirm

documentation/en/default-lotus-provider-config.toml
```diff
@@ -1,2 +1,190 @@
...
+  # 
+  # WARNING: Setting this value too high may make the node crash by running out of stack
+  # WARNING: Setting this value too low may make sector challenge reading much slower, resulting in failed PoSt due
+  # to late submission.
+  # 
+  # After changing this option, confirm that the new value works in your setup by invoking
+  # 'lotus-miner proving compute window-post 0'
+  #
+  # type: int
+  #ParallelCheckLimit = 0
```
rjan90, line 85: I think this value should be 32 to be kept in-line with what we already have?

documentation/en/default-lotus-provider-config.toml
```diff
@@ -1,2 +1,190 @@
...
+
+  # Maximum amount of time a proving pre-check can take for a sector. If the check times out the sector will be skipped
+  # 
+  # WARNING: Setting this value too low risks in sectors being skipped even though they are accessible, just reading the
+  # test challenge took longer than this timeout
+  # WARNING: Setting this value too high risks missing PoSt deadline in case IO operations related to this sector are
+  # blocked (e.g. in case of disconnected NFS mount)
+  #
+  # type: Duration
+  #SingleCheckTimeout = "0s"
```
rjan90, line 95: This should be 10m0s if we want it to be a 1:1 mapping of what we currently have.

documentation/en/default-lotus-provider-config.toml
```diff
@@ -1,2 +1,190 @@
...
+  # Maximum amount of time a proving pre-check can take for an entire partition. If the check times out, sectors in
+  # the partition which didn't get checked on time will be skipped
+  # 
+  # WARNING: Setting this value too low risks in sectors being skipped even though they are accessible, just reading the
+  # test challenge took longer than this timeout
+  # WARNING: Setting this value too high risks missing PoSt deadline in case IO operations related to this partition are
+  # blocked or slow
+  #
+  # type: Duration
+  #PartitionCheckTimeout = "0s"
```
rjan90, line 106: This should be 20m0s if we want it to be a 1:1 mapping of what we currently have.

documentation/en/default-lotus-provider-config.toml
snadrus: good catch! I was not setting defaualts for everything.

snadrus:
I'm pretty sure the remaining tests have more to do with NV21 deps leaking in than with config code which touches nothing but lotus-provider & config types.

cmd/lotus-provider/config.go
magik6k: Could have a separate `--title` flag (much like kubo has `--stdin-name` for ipfs-add), and make that flag non-optional when pulling input from stdin

cmd/lotus-bench/main.go
```diff
@@ -776,9 +779,7 @@ func runSeals(sb *ffiwrapper.Sealer, sbfs *basicfs.Provider, numSectors int, par
 		start := time.Now()
 		log.Infof("[%d] Writing piece into sector...", i)
 
-		r := rand.New(rand.NewSource(100 + int64(i)))
-
-		pi, err := sb.AddPiece(context.TODO(), sid, nil, abi.PaddedPieceSize(sectorSize).Unpadded(), r)
```
magik6k, line 781: Why do we need this change? I don't think anything about using deterministic rand in a benchmark is wrong, and it makes the benchmark more verifiable

magik6k (commented):
Other than a few small comments LGTM

cmd/lotus-provider/config.go
```diff
@@ -1,61 +1,221 @@
...
+		}
+		meta, err := toml.Decode(text, &lp)
+		if err != nil {
+			return nil, fmt.Errorf("could not read layer, bad toml %s: %w", layer, err)
+		}
+		for _, k := range meta.Keys() {
+			have = append(have, strings.Join(k, " "))
+		}
+	}
+	_ = have // FUTURE: verify that required fields are here.
```
shrenujbansal, line 219: If you add verification of fields here, does this then need to be updated every time a config is added/removed?

orjanCheckThisIn.toml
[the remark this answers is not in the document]
snadrus: oops. That was a demo file.

chain/gen/genesis/miners.go
```diff
@@ -247,7 +247,8 @@ func SetupStorageMiners(ctx context.Context, cs *store.ChainStore, sys vm.Syscal
 			}
 
 			params := &markettypes.PublishStorageDealsParams{}
-			for _, preseal := range m.Sectors {
+			for _, presealTmp := range m.Sectors {
+				preseal := presealTmp
```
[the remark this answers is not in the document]
snadrus: There were a lot of CI failures showing up, and nobody was reviewing it. 
I found later that Mikers had changed the linter settings which may have caused the warnings.

cmd/lotus-provider/config.go
snadrus: There is a lot more to it if we have this config unified across Boost and potential 3rd-party consumers, so I think it's not the time to take on all the details now. 
I am leaning toward [JSONSchema](https://json-schema.org/) as a generic way to accept 3rd-party config requirements that we can impose here.

lib/harmony/harmonydb/sql/20230919.sql
snadrus: along with being convention, internally most databases will create this column if you do not. The only way to access it is at creation time.

PR #11297, opened 2023-09-26
chore: config: Update todo in UseSyntheticPoRep

rjan90:
## Proposed Changes
Updates the UseSyntheticPoRep config with actual numbers of what the cache-size is (11GiB) after completing PreCommit 2.

## Checklist

- [ ] CI is green

### DIFF
--- a/documentation/en/default-lotus-miner-config.toml
+++ b/documentation/en/default-lotus-miner-config.toml
@@ -666,9 +666,9 @@
   # env var: LOTUS_SEALING_TERMINATEBATCHWAIT
   #TerminateBatchWait = "5m0s"
 
-  # UseSyntheticPoRep will reduce the amout of data held on disk in the WaitSeed phase to 32GiB
-  # at the cost of having to precompute the synthetic challenges.
-  # TODO: put estimated costs here
+  # UseSyntheticPoRep, when set to true, will reduce the amount of cache data held on disk after the completion of PreCommit 2 to 11GiB.
+  # This is achieved by precomputing "synthetic" challenges from CommR.
+  # Note that this comes at a cost of additional computation to generate these synthetic challenges.
   #
   # type: bool
   # env var: LOTUS_SEALING_USESYNTHETICPOREP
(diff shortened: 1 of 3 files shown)


documentation/en/default-lotus-miner-config.toml
```diff
@@ -666,9 +666,9 @@
   # env var: LOTUS_SEALING_TERMINATEBATCHWAIT
   #TerminateBatchWait = "5m0s"
 
-  # UseSyntheticPoRep will reduce the amout of data held on disk in the WaitSeed phase to 32GiB
-  # at the cost of having to precompute the synthetic challenges.
-  # TODO: put estimated costs here
+  # UseSyntheticPoRep, when set to true, will reduce the amount of cache data held on disk after the completion of PreCommit 2 to 11GiB.
+  # This is achieved by precomputing "synthetic" challenges from CommR.
```
magik6k, line 670: It’s not computed from commr, it’s reading a lot of challenges from the whole cache, and later we’re picking a few from that set in C1

documentation/en/default-lotus-miner-config.toml
```diff
@@ -666,9 +666,9 @@
...
+  # Note that this comes at a cost of additional computation to generate these synthetic challenges.
```
magik6k, line 671: Do we have some number to put on this as a rough guidance?

documentation/en/default-lotus-miner-config.toml
rjan90: Yeah, I can probably update the wording to "Based on the CommR, the Storage Providers generates a list of "Synthetic Challenges" as written in the [FIP itself](https://fips.filecoin.io/FIPS/fip-0059.html) says that:

Also happy to discard that sentence, as I´m unsure what value that sentence actually brings to someone reading the config.

documentation/en/default-lotus-miner-config.toml
rjan90: I do not have any numbers right now, but the additional costs should be minimal. I can see if we can get a SynthPoRep vs non SynthPoRep sealing benchmark done

rjan90: As discussed, removed:
> Note that this comes at a cost of additional computation to generate these synthetic challenges.

Given that costs are negligible, and we currently do not have any good estimations on the diff here between SynthPoReps vs Non-SynthPoRep

PR #11321, opened 2023-10-10
feat: miner: defensive check for equivocation

arajasek:
## Related Issues

@Stebalien was concerned that this could cause miners to fail to produce blocks when if they see a reorg of height > 1 partway through the mining loop. The issue is that if the tipset at the height of the base's parent changes, the intersection between old and new base MUST be null, and the miner will fail.

## Proposed Changes

Weaken the check against equivocation a little -- if the intersection is null, just use the old base (even though it risks "allowing" some equivocation.).

## Checklist

- [ ] CI is green

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -594,7 +594,7 @@ func (m *Miner) mineOne(ctx context.Context, base *MiningBase) (minedBlock *type
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


Stebalien:
I think we need to go with this for now, but we'll need to fix it properly later.

Stebalien (approved):

Issue #11325, opened 2023-10-10
Lack of transaction receipts (eth_getTransactionReceipt) for some of  FEVM transactions


MarcinZakrzewski13:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I did not make any code changes to lotus.

### Lotus component

- [X] lotus Ethereum RPC
- [ ] lotus FVM - Lotus FVM interactions
- [ ] FEVM tooling
- [ ] Other

### Lotus Version

```text
Daemon:  1.23.3+mainnet+git.7bb1f98ac+api1.5.0
Local: lotus version 1.23.3+mainnet+git.7bb1f98ac
```

### Repro Steps

Request
```
curl --request POST --url https://node-URI --header 'Content-Type: application/json' --header 'X-API-Key: api-key' --data '{"jsonrpc": "2.0","method": "eth_getTransactionReceipt","params": ["0x197bb7dc4af0fa587b913827f6c626d695a5205427656e26f7637fba03f88325"],"id": 1}' 
```
Response
```
{"jsonrpc":"2.0","result":null,"id":1}
```

### Describe the Bug

We are indexing Filecoin blocks using Lotus node. We have found that for transactions in FEVM we did not get transaction receipts.

Today, we have indexed data from blocks in range from 2_683_348 to 2_846_394 searching of transactions with missing receipt.
In this range we found 53_082 transactions without receipt in 11_068 blocks.

First block with transaction which has not receipt on node: 2_683_360
Last block with transaction which has not receipt on node: 2_792_027

Is it a node issue or something we don't see?

**_Samples_**

**Transactions which have not receipt on node:**

`2683360,0x197bb7dc4af0fa587b913827f6c626d695a5205427656e26f7637fba03f88325,bafy2bzacebctgcd4fi5m23446hzhuel7mbxnotffkkufuqgpcwgbisduy3ujk`
Explorer: https://filfox.info/en/message/0x197bb7dc4af0fa587b913827f6c626d695a5205427656e26f7637fba03f88325?t=3
Node:     
```
{"jsonrpc":"2.0","result":null,"id":1}
```

[...]

PR #11329, opened 2023-10-11
fix: api: return errors on failure to lookup an eth txn receipt

Stebalien:
## Related Issues

Part of #11325.

## Proposed Changes

Return errors on failure to lookup an eth txn receipt.

All these cases here are actually errors and returning `nil` makes this hard to debug. We likely returned nil in the past to be "best effort" but, as far as I can tell, we should only hit these error cases if something is actually wrong.

## Checklist

- [x] CI is green


fridrik01 (approved):
LGTM, there is though failing tests but doesn't likely to be related

Stebalien:
Yeah, there's an issue with drand testnet right now.

PR #11353, opened 2023-10-25
feat: sealing: load SectorsSummary from sealing SectorStats instead of calling API each time

FlorianRuen:
## Related Issues
https://github.com/filecoin-project/boost/issues/1733
https://github.com/filecoin-project/boost/issues/1747
(**summary of these issues is available in `Additional Info` below**)

## Proposed Changes
Based on the discussions in the two issues mentioned above, and initial issue by @ze42 and @cryptowhizzard, I propose to retrieve the information from the stats (prometheus) rather than from the API. 

This PR was quickly discussed by @s0nik42 with @magik6k, and here is a list of changes mades:

- Create a method in `Sealing` called `SectorsSummary` in order to return the stats by states available in `SectorStats`
- Update the SectorsSummary method in StorageMinerAPI, to use the previous method. This allows to return the result more quickly (already updated with each state modification), and to avoid recalculating it from the list of sectors each time.

## Additional Info
This is a quick summary to explain the why and why of this change

1. I provide this change, because @cryptowhizzard report a boost issue when receiving 200+ online deals.
2. After some investigations, the problem is caused by the deal filter, that need to compute a lot of info = massive delays.
3. Going deeper into the deal filter code, the call that causes this delay is `SectorsSummary`, that will fill the proposal `SealingPipelineState.SectorStates`, because it iterates over all sectors to compute the states
4. This call can be very slow, espececially on 30 PB+ like @cryptowhizzard 
5. But the sector states are already available in Prometheus, each time a sector is updated
6. This change proposes to get them directly from Prometheus, in order to drastically reduce the delays

Once this is done, our wish with CIDgravity is to add information in the proposal, information available on boost directly (checkpoints, [see issue #1747 here](https://github.com/filecoin-project/boost/issues/1747))

## Checklist

- [x] CI is green

### DIFF
--- a/storage/pipeline/sealing.go
+++ b/storage/pipeline/sealing.go
@@ -304,6 +304,10 @@ func (m *Sealing) TerminateSector(ctx context.Context, sid abi.SectorNumber) err
 	return m.sectors.Send(uint64(sid), SectorTerminate{})
 }
 
+func (m *Sealing) SectorsSummary(ctx context.Context) map[SectorState]int64 {
+	return m.stats.byState
+}
+
 func (m *Sealing) TerminateFlush(ctx context.Context) (*cid.Cid, error) {
 	return m.terminator.Flush(ctx)
 }
--- a/storage/pipeline/sealing.go
+++ b/storage/pipeline/sealing.go
@@ -304,8 +304,18 @@ func (m *Sealing) TerminateSector(ctx context.Context, sid abi.SectorNumber) err
 	return m.sectors.Send(uint64(sid), SectorTerminate{})
 }
 
-func (m *Sealing) SectorsSummary(ctx context.Context) map[SectorState]int64 {
-	return m.stats.byState
+func (m *Sealing) SectorsSummary(ctx context.Context) map[api.SectorState]int {
+	m.stats.lk.Lock()
+	defer m.stats.lk.Unlock()
+
+	out := make(map[api.SectorState]int)
+
+	for st, count := range m.stats.byState {
+		state := api.SectorState(st)
+		out[state] = int(count)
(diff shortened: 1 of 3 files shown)


FlorianRuen:
It seems the failed checks are not related to the changes I made
On the master branch, using `go tests`, the test also fail with same error

storage/pipeline/sealing.go
```diff
@@ -304,6 +304,10 @@ func (m *Sealing) TerminateSector(ctx context.Context, sid abi.SectorNumber) err
 	return m.sectors.Send(uint64(sid), SectorTerminate{})
 }
 
+func (m *Sealing) SectorsSummary(ctx context.Context) map[SectorState]int64 {
+	return m.stats.byState
```
magik6k, line 308: * You have to copy this map before returning from this function, otherwise the map may be read and modified in parallel which will crash the process
  * You can move the map copy from `StorageMinerAPI.SectorsSummary` into here for that
* When accessing the `m.stats.byState` map in this function you have to hold the `m.stats.lk` lock

magik6k (commented):
Sorry for taking so long to get to this. Also thanks for the PR!

One thing to address, the current version can make lotus-miner crash occasionally.

storage/pipeline/sealing.go
FlorianRuen: @magik6k thanks for the review, just patched by moving the map copy to storage/pipeline/sealing.go and add the lock management before the copy with

```
m.stats.lk.Lock()
defer m.stats.lk.Unlock()
```

magik6k (approved):
Thanks for the PR!

PR #11358, opened 2023-10-27
Simpleharmony

snadrus:
Make the harmonytask system compatible with WinningPost. 

- Use a 3 second timeout.
- Remove "bump" and URLs as they complicate and risk causing overlapping adds
- "Follow" (task creator hook) is same-process & db.

The simpler design means that the existing, (passing) unit tests now comprehensively cover the use-cases.


journal/fsjournal/fs.go
```diff
@@ -37,7 +37,11 @@ type fsJournal struct {
 // OpenFSJournal constructs a rolling filesystem journal, with a default
 // per-file size limit of 1GiB.
 func OpenFSJournal(lr repo.LockedRepo, disabled journal.DisabledEvents) (journal.Journal, error) {
-	dir := filepath.Join(lr.Path(), "journal")
+	return OpenFSJournal2(lr.Path(), disabled)
+}
+
+func OpenFSJournal2(path string, disabled journal.DisabledEvents) (journal.Journal, error) {
```
magik6k, line 43: Could call it OpenFSJournalPath or something more descriptive than '2'

storage/wdpost/wdpost_task.go
magik6k, line 1: I moved this file to `provider/lpwindow/task.go` (mostly to keep storage/wdpost free of changes which don't need to be there + to have an easier time extracting lotus-provider work to a separate repo if that ever needs to happen)

magik6k: (so if there are any changes here, they should be merged into provider/lpwindow/task.go, and this file shouldn't be committed)

magik6k (approved):
Nothing blocking, just some things to clean up later

~/.lotus/journal/lotus-journal.ndjson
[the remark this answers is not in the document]
magik6k: Removed in https://github.com/filecoin-project/lotus/commit/f4c061bea880f6a1df93cb30e548ab3cb399865a

Issue #10547, opened 2023-03-23
Prometheus metric for pending mpool messages. 


marshyonline:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [X] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

At present, the only way to monitor stuck messages in the mpool is to use CLI-based tools.
It would be great if we could have 2 counters added to the exported metrics:

+ Total current message in the mpool
+ Total local (like the --local command outputs) messages in the mpool.

### Describe the solution you'd like

as above

### Additional context

It is common now with GAS fluctuations that's messages need to be replaced - this will aid in monitoring how much is in the pool and could be used to fire alerts in the middle of the night so someone isn't waking up to 500 stuck messages as I have recently.

rjan90:
Opened a PR for exporting the current amount of messages in the Mpool here: https://github.com/filecoin-project/lotus/pull/11361.

I will look into if there are some easy ways of exporting the local message in the mpool metric as well

PR #11361, opened 2023-10-31
feat: metric: export Mpool message count

rjan90:
## Related Issues
Partially #10547.  Also nice to have when investigating things like MpoolSelect times, syncing, etc, and its potential correlations with size of the Mpool.

## Proposed Changes
Recording the amount of pending messages in the mpool. The metric is updated after messages are removed from the pool (which happens on HeadChange or pruneMessage).

## Additional Info
Testing: in `localhost:1234/debug/metrics`, you now have:

```
# HELP lotus_mpool_message_count Number of messages in the mpool
# TYPE lotus_mpool_message_count gauge
lotus_mpool_message_count 180
```

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -1022,6 +1022,9 @@ func (mp *MessagePool) addLocked(ctx context.Context, m *types.SignedMessage, st
 		}
 	})
 
+	// Record the current size of the Mpool
+	stats.Record(ctx, metrics.MpoolMessageCount.M(int64(mp.currentSize)))
+
 	return nil
 }
 
--- a/chain/messagepool/messagepool.go
+++ b/chain/messagepool/messagepool.go
@@ -21,6 +21,7 @@ import (
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/minio/blake2b-simd"
 	"github.com/raulk/clock"
+	"go.opencensus.io/stats"
 	"golang.org/x/xerrors"
 
 	ffi "github.com/filecoin-project/filecoin-ffi"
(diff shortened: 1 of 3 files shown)


chain/messagepool/messagepool.go
```diff
@@ -1213,6 +1214,9 @@ func (mp *MessagePool) remove(ctx context.Context, from address.Address, nonce u
 			return
 		}
 	}
+
+	// Record the current size of the Mpool
+	stats.Record(ctx, metrics.MpoolMessageCount.M(int64(mp.currentSize)))
```
Stebalien, line 1219: Any reason we're not updating this on add as well?

rjan90: I was considering it, but opted out because addLocked looks like it can be called fairly frequently in a very short amount of time, and I did not want to add "unnecessary" code there. The default metric scrape interval is 10s, so having it just on remove (HeadChange or pruneMessage) seemed adequate.

Happy to add it there as well if you think its okay/more correct?

Stebalien: My concern is, e.g., if a node gets stuck (or slow) we won't get a picture of the actual number of messages and could draw the wrong conclusions.

I'm also not sure if doing it on every add is a huge issue. Especially given that we emit a journal event and and publish the change on a pubsub channel.

https://github.com/filecoin-project/lotus/blob/29fbfaf13df2fa35f84fd45dabd98b09324debfe/chain/messagepool/messagepool.go#L1012-L1022

Unless there's some hot path I'm missing here, I'd just go for accuracy.

rjan90: Agreed! Added it here: https://github.com/filecoin-project/lotus/pull/11361/commits/d121760172cbbcf9e0dd0f4ba963450a0ffc3bb1

Stebalien (approved):

Stebalien:
Could you add a quick entry to the changelog (then merge)?

PR #11363, opened 2023-10-31
feat: chain: light-weight patch to fix calibrationnet

arajasek:
## Related Issues

The calibration test network upgraded to nv21 / actors v12 2 weeks ago. Further testing on the testnet revealed a minor bug in a new feature that has since been patched: https://github.com/filecoin-project/builtin-actors/pull/1455.

Unfortunately, this is consensus-critical, and so needs a coordinated upgrade on the testnet in order to land the new code. This PR proposes a relatively easy way to achieve this.

## Proposed Changes

- Create a new `UpgradeWatermelonFixHeight` upgrade that ONLY exists for calibrationnet (disabled for all others)
- At that height, upgrade all miner actors to the new, fixed miner actor CID
- Include the patched actors bundle as the v12 bundle for calibrationnet
- Also, explicitly add the "buggy" actors bundle to v12.tar.zst
- Also, explicitly add the "buggy" miner actor info to `actorMeta`
- Tweak the "real" v12 migration (that already ran) to upgrade to the buggy bundle (this is only relevant for syncing calibnet from scratch).

## Additional Info

- Needs to be tested, we'll do that by:
  - [x] simulating this on calibnet 
  - [ ] actually running this on a butterflynet
- Could also consider:
  - Doing a "full" upgrade on calibnet with new network / actors version, etc.
  - Leaving calibnet as-is for now, it'll get patched in the next upgrade.

## Checklist

- [ ] CI is green


build/params_calibnet.go
```diff
@@ -82,6 +82,9 @@ const UpgradeThunderHeight = UpgradeLightningHeight + 3120
 // 2023-10-19T13:00:00Z
 const UpgradeWatermelonHeight = 1013134
 
+// TODO: SET THE PATCH HEIGHT EPOCH
```
rjan90, line 85: My proposal would be: epoch 1070494 - which should be 2023-11-07T13:00:00Z

chain/actors/manifest.go
```diff
@@ -51,6 +51,10 @@ func RegisterManifest(av actorstypes.Version, manifestCid cid.Cid, entries map[s
 	}
 }
 
+func AddActorMeta(name string, codeId cid.Cid, av actorstypes.Version) {
+	actorMeta[codeId] = actorEntry{name: name, version: av}
```
Stebalien, line 55: Take the lock.

chain/consensus/filcns/upgrades.go
```diff
@@ -1928,6 +1973,132 @@ func upgradeActorsV12Common(
...
+		if inActor.Address != outActor.Address && inActor.Address.String() != outActor.Address.String() {
+			return xerrors.Errorf("mismatched address for actor %s: %s != %s", a, inActor.Address, outActor.Address)
+		}
+
+		if inActor.Head != outActor.Head {
+			return xerrors.Errorf("mismatched head for actor %s", a)
+		}
+
+		// This is the hard-coded "buggy" miner actor Code ID
+		if inActor.Code != cid.MustParse("bafk2bzacecnh2ouohmonvebq7uughh4h3ppmg4cjsk74dzxlbbtlcij4xbzxq") && inActor.Code != outActor.Code {
```
arajasek, line 2081: make these constants

Stebalien (approved):

PR #11385, opened 2023-11-07
fix: eth: remove trace sanity check

Stebalien:
## Proposed Changes

We added this check as a self-test to make sure that the message indices in our trace matched up with those in the block. Unfortunately, this only works for blocks on the main-chain, and breaks tracing of uncle blocks (and tracing of head).

## Checklist

- [x] CI is green

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -842,16 +842,6 @@ func (a *EthModule) EthTraceBlock(ctx context.Context, blkNum string) ([]*ethtyp
 		return nil, xerrors.Errorf("failed when calling ExecutionTrace: %w", err)
 	}
 
-	tsParent, err := a.ChainAPI.ChainGetTipSetByHeight(ctx, ts.Height()+1, a.Chain.GetHeaviestTipSet().Key())
-	if err != nil {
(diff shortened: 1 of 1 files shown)


Stebalien:
Hm. Apparently we're hitting this check in production. I'll have to debug this.

Stebalien:
Nevermind, this code was the problem. We are attempting to lookup the _parent_ messages by looking up the "next" epoch's tipset, then looking backwards. But, if we have a null tipset, we'd look up the wrong tipset.

So, removing this code is the correct thing to do.

hawkaa (approved):
LGTM!

fridrik01:
@Stebalien So this failed when we were tracing a block which following tipset was a null tipset? Could we have a method for fetching the next non-null tipset for a given epoch (it may be useful in other cases) and then we could have sanity checks like this

Stebalien:
@fridrik01 this logic doesn't work in the presence of forks either. E.g.:

Given:

```
A <- B <- C     <-- C is head
 \
  \- X <- Y     <-- fork
```

If we try to trace `X`, we'll end up looking up `C` as the next block instead of `Y`.

fridrik01 (approved):

Issue #11355, opened 2023-10-26
Improve Eth JSON-RPC API handling of native messages


Stebalien:
E.g., code like https://github.com/filecoin-project/lotus/blob/6e22c08c1d24ebd951ddfa8b202ef4e649013583/node/impl/full/eth_utils.go#L483-L499

Right now, we:

1. Blindly convert without translating parameters, etc.
2. Discard, e.g., the message number.
3. Ignore errors converting from/to addresses.

Instead, we need to:

1. Translate parameters & return values the same way we translate them in the tracing API. I.e., we need to call these two methods: https://github.com/filecoin-project/lotus/blob/6e22c08c1d24ebd951ddfa8b202ef4e649013583/node/impl/full/eth_trace.go#L293-L353
2. Try harder to lookup the correct addresses. Although there will likely be some impossible cases.

Stebalien:
> Ignore errors converting from/to addresses.

So, an additional issue here is that we lookup addresses with an "empty tsk". That means we'll use whatever we consider to be our latest "head" instead of the state-tree after the message executed, and will fail (intermittently) on re-org.

PR #11387, opened 2023-11-07
fix: eth: use the correct state-tree when resolving addresses

Stebalien:
## Related Issues

Part of #11355

## Proposed Changes

We need to always use the state-tree from the tipset _after_ the message executed (i.e., the state-tree produced by executing the tipset). If we use any other state-tree, we might not find the address we're trying to resolve.

This change also has some implication for pending messages: there's no guarantee we'll be able to generate a 0x-style address for a pending native message. So, instead of trying, I've removed support for pending native messages from the Eth API. Messages from EthAccounts will still work, and native messages will still show up in blocks/traces, they just won't show up as "pending". Which should affect exactly nobody.

I'm also taking this opportunity to cleanup some edge-cases:

1. Pass contexts where appropriate.
2. Remove all state access from `ethTxHashFromSignedMessage`.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/types/ethtypes/eth_transactions.go
+++ b/chain/types/ethtypes/eth_transactions.go
@@ -62,9 +62,14 @@ type EthTxArgs struct {
 // - BlockHash
 // - BlockNumber
 // - TransactionIndex
-// - From
 // - Hash
 func EthTxFromSignedEthMessage(smsg *types.SignedMessage) (EthTx, error) {
+	// The from address is always an f410f address, never an ID or other address.
+	if !IsEthAddress(smsg.Message.From) {
+		return EthTx{}, xerrors.Errorf("sender must be an eth account, was %s", smsg.Message.From)
+	}
+
+	// Probably redundant, but we might as well check.
 	if smsg.Signature.Type != typescrypto.SigTypeDelegated {
 		return EthTx{}, xerrors.Errorf("signature is not delegated type, is type: %d", smsg.Signature.Type)
 	}
@@ -79,10 +84,18 @@ func EthTxFromSignedEthMessage(smsg *types.SignedMessage) (EthTx, error) {
 		return EthTx{}, xerrors.Errorf("failed to recover signature: %w", err)
 	}
 
+	from, err := EthAddressFromFilecoinAddress(smsg.Message.From)
+	if err != nil {
+		// This should be impossible as we've already asserted that we have an EthAddress
+		// sender...
+		return EthTx{}, xerrors.Errorf("sender was not an eth account")
+	}
+
 	return EthTx{
 		Nonce:                EthUint64(txArgs.Nonce),
 		ChainID:              EthUint64(txArgs.ChainID),
 		To:                   txArgs.To,
+		From:                 from,
 		Value:                EthBigInt(txArgs.Value),
 		Type:                 Eip1559TxType,
 		Gas:                  EthUint64(txArgs.GasLimit),
--- a/node/impl/full/eth_utils.go
+++ b/node/impl/full/eth_utils.go
@@ -21,6 +21,7 @@ import (
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain/actors"
+	"github.com/filecoin-project/lotus/chain/state"
 	"github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
@@ -190,7 +191,8 @@ func newEthBlockFromFilecoinTipSet(ctx context.Context, ts *types.TipSet, fullTx
 
 	bn := ethtypes.EthUint64(ts.Height())
 
-	blkCid, err := ts.Key().Cid()
+	tsk := ts.Key()
+	blkCid, err := tsk.Cid()
 	if err != nil {
 		return ethtypes.EthBlock{}, err
 	}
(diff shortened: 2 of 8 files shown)


Stebalien:
Based on https://github.com/filecoin-project/lotus/pull/11385, merge that first.

chain/types/ethtypes/eth_transactions.go
```diff
@@ -62,9 +62,14 @@ type EthTxArgs struct {
 // - BlockHash
 // - BlockNumber
 // - TransactionIndex
-// - From
```
Stebalien, line 65: I'm not sure why we weren't filling this field, but it's easy enough to do so and it saves some work elsewhere.

chain/types/ethtypes/eth_transactions.go
```diff
@@ -62,9 +62,14 @@ type EthTxArgs struct {
...
 // - Hash
 func EthTxFromSignedEthMessage(smsg *types.SignedMessage) (EthTx, error) {
+	// The from address is always an f410f address, never an ID or other address.
+	if !IsEthAddress(smsg.Message.From) {
```
Stebalien, line 68: I think, at one point, we thought we were going to support sending "eth transactions" from, e.g., ID addresses. At this point, they can only be sent from f410f addresses, so we can assert this up-front.

chain/types/ethtypes/eth_transactions.go
```diff
@@ -79,10 +84,18 @@ func EthTxFromSignedEthMessage(smsg *types.SignedMessage) (EthTx, error) {
 		return EthTx{}, xerrors.Errorf("failed to recover signature: %w", err)
 	}
 
+	from, err := EthAddressFromFilecoinAddress(smsg.Message.From)
```
Stebalien, line 87: That check above lets us convert the address here without having to potentially look it up in the state.

node/impl/full/eth_utils.go
```diff
@@ -424,21 +431,22 @@ func ethTxHashFromMessageCid(ctx context.Context, c cid.Cid, sa StateAPI) (ethty
...
 	if smsg.Signature.Type == crypto.SigTypeDelegated {
-		ethTx, err := newEthTxFromSignedMessage(ctx, smsg, sa)
+		tx, err := ethtypes.EthTxFromSignedEthMessage(smsg)
 		if err != nil {
-			return ethtypes.EmptyEthHash, err
+			return ethtypes.EthHash{}, xerrors.Errorf("failed to convert from signed message: %w", err)
 		}
-		return ethTx.Hash, nil
+
+		return tx.TxHash()
```
Stebalien, line 441: This does the same thing as the old code, this new code just makes it clear that there's no state access.

node/impl/full/eth_utils.go
```diff
@@ -576,7 +583,6 @@ func newEthTxFromMessageLookup(ctx context.Context, msgLookup *api.MsgLookup, tx
 		ti = ethtypes.EthUint64(txIdx)
 	)
 
-	tx.ChainID = ethtypes.EthUint64(build.Eip155ChainId)
```
Stebalien, line 579: This line was redundant.

node/impl/full/txhashmanager.go
```diff
@@ -84,13 +85,18 @@ func (m *EthTxHashManager) ProcessSignedMessage(ctx context.Context, msg *types.
 		return
 	}
 
-	ethTx, err := newEthTxFromSignedMessage(ctx, msg, m.StateAPI)
+	ethTx, err := ethtypes.EthTxFromSignedEthMessage(msg)
```
Stebalien, line 88: Again, I'm just removing the state access (this is always a message from an EthAccount, so we never need to look at the state to resolve addresses).

node/impl/full/eth_utils.go
```diff
@@ -453,21 +461,14 @@ func newEthTxFromSignedMessage(ctx context.Context, smsg *types.SignedMessage, s
 		if err != nil {
 			return ethtypes.EthTx{}, xerrors.Errorf("failed to calculate hash for ethTx: %w", err)
 		}
-
-		fromAddr, err := lookupEthAddress(ctx, smsg.Message.From, sa)
-		if err != nil {
-			return ethtypes.EthTx{}, xerrors.Errorf("failed to resolve Ethereum address: %w", err)
-		}
-
-		tx.From = fromAddr
```
fridrik01, line 462: This should be fine I guess since we are now setting `From` inside `ethtypes.EthTxFromSignedEthMessage`

node/impl/full/txhashmanager.go
```diff
@@ -84,13 +85,18 @@ func (m *EthTxHashManager) ProcessSignedMessage(ctx context.Context, msg *types.
...
 		return
 	}
+	txHash, err := ethTx.TxHash()
+	if err != nil {
+		log.Errorf("error hashing transaction: %s", err)
+		return
+	}
 
-	err = m.TransactionHashLookup.UpsertHash(ethTx.Hash, msg.Cid())
+	err = m.TransactionHashLookup.UpsertHash(txHash, msg.Cid())
```
fridrik01, line 99: Were we inserting the wrong hash into `eth_tx_hashes`?

fridrik01 (approved):
LGTM

node/impl/full/eth_utils.go
Stebalien: Yeah, that was the idea.

node/impl/full/txhashmanager.go
Stebalien: No, I'm just refactoring this to make it clear that it doesn't access any state.

Issue #11381, opened 2023-11-03
Lotus Provider Migrate tool


snadrus:
## User Story
The SPs want an easy way to get started using lotus-provider. Today's setup is too complicated, so a tool (safe for every configuration) will help

## Acceptance Criteria
The tool (once known about from documentation) will provide a self-explanatory experience to enable a competent SP to get started with little other information. 

```[tasklist]
### Deliverables
- [ ] Lotus-Miner CLI command that will contact the server and complete the setup.
```

## Technical Breakdown
```[tasklist]
### Development
- [ ] CLI
- [ ] Miner API
- [ ] Config creation in the database
- [ ] Convenience: The remaining setup including the exact command switches needed to run lotus-provider.
```

```[tasklist]
### Testing
- [ ] Pretend you're an beginner SP employee 
- [ ] Presume the documentation points to this tool 
- [ ] Use the tool as a first-time migrator
- [ ] Try the resulting config = following command.
```

rjan90:
Closing as completed by: https://github.com/filecoin-project/lotus/pull/11388.

The lotus-provider has a `lotus-provider config from-miner` cmd, to help aid in the migration process from Lotus-Miner --> Lotus-Provider. Completed by

PR #11388, opened 2023-11-07
cli: lotus-shed provider from-miner

snadrus:
## Related Issues
https://github.com/filecoin-project/lotus/issues/11381

## Proposed Changes
cli: lotus-shed provider from-miner

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus-provider/migrate.go
+++ b/cmd/lotus-provider/migrate.go
@@ -34,14 +34,14 @@ var configMigrateCmd = &cli.Command{
 			Usage:   fmt.Sprintf("Specify miner repo path. flag(%s) and env(LOTUS_STORAGE_PATH) are DEPRECATION, will REMOVE SOON", FlagMinerRepoDeprecation),
 		},
 		&cli.StringFlag{
-			Name:    "layer",
-			Aliases: []string{"l"},
+			Name:    "to-layer",
+			Aliases: []string{"t"},
 			Usage:   "The layer name for this data push. 'base' is recommended for single-miner setup.",
 		},
 		&cli.BoolFlag{
-			Name:    "overwrite",
-			Aliases: []string{"o"},
-			Usage:   "Use this with layer to overwrite an existing layer",
+			Name:    "replace",
+			Aliases: []string{"r"},
+			Usage:   "Use this with --to-layer to replace an existing layer",
 		},
 	},
 	Action: fromMiner,
@@ -94,7 +94,7 @@ func fromMiner(cctx *cli.Context) (err error) {
 		return fmt.Errorf("miner cannot reach the db. Ensure the config toml's HarmonyDB entry"+
 			" is setup to reach Yugabyte correctly: %s", err.Error())
 	}
-	name := cctx.String("layer")
+	name := cctx.String("to-layer")
 	if name == "" {
 		name = fmt.Sprintf("mig%d", len(titles))
 	} else {
--- a/cmd/lotus-provider/migrate.go
+++ b/cmd/lotus-provider/migrate.go
@@ -3,44 +3,48 @@ package main
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
+	"errors"
 	"fmt"
 	"os"
 	"path"
 	"strings"
 
 	"github.com/BurntSushi/toml"
+	"github.com/samber/lo"
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+
 	cliutil "github.com/filecoin-project/lotus/cli/util"
 	"github.com/filecoin-project/lotus/lib/harmony/harmonydb"
 	"github.com/filecoin-project/lotus/node/config"
+	"github.com/filecoin-project/lotus/node/modules"
 	"github.com/filecoin-project/lotus/node/repo"
-	"github.com/samber/lo"
-	"github.com/urfave/cli/v2"
 )
 
-var providerCmd = &cli.Command{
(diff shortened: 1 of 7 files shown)


cmd/lotus-shed/provider.go
```diff
@@ -0,0 +1,180 @@
...
+	defer cc()
+	addr, err := sm.ActorAddress(ctx)
+	if err != nil {
+		return fmt.Errorf("could not read actor address: %w", err)
+	}
+
+	lpCfg.Addresses.MinerAddresses = []string{addr.String()}
+
+	// TODO Apis.StorageSecret  storage repo and reading the rpc secret
+	//lr.??
```
snadrus, line 122: @magik6k I could not find this one

magik6k: This comes from the keystore, roughly `base64Enc(lr.KeyStore().Get(modules.JWTSecretName).PrivateKey)`

cmd/lotus-shed/provider.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// Populate API Key
+	_, header, err := cliutil.GetRawAPI(cctx, repo.FullNode, "v0")
+	if err != nil {
+		return fmt.Errorf("cannot read API: %w", err)
+	}
+
+	lpCfg.Apis.FULLNODE_API_INFO = []string{header.Get("Authorization")[7:]}
```
magik6k, line 133: TBH I'd rename this config from `FULLNODE_API_INFO` to something more consistent with the rest of the values. `ChainApiInfo` sounds decent to me.

cmd/lotus-shed/provider.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// Enable WindowPoSt
+	lpCfg.Subsystems.EnableWindowPost = true
+	msg += `\nBefore running lotus-provider, ensure any miner/worker answering of WindowPost is disabled.\n`
```
magik6k, line 137: Would be nice if the message said how to do that / what value to set

cmd/lotus-shed/provider.go
```diff
@@ -0,0 +1,183 @@
...
+	configTOML := &bytes.Buffer{}
+	if err = toml.NewEncoder(configTOML).Encode(lpCfg); err != nil {
+		return err
+	}
+
+	if !lo.Contains(titles, "base") {
+		_, err = db.Exec(ctx, "INSERT INTO harmony_config (title, config) VALUES ('base', '')", "base")
+		if err != nil {
+			return err
+		}
```
magik6k, line 149: Maybe this could be done as part of schema creation?

magik6k (commented):
This will make the setup so much better!

Couple of comments, otherwise good to go.

cmd/lotus-shed/provider.go
snadrus: I'm torn on the whole "base is required" thing: On one hand, it would be nice to ensure that a layer is always ran that *could* cover the basics, so I made "base" layer always run even without CLI config, and it won't start without one. On the other hand, it's another hurdle that may not mean anything. I could more easily remove the requirement if you don't think it's worthwhile. Let me know your thoughts here.

cmd/lotus-shed/provider.go
snadrus: On the miner, disable built-ins via the config with "DisableBuiltinWindowPoSt = true" and ensure no worker has windowpost enabled by either windowpost CLI arg nor LOTUS_WORKER_WINDOWPOST env, correct?

cmd/lotus-shed/provider.go
snadrus: UX: 
Guide help says use 'base'
A naive run is non-destructive by writing a new layer
Either of these populates base with something that will work. 

Ignoring the migrator & doing DiY config leaves you with a start-up error: "no base found".

The result should be a straightforward way to understand that a base layer exists OR a complete ignoring of it.

cmd/lotus-provider/migrate.go
```diff
@@ -0,0 +1,202 @@
...
+		return fmt.Errorf("miner cannot reach the db. Ensure the config toml's HarmonyDB entry"+
+			" is setup to reach Yugabyte correctly: %s", err.Error())
+	}
+	name := cctx.String("to-layer")
+	if name == "" {
+		name = fmt.Sprintf("mig%d", len(titles))
+	} else {
+		if lo.Contains(titles, name) && !cctx.Bool("overwrite") {
+			return errors.New("the overwrite flag is needed to replace existing layer: " + name)
+		}
```
magik6k, line 103: Nit: this could probably be an sql query with a where
```
err = db.Select(ctx, &titles, `SELECT title FROM harmony_config WHERE title = ? AND LENGTH(config) > 0`, name)
if err != nil {
    return fmt.Errorf("miner cannot reach the db. Ensure the config toml's HarmonyDB entry"+
        " is setup to reach Yugabyte correctly: %s", err.Error())
}

// Check if the specific title exists and handle the overwrite flag
if len(titles) > 0 && !cctx.Bool("overwrite") {
    return errors.New("the overwrite flag is needed to replace existing layer: " + name)
}
```

magik6k (approved):
1 nit, otherwise good to go

cmd/lotus-provider/migrate.go
snadrus: yes for here, but I use it later at 
if !lo.Contains(titles, "base") {

PR #11399, opened 2023-11-08
fix: nv21fix migration: correctly upgrade system actor

arajasek:
## Related Issues

The calibrationnet is currently in an inconsistent state, where there are miner actor CIDs that are not present in the system actor state. We need to backfix the `upgradeActorsV12Fix` to ALSO migrate the system actor state, this will leave us in a correct state.

## Proposed Changes

- Correctly migrate the system actor state at the v12 "fix"
- Ask users / miners to rollback history and reperform the migration.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/vm/fvm.go
+++ b/chain/vm/fvm.go
@@ -92,6 +92,15 @@ func (x *FvmExtern) VerifyConsensusFault(ctx context.Context, a, b, extra []byte
 		log.Info("invalid consensus fault: submitted blocks are the same")
 		return ret, totalGas
 	}
+
+	// workaround chain halt
+	if build.IsNearUpgrade(blockA.Height, build.UpgradeWatermelonFixHeight) {
+		return ret, totalGas
+	}
+	if build.IsNearUpgrade(blockB.Height, build.UpgradeWatermelonFixHeight) {
+		return ret, totalGas
+	}
+
 	// (1) check conditions necessary to any consensus fault
 
 	// were blocks mined by same miner?
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -324,7 +324,7 @@ minerLoop:
 					"block-time", btime, "time", build.Clock.Now(), "difference", build.Clock.Since(btime))
 			}
 
-			if os.Getenv("LOTUS_MINER_NO_SLASHFILTER") != "_yes_i_know_i_can_and_probably_will_lose_all_my_fil_and_power_" {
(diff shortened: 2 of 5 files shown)


arajasek:
TODO: 

- disable lotus migration caching
- confirm migration by comparing against "real" migrations in GST
- run on a devnet

raulk (approved):
Would be great if we could run a test in a local node against calibrationnet, but running the upgrade in a devnet would be enough. We won't get to test the rollback and redo mechanics that will occur in Calibnet, but folks have done this operation before and the code hasn't changed, so from the risk/gain perspective, doesn't make me too uncomfortable.

miner/miner.go
```diff
@@ -324,7 +324,7 @@ minerLoop:
 					"block-time", btime, "time", build.Clock.Now(), "difference", build.Clock.Since(btime))
 			}
 
-			if os.Getenv("LOTUS_MINER_NO_SLASHFILTER") != "_yes_i_know_i_can_and_probably_will_lose_all_my_fil_and_power_" {
+			if os.Getenv("LOTUS_MINER_NO_SLASHFILTER") != "_yes_i_know_i_can_and_probably_will_lose_all_my_fil_and_power_" && !build.IsNearUpgrade(base.TipSet.Height(), build.UpgradeWatermelonFixHeight) {
```
Stebalien, line 327: ditto

chain/vm/fvm.go
```diff
@@ -92,6 +92,15 @@ func (x *FvmExtern) VerifyConsensusFault(ctx context.Context, a, b, extra []byte
 		log.Info("invalid consensus fault: submitted blocks are the same")
 		return ret, totalGas
 	}
+
+	// workaround chain halt
```
LesnyRumcajs, line 96: This will also apply on other networks, not only calibnet, right?

Stebalien (commented):
Not blocking a merge.

Stebalien (approved):

LesnyRumcajs:
@arajasek will the check 
```go
		if inActor.Head != outActor.Head {
			return xerrors.Errorf("mismatched head for actor %s", a)
		}
```
still hold?

arajasek:
> @arajasek will the check
> 
> ```go
> 		if inActor.Head != outActor.Head {
> 			return xerrors.Errorf("mismatched head for actor %s", a)
> 		}
> ```
> 
> still hold?

I've added an exception for it, thank you!

PR #11401, opened 2023-11-08
fix: build: an epoch is near an upgrade iff the upgrade is enabled

Stebalien:
## Proposed Changes

Only consider an epoch near an upgrade if the upgrade is enabled (the upgrade epoch is >= 0). This isn't a huge deal, but we'd otherwise consider the first finality epochs (or so) to be "near" any disabled upgrade.

## Checklist

- [x] CI is green

### DIFF
--- a/build/isnearupgrade.go
+++ b/build/isnearupgrade.go
@@ -5,5 +5,5 @@ import (
 )
 
 func IsNearUpgrade(epoch, upgradeEpoch abi.ChainEpoch) bool {
-	return epoch > upgradeEpoch-Finality && epoch < upgradeEpoch+Finality
+	return upgradeEpoch >= 0 && epoch > upgradeEpoch-Finality && epoch < upgradeEpoch+Finality
 }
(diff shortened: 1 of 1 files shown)


build/isnearupgrade.go
```diff
@@ -5,5 +5,5 @@ import (
 )
 
 func IsNearUpgrade(epoch, upgradeEpoch abi.ChainEpoch) bool {
-	return epoch > upgradeEpoch-Finality && epoch < upgradeEpoch+Finality
+	return upgradeEpoch >= 0 && epoch > upgradeEpoch-Finality && epoch < upgradeEpoch+Finality
```
fridrik01, line 8: This is probably easier to read: 

```suggestion
	if upgradeEpoch < 0 {
	    return false
	}
	return epoch > upgradeEpoch-Finality && epoch < upgradeEpoch+Finality
```

fridrik01 (approved):

build/isnearupgrade.go
Stebalien: fair enough

PR #11419, opened 2023-11-15
Fixes for lotus-provider, and conveniences

snadrus:
## Related Issues
https://github.com/FILCAT/Space-Panthers/discussions/11

## Proposed Changes
- make install now installs lotus-provider
- configuration provides more defaults
- cli tester for windowpost

## Additional Info
For testing, these top-level commands (flags available):
lotus-provider test window-post here
    Good for: CLI testing to ensure configuration layers & basic functionality work as expected. 
    Does not post to chain. 

lotus-provider test window-post task
   Good for: E2E testing of a lotus-provider that's enabled for the window-post task. 
   Does not post to chain. 

## Checklist

- [ ] CI is green


provider/lpwindow/compute_task.go
```diff
@@ -74,10 +74,10 @@ type WdPostTask struct {
...
 
 type wdTaskIdentity struct {
-	Sp_id                uint64
-	Proving_period_start abi.ChainEpoch
-	Deadline_index       uint64
-	Partition_index      uint64
+	SpID               uint64
+	ProvingPeriodStart abi.ChainEpoch
+	DeadlineIndex      uint64
+	PartitionIndex     uint64
```
magik6k, line 80: This goes into db.Select, I'm not sure if pgx will handle this correctly (does it only care about order?).

cmd/lotus-provider/proving.go
```diff
@@ -0,0 +1,107 @@
...
+
+var provingCompute = &cli.Command{
+	Name:  "compute",
+	Usage: "Compute a proof-of-spacetime for a sector (requires the sector to be pre-sealed)",
+	Subcommands: []*cli.Command{
+		provingComputeWindowPoStCmd,
+	},
+}
+
+var provingComputeWindowPoStCmd = &cli.Command{
```
magik6k, line 36: Might be nice to have a way to submit a mock-compute task to the database, so the computation isn't done in the local cli, but on the real workers (which imo is what SPs will expect in most cases - running check tasks on the cluster with proper scheduling, but no submitting things to the chain)

magik6k (commented):
Some nits / Qs

rjan90:
It would be nice to add these to the default config, as they are needed for getting lotus-provider configured, but are currently missing from the basic config.toml:

```
[Addresses]
MinerAddresses = [ "MyMinerID"]

[Apis]
ChainApiInfo = ["myJWT"]
StorageRpcSecret = "mySecret"
```

cmd/lotus-provider/run.go
```diff
@@ -345,3 +251,144 @@ func StorageAuth(apiKey string) (sealer.StorageAuth, error) {
...
+
+	de, err := journal.ParseDisabledEvents(cfg.Journal.DisabledEvents)
+	if err != nil {
+		return nil, err
+	}
+	j, err := fsjournal.OpenFSJournalPath(cctx.String("journal"), de)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = j.Close() }()
```
magik6k, line 322: Definitely wrong, need to return a closer

cmd/lotus-provider/run.go
```diff
@@ -345,3 +251,144 @@ func StorageAuth(apiKey string) (sealer.StorageAuth, error) {
...
+
+	full, fullCloser, err := cliutil.GetFullNodeAPIV1LotusProvider(cctx, cfg.Apis.ChainApiInfo)
+	if err != nil {
+		return nil, err
+	}
+	defer fullCloser()
```
magik6k, line 328: Same here, need to defer the closer

magik6k (commented):
Some more problems here

cmd/lotus-provider/run.go
snadrus: how about managing it instead?

provider/lpwindow/compute_task.go
snadrus: It does a matching game where it replaces underscores with capitals and id --> ID.
but when in doubt, you can use struct tags:
`db:"sp_id"`

cmd/lotus-provider/proving.go
snadrus: that makes sense. Would you want it to do that with live updates, or just direct them to logs and database to evaluate the success?

magik6k: We could have either this or a separate command poll for success

cmd/lotus-provider/run.go
```diff
@@ -345,3 +253,147 @@ func StorageAuth(apiKey string) (sealer.StorageAuth, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			fullCloser()
+			_ = j.Close()
+		}
```
magik6k, line 334: this isn't ideal as it won't give us the correct shutdown order, but I don't think we care about that much in lotus-provider

node/builder_chain.go
```diff
@@ -60,7 +60,6 @@ var ChainNode = Options(
 	Override(new(dtypes.NetworkName), modules.NetworkName),
 	Override(new(modules.Genesis), modules.ErrorGenesis),
 	Override(new(dtypes.AfterGenesisSet), modules.SetGenesis),
-	Override(SetGenesisKey, modules.DoSetGenesis),
```
magik6k, line 63: This definitely needs to be kept.

provider/lpmessage/sender.go
```diff
@@ -39,16 +40,17 @@ type Sender struct {
 	api    SenderAPI
 	signer SignerAPI
 
-	db *harmonydb.DB
+	db     *harmonydb.DB
+	noSend bool // For early testing with only 1 lotus-provider
```
magik6k, line 44: I don't think we want a global noSend, imo it would be better to have a flag on PoSt tasks which says 'this is mock execution, don't send'

cmd/lotus-provider/run.go
magik6k: (just a nit, not blocking merge)

magik6k (commented):
This is objectively better, just two small things to fix

magik6k:
(also some conflicts)

snadrus:
I'm going to propose a different CLI: 

lotus-provider test window-post task
and 
lotus-provider test window-post here

That is a bit clearer on what's going on with the two test commands, and provides an avenue for future integration test behavior.

As a side-note, I also made it possible to have titled sql files since I was starting to get lost in all of them. I didn't persist the titles to the database because that seems pointless as it's a developer convenience that doesn't really have an analog at runtime.

provider/lpwindow/compute_task.go
```diff
@@ -150,7 +150,7 @@ func (t *WdPostTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done
 		return false, err
 	}
 
-	postOut, err := t.doPartition(context.Background(), ts, maddr, deadline, partIdx)
+	postOut, err := t.DoPartition(context.Background(), ts, maddr, deadline, partIdx)
```
magik6k, line 153: So below here we need to capture no_send and add it to wdpost_proofs, which will later be picked up by the submit task - which will have a different taskID than the compute task.

Alternatively we can just put a no_send on wdpost_proofs, and simply never pick that up in the submit task.

Not blocking the merge on this, but should be fixed, otherwise we'll do the send in test mode.

magik6k (approved):
1 think likely to fix, but this PR does make a lot of things better enough to merge now

PR #11421, opened 2023-11-15
fix: eth: avoid panicing on start if we don't have a head

Stebalien:
-------

## Proposed Changes

Check if we actually have a head before trying to pre-fill the tipset cache in the Eth module

This case shouldn't happen unless the node gets corrupted (I think?). Or maybe when starting at genesis? But there's no reason to panic regardless.

## Additional Info

This really isn't the right place to do this prefill, but I don't know where it should be moved. We could move it, e.g., into `ChainStore.Start`, but then we'd run this on _all_ nodes. Is that OK?

## Checklist

- [x] CI is green


Stebalien:
Testing an alternative in https://github.com/filecoin-project/lotus/pull/11426.

PR #11433, opened 2023-11-20
fix: eth: handle unresolvable addresses

Stebalien:
## Related Issues

Part of #11355.

## Proposed Changes

Correctly handle "unresolvable" to/from addresses in top-level messages in the Ethereum API. Specifically:

1. Fail if we can't resolve the from address. As far as I can tell, this should be impossible (the message statically couldn't have been included in the block if the sender didn't exist).
2. If we can't resolve the "to" address to an ID, use "max uint64" as the ID (`0xff0000000000000000000000ffffffffffffffff`). This will only happen if the transaction was reverted. It'll be a little confusing, but the alternative is to (a) use an empty address (will look like a contract creation, which is definitely wrong) or (b) use a random/hashed address which will likely be more confusing as it won't be "obviously weird".

## Additional Info

- I haven't touched the tracing API yet.
- I also refactored the tests as the single "native eth txn" test was rather long.

## Checklist

- [ ] CI is green


node/impl/full/eth_utils.go
```diff
@@ -482,19 +500,33 @@ func newEthTxFromSignedMessage(smsg *types.SignedMessage, st *state.StateTree) (
...
+	from, err := lookupEthAddress(msg.From, st)
+	if err != nil {
+		return ethtypes.EthTx{}, xerrors.Errorf("failed to lookup sender address %s when converting a native message to an eth txn: %w", msg.From, err)
+	}
+	// Lookup the to address. If the recipient doesn't exist, we replace the address with a
+	// known sentinel address.
+	to, err := lookupEthAddress(msg.To, st)
+	if err != nil {
+		if !errors.Is(err, types.ErrActorNotFound) {
+			return ethtypes.EthTx{}, xerrors.Errorf("failed to lookup receiver address %s when convertin a native message to an eth txn: %w", msg.To, err)
```
fridrik01, line 523:
```suggestion
			return ethtypes.EthTx{}, xerrors.Errorf("failed to lookup receiver address %s when converting a native message to an eth txn: %w", msg.To, err)
```

fridrik01 (approved):
LGTM

PR #11434, opened 2023-11-21
deps: libp2p:bump libp2p version to v0.30

jennijuju:
libp2p v.0.30 has been integrated in lotus and tested with v1.25.0 train for over a month.

Bringing the libp2p dependency more up to date so that the one in the mandatory release doesn't fall behind too much.


chain/badtscache.go
```diff
@@ -43,7 +43,7 @@ func (bbr BadBlockReason) String() string {
 }
 
 func NewBadBlockCache() *BadBlockCache {
-	cache, err := lru.NewARC[cid.Cid, BadBlockReason](build.BadBlockCacheSize)
+	cache, err := arc.NewARC[cid.Cid, BadBlockReason](build.BadBlockCacheSize)
```
aarshkshah1992, line 46: We don't really need this for the libp2p upgrade, right ? Do you think we can do this in a separate PR ?

chain/events/message_cache.go
```diff
@@ -14,11 +14,11 @@ type messageCache struct {
 	api EventAPI
 
 	blockMsgLk    sync.Mutex
-	blockMsgCache *lru.ARCCache[cid.Cid, *api.BlockMessages]
+	blockMsgCache *arc.ARCCache[cid.Cid, *api.BlockMessages]
 }
 
 func newMessageCache(a EventAPI) *messageCache {
-	blsMsgCache, _ := lru.NewARC[cid.Cid, *api.BlockMessages](500)
+	blsMsgCache, _ := arc.NewARC[cid.Cid, *api.BlockMessages](500)
```
aarshkshah1992, line 21: Same as above, we don't really need this as part of the libp2p upgrade.

aarshkshah1992:
Have raised https://github.com/filecoin-project/lotus/pull/11438 for just the libp2p bump to see if that fails the CI.

chain/badtscache.go
marten-seemann: Yes you need this, due to a breaking change in that library introduced in a patch release (see https://github.com/hashicorp/golang-lru/issues/142). I agree that this cherry-pick should be left as is.

Stebalien:
Ok, so, we're simply not downloading the params in those cases. So... I'm backporting a small change to the unit tests to see if it matters. Otherwise... I'm not sure.

Stebalien:
Ok, now _both_ cases should be fixed.

aarshkshah1992:
@jennijuju Can we go ahead and merge this ?

aarshkshah1992 (approved):

marten-seemann (approved):

jennijuju:
thank you all!

PR #11440, opened 2023-11-21
chore: actors: update v12 to the final release

Stebalien:
Update actors to the final v12 release. This doesn't change any of the CIDs as the code is the same.


jennijuju:
from slack
> zst is still non-deterministic for some reason.

with ^ should we expect to see a v12.zst change in diff?

Stebalien:
> with ^ should we expect to see a v12.zst change in diff?

I reverted that change because the code CIDs are the same either way.

jennijuju:
confirmed the bundle cars ✅

jennijuju (approved):

PR #11442, opened 2023-11-22
lp: task test better

snadrus:
## Related Issues
Comments-related idea to make the "task" tester request-response style. This should also enable E2E testing without log reading. 

## Checklist

- [ ] CI is green

### DIFF
--- a/lib/harmony/harmonydb/sql/20231120-testing1.sql
+++ b/lib/harmony/harmonydb/sql/20231120-testing1.sql
@@ -2,5 +2,6 @@ CREATE TABLE harmony_test (
     task_id bigint         
         constraint harmony_test_pk
             primary key,
-    options text
+    options text,
+    result text
 );
--- a/cmd/lotus-provider/proving.go
+++ b/cmd/lotus-provider/proving.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/urfave/cli/v2"
 	"golang.org/x/xerrors"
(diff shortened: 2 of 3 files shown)


provider/lpwindow/compute_task.go
```diff
@@ -161,6 +161,14 @@ func (t *WdPostTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done
...
+	testTaskID := 0
+	testTaskIDCt := 0
+	if err = t.db.QueryRow(context.Background(), `SELECT COUNT(*) FROM harmony_test WHERE task_id = $1`, taskID).Scan(&testTaskIDCt); err != nil {
+		return false, xerrors.Errorf("querying for test task: %w", err)
+	}
+	if testTaskIDCt == 1 {
+		testTaskID = int(taskID)
+	}
 	// Insert into wdpost_proofs table
 	n, err := t.db.Exec(context.Background(),
```
magik6k, line 173: We want to insert into the harmony_test table here, to avoid passing through the submit task, which won't run until the deadline open time - which can be up to a day of waiting.

It a PoSt test run SPs just want to get the compute result as soon as it's computed, and see the proof / skipped sector number. Submit task doesn't add much other than waiting for the correct time + wrapping the PoSt into a chain message, but neither of those useful to SP operators

snadrus: The Sender works for Do(), so we avoid the submit task. It also gets the nonce, which I figured may be a nice addition to show how completed the (likely inaccurate) windowpost test got.

harmony_test is both input and output to this task: 
 - input if this task is a test, in Do()
 - output the test result (below) and consider complete, in Send()

lib/harmony/harmonydb/sql/20231120-testing1.sql
```diff
@@ -2,5 +2,7 @@ CREATE TABLE harmony_test (
     task_id bigint         
         constraint harmony_test_pk
             primary key,
-    options text
-);
\ No newline at end of file
+    options text,
+    result text
```
magik6k, line 6: nit: could be jsonb for the result so that it's easier to query.

magik6k (approved):

PR #11446, opened 2023-11-22
feat: chain import: don't walk to genesis - 2-3x faster snapshot import

magik6k:
## Related Issues
This cherrypicks https://github.com/filecoin-project/lotus/pull/10821/commits/921a3a02d47abaabd97096f8647bbc3f7a07e2ae from https://github.com/filecoin-project/lotus/pull/10821

## Proposed Changes
* Snapshot cars are written from chain top to bottom
* We know what the top block header is because that's the car root
* We know that the links in that block header will lead us to the correct parent
* So when we see parents of the last seen block, we record those
* Snapshots contain all block headers to genesis, so if we repeat this process for the whole import we never have to read a single block to find the genesis block
* This way we can instantly set the correct genesis after snapshot import without having to walk there, saving roughly 17 000 000  useless reads from chainstore
  * (This is what causes the hand after chain import after the 'flushing validation cache' log)
  * For me it makes the mainnet snapshot import 2-3x faster

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/store/snapshot.go
+++ b/chain/store/snapshot.go
@@ -60,7 +60,7 @@ func (cs *ChainStore) Export(ctx context.Context, ts *types.TipSet, inclRecentRo
 	})
 }
 
-func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (*types.TipSet, error) {
+func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (head *types.TipSet, tail *types.BlockHeader, err error) {
 	// TODO: writing only to the state blockstore is incorrect.
 	//  At this time, both the state and chain blockstores are backed by the
 	//  universal store. When we physically segregate the stores, we will need
@@ -69,7 +69,7 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (*types.TipSet, e
 
 	br, err := carv2.NewBlockReader(r)
 	if err != nil {
-		return nil, xerrors.Errorf("loadcar failed: %w", err)
+		return nil, nil, xerrors.Errorf("loadcar failed: %w", err)
 	}
 
 	s := cs.StateBlockstore()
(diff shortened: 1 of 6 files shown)


chain/store/snapshot.go
```diff
@@ -60,7 +60,7 @@ func (cs *ChainStore) Export(ctx context.Context, ts *types.TipSet, inclRecentRo
 	})
 }
 
-func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (*types.TipSet, error) {
+func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (head *types.TipSet, tail *types.BlockHeader, err error) {
```
Stebalien, line 63: nit: `tail` -> `genesis`?

chain/store/snapshot.go
```diff
@@ -80,27 +80,44 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (*types.TipSet, e
 		putThrottle <- nil
 	}
 
+	nextTailCid := br.Roots[0]
```
Stebalien, line 83: Panics when empty

chain/store/snapshot.go
```diff
@@ -80,27 +80,44 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (*types.TipSet, e
...
+		}
+
+		// check for header block, looking for genesis
+		if blk.Cid() == nextTailCid && tailBlock.Height != 0 {
+			if err := tailBlock.UnmarshalCBOR(bytes.NewReader(blk.RawData())); err != nil {
+				return nil, nil, xerrors.Errorf("failed to unmarshal tail block: %w", err)
+			}
+			if len(tailBlock.Parents) > 0 {
+				nextTailCid = tailBlock.Parents[0]
+			}
```
Stebalien, line 112: else?

chain/store/snapshot.go
```diff
@@ -113,13 +130,17 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (*types.TipSet, e
 	// check errors
 	for i := 0; i < parallelPuts; i++ {
 		if lastErr := <-putThrottle; lastErr != nil {
-			return nil, lastErr
+			return nil, nil, lastErr
 		}
 	}
 
+	if tailBlock.Height != 0 {
```
Stebalien, line 137: What about the json genesis thing? (I assume you've tested this, I just wanted to make sure we're not missing something)

magik6k: The first block with the BlockHeader type is at height 0, so here we just check for that - https://filfox.info/en/block/bafy2bzacecnamqgqmifpluoeldx7zzglxcljo6oja4vrmtj7432rphldpdmm2

Stebalien (approved):

Issue #11411, opened 2023-11-12
ETH API: Remix not able to estimate gas because of missing support for block and state override params


maciejwitowski:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
-
```

### Repro Steps

1. Go to https://remix.ethereum.org/
2. Try to deploy any contract 
3. You will see this error 
![image](https://github.com/filecoin-project/lotus/assets/17002645/2f2682b6-0835-4355-80ef-72136b67bad4)

### Describe the Bug

The error comes from the fact that Remix started passing block and/or state override params to [geth API](https://github.com/ethereum/go-ethereum/blob/f265cc24b4b17b55b468b315303a3a599fc01d81/internal/ethapi/api.go#L1318) to estimate gas which ETH API in Lotus doesn't support 

### Logging Information

```text
-
```

longfeiWan9:
This also affects hardhat. It will fail to deploy and invoke smart contracts using hardhat.

cxytz01:
https://filecoinproject.slack.com/archives/C017HM9BJ8Z/p1700823496984119

I also met the bug -- when using the abigen tool to generate the Golang code, it will fail when calling a contract function.

jinqiao0423:
Please have a fix soon, this one affects a lot of features

cryshado:
> This also affects hardhat. It will fail to deploy and invoke smart contracts using hardhat.

I agree, is an actual problem on 
`"hardhat": "^2.19.1"`

maciejwitowski:
[...]

PR #11462, opened 2023-11-28
api: Add block param to eth_estimateGas

fridrik01:
## Related Issues
https://github.com/filecoin-project/lotus/issues/11411

## Proposed Changes
Geth implementation has added paramters to `eth_estimateGas` RPC endpoint and from reports from users on slack ([thread1](https://filecoinproject.slack.com/archives/C017HM9BJ8Z/p1700823496984119), [thread2](https://filecoinproject.slack.com/archives/C0370V3JTPB/p1699710397657659), [thread3](https://filecoinproject.slack.com/archives/C029MT4PQB1/p1700791457984349)) it seems that Remix has been updated to use those extra parameters, causing RPC errors in Lotus since it does not recognize these parameters. 

This PR adds support for a second  `ethtypes.EthBlockNumberOrHash` parameter  which if set will use the block specified when calculating estimateGas. Note that Geth supports an optional 3rd parameter which this PR does not implement as supporting that would require considerable effort and is left for future updates in case its needed.

## Test Plan

- I compiled lotus on calibnet, downloaded and imported calibnet snapshots
- Started lotus 
- Switched Metamask to use my local lotus node  
- Went to https://www.glif.io/ and connected my wallet, then translated my address into a "t address"
- Deposited some tFIL to my MM address using https://faucet.calibration.fildev.network/funds.html and using that "t address"
- Went to https://remix.ethereum.org/ and set Metamask as the provider
- Set a breakpoint in `EthEstimateGas` in lotus and attached my vscode to the running lotus instance
- Deployed the contract from remix
- Stepped through the `EthEstimateGas` in the debugger and observed that it went through successfully 
- Observed success from remix in deploying the contract (see screenshot):

image

- Also tested deploying a contract using the "Compile and Run script" in the "Solidity Compiler" tab in Remix UI which apparently has not been updated with the extra param as it only sends one parameter. Confirmed that it successfully deployed the contract as well.

## Checklist

- [x] CI is green


gateway/proxy_eth.go
```diff
@@ -353,13 +353,13 @@ func (gw *Node) EthMaxPriorityFeePerGas(ctx context.Context) (ethtypes.EthBigInt
 	return gw.target.EthMaxPriorityFeePerGas(ctx)
 }
 
-func (gw *Node) EthEstimateGas(ctx context.Context, tx ethtypes.EthCall) (ethtypes.EthUint64, error) {
+func (gw *Node) EthEstimateGas(ctx context.Context, tx ethtypes.EthCall, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthUint64, error) {
```
Stebalien, line 356: Does this still work if the user _doesn't_ pass `blkParam`? IIRC, this is why, e.g., methods like `EthFeeHistory` take a `jsonrpc.RawParams`. That lets one decode a variable number of params.

E.g., see https://github.com/filecoin-project/lotus/blob/3283cb70a776d5fba26658b38e3e9f000e9b6f3f/node/impl/full/eth.go#L692

https://github.com/filecoin-project/lotus/blob/3283cb70a776d5fba26658b38e3e9f000e9b6f3f/chain/types/ethtypes/eth_types.go#L803-L835

fridrik01: It requires passing `blkParam` so yeah, this will probably break other tools since now the expect two params. I will update the PR tomorrow morning to support this as optional param (works with 1 and 2 params)

fridrik01:
Moving back to draft until I make changes to support both 1 and 2 params

chain/types/ethtypes/eth_types.go
```diff
@@ -799,6 +799,45 @@ func GetContractEthAddressFromCode(sender EthAddress, salt [32]byte, initcode []
 	return ethAddr, nil
 }
 
+// EthEstimateGasParams handles raw jsonrpc params for eth_estimateGas
+type EthEstimateGasParams struct {
+	Tx       EthCall
+	BlkParam *EthBlockNumberOrHash
```
Stebalien, line 805: nit: it might be easier if we set this to "latest" by default rather than making it nullable.

fridrik01: I actually prefer to handle that logic in the eth handler

Stebalien: Fair enough.

Stebalien (approved):

Stebalien:
Can we add a test or two by specifying a block param in a normal test.

Issue #10001, opened 2023-01-13
add method support Allocation: DataCap allocated by a client to a specific piece of data.


beck-8:
As the title suggests, I didn't see this method, or I didn't find it?
If there is no support, just remind me here. Because I know that it will be no problem to support this method within 210 days after the last upgrade.

Related issues https://github.com/filecoin-project/lotus/issues/9729

PR #11466, opened 2023-11-29
feat: extend datacap expiration

swift-mx:
## Related Issues
This is Droplet extend datacap expiration function https://github.com/ipfs-force-community/droplet/blob/a9bceef4afcd3141ee929e72483ba2f812069c02/cmd/droplet-client/datacap.go#L35

### DIFF
--- a/cli/client.go
+++ b/cli/client.go
@@ -36,11 +36,14 @@ import (
 	"github.com/filecoin-project/go-fil-markets/storagemarket"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
+	builtin2 "github.com/filecoin-project/go-state-types/builtin"
+	verifregtypes "github.com/filecoin-project/go-state-types/builtin/v9/verifreg"
 
 	"github.com/filecoin-project/lotus/api"
 	lapi "github.com/filecoin-project/lotus/api"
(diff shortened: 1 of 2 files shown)


LexLuthr (changes requested):
@swift-mx Can we please increase the scope of this PR to cover all cases of the claim extension?
1. Extend all claims for a miner ID
2. Extend all claims for multiple miner IDs
3. Extend specified claims for a miner ID
4. Extend specific claims for specific miner ID

1. In case, wallet address is different from claim.Client then we should filter out such claims and in the end ask client if they wish to extend such claims as well. Extending such claims require client to send an allocation message and client will end up spending the data cap. This could be a prompt with all the info and final data cap requirements asking client to choose between "Yes" or "No". You can also add a `--force` flag to skip this check.

TermMax should default to `verifregtypes.MaximumVerifiedAllocationTerm` if not provided by the user.

LexLuthr:
Please note that we have one more PR to implement claim extension https://github.com/filecoin-project/lotus/pull/11545

I would request you and @nicelove666 to work together to avoid duplicating efforts and expanding the scope of the command as requested.

beck-8:
https://github.com/filecoin-project/lotus/issues/10001

LexLuthr:
Closing this there has been no input from author. https://github.com/filecoin-project/lotus/pull/11711 supersedes this now.

PR #11471, opened 2023-12-01
fix: Eth API: accept input data in call arguments under field 'input'.

raulk:
## Proposed Changes

The correct name for this field is 'input' according to the Ethereum specs \[0]. However, for the longest time, clients have been using 'data' and servers have been lenient to accept both, preferring 'input' over 'data' when both appear.

Our lack of support for 'input' had gone unnoticed until go-ethereum decided to adjust their ethclient implementation to issue eth_call and eth_estimateGas requests with the 'input' field instead of 'data' \[1]. This suddenly broke apps using this client against Lotus' Eth API.

Here's the fix.

\[0]: https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.yaml#L33-L35
\[1]: https://github.com/ethereum/go-ethereum/pull/28078

## Checklist

- [ ] CI is green


chain/types/ethtypes/eth_types.go
```diff
@@ -225,7 +225,8 @@ type EthCall struct {
 	Gas      EthUint64   `json:"gas"`
 	GasPrice EthBigInt   `json:"gasPrice"`
 	Value    EthBigInt   `json:"value"`
-	Data     EthBytes    `json:"data"`
+	Data     *EthBytes   `json:"data"`
+	Input    *EthBytes   `json:"input"`
 }
```
Stebalien, line 230: We should just write a custom `UnmarshalJSON` function that aliases the field on decode.

Stebalien (changes requested):

chain/types/ethtypes/eth_types.go
Stebalien: Alternative version https://github.com/filecoin-project/lotus/pull/11505

Stebalien:
Fixed in #11505.

PR #11473, opened 2023-12-03
fix: lp api handler for stop

snadrus:
## Proposed Changes
implement an API handler like Worker does, but Provider-specific. Enable "stop" call for CLI to work. 

## Additional Info

I reused the endpoint shape that Common uses, so the client uses the Common API even though other funcs available to it will not work here. 

## Checklist

- [ ] CI is green


api/permissioned.go
```diff
@@ -41,6 +41,12 @@ func PermissionedWorkerAPI(a Worker) Worker {
 	return &out
 }
 
+func PermissionedAPI[T any](a T) T {
```
magik6k, line 44: yay generics exist now; We could drop the other ones now, but that's a separate PR, not now

PR #11492, opened 2023-12-06
fix: send errs then retries despite being sent

snadrus:
## Proposed Changes
1. Fix a SQL query bug that was erroring out Send() after sending, causing weird retry. 
2. Added a backstop to stop retrying a send when the error is that it's already sent. 

## Checklist

- [ ] CI is green

### DIFF
--- a/provider/lpmessage/sender.go
+++ b/provider/lpmessage/sender.go
@@ -281,7 +281,7 @@ func NewSender(api SenderAPI, signer SignerAPI, db *harmonydb.DB) (*Sender, *Sen
 // API nodes
 //
 // Send is also currently more strict about required parameters than MpoolPushMessage
-func (s *Sender) Send(ctx context.Context, msg *types.Message, mss *api.MessageSendSpec, reason string) (cid.Cid, error) {
+func (s *Sender) Send(ctx context.Context, msg *types.Message, mss *api.MessageSendSpec, reason string, taskID harmonytask.TaskID) (cid.Cid, error) {
 	if mss == nil {
 		return cid.Undef, xerrors.Errorf("MessageSendSpec cannot be nil")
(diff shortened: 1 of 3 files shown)


provider/lpmessage/sender.go
```diff
@@ -281,7 +281,7 @@ func NewSender(api SenderAPI, signer SignerAPI, db *harmonydb.DB) (*Sender, *Sen
 // API nodes
 //
 // Send is also currently more strict about required parameters than MpoolPushMessage
-func (s *Sender) Send(ctx context.Context, msg *types.Message, mss *api.MessageSendSpec, reason string) (cid.Cid, error) {
+func (s *Sender) Send(ctx context.Context, msg *types.Message, mss *api.MessageSendSpec, reason string, taskID harmonytask.TaskID) (cid.Cid, error) {
```
magik6k, line 284: This task id is not what should go into the message_sends table

provider/lpmessage/sender.go
```diff
@@ -350,7 +350,10 @@ func (s *Sender) Send(ctx context.Context, msg *types.Message, mss *api.MessageS
 		var sigCidStr, sendError string
 		var sendSuccess *bool
 
-		err = s.db.QueryRow(ctx, `select signed_cid, send_success, send_error from message_sends where send_task_id = $1`, taskAdder).Scan(&sigCidStr, &sendSuccess, &sendError)
+		err = s.db.QueryRow(ctx, `
```
magik6k, line 353: oh, we should be persisting the taskID in the call to `taskAdder` a few lines above - this is the ID of the message send task (which is separate from the recover/submit task IDs)

magik6k (commented):
This is still incorrect

PR #11532, opened 2023-12-15
lpwinning: Don't say a task is done when persistNoWin fails

magik6k:
## Proposed Changes
This PR fixes a small bug where occasionally in epochs where we didn't win a block we would, for whatever reason, fail to persist that we didn't won on a given mining base, but still would mark the task as successfully executed.

This could lead to an edge-case where if we could win in the next epoch mining on the same base, we wouldn't try, because that could trigger a slashing condition

## Checklist

- [ ] CI is green

### DIFF
--- a/provider/lpwinning/winning_task.go
+++ b/provider/lpwinning/winning_task.go
@@ -167,7 +167,7 @@ func (t *WinPostTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
 			return false, xerrors.Errorf("persist no win: no rows updated")
 		}
 
-		return false, nil
+		return true, nil
 	}
 
 	// ensure we have a beacon entry for the epoch we're mining on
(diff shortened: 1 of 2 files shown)


provider/lpwinning/winning_task.go
```diff
@@ -156,13 +156,18 @@ func (t *WinPostTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
...
+			return false, xerrors.Errorf("marking base as not-won: %w", err)
 		}
+		log.Debugw("persisted no-win", "rows", n)
 
-		return nil
+		if n == 0 {
+			return false, xerrors.Errorf("persist no win: no rows updated")
+		}
+
+		return false, nil
```
snadrus, line 170: This func only ever returns false

snadrus (changes requested):

snadrus (approved):

Issue #11530, opened 2023-12-14
Syncer: don't fetch 900 blocks each time


arajasek:
As noted in the comment [here](https://github.com/filecoin-project/lotus/blob/07a61942d6f004460b62f36866d8b8b804a25a55/chain/sync.go#L892), we _always_ fetch 900 blocks from our peers when asked to syncFork. This is wasteful, and can cause us to fall out of sync / pressure our peers.

This isn't trivial to resolve, however. I propose the following optimizations:
- check if `incoming` is already built on a chain we have synced. If so, fetch nothing.
  - Note that this isn't easy to do, and must be reasoned about carefully. In particular, simply finding `incoming.Parents()` in your store is NOT sufficient evidence that we have the entire chain.
- fetch blocks with some exponential increase
  - a good "starting" number of blocks would be `incoming.Height() - known.Height() + 10` 
  - when those are fetched, check if we have enough to find our common ancestor (if so, we're done)
  - if not, double the amount and fetch again (max of 900)

Stebalien:
> In particular, simply finding incoming.Parents() in your store is NOT sufficient evidence that we have the entire chain.

Could you expand on this? I thought we relied on this guarantee.

> fetch blocks with some exponential increase

This sounds like the right approach. Honestly, we can probably start at 2 epochs, most forks are very short.

arajasek:
> > In particular, simply finding incoming.Parents() in your store is NOT sufficient evidence that we have the entire chain.
> 
> Could you expand on this? I thought we relied on this guarantee.

I think it's _likely_ the case that you have the entire chain, but not necessarily so. Say that you're at height 100, and were "stuck" for some reason for 10 epochs. When you return, one peer gossips the tipset at height 109, while another gossips height 110. In this case, when processing 110 as `incoming`, we WILL have its parents, but not 101-108.

> > fetch blocks with some exponential increase
> 
> This sounds like the right approach. Honestly, we can probably start at 2 epochs, most forks are very short.

I agree. I'm going to start by optimizing what appear to be nearly 100% of calls to `syncFork`, which is the 1 epoch case. In this case you actually don't need to fetch either zero blocks (if you already have all the parents, which is 90% of the time, based on analysis), or fetch one tipset (`incoming.Parents()`).

Stebalien:
[...]

PR #11533, opened 2023-12-15
feat: syncer: optimize syncFork for one-epoch forks

arajasek:
## Related Issues

Towards https://github.com/filecoin-project/lotus/issues/11530#issuecomment-1858361838

Logging and analysis have shown that nearly 100% of calls to syncFork are for a very specific case: when `incoming.Parents() ⊂ known`. Currently, we fetch 900 tipsets in this case. We _actually_ need to fetch ZERO.

## Proposed Changes

Check if there are any common blocks between `known` and `incoming.Parents()`. If there are, we simply need to return [incoming.Parents, incoming]. We _know_ we already have `incoming`, so we just need `incoming.Parents`. We now check our local store to fetch `incoming.Parents` (which according to experiments we almost always do), and if we don't we fallback onto the network requesting the ONE tipset.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/sync.go
+++ b/chain/sync.go
@@ -886,6 +886,35 @@ func (syncer *Syncer) syncFork(ctx context.Context, incoming *types.TipSet, know
 		}
 	}
 
+	incomingParentsTsk := incoming.Parents()
+	commonParent := false
+	for _, incomingParent := range incomingParentsTsk.Cids() {
+		if known.Contains(incomingParent) {
+			commonParent = true
+		}
+	}
+
+	if commonParent {
+		// known contains at least one of incoming's Parents => the common ancestor is known's Parents (incoming's Grandparents)
+		// in this case, we need to return {incoming, incoming.Parents()}
+		incomingParents, err := syncer.store.LoadTipSet(ctx, incomingParentsTsk)
+		if err != nil {
+			// fallback onto the network
+			tips, err := syncer.Exchange.GetBlocks(ctx, incoming.Parents(), 1)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


Stebalien (approved):

chain/sync.go
```diff
@@ -886,6 +886,35 @@ func (syncer *Syncer) syncFork(ctx context.Context, incoming *types.TipSet, know
...
+	incomingParentsTsk := incoming.Parents()
+	commonParent := false
+	for _, incomingParent := range incomingParentsTsk.Cids() {
+		if known.Contains(incomingParent) {
+			commonParent = true
+		}
+	}
+
+	if commonParent {
+		// known contains at least one of incoming's Parents => the common ancestor is known's Parents (incoming's Grandparents)
```
magik6k, line 898: Pretty sure this is a typo, as I parse it we only have the guarantee about common grandparents

2023-12-15-223158_1222x936_scrot


```suggestion
		// known contains at least one of incoming's Parents => the common ancestor is known's Grandparents (also incoming's Grandparents)
```

magik6k: (the logic does appear to be correct tho)

magik6k (approved):

chain/sync.go
arajasek: @magik6k No, I don't think this is right? The logic in 891-895 is checking whether `incoming.Parents()` has any blocks in common with `known` (so `known = {A,B,C,D}` in your diagram).

arajasek:
Merging cuz the only outstanding discussion is about a comment, but happy to revisit.

Issue #11457, opened 2023-11-25
Implement API for clients to subscribe to and consume built-in Actor events 


aarshkshah1992:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [X] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Once https://github.com/filecoin-project/builtin-actors/issues/1464 lands, built-in Actors will start emitting important lifecycle events that clients can subscribe to for tracking sector and deal lifecycle events.

This issue proposes implementing an API in Lotus for clients to be able to subscribe to these events.

### Describe the solution you'd like

WIP

### Describe alternatives you've considered

WIP

### Additional context

WIP

aarshkshah1992:
@Stebalien Is there an easy way to hook into message execution so we can collect and emit executed messages that match the `type MessageFilter struct {
	// Matches one of these addresses, or any address if empty.
	From   []address.Address
	// Matches one of these addresses, or any address if empty.
	To     []address.Address
	// Matches one of these methods, or any method if empty.
	Method []abi.MethodNum
}`

above ?

Stebalien:
> @Stebalien Is there an easy way to hook into message execution so we can collect and emit executed messages that match the MessageFilter filter defined in https://github.com/filecoin-project/lotus/pull/3 above ?

I believe we discussed this sync but leaving a note here:

1. We can punt this for now.
2. We'd use a lotus event observer (to watch for chain changes), same as we do when indexing events.

PR #11540, opened 2023-12-19
feat: events: Lotus API to consume events (smart contract + built-in Actor events)

aarshkshah1992:
## TODO
- [x] Rebase on feat/nv22 after the FFI/ref-fvm/actor changes have been bubbled up to the feat/nv22 branch

## Related Issues
- For https://github.com/filecoin-project/lotus/issues/11457
- FIP-discussion for built-in Actor events at https://github.com/filecoin-project/FIPs/discussions/754
- [FIP-0083](https://github.com/filecoin-project/FIPs/blob/7eb8da9f5d50f2664e059bd56c930272d6189a57/FIPS/fip-0083.md) for built-in Actor events 

## Proposed Changes
This PR is for https://github.com/filecoin-project/lotus/issues/11457.  It implements a new API in Lotus for clients to consume **_both_** smart contract and the upcoming built-in Actor events in [FIP-0083](https://github.com/filecoin-project/FIPs/blob/7eb8da9f5d50f2664e059bd56c930272d6189a57/FIPS/fip-0083.md)

Note that this is the first draft and it would be great to get community feedback on improvements we can make to the API here.

The existing ETH events API will remain as is and no changes have been made to it.

### API Definition

```
type ActorEventAPI interface {
	GetActorEvents(ctx context.Context, filter *types.ActorEventFilter) ([]*types.ActorEvent, error)
	SubscribeActorEvents(ctx context.Context, filter *types.SubActorEventFilter) (<-chan *types.ActorEvent, error)
}

type ActorEventBlock struct {
	// what value codec does client want to match on ?
	Codec uint64 `json:"codec"`
	// data associated with the "event key"
	Value []byte `json:"value"`
}

type SubActorEventFilter struct {
	Filter  ActorEventFilter `json:"filter"`
	Prefill bool             `json:"prefill"`
}

type ActorEventFilter struct {
	// Matches events from one of these actors, or any actor if empty.
	// TODO: Should we also allow Eth addresses here?
	// For now, this MUST be a Filecoin address.
	Addresses []address.Address `json:"address"`

	// Matches events with the specified key/values, or all events if empty.
	// If the `Blocks` slice is empty, matches on the key only.
	Fields map[string][]ActorEventBlock `json:"fields"`

	// Epoch based filtering ?
	// Start epoch for the filter; -1 means no minimum
	MinEpoch abi.ChainEpoch `json:"minEpoch,omitempty"`

	// End epoch for the filter; -1 means no maximum
	MaxEpoch abi.ChainEpoch `json:"maxEpoch,omitempty"`
}

type ActorEvent struct {
	Entries     []EventEntry
	EmitterAddr address.Address // f4 address of emitter
	Reverted    bool
	Height      abi.ChainEpoch
	TipSetKey   cid.Cid // tipset that contained the message
	MsgCid      cid.Cid // cid of message that produced event
}
```

There are some changes/details we need to drive consensus on which I've left as review comments on the PR.

## Checklist

- [ ] CI is green


chain/types/actor_event.go
```diff
@@ -0,0 +1,47 @@
...
+type SubActorEventFilter struct {
+	Filter  ActorEventFilter `json:"filter"`
+	Prefill bool             `json:"prefill"`
+}
+
+type ActorEventFilter struct {
+	// Matches events from one of these actors, or any actor if empty.
+	// TODO: Should we also allow Eth addresses here?
+	// For now, this MUST be a Filecoin address.
+	Addresses []address.Address `json:"address"`
```
aarshkshah1992, line 26: Should we allow clients to specify both ETH & Filecoin addresses to filter by in this new API ?

node/modules/actorevent.go
```diff
@@ -111,6 +146,7 @@ func EthEventAPI(cfg config.FevmConfig) func(helpers.MetricsCtx, repo.LockedRepo
 					return address.Undef, false
 				}
 				// we have an f4 address, make sure it's assigned by the EAM
+				// What happens when we introduce events for built-in Actor events here ?
```
aarshkshah1992, line 149: It's better to use a shared FilterManager/Indexer between the ETH events API and this new API.

However, the existing address resolver in the ETH API has this check which I believe will fail for built-in actor addresses. Can we remove this check if we're going to share this resolver b/w the two APIs or should we refactor the code so that both APIs can pass in their own address resolvers ?

node/modules/actorevent.go
```diff
@@ -64,6 +64,41 @@ func EthEventAPI(cfg config.FevmConfig) func(helpers.MetricsCtx, repo.LockedRepo
...
+
+				return nil
+			},
+		})
+
+		return ee, nil
+	}
+}
+
+func EventFilterManager(cfg config.FevmConfig) func(helpers.MetricsCtx, repo.LockedRepo, fx.Lifecycle, *store.ChainStore, *stmgr.StateManager, EventAPI, full.ChainAPI) (*filter.EventFilterManager, error) {
```
aarshkshah1992, line 98: This refactor enables the existing ETH events API & the new events API to share the filter manager/Indexer.

chain/types/actor_event.go
```diff
@@ -0,0 +1,47 @@
...
+	// If the `Blocks` slice is empty, matches on the key only.
+	Fields map[string][]ActorEventBlock `json:"fields"`
+
+	// Epoch based filtering ?
+	// Start epoch for the filter; -1 means no minimum
+	MinEpoch abi.ChainEpoch `json:"minEpoch,omitempty"`
+
+	// End epoch for the filter; -1 means no maximum
+	MaxEpoch abi.ChainEpoch `json:"maxEpoch,omitempty"`
+}
```
aarshkshah1992, line 38: I think clients can just use 0 as `MinEpoch` to specify "earliest" and -1 in both `MinEpoch` and `MaxEpoch`s to signify no-ops.

chain/types/actor_event.go
```diff
@@ -0,0 +1,47 @@
...
+type ActorEventBlock struct {
+	// what value codec does client want to match on ?
+	Codec uint64 `json:"codec"`
+	// data associated with the "event key"
+	Value []byte `json:"value"`
+}
+
+type SubActorEventFilter struct {
+	Filter  ActorEventFilter `json:"filter"`
+	Prefill bool             `json:"prefill"`
```
aarshkshah1992, line 19: If client is subscribing to events using the `SubscribeActorEvents(ctx context.Context, filter *types.SubActorEventFilter) (<-chan *types.ActorEvent, error)
}` API, they can use the `Prefill` flag here to recieve pre-existing events that match the filter before recieving new/real-time events.

api/docgen/docgen.go
```diff
@@ -406,6 +406,42 @@ func init() {
...
+		Value: []byte("data"),
+	})
+
+	addExample(&types.ActorEventFilter{
+		Addresses: []address.Address{addr},
+		Fields: map[string][]types.ActorEventBlock{
+			"abc": {
+				{
+					Codec: 0x51,
+					Value: []byte("data"),
```
rvagg, line 421: it just so happens that if you stick an extra `d` in front of `data` that you turn it into valid cbor for the string `"data"`, maybe making it valid would be a bonus here for docs?

rvagg: FYI:

```
$ cborg bin2diag ddata
64                                                # string(4)
  64617461                                        #   "data"
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"github.com/ipfs/go-cid"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+)
+
+type ActorEventBlock struct {
+	// what value codec does client want to match on ?
+	Codec uint64 `json:"codec"`
```
rvagg, line 12: How about this - we only want `Codec` to be one of two types at the moment, and I think we ought to be in control of that. We also have a type for codecs, so we could do this:

```suggestion
	"github.com/ipfs/go-cid"
	"github.com/multiformats/go-multicodec"

	"github.com/filecoin-project/go-address"
	"github.com/filecoin-project/go-state-types/abi"
)

type EventCodec multicodec.Code

// Currently supported codecs for actor event data
const (
	Raw  EventCodec = EventCodec(multicodec.Raw)
	Cbor EventCodec = EventCodec(multicodec.Cbor)
)

type ActorEventBlock struct {
	// what value codec does client want to match on ?
	Codec EventCodec `json:"codec"`
```

chain/types/actor_event.go
rvagg: wouldn't that be implied, and controlled by the min-epoch in the filter though? or do we want to be doubly sure and have a default `-f` behaviour so we're not being wasteful?

chain/types/actor_event.go
rvagg: Does this work now if you just bundle up an eth address into an `address.Address` here, or do we have something that stops that being useful later on? Since an `ethtypes.EthAddress` can also be an `address.Address`, can we just handle it all in one? Or do we have precedent for handling eth addresses differently on external APIs?

chain/types/actor_event.go
rvagg: I'd vote for having one way to do a thing. How about:

* `<-1` is an error
* `-1` means "earliest you have"
* `>=0` means from precisely that epoch _and_ if you don't have events for that epoch then you should error

Or, is there a use-case for "this epoch, but later if you don't have that one"? Unfortunately that leads to lossy information for the user: you don't get to know whether the first epoch the remote gives you is the first it _has_ or the first that there _are_ events for.

documentation/en/api-v1-unstable-methods.md
```diff
@@ -8626,6 +8685,61 @@ Response:
 }
 ```
 
+## Subscribe
+
+
+### SubscribeActorEvents
```
bajtos, line 8691: How does this JSON RPC endpoint differ from `GetActorEvents` described above? They both seem to provide the same functionality when called over JSON RPC: "Give me a list of events you know about now". I would expect `SubscribeActorEvents` to provide a long-living stream where new events are returned as they are added to the chain.

I am not familiar with ETH JSON RPC, maybe I am missing something?

documentation/en/api-v1-unstable-methods.md
```diff
@@ -3383,6 +3387,61 @@ Response:
...
+Response:
+```json
+[
+  {
+    "Entries": [
+      {
+        "Flags": 7,
+        "Key": "string value",
+        "Codec": 42,
+        "Value": "Ynl0ZSBhcnJheQ=="
```
bajtos, line 3429: It would be nice if this documentation explained how to interpret Codec and how to parse Value for common events that clients may want to observe.

For example:
- Link to relevant specs describing known events and their key/codec/value.
- Assuming this example is based on such an event, can you improve this example to link to that event docs and show what we get after parsing the value `Ynl0ZSBhcnJheQ==` using the code `42`.

documentation/en/api-v1-unstable-methods.md
```diff
@@ -3383,6 +3387,61 @@ Response:
...
+      }
+    ],
+    "EmitterAddr": "f01234",
+    "Reverted": true,
```
bajtos, line 3433: Maybe this is already documented elsewhere, but what does it mean when an event is reverted?

It would be nice to add a bit more documentation for the other event fields like Height, TipSetKey and MsgCid too. Copying the comments you have in your Go source code is a good start.

bajtos (commented):
I came here via a ping from @rjan90; thank you!

While we have a vague idea that this new API could be useful to us, I haven't figured out any details yet. It's difficult to tell whether the proposed API would work for us or not. 

Having written that, I don't see any obvious problems that would make our work harder with this PR than it would be without it. 🤷🏻

documentation/en/api-v1-unstable-methods.md
aarshkshah1992: @bajtos 

Given how Filecoin tipsets work, `reverted` basically means that the transition implied by the event no longer holds true as the tipset/block that caused the event has been reverted i.e. transactions in the block that were applied have been reverted and the state transition caused by those transactions no longer hold true.

documentation/en/api-v1-unstable-methods.md
aarshkshah1992: Hey yes,

We will document this once we have rough consensus on the API from other partners/code reviews from devs.

documentation/en/api-v1-unstable-methods.md
aarshkshah1992: @bajtos 

1. The `GetActorEvents ` is a request/response API. You send it an event filter for the events you are interested in and you get back a response containing all the events that match your filter. 

2. The `SubscribeActorEvents ` is a streaming/subscription API. You send it an event filter and it sends back a long lived stream of events. Events are written to this stream as they are emitted/reverted. You can also use this API to get back a list of "historical events that match the filter" before it starts sending real time events.

Let me know if you need more information.

bajtos: > @bajtos
> 
Thank you for the explanation, @aarshkshah1992! What you describes matches my mental model of what I would expect from these APIs.

Could you perhaps capture the information from your comment in the docs?

My confusion comes from the fact that the example response for `SubscribeActorEvents` looks exactly the same as the example response for `GetActorEvents`.

As a user of this new API, I would find helpful if the docs for `SubscribeActorEvents` provided an example of "long lived stream of events". (If this is a well-known concept, then a link to existing docs describing it is sufficient.)

I am approaching this as a newbie user with little knowledge about ETH JSON RPC API and how Filecoin/Lotus works under the hood. Maybe your work is for a different audience. In which case, feel free to ignore my suggestions.

documentation/en/api-v1-unstable-methods.md
bajtos: Makes sense, thanks for the explanation 👍🏻 

Would you know where to find more documentation about when a tipset/block can get reverted?

api/api_full.go
```diff
@@ -882,6 +882,10 @@ type FullNode interface {
 
 	RaftState(ctx context.Context) (*RaftStateData, error) //perm:read
 	RaftLeader(ctx context.Context) (peer.ID, error)       //perm:read
+
+	// Actor events
+	GetActorEvents(ctx context.Context, filter *types.ActorEventFilter) ([]*types.ActorEvent, error)               //perm:read
```
jennijuju, line 887: can we add API description in comments?

chain/types/actor_event.go
arajasek: I would only support Filecoin-formatted addresses here for now. Also parsing for EthAddresses here and converting could be nice, but:
- I suspect we'll run into unexpected edge-cases
- It strikes me as strictly an enhancement, so let's descope from this.

chain/types/actor_event.go
```diff
@@ -0,0 +1,47 @@
...
+	// Start epoch for the filter; -1 means no minimum
+	MinEpoch abi.ChainEpoch `json:"minEpoch,omitempty"`
+
+	// End epoch for the filter; -1 means no maximum
+	MaxEpoch abi.ChainEpoch `json:"maxEpoch,omitempty"`
+}
+
+type ActorEvent struct {
+	Entries     []EventEntry
+	EmitterAddr address.Address // f4 address of emitter
```
arajasek, line 42: I don't think this will always be f4 anymore.

node/impl/full/eth.go
```diff
@@ -1352,7 +1353,20 @@ func (e *EthEvent) installEthFilterSpec(ctx context.Context, filterSpec *ethtype
...
-	return e.EventFilterManager.Install(ctx, minHeight, maxHeight, tipsetCid, addresses, keys)
+	return e.EventFilterManager.Install(ctx, minHeight, maxHeight, tipsetCid, addresses, keysToKeysWithCodec(keys), true)
+}
+
+func keysToKeysWithCodec(keys map[string][][]byte) map[string][]types.ActorEventBlock {
+	keysWithCodec := make(map[string][]types.ActorEventBlock)
+	for k, v := range keys {
+		for _, vv := range v {
+			keysWithCodec[k] = append(keysWithCodec[k], types.ActorEventBlock{
+				Codec: uint64(multicodec.Raw),
```
arajasek, line 1364: Comment explaining assumption here please.

node/modules/actorevent.go
arajasek: I'd propose:
- preferring f4 addresses where possible (this is a fast check anyway)
- if no f4 address found, try to find an f1 / f3 address in the init actor (this can be left as a TODO for now, since IIUC accounts will never be emitting events)
- return ID otherwise

I think this should get what we want without changing any existing Eth behaviour, but this is more a @Stebalien department.

arajasek (commented):
Broadly looks good, some notes.

node/modules/actorevent.go
Stebalien: Honestly, I think this code is just wrong.

1. We should record the _ID_ in the index (resolve nothing). The ActorID plus the tipset is the "canonical" identifier for the actor.
2. When querying, resolve addresses to IDs, then query. As long as we limit ourselves to querying for non-reverted events, there shouldn't be any issue of ID reuse.
3. When subscribing, we'll have to resolve the filter's addresses to IDs at every tipset, but that shouldn't be a huge deal.

Of course... that'll likely require a migration. A short-term solution is to just use the `actor.Address` if specified (drop the "is delegated" check), otherwise, use the ID. That should "work"-ish, but won't be great because it, e.g., won't work for f2 addresses.

As @arajasek points out, we can punt on f1/f3 for now. However, IMO, we should punt on that forever and just "fix" this the correct way by indexing by ID.

node/config/def.go
```diff
@@ -109,6 +109,7 @@ func DefaultFullNode() *FullNode {
 		Cluster: *DefaultUserRaftConfig(),
 		Fevm: FevmConfig{
 			EnableEthRPC:                 false,
+			EnableActorEventsAPI:         false,
```
jennijuju, line 112: It seems weird for this to live under FEVM config

node/config/types.go
```diff
@@ -777,6 +777,8 @@ type FevmConfig struct {
 	// This will also enable the RealTimeFilterAPI and HistoricFilterAPI by default, but they can be disabled by config options above.
 	EnableEthRPC bool
 
+	EnableActorEventsAPI bool
```
jennijuju, line 780: add description in comments plz

node/config/def.go
aarshkshah1992: Events is a functionality offered by the FVM that built-in Actor uses. Do we differentiate between the `FVM` and the `FEVM` in Lotus ?

chain/types/actor_event.go
aarshkshah1992: Yeh let's just go ahead with Filecoin addresses for now and revisit later if support ETH addresses becomes necessary.

node/modules/actorevent.go
aarshkshah1992: Following @arajasek 's suggestion here for now and returning ID Addresses if we can't find a robust/f4 address for the given Actor ID.

chain/types/actor_event.go
aarshkshah1992: I think we want to ensure clients explicitly ask for historical events so as to not be wasteful. 

A min-epoch of -1 could still mean that "I don't want historical events but I want everything starting from now" so best to have a `Prefill` flag.

documentation/en/api-v1-unstable-methods.md
aarshkshah1992: @bajtos Yes, that's a good point. Will ping you once we solid docs available for these new APIs.

api/docgen/docgen.go
aarshkshah1992: Done.

chain/types/actor_event.go
aarshkshah1992: Removed this assumption

node/config/types.go
aarshkshah1992: Done.

chain/types/actor_event.go
rvagg: yeah, well since these are for the JSON API maybe it's not that much of a change since these will be mostly constructed by the JSON decoder :shrug: probably not a big improvement

chain/types/actor_event.go
rvagg: worth documenting this since you needed to explain it in the comments

chain/types/actor_event.go
aarshkshah1992: This makes sense to me.

aarshkshah1992: @rvagg Actually, replaced it with what we have for the existing ETH events API. Good to have this consistent across both the event APIs.

node/modules/actorevent.go
```diff
@@ -91,61 +124,57 @@ func EthEventAPI(cfg config.FevmConfig) func(helpers.MetricsCtx, repo.LockedRepo
...
 			},
 		})
 
+		return fm, nil
+	}
+}
+
+func ActorEventAPI(cfg config.FevmConfig) func(helpers.MetricsCtx, repo.LockedRepo, fx.Lifecycle, *filter.EventFilterManager, *store.ChainStore, *stmgr.StateManager, EventAPI, *messagepool.MessagePool, full.StateAPI, full.ChainAPI) (*full.ActorEvent, error) {
+	return func(mctx helpers.MetricsCtx, r repo.LockedRepo, lc fx.Lifecycle, fm *filter.EventFilterManager, cs *store.ChainStore, sm *stmgr.StateManager, evapi EventAPI, mp *messagepool.MessagePool, stateapi full.StateAPI, chainapi full.ChainAPI) (*full.ActorEvent, error) {
+		ee := &full.ActorEvent{
```
arajasek, line 166: Out of curiosity, what happens if we just return `nil, nil` (or even `nil, error`) if this method is called with `EnableActorEventsAPI = false`? 

That _seems_ more correct to me, since there's just no functionality here if `EnableActorEventsAPI` is false, and so this should never be called in the first place that way. But I'm not sure whether that breaks something?

Regardless, this seems correct as written.

api/api_full.go
```diff
@@ -885,6 +885,22 @@ type FullNode interface {
...
 	RaftLeader(ctx context.Context) (peer.ID, error)       //perm:read
+
+	// Actor events
+
+	// GetActorEvents returns all FEVM Actor events that match the given filter
+	// This is a request/response API. Please see the docs for `ActorEventFilter` for a detailed
+	// description of the filter options.
+	GetActorEvents(ctx context.Context, filter *types.ActorEventFilter) ([]*types.ActorEvent, error) //perm:read
+
+	// SubscribeActorEvents returns a long lived stream of all FEVM Actor events that match the given filter
```
rvagg, line 896:
```suggestion
	// SubscribeActorEvents returns a long-lived stream of all FEVM Actor events that match the given filter
```

api/api_full.go
```diff
@@ -885,6 +885,22 @@ type FullNode interface {
 
 	RaftState(ctx context.Context) (*RaftStateData, error) //perm:read
 	RaftLeader(ctx context.Context) (peer.ID, error)       //perm:read
+
+	// Actor events
+
+	// GetActorEvents returns all FEVM Actor events that match the given filter
+	// This is a request/response API. Please see the docs for `ActorEventFilter` for a detailed
+	// description of the filter options.
```
rvagg, line 893:
```suggestion
	// GetActorEvents returns all FEVM Actor events that match the given filter.
	// This is a request/response API. Please see the docs for `ActorEventFilter` for a detailed
	// description of the filter options.
```

api/api_full.go
```diff
@@ -885,6 +885,22 @@ type FullNode interface {
...
+	GetActorEvents(ctx context.Context, filter *types.ActorEventFilter) ([]*types.ActorEvent, error) //perm:read
+
+	// SubscribeActorEvents returns a long lived stream of all FEVM Actor events that match the given filter
+	// Events that match the given filter are written to the stream in real-time as they are emitted from the FEVM
+	// The response stream is closed when the client disconnects or if there is an error while writing an event to the stream
+	// This API also allows clients to read all historical events matching the given filter before
+	// any real-time events are written to the response stream
+	// Please see the docs for `SubActorEventFilter` for a detailed description of the filter options.
+	// NOTE: THIS API IS ONLY SUPPORTED OVER WEBSOCKETS FOR NOW
```
rvagg, line 902:
```suggestion
	// SubscribeActorEvents returns a long-lived stream of all FEVM Actor events that match the given filter.
	// Events that match the given filter are written to the stream in real-time as they are emitted from the FEVM.
	// The response stream is closed when the client disconnects or if there is an error while writing an event to the stream.
	// This API also allows clients to read all historical events matching the given filter before
	// any real-time events are written to the response stream.
	// Please see the docs for `SubActorEventFilter` for a detailed description of the filter options.
	// NOTE: THIS API IS ONLY SUPPORTED OVER WEBSOCKETS FOR NOW	
```

api/api_full.go
rvagg: @rjan90 is this an appropriate place to add "experimental"? Maybe like:

```suggestion
	// GetActorEvents returns all FEVM Actor events that match the given filter
	// This is a request/response API. Please see the docs for `ActorEventFilter` for a detailed
	// description of the filter options.
	// NOTE: This is an EXPERIMENTAL API and is subject to change in future releases.
```

api/docgen/docgen.go
```diff
@@ -414,6 +414,42 @@ func init() {
...
+
+	addExample(&types.ActorEventFilter{
+		Addresses: []address.Address{addr},
+		Fields: map[string][]types.ActorEventBlock{
+			"abc": {
+				{
+					Codec: 0x51,
+					Value: []byte("ddata"),
+				},
+			},
```
rvagg, line 431: I'm not convinced this is a great API; It feels really clunky to have to specify the codec here, which is encouraged by the clunky nature of the events themselves of course (!).

We do have the option of leaving the Codec off and just keeping it `map[string][][]byte` and matching on that, the bytes are going to remain the same, identical to the `Value` they provide here. It will lack the precision of saying "this value, which is _this codec_", but that's not too much different from just "this value".

It's not like we're doing anything fancy with the `Codec` here of course—e.g. decoding their `Value` and re-encoding it to the codec specified by the event, the `Value` is the same.

chain/types/actor_event.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"github.com/ipfs/go-cid"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+)
+
+type ActorEventBlock struct {
+	// what value codec does client want to match on ?
```
rvagg, line 11:
```suggestion
	// The value codec to match when filtering event values.
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,69 @@
...
+	Codec uint64 `json:"codec"`
+
+	// value we want to match on associated with the corresponding "event key"
+	// should be a byte array encoded with the above codec
+	// assumes base64 encoding for marshalling/um-marshalling byte arrays to json
```
rvagg, line 16:
```suggestion
	// The value to want to match on associated with the corresponding "event key"
	// when filtering events.
	// Should be a byte array encoded with the specified codec.
	// Assumes base64 encoding when converting to/from JSON strings.
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,69 @@
...
+	Value []byte `json:"value"`
+}
+
+type SubActorEventFilter struct {
+	Filter ActorEventFilter `json:"filter"`
+
+	// If true, historical events that match the given filter will be written to the response stream
```
rvagg, line 23:
```suggestion
	// If true, all available matching historical events will be written to the response stream
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,69 @@
...
+	Prefill bool `json:"prefill"`
+}
+
+type ActorEventFilter struct {
+	// Matches events from one of these actors, or any actor if empty.
+	// For now, this MUST be a Filecoin address.
+	Addresses []address.Address `json:"address"`
+
+	// Matches events with the specified key/values, or all events if empty.
+	// If the `Blocks` slice is empty, matches on the key only.
```
rvagg, line 34:
```suggestion
	// If the value is an empty slice, the filter will match on the key only.
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,69 @@
...
+	// Restricts events returned to those emitted from messages contained in this tipset.
+	// If `TipSetKey` is present in the filter criteria, then neither `FromBlock` nor `ToBlock` are allowed.
+	TipSetKey *cid.Cid `json:"tipset_cid,omitempty"`
+}
+
+type ActorEvent struct {
+	// event logs
+	Entries []EventEntry `json:"entries"`
+
+	// filecoin address of the actor that emitted this event
```
rvagg, line 54:
```suggestion
	// Filecoin address of the actor that emitted this event.
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,69 @@
...
+	EmitterAddr address.Address `json:"emitter"`
+
+	// reverted is set to true if the message that produced this event was reverted because of a network re-org
+	// in that case, the event should be considered as reverted as well
```
rvagg, line 58:
```suggestion
	// Reverted is set to true if the message that produced this event was reverted because of a network re-org
	// in that case, the event should be considered as reverted as well.
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,69 @@
...
+	Reverted bool `json:"reverted"`
+
+	// height of the tipset that contained the message that produced this event
```
rvagg, line 61:
```suggestion
	// Height of the tipset that contained the message that produced this event.
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,69 @@
...
+	Height abi.ChainEpoch `json:"height"`
+
+	// cid of the tipset that contained the message that produced this event
```
rvagg, line 64:
```suggestion
	// CID of the tipset that contained the message that produced this event.
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,69 @@
...
+	TipSetKey cid.Cid `json:"tipset_cid"`
+
+	// cid of message that produced this event
+	MsgCid cid.Cid `json:"msg_cid"`
```
rvagg, line 68: we have "Cid" for this but "Key" for the tipset, is this following existing conventions, or should we align them?

api/docgen/docgen.go
aarshkshah1992: @rvagg 

Yeah this is true. I think just a nice way for clients to ensure they're using the right codec as we use different ones for smart contract vs built-in Actor events so they're not left wondering why their filter isn't matching anything. I think best to keep it for now to make it explicit. We can remove this later once our experimental API gets some feedback.

node/modules/actorevent.go
aarshkshah1992: @arajasek Actually, this is never called if `EnableActorEventsAPI` if false

See

```
ApplyIf(isFullNode,
			If(cfg.Fevm.EnableActorEventsAPI,
				Override(new(full.ActorEventAPI), modules.ActorEventAPI(cfg.Fevm)),
			),
			If(!cfg.Fevm.EnableActorEventsAPI,
				Override(new(full.ActorEventAPI), &full.ActorEventDummy{}),
			),
		),
```

I think this is just a nice to have sanity check.

chain/types/actor_event.go
aarshkshah1992: @rvagg This looks like the convention as the usage is similar in other places in the codebase.

node/modules/actorevent.go
aarshkshah1992: @Stebalien Why wont it work for f2 addresses ?

node/modules/actorevent.go
arajasek: I agree, but in that case I think this method should just fail if called this way. But it doesn't matter too much.

node/modules/actorevent.go
Stebalien: Because we don't have a table mapping IDs to f2 addresses and maintaining one would be annoying (given reverts, etc.). So, it _can_ be done, but it's inefficient (you'd need to scan the entire address resolution table in the init actor for each lookup).

chain/types/actor_event.go
```diff
@@ -0,0 +1,70 @@
...
+	// If the value is an empty slice, the filter will match on the key only, accepting any value.
+	Fields map[string][]ActorEventBlock `json:"fields"`
+
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
+	// Optional, default: "latest".
+	FromBlock *string `json:"fromBlock,omitempty"`
+
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
+	// Optional, default: "latest".
+	ToBlock *string `json:"toBlock,omitempty"`
```
rvagg, line 44: Are we restricted with these to make optional be pointers? We have `omitempty` so plain strings should be OK for RPC. Making pointers from strings is pretty annoying if you have constant forms of them, like `"earliest"`.

```suggestion
	FromBlock string `json:"fromBlock,omitempty"`

	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
	// Optional, default: "latest".
	ToBlock string `json:"toBlock,omitempty"`
```

node/impl/full/eth.go
```diff
@@ -1262,6 +1263,59 @@ func (e *EthEvent) EthGetFilterLogs(ctx context.Context, id ethtypes.EthFilterID
...
 
+func parseBlockRange(heaviest abi.ChainEpoch, fromBlock, toBlock *string, maxRange abi.ChainEpoch) (minHeight abi.ChainEpoch, maxHeight abi.ChainEpoch, err error) {
+	if fromBlock == nil || *fromBlock == "latest" {
+		minHeight = heaviest
+	} else if *fromBlock == "earliest" {
+		minHeight = 0
+	} else {
+		if !strings.HasPrefix(*fromBlock, "0x") {
+			return 0, 0, xerrors.Errorf("FromBlock is not a hex")
+		}
```
rvagg, line 1274: This seems unnecessarily onerous to me. Maybe we could be more lax on this and just check the strings against a regex or something similar (`/(0x)?[0-9a-f/i`); then if `EthUint64FromHex` requires the `0x` we could prepend it. Or just make `EthUint64FromHex` more relaxed in the process.

Alternatively, at least we should list that it must be prefixed with `0x` in the docs.

chain/types/actor_event.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// If true, all available matching historical events will be written to the response stream
+	// before any new real-time events that match the given filter are written.
+	Prefill bool `json:"prefill"`
+}
+
+type ActorEventFilter struct {
+	// Matches events from one of these actors, or any actor if empty.
+	// For now, this MUST be a Filecoin address.
+	Addresses []address.Address `json:"address"`
```
rvagg, line 32:
```suggestion
	Addresses []address.Address `json:"addresses"`
```

chain/types/actor_event.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Matches events with the specified key/values, or all events if empty.
+	// If the value is an empty slice, the filter will match on the key only, accepting any value.
+	Fields map[string][]ActorEventBlock `json:"fields"`
+
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
+	// Optional, default: "latest".
```
rvagg, line 39: by default, the call is only "give me events from the current epoch" isn't it? I wonder if different defaults makes sense for `Get` vs `Subscribe`?

chain/types/actor_event.go
```diff
@@ -0,0 +1,70 @@
...
+	// Assumes base64 encoding when converting to/from JSON strings.
+	Value []byte `json:"value"`
+}
+
+type SubActorEventFilter struct {
+	Filter ActorEventFilter `json:"filter"`
+
+	// If true, all available matching historical events will be written to the response stream
+	// before any new real-time events that match the given filter are written.
+	Prefill bool `json:"prefill"`
```
rvagg, line 26: We already have `FromBlock` in the filter, so we have two ways to signal a similar thing. So we can have a situation like `Prefill=true` + `FromBlock="latest"`, or `Prefill=false` + `FromBlock="earliest"`. In fact, the only really meaningful value of `FromBlock` with a `Prefill=false` would be a future epoch, all other possible values yield the same result.

So I wonder if we should just remove `SubActorEventFilter` entirely and rely on `FromBlock` to do the prefilling?

chain/types/actor_event.go
```diff
@@ -0,0 +1,70 @@
...
+	// For now, this MUST be a Filecoin address.
+	Addresses []address.Address `json:"address"`
+
+	// Matches events with the specified key/values, or all events if empty.
+	// If the value is an empty slice, the filter will match on the key only, accepting any value.
+	Fields map[string][]ActorEventBlock `json:"fields"`
+
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
+	// Optional, default: "latest".
+	FromBlock *string `json:"fromBlock,omitempty"`
```
rvagg, line 40: Should we `s/Block/Epoch`? I'm not entirely clear on idiomatic nomenclature around these things yet but we seem to be mixing two ideas.

chain/types/actor_event.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
+	// Optional, default: "latest".
+	ToBlock *string `json:"toBlock,omitempty"`
+
+	// Restricts events returned to those emitted from messages contained in this tipset.
+	// If `TipSetKey` is present in the filter criteria, then neither `FromBlock` nor `ToBlock` are allowed.
+	TipSetKey *cid.Cid `json:"tipset_cid,omitempty"`
```
rvagg, line 48:
```suggestion
	TipSetCid *cid.Cid `json:"tipsetCid,omitempty"`
```

Since `TipSetKey` is already a concrete type, we'd better be clear with what this is. Also matching casing style.

node/modules/actorevent.go
aarshkshah1992: Punting this for now in the first draft.

chain/types/actor_event.go
aarshkshah1992: What defaults do you have in mind ? My suggestion would be defer to actual user feedback here and make changes in a subsequent release once people have tried this out. F

or now, I've just picked this up from the existing ETH events API which has been running the wild for some time.

chain/types/actor_event.go
aarshkshah1992: Hmm, again, I've just borrowed the convention of the existing ETH events API. Happy to change it if you think it'll make things simple here.

api/api_full.go
```diff
@@ -885,6 +885,22 @@ type FullNode interface {
 
 	RaftState(ctx context.Context) (*RaftStateData, error) //perm:read
 	RaftLeader(ctx context.Context) (peer.ID, error)       //perm:read
+
+	// Actor events
+
+	// GetActorEvents returns all FEVM Actor events that match the given filter.
+	// This is a request/response API. Please see the docs for `ActorEventFilter` for a detailed
```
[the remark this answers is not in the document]
aarshkshah1992: I see above that by convention, we don't document arguments. I am removing that line for now. I think a comprehensive doc here would belong to the integration doc that we're putting together for this.

api/api_full.go
```diff
@@ -885,6 +885,22 @@ type FullNode interface {
...
+	// description of the filter options.
+	GetActorEvents(ctx context.Context, filter *types.ActorEventFilter) ([]*types.ActorEvent, error) //perm:read
+
+	// SubscribeActorEvents returns a long-lived stream of all FEVM Actor events that match the given filter.
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

node/impl/full/eth.go
aarshkshah1992: I changed it so that atleast for the Actor Events API, we prepend this if it does not exist. Don't want to touch the ETH events API here.

chain/types/actor_event.go
aarshkshah1992: Yeah renamed these to use `Epoch`.

chain/types/actor_event.go
aarshkshah1992: I agree, I've removed the `SubActorEventFilter `. We will now Prefill if FromEpoch < Current Epoch.

node/impl/full/eth.go
aarshkshah1992: Got rid of all this by changing the `FromEpoch` and `ToEpoch` types in the filter to `abi.ChainEpoch`.

chain/types/actor_event.go
aarshkshah1992: @rvagg After making this change and writing a test to see how it feels, I think it's best if we keep this explicit for the user so they know what they're asking for.

I am adding some docs to point out the edge case you mention here.

node/impl/full/eth.go
aarshkshah1992: Addressed this by not making "0x" mandatory in the input. The docs do mention that these must be hex encoded.

aarshkshah1992:
itests will be green after we merge the FFI/ref-fvm deps changes for NV22 to this branch.

chain/types/actor_event.go
rvagg: I'm mainly wondering, from a user perspective, what the expectation with a call like `GetActorEvents` is? The current implementation just gives them any for this current epoch. Maybe that's reasonable and more than that would be way too much. I'm tempted to think that users will come to this imagining that the default filter is "just give me everything!". But in fact it's actually "just give me everything .. in the current epoch". For `Subscribe` I always think of those kinds of APIs in terms of `tail -f`, where fetching historical stuff _is_ something you go out of your way to do. Then we have the doubling up of `Prefill` for `Subscribe`, so even if we set this default to `"earliest"` for both and `Prefill` remains `false` then you'd get the `tail -f` behaviour for default `Subscribe` and the maximal output with default `Get`.

My analogies are something like:

```
# "Get" system logs - all the things!
cat /var/log/syslog
# Follow / "Subscribe" system logs - only the new stuff
tail -f /var/log/syslog
# "Get" system logs - with filters applied
cat /var/log/syslog | grep '^boop:\|ERROR'
```

Not a big deal, we can always change it later. Although it'd be quite a surprise change in default to go from minimal to maximal.

Stebalien:
> itests will be green after we merge the FFI/ref-fvm deps changes for NV22 to this branch.

Merged.

rvagg:
I've merged feat/nv22 back in here, but it also needs a newer builtin-actors. Unfortunately, the v13.0.0-rc.1 already has the optional AggregateProofType change which requires the additional fixes in go-state-types and back here, so it has to be [v13.0.0-dev.0](https://github.com/filecoin-project/builtin-actors/releases/tag/v13.0.0-dev.0); which can be installed with:

```
(cd build/actors/; ./pack.sh v13 v13.0.0-dev.0)
```

Then this works as is.

rvagg:
@aarshkshah1992 latest commit from me is a cleanup and doc of the test. I've tried to structure it a bit better to be more easily understood since there's so many things going on in there.

I got to simplify a bunch of things while in there, including that height problem! Switching to using `ChainGetMessagesInTipset` and then following that up with a general `StateSearchMsg` (can't use the actual tipset for that call, has to be `types.EmptyTSK`, I guess because receipts come after the tipset..?). Then you end up with exactly the same events as the `GetActorEvents` produces and a simple `require.Equal(t, expected, actual)` works for comparing.

I think I also made the whole allocation->claim testing thing a bit clearer and actually tested the claim is what we expect. Which then verifies that data collection for the DDO flow is possible with this workflow.

rvagg:
Some notes related to the differences to the Eth events API for this one, for anyone reviewing this who may think this is important:

* `FromEpoch` instead of `FromBlock`, `ToEpoch` instead of `ToBlock` - we're still using hex to communicate the epoch but we take it as a string so we can also do `"earliest"` and `"latest"`.
  - The hex is still something we could revisit, is there a good reason to use hex for this when it's just an epoch number? The returned `ActorEvent` object has a plain epoch integer in it, maybe we should be mirroring that and accepting a plain number (as a string) instead of a hex?
  - It's named `height` in `ActorEvent` returned by you use `FromEpoch` and `ToEpoch`, should we make the terms consistent?
* `FromEpoch` and `ToEpoch` are `string` with `json:omitempty`, in the Eth API they are `*string` with `json:omitempty`. Does this matter? Empty string signals default and you don't have to mess with pointers to literals. For RPC it should be essentially the same I believe.
* There's no direct equivalent to `EthLog` in here, although `ActorEvent` as some of the information; you're stuck with a raw `EventEntry` slice to decode yourself. We _could_ do something like "topics" in `EthLog` and extract the `$type` up to the top level. But maybe that's something we can do later.

rjan90:
Given that the TODO of `Rebase on feat/nv22 after the FFI/ref-fvm/actor changes have been bubbled up to the feat/nv22` branch now is done, I think this PR now is ready for merge @aarshkshah1992 ?

rvagg:
I'm fine with this getting merged, probably needs to be squashed up because it's a bit of a mess.

I have some additional proposals for the API but they can be done separately.

rvagg:
https://github.com/filecoin-project/lotus/pull/11618 has all of this and is condensed and is receiving updates. So moving discussion to there.

PR #11546, opened 2023-12-21
fix: potential null pointer error

60ke:
if  cs.heaviest is nil cause panic  ,when call chain/stmgr/execute.go:TipSetState -> tryLookupTipsetState -> GetTipsetByHeight

### DIFF
--- a/chain/store/store.go
+++ b/chain/store/store.go
@@ -1248,6 +1248,9 @@ func (cs *ChainStore) GetTipsetByHeight(ctx context.Context, h abi.ChainEpoch, t
 
 	if ts == nil {
 		ts = cs.GetHeaviestTipSet()
+		if ts == nil {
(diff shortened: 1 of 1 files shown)


magik6k:
How are you able to get chainstore into a state where the heaviest tipset is nil? it's loaded on startup, so it should be impossible for it to be nil.

Stebalien:
> How are you able to get chainstore into a state where the heaviest tipset is nil? it's loaded on startup, so it should be impossible for it to be nil.

It's possible in very special cases at startup. I can't remember how, exactly. But I've seen it happen.

60ke:
> How are you able to get chainstore into a state where the heaviest tipset is nil? it's loaded on startup, so it should be impossible for it to be nil.

It does not appear in lotus；But my project relies on some api of lotus，It appeared when I called it.

PR #11565, opened 2024-01-10
feat: exchange: change GetBlocks to always fetch the requested number of tipsets

### LINKED ISSUE #11529: Syncer: Handle partial responses correctly
The syncer relies on the chain exchanger to fetch blocks it needs for syncing. The chain exchanger, which fetches blocks from peers, can receive either full or partial responses. AFAIC tell, we always treat partial responses as "successful", returning them to the syncer.

This can lead to undefined / unexpected behaviour. We need to audit calls to the `GetBlocks` method, and reason about partial responses. As a concrete first step, the call to `GetBlocks` in `syncFork` ([here](https://github.com/filecoin-project/lotus/blob/07a61942d6f004460b62f36866d8b8b804a25a55/chain/sync.go#L892)) doesn't check for partial responses -- it should, and fetch more blocks assuming it needs them.

arajasek:
## Related Issues

Fixes #11529 

## Proposed Changes

Change `GetBlocks` to always return `count` tipsets, unless we hit genesis. Partial responses are treated as correct, but the exchanger will still request more tipsets (instead of returning fewer than requested tipsets to the caller).

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/exchange/client.go
+++ b/chain/exchange/client.go
@@ -284,16 +284,18 @@ func (c *client) validateCompressedIndices(chain []*BSTipSet) error {
 				len(msgs.SecpkIncludes), blocksNum)
 		}
 
+		blsLen := uint64(len(msgs.Bls))
+		secpLen := uint64(len(msgs.Secpk))
 		for blockIdx := 0; blockIdx < blocksNum; blockIdx++ {
 			for _, mi := range msgs.BlsIncludes[blockIdx] {
-				if int(mi) >= len(msgs.Bls) {
+				if mi >= blsLen {
 					return xerrors.Errorf("index in BlsIncludes (%d) exceeds number of messages (%d)",
 						mi, len(msgs.Bls))
 				}
 			}
 
 			for _, mi := range msgs.SecpkIncludes[blockIdx] {
-				if int(mi) >= len(msgs.Secpk) {
+				if mi >= secpLen {
 					return xerrors.Errorf("index in SecpkIncludes (%d) exceeds number of messages (%d)",
 						mi, len(msgs.Secpk))
 				}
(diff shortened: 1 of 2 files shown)


chain/exchange/client.go
```diff
@@ -341,12 +361,16 @@ func (c *client) GetFullTipSet(ctx context.Context, peer peer.ID, tsk types.TipS
...
+	}
+
+	fullTipsets := validRes.toFullTipSets()
+
+	if len(fullTipsets) == 0 {
+		return nil, xerrors.New("unexpectedly got no tipsets in exchange")
 	}
 
-	return validRes.toFullTipSets()[0], nil
-	// If `doRequest` didn't fail we are guaranteed to have at least
```
arajasek, line 348: This was true, but I still see no reason not to check this.

chain/exchange/interfaces.go
```diff
@@ -28,8 +28,8 @@ type Server interface {
 // used by the Syncer.
 type Client interface {
 	// GetBlocks fetches block headers from the network, from the provided
-	// tipset *backwards*, returning as many tipsets as the count parameter,
-	// or less.
+	// tipset *backwards*, returning as many tipsets as the count parameter.
+	// The ONLY case in which we return fewer than `count` tipsets is if we hit genesis.
 	GetBlocks(ctx context.Context, tsk types.TipSetKey, count int) ([]*types.TipSet, error)
 
 	// GetChainMessages fetches messages from the network, starting from the first provided tipset
```
arajasek, line 35: Note that we are NOT changing `GetChainMessages` behaviour -- it might still return fewer than `count` epochs worth of messages.

PR #11572, opened 2024-01-12
feat: implement FIP-0063

arajasek:
## Related Issues

Implements FIP-0063.

## Proposed Changes

- Introduce the new entry to the Drand Schedule, switching to Quicknet at PineappleHeight + 1
- Refactor `VerifyEntry` to receive the previous entry's signature only, not the `Entry` struct. This signature field is nil for Quicknet entries.
- Bumping drand deps to pull in Quicknet logic

## Checklist

- [ ] CI is green


build/drand.go
```diff
@@ -27,15 +27,18 @@ const (
 	DrandDevnet
 	DrandLocalnet
 	DrandIncentinet
+	DrandQuicknet
 )
 
 var DrandConfigs = map[DrandEnum]dtypes.DrandConfig{
 	DrandMainnet: {
+		//FUCK: add storswift endpoint
```
arajasek, line 35: @cluelessuk can we do this now?

chain/beacon/beacon.go
```diff
@@ -43,31 +45,31 @@ type BeaconPoint struct {
...
-			}
-			err := currBeacon.VerifyEntry(h.BeaconEntries[1], h.BeaconEntries[0])
-			if err != nil {
-				return xerrors.Errorf("beacon at fork point invalid: (%v, %v): %w",
-					h.BeaconEntries[1], h.BeaconEntries[0], err)
-			}
-			return nil
-		}
-	}
+	//{
```
arajasek, line 54: I think we can drop all of this code? It was intended for a beacon upgrade, but I don't think we want to do it this way.

chain/beacon/drand/drand.go
```diff
@@ -164,33 +174,28 @@ func (db *DrandBeacon) getCachedValue(round uint64) *types.BeaconEntry {
...
 
-func (db *DrandBeacon) VerifyEntry(curr types.BeaconEntry, prev types.BeaconEntry) error {
-	if prev.Round == 0 {
-		// TODO handle genesis better
-		return nil
-	}
-
-	if curr.Round != prev.Round+1 {
-		return xerrors.Errorf("invalid beacon entry: cur (%d) != prev (%d) + 1", curr.Round, prev.Round)
-	}
```
arajasek, line 175: It's safe to drop this for both "current" Drand and quicknet, right?

chain/beacon/drand/drand.go
```diff
@@ -218,6 +223,10 @@ func (db *DrandBeacon) maxBeaconRoundV2(latestTs uint64) uint64 {
 	// we take the time from genesis divided by the periods in seconds, that
 	// gives us the number of periods since genesis.  We also add +1 because
 	// round 1 starts at genesis time.
+	// TODO: Confirm that we are still fetching the right round for Filecoin
```
arajasek, line 226: Gonna experimentally confirm this, but special review here is requested.

build/drand.go
AnomalRoil: No reason not to, afaik

CluEleSsUK: yep! it's up an working

build/params_mainnet.go
```diff
@@ -15,9 +15,11 @@ import (
 	builtin2 "github.com/filecoin-project/specs-actors/v2/actors/builtin"
 )
 
+// FUCK: add a block to update to the new network
```
CluEleSsUK, line 18:
```suggestion
```

chain/beacon/drand/drand.go
CluEleSsUK: I think so - it doesn't seems like in the current implementation the previous beacon should be cached, so there's no way for someone to e.g. provide a historical beacon that gets validated, though might be worth confirming

CluEleSsUK: hmm well actually we only check the `MaxBeaconRoundForEpoch` - perhaps it _is_ possible to provide a historical beacon without a check of the round?

build/drand.go
```diff
@@ -44,6 +46,21 @@ var DrandConfigs = map[DrandEnum]dtypes.DrandConfig{
 		},
 		ChainInfoJSON: `{"public_key":"868f005eb8e6e4ca0a47c8a77ceaa5309a47978a7c71bc5cce96366b5d7a569937c529eeda66c7293784a9402801af31","period":30,"genesis_time":1595431050,"hash":"8990e7a9aaed2ffed73dbd7092123d6f289930540d7651336225dc172e51b2ce","groupHash":"176f93498eac9ca337150b46d21dd58673ea4e3581185f869672e59fa4cb390a"}`,
 	},
+	DrandQuicknet: {
+		Servers: []string{
+			"https://api.drand.sh",
```
Kubuxu, line 51: How are the same servers used for different drand networks?

chain/beacon/beacon.go
```diff
@@ -25,7 +25,9 @@ type Schedule []BeaconPoint
 func (bs Schedule) BeaconForEpoch(e abi.ChainEpoch) RandomBeacon {
 	for i := len(bs) - 1; i >= 0; i-- {
 		bp := bs[i]
-		if e >= bp.Start {
+		// The new network version kicks in at one _after_ the upgrade height itself,
+		// so we switch to the new beacon when _strictly_ greater than bp.Start
+		if e > bp.Start {
```
Kubuxu, line 30: Won't this break existing migration point?

chain/beacon/beacon.go
Kubuxu: We cannot as the existing chain will fail to verify at the migration point.

chain/beacon/drand/drand.go
Kubuxu: I think it is safe although it

chain/sync.go
```diff
@@ -704,25 +704,25 @@ func (syncer *Syncer) collectHeaders(ctx context.Context, incoming *types.TipSet
...
-		// ensure consistency of beacon entires
+		// ensure consistency of beacon entries
 		targetBE := incoming.Blocks()[0].BeaconEntries
 		sorted := sort.SliceIsSorted(targetBE, func(i, j int) bool {
 			return targetBE[i].Round < targetBE[j].Round
 		})
 		if !sorted {
-			syncer.bad.Add(incoming.Cids()[0], NewBadBlockReason(incoming.Cids(), "wrong order of beacon entires"))
-			return nil, xerrors.Errorf("wrong order of beacon entires")
+			syncer.bad.Add(incoming.Cids()[0], NewBadBlockReason(incoming.Cids(), "wrong order of beacon entries"))
```
Kubuxu, line 713: I recognise these typos :D

chain/beacon/beacon.go
arajasek: Oh, you're right, I forgot that we upgraded from incentinet to mainnet. I'll see how best to fix that up.

build/drand.go
arajasek: I assume it has to do with the info in `ChainInfoJSON`, but I'll let @CluEleSsUK @AnomalRoil confirm that.

chain/beacon/drand/drand.go
Stebalien: This _should_ give us the drand beacon generated when `latestTs` was produced, so yes?

Stebalien: Are we sure we want "at least 30 seconds"? We didn't actually have the option of forcing propagation delays before... (probably too late to bikeshed this now).

chain/beacon/drand/drand.go
```diff
@@ -68,6 +69,10 @@ func (l *logger) Named(s string) dlog.Logger {
 	return &logger{l.SugaredLogger.Named(s)}
 }
 
+func (l *logger) AddCallerSkip(skip int) dlog.Logger {
```
Stebalien, line 72: why?

build/drand.go
CluEleSsUK: Precisely

chain/beacon/drand/drand.go
Kubuxu: I think drand requires it as part of their logger interface.

build/drand.go
```diff
@@ -44,6 +46,21 @@ var DrandConfigs = map[DrandEnum]dtypes.DrandConfig{
...
+			"https://api2.drand.sh",
+			"https://api3.drand.sh",
+			"https://drand.cloudflare.com",
+			"https://api.drand.secureweb3.com:6875", // Storswift
```
Kubuxu, line 55: Is this endpoint monitored by any of the drand monitoring infra? We were burned in the past by 3rd party bootstrappers going offline.

chain/beacon/drand/drand.go
Stebalien:
```
	return &logger{l.SugaredLogger.With(zap.AddCallerSkip(skip))}
```

I guess my point was, why are we stubbing this.

chain/beacon/beacon.go
Kubuxu: As long as we switch the same in miner, no issue. It's mostly just code/system smell.

chain/beacon/beacon.go
arajasek: Why would it?

chain/beacon/drand/drand.go
arajasek: I asked Protocol Design gang about this, and they basically said "leave at 30 for now, change later maybe" (which is kinda how I feel about it too).

arajasek: Confirmed this manually, dropping the TODO.

build/drand.go
CluEleSsUK: the endpoint, no, though we get some alerting if their node goes down (which is normally a good indicator).
The client uses all the addresses as fallback so it shouldn’t cause an outage though - what was the issue in the past?

chain/beacon/drand/drand.go
arajasek: Sorry, I think this was a WIP thing from the Drand folks.

chain/beacon/beacon.go
arajasek: Ah, dammit, I keep forgetting we already had a drand fork on mainnet lol.

build/drand.go
arajasek: @CluEleSsUK @AnomalRoil ?

chain/beacon/beacon.go
arajasek: I like that, I'm actually inclined to switch _later_ than the upgrade height (get your migration and other nonsense done first, then we switch beacons).

Stebalien: I'm just thinking we may want to ship two upgrades at the same time instead of trying to bundle them into one.

build/drand.go
Kubuxu: The general issue in the past was that we had a significant number of third-party bootstrappers (let's say 70% of all), and the majority of them were down.

It didn't cause an outage per se, but it slowed down bootstraping.

Stebalien (approved):

chain/beacon/beacon.go
```diff
@@ -43,20 +43,21 @@ type BeaconPoint struct {
...
-	VerifyEntry(types.BeaconEntry, types.BeaconEntry) error
+	VerifyEntry(entry types.BeaconEntry, prevEntrySig []byte) error
 	MaxBeaconRoundForEpoch(network.Version, abi.ChainEpoch) uint64
 }
 
 func ValidateBlockValues(bSchedule Schedule, nv network.Version, h *types.BlockHeader, parentEpoch abi.ChainEpoch,
 	prevEntry types.BeaconEntry) error {
-	{
+	// Before nv22 we had "chained" beacons, and so required two entries at a fork
+	if nv < network.Version22 {
```
hanabi1224, line 53: If drand network upgrade does not align with chain network upgrade, should we check the exact drand network upgrade epoch here?

chain/beacon/beacon.go
```diff
@@ -79,7 +80,12 @@ func ValidateBlockValues(bSchedule Schedule, nv network.Version, h *types.BlockH
 		return xerrors.Errorf("expected to have beacon entries in this block, but didn't find any")
 	}
 
-	// Verify that the last beacon entry's round corresponds to the round we expect
+	if nv < network.Version22 && prevEntry.Round == 0 {
```
hanabi1224, line 83: Ditto, should we decouple drand network version and chain network version here?

chain/beacon/beacon.go
Kubuxu: This is a fork logic which works off the DrandSchedule. So, the meaning of this is to disable the old fork logic before switching to the new Drand network.

build/params_butterfly.go
```diff
@@ -16,7 +16,8 @@ import (
 )
 
 var DrandSchedule = map[abi.ChainEpoch]DrandEnum{
-	0: DrandMainnet,
+	0:                           DrandMainnet,
+	UpgradePineappleHeight + 10: DrandQuicknet,
```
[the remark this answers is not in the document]
arajasek: Yes, certainly.

chain/beacon/beacon.go
arajasek: Similar to above, I don't think so, no. From nv22 onwards we'll be using "unchained" randomness, and so can simply verify round 0 too.

Stebalien (approved):

PR #11579, opened 2024-01-15
fix: docker: Dockerfile import snapshot

vidbregar:
## Proposed Changes
- Updates Dockerfile to default to a working snapshot endpoint
  ```bash
  curl -I https://snapshots.mainnet.filops.net/minimal/latest
  HTTP/1.1 503 Service Unavailable: Back-end server is at capacity
  Connection: keep-alive
  ```
  ```bash
    curl -I https://forest-archive.chainsafe.dev/latest/mainnet/
    HTTP/2 302
    ...
  ```
- Since 1.25.2, entrypoint script requires `yes` to import the snapshot

## Checklist

- [ ] CI is green

### DIFF
--- a/scripts/docker-lotus-entrypoint.sh
+++ b/scripts/docker-lotus-entrypoint.sh
@@ -1,18 +1,21 @@
 #!/usr/bin/env bash
 
-if [ ! -z $DOCKER_LOTUS_IMPORT_SNAPSHOT ]; then
+if [ ! -z "$DOCKER_LOTUS_IMPORT_SNAPSHOT" ]; then
 	GATE="$LOTUS_PATH"/date_initialized
 	# Don't init if already initialized.
 	if [ ! -f "$GATE" ]; then
 		echo importing minimal snapshot
-		/usr/local/bin/lotus daemon --import-snapshot "$DOCKER_LOTUS_IMPORT_SNAPSHOT" --halt-after-import
+		/usr/local/bin/lotus daemon \
+			--import-snapshot "$DOCKER_LOTUS_IMPORT_SNAPSHOT" \
+			--remove-existing-chain=false \
+			--halt-after-import
 		# Block future inits
 		date > "$GATE"
 	fi
 fi
 
 # import wallet, if provided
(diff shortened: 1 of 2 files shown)


scripts/docker-lotus-entrypoint.sh
```diff
@@ -1,18 +1,18 @@
 #!/usr/bin/env bash
 
-if [ ! -z $DOCKER_LOTUS_IMPORT_SNAPSHOT ]; then
+if [ ! -z "$DOCKER_LOTUS_IMPORT_SNAPSHOT" ]; then
 	GATE="$LOTUS_PATH"/date_initialized
 	# Don't init if already initialized.
 	if [ ! -f "$GATE" ]; then
 		echo importing minimal snapshot
-		/usr/local/bin/lotus daemon --import-snapshot "$DOCKER_LOTUS_IMPORT_SNAPSHOT" --halt-after-import
+		printf 'yes\n' | /usr/local/bin/lotus daemon --import-snapshot "$DOCKER_LOTUS_IMPORT_SNAPSHOT" --halt-after-import
```
Stebalien, line 8: This actually seems like a bug. We should only be prompting the user if there's actually a chain to remove.

Stebalien: https://github.com/filecoin-project/lotus/pull/11582

Stebalien: Fixed on master, this change shouldn't be necessary anymore. Although it may help to pass `--remove-existing-chain=false` (which will cause an immediate failure if there's an existing change, which there shouldn't be given that the node isn't initialized).

Stebalien:
@vidbregar can you rebase on master and address my comment?

scripts/docker-lotus-entrypoint.sh
vidbregar: Rebased and added the flag. Although, I'm not sure what happened to the unreleased changelog.

Stebalien (approved):

Stebalien:
Thanks!

PR #11566, opened 2024-01-10
fix: chain: use latest go-state-types types for miner UI

arajasek:
## Related Issues

@jimpick pointed out that the SectorOnChainInfo returned over the API has the wrong field names.

## Proposed Changes

Use {{.latestVersion}} types which will keep us up to date.

## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] If the PR affects users (e.g., new feature, bug fix, system requirements change), update the CHANGELOG.md and add details to the UNRELEASED section.
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #11586, opened 2024-01-17
Add missing fields for v12 miner

jimpick:
## Related Issues

https://github.com/filecoin-project/lotus/pull/11566

## Proposed Changes

Add data for `PowerBaseEpoch` and `ReplacedDayReward` fields so they appear in the JSON-RPC API.


Stebalien (approved):

Stebalien (changes requested):
Ah, that doesn't work as those lines are autogenerated. I'll push the correct fix to the upstream branch.

Stebalien:
Ok, fixed by pushing directly to that branch. Thanks for catching it!

PR #11588, opened 2024-01-18
fix: eth: re-execute tipsets on missing events

### LINKED ISSUE #11335: FEVM: Re-execute tipset on missing events
If an Eth JSON-RPC API like `EthGetTransactionReceipt` fails because we're missing the transaction's events (e.g., because we enabled the Ethereum API _after_ we processed the tipset, we should re-execute the tipset.

This will cause some work on otherwise pure "lookup" API requests however, this should only happen _once_ and it should only happen for "old" tipsets. Although we _do_ need to make sure that it only happens once (we may need some way to lock/deduplicate such requests?).

See https://github.com/filecoin-project/lotus/issues/11325.

Stebalien:
## Related Issues

fixes #11335

## Proposed Changes

This will re-execute tipsets to forcibly re-compute and store events when they're missing. This is effectively lazy backfilling of events.

NOTE: This _won't_ backfill the index itself, it'll just give us the events.

## Checklist

- [ ] CI is green

### DIFF
--- a/node/impl/full/eth_utils.go
+++ b/node/impl/full/eth_utils.go
@@ -654,7 +654,7 @@ func newEthTxFromMessageLookup(ctx context.Context, msgLookup *api.MsgLookup, tx
 	return tx, nil
 }
 
-func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLookup, events []types.Event, cs *store.ChainStore, sa StateAPI) (api.EthTxReceipt, error) {
+func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLookup, ca ChainAPI, sa StateAPI) (api.EthTxReceipt, error) {
 	var (
 		transactionIndex ethtypes.EthUint64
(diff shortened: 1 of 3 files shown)


node/impl/full/eth_utils.go
```diff
@@ -731,6 +731,24 @@ func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLook
 		receipt.ContractAddress = &addr
 	}
 
+	var events []types.Event
+	if rct := lookup.Receipt; rct.EventsRoot != nil {
+		events, err = ca.ChainGetEvents(ctx, *rct.EventsRoot)
+		if err != nil {
```
snissn, line 737: should we match a specific error here, or is just any error correct enough ?

Stebalien: Worst case, we'll waste some CPU if the datastore is corrupted. Then we'll return the same error below when we try to fetch the events again. Unfortunately, matching errors is kind of error prone.

aarshkshah1992 (approved):
lgtm

PR #11595, opened 2024-01-23
fix: sync: atomically switch chains when checkpointing

Stebalien:
-------

## Related Issues

F3

## Proposed Changes

We can now atomically switch chains when checkpointing. Previously, we'd call `SetHead` followed by `SetCheckpoint`. Unfortunately, that's not atomic and the "head" could have reverted before we called `SetCheckpoint` (causing the latter to fail).

Now, we just call `SetCheckpoint` and let `SetCheckpoint` adjust our head. This changes the behavior of `ChainStore.SetCheckpoint`, but `Syncer.SyncCheckpoint` is the only caller anyways.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/checkpoint.go
+++ b/chain/checkpoint.go
@@ -24,34 +24,20 @@ func (syncer *Syncer) SyncCheckpoint(ctx context.Context, tsk types.TipSetKey) e
 		ts = tss[0]
 	}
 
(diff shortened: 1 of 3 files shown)


Kubuxu (approved):
SGWM

Issue #11356, opened 2023-10-26
[meta] Path to marking the tracing APIs as stable


Stebalien:
- Make the other Ethereum APIs agree with them https://github.com/filecoin-project/lotus/issues/11355 (in terms of how they encode native messages).
- Fix issues around tracing native actor creation:
  - [ ] https://github.com/filecoin-project/lotus/issues/11219
  - [ ] https://github.com/filecoin-project/lotus/issues/11218
- Gracefully handle failures EVM contract creation failures https://github.com/filecoin-project/lotus/issues/11217

PR #11596, opened 2024-01-24
feat: eth: remove support for ETH tracing APIs

Stebalien:
## Related Issues

https://github.com/filecoin-project/lotus/issues/11356

## Proposed Changes

Unfortunately, these APIs turned out to be very difficult to correctly support given all the differences between Filecoin and Ethereum. The fact of the matter is that Filecoin is not Ethereum. We strive for Ethereum compatibility from the perspective of _builders_, but that doesn't mean that the Filecoin chain can be accurately indexed as if it were an Ethereum chain.

Issues with the Eth-JSON RPC API include:

1. Mapping contract creation semantics from Filecoin to Ethereum requires mapping multiple calls into a single "CREATE" operation.
2. Address translation from Filecoin Actor IDs to Ethereum addresses is error prone, especially in the face of reverts.
3. Representing native calls in an EVM friendly way is non-trivial.

That's not to say that all this is impossible to fix. However:

1. This code would require significant testing before we'd feel comfortable recommending that it be used for indexing.
2. Any indexers built on this code would likely be sub-par.
3. There's no guarantee that we won't make future changes that are fundamentally incompatible with Ethereum-style tracing.

## Additional Info

This is a request for comments, not a final proposal. But "do nothing" isn't really an option. We either need to remove it and or fix it.

For some context, I also have a 90% done patch to "fix" most of the issues in the trace API (required some changes in the FVM itself to expose more information in traces). Unfortunately, that's 90% feature-complete not 90% tested. There _are_ no tests right now and writing enough tests to be sure of this API would take quite a bit of work. And even then, I this isn't something I want to maintain.

## Checklist

- [ ] CI is green


Stebalien:
Ok, we've decided that we need this.

PR #11598, opened 2024-01-24
feat: drand: refactor round verification

arajasek:
## Related Issues

This is a refactor of some of the beacon logic in preparation of FIP-0063, although this doesn't actually implement any logic towards that. 

We currently assume that the beacon entries included in blocks must make a contiguous chain (no skipped rounds). This is currently needed, because we're using chained randomness.

## Proposed Changes

This PR doesn't lead to any change in behaviour, but changes the logic to instead rely on expected _rounds_. We expect a certain _round_ for every epoch, and include the beacon entry corresponding to that round in our blocks.

This makes the logic easier to extend when we switch to unchained randomness with a different period than Filecoin.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/beacon/beacon.go
+++ b/chain/beacon/beacon.go
@@ -79,11 +79,21 @@ func ValidateBlockValues(bSchedule Schedule, nv network.Version, h *types.BlockH
 		return xerrors.Errorf("expected to have beacon entries in this block, but didn't find any")
 	}
 
+	// Verify that the last beacon entry's round corresponds to the round we expect
 	last := h.BeaconEntries[len(h.BeaconEntries)-1]
 	if last.Round != maxRound {
 		return xerrors.Errorf("expected final beacon entry in block to be at round %d, got %d", maxRound, last.Round)
 	}
 
+	// Verify that all other entries' rounds are as expected for the epochs in between parentEpoch and h.Height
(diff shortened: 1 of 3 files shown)


chain/beacon/drand/drand.go
```diff
@@ -170,10 +170,6 @@ func (db *DrandBeacon) VerifyEntry(curr types.BeaconEntry, prev types.BeaconEntr
 		return nil
 	}
 
-	if curr.Round != prev.Round+1 {
```
arajasek, line 173: this check is now covered in the beacon logic itself, no longer drand specific.

chain/beacon/beacon.go
```diff
@@ -132,18 +142,19 @@ func BeaconEntriesForBlock(ctx context.Context, bSchedule Schedule, nv network.V
 		prev.Round = maxRound - 1
 	}
 
-	cur := maxRound
+	currEpoch := epoch
 	var out []types.BeaconEntry
-	for cur > prev.Round {
-		rch := beacon.Entry(ctx, cur)
+	for currEpoch > parentEpoch {
```
Stebalien, line 147:
```suggestion
	for currEpoch := epoch; currEpoch > parentEpoch; curEpoch-- {
```

Stebalien (approved):

PR #11603, opened 2024-01-30
test: DDO onboarding non-market verified data

rvagg:
WIP. I'm planning on adding some events testing in this. This could also do with some refactoring, maybe moving some of this into the testkit because it's quite verbose but also some of it is repeated from other verifreg testing.


itests/direct_data_onboard_test.go
```diff
@@ -140,6 +145,23 @@ func TestOnboardRawPieceVerified(t *testing.T) {
...
+			MaxEpoch: -1,
+		},
+		Prefill: true,
+	})
+	require.NoError(t, err)
+
+	events := make([]types.ActorEvent, 0)
+	go func() {
+		for e := range evtChan {
+			fmt.Printf("%s Got ActorEvent: %+v", time.Now().Format(time.StampMilli), e)
```
rvagg, line 160: currently doesn't show anything because they get filtered out due to address translation in the event filter

itests/direct_data_onboard_test.go
```diff
@@ -293,6 +315,86 @@ func TestOnboardRawPieceVerified(t *testing.T) {
...
 	require.Len(t, allocations, 0)
+
+	evts, err := miner.FullNode.GetActorEvents(ctx, &types.ActorEventFilter{
+		MinEpoch: -1,
+		MaxEpoch: -1,
+	})
+	require.NoError(t, err)
+
+	for _, evt := range evts {
+		fmt.Printf("Got ActorEvent: %+v", evt)
```
rvagg, line 326: also doesn't show anything for the same reason

itests/direct_data_onboard_test.go
```diff
@@ -293,6 +315,86 @@ func TestOnboardRawPieceVerified(t *testing.T) {
...
+	}
+
```
rvagg, line 328: from here below we use the existing lotus APIs to get receipts and their event logs for the whole chain, which works

itests/direct_data_onboard_test.go
```diff
@@ -102,6 +118,310 @@ func TestOnboardRawPiece(t *testing.T) {
...
+	mkVerifier(ctx, t, client.FullNode.(*api.FullNodeStruct), rootAddr, verifier1Addr)
+
+	// assign datacap to a client
+	initialDatacap := big.NewInt(10000)
+
+	params, err := actors.SerializeParams(&verifregtypes13.AddVerifiedClientParams{Address: verifiedClientAddr, Allowance: initialDatacap})
+	require.NoError(t, err)
+
+	msg := &types.Message{
```
arajasek, line 212: we really need to extract some of this logic into a helper to be reused by multiple tests

itests/direct_data_onboard_test.go
```diff
@@ -102,6 +123,368 @@ func TestOnboardRawPiece(t *testing.T) {
...
+	}
+	claimKeyCbor, err := ipld.Encode(basicnode.NewString("claim"), dagcbor.Encode)
+	require.NoError(t, err)
+
+	for _, event := range events {
+		entryStrings := []string{
+			fmt.Sprintf("height=%d", event.Height),
+			fmt.Sprintf("msg=%s", event.MsgCid),
+			fmt.Sprintf("emitter=%s", event.EmitterAddr),
+			fmt.Sprintf("reverted=%t", event.Reverted),
```
aarshkshah1992, line 407: This information is not persisted on message receipts. How will client be informed about reverted events if they only depend on message receipts ?

itests/direct_data_onboard_test.go
aarshkshah1992: This is fixed.

itests/direct_data_onboard_test.go
aarshkshah1992: Fixed

aarshkshah1992:
Addes tests for the events API.

aarshkshah1992 (approved):
lgtm.

itests/direct_data_onboard_test.go
```diff
@@ -102,6 +125,458 @@ func TestOnboardRawPiece(t *testing.T) {
...
+	for _, evt := range ptr {
+		evts = append(evts, *evt)
+	}
+	return evts
+}
+
+func matchEvents(t *testing.T, exp []types.ActorEvent, actual []types.ActorEvent) {
+	// height and tipset cid can mismatch because expected events are sourced using APIs that can put in different tipsets
+	for i := range exp {
+		exp[i].Height = 0
```
aarshkshah1992, line 409: @rvagg The tipset/height we put in the `buildActorEventsFromMessages ` function below on these events is always one ahead of the actual tipset/height these events are actually emitted at (as recorded by the `CollectEvents` function in lotus/chain/events/index.go).

I tried digging into it and it looks like this is because the ` node.StateSearchMsg` function can return messages from a previous tipset as well. Let me know what you think here. I think this is a non-issue.

rvagg: Oh, that's very interesting. Probably a non issue as far as the tests are concerned but this is really helpful to know as I write docs.

aarshkshah1992:
itests will be green once the FFI/ref-fvm deps changes for NV22 are merged into this branch.

aarshkshah1992:
@rvagg Merging this so we can iron out the final parts in https://github.com/filecoin-project/lotus/pull/11540.

PR #11608, opened 2024-02-03
feat: fvm: update the FVM/FFI to v4.1

Stebalien:
## Proposed Changes

This:

1. Adds nv22 support.
2. Updates the message tracing format.

## Additional Info

I'm making this against master and not against the feat/nv22 branch because I need to base the tracing fixes branch on it.

## Checklist

- [ ] CI is green


chain/types/execresult.go
```diff
@@ -38,10 +40,11 @@ type ReturnTrace struct {
 }
 
 type ExecutionTrace struct {
-	Msg        MessageTrace
-	MsgRct     ReturnTrace
-	GasCharges []*GasTrace      `cborgen:"maxlen=1000000000"`
-	Subcalls   []ExecutionTrace `cborgen:"maxlen=1000000000"`
+	Msg          MessageTrace
+	MsgRct       ReturnTrace
+	InvokedActor *ActorTrace      `json:",omitempty"`
```
rvagg, line 45: Is this nullable to make it backward compatible, do we guarantee it exists on traces from this point forward? There's usage in `node/impl/full/eth_trace.go` that assumes it exists and will panic otherwise (3 instances of `et.InvokedActor.State.Code`).

Stebalien: No, it can definitely not be set. Nice catch!

Stebalien: Fixed by simply skipping these traces. It can only happen if we fail to resolve the target actor, and that's not a case that's even possible in the EVM (i.e., the EVM would have aborted the call before it even hit the tracing logic).

rvagg (approved):
do we store these anywhere, so does this require a migration of some kind?

Stebalien:
We don't. However, this is API breaking. @arajasek what kind of guarantees do we make for the trace API?

Stebalien:
Specifically https://github.com/filecoin-project/lotus/pull/11608/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR7-R49

arajasek (approved):
I think this is fine, we can write up a little integration guide based on the info in the changelog.

PR #11609, opened 2024-02-03
feat: api: improve the correctness of Eth's trace_block

### LINKED ISSUE #11217: TraceAPI: Support failures while creating contracts
Left as TODO in: https://github.com/filecoin-project/lotus/pull/11100
Relevant Comment: https://github.com/filecoin-project/lotus/pull/11100/files#r1305971682

We need to handle failures in contract creations and support resurrections on an existing but dead EVM actor)

We have a few cases:
- Failures. I.e., we may fail halfway through this path for some reason and need to treat it as a failed to creation.
- Resurrection! EAM "create" and/or "create2" may not call the init actor but may, instead call "resurrect" on an existing but dead EVM actor. Let's handle that later, but it's an important part of "EVM-style" code upgrades (which are not the same as FVM code upgrades).

### LINKED ISSUE #11218: TraceAPI: Support sub-calls from "native" contract constructors
Left as TODO in: https://github.com/filecoin-project/lotus/pull/11100
Relevant comment: https://github.com/filecoin-project/lotus/pull/11100#discussion_r1308033493

### LINKED ISSUE #11219: TraceAPI: Support tracing messages that create native actors
Left as TODO in: https://github.com/filecoin-project/lotus/pull/11100
Relevant comment: https://github.com/filecoin-project/lotus/pull/11100#discussion_r1302442288

### LINKED ISSUE #11355: Improve Eth JSON-RPC API handling of native messages

### LINKED ISSUE #11356: [meta] Path to marking the tracing APIs as stable

Stebalien:
## Related Issues

fixes #11217
fixes #11355
fixes #11356
fixes #11219
fixes #11218
fixes https://github.com/filecoin-project/ref-fvm/issues/1793

## Proposed Changes

This tracing API is supposed to translate FVM traces to ETH traces, translating FVM semantics (calls, actor creation, etc.) into ETH semantics.

This PR improves the current implementation (I hope):

- Improve encoding/decoding of parameters and return values:
  - Encode "native" parameters and return values with Solidity ABI.
  - Correctly decode parameters to "create" calls.
  - Use the correct (ish) output for "create" calls.
  - Handle all forms of "create".
- Make robust with respect to reverts:
  - Use the actor ID/address from the trace instead of looking it up in the state-tree (may not exist in the state-tree due to a revert).
  - Gracefully handle failed actor/contract creation.
- Improve performance:
  - We avoid looking anything up in the state-tree when translating the trace, which should significantly improve performance.
- Improve code readability:
  - Remove all "backtracking" logic.
  - Use an "environment" struct to store temporary state instead of attaching it to the trace.
- Fix random bugs:
  - Fix an allocation bug in the "address" logic (need to set the capacity before modifying the slice).
  - Improved error checking/handling.

There are still some "not great" things here, but I'm reasonably happy with the result.

## Limitations

- No "rewards". We could, but I don't yet do that and don't intend to fix that in this patch.
  - https://github.com/filecoin-project/lotus/issues/11633
- No "selfdestruct". Given access to intermediate state along with the state-root of every actor before/actor, we could fish out this information. But it's going to be a non-trivial amount of work.
  - https://github.com/filecoin-project/lotus/issues/11634
- All "create" calls return `0xfe` as the contract's bytecode instead of the correct. To fix this, we need intermediate state (state discarded during block execution). This _could_ be fixed in the future.
  - https://github.com/filecoin-project/lotus/issues/11635

## Additional Information

Instead of describing how this logic is supped to work in the PR description, I'm hoping that the comments I've left inline will suffice. If they don't, we need to fix that.

The main entry-point is `buildTraces` in `eth_trace.go`.

## Checklist

- [x] CI is green


node/impl/full/eth_trace.go
```diff
@@ -38,212 +41,438 @@ func decodePayload(payload []byte, codec uint64) (ethtypes.EthBytes, error) {
...
-						parent.LastByteCode = trace
-					}
-
-					return nil
-				}
+				// We're _supposed_ to include the contracts bytecode here, but we
+				// can't do that reliably (e.g., if some part of the trace reverts).
+				// So we don't try and include a sentinel "impossible bytecode"
+				// value (the value specified by EIP-3541).
+				trace.Result.Output = []byte{0xFE}
```
Stebalien, line 395: So, I'm supposed to return the contract's bytecode here. Unfortunately, I can't do that reliably because reverts (the bytecode may not actually exist in the state after the block is executed).

Options include:

1. Adding this information to the trace (that's how we've fixed all of our other "revert" issues). But, I _really_ don't want to do that.
2. Keep this as 0xFE.
3. Omit `Output` entirely?
4. Return `null`?

node/impl/full/eth_trace.go
```diff
@@ -38,212 +41,438 @@ func decodePayload(payload []byte, codec uint64) (ethtypes.EthBytes, error) {
...
+	//    method, we know something special is going on. We look for calls related to
+	//    DELEGATECALL and drop everything else (everything else includes calls triggered by,
+	//    e.g., EXTCODEHASH).
+
+	// If we don't have sufficient funds, or we have a fatal error, or we have some
+	// other syscall error: skip the entire trace to mimic Ethereum (Ethereum records
+	// traces _after_ checking things like this).
+	//
+	// NOTE: The FFI currently folds all unknown syscall errors into "sys assertion
+	// failed" which is turned into SysErrFatal.
```
Stebalien, line 174: Maybe we should fix this?

node/impl/full/eth_trace.go
```diff
@@ -38,212 +41,438 @@ func decodePayload(payload []byte, codec uint64) (ethtypes.EthBytes, error) {
...
+				// Otherwise, this can happen if creation fails early (bad params,
+				// out of gas, contract already exists, etc.). The EVM wouldn't
+				// trace such cases, so we don't either.
+				//
+				// NOTE: It's actually impossible to run out of gas before calling
+				// initcode in the EVM (without running out of gas in the calling
+				// contract), but this is an equivalent edge-case to InvokedActor
+				// being nil, so we treat it the same way and skip the entire
+				// operation.
+				return nil, nil, nil
```
Stebalien, line 285: Thoughts on better ways to handle this?

Stebalien:
This is now working (tested locally). However, I still need to figure out how to write real tests. My thinking is:

1. Unit test some of the encoding/decoding functions.
2. Dump a bunch of FVM traces and their corresponding Eth traces to files and then test to make sure we (a) get the same results every time and (b) everything (paths, etc.). Make sense.

Thoughts on this plan?

Stebalien:
You should be able to test with:

```bash
trace_block() {
    local blk="$(printf '0x%x' $1)"
    local req="{\"jsonrpc\": \"2.0\", \"id\": 0, \"method\": \"trace_block\", \"params\": [\"$blk\"]}"
    curl -H "Content-Type: application/json" -X POST --data "$req"http://localhost:1234/rpc/v1
}
```

Assuming you have a local node with the Eth API enabled.

documentation/en/api-v1-unstable-methods.md
```diff
@@ -3110,7 +3110,8 @@ Response:
     "traceAddress": [
       123
     ],
-    "Type": "string value",
+    "type": "string value",
```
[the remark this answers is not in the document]
i-norden: This is correct based on the erigon code.

node/impl/full/eth_trace.go
```diff
@@ -38,217 +37,441 @@ func decodePayload(payload []byte, codec uint64) (ethtypes.EthBytes, error) {
...
+	//    e.g., EXTCODEHASH).
+
+	// If we don't have sufficient funds, or we have a fatal error, or we have some
+	// other syscall error: skip the entire trace to mimic Ethereum (Ethereum records
+	// traces _after_ checking things like this).
+	//
+	// NOTE: The FFI currently folds all unknown syscall errors into "sys assertion
+	// failed" which is turned into SysErrFatal.
+	if len(addr) > 0 {
+		switch et.MsgRct.ExitCode {
```
i-norden, line 175: Can this guard be moved up to the `buildTraces` level?

chain/types/ethtypes/eth_types.go
```diff
@@ -987,13 +991,8 @@ type EthTrace struct {
...
 	TraceAddress []int          `json:"traceAddress"`
-	Type         string         `json:"Type"`
-
-	Parent *EthTrace `json:"-"`
-
-	// if a subtrace makes a call to GetBytecode, we store a pointer to that subtrace here
-	// which we then lookup when checking for delegatecall (InvokeContractDelegate)
-	LastByteCode *EthTrace `json:"-"`
+	Type         string         `json:"type"`
+	Error        string         `json:"error,omitempty"`
```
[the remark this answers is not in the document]
i-norden: That looks correct to me based on erigon. Erigon trace result also "omitempty" on blockhash, blockNumber, transactionHash, and transactionPosition.

node/impl/full/eth_trace.go
Stebalien: You know, now that i think of it, no. Because:

1. We only flush the state at the end of the block, not at the end of every transaction.
2. A contract can be repeatedly created/destroyed (with the same init code but different _bytecode_), once per transaction.

That is...

TxA:

1. ContractA: calls Create2 with INITCODE.
2. ContractT (new): INITCODE loads the bytecode from the caller, and returns it.
3. ContractA: does some stuff, then calls back into ContractT.
4. ContractT: Calls SELFDESTRUCT.

TxB:

1. ContractB: calls Create2 with INITCODE (same INITCODE as above).
2. ContractT (re-created): INITCODE loads the bytecode from the caller (ContractB this time) and returns it. This is now _different_ bytecode.
3. ContractB: does some stuff, then calls back into ContractT.
4. ContractT: Calls SELFDESTRUCT.

By the end of the block, even with no reverted transactions, the ContractT could have changed its bytecode any number of times.

Stebalien: This is fixable, it just requires even more changes to the FVM:

1. Record an actors state-object when it returns (instead of only when we call it).
2. Add a flag to keep all the intermediate state, and use it when tracing. But that could produce a _lot_ of garbage and I'd only be comfortable doing it if we passed in a special _separate_ datastore per-trace.

The upside is that this would give us nearly complete information on intermediate states in the system. So it's not a _bad_ idea, just something that would likely have to happen in a followup PR.

documentation/en/api-v1-unstable-methods.md
```diff
@@ -3110,7 +3110,8 @@ Response:
...
+    "error": "string value",
     "blockHash": "0x37690cfec6c1bf4c3b9288c7a5d783e98731e90b0a4c177c2a374c7a9427355e",
     "blockNumber": 9,
     "transactionHash": "0x37690cfec6c1bf4c3b9288c7a5d783e98731e90b0a4c177c2a374c7a9427355e",
```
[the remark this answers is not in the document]
Stebalien: We _should_ only be including those in the top-level trace.

Stebalien:
@i-norden thanks for pointing me at erigon, I'll have to make a few fixes:

1. We don't support suicide traces. Oh well.
2. The top-level "type" should be "call" for all calls and only the action's "callType" should depend on the call type.
3. I need a totally different trace action for create, as far as I can tell.
4. I likely need to convert the reward messages as well...

node/impl/full/eth_trace.go
```diff
@@ -38,217 +37,441 @@ func decodePayload(payload []byte, codec uint64) (ethtypes.EthBytes, error) {
...
 		return nil
 	}
 
-	// Skip the trace if we never reached the point where we invoked this actor.
+	subEnv := &environment{
+		caller: traceToAddress(recurseInto.InvokedActor),
+		isEVM:  builtinactors.IsEvmActor(recurseInto.InvokedActor.State.Code),
+		traces: env.traces,
+	}
+	addr = addr[:len(addr):len(addr)]
```
fridrik01, line 114: maybe add a comment here

node/impl/full/eth_trace.go
```diff
@@ -38,217 +37,441 @@ func decodePayload(payload []byte, codec uint64) (ethtypes.EthBytes, error) {
...
+		return "create", input, nil
+	default:
+		return "", nil, xerrors.Errorf("unexpected CREATE method %d", msg.Method)
+	}
+}
+
+// buildTrace processes the passed execution trace and updates the environment, if necessary.
+//
+// On success, it returns a trace to add (or nil to skip) and the trace recurse into (or nil to skip).
+func buildTrace(env *environment, addr []int, et *types.ExecutionTrace) (*ethtypes.EthTrace, *types.ExecutionTrace, error) {
```
fridrik01, line 155: this function is really long and complex and we should consider refactoring into handlers for different code paths

node/impl/full/eth_trace.go
fridrik01: Seems fine to ignore to me :S

fridrik01 (approved):
This seems like a solid improvement!. The implementation is getting really quite complex though, but I appreciated the nice comments you put in there. The idea of dumping bunch of fvm traces and corrseponding eth traces as you mention sounds like a great idea to me and would help maintain this code going forward.

fridrik01:
> This is now working (tested locally). However, I still need to figure out how to write real tests. My thinking is:
> 
> 1. Unit test some of the encoding/decoding functions.
> 2. Dump a bunch of FVM traces and their corresponding Eth traces to files and then test to make sure we (a) get the same results every time and (b) everything (paths, etc.). Make sense.
> 
> Thoughts on this plan?

That makes sense, especially 2) to me

Stebalien:
Ok, I've now refactored this according to the feedback:

1. I've split it into multiple (many) functions.
2. It now accurately handles call v. create with two separate sets of actions & results.
  1. The top-level "type" is now one of "call" or "create".
  2. The actual call type is only recorded in "callType".
  3. The "create" fields are named correctly.

Limitations:

1. I don't handle "rewards". We could, but I don't yet do that and don't intend to fix that in this patch.
2. I don't handle "selfdestruct". We don't have enough information.

Changes:

- On failure, if I fail to decode a revert reason, I just leave it blank instead of abi-encoding it as a "native" return value.
- I'm now setting the "error" message to "Reverted" on revert, because tools likely expect that specific error message.

chain/actors/builtin/evm/evm.go
```diff
@@ -16,6 +17,17 @@ import (
 
 var Methods = builtin12.MethodsEVM
 
+const (
+	ErrReverted exitcode.ExitCode = iota + 33
```
fridrik01, line 21: its not clear where the 33 is coming from?

chain/actors/builtin/evm/evm.go
```diff
@@ -16,6 +17,17 @@ import (
...
+	ErrInvalidInstruction
+	ErrUndefinedInstruction
+	ErrStackUnderflow
+	ErrStackOverflow
+	ErrIllegalMemoryAccess
+	ErrBadJumpdest
+	ErrSelfdestructFailed
+)
```
fridrik01, line 29: where are these used?

fridrik01 (approved):
lgtm, some minor nits

chain/actors/builtin/evm/evm.go
Stebalien: https://github.com/filecoin-project/lotus/blob/cf910a8d1a25d46267412b3e5f7a569ef16b592d/node/impl/full/eth_trace.go#L142-L159

Stebalien: They're returned by the EVM on error: https://github.com/filecoin-project/builtin-actors/blob/6e781444cee5965278c46ef4ffe1fb1970f18d7d/actors/evm/src/lib.rs#L35-L42

Stebalien:
Issue for writing more tests: https://github.com/filecoin-project/lotus/issues/11638

PR #11615, opened 2024-02-09
docs: add first pass of actor events and DDO documentation

rvagg:
These started as the same doc but they're different enough that they've been split up. They don't necessarily belong here either, feedback welcome on what to do with them! Working through this has been quite an educational experience and has lead to some changes to the APIs already and I want to propose a few more as a result.

* **Actor Events and Lotus APIs** is intended to be a general doc describing what the builtin actor events are and how to consume them. It will evolve over time as more events are added in future FIPs so should be a good single source of information about all of the events available and what they look like.
* **Data Onboarding Visibility** is intended as a kind of integration guide - what to do in a post-nv22 world where DDO starts making data accounting more difficult. How to access the information, where the best source of truth is and what not to rely on. It's got a basic primer on DDO and the way it changes the way we need to view data stored on Filecoin.


documentation/en/actor-events-api.md
```diff
@@ -0,0 +1,434 @@
...
+
+Typically events contain entires that use either use `0x01` or `0x03` flags.
+
+The structured logging style of composition should be seen in contrast to an alternative representation as a plain map or struct where the keys represent the fields of the event and the values represent the values of those fields. Some entries may duplicate keys, in which case that particular field of the event could be represented as an array. Builtin actor events are sufficiently well defined that translation to such a format is possible, but left up to the user.
+
+## Querying Lotus for ActorEvents
+
+Two Lotus APIs are provided that can be used to obtain direct access to events stored on the node being queried (a node may not have all historical events stored and available for query):
+
+_TODO: update links_
```
rjan90, line 75: What is blocking us from updating the links?

rvagg: 1. I wanted to check whether these docs are published anywhere else or if pointing into the repo is appropriate
2. These aren't in `master` yet, I'm hessitant to point a link to something that's not merged. It's on feat/nv22 and I don't know the timeframe for merging that.

Could update it now, but the links won't work till it's all landed.

documentation/en/actor-events-api.md
```diff
@@ -0,0 +1,434 @@
...
+    * [Deal completed](#deal-completed)
+  * [Miner actor events](#miner-actor-events)
+    * [Sector precommitted](#sector-precommitted)
+    * [Sector activated](#sector-activated)
+    * [Sector updated](#sector-updated)
+    * [Sector terminated](#sector-terminated)
+
+## Background
+
+Actor events are a fire-and-forget mechanism for actors in Filecoin to signal events that occur during execution of their methods to external observers. The events themselves are not stored on chain, but are emitted by the actor and may be stored in a node's event store. They are intended to be used by tooling and applications that need to observe and react to events that occur within the chain.
```
jennijuju, line 28:
```suggestion
Actor events are a fire-and-forget mechanism for actors in Filecoin to signal events that occur during execution of their methods to external observers. The events themselves are not stored on chain (in state), but are emitted by the actor and may be stored in a node's event store. They are intended to be used by tooling and applications that need to observe and react to events that occur within the chain.
```

documentation/en/actor-events-api.md
```diff
@@ -0,0 +1,434 @@
...
+
+The FEVM already has this capability and these events for builtin actors have been added to support a range of new features, starting at network version 22 with a focus on some information gaps for consumers of data onboarding metrics due to the introduction of Direct Data Onboarding (DDO), plus some additional events related to data onboarding, deal lifecycles, sector lifecycles, and DataCap activity. Additional events are expected to be added in the future to support other features and use cases.
```
jennijuju, line 30:
```suggestion
The FVM already has this capability and these events for builtin actors have been added to support a range of new features, starting at network version 22 with a focus on some information gaps for consumers of data onboarding metrics due to the introduction of Direct Data Onboarding (DDO), plus some additional events related to data onboarding, deal lifecycles, sector lifecycles, and DataCap activity. Additional events are expected to be added in the future to support other features and use cases.
```

documentation/en/actor-events-api.md
```diff
@@ -0,0 +1,434 @@
...
+
+Builtin actor events share basic similarities to the existing events emitted by the FEVM, but are structured differently to reflect Filecoin-specific concerns. There are also new APIs in Lotus to support querying for these events that bear some similarities to the existing FEVM `Eth*` APIs for querying events but are unique to builtin actors.
```
jennijuju, line 32:
```suggestion
Built-in actor events share basic similarities to the existing events emitted by FEVM-based user actors but are structured differently to reflect Filecoin-specific concerns. There are also new APIs in Lotus to support querying for these events that bear some similarities to the existing FEVM `Eth*` APIs for querying events but are unique to builtin actors.
```

jennijuju: > but are unique to builtin actors.

is this true now that we went for the generic approach with GetActorEvents & SubscribeActorEvents -> which supports *both* builtin actors and user smart contract events?

documentation/en/actor-events-api.md
```diff
@@ -0,0 +1,434 @@
...
+
+## ActorEvent structure
+
+Introduced in [FIP-0049](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0049.md), events use a structured logging style of composition, containing a list of entries that define properties of the event. The log entries are described below as `EventEntry` and have the same schema for FEVM and builtin actor events. `ActorEvent` is specifically for representing builtin actor events and includes the list of entries, the actor that emitted the event, and some metadata about the event.
```
jennijuju, line 36:
```suggestion
Introduced in [FIP-0049](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0049.md), events use a structured logging style of composition, containing a list of entries that define properties of the event. The log entries are described below as `EventEntry` and have the same schema for user deployed and builtin actor events. `ActorEvent` is specifically for representing builtin actor events and includes the list of entries, the actor that emitted the event, and some metadata about the event.
```

documentation/en/actor-events-api.md
```diff
@@ -0,0 +1,434 @@
...
+
+type ActorEventValue struct {
+  codec Int   # typically the CBOR codec (0x51)
+  value Bytes # typically the CBOR encoded value
+}
+```
+
+## Retrieving events from message receipts
+
+The Lotus API `ChainGetEvents` can be used to retrieve events given an event root CID. This CID is attached to the message receipt that generated the events. The `StateSearchMsg` API can be used to retrieve the message receipt given a message CID, the receipt contains the `EventsRoot` CID. The events returned from `ChainGetEvents` contain roughly the same information as the `ActorEvent` structure, including the `EventEntry` log array.
```
jennijuju, line 110: TIL this exists

documentation/en/data-onboarding-visibility.md
```diff
@@ -0,0 +1,71 @@
...
+* [DDO information flow](#ddo-information-flow)
+* [Relevant message contents](#relevant-message-contents)
+* [Relevant builtin actor events](#relevant-builtin-actor-events)
+
+## Introduction and background
+
+**Direct Data Onboarding** (DDO) as defined in **[FIP-0076](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0076.md)** provides an optional data onboarding path that is both gas-efficient and paves a path toward eventual smart contract mediated onboarding mechanisms. The existing market actor and market actor mediated onboarding pathway remains largely unchanged but is now optional; it is anticipated that the gas savings alone will see a significant shift away from use of the market actor.
+
+Historically, a large amount of tooling was built around Filecoin that makes use of the market actor to quantify various data-related metrics. A shift in behaviour toward data onboarding that bypasses the market actor requires adaption in order to have continuity with the some of the data-related metrics being collected. This will continue to be true as Filecoin evolves to enable onboarding mechanisms mediated by smart contracts.
```
jennijuju, line 12:
```suggestion
Historically, a large amount of tooling was built around Filecoin that makes use of the market actor (f05) to quantify various data-related metrics. A shift in behaviour toward data onboarding that bypasses the market actor requires adaption in order to have continuity with the some of the data-related metrics being collected. This will continue to be true as Filecoin evolves to enable onboarding mechanisms mediated by smart contracts.
```

documentation/en/data-onboarding-visibility.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Deal-lifecycle actor events** as defined in **[FIP-0083](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0083.md)** and detailed in [Actor Events and Lotus APIs](./actor-events-api.md) introduced the first batch of fire-and-forget externally observable events for builtin actors. The FEVM already had this capability and these new first events for builtin actors were added to increase the visibility of information around data onboarding, particularly in light of the introduction of DDO which will require metrics gatherers to rely on mechanisms other than the market actor to collect data.
```
jennijuju, line 14:
```suggestion
**Deal-lifecycle actor events** as defined in **[FIP-0083](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0083.md)** and detailed in [Actor Events and Lotus APIs](./actor-events-api.md) introduced the first batch of fire-and-forget externally observable events for builtin actors. The FVM already had this capability and these new first events for builtin actors were added to increase the visibility of information around data onboarding, particularly in light of the introduction of DDO which will require metrics gatherers to rely on mechanisms other than the market actor to collect data.
```

documentation/en/data-onboarding-visibility.md
```diff
@@ -0,0 +1,71 @@
...
+
+Actor events are an optional method for gaining insight into data onboarding metrics. Messages may also be used as a source of truth for data onboarding metrics, but are more complex to consume and may not be suitable for some workflows. For verified data, the verified registry actor should be used as the primary source of truth for data; however FIP-0076 introduces the possibility of "sparkling data" which is not verified and not mediated through the builtin market actor or possibly any other actor. This data currently still requires piece commitments to be detailed as part of a miner's sector commitments, and so may be observed through messages and actor events that carry sector commitment piece manifests.
```
jennijuju, line 16:
```suggestion
Actor events are an optional method for gaining insight into data onboarding activities & data/storage lifecycle. Messages may also be used as a source of truth for data onboarding metrics, but are more complex to consume and may not be suitable for some workflows. For verified data, the verified registry actor should be used as the primary source of truth for data; however FIP-0076 introduces the possibility of "sparkling data" which is not verified and not mediated through the builtin market actor or possibly any other actor. This data currently still requires piece commitments to be detailed as part of a miner's sector commitments, and so may be observed through messages and actor events that carry sector commitment piece manifests.
```

documentation/en/data-onboarding-visibility.md
```diff
@@ -0,0 +1,71 @@
...
+
+There are two possible additions to this flow:
+
+- Prior to PreCommit, an SP publishes a storage deal to the builtin market actor, in which case deal information exists on chain as it does with non-DDO deals today; then
+  - at ProveCommit, or ReplicaUpdate, the SP can notify an actor of the commitment. Currently this can only be the builtin market actor (in the future this may be a list of arbitrary user defined actors), in which case it will be used to activate a deal previously proposed on chain.
+- At ProveCommit, or ReplicaUpdate, the SP can claim DataCap that was previously allocated by the client for a particular piece.
+
+💡 **The builtin market actor should not be used as single a source of truth regarding data onboarding metrics.** The builtin market actor is only a source of truth for data onboarding mediated by the builtin market actor.
```
jennijuju, line 33:
```suggestion
💡 **The builtin market actor should not be used as single a source of truth regarding data onboarding activities.** The builtin market actor is only a source of truth for data onboarding mediated by the builtin market actor.
```

jennijuju:
This is a great start - thank you 

could you please add add an API call example, with payload and response - for example https://docs.filecoin.io/reference/exchanges/exchange-integration#api-examples

documentation/en/actor-events-api.md
rvagg: bumped to `master` even though they're not there .. they _will_ be

documentation/en/actor-events-api.md
rvagg: bit more complicated than that, I've adjusted this para to read:

Actor events are a fire-and-forget mechanism for actors in Filecoin to signal events that occur during execution of their methods to external observers. Actor events are intended to be used by tooling and applications that need to observe and react to events that occur within the chain. The events themselves are not stored in chain state, although a root CID for an array (AMT) of all events emitted for a single message is recorded on message receipts, which are themselves referenced as an array (AMT) in the `ParentMessageReceipts` in each `BlockHeader` of a tipset. A node may optionally retain historical events for querying, but this is not guaranteed and not essential as it does not affect the chain state.

documentation/en/actor-events-api.md
rvagg: yeah, revisiting some of this with fresh, and more informed eyes, I see a bunch of stuff that's not _quite_ right. Have reworked this para and some other stuff around it

documentation/en/actor-events-api.md
```diff
@@ -0,0 +1,432 @@
...
+```
+
+A `flags` field is used to convey metadata or hints about the entry, currently this is used to provide an indication of the suitability of that field for indexing when stored or when used for querying.
+
+* A `flag` of `0x00` indicates that neither the key nor value are suitable for indexing.
+* A `flag` of `0x01` indicates that the key only is suitable for indexing.
+* A `flag` of `0x02` indicates that the value is suitable for indexing.
+* A `flag` of `0x03` indicates that both the key and value are suitable for indexing.
+
+Suitability for indexing is a hint, and will typically indicate the uniqueness of the content. For example if a value has an arbitrary number, such as a `piece-size`, it's unlikely to be suitable for building an index for. However, if a field such as a `"provider"` contains an actor ID, it is likely to be suitable for indexing because queries may be made for all events from a particular actor.
```
raulk, line 65: I think the logic here is a bit off, and carries some incorrect assumptions. It's not the uniqueness/repetitiveness of a value that determines whether an index is appropriate or not, but rather its queriability. One can perfectly build an index for high variability or unique data, DBMS do this all the time. But our query APIs only offer equality predicates today, so that's what influences what we chose to tag as indexed, eg makes sense to index a CID, but it makes sense to index an arbitrary size. If we exposed gt, lt, etc predicates in queries, indexing the latter would become more worthwhile.

rvagg: @raulk OK, good input, but can you suggest new wording here please to reflect that?

```suggestion
Suitability for indexing is a hint, and will typically indicate the queriability content.
```

That's about the best I have once you boil it down. From a _user_ perspective, what do we have to say about this field? We're not even using it internally and so far have no plans to. If we can imagine users relying on it, how do we describe it? Maybe leave it simple and move on.

rvagg: I integrated this into the sentence above the flags description since it already says something similar. I removed this paragraph entirely.

rjan90 (approved):

PR #11616, opened 2024-02-12
Add lotus-provider to build to match install

rvagg:
There's currently a mismatch between `build` and `install` in that `install` always tries to install `lotus-provider` and will error if it's not built, but `build` doesn't build it for you; so a `make clean && make build && sudo make install` will always fail.

This adds `lotus-provider` to `build`, but the alternative is to remove `lotus-provider` from `install`. That option may be more appropriate since it's still under development?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ lotus-gateway: $(BUILD_DEPS)
 .PHONY: lotus-gateway
 BINS+=lotus-gateway
 
-build: lotus lotus-miner lotus-worker 
+build: lotus lotus-miner lotus-worker lotus-provider
 	@[[ $$(type -P "lotus") ]] && echo "Caution: you have \
 an existing lotus binary in your PATH. This may cause problems if you don't run 'sudo make install'" || true
 


rjan90 (approved):
Confirmed to fix the error: 
```
go build  -ldflags="-X=github.com/filecoin-project/lotus/build.CurrentCommit=+git.16b3f12c8.dirty" -o lotus ./cmd/lotus
rm -f lotus-miner
go build  -ldflags="-X=github.com/filecoin-project/lotus/build.CurrentCommit=+git.16b3f12c8.dirty" -o lotus-miner ./cmd/lotus-miner
rm -f lotus-worker
go build  -ldflags="-X=github.com/filecoin-project/lotus/build.CurrentCommit=+git.16b3f12c8.dirty" -o lotus-worker ./cmd/lotus-worker
rm -f lotus-provider
go build  -ldflags="-X=github.com/filecoin-project/lotus/build.CurrentCommit=+git.16b3f12c8.dirty" -o lotus-provider ./cmd/lotus-provider
Caution: you have an existing lotus binary in your PATH. This may cause problems if you don't run 'sudo make install'
```

snadrus (approved):

PR #11618, opened 2024-02-13
feat: events: Add Lotus APIs to consume smart contract and built-in actor events

rvagg:
## Related Issues
- For https://github.com/filecoin-project/lotus/issues/11457
- FIP-discussion for built-in Actor events at https://github.com/filecoin-project/FIPs/discussions/754
- [FIP-0083](https://github.com/filecoin-project/FIPs/blob/7eb8da9f5d50f2664e059bd56c930272d6189a57/FIPS/fip-0083.md) for built-in Actor events 

## Proposed Changes
This work implements a new API in Lotus for clients to consume both smart contract and the upcoming built-in Actor events in [FIP-0083](https://github.com/filecoin-project/FIPs/blob/7eb8da9f5d50f2664e059bd56c930272d6189a57/FIPS/fip-0083.md)

Replaces https://github.com/filecoin-project/lotus/pull/11540 but with a linear git history and it's actually mergeable. I've squashed it down to a few major commits, retaining (mostly) the provenance. The result is identical to https://github.com/filecoin-project/lotus/pull/11540 except for two very minor example doc changes (`"ddata"` instead of `"data"` for some cbor, it was done for one of the 3 and I didn't notice until now).

## Additional Info
This would be a good one to rebase-merge since it's already linear and sits at the top of `feat/nv22`.

## Checklist

- [ ] CI is green


rjan90 (approved):
LGTM!

documentation/en/default-lotus-config.toml
```diff
@@ -330,6 +330,13 @@
   # env var: LOTUS_FEVM_ENABLEETHRPC
   #EnableEthRPC = false
 
+  # EnableActorEventsAPI enables the Actor events API that enables clients to consume events emitted by (smart contracts + built-in Actors).
+  # This will also enable the RealTimeFilterAPI and HistoricFilterAPI by default, but they can be disabled by config options above.
+  #
+  # type: bool
+  # env var: LOTUS_FEVM_ENABLEACTOREVENTSAPI
+  #EnableActorEventsAPI = false
```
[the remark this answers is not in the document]
rvagg: Discussion started but not resolved in https://github.com/filecoin-project/lotus/pull/11540#pullrequestreview-1841106215

We currently have a `Fevm` section in config, not `Fvm`. Do you think this warrants starting a new config section for one item?

chain/types/actor_event.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// Filecoin address of the actor that emitted this event.
+	EmitterAddr address.Address `json:"emitter"`
+
+	// Reverted is set to true if the message that produced this event was reverted because of a network re-org
+	// in that case, the event should be considered as reverted as well.
+	Reverted bool `json:"reverted"`
+
+	// Height of the tipset that contained the message that produced this event.
+	Height abi.ChainEpoch `json:"height"`
```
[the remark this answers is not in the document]
rvagg: Yeah, that was me, Aarsh originally had it consistent with the FEVM events API but to me "Block" seems EVM but "Height" seems Filecoin. But I'm a n00b at the terminology so what do you suggest? I like consistency, but "Height" seems to be winning in most public Lotus APIs.

documentation/en/default-lotus-config.toml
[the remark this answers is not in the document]
rvagg: I think maybe an `Events` top level config option would be appropriate but that logically should move all of the current `Fevm.Events` options into it as well since this feature builds on top of it. But that probably needs a config section deprecation and duplicate handling during a deprecation period. I can do that if it makes sense but it seems heavy-handed.

Maybe the alternative is to put this in a new `Events` section alone for now and do a move of the other options at a later date as a cleanup.

node/builder_chain.go
```diff
@@ -277,6 +281,15 @@ func ConfigFullNode(c interface{}) Option {
 			),
 		),
 
+		ApplyIf(isFullNode,
+			If(cfg.Fevm.EnableActorEventsAPI,
+				Override(new(full.ActorEventAPI), modules.ActorEventAPI(cfg.Fevm)),
```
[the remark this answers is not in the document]
rvagg: yeah, tried clarifying a bit here https://github.com/filecoin-project/lotus/pull/11626

documentation/en/default-lotus-config.toml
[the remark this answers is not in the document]
Stebalien: I agree with a top-level `Events` config option, although I'd keep the ETH-specific options where they are (although... maybe they all generalize?).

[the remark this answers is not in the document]
Stebalien: > Maybe the alternative is to put this in a new Events section alone for now and do a move of the other options at a later date as a cleanup.

Yeah, this sounds like a good idea.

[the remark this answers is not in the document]
jennijuju: Events top level config makes sense. make sure we have decoupled eth api enablement vs. native event subscription enablement, while keeping the coupling between eth api enablement & FEVM event enablement

chain/types/actor_event.go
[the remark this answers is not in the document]
Stebalien: Height is fine. It's kind of incorrect given null blocks, but it is what we use. However, we may want to change the other cases from `Epoch` to `Height` to be consistent.

rvagg:
Notes from zoom chat just now with @Stebalien:

* `Epoch` -> `Height`
* `Height` should take an int
  * `"latest"` and `"earliest"` go and are replaced by:
  * empty/`nil` = latest (or next tipset), from and to, for both get and subscribe
  * `0` meaning actual `0`
  * Warn that in future this may error if you specify a tipset earlier than which we have indexed events for (see below)
* `Prefill` for subscription stays and means ~the same thing, just note that `FromHeight` in the subscription won't go any earlier unless `Prefill` is specified, otherwise it'll be rounded up to latest. (What about `FromHeight` in the future? That should be OK too but need to confirm.)
* Config: top level "Events" with just that one item; future likely migrate most of the Fevm events stuff there.
* TipSetKey instead of TipSetCid

* Later TODO: add a "got events for tipset" for every tipset processed into the events index db, needs a minor migration - version + new table with just "tipset" processed. So we know whether we can provide events (which will be `[]` for earlier tipsets anyway but we don't know when we started processing events; also could we hardwire fevm start date to the earliest here? Figure that out later.).
* Document Addr->AddrID for ethereum events change to come in the future, still be `Address` but will be an ID address in future.

chain/types/actor_event.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// CID of the tipset that contained the message that produced this event.
+	TipSetCid cid.Cid `json:"tipsetCid"`
```
Stebalien, line 68: Discussed sync: Let's use `tipsetKey`, if possible. If that's not possible/efficient, oh well. But let's try to use a key for consistency.

chain/types/actor_event.go
```diff
@@ -0,0 +1,72 @@
...
+	// If the value is an empty slice, the filter will match on the key only, accepting any value.
+	Fields map[string][]ActorEventBlock `json:"fields,omitempty"`
+
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
+	// Optional, default: "latest".
+	FromEpoch string `json:"fromEpoch,omitempty"`
+
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
+	// Optional, default: "latest".
+	ToEpoch string `json:"toEpoch,omitempty"`
```
Stebalien, line 46: Epoch -> Height for consistency.

chain/types/actor_event.go
```diff
@@ -0,0 +1,72 @@
...
+	// For now, this MUST be a Filecoin address.
+	Addresses []address.Address `json:"addresses,omitempty"`
+
+	// Matches events with the specified key/values, or all events if empty.
+	// If the value is an empty slice, the filter will match on the key only, accepting any value.
+	Fields map[string][]ActorEventBlock `json:"fields,omitempty"`
+
+	// Interpreted as an epoch (in hex) or one of "latest" for last mined block, "earliest" for first,
+	// Optional, default: "latest".
+	FromEpoch string `json:"fromEpoch,omitempty"`
```
Stebalien, line 42: As discussed:

- Use integers.
- Omit to mean "latest" in both cases. Or, for subscription, to mean "the next tipset".

chain/types/actor_event.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+type SubActorEventFilter struct {
+	Filter ActorEventFilter `json:"filter"`
+
+	// If true, all available matching historical events will be written to the response stream
+	// before any new real-time events that match the given filter are written.
+	// If `Prefill` is true and `FromEpoch` is set to latest, the pre-fill operation will become a no-op.
+	// if `Prefill` is false and `FromEpoch` is set to earliest, historical events will still be sent to the client.
+	Prefill bool `json:"prefill"`
```
Stebalien, line 28: TODO: remove. Prefill is implied if `FromEpoch` (now `FromHeight`) is non-nil.

chain/types/actor_event.go
```diff
@@ -0,0 +1,72 @@
+package types
+
+import (
+	"github.com/ipfs/go-cid"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+)
+
+type ActorEventBlock struct {
```
Stebalien, line 10: not now: in the future, I'd like to consider automatically translating cbor to json when requested by the user.

chain/types/actor_event.go
```diff
@@ -0,0 +1,72 @@
...
+	// If `TipSetCid` is present in the filter criteria, then neither `FromEpoch` nor `ToEpoch` are allowed.
+	TipSetCid *cid.Cid `json:"tipsetCid,omitempty"`
+}
+
+type ActorEvent struct {
+	// Event entries in log form.
+	Entries []EventEntry `json:"entries"`
+
+	// Filecoin address of the actor that emitted this event.
+	EmitterAddr address.Address `json:"emitter"`
```
Stebalien, line 58: nit: just call it `Emitter`?

Stebalien: We discussed changing this to ID, but decided to punt on that for now.

node/impl/full/actor_events.go
```diff
@@ -0,0 +1,226 @@
...
+					log.Errorf("closing event subscription due to slow reader")
+					return
+				}
+			}
+		}
+
+		in := make(chan interface{}, 256)
+		fm.SetSubChannel(in)
+
+		for {
```
Stebalien, line 154: nit: `for ctx.Err() == nil`. Select _doesn't_ have a priority.

node/impl/full/actor_events.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	fm, err := a.EventFilterManager.Install(ctx, params.MinHeight, params.MaxHeight, params.TipSetCid, f.Filter.Addresses, f.Filter.Fields, false)
+	if err != nil {
+		return nil, err
+	}
+
+	out := make(chan *types.ActorEvent, 25)
```
Stebalien, line 119: Hm. Given that we're not batching by block, this buffer is _tiny_ (likely way too tiny) to simply disconnect the reader if we get stuck.

We likely need some leniency. E.g., the user gets to block for a few milliseconds at a time before we kill the reader. Or, even, the user gets some form of "budget" where they get to block for, say, 10% of the time? I.e., if they get to block for 3 seconds every 30s, they'll never _really_ fall behind.

Or maybe just a larger buffer? I'm just worried about this API may be completely unusable in practice (given any amount of RPC latency).

Stebalien: A somewhat complex algorithm that should "work" (after prefill) is:

1. Read from the event channel into a slice.
2. Record the height of the event from the _newest_ tipset observed. Don't make reduce this height if we hit a revert.
3. If the height increases and the slice isn't empty, disconnect the reader.

Except when the node is "catching up" (probably not something we care about anyways), this will always give the reader a full block time to receive events but won't let them fall behind.

However, the downside is memory. _Ideally_ the event filter manager would return _batches_ of events (the same slice for every subscriber).

Stebalien: In terms of implementation, one usually does this as follows:

```go
var buffer []*types.ActorEvent
for ctx.Err() == nil {
    var ev *types.ActorEvent
    var ok bool
    if len(buffer) > 0 {
        select {
        case ev, ok = <-in:
        case out<-buffer[0]:
            buffer[0] = nil
            buffer = buffer[1:]
            continue
        case <-ctx.Done():
            return
        }
    } else {
        select {
        case ev, ok = <-in:
        case <-ctx.Done():
            return
        }
    }
    if !ok {
        return
    }
    // check if we need to disconnect the client because they've fallen behind.
    // ...
    
    // Otherwise, add the event to the buffer
    buffer = append(buffer, ev)
}
```

Stebalien: But... this is an implementation thing. We _can_ fix this in a followup patch (no API changes).

node/impl/full/eth.go
```diff
@@ -1262,7 +1263,60 @@ func (e *EthEvent) EthGetFilterLogs(ctx context.Context, id ethtypes.EthFilterID
 	return nil, xerrors.Errorf("wrong filter type")
 }
 
-func (e *EthEvent) installEthFilterSpec(ctx context.Context, filterSpec *ethtypes.EthFilterSpec) (*filter.EventFilter, error) {
+func parseBlockRange(heaviest abi.ChainEpoch, fromBlock, toBlock *string, maxRange abi.ChainEpoch) (minHeight abi.ChainEpoch, maxHeight abi.ChainEpoch, err error) {
```
Stebalien, line 1266: As discussed, all of this can be reverted. I think?

node/impl/full/actor_events.go
Stebalien: For pre-filling, we'll likely just need some kind of minimum rate (e.g., one event per millisecond, maybe averaged out over some period of time).

node/impl/full/actor_events.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+var _ ActorEventAPI = (*ActorEventHandler)(nil)
+
+type ActorEventsAPI struct {
+	fx.In
+	ActorEventAPI
+}
+
+func (a *ActorEventHandler) GetActorEvents(ctx context.Context, filter *types.ActorEventFilter) ([]*types.ActorEvent, error) {
```
Stebalien, line 42: We should have a limit here, although we may want to install that limit in the gateway? Do we even expose this API on the gateway?

I.e., I'm thinking:

1. A max number of epochs.
2. A max number of events to collect before aborting.

But, given those limits... we may want to change the API? E.g.:

1. Return an object specifying the range of epochs we returned events from.
2. Guarantee that we'll either return all events from an epoch, or none.

(I'm starting to become more convinced that we should have batched by epoch...).

Stebalien: Really, we likely need:

1. A `Limit` field in the `ActorEventFilter` (max number of events).
2. Something in the _gateway_ that caps the `Limit`.

chain/types/actor_event.go
rvagg: @Stebalien I have a minor mismatch from our discussion with `Prefill`, but I think maybe I'm remembering it wrong since we did a couple of loops. We came to the conclusion that the use-case for "earliest you have" isn't clear so we're doing away with that (eventually to error if you specify a height we don't actually have). Without `Prefill` on the subscribe then I think we can collapse `SubActorEventFilter` and do away with it entirely so the same `ActorEventFilter` is provided to both APIs. It does mean that you must use `FromEpoch` as a signal to decide whether to fish around in historical events or not. The only differences between the two APIs is (1) the `chan` and (2) the ability to reach into the future for _both_ `Height` parameters for `Subscribe` whereas they are both maxxed at the current heaviest for `Get`.

chain/types/actor_event.go
rvagg: Need to decide what to do if this and/or `From` will cause an error if they are later than the current heaviest. If we're going to do that with earliest eventually then it may be something to consider. It just means you can't be sloppy with your calls other than `nil`.

node/impl/full/actor_events.go
rvagg: Already have the latter, `cfg.Fevm.Events.MaxFilterHeightRange` exists already as part of fevm events and we're piggy backing that, defaults to 2880. Because it's old, it's going to be difficult to move it to the new top-level `Events` object but it would be nice for them to sit together ... maybe we need to consider a config "migration", or at least having a config unifying thing run at startup to support old + new for these events values.

chain/types/actor_event.go
Stebalien: > I think we can collapse SubActorEventFilter and do away with it entirely so the same ActorEventFilter is provided to both APIs.

Yes.

> It does mean that you must use FromEpoch as a signal to decide whether to fish around in historical events or not

Yes.

> The only differences between the two APIs is

That and, for "get" `FromEpoch` and `ToEpoch` default to "latest". For "subscribe", they default to "next" (i.e., the epoch after the current epoch. Subtle difference that nobody's going to care about but still technically a difference.

node/impl/full/actor_events.go
Stebalien: That's probably fine? But then we need to make sure to apply that in the gateway proxy for the events API.

Or are we applying that deep in the system? If so... ugh, that's not how we do things... but oh well.

chain/types/actor_event.go
Stebalien: > If we're going to do that with earliest eventually then it may be something to consider.

I'd say yes? You can just specify "nil" if you don't care.

node/impl/full/actor_events.go
rvagg: No, thankfully not, it's at the parse-inputs phase, which is one main purpose of the `parseBlockRange`, which itself was an extraction of existing code for fevm event filter code.

What I'm doing now is keeping that function intact for the fevm case so we leave that functionality as is, but implementing a simpler one for `parseHeightRange` for this API that applies the same limit but also gets us from inputs to the internal filter representation. That turns out to be a bit tricky to unify because the `"earliest"` and `"latest"` stuff comes from the existing fevm events filter API. So I'm just trying to make sure tests cover the edges of the conversions.

node/impl/full/actor_events.go
rvagg: @Stebalien a case I've hit that I'm unsure about in terms of what patterns to follow with existing APIs - what about the case where we can clearly identify a halting condition, do we close the channel? If I specify a `ToHeight` that's 100 epochs in the future (or even do a subscribe that has the range entirely historical), does the API terminate itself? It seems like _yes_ is the right answer to this, at least that's what I'd expect.

Stebalien: > If I specify a ToHeight that's 100 epochs in the future

You mean on subscribe? Ideally you'd terminate when hitting the target height. Although it would nice to be able to signal success/error... i think that's a limitation of our JSON-RPC API more than anything.

node/impl/full/actor_events.go
```diff
@@ -6,14 +6,14 @@ import (
 	"time"
 
 	"github.com/ipfs/go-cid"
+	"github.com/raulk/clock"
```
Stebalien, line 9: Gah.... upstream is now unmaintained and we have two forks of this library, where unmaintained upstream has nice features like `clock.WithTimeout(ctx, timeout)`.

Upstream: https://github.com/benbjohnson/clock

Now I'm going to go and update raul's fork...

rvagg: yeah, I would have used benbjohnson/clock, but that got dropped last year; raulk/clock is used throughout lotus so I just went with it

Stebalien: You know, upstream works (even if it's archived). IMO, we should just use it. https://github.com/filecoin-project/lotus/issues/11677

node/impl/full/actor_events.go
```diff
@@ -321,30 +338,30 @@ func (a *ActorEventHandler) SubscribeActorEvents(ctx context.Context, evtFilter
...
+					if params.MaxHeight > 0 && currentHeight > params.MaxHeight {
+						// we've reached the filter's MaxHeight, we're done so we can close the channel
 						return
 					}
 				}
 			}
+
+			if a.clock.Now().After(nextBacklogHeightUpdate) {
+				minBacklogHeight = a.chain.GetHeaviestTipSet().Height() - 1
+				nextBacklogHeightUpdate = a.clock.Now().Add(a.blockDelay)
```
Stebalien, line 351: I wouldn't bother with time here. Just wait till we see a new event, and check the height.

node/impl/full/actor_events.go
```diff
@@ -195,54 +237,41 @@ func (a *ActorEventHandler) SubscribeActorEvents(ctx context.Context, evtFilter
...
+					timer.Stop()
 					return
 				}
 			}
+			timer.Stop()
+		}
+
+		// for the case where we have a MaxHeight set, we don't get a signal from the filter when we
+		// reach that height, so we need to check it ourselves, do it now but also in the loop
+		if params.MaxHeight > 0 && a.chain.GetHeaviestTipSet().Height() > params.MaxHeight {
```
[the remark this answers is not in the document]
rvagg: This point is the same for the `nextBacklogHeightUpdate` comment below: in the case of a restrictive filter, we may not get an event to use for our check for a long time, or ever, so we can't rely on looking at new events to check for height. e.g. the itest in here won't terminate if we don't have these checks because it does a subscribe with a height termination condition and no further activity. This is a reasonable case when you consider how specific a filter can be.

node/impl/full/actor_events.go
```diff
@@ -321,30 +338,30 @@ func (a *ActorEventHandler) SubscribeActorEvents(ctx context.Context, evtFilter
 				case <-ctx.Done():
 					return
 				case <-ticker.C:
-					if params.MaxHeight > 0 && a.Chain.GetHeaviestTipSet().Height() > params.MaxHeight {
+					currentHeight := a.chain.GetHeaviestTipSet().Height()
+					if params.MaxHeight > 0 && currentHeight > params.MaxHeight {
```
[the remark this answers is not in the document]
rvagg: yep, I think this could be dealt with if we wanted to mess with the `EventFilterManager` but I'm hesitant to get too much more tangled up in there

it would also be helpful to batch by height, as you've suggested, but that would be done in `EventFilterManager` too

node/impl/full/actor_events.go
```diff
@@ -195,54 +237,41 @@ func (a *ActorEventHandler) SubscribeActorEvents(ctx context.Context, evtFilter
...
-			if err := a.EventFilterManager.Remove(ctx, fm.ID()); err != nil {
+			if err := a.eventFilterManager.Remove(ctx, fm.ID()); err != nil {
 				log.Warnf("failed to remove filter: %s", err)
 			}
 		}()
 
+		// When we start sending real-time events, we want to make sure that we don't fall behind more
+		// than one epoch's worth of events (approximately). Capture this value now, before we send
+		// historical events to allow for a little bit of slack in the historical event sending.
+		minBacklogHeight := a.chain.GetHeaviestTipSet().Height() - 1
```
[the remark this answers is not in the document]
rvagg: it's not `panic` racy though, right? it just needs to be approximate, we need an end condition and if it ends up being 2, or even 3 epochs behind rather than 1 then that's OK, as long as there's a termination condition

node/impl/full/actor_events.go
Stebalien: Ah, I see what you mean...

node/impl/full/actor_events.go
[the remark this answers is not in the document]
Stebalien: Discussing it below.

node/impl/full/actor_events.go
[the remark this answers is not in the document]
Stebalien: Hm. You're right. Can we check _before_ we start the filter? My concern here and above is that we:

1. Install the filter.
2. _Then_ we get the tipset height.

The tipset could advance in the meantime and we could end up missing some events.

I mean, we could miss events _anyways_ due to reorgs, but that's less likely. So we're probably good (after filing an issue) to land this as long as we check the tipset height first.

node/impl/full/actor_events.go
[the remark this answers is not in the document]
Stebalien: You're right.

Stebalien (approved):
LGTM with:

1. File an issue to fix the timing stuff.
2. Get the initial "heaviest tipset" before we subscribe to reduce the race a bit.

api/api_full.go
```diff
@@ -906,6 +906,28 @@ type FullNode interface {
 
 	RaftState(ctx context.Context) (*RaftStateData, error) //perm:read
 	RaftLeader(ctx context.Context) (peer.ID, error)       //perm:read
+
+	// Actor events
+
+	// GetActorEvents returns all FVM and built-in Actor events that match the given filter.
```
jennijuju, line 912:
```suggestion
	// GetActorEvents returns all FEVM user actor and built-in Actor events that match the given filter.
```

(technically everything on filecoin is executed by FVM 😄 i call this a side effect of too-good-of-branding 😸

api/api_full.go
```diff
@@ -906,6 +906,28 @@ type FullNode interface {
...
+	// This is a request/response API.
+	// Results available from this API may be limited by the MaxFilterResults and MaxFilterHeightRange
+	// configuration options and also the amount of historical data available in the node.
+	GetActorEvents(ctx context.Context, filter *types.ActorEventFilter) ([]*types.ActorEvent, error) //perm:read
+
+	// SubscribeActorEvents returns a long-lived stream of all FVM and built-in Actor events that
```
jennijuju, line 918:
```suggestion
	// SubscribeActorEvents returns a long-lived stream of all FEVM user actor and built-in Actor events that
```
```suggestion
	// SubscribeActorEvents returns a long-lived stream of all FVM and built-in Actor events that
```

chain/events/filter/event.go
```diff
@@ -27,22 +27,32 @@ func isIndexedValue(b uint8) bool {
...
+
+	TakeCollectedEvents(context.Context) []*CollectedEvent
+	CollectEvents(context.Context, *TipSetEvents, bool, AddressResolver) error
+}
+
+type eventFilter struct {
+	id        types.FilterID
+	minHeight abi.ChainEpoch // minimum epoch to apply filter or -1 if no minimum
+	maxHeight abi.ChainEpoch // maximum epoch to apply filter or -1 if no maximum
+	tipsetCid cid.Cid
```
jennijuju, line 43: can you remind me how can one get TipsetCid? (which is different from tipset keys CID?

chain/events/filter/event.go
```diff
@@ -27,22 +27,32 @@ func isIndexedValue(b uint8) bool {
...
+type EventFilter interface {
+	Filter
+
+	TakeCollectedEvents(context.Context) []*CollectedEvent
+	CollectEvents(context.Context, *TipSetEvents, bool, AddressResolver) error
+}
+
+type eventFilter struct {
+	id        types.FilterID
+	minHeight abi.ChainEpoch // minimum epoch to apply filter or -1 if no minimum
```
jennijuju, line 41: Nit - shouldn't we use CalmelCase here?

chain/events/filter/event.go
```diff
@@ -27,22 +27,32 @@ func isIndexedValue(b uint8) bool {
...
+	maxHeight abi.ChainEpoch // maximum epoch to apply filter or -1 if no maximum
+	tipsetCid cid.Cid
+	addresses []address.Address // list of f4 actor addresses that are extpected to emit the event
```
jennijuju, line 44: Why is this only applicable to f4 actors..?

chain/events/filter/event.go
rvagg: This type is identical to what was there before, just made private and the public one is replaced with an interface. lower-case camelCase seems to be standard in this package, for private types. 🤷

jennijuju: oh is it a go thing for interfaces vs struct?

documentation/en/api-v1-unstable-methods.md
```diff
@@ -8758,6 +8825,75 @@ Response:
 }
 ```
 
+## Subscribe
+
+
+### SubscribeActorEvents
+SubscribeActorEvents returns a long-lived stream of all FVM and built-in Actor events that
```
jennijuju, line 8832:
```suggestion
SubscribeActorEvents returns a long-lived stream of all FEVM user acor and built-in Actor events that
```

chain/events/filter/event.go
rvagg: `TipSetKey#Cid()` does it, bundles up the raw bytes of each CID, concatenates them, and makes a CID from those bytes.

This internal API is serving the existing Eth filter API which can take an Eth `BlockHash`, in which case it converts that to a tipset CID from the internal index, and it's now serving this new API which takes a `TipSetKey` which it does a `.Cid()` on to get the CID.

wew .. that's not confusing at all!

chain/events/filter/event.go
rvagg: outdated comment from eth-specific api, removing

chain/events/filter/event.go
rvagg: public vs private, none of these fields are intended to be used outside of this package, although with a private struct they can't be anyway - so they could be either way here and it wouldn't make a difference; but looking in this package all private structs have lower-case field names so I'll stick with that

api/api_full.go
rvagg: "user-programmed and built-in actor events"

jennijuju (approved):

PR #11623, opened 2024-02-14
fix: state: remove watermelon migration hack

Stebalien:
## Proposed Changes

Remove migration cache skip hack from the StateManager.

We did this to avoid using cached migrations in the borked watermelon upgrade on calibrationnet, but we don't need to keep this in the code (even for users syncing from scratch).

## Checklist

- [x] CI is green


rvagg (approved):
I don't know the history here but this seems fine to me given the description

PR #11626, opened 2024-02-19
Rename internal events modules for clarity

rvagg:
Based on feedback @ https://github.com/filecoin-project/lotus/pull/11618#discussion_r1488356419 I thought that it worth doing some naming clarification. The other change I want to make but it would make for a noisy diff, is to move the `EthEventHandler` out of eth.go and into eth_events.go but that's not a big deal.

The clarity mostly shows up in node/builder_chain.go when it all comes together.


rvagg:

PR #11627, opened 2024-02-19
Avoid duplicate messages when looking for receipts

rvagg:
From CI failure in https://github.com/filecoin-project/lotus/pull/11626

I think it's possible to get the same tipset for a height, somehow; or maybe it's too early to be getting the final tipset for the head's height? I'd like an explanation for this if someone has it.


rvagg:

PR #11628, opened 2024-02-19
Add `unittests` make target

rvagg:
Is there a better way of doing this currently that I'm not seeing? I can see in the CircleCI config that we're explicitly listing packages and grouping them to be run on different machines.


Stebalien (approved):
I usually just run relevant unit tests directly (and let CI run "everything"). But this seems reasonable.

PR #11629, opened 2024-02-19
Extract a "Type" from an event log if possible

rvagg:
This is one of the improvements for actor events that I suggested in https://github.com/filecoin-project/lotus/pull/11540#issuecomment-1933690898. Where the event log entries have a `"$type"` entry (at least one, picking the first), we assume CBOR, decode it and use it as a `Type` on the top level `ActorEvent` object we return to the user. It makes it easier to deal with these as a user if you have more than one type coming out and want to do something specific with them. Otherwise you're stuck doing this yourself to find the event types you want and it's kind of hostile because (a) it's log form and (b) it's CBOR.


rvagg:

rvagg:
Going to close this for now, maybe revisit later as a usability enhancement. Steb suggested a preference for KISS and let the user deal with this, at least for now.

Issue #11614, opened 2024-02-08
Add Verifreg APIs to support tooling migration 


jennijuju:
Many ecosystem tooling has being tracking datacap activities by monitoring f05 via StateMarketDeals, instead of use the verifreg state. It has been fine given all fil+ pieces can only be onboarded via f05, however DDO will change that in the upcoming nv22. We want to help toolings to migrate their system by exposing APIs that allow them to the datacap activities, and to be more precise, all allocation/claims. 

There are two options we can consider:
- Option 1: adding an API that returns all verified clients, a.k.a all the ActorIds that holds datacap (can get from f07 state https://github.com/helix-onchain/filecoin/blob/666ac555876ec60370f507f8cded17d4af2ece85/frc46_token/src/token/state.rs#L115). The get all allocations & claims for each client via StateGetAllocations & StateGetClaims.
- Option 2: adding 2 new APIs that returns **all** allocations/claims: StateGetAllAllocations & StateGetAllClaims but reading through and returning `allocations` and `claims` from f06. (similar to  StateGetAllocations/StateGetClaims just without a given client address)

rvagg:
@jennijuju @rjan90 maybe this issue should close/replace these:

* https://github.com/filecoin-project/lotus/issues/11249
* https://github.com/filecoin-project/lotus/issues/11173 (some good background context in here tho)

willscott:
Do we have a timeline for when we can an expected API for this functionality where be formalized and when it might become available?

tooling will really want access to this without digging into verifreg state manually before NV22 or we risk disruption of dashboards. We need some time to update things on top of it as well.

rvagg:
I'll take a look into this next week as I continue work on related things.

> tooling will really want access to this without digging into verifreg state manually

Can you give examples of the types of access that'll be needed? From what I understand, this kind of information would already be collected and stored in queryable form by Lily, and consumers of that information don't seem to want to be querying Lotus directly for additional information. I wouldn't mind finding direct consumers of Lotus APIs and understand their current workflows to figure out what appropriate augmentation for DDO looks like. This has been one of the challenges to date—finding users that want queryable APIs for this stuff and not just "store messages/receipts in my own format and query later" - or a variation "download relevant state snapshot as a json file from glif and query it in my own way".

willscott:
[...]

PR #11631, opened 2024-02-19
feat: api: new verified registry methods to get all allocations and claims

LexLuthr:
## Related Issues

Fixes https://github.com/filecoin-project/lotus/issues/11614

## Proposed Changes

- This PR create 2 new methods in verified registry actor called `GetAllAllocations and GetAllClaims` which lists all the available allocations and claims in the actor.
- These methods are further plugged into Lotus API methods `StateGetAllAllocations` and `StateGetAllClaims`.
- Lotus CLI has been updated to get the same information.
- ~~Lotus geteway API has also been updated to support the new methods.~~

## Checklist

- [x] CI is green


itests/verifreg_test.go
```diff
@@ -503,3 +503,140 @@ func makeVerifier(ctx context.Context, t *testing.T, api *impl.FullNodeAPI, root
...
+
+	// check datacap balance
+	dcap, err = api.StateVerifiedClientStatus(ctx, verifiedClientAddr, types.EmptyTSK)
+	require.NoError(t, err)
+	require.Nil(t, dcap)
+
+	allocations, err := api.StateGetAllAllocations(ctx, types.EmptyTSK)
+	require.NoError(t, err)
+
+	require.Equal(t, 2, len(allocations))
```
rvagg, line 641: would be good to check the returns here rather than just the length, maybe give them different piece CIDs and check those You could use `baga6ea4seaaqa` and `baga6ea4seaaqc` which are ~valid piece CIDs with different single byte digests.

There's also an itest in the actor events PR that'll introduce the whole allocation + claim lifecycle that we can piggy-back when we have both of these merged to check both allocations and claims at different stages.

cli/filplus.go
```diff
@@ -799,3 +801,149 @@ var filplusSignRemoveDataCapProposal = &cli.Command{
 		return nil
 	},
 }
+
+var filplusListAllAllocationsCmd = &cli.Command{
```
rvagg, line 805: Not really suggestions, just thoughts inviting discussion: These could be integrated into `list-allocations` and `list-claims` just by making the client optional. Also as they are, I'm not sure the `--expired` makes as much sense for these `all` listings as stand-alone commands.

rvagg:
lgtm, with some suggestions

cli/filplus.go
LexLuthr: I has a similar initial though but I wasn't sure if that would make things easier or more tricky. I will update the CLI as per your suggestion.

itests/verifreg_test.go
LexLuthr: @rvagg Can you please provide me the link to the said PR? I can wait till it lands and create a more inclusive itest for this. Meanwhile, I will add the more detailed checks for the allocations.

cli/filplus.go
LexLuthr: I have updated the CLI and added a JSON flag to allow json output as well.

itests/verifreg_test.go
rvagg: https://github.com/filecoin-project/lotus/pull/11618 is the latest iteration of it with a cleaner commit history (https://github.com/filecoin-project/lotus/pull/11540 has the gory details of the process)

It's much bigger in scope than this PR so I wouldn't want to hold this one up with that. You're welcome to rebase onto it to do the itest but I think it might just be best to notch this up as a TODO for then.

rvagg (approved):
lgtm, the only thing I'd note for other reviewers is that this is a breaking change in that the `filplus list-allocations` and `filplus list-claims` CLI commands are going to output tables with `ClaimID` and `AllocationID` instead of just `ID` now. I'd assume that this is minor enough, and also an improvement, so it shouldn't be a problem for people to adapt to if they're using this output programatically.

chain/actors/builtin/verifreg/state.go.template
```diff
@@ -145,6 +145,21 @@ func (s *state{{.v}}) GetAllocations(clientIdAddr address.Address) (map[Allocati
 {{end}}
 }
 
+func (s *state{{.v}}) GetAllAllocations() (map[AllocationId]Allocation, error) {
+{{if (le .v 12)}}
```
rvagg, line 149:
```suggestion
{{if (le .v 8)}}
```

chain/actors/builtin/verifreg/state.go.template
```diff
@@ -170,6 +185,22 @@ func (s *state{{.v}}) GetClaims(providerIdAddr address.Address) (map[ClaimId]Cla
 {{end}}
 }
 
+func (s *state{{.v}}) GetAllClaims() (map[ClaimId]Claim, error) {
+{{if (le .v 12)}}
```
rvagg, line 189:
```suggestion
{{if (le .v 8)}}
```

rvagg: Sorry, late review but after trying this out I realised this could be supported all the way back to v9, so there's no point in holding it off until we get to v13.

itests/verifreg_test.go
LexLuthr: I have added a TODO to the test. I can update it later once events PR is in place.

Stebalien:
@LexLuthr could you add a quick note in the CHANGELOG.md about the changes to the `filplus list-allocations` and `filplus list-claims` command output? With that it should be good to go.

rvagg:
@willscott FYI.

After a bit of messing around (long story) I managed to use these APIs to get some data. Mainnet right now has allocations as a ~190Mb JSON file and claims as a ~12Gb JSON file.

Allocations is a map of the allocation ID to the `Allocation` struct:

```json
{
  "1": {
    "Client": 215074,
    "Provider": 1353593,
    "Data": {
      "/": "baga6ea4seaqimrrmdbqi6yomnkjpwonkeuvmcmqzgf4dnosdsagetxcddify4di"
    },
    "Size": 17179869184,
    "TermMin": 1463040,
    "TermMax": 1555200,
    "Expiration": 2414101
  },
  "10": {
    "Client": 215074,
    "Provider": 1353593,
    "Data": {
      "/": "baga6ea4seaqiix3rs3p3qkkitpggsuq276qwxz7mu22mx35wxq35snt7l3dk2di"
    },
    "Size": 17179869184,
    "TermMin": 1463040,
    "TermMax": 1555200,
    "Expiration": 2414101
  },
...
```

Claims is a map of claim ID to the `Claim` struct:

```json
{
  "100000": {
    "Provider": 1970213,
    "Client": 1948518,
    "Data": {
      "/": "baga6ea4seaqhmk73sc4kewlkdwhxlrvr3eyqyvmky7hhr2v2shqjmrlylgk22fi"
    },
    "Size": 34359738368,
    "TermMin": 1512000,
    "TermMax": 1771200,
    "TermStart": 2385954,
    "Sector": 110733
  },
  "1000000": {
    "Provider": 1964215,
    "Client": 1940441,
    "Data": {
      "/": "baga6ea4seaqdiugqdshc7ejkn2rypywiuth4auvx5bijf4hhll7hbswsxpk5mbq"
    },
    "Size": 34359738368,
    "TermMin": 1541862,
    "TermMax": 1801062,
    "TermStart": 2417926,
    "Sector": 11634
  },
...
```

Should be enough info there for piece accounting. But it's a very hefty way to consume it. Thankfully it's straight out of state with very little munging so it's not too inefficient to dump if you have the resources.

Events are going to be the most efficient way to consume these though - maybe start with a snapshot of this then consume events from that point on to update your local mirror of claim state. See https://github.com/filecoin-project/lotus/pull/11618 which has an itest with both `allocation` and `claim` events.

api/v0api/gateway.go
```diff
@@ -65,8 +65,10 @@ type Gateway interface {
 	StateGetAllocationForPendingDeal(ctx context.Context, dealId abi.DealID, tsk types.TipSetKey) (*verifregtypes.Allocation, error)
 	StateGetAllocation(ctx context.Context, clientAddr address.Address, allocationId verifregtypes.AllocationId, tsk types.TipSetKey) (*verifregtypes.Allocation, error)
 	StateGetAllocations(ctx context.Context, clientAddr address.Address, tsk types.TipSetKey) (map[verifregtypes.AllocationId]verifregtypes.Allocation, error)
+	StateGetAllAllocations(ctx context.Context, tsk types.TipSetKey) (map[verifregtypes.AllocationId]verifregtypes.Allocation, error)
 	StateGetClaim(ctx context.Context, providerAddr address.Address, claimId verifregtypes.ClaimId, tsk types.TipSetKey) (*verifregtypes.Claim, error)
 	StateGetClaims(ctx context.Context, providerAddr address.Address, tsk types.TipSetKey) (map[verifregtypes.ClaimId]verifregtypes.Claim, error)
+	StateGetAllClaims(ctx context.Context, tsk types.TipSetKey) (map[verifregtypes.ClaimId]verifregtypes.Claim, error)
```
Stebalien, line 71: These new endpoints return too much data at once to safely add to the gateway (requires too much buffering). Do you _need_ these on an API gateway? If so, we need to make them stream their results back via a channel.

willscott: I would expect these to end up being accessed in the same way as statemarketdeals is today - e.g. https://lotus.filecoin.io/lotus/developers/glif-nodes/

jennijuju: I agree. you are right -  I mixed glifs customized  end point  for statemarketdeal vs their gateway endpoint in my original issue. @LexLuthr could you please remove gateways? Thanks

LexLuthr: I have removed the gateway APIs.

LexLuthr:
@Stebalien I have updated the CHANGELOG.md with API and CLI changes.

cli/filplus.go
```diff
@@ -334,14 +400,81 @@ var filplusListClaimsCmd = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
-		providerAddr, err := address.NewFromString(cctx.Args().Get(0))
-		if err != nil {
-			return err
-		}
+		writeOut := func(tsHeight abi.ChainEpoch, claims map[verifreg.ClaimId]verifreg.Claim, json, expired bool) error {
+			// Map Keys. Corresponds to the standard tablewriter output
+			allocationID := "ClaimID"
```
rvagg, line 405:
```suggestion
			claimID := "ClaimID"
```

cli/filplus.go
```diff
@@ -334,14 +400,81 @@ var filplusListClaimsCmd = &cli.Command{
...
+			data := "Data"
+			size := "Size"
+			tMin := "TermMin"
+			tMax := "TermMax"
+			tStart := "TermStart"
+			sector := "Sector"
+
+			// One-to-one mapping between tablewriter keys and JSON keys
+			tableKeysToJsonKeys := map[string]string{
+				allocationID: strings.ToLower(allocationID),
```
rvagg, line 417:
```suggestion
				claimID: strings.ToLower(claimID),
```

cli/filplus.go
```diff
@@ -334,14 +400,81 @@ var filplusListClaimsCmd = &cli.Command{
...
+				tMin:         strings.ToLower(tMin),
+				tMax:         strings.ToLower(tMax),
+				tStart:       strings.ToLower(tStart),
+				sector:       strings.ToLower(sector),
+			}
 
-		providerIdAddr, err := api.StateLookupID(ctx, providerAddr, types.EmptyTSK)
-		if err != nil {
-			return err
+			var allocs []map[string]interface{}
```
rvagg, line 428:
```suggestion
			var claims []map[string]interface{}
```

cli/filplus.go
```diff
@@ -334,14 +400,81 @@ var filplusListClaimsCmd = &cli.Command{
...
+
+			for key, val := range claims {
+				if tsHeight > val.TermMax || !expired {
+					alloc := map[string]interface{}{
+						allocationID: key,
```
rvagg, line 433:
```suggestion
					claim := map[string]interface{}{
						claimID: key,
```

cli/filplus.go
```diff
@@ -334,14 +400,81 @@ var filplusListClaimsCmd = &cli.Command{
...
+						provider:     val.Provider,
+						client:       val.Client,
+						data:         val.Data,
+						size:         val.Size,
+						tMin:         val.TermMin,
+						tMax:         val.TermMax,
+						tStart:       val.TermStart,
+						sector:       val.Sector,
+					}
+					allocs = append(allocs, alloc)
```
rvagg, line 443:
```suggestion
					claims = append(claims, claim)
```

cli/filplus.go
```diff
@@ -334,14 +400,81 @@ var filplusListClaimsCmd = &cli.Command{
...
+				var jsonAllocs []map[string]interface{}
+				for _, alloc := range allocs {
+					jsonAlloc := make(map[string]interface{})
+					for k, v := range alloc {
+						jsonAlloc[tableKeysToJsonKeys[k]] = v
+					}
+					jsonAllocs = append(jsonAllocs, jsonAlloc)
+				}
+				// then return this!
+				return PrintJson(jsonAllocs)
```
rvagg, line 458:
```suggestion
				// get a new list of claims with json keys instead of tablewriter keys
				var jsonClaims []map[string]interface{}
				for _, claim := range claims {
					jsonClaim := make(map[string]interface{})
					for k, v := range claim {
						jsonAlloc[tableKeysToJsonKeys[k]] = v
					}
					jsonClaims = append(jsonClaims, jsonClaim)
				}
				// then return this!
				return PrintJson(jsonClaims)
```

cli/filplus.go
```diff
@@ -334,14 +400,81 @@ var filplusListClaimsCmd = &cli.Command{
...
+			}
+			// Init the tablewriter's columns
+			tw := tablewriter.New(
+				tablewriter.Col(allocationID),
```
rvagg, line 462:
```suggestion
				tablewriter.Col(claimID),
```

cli/filplus.go
```diff
@@ -334,14 +400,81 @@ var filplusListClaimsCmd = &cli.Command{
...
+				tablewriter.Col(size),
+				tablewriter.Col(tMin),
+				tablewriter.Col(tMax),
+				tablewriter.Col(tStart),
+				tablewriter.NewLineCol(sector))
+			// populate it with content
+			for _, alloc := range allocs {
+
+				tw.Write(alloc)
+			}
```
rvagg, line 475:
```suggestion
			for _, claim := range claims {
				tw.Write(claim)
			}
```

cli/filplus.go
```diff
@@ -253,14 +259,77 @@ var filplusListAllocationsCmd = &cli.Command{
...
+				tablewriter.Col(client),
+				tablewriter.Col(provider),
+				tablewriter.Col(pieceCid),
+				tablewriter.Col(pieceSize),
+				tablewriter.Col(tMin),
+				tablewriter.Col(tMax),
+				tablewriter.NewLineCol(expr))
+			// populate it with content
+			for _, alloc := range allocs {
+
```
rvagg, line 328:
```suggestion
```

cli/filplus.go
```diff
@@ -253,14 +259,77 @@ var filplusListAllocationsCmd = &cli.Command{
...
+						tMin:         val.TermMin,
+						tMax:         val.TermMax,
+						expr:         val.Expiration,
+					}
+					allocs = append(allocs, alloc)
+				}
+			}
+
+			if json {
+				// get a new list of wallets with json keys instead of tablewriter keys
```
rvagg, line 304:
```suggestion
				// get a new list of allocations with json keys instead of tablewriter keys
```

rvagg (changes requested):
sorry, just noticed these

LexLuthr:
@rvagg This was entirely my fault. I created the JSON output for Allocations and forgot to rename the internals for claims. I have updated the internals for claims.

rvagg (approved):
fine by me

I don't have an opinion on the json output but I'll note that it'll be different from an API call's output which is what I've pasted the example of above, i.e. a map with claim/allocation ID as the key rather than an array. Maybe it's good to have alternative output? :shrug: Maybe we'll find out what's actually wanted.

Stebalien (approved):

PR #11641, opened 2024-02-22
Add systemd memory note on install and in config

rvagg:
I got caught by this without noticing for a couple of weeks, wondering why my node was behaving so difficult when I tried to do anything non-trivial. I'm also wondering if raising this in general might be a good idea at this point since it seems to be pushing the limit with the size of the blockstore even with minimal history.

### DIFF
--- a/scripts/lotus-daemon.service
+++ b/scripts/lotus-daemon.service
@@ -11,6 +11,8 @@ Restart=always
 RestartSec=10
 
 MemoryAccounting=true
+# Depending on the complexity of API usage of this node, you may need to adjust the following
+# memory options. Please monitor the node's performance and adjust accordingly.
 MemoryHigh=8G
 MemoryMax=10G
 LimitNOFILE=8192:10240
--- a/Makefile
+++ b/Makefile
@@ -246,7 +246,10 @@ install-daemon-service: install-daemon
 	install -C -m 0644 ./scripts/lotus-daemon.service /etc/systemd/system/lotus-daemon.service
 	systemctl daemon-reload
 	@echo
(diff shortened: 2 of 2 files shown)


scripts/lotus-daemon.service
```diff
@@ -11,6 +11,8 @@ Restart=always
 RestartSec=10
 
 MemoryAccounting=true
+# Depending on the complexity of API usage of this node, you may need to adjust the following
+# memory options. Please monitor the node's performance and adjust accordingly.
 MemoryHigh=8G
 MemoryMax=10G
```
rvagg, line 17:
```suggestion
MemoryHigh=16G
MemoryMax=20G
```

:shrug: my node was struggling to even stay in sync so it would be good to lift this at bit. Suggestions? 12/16 maybe as a compromise if we're still aiming for laptop installs?

rvagg: Even just having a `MemoryHigh` is probably causing heavy performance degradation aside from the memory limitation since the process doesn't get to know this pressure is being applied except that allocation calls are slower and reclamation efforts are more aggressive. Ideally this would be removed entirely but it does get very hungry on some of the big API calls.

rvagg:
@f8-ptrk suggested on Slack that we just remove this entirely because they reflect ancient times when these limits were practical. I'm happy to do that, maybe leave a note in there with them commented out suggesting they can be used if it's too much of a hog. Thoughts?

rvagg:
:thumbsup: done. I've removed the note from Makefile that I had in there, we have to assume people are running this and are fairly serious about allocating resources. I've left the Makefile print formatting though. Note left in the service config with those options commented out. I left the fd limits, I think they're probably high enough as they are unless someone thinks otherwise.

Stebalien (approved):
Makes sense. This will affect docker images but... users should apply memory limits to docker itself.

PR #11647, opened 2024-02-22
harmony storage

snadrus:
## Related Issues
This enables HarmonyTask tasks to express (kind-of-cleanly) the storage needs, the system to atomically reserve each need, and provide the paths to Do() without any API change to existing tasks. 

## Proposed Changes
- Storage is an interface-like struct on TaskTypeDetails that's used to ensure tasks have what they need
- Storagemgr, A helper, manages locking and naming. 

## Checklist

- [ ] CI is green


lib/harmony/resources/storagemgr/storagemgt.go
```diff
@@ -0,0 +1,251 @@
...
+		if strings.Contains(p.LocalPath, "cache") {
+			t = UsageCache
+		}
+		if strings.Contains(p.LocalPath, "tmp") {
+			t = UsageTemporary
+		}
+		purposes[t] = append(purposes[t], path(p.LocalPath))
+	}
+
+	return &StorageMgr{}
```
magik6k, line 68: https://github.com/filecoin-project/lotus/tree/master/storage/sealer#package-paths has some rough description of how the storage subsystem roughly works.

Basically there can be multiple storage paths attached to a given node, and each storage path can hold one of 5 "file types" (sealed data, unsealed data, etc.)

Then each of those paths can say whether it wants to support Sealing on Long Term storage (neither and both are valid combinations)

Here we're mostly reimplementing what exists in https://github.com/filecoin-project/lotus/blob/master/storage/paths/local.go#L402, and in general what the Local paths subsystem is doing, and I'd really rather not do that given how much battle testing that part of the code has seen - e.g. there are no syscalls in the calls in the calls relevant to scheduling, because those will hang (e.g. try to start 10 32G mv-s on an NFS share, then see `df -h` hang; dying but not dead yet ssds/hdds are also similar)

The existing paths.Local Reserve mechanism also is going to send the reservation stats to the sector index, which then will show that information in the storage cli / prometheus / will be query-able from postgres;

Both impls are missing one thing - paths can be shared between nodes - e.g. one can have one GPU box with 10 GPUs, one raided nvme mount - then neither of the systems will be able to account properly for reserved storage in the shared mount because there are 10 processes with independent in-memory accounting.

(but now that we have the db storage index it should be reasonably easy to change paths.Local to do the reservation accounting through the database)

lib/harmony/resources/storagemgr/storagemgt.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+}
+
+type consumption struct {
+	Location string
+	Need     uint64
+}
+
+func getJSON(path path) []consumption {
+	b, err := os.ReadFile(string(path) + "/claims.json")
```
magik6k, line 101: * Should use filepath.Join (rest of the storage subsystem does pretty religously, and that will make our lives so much easier when we'll want to build super light storage-only SaaS backed thing for windows)

magik6k: Also I'd avoid putting anything in the scheduling codepath that makes any syscalls to the storage paths; Those will be backed by devices which are constantly hammered, and you will see basic things like statfs hang for 30 minutes

IMO reservations should live in the DB, addressed like the rest of the storage subsystem by existing pathIDs (or just use/interface with the existing storage subsystem which has reservation mechanism)

lib/harmony/resources/storagemgr/storagemgt.go
```diff
@@ -0,0 +1,251 @@
...
+	claimsMx.Lock()
+	defer claimsMx.Unlock()
+
+	j := getJSON(path)
+	j = append(j, consumption)
+	b, err := json.Marshal(j)
+	if err != nil {
+		return "", xerrors.Errorf("error encoding claims: %w", err)
+	}
+	err = os.WriteFile(string(path)+"/claims.json", b, 0644)
```
magik6k, line 168: Nit: I'd put `claims.json` in a global const

lib/harmony/resources/resources.go
```diff
@@ -24,6 +24,20 @@ type Resources struct {
 	Gpu       float64
 	Ram       uint64
 	MachineID int
+	Storage
+}
+
+// Optional Storage management.
+// See storagemgr/storagemgt.go for more details.
+type Storage struct {
```
magik6k, line 32: I do like how abstract this interface is, probably could be just implemented with the existing storage subsystem.

In .Do how do I get the storage path I'm supposed to use?

lib/harmony/harmonytask/task_type_handler.go
```diff
@@ -116,6 +118,15 @@ top:
 		}
 	}
 
+	var location string
+	if c := h.TaskTypeDetails.Cost.Storage.Claim; c != nil {
+		if location, err = c(int(*tID)); err != nil {
```
magik6k, line 123: This won't work nicely with paths shared between nodes:
* Start 100 tasks
* 10 machines try to claim on one path, let's say 2 succeed, rest will fail the task they have claimed

lib/harmony/resources/resources.go
magik6k: How about an even more minimal storage interface which allows the task to bring its own storage system:

```go
// technically now this interface would be general enough to just call it "TaskResource"
type Storage interface {
	// taskID may be used to see if some/all of sector files are already in the local storage (possible edge-case when retries of e.g. transfers happen)
	HasCapacity func(taskID int) bool
  
	// Called at the CanAccept scheduling stage, does not have to run on this machine
	Claim func(taskID int) (err error)

	// This allows some other system to consider the task done.
	// It's up to the caller to remove the data, if that applies.
	// May be called on the task which wasn't run in case this node fails to claim it
	MarkComplete func(taskID int) error
}
```

Then it would be really easy to implement exactly this interface as a direct extension to `paths.Local`, and it each task which wants to use the storage we could just roughly:

```go
harmonytask.TaskTypeDetails{
		Max:  s.max,
		Name: "SDR",
		Cost: ...
		// returns harmonytask.Storage
		Storage: task.localStorage.TaskResources(storiface.FTCache, storiface.FSOverheadSeal)
		MaxFailures: 2,
		Follows:     nil,
	}
```

Then in task .Do, like in the sealing pipeline the task would just call a version of `task.localStorage.AcquireSector` with the task ID which would give it the correct paths

snadrus: > how do I get the storage path...
Oops, I thought Do() gets a TaskEngine arg like CanAccept() has. I will need to add it. 
Then one does:
locationsCSV := te.GetWorkingLocation(taskID)
locations = strings.Split(",") // When using storagemgt.CombeFuncs() it will be in the order specified.

lib/harmony/harmonytask/task_type_handler.go
```diff
@@ -116,6 +118,15 @@ top:
...
+			log.Infow("did not accept task", "task_id", strconv.Itoa(int(*tID)), "reason", "storage claim failed", "name", h.Name, "error", err)
+			return false
```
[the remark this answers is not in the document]
magik6k: Isn’t the claiming what we do with `h.TaskEngine.db.Exec(h.TaskEngine.ctx, "UPDATE harmony_task SET owner_id=$1 WHERE id=$2 AND owner_id IS NULL", h.TaskEngine.ownerID, *tID)` a few lines above?

cmd/lotus-provider/deps/deps.go
```diff
@@ -253,6 +253,9 @@ Get it with: jq .PrivateKey ~/.lotus-miner/keystore/MF2XI2BNNJ3XILLQOJUXMYLUMU`,
 			}
 		}
 	}
+
+	fmt.Println("last line of populate")
+
```
magik6k, line 258: Looks like debugging print snuck in

```suggestion
```

lib/harmony/harmonytask/task_type_handler.go
```diff
@@ -95,12 +97,33 @@ top:
 		return false
 	}
 
+	var location string
```
magik6k, line 100: location is no longer used, can drop it, along with LocationMap

lib/harmony/harmonytask/task_type_handler.go
```diff
@@ -95,12 +97,33 @@ top:
...
+	releaseStorage := func() {
+	}
+	if h.TaskTypeDetails.Cost.Storage != nil {
+		if c := h.TaskTypeDetails.Cost.Storage.Claim; c != nil {
```
magik6k, line 104: .Claim can't be nil now that .Storage is an interface

lib/harmony/harmonytask/task_type_handler.go
```diff
@@ -95,12 +97,33 @@ top:
...
 	if from != workSourceRecover {
 		// 4. Can we claim the work for our hostname?
 		ct, err := h.TaskEngine.db.Exec(h.TaskEngine.ctx, "UPDATE harmony_task SET owner_id=$1 WHERE id=$2 AND owner_id IS NULL", h.TaskEngine.ownerID, *tID)
 		if err != nil {
 			log.Error(err)
+
+			releaseStorage()
 			return false
 		}
 		if ct == 0 {
```
magik6k, line 129: We also must call releaseStorage if we didn't accept the taks because it was already taken

lib/harmony/harmonytask/task_type_handler.go
```diff
@@ -247,5 +271,11 @@ func (h *taskTypeHandler) AssertMachineHasCapacity() error {
 	if r.Gpu-h.Cost.Gpu < 0 {
 		return errors.New("Did not accept " + h.Name + " task: out of available GPU")
 	}
+
+	if h.TaskTypeDetails.Cost.Storage != nil {
+		if has := h.TaskTypeDetails.Cost.Storage.HasCapacity; has != nil && !has() {
```
magik6k, line 276: HasCapacity can't be nil, it's in an interface.

lib/harmony/resources/resources.go
```diff
@@ -24,6 +24,20 @@ type Resources struct {
...
+// See storagemgr/storagemgt.go for more details.
+type Storage interface {
+	HasCapacity() bool
+
+	// This allows some other system to claim space for this task.
+	Claim(taskID int) error
+
+	// This allows some other system to consider the task done.
+	// It's up to the caller to remove the data, if that applies.
+	MarkComplete() error
```
magik6k, line 40: MarkComplete should also have the taskID param

lib/harmony/resources/resources.go
```diff
@@ -24,6 +24,20 @@ type Resources struct {
 	Gpu       float64
 	Ram       uint64
 	MachineID int
+	Storage
+}
+
+// Optional Storage management.
+// See storagemgr/storagemgt.go for more details.
```
magik6k, line 31:
```suggestion
```

magik6k (commented):
Some comments, but this looks much closer.

magik6k (approved):
With the last commit this is sufficient to implement storage reservation mechanism for tasks in the sealing pipeline; I'll open a PR on top of this one with impl ported from my previous PR with the better harmony integration

PR #11649, opened 2024-02-23
Adjust actor event API after review

rvagg:
From review discussions in https://github.com/filecoin-project/lotus/pull/11618, WIP

* `TipSetCid` -> `TipSetKey` (this is still nillable on the filter because omitempty doesn't play as nice, but should confirm that this is reasonable given existing API precedents)
* `EmitterAddr` -> `Emitter`
  * [x] Document that this will end up being an ID address in the future
* `Epoch` -> `Height`
* `FromHeight` and `ToHeight` take nillable `abi.ChainEpoch`s (ints) that must be >= 0 when set
* Heights are interpreted as being exact, there is no longer a "whatever you have"
  * [x] Documented that this will be strict in the future after we can say for sure whether or not we have those events
* Removed `Prefill` option in `SubscribeActorEvents` API and collapsed the argument structure, `FromHeight` will do that work implicitly.
* `nil` for `FromHeight` will:
  * For `GetActorEvents` be interpreted to mean "current epoch"
  * [x] For `SubscribeActorEvents` will be interpreted to mean "next epoch" (TODO: confirm we want to do this because current impl I think will just do a prefill of current epoch)
* `nil` for `ToHeight` will:
  * For `GetActorEvents` be interpreted to mean "current epoch"
  * For `SubscribeActorEvents` be interpreted to mean that there is no height-based stopping condition
* `EnableActorEventsAPI` config option has been moved to a new ~`ActorEvents` top-level section; could also just be `Events` as per previous discussions?~ `Events` section. (future TODO of moving other events config options into there so it won't be lonely)
* [x] Halt `SubscribeActorEvents` when we `ToHeight` is set and we have reached that position; currently will just stay open giving you nothing.
* [ ] ~Add `Limit` option to the filter to limit the number of epochs to process, counting from whatever the start epoch is calculated to be~ _(Discussed [here](https://github.com/filecoin-project/lotus/pull/11618/files#r1498570317), I'm not sure I see a use for this given the global limit that's in place and the way you can use From and To as a user)._
* [x] Make send-buffer handling for `SubscribeActorEvents` smarter:
  * [x] Prefill should have a minimum rate that is acceptable to send events, abort if they are too slow (perhaps calculate based on the speed at which you receive an epoch—you should be able to catch up in reasonable time)
  * [x] Ongoing subscription shouldn't end up lagging but should be able to deal with large numbers of events per epoch
  * [ ] Tests for all this fancy footwork


rvagg:
@Stebalien I'm going to call this one "ready for review", but it's lacking some unit testing for the new stream backlog handling code (and maybe some other minor things). I don't want to invest in figuring out how to build something for that without getting agreement on the approach.

rvagg:
Alternatively, we can just merge this in to https://github.com/filecoin-project/lotus/pull/11618 and go from there. Whatever's easiest to review.

chain/types/actor_event.go
```diff
@@ -37,25 +27,30 @@ type ActorEventFilter struct {
...
-	ToEpoch string `json:"toEpoch,omitempty"`
+	// The height of the latest tipset to include in the query. If empty, the query ends at the
+	// latest tipset.
+	ToHeight *abi.ChainEpoch `json:"toHeight,omitempty"`
 
 	// Restricts events returned to those emitted from messages contained in this tipset.
-	// If `TipSetCid` is present in the filter criteria, then neither `FromEpoch` nor `ToEpoch` are allowed.
-	TipSetCid *cid.Cid `json:"tipsetCid,omitempty"`
+	// If `TipSetKey` is legt empty in the filter criteria, then neither `FromHeight` nor `ToHeight` are allowed.
+	TipSetKey *TipSetKey `json:"tipsetKey,omitempty"`
```
rvagg, line 43: Pointer here may not be idiomatic; I just wanted to avoid an unnecessary `[]` in the json form, mainly. I think maybe if `[]` is omitted on the input side that it may still resolve to the zero-value here, I'm not sure.

node/impl/full/actor_events.go
```diff
@@ -39,188 +40,320 @@ type ActorEventsAPI struct {
...
-		if len(f.FromEpoch) != 0 || len(f.ToEpoch) != 0 {
-			return nil, fmt.Errorf("cannot specify both TipSetCid and FromEpoch/ToEpoch")
+func (a *ActorEventHandler) parseFilter(f types.ActorEventFilter) (*filterParams, error) {
+	if f.TipSetKey != nil && !f.TipSetKey.IsEmpty() {
+		if f.FromHeight != nil || f.ToHeight != nil {
+			return nil, fmt.Errorf("cannot specify both TipSetKey and FromHeight/ToHeight")
 		}
 
+		tsk := types.EmptyTSK
+		if f.TipSetKey != nil {
```
Stebalien, line 94: redundant

node/impl/full/actor_events.go
```diff
@@ -39,188 +40,320 @@ type ActorEventsAPI struct {
...
-		if f.Prefill {
-			evs, err := getCollected(ctx, fm)
-			if err != nil {
-				log.Errorf("failed to get collected events: %w", err)
-				return
+		// Handle any historical events that our filter may have picked up -----------------------------
+
+		evs, minEpoch, maxEpoch := getCollected(ctx, fm)
+		if len(evs) > 0 {
+			// must be able to send events at least twice as fast as they were generated
```
Stebalien, line 207: You know, I'd give them a single block time and that's it. We already have a limit on the number of events they can collect, and that'll keep us from backing up our main channel.

It'll also simplify this code quite a bit.

node/impl/full/actor_events.go
```diff
@@ -39,188 +40,320 @@ type ActorEventsAPI struct {
...
+				eventsPerSecond = 1
 			}
 
+			// send events from evs to the out channel and ensure we don't do it slower than eventsPerMs
+			ticker := time.NewTicker(time.Second)
+			defer ticker.Stop()
+
+			const maxSlowTicks = 3 // slightly forgiving, allow 3 slow ticks (seconds) before giving up
+			slowTicks := 0
+			sentEvents := 0.0
```
Stebalien, line 225: Given that we're doing a single batch here, there's no need for a ticker. We can just have a global timeout.

Stebalien: (I mean, the _right_ way to handle slow reads is in the JSON-RPC library... anyways)

node/impl/full/actor_events.go
```diff
@@ -39,188 +40,308 @@ type ActorEventsAPI struct {
...
-		for {
-			select {
-			case val, ok := <-in:
-				if !ok {
-					// Shutting down.
-					return
-				}
+		var buffer []*types.ActorEvent
+		const α = 0.2                        // decay factor for the events per height EMA
+		var eventsPerHeightEma float64 = 256 // exponential moving average of events per height, initially guess at 256
```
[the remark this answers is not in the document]
Stebalien: See above. IMO, we wait one block time (30s) then give up.

Stebalien (commented):
_Ideally_ we'd be able to:

1. Query for historical events.
2. Send them.
3. Query for more.
4. Send them.
5. Etc...

Then, when we're all caught up, we'd subscribe. But that seems like a lot of work. Also, I assume the limits will prevent that from really being an issues (we should never have to send _that_ many events.

node/impl/full/actor_events.go
rvagg: I won't admit how long I looked at this wondering what Go magic I was missing .. before noticing the enclosing `if`

rvagg:
Pretty rough coming up with nice unit tests for `SubscribeActorEvents` given they're time-based, but I think I got there.

In response to last round of feedback we have two things for keeping track of "are you getting the events fast enough":

* Historical events must all be received within one block's time, you have ~30s to receive what you've asked for before we consider you delinquent (really, use `GetActorEvents` for historical stuff).
* For real-time events we (conceptually) keep track of the oldest event you haven't yet received and if it takes you longer than a block's time to get it, then you're delinquent and you get disconnected.

Basically: don't fall more than ~one block's time behind in receiving your events and you'll be dine.

Bunch of related refactorings in the last commit.

rvagg:
Going to take this back to #11618 and proceed from there

Issue #3256, opened 2020-08-23
Starting v0.4.2 fails with error lookup _dnsaddr.pl-sin.testnet.drand.sh on 10.0.0.254:53: no such host


astudnev:
```
studnev@node1-2:~$ /usr/local/bin/lotus daemon
2020-08-23T19:30:19.561Z	INFO	main	lotus/daemon.go:166	lotus repo: /home/studnev/.lotus
2020-08-23T19:30:19.561Z	INFO	repo	repo/fsrepo.go:102	Initializing repo at '/home/studnev/.lotus'
2020-08-23T19:30:19.561Z	INFO	build	go-paramfetch@v0.0.2-0.20200701152213-3e0f0afdc261/paramfetch.go:173	Fetching /var/tmp/filecoin-proof-parameters/v27-proof-of-spacetime-fallback-merkletree-poseidon_hasher-8-0-0-5294475db5237a2e83c3e52fd6c2b03859a1831d45ed08c4f35dbf9a803165a9.vk from https://proofs.filecoin.io/ipfs/
2020-08-23T19:30:19.561Z	INFO	build	go-paramfetch@v0.0.2-0.20200701152213-3e0f0afdc261/paramfetch.go:191	GET https://proofs.filecoin.io/ipfs/QmQJKmvZN1a5cQ1Nw6CDyXs3nuRPzvyU5NvCFMUL2BfcZC
 3.00 KiB / 3.00 KiB [======================================================================================================================================] 100.00% 42.13 MiB/s 0s
2020-08-23T19:30:19.767Z	INFO	build	go-paramfetch@v0.0.2-0.20200701152213-3e0f0afdc261/paramfetch.go:138	Parameter file /var/tmp/filecoin-proof-parameters/v27-proof-of-spacetime-fallback-merkletree-poseidon_hasher-8-0-0-5294475db5237a2e83c3e52fd6c2b03859a1831d45ed08c4f35dbf9a803165a9.vk is ok
2020-08-23T19:30:19.767Z	INFO	build	go-paramfetch@v0.0.2-0.20200701152213-3e0f0afdc261/paramfetch.go:173	Fetching /var/tmp/filecoin-proof-parameters/v27-stacked-proof-of-replication-merkletree-poseidon_hasher-8-0-0-sha256_hasher-ecd683648512ab1765faa2a5f14bab48f676e633467f0aa8aad4b55dcb0652bb.vk from https://proofs.filecoin.io/ipfs/
[...]

PR #11657, opened 2024-02-27
Remove singapore drand testnet host from config

rvagg:
Not sure what the history is but this host isn't resolvable.

Found this confirmation that it doesn't exist anymore: https://github.com/filecoin-project/lotus/issues/3256#issuecomment-678817335


Stebalien (approved):

Stebalien:
@GlacierWalrus I assume these are dead and need to be removed from lotus-infra as well?

Stebalien:
Venus will also need to remove them.

GlacierWalrus:
@Stebalien lotus-infra is no more.

Forest (Chainsafe) may need to make some changes, will ping them.

PR #11661, opened 2024-02-27
feat: ipfs: remove kubo client backend

Stebalien:
## Proposed Changes

Remove the kubo client backend. This appears to be deprecated and unused? Basically, I'm _trying_ to kill off https://github.com/filecoin-project/kubo-api-client so I don't have to update it.

## Additional Info

This was used to run an internal IPFS node, for some kind of client deal thing.

Maybe we can also get rid of:

https://github.com/filecoin-project/lotus/blob/086ed0c279402e7a9ab30d894912389abee91d61/node/modules/client.go#L91-L95

But... I'm not sure what to do about:

https://github.com/filecoin-project/lotus/blob/086ed0c279402e7a9ab30d894912389abee91d61/node/modules/graphsync.go#L27

## Checklist

- [ ] CI is green


node/builder_chain.go
```diff
@@ -227,13 +226,6 @@ func ConfigFullNode(c interface{}) Option {
 
 		Override(new(dtypes.ClientBlockstore), modules.ClientBlockstore),
 
-		If(cfg.Client.UseIpfs,
```
rvagg, line 230: I believe this is the only place where `UseIpfs` is actually used, so you may as well delete that too eh?

Stebalien: AH, I thought I did.

rvagg (approved):

Stebalien:
@rvagg I'm also looking for insight into:

- Whether or not it's safe to remove this code (some history, if you have it).
- If we can maybe remove the client retrieval code entirely? Could be done in a new PR.

rvagg:
I don't have enough context but my understanding from a distance _was_ that we are no longer recommending people use the retrieval code in here; you just need to look at the last updates to go-fil-markets to see how stale it all is. Boost forked it, trimmed it down to the bone, and have just removed it entirely from their stack. Lassie has been integrated in various places that do retrievals now and could be done here as Hannah said.

But, I just don't know the current state of our recommendations to users, regarding either side of the market, we still have code laying around to make deals and to perform retrievals. Do we have consistent messaging? We should just rip it all out, but we need good answers for people as a replacement.

I think we can also rip out graphsync while in there, afaik we don't use it for the chain and I think even the lotus-miner recommendations these days is to defer to separate markets software.

:shrug: I suspect that @magik6k might have more context on this and could help? Otherwise I say just do it.

hannahhoward:
Seems like:
1. You can remove the Kubo Blockstore to remove the annoying side repo immediately
2. You SHOULD probably remove all the client retrieval software, which would allow you to remove a lot of other things (like the client blockstore and graphsync). But perhaps that needs slightly more planning and at least an FYI shared with the Lotus community?

I would try to just remove the client blockstore seperately.

rvagg:
+1 to that, and we can stick lassie in there for CLI retrieval if people still want/need a client in here somewhere. Mostly we should just refer people to lassie itself I think.

Stebalien:
Ok, so, this code _isn't_ used for retrieval, it's used to make deals. Specifically, it means you can:

1. `ipfs add -r foo` to get a `$CID`.
2. `lotus client deal $CID ...`.

As described in https://lotus.filecoin.io/tutorials/lotus/import-data-from-ipfs/. That actually seems like it might still be useful? But it also means that the current kubo dependency is overkill: all we need is the ability to read/write blocks. On the other hand... I'm now really annoyed that those dependencies got merged.

magik6k:
This dealmaking pathway is deprecated anyways, the new version of boost is dropping go-fil-markets entirely anyways, so it won't even be possible to make deals with SPs through this mechanism anymore.

Dropping deals-from-ipfs SGTM

PR #11664, opened 2024-02-28
feat: curio deal cache (Piece Park)

magik6k:
## Related Issues
https://github.com/filecoin-project/lotus/pull/11534 Has shipped with a very simple lotus-provider<->boost adapter with no buffers which relied on connections staying alive for the duration of sectors getting sealed. This has proven to be wildly optimistic, and horribly unreliable when run in practice on mainnet.

## Proposed Changes
This PR implements a new task type `ParkPiece` and new storage type `piece`.

When boost wants to add a piece to a sector with lotus-provider it will still talk to the adapter process, but now the adapter process will start by "parking" the piece in local lotus-provider operated area of storage (same storage system as sectors, segregated in the `piece` subdirectory in each storage path; path type filters apply of course)

This means that the data flow will match lotus-miner more closely, which is how boost was designed, so this approach should result in much more robust boost<->LP interactions. It also means that there is now one more, technically redundant data copy, but I really don't see how to avoid it without fundamental changes to boost.

## Additional Info
The "Piece Park" will also be needed when we reimplement a SP market node in curio/lotus-provider for "normal" online deals. We will still be able to support trusted-online/offline deals which fetch data directly into TreeD/Replica at the SDRTrees stage.

The way Piece Park is set up should allow for later sharing unsealed pieces between SP IDs, which boils down to never binding a piece to a specific SP ID, just to the content address (PieceCID)

TODO:
- [x] Integrate into the boost-proxy util
- [x] Piece GC poller (rm pieces with no refs)
- [x] Test on a devnet
- [x] Test on mainnet

## Checklist

- [ ] CI is green


documentation/en/default-lotus-provider-config.toml
```diff
@@ -25,6 +25,18 @@
...
   #WinningPostMaxTasks = 0
 
+  # EnableParkPiece enables the "piece parking" task to run on this node. This task is responsible for fetching
+  # pieces from the network and storing them in the storage subsystem until sectors are sealed. This task is
+  # only applicable when integrating with boost, and should be enabled on nodes which will hold deal data
+  # from boost until sectors containing the related pieces have the TreeD/TreeR constructed.
+  # Note that future Curio implementations will have a separate task type for fetching pieces from the internet.
+  #
+  # type: bool
+  #EnableParkPiece = false
```
snadrus, line 35: Should we add it to the "storage" default layer?

provider/lppiece/task_park_piece.go
```diff
@@ -0,0 +1,195 @@
...
+
+		return true, nil
+	}
+
+	// If no URL is found, this indicates an issue since at least one URL is expected.
+	return false, xerrors.Errorf("no data URL found for piece_id: %d", pieceData.PieceID)
+}
+
+func (p *ParkPieceTask) CanAccept(ids []harmonytask.TaskID, engine *harmonytask.TaskEngine) (*harmonytask.TaskID, error) {
+	id := ids[0]
```
snadrus, line 173: Should we limit this to nodes with space available in their "piece" storage areas?

storage/paths/remote.go
```diff
@@ -747,6 +747,46 @@ func (r *Remote) Reader(ctx context.Context, s storiface.SectorRef, offset, size
 	return nil, nil
 }
 
+func (r *Remote) ReaderSeq(ctx context.Context, s storiface.SectorRef, ft storiface.SectorFileType) (io.ReadCloser, error) {
```
snadrus, line 750: Comment please

storage/sealer/storiface/storage.go
```diff
@@ -20,6 +20,17 @@ type SectorRef struct {
 
 var NoSectorRef = SectorRef{}
 
+// PieceNumber is a reference to a piece in the storage system; mapping between
+// pieces in the storage system and piece CIDs is maintained by the storage index
```
snadrus, line 24: why not use CIDs in the storage system?

snadrus (approved):

storage/sealer/storiface/storage.go
magik6k: Mainly because it's a much much bigger change to the storage system (which today operates on actorid-number tuples).

There are other very slight advantages to number-based names, mostly that allow us to duplicate the same piece across storage easier if need be, but that is not the main reason.

magik6k: fwiw boost maintains a similar mapping of CommD->actor/sector/offset, here I'm just extending this model to work directly on pieces with minimal change to the storage system

documentation/en/default-lotus-provider-config.toml
magik6k: Probably not storage.

Before we move on from boost this will be somewhat awkward, basically this data should live on a node next to boost (so that data copy cost is minimized) or on a node next to/on nodes which execute SDRTrees. Also possible to put this buffer on a separate node type, possibly with cheaper-per-churn-gb-storage, but given how not-scalable boost is anyways that probably doesn't really make sense.

In the future when we integrate the data fetching parts of boost into Curio, most piece-related activities will just happen on the nodes which handle deal downloads+verification.

If we want a simple default that "just works" I'd either put this on configs which enable SDRTrees, or, if we decide to move the boost api adapter (lotus-shed/lpdeal) to core curio (which we might want to do, I'm still thinking through the best approach here), one would just also run the ParkPiece on that node.

(In the future-future when we implement unsealed copy sharing, retrievals probably will become primarly served from PiecePark, then it will make sense to shift the default to enabling this task on storage nodes, but at the same time this specific task will probably become less commonly used in setups, as we'll no longer rely on boost, and curio-native market will have different ways of getting pieces into the system)

snadrus: 👍

snadrus: By concern is creeping-in user-complexity. I would like a very small number of new configurations that the SPs need to be aware of. That gets people started faster and improves engagement as we grow the technologies. 
I think it's understandable that it's not perfectly optimal while it's new. 
I recommended 'storage' because this is a low-CPU, higher-reliability-needed operation.

magik6k: Yeah, putting in storage would make this simpler, but this tasks is pretty high bandwidth

PR #11668, opened 2024-03-01
feat: mpool: Cache actors in lite mode

magik6k:
## Related Issues
Mpool startup in lite mode for some reason requests the same actor head 1000s of times (probably for each message ever sent)

## Additional Info
This makes my node start in ~2 seconds instead of roughly 20min when backed by a lite node

Likely there is a better fix to the underlying issue like cleaning up messages in mpool after finality, but this is way simpler and also solves the issue.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/messagepool/provider.go
+++ b/chain/messagepool/provider.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"time"
 
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/ipfs/go-cid"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"golang.org/x/xerrors"
@@ -16,6 +17,8 @@ import (
 	"github.com/filecoin-project/lotus/chain/stmgr"
 	"github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/lib/must"
+	"github.com/filecoin-project/lotus/lib/result"
 )
 
 var (
@@ -39,9 +42,18 @@ type Provider interface {
 	IsLite() bool
 }
 
+type actorCacheKey struct {
+	types.TipSetKey
+	address.Address
+}
+
+var nonceCacheSize = 128
+
 type mpoolProvider struct {
 	sm *stmgr.StateManager
 	ps *pubsub.PubSub
+
+	liteActorCache *lru.Cache[actorCacheKey, result.Result[*types.Actor]]
 
 	lite MpoolNonceAPI
 }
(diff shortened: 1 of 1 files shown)


chain/messagepool/provider.go
```diff
@@ -39,10 +42,19 @@ type Provider interface {
...
+	address.Address
+}
+
+var nonceCacheSize = 128
+
 type mpoolProvider struct {
 	sm *stmgr.StateManager
 	ps *pubsub.PubSub
 
+	liteActorCache *lru.Cache[nonceCacheKey, result.Result[*types.Actor]]
```
magik6k, line 56: Caching results also caches errors, and that is precisely the point - in lite mode if the node hasn't run for a while the head it is requesting may be behind lotus-gateway lookback limit, so every request would fail in the same way.

chain/messagepool/provider.go
```diff
@@ -53,18 +65,31 @@ func NewProvider(sm *stmgr.StateManager, ps *pubsub.PubSub) Provider {
...
+		liteActorCache: must.One(lru.New[nonceCacheKey, result.Result[*types.Actor]](nonceCacheSize)),
+	}
 }
 
 func (mpp *mpoolProvider) IsLite() bool {
 	return mpp.lite != nil
 }
 
-func (mpp *mpoolProvider) getActorLite(addr address.Address, ts *types.TipSet) (*types.Actor, error) {
+func (mpp *mpoolProvider) getActorLite(addr address.Address, ts *types.TipSet) (act *types.Actor, err error) {
```
rvagg, line 80: you're not using `act` or `err` below your `defer`, they're always going to be unset as far as I can see; what am I missing?

chain/messagepool/provider.go
```diff
@@ -39,10 +42,19 @@ type Provider interface {
 	IsLite() bool
 }
 
+type nonceCacheKey struct {
```
rvagg, line 45: is this the right name? it _may_ reference a nonce lookup error but from from a use-perspective ("getActorLite") it's in the path of getting the actor where the nonce is just a secondary step, so maybe this would be better as `actorLookupKey`

chain/messagepool/provider.go
magik6k: Nope, those values are set on return and accessible in defer. See https://go.dev/play/p/4qc2WmSDnZw

magik6k: (iiuc this behavior is actually the reason why named returns were conceived of - so that you can access return vals in defer)

rvagg: that's wild, didn't know that! thanks

PR #11673, opened 2024-03-04
test(actor events): expanded testing

rvagg:
A few things going on here:

1. Refactored the big DDO verified data test since it's doing so much and getting a bit out of hand; hopefully it's a little cleaner now with some logic extracted to functions
2. Moved that test to its own file, its size feels like it warrants it, and it's largely events focused rather than being about verified DDO tbqh
3. Adding additional testing of other FIP0083 events as I can manage it. Some are a bit difficult to trigger so I may not get all the pokemons.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -670,6 +670,12 @@ workflows:
           suite: itest-direct_data_onboard
           target: "./itests/direct_data_onboard_test.go"
       - test:
+          name: test-itest-direct_data_onboard_verified
+          requires:
+            - build
+          suite: itest-direct_data_onboard_verified
+          target: "./itests/direct_data_onboard_verified_test.go"
+      - test:
           name: test-itest-dup_mpool_messages
           requires:
             - build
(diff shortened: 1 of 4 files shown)


rvagg:
Got all of the events I could easily get without adding too much extra workflow. The hard ones are those that involve longer lived sectors/deals and I don't see a good way to override some of the defaults (like min term of a claim, to make a really short one). Will add deal-activated separately when we get that bubbled up through a new actors bundle.

* [x] `verifier-balance`
* [x] `allocation`
* [x] `allocation-removed`
* [x] `claim`
* [ ] `claim-updated`
* [ ] `claim-removed`
* [x] `deal-published`
* [ ] `deal-activated` (TODO: see https://github.com/filecoin-project/builtin-actors/pull/1526)
* [ ] `deal-terminated`
* [ ] `deal-completed`
* [x] `sector-precommitted`
* [x] `sector-activated`
* [ ] `sector-updated`
* [x] `sector-terminated`

masih (approved):

jennijuju (approved):

PR #11674, opened 2024-03-04
Update CODEOWNERS

jennijuju:
These are the folks who are actively contributing to lotus maintenance, issue triaging, PR reviews and development. 

Made this change to decouple lotus ownership with other filecoin core repo's ownership and maintainership that comes with lotus-maintainer perm

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,4 +3,4 @@
 
 # Global owners
 # Ensure maintainers team is a requested reviewer for non-draft PRs
-*                       @filecoin-project/lotus-maintainers
+*                       @jennijuju @Stebalien @snadrus @aarshkshah1992 @magik6k @zengroud0 @arajasek @rjan90 @masih @rvagg


rjan90 (approved):
lgtm!

rvagg (approved):

Stebalien:
Can we make a new github team and use that? E.g., so we can reuse it elsewhere?

jennijuju:
> Can we make a new github team and use that? E.g., so we can reuse it elsewhere?

I think one of the goal is to not assume all lotus active contributors are by default good maintainers for other repos - like reffvm. 
happy to add a team later on if that’s proven to be the case!

Stebalien:
> I think one of the goal is to not assume all lotus active contributors are by default good maintainers for other repos - like reffvm.
happy to add a team later on if that’s proven to be the case!

Yeah, makes sense. And I guess lotus is basically a monorepo so we don't really need to add the same codeowners to multiple repos.

Issue #11683, opened 2024-03-05
Automate sanity check of the bundle CIDs


masih:
## User Story
As a developer I want Continuous integration to assert that the bundled CIDs in code are correct and correspond to the right root CID / version. This is because it rules out human error considering the sensitivity of the bundles as it is currently done manually before each release. Additionally, 

## Acceptance Criteria
Add a GitHub Actions workflow that sanity checks the bundles for a given network version. The network version may be inferred from the branch name, since Lotus release branches seem to follow `release/<version>` pattern. The sanity check invoves:

* download the bundles from actor release assets (cars) 
* confirm the car root CID s and actor cids in car matches 
   * what we have in builtin-acrors-gen release, e.g. [1.26.0](https://github.com/filecoin-project/lotus/blob/release/v1.26.0/build/builtin_actors_gen.go), and 
   * what’s generated by `lotus state actor-cids --network-version=22` 

The implementation is acceptable when the CI workflow is triggered on push to specific branch names for the bundled CIDs and passes the condition described above, failing otherwise.

PR #11684, opened 2024-03-05
test: build: Add tests to sanity check embedded built-in actors metadata

masih:
## Related Issues
#11683 

## Proposed Changes
Add tests that assert the embedded built-in actors metadata is correct:

* the corresponding CAR file is present in built-in actors released
  assets as a CAR file.
* the corresponding tar.zst is locally present and matches the expected
  SHA-256 digest 
* manifest CID is the only root CID in the corresponding CAR file.
* actor CIDs are present in the corresponding CAR file.

Configure a dedicated CI workflow to only run the tests on:
 * `release/*` branches,
 * modification to `build/builtin_actors_gen.go`,
 * or embedded ZST files.

## Additional Info
Note: only the latest network version seem to have `BundleGitTag`. Tests depend on this value to fetch the right CAR file. 

## Checklist

- [x] CI is green

### DIFF
--- a/build/builtin_actors_gen_test.go
+++ b/build/builtin_actors_gen_test.go
@@ -1,0 +1,107 @@
+//go:build release
+// +build release
+
+package build_test
+
+import (
+	"archive/tar"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/DataDog/zstd"
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-car/v2"
+	"github.com/stretchr/testify/require"
+
+	actorstypes "github.com/filecoin-project/go-state-types/actors"
+
+	"github.com/filecoin-project/lotus/build"
+)
+
+func TestEmbeddedBuiltinActorsMetadata(t *testing.T) {
+	subjectsByVersionByNetworks := make(map[actorstypes.Version]map[string]*build.BuiltinActorsMetadata)
+	for _, subject := range build.EmbeddedBuiltinActorsMetadata {
+		if subject.BundleGitTag == "" {
+			// BundleGitTag is required to verify the SHA-256 checksum.
+			// The pack script only includes this for the latest network version, and it is good enough to only
+			// check the latest network version metadata. Hence the skip.
+			continue
+		}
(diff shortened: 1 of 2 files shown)


build/builtin_actors_gen_test.go
```diff
@@ -0,0 +1,59 @@
...
+func TestEmbeddedBuiltinActorsMetadata(t *testing.T) {
+	for _, subject := range build.EmbeddedBuiltinActorsMetadata {
+		if subject.BundleGitTag == "" {
+			// Only the latest network version specifies BundleGitTag.
+			// Move on to the next Metadata instead of skipping in order to reduce test log verbosity.
+			continue
+		}
+		subject := subject
+		t.Run(fmt.Sprintf("%s/%v", subject.Network, subject.Version), func(t *testing.T) {
+			carURL := fmt.Sprintf("https://github.com/filecoin-project/builtin-actors/releases/download/%s/builtin-actors-%s.car", subject.BundleGitTag, subject.Network)
```
Stebalien, line 26: We already have this car file in `build/actors`.

- If we're trying to ensure that our bundle is correct, I'd consider downloading the sha256 file instead and checking that against our local bundle.
- Otherwise, I assume running "make gen" and checking the diff _should_ validate that everything is correct.

I'd just like to avoid having to constantly re-download bundles in CI. Unless there's a reason for this that I'm missing.

Stebalien: Ok, you know what? That's ~10s even on my terrible connection. There's little reason not to merge this unless it shows up as actually _costing_ something.

Stebalien (approved):

rvagg (approved):
Just want to note that I've been running unit tests locally (per https://github.com/filecoin-project/lotus/pull/11628) and wouldn't mind if we could try and keep them fairly minimal, and even squish down some of the slow ones we currently have. This is only ~9s for me, which is not the worst, but it all stacks up.

Maybe we should do a pass some time and do a skip based on `os.Getenv("CI")` to make it more practical. This one would probably be in that category.

jennijuju:
Can this to be only run in CI when  a PR is open against releases branch?

masih:
> `os.Getenv("CI")`

I have a better idea; I'll change this PR to:

* compare sha256 of files that are already downloaded as Steven suggested.
* only run it if either `builtin_actors_gen.go` or zst files in `/build/actors` change.

That should significantly shave time off the 10 seconds it takes to run this tests which seems to be an issue.

Issue #11594, opened 2024-01-23
Index Actor Events by ID


Stebalien:
Currently, we index actor events by the emitter's f4 address. This will have to change in https://github.com/filecoin-project/lotus/pull/11540 to accommodate non-EVM actors, but really, we simply shouldn't be resolving addresses when we index events.

Instead:

1. We should record the ID in the index (resolve nothing). The ActorID plus the tipset is the "canonical" identifier for the actor.
2. When querying for events, resolve addresses to IDs using the latest tipset, then query. As long as we limit ourselves to querying for non-reverted events, there shouldn't be any issue of ID reuse.
3. When subscribing, we'll have to resolve the filter's addresses to IDs at every tipset.

Unlike the current system, this will work correctly with _every_ address type, will result in more performant indexing, and will reduce the size of the database.

Unfortunately, the correct way to do this involves a migration.

masih:
@rvagg if you haven't started working on this, I might as well take this one too considering I am working on https://github.com/filecoin-project/lotus/issues/11640 ?

rvagg:
@masih I had a look at my code and it wasn't quite as terrible as I imagined. I've tidied it up a little and made it all pass with the latest v1.26 @ https://github.com/filecoin-project/lotus/pull/11694. It's just a draft and shouldn't necessarily be _the_ way this ticket gets done. I have some other things to get done in the meantime so I've removed myself from assignee here but could come back to this if you don't happen to get to it in the course of what you're working on. I started looking at the migration problem but didn't get too far. I really want good tests for migrations and there are none in there yet.

masih:
@Stebalien Question: What is the correct thing to do when migrating the reverted events `emitter_addr`? We can't resolve those address to actor IDs right? Which means some default value for emitter, e.g. `0` for events that cannot be migrated?

Stebalien:
>  We can't resolve those address to actor IDs right?

1. Reverted events exist in reverted tipsets. If we still have the tipset, we _should_ be able to resolve the addresses relative to those tipsets.
2. It might just be faster to re-index from the events in the receipts.

But, honestly, I'd just delete the reverted events from the database in this case. I'm not quite sure why we record them anyways given that, unless I'm mistaken, we only _return_ reverted events from the "subscribe" API. I guess we should add "remove reverted events from the database" to the TODO list...

masih:
> remove reverted events from the database

[...]

PR #11694, opened 2024-03-07
events: index actor events by actor ID rather than f4 addr

rvagg:
Ref: https://github.com/filecoin-project/lotus/issues/11594

WIP, notes:

* This doesn't have a migration, it just starts from a new schema, I started looking at what that would do and there's a little bit of work on a test framework for index migrations. That looked like a fun problem to try and solve and have tests for it.
* There's an unresolved question in this: what kind of address do we provide back to the user on their events? This PR has a two-way translation, such that your inputs can be non-ID addresses and they'll get translated, but your events will be translated to addresses if they can be, or ID addresses if not. I'm unsure of what the appropriate UX is here.
* Revisiting this after a month I'm not so confident in the timing of the look-ups - are they being done at the correct tipset wherever we do a translation (see https://github.com/filecoin-project/lotus/issues/11594 for notes about this).

### DIFF
--- a/chain/events/filter/event.go
+++ b/chain/events/filter/event.go
@@ -3,13 +3,13 @@ package filter
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"math"
 	"sync"
 	"time"
 
 	"github.com/ipfs/go-cid"
 	cbg "github.com/whyrusleeping/cbor-gen"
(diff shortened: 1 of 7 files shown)


rvagg:
better version @ https://github.com/filecoin-project/lotus/pull/11723

PR #11695, opened 2024-03-07
chore: scripts: Fix go imports at least 2X faster

masih:
## Related Issues
None.

## Proposed Changes
`fiximports` is run up to four times across various make targets. The previous implementation used various commands to find go files, exclude irrelevant ones, etc. all in bash. It was taking ~25 seconds every time on my machine.

The changes here re-implement the bash script in pure go, with 1:1 functionality. The work here uses native go implementation to sort and adjust imports, and uses `goimports` under the hood to format imports just like the bash script. Compared to the previous bash version, this is at least 2X faster, taking ~10 seconds to run on my machine. 

## Additional Info
The convention this repository follows is to group filecoin-project imports across 2 additional separate groups. The purpose of fixing imports is to indeed have a consistent formatting across the files. However, it is unclear why the choice of grouping is very particular. Relaxing the grouping to simply SDK and non-SDK groups can reduce the time it takes to fix them to ~5 seconds. Additionally, this means a lot of the generated files would not need import reformatting at all.

I think we should consider a simpler more straightforward import formatting. To be discussed with wider dev team.

## Checklist

- [x] CI is green


snadrus:
Yes I think we should go with the standard approach for imports. We don't need to be doing anything real fancy.

snadrus (approved):
Consider the simpler goimports default layout.

rvagg:
> I think we should consider a simpler more straightforward import formatting. To be discussed with wider dev team.

yeah, I agree, I did a quick pass on this script recently to make it slightly faster and less messy by changing the `sed` invocation, but the whole purpose of it is questionable. @Stebalien did some more investigation and has opinions.

I like consistency, but this script is super annoying and slows down development.

Stebalien:
Honestly, I'm fine if we just garble them up for generated files as long as the sorting is consistent.

Stebalien:
(if we only apply the import sorting to _generated_ files, that is)

scripts/fiximports/main.go
```diff
@@ -0,0 +1,92 @@
...
+			strings.HasSuffix(path, "extern/filecoin-ffi"),
+			strings.HasSuffix(path, "extern/test-vectors"):
+			return filepath.SkipDir
+		case // Skip directories, generated cborgen go files and any other non-go files.
+			info.IsDir(),
+			strings.HasSuffix(info.Name(), "_cbor_gen.go"),
+			!strings.HasSuffix(info.Name(), ".go"):
+			return nil
+		}
+		return fixGoImports(path)
```
magik6k, line 43: I'd spin this up in a goroutine, should make the whole thing much fasterer

scripts/fiximports/main.go
```diff
@@ -0,0 +1,92 @@
...
+	}); err != nil {
+		fmt.Printf("Error fixing go imports: %v\n", err)
```
Stebalien, line 45: Exit with a non-zero status on failure.

scripts/fiximports/main.go
```diff
@@ -0,0 +1,92 @@
...
+		_ = sourceFile.Close()
+		if err := replaceFileContent(path, formatted); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func replaceFileContent(path string, replacement []byte) error {
+	target, err := os.OpenFile(path, os.O_RDWR|os.O_TRUNC, 0666)
```
Stebalien, line 78: Why re-open? We can open read/write then call `file.Seek(0, 0)` and `file.Truncate(0)`.

scripts/fiximports/main.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+func main() {
+	if err := filepath.Walk(".", func(path string, info fs.FileInfo, err error) error {
+		switch {
+		case err != nil:
+			return err
+		case // Skip ffi and test vectors in "./extern" directory and their contents.
+			strings.HasSuffix(path, "extern/filecoin-ffi"),
+			strings.HasSuffix(path, "extern/test-vectors"):
```
Stebalien, line 35: I'd be a little more strict here. I.e., path is exactly `./extern/...` (in case we have other extern directories eventually). Not critical, but...

Stebalien (changes requested):
Only critical change is the non-zero exit code. But it would be nice to make the other changes as well (including @magik6k's point about doing all of this in parallel).

scripts/fiximports/main.go
masih: Just to avoid opening files in RW when they might not need to be. In terms of speed it made little difference on my machine. 

Happy to change it if you think the implementation is too defensive.

scripts/fiximports/main.go
masih: Yep I have tried that with errgroups. it's about 4x faster than current 👍

scripts/fiximports/main.go
Stebalien: Honestly, opening RW is safer IMO (e.g., if the user calls `git checkout` while running make, we'll clobber the now deleted file instead of clobbering the newly checked out file).

scripts/fiximports/main.go
Stebalien: Hm. Actually, we should probably exclude `extern` entirely. But my point was more: exclude `./extern/` _only_, not everything that happens to end with `extern/...`.

scripts/fiximports/main.go
masih: Sigh... goimports programatic API is not thread safe unfortunately.

Stebalien: Wow, that's a nasty API... We could do it in two steps (this time, we probably would close the files after reading them), but that's probably more trouble than it's worth for now.

masih: yep. I think the thing to do is to simplify our formatting to two groups of SDK and no SDK, and frankly use another tool to sort imports.  `goimports` seems very slow in addressing the feature requests.

scripts/fiximports/main.go
```diff
@@ -0,0 +1,92 @@
...
+
+func replaceFileContent(target *os.File, replacement []byte) error {
+	if _, err := target.Seek(0, io.SeekStart); err != nil {
+		return err
+	}
+	written, err := target.Write(replacement)
+	if err != nil {
+		return err
+	}
+	return target.Truncate(int64(written))
```
Stebalien, line 84: nit: I'd _generally_ recommend that you truncate then write. It shouldn't make any difference at the OS level, but it'll reduce the chances of someone observing a garbage version.

(well, actually, it might make a difference with a CoW filesystem...)

But this is getting _deep_ into nit forest.

Stebalien (approved):

rvagg:
👌 nice, almost exactly 2x faster

scripts/fiximports/main.go
masih: Interesting; thank you for pointing that out

PR #11696, opened 2024-03-08
fix: api: Length check the array sent to RPC.

parthshah1:
## Related Issues

When calling **eth_feeHistory**, it takes input a rewardPercentile array. Currently, there isn't any restriction on the length of the array, can result in high computation and potentially leading to OOM issues in RPC nodes. The fix is to set a hard limit of the array to 100. 
## Proposed Changes

Changing the length of RewardPercentiles Array to upto 100.
## Checklist

- [ ] CI is green

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -692,7 +692,7 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
 	rewardPercentiles := make([]float64, 0)
 	if params.RewardPercentiles != nil {
 		if len(*params.RewardPercentiles) > maxEthFeeHistoryRewardPercentiles {
-			return ethtypes.EthFeeHistory{}, fmt.Errorf("Length of the reward percentile array cannot be greater than 100")
+			return ethtypes.EthFeeHistory{}, errors.New("length of the reward percentile array cannot be greater than 100")
 		}
 		rewardPercentiles = append(rewardPercentiles, *params.RewardPercentiles...)
 	}
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -687,6 +687,9 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
 	if params.BlkCount > 1024 {
 		return ethtypes.EthFeeHistory{}, fmt.Errorf("block count should be smaller than 1024")
 	}
+	if len(*params.RewardPercentiles) >= 100 {
+		return ethtypes.EthFeeHistory{}, fmt.Errorf("Length cannot be greater than 100")
+	}
 	rewardPercentiles := make([]float64, 0)
 	if params.RewardPercentiles != nil {
 		rewardPercentiles = append(rewardPercentiles, *params.RewardPercentiles...)
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -40,6 +40,7 @@ import (
 )
 
 var ErrUnsupported = errors.New("unsupported method")
+
 const maxEthFeeHistoryRewardPercentiles = 100
 
 type EthModuleAPI interface {
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -40,7 +40,7 @@ import (
 )
 
 var ErrUnsupported = errors.New("unsupported method")
-var RewardPercentileArrayLimit = 100
+var RewardPercentileArrayLimit int = 100
 
 type EthModuleAPI interface {
 	EthBlockNumber(ctx context.Context) (ethtypes.EthUint64, error)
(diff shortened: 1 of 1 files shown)


node/impl/full/eth.go
```diff
@@ -687,6 +687,9 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
 	if params.BlkCount > 1024 {
 		return ethtypes.EthFeeHistory{}, fmt.Errorf("block count should be smaller than 1024")
 	}
+	if len(*params.RewardPercentiles) >= 100 {
+		return ethtypes.EthFeeHistory{}, fmt.Errorf("Length cannot be greater than 100")
+	}
 	rewardPercentiles := make([]float64, 0)
 	if params.RewardPercentiles != nil {
```
rvagg, line 694:
```suggestion
	rewardPercentiles := make([]float64, 0)
	if params.RewardPercentiles != nil {
		if len(*params.RewardPercentiles) >= 100 {
			return ethtypes.EthFeeHistory{}, fmt.Errorf("Length cannot be greater than 100")
		}
```

@parthshah1 you've introduced a panic DoS in your DoS fix!

rvagg:
Seems reasonable, except that I don't know what a reasonable limit on this would be; geth doesn't appear to have one, at least when you get this deep: https://github.com/ethereum/go-ethereum/blob/c41105ce80f12f60ec4bf6c65c4c59c6bf4a86e7/eth/gasprice/feehistory.go#L200 and it event does an `8*len` allocation at one point. Maybe there's a total params limit size, which I guess would be appropriate for us too if we don't have one.

node/impl/full/eth.go
parthshah1: I had a nudge. Just saw CI logs. Apologies.

parthshah1:
@rvagg. I had a discussion about the limits with @Stebalien, that maybe 100 would be reasonable number since one would not need more than that to calculate the eth_feehistory. It also aims to solve computation issues this might cause with huge arrays.  

> it event does an 8*len allocation at one point. Maybe there's a total params limit size, which I guess would be appropriate for us too if we don't have one.

I was not aware of such a case in geth. I am not too expert in this area and how often, how much data it is expected to compute. I will agree to what you think should be right. Though, imo keeping the hard fix on the length won't harm.

rvagg (approved):
this seems fine to me, we can bump it up if there's a genuine request for more

node/impl/full/eth.go
```diff
@@ -689,6 +689,9 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
 	}
 	rewardPercentiles := make([]float64, 0)
 	if params.RewardPercentiles != nil {
+		if len(*params.RewardPercentiles) >= 100 {
```
Stebalien, line 692:
```suggestion
		if len(*params.RewardPercentiles) > 100 {
```

(to match the error description)

Stebalien: I'd also make the limit a constant at the top of the file.

Stebalien (commented):
Nits, but this looks correct.

node/impl/full/eth.go
parthshah1: Done.

node/impl/full/eth.go
```diff
@@ -40,6 +40,7 @@ import (
 )
 
 var ErrUnsupported = errors.New("unsupported method")
+var RewardPercentileArrayLimit int = 100
```
Stebalien, line 43: nit `maxEthFeeHistoryRewardPercentiles` is a little more descriptive (not a limit on the array size, a limit on something we're doing).

Also, `const ... = 100`, no need for this to be variable.

Stebalien (approved):

node/impl/full/eth.go
parthshah1: Made appropriate changes.

Stebalien (approved):

Stebalien:
Thank you! Sorry for all the nits.

One final thing: it looks like some of the lints are unhappy. I _think_ you just need to run `go fmt` and fix some spelling.

rvagg (approved):

node/impl/full/eth.go
```diff
@@ -689,6 +691,9 @@ func (a *EthModule) EthFeeHistory(ctx context.Context, p jsonrpc.RawParams) (eth
 	}
 	rewardPercentiles := make([]float64, 0)
 	if params.RewardPercentiles != nil {
+		if len(*params.RewardPercentiles) > maxEthFeeHistoryRewardPercentiles {
+			return ethtypes.EthFeeHistory{}, fmt.Errorf("Length of the reward percentile array cannot be greater than 100")
```
rvagg, line 695:
```suggestion
			return ethtypes.EthFeeHistory{}, errors.New("length of the reward percentile array cannot be greater than 100")
```

Issue #11679, opened 2024-03-05
Move some of Fevm.Events.* config options to top-level Events


rvagg:
https://github.com/filecoin-project/lotus/pull/11618 introduces an `Events` top-level, config option. `Fevm.Events` is full of items that are not FEVM-specific and should be moved:

* `DisableRealTimeFilterAPI`
* `DisableHistoricFilterAPI`
* `FilterTTL` (needs a doc update)
* `MaxFilters` (needs doc update)
* `MaxFilterResults`
* `MaxFilterHeightRange`
* `DatabasePath`

My proposal is to copy all of these into the new top-level `Events` and do a reconcile at start-up, prioritising the new top-level version if set. If `Fevm.Events.*` items are set we'll print a warning to stdout (or via log). Then in a future breaking change, maybe next network version, we remove `Fevm.Events` entirely.

The blast-radius of changes is people using the Ethereum-compatible log subscription today, which I believe is relatively small. The risk with v1.26.0 if we don't get to this is that we have a new audience of oracle developers using generic actor events APIs who will have broken configs. So sooner than later would be good.

rvagg:
https://github.com/filecoin-project/lotus/pull/11698

PR #11698, opened 2024-03-08
feat(config): move Fevm.Events->Events, implement soft deprecation

rvagg:
Closes: https://github.com/filecoin-project/lotus/issues/11679

The `Fevm.Events` config options don't make sense under `Fevm`, particularly now that they are for "actors" in general. @jennijuju and I thought that getting them moved to the new top-level `Events` in v1.26.0 is going to be the least painful option because the number of users who currently touch those options is restricted to a small number that are using the eth APIs. But now, with the new actor events APIs, the circle of users who may be touching these will increase. So best act now before we have more people with these config options set in the wrong place.

Although, I've implemented a **soft** deprecation here. Your old config will still continue to work but you'll get a stderr warning about it and how it will be removed in a future update. Your settings will be automatically migrated to the new location. So this technically shouldn't be breaking for anyone.

* Introduce a `moved:"To.New.Config"` tag which prints a stderr warning when you use one of these, but will move any set value to the new location if the new location isn't already set itself.
* Look for `X is DEPRECATED` to hold certain fields back from documentation.
* Use `toml:"omitempty"` to prevent the default config output from having these deprecated values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,7 +136,7 @@ Additionally, Filecoin is not Ethereum no matter how much we try to provide API/
 
 [FIP-0049](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0049.md) introduced _Actor Events_ that can be emitted by user programmed actors. [FIP-0083](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0083.md) introduces new events emitted by the builtin Verified Registry, Miner and Market Actors. These new events for builtin actors are being activated with network version 22 to coincide with _Direct Data Onboarding_ as defined in [FIP-0076](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0076.md) which introduces additional flexibility for data onboarding. Sector, Deal and DataCap lifecycles can be tracked with these events, providing visibility and options for programmatic responses to changes in state.
 
-Actor events are available on message receipts, but can now be retrieved from a node using the new `GetActorEvents` and `SubscribeActorEvents` methods. These methods allow for querying and subscribing to actor events, respectively. They depend on the Lotus node both collecting events (with `Fevm.Events.RealTimeFilterAPI` and `Fevm.Events.HistoricFilterAPI`) and being enabled with the new configuration option `Events.EnableActorEventsAPI`. Note that a Lotus node can only respond to requests for historic events that it retains in its event store.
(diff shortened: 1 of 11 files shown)


masih (approved):

jennijuju:
capture this in the changelog maybe?

jennijuju (approved):

rvagg:
Added changelog docs about this. Also tested it out. Nothing in my systemd journal if I have none of the options set. If I set some of them then I get:

```
Mar 11 13:15:35 fletcher lotus-calibnet[2677403]: WARNING: Use of deprecated configuration option 'Fevm.Events.DisableRealTimeFilterAPI' will be removed in a future release, use 'Events.DisableRealTimeFilterAPI' instead
Mar 11 13:15:35 fletcher lotus-calibnet[2677403]: WARNING: Use of deprecated configuration option 'Fevm.Events.DisableHistoricFilterAPI' will be removed in a future release, use 'Events.DisableHistoricFilterAPI' instead
Mar 11 13:15:35 fletcher lotus-calibnet[2677403]: WARNING: Use of deprecated configuration option 'Fevm.Events.FilterTTL' will be removed in a future release, use 'Events.FilterTTL' instead
Mar 11 13:15:35 fletcher lotus-calibnet[2677403]: WARNING: Use of deprecated configuration option 'Fevm.Events.MaxFilters' will be removed in a future release, use 'Events.MaxFilters' instead
Mar 11 13:15:35 fletcher lotus-calibnet[2677403]: WARNING: Use of deprecated configuration option 'Fevm.Events.MaxFilterResults' will be removed in a future release, use 'Events.MaxFilterResults' instead
Mar 11 13:15:35 fletcher lotus-calibnet[2677403]: WARNING: Use of deprecated configuration option 'Fevm.Events.MaxFilterHeightRange' will be removed in a future release, use 'Events.MaxFilterHeightRange' instead
```

If I set `Fevm.Events.DisableRealTimeFilterAPI` to `true` then my `eth_getLogs` stops working:

```
$ curl -s -X POST   -H "Content-Type: application/json"   --data '{"method":"Filecoin.EthGetLogs", "params": [{}],"id":1,"jsonrpc":"2.0"}' http://localhost:1235/rpc/v1
{"jsonrpc":"2.0","id":1,"error":{"code":1,"message":"method not supported"}}
```

If I move that to `Events.DisableRealTimeFilterAPI` then it still doesn't work.

If I set `Fevm.Events.DisableRealTimeFilterAPI` to `false` and `Events.DisableRealTimeFilterAPI` to `true` then it works. As expected.

Issue #11680, opened 2024-03-05
EventFilterManager needs to deal in epochs


rvagg:
`ToHeight` should lead to a stop-condition, or at least be communicated to a consumer that a stop-condition has been reached. When subscribing, you should be able to receive a signal that there are no more events because you've reached the maximum height and there won't be anymore. See `SubscribeActorEvents` internals for the juggling needed to work around this.

Stretch goal: organise the API by epoch: batch them up such that you get all events for a particular epoch in one call, that way you can make some smarter consumption decisions on the outside and maybe even offer APIs that have this information in them.

Currently there's no way to know precisely how many events are in an epoch; even if you do a `GetActorEvents` there's a maximum number of events we'll give you according to config option and you have no insight. Perhaps a better API would give you a block, `{ "height": 12345, "totalEvents": 56, "matchedEvents": 45, "events": [ ... ] }` ?

rvagg:
@masih FYI this is related to events index, when you have your head in there notice how there's no easy stop condition, you just install a filter and it gets applied until you remove it .. but how do you know when to remove it? You have to watch from outside. `EventFilterManager` knows both the current height and `maxHeight` of a filter. Unfortunately reverts make this complicated, I'm not clear on how to handle that. Maybe it's entirely appropriate to handle this outside because you can't assume you won't be rolling back and forward again?

The related concern is that events are already grouped by tipset - we load all messages then their receipts for an entire tipset and process them in a batch in `CollectEvents` but we get them one at a time with the subscription channel _or_ all matching for a filter with `TakeCollectedEvents` (check out how `maxResults` messes things up too). Perhaps if we rearranged this whole thing to chunk by tipset then we might be able to have better downstream handling?

I'm not sure how this issue should get resolved, for now I think it's more of a bookmark of a set of concerns that would be nice to deal with somehow in the guts of this thing.

PR #11707, opened 2024-03-12
feat(events): introduce compact JSON form of EventEntry

rvagg:
Optional compact form, can round-trip as either full standard Go style JSON or compact tuple struct with decoded "value" field represented as dag-json. Currently turned on as strict default for `GetActorEvents` and `SubscribeActorEvents`.

Bear with me, I'll explain the reasons for some of the crazy in the code here, but first what I'm trying to achieve here.

Here's some calibnet events from nv22 as they are directly out of the json response from `Filecoin.GetActorEvents`, formatted for emphasis on the pieces that you're expected to decode to work out what the event is doing (i.e. the entries):

Built-in:

```json
{"entries":[
  {"Flags":3,"Key":"$type","Codec":81,"Value":"bmRlYWwtcHVibGlzaGVk"},
  {"Flags":3,"Key":"id","Codec":81,"Value":"GgACyyc="},
  {"Flags":3,"Key":"client","Codec":81,"Value":"GQS2"},
  {"Flags":3,"Key":"provider","Codec":81,"Value":"GQSG"}
],"emitter":"t05","reverted":false,"height":1427997,"tipsetKey":[{"/":"bafy2bzacedsp7zcxejngyftpu6hnr72ahsxyrycn5bjtfsp3vwxgvhs2hbpag"},{"/":"bafy2bzaceduo4goajlqtkarcozgm4hl6yla5kgajypguddk4embb6oxj2ah22"},{"/":"bafy2bzaceafs5ewv73nqxqhqqveanv74il5jxpvkp5oymoje6prtqh6b6hzje"}],"msgCid":{"/":"bafy2bzaceatdcipsape77hnf4lvaslwgkx4axbmro4fsgq3dvfk5ri4nbpoec"}}
```

FEVM:

```json
{"entries":[
  {"Flags":3,"Key":"t1","Codec":85,"Value":"mzhAeIhrixxtxKRVUgms9lTj+bEmu4rNhD4BOizSMmc="},
  {"Flags":3,"Key":"t2","Codec":85,"Value":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADc="},
  {"Flags":3,"Key":"d","Codec":85,"Value":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAKSfDFPBaMbbj1gAAAAAAAAAAAAAAAAAAAAAAAAAAAAApJ8MU8FoxtuPWAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABl7xEw"}
],"emitter":"t410fazjdjy3l32rrlnavap6uljckuvjl7f4ziticsca","reverted":false,"height":1427998,"tipsetKey":[{"/":"bafy2bzaceb7dl5fqr32biq75c6f5e5lwuc7ikdgrztmetweiiwahrgsmvflp4"},{"/":"bafy2bzacebjzpdk4aybiwfpcjs6fmkxf7me452f4o4tcs7wugiambehsnspde"},{"/":"bafy2bzacecw4f5n6xlw3jqui2pw7eoaga3cxyepisestfztbizcfnzwd5kise"}],"msgCid":{"/":"bafy2bzaceahofvwlvtzcyfx75v26boqnai2t6q5uqogdb6bkepcfu2rqlmmyg"}}
```

With this branch, these same events come out looking like this:

Built-in:

```json
{"emitter":"t05","entries":[
  [3,81,"$type","deal-published"],
  [3,81,"id",183079],
  [3,81,"client",1206],
  [3,81,"provider",1158]
],"height":1427997,"msgCid":{"/":"bafy2bzaceatdcipsape77hnf4lvaslwgkx4axbmro4fsgq3dvfk5ri4nbpoec"},"reverted":false,"tipsetKey":[{"/":"bafy2bzacedsp7zcxejngyftpu6hnr72ahsxyrycn5bjtfsp3vwxgvhs2hbpag"},{"/":"bafy2bzaceduo4goajlqtkarcozgm4hl6yla5kgajypguddk4embb6oxj2ah22"},{"/":"bafy2bzaceafs5ewv73nqxqhqqveanv74il5jxpvkp5oymoje6prtqh6b6hzje"}]},
```

FEVM:

```json
{"emitter":"t410fazjdjy3l32rrlnavap6uljckuvjl7f4ziticsca","entries":[
  [3,85,"t1",{"/":{"bytes":"mzhAeIhrixxtxKRVUgms9lTj+bEmu4rNhD4BOizSMmc"}}],
  [3,85,"t2",{"/":{"bytes":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADc"}}],
  [3,85,"d",{"/":{"bytes":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAKSfDFPBaMbbj1gAAAAAAAAAAAAAAAAAAAAAAAAAAAAApJ8MU8FoxtuPWAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABl7xEw"}}]
],"height":1427998,"msgCid":{"/":"bafy2bzaceahofvwlvtzcyfx75v26boqnai2t6q5uqogdb6bkepcfu2rqlmmyg"},"reverted":false,"tipsetKey":[{"/":"bafy2bzaceb7dl5fqr32biq75c6f5e5lwuc7ikdgrztmetweiiwahrgsmvflp4"},{"/":"bafy2bzacebjzpdk4aybiwfpcjs6fmkxf7me452f4o4tcs7wugiambehsnspde"},{"/":"bafy2bzacecw4f5n6xlw3jqui2pw7eoaga3cxyepisestfztbizcfnzwd5kise"}]},
```

Note the built-in forms become significantly more useful, no matter how you consume these. Even with `curl` you can use `jq` to mess with them without involving a CBOR decoder.

The changes being:

* "Tuple struct" formatting of the `EventEntry` so they're more compact, but we still retain flags and codec code but put them up front (`[flag,codec,key,value]`)
* We _decode_ the value using the codec (if we can) and present the value field as it comes out
* We use dag-json to do the JSONification, so we can properly represent bytes without the ambiguity of the Go standard JSON formatting (i.e. is ths string bytes as base64? is it at string that just happens to be base64?), but also allows us to represent objects as complex as we need (currently there's only scalars here and all of the FEVM events are `raw` for now so will always be bytes)).

About the complexity:

* This all works transparently with go-jsonrpc, these objects will round-trip faithfully, even with the value decoding
* This is currently an **option**, I was imagining it being a second argument to `GetActorEvents` and `SubscribeActorEvents`; but **if we land this before 1.26.0 final then it's not a breaking change and we could make it default**. If we do that, then maybe a third of the complexity in here can go away because this assumes being able to switch between, and gracefully handle, both forms.


chain/types/actor_event.go
```diff
@@ -65,3 +81,286 @@ type ActorEvent struct {
...
+	}
+	value, err := base64.StdEncoding.DecodeString(value64)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode base64 value: %w", err)
+	}
+	return &EventEntry{
+		Flags: uint8(flags),
+		Key:   key,
+		Codec: uint64(codec),
+		Value: value,
```
aarshkshah1992, line 220: @rvagg Where do we decode this using the CBOR decoder ?

aarshkshah1992: Oh I see it's in `eventEntryCompactFromAny`

chain/types/event.go
```diff
@@ -32,4 +32,7 @@ type EventEntry struct {
 	Value []byte
 }
 
+// TODO: implement EventEntry#UnmarshalJSON and EventEntry#MarshalJSON to allow for both compact and
+// non-compact forms as per ActorEvent#UnmarshalJSON and ActorEvent#MarshalJSON
+
```
aarshkshah1992, line 37: We can get rid of this if we make the compact form the only available form right ? The `ActorEvent` encoders and decoders already handle the compaction for Event entries.

rvagg: No, not quite, because there are existing API methods that use this via the `Event` type that we don't use here. Such as `ChainGetEvents`.

What I would do I think is implement an unmarshaller that takes either form.

Also, we should probably also do this same treatment for the filter, currently I'm constructing queries with filters like this:

```json
{"method":"Filecoin.GetActorEvents", "params": [{"fromHeight":1427974,"toHeight":-1,"fields":{"$type":[{"Codec":81,"Value":"cHZlcmlmaWVyLWJhbGFuY2U="},{"Codec":81,"Value":"amFsbG9jYXRpb24="},{"Codec":81,"Value":"cmFsbG9jYXRpb24tcmVtb3ZlZA=="},{"Codec":81,"Value":"ZWNsYWlt"},{"Codec":81,"Value":"bWNsYWltLXVwZGF0ZWQ="},{"Codec":81,"Value":"bWNsYWltLXJlbW92ZWQ="},{"Codec":81,"Value":"bmRlYWwtYWN0aXZhdGVk"},{"Codec":81,"Value":"b2RlYWwtdGVybWluYXRlZA=="},{"Codec":81,"Value":"bmRlYWwtY29tcGxldGVk"},{"Codec":81,"Value":"c3NlY3Rvci1wcmVjb21taXR0ZWQ="},{"Codec":81,"Value":"cHNlY3Rvci1hY3RpdmF0ZWQ="},{"Codec":81,"Value":"bnNlY3Rvci11cGRhdGVk"},{"Codec":81,"Value":"cXNlY3Rvci10ZXJtaW5hdGVk"}]}}],"id":1,"jsonrpc":"2.0"}
```

I'd like to do `[81,"sector-terminated"]`, and it would be nice if that was flexible and took either form too I think

rvagg: FYI, this is an existing API that we probably don't want to break:

```
$ curl -s -X POST   -H "Content-Type: application/json"   --data '{"method":"Filecoin.ChainGetEvents", "params": [{"/":"bafy2bzaceahc33tda6xtwxx7agspupiiwrgjwkniie5x3epz76tstv5nhdn5y"}],"id":1,"jsonrpc":"2.0"}' http://localhost:1235/rpc/v1
{"jsonrpc":"2.0","result":[{"Emitter":6,"Entries":[{"Flags":3,"Key":"$type","Codec":81,"Value":"amFsbG9jYXRpb24="},{"Flags":3,"Key":"id","Codec":81,"Value":"GXxO"},{"Flags":3,"Key":"client","Codec":81,"Value":"GQPz"},{"Flags":3,"Key":"provider","Codec":81,"Value":"GQPo"}]}],"id":1}
```

chain/types/actor_event.go
```diff
@@ -65,3 +81,286 @@ type ActorEvent struct {
 	// CID of message that produced this event.
 	MsgCid cid.Cid `json:"msgCid"`
 }
+
+// AsCompactEncoded will trigger alternate JSON encoding for ActorEvents, where the event entries
+// are encoded as a list of tuple representation structs, rather than a list of maps, values are
+// decoded using the specified codec where possible, and they are encoded using dag-json form so
+// bytes are represented using the `{"/":{"bytes":"base64"}}` form rather than Go standard base64
+// encoding.
+func (ae ActorEvent) AsCompactEncoded() ActorEvent {
```
aarshkshah1992, line 90: Why do we need this ? I don't see us not using this in any of the APIs. Is it so that we can ultimately allow users to specify this ? I think the best thing to do would be to ship this PR with 1.26 so we can get rid of this optionality.

aarshkshah1992:
@rvagg This is a major UX improvement for Actor Events and the code already looks really tight ! The biggest win for clients is not having to decode CBOR values.

While we shouldn't change anything on the smart contract events, shipping this for Actor events is a great win for the events work in 1.26.

chain/types/actor_event.go
arajasek: Yeah, I hate the switching here, agree with @aarshkshah1992

chain/types/actor_event.go
```diff
@@ -65,3 +81,286 @@ type ActorEvent struct {
...
+	}
+	builder := actorEventProto.Representation().NewBuilder()
+	if err := builder.AssignNode(nd); err != nil {
+		return err
+	}
+	aePtr := bindnode.Unwrap(builder.Build())
+	aec, _ := aePtr.(*ActorEvent) // safe to assume type
+	*ae = *aec
+
+	// check if we were encoded in compact form and set the flag accordingly
```
arajasek, line 109: I appreciate the completeness here, but I do think this is overkill.

chain/types/event.go
arajasek: Not super relevant to this PR, but it might be worth just deprecating the uncompacted format -- perhaps renaming the existing methods to `ChainGetEventsLegacy` (so easy breaking change to integrate initially), and then killing them in 3 months. I'm generally opposed to having two different formats for something so user-facing, it's only gonna be more detail for users to wrap their heads around.

Buuutt...future discussion, doesn't affect this changeset.

arajasek (commented):
This generally looks correct, and is honestly a great improvement. Thank you!

I would strongly favour landing this in 1.26.0, removing the need to switch between the two forms. That logic is...gnarly, and definitely doesn't seem worth the effort to me. I'm typically opposed to rushing things at this stage of a release, but this is a risk-free area of the code, we can correct any issues here fairly easily.

One option we can consider is to simply descope this entire system from the entire 1.26.0 release, and land it in master (or some tagged feature release). That can give us a bit more time to look at this / think this through, while still ensuring that the functionality is there on mainnet upgrade day (just not in the "minimum" release).

rvagg:
There's one tiny problem this doesn't deal with that ends up being a bit important—the case of a poorly encoded `Value` field from a user programmed actor. IIRC in here I deal with that case by just representing it as bytes. Unfortunately that case is lossy, as is the case of an unknown future codec that we're not prepared to deal with. You can't _quite_ round-trip because you've lost information about the original form. You couldn't use this for example to reconstruct an events AMT and get the event root in all cases (which isn't something we'd probably expect to be done, but it _could_ be done to validate that an API gave you the events you cared about).

From discussion today, the proposal is:

* Two separate API pairs: `GetActorEvents` + `SubscribeActorEvents` and `GetActorEventsRaw` + `SubscribeActorEventsRaw`.
* Separate return types, and likely separate filter parameters for both that give us the flexibility to give _nice_ decoded forms to the user except in the case where they want the raw, unaltered forms, perhaps to stick in a database, or perhaps to decode in their funky format without us molesting them.
* Drop `Flags` and `Codec` from the nice forms, if you want the gory details go and ask for it with the `Raw` forms.
* Ignore events that we can't decode - either because we don't recognise the codec or it's not decodable.

@arajasek's point about `ChainGetEvents` is a good one, maybe we do the same with it? `ChainGetEvents` vs `ChainGetEventsRaw`? Or we could just leave it alone.

As for descoping this for 1.26.0, that idea has merit except that one of the things we're aiming for is to help fill a data void around DDO, it wouldn't be the end of the world without it. But we are marking this clearly as "EXPERIMENTAL: may change in a future release". I'm suspecting we may even want to change the return value at some point as we find this one lacking; see https://github.com/filecoin-project/lotus/issues/11680 for some initial thoughts on what we may end up wanting to do.

rvagg:
Converted to draft for now, needs a bit of a rethink: this might be better if it's lossless (i.e. you could faithfully reverse the process to get the raw event data) and could break a little harder to deal with problems like pagination and signalling max-results that we don't have on the `*Raw()` variants.

rvagg:
will revisit this later, needs a bit more thought and work

PR #11717, opened 2024-03-13
chore: events: Prevent DDL re-execution during event index schema migrations

masih:
## Related Issues
* #11594 

## Proposed Changes

TL;DR: no net functionality change but a more conservative DDL execution for existing event index schema migrations.

This enhancement optimizes the schema migration process for the event index by preventing the redundant execution of Data Definition Language (DDL) statements that define the event schema. Traditionally, these DDL statements were grouped into a single slice, reflecting the most current version of the event index schema. With each migration, this slice was updated to the latest schema iteration, executing all statements in bulk. Initially, this method sufficed as migrations were focused on adding indices to existing table columns.

However, as the database schema evolves to meet new requirements, such as the forthcoming migrations that involve changes to table schemas (notably, indexing events by emitter actor ID instead of addresses), the prior approach of bulk execution of DDL statements becomes unsuitable: it will no longer be safe to repeatedly execute DDL statements in previous migrations, because the upcoming one changes `event` table column structure. To address this issue, the work here has isolated the event index schema migrations on a per-version basis. This adjustment ensures that only the necessary DDL statements are executed during each migration, avoiding the inefficiencies and potential errors associated with redundant executions.

## Additional Info
The work here:
* should also minimize the refactoring required for future migrations, in favour of smaller more manageable PRs.
* makes changes to the existing code to respect the provided context during migration execution.

## Checklist

- [x] CI is green

### DIFF
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -45,8 +45,8 @@ var ddls = []string{
 		reverted INTEGER NOT NULL
 	)`,
 
-	`CREATE INDEX IF NOT EXISTS height_tipset_key ON event (height,tipset_key)`,
-	`CREATE INDEX IF NOT EXISTS event_emitter_addr ON event (emitter_addr)`,
+	createIndexEventHeightTipsetKey,
+	createIndexEventEmitterAddr,
 
 	`CREATE TABLE IF NOT EXISTS event_entry (
 		event_id INTEGER,
@@ -57,11 +57,11 @@ var ddls = []string{
 		value BLOB NOT NULL
 	)`,
 
-	`CREATE INDEX IF NOT EXISTS event_entry_key_index ON event_entry (key)`,
+	createIndexEventEntryKey,
 
 	// metadata containing version of schema
 	`CREATE TABLE IF NOT EXISTS _meta (
-    	version UINT64 NOT NULL UNIQUE
+		version UINT64 NOT NULL UNIQUE
 	)`,
 
 	`INSERT OR IGNORE INTO _meta (version) VALUES (1)`,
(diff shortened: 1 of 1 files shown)


chain/events/filter/index.go
```diff
@@ -306,38 +355,21 @@ func NewEventIndex(ctx context.Context, path string, chainStore *store.ChainStor
...
-
 			err = eventIndex.migrateToVersion2(ctx, chainStore)
 			if err != nil {
 				_ = db.Close()
 				return nil, xerrors.Errorf("could not migrate sql data to version 2: %w", err)
 			}
-
-			// to upgrade to version version 2 we only need to create an index on the event table
-			// which means we can just recreate the schema (it will not have any effect on existing data)
-			for _, ddl := range ddls {
```
rvagg, line 318: k, so if I'm following the logic here correctly: you've duplicated index creation into the v2 migration so that they can be done in the same transaction+context so that we don't double up?

My only (minor) concern is that you now have duplicated all the `CREATE INDEX`es from the DDL into the migrations. Perhaps we could just define them once, and reference them in the main `ddl` slice and also in the migrations, just to avoid potential mismatch problems and set the precedent for the next upgrade?

masih: > duplicated index creation into the v2 migration

In both v2 and v3 migrations. 

> so that they can be done in the same transaction+context so that we don't double up?

such that the migration process does not blindly re-execute all of DDLs as part of migration. Instead it would execute just enough to reach the equivalent state as the DDL statements. The rationale is that DDL slice is a singular filed of event index; it represents the latest schema. If the schema migration requires table structure change, then we can no longer execute all DDL s as part of migrations v2 and v3. Hence this PR.

> reference them in the main `ddl`

Done. Please let me know if there is anything else that I might have missed.

rvagg (approved):

PR #11723, opened 2024-03-14
feat: events: Index events by emitter actor ID

### LINKED ISSUE #11594: Index Actor Events by ID

masih:
## Related Issues
Fixes #11594 

## Proposed Changes
Changes the chain event indexing mechanism to index events 
solely by their emitter actor ID, instead of their address. This change 
means: 1) index can accommodate non-EVM actors, and 2) improved
performance and efficiency due to reduced number of resolutions and
storage footprint.

## Additional Info
The work introduces a caching mechanism to momentarily cache the
resolutions to/fro actor ID and address. Further experimentation is
needed to tune the cache for optimal performance. For now, the
introduction of a unified cache reduces lookup time and memory
allocation foot print compared to the previous implementation where
ephemeral caches were being instantiated per event filter.

## Checklist

- [ ] CI is green


chain/events/filter/resolver.go
```diff
@@ -0,0 +1,118 @@
...
+		resolution.addr, resolution.err = c.delegate(ctx, k, ts)
+		c.cache.Add(key, resolution)
+	}
+	return resolution.addr, resolution.err
+}
+
+func (c *cachedTipSetInfoResolver[K, V]) newTipSetInfoKey(k K, ts *types.TipSet) string {
+	kk := c.keyer(k)
+	if ts == nil {
+		return string(kk)
```
aarshkshah1992, line 101: Why do we not use a pooled buffer if tipset is absent ?

chain/events/filter/store.go
```diff
@@ -15,7 +15,7 @@ import (
 type Filter interface {
 	ID() types.FilterID
 	LastTaken() time.Time
-	SetSubChannel(chan<- interface{})
+	SetSubChannel(chan<- any)
```
aarshkshah1992, line 18: Why do we need this change ?

chain/events/filter/resolver.go
```diff
@@ -0,0 +1,118 @@
+package filter
```
aarshkshah1992, line 1: We need some solid unit tests for this cache.

chain/events/filter/index.go
```diff
@@ -520,18 +655,26 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
...
-			values = append(values, addr.Bytes())
+	if len(emitters) > 0 {
+		subclauses := make([]string, 0, len(emitters))
+		for _, emitter := range emitters {
+			subclauses = append(subclauses, "emitter=?")
+			values = append(values, emitter)
 		}
 		clauses = append(clauses, "("+strings.Join(subclauses, " OR ")+")")
+		// Explicitly exclude reverted events, since at least one emitter is present and reverts cannot be considered.
+		excludeReverted = true
```
aarshkshah1992, line 672: What does this mean ? Why does having atleast one matching emitter with the filter lead us to exclude reverted events ?

chain/events/filter/index.go
```diff
@@ -236,39 +241,189 @@ func (ei *EventIndex) migrateToVersion2(ctx context.Context, chainStore *store.C
...
+	if err != nil {
+		return xerrors.Errorf("begin transaction: %w", err)
+	}
+	defer func() { _ = tx.Rollback() }()
 
-	log.Infof("Successfully migrated events to version 2 in %s", time.Since(now))
+	// create index on event.emitter_addr.
+	_, err = tx.ExecContext(ctx, "CREATE INDEX IF NOT EXISTS event_emitter_addr ON event (emitter_addr)")
+	if err != nil {
+		return xerrors.Errorf("create index event_emitter_addr: %w", err)
```
aarshkshah1992, line 306: Why are we changing code in prior migrations ? Is the existing idea of recreating everything so the indices also get recreated bad ?

aarshkshah1992 (commented):
Will do one more pass once the tests are up.

chain/events/filter/event.go
```diff
@@ -27,49 +27,49 @@ func isIndexedValue(b uint8) bool {
...
+	addresses []address.Address // list of actor addresses that are expected to emit the event
 
 	keysWithCodec map[string][]types.ActorEventBlock // map of key names to a list of alternate values that may match
 	maxResults    int                                // maximum number of results to collect, 0 is unlimited
 
 	mu        sync.Mutex
 	collected []*CollectedEvent
 	lastTaken time.Time
-	ch        chan<- interface{}
+	ch        chan<- any
```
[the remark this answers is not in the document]
masih: Less characters in code base and a more concise go code fit for the version of go supported by this repo.

In fact, I think we should do a mass find replace for all such instances across the repo.

chain/types/actor_event.go
```diff
@@ -47,9 +47,7 @@ type ActorEvent struct {
 	// Event entries in log form.
 	Entries []EventEntry `json:"entries"`
 
-	// Filecoin address of the actor that emitted this event.
-	// NOTE: In a future upgrade, this will change to always be an ID address. Currently this will be
-	// either the f4 address, or ID address if an f4 is not available for this actor.
+	// Emitter is the ID address of the actor that emitted this event.
 	Emitter address.Address `json:"emitter"`
```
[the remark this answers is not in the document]
masih: Happy to do it if there won't be any breaking changes or requirements for FIP?

chain/events/filter/store.go
masih: See earlier comment.

chain/events/filter/index.go
masih: Because we cannot confidently resolve the address for reverted events.

chain/events/filter/index.go
masih: Becase:
* this index is created for a column that will no longer exist once this PR lands, i.e. `emitter_addr`
* the `ddls` slice always points to the latest schema, i.e. the one that replaces `emitter_addr` column with `emitter` actor ID.

Therefore, we change the index creation statement that made sense for v3 here, where the `ddls` slice has the new version, i.e. index over `emitter` column.

chain/events/filter/resolver.go
masih: From design perspective, because `keyer` is beyond the scope of what poole buffer is used for. 

From implementation perspective, using pooling in the keyer won't make much difference since getting a key for an actor ID or address is implemented deep inside those structs. It would entail a larger refactor for little gain.

node/modules/actorevent.go
```diff
@@ -98,6 +98,43 @@ func EventFilterManager(cfg config.FevmConfig) func(helpers.MetricsCtx, repo.Loc
...
+		const (
+			// TODO: Consider introducing configuration parameters for constants below.
+			resolverCacheSize      = 1024
+			resolverCacheExpiry    = time.Minute
+			resolverCacheNilTipSet = false
+		)
+		actorResolver := filter.NewCachedActorResolver(
+			func(ctx context.Context, emitter address.Address, ts *types.TipSet) (abi.ActorID, error) {
+				var addr address.Address
+				if emitter.Protocol() == address.ID {
```
arajasek, line 110: Consider moving this short-circuit logic into the `StateManager` itself? Any caller of that method should benefit from it.

node/modules/actorevent.go
```diff
@@ -98,6 +98,43 @@ func EventFilterManager(cfg config.FevmConfig) func(helpers.MetricsCtx, repo.Loc
...
+					addr = emitter // already an ID address
+				} else {
+					var err error
+					addr, err = sm.LookupID(ctx, emitter, ts)
+					if err != nil {
+						return 0, err
```
arajasek, line 116: nit: chain errors, it'll make your debugging life easier.

chain/events/filter/resolver.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/lotus/chain/types"
+)
+
+type (
+	// AddressResolver resolves an actor ID to its address at a given TipSet, or the latest TipSet if nil.
+	AddressResolver func(context.Context, abi.ActorID, *types.TipSet) (address.Address, error)
```
arajasek, line 20: So, this needs to be specified a bit more. Unfortunately, in the post-FEVM land, an actor can have _both_ a "delegated" (f4) address AND an "actor" (f2) address. So basically, when "resolving" an ID address, the possibilities are:
ID -> f1 (secp address)
ID -> f3 (BLS address)
ID -> f2 ("actor" address, this is true for miners, multisigs, etc.)
ID -> f2 AND f4 ("actor" and "delegated" address, I think this is true of EVM actors)
(and maybe) ID -> f4 (just a "delegated" address, this is true of...EthAccounts? I'm not sure of this one.)

So, when doing any resolution starting from an ID address, you need to decide which one you want to prefer if _both_ are present. My instinct is to say that you should prefer the f4 address in this case (it exists for Ethereum-like behaviour, so we might as well use it for Ethereum events, while falling back on robust f2 addresses for everything else). But this is probably more a @Stebalien question.

(This wasn't a problem before, because ONLY Ethereum-like things were emitting events, so the AddressResolver could exclusively rely on f4 addresses).

node/modules/actorevent.go
```diff
@@ -98,6 +98,43 @@ func EventFilterManager(cfg config.FevmConfig) func(helpers.MetricsCtx, repo.Loc
...
+				}
+				return abi.ActorID(actor), nil
+			}, resolverCacheSize, resolverCacheExpiry, resolverCacheNilTipSet)
+		addressResolver := filter.NewCachedAddressResolver(
+			func(ctx context.Context, emitter abi.ActorID, ts *types.TipSet) (address.Address, error) {
+				idAddr, err := address.NewIDAddress(uint64(emitter))
+				if err != nil {
+					return address.Undef, err
+				}
+				actor, err := sm.LoadActor(ctx, idAddr, ts)
```
arajasek, line 131: See note in `resolver.go` above. This is likely incorrect.

arajasek (commented):
I left a ramble below on the different possible address resolutions mostly for informative purposes, but I'm a little confused about why we need an `AddressResolver` at all. I fee like if we're doing this correctly, we should _only_ ever be going in one direction (to ID addresses).

chain/events/filter/event.go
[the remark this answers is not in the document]
Stebalien: Eventually. But ideally after we merge everything we can.

chain/events/filter/resolver.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/hashicorp/golang-lru/v2/expirable"
+	"lukechampine.com/blake3"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/lotus/chain/types"
+)
+
+type (
```
Stebalien, line 17: We do this for consts but not types.

chain/events/filter/resolver.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+)
+
+var cacheHasherPool = sync.Pool{
+	New: func() any {
+		return blake3.New(32, nil)
+	},
+}
+
+func NewCachedActorResolver(delegate ActorResolver, size int, expiry time.Duration, cacheNilTipSet bool) ActorResolver {
```
Stebalien, line 38: In terms of caching, I believe we will get a bigger speed up by keeping the same state tree around while resolving addresses within a single tipset.

Stebalien: But this probably is not a bottleneck.

chain/events/filter/index.go
```diff
@@ -317,11 +322,111 @@ func (ei *EventIndex) migrateToVersion3(ctx context.Context) error {
...
+			emitterAddr []byte
+		}
+		if err := rows.Scan(&row.id, &row.emitterAddr); err != nil {
+			return xerrors.Errorf("read event emitter_addr row: %w", err)
+		}
+		addr, err := address.NewFromBytes(row.emitterAddr)
+		if err != nil {
+			return xerrors.Errorf("parse emitter_addr: %w", err)
+		}
+		emitter, err := actorResolver(ctx, addr, head)
```
Stebalien, line 382: I'd implement address resolution manually here. Ideally, we'd load the init actor once (https://github.com/filecoin-project/lotus/blob/2e75f3b796dd1396baa37d73ea99dc3407b4da8d/chain/actors/builtin/init/init.go#L32, , then load the address map once (https://github.com/filecoin-project/lotus/blob/2e75f3b796dd1396baa37d73ea99dc3407b4da8d/chain/actors/builtin/init/v13.go#L117), then use that for the rest of this function.

chain/events/filter/index.go
Stebalien: We should just always exclude reverted events when prefilling (for now). The option to not exclude them only exists for testing.

I've filed an issue. https://github.com/filecoin-project/lotus/issues/11770

chain/stmgr/stmgr.go
```diff
@@ -570,3 +570,21 @@ func (sm *StateManager) GetRandomnessDigestFromTickets(ctx context.Context, rand
 
 	return r.GetChainRandomness(ctx, randEpoch)
 }
+
+func (sm *StateManager) ResolveActorID(ctx context.Context, emitter address.Address, ts *types.TipSet) (abi.ActorID, error) {
```
Stebalien, line 574:
```suggestion
func (sm *StateManager) ResolveActorID(ctx context.Context, addr address.Address, ts *types.TipSet) (abi.ActorID, error) {
```

chain/stmgr/stmgr.go
```diff
@@ -570,3 +570,21 @@ func (sm *StateManager) GetRandomnessDigestFromTickets(ctx context.Context, rand
...
+	var addr address.Address
+	if emitter.Protocol() == address.ID {
+		addr = emitter // already an ID address
+	} else {
+		var err error
+		addr, err = sm.LookupID(ctx, emitter, ts)
```
Stebalien, line 580: If we introduce a new function, we should make a lookup ID use this new function instead of the other way around.

I would also prefer to do a pre-factor where we rename `LookupID` to `LookupIDAddress`.

chain/types/actor_event.go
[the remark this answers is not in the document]
Stebalien: Is an API breaking change, not a networking change. Not great, but it's allowed.

Thoughts @rvagg?

chain/types/ethtypes/eth_types.go
```diff
@@ -642,7 +643,7 @@ func (e *EthAddressList) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-// TopicSpec represents a specification for matching by topic. An empty spec means all topics
+// EthTopicSpec represents a specification for matching by topic. An empty spec means all topics
```
Stebalien, line 646: I have a lot of little cleanups here. Can we break them out into separate pull requests? Maybe a few pull requests.

chain/events/filter/event.go
```diff
@@ -27,49 +27,48 @@ func isIndexedValue(b uint8) bool {
...
 }
 
 type eventFilter struct {
+	ActorResolver
 	id        types.FilterID
 	minHeight abi.ChainEpoch // minimum epoch to apply filter or -1 if no minimum
 	maxHeight abi.ChainEpoch // maximum epoch to apply filter or -1 if no maximum
 	tipsetCid cid.Cid
-	addresses []address.Address // list of actor addresses that are extpected to emit the event
+	addresses []address.Address // list of actor addresses that are expected to emit the event
```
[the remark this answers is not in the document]
masih: That means an event filter would only support filtering by ID and no longer by address. 

Is that OK?

chain/events/filter/index.go
masih: > Can we just drop this index from this migration?

Not sure I follow. That index is the whole point of [Version 3 migration](https://github.com/filecoin-project/lotus/commit/95fb198597b8d996b396281da8c975f5c9c858b6).

Version 4, i.e. the changes in this PR, [drop it](https://github.com/filecoin-project/lotus/blob/ba2e7a19b8d911ecb945a41e87d0c9df39f908e6/chain/events/filter/index.go#L400) first since then [recreate it](https://github.com/filecoin-project/lotus/blob/ba2e7a19b8d911ecb945a41e87d0c9df39f908e6/chain/events/filter/index.go#L410) on the newly populated `emitter` column if that's what you are suggesting?

chain/stmgr/stmgr.go
masih: > we should make a lookup ID use this new function instead of the other way around.

This would mean translating address to ID and back again in `LookupID`, because the underlaying `StateTree.LookupID` API returns address (as well as state snapshot APIs). Right?

chain/events/filter/index.go
masih: @Stebalien I'd love your help on how to keep the index migration tests running after this change. I _think_ I'll have to export `lookupIDFunc` in `StateTree`, unless there is another way you can think of?

chain/events/filter/event.go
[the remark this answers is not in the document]
Stebalien: Ah, sorry, I was not paying attention. Keep it as is.

chain/stmgr/stmgr.go
Stebalien: > OK to refactor them too?

Yes, although it would be great if this could be done in a separate patch.

> This would mean translating address to ID and back again in LookupID, because the underlaying StateTree.LookupID API returns address (as well as state snapshot APIs). Right?

Ah, sorry, I mean the new `LookupIDAddress` function. That is, we take the current `LookupID` function and split it into a `LookupIDAddress` function and a `LookupID` function and have the `LookupIDAddress` address function called `LookupID` function.

chain/events/filter/event.go
```diff
@@ -188,19 +179,29 @@ func (f *eventFilter) matchTipset(te *TipSetEvents) bool {
...
+		if err != nil {
+			// Cannot match against addr; skip.
+			continue
 		}
+		emitters[emitter] = struct{}{}
 	}
-	return false
+	if len(emitters) == 0 {
+		// There are no actors that match the specified addresses in the current tipset.
+		return nil, xerrors.New("no actors found for any addresses")
```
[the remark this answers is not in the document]
rvagg: Maybe this function needs a third return, bool, to indicate this case - we don't want it to fall back to matching all addresses. For the prefill: if you've specified an address and it doesn't resolve, you don't match anything. If you've specified 2 addresses and one resolves, you get events for the one that does. After prefill, we could keep trying to look up the address but still shortcut out of matching anything if there are non resolving for that tipset.

chain/events/filter/index.go
rvagg: I suppose that once this code goes live, then the only pathway leads up to a v4 schema anyway and you have a "drop if exists" for this index in there, so removing it entirely from a v3 migration when v4 is live is basically a noop as far as the end result is concerned. If you run this code on a v2 (which I think 1.26.0 is), you'll end up with a v4 anyway and this index will have been temporarily added then removed to get there.

chain/events/filter/index.go
rvagg: I think you'd just have to use the `chainStore` directly for this - wrap it in `cbor.NewCborStore`, wrap that in the `AdtStore` like `StateTree#lookupIDinternal` does and proceed from there, `Load`ing the init HAMT and doing a `ResolveAddress` on init state while holding on to the HAMT the whole time. For testing, just build a custom HAMT to hold the addresses you want and pass it in via the `chainStore`. Sounds complicated but would that not work here?

rvagg: I wonder though is there a path to having an address that _doesn't_ resolve to a new ActorID at this point though because the error is a hard fatal and will be annoying for users to encounter.

If there was a revert or two after you shut down your node, upgraded your lotus and tried to do a migration to v4 and encountered an address that failed to resolve because your head tipset is on a diverging branch? You'd be in trouble. Maybe it would be good in this case to error log it but drop the row entirely?

Is there a _right_ thing to do in this case? If we fail to look one up, should we try to load the state from the tipset in the event line and fish around in its init actor?

chain/events/filter/index.go
```diff
@@ -317,11 +322,111 @@ func (ei *EventIndex) migrateToVersion3(ctx context.Context) error {
...
+		return xerrors.Errorf("drop event.emitter_addr column: %w", err)
+	}
+
+	// Create a new index index fo event.emitter, replacing the deleted event.emitter_addr.
+	if _, err = tx.ExecContext(ctx, createIndexEventEmitter); err != nil {
+		return xerrors.Errorf("create event_emitter index: %w", err)
+	}
+
+	// Increment the schema version in _meta table to 4.
+	if _, err = tx.Exec("INSERT OR IGNORE INTO _meta (version) VALUES (4)"); err != nil {
```
rvagg, line 415: do these `OR IGNORE`s make sense to you for the versions in the migrations? I think I'd want to know if I got here and it was already at `4`, even if it was an error log and not fatal.

chain/events/filter/index.go
```diff
@@ -557,18 +658,26 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 		}
 	}
 
-	if excludeReverted {
-		clauses = append(clauses, "event.reverted=?")
-		values = append(values, false)
+	// Resolve emitters from filter addresses at the latest TipSet.
+	// TODO investigate if it is safe to use the current tipset; what about finality?
```
rvagg, line 662: Can you elaborate on the concern here? I would imagine that for prefill that you just want a best guess, and current head is that best guess and should represent the correct state as far as we know it up until this point. Is there even a reasonable alternative to doing this?

chain/events/filter/index.go
rvagg: I think until we've done proper work to resolve #11770 (see thread there) you'd have to _include_ reverted even in the case of having emitters to consider. I think you need to give maximal information to someone seeking to do a look-back because they're likely trying to catch up _unless_ they exclude reverted themselves in which case they are not doing a "tell me what happened since I last looked" catch-up but a "tell me how it is now".

chain/events/filter/resolver.go
rvagg: I can imagine this being helpful for the case of a public node operator serving event lookups for popular event types and addresses. It's probably not going to help much for general use but should scale in a multi-user situation.

chain/types/actor_event.go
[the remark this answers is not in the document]
rvagg: As preference I think I like `Address` here, but that's not a strong opinion. I don't like breaking APIs very much though, this would go from a JSON string to an int.

`git grep abi.ActorID -- api/` shows it being used directly in only 2 API calls, for miner worker. `git grep abi.ActorID -- chain/types` shows almost nothing. **Except** `type Event` which is essentially what we extend to make `ActorEvent` by adding tipset and message info. It's also what we return on `ChainGetEvents`, so that API is returning an int. So there's an argument for consistency there.

:shrug: I introduced `*Raw()` so we could break stuff when we do non-`Raw()` forms of them because breakage is just annoying.

How about this: I think we're going to have to rework a lot of these API guts anyway to make it tipset focused, kind of like https://github.com/filecoin-project/lotus/issues/11680 and what's been described in https://github.com/filecoin-project/lotus/issues/11770#issuecomment-2015678344. So instead of breaking this now, leave a note in here about it may break in the future to become an `ActorID`.

chain/events/filter/index.go
masih: The issue with that is address resolution; events database may contain reverted events without emitter actor ID. Therefore, we cannot confidently infer whether those events should be included in pre-filling or not. Does that make sense?

chain/events/filter/index.go
masih: I am not sure if it is worth making optimisations here. Instead, my preference is to execute every migration unless the schema does not exist. That simple rule would make it a lot easier to reason about stateful migrations in general.

Unless there is significant performance overhead, my sense is to stick to the above migration rule.

chain/events/filter/index.go
masih: It keeps to the existing convention that's present across previous schema migrations. Apart from that removing `OR IGNORE` here will not cause the migration to fail. Right?

Other than that, the scenario you described is not currently possible through the code executed by Lotus. Unless the user modifies the DB externally while the migration is running or runs two instances of lotus at the same time.

chain/stmgr/stmgr.go
masih: @Stebalien The changes you suggested above is causing enough churn across the code and diverges enough from the purpose of this PR that I prefer doing it in a separate PR. Unless you think it's a blocker for this one?

chain/events/filter/index.go
Stebalien: Well, after the migration, they'll have IDs, they just won't have addresses. But you're right that we won't be able to confidently pre-fill (quickly, at least).

I'm going to have to disagree with @rvagg here. Our handling of reverted events in the current APIs is not very useful, as discussed in the linked issue.

chain/events/filter/index.go
Stebalien: I'm more concerned about keeping the code clean and easy to read. I'm not a fan of moving this SQL statement into this migration as a string literal and then just tossing the result.

masih: IMHO reasoning about immutable migrations is much easier than trying to optimise it per new migration. 
But it looks like I am out voted here. So I i'll drop the index migration at v3 in favour of combining it into v4.

chain/types/actor_event.go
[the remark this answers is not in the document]
masih: @Stebalien Just to clarify, your preference is to user `ActorID` type instead of `address.Address`. Correct?

chain/events/filter/event.go
```diff
@@ -27,49 +27,48 @@ func isIndexedValue(b uint8) bool {
...
 type EventFilter interface {
 	Filter
 
 	TakeCollectedEvents(context.Context) []*CollectedEvent
-	CollectEvents(context.Context, *TipSetEvents, bool, AddressResolver) error
+	CollectEvents(context.Context, *TipSetEvents, bool) error
 }
 
 type eventFilter struct {
+	ActorResolver
```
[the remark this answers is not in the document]
masih: Unexported `ActorResolver`.

> why add this field on the event filter itself?

I think either event filter keeps a reference to manager to access the resolver or takes the resolver reference directly. The latter seems like a more concise reference for what it is used today.

chain/stmgr/stmgr.go
masih: Adressed here: https://github.com/filecoin-project/lotus/pull/11919

chain/types/actor_event.go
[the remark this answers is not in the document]
Stebalien: Sorry, I meant leaving it as an `Address` (I was agreeing with rod). I'd kind of prefer ActorID from an engineering standpoint, but Address is better from a consistency standpoint.

chain/events/filter/index.go
```diff
@@ -364,6 +370,149 @@ func (ei *EventIndex) migrateToVersion4(ctx context.Context) error {
 	ei.vacuumDBAndCheckpointWAL(ctx)
 
 	log.Infof("Successfully migrated events to version 3 in %s", time.Since(now))
+	log.Infof("successfully migrated event index from version 2 to version 3 in %s", time.Since(now))
+	return nil
+}
+
+// migrateToVersion4 migrates the schema from version 3 to version 4: indexing events by emitter actor ID.
```
rvagg, line 377:
```suggestion
// migrateToVersion5 migrates the schema from version 4 to version 3: indexing events by emitter actor ID.
```

chain/events/filter/index.go
rvagg: OK, so the alternative you're suggesting would be to rewind to finality and resolve on that?

I think we might be getting into the realm of guessing user-intention here unfortunately and we don't have enough signal for that—we'd probably need a "resolve tipset" to get it exactly the way they intended. I think as long as we're doing it consistently against the latest tipset then we should be pretty likely to get what the user wants. Although my experience doesn't give me any useful hypotheticals to test that theory against! If it ever comes up as a complaint then maybe we consider the options then?

chain/events/filter/index.go
```diff
@@ -627,20 +773,29 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
...
+		for emitter := range emitters {
+			subclauses = append(subclauses, "emitter=?")
+			values = append(values, emitter)
 		}
 		clauses = append(clauses, "("+strings.Join(subclauses, " OR ")+")")
 	}
 
+	// Always exclude reverted events when prefilling. See:
+	// - https://github.com/filecoin-project/lotus/issues/11770
+	clauses = append(clauses, "event.reverted=?")
```
rvagg, line 796: I disagree with this as an outcome to 11770, see https://github.com/filecoin-project/lotus/issues/11770#issuecomment-2015678344

I think we have to leave this out for now and resolve it separately. I like the idea of an API where you can say "this is where I was last, catch me up". But we don't have that yet so the best we can do is get everything from the API and build a state representation locally by consuming reverted events to undo existing state we have already collected.

```suggestion
```

chain/events/filter/index.go
```diff
@@ -364,6 +370,149 @@ func (ei *EventIndex) migrateToVersion4(ctx context.Context) error {
 	ei.vacuumDBAndCheckpointWAL(ctx)
 
 	log.Infof("Successfully migrated events to version 3 in %s", time.Since(now))
+	log.Infof("successfully migrated event index from version 2 to version 3 in %s", time.Since(now))
```
rvagg, line 373:
```suggestion
	log.Infof("successfully migrated event index from version 3 to version 4 in %s", time.Since(now))
```

bad rebase I think, and removing the vacuum from this migration since you're doing it again in 5

chain/events/filter/index.go
```diff
@@ -364,6 +370,149 @@ func (ei *EventIndex) migrateToVersion4(ctx context.Context) error {
...
+		return xerrors.Errorf("increment _meta version: %w", err)
+	}
+
+	if err = tx.Commit(); err != nil {
+		return xerrors.Errorf("commit transaction: %w", err)
+	}
+
+	ei.vacuumDBAndCheckpointWAL(ctx)
+
+	log.Infof("successfully migrated event index from version 3 to version 4 in %s", time.Since(now))
```
rvagg, line 515:
```suggestion
	log.Infof("Successfully migrated event index from version 4 to version 5 in %s", time.Since(now))
```

chain/events/filter/index.go
```diff
@@ -364,6 +370,149 @@ func (ei *EventIndex) migrateToVersion4(ctx context.Context) error {
...
+		return xerrors.Errorf("drop event.emitter_addr column: %w", err)
+	}
+
+	// Create a new index index fo event.emitter, replacing the deleted event.emitter_addr.
+	if _, err = tx.ExecContext(ctx, createIndexEventEmitter); err != nil {
+		return xerrors.Errorf("create event_emitter index: %w", err)
+	}
+
+	// Increment the schema version in _meta table to 4.
+	if _, err = tx.Exec("INSERT OR IGNORE INTO _meta (version) VALUES (4)"); err != nil {
```
rvagg, line 505:
```suggestion
	if _, err = tx.Exec("INSERT OR IGNORE INTO _meta (version) VALUES (5)"); err != nil {
```

rvagg:
This seems pretty close, albeit a scary migration and big change.

With a few of my diff suggestions applied and rebased to current `master` I've run it on my calibnet node which has a complete events db (the 3->4 upgrade timing is not quite right cause I already manually added the indexes from that migration in this db):

```json
{"level":"info","ts":"2024-04-30T14:59:54.325+1000","logger":"filter","caller":"filter/index.go:588","msg":"Upgrading event index from version 3 to version 4"}
{"level":"info","ts":"2024-04-30T14:59:54.325+1000","logger":"filter","caller":"filter/index.go:370","msg":"Successfully migrated event index from version 3 to version 4 in 31.81µs"}
{"level":"info","ts":"2024-04-30T14:59:54.325+1000","logger":"filter","caller":"filter/index.go:598","msg":"Upgrading event index from version 4 to version 5"}
{"level":"info","ts":"2024-04-30T14:59:59.709+1000","logger":"filter","caller":"filter/index.go:520","msg":"Performing DB vacuum and wal checkpointing to free up space after the migration"}
{"level":"info","ts":"2024-04-30T15:00:07.075+1000","logger":"filter","caller":"filter/index.go:512","msg":"Successfully migrated event index from version 4 to version 5 in 12.749369829s"}
```

So 👍 that the migration was successful at least.

I then ran a couple of queries before and after to see what the difference is:

Just built-in actor events:

```
curl -s -X POST -H "Content-Type: application/json" --data '{"method":"Filecoin.GetActorEventsRaw","params":[{"fromHeight":1441214,"toHeight":1444094,"fields":{"$type":[{"Codec":81,"Value":"cHZlcmlmaWVyLWJhbGFuY2U="},{"Codec":81,"Value":"amFsbG9jYXRpb24="},{"Codec":81,"Value":"cmFsbG9jYXRpb24tcmVtb3ZlZA=="},{"Codec":81,"Value":"ZWNsYWlt"},{"Codec":81,"Value":"bWNsYWltLXVwZGF0ZWQ="},{"Codec":81,"Value":"bWNsYWltLXJlbW92ZWQ="},{"Codec":81,"Value":"bmRlYWwtcHVibGlzaGVk"},{"Codec":81,"Value":"bmRlYWwtYWN0aXZhdGVk"},{"Codec":81,"Value":"b2RlYWwtdGVybWluYXRlZA=="},{"Codec":81,"Value":"bmRlYWwtY29tcGxldGVk"},{"Codec":81,"Value":"c3NlY3Rvci1wcmVjb21taXR0ZWQ="},{"Codec":81,"Value":"cHNlY3Rvci1hY3RpdmF0ZWQ="},{"Codec":81,"Value":"bnNlY3Rvci11cGRhdGVk"},{"Codec":81,"Value":"cXNlY3Rvci10ZXJtaW5hdGVk"}]}}],"id":1,"jsonrpc":"2.0"}' http://localhost:1235/rpc/v1
```

This one produced no difference, so that's good! They're already actor IDs anyway.

Same epoch span, but all events:

```
curl -s -X POST -H "Content-Type: application/json" --data '{"method":"Filecoin.GetActorEventsRaw","params":[{"fromHeight":1441214,"toHeight":1444094}],"id":1,"jsonrpc":"2.0"}' http://localhost:1235/rpc/v1
```

This one has differences, but formatting JSON and diffing just shows the addresses changed.

```
...
---
>   "emitter": "t067651",
455026c455026
<   "emitter": "t410f43gi2ue3ffyued2tnzggj4bp46sbm4oqefvwm7i",
---
>   "emitter": "t067646",
455065c455065
<   "emitter": "t410fazjdjy3l32rrlnavap6uljckuvjl7f4ziticsca",
---
>   "emitter": "t067651",
```

However, using the more common `eth_getLogs` for the same range, fails:

```
curl -s -X POST -H "Content-Type: application/json" --data '{"method":"Filecoin.EthGetLogs","params":[{"fromBlock":"0x15fdbe","toBlock":"0x1608fe"}],"id":1,"jsonrpc":"2.0"}' http://localhost:1235/rpc/v1
```

```json
{"jsonrpc":"2.0","id":1,"error":{"code":1,"message":"state get actor: actor not found"}}
```

Logs have:

```json
{"level":"warn","ts":"2024-04-30T15:21:51.735+1000","logger":"rpc","caller":"go-jsonrpc@v0.3.1/handler.go:444","msg":"error in RPC call to 'Filecoin.EthGetLogs': state get actor:\n    github.com/filecoin-project/lotus/node/impl/full.ethFilterResultFromEvents\n        /home/rvagg/go/src/github.com/filecoin-project/lotus/node/impl/full/eth_events.go:123\n  - actor not found"}
```

Extra debugging shows it tripping up on this event:

```json
{"level":"warn","ts":"2024-04-30T15:34:02.894+1000","logger":"fullnode","caller":"full/eth_events.go:124","msg":"failed to get actor","actor":"t084452","event":{"Entries":[{"Flags":3,"Key":"t1","Codec":85,"Value":"j6pwh4ZxzNIS0gdxt5XFCvj9P/bPJ/S95X5dTeCutnM="},{"Flags":3,"Key":"d","Codec":85,"Value":"<snipped>"}],"Emitter":84452,"EventIdx":1,"Reverted":false,"Height":1441489,"TipSetKey":[{"/":"bafy2bzacedlxximvilreromjdsbqdd5irj7j6plhu57rm64xexkhusptdorj2"},{"/":"bafy2bzacedi3vqjodd63k7cycvvnnjvmcdkkcfvmhn67bus7szwc4a4x42uda"}],"MsgIdx":1,"MsgCid":{"/":"bafy2bzacebl5mfp7ekw56xdetv7vscukqnbhos5t7jbg3wmtb3z4nmvstnkfw"}},"error":"actor not found"}
```

Which isn't a reverted event, but apparently not found in state. I'll investigate further after I fix an error I found with my v4 migration!

rvagg:
OK, here's the problem:

The actor exists at the current head:

```sh
curl -s -X POST -H "Content-Type: application/json" --data '{"method":"Filecoin.StateGetActor","params":["t084452",[]],"id":1,"jsonrpc":"2.0"}' http://localhost:1235/rpc/v1
```

```json
{"jsonrpc":"2.0","result":{"Code":{"/":"bafk2bzacec5ibmbtzuzjgwjmksm2n6zfq3gkicxqywwu7tsscqgdzajpfctxk"},"Head":{"/":"bafy2bzacebezeusg7k3z4g2j2cl3uvx5wvydgvbn7f2p5hzgu3mmvfmn4ldqw"},"Nonce":0,"Balance":"0","Address":"t410fczebosgrdb3oapy37zzmwpli7o5tjdvi7yfv2dy"},"id":1}
```

But it doesn't exist at the tipset of the event:

```sh
curl -s -X POST -H "Content-Type: application/json" --data '{"method":"Filecoin.StateGetActor","params":["t084452",[{"/":"bafy2bzacedlxximvilreromjdsbqdd5irj7j6plhu57rm64xexkhusptdorj2"},{"/":"bafy2bzacedi3vqjodd63k7cycvvnnjvmcdkkcfvmhn67bus7szwc4a4x42uda"}]],"id":1,"jsonrpc":"2.0"}' http://localhost:1235/rpc/v1
```

```json
{"jsonrpc":"2.0","id":1,"error":{"code":1,"message":"actor not found"}}
```

But if we fast-forward by one epoch, it does exist:

```sh
curl -s -X POST -H "Content-Type: application/json" --data '{"method":"Filecoin.StateGetActor","params":["t084452",[{"/":"bafy2bzaceckcfvpj5uxarwoky37uinq2kcbxuijliisloxvopf2xc75mqvvfc"},{"/":"bafy2bzacebjhy3nw3abeasz2symc4y4ze4w76tdgpl6b4e23j6unhr2qmqwym"},{"/":"bafy2bzaceavtrohq4y7pprr3m4ecjxauo66j3xxfcq6harwpg3qxus2qbxepg"},{"/":"bafy2bzacec27r2slvaysjzqac4k6oy5yesqogpqgoncvl3phrfkdkcl6uqu5g"}]],"id":1,"jsonrpc":"2.0"}' http://localhost:1235/rpc/v1
```

```json
{"jsonrpc":"2.0","result":{"Code":{"/":"bafk2bzacec5ibmbtzuzjgwjmksm2n6zfq3gkicxqywwu7tsscqgdzajpfctxk"},"Head":{"/":"bafy2bzacedyonsvsf6f3ufayfdaaiepvl23dt2c5btfbmhlqlcpepetqavirk"},"Nonce":0,"Balance":"0","Address":"t410fczebosgrdb3oapy37zzmwpli7o5tjdvi7yfv2dy"},"id":1}
```

Deferred execution yay.

I'm not sure the best way to deal with this. Perhaps we could:

1. Call `StateGetActor` without a tipset and cross our fingers? That probably won't work for reverted events though.
2. Use a combination of `GetTipSetFromKey` and `GetTipSetFromHeight` to get the tipset+1 and use that when we call `StateGetActor`?
3. ... something else I'm not aware of that we can do?

chain/types/actor_event.go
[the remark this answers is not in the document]
masih: Got it thanks, I too think it should (eventually) be ActorID but it seems like we use address more consistently across Lotus APIs than ActorID even though in this case the address should/will always be an ID address.

chain/events/filter/index.go
masih: Progressed with always excluding reverted events during backfill. Please mark as unresolved if there is anything that I have missed here. thanks

chain/events/filter/index.go
masih: Right thanks

rvagg:
I just found this, which will need to be included in this PR: https://github.com/filecoin-project/lotus/blob/6bbe09041101845b3636f89390e2e2320cbd9af8/cmd/lotus-shed/indexes.go#L110

rvagg:
OK, with https://github.com/filecoin-project/lotus/pull/11953 applied, the `Filecoin.EthGetLogs` (`eth_getLogs`) call above works now, which is good.

However, it takes 1.16 seconds to get 10198 events.

Without this PR and needing to do the address translation on request, the same call for 10198 events takes 0.56s seconds to fetch. So we've doubled the time due to the need to load state and map addresses on output.

masih:
That's the time measurement on a warm instance with caches populated?

Stebalien:
> However, it takes 1.16 seconds to get 10198 events.

The issue is likely that we're constantly re-loading the init actor and address map. If we could cache that (even within a single trace), we'd get much better performance (https://github.com/filecoin-project/lotus/pull/11723#discussion_r1534688141).

Although this is something we should be able to look at with a pprof profile.

masih:
Time to either close this PR and salvage what's useful or get it landed. 

@rvagg is this PR still relevant/worth brining up-to-date or shall we close it off?

rvagg:
I've raised in #12421 that we may want to consider pulling it into there. I still don't know. After tinkering with this stuff for months now I don't really feel any clearer on whether this change is worth it for anything other than saving bytes on disk! It just seems like we're shifting the complexity around whichever way we do it. Maybe @Stebalien can enlighten us before #12421 lands because doing it when while replacing the indexer entirely would be a good time to make it happen so it doesn't need yet another migration.

masih:
Reopen if needed, or salvage what you will from this PR.

PR #11700, opened 2024-03-11
feat: add StateMinerDeadlines to gateway

qwdsds:
## Proposed Changes

Add StateMinerDeadlines to gateway.

https://github.com/filecoin-project/lotus/discussions/11693

## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] If the PR affects users (e.g., new feature, bug fix, system requirements change), update the CHANGELOG.md and add details to the UNRELEASED section.
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #11725, opened 2024-03-15
fix(events,gateway): check that filter.FromHeight isn't too far back for gateway calls

rvagg:
TIL about this limiting (from #11700), this seems appropriate to do for the new events calls.

### DIFF
--- a/gateway/proxy_fil.go
+++ b/gateway/proxy_fil.go
@@ -441,12 +441,22 @@ func (gw *Node) GetActorEvents(ctx context.Context, filter *types.ActorEventFilt
 	if err := gw.limit(ctx, stateRateLimitTokens); err != nil {
 		return nil, err
 	}
+	if filter != nil && filter.FromHeight != nil {
+		if err := gw.checkTipSetHeight(ctx, *filter.FromHeight, types.EmptyTSK); err != nil {
+			return nil, err
+		}
+	}
 	return gw.target.GetActorEvents(ctx, filter)
 }
 
 func (gw *Node) SubscribeActorEvents(ctx context.Context, filter *types.ActorEventFilter) (<-chan *types.ActorEvent, error) {
 	if err := gw.limit(ctx, stateRateLimitTokens); err != nil {
 		return nil, err
+	}
+	if filter != nil && filter.FromHeight != nil {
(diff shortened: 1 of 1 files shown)


gateway/proxy_fil.go
```diff
@@ -441,13 +441,23 @@ func (gw *Node) GetActorEvents(ctx context.Context, filter *types.ActorEventFilt
 	if err := gw.limit(ctx, stateRateLimitTokens); err != nil {
 		return nil, err
 	}
+	if filter != nil && filter.FromHeight != nil {
```
Stebalien, line 444: I guess "filter == nil" means "current tipset"?

gateway/proxy_fil.go
```diff
@@ -441,13 +441,23 @@ func (gw *Node) GetActorEvents(ctx context.Context, filter *types.ActorEventFilt
...
+		if err := gw.checkTipSetHeight(ctx, *filter.FromHeight, types.EmptyTSK); err != nil {
```
Stebalien, line 445: Do we also need to apply some default "max events" limit? Maybe the lookback is enough?

gateway/proxy_fil.go
rvagg: yeah, it ends up being the same as `filter.FromHeight == nil` because we just insert a zero value filter in place; do we need to account for the "last finalized" tipset in this limiting? can you have a lookback limit of 0?

gateway/proxy_fil.go
rvagg: no, we have the `MaxFilterHeightRange` and `MaxFilterResults` in config that defaults to 2880 and 10000 respectively, which is checked wherever you supply a filter (both here and the eth forms)

(aside: the `MaxFilterResults` is pretty annoying because it treats the collected events as a FIFO with that as the limit, and the user doesn't get told that they reached the limit, it just silently bumps out previously found events to stay within the limit, we need to fix that I think. Notes about that included in https://github.com/filecoin-project/lotus/issues/11680)

gateway/proxy_fil.go
Stebalien: I think we can leave it as-is.

Stebalien (approved):

PR #11741, opened 2024-03-19
feat(events): add "Raw" suffix to {Get,Subscribe}ActorEvents

rvagg:
This is done with the intention to add new {Get,Subscribe}ActorEvents in a future release (i.e. soon!) with both decoded values (dag-json represented) and simplified (no flags or codec). But because this comes with some trade-offs wrt fidelity of information (e.g. likely needing to drop events with badly encoded values, and not retaining original codec), we need to also have a Raw form of these APIs for consumers that want to take on the burden of consuming them as they are.

The changes being proposed in https://github.com/filecoin-project/lotus/pull/11707 roughly outline the functionality of the new methods, but further work needs to be done on the return types and argument types. Rather than attempt to rush that for a v1.26.0 release, **I'm proposing that we make this rename change for inclusion in v1.26.0 in order to avoid a future rename to make way for #11707 or to come up with new names for that functionality.**

## Related Issues

https://github.com/filecoin-project/lotus/pull/11707

## Proposed Changes

1. Rename `GetActorEvents` to `GetActorEventsRaw`
2. Rename `SubscribeActorEvents` to `SubscribeActorEventsRaw`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,20 +132,22 @@ Additionally, Filecoin is not Ethereum no matter how much we try to provide API/
 
 [handlefilecoinmethod]: https://fips.filecoin.io/FIPS/fip-0054.html#handlefilecoinmethod-general-handler-for-method-numbers--1024
 
-### GetActorEvents and SubscribeActorEvents
+### GetActorEventsRaw and SubscribeActorEventsRaw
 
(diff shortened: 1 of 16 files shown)


Stebalien (approved):

aarshkshah1992 (approved):
lgtm

PR #11764, opened 2024-03-21
fix: CLI: adjust TermMax for extend-claim used by a different client

LexLuthr:
## Proposed Changes

This PR fixes the TermMax for claim extensions requested by a client different than the original client.

## Checklist

- [x] CI is green


LexLuthr:
Test with Datacap:

Before
```
ID  Provider  Client  Data                                                              Size     TermMin  TermMax  TermStart  Sector  
1   1000      102     baga6ea4seaqlsmme4a6x2pfvlstsoh2eyqwfxvfhudcmnf5ndiw56leqa3fiaoa  8388608  3675537  5256000  0          0       
2   1000      1009    baga6ea4seaqpfonfro6j3gcw4uxkxbivlqn2fghx5dpulc6sbi5noz7bcvzmuiq  2048     518400   5256000  13016      1
```

After
```
ID  Provider  Client  Data                                                              Size     TermMin  TermMax  TermStart  Sector  
1   1000      102     baga6ea4seaqlsmme4a6x2pfvlstsoh2eyqwfxvfhudcmnf5ndiw56leqa3fiaoa  8388608  3675537  5288893  0          0       
2   1000      1009    baga6ea4seaqpfonfro6j3gcw4uxkxbivlqn2fghx5dpulc6sbi5noz7bcvzmuiq  2048     518400   5275877  13016      1 
```

New TerMax is more than 5 years

LexLuthr:
CC: @beck-8 Please review the code and provide a +1 if everything looks good.

beck-8:
batchSize to a default value. And it is better if the user can fill in the value by himself/herself

LexLuthr:
@beck-8 I have pushed the changes. Please test it and let me know if any issues.

LexLuthr:
@rvagg Can you please review? I would like to merge before diff starts to grow.

beck-8:
> @rvagg Can you please review? I would like to merge before diff starts to grow.

Have you seen my comment? I don’t know if my question is correct.

LexLuthr:
@beck-8 Which comment? I don't see any comments on this PR.

beck-8:
> @beck-8 Which comment? I don't see any comments on this PR.

image

LexLuthr:
@beck-8 You never posted that comment. I think you forgot to publish the review. 
Good catch! Working on the refactor to fix this.

cli/filplus.go
```diff
@@ -1372,7 +1351,41 @@ func CreateExtendClaimMsg(ctx context.Context, api api.FullNode, pcm map[verifre
...
+			ncparams, err := actors.SerializeParams(&verifregtypes13.AllocationRequests{
+				Extensions: batch,
+			})
+			if err != nil {
+				return nil, xerrors.Errorf("failed to searialise the parameters: %s", err)
+			}
+
+			transferParams, err := actors.SerializeParams(&datacap2.TransferParams{
+				To:           builtin.VerifiedRegistryActorAddr,
+				Amount:       big.Mul(rDataCap, builtin.TokenPrecision),
```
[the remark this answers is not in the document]
LexLuthr: Pushed a commit. Please test again and let me know. I have also reduced the batch size based on the batchSize used for commit messages.
@rvagg Can you also provide us an efficient batch size for this message? I am not familiar with gas calculations required for batching.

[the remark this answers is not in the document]
beck-8: I think there's no problem (I didn't look carefully)

cli/filplus.go
```diff
@@ -934,7 +934,11 @@ var filplusSignRemoveDataCapProposal = &cli.Command{
 
 var filplusExtendClaimCmd = &cli.Command{
 	Name:  "extend-claim",
-	Usage: "extend claim expiration (TermMax)",
+	Usage: "extends claim expiration (TermMax)",
+	UsageText: `Extends claim expiration (TermMax).
+If the client is original client then claim can be extended to Maximum 5 years and no Datacap is required.
+If the client id different then claim can be extended up to Maximum 5 years from now and Datacap is required.
```
rvagg, line 940:
```suggestion
If the client is original client then claim can be extended to maximum 5 years and no DataCap is required.
If the client id different then claim can be extended up to maximum 5 years from now and DataCap is required.
```

cli/filplus.go
```diff
@@ -1162,17 +1171,23 @@ func CreateExtendClaimMsg(ctx context.Context, api api.FullNode, pcm map[verifre
...
+							TermMax:  head.Height() + tmax - claim.TermStart,
+						}
+						newClaims[req] = big.NewInt(int64(claim.Size))
 						rDataCap.Add(big.NewInt(int64(claim.Size)).Int, rDataCap.Int)
-						continue
 					}
+					// If new duration shorter than the original duration then do nothing
+					continue
+				}
+				// For original client, compare duration(TermMax) and claim should be already be expired
```
rvagg, line 1189:
```suggestion
				// For original client, compare duration(TermMax) and claim should not already be expired
```

cli/filplus.go
```diff
@@ -1204,17 +1219,23 @@ func CreateExtendClaimMsg(ctx context.Context, api api.FullNode, pcm map[verifre
...
+						TermMax:  head.Height() + tmax - claim.TermStart,
+					}
+					newClaims[req] = big.NewInt(int64(claim.Size))
 					rDataCap.Add(big.NewInt(int64(claim.Size)).Int, rDataCap.Int)
-					continue
 				}
+				// If new duration shorter than the original duration then do nothing
+				continue
+			}
+			// For original client, compare duration(TermMax) and claim should be already be expired
```
rvagg, line 1237:
```suggestion
			// For original client, compare duration(TermMax) and claim should not already be expired
```

rvagg:
This is fine I think. But there seems to be a third option that's not covered here, you have "my claim, extend up to 5 years from start", and "not my claim, extend up to 5 years from now, pay datacap", but not "my claim, extend up to 5 years from now, pay datacap", which I believe is possible. There's no client check when doing a `ClaimExtensionRequest` and the ability to extend by 5 years from now by paying datacap seems to be something that a client might want to be doing.

Maybe that doesn't have to be in this PR but I can see it being a concern so perhaps think about how that might work in terms of UX and make sure that it'd be additive to the command rather than having to break it.

You have a conflict now with itest/direct_data_onboard_verified_test.go, I did so updates in https://github.com/filecoin-project/lotus/pull/11766, they got rewritten to work on `release/v1.26.0` and then pulled back in to master. It should be straightforward I think.

beck-8:
i think maybe "my claim, extend up to 5 years from start" = "my claim, extend up to 5 years from now, pay datacap" ？

Normal people will not use parameters less than 5 years.

rvagg:
start != now though, https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0045.md#renewal-of-claim-term-by-spending-datacap, this is the perpetual-storage case where you can keep spending datacap to go beyond max term. I don't think this is urgent, by any means, maybe we have a few years before this will be something that makes sense doing given sector lifetimes, but it seems like a use-case that will at least eventually be desired

LexLuthr:
@rvagg Thank you for the review. I will make all the suggested changes. The third case makes sense in terms of Verifreg actor but Sector lifetime is currently limited to 5 years. I think we should defer it to future when sector lifetime has been extended or resealing is in place.

rvagg (approved):

itests/direct_data_onboard_verified_test.go
```diff
@@ -873,14 +872,14 @@ func TestVerifiedDDOExtendClaim(t *testing.T) {
...
+	msgs, err = cli.CreateExtendClaimMsg(ctx, client.FullNode, pcm, []string{}, verifiedClientAddr2, builtin.EpochsInYear*4, false, true, 100)
 	require.NoError(t, err)
 	require.Nil(t, msgs)
 
 	newclaim, err = client.StateGetClaim(ctx, miner.ActorAddr, verifreg.ClaimId(allocationId), types.EmptyTSK)
 	require.NoError(t, err)
 	require.NotNil(t, newclaim)
-	require.EqualValues(t, newclaim.TermMax, verifregtypes13.MaximumVerifiedAllocationTerm)
 
-	// TODO: check "claim-updated" message
```
rvagg, line 885: actually, would you mind leaving this TODO here, it's a note to me, and it should say "event" rather than "message"

LexLuthr: Sure

rvagg (approved):

Issue #11780, opened 2024-03-25
Makefile target to run itests


rvagg:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Be able to execute the same paths CI runs but locally if you have the resources to do so.

### Describe the solution you'd like

We now have `make unittests` to run everything but itests, we should have a `make itests` to do the itests. I'm not sure if a `make test` is a good idea though, we don't really want to give the impression this is a good idea for any contributor to run (maybe it could print documentation though).

Currently itests are each run separately, and they are configured independently in the CI config. It would not be ideal to have yet another place to have to remember to add your new itest to make it run, so some deduplication is needed. Either the Makefile target reads CI config (new GitHub Actions one) to figure out what and how to run, or we make CI config and Makefile depend on some new configuration, perhaps a text file in the itests directory, or perhaps we decorate itests in source and extract metadata from there.

BigLep:
As an example of the impact of this problem, ecosystem partners are forced to to open a fake PR to get itests to invoke.  https://github.com/filecoin-project/lotus/pull/11987 is an example.  https://filecoinproject.slack.com/archives/CP50PPW2X/p1714698675147429?thread_ts=1711282222.982289&cid=CP50PPW2X is a #fil-lotus-dev thread on this.

PR #11782, opened 2024-03-26
ci: configure itests in build constraints, detect in Actions script

rvagg:
This is a stacked PR against #11762 and I'm exploring ways to configure itests inline, mainly because I'd like to not restrict this to the CI config, but also make it available in other ways, such as in the `Makefile`, as discussed in #11780. I can also imagine this being used to build some kind of fancy Docker or k8s config to do something similar, particularly as we get YugabyteDB involved.

A `make itests` would ideally not copy config that exists in the Actions config already (adding a new itest shouldn't involve wiring it up in multiple places—ideally no places but the itest itself!). It's also not really practical to be parsing yaml from `make`. So my thought here is to just use Go build constraints since they are an existing annotation pattern already. You could even use them to run them as a group with `go test`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,6 +53,7 @@ jobs:
                 "skip_conformance": "0"
               }
             ]
+        run: |
           # Mapping from test group names to custom runner labels
           # The jobs default to running on the default hosted runners (4 CPU, 16 RAM).
           # We use self-hosted xlarge (4 CPU, 8 RAM; and large - 2 CPU, 4 RAM) runners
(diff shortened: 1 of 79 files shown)


rvagg:
good feedback @galargh, the runner size is quite leaky and it's good to hear that it's not specifically about the tests but the resources, I can adjust to that.

I'll hold off on taking this further, it's not urgent, I'd just like to get toward a better local way of running some of these that doesn't involve duplicating information we have in CI config

Stebalien:
I think the first step here is to start trying to reduce the number of tests that actually need custom runners. If we can get it down to a reasonable number, putting those exceptions in the CI config isn't terrible.

Issue #11734, opened 2024-03-18
Migrate CI from CircleCI to GitHub Actions


galargh:
## User Story

GitHub Actions provides Developer Experience better suited for Lotus development team going forward. 

## Acceptance Criteria

- [x] The p50 and p95 success rate of the new GitHub Actions workflows is the same or higher than their CircleCI counterparts
- [x] The p50 and p95 execution time of the new GitHub Actions workflows is the same or smaller than their CircleCI counterparts (*p90 is worse for the **Test** workflow but it is due to an outlier caused by an issue with self-hosted runners that is being addressed)

```[tasklist]
### Deliverables (GitHub Actions workflows)
- [x] docs-check
- [x] gen-check
- [x] gofmt
- [x] lint-all
- [x] mod-tidy-check
- [x] test
- [x] test-conformance
- [x] test-itest-*
- [x] test-unit-*
- [x] docker-build
- [x] docker-push
- [x] release
```

## Technical Breakdown
```[tasklist]
- [x] Create GitHub Actions jobs performing checks and generation (docs-check, gen-check, gofmt, lint-all, mod-tidy-check)
- [x] Provision self-hosted runners (large) for use in lotus (to accommodate expected high parallelism demands)
- [x] Create GitHub Actions jobs executing tests (test, test-conformance)
- [x] Create GitHub Actions jobs executing tests that require high parallelism (test-itest-*, test-unit-*)
- [x] Create GitHub Actions jobs executing docker builds (docker build, docker push)
- [x] Create GitHub Actions jobs performing release (release (dry), release)
- [x] Gather and compare CI stats for CircleCI job runs and their GitHub Actions counterparts*
- [ ] Decommission CircleCI jobs after 1-2 weeks of stable execution of GitHub Actions counterparts*
- [ ] Update lotus-priv clone
```

```[tasklist]
### Testing
- [x] Run the new GitHub Actions workflows in shadow mode for 1-2 weeks
- [x] Gather success rate and execution time metrics for both the CircleCI and GitHub Actions workflows
- [x] Compare and prepare a report
```

Stebalien:
Any idea what caused the apt locking issue in https://github.com/filecoin-project/lotus/actions/runs/8529890853/job/23366624530?pr=11808?

galargh:
> Any idea what caused the apt locking issue in https://github.com/filecoin-project/lotus/actions/runs/8529890853/job/23366624530?pr=11808?

[...]

PR #11785, opened 2024-03-26
ci: ci: migrate the release workflow to github actions

galargh:
## Related Issues

https://github.com/filecoin-project/lotus/issues/11734

## Proposed Changes

This PR creates two new GitHub Actions workflows (`docker.yml` and `release.yml`) that perform the following jobs previously handled exclusively by CircleCI:
- docker-*
- release-*

## Additional Info

You can find the example of a Docker publish at: https://github.com/filecoin-project/lotus/actions/runs/8438620518
You can find the example of a release publish at: https://github.com/filecoin-project/lotus/actions/runs/8438620515

Note that both Docker as well as release will operate exclusively in a dry run mode until CircleCI is deprecated. This is to prevent duplicate publishing.

The newly added workflows need the following vars/secrets set up before they can be brought out of the dry run mode:
- vars.DOCKERHUB_USERNAME
- secrets. DOCKERHUB_TOKEN
- secrets.GORELEASER_KEY

## Checklist

- [ ] CI is green


.github/workflows/docker.yml
```diff
@@ -0,0 +1,101 @@
+name: Docker
+
+on:
+  push:
+    branches:
+      - ci/*
```
Stebalien, line 6: testing?

.github/workflows/release.yml
```diff
@@ -0,0 +1,112 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - ci/*
```
Stebalien, line 6: remove?

.github/workflows/release.yml
```diff
@@ -0,0 +1,112 @@
...
+      - release/*
```
Stebalien, line 7: Shouldn't we only look at tags? Really, don't we trigger on release?

.github/workflows/docker.yml
galargh: You're right! The docker builds were not triggered on `ci/*` in CircleCI. They were only triggered on:
- push to master (channel=master)
- push to release/* (dry run)
- tag push (channel=candidate or channel=stable)
- schedule (channel=nightly)

Let's replace the ci/* push trigger with master push trigger that we skipped.

```suggestion
      - master
```

.github/workflows/release.yml
galargh: The release, unlike docker publish, was triggered on pushes to ci/* branches in CircleCI (see https://github.com/filecoin-project/lotus/blob/78d9d9baa054448f2dfb192101eb9a9fbdd0ac9b/.circleci/config.yml#L1112).

All in all, releases were triggered on:
- push to ci/* (dry run)
- push to release/* (dry run)
- tag push

I think whether we perform the dry runs or not comes down to the team preference.

```suggestion
```

.github/workflows/release.yml
galargh: Following the thought from https://github.com/filecoin-project/lotus/pull/11785#discussion_r1541430018, the tag push would trigger a proper release while a push to either ci/* or release/* branch would trigger a dry run release to see if everything's OK.

.github/workflows/release.yml
Stebalien: Ah, I see. @arajasek Do we actually use this?

.github/workflows/release.yml
Stebalien: Can we trigger the full run on github releases only (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) to prevent accidental tag pushes from turning into releases.

.github/workflows/release.yml
```diff
@@ -0,0 +1,112 @@
...
+          name: kubo
+          version: v0.16.0
+      - uses: goreleaser/goreleaser-action@7ec5c2b0c6cdda6e8bbb49444bc797dd33d74dd8 # v5.0.0
+        with:
+          distribution: goreleaser-pro
+          version: latest
+          args: release --clean --debug ${{ env.PUBLISH == 'false' && '--snapshot' || '' }}
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
```
Stebalien, line 107: Is there any way to avoid passing tokens when we're doing a dry-run? These tokens are pretty powerful and we don't have any additional restrictions on `ci/` branches.

.github/workflows/release.yml
Stebalien: Okay, actually this is kind of nice to have. But I'd like to find some way to avoid passing tokens because they're really powerful. I guess I can also just restrict pushes these branches, but in that case, I'd probably rename it something like test-release or something like that.

.github/workflows/release.yml
galargh: Of course, no problem - we can use empty values when we're not publishing like so:

```suggestion
          GITHUB_TOKEN: ${{ env.PUBLISH != 'false' && github.token || '' }}
          GORELEASER_KEY: ${{ env.PUBLISH != 'false' && secrets.GORELEASER_KEY || '' }}
```

.github/workflows/release.yml
galargh: If you don't mind, I'd prefer to keep the triggers consistent with CircleCI during the trial period. This will make it easier for us to verify whether we're ready to deprecate CircleCI. Once we do that, I totally agree that it'd be a good idea to explore how we could improve and secure the release process further. Would that be alright?

And yes, we could definitely explore using the release trigger instead of the tag one in the future. It might need a slight rework of the workflow and/or underlying scripts since the context would be slightly different, but it's not an issue.

.github/workflows/release.yml
galargh: I applied the suggestion from https://github.com/filecoin-project/lotus/pull/11785/files#r1542909114. That way, we'll only pass the tokens to goreleaser action if it's a run where publish=true.

.github/workflows/release.yml
Stebalien: Sg.

.github/workflows/release.yml
```diff
@@ -0,0 +1,112 @@
...
+  run:
+    shell: bash
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  # This enables the workflow to create GitHub releases
+  contents: write
```
Stebalien, line 22: can we grant write access in the release job only?

```yaml
jobs:
  release:
    permissions:
      contents: write
```

Stebalien (approved):
One more paranoid suggestion, otherwise LGTM.

.github/workflows/release.yml
galargh: There should be a way to do that. I'll check and update accordingly today.

galargh: Of course! I'm on it.

galargh: Done - https://github.com/filecoin-project/lotus/pull/11785/commits/b3c24acf4bdfbf2e9331916f06adc7be3d5b76ea ✅

PR #11796, opened 2024-03-28
Curio Network Optimizaton

snadrus:
## Related Issues
Network-level optimization has not been a priority. Here is a straightforward way we could implement it. 

## Proposed Changes
CanAccept's end (once hard requirements are met) can include a "soft requirements" section managed by taskhelp.NetworkOptimizer which skips rounds if soft requirements are too few. 

## Checklist

- [ ] CI is green

### DIFF
--- a/lib/harmony/taskhelp/common.go
+++ b/lib/harmony/taskhelp/common.go
@@ -1,5 +1,19 @@
 package taskhelp
 
+import (
+	"context"
+	"time"
+
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/samber/lo"
+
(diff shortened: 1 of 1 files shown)


magik6k (commented):
I quite like this approach.

Would be nice to ship this initially with some machine-resource-based funcs - e.g. "machine load over 50%" and "machine load over 80%" - then give SPs a config toggle to say "prefer spreading load - less utilized machines accept tasks first" and "prefer consolidating load - more utilized machines accept tasks first" (in the latter more of the cluster can idle/sleep when not needed)

I can see this applied to almost all heavier tasks that aren't latency-sensitive (like winningPoSt which needs to happen in <30s)

Issue #11703, opened 2024-03-11
DisputWindowPost executes successfully but without a reward!


arvin-lau:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### Lotus Version

```text
Daemon:  1.25.2+mainnet+git.60a61b483.dirty+api1.5.0
Local: lotus version 1.25.2+mainnet+git.60a61b483.dirty
```

### Repro Steps

https://filfox.info/en/message/bafy2bzacedeupksdrvt7ujn62xrk6peinaa27czhakkr3osyut2jzqym4pihq?t=1

### Describe the Bug

Sending DisputWindowPost succeeded but no reward, the last DisputWindowPust succeeded with a reward, I'm wondering if the reward was canceled or if there was a problem with the lotus

### Logging Information

```text
https://filfox.info/en/message/bafy2bzacedeupksdrvt7ujn62xrk6peinaa27czhakkr3osyut2jzqym4pihq?t=1
```

Stebalien:
The reward comes from the Miner Actor's balance, and this Miner had no balance. It looks like they:

1. Submitted a bad window post.
2. Terminated the sectors in that window post (burning some FIL, but also unlocking their pledge).
3. Withdrew their pledge.

This is definitely an... interesting edge case. But there isn't really a way to "pay out" in this case because there simply isn't any FIL.

arvin-lau:
> The reward comes from the Miner Actor's balance, and this Miner had no balance. It looks like they:
> 
> 1. Submitted a bad window post.
[...]

PR #11800, opened 2024-03-31
refactor: cli: Avoid disputing WindowedPoSt messages when the target miner has no balance to pay rewards

### LINKED ISSUE #11715: Avoid disputing WindowedPoSt messages when the target Miner has no balance to pay rewards
### Checklist

### Lotus component

- [ ] lotus JSON-RPC API
- [X] Other

In https://github.com/filecoin-project/lotus/issues/11703, a user came across an edge case where an SP:

1. Submitted a bad PoSt.
2. Terminated their sectors.
3. Withdrew all funds (after paying all fees).

This left the SP's miner actor with a bad proof but without any funds to pay for disputes. On the other hand, given that the sectors have been terminated and termination fees were already paid we don't really care about fining the SP for the bad proofs anyways.

But... we still don't want to submit dispute messages in this case because doing so isn't free and doesn't really serve a purpose.

Before sending a dispute to a miner actor, validate that the miner actor has a non-zero balance. If it has a zero balance, skip it.

We'll likely want to skip/check at the top of this loop by calling `api.WalletBalance` on the target miner actor's address:

https://github.com/filecoin-project/lotus/blob/81e65dba5676a93d077d0d420e56eeb9f9a0d27f/cli/disputer.go#L240

It would be better to determine if sending the dispute is likely to be profitable, but that's significantly more challenging (we'd need to look at gas costs, calculate expected fines/rewards, etc.). Simply checking if the balance is 0 should be sufficient in most cases.

### Additional context

_No response_

Nagaprasadvr:
Avoid disputing WindowedPoSt messages when the target Miner has no balance to pay rewards

## Related Issues
fixes #11715

## Proposed Changes
In https://github.com/filecoin-project/lotus/issues/11703, a user came across an edge case where an SP:

1. Submitted a bad PoSt.
2. Terminated their sectors.
3. Withdrew all funds (after paying all fees).

This left the SP's miner actor with a bad proof but without any funds to pay for disputes. On the other hand, given that the sectors have been terminated and termination fees were already paid we don't really care about fining the SP for the bad proofs anyways.

But... we still don't want to submit dispute messages in this case because doing so isn't free and doesn't really serve a purpose.

So add a check to fetch miner wallet balance and only send disputing messages if balance is non zero.

## Additional Info
look at disputer.go line no 241 onwards

### DIFF
--- a/cli/disputer.go
+++ b/cli/disputer.go
@@ -238,15 +238,6 @@ var disputerStartCmd = &cli.Command{
 
 			// TODO: Parallelizeable
 			for _, dl := range dls {
-				// CHECK: if miner waller balance is zero then skip sending dispute message
-				walletBalance, err := api.WalletBalance(ctx, dl.miner)
-				if err != nil {
-					return xerrors.Errorf("failed to get wallet balance while checking to send dispute messages to miner %w: %w", dl.miner, err)
-				}
-				if walletBalance.Equals(big.Zero()) {
-					disputeLog.Warnw("wallet balance is zero, skipping dispute message", "wallet", dl.miner)
-					return nil
-				}
 				fullDeadlines, err := api.StateMinerDeadlines(ctx, dl.miner, tsk)
 				if err != nil {
 					return xerrors.Errorf("failed to load deadlines: %w", err)
@@ -260,6 +251,9 @@ var disputerStartCmd = &cli.Command{
 				proofsChecked += disputableProofs
 
 				ms, err := makeDisputeWindowedPosts(ctx, api, dl, disputableProofs, fromAddr)
+				if ms == nil {
+					continue
+				}
 				if err != nil {
 					return xerrors.Errorf("failed to check for disputes: %w", err)
 				}
@@ -370,6 +364,15 @@ var disputerStartCmd = &cli.Command{
 // for a given miner, index, and maxPostIndex, tries to dispute posts from 0...postsSnapshotted-1
 // returns a list of DisputeWindowedPoSt msgs that are expected to succeed if sent
 func makeDisputeWindowedPosts(ctx context.Context, api v0api.FullNode, dl minerDeadline, postsSnapshotted uint64, sender address.Address) ([]*types.Message, error) {
+	// CHECK: if miner waller balance is zero then skip sending dispute message
+	walletBalance, err := api.WalletBalance(ctx, dl.miner)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get wallet balance while checking to send dispute messages to miner %w: %w", dl.miner, err)
+	}
+	if walletBalance.Equals(big.Zero()) {
+		disputeLog.Warnw("wallet balance is zero, skipping dispute message", "wallet", dl.miner)
+		return nil, nil
+	}
 	disputes := make([]*types.Message, 0)
 
 	for i := uint64(0); i < postsSnapshotted; i++ {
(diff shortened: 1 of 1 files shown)


Stebalien (approved):

cli/disputer.go
```diff
@@ -238,6 +238,15 @@ var disputerStartCmd = &cli.Command{
...
 			// TODO: Parallelizeable
 			for _, dl := range dls {
+				// CHECK: if miner waller balance is zero then skip sending dispute message
+				walletBalance, err := api.WalletBalance(ctx, dl.miner)
+				if err != nil {
+					return xerrors.Errorf("failed to get wallet balance while checking to send dispute messages to miner %w: %w", dl.miner, err)
+				}
+				if walletBalance.Equals(big.Zero()) {
+					disputeLog.Warnw("wallet balance is zero, skipping dispute message", "wallet", dl.miner)
+					return nil
```
Stebalien, line 248: This will skip processing the rest of the tipset. Instead, I think we need an early return of `nil, nil` inside "`makeDisputeWindowedPosts` if the miner has no balance.

Stebalien (changes requested):

cli/disputer.go
Nagaprasadvr: will add this change soon

cli/disputer.go
```diff
@@ -251,6 +251,9 @@ var disputerStartCmd = &cli.Command{
 				proofsChecked += disputableProofs
 
 				ms, err := makeDisputeWindowedPosts(ctx, api, dl, disputableProofs, fromAddr)
+				if ms == nil {
+					continue
+				}
```
Stebalien, line 256: Let this continue to the end of the loop. If this is empty, we won't add any messages to dispute, but we still need to add the deadline to the deadline map (caching, IIRC).

cli/disputer.go
```diff
@@ -361,6 +364,15 @@ var disputerStartCmd = &cli.Command{
 // for a given miner, index, and maxPostIndex, tries to dispute posts from 0...postsSnapshotted-1
 // returns a list of DisputeWindowedPoSt msgs that are expected to succeed if sent
 func makeDisputeWindowedPosts(ctx context.Context, api v0api.FullNode, dl minerDeadline, postsSnapshotted uint64, sender address.Address) ([]*types.Message, error) {
+	// CHECK: if miner waller balance is zero then skip sending dispute message
+	walletBalance, err := api.WalletBalance(ctx, dl.miner)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get wallet balance while checking to send dispute messages to miner %w: %w", dl.miner, err)
+	}
+	if walletBalance.Equals(big.Zero()) {
```
Stebalien, line 372:
```suggestion
	if walletBalance.IsZero() {
```

(I think)

Stebalien (changes requested):
Thanks!

cli/disputer.go
Nagaprasadvr: done

cli/disputer.go
Nagaprasadvr: done

cli/disputer.go
```diff
@@ -255,7 +255,9 @@ var disputerStartCmd = &cli.Command{
 					return xerrors.Errorf("failed to check for disputes: %w", err)
 				}
 
-				dpmsgs = append(dpmsgs, ms...)
+				if ms != nil {
```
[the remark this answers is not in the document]
Nagaprasadvr: removed check

Stebalien (approved):

PR #11808, opened 2024-04-02
test: eth: deflake EthBlockHashesCorrect_MultiBlockTipset

Stebalien:
Increase the chances of having multiple blocks at the same height by aligning block times across test miners. Also, make the block hash test faster.

### DIFF
--- a/itests/eth_block_hash_test.go
+++ b/itests/eth_block_hash_test.go
@@ -26,30 +26,36 @@ import (
 func TestEthBlockHashesCorrect_MultiBlockTipset(t *testing.T) {
 	// miner is connected to the first node, and we want to observe the chain
 	// from the second node.
-	blocktime := 250 * time.Millisecond
+	blocktime := 100 * time.Millisecond
 	n1, m1, m2, ens := kit.EnsembleOneTwo(t,
 		kit.MockProofs(),
 		kit.ThroughRPC(),
 	)
 	ens.InterconnectAll().BeginMining(blocktime)
 
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
-	n1.WaitTillChain(ctx, kit.HeightAtLeast(abi.ChainEpoch(5)))
-	defer cancel()
+	{
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		n1.WaitTillChain(ctx, kit.HeightAtLeast(abi.ChainEpoch(5)))
+		cancel()
+	}
 
 	var n2 kit.TestFullNode
 	ens.FullNode(&n2, kit.ThroughRPC()).Start().Connect(n2, n1)
 
-	// find the first tipset where all miners mined a block.
-	ctx, cancel = context.WithTimeout(context.Background(), 5*time.Minute)
-	n2.WaitTillChain(ctx, kit.BlocksMinedByAll(m1.ActorAddr, m2.ActorAddr))
-	defer cancel()
+	{
+		// find the first tipset where all miners mined a block.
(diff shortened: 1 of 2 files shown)


Stebalien:
IMO, the miner itself needs better synchronization logic. But this fixes the immediate issues with the test.

itests/eth_block_hash_test.go
```diff
@@ -26,30 +26,36 @@ import (
 func TestEthBlockHashesCorrect_MultiBlockTipset(t *testing.T) {
 	// miner is connected to the first node, and we want to observe the chain
 	// from the second node.
-	blocktime := 250 * time.Millisecond
+	blocktime := 100 * time.Millisecond
```
Stebalien, line 29: This is reverting a change I made a while back while attempting to deflake this before.

itests/eth_block_hash_test.go
```diff
@@ -26,30 +26,36 @@ import (
...
+		// find the first tipset where all miners mined a block.
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+		n2.WaitTillChain(ctx, kit.BlocksMinedByAll(m1.ActorAddr, m2.ActorAddr))
+		cancel()
+	}
 
 	head, err := n2.ChainHead(context.Background())
 	require.NoError(t, err)
 
+	ctx := context.Background()
```
Stebalien, line 55: I'm just cleaning up contexts here.

itests/eth_block_hash_test.go
```diff
@@ -26,30 +26,36 @@ import (
...
+
 	// let the chain run a little bit longer to minimise the chance of reorgs
-	n2.WaitTillChain(ctx, kit.HeightAtLeast(head.Height()+50))
+	n2.WaitTillChain(ctx, kit.HeightAtLeast(head.Height()+10))
```
Stebalien, line 58: This delay was unnecessarily long in an attempt to make this test less flaky, but this wasn't the real problem.

itests/eth_block_hash_test.go
```diff
@@ -26,30 +26,36 @@ import (
 func TestEthBlockHashesCorrect_MultiBlockTipset(t *testing.T) {
```
Stebalien, line 26: All my changes to this test are just making it faster, not fixing any issues.

itests/kit/blockminer.go
```diff
@@ -256,8 +256,6 @@ func (bm *BlockMiner) MineBlocksMustPost(ctx context.Context, blocktime time.Dur
 }
 
 func (bm *BlockMiner) MineBlocks(ctx context.Context, blocktime time.Duration) {
-	time.Sleep(time.Second)
```
Stebalien, line 259: This was staggering the time at which block miners started, reducing the chances of having blocks tipsets with multiple blocks.

It was also just unnecessarily slowing down tests.

itests/kit/blockminer.go
```diff
@@ -278,8 +276,11 @@ func (bm *BlockMiner) MineBlocks(ctx context.Context, blocktime time.Duration) {
 			default:
 			}
 
+			now := time.Duration(time.Now().UnixNano())
+			delay := blocktime - (now % blocktime)
+
 			select {
-			case <-time.After(blocktime):
+			case <-time.After(delay):
```
Stebalien, line 283: This is the real fix. I'm now aligning block times to a multiple of the block time so multiple miners will actually produce blocks at the same time.

rvagg:
@Stebalien can you describe the flake itself? I don't recall what this was failing with and can't find one of the failure records.

rvagg (approved):
Goodo, nicely diagnosed!

PR #11816, opened 2024-04-03
fix: eth rpc: ETH Subscription RPC API should return blocks that are one epoch behind the current epoch

### LINKED ISSUE #11807: newHeads subscription compatibility with deferred execution model
### Checklist

- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.

### Lotus component

- [X] lotus JSON-RPC API
- [ ] Other

### Lotus Version

```text
1.26.1+mainnet+git.9dc9a5cf4.dirty
```

### Repro Steps

1. Subscribe to newHeads via eth_subscribe
2. When you receive new block from the subscription, immediately send eth_getBlockByNumber(block_num, true)
3. See error `requested a future epoch (beyond 'latest')`

### Describe the Bug

So as per documentation: https://docs.filecoin.io/smart-contracts/filecoin-evm-runtime/difference-with-ethereum

Filecoin returns for latest block n - 1 for compatibility reason, should it also delay the subscription event?

### Logging Information

```text
-
```

aarshkshah1992:
## Related Issues
Closes #11807 

## Proposed Changes
ETH Subscription RPC API should return blocks that are one epoch behind the current epoch to match the contract of the `eth_getBlockByNumber` API

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/eth_filter_test.go
+++ b/itests/eth_filter_test.go
@@ -137,6 +137,40 @@ func TestEthNewPendingTransactionFilter(t *testing.T) {
 	}
 }
 
+func TestEthNewHeadsSub(t *testing.T) {
+	require := require.New(t)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	kit.QuietAllLogsExcept("events", "messagepool")
+
+	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC(), kit.WithEthRPC())
+	ens.InterconnectAll().BeginMining(10 * time.Millisecond)
+
+	// install filter
+	subId, err := client.EthSubscribe(ctx, res.Wrap[jsonrpc.RawParams](json.Marshal(ethtypes.EthSubscribeParams{EventType: "newHeads"})).Assert(require.NoError))
+	require.NoError(err)
+
+	err = client.EthSubRouter.AddSub(ctx, subId, func(ctx context.Context, resp *ethtypes.EthSubscriptionResponse) error {
+		rs := *resp
+		block, ok := rs.Result.(map[string]interface{})
+		require.True(ok)
+		blockNumber, ok := block["number"].(string)
+		require.True(ok)
+
+		blk, err := client.EthGetBlockByNumber(ctx, blockNumber, false)
+		require.NoError(err)
+		require.NotNil(blk)
+		// block hashes should match
+		require.Equal(block["hash"], blk.Hash.String())
+
+		return nil
+	})
+	require.NoError(err)
+	time.Sleep(2 * time.Second)
+}
+
 func TestEthNewPendingTransactionSub(t *testing.T) {
 	require := require.New(t)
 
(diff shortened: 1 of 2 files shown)


node/impl/full/eth_events.go
```diff
@@ -341,6 +343,35 @@ func (e *ethSubscription) start(ctx context.Context) {
 					e.send(ctx, r)
 				}
 			case *types.TipSet:
+				// Get current chain epoch
+				head, err := e.ChainAPI.ChainHead(ctx)
```
Stebalien, line 347: Unfortunately, anything that looks at the head here is going to be racy because there could have been a reorg in the meantime

I think we can just return the tipset's parent here. Although we probably need to remember the last tip set we returned, so we avoid returning it multiple times in a row (e.g., if the "pending" tipset changes).

aarshkshah1992: @Stebalien Please can you explain the race here in more detail ? Even if the head changes by the time we get here, the epoch should still remain the same right ? I am interested in the epoch of the head here and not the head itself.

Stebalien: I don't think that's guaranteed due to shenanigans with null tipsets and weights. Finally, all you need is a heavier chain (more blocks). If I recall correctly, height is a factor in weight, but there are a bunch of other factors as well.

Regardless, it's simpler to just return the parent here. It means we don't have to keep some kind of buffer or anything.

itests/eth_filter_test.go
```diff
@@ -137,6 +137,40 @@ func TestEthNewPendingTransactionFilter(t *testing.T) {
...
+	kit.QuietAllLogsExcept("events", "messagepool")
+
+	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC(), kit.WithEthRPC())
+	ens.InterconnectAll().BeginMining(10 * time.Millisecond)
+
+	// install filter
+	subId, err := client.EthSubscribe(ctx, res.Wrap[jsonrpc.RawParams](json.Marshal(ethtypes.EthSubscribeParams{EventType: "newHeads"})).Assert(require.NoError))
+	require.NoError(err)
+
+	err = client.EthSubRouter.AddSub(ctx, subId, func(ctx context.Context, resp *ethtypes.EthSubscriptionResponse) error {
```
rvagg, line 155: this could probably do with a signal that we can check it was called before the test ends, no? maybe just a chan we can close and then instead of sleeping for 2 seconds, `require.Eventually` for it to close for 2 seconds.

itests/eth_filter_test.go
```diff
@@ -137,6 +137,40 @@ func TestEthNewPendingTransactionFilter(t *testing.T) {
 	}
 }
 
+func TestEthNewHeadsSub(t *testing.T) {
```
rvagg, line 140: can you document inside the test here, maybe near the assertions in the listener, how it relates to the fix? it doesn't seem to be testing much other than the ability to look up block numbers as they come in. Is that because it'd be too hard (racy) to test the relationship between the block we've received and the current head so we can't really do a "are we behind by one" properly?

node/impl/full/eth_events.go
aarshkshah1992: Makes sense.

aarshkshah1992:
Closed in favour of https://github.com/filecoin-project/lotus/pull/11858.

PR #11825, opened 2024-04-04
fix: paths/local: Fix on-disk storage accounting in new reservations

magik6k:
## Proposed Changes
Previously `Reserve` would instruct `stat` to include any sector files already on disk to lower what it consider as reserved, which in turn would lower Available storage less - The issue is that the new reservation isn't reflected in p.reserved at the time when stat is called, which means that usually we'd hit the `if stat.Reserved < 0 {.. stat.Reserved = 0` condition, making the accounting inaccurate.

This PR makes stat just return how much data is on-disk for the new reservation, and shifts the accounting to the Reserve function.

## Checklist

- [ ] CI is green


snadrus:
A little more doc would be nice.

Issue #11823, opened 2024-04-04
Actor event entries do not retain order when passing through event index


rvagg:
The event index doesn't care about order of "entries" that come with events, as per schema: https://github.com/filecoin-project/lotus/blob/a1264c8c8402b868a69dc34311fd09cebd669f93/chain/events/filter/index.go#L51-L58

Unfortunately, order now matters. As per [FIP-0083](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0083.md#sector-activated) both `sector-activated` and `sector-updated` rely on `piece-cid` and `piece-size` pairings to come out in order such that you can join them up to get the "PieceInfo". When we rely on sqlite table ordering, we get a shuffling of the array of events which loses pairings.

The result is that, when using `GetActorEventsRaw` (and `SubscribeActorEventsRaw` for historical events) we get something like this (from calibnet msg `bafy2bzacec2e6fiovhcjrprjgi76rhoqhr6bt7ag7dsejng5sefuksyvmvpwu`):

```json
[
  {
    "Flags": 1,
    "Key": "piece-size",
    "Codec": 81,
    "Value": "GwAAAAgAAAAA"
  },
  {
    "Flags": 3,
    "Key": "$type",
    "Codec": 81,
    "Value": "bnNlY3Rvci11cGRhdGVk"
  },
  {
    "Flags": 3,
    "Key": "piece-cid",
    "Codec": 81,
    "Value": "2CpYKAABgeIDkiAg8pxbycLUKlq7v8IZyXva5vs6aSxy9/mz5t2ygXvZjAw="
  },
  {
    "Flags": 3,
    "Key": "sector",
    "Codec": 81,
    "Value": "GScT"
  },
  {
    "Flags": 3,
    "Key": "unsealed-cid",
    "Codec": 81,
    "Value": "2CpYKAABgeIDkiAg8pxbycLUKlq7v8IZyXva5vs6aSxy9/mz5t2ygXvZjAw="
  }
]
```

When loading the same from the from the AMT with `ChainGetEvents`, we see the original order:

```json
[
  {
    "Flags": 3,
    "Key": "$type",
    "Codec": 81,
    "Value": "bnNlY3Rvci11cGRhdGVk"
  },
  {
    "Flags": 3,
    "Key": "sector",
    "Codec": 81,
    "Value": "GScT"
  },
  {
    "Flags": 3,
    "Key": "unsealed-cid",
    "Codec": 81,
[...]

PR #11829, opened 2024-04-04
fix: actor events: read entries from event index in order of their insertion

aarshkshah1992:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/11823

## Proposed Changes
Event entries should be read from the event index in the order of their insertion

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -53,7 +53,8 @@ var ddls = []string{
 		flags BLOB NOT NULL,
 		key TEXT NOT NULL,
 		codec INTEGER,
-		value BLOB NOT NULL
+		value BLOB NOT NULL,
+		ordering INTEGER NOT NULL
 	)`,
 
 	// metadata containing version of schema
@@ -63,6 +64,7 @@ var ddls = []string{
 
 	`INSERT OR IGNORE INTO _meta (version) VALUES (1)`,
 	`INSERT OR IGNORE INTO _meta (version) VALUES (2)`,
+	`INSERT OR IGNORE INTO _meta (version) VALUES (3)`,
 }
 
 var (
@@ -70,11 +72,11 @@ var (
 )
 
 const (
-	schemaVersion = 2
+	schemaVersion = 3
 
 	eventExists          = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
 	insertEvent          = `INSERT OR IGNORE INTO event(height, tipset_key, tipset_key_cid, emitter_addr, event_index, message_cid, message_index, reverted) VALUES(?, ?, ?, ?, ?, ?, ?, ?)`
-	insertEntry          = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
+	insertEntry          = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value, ordering) VALUES(?, ?, ?, ?, ?, ?, ?)`
 	revertEventsInTipset = `UPDATE event SET reverted=true WHERE height=? AND tipset_key=?`
 	restoreEvent         = `UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
 )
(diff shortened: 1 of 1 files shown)


chain/events/filter/index.go
```diff
@@ -53,7 +53,8 @@ var ddls = []string{
 		flags BLOB NOT NULL,
 		key TEXT NOT NULL,
 		codec INTEGER,
-		value BLOB NOT NULL
+		value BLOB NOT NULL,
+		ordering INTEGER NOT NULL
```
Stebalien, line 57: nit: index?

chain/events/filter/index.go
```diff
@@ -557,7 +602,7 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 		s = s + " WHERE " + strings.Join(clauses, " AND ")
 	}
 
-	s += " ORDER BY event.height DESC"
+	s += "ORDER BY event.height DESC"
```
Stebalien, line 605: I think we need the space, right?

chain/events/filter/index.go
```diff
@@ -572,9 +617,27 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 		return xerrors.Errorf("exec prefill query: %w", err)
 	}
 
-	var ces []*CollectedEvent
+	// create struct for holding EventEntry with ordering
+	type eventEntryWithOrdering struct {
+		types.EventEntry
+		ordering uint64
```
Stebalien, line 623: I don't think we need to select this field, we just need to order by it when querying.

chain/events/filter/index.go
```diff
@@ -672,8 +739,37 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
...
 	// sort it into height order
-	sort.Slice(ces, func(i, j int) bool { return ces[i].Height < ces[j].Height })
-	f.setCollectedEvents(ces)
+	sort.Slice(ces, func(i, j int) bool { return ces[i].height < ces[j].height })
+
+	var collectedEvents []*CollectedEvent
+	// sort the entries in the collected event by ordering
+	for _, ce := range ces {
+		ce := ce
+		sort.Slice(ce.entries, func(i, j int) bool { return ce.entries[i].ordering < ce.entries[j].ordering })
```
Stebalien, line 748: Let's do this in the query.

chain/events/filter/index.go
```diff
@@ -672,8 +739,37 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 
 	// collected event list is in inverted order since we selected only the most recent events
 	// sort it into height order
-	sort.Slice(ces, func(i, j int) bool { return ces[i].Height < ces[j].Height })
-	f.setCollectedEvents(ces)
+	sort.Slice(ces, func(i, j int) bool { return ces[i].height < ces[j].height })
```
Stebalien, line 742: I'm not sure why we bother with this at all...

chain/events/filter/index.go
```diff
@@ -118,6 +120,35 @@ func (ei *EventIndex) initStatements() (err error) {
...
+	}
+	// rollback the transaction (a no-op if the transaction was already committed)
+	defer tx.Rollback() //nolint:errcheck
+
+	if _, err = tx.ExecContext(ctx, "ALTER TABLE event_entry ADD COLUMN ordering INTEGER NOT NULL DEFAULT 0"); err != nil {
+		return xerrors.Errorf("increment _meta version: %w", err)
+	}
+
+	// Add the index on event_id and ordering fields
+	if _, err = tx.ExecContext(ctx, "CREATE INDEX IF NOT EXISTS idx_event_id_ordering ON event_entry (event_id, ordering)"); err != nil {
```
Stebalien, line 136: We don't need to index on ordering, right?

chain/events/filter/index.go
```diff
@@ -118,6 +120,35 @@ func (ei *EventIndex) initStatements() (err error) {
...
 
+func (ei *EventIndex) migrateToVersion3(ctx context.Context) error {
+	tx, err := ei.db.Begin()
+	if err != nil {
+		return xerrors.Errorf("begin transaction: %w", err)
+	}
+	// rollback the transaction (a no-op if the transaction was already committed)
+	defer tx.Rollback() //nolint:errcheck
+
+	if _, err = tx.ExecContext(ctx, "ALTER TABLE event_entry ADD COLUMN ordering INTEGER NOT NULL DEFAULT 0"); err != nil {
```
Stebalien, line 131: I'd prefer to actually fill these (may be possible in a single statement?).

chain/events/filter/index.go
aarshkshah1992: @Stebalien For some reason, if I add this to the ordering, we get way more records than we want and the test starts failing. I spent time debugging it before giving up. 

I had changed the current ordering to  	s += "ORDER BY event.height DESC, event_entry.ordering ASC".

chain/events/filter/index.go
aarshkshah1992: I agree. It was already here so didn't want to mess with it.

chain/events/filter/index.go
aarshkshah1992: Done.

chain/events/filter/index.go
aarshkshah1992: Done.

chain/events/filter/index.go
aarshkshah1992: Done.

aarshkshah1992:
This should be closed once https://github.com/filecoin-project/lotus/pull/11834 is merged.

PR #11834, opened 2024-04-05
fix(events): order entries by insertion order when selecting

rvagg:
Fixes: https://github.com/filecoin-project/lotus/issues/11823

Replaces: https://github.com/filecoin-project/lotus/pull/11829

I went down the `rowid` rabbit hole that @Stebalien mentioned in #11823, in https://www.sqlite.org/lang_createtable.html#rowid:

> The rowid value can be accessed using one of the special case-independent names "rowid", "oid", or "_rowid_" in place of a column name.

So, it's there already, let's use it! This requires no migration, and just sorts on the way out. The example message I had in #11823 now looks like this from `GetActorEventsRaw`:

```json
  "entries": [
    {
      "Flags": 3,
      "Key": "$type",
      "Codec": 81,
      "Value": "bnNlY3Rvci11cGRhdGVk"
    },
    {
      "Flags": 3,
      "Key": "sector",
      "Codec": 81,
      "Value": "GScT"
    },
    {
      "Flags": 3,
      "Key": "unsealed-cid",
      "Codec": 81,
      "Value": "2CpYKAABgeIDkiAg8pxbycLUKlq7v8IZyXva5vs6aSxy9/mz5t2ygXvZjAw="
    },
    {
      "Flags": 3,
      "Key": "piece-cid",
      "Codec": 81,
      "Value": "2CpYKAABgeIDkiAg8pxbycLUKlq7v8IZyXva5vs6aSxy9/mz5t2ygXvZjAw="
    },
    {
      "Flags": 1,
      "Key": "piece-size",
      "Codec": 81,
      "Value": "GwAAAAgAAAAA"
    }
  ],
```

And, from what I can see, `$type` is coming first on all of them.

I think it's that simple. Will be testing more today with my tooling to make sure my assumptions hold.

### DIFF
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -557,7 +557,7 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 		s = s + " WHERE " + strings.Join(clauses, " AND ")
 	}
 
-	s += " ORDER BY event.height DESC"
+	s += " ORDER BY event.height DESC, event_entry._rowid_ ASC"
 
 	stmt, err := ei.db.Prepare(s)
 	if err != nil {
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -557,6 +557,7 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 		s = s + " WHERE " + strings.Join(clauses, " AND ")
 	}
 
+	// retain insertion order of event_entry rows with the implicit _rowid_ column
 	s += " ORDER BY event.height DESC, event_entry._rowid_ ASC"
 
 	stmt, err := ei.db.Prepare(s)
--- a/itests/direct_data_onboard_test.go
+++ b/itests/direct_data_onboard_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/codec/dagcbor"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
 	"github.com/ipld/go-ipld-prime/node/basicnode"
 	"github.com/multiformats/go-multicodec"
 	"github.com/stretchr/testify/require"
@@ -24,6 +25,7 @@ import (
 	"github.com/filecoin-project/go-state-types/exitcode"
 	"github.com/filecoin-project/go-state-types/network"
 
+	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/market"
 	minertypes "github.com/filecoin-project/lotus/chain/actors/builtin/miner"
 	"github.com/filecoin-project/lotus/chain/consensus/filcns"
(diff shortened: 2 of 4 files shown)


itests/direct_data_onboard_test.go
```diff
@@ -282,21 +289,32 @@ func TestOnboardMixedMarketDDO(t *testing.T) {
...
-						case "deal-published", "deal-activated":
-							expectedEntries := []types.EventEntry{
-								{Flags: 0x03, Codec: uint64(multicodec.Cbor), Key: "$type", Value: keyBytes},
-								{Flags: 0x03, Codec: uint64(multicodec.Cbor), Key: "id", Value: must.One(ipld.Encode(basicnode.NewInt(2), dagcbor.Encode))},
-								{Flags: 0x03, Codec: uint64(multicodec.Cbor), Key: "client", Value: must.One(ipld.Encode(basicnode.NewInt(int64(clientId)), dagcbor.Encode))},
-								{Flags: 0x03, Codec: uint64(multicodec.Cbor), Key: "provider", Value: must.One(ipld.Encode(basicnode.NewInt(int64(minerId)), dagcbor.Encode))},
-							}
-							require.ElementsMatch(t, expectedEntries, event.Entries)
+				require.True(t, len(event.Entries) > 0)
+				if event.Entries[0].Key == "$type" && bytes.Equal(event.Entries[0].Value, keyBytes) {
```
rvagg, line 293: important test change here - `$type` should be first

itests/direct_data_onboard_test.go
```diff
@@ -282,21 +289,32 @@ func TestOnboardMixedMarketDDO(t *testing.T) {
...
+							{Flags: 0x03, Codec: uint64(multicodec.Cbor), Key: "$type", Value: must.One(ipld.Encode(basicnode.NewString("sector-activated"), dagcbor.Encode))},
+							{Flags: 0x03, Codec: uint64(multicodec.Cbor), Key: "sector", Value: must.One(ipld.Encode(basicnode.NewInt(int64(rawSector.Sector)), dagcbor.Encode))},
+							{Flags: 0x03, Codec: uint64(multicodec.Cbor), Key: "unsealed-cid", Value: must.One(ipld.Encode(basicnode.NewLink(cidlink.Link{Cid: expectCommD}), dagcbor.Encode))},
+							{Flags: 0x03, Codec: uint64(multicodec.Cbor), Key: "piece-cid", Value: must.One(ipld.Encode(basicnode.NewLink(cidlink.Link{Cid: marketPiece.PieceCID}), dagcbor.Encode))},
+							{Flags: 0x01, Codec: uint64(multicodec.Cbor), Key: "piece-size", Value: must.One(ipld.Encode(basicnode.NewInt(int64(marketPieceSize.Padded())), dagcbor.Encode))},
+							{Flags: 0x03, Codec: uint64(multicodec.Cbor), Key: "piece-cid", Value: must.One(ipld.Encode(basicnode.NewLink(cidlink.Link{Cid: rawPiece.PieceCID}), dagcbor.Encode))},
+							{Flags: 0x01, Codec: uint64(multicodec.Cbor), Key: "piece-size", Value: must.One(ipld.Encode(basicnode.NewInt(int64(rawPieceSize.Padded())), dagcbor.Encode))},
```
rvagg, line 313: important test change here - this is a sector activation with 2 pieces, we're asserting they are strictly in this order so the `piece-cid` + `piece-size` pairings are intact. Note also the `require.Equal` below (I've also replaced all the `require.ElementsMatch` elsewhere that operate on event elements).

aarshkshah1992 (approved):

rjan90 (approved):

chain/events/filter/index.go
```diff
@@ -557,7 +557,8 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 		s = s + " WHERE " + strings.Join(clauses, " AND ")
 	}
 
-	s += " ORDER BY event.height DESC"
+	// retain insertion order of event_entry rows with the implicit _rowid_ column
+	s += " ORDER BY event.height DESC, event_entry._rowid_ ASC"
```
Stebalien, line 561: Maybe we _should_ do a migration where we rename `_rowid_`? Can happen later.

PR #11852, opened 2024-04-08
fix(build): disable phoenix drand upgrade if LOTUS_DISABLE_DRAGON

rvagg:
This is a bit too late for 1.26.0 and 1.26.1 but it at least gives mitigation if there was some reason to use this env var to disable proper nv22 upgrade.

### DIFF
--- a/build/params_mainnet.go
+++ b/build/params_mainnet.go
@@ -132,6 +132,7 @@ func init() {
 
 	if os.Getenv("LOTUS_DISABLE_DRAGON") == "1" {
 		UpgradeDragonHeight = math.MaxInt64 - 1
+		delete(DrandSchedule, UpgradePhoenixHeight)
 		UpgradePhoenixHeight = math.MaxInt64
 	}
 


rvagg:
OK, this works, but the hack in `DrandConfigSchedule()` is less than ideal since it has to check for `DrandQuicknet` rather than the `UpgradePhoenixHeight` since that's already lost and I didn't want to hard-wire in `3817920+120`.

Probably too late for inclusion in a release though unless someone else is around to review?

build/drand.go
```diff
@@ -11,6 +12,9 @@ type DrandEnum int
 func DrandConfigSchedule() dtypes.DrandSchedule {
 	out := dtypes.DrandSchedule{}
 	for start, network := range DrandSchedule {
+		if os.Getenv("LOTUS_DISABLE_DRAGON") == "1" && network == DrandQuicknet {
```
Stebalien, line 15: Filter out `start < 0`? We always set disabled upgrade heights to 0.

build/params_mainnet.go
```diff
@@ -133,6 +133,7 @@ func init() {
 	if os.Getenv("LOTUS_DISABLE_DRAGON") == "1" {
 		UpgradeDragonHeight = math.MaxInt64 - 1
 		UpgradePhoenixHeight = math.MaxInt64
+		delete(DrandSchedule, UpgradePhoenixHeight)
```
Stebalien, line 136: This _shouldn't_ be necessary. (it won't actually work given the order of operations?)

build/drand.go
Stebalien: Gah!!! we should be setting those to -1, not infinity. Oh well...

Stebalien (commented):
Note: I believe the existing code "works" (technically) because we're setting the upgrade infinitely far into the future. Right?

build/drand.go
rvagg: too late cause it gets set as `3817920+120` before this, or init gets called

build/params_mainnet.go
rvagg: actually, if I just move it up one line then we're all good and I can remove the drand.go hack, done that now

rvagg:
> Note: I believe the existing code "works" (technically) because we're setting the upgrade infinitely far into the future. Right?

no, it doesn't work because we set this at the top of params:

```
var DrandSchedule = map[abi.ChainEpoch]DrandEnum{
	0:                    DrandIncentinet,
	UpgradeSmokeHeight:   DrandMainnet,
	UpgradePhoenixHeight: DrandQuicknet,
}
```

so `UpgradePhoenixHeight` is used before we change it in `init`.

But it worked in the version you reviewed and I just simplified it a lot in the latest iteration by moving the `delete` up one line. Can you 👁️ again please @Stebalien ?

PR #11861, opened 2024-04-09
ci: ci: create gh workflow that updates sorted pr checks

galargh:
## Related Issues

https://github.com/filecoin-project/lotus/issues/11734

## Proposed Changes

I propose to introduce a workflow that is going to post a nicely formatted list of completed checks on a PR once all the GitHub Actions workflow runs finish for the PR in question. 

## Additional Info

The issue with GitHub PR Checks was originally reported here - https://filecoinproject.slack.com/archives/CP50PPW2X/p1711707107172389

Unfortunately, it is unlikely to be resolved by GitHub any time soon - https://github.com/orgs/community/discussions/7885

We developed an alternative solution to the problem that relies on posting a sticky comment to a PR which a sorted/grouped list of checks after all the GitHub Actions workflows are finished - https://github.com/ipdxco/sorted-pr-checks

Please note that the comment only includes information about GitHub Actions workflows, it does NOT include CircleCI checks.

We can choose between two(three) flavours of the comment:
A. Sorted by result and then alphabetically - https://github.com/filecoin-project/lotus/pull/11861#issuecomment-2044753484
B. Grouped by result and then sorted alphabetically within the group - https://github.com/filecoin-project/lotus/pull/11861#issuecomment-2044919152
C. Create a new comment template

Given the number of checks that we have here, I'd suggest the grouped comment.

There is one more alternative solution to the problem. The Refined GitHub extension - https://github.com/refined-github/refined-github - implements GitHub status checks sorting. While the extension is really useful and makes using GitHub nicer, I don't think we can rely on it and expect all contributors to install it. That's why I'd suggest proceeding with the introduction of this workflow nonetheless.

## Checklist

- [x] CI is green

### DIFF
--- a/.github/workflows/sorted-pr-checks.yml
+++ b/.github/workflows/sorted-pr-checks.yml
@@ -29,3 +29,4 @@ jobs:
     uses: ipdxco/sorted-pr-checks/.github/workflows/comment.yml@v1
     with:
       pull_number: ${{ github.event.inputs.pull_number || github.event.workflow_run.pull_requests[0].number }}
+      template: grouped_by_result
--- a/.github/workflows/sorted-pr-checks.yml
+++ b/.github/workflows/sorted-pr-checks.yml
@@ -1,0 +1,31 @@
+name: Comment with sorted PR checks
+
+on:
+  workflow_dispatch:
+    inputs:
+      pull_number:
+        description: 'Pull request number'
+        required: true
+  workflow_run:
+    workflows:
+      - Build
+      - Check
+      - CodeQL
+      - Test
+    types:
+      - requested
+      - completed
+
+permissions:
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.inputs.pull_number || github.event.workflow_run.pull_requests[0].number || 'unknown' }}
+  cancel-in-progress: true
+
+jobs:
+  comment:
+    if: github.event.inputs.pull_number || github.event.workflow_run.pull_requests[0]
+    uses: ipdxco/sorted-pr-checks/.github/workflows/comment.yml@v1
+    with:
+      pull_number: ${{ github.event.inputs.pull_number || github.event.workflow_run.pull_requests[0].number }}
--- a/.github/workflows/sorted-pr-checks.yml
+++ b/.github/workflows/sorted-pr-checks.yml
@@ -29,4 +29,4 @@ jobs:
     uses: ipdxco/sorted-pr-checks/.github/workflows/comment.yml@v1
     with:
       pull_number: ${{ github.event.inputs.pull_number || github.event.workflow_run.pull_requests[0].number }}
-      template: grouped_by_result
+      template: unsuccessful_only


Stebalien:
Is there any way to just post failures? Although I guess they're fine if we group them.

Stebalien:
Eh, I don't have strong preferences.

masih:
I am not sure if this is worth having, TBH. It seems like a lot of duplicate information?

I believe the PR author would get an email notification of the failed CI workflow anyway and that should make it obvious what failed. 

I we absolutely want this, then only comment when there is a failure and only include failures?

galargh:
> Is there any way to just post failures? Although I guess they're fine if we group them.

Yes, of course. I added a new template for the comment that would skip successful checks. The comment, if something fails, would look something like this - https://github.com/ipdxco/sorted-pr-checks/pull/5#issuecomment-2049528780

> I am not sure if this is worth having, TBH. It seems like a lot of duplicate information?

Yes, all the information about the checks is available but not easily discoverable. The comment provides an easier way to find what failed among almost a hundred checks. The need for this addition was raised in https://filecoinproject.slack.com/archives/CP50PPW2X/p1711707107172389

> I believe the PR author would get an email notification of the failed CI workflow anyway, and that should make what failed obvious.

Yes - https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs. However, that information is limited to entire workflow statuses, and it makes it only to the one who triggered the workflow (in particular, reviewers don't get it).

> If we absolutely want this, then only comment when there is a failure and only include failures?

We added that option here - https://github.com/ipdxco/sorted-pr-checks/pull/5#issuecomment-2049528780 I'd suggest still keeping the comment if all the checks have passed. It's an easier lifecycle to manage, and a comment explicitly saying everything's OK means that users do not have to interpret what lack of comment means.

Let us know whether you'd like to proceed with this addition and, if so, with which template :)

Stebalien:
Honestly, a single short comment seems reasonable. I personally prefer the one that only lists the failures (or a single "all good"). However, we _probably_ want to collapse the list if we have more than, say, 10 failures (e.g., when one pushes a bad commit and everything fails).

rvagg:
I don't mind a sticky comment with the info, it'll make notifications noisy but if it's consistent then that's probably OK.

A sticky with only failures might be an improvement though, then one only shows up on failures. My notifications already get Circle failure messages but they come in separately to the issue notifications (I use email for these and like the threads to line up). Having failures in a message would probably be a solid improvement and line up nicely in the notifications.

.github/workflows/sorted-pr-checks.yml
```diff
@@ -0,0 +1,32 @@
...
+  group: ${{ github.workflow }}-${{ github.event.inputs.pull_number || github.event.workflow_run.pull_requests[0].number || 'unknown' }}
+  cancel-in-progress: true
+
+jobs:
+  comment:
+    if: github.event.inputs.pull_number || github.event.workflow_run.pull_requests[0]
+    uses: ipdxco/sorted-pr-checks/.github/workflows/comment.yml@v1
+    with:
+      pull_number: ${{ github.event.inputs.pull_number || github.event.workflow_run.pull_requests[0].number }}
+      template: grouped_by_result
```
galargh, line 32: This is how we can switch to listing unsuccessful checks only. At the moment, that setting doesn't group the failures, though. If you want us to add that, just let me know :)

```suggestion
      template: unsuccessful_only
```

rvagg: @galargh I think the general _vibe_ of the thread here is that we should do `unsuccessful_only` but it's good to know that `grouped_by_result` is a possibility if we want it. I assume that this means the sticky will only appear if there's unsuccessful and otherwise there'll be no noise?

galargh: It will always appear, even if all the checks pass. The comment appears as early as the workflows start to inform the user they are in progress. Once they're all finished, the action changes the content to match the workflows' outcomes. It's easier to implement that way and I believe it creates better expectations for the end user as they can always check the results in the comment.

rvagg (approved):
well, I guess we can give it a try and see how it feels

PR #11866, opened 2024-04-10
fix: local storage reservations fixes

magik6k:
## Proposed Changes
More debugging of accounting issues in the storage reservation mechanism in Curio.

* This PR adds a lot more debug logging to storage reservation code
* Fixes a bug in curio task storage where in some cases we'd release storage multiple times
  * Fixed with sync.once as other ways of fixing it would be either very complicated or would trade robustness in favour of sometimes maybe not releasing storage
* Also a small drive-by fix which drops existing SDR layers on sdr restart - seems like this was one of the causes of corrupted sectors

Currently testing on my mainnet setup, seems to work very well so far

## Checklist

- [ ] CI is green

### DIFF
--- a/storage/paths/local.go
+++ b/storage/paths/local.go
@@ -507,7 +507,7 @@ func (st *Local) Reserve(ctx context.Context, sid storiface.SectorRef, ft storif
 
 		resID := sectorFile{sid.ID, fileType}
 
-		log.Debugw("reserve add", "id", id, "sector", sid, "fileType", fileType, "overhead", overhead)
+		log.Debugw("reserve add", "id", id, "sector", sid, "fileType", fileType, "overhead", overhead, "reserved-before", p.reserved, "reserved-after", p.reserved+overhead)
 
 		p.reserved += overhead
 		p.reservations[resID] = overhead
@@ -519,7 +519,7 @@ func (st *Local) Reserve(ctx context.Context, sid storiface.SectorRef, ft storif
 			st.localLk.Lock()
 			defer st.localLk.Unlock()
 
-			log.Debugw("reserve done", "id", id, "sector", sid, "fileType", fileType, "overhead", overhead)
+			log.Debugw("reserve done", "id", id, "sector", sid, "fileType", fileType, "overhead", overhead, "reserved-before", p.reserved, "reserved-after", p.reserved-overhead)
 
 			p.reserved -= overhead
 			delete(p.reservations, resID)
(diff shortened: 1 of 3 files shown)


storage/paths/local.go
```diff
@@ -438,7 +468,24 @@ func (st *Local) Reserve(ctx context.Context, sid storiface.SectorRef, ft storif
...
+	done := func() {
+		if doneCalled {
+			curStack := make([]byte, 20480)
+			curStackN := runtime.Stack(curStack, false)
+
+			log.Errorw("double done call", "sector", sid, "fileType", ft, "prevStack", string(firstDonebuf[:firstDoneN]), "curStack", string(curStack[:curStackN]))
+		}
+
+		firstDonebuf = make([]byte, 20480)
+		firstDoneN = runtime.Stack(firstDonebuf, false)
```
snadrus, line 484: firstDonebuf = firstDonebuf[:runtime.Stack(firstDonebuf, false)]
Then you can get rid of firstDoneN

storage/paths/local.go
```diff
@@ -438,7 +468,24 @@ func (st *Local) Reserve(ctx context.Context, sid storiface.SectorRef, ft storif
...
+
+		doneCalled = true
+	}
+
 	deferredDone := func() { done() }
 	defer func() {
 		st.localLk.Unlock()
```
snadrus, line 491: Lets get rid of the confusing done(): 
1. Name Reserve()'s returns: (release func(), err error)
2. Here, if err != nil { log_the_stack } // is this the intent of the first done() ? 
3. The L527 done() could be inline. 
A double-done should be impossible.

snadrus (changes requested):

storage/paths/local.go
magik6k: > Here, if err != nil { log_the_stack } // is this the intent of the first done() ?

No, the intent of this stack logging was identifying where this double call was happening.

Can drop that code, but decided to leave as it doesn't do much harm, and if we ever hit that again the visibility will be very handy.

magik6k: > The L527 done() could be inline.

Can you elaborate what do you mean by that? That code just appends some cleanup logic to the `done` call chain; The call chain will only ever be called once by the defer if Reserve fails at any later point, or if it succeeds, by the release return.

The bug here was that in the successful case the release function could be called more than once by the task_storage code, which was a bug in that code.

snadrus (approved):

snadrus (approved):

PR #11869, opened 2024-04-12
feat: curio: sectors UI

snadrus:
## Related Issues
Resolves an item for the Curio beta epic.
Along with 'view', it offers 'terminate'. 
image

## Checklist

- [ ] CI is green


LexLuthr (changes requested):
I would suggest a few changes
1. We should convert sealed column to PresentOnDisk
2. Move unsealed next to sealed column
3. We don't need Fil/Day as the number would be really low to have any useful impact
4. Seal Proof is not required.

I could not find where we are getting the Size as type Sector{} doesn't seems to have one.

LexLuthr:
We also need to add another column to UI for "Proving" on chain. A sector can be on-chain but not proving.

curiosrc/web/static/sector/index.html
```diff
@@ -0,0 +1,117 @@
...
+      ],
+      order: [[1, 'asc']],
+      select: {
+        style: 'multi',
+        selector: 'td:first-child',
+        items: 'row',
+        rows: '%d rows selected',
+        headerCheckbox: true,
+      },
+      scrollY:        200,
```
LexLuthr, line 109: We should probably increase this to 1000. Current scroller is too small.

curiosrc/web/api/sector/sector.go
```diff
@@ -0,0 +1,197 @@
...
+		sNum uint64
+	}
+	sectorIdx := map[sectorID]int{}
+	for i, s := range sectors {
+		sectors[i].HasSealed = s.SectorFiletype&int(storiface.FTSealed) != 0 || s.SectorFiletype&int(storiface.FTUpdate) != 0
+		sectors[i].HasUnsealed = s.SectorFiletype&int(storiface.FTUnsealed) != 0
+		sectors[i].HasSnap = s.SectorFiletype&int(storiface.FTUpdate) != 0
+		if ss, err := s.SealProof.SectorSize(); err == nil {
+			sectors[i].SealInfo = ss.ShortString()
+		}
```
LexLuthr, line 101: This should be moved further down to when we have chainInfo. Currently, this displays wrong value.

curiosrc/web/static/sector/index.html
snadrus: hows  window.innerHeight - 150

snadrus:
> I would suggest a few changes
> 
> 1. We should convert sealed column to PresentOnDisk
> 2. Move unsealed next to sealed column
> 3. We don't need Fil/Day as the number would be really low to have any useful impact
> 4. Seal Proof is not required.
> 

I'd like to stick with the has-sealed and has-unsealed language as soon having one will mean that we can (automatically?) generate the other.  I want these to seem like 2 sides of the same coin (or a kind of backup) rather than two totally different concepts. 
Imagine if everywhere the answer was "false" we could have a top-level button that said "regenerate". Then it would be as simple as 
 - "search: has_sealed=false", 
 - "select-all"
 - "regenerate"

If the language was totally different between the two, then it would make things more confusing.

LexLuthr:
The naming convention sounds good based on your explanation.
I have halted the testing due a functionality bug in storage redeclaration. The sectors missing files are not updated in the DB. It prevents us from gauging the correct behaviour of this change under different scenarios. Let's keep this PR on hold till we have some fix for that bug.

LexLuthr (approved):

LexLuthr:
@magik6k Can you test the rendering on your miner? We don't have any good test setup with a lot of sectors.

LexLuthr:
Apart from adding 3 new columns for Proving, DealWeight and Deals, recent commit also reorders the row based on a hidden column called "Flag". We can flag a sector for any reason like not terminated by local sector is missing or it has expired. We reorder to quickly show SPs if they need to take action for any sector.  
Screenshot 2024-04-16 at 5 05 02 PM

LexLuthr:
@snadrus Testing results:
1. Terminate & Delete doesn't work for sector which are not yet on-chain. 
2. We should also check how terminate and remove is handled mid-task.

snadrus:
> @snadrus Testing results:
> 1. Terminate & Delete doesn't work for sector which are not yet on-chain. 
> 2. We should also check how terminate and remove is handled mid-task. 
> 3. We need to exclude piece type files when fetching all sector files. 

1. Not really a problem. 
2. Maybe just not allow it?
3. How are those getting through today's filter?

LexLuthr:
@snadrus We need a way to remove sectors safely. I will create a new issue for it. 
I have pushed 2 commits which fix all knows issues with Sector UI.
We have 2 pending issues left
1. OnChainInfo doesn't have deal IDs anymore. If this is post DDO then we can just estimate this from local data.
2. DDO testing failed as Market RPC has some bug which prevents AP.

PR #11884, opened 2024-04-16
feat: curio: storage index gc task

magik6k:
## Proposed Changes
* Add a harmonytask helper for defining "singleton" tasks - tasks which run as at most one instance across the whole cluster
  * This construction also helps with scheduled tasks which only run every so often without requiring a separate table for each task - which will be handy for other periodic cleanup tasks.
* Implement a storage path endpoint cleanup task
  * This task scans the sector index pinging all endpoints, and removes the ones which appear to be dead for a long time
    * (lotus-miner never had that, but because the index there was in-memory, cleanup was done by restarting the process..)
  * Curio needs this because the index is now in the database, and after migrating from lotus-miner there are guaranteed to be stale endpoints in the sector DBIndex. The remote storage code can cope with those, but it's not great to keep dead references in the index.

## Additional Info
TODO:
* [ ] Do a dry run on a setup with stale references
* [ ] Do a real run
* [ ] Increase intervals, drop the dryrun guard

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/curio/tasks/tasks.go
+++ b/cmd/curio/tasks/tasks.go
@@ -3,7 +3,7 @@ package tasks
 
 import (
 	"context"
-	"github.com/filecoin-project/lotus/curiosrc/gc"
+
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/samber/lo"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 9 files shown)


snadrus (approved):

snadrus:
This is a great basis for a "follows" task that reviews sector paths to determine what was on a defunct storage path. It could:
- warn
- increase the count of available copies
- unseal to recreate the pieces and/or unsealed copy/copies.
- re-seal (with appropriate context) to recreate what sealed copies were lost.

We could also verify daily. We could verify copy count too.

PR #11905, opened 2024-04-19
fix: ETH RPC API: ETH Call should use the parent state root of the subsequent tipset

### LINKED ISSUE #11205: `eth_call` incorrectly reconciling deferred (Filecoin) and immediate (Ethereum) execution
Context: https://github.com/cerc-io/stack-orchestrator/pull/509#issuecomment-1691735496

Ethereum clients and libraries expect that calling `eth_call` with a given epoch (e.g. 1234) will apply the call on the **_output_** state of the supplied epoch. In Ethereum, the output is computed immediately and the root hash is part of the block header of height 1234. In Filecoin, deferred execution causes the output of epoch 1234 to be calculated on the next non-null round, normally 1235 (but not necessarily).

The logic [here](https://github.com/filecoin-project/lotus/blob/41fce94db48fbed79fe47549cc7b5a0c3c78c708/node/impl/full/eth.go#L892) is flawed.

- The correct behaviour is to apply the call on the parent state root of the next tipset (execution tipset).
	- If this supplied epoch is the "head" epoch, we should error out, thus preserving consistency with the rest of the Eth module (ie. the tipset has still not been executed and is not "exposed" to Eth endpoint clients).
- We should not be doing any trickery with individual messages (e.g. finding prior messages, filtering, etc.) down the line, _at all_ (see callees of this function).
- The `LOTUS_SKIP_APPLY_TS_MESSAGE_CALL_WITH_GAS` env var introduced in #10546 must go.
	- This causes inconsistent behaviour across RPC endpoints in the ecosystem depending on how this env var is configured.
	- When enabled, it _deliberately_ fails to meet the required functionality / expectations of `eth_call`.
	- Due to a chain of circumstances, all new RPC endpoints being set up in the Filecoin ecosystem have this flag enabled.
	- Therefore, all new endpoints are failing to meet expectations.

aarshkshah1992:
## Related Issues
Closes #11205 

## Proposed Changes
ETH Call should use the parent state root of the subsequent tipset. This can be obtained from the `TipsetState` on the `Statemanager`

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/call.go
+++ b/chain/stmgr/call.go
@@ -48,12 +48,17 @@ func (sm *StateManager) Call(ctx context.Context, msg *types.Message, ts *types.
 		msg.Value = types.NewInt(0)
 	}
 
-	return sm.callInternal(ctx, msg, nil, ts, cid.Undef, sm.GetNetworkVersion, false, false)
+	return sm.callInternal(ctx, msg, nil, ts, cid.Undef, sm.GetNetworkVersion, false, false, false)
+}
+
+// ApplyOnStateWithGas applies the given message on top of the given state root with gas tracing enabled
+func (sm *StateManager) ApplyOnStateWithGas(ctx context.Context, stateCid cid.Cid, msg *types.Message, ts *types.TipSet) (*api.InvocResult, error) {
+	return sm.callInternal(ctx, msg, nil, ts, stateCid, sm.GetNetworkVersion, true, false, true)
 }
 
 // CallWithGas calculates the state for a given tipset, and then applies the given message on top of that state.
 func (sm *StateManager) CallWithGas(ctx context.Context, msg *types.Message, priorMsgs []types.ChainMsg, ts *types.TipSet, applyTsMessages bool) (*api.InvocResult, error) {
-	return sm.callInternal(ctx, msg, priorMsgs, ts, cid.Undef, sm.GetNetworkVersion, true, applyTsMessages)
+	return sm.callInternal(ctx, msg, priorMsgs, ts, cid.Undef, sm.GetNetworkVersion, true, applyTsMessages, false)
 }
 
 // CallAtStateAndVersion allows you to specify a message to execute on the given stateCid and network version.
@@ -65,13 +70,13 @@ func (sm *StateManager) CallAtStateAndVersion(ctx context.Context, msg *types.Me
 		return v
 	}
 
-	return sm.callInternal(ctx, msg, nil, nil, stateCid, nvGetter, true, false)
+	return sm.callInternal(ctx, msg, nil, nil, stateCid, nvGetter, true, false, false)
 }
 
 //   - If no tipset is specified, the first tipset without an expensive migration or one in its parent is used.
 //   - If executing a message at a given tipset or its parent would trigger an expensive migration, the call will
 //     fail with ErrExpensiveFork.
-func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, priorMsgs []types.ChainMsg, ts *types.TipSet, stateCid cid.Cid, nvGetter rand.NetworkVersionGetter, checkGas, applyTsMessages bool) (*api.InvocResult, error) {
+func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, priorMsgs []types.ChainMsg, ts *types.TipSet, stateCid cid.Cid, nvGetter rand.NetworkVersionGetter, checkGas, applyTsMessages bool, noReExec bool) (*api.InvocResult, error) {
 	ctx, span := trace.StartSpan(ctx, "statemanager.callInternal")
 	defer span.End()
 
@@ -117,24 +122,6 @@ func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, pr
 	if stateCid == cid.Undef {
 		stateCid = ts.ParentState()
 	}
-	tsMsgs, err := sm.cs.MessagesForTipset(ctx, ts)
-	if err != nil {
-		return nil, xerrors.Errorf("failed to lookup messages for parent tipset: %w", err)
-	}
-
-	if applyTsMessages {
-		priorMsgs = append(tsMsgs, priorMsgs...)
-	} else {
-		var filteredTsMsgs []types.ChainMsg
-		for _, tsMsg := range tsMsgs {
-			//TODO we should technically be normalizing the filecoin address of from when we compare here
-			if tsMsg.VMMessage().From == msg.VMMessage().From {
-				filteredTsMsgs = append(filteredTsMsgs, tsMsg)
-			}
-		}
-		priorMsgs = append(filteredTsMsgs, priorMsgs...)
-	}
-
 	// Technically, the tipset we're passing in here should be ts+1, but that may not exist.
 	stateCid, err = sm.HandleStateForks(ctx, stateCid, ts.Height(), nil, ts)
 	if err != nil {
@@ -169,18 +156,38 @@ func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, pr
 	if err != nil {
 		return nil, xerrors.Errorf("failed to set up vm: %w", err)
 	}
-	for i, m := range priorMsgs {
-		_, err = vmi.ApplyMessage(ctx, m)
-		if err != nil {
-			return nil, xerrors.Errorf("applying prior message (%d, %s): %w", i, m.Cid(), err)
-		}
-	}
 
-	// We flush to get the VM's view of the state tree after applying the above messages
-	// This is needed to get the correct nonce from the actor state to match the VM
-	stateCid, err = vmi.Flush(ctx)
-	if err != nil {
-		return nil, xerrors.Errorf("flushing vm: %w", err)
+	if !noReExec {
+		tsMsgs, err := sm.cs.MessagesForTipset(ctx, ts)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to lookup messages for parent tipset: %w", err)
+		}
+
+		if applyTsMessages {
+			priorMsgs = append(tsMsgs, priorMsgs...)
+		} else {
+			var filteredTsMsgs []types.ChainMsg
+			for _, tsMsg := range tsMsgs {
+				//TODO we should technically be normalizing the filecoin address of from when we compare here
+				if tsMsg.VMMessage().From == msg.VMMessage().From {
+					filteredTsMsgs = append(filteredTsMsgs, tsMsg)
+				}
+			}
+			priorMsgs = append(filteredTsMsgs, priorMsgs...)
+		}
+		for i, m := range priorMsgs {
+			_, err = vmi.ApplyMessage(ctx, m)
+			if err != nil {
+				return nil, xerrors.Errorf("applying prior message (%d, %s): %w", i, m.Cid(), err)
+			}
+		}
+
+		// We flush to get the VM's view of the state tree after applying the above messages
+		// This is needed to get the correct nonce from the actor state to match the VM
+		stateCid, err = vmi.Flush(ctx)
+		if err != nil {
+			return nil, xerrors.Errorf("flushing vm: %w", err)
+		}
 	}
 
 	stTree, err := state.LoadStateTree(cbor.NewCborStore(buffStore), stateCid)
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -980,14 +980,14 @@ func (a *EthModule) applyMessage(ctx context.Context, msg *types.Message, tsk ty
 		return nil, xerrors.Errorf("cannot get tipset: %w", err)
 	}
 
-	applyTsMessages := true
-	if os.Getenv("LOTUS_SKIP_APPLY_TS_MESSAGE_CALL_WITH_GAS") == "1" {
-		applyTsMessages = false
-	}
-
 	// Try calling until we find a height with no migration.
 	for {
-		res, err = a.StateManager.CallWithGas(ctx, msg, []types.ChainMsg{}, ts, applyTsMessages)
+		st, _, err := a.StateManager.TipSetState(ctx, ts)
+		if err != nil {
+			return nil, xerrors.Errorf("cannot get tipset state: %w", err)
+		}
+
+		res, err = a.StateManager.ApplyOnStateWithGas(ctx, st, msg, ts)
 		if err != stmgr.ErrExpensiveFork {
 			break
 		}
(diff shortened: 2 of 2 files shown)


jennijuju:
@aarshkshah1992  - can you please assign a reviewer?

Stebalien:
Well, this is bringing back past nightmares with respect to https://github.com/filecoin-project/lotus/pull/10216.

aarshkshah1992:
@Stebalien Are you saying that this wont work ?

chain/stmgr/call.go
```diff
@@ -64,14 +69,13 @@ func (sm *StateManager) CallAtStateAndVersion(ctx context.Context, msg *types.Me
...
-
-	return sm.callInternal(ctx, msg, nil, nil, stateCid, nvGetter, true, false)
+	return sm.callInternal(ctx, msg, nil, nil, stateCid, nvGetter, true, false, false)
 }
 
 //   - If no tipset is specified, the first tipset without an expensive migration or one in its parent is used.
 //   - If executing a message at a given tipset or its parent would trigger an expensive migration, the call will
 //     fail with ErrExpensiveFork.
-func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, priorMsgs []types.ChainMsg, ts *types.TipSet, stateCid cid.Cid, nvGetter rand.NetworkVersionGetter, checkGas, applyTsMessages bool) (*api.InvocResult, error) {
+func (sm *StateManager) callInternal(ctx context.Context, msg *types.Message, priorMsgs []types.ChainMsg, ts *types.TipSet, stateCid cid.Cid, nvGetter rand.NetworkVersionGetter, checkGas, applyTsMessages bool, noReExec bool) (*api.InvocResult, error) {
```
rvagg, line 78: The negative `no` here is awkward, because you further go on to negate it with `!noReExec`, so we have the first cognitive hurdle. We also have two interacting booleans, an additional cognitive hurdle, and one where the second bool can negate the first, yet another hurdle.

Maybe an enum should be used here to express the 3 cases, replacing both booleans, and then we get to name the strategies better:

```go
type execMessageStrategy int

const (
	execNoMessages         execMessageStrategy = iota // apply no prior or current tipset messages
	execAllMessages                                   // apply all prior and current tipset messages
	execSameSenderMessages                            // apply all prior messages and any current tipset messages from the same sender
)
```

node/impl/full/eth.go
```diff
@@ -980,14 +980,14 @@ func (a *EthModule) applyMessage(ctx context.Context, msg *types.Message, tsk ty
...
 
-	applyTsMessages := true
-	if os.Getenv("LOTUS_SKIP_APPLY_TS_MESSAGE_CALL_WITH_GAS") == "1" {
-		applyTsMessages = false
-	}
-
 	// Try calling until we find a height with no migration.
 	for {
-		res, err = a.StateManager.CallWithGas(ctx, msg, []types.ChainMsg{}, ts, applyTsMessages)
+		st, _, err := a.StateManager.TipSetState(ctx, ts)
```
rvagg, line 985: Working through the logic here - `TipSetState` will end up executing the messages on top of parent state if the requested tipset doesn't already have a height+1 not on a different fork that we can use to get the state from.

When you get down to the bottom of `tipSetState()`, having failed to find a valid existing height+1 state, you run `ExecuteTipSet`, which itself will call `HandleStateForks` which the `callInternal()` code above in this PR does so a bunch of work to avoid calling if it can tell that there's an "expensive" migration.

So we might need to port that logic down to here to handle the expensive-migration case.

But other than that, I think this results in roughly the same outcome of having all current tipset messages applied to produce a new state and then applying _only_ the requested `msg` on top of that.

I'm not sure if/how null rounds come in to play with all of this though, maybe it doesn't matter as per the discussion in #11205.

aarshkshah1992:
Hey @raulk 

This PR needs a review from someone who understands the ETH RPC stack really well. But @Stebalien is fully focused on F3 currently. Would you or Mikers or Fridrick be able to help review this by any chance ?

node/impl/full/eth.go
aarshkshah1992: @rvagg 

Given that `TipSetState ` already handles state forks for us, can we just not re-do it in `callInternal` for our use case ? We can control that behaviour in `callInternal` with a flag.

chain/stmgr/call.go
aarshkshah1992: This is done.

node/impl/full/eth.go
rvagg: > already handles state forks for us,

but that's the problem, imagine if we did this across an expensive migration boundary and triggered a migration just to get the state at this tipset.

All of this logic exists to check that condition: https://github.com/filecoin-project/lotus/blob/187e837ab8bf8261bc704daeb42de7348c294274/chain/stmgr/call.go#L104-L132

But that's only applied after you've potentially executed that tipset because `TipSetState` can result in an `ExecuteTipSet` if it doesn't already have it handy, that executes `ApplyBlocks` which in turn runs `HandleStateForks`.

So I **_think_**, if you run this, it'll still do what it says in the comment at the top of the `for` ("Try calling until we find a height with no migration."), but it'll actually perform a migration _before_ it checks whether there's an expensive migration.

I believe what we want is to run an `hasExpensiveForkBetween(pts.Height(), ts.Height()+1)` ourselves and walk back up parents until it's `false`. But currently `hasExpensiveForkBetween` is a private method of `StateManager`.

aarshkshah1992: Ahaa good catch @rvagg I have fixed this.

aarshkshah1992:
@rvagg Have addressed both your review comments. Please can you take one final look at the PR ?

node/impl/full/eth.go
```diff
@@ -980,25 +980,32 @@ func (a *EthModule) applyMessage(ctx context.Context, msg *types.Message, tsk ty
...
+			}
+			if !a.StateManager.HasExpensiveForkBetween(pts.Height(), ts.Height()+1) {
+				break
+			}
+			ts, err = a.Chain.GetTipSetFromKey(ctx, ts.Parents())
+			if err != nil {
+				return nil, xerrors.Errorf("getting parent tipset: %w", err)
+			}
+		}
 	}
```
rvagg, line 998: Actually I think the correct thing to do is not loop because the request is for an explicit tipset, so we can't just adjust the tipset they want. I believe the right behaviour is to just check whether there's an expensive fork and error if not. Perhaps like this:

```suggestion
	if ts.Height() > 0 {
		pts, err := a.Chain.GetTipSetFromKey(ctx, ts.Parents())
		if err != nil {
			return nil, xerrors.Errorf("failed to find a non-forking epoch: %w", err)
		}
		// Check for expensive forks from the parents to the tipset, including nil tipsets
		if a.StateManager.HasExpensiveForkBetween(pts.Height(), ts.Height()+1) {
			return nil, stmgr.ErrExpensiveFork
		}
	}
```

Then the call just won't work for those major upgrade epochs. I don't know if that's going to cause complaints but it's only a few every year.

aarshkshah1992: @rvagg But the pre-existing behaviour does loop back till it can find an appropriate tipset. I am not sure we should change that behaviour.

rvagg: only if `ts` is not supplied; but I think here we always have a `ts` so we're in the `else` part of this block: https://github.com/filecoin-project/lotus/blob/f75bdd1f262483ec13e82603687f7cf8b9d893b8/chain/stmgr/call.go#L104-L132

rvagg: @aarshkshah1992 ping on this; I think we should be able to get this wrapped up after dealing with this as long as we can get an expert reviewer to eyeball it.

momack2:
@rvagg @aarshkshah1992 can we get a status update here? What's needed to get this fully reviewed?

rvagg:
@momack2 just waiting on expert eyes to review so the blind don't lead the blind down a dodgy alleyway

node/impl/full/eth.go
rvagg: I think we agree that the loop can go, there's no need to "walk". The `if `ts.Height()` can go too as per @raulk's comment of there always being a height, the remaining concern here is about expensive forks and whether to do anything about them.

@raulk the current code path for this hits that check already, the reason for inserting it in here is that the new code bypasses it by taking shortcuts to the execution. This block was intended to restore the existing behaviour of returning an `ErrExpensiveFork` if you happen to hit an upgrade epoch with the call.

Is that an unreasonable thing to do for this call? Would there be a security concern if a public API allowed users to make this call repeatedly across that boundary? tbh I'm not sure of the implications of any of this other than the existing code seems to go to some effort to ensure that you don't try and load state for a manual execution where that might involve a big migration.

aarshkshah1992: @rvagg 

- I see your point about removing the loop. Yeah, we always have a tipset present (is resolved to the head if none is given by the user, same for pending). Fixed this.

node/impl/full/eth.go
```diff
@@ -980,25 +980,32 @@ func (a *EthModule) applyMessage(ctx context.Context, msg *types.Message, tsk ty
 		return nil, xerrors.Errorf("cannot get tipset: %w", err)
 	}
 
-	applyTsMessages := true
-	if os.Getenv("LOTUS_SKIP_APPLY_TS_MESSAGE_CALL_WITH_GAS") == "1" {
-		applyTsMessages = false
+	// Walk back until we find a tipset where no state migration is needed
```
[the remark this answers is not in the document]
aarshkshah1992: @raulk Have removed this. If the user specifies a tipset that causes an expensive fork, we simply error out.

Note: The erroring out on an expensive fork is pre-existing behaviour that we are not changing  with this PR.

aarshkshah1992:
@raulk I agree. We have removed the looping so if the user specifies a tipset that can cause an expensive fork, we just error out the request(should be few of those as Rodd mentions).

aarshkshah1992:
@momack2 We were waiting on a review from someone who understands the ETH<>Filecoin stack well. Now that Raul has reviewed it, we can land this. Have addressed the latest round of review. Will merge on a green tick.

rvagg (approved):
I wouldn't mind hearing @raulk's response to the fact that we error on a migration epoch if he has time to chime in. But, it's also something that can be easily changed at a future date if we hear objections after this lands.
It might be good to bring up in a sync meeting actually, just that specific point.
Otherwise this lgtm.

aarshkshah1992:
Will address @raulk 's response in a future PR when he takes a look at this as we're not changing any pre-existing behaviour here wrt handling migration epochs.

raulk (changes requested):

node/impl/full/eth.go
```diff
@@ -980,25 +980,26 @@ func (a *EthModule) applyMessage(ctx context.Context, msg *types.Message, tsk ty
...
+		pts, err := a.Chain.GetTipSetFromKey(ctx, ts.Parents())
 		if err != nil {
-			return nil, xerrors.Errorf("getting parent tipset: %w", err)
+			return nil, xerrors.Errorf("failed to find a non-forking epoch: %w", err)
+		}
+		// Check for expensive forks from the parents to the tipset, including nil tipsets
+		if a.StateManager.HasExpensiveForkBetween(pts.Height(), ts.Height()+1) {
+			return nil, stmgr.ErrExpensiveFork
 		}
 	}
```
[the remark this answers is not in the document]
aarshkshah1992: @raulk 

I agree with what you say here but the work to fix the event index is of a larger scope than just fixing the ETH_CALL API. This PR aims to fix two problems with the `ETH_CALL` API:

1. Get rid of the `LOTUS_SKIP_APPLY_TS_MESSAGE_CALL_WITH_GAS ` flag that is almost invariably turned on by all node operators thus hampering the correctness of this API (you don't apply the user tx on top of the state at the end of the epoch)

2. Ensure that `ETH_CALL` never re-executes transactions for an epoch if it already has the state for them.

Both these problems are solved by the PR as it stands right now. We can do a follow-up workstream to prioritise "fixing" the events index so we don't re-execute txns in `TipSetState` but we can merge this PR without that.

cc @snissn @trruckerfling

aarshkshah1992:
@raulk Can we go ahead and merge this one given that we're now co-ordinating on a separate workstream for events ?

aarshkshah1992:
@raulk Vulcanise is also running into some of the rough edges related to the events. So we are starting the work with fixing one of the issues they're running into i.e. async indexing of events leading to races where client discovers a block but does not get any events for them because they're yet to be indexed. 

However, please note that we don't have dedicated epic to fix all of the issues you and @rvagg have enlisted. If doing all of those is important, would be great to do a prioritsation epic for this workstream with @jennijuju and @rvagg.

snissn:
@aarshkshah1992 this PR sped the eth_call RPC call up considerably!

Before:
```
ab -p postdata.json -T application/json -c 10 -n 1000 http://127.0.0.1:1234/rpc/v1
This is ApacheBench, Version 2.3 <$Revision: 1879490 $>
                                                                                 
Benchmarking 127.0.0.1 (be patient)                                                        

Server Software:        
Server Hostname:        127.0.0.1
Server Port:            1234

Document Path:          /rpc/v1
Document Length:        103 bytes

Concurrency Level:      10
Time taken for tests:   127.168 seconds
Complete requests:      1000
Failed requests:        0
Total transferred:      221000 bytes
Total body sent:        394000
HTML transferred:       103000 bytes
Requests per second:    7.86 [#/sec] (mean)
Time per request:       1271.684 [ms] (mean)
Time per request:       127.168 [ms] (mean, across all concurrent requests)
Transfer rate:          1.70 [Kbytes/sec] received
                        3.03 kb/s sent
                        4.72 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:   302 1264 589.7   1112    3408
Waiting:      302 1264 589.7   1112    3408
Total:        302 1264 589.8   1112    3408

Percentage of the requests served within a certain time (ms)
  50%   1112
  66%   1415
  75%   1647
  80%   1785
  90%   2143
  95%   2402
  98%   2741
  99%   2865
 100%   3408 (longest request)
```

after:
```
$ ab -p postdata.json -T application/json -c 10 -n 1000 http://127.0.0.1:1234/rpc/v1
This is ApacheBench, Version 2.3 <$Revision: 1879490 $>

Server Software:        
Server Hostname:        127.0.0.1
Server Port:            1234

Document Path:          /rpc/v1
Document Length:        103 bytes

Concurrency Level:      10
Time taken for tests:   1.524 seconds
Complete requests:      1000
Failed requests:        0
Total transferred:      221000 bytes
Total body sent:        394000
HTML transferred:       103000 bytes
Requests per second:    656.21 [#/sec] (mean)
Time per request:       15.239 [ms] (mean)
Time per request:       1.524 [ms] (mean, across all concurrent requests)
Transfer rate:          141.62 [Kbytes/sec] received
                        252.49 kb/s sent
                        394.11 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:     4   15  23.0      8     221
Waiting:        4   15  22.9      8     219
Total:          4   15  23.0      8     221

Percentage of the requests served within a certain time (ms)
  50%      8
  66%     10
  75%     13
  80%     16
  90%     29
  95%     49
  98%     81
  99%    131
 100%    221 (longest request)         
```

PR #11919, opened 2024-04-23
Refactor `LookupID*` APIs in `StateManager` and `StateTree`

masih:
## Related Issues
* #11594 

## Proposed Changes

The naming of `LookupID` can cause confusion when resolving actor IDs vs
 ID addresses. To avoid this:

* Refactor `StateTree` `LookupID` to `LookupIDAddress`, because it returns ID address.
* Refactor `StateManager` `LookupID` to `LookupIDAddress` because it also returns ID address via a chain call to `StateTree`.
* Introduce a new API `StateManager` dedicated to resolving address to actor ID, called `LookupID` which returns `abi.ActorID`.

## Additional Info

For context, see:
 * https://github.com/filecoin-project/lotus/pull/11723#discussion_r1534728607
## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/stmgr.go
+++ b/chain/stmgr/stmgr.go
@@ -49,7 +49,7 @@ type StateManagerAPI interface {
 	Call(ctx context.Context, msg *types.Message, ts *types.TipSet) (*api.InvocResult, error)
 	GetPaychState(ctx context.Context, addr address.Address, ts *types.TipSet) (*types.Actor, paych.State, error)
 	LoadActorTsk(ctx context.Context, addr address.Address, tsk types.TipSetKey) (*types.Actor, error)
-	LookupID(ctx context.Context, addr address.Address, ts *types.TipSet) (address.Address, error)
+	LookupIDAddress(ctx context.Context, addr address.Address, ts *types.TipSet) (address.Address, error)
 	ResolveToDeterministicAddress(ctx context.Context, addr address.Address, ts *types.TipSet) (address.Address, error)
 }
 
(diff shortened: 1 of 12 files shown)


chain/stmgr/stmgr.go
```diff
@@ -400,13 +400,30 @@ func (sm *StateManager) GetBlsPublicKey(ctx context.Context, addr address.Addres
 	return kaddr.Payload(), nil
 }
 
-func (sm *StateManager) LookupID(ctx context.Context, addr address.Address, ts *types.TipSet) (address.Address, error) {
+func (sm *StateManager) LookupIDAddress(_ context.Context, addr address.Address, ts *types.TipSet) (address.Address, error) {
+	// Check for the fast route first to avoid unnecessary CBOR store instantiation and state tree load.
+	if addr.Protocol() == address.ID {
```
ZenGround0, line 405: nice

chain/stmgr/stmgr.go
```diff
@@ -400,13 +400,30 @@ func (sm *StateManager) GetBlsPublicKey(ctx context.Context, addr address.Addres
...
-	return state.LookupID(addr)
+	return state.LookupIDAddress(addr)
+}
+
+func (sm *StateManager) LookupID(ctx context.Context, addr address.Address, ts *types.TipSet) (abi.ActorID, error) {
+	idAddr, err := sm.LookupIDAddress(ctx, addr, ts)
+	if err != nil {
+		return 0, xerrors.Errorf("state manager lookup id: %w", err)
+	}
+	actor, err := address.IDFromAddress(idAddr)
```
ZenGround0, line 422: nit: I'd call this `id` instead of `actor`

ZenGround0 (approved):

rvagg (approved):
sgtm

It has me wondering whether we should have a label, or some other way to identify breaking changes so we can keep them separate and call them out specifically. The distinction between breaking _code_ API changes vs breaking _RPC_ API changes also seems worth separating out too, because this change is going to be much less impactful than an RPC API change. @rjan90 what do you think? Maybe at a minimum we should ensure we add a CHANGELOG entry in a PR like this to call out the change for anyone using lotus as a library.

PR #11926, opened 2024-04-24
feat: curio: alertManager

LexLuthr:
## Proposed Changes

This PRs adds a new singleton task for alerting.
The alerting task can create an incident for a previously configured service.
User should get critical alerts via SMS, email and call. 

- [x] Test in a running Curio cluster before merging

## Additional Info

A new page about alerting is required in Curio docs. We also need to create a section on how to sign up and configure PagerDuty.

## Checklist

- [ ] CI is green

### DIFF
--- a/curiosrc/alertmanager/alerts.go
+++ b/curiosrc/alertmanager/alerts.go
@@ -6,10 +6,12 @@ import (
 	"strings"
 
 	"github.com/BurntSushi/toml"
+	"github.com/dustin/go-humanize"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
 
 	"github.com/filecoin-project/lotus/node/config"
 )
@@ -20,7 +22,7 @@ import (
 // It queries the database for the configuration of each layer and decodes it using the toml.Decode function.
 // It then iterates over the addresses in the configuration and curates a list of unique addresses.
 // If an address is not found in the chain node, it adds an alert to the alert map.
-// If the balance of an address is below 5 Fil, it adds an alert to the alert map.
+// If the balance of an address is below MinimumWalletBalance, it adds an alert to the alert map.
 // If there are any errors encountered during the process, the err field of the alert map is populated.
 func balanceCheck(al *alerts) {
 	Name := "Balance Check"
@@ -267,22 +269,43 @@ func permanentStorageCheck(al *alerts) {
 		return
 	}
 
-	var totalRequiredSpace int64
+	type sm struct {
+		s    sector
+		size int64
+	}
+
+	sectorMap := make(map[sm]bool)
+
 	for _, sec := range sectors {
+		space := int64(0)
 		sec := sec
 		sectorSize, err := sec.Proof.SectorSize()
 		if err != nil {
-			totalRequiredSpace += int64(64 << 30)
+			space = int64(64<<30)*2 + int64(200<<20) // Assume 64 GiB sector
+		} else {
+			space = int64(sectorSize)*2 + int64(200<<20) // sealed + unsealed + cache
 		}
-		totalRequiredSpace += int64(sectorSize)
+
+		key := sm{s: sec, size: space}
+
+		sectorMap[key] = false
+
+		for _, strg := range storages {
+			if space > strg.Available {
+				strg.Available -= space
+				sectorMap[key] = true
+			}
+		}
 	}
 
-	var totalAvailableSpace int64
-	for _, storage := range storages {
-		totalAvailableSpace += storage.Available
+	missingSpace := big.NewInt(0)
+	for sec, accounted := range sectorMap {
+		if !accounted {
+			big.Add(missingSpace, big.NewInt(sec.size))
+		}
 	}
 
-	if totalAvailableSpace < totalRequiredSpace {
-		al.alertMap[Name].alertString = fmt.Sprintf("Insufficient storage space for sealing sectors. Required: %d bytes, Available: %d bytes", totalRequiredSpace, totalAvailableSpace)
+	if missingSpace.GreaterThan(big.NewInt(0)) {
+		al.alertMap[Name].alertString = fmt.Sprintf("Insufficient storage space for sealing sectors. Additional %s required.", humanize.Bytes(missingSpace.Uint64()))
 	}
 }
--- a/curiosrc/alertmanager/alerts.go
+++ b/curiosrc/alertmanager/alerts.go
@@ -1,0 +1,288 @@
+package alertmanager
+
+import (
+	"database/sql"
+	"fmt"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/lotus/node/config"
+)
+
+// balanceCheck retrieves the machine details from the database and performs balance checks on unique addresses.
+// It populates the alert map with any errors encountered during the process and with any alerts related to low wallet balance and missing wallets.
+// The alert map key is "Balance Check".
+// It queries the database for the configuration of each layer and decodes it using the toml.Decode function.
+// It then iterates over the addresses in the configuration and curates a list of unique addresses.
+// If an address is not found in the chain node, it adds an alert to the alert map.
+// If the balance of an address is below 5 Fil, it adds an alert to the alert map.
+// If there are any errors encountered during the process, the err field of the alert map is populated.
+func balanceCheck(al *alerts) {
+	Name := "Balance Check"
+	al.alertMap[Name] = &alertOut{}
+
+	// MachineDetails represents the structure of data received from the SQL query.
+	type machineDetail struct {
+		ID          int
+		HostAndPort string
+		Layers      string
+	}
+	var machineDetails []machineDetail
+
+	// Get all layers in use
+	err := al.db.Select(al.ctx, &machineDetails, `
+				SELECT m.id, m.host_and_port, d.layers
+				FROM harmony_machines m
+				LEFT JOIN harmony_machine_details d ON m.id = d.machine_id;`)
+	if err != nil {
+		al.alertMap[Name].err = xerrors.Errorf("getting config layers for all machines: %w", err)
+		return
+	}
+
+	// UniqueLayers takes an array of MachineDetails and returns a slice of unique layers.
+
+	layerMap := make(map[string]bool)
+	var uniqueLayers []string
+
+	// Get unique layers in use
+	for _, machine := range machineDetails {
+		machine := machine
+		// Split the Layers field into individual layers
+		layers := strings.Split(machine.Layers, ",")
+		for _, layer := range layers {
+			layer = strings.TrimSpace(layer)
+			if _, exists := layerMap[layer]; !exists && layer != "" {
+				layerMap[layer] = true
+				uniqueLayers = append(uniqueLayers, layer)
+			}
+		}
+	}
+
+	addrMap := make(map[string]bool)
+	var uniqueAddrs []string
+
+	// Get all unique addresses
+	for _, layer := range uniqueLayers {
+		text := ""
+		cfg := config.DefaultCurioConfig()
+		err := al.db.QueryRow(al.ctx, `SELECT config FROM harmony_config WHERE title=$1`, layer).Scan(&text)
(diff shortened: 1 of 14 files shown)


curiosrc/alertmanager/task_alert.go
```diff
@@ -0,0 +1,395 @@
...
+
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/lib/harmony/harmonydb"
+	"github.com/filecoin-project/lotus/lib/harmony/harmonytask"
+	"github.com/filecoin-project/lotus/lib/harmony/resources"
+	"github.com/filecoin-project/lotus/node/config"
+	"github.com/filecoin-project/lotus/storage/ctladdr"
+)
+
+var MinimumBalance = types.MustParseFIL("5 FIL")
```
snadrus, line 29: How about 0.01 FIL? 
When it's at $60/FIL, it's going to be hard to meet 5 in all the different wallets.

curiosrc/alertmanager/task_alert.go
```diff
@@ -0,0 +1,395 @@
+package alertmanager
```
snadrus, line 1: + // Nobody associated with this software's development has any business relationship to pagerduty.
// This is provided as a convenient trampoline to SP's alert system of choice.

curiosrc/alertmanager/task_alert.go
```diff
@@ -0,0 +1,395 @@
...
+	client := &http.Client{}
+	var resp *http.Response
+
+	for i := 0; i < 5; i++ { // Maximum of 5 retries
+		resp, err = client.Do(req)
+		if err != nil {
+			time.Sleep(time.Duration(2*i) * time.Second) // Exponential backoff
+			continue
+		}
+		defer resp.Body.Close()
```
snadrus, line 194: defer func() {_= resp.Body.Close}()

curiosrc/alertmanager/task_alert.go
```diff
@@ -0,0 +1,395 @@
...
+		for _, tf := range taskFailures {
+			if tf.Failures > 10 {
+				al.alertMap[Name].alertString += fmt.Sprintf("Machine: %s, Failures: %d. ", tf.Machine, tf.Failures)
+			}
+		}
+	}
+	return
+}
+
+func storageCheck(al *alerts) {
```
snadrus, line 393: This one is also quite fuzzy. The best number would be "2 days at max fill speed" which we can't really know. But we can start with a few TB and adjust it as we go. 
We also may want to focus on storage that allows sectors. 
The sealing pipeline should probably get another alert related to dead sectors.

curiosrc/alertmanager/task_alert.go
```diff
@@ -0,0 +1,395 @@
...
+								GROUP BY completed_by_host_and_port
+								ORDER BY failed_tasks_count DESC;`, AlertMangerInterval.Minutes())
+	if err != nil {
+		al.alertMap[Name].err = xerrors.Errorf("getting failed task count for all machines: %w", err)
+		return
+	}
+
+	if len(taskFailures) > 0 {
+		for _, tf := range taskFailures {
+			if tf.Failures > 10 {
```
snadrus, line 385: I'm feeling like a  per-machine alert is nice, but we probably also want a per-task-type. 
Even one failed seal is worth knowing about.

curiosrc/alertmanager/task_alert.go
```diff
@@ -0,0 +1,395 @@
...
+	"github.com/filecoin-project/lotus/lib/harmony/harmonydb"
+	"github.com/filecoin-project/lotus/lib/harmony/harmonytask"
+	"github.com/filecoin-project/lotus/lib/harmony/resources"
+	"github.com/filecoin-project/lotus/node/config"
+	"github.com/filecoin-project/lotus/storage/ctladdr"
+)
+
+var MinimumBalance = types.MustParseFIL("5 FIL")
+
+const AlertMangerInterval = 10 * time.Minute
```
snadrus, line 31: This has been dropping. It started out at 1 hour. What's the logic here?
I'm concerned we may slice too thin &  miss out.

node/config/types.go
```diff
@@ -1109,3 +1110,13 @@ type FaultReporterConfig struct {
 	// rewards. This address should have adequate funds to cover gas fees.
 	ConsensusFaultReporterAddress string
 }
+
+type CurioAlerting struct {
+	// PagerDutyEventURL is URL for PagerDuty's Events API v2 URL. Events sent to this API URL are ultimately
+	// routed to a PagerDuty service and processed.
+	PagerDutyEventURL string
+
+	// PageDutyIntegrationKey is the integration key for a pager duty service. You can find this unique service
```
snadrus, line 1119: Please add dot-com URL because translation will eat this unrecognizable.

node/config/types.go
```diff
@@ -1109,3 +1110,13 @@ type FaultReporterConfig struct {
 	// rewards. This address should have adequate funds to cover gas fees.
 	ConsensusFaultReporterAddress string
 }
+
+type CurioAlerting struct {
+	// PagerDutyEventURL is URL for PagerDuty's Events API v2 URL. Events sent to this API URL are ultimately
+	// routed to a PagerDuty service and processed.
```
snadrus, line 1116: + // The default is sufficient for integration with the stock commercial Pager Duty company's service.

snadrus (changes requested):

curiosrc/alertmanager/task_alert.go
LexLuthr: I was actually thinking about making this a config param so users can adjust based on their setup. it cover what you described as scenarios where I am not sealing and just proving so I shouldn't need to keep sealing wallets full.

curiosrc/alertmanager/task_alert.go
LexLuthr: Yes. I will add this line.

curiosrc/alertmanager/task_alert.go
LexLuthr: I haven't though this one thorough yet. But, yes, plan was to start with only permanent storage path and not sealing paths. Sealing path are covered by reservations. Maybe start with 100 GiB free for each storage path to allow.

curiosrc/alertmanager/alerts.go
```diff
@@ -0,0 +1,288 @@
...
+	"github.com/filecoin-project/lotus/node/config"
+)
+
+// balanceCheck retrieves the machine details from the database and performs balance checks on unique addresses.
+// It populates the alert map with any errors encountered during the process and with any alerts related to low wallet balance and missing wallets.
+// The alert map key is "Balance Check".
+// It queries the database for the configuration of each layer and decodes it using the toml.Decode function.
+// It then iterates over the addresses in the configuration and curates a list of unique addresses.
+// If an address is not found in the chain node, it adds an alert to the alert map.
+// If the balance of an address is below 5 Fil, it adds an alert to the alert map.
```
snadrus, line 23: below the minimum configured balance (default 5 Fil)

curiosrc/alertmanager/alerts.go
```diff
@@ -0,0 +1,288 @@
...
+			totalRequiredSpace += int64(64 << 30)
+		}
+		totalRequiredSpace += int64(sectorSize)
+	}
+
+	var totalAvailableSpace int64
+	for _, storage := range storages {
+		totalAvailableSpace += storage.Available
+	}
+
```
snadrus, line 284: totalRequiredSpace += largestSectorSize * len(storages)
 // without this, bin packing differences and half-sector aggregations would make this thing never alert.

snadrus: That could be too much (especially if every hard drive is a storage). 
Your alternative would be to do bin-packing by remembering how many of each sector size, and burn-down the list with some arrangement.

curiosrc/alertmanager/task_alert.go
```diff
@@ -0,0 +1,220 @@
...
+	Summary       string      `json:"summary"`
+	Severity      string      `json:"severity"`
+	Source        string      `json:"source"`
+	Component     string      `json:"component,omitempty"`
+	Group         string      `json:"group,omitempty"`
+	Class         string      `json:"class,omitempty"`
+	CustomDetails interface{} `json:"custom_details,omitempty"`
+}
+
+type pdData struct {
```
snadrus, line 67: please just move this inside of sendAlert(). It doesn't go anywhere else so it may-as-well be local.

documentation/en/default-curio-config.toml
```diff
@@ -364,3 +364,38 @@
...
+  # type: string
+  #PageDutyIntegrationKey = ""
+
+  # MinimumWalletBalance is the minimum balance all active wallets. If the balance is below this value, an
+  # alerts will be triggered for the wallet
+  #
+  # type: types.FIL
+  #MinimumWalletBalance = "5 FIL"
+
+  [Alerting.PagerDutyEventURL]
```
snadrus, line 381: this should be a url string, not a section

node/config/def.go
```diff
@@ -328,6 +329,14 @@ const (
 	ResourceFilteringDisabled = ResourceFilteringStrategy("disabled")
 )
 
+func PGDutyURL() *url.URL {
+	ret, err := url.Parse("https://events.pagerduty.com/v2/enqueue")
```
snadrus, line 333: config.alerting.url ?

node/config/types.go
```diff
@@ -1109,3 +1112,18 @@ type FaultReporterConfig struct {
...
+	// The default is sufficient for integration with the stock commercial Pager Duty company's service.
+	PagerDutyEventURL *url.URL
```
snadrus, line 1120: string, not *url.URL

snadrus (changes requested):

curiosrc/alertmanager/task_alert.go
LexLuthr: Moved

curiosrc/alertmanager/alerts.go
LexLuthr: This was poor logic. I have updated the code block to account for each sector individually on storages. If all sectors are accounted for then no alerts but if we have a sector which could not be places anywhere, then an alert is created.

node/config/types.go
snadrus: change comment to:
This enables a pagerduty.com integration

curiosrc/alertmanager/alerts.go
```diff
@@ -0,0 +1,311 @@
...
+		Number abi.SectorNumber        `db:"sector_number"`
+		Proof  abi.RegisteredSealProof `db:"reg_seal_proof"`
+	}
+
+	var sectors []sector
+
+	err = al.db.Select(al.ctx, &sectors, `
+								SELECT sp_id, sector_number, reg_seal_proof
+								FROM sectors_sdr_pipeline
+								WHERE after_finalize = FALSE;`)
```
magik6k, line 266: Probably should be
```suggestion
								WHERE after_move_storage = FALSE;`)
```

magik6k (approved):
Looks really good. Just one nit.

Not in this PR, but would be neat to reuse the config parsing logic in `balanceCheck` and gather all miner-ids served by the cluster, then check that winningPoSt is happenning correctly based on entries being added to the `mining_tasks` table (we expect ~2880 entries in last 24h per minerID based on the base_compute_time column, the check should probably be for a smaller window tho)

LexLuthr:
It's alive!!!

Screenshot 2024-05-07 at 11 48 31 PM

Pushing the suggested changes and some minor changes in another commit.

curiosrc/alertmanager/alerts.go
LexLuthr: The idea was to generate an alert a little early but it doesn't really allow much time in reality. I will update this.

LexLuthr:
> Looks really good. Just one nit.
> 
> Not in this PR, but would be neat to reuse the config parsing logic in `balanceCheck` and gather all miner-ids served by the cluster, then check that winningPoSt is happenning correctly based on entries being added to the `mining_tasks` table (we expect ~2880 entries in last 24h per minerID based on the base_compute_time column, the check should probably be for a smaller window tho)

I think a better check would be to just check that all "win" instances resulted in a block which was accepted by the chain. This would help see if SPs are loosing blocks due to any issues. 

I have extracted the address function for future use.

snadrus (approved):

magik6k (approved):

magik6k:
> I think a better check would be to just check that all "win" instances resulted in a block which was accepted by the chain. This would help see if SPs are loosing blocks due to any issues.

That's another good one, but the former does tell you that you have some machines in the cluster taking care of winningPoSt at all times

Issue #11927, opened 2024-04-25
`lotus-miner sectors extend` unable to renew DC sectors


beck-8:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [X] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
v1.26.2
```

### Repro Steps

1. Run 'lotus-miner sectors extend --from 3903360 --to 4218480 --new-expiration 4227120 --really-do-it --max-sectors 1'
2. See error '...'
...

### Describe the Bug

```bash
root@hostname:~# lotus-miner sectors extend --from 3903360 --to 4218480 --new-expiration 4227120 --really-do-it --max-sectors 1
Extending 0 sectors: bafy2bzaceceyp6jcug4yooqkzrduaivdv6qey7b2if4sf5fsntwwsqb4xxxxx
Extending 2349 sectors: ERROR: mpool push message: GasEstimateMessageGas error: call ran out of gas
```

Problem 1: An empty structure message is generated with no parameters
Problem 2: Cannot be used normally after the second message (out of gas)

### Logging Information

```text
As above.
```

beck-8:
This problem will occur when `--max-sectors` is less than 2349.

beck-8:
Because the original batch operation unit is 2349, and it was CC before, the gas consumption is not that much, and the normal renewal is also more than 2349 sectors, so there is no problem. Now when renewing DC, this part is no longer satisfied, and the number needs to be less than 2349.

beck-8:
No, some of the logic was broken when it was refactored. It's not the batch problem I mentioned earlier, I will fix it later.

[...]

PR #11928, opened 2024-04-25
fix(cli): make lotus-miner sectors extend command resilient to higher gas

beck-8:
## Related Issues

https://github.com/filecoin-project/lotus/issues/11927

## Checklist

- [ ] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
   - default to ProveCommit aggregation
   - remove config options: AggregateCommits, AggregateAboveBaseFee, BatchPreCommitAboveBaseFee
 - feat(paych): add EnablePaymentChannelManager config option and disable payment channel manager by default ([filecoin-project/lotus#13139](https://github.com/filecoin-project/lotus/pull/13139))
+- fix: cli: lotus-miner sectors extend command ([filecoin-project/lotus#11927](https://github.com/filecoin-project/lotus/pull/11928))
 
 # Node v1.33.0 / 2025-05-08
 The Lotus v1.33.0 release introduces experimental v2 APIs with F3 awareness, featuring a new TipSet selection mechanism that significantly enhances how applications interact with the Filecoin blockchain. This release candidate also adds F3-aware Ethereum APIs via the /v2 endpoint.  All of the /v2 APIs implement intelligent fallback mechanisms between F3 and Expected Consensus and are exposed through the Lotus Gateway.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
   - default to ProveCommit aggregation
   - remove config options: AggregateCommits, AggregateAboveBaseFee, BatchPreCommitAboveBaseFee
 - feat(paych): add EnablePaymentChannelManager config option and disable payment channel manager by default ([filecoin-project/lotus#13139](https://github.com/filecoin-project/lotus/pull/13139))
-- fix: cli: lotus-miner sectors extend command ([filecoin-project/lotus#11927](https://github.com/filecoin-project/lotus/pull/11928))
+- fix(cli): make lotus-miner sectors extend command resilient to higher gas ([filecoin-project/lotus#11927](https://github.com/filecoin-project/lotus/pull/11928))
 
 # Node v1.33.0 / 2025-05-08
 The Lotus v1.33.0 release introduces experimental v2 APIs with F3 awareness, featuring a new TipSet selection mechanism that significantly enhances how applications interact with the Filecoin blockchain. This release candidate also adds F3-aware Ethereum APIs via the /v2 endpoint.  All of the /v2 APIs implement intelligent fallback mechanisms between F3 and Expected Consensus and are exposed through the Lotus Gateway.
--- a/cli/spcli/sectors.go
+++ b/cli/spcli/sectors.go
@@ -627,6 +627,7 @@ func SectorsExtendCmd(getActorAddress ActorAddressGetter) *cli.Command {
 			&cli.Int64Flag{
 				Name:  "max-sectors",
 				Usage: "the maximum number of sectors contained in each message",
+				Value: 500,
 			},
 			&cli.BoolFlag{
 				Name:  "really-do-it",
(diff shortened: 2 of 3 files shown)


beck-8 (commented):
This is still incorrect.

rjan90:
Thank you so much for the PR. I think this PR just needs a `make gen` (run the command while in your /lotus folder), to pass the final check.

cli/spcli/sectors.go
```diff
@@ -876,84 +877,96 @@ func SectorsExtendCmd(getActorAddress ActorAddressGetter) *cli.Command {
...
-					sectorsWithoutClaimsCount, err := sectorsWithoutClaimsToExtend.Count()
-					if err != nil {
-						return xerrors.Errorf("failed to count cc sectors: %w", err)
-					}
+						sectorsInDecl := int(sectorsWithoutClaimsCount) + len(sectorsWithClaims)
+						scount += sectorsInDecl
 
-					sectorsInDecl := int(sectorsWithoutClaimsCount) + len(sectorsWithClaims)
-					scount += sectorsInDecl
+						if scount > addrSectors || len(p.Extensions) >= declMax {
```
LexLuthr, line 955: I don't see a point for this check. If we are already batching and checking that batch size is smaller than policy max then what's the point of this check?

LexLuthr (approved):
Mostly looks good with some questions

cli/spcli/sectors.go
beck-8: I understand what you mean, this part is the original logic, so I didn’t change it.
The first half of this judgment is redundant, and the second half of the inspection may still be useful. I had no one to communicate with at the time, so I chose not to adjust it in a conservative strategy.

beck-8: Should I delete the redundant first half?

LexLuthr: Yes

cli/spcli/sectors.go
```diff
@@ -876,84 +877,97 @@ func SectorsExtendCmd(getActorAddress ActorAddressGetter) *cli.Command {
...
+						sectorsInDecl := int(sectorsWithoutClaimsCount) + len(sectorsWithClaims)
+						scount += sectorsInDecl
 
-					sectorsInDecl := int(sectorsWithoutClaimsCount) + len(sectorsWithClaims)
-					scount += sectorsInDecl
+						if len(p.Extensions) >= declMax {
+							params = append(params, p)
+							p = miner.ExtendSectorExpiration2Params{}
+							scount = sectorsInDecl
+						}
```
LexLuthr, line 960: I would suggest a complete refactor to get rid of this. If we are batching then we simply should determine the batch size correctly and apply the policies afterward. 

```suggestion
						if len(p.Extensions) >= declMax {
							params = append(params, p)
							p = miner.ExtendSectorExpiration2Params{}
							scount = sectorsInDecl
						}
```

cli/spcli/sectors.go
beck-8: No, I realized, there is nothing wrong here, it is completely wrong to remove the first half.

cli/spcli/sectors.go
beck-8: I can only help him get back on the right path, but I can't reconstruct it.

cli/spcli/sectors.go
beck-8: The function here is to determine whether a new parameter should be opened. The function of the previous batching is to reduce the input batch size.

beck-8: It's raining, 2300ml at a time
My barrel only has 500ml, and every time it rains 2300ml. My barrel is full, and there is still 1800ml (2300-500) left, and my other barrel cannot hold it.
So I divided the 2300ml each time into 500ml each time, so that my barrel can work normally (at this time, I also need to check whether the barrel is full)

BigLep:
2025-02-10 triage conversation: Venus is going to pick this up to get it over the line.

beck-8:
Is there any news here?

BigLep:
@beck-8 : thanks for the ping.

@tediou5 : has @filecoin-project/lotus-miner-maintainers looked at this?

tediou5 (approved):
@beck-8 Thanks a lot for your PR; it looks good to me. However, you might need to resolve the conflicts first.

Sorry I overlooked it earlier. I’ll push it forward and get it merged ASAP.

cli/spcli/sectors.go
```diff
@@ -879,90 +880,97 @@ func SectorsExtendCmd(getActorAddress ActorAddressGetter) *cli.Command {
...
+									claimExpiration := claim.TermStart + claim.TermMax
+									// can be maintained in the extended sector
+									if claimExpiration > newExp {
+										claimIdsToMaintain = append(claimIdsToMaintain, claimId)
+									} else {
+										sectorInfo, ok := activeSectorsInfo[sectorNumber]
+										if !ok {
+											return xerrors.Errorf("failed to find sector in active sector set: %w", err)
+										}
+										if !cctx.Bool("drop-claims") ||
```
tediou5: This approach certainly works. I'm just putting it up for discussion. While the code is cleaner, the error messages are less informative. What do you think?

beck-8: This PR has been put on hold for too long. I may have forgotten some information. Can you be more specific?

tediou5: Previously, the logic produced separate error logs for each error, but this has now been simplified into a single message. While the previous approach seemed rather redundant, I personally prefer the original method from the user’s perspective.

beck-8: Oh, I saw it.

tediou5 (approved):
Thank you so much for the PR!

Sorry this took so long. I’ll wait a bit in case others are interested, but it should be merged soon. Thanks again for your contribution!

beck-8:
Don't worry, the careful review is what we want to see.

tediou5:
@beck-8 By the way, with FIP-0100 implemented and the batch balancer removed, we probably don’t need to worry about that earlier issue anymore.

beck-8:
> @beck-8 By the way, with FIP-0100 implemented and the batch balancer removed, we probably don’t need to worry about that earlier issue anymore.

`out of gas` ?
Is this part certain? FIP100 only reduces the gas proportion of some users, but the block limit of the blockchain does not change. I understand that if I have too many sectors in one batch, there will still be a limit.

tediou5:
> but the block limit of the blockchain does not change. I understand that if I have too many sectors in one batch, there will still be a limit.

Yes, the block limit still exists, but since messages now contain only the basic gas and the rest has been shifted to a daily fee, a batch message is now almost impossible to hit the block limit.

I can go check the data to confirm. That said, I don’t see a problem with this PR; having it configurable is always good. We might just set a more reasonable default value.

cli/spcli/sectors.go
```diff
@@ -879,90 +880,103 @@ func SectorsExtendCmd(getActorAddress ActorAddressGetter) *cli.Command {
...
+						sectorsInDecl := int(sectorsWithoutClaimsCount) + len(sectorsWithClaims)
+						scount += sectorsInDecl
 
-					sectorsInDecl := int(sectorsWithoutClaimsCount) + len(sectorsWithClaims)
-					scount += sectorsInDecl
+						if scount > addrSectors || len(p.Extensions) >= policy.DeclarationsMax {
+							params = append(params, p)
+							p = miner.ExtendSectorExpiration2Params{}
+							scount = sectorsInDecl
+						}
```
tediou5: Since you already pre-group the messages, the logic here is somewhat redundant; we could process it directly when building the batch above. It still works as is, so this is just a suggestion. Perhaps we can submit a separate PR later to optimize this section of the logic.

tediou5 (approved):
Alright, my bad. I checked again, and while it’s lower, it’s still quite high. Setting it to 500 seems fairly reasonable.

Here are something that could be improved, but I don’t think we need to handle it in this PR.

cli/spcli/sectors.go
beck-8: I forgot what the reason for this part was. I looked back on the history and found that `scount > addrSectors` had been deleted before, but then this logic was added back. I think I must have tested a special logic? But I really forgot.

beck-8: I tested it, and there is no problem with the logic here. The logic I optimized is input, reducing the size of each input. This check logic must also exist. Here is to check the size of the bucket.

beck-8: https://github.com/filecoin-project/lotus/pull/11928#discussion_r1624524143

At present, I think there is no problem. But as they said, this part of the logic needs to be reconstructed, but I don't have enough experience to reconstruct it, so I can only make it run on the right track as much as possible.

tediou5: You’re absolutely right. As I mentioned earlier, it works fine as is, it just could be optimized.

This logic was originally meant to split oversized messages into groups.

However, since you now pre-group them earlier, `scount` always equals `addrSectors` when the first batch finishes, and after each subsequent batch is processed it’s appended to the newly created `p`.

Therefore, the conditional check here can be removed, you just need to create a new `p` at the start of each batch and append it once the batch is done.

beck-8: please check @tediou5

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
   - default to ProveCommit aggregation
   - remove config options: AggregateCommits, AggregateAboveBaseFee, BatchPreCommitAboveBaseFee
 - feat(paych): add EnablePaymentChannelManager config option and disable payment channel manager by default ([filecoin-project/lotus#13139](https://github.com/filecoin-project/lotus/pull/13139))
+- fix: cli: lotus-miner sectors extend command ([filecoin-project/lotus#11927](https://github.com/filecoin-project/lotus/pull/11928))
```
rvagg:
```suggestion
- fix(cli): make lotus-miner sectors extend command resilient to higher gas ([filecoin-project/lotus#11927](https://github.com/filecoin-project/lotus/pull/11928))
```

I'm trying to make this more descriptive in the squashed commit message and in CHANGELOG, does this sound right for what we're trying to achieve here @beck-8 @tediou5 ?

tediou5:
> /usr/bin/ld: warning: 81a71fbc30f7fcce-sha256_x64.o: missing .note.GNU-stack section implies executable stack
> /usr/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker

CI seems to be having issues, I’ve run into this a few times.

rvagg:
> I’ve run into this a few times

this has been my life for a while now .. Go compiler thing on some Linux variants; I'm blind to it these days but maybe I shouldn't be!

rvagg (approved):

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
   - default to ProveCommit aggregation
   - remove config options: AggregateCommits, AggregateAboveBaseFee, BatchPreCommitAboveBaseFee
 - feat(paych): add EnablePaymentChannelManager config option and disable payment channel manager by default ([filecoin-project/lotus#13139](https://github.com/filecoin-project/lotus/pull/13139))
```
eddieperez88, line 33: How

eddieperez88:
I can go check the data to confirm but looks great

PR #11930, opened 2024-04-25
feat: ETH compatibility in Filecoin : Support legacy Homestead Ethereum transactions and legacy EIP-155 transactions in Filecoin

### LINKED ISSUE #11859: Filecoin should support legacy Ethereum transactions
### Checklist

### Lotus component

- [X] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus JSON-RPC API
- [ ] Other

This issue has been created to add support for legacy transactions to Filecoin so that contracts and user transactions are not scoped to a specific chain. Specifically, this issue tracks FIP discussion at https://github.com/filecoin-project/FIPs/discussions/962.

Option #3 on https://github.com/filecoin-project/FIPs/discussions/962 

"We could also prepend a transaction type byte to the signature field (for Ethereum transactions only). IMO, this is the best solution, but it may break any tooling manually signing/validating messages from Ethereum accounts. On the other hand, I'm doubtful any tooling like this exists, so we're probably fine." 

_No response_

### Additional context

_No response_

aarshkshah1992:
## Related Issues
Closes #11859 

## Proposed Changes
- Adds support for legacy Homestead Ethereum transactions to Filecoin
- FIP is at https://github.com/filecoin-project/FIPs/pull/995
- FIP discussion at https://github.com/filecoin-project/FIPs/discussions/962
- Note that all ETH RPC APIs that return transaction info will exhibit the following behaviour for legacy transactions:
    - return nil for the `MaxFeePerGas` and `MaxPriorityFeePerGas`
    - `ChainID` will be 0x00
    - `TxType` will be 0x00
-  For EIP-1559 transactions, all ETH RPC APIs will:
    - return nil for `GasPrice`    

## Checklist

- [ ] CI is green


aarshkshah1992:
Should we just make all ETH RPC APIs return all three of `GasPrice`, `MaxFeePerGas ` and `MaxPriorityFeePerGas ` as non-nil ? For legacy transactions, we can just set the latter two to the value of `GasPrice`.

Users can always detect that a transaction is a legacy transaction if the `ChainID` and `TxType` fields are both set to 0.

chain/types/ethtypes/eth_legacy_homestead_transactions.go
```diff
@@ -0,0 +1,368 @@
...
+		Params:     mi.params,
+	}, nil
+}
+
+func (tx *EthLegacyHomesteadTxArgs) ToVerifiableSignature(sig []byte) ([]byte, error) {
+	if len(sig) != legacyHomesteadTxSignatureLen {
+		return nil, fmt.Errorf("signature should be %d bytes long, but got %d bytes", legacyHomesteadTxSignatureLen, len(sig))
+	}
+	if sig[0] != LegacyHomesteadEthTxSignaturePrefix {
+		return nil, fmt.Errorf("expected signature prefix 0x01, but got 0x%x", sig[0])
```
ZenGround0, line 92: nit: 0x01 => 0x%x and add LegacyHomesteadEthTxSignaturePrefix as argument

chain/types/ethtypes/eth_legacy_homestead_transactions.go
```diff
@@ -0,0 +1,368 @@
...
+		GasFeeCap:  tx.GasPrice,
+		GasPremium: tx.GasPrice,
+		Method:     mi.method,
+		Params:     mi.params,
+	}, nil
+}
+
+func (tx *EthLegacyHomesteadTxArgs) ToVerifiableSignature(sig []byte) ([]byte, error) {
+	if len(sig) != legacyHomesteadTxSignatureLen {
+		return nil, fmt.Errorf("signature should be %d bytes long, but got %d bytes", legacyHomesteadTxSignatureLen, len(sig))
```
ZenGround0, line 89: Not sure but might be instructive to add a little more information to avoid confusion with the literal length of eth homestead signatures being 1 byte less.
```suggestion
		return nil, fmt.Errorf("signature should be %d bytes long (1 byte metadata, %d bytes sig data), but got %d bytes", legacyHomesteadTxSignatureLen, legacyHomesteadTxSignatureLen - 1, len(sig))
```

ZenGround0:
> Should we just make all ETH RPC APIs return all three of GasPrice, MaxFeePerGas  and MaxPriorityFeePerGas  as non-nil ? For legacy transactions, we can just set the latter two to the value of GasPrice.

This sounds like a good approach to me.

chain/types/ethtypes/eth_transactions.go
```diff
@@ -36,189 +61,121 @@ type EthTx struct {
...
 
-	if len(msg.Params) > 0 {
-		paramsReader := bytes.NewReader(msg.Params)
-		params, err = cbg.ReadByteArray(paramsReader, uint64(len(msg.Params)))
-		if err != nil {
-			return EthTxArgs{}, xerrors.Errorf("failed to read params byte array: %w", err)
+	// Process based on the signature data length.
+	switch len(smsg.Signature.Data) {
+	case 66: // Legacy Homestead transaction
+		if smsg.Signature.Data[0] != LegacyHomesteadEthTxSignaturePrefix {
```
ZenGround0, line 129: Is there a reason this first byte value has to be specified?  Is it just keeping things clean?  I don't see a real reason we can't just throw it away without checking.

aarshkshah1992: @ZenGround0 
1. We do plan to use this extended signature length for supporting more tx types down the road if need be. For now, there is only type with this extended length and that is the legacy transaction type.So, probably does not hurt to just explicity sanity check here. Either is fine.

chain/types/ethtypes/eth_legacy_homestead_transactions.go
aarshkshah1992: Done.

chain/types/ethtypes/eth_legacy_homestead_transactions.go
aarshkshah1992: Done.

chain/messagesigner/messagesigner.go
```diff
@@ -196,7 +196,7 @@ func (ms *MessageSigner) dstoreKey(addr address.Address) datastore.Key {
 
 func SigningBytes(msg *types.Message, sigType address.Protocol) ([]byte, error) {
 	if sigType == address.Delegated {
-		txArgs, err := ethtypes.EthTxArgsFromUnsignedEthMessage(msg)
+		txArgs, err := ethtypes.Eth1559TxArgsFromUnsignedFilecoinMessage(msg)
```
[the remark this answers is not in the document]
aarshkshah1992: @ZenGround0 Yes, legacy txs should never hit this code path as we only ever expect Lotus to recieve already signed legacy txns. We don't want to create and sign legacy txns from within Lotus.

The API to send signed txns (legacy and new) to Lotus is the `EthSendRawTransaction` API and that only expects signed messages.

chain/types/ethtypes/eth_legacy_homestead_transactions_test.go
```diff
@@ -0,0 +1,258 @@
...
+			"0x7778cde41a8a37f6a087622b38bc201bd3e7df06dce067569d4def1b53dba98c",
+			"0x1b",
+			false,
+		},
+		{
+			"0x00",
+			"",
+			"",
+			"",
+			true,
```
ZenGround0, line 110: Could we get a test with v = 0x1c? 
Also a test with valid signature but v != 0x1c or 0x1b

itests/eth_legacy_transactions_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	client.EVM().SignLegacyTransaction(&tx, key.PrivateKey)
+	// Mangle signature
+	tx.V.Int.Xor(tx.V.Int, big.NewInt(1).Int)
+
+	signed, err := tx.ToRlpSignedMsg()
+	require.NoError(t, err)
+	// Submit transaction with bad signature
+	_, err = client.EVM().EthSendRawTransaction(ctx, signed)
+	require.Error(t, err)
```
ZenGround0, line 71: 🥇

ZenGround0: These tests are great

node/impl/full/eth_utils.go
```diff
@@ -251,7 +251,11 @@ func newEthBlockFromFilecoinTipSet(ctx context.Context, ts *types.TipSet, fullTx
 			return ethtypes.EthBlock{}, xerrors.Errorf("failed to convert msg to ethTx: %w", err)
 		}
 
-		tx.ChainID = ethtypes.EthUint64(build.Eip155ChainId)
+		// Don't override the chainID if this is not an EIP-1559 transaction.
+		// The chainID has already been set to 0 for leagcy ETH transactions.
```
ZenGround0, line 255: nit: legacy

node/impl/full/eth_utils.go
```diff
@@ -251,7 +251,11 @@ func newEthBlockFromFilecoinTipSet(ctx context.Context, ts *types.TipSet, fullTx
...
+		if smsg.Signature.Type != crypto.SigTypeDelegated {
```
ZenGround0, line 256: Im confused by this, won't the Signature.Type be delegated for both Eip155 txs and legacy txs?  Doesn't this code avoid setting the ChainID for Eip155 Txs as well as legacy?

node/impl/full/eth_utils.go
```diff
@@ -713,7 +720,18 @@ func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLook
 	}
 
 	baseFee := parentTs.Blocks()[0].ParentBaseFee
-	gasOutputs := vm.ComputeGasOutputs(lookup.Receipt.GasUsed, int64(tx.Gas), baseFee, big.Int(tx.MaxFeePerGas), big.Int(tx.MaxPriorityFeePerGas), true)
+
+	gasFeeCap, err := tx.GasFeeCap()
+	if err != nil {
+		return api.EthTxReceipt{}, xerrors.Errorf("failed to get gas fee cap: %w", err)
```
ZenGround0, line 726: Do we want to error here given a legacy transaction?  Not sure who calls this but if we expect legacy txs then it would probably be better to handle explicitly.  I think this might be what youre top level comment is referring to

ZenGround0 (approved):
I would really like someone to double check the new signature logic as the only way this change can go really bad is if we get that wrong.  But to me everything looks great.

chain/types/ethtypes/eth_transactions.go
```diff
@@ -21,8 +19,35 @@ import (
...
 )
 
-const Eip1559TxType = 2
+const (
+	// LegacyHomesteadEthTxSignaturePrefix defines the prefix byte used to identify the signature of a legacy Homestead Ethereum transaction.
+	LegacyHomesteadEthTxSignaturePrefix = 0x01
+	Eip1559TxType                       = 0x2
 
+	ethLegacyHomesteadTxType    = 0x0
+	ethLegacyHomesteadTxChainID = 0x0
```
rvagg, line 28:
```suggestion
	LegacyHomesteadEthTxSignaturePrefix = 0x01
	Eip1559TxType                       = 0x02

	ethLegacyHomesteadTxType    = 0x00
	ethLegacyHomesteadTxChainID = 0x00
```

or the other way, just for consistency

itests/eth_legacy_transactions_test.go
```diff
@@ -0,0 +1,302 @@
...
+	kit.SendFunds(ctx, t, client, deployer, types.FromFil(1000))
+
+	gasParams, err := json.Marshal(ethtypes.EthEstimateGasParams{Tx: ethtypes.EthCall{
+		From: &ethAddr,
+		Data: contract,
+	}})
+	require.NoError(t, err)
+
+	gaslimit, err := client.EthEstimateGas(ctx, gasParams)
+	require.NoError(t, err)
```
rvagg, line 50: I don't see `contract` or `gasParams` reused in this test below here, so is this step strictly necessary to get `gaslimit`? Can we simplify and remove the contract from this test to make it less confusing, or is the confusion unavoidable in what we need to do? Or, is `Data: contract` meant to be set on `EthLegacyHomesteadTxArgs`?

itests/eth_legacy_transactions_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	// install contract
+	contractHex, err := os.ReadFile("./contracts/SimpleCoin.hex")
+	require.NoError(t, err)
+
+	contract, err := hex.DecodeString(string(contractHex))
+	require.NoError(t, err)
```
rvagg, line 123: these could be moved in to `deployLegacyContractTx` since it's the same for both cases, no?

itests/eth_legacy_transactions_test.go
```diff
@@ -0,0 +1,302 @@
...
+	require.EqualValues(t, 0, ethTx.ChainID)
+	require.EqualValues(t, ethtypes.EthBytes{}, ethTx.Input)
+	require.EqualValues(t, tx.GasLimit, ethTx.Gas)
+	require.EqualValues(t, tx.GasPrice, *ethTx.GasPrice)
+	require.EqualValues(t, tx.R, ethTx.R)
+	require.EqualValues(t, tx.S, ethTx.S)
+	require.EqualValues(t, tx.V, ethTx.V)
+}
+
+func TestLegacyContractDeploymentValidSignature(t *testing.T) {
```
rvagg, line 109: this case seems to be duplicated in the next, what's the justification for having a separate one just for this if we're keeping the invalid sig in the next case?

chain/types/ethtypes/eth_1559_transactions.go
```diff
@@ -0,0 +1,384 @@
...
+		value,
+		tx.Input,
+		[]interface{}{}, // access list
+	}
+	return res, nil
+}
+
+func parseEip1559Tx(data []byte) (*Eth1559TxArgs, error) {
+	if data[0] != Eip1559TxType {
+		return nil, xerrors.Errorf("not an EIP-1559 transaction: first byte is not 2")
```
rvagg, line 265:
```suggestion
		return nil, xerrors.Errorf(fmt.Sprintf("not an EIP-1559 transaction: first byte is not %d", Eip1559TxType))
```

not important, just reducing magic

rvagg: or even `0x%x` perhaps

chain/types/ethtypes/eth_1559_transactions_test.go
```diff
@@ -0,0 +1,241 @@
...
+	signature := &crypto1.Signature{
+		Type: crypto1.SigTypeDelegated,
+		Data: sig,
+	}
+
+	err = sigs.Verify(signature, from, msg)
+	require.NoError(t, err)
+}
+
+func prepareTxTestcases() ([]TxTestcase, error) {
```
rvagg, line 211: any idea where these came from and if it would be easy to compile a similar list for legacy?

chain/types/ethtypes/eth_legacy_homestead_transactions.go
```diff
@@ -0,0 +1,369 @@
...
+	"github.com/filecoin-project/go-address"
+	gocrypto "github.com/filecoin-project/go-crypto"
+	"github.com/filecoin-project/go-state-types/big"
+	typescrypto "github.com/filecoin-project/go-state-types/crypto"
+
+	"github.com/filecoin-project/lotus/chain/types"
+)
+
+const (
+	legacyHomesteadTxSignatureLen = 66
```
rvagg, line 18: We could also have one for eip1559 too to avoid the `65` magic number in that file

chain/types/ethtypes/eth_transactions.go
```diff
@@ -21,8 +19,35 @@ import (
...
+)
+
+// EthTransaction defines the interface for Ethereum-like transactions.
+// It provides methods to convert transactions to various formats,
+// retrieve transaction details, and manipulate transaction signatures.
+type EthTransaction interface {
+	Sender() (address.Address, error)
+	Signature() (*typescrypto.Signature, error)
+	InitialiseSignature(sig typescrypto.Signature) error
```
rvagg, line 37: I do love a good english spelling code interface, I wonder if this on a public interface will cause problems for our American friends?

chain/types/ethtypes/eth_transactions.go
```diff
@@ -227,347 +184,90 @@ func (tx *EthTxArgs) ToSignedMessage() (*types.SignedMessage, error) {
...
-func (tx *EthTx) TxHash() (EthHash, error) {
-	ethTxArgs := tx.ToEthTxArgs()
-	return (&ethTxArgs).TxHash()
-}
 
-func (tx *EthTxArgs) TxHash() (EthHash, error) {
-	rlp, err := tx.ToRlpSignedMsg()
-	if err != nil {
-		return EmptyEthHash, err
+	if data[0] > 0x7f {
```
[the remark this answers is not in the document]
rvagg: This is found in 2 places in this diff so should probably be documented in both places. I think the best resource might be: https://eips.ethereum.org/EIPS/eip-2718#backwards-compatibility, and then specifically EIP-1559 builds on that by using `0x02` as its prefix @ https://eips.ethereum.org/EIPS/eip-1559#abstract

Also interesting is that `[0xc0, 0xfe]` is specifically called out as the logical range for legacy transactions. I don't know if we need to add that check because out of that range should either be picked up by bad sig or bad RLP encoding, but it's an option apparently.

chain/types/ethtypes/eth_transactions.go
```diff
@@ -36,189 +61,121 @@ type EthTx struct {
...
 	}
 
-	if len(msg.Params) > 0 {
-		paramsReader := bytes.NewReader(msg.Params)
-		params, err = cbg.ReadByteArray(paramsReader, uint64(len(msg.Params)))
-		if err != nil {
-			return EthTxArgs{}, xerrors.Errorf("failed to read params byte array: %w", err)
+	// Process based on the signature data length.
+	switch len(smsg.Signature.Data) {
+	case 66: // Legacy Homestead transaction
```
rvagg, line 128: these could both be replaced with consts which I think could be exposed publicly too, no reason to keep them hidden as long as we're also publicly exposing the signature data

chain/types/ethtypes/eth_transactions.go
```diff
@@ -227,347 +184,90 @@ func (tx *EthTxArgs) ToSignedMessage() (*types.SignedMessage, error) {
...
 	}
 
-	return EthHashFromTxBytes(rlp), nil
-}
-
-func (tx *EthTxArgs) ToRlpSignedMsg() ([]byte, error) {
-	packed1, err := tx.packTxFields()
-	if err != nil {
-		return nil, err
+	if data[0] == 1 {
```
rvagg, line 200: looks like this whole section could be a `switch`

rvagg:
This looks pretty good, as far as my knowledge of such things go. It would be nice to add a corpus of legacy transactions to decode for the tests like the existing eip1559 ones, but if we don't have a ready source of that I don't think spending time collecting them is going to add enough value for the effort.

It'd also be nice to have a label that we could mark PRs like this as "API breaking" to make sure we don't slip it into a semver-patch.

jennijuju:
@raulk @fridrik01  could you please help take a quick look on this PR by any chance?

chain/types/ethtypes/eth_transactions.go
aarshkshah1992: Eh should be fine. They can lookup the Oxford dictionary :)

itests/eth_legacy_transactions_test.go
aarshkshah1992: You are right, we don't need the contract at all here. I have fixed this test by estimating  the gas costs for doing a value transfer instead.

itests/eth_legacy_transactions_test.go
aarshkshah1992: Done.

itests/eth_legacy_transactions_test.go
aarshkshah1992: You are right.

I added the invalid signature case to the next test and removed this test.

chain/types/ethtypes/eth_legacy_homestead_transactions.go
aarshkshah1992: Removed this and added exported constants for both the txn types.

chain/types/ethtypes/eth_transactions.go
aarshkshah1992: Done.

chain/types/ethtypes/eth_transactions.go
aarshkshah1992: Done.

node/impl/full/eth_utils.go
aarshkshah1992: @ZenGround0 This wont error for legacy transactions as it will return the `GasPrice` instead of `MaxFeePerGas`.

```

func (tx *EthTx) GasFeeCap() (EthBigInt, error) {
	if tx.GasPrice == nil && tx.MaxFeePerGas == nil {
		return EthBigInt{}, fmt.Errorf("gas fee cap is not set")
	}
	if tx.MaxFeePerGas != nil {
		return *tx.MaxFeePerGas, nil
	}
	return *tx.GasPrice, nil
}
```

Yeah this is what I am talking about in my top level comment. Setting the `MaxFeePerGas` on legacy txns feels wrong and `go-ethereum` does not do it either.

I am just waiting for feedback from @raulk or someone else with ETH experience and am happy to change it. However, this is not a consensus breaking change and so can easily be changed in a feature release.

node/impl/full/eth_utils.go
aarshkshah1992: Actually we don't need this at all as `ChainID` is already set on tx in all cases by the `newEthTxFromSignedMessage` method.

node/impl/full/eth_utils.go
aarshkshah1992: removed this completely. See below.

chain/types/ethtypes/eth_legacy_homestead_transactions.go
```diff
@@ -0,0 +1,369 @@
...
+	Input    []byte      `json:"input"`
+	V        big.Int     `json:"v"`
+	R        big.Int     `json:"r"`
+	S        big.Int     `json:"s"`
+}
+
+func (tx *EthLegacyHomesteadTxArgs) ToEthTx(smsg *types.SignedMessage) (EthTx, error) {
+	from, err := EthAddressFromFilecoinAddress(smsg.Message.From)
+	if err != nil {
+		// This should be impossible as we've already asserted that we have an EthAddress
```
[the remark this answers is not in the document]
aarshkshah1992: Removed this comment from here as it was here from prior code but I would like to keep this sanity check here.

aarshkshah1992:
@rvagg @ZenGround0 Have addressed all your review comments so far. Testing this on a local node now.

rvagg (approved):
sgtm as far as my review is worth it, pending more functional testing; also a shame we don't have more legacy bytes to run through this but I understand the difficulty

aarshkshah1992:
Tested and works with Metamask Legacy Transactions

image

aarshkshah1992:
Works with Coinbase Wallet as well

image

aarshkshah1992:
We need EIP-155 transactions to enable support for Glif, Coinbase Wallet, Phantom Wallet etc with EVM based Filecoin transactions as they use EIP-155 transactions under the hood. This is a big unlock for the ecosystem:
The code difference with Homestead transactions entails two things:
a) Enabling chain replay protection by extracting the ChainID from the V value in the signature and making sure it matches the ChainID of the node (as you have pointed out above)

b) There is a difference in how we construct the digest for signature verification for EIP-155 transactions namely that we include the ChainID in the digest but not in the serialised transaction that is broadcast among nodes (as the digest can be reconstructed by extracting the ChainID from the V value of the signature)

aarshkshah1992:
@rvagg @ZenGround0 @rjan90 

Broke this down into two PRs.

1) https://github.com/filecoin-project/lotus/pull/11969 -> Enable support for legacy Homestead transactions in Filecoin. This PR contains the reviewed portion of this PR which has everything we need to enable legacy Homestead transactions.

2) https://github.com/filecoin-project/lotus/pull/11970 -> Implements EIP-155 transactions in Filecoin.

We will merge both of these to an integration and then merge the integration branch to master.

PR #11932, opened 2024-04-26
fix(events): check for sync-in-progress

rvagg:
We use `currentHeight` to figure out where we are to know whether we need to backfill a filter or not. Unfortunately, we collect this from `Apply()` so if you install a filter _prior_ to processing a message, we can't know where we are and what we should do about it.

The current behaviour is that it will fail the `minHeight < currentHeight` test further down in this file, not perform a pre-fill, and we end up with an empty filter (initially). So any calls to `GetActorEventsRaw` or `SubscribeActorEventsRaw` wanting backfill will not get any of it and will return a `null` to the user. But, this behaviour is the same as "there are no events in this range", so the user has no way of differentiating between "node is starting" and "no events".

This new behaviour will send an error in response to a `GetActorEventsRaw` or `SubscribeActorEventsRaw` or even `eth_getLogs` if the node is still starting and we haven't seen our first `Apply`.

An alternative to this behaviour is to go ask the chain what the current head is, but I don't think we currently have the means to do this in `EventFilterManager`, we only have a `ChainStore` to work with.


rvagg:
Derp, I looked and couldn't see an obvious way to get it; I didn't look long enough! Thanks @ribasushi.

Updated the PR to update `m.currentHeight` instead of erroring if we haven't recorded it yet.

Issue #11909, opened 2024-04-20
Index change proposals for the events sqlite database


rvagg:
Looking at the schema and indexes as they are on master now, and ignoring ongoing work to modify the database for the actor id work, combined with the main `SELECT` pattern we do, our indexes are out of whack and aren't doing what we need. I propose modifications, needing yet another migration of course.

## Current setup & use

Current indexes on master:

```sql
CREATE INDEX IF NOT EXISTS height_tipset_key ON event (height,tipset_key)
CREATE INDEX IF NOT EXISTS event_emitter_addr ON event (emitter_addr)
CREATE INDEX IF NOT EXISTS event_entry_key_index ON event_entry (key)
```

The big `SELECT` in `prefillFilter` @ https://github.com/filecoin-project/lotus/blob/1a789d3acbed6a6c89ab6d92a454218b293acf0c/chain/events/filter/index.go#L540-L617 does the following:

* Optional `tipset_key_cid` match (mutually exclusive with `height` match)
* Optional `height` match (mutually exclusive with `tipset_key_cid` match)
* Optional `reverted` match
* Optional bulk `OR` match on `emitter_addr`
* Optional matching of the `keysWithCodec` structure, which involves `key`, `codec` and `value`. When provided, sets up a `JOIN` between `event` and `event_entry` on `event_id`.
  * Always match `event_entry.indexed=1` and the `event_entry.key` for each `keysWithCodec`
  * For each `keysWithCodec` value, match both `event_entry.value` and `event_entry.codec` with a bulk `OR`.

## Proposal

1. `height_tipset_key` is not very useful, we never match on `tipset_key`: `DROP INDEX IF EXISTS height_tipset_key;`
2. Index `tipset_key_cid` by itself: `CREATE INDEX IF NOT EXISTS event_tipset_key_cid ON event (tipset_key_cid);`
3. Index `height` by itself: `CREATE INDEX IF NOT EXISTS event_height ON event (height);`
4. Index `reverted` by itself: `CREATE INDEX IF NOT EXISTS event_reverted ON event (reverted);`
5. Index `indexed` and `key` as a pair: `CREATE INDEX IF NOT EXISTS event_entry_indexed_key ON event_entry (indexed, key);`
6. Because `key` is now indexed along with `indexed`, we drop the stand-alone `key` index introduced in https://github.com/filecoin-project/lotus/pull/11477: `DROP INDEX IF EXISTS event_entry_key_index;`
[...]

PR #11934, opened 2024-04-26
feat(events): adjust indexes in event index db to match query patterns

rvagg:
Introduces a v4 migration that just adjusts indexes.

Copies some improvements from https://github.com/filecoin-project/lotus/pull/11723

Closes: https://github.com/filecoin-project/lotus/issues/11909

/cc @i-norden

### DIFF
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -26,7 +26,7 @@ var pragmas = []string{
 	"PRAGMA temp_store = memory",
 	"PRAGMA mmap_size = 30000000000",
 	"PRAGMA page_size = 32768",
-	"PRAGMA auto_vacuum = NONE",
+	"PRAGMA auto_vacuum = NONE", // not useful until we implement GC
 	"PRAGMA automatic_index = OFF",
 	"PRAGMA journal_mode = WAL",
 	"PRAGMA read_uncommitted = ON",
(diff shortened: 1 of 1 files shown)


chain/events/filter/index.go
```diff
@@ -266,20 +274,7 @@ func (ei *EventIndex) migrateToVersion2(ctx context.Context, chainStore *store.C
 		return xerrors.Errorf("commit transaction: %w", err)
 	}
 
-	// during the migration, we have likely increased the WAL size a lot, so lets do some
```
rvagg, line 269: I removed the vacuum entirely from this migration, leaving it only in the last migration so we only do one rewrite at the very end

masih (approved):

rvagg:
Updated my calibnet node with `master`:

```
{"level":"info","ts":"2024-04-29T14:02:24.696+1000","logger":"filter","caller":"filter/index.go:432","msg":"Upgrading event index from version 2 to version 3"}
{"level":"info","ts":"2024-04-29T14:02:25.046+1000","logger":"filter","caller":"filter/index.go:313","msg":"Successfully migrated event index from version 2 to version 3 in 349.827948ms"}
{"level":"info","ts":"2024-04-29T14:02:25.046+1000","logger":"filter","caller":"filter/index.go:442","msg":"Upgrading event index from version 3 to version 4"}
{"level":"info","ts":"2024-04-29T14:02:25.046+1000","logger":"filter","caller":"filter/index.go:374","msg":"Performing DB vacuum and wal checkpointing to free up space after the migration"}
{"level":"info","ts":"2024-04-29T14:02:32.881+1000","logger":"filter","caller":"filter/index.go:366","msg":"Successfully migrated events to version 3 in 7.835114071s"}
```

fixing that last message in https://github.com/filecoin-project/lotus/pull/11937, my bad in this PR, and I was being so careful making everything consistent!

PR #11945, opened 2024-04-30
fix(events): load actor from tipset state, not parent state

rvagg:
PR against https://github.com/filecoin-project/lotus/pull/11723 to load state of current tipset instead of parent.


node/impl/full/eth_events.go
```diff
@@ -118,35 +118,43 @@ func ethFilterResultFromEvents(ctx context.Context, evs []*filter.CollectedEvent
...
-		actor, err := sa.StateGetActor(ctx, emitterAddr, ev.TipSetKey)
+		ts, err := sa.Chain.GetTipSetFromKey(ctx, ev.TipSetKey)
+		if err != nil {
+			return nil, xerrors.Errorf("loading tipset: %w", err)
+		}
+		stateRoot, _, err := sa.StateManager.TipSetState(ctx, ts)
+		if err != nil {
+			return nil, xerrors.Errorf("loading tipset state: %w", err)
+		}
+		actor, err := sa.StateManager.LoadActorRaw(ctx, emitterAddr, stateRoot)
```
rvagg, line 129: this is the relevant section that I'm testing out

The problem with `sa.StateGetActor(tsk)` is that it it loads `ParentState` but the event may arise from a tipset where the actor id was first introduced, so we fail with an "unknown actor" error. But if we fast-forward by one then it's there.

rvagg:
see comment in https://github.com/filecoin-project/lotus/pull/11905/files#r1585930424 regarding `hasExpensiveForkBetween`; do we want to try and handle that here? Maybe error if you call this when there's an expensive migration to be run on this tipset.

PR #11950, opened 2024-04-30
fix: curio: node UI & darwin gpu count

snadrus:
## Related Issues

Fixed Node page:
image

Also added hardware to cluster machines for relate-ability.
image

Reduced the GPU count for Darwin devices just-in-case someone connected to a Macbook to a bigger system (say, for GUI support) so the Macbook does not get all the network's tasks. 

## Checklist

- [ ] CI is green


LexLuthr (approved):
I think we are missing a `go mod tidy`. Master already have this.

extern/filecoin-ffi
magik6k, line 1: Committed by accident

magik6k (changes requested):
Need to undo ffi commit

extern/filecoin-ffi
magik6k:
```
git checkout master
git pull
make clean deps
git checkout thisBranch
git commit --ammend --no-edit -- extern/filecoin-ffi
```

snadrus: I didn't want a confusion in the history when merged back to master as future changes could experience a merge conflict, so since it was the only thing in the most recent commit, I forced the branch to the previous commit.

magik6k:
The linter is sad
```
Error: curiosrc/web/api/config/config.go:47:10: Error return value of `w.Write` is not checked (errcheck)
		w.Write([]byte("Layer already exists"))
```

magik6k (commented):
Other than the linter thing, Looks Good

curiosrc/web/hapi/simpleinfo.go
```diff
@@ -580,7 +584,7 @@ func (a *app) clusterMachineSummary(ctx context.Context) ([]machineSummary, erro
 	}
 
 	// Then machine summary
-	rows, err := a.db.Query(ctx, "SELECT id, host_and_port, last_contact FROM harmony_machines order by host_and_port asc")
+	rows, err := a.db.Query(ctx, "SELECT id, host_and_port, CURRENT_TIMESTAMP - last_contact, cpu, ram, gpu AS last_contact FROM harmony_machines order by host_and_port asc")
```
magik6k, line 587: Nit: doesn't really matter due to how we select out of this, but technically should be

```suggestion
	rows, err := a.db.Query(ctx, "SELECT id, host_and_port, CURRENT_TIMESTAMP - last_contact AS last_contact, cpu, ram, gpu FROM harmony_machines order by host_and_port asc")
```

storage/paths/local.go
```diff
@@ -462,7 +462,7 @@ func (st *Local) reportStorage(ctx context.Context) {
 	}
 }
 
-func (st *Local) Reserve(ctx context.Context, sid storiface.SectorRef, ft storiface.SectorFileType, storageIDs storiface.SectorPaths, overheadTab map[storiface.SectorFileType]int, minFreePercentage float64) (userRelease func(), err error) {
+func (st *Local) Reserve(ctx context.Context, sid storiface.SectorRef, ft storiface.SectorFileType, storageIDs storiface.SectorPaths, overheadTab map[storiface.SectorFileType]int, minFreePercentage float64) (release func(), err error) {
```
magik6k, line 465: This is very wrong, I'd either not name those returns or would keep the userRelease name.

If we name this `return` and change the `release := func() {}` line below to `=` you'll get panics on error - because the func will return `nil, err..`, then the defer in L478 will run a nil `release` (which was just set to nil in the return..)

magik6k (changes requested):
The local.go change is wrong, other than that looks good now

magik6k (approved):

Issue #11630, opened 2024-02-19
Values of `logIndex` field don't conform to geth


ilyalukyanov:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
N/A
```

### Repro Steps

1. Call:
```bash
curl --silent --location 'https://api.node.glif.io/rpc/v1' --header 'Content-Type: application/json' --data '{
    "jsonrpc": "2.0",
    "id": 5,
    "method": "eth_getLogs",
    "params": [
        {
            
            "fromBlock": "0x37e517",
            "toBlock": "0x37e517"
            
        }
    ]
}'
```
for any recent tipset that has multiple transactions at least two of which emitted logs. To reproduce the issue, transactions may or may not need to be in different Filecoin blocks - I didn't narrow that down further.
2. Multiple returned items (logs) will have e.g. `logIndex: 0`
   

### Describe the Bug

Filecoin interpretation of `logIndex` is inconsistent with geth. In geth `logIndex` should be unique within a block. It represents a sequential index of a log within a block (https://github.com/ethereum/go-ethereum/blob/034bc4669ffe92b95155c8331334f47fa8bb4333/core/types/log.go#L49). Thus, the expected behaviour is that `logIndex` should be a unique incremental number for each log within a tipset.

For example, this request:
```bash
curl --silent --location 'https://api.node.glif.io/rpc/v1' --header 'Content-Type: application/json' --data '{
    "jsonrpc": "2.0",
    "id": 2,
    "method": "eth_getLogs",
    "params": [
        {
[...]

PR #11952, opened 2024-05-01
Event index should be unique for tipsets

aarshkshah1992:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/11630

## Proposed Changes
The event index for logs should be unique for all events in the same tipset to confirm with geth behaviour.

## Checklist

- [ ] CI is green


chain/events/filter/index.go
```diff
@@ -356,6 +356,16 @@ func (ei *EventIndex) migrateToVersion4(ctx context.Context) error {
 		}
 	}
 
+	stmtEventIndexUpdate, err := tx.PrepareContext(ctx, "UPDATE event SET event_index = (SELECT COUNT(*) FROM event e2 WHERE e2.tipset_key_cid = event.tipset_key_cid AND e2.id <= event.id) - 1")
```
rvagg, line 359: I'm a bit worried about bundling this into the same migration. I'd be inclined to just introduce a v5 and stack them up just to be safe for anyone running on `master`. If we always move `ei.vacuumDBAndCheckpointWAL(ctx)` to the very last migration so we only do it once then the cost is pretty low.

Maybe this one isn't so critical but we've found from experience that people run on `master` and we don't have a good way of dealing with these migrations.

rvagg (commented):
👌 yep, looks good, except for the migration concern, will take that discussion to Slack

chain/events/filter/index.go
aarshkshah1992: @rvagg I agree that migrating to version 5 makes more sense here.

rvagg (approved):
Approving without running it. If this works on calibnet and you manually inspect the DB and see the migration worked then all good. The rest looks good.

aarshkshah1992:
Tested this on calibnet

```
SELECT a.event_index, b.event_index FROM event a INNER JOIN event b ON a.tipset_key_cid = b.tipset_key_cid WHERE a.id != b.id LIMIT 120;
```

This query returns rows with duplicate event_indexes for master but does NOT do so for the current branch and returns sequential event_indices instead

```
4|0
4|1
4|2
4|3
4|5
4|6
4|7
4|8
4|9
4|10
4|11
4|12
4|13
4|14
4|15
5|0
5|1
5|2
5|3
5|4
5|6
5|7
5|8
5|9
5|10
5|11
5|12
5|13
5|14
5|15
6|0
6|1
6|2
6|3
6|4
6|5
6|7
6|8
```
......

Issue #11962, opened 2024-05-06
Flags field is dropped in SectorOnChainInfo conversions


hanabi1224:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
lotus version 1.27.0-dev+calibnet+git.10f7b6ed5
```

### Repro Steps

The `Flags` field is not converted in the code at https://github.com/filecoin-project/lotus/blob/master/chain/actors/builtin/miner/v12.go#L533
```go
func fromV12SectorOnChainInfo(v12 miner12.SectorOnChainInfo) SectorOnChainInfo {
	info := SectorOnChainInfo{
		SectorNumber:          v12.SectorNumber,
		SealProof:             v12.SealProof,
		SealedCID:             v12.SealedCID,
		DealIDs:               v12.DealIDs,
		Activation:            v12.Activation,
		Expiration:            v12.Expiration,
		DealWeight:            v12.DealWeight,
		VerifiedDealWeight:    v12.VerifiedDealWeight,
		InitialPledge:         v12.InitialPledge,
		ExpectedDayReward:     v12.ExpectedDayReward,
		ExpectedStoragePledge: v12.ExpectedStoragePledge,

		SectorKeyCID: v12.SectorKeyCID,

		PowerBaseEpoch:    v12.PowerBaseEpoch,
		ReplacedDayReward: v12.ReplacedDayReward,
	}
	return info
}
```

### Describe the Bug

When posting the below payload to `http://localhost:1234/rpc/v0`
```json
[...]

PR #11963, opened 2024-05-06
fix: copy Flags field from SectorOnChainInfo

rvagg:
Fixes: https://github.com/filecoin-project/lotus/issues/11962

~~not sure if v12 is worth updating here, included just in case—is it worth it?~~ whoops, codegen, I've done it for all possible, which is v12 and v13.

### DIFF
--- a/chain/actors/builtin/miner/state.go.template
+++ b/chain/actors/builtin/miner/state.go.template
@@ -587,6 +587,7 @@ func fromV{{.v}}SectorOnChainInfo(v{{.v}} miner{{.v}}.SectorOnChainInfo) SectorO
 		{{if (ge .v 12)}}
 		PowerBaseEpoch:        v{{.v}}.PowerBaseEpoch,
 		ReplacedDayReward:     v{{.v}}.ReplacedDayReward,
+		Flags:                 SectorOnChainInfoFlags(v{{.v}}.Flags),
                 {{end}}
 	}
 	return info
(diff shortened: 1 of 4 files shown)


chain/actors/builtin/account/account.go
```diff
@@ -1,25 +1,33 @@
...
-	actorstypes "github.com/filecoin-project/go-state-types/actors"
-	builtin13 "github.com/filecoin-project/go-state-types/builtin"
 	"github.com/filecoin-project/go-state-types/cbor"
+
 	"github.com/filecoin-project/go-state-types/manifest"
+	"github.com/filecoin-project/lotus/chain/actors/adt"
+	"github.com/filecoin-project/lotus/chain/types"
+
 	builtin0 "github.com/filecoin-project/specs-actors/actors/builtin"
+
```
ZenGround0, line 17: I'd prefer if we didn't do this imports related change throughout.

ZenGround0: I don't think it's standard for go imports and it's adding a lot of noise.  Is this your editor maybe?

rvagg: oh, yeah, that is odd .. not my editor because I didn't even open most of these files, I only touched chain/actors/builtin/miner/state.go.template; something fishy though! will Draft this and work out what's up

rvagg: oh, right, I know what it was - `make actors-code-gen` was a silly mistake, that should just be internal to `make gen` or `make actors-gen` because it doesn't include `scripts/fiximports` that would fix this.

rvagg:
What am I not seeing in this lint job output? https://github.com/filecoin-project/lotus/actions/runs/8968065046/job/24630013957 I've re-run it but it's failing in the same way but I can't see it saying anything about the code.

ZenGround0 (approved):

rvagg:
linter problems fixed @ #11968

rvagg:
... and actors v14 gen added

PR #11944, opened 2024-04-30
fix: curio: gui listen conflict

strahe:
## Related Issues

#11941 
## Proposed Changes

1. Change the default value of `Subsystems.GuiAddress` from `:4701` to `127.0.0.1:4701`.
2. Add a `--gui-listen` flag to `curio run` and keep it in sync with `curio web`.
## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] If the PR affects users (e.g., new feature, bug fix, system requirements change), update the CHANGELOG.md and add details to the UNRELEASED section.
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

Issue #11941, opened 2024-04-30
Curio web and worker listening ports are confused


strahe:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### Lotus Version

```text
curio version 1.27.0-dev+calibnet+git.10877d2e6
```

### Repro Steps

1. Run '...'
2. Do '...'
3. See error '...'
...

### Describe the Bug

1. The default listening address for curio web is ":4701", which is interpreted as "localhost:4701", but the actual listening address is "0.0.0.0:4701", which does not match the actual intention.
2. When running curio run --layers "gui,xxx", it is not possible to customize the guiAddress. When running curio web --layers "xxx" --listen "xxx", the --listen option is applied to the worker rpc instead of the web.

### Logging Information

```text
2024-04-30T03:39:59.797Z        WARN    harmonytask     resources/resources.go:134      curio's defaults are for running alone. Use task maximums or CGroups.
2024-04-30T03:40:17.194Z        INFO    harmonytask     resources/getGPU.go:25  GPUs    {"list": []}
2024-04-30T03:40:17.236Z        INFO    harmonytask     resources/resources.go:90       Cleaned up machines     {"count": 0}
2024-04-30T03:40:17.248Z        INFO    curio/rpc       rpc/rpc.go:272  Setting up RPC server at 127.0.0.1:12300
2024-04-30T03:40:17.251Z        INFO    curio/chainsched        chainsched/chain_sched.go:69    restarting chain scheduler
[...]

PR #11965, opened 2024-05-06
fix: curio: gui listen conflict

strahe:
## Related Issues

https://github.com/filecoin-project/lotus/issues/11941
Previous PR： https://github.com/filecoin-project/lotus/pull/11944
## Proposed Changes

1. Change the default value of `Subsystems.GuiAddress` from `:4701` to `127.0.0.1:4701`.
2. Rename the `--listen` parameter of `curio web` to `--gui-listen`, making it common and clear in both `curio web` and `curio run`.

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/curio/deps/deps.go
+++ b/cmd/curio/deps/deps.go
@@ -223,6 +223,9 @@ func (deps *Deps) PopulateRemainingDeps(ctx context.Context, cctx *cli.Context, 
 		if err != nil {
 			return xerrors.Errorf("populate config: %w", err)
 		}
+		if cctx.IsSet("gui-listen") {
+			deps.Cfg.Subsystems.GuiAddress = cctx.String("gui-listen")
+		}
 	}
 
 	log.Debugw("config", "config", deps.Cfg)
--- a/cmd/curio/run.go
+++ b/cmd/curio/run.go
@@ -37,6 +37,12 @@ var runCmd = &cli.Command{
 			Value:   "0.0.0.0:12300",
 			EnvVars: []string{"LOTUS_WORKER_LISTEN"},
 		},
+		&cli.StringFlag{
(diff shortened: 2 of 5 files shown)


snadrus:
The most common use is (we presume) to stand up a server that is likely to be used for long-term monitoring.
You can set  config.Subsystems.GuiAddress by editing this on base, then `curio run --layers=gui`. This edit is even possible on the GUI.

strahe:
I am more inclined to write all configurations into the database, except for the listening address. This is because the listening address is a one-time parameter when running the service, and it cannot be modified while the service is running. Therefore, writing it into the database or modifying it in the database will not take effect immediately. A better approach would be to specify it as a parameter when running the service, as is currently done.

LexLuthr:
I was thinking about this from user perspective and @strahe is correct. This is confusing at best. Our current system allows for a weird setup where depending upon the layer selected, we might not get a full picture of the cluster in GUI. Perhaps, we should rethink this command a little. I would suggest the following:
1. Remove GUI from the run command. (ignore even if the layer is provided)
2. `web` subcommand should ignore all the other subsystems and then run the GUI server(not use run action). We can add a `listen-gui` flag here and allow users to override the port supplied from the DB. This allows users to not care about the port number resulting from the different stacking possibilities. 
3. Since, UI doesn't really require any resources, running more than one Curio on the same node should not present a challenge. The only exception is the market-connect module in Curio. We can move that inside the startTask to allow user flexibility about starting market module only wherever they desire allowing dedicated piecePark nodes for large clusters.

This approach is less sophisticated but does allow for a more user friendly op in my opinion. WDYT?

snadrus:
Thanks. I've adapted it into https://github.com/filecoin-project/lotus/pull/12013

Issue #11967, opened 2024-05-06
Deprecated Linters Causing failed checks


rjan90:
During the execution of [golangci-lint](https://github.com/filecoin-project/lotus/actions/runs/8969938200/job/24632349545) when submitting a PR, several linters are reported as deprecated and fully inactivated, causing the process to fail. It suggest that we:

- Replace `varcheck` with `unused`.
- Replace `deadcode` with `unused`.
- Replace `scopelint` with `exportloopref`.
- Replace `golint` with `revive`.

rvagg:
Some context, latest update turned deprecation of these into an error, see section 5 in https://golangci-lint.run/product/changelog/#1580

varcheck and deadcode recommended replacement with unused here: https://golangci-lint.run/product/changelog/#v1490

discussion on scopelint: https://github.com/golangci/golangci-lint/issues/1041 - exportloopref and looppointer come up here, but looppointer was rejected @ https://github.com/golangci/golangci-lint/pull/1924

revive mentioned as drop-in replacement to golint: https://golangci-lint.run/usage/linters/#revive

unfortunately revive is quite a bit pickier than golint and throws up a bunch of errors when implementing, so we'll have to fix up a bunch of problems

rvagg:
https://github.com/filecoin-project/lotus/pull/11968 an attempt at what it might take .. a lot, without global exclusions

PR #11968, opened 2024-05-07
chore: lint: update linter settings, fix lint errors

rvagg:
Ref: https://github.com/filecoin-project/lotus/issues/11967

This isn't complete, I just banged through a few iterations of it to see the kinds of changes we're being asked to make. I think a better approach here is to add some exclusion error regexes. Most of the issues fall into two categories:

1. Named but unused parameters
2. Acronym capitalisation - Api -> API, Http -> HTTP, Url -> URL, Id -> ID

The latter suck in particular because a lot of them are on API methods. I've started adding explicit exclusions for those in here, but we could opt for a global exclusion (minimal might be something like `var-naming: var [A-Z].+ should be ` to capture public variables).

Or we could just hammer it with exclusions to minimise the diff in this PR in particular, even turn off `var-naming` entirely.

### DIFF
--- a/chain/checkpoint.go
+++ b/chain/checkpoint.go
@@ -25,15 +25,15 @@ func (syncer *Syncer) SyncCheckpoint(ctx context.Context, tsk types.TipSetKey) e
 	}
 
 	hts := syncer.ChainStore().GetHeaviestTipSet()
-	if hts.Equals(ts) {
-		// Current head, no need to switch.
-	} else if anc, err := syncer.store.IsAncestorOf(ctx, ts, hts); err != nil {
-		return xerrors.Errorf("failed to walk the chain when checkpointing: %w", err)
-	} else if anc {
-		// New checkpoint is on the current chain, we definitely have the tipsets.
-	} else if err := syncer.collectChain(ctx, ts, hts, true); err != nil {
-		return xerrors.Errorf("failed to collect chain for checkpoint: %w", err)
-	}
+	if !hts.Equals(ts) {
+		if anc, err := syncer.store.IsAncestorOf(ctx, ts, hts); err != nil {
+			return xerrors.Errorf("failed to walk the chain when checkpointing: %w", err)
+		} else if !anc {
+			if err := syncer.collectChain(ctx, ts, hts, true); err != nil {
+				return xerrors.Errorf("failed to collect chain for checkpoint: %w", err)
+			}
+		} // else new checkpoint is on the current chain, we definitely have the tipsets.
+	} // else current head, no need to switch.
 
 	if err := syncer.ChainStore().SetCheckpoint(ctx, ts); err != nil {
 		return xerrors.Errorf("failed to set the chain checkpoint: %w", err)
--- a/itests/kit/client.go
+++ b/itests/kit/client.go
@@ -141,7 +141,10 @@ func createRandomFile(rseed, size int) ([]byte, string, error) {
 		size = 1600
 	}
 	data := make([]byte, size)
-	_, _ = rand.New(rand.NewSource(int64(rseed))).Read(data)
+	_, err := rand.New(rand.NewSource(int64(rseed))).Read(data)
+	if err != nil {
+		return nil, "", err
+	}
 
 	dir, err := os.MkdirTemp(os.TempDir(), "test-make-deal-")
 	if err != nil {
--- a/itests/kit/client.go
+++ b/itests/kit/client.go
@@ -141,7 +141,7 @@ func createRandomFile(rseed, size int) ([]byte, string, error) {
 		size = 1600
 	}
 	data := make([]byte, size)
-	rand.New(rand.NewSource(int64(rseed))).Read(data)
+	_, _ = rand.New(rand.NewSource(int64(rseed))).Read(data)
 
 	dir, err := os.MkdirTemp(os.TempDir(), "test-make-deal-")
 	if err != nil {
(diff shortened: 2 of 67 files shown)


storage/paths/db_index.go
```diff
@@ -49,8 +49,8 @@ func NewDBIndex(al *alerting.Alerting, db *harmonydb.DB) *DBIndex {
 func (dbi *DBIndex) StorageList(ctx context.Context) (map[storiface.ID][]storiface.Decl, error) {
 
 	var sectorEntries []struct {
-		StorageId      string
-		MinerId        sql.NullInt64
+		StorageID      string
+		MinerID        sql.NullInt64
```
magik6k, line 53: Are you sure this won't mess with pgx.Select?

rvagg:
I think I agree, I was getting slightly angry at it telling me that `HttpServer` _must_ be `HTTPServer`. I'll work on some exclusions, maybe turn off `var-naming` entirely.

ZenGround0:
>  but we could opt for a global exclusion (minimal might be something like var-naming: var [A-Z].+ should be  to capture public variables).

I like the sound of a global exclusion for this

snadrus:
I'm alright turning off both of these things. Because these are not actual problems, but style matters

rvagg:
I re-did this PR from scratch, and revisited the whole config. Here's the choices I've made:

* Removed the deprecated linters `golint`, `varcheck`, `deadcode` and `scopelint`, replacing with their equivalents `exportloopref` and `revive`
* Removed `golint` specific path exclusions
* Remove `goconst` - it had two separate exclusions that meant it wasn't being used anyway, I think we're experienced enough to not need this
* All message exclusions are more explicit regexes now, they're harder to find ways around (harder to subvert) and it's much clearer in the config the exclusion decisions we're making
* I decided to ditch all exclusions to unchecked errors except for 3 explicit ones in node/repo/fsrepo.go and 2 in paychmgr/simple.go. This mostly required adding `_ =` in place of the explicit exclusions and `logger.SetLogLevel` had a bunch of unchecked but we have far more `_ =` in the repo so I'm just making that consistent.
* A bunch of minor changes that are obvious in the diff
* I decided not to exclude most of the new `empty-block` errors, but instead I ended up rewriting some logic. I get why some people like empty blocks for comments, but I don't believe they ultimately aid in clarity. **Reviewers please pay careful attention to my changes to make sure I didn't change any logic; some are non-trivial.**

magik6k (commented):
Only looked at the SP side of things, but looks good to me

chain/checkpoint.go
```diff
@@ -25,15 +25,15 @@ func (syncer *Syncer) SyncCheckpoint(ctx context.Context, tsk types.TipSetKey) e
...
-		// Current head, no need to switch.
-	} else if anc, err := syncer.store.IsAncestorOf(ctx, ts, hts); err != nil {
-		return xerrors.Errorf("failed to walk the chain when checkpointing: %w", err)
-	} else if anc {
-		// New checkpoint is on the current chain, we definitely have the tipsets.
-	} else if err := syncer.collectChain(ctx, ts, hts, true); err != nil {
-		return xerrors.Errorf("failed to collect chain for checkpoint: %w", err)
-	}
+	if !hts.Equals(ts) {
+		if anc, err := syncer.store.IsAncestorOf(ctx, ts, hts); err != nil {
```
ZenGround0, line 29: Might just be me but I find this style a lot more readable.

itests/kit/client.go
```diff
@@ -141,7 +141,7 @@ func createRandomFile(rseed, size int) ([]byte, string, error) {
 		size = 1600
 	}
 	data := make([]byte, size)
-	rand.New(rand.NewSource(int64(rseed))).Read(data)
+	_, _ = rand.New(rand.NewSource(int64(rseed))).Read(data)
```
ZenGround0, line 144: Maybe propagate this error ?

markets/dagstore/miner_api.go
```diff
@@ -201,7 +201,7 @@ func (m *minerAPI) GetUnpaddedCARSize(ctx context.Context, pieceCid cid.Cid) (ui
 		return 0, xerrors.Errorf("no storage deals found for piece %s", pieceCid)
 	}
 
-	len := pieceInfo.Deals[0].Length
+	l := pieceInfo.Deals[0].Length
 
-	return uint64(len), nil
+	return uint64(l), nil
```
ZenGround0, line 206: Maybe remove the intermediate `l` and just `return uint64(pieceInfo.Deals[0].Length)`

markets/dagstore/miner_api_test.go
```diff
@@ -129,9 +129,9 @@ func TestLotusAccessorGetUnpaddedCARSize(t *testing.T) {
 
 	// Check that the data length is correct
 	//stm: @MARKET_DAGSTORE_GET_UNPADDED_CAR_SIZE_001
-	len, err := api.GetUnpaddedCARSize(ctx, cid1)
+	l, err := api.GetUnpaddedCARSize(ctx, cid1)
```
ZenGround0, line 132: Maybe `size` `l` is just too close to `1` for me

ZenGround0 (approved):

node/impl/full/eth_utils.go
```diff
@@ -410,7 +411,6 @@ func lookupEthAddress(addr address.Address, st *state.StateTree) (ethtypes.EthAd
 		return ethAddr, nil
 	}
 
-	// Otherwise, use the masked address.
```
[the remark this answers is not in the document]
rvagg: good catch, I'd rather not lose it, https://github.com/filecoin-project/lotus/pull/12007

PR #11969, opened 2024-05-07
feat: ETH compatibility in Filecoin : Support Homestead and EIP-155 Ethereum transactions("legacy" transactions) in Filecoin

aarshkshah1992:
## Related Issues

For https://github.com/filecoin-project/lotus/issues/11859

## Proposed Changes
FIP discussion at https://github.com/filecoin-project/FIPs/discussions/962. 
FIP proposal at  https://github.com/filecoin-project/FIPs/pull/995

Implement Ethereum Homestead transactions and EIP-155 transactions(both are referred to as "legacy" transactions) in Filecoin.

## Checklist

- [ ] CI is green


ZenGround0 (approved):

chain/types/ethtypes/eth_legacy_155_transactions.go
```diff
@@ -0,0 +1,291 @@
...
+var _ EthTransaction = (*EthLegacy155TxArgs)(nil)
+
+// EthLegacy155TxArgs is a legacy Ethereum transaction that uses the EIP-155 chain replay protection mechanism
+// by incorporating the chainId in the signature.
+// See how the `V` value in the signature is derived from the chainId at
+// https://github.com/ethereum/go-ethereum/blob/86a1f0c39494c8f5caddf6bd9fbddd4bdfa944fd/core/types/transaction_signing.go#L424
+// For EthLegacy155TxArgs, the digest that is used to create a signed transaction includes the `ChainID` but the serialised RLP transaction
+// does not include the `ChainID` as an explicit field. Instead, the `ChainID` is included in the V value of the signature as mentioned above.
+type EthLegacy155TxArgs struct {
+	LegacyTx *EthLegacyHomesteadTxArgs
```
arajasek, line 23: As written, this is a guarantee to cause bugs in the future when people accidentally call the "inner" methods on `EthLegacyHomesteadTxArgs` instead of the correct methods (on `EthLegacy155TxArgs`). Please rewrite this, you could:

- create an inner struct that's just the fields (no logic), and then have two separate structs that contain that struct
- just duplicate the fields in `EthLegacy155TxArgs`, it's fine to do so
- if you really want to keep the existing pattern, make `LegacyTx` private, and rewrite external usages of it to access it properly (eg. by creating a constructor)

arajasek (changes requested):
Some notes:
- The new logic for the two types of Homestead transactions seems correct, assuming the FIP is correctly specified. I'll stare at it more in second review, but everything seems to do what it should.
- I'm not too strongly in favour of this change, it's adding a lot more edgecase-handling to what is already one of the gnarliest parts of the Filecoin protocol. I see the strong support for it in the FIP discussion, and this is about as lightweight an approach as can be expected for this change, so I'm happy to disagree-and-commit. But I do think this is a step in the wrong direction.
  - In particular, I don't think we should lean into an attitude of "[well, it's just in the client, so it's okay](https://github.com/filecoin-project/FIPs/discussions/962#discussioncomment-8813680)". It's still very much sharp edges in the protocol, and the fact that it's in the client only means that it needs to be (potentially incorrectly) re-implemented by Forest and Venus too.
- Directly related to the previous point, I unfortunately don't think this PR is safe to land as-is, because I strongly suspect it unintentionally breaks existing behaviour (in subtle ways). In general, it seems like this PR will tighten conditions on Eth1559 transactions. I don't think there's anything WRONG with the changes, but they'll still break consensus on nv22 (so, if anything, we'll have to put them behind a network version check).
  - As an example (the only one I can see so far, though there might be more), I'm pretty sure we can have messages with ID addresses as the sender carrying a Delegated signature today. I think that behaviour will change with this PR (it'll become invalid).

ZenGround0:
> But I do think this is a step in the wrong direction.

When you say this are you referring to the FIP change itself? 

> As an example (the only one I can see so far, though there might be more), I'm pretty sure we can have messages with ID addresses as the sender carrying a Delegated signature today. I think that behaviour will change with this PR (it'll become invalid).

Can you explain this more.  My understanding is that EIP1559 Txs won't have any behavior changes at all with this change.  How does this PR/FIP impact ID address sender of EIP1559 Tx?

chain/consensus/signatures.go
```diff
@@ -18,34 +18,41 @@ import (
...
+	signatureCopy := msg.Signature
 
-	typ := msg.Signature.Type
-	switch typ {
+	switch signatureType {
 	case crypto.SigTypeDelegated:
-		txArgs, err := ethtypes.EthTxArgsFromUnsignedEthMessage(&msg.Message)
+		signatureCopy.Data = make([]byte, len(msg.Signature.Data))
+		copy(signatureCopy.Data, msg.Signature.Data)
+		ethTx, err := ethtypes.EthTransactionFromSignedFilecoinMessage(msg)
```
ZenGround0, line 28: Coming back to this again I'm having some paranoid thoughts.  Does this PR open up forking as written?  
We release this code before nv23, someone comes in with a legacy Tx, sends it over to an SP with updated software.  That SP includes in a block.  Now an SP who is not updated yet reads the block.  It has this weird signature with one extra byte.  It doesn't validate per its code.  SP rejects block.

I could be off base and missing some subtlty that prevents this.  Just want to think it over with you before merge.

ZenGround0 (changes requested):

chain/consensus/signatures.go
snissn: I agree with @arajasek and @ZenGround0 that this PR would definitely create consensus issues if some nodes are able to execute a transaction and other nodes would reject it. Was there any discussion in the FIP process around how to handle the consensus breaking aspects of supporting a new transaction type?

aarshkshah1992: @snissn @ZenGround0 I have update this PR to only accept legacy txs if the node is on NV23. If the Node still hasn't upgraded to NV23, it will fail to accept legacy txns. 

I have also added an itest for it. It's in test `TestLegacyEIP155ValueTransferValidSignatureFailsNV22`. I think this should suffice. Let me know if you have any other concerns.

chain/types/ethtypes/eth_legacy_155_transactions.go
aarshkshah1992: Fixed this by making it private.

chain/types/ethtypes/eth_transactions.go
```diff
@@ -227,443 +212,386 @@ func (tx *EthTxArgs) ToSignedMessage() (*types.SignedMessage, error) {
...
+		}
+	} else if msg.Method == builtintypes.MethodsEVM.InvokeContract {
+		addr, err := EthAddressFromFilecoinAddress(msg.To)
+		if err != nil {
+			return nil, nil, err
+		}
+		to = &addr
+	} else {
+		return nil, nil,
+			fmt.Errorf("invalid methodnum %d: only allowed method is InvokeContract(%d)",
```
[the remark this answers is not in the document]
aarshkshah1992: Fixed.

aarshkshah1992:
@arajasek @ZenGround0 

- I agree that this FIP needs to be implemented by both Forrest and Venus but that should not be a reason to not ship a protocol change that has been requested by multiple users. We need to build for what they need.

**_- I have added network version gating to this and so nodes will not accept legacy txs if they are not on NV23._**

```
In general, it seems like this PR will tighten conditions on Eth1559 transactions. I don't think there's anything wrong with the changes, but they'll still break consensus on nv22 (so, if anything, we'll have to put them behind a network version check).
As an example (the only one I can see so far, though there might be more), I'm pretty sure we can have messages with ID addresses as the sender carrying a Delegated signature today. I think that behaviour will change with this PR (it'll become invalid).
```

This is the part that I do not understand. Why will messages with ID addresses and a delegated signature become invalid with this PR ? I

aarshkshah1992:
So tested this with CB Wallet on devnet after implementing Network Version Gating and looks good now. Transactions fail for NV22(when devnet epoch < NV23 epoch) but work for NV23(when devnet epoch >= NV23 epoch) !

Screenshot 2024-05-23 at 6 22 37 PM
Screenshot 2024-05-23 at 6 32 12 PM

chain/messagepool/messagepool.go
```diff
@@ -824,7 +824,16 @@ func (mp *MessagePool) VerifyMsgSig(m *types.SignedMessage) error {
 		return nil
 	}
 
-	if err := consensus.AuthenticateMessage(m, m.Message.From); err != nil {
+	mp.curTsLk.RLock()
+	curTs := mp.curTs
+	mp.curTsLk.RUnlock()
+	epoch := curTs.Height() + 1
```
ZenGround0, line 828: I'm a little worried about how this works for null epochs during network upgrade.

ZenGround0: Actually no Im not

chain/messagepool/messagepool.go
```diff
@@ -824,7 +824,16 @@ func (mp *MessagePool) VerifyMsgSig(m *types.SignedMessage) error {
 		return nil
 	}
 
-	if err := consensus.AuthenticateMessage(m, m.Message.From); err != nil {
+	mp.curTsLk.RLock()
```
snissn, line 825: from a thread safety perspective this block looks ok to me

arajasek:
> > But I do think this is a step in the wrong direction.
> 
> When you say this are you referring to the FIP change itself?

Yes.

> 
> > As an example (the only one I can see so far, though there might be more), I'm pretty sure we can have messages with ID addresses as the sender carrying a Delegated signature today. I think that behaviour will change with this PR (it'll become invalid).
> 
> Can you explain this more. My understanding is that EIP1559 Txs won't have any behavior changes at all with this change. How does this PR/FIP impact ID address sender of EIP1559 Tx?

EIP-1559 Txs _shouldn't_ have any behaviour changes with this FIP. I'm concerned that this PR _will_ cause some changes (example above) due to the refactor, which is consensus-breaking.

chain/consensus/signatures.go
```diff
@@ -16,36 +16,51 @@ import (
...
 		if err != nil {
-			return xerrors.Errorf("failed to reconstruct eth transaction: %w", err)
+			return xerrors.Errorf("failed to reconstruct Ethereum transaction: %w", err)
 		}
-		roundTripMsg, err := txArgs.ToUnsignedMessage(msg.Message.From)
+
+		if ethTx.Type() != ethtypes.EIP1559TxType {
```
arajasek, line 33: nit: use && instead of nested ifs

arajasek:
> @arajasek @ZenGround0
> 
>     * I agree that this FIP needs to be implemented by both Forrest and Venus but that should not be a reason to not ship a protocol change that has been requested by multiple users. We need to build for what they need.

The objection isn't that 3 implementations have to ship the change, the objection is that this is further messing up what is already a pretty janky part of the Filecoin protocol.

> This is the part that I do not understand. Why will messages with ID addresses and a delegated signature become invalid with this PR ? I

Because of the refactor. You're now calling `EthTransactionFromSignedFilecoinMessage` instead of `EthTxArgsFromUnsignedEthMessage` as part of signature authentication, which is asserting that the sender is an f4 address.

aarshkshah1992:
@arajasek 
> > This is the part that I do not understand. Why will messages with ID addresses and a delegated signature become invalid with this PR ? I
> 
> Because of the refactor. You're now calling `EthTransactionFromSignedFilecoinMessage` instead of `EthTxArgsFromUnsignedEthMessage` as part of signature authentication, which is asserting that the sender is an f4 address.

The same behaviour is also enforced  in master for EIP-1559 transactions. There is no behaviour change for EIP-1559 txns in this PR.

go.mod
```diff
@@ -141,6 +141,7 @@ require (
 	github.com/samber/lo v1.39.0
 	github.com/stretchr/testify v1.9.0
 	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
+	github.com/test-go/testify v1.1.4
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

chain/consensus/signatures.go
```diff
@@ -16,36 +16,51 @@ import (
 // SignedMessage was signed by the indicated Address, computing the correct
 // signature payload depending on the signature type. The supplied Address type
 // must be recognized by the registered verifier for the signature type.
-func AuthenticateMessage(msg *types.SignedMessage, signer address.Address) error {
+func AuthenticateMessage(msg *types.SignedMessage, signer address.Address, nv network.Version) error {
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien I have addressed this.

chain/consensus/signatures.go
aarshkshah1992: Removed this.

aarshkshah1992:
@Stebalien Pls can you approve this if lgty ?

arajasek:
> @arajasek
> 
> > Because of the refactor. You're now calling `EthTransactionFromSignedFilecoinMessage` instead of `EthTxArgsFromUnsignedEthMessage` as part of signature authentication, which is asserting that the sender is an f4 address.
> 
> The same behaviour is also enforced in master for EIP-1559 transactions. There is no behaviour change for EIP-1559 txns in this PR. See
> 
> https://github.com/filecoin-project/lotus/blob/18795706288ec76292988116a8b3f0978c7290de/chain/types/ethtypes/eth_transactions.go#L68
> in master.

Please read the code and my message more carefully. We are calling `EthTxArgsFromUnsignedEthMessage` on `master`, which does not assert that the sender is an f4 address. AFAIC tell nothing currently running on mainnet does. 

With this refactor, we will start asserting that the sender of EIP-1559 transactions is an f4 address. If I'm right about all of this, that would be a consensus break. I may not be right, but we need to be sure of that, and this PR cannot land until we are.

aarshkshah1992:
Oh hey @arajasek 

I see what you mean. We do NOT in fact allow EIP-1559 messages where sender is a non-ETH address.

See: https://github.com/filecoin-project/lotus/blob/18795706288ec76292988116a8b3f0978c7290de/lib/sigs/delegated/init.go#L65

https://github.com/filecoin-project/lotus/blob/18795706288ec76292988116a8b3f0978c7290de/lib/sigs/delegated/init.go#L70

Basically, the sender of a delegated signature type MUST be an f4 address.

aarshkshah1992:
Oh hey @arajasek

I see what you mean. We do NOT in fact allow EIP-1559 messages where sender is a non-ETH(f4) address.

See:

[lotus/lib/sigs/delegated/init.go](https://github.com/filecoin-project/lotus/blob/18795706288ec76292988116a8b3f0978c7290de/lib/sigs/delegated/init.go#L65)

Line 65 in [1879570](https://github.com/filecoin-project/lotus/commit/18795706288ec76292988116a8b3f0978c7290de)

 maybeaddr, err := address.NewDelegatedAddress(builtin.EthereumAddressManagerActorID, addrHash[12:]) 
[lotus/lib/sigs/delegated/init.go](https://github.com/filecoin-project/lotus/blob/18795706288ec76292988116a8b3f0978c7290de/lib/sigs/delegated/init.go#L70)

Line 70 in [1879570](https://github.com/filecoin-project/lotus/commit/18795706288ec76292988116a8b3f0978c7290de)

 if maybeaddr != a { 
Basically, the sender of a delegated signature type MUST be an f4 address.

arajasek:
> Oh hey @arajasek
> 
> I see what you mean. We do NOT in fact allow EIP-1559 messages where sender is a non-ETH(f4) address and fail these transactions during signature verification.
> 
> See:
> 
> https://github.com/filecoin-project/lotus/blob/18795706288ec76292988116a8b3f0978c7290de/lib/sigs/delegated/init.go#L65
> 
> Basically, the sender of a delegated signature type **_MUST_** be an f4 address for the signature verification to succeed.

@aarshkshah1992 Thanks for digging further! I’m not quite convinced yet -- the line you linked to extracts the address from the _signature_ carried by the message (which can obviously never be an ID address). But what happens if the _sender field_ of a message is an ID address? That case will definitely not be supported (invalid message) _after_ this refactor, because of the additional checks in `EthTransactionFromSignedFilecoinMessage`. But is that possible on mainnet today? If so, this refactor is unsafe.

aarshkshah1992:
Oh hey @arajasek 

- The line immediately below that at https://github.com/filecoin-project/lotus/blob/cf8d187817a661278e149ac3ca78c2144906a323/lib/sigs/delegated/init.go#L70 ensures that the address passed down to the `Verify` function matches the address extracted from the signature and the address extracted from the signature has to be an ID address as per https://github.com/filecoin-project/lotus/blob/cf8d187817a661278e149ac3ca78c2144906a323/lib/sigs/delegated/init.go#L65

- You can see that the sender of the message is indeed passed as an argument to the signature Verify function at https://github.com/filecoin-project/lotus/blob/cf8d187817a661278e149ac3ca78c2144906a323/chain/consensus/signatures.go#L47. That is the sender/signer of the message which is then matched to the address extracted from the signature.

arajasek:
> The line immediately below that at https://github.com/filecoin-project/lotus/blob/cf8d187817a661278e149ac3ca78c2144906a323/lib/sigs/delegated/init.go#L70
>        ensures that the address passed down to the `Verify` function matches the address extracted from the signature and the address extracted from the signature **has to be an ID address** as per https://github.com/filecoin-project/lotus/blob/cf8d187817a661278e149ac3ca78c2144906a323/lib/sigs/delegated/init.go#L65

I assume you mean has to be a delegated (f4) address here?
 
>  You can see that the sender of the message is indeed passed as an argument to the signature Verify function at https://github.com/filecoin-project/lotus/blob/cf8d187817a661278e149ac3ca78c2144906a323/chain/consensus/signatures.go#L47

Is the _sender_ of the message being passed? I see the _signer_ being passed as an argument here, but going up one more level it looks like that has already been resolved to a deterministic address. AFAIC tell this confirms my concern?

arajasek:
To break this down:

- Using `master` today, can I send a message carrying an f4 signature on mainnet whose `From` field is an ID address? 
  - Yes / No / I don't know.
- With this refactor, can I send a(n EIP-1559) message carrying an f4 signature on mainnet whose `From` field is an ID address?
  - Yes / No / I don't know.

If the answer to either question is `I don't know`, we need to figure out the actual answer (with confidence). If the answer to both questions isn't the same (both Yes or both No), we have a consensus split caused by this change, and so it isn't safe to land.

(I think the answers are Yes and No respectively, which would be a problem if true, but that's what needs to be confirmed).

chain/types/ethtypes/eth_transactions.go
```diff
@@ -113,10 +113,6 @@ func EthTransactionFromSignedFilecoinMessage(smsg *types.SignedMessage) (EthTran
 	if smsg == nil {
 		return nil, errors.New("signed message is nil")
 	}
-	// Validate the sender's address format.
```
aarshkshah1992: Mpool validation/Block validation already enforces that the sender should either be an f4 address or an ID address that can be resolved to an f4 address. No harm in allowing legacy txns to have ID addresses as sender.

aarshkshah1992:
- So based on https://github.com/filecoin-project/lotus/pull/11969#issuecomment-2138155588, I wrote some tests on master
- The test is at https://github.com/filecoin-project/lotus/pull/12061 (has some new code in mpool validation to get it to work)
- You can't get EIP-1559 messages with non-ID addresses through the Lotus mpool verification implementation -> it has to have an f4 address for mpool inclusion
- However, if an incoming block has an EIP-1559 transaction with the sender as an ID address (albeit one that can be resolved to a f4 address) -> it will go through and you are right here :+1:
- So, if a Miner decides to mine a block using their own custom mpool implementation that does not use/copy the Lotus mpool implementation -> EIP-1559 transactions with ID senders will indeed go through on master but fail on our branch with the legacy TX implementation (tried that as well)
- So yeah, this is indeed a consensus breaking change for the protocol :+1:
- I've fixed  this in the latest [commit](https://github.com/filecoin-project/lotus/pull/11969/commits/270d8ea2d96702603dd3194bbe27aa4e4f7633dc) on the PR  -> basically have completely removed the "sender should be an f4 address" check. If the sender is an ID address that can not be resolved to an f4 address, signature verification will anyways fail. And we want to keep the legacy tx behaviour with the EIP-1559 tx behaviour.

arajasek:
Thanks a lot for looking into this in detail! I'll give it a second review today, but it _should_ be good to go.

chain/types/ethtypes/eth_transactions.go
arajasek: Consider making this explicit in the FIP, I don't think it currently is.

arajasek (approved):

rvagg (approved):
:ship:

ZenGround0 (approved):

Stebalien:
The network-version checks look correct to me.

I would, however, recommend making them a bit more generic (can also happen in a followup so we can get this merged). Basically, combine the `IsValidSecpkSigType`, the "is delegate signature", and the eth check into a single `IsValidSignedMessageForSending` check.

itests/kit/evm.go
```diff
@@ -44,6 +44,49 @@ func (f *TestFullNode) EVM() *EVM {
...
+	}
+
+	signature.Data = append(signature.Data[:65], vVal.Int.Bytes()...)
+
+	err = tx.InitialiseSignature(*signature)
+	require.NoError(e.t, err)
+}
+
+// SignLegacyHomesteadTransaction signs a legacy Homstead Ethereum transaction in place with the supplied private key.
+func (e *EVM) SignLegacyHomesteadTransaction(tx *ethtypes.EthLegacyHomesteadTxArgs, privKey []byte) {
```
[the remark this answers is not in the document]
aarshkshah1992: Will address this in a follow up.

chain/types/ethtypes/eth_transactions.go
aarshkshah1992: Filed at https://github.com/filecoin-project/FIPs/pull/1019.

PR #11976, opened 2024-05-08
feat:state:Ignore market balance after nv23

### LINKED ISSUE #11921: Implement FIP 0065
### Checklist

### Lotus component

- [ ] lotus JSON-RPC API
- [ ] Other

FIP 0065 ignoring f05's contribution to locked FIL is accepted.

Implement the FIP

_No response_

### Additional context

_No response_

ZenGround0:
## Related Issues

Closes #11921 
## Proposed Changes

Drop builtin market calculation from locked balance calculation after nv23

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/stmgr/supply.go
+++ b/chain/stmgr/supply.go
@@ -308,6 +308,7 @@ func GetFilLocked(ctx context.Context, st *state.StateTree, nv network.Version) 
 	if nv >= network.Version23 {
 		return getFilPowerLocked(ctx, st)
 	}
+
 	filMarketLocked, err := getFilMarketLocked(ctx, st)
 	if err != nil {
 		return big.Zero(), xerrors.Errorf("failed to get filMarketLocked: %w", err)
@@ -391,6 +392,8 @@ func (sm *StateManager) GetVMCirculatingSupplyDetailed(ctx context.Context, heig
 func (sm *StateManager) GetCirculatingSupply(ctx context.Context, height abi.ChainEpoch, st *state.StateTree) (abi.TokenAmount, error) {
 	circ := big.Zero()
 	unCirc := big.Zero()
+	nv := sm.GetNetworkVersion(ctx, height)
+
 	err := st.ForEach(func(a address.Address, actor *types.Actor) error {
 		// this can be a lengthy operation, we need to cancel early when
 		// the context is cancelled to avoid resource exhaustion
@@ -419,18 +422,22 @@ func (sm *StateManager) GetCirculatingSupply(ctx context.Context, height abi.Cha
 			unCirc = big.Add(unCirc, actor.Balance)
 
 		case a == market.Address:
-			mst, err := market.Load(sm.cs.ActorStore(ctx), actor)
-			if err != nil {
-				return err
-			}
+			if nv >= network.Version23 {
+				circ = big.Add(circ, actor.Balance)
+			} else {
+				mst, err := market.Load(sm.cs.ActorStore(ctx), actor)
+				if err != nil {
+					return err
+				}
 
-			lb, err := mst.TotalLocked()
-			if err != nil {
-				return err
-			}
+				lb, err := mst.TotalLocked()
+				if err != nil {
+					return err
+				}
 
-			circ = big.Add(circ, big.Sub(actor.Balance, lb))
-			unCirc = big.Add(unCirc, lb)
+				circ = big.Add(circ, big.Sub(actor.Balance, lb))
+				unCirc = big.Add(unCirc, lb)
+			}
 
 		case builtin.IsAccountActor(actor.Code) ||
 			builtin.IsPaymentChannelActor(actor.Code) ||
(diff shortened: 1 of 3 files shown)


chain/stmgr/supply.go
```diff
@@ -303,11 +303,14 @@ func getFilPowerLocked(ctx context.Context, st *state.StateTree) (abi.TokenAmoun
...
 }
 
-func GetFilLocked(ctx context.Context, st *state.StateTree) (abi.TokenAmount, error) {
-
-	filMarketLocked, err := getFilMarketLocked(ctx, st)
-	if err != nil {
-		return big.Zero(), xerrors.Errorf("failed to get filMarketLocked: %w", err)
+func GetFilLocked(ctx context.Context, st *state.StateTree, height abi.ChainEpoch) (abi.TokenAmount, error) {
+	filMarketLocked := big.Zero()
+	if height <= build.UpgradeAussieHeight {
```
rvagg, line 308: how about we switch this around with a view to the future, where the common case is `height > build.UpgradeAussieHeight` and comes first, where we just do a simple one-line `return getFilPowerLocked(ctx, st)`, then we don't need `filMarketLocked=0`.

chain/stmgr/supply.go
```diff
@@ -303,11 +303,14 @@ func getFilPowerLocked(ctx context.Context, st *state.StateTree) (abi.TokenAmoun
 	return pst.TotalLocked()
 }
 
-func GetFilLocked(ctx context.Context, st *state.StateTree) (abi.TokenAmount, error) {
-
-	filMarketLocked, err := getFilMarketLocked(ctx, st)
-	if err != nil {
-		return big.Zero(), xerrors.Errorf("failed to get filMarketLocked: %w", err)
+func GetFilLocked(ctx context.Context, st *state.StateTree, height abi.ChainEpoch) (abi.TokenAmount, error) {
```
arajasek, line 306: Please refactor this to receive a network version instead, and make the change based on that. I know we have a lot of height comparisons in this area of the code, but that's quite bad and we shouldn't be adding any more.

The `StateManager` can lookup network versions based on the provided height and supply it here.

arajasek (changes requested):

itests/supply_test.go
```diff
@@ -0,0 +1,204 @@
...
+	require.NoError(t, err)
+	ts1, err := fullNode1.ChainGetTipSetByHeight(ctx, max, types.EmptyTSK)
+	require.NoError(t, err)
+
+	nv22Supply, err := fullNode0.StateVMCirculatingSupplyInternal(ctx, ts0.Key())
+	require.NoError(t, err, "Failed to fetch nv22 circulating supply")
+	nv23Supply, err := fullNode1.StateVMCirculatingSupplyInternal(ctx, ts1.Key())
+	require.NoError(t, err, "Failed to fetch nv23 circulating supply")
+
+	// Unfortunately there's still some non-determinism in supply dynamics so check for equality within a tolerance
```
ZenGround0, line 144: This is a huge pain and I'm not sure where the non-determinism is coming from -- we are measuring two identically set up test networks with a single miner at the exact same epoch.  If I had to guess I'd say sensitivity of IP function to variations in message landing time.

Though I would love to understand why it is happening and eliminate non determinism from the itest framework, for the purposes of testing FIP 0065 isolating out the non-determinism is probably a lot of work for something that is ultimately not important.

itests/supply_test.go
```diff
@@ -0,0 +1,204 @@
...
+		Client:               client,
+		Provider:             provider,
+		ClientCollateral:     big.Zero(),
+		ProviderCollateral:   providerLocked,
+		StartEpoch:           10000,
+		EndEpoch:             10000 + abi.ChainEpoch(duration),
+		StoragePricePerEpoch: ppe,
+	}
+	buf := bytes.NewBuffer(nil)
+	proposal.MarshalCBOR(buf)
```
rvagg, line 184:
```suggestion
	require.NoError(t, proposal.MarshalCBOR(buf))
```

itests/supply_test.go
rvagg: that's kind of weird, is the difference unstable across different test runs? this wouldn't be due to the f099 changes would it?

rvagg (approved):
👍, although the need for a difference tolerance seems a bit concerning without an explanation

itests/supply_test.go
ZenGround0: Any ensemble run has the difference.  So between test runs and between the two ensembles in every test run.

Tbh it's not particularly surprising from the reputation of economics in new networks being difficult to work with.  And the sensitivity to message timing of filtering on power and circ supply[ within actors](https://github.com/filecoin-project/ref-fvm/blob/d10baf82857df234b37404507df8bbaa581990ac/shared/src/smooth/alpha_beta_filter.rs) is a hypothesis I'm reasonably ok with.

ZenGround0: The difference in locked balance is almost certainly coming from differences in total InitialPledge between runs as Burnt is kept in a different category.

anorth:
Thanks!

Issue #11635, opened 2024-02-20
Eth API: Trace installed bytecode in the Ethereum JSON-RPC `trace_block` output


Stebalien:
The Ethereum JSON-RPC `trace_block` API currently returns `0xfe` as the EVM bytecode for all newly created actors:

1. For non-EVM actors, this makes sense as there's way to return any reasonable bytecode.
2. For EVM actors, there's just no reliable way to get the bytecode because an EVM smart contract's bytecode can change multiple times over the course of the block, even without reverts, by invoking selfdestruct and re-deploying. Even [EIP-6780](https://eips.ethereum.org/EIPS/eip-6780) will maintain this behavior.

To fix this, we need some way to get the bytecode of an EVM smart contract in the middle of a block. We'll need to implement both:

- https://github.com/filecoin-project/ref-fvm/issues/1994
- https://github.com/filecoin-project/ref-fvm/issues/1995

Then, every time we see a "create" event in the trace, we can lookup the state of the smart contract after it finished running the constructor to retrieve the bytecode.

aarshkshah1992:
@rjan90 @jennijuju Why has this been prioritised ?

Stebalien:
An interim solution is to _only_ handle smart contracts that live beyond the end of the block. Basically:

1. While we process the trace, for each address (ID) at which a smart contract is deployed, record the last transaction that successfully (not reverted) created a smart contract there.
2. After processing the trace, for each of those addresses, lookup the associated smart contract.
3. If the smart contract has self-destructed and/or doesn't exist, ignore it.
4. If the smart contract exists and is alive, get its bytecode and _replace_ the "fake" bytecode we recorded in the trace with the real bytecode.

Note: we're only altering the trace of the _last successful_ attempt to deploy a contract at any given address. Given that we won't have any intermediate and/or reverted state, that's all we can do. On the other hand, that's probably sufficient in most cases.

Stebalien:
Basically:

1. If a call and/or sub-call hasn't been reverted, pass a non-nil `deployedCode` map with a type `map[ActorID]*EthBytes` through the tracing logic.
2. When processing a contract creation and `deployedCode` is non-nil, record `deployedCode[contractID] = &crateTraceAction.Code` _even if there's already an entry for `contractID` in this mapping_.
3. After processing the trace, for each `contractID, codePtr` in `deployedCode`, lookup the associated contract, assert that it isn't dead, and set `*codePtr = contract.GetBytecode()`.

[...]

PR #11984, opened 2024-05-10
feat: ETH TRACE API: Trace installed bytecode in the Ethereum JSON-RPC trace_block output

aarshkshah1992:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/11635.

## Checklist

- [ ] CI is green

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -855,18 +855,20 @@ func (a *EthModule) EthTraceBlock(ctx context.Context, blkNum string) ([]*ethtyp
 		return nil, xerrors.Errorf("failed load computed state-tree: %w", err)
 	}
 
-	cid, err := ts.Key().Cid()
+	tsCid, err := ts.Key().Cid()
 	if err != nil {
 		return nil, xerrors.Errorf("failed to get tipset key cid: %w", err)
 	}
 
-	blkHash, err := ethtypes.EthHashFromCid(cid)
+	blkHash, err := ethtypes.EthHashFromCid(tsCid)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to parse eth hash from cid: %w", err)
 	}
 
 	allTraces := make([]*ethtypes.EthTraceBlock, 0, len(trace))
 	msgIdx := 0
+	deployedCode := make(map[abi.ActorID]*cid.Cid)
+
 	for _, ir := range trace {
 		// ignore messages from system actor
 		if ir.Msg.From == builtinactors.SystemActorAddr {
@@ -888,7 +890,12 @@ func (a *EthModule) EthTraceBlock(ctx context.Context, blkNum string) ([]*ethtyp
 			return nil, xerrors.Errorf("when processing message %s: %w", ir.MsgCid, err)
 		}
 
-		err = buildTraces(env, []int{}, &ir.ExecutionTrace)
+		var deployed map[abi.ActorID]*cid.Cid
+		if ir.MsgRct.ExitCode.IsSuccess() {
+			deployed = deployedCode
+		}
+
+		err = buildTraces(env, []int{}, &ir.ExecutionTrace, deployed)
 		if err != nil {
 			return nil, xerrors.Errorf("failed building traces for msg %s: %w", ir.MsgCid, err)
 		}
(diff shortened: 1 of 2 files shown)


aarshkshah1992:
@Stebalien Does this look correct ?

node/impl/full/eth_trace.go
```diff
@@ -498,6 +499,13 @@ func traceEthCreate(env *environment, addr []int, et *types.ExecutionTrace) (*et
 		// So we don't try and include a sentinel "impossible bytecode"
 		// value (the value specified by EIP-3541).
 		output = []byte{0xFE}
+		if deployedCode != nil {
+			var getBytecodeReturn evm12.GetBytecodeReturn
+			if err := getBytecodeReturn.UnmarshalCBOR(bytes.NewReader(et.MsgRct.Return)); err == nil && getBytecodeReturn.Cid != nil {
```
aarshkshah1992, line 504: @Stebalien 

Is this correct ? Can I get the byte code from the return value of the receipt on the top level transaction ?

aarshkshah1992: How else can I get the `cid` of the code here ?

node/impl/full/eth_trace.go
```diff
@@ -498,6 +499,13 @@ func traceEthCreate(env *environment, addr []int, et *types.ExecutionTrace) (*et
...
+				deployedCode[abi.ActorID(id)] = getBytecodeReturn.Cid
```
aarshkshah1992, line 505: @Stebalien This is not correct. How do I get the byte code of the contract here ?

aarshkshah1992: Do I need to make a call to the `GetBytecode` method of the contract Actor ?
Looks like none of the trace types we here have the byte code in them.

jennijuju:
cc @eshon this PR should fix the issue you brought up in our 1:1!

node/impl/full/eth.go
```diff
@@ -888,7 +890,12 @@ func (a *EthModule) EthTraceBlock(ctx context.Context, blkNum string) ([]*ethtyp
 			return nil, xerrors.Errorf("when processing message %s: %w", ir.MsgCid, err)
 		}
 
-		err = buildTraces(env, []int{}, &ir.ExecutionTrace)
+		var deployed map[abi.ActorID]*cid.Cid
+		if ir.MsgRct.ExitCode.IsSuccess() {
+			deployed = deployedCode
+		}
```
Stebalien, line 896: We need to do this per sub-trace. The top-level message can succeed while sub-messages cain fail.

Basically, move this into `buildTraces` and, if the receipt of the currently processed message has a non-zero exit code, set this to nil for the sub-call.

node/impl/full/eth.go
```diff
@@ -904,9 +911,62 @@ func (a *EthModule) EthTraceBlock(ctx context.Context, blkNum string) ([]*ethtyp
...
+		return nil, xerrors.Errorf("failed to update contract addresses: %w", err)
+	}
+
 	return allTraces, nil
 }
 
+func (a *EthModule) updateContractAddresses(ctx context.Context, ts *types.TipSet, allTraces []*ethtypes.EthTrace, deployedCode map[abi.ActorID]*cid.Cid) error {
+	// Update Contract Addresses if contract still exists
+	for _, trace := range allTraces {
+		switch r := trace.Result.(type) {
```
Stebalien, line 930: This isn't correct, we want to update the last one _only_. The idea was to record a pointer to `r.Code` in `deployedCode`, so we could set it without even having to look at the traces. That is:

```go
deployedCode := make(map[abi.ActorID]*EthBytes)
```

Then, inside the `buildTraces` logic, we'd record `deployedCode[receiverID] = &r.Code`.

Finally, we'd iterate over `deployedCode` and set `*deployedCode[receiverID] = loadByteCode(receiverID)`, as long as the receiver is still live.

node/impl/full/eth_trace.go
Stebalien: Sorry, I missed this question. You can actually just read the bytecode from the EVM actor state (load the actor and call `GetBytecode` using the actor shims.

However, you have to do this _after_ building the trace (see https://github.com/filecoin-project/lotus/pull/11984#discussion_r1612060558)

BigLep:
@aarshkshah1992 : I'm moving this back to draft given it hasn't been updated for a few months and I'm not aware of it being a priority currently.

aarshkshah1992:
@BigLep Closing this as Im not sure if this is correct and it needs engagement from Steb if we get to it again.

Stebalien:
I've restored the branch here because the code may be useful in the future.

PR #11994, opened 2024-05-15
feat: curio ffiselect: Isolate gpu calls in a subprocess

snadrus:
## Related Issues

The GPU picker library "ffiselect" will run ffi calls on an assigned GPU using a subprocess. It does not respect deadlines, but will survive fatal behavior of the subprocess.

This PR also removes localWorker (tech debt) from curio and copies code from there and the sealer into Curiosrc near the usage. 

## Checklist

- [ ] CI is green


snadrus:
@magik6k  what should I put in as an argument to cmd/curio/ffi.go call to ffiwrapper.New() ?

snadrus:
@magik6k please review my approach. 
1. how do I avoid calling WindowPoSt in a loop?
2. WInningPoSt: Should I bring GenerateSingleVanillaProof into the FFI Select wrapper?

cmd/curio/ffi.go
```diff
@@ -0,0 +1,116 @@
...
+		}()
+		var callInfo ffiselect.FFICall
+		if err := gob.NewDecoder(os.Stdin).Decode(&callInfo); err != nil {
+			return err
+		}
+
+		// wasteful, but *should* work
+		depnd, err := deps.GetDeps(cctx.Context, cctx)
+
+		w, err := ffiwrapper.New(&sectorProvider{index: depnd.Si, stor: depnd.LocalStore})
```
magik6k, line 46: This wrapper definitely shouldn't wrap around the ffiwrapper. It absolutely can't have it's own storage/path Index/Remote/Local instances, those must be in the same process as the scheduler, otwerwise we break storage reservations.

This command should allow for invoking methods one layer deeper, like [SealPreCommitPhase2](https://github.com/filecoin-project/lotus/blob/3d6ae227ef031c06f2c5062083b5d1059766db78/curiosrc/ffi/sdr_funcs.go#L259) - that will be much, much simpler as we won't need connect to the db (no GetDeps above) and interact with the storage index for anything. Also storage reservations won't need a rewrite.

The wrapper should be integrated in `curiosrc/ffi`, not in tasks; the curio ffi wrapper is meant to abstract interactions with the proofs library down to "run this algorithm on this sector-id now"

magik6k (changes requested):
I feel quite strongly that we're passing those calls one layer of abstraction above where we should be passing them.

cmd/curio/ffi.go
snadrus: I connected it at the FFI level, but I still like reflection so I made "ffidirect" under "ffiselect" as we will need to update both together. 

 I removed the dependency on LocalWorker and copied-over the related funcs to get access to where the edits were needed. That's why "package deps" was touched. 
This tech debt clean-up simplifies the mental model in this area and helps us avoid importing skeletons.

curiosrc/winning/winning_task.go
```diff
@@ -433,6 +435,47 @@ func (t *WinPostTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
...
+	vproofs := make([][]byte, len(sectors))
+	var rerr error
+
+	for i, s := range sectors {
+		go func(i int, s storiface.PostSectorChallenge) {
+			defer wg.Done()
+
+			vanilla, err := t.paths.GenerateSingleVanillaProof(ctx, mid, s, ppt)
+			if err != nil {
+				rerr = multierror.Append(rerr, xerrors.Errorf("get winning sector:%d,vanila failed: %w", s.SectorNumber, err))
```
magik6k, line 460: This is technically a racy assignment; Realistically not a problem because most (all?) of the time we just prove one sector and usually don't fail here; Still would be more correct to wrap this in some lock

curiosrc/window/compute_do.go
```diff
@@ -25,12 +25,16 @@ import (
...
 	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/lib/ffiselect"
 	"github.com/filecoin-project/lotus/storage/sealer"
 	"github.com/filecoin-project/lotus/storage/sealer/storiface"
 )
 
 const disablePreChecks = false // todo config
 
+// TODO Surface into config?
+const challengeReadTimeout = 0
```
magik6k, line 36: This is in the config, also we really should wire this up.

https://github.com/filecoin-project/lotus/blob/master/documentation/en/default-curio-config.toml#L246

Without this if some sectors are e.g. on a blocked nfs mount we'll fail all sectors in the partition instead of just the sectors on the blocking mount.

lib/ffiselect/ffiselect.go
magik6k, line 1: Would be really nice to have a "SelfTest" function and call it on node startup to confirm that this mechanism works and there aren't issues caused by how the binary was deployed.

lib/ffiselect/ffiselect.go
```diff
@@ -0,0 +1,154 @@
...
+func SealPreCommitPhase2(
+	phase1Output []byte,
+	cacheDirPath string,
+	sealedSectorPath string,
+) (sealedCID cid.Cid, unsealedCID cid.Cid, err error) {
+	val, err := call("SealPreCommitPhase2", phase1Output, cacheDirPath, sealedSectorPath)
+	if err != nil {
+		return cid.Undef, cid.Undef, err
+	}
+	return val[0].(cid.Cid), val[1].(cid.Cid), nil
```
magik6k, line 125: Are you sure that the Cid casts actually work? Under the hood CIDs are `type Cid struct{ str string }`

lib/ffiselect/ffiselect.go
```diff
@@ -0,0 +1,154 @@
...
+	phase1Output []byte,
+	sectorNum abi.SectorNumber,
+	minerID abi.ActorID,
+) ([]byte, error) {
+	val, err := call("SealCommitPhase2", phase1Output, sectorNum, minerID)
+	if err != nil {
+		return nil, err
+	}
+
+	return val[0].([]byte), val[1].(error)
```
magik6k, line 138: This error cast will either return `interface<Type: error, Value: nil>`, which when you'll check the returned err will tell you that the error is not in fact nil, or if `val[1]` is a real nil, the cast will panic
```suggestion
	return val[0].([]byte), nil
```

magik6k (commented):
The linter does appear to be correctly complainy about how we return errors from the ffiselect wrapper.

Some minor things to address, overall looks really good, can't wait to give this a try!

curiosrc/winning/winning_task.go
snadrus: mind that I got rid of the multi-err since it's typically only 1 sector which can only make 1 err?

lib/ffiselect/ffiselect.go
snadrus: Cid has MarshalBinary & UnmarshalBinary which Gob can use to round-trip. This type cast only gets away from the 'any' type.

lib/ffiselect/ffiselect.go
snadrus: done. It round-trips a novel Cid.

magik6k (approved):

snadrus:
Your changes look good @magik6k. I also added an always-ran "test" to verify everything looks sharp. I realize that part of that could be a unit test instead, but I'm not sure if that's working well at the moment.

magik6k (approved):
Looks good now, and works on my machine. Flaky curio test is not from this PR, will fix in a followup

PR #11998, opened 2024-05-16
feat: api: add SectorNumber to MarketDealState (nv22)

rvagg:


rvagg:
haha, same PR! and I even reviewed that one .. but I do see a bug in that other one

rjan90 (approved):

api/api_full.go
```diff
@@ -1166,13 +1166,15 @@ type MarketBalance struct {
 }
 
 type MarketDealState struct {
-	SectorStartEpoch abi.ChainEpoch // -1 if not yet included in proven sector
-	LastUpdatedEpoch abi.ChainEpoch // -1 if deal state never updated
-	SlashEpoch       abi.ChainEpoch // -1 if deal never slashed
+	SectorNumber     abi.SectorNumber // 0 if not yet included in proven sector (0 is also a valid sector number).
```
LexLuthr, line 1169: Can we do this to -1. Having 0 as valid sector creates a lot of issues in checks. Boost already have a form on this issue for indexing and deal announcement where deal fails in AP(). I would love to have uniformity of -1 for sector number as much as possible.

rvagg: sadly not if we're going to stick with `abi.SectorNumber`, and this was a limitation of FIP0076 too

```go
type SectorNumber uint64
```

It's in state as `0` so even if we introduced a new `int64` for this, we'd still be loading >=`0` from state to fetch this.

LexLuthr: 😞 I can work with this. I will handle the exception on our side.

PR #12015, opened 2024-05-20
chore : CoreTypes : Remove stray C dependency

ribasushi:
Without this change the following test fails:

`CGO_ENABLED=0 go test -count=1 ./chain/types_test.go`

Without this fix importing the chain types into a fully-static executable is not possible.

I am not adding a test, as that should be done within the CI harness somehow (there is no way to force-disable compile-time flags from a test itself). Someone more familiar with how tests are run on a macro-level (@rvagg ?) should add a one-liner somewhere so things do not regress in the future.

### DIFF
--- a/build/builtin_actors.go
+++ b/build/builtin_actors.go
@@ -12,10 +12,10 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/DataDog/zstd"
 	"github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
 	"github.com/ipld/go-car"
+	"github.com/klauspost/compress/zstd"
 	"golang.org/x/xerrors"
 
 	actorstypes "github.com/filecoin-project/go-state-types/actors"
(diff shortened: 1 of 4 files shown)


rvagg (approved):
🤷 I don't have any bright ideas for locking in testing that would prevent regressions, we have unit tests all over the place that need cgo so it can't just be a re-run with a blanket exclusion, it'd have to be specific.

rvagg:
@ribasushi how about removing the dependency completely so it'll be obvious if it ever comes back?

build/builtin_actors_gen_test.go still has a reference, which seems a bit redundant since we're now using 2 libraries for the same code paths .. perhaps it helps test correctness but do we think there's a problem?

cmd/lotus/daemon.go of course is the big one, it'd be interesting to get an idea of the speed difference importing a snapshot between the two implementations, if it's nearly as good, or better, then we may as well ditch the dependency entirely?

Stebalien:
@rvagg unfortunately, badger v2 pulls in github.com/DataDog/zstd. So we're not going to be able to _completely_ ditch it. But:

1. I'd definitely remove it from `builtin_actors_gen_test.go`.
2. I'd leave it in `cmd/lotus/daemon.go` unless someone wants to benchmark snapshot imports (snapshot importing may be io bound, but I'm not sure).

rvagg:
@ribasushi would you mind at least removing the test dependency?

rvagg:
We could build this into the test.yml workflow - add another `utests` item with a specific list and add a new property like `yugabytedb` and `parameters` to inject the cgo disable env var. But ./chain/types_test.go doesn't buy you much, just `Message` and `SignedMessage`, right?

If you want to make a list of tests to run I can try and work on something but there's just not much to ./chain/types_test.go that it doesn't seem worth the hassle to me.

Issue #12009, opened 2024-05-17
[DX Streamline] Lean NI-PoRep integration testing


rvagg:
As part of the effort to decouple lotus node development from lotus-miner development, this task aims to implement integration testing of the [FIP-0090](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0090.md) (and likely [FIP-0092](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0092.md)) NI-PoRep feature without requiring full lotus-miner implementation of the feature. This will likely lead to changes in the itest kit setup and more verbose testing in the short term, but a better understanding of the longer-term implications of decoupling.

rjan90:
Reopening this ticket. 

#12017 delivered the generic testkit manual miner that can be used to onboard CC sectors and Snap deal sectors which was a pre-requisite for the lean NI-PoRep integration testing.

https://github.com/filecoin-project/lotus/pull/12045 is planned to deliver the NI-PoRep part.

rjan90:
Closing as this has been completed

PR #12005, opened 2024-05-16
cleanup: miner: remove markets and deal-making from Lotus Miner

aarshkshah1992:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/12004

## Proposed Changes
The use of legacy markets was deprecated from Lotus Miner a long time ago in favour of Boost. This PR removes **all** legacy markets related code from the Lotus Miner. Lotus no longer depends on `go-graphsync`, `go-data-transfer`, `go-fil-markets`, `dagstore` and `index-provider` .

All Markets related functionality has been removed from the Lotus Miner CLI and the API. Specifically, the following Market APIs have been removed from a Lotus Miner API as part of this PR:

```
         MarketImportDealData(ctx context.Context, propcid cid.Cid, path string) error //perm:write
	// MarketListRetrievalDeals is deprecated, returns empty list
	MarketListRetrievalDeals(ctx context.Context) ([]struct{}, error)                                                                                                                    //perm:read
	MarketGetDealUpdates(ctx context.Context) (<-chan storagemarket.MinerDeal, error)                                                                                                    //perm:read
	MarketListIncompleteDeals(ctx context.Context) ([]storagemarket.MinerDeal, error)                                                                                                    //perm:read
	MarketSetAsk(ctx context.Context, price types.BigInt, verifiedPrice types.BigInt, duration abi.ChainEpoch, minPieceSize abi.PaddedPieceSize, maxPieceSize abi.PaddedPieceSize) error //perm:admin
	MarketGetAsk(ctx context.Context) (*storagemarket.SignedStorageAsk, error)                                                                                                           //perm:read
	MarketSetRetrievalAsk(ctx context.Context, rask *retrievalmarket.Ask) error                                                                                                          //perm:admin
	MarketGetRetrievalAsk(ctx context.Context) (*retrievalmarket.Ask, error)                                                                                                             //perm:read
	MarketListDataTransfers(ctx context.Context) ([]DataTransferChannel, error)                                                                                                          //perm:write
	MarketDataTransferUpdates(ctx context.Context) (<-chan DataTransferChannel, error)                                                                                                   //perm:write
	// MarketDataTransferDiagnostics generates debugging information about current data transfers over graphsync
	MarketDataTransferDiagnostics(ctx context.Context, p peer.ID) (*TransferDiagnostics, error) //perm:write
	// MarketRestartDataTransfer attempts to restart a data transfer with the given transfer ID and other peer
[...]

PR #12017, opened 2024-05-20
test: actors: manual CC onboarding and proving integration test

rvagg:
WIP, working toward https://github.com/filecoin-project/lotus/issues/12009

Starting here with CC onboarding by manually crafting messages with the current flow: 

1. `PreCommitSectorBatch2`
2. `ProveCommitSectors3`
3. `SubmitWindowedPoSt`
4. Confirm by checking miner power


itests/manual_cc_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+		client kit.TestFullNode
+		minerA kit.TestMiner // A has some genesis sector, just keeps power
+		minerB kit.TestMiner // B pledges some sector
+	)
+
+	opts := []kit.NodeOpt{kit.WithAllSubsystems()}
+	ens := kit.NewEnsemble(t, kit.MockProofs()).
+		FullNode(&client, opts...).
+		Miner(&minerA, &client, opts...).
```
aarshkshah1992, line 47: Right so we just give A some genesis sectors with power here but then don't add more sectors to A, right ?

itests/manual_cc_test.go
```diff
@@ -0,0 +1,216 @@
...
+	t.Logf("MinerB RBP: %v, QaP: %v", p.MinerPower.QualityAdjPower.String(), p.MinerPower.RawBytePower.String())
+	require.True(t, p.MinerPower.RawBytePower.IsZero())
+
+	t.Log("Submitting ProveCommitSector...")
+
+	bSectorNum := preCommitParams.Sectors[0].SectorNumber
+
+	proveCommitParams := miner14.ProveCommitSectors3Params{
+		SectorActivations:        []miner14.SectorActivationManifest{{SectorNumber: bSectorNum}},
+		SectorProofs:             [][]byte{{0xde, 0xad, 0xbe, 0xef}},
```
aarshkshah1992, line 119: Where did you get these values from ?

itests/manual_cc_test.go
```diff
@@ -0,0 +1,216 @@
...
+	req.NoError(err)
+	require.True(t, r.Receipt.ExitCode.IsSuccess())
+
+	t.Log("Checking power after PoSt...")
+
+	p, err = client.StateMinerPower(ctx, maddrB, r.TipSet)
+	req.NoError(err)
+	// require.True(t, p.MinerPower.RawBytePower.IsZero())
+	t.Logf("MinerB RBP: %v, QaP: %v", p.MinerPower.QualityAdjPower.String(), p.MinerPower.RawBytePower.String())
+	req.Equal(uint64(2<<10), p.MinerPower.RawBytePower.Uint64())    // 2kiB RBP
```
aarshkshah1992, line 214: Do you start getting power only after submitting the first WindowPost ?

itests/manual_cc_test.go
rvagg: Yes, and there's an additional action item here at some stage to try and do this outside of the miner framework - there's both presealing and regular chain iteration happening behind the scenes for `minerA`. Not so important now, and maybe it just means tinkering with the test kit to decouple it a little more.

itests/manual_cc_test.go
rvagg: They could be anything for now, this is using mock proofs which have a "always OK" effect when calling down in to the lower-level proofs (FFI) functions. Next step here would be to replace the mock proofs with real ones and do the FFI calls manually as well. We'll have to consider how much that buys us for NI-PoRep and whether it's worth the hassle. It would be good to do at some point though.

itests/manual_cc_test.go
rvagg: Yes, I believe this is how the protocol is intended to work - precommit, commit, then prove before you get power, and we have to wait for the right window to submit the proof, hence the complicated deadline wait logic.

itests/manual_cc_test.go
rjan90: If I understand correctly, I think it would be really nice to replace the mock proofs with real ones, as I suspect there is a lot of edge case testing that will be hard to do if the proofs always have a "always OK" effect, like: "Onboard a sectors, submit a fake/bad proofs, confirm that the disputer catches it"

rvagg: yeah, I'm in the process of figuring out how on earth to do that, it's buried under layers of abstraction that I'm trying to grok

jennijuju:
NICE

itests/manual_cc_test.go
aarshkshah1992: @rjan90 But looks like our existing itests use only mock proofs.

aarshkshah1992: So we've never done this kind of integration testing before.

aarshkshah1992:
@rvagg This test is failing for me locally after the force push.

itests/manual_onboarding_test.go
```diff
@@ -0,0 +1,534 @@
...
+	head, err := client.ChainHead(ctx)
+	req.NoError(err)
+
+	sp, err := client.StateSectorPartition(ctx, minerAddr, sectorNumber, head.Key())
+	req.NoError(err)
+
+	di, err := client.StateMinerProvingDeadline(ctx, minerAddr, head.Key())
+	req.NoError(err)
+
+	disputeEpoch := di.Challenge + miner14.WPoStDisputeWindow + 5
```
aarshkshah1992, line 511:
```
// WPoStDisputeWindow is the period after a challenge window ends during which
// PoSts submitted during that period may be disputed.
```

Now with 

```
disputeEpoch := di.Challenge + miner14.WPoStDisputeWindow + 5
```
This makes sense as we want to ensure that the WdPost message submitted by the Miner has achieved finality before we dispute it. But why do we use `di.Challenge` here instead of `di.Open`?

aarshkshah1992 (commented):
Great Great stuff Rodd !

rvagg:
Rebased on top of current #12005 which has master merged back in. Also pulled in #12049 as a squashed commit. We can continue iterating from here.

aarshkshah1992 (approved):
lgtm and merging to mastern as discussed offline

aarshkshah1992:
The test failure is a flaky.

PR #12025, opened 2024-05-21
chore: deps: update ipni

Stebalien:
## Proposed Changes

Update IPNI and indexer-related deps. Removes support for the data-transfer component.

## Additional Info

I have _no idea_ how the removal of the data transfer component interacts with lotus and/or whether or not IPNI is even useful in lotus without it.

## Checklist

- [ ] CI is green


node/modules/storageminer_idxprov.go
```diff
@@ -77,7 +77,6 @@ func IndexProvider(cfg config.IndexProviderConfig) func(params IdxProv, marketHo
 			ma := address.Address(maddr)
 			opts = append(opts,
 				engine.WithPublisherKind(engine.DataTransferPublisher),
-				engine.WithDataTransfer(dt),
```
Stebalien, line 80: @masih so... what's this mean for users?

Stebalien:
It looks like, at a minimum, the urfav update is causing quite a few issues.

node/modules/storageminer_idxprov.go
rvagg: not much, pull out whatever you want be cause this whole thing, and its dependency, is going when we land https://github.com/filecoin-project/lotus/pull/12005 :soon:

Stebalien: Ooh. I didn't see that. I'll close this and wait.

masih: The default transport for syncing ad chains used to be data transfer a long long time ago. Now it is moved to use HTTP over libp2p as default with data transfer support being entirely deprecated and eventually removed.

Please feel free to remove all things data transfer related in the context of IPNI. Indexer servers are already upgraded to support the latest.

masih: > `engine.WithPublisherKind(engine.DataTransferPublisher)`

The line above needs to be updated to use HTTP over libp2p by default. Otherwise removing `WithDataTransfer` would cause all sorts of issues since index-provider engine would try to instantiate its own, which will cause a protocol ID registration conflict since every new instance of data transfer does that IIRC.

Issue #12010, opened 2024-05-17
[DX Streamline] Lotus node-only release process


rvagg:
Current [workflow](https://github.com/filecoin-project/lotus/blob/master/.github/workflows/release.yml) runs `make deps lotus lotus-miner lotus-worker` and [GoReleaser config](https://github.com/filecoin-project/lotus/blob/master/.goreleaser.yaml) bundles the 3 binaries into a single tgz for uploading to the release page.

We need a streamlined way to produce `lotus`-only releases. Possibly using some variation of the IPDX `version.json` tooling ([e.g.](https://github.com/filecoin-project/lassie/pull/471)) such that it simplifies some or all of the steps outlined in the [release flow checklist](https://github.com/filecoin-project/lotus/blob/master/documentation/misc/RELEASE_ISSUE_TEMPLATE.md).

e.g. A pull request changing `build/version.go` will trigger other actions in that release flow, including setting up a draft release page, generating a basic changelog, running gen and docsgen-cli etc. When the PR is closed, it's tagged and a release is published. Somewhere in there, a flag that says "lotus-only", or the inverse, should indicate whether to include the additional components.

e.g. Alternatively, just work within the current process but add optionality to GoReleaser and release.yml that can switch between two modes with a simple flag from somewhere.

rvagg:
Outstanding question: all components use api/version.go to determine their build version, should we separate this out to have the daemon using a different version number. If we tie them together, even if we only release a daemon binary, that tag would give you a compile of lotus-miner that has the same version, yet it may not be something we really want people to compile and use because. Maybe we need a separate file for both.

galargh:
> We need a streamlined way to produce lotus-only releases. Possibly using some variation of the IPDX version.json tooling (https://github.com/filecoin-project/lassie/pull/471) such that it simplifies some or all of the steps outlined in the [release flow checklist](https://github.com/filecoin-project/lotus/blob/master/documentation/misc/RELEASE_ISSUE_TEMPLATE.md).

Yes, using something like a `version.json` file would definitely help. The big advantage of it is that it's easy to parse and use the information on the version in other parts of the release flow. In particular, it would be possible for us to get the version information from `version.json` in `build/version.go`.

[...]

PR #12035, opened 2024-05-23
feat!: build: separate miner and node version strings

rvagg:
_Proposal asking for feedback._

This removes `build.BuildVersion` and `build.UserVersion()` and replaces them with `build.NodeBuildVersion`, `build.MinerBuildVersion` and `build.NodeUserVersion()`, `build.MinerUserVersion()` so they can be incremented separately.

Why? Because as per https://github.com/filecoin-project/lotus/issues/12010 we would like to be able to release them separately. This has two problems:

1. Once we fix the version number in here without a `-dev` and tag it, you can build both miner and node with that version string and it looks official. But, we may be prepared to release one and not the other (node vs miner) and may consider one of them not release-ready so not deserving of an official release.
2. We would like to automate the build and release process such that it only produces the binaries we want, either node or miner, not both, so we can release them separately and clearly signal such.

I'm thinking that we could copy the ipdx `version.json` pattern implemented by https://github.com/ipdxco/unified-github-workflows/blob/main/.github/workflows/release-check.yml but with some minor tweaks.

Instead of doing the version.json check:

```sh
version="$(gh api -X GET "repos/$HEAD_FULL_NAME/contents/version.json" -f ref="$HEAD_SHA" --jq '.content' | base64 -d | jq -r '.version')"
```

We could run two checks on PRs (probably just on the `releases` branch?):

```sh
node_version="$(gh api -X GET "repos/$HEAD_FULL_NAME/contents/build/version.go" -f ref="$HEAD_SHA" --jq '.content' | base64 -d | grep '^const NodeBuildVersion = ' | cut -d\" -f2)"
miner_version="$(gh api -X GET "repos/$HEAD_FULL_NAME/contents/build/version.go" -f ref="$HEAD_SHA" --jq '.content' | base64 -d | grep '^const MinerBuildVersion = ' | cut -d\" -f2)"
```

Then, for both:

1. If it ends in `-dev`, no more need for further checks.
2. If there isn't a tag for that version and proceed with roughly the same things it does now, including making a draft release that can be further edited (could even build in some smarts to copypasta the versioned section of the changelog, but that'd be a stretch-goal).
3. It needs to remember whether it's `Node` or `Miner` and then either run `make deps lotus` or `make deps lotus-miner lotus-worker` and also probably do something different on the release draft.

I'm not sure yet what it should do if they both pass that, probably that just shouldn't be allowed and it has an error condition with a "nope" comment in the PR.

### DIFF
--- a/api/docgen-openrpc/openrpc.go
+++ b/api/docgen-openrpc/openrpc.go
@@ -106,7 +106,7 @@ func NewLotusOpenRPCDocument(Comments, GroupDocs map[string]string) *go_openrpc_
 			title := "Lotus RPC API"
 			info.Title = (*meta_schema.InfoObjectProperties)(&title)
 
-			version := build.BuildVersion
+			version := build.NodeBuildVersion
 			info.Version = (*meta_schema.InfoObjectVersion)(&version)
 			return info
 		},
--- a/build/panic_reporter.go
+++ b/build/panic_reporter.go
@@ -28,13 +28,31 @@ var PanicReportingPath = "panic-reports"
 // the lotus journal to be included in the panic report.
 var PanicReportJournalTail = defaultJournalTail
 
-// GeneratePanicReport produces a timestamped dump of the application state
+// GenerateNodePanicReport produces a timestamped dump of the application state
 // for inspection and debugging purposes. Call this function from any place
 // where a panic or severe error needs to be examined. `persistPath` is the
 // path where the reports should be saved. `repoPath` is the path where the
 // journal should be read from. `label` is an optional string to include
 // next to the report timestamp.
-func GeneratePanicReport(persistPath, repoPath, label string) {
+//
+// This function should be called for panics originating from the Lotus daemon.
+func GenerateNodePanicReport(persistPath, repoPath, label string) {
+	generatePanicReport(NodeUserVersion(), persistPath, repoPath, label)
+}
+
+// GenerateMinerPanicReport produces a timestamped dump of the application state
+// for inspection and debugging purposes. Call this function from any place
+// where a panic or severe error needs to be examined. `persistPath` is the
+// path where the reports should be saved. `repoPath` is the path where the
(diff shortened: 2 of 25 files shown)


galargh:
The split in `build/version.go` is definitely needed given the plans to support separate releases. In my opinion, we do need a little bit more separation between the tags associated with the main lotus releases and the ones for miner. 

One idea I proposed here (https://github.com/filecoin-project/lotus/issues/12010#issuecomment-2129792312) is to use prefixes to make the distinction. As in, the main lotus releases would happen on `v1.29.0`, `v1.29.1-dev`, `v1.29.2-rc3`, etc. while the miner would be on a separate cycle `miner-v1.29.0`, `miner-v1.29.1-dev`, `miner-v1.29.2-rc3`.

I'm afraid otherwise it might be confusing that there exists a release tag, but there is no lotus or lotus-miner published for it. 

As for the automation part, once we solve the above and agree on the source of truth (`version.json`, `CHANGELOG.md`, etc.), it won't be a problem. We'll adjust the workflow specifically to lotus' needs.

> I'm not sure yet what it should do if they both pass that, probably that just shouldn't be allowed and it has an error condition with a "nope" comment in the PR.

It's either not allowed or create both releases - we could easily accomplish that with using a matrix configuration for executing the release job for example. I'd lean towards do both myself.

rjan90:
Did a quick test on a devnet, setting a different version for miner and the lotus node, which worked well:

```
./lotus-miner -v
lotus-miner version 1.27.0+2k+git.67c3a6fd0.dirty
```

```
./lotus -v
lotus version 1.27.4-rc1+2k+git.67c3a6fd0.dirty
```
```
./lotus-worker -v
lotus-worker version 1.27.0+2k+git.67c3a6fd0.dirty
```

Also ran `make docsgen-cli`, and the documentation output correctly had seperate versions for the daemon and the miner.

rvagg:
k, well let's make it so

PR #12040, opened 2024-05-24
[WIP] Refactor Manual CC onbaording itest and add docs

aarshkshah1992:
## Checklist

- [ ] CI is green

### DIFF
--- a/itests/kit/ensemble.go
+++ b/itests/kit/ensemble.go
@@ -321,6 +321,10 @@ func (n *Ensemble) MinerEnroll(minerNode *TestMiner, full *TestFullNode, opts ..
 		PresealSectors: presealSectors,
 		options:        options,
 		RemoteListener: rl,
+
+		cacheDirPath:       options.cacheDirPath,
+		unsealedSectorPath: options.unsealedSectorPath,
+		sealedSectorPath:   options.sealedSectorPath,
 	}
 
 	minerNode.Libp2p.PeerID = peerId
--- a/itests/kit/node_full.go
+++ b/itests/kit/node_full.go
@@ -55,6 +55,18 @@ type TestFullNode struct {
 	options nodeOpts
 }
 
+func (f *TestFullNode) GetChainHead(ctx context.Context, t *testing.T) *types.TipSet {
+	head, err := f.ChainHead(ctx)
+	require.NoError(t, err)
+	return head
+}
+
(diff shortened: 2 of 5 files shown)


itests/kit/ensemble.go
```diff
@@ -321,6 +321,10 @@ func (n *Ensemble) MinerEnroll(minerNode *TestMiner, full *TestFullNode, opts ..
 		PresealSectors: presealSectors,
 		options:        options,
 		RemoteListener: rl,
+
+		cacheDirPath:       options.cacheDirPath,
+		unsealedSectorPath: options.unsealedSectorPath,
+		sealedSectorPath:   options.sealedSectorPath,
```
rvagg, line 327: if we're doing these in the ensemble then we should make it a map of `SectorNumber` to string so we can have more than one sector per miner

rvagg:
I'm not sure it's going to be this easy to extract the functions out because they currently assume a serial execution, but once we add even one more sector (in a new miner or the same) then everything gets parallel; we need to have some scheduling utilities so we know what to do when - two miners at the same time will have different deadlines for their sectors but we need to know what to do next and can't just wait for a particular miner; so we're back to the lotus-miner scheduling problem. Either we do it manually inside the test or build some basic scheduling smarts in the ensemble to do it.

rvagg:
We could extract a "window post" goroutine here that we spin off for each miner after we've onboarded a sector - it could simply figure out which sectors the miner has onboarded, calculate their deadlines, and then do basic posts for them. Perhaps even we could limit it to a single sector for now, so one miner per sector is our testing limitaiton to begin with, which would make this a bit easier to begin with.

We'd also want to be able to interact with it, perhaps be able to query "have you submitted first post", so that we can wait until it has before testing power.

Cancelling via context should be fine, probably feeding errors back to the main goroutine via a channel for nice `*testint.T` interaction might help too.

I'm not sure yet about precommit and provecommit. I'm currently writing some code with a `minerC` that does an NI-PoRep, and it's already a bit complicated. It might be easier to just treat them as sequential operations for now -- onboard a sector for `minerB`, then spin off the windowpost goroutine for that miner, then onboard a sector for `minerC` and spin off the windowpost goroutine for that miner, then have a sleep loop that checks on them both to see if they've both submitted at least their first post and then check power and exit.

PR #12045, opened 2024-05-27
test: actors: manual Ni-PoRep onboarding

rvagg:
Also adds multi-miner support to the manual test, including async PoST scheduling.

This depends on a bunch of changes elsewhere to work and needs manual wiring up:

- [ ] https://github.com/filecoin-project/builtin-actors/pull/1546 which I've included here as an actors bundle (DO NOT MERGE WITH THIS)
- [ ] https://github.com/filecoin-project/go-state-types/pull/270
- [ ] https://github.com/filecoin-project/rust-filecoin-proofs-api/pull/103
- [ ] https://github.com/filecoin-project/filecoin-ffi/pull/456
- [ ] https://github.com/filecoin-project/ref-fvm/pull/1999 rebased on master to add some stuff and remove bls pairing; also some awkward manual wiring I had to do in Cargo.toml to make this compile, which should go away when released


rvagg:
Oh, and this currently doesn't get past the actors call for NI-PoRep `WithRealProofs`, it fails with:

```
sector proof size 14164 exceeds max 192
```

I believe we should have a single groth16 snark proof that rolls up the new PoRep for the batch (single) case so I that should be `192` as far as I understand. But somehow we're getting this massive proof for a single sector.

Thread about this in Slack @ https://filecoinproject.slack.com/archives/C06EZQ2DTD0/p1716637350693809

rvagg:
FWIW this now passes if I comment out this line in actors: https://github.com/filecoin-project/builtin-actors/blob/308f0c06d01ae2fdc017fbe022a077bb972c3991/actors/miner/src/lib.rs#L2004

Discussion still ongoing regarding what the expected sizes are for proofs, but that check probably just needs to be changed to account for single NI-PoRep proof sizes which are much bigger than plain proofs.

rvagg:
Added support for NI-PoRep in `lotus-bench` so I can generate `--non-interactive` proofs and do some testing.

I did break `SealProofTypeFromSectorSize` in `chain/actors/builtin/miner` though, it currently has a `synthetic bool` argument, I changed it to:

```go
type SealProofVariant int

const (
	SealProofVariant_Standard SealProofVariant = iota
	SealProofVariant_Synthetic
	SealProofVariant_NonInteractive
)

// ...

func SealProofTypeFromSectorSize(ssize abi.SectorSize, nv network.Version, variant SealProofVariant) (abi.RegisteredSealProof, error) {
// ...
}
```

And it does slightly more rigorous checking on what's possible—the variant _must_ be valid for the network version.

rvagg:
Status of this is that (for the 3rd time) I'm going to drop this branch and push a new one soon. I've started again locally on top of master because most of the churn that went into this branch has been merged in the `UnmanagedMiner` work _and_ we have the markets and client removal stuff on top of that so there's a huge diff and it's almost impossible to reconcile now. Plus, NI-PoRep will go into a new, separate itest now that we have pulled all the shared `UnmanagedMiner` stuff into the testkit.

rvagg:
Replaced by https://github.com/filecoin-project/lotus/pull/12105

PR #12049, opened 2024-05-27
Manual CC & Snapdeals onboarding (No NI-Porep)

aarshkshah1992:
## Related Issues
This PR aims to refactor https://github.com/filecoin-project/lotus/pull/12017 by creating a generic testkit manual miner that can be used to onboard CC sectors and Snap deal sectors. It does not have any of the NI-PoRep tests introduced in https://github.com/filecoin-project/lotus/pull/12045. 

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/kit/node_unmanaged.go
+++ b/itests/kit/node_unmanaged.go
@@ -1,17 +1,26 @@
 package kit
 
 import (
+	"bytes"
 	"context"
+	"crypto/rand"
 	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	ffi "github.com/filecoin-project/filecoin-ffi"
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/builtin"
 	miner14 "github.com/filecoin-project/go-state-types/builtin/v14/miner"
 	"github.com/filecoin-project/go-state-types/crypto"
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/chain/actors"
+	"github.com/filecoin-project/lotus/chain/actors/policy"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/ipfs/go-cid"
+	"github.com/stretchr/testify/require"
 	cbg "github.com/whyrusleeping/cbor-gen"
-	"testing"
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/lotus/chain/wallet/key"
@@ -23,8 +32,19 @@ import (
 // infrastructure, all tasks must be manually executed, managed and scheduled by
 // the test or test kit.
 type TestUnmanagedMiner struct {
-	t       *testing.T
-	options nodeOpts
+	t                 *testing.T
+	options           nodeOpts
+	cacheDir          string
+	unsealedSectorDir string
+	sealedSectorDir   string
+	sectorSize        abi.SectorSize
+
+	CacheDirPaths       map[abi.SectorNumber]string
+	UnsealedSectorPaths map[abi.SectorNumber]string
+	SealedSectorPaths   map[abi.SectorNumber]string
+	SealedCids          map[abi.SectorNumber]cid.Cid
+	UnsealedCids        map[abi.SectorNumber]cid.Cid
+	SealTickets         map[abi.SectorNumber]abi.SealRandomness
 
 	ActorAddr address.Address
 	OwnerKey  *key.Key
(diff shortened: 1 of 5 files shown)


itests/kit/node_unmanaged.go
```diff
@@ -9,12 +17,15 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
+const sectorSize = abi.SectorSize(2 << 10) // 2KiB
```
rvagg, line 20: There's already a `sectorSize` in the test kit, I just couldn't figure out a way to feed that back, I wanted to ask the kit "what sector size am I using?" but I think 2k just gets hard-wired everywhere in itests now, which is a little unfortunate. Maybe we should make a `SectorSize()` method on the miner so we can: use the 2k default that the test kit uses, have a `SectorSize()` option, and also have a `SectorSize()` method on the miner so we can work with it.

But also this brought up another, adjacent, issue when I was looking at it, we have this line: https://github.com/filecoin-project/lotus/blob/6302607650f9191bcc471547c595d712875b2f0f/itests/kit/ensemble.go#L284-L286

But we also have `kit.TestSpt` and I'm adding `kit.TestSptNi`, so hard-wired seal proof types; I don't think they should be hard-wired but maybe we add new methods `SealProofType()` and `SealProofTypeNi()` to get them instead of these constants.

itests/kit/node_opts.go
```diff
@@ -82,6 +84,13 @@ type OptBuilder func(activeNodes []*TestFullNode) node.Option
 // NodeOpt is a functional option for test nodes.
 type NodeOpt func(opts *nodeOpts) error
 
+func WithMockProofs() NodeOpt {
```
rvagg, line 87: there's already one on the ensemble itself, https://github.com/filecoin-project/lotus/blob/6302607650f9191bcc471547c595d712875b2f0f/itests/kit/ensemble_opts.go#L38

I think we should be able to fetch `n.options.mockProofs` when we instantiate `TestUnmanagedMiner` so there's no need for a new option, use what's there, just expose it in the `TestUnmanagedMiner` struct.

itests/kit/node_unmanaged.go
```diff
@@ -23,4 +34,78 @@ type TestUnmanagedMiner struct {
 		PeerID  peer.ID
 		PrivKey libp2pcrypto.PrivKey
 	}
+
+	SealedCids, UnsealedCids map[abi.SectorNumber]cid.Cid
+	SealTickets              map[abi.SectorNumber]abi.SealRandomness
+}
+
+func (tm *TestUnmanagedMiner) ManualOnboardingGeneratePreCommit(
```
rvagg, line 42: let's get rid of `ManualOnboarding`, I just prefixed with those names because utility functions in `itest/` need to be unique because they are all run as `go test ./itest/test-thing.go` so it's both a shared namespace when we dev, but a separate namespace when they are run (ugh), but that's also why it's good to get these out into the kit so they can be shared!

itests/kit/node_unmanaged.go
```diff
@@ -23,4 +34,78 @@ type TestUnmanagedMiner struct {
...
+		sealedSectorPath,
+	)
+	req.NoError(err)
+
+	tm.t.Logf("Unsealed CID: %s", unsealedCid)
+	tm.t.Logf("Sealed CID: %s", sealedCid)
+
+	tm.SealTickets[sectorNumber] = sealTickets
+	tm.SealedCids[sectorNumber] = sealedCid
+	tm.UnsealedCids[sectorNumber] = unsealedCid
```
rvagg, line 110: 👌 this is great, so we just ask it to make the proof and hold on to it, then for ni-porep we can reuse it for the prove commit, but for the standard path we submit the message.

one thing I don't know yet is whether we need a new cache dir for each run of this; it may be that the per-miner cache dir is not enough, and it should be per-commit

aarshkshah1992: @rvagg Yeah we do need a new cache dir for each run. Have added it to the latest commit.

itests/kit/node_unmanaged.go
aarshkshah1992: Done.

itests/kit/node_opts.go
aarshkshah1992: I have removed tests with mock proofs completely for now. Will add it back as a separate test.

aarshkshah1992:
@rvagg This is ready for a review. We can merge this and I can then continue work on top of this.

itests/kit/node_unmanaged.go
rvagg: this still needs to go, it's in the opts, we should only have it once

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+	}
+	tm.Libp2p.PeerID = peerId
+	tm.Libp2p.PrivKey = privkey
+
+	return &tm
+}
+
+func (tm *TestUnmanagedMiner) AssertNoPower(ctx context.Context) {
+	p := tm.CurrentPower(ctx)
+	tm.t.Logf("MinerB RBP: %v, QaP: %v", p.MinerPower.QualityAdjPower.String(), p.MinerPower.RawBytePower.String())
```
rvagg, line 123: log line has `MinerB`, could either remove this entirely or replace it with `f0` actor address—which would be helpful because you can then correlate it with the order they were instantiated

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+	p, err := tm.FullNode.StateMinerPower(ctx, tm.ActorAddr, head.Key())
+	require.NoError(tm.t, err)
+
+	return p
+}
+
+func (tm *TestUnmanagedMiner) AssertPower(ctx context.Context, raw uint64, qa uint64) {
+	req := require.New(tm.t)
+	p := tm.CurrentPower(ctx)
+	tm.t.Logf("MinerB RBP: %v, QaP: %v", p.MinerPower.QualityAdjPower.String(), p.MinerPower.RawBytePower.String())
```
rvagg, line 140: ditto log line

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+		_ = sealedSectorFile.Close()
+	}()
+
+	// Update paths for the sector
+	tm.sealedSectorPaths[sectorNumber] = sealedSectorFile.Name()
+	tm.unsealedSectorPaths[sectorNumber] = unsealedSectorFile.Name()
+	tm.cacheDirPaths[sectorNumber] = filepath.Join(tm.cacheDir, fmt.Sprintf("%d", sectorNumber))
+
+	// Ensure the cache directory exists
+	_ = os.Mkdir(tm.cacheDirPaths[sectorNumber], 0755)
```
rvagg, line 194: I think this isn't even necessary, I've been running proofs manually and it seems to make the cache itself, perhaps try removing this and see if it passes

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+	return publicPieces
+}
+
+func (tm *TestUnmanagedMiner) makeAndSaveCCSector(_ context.Context, sectorNumber abi.SectorNumber) {
+	requirements := require.New(tm.t)
+
+	// Create cache directory
+	cacheDirPath := filepath.Join(tm.cacheDir, fmt.Sprintf("%d", sectorNumber))
+	requirements.NoError(os.Mkdir(cacheDirPath, 0755))
+	tm.t.Logf("MinerB: Sector %d: created cache directory at %s", sectorNumber, cacheDirPath)
```
rvagg, line 205: `MinerB`

rvagg: 2 more in this function

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+
+	return publicPieces
+}
+
+func (tm *TestUnmanagedMiner) makeAndSaveCCSector(_ context.Context, sectorNumber abi.SectorNumber) {
+	requirements := require.New(tm.t)
+
+	// Create cache directory
+	cacheDirPath := filepath.Join(tm.cacheDir, fmt.Sprintf("%d", sectorNumber))
+	requirements.NoError(os.Mkdir(cacheDirPath, 0755))
```
rvagg, line 204: try removing?

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+	req := require.New(tm.t)
+	sectorNumber := tm.currentSectorNum
+	tm.currentSectorNum++
+
+	// --------------------Create pre-commit for the CC sector -> we'll just pre-commit `sector size` worth of 0s for this CC sector
+
+	// Step 1: Wait for the pre-commitseal randomness to be available (we can only draw seal randomness from tipsets that have already achieved finality)
+	preCommitSealRand := tm.waitPreCommitSealRandomness(ctx, sectorNumber)
+
+	// Step 2: Write empty 32 bytes that we want to seal i.e. create our CC sector
```
rvagg, line 236: "CC" and 32 bytes? isn't this onboarding actual piece (non-CC) that's the size of unpadded 2KiB sector?

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+	tm.sealedSectorPaths[sectorNumber] = sealedSectorPath
+	tm.cacheDirPaths[sectorNumber] = cacheDirPath
+}
+
+func (tm *TestUnmanagedMiner) OnboardSectorWithPiecesAndRealProofs(ctx context.Context, proofType abi.RegisteredSealProof) (abi.SectorNumber, chan WindowPostResp) {
+	req := require.New(tm.t)
+	sectorNumber := tm.currentSectorNum
+	tm.currentSectorNum++
+
+	// --------------------Create pre-commit for the CC sector -> we'll just pre-commit `sector size` worth of 0s for this CC sector
```
rvagg, line 231: this function's docs all talk about "CC" but there's a piece in here, right?

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+
+	respCh := make(chan WindowPostResp, 1)
+
+	go tm.wdPostLoop(ctx, sectorNumber, respCh)
+
+	return sectorNumber, respCh
+}
+
+func (tm *TestUnmanagedMiner) wdPostLoop(ctx context.Context, sectorNumber abi.SectorNumber, respCh chan WindowPostResp) {
+	go func() {
```
rvagg, line 341: This goroutine is using `*testing.T`, which isn't safe unless you're only using `t.Error()`, which it's not (I think `t.Log*` is OK though). For a start, `tm.FullNode.WaitTillChain` uses `require.NoError` and all the `require.*` methods end in `t.FailNow` which isn't OK. The `assert.*` variants don't `t.FailNow` afaik though.

See https://pkg.go.dev/testing#T.FailNow

That's why I had all that awkward infrastructure to (a) send errors back to the main goroutine (although that may not have been necessary if it just used `assert.NoError` and friends, or just `t.Error` and then get the main thread to fail the test with channel-closed signal or something like that) and (b) increment the chain manually without `WaitTillChain`.

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+		case respCh <- WindowPostResp{SectorNumber: sectorNumber, Error: err}:
+		case <-ctx.Done():
+			return
+		default:
+		}
+	}()
+}
+
+func (tm *TestUnmanagedMiner) SubmitPostDispute(ctx context.Context, sectorNumber abi.SectorNumber) error {
+	tm.t.Logf("MinerB(%s): Starting dispute submission for sector %d", tm.ActorAddr, sectorNumber)
```
rvagg, line 367: `MinerB` all through this method

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+	enc, aerr := actors.SerializeParams(params)
+	require.NoError(tm.t, aerr)
+
+	_, err = tm.FullNode.MpoolPushMessage(ctx, &types.Message{
+		To:     tm.ActorAddr,
+		From:   tm.OwnerKey.Address,
+		Value:  types.FromFil(1),
+		Method: builtin.MethodsMiner.DisputeWindowedPoSt,
+		Params: enc,
+	}, nil)
```
rvagg, line 405: `submitMessage()` should be good for this, right?

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+
+	sectorProof, err := ffi.SealCommitPhase2(scp1, sectorNumber, actorId)
+	req.NoError(err)
+
+	tm.t.Logf("Miner %s: Got proof type %d sector proof of length %d for sector %d", tm.ActorAddr, proofType, len(sectorProof), sectorNumber)
+
+	return sectorProof
+}
+
+func (tm *TestUnmanagedMiner) submitMessage(
```
rvagg, line 740: I think this should be refactored to return an error rather than using `tm.t` because it's used inside the PoST goroutine

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +60,730 @@ type TestUnmanagedMiner struct {
...
+	msg, err := tm.FullNode.StateWaitMsg(ctx, m.Cid(), 2, api.LookbackNoLimit, true)
+	require.NoError(tm.t, err)
+	tm.t.Logf("Message with CID: %s has been confirmed on-chain for miner %s", m.Cid(), tm.ActorAddr)
+
+	return msg
+}
+
+func requireTempFile(t *testing.T, fileContentsReader io.Reader, size uint64) *os.File {
+	// Create a temporary file
+	tempFile, err := os.CreateTemp("", "")
```
rvagg, line 771: these don't get cleaned up by the testing infra because they're not done inside a temp dir made with `t.TempDir`; I think we either need to pass in a temp dir, or make a new one here with `t.TempDir` and put it inside of that so we get proper cleanup

itests/kit/node_opts.go
rvagg: I'm a little concerned about the complete removal of mock tests from this because there's so many places where you need to bypass ffi calls when you are running with mock proofs; I'll trust that you've got that under control but it is harder now the logic has been spread around a bit more.

itests/kit/node_unmanaged.go
aarshkshah1992: @rvagg The tests fail for me without this. I'm surprised it works for you. Are you sure you don't have a `mkDir` anywhere ? 

```
    node_unmanaged.go:636: Miner f01003: Running proof type 5 SealPreCommitPhase1 for sector 101...
    node_unmanaged.go:652: 
        	Error Trace:	/Users/aarshshah/src/filecoin/lotus/itests/kit/node_unmanaged.go:652
        	            				/Users/aarshshah/src/filecoin/lotus/itests/kit/node_unmanaged.go:240
        	            				/Users/aarshshah/src/filecoin/lotus/itests/manual_onboarding_test.go:77
        	            	No such file or directory (os error 2)
        	Test:       	TestManualCCOnboarding
        	Messages:   	Miner f01003: SealPreCommitPhase1 failed for sector 101
```

itests/kit/node_unmanaged.go
aarshkshah1992: Test fails. See my comment above.

itests/kit/node_unmanaged.go
aarshkshah1992: Fixed this by creating a `WaitTillChainOrError` function that propagates the error back to the caller

itests/kit/node_unmanaged.go
aarshkshah1992: Fixed this.

itests/kit/node_unmanaged.go
aarshkshah1992: Fixed all these log lines.

itests/kit/node_unmanaged.go
aarshkshah1992: Fixed.

itests/kit/node_unmanaged.go
aarshkshah1992: Fixed.

itests/kit/node_unmanaged.go
aarshkshah1992: Fixed.

itests/kit/node_unmanaged.go
aarshkshah1992: Fixed.

itests/kit/node_unmanaged.go
aarshkshah1992: Fixed this. Also made `submitMessage` propagate the error back to the caller.

itests/kit/node_unmanaged.go
aarshkshah1992: Done.

itests/kit/node_unmanaged.go
rvagg: oh, maybe it does then, I kept a mkdir in my local version but I thought at one stage I saw it doing it itself without needing me to, probably mistaken

itests/kit/node_unmanaged.go
```diff
@@ -24,3 +59,848 @@ type TestUnmanagedMiner struct {
...
+	tm.proofType[sectorNumber] = proofType
+
+	respCh := make(chan WindowPostResp, 1)
+
+	go tm.wdPostLoop(ctx, sectorNumber, respCh, false)
+
+	return sectorNumber, respCh
+}
+
+func (tm *TestUnmanagedMiner) wdPostLoop(ctx context.Context, sectorNumber abi.SectorNumber, respCh chan WindowPostResp, withMockProofs bool) {
```
rvagg, line 453: this isn't a loop anymore, it just does one post then exits

which is going to be a problem because we're going to have additions (including ni-porep) that will take longer to onboard so existing sectors need to continue to post

which has implications for other changes in here -- the channel has a buffer of 1, but what if we just want to make it continue to post without waiting for a response? I think my original version had a channel for _first_ post because that's the only one we actually care about so we can check power.

It also disputed all posts to make sure it did the right thing on each post. I don't think that's as critical to keep in the loop, but it was a nice feature that you always got an assertion that your post was valid without having to include additional logic. The only catch with that is when you want to assert that a post _should not_ be valid, but that could also just be a boolean "expect valid" (to check each time and expect a failure to dispute) or "submit dispute" (to check each time or not check at all and make it the caller's responsibility).

I was just doing this last week with ni-porep, experimenting with a test case where the post was invalid and making sure the dispute didn't error to make sure that the dispute mechanism was working.

**However** I'm actually not sure about the value of adding tests for disputing succeeding, because we have itests/wdpost_dispute_test.go already. I don't know that ni-porep adds anything novel that would need to add a new dispute check on windowpost, we'd need to check on that. But I still think it's nice to add a dispute mechanism when we're not using fake proofs, even if we can't flip between "expect valid" and "expect invalid".

rvagg:
I've gone ahead and squashed this down and included it in #12017 and also rebased that branch on the current https://github.com/filecoin-project/lotus/pull/12005 which has master in it, so we don't get too far out on a limb! We can keep iterating from there rather than having these big chains of PRs and commits.

PR #12062, opened 2024-05-31
refactor(eth): attach ToFilecoinMessage converter to EthCall type

raulk:
Small refactor to clean things up.

* This converter is better placed as an EthCall method instead of a package-level function.
* The resulting package/module import graph is sane.
* Exporting this as a public method is very helpful for those who use Lotus as a library.

### DIFF
--- a/chain/types/ethtypes/eth_types.go
+++ b/chain/types/ethtypes/eth_types.go
@@ -23,6 +23,8 @@ import (
 	builtintypes "github.com/filecoin-project/go-state-types/builtin"
 
 	"github.com/filecoin-project/lotus/build"
+	"github.com/filecoin-project/lotus/chain/actors"
+	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/lib/must"
 )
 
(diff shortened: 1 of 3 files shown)


Stebalien (approved):

Stebalien:
Er, sorry, I didn't see the comment from @rjan90.

Stebalien:
Test failure looks real.

raulk:
@rjan90 thanks for running make! ❤️ 
@Stebalien -- yep, it looks real. I'm investigating.

rjan90:
Hi @raulk, I'm considering whether it's still worthwhile to finalize this change, or if this PR should get moved back to draft/be closed. I can't recall the specific test that was failing, but it appears to have been resolved, as all checks are currently passing. What do you think?

rvagg:
👍 this would be a good change; unfortunately things have moved around a bit, the original function being refactored is now in node/impl/eth/utils.go, so this would need a little bit of work.

rjan90:
I've reopened the pull request with the changes here: https://github.com/filecoin-project/lotus/pull/12844. Starting from the master branch seemed to be the simplest approach.

rjan90:
Superseded by: https://github.com/filecoin-project/lotus/pull/12844

PR #12063, opened 2024-06-03
feat: test: MineBlocksMustPost can watch for >1 miners

rvagg:
@aarshkshah1992 this is on top of the unmanaged miner work in #12017, There's 2 things in here:

1. `MineBlocksMustPost` is now able to watch for posts from more than just itself. So when we add UnmanagedMiners we can add them to the list of miners it checks posts for. It just looks for whichever is the earliest deadline end and then makes sure there is a post in there for the sectors (most will have zero sectors).
2. Make the UnmanagedMiners continue to post until context cancel, so we won't run into trouble with them missing post. This also means I've changed the channel return struct to have a bool for first-post. It only sends on the channel if it's first post _or_ there's an error. It also means that an error may not go noticed by the main test function but I think that might be OK after the first post? It'll get logged anyway. We could add a channel slurp at the end of the test to make sure we don't miss anything on each of these. I had that in my original test.

### DIFF
--- a/itests/kit/blockminer.go
+++ b/itests/kit/blockminer.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-jsonrpc"
 	"github.com/filecoin-project/go-state-types/abi"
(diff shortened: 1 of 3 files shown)


rvagg:
I think an additional improvement can be made here to get started on the second miner as soon as we've onboarded the first's sector and then asserting power after both have activated. The original test tried to do something like this. `activateAndAssertPower` holds us up until miner B is proven before we even get started on miner C. But we can do them in parallel and only check at the end.

rvagg:
I was able to just move minerB's `waitTillActivatedAndAssertPower` down to just before minerC's and it works fine because they just wait for the channel and we now have first-post-only and a channel buffer of 1 so it's safe to defer checking. Also did some other minor fixups and cranked the blocktime back up to 2ms so it should be much faster now.

I've also marked the `WithRealProofs` test as `Expensive` so it'll be skipped unless `CI=true` or `LOTUS_RUN_EXPENSIVE_TESTS=1`.

PR #12065, opened 2024-06-04
src: lint: bump golangci-lint to 1.59, address unchecked fmt.Fprint*

rvagg:
Another attempt at bumping the linter.

In 1.59, they removed a blanket `Ignore: "fmt:.*",` and instead now rely on a more targeted ignore set of rules based on the types that `fmt.Fprint*()` prints into. The main reason this bites us hard here is because when we print from `urfave/cli` `Command` runners, we use the built-in context `AppWriter` and `ErrWriter`. `AppWriter` is _usually_ `os.Stdout` which is not excluded in the new rules, `ErrWriter` is _usually_ `os.Stderr` but it could also be something else and the new rules can't differentiate and neither can we.

I'm fine with the more targeted ruleset. It means having to opt-in to ignores if you have good reason to. Mostly I've gone with `_, _ = ` ignores, but there's a few places of blocks of `fmt.Fprintf` lines where I've gone with `//nolint:errcheck`.

In terms of landing this, it will have conflicts with the markets removal PRs, notably there's a big change set in paych.go which is disappearing with those PRs. This PR isn't urgent and it may be easier to rebase this on those changes once they land than the other way around.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -57,7 +57,7 @@ jobs:
           submodules: 'recursive'
       - uses: ./.github/actions/install-system-dependencies
       - uses: ./.github/actions/install-go
-      - run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.58.2
+      - run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.0
       - run: make deps
       - run: golangci-lint run -v --timeout 10m --concurrency 4
   check-fmt:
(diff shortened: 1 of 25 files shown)


rjan90 (approved):

rjan90:
> In terms of landing this, it will have conflicts with the markets removal PRs, notably there's a big change set in paych.go which is disappearing with those PRs. This PR isn't urgent and it may be easier to rebase this on those changes once they land than the other way around.

Agree on landing this PR after the markets removal PR has landed.

PR #12066, opened 2024-06-04
Implement snap deals test for manual sector onboarding

aarshkshah1992:
## Related Issues
Implements a test for Snap Deals (by upgrading an active CC sector with actual pieces and then using DDO for sector upgrade) with real proofs.

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/kit/node_unmanaged.go
+++ b/itests/kit/node_unmanaged.go
@@ -349,6 +349,112 @@ func (tm *TestUnmanagedMiner) OnboardSectorWithPiecesAndMockProofs(ctx context.C
 	return sectorNumber, respCh
 }
 
+func (tm *TestUnmanagedMiner) mkStagedFileWithPieces(_ context.Context, sectorNumber abi.SectorNumber, pt abi.RegisteredSealProof) ([]abi.PieceInfo, string) {
+	paddedPieceSize := abi.PaddedPieceSize(tm.options.sectorSize)
+	unpaddedPieceSize := paddedPieceSize.Unpadded()
+
+	// Generate random bytes for the piece
+	randomBytes := make([]byte, unpaddedPieceSize)
+	_, err := io.ReadFull(rand.Reader, randomBytes)
+	require.NoError(tm.t, err)
+
+	// Create a temporary file for the first piece
+	pieceFileA := requireTempFile(tm.t, bytes.NewReader(randomBytes), uint64(unpaddedPieceSize))
+
+	// Generate the piece CID from the file
+	pieceCIDA, err := ffi.GeneratePieceCIDFromFile(pt, pieceFileA, unpaddedPieceSize)
+	require.NoError(tm.t, err)
+
+	// Reset file offset to the beginning after CID generation
+	_, err = pieceFileA.Seek(0, io.SeekStart)
+	require.NoError(tm.t, err)
+
+	unsealedSectorFile := requireTempFile(tm.t, bytes.NewReader([]byte{}), 0)
+	defer func() {
+		_ = unsealedSectorFile.Close()
+	}()
+
+	// Write the piece to the staged sector file without alignment
+	writtenBytes, pieceCID, err := ffi.WriteWithoutAlignment(pt, pieceFileA, unpaddedPieceSize, unsealedSectorFile)
+	require.NoError(tm.t, err)
+	require.EqualValues(tm.t, unpaddedPieceSize, writtenBytes)
+	require.True(tm.t, pieceCID.Equals(pieceCIDA))
+
+	// Create a struct for the piece info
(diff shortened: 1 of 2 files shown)


aarshkshah1992:
@rvagg The only thing missing here is submitting a WdPost after the sector snap proof is submitted. We will have to do some refactoring here to de-race the CC proof WdPost loop from the Snap WdPost loop. Is it needed/worth the complexity ? We've already tested that the WdPost for CC works and the snap proof is verified successfully by the chain here. Lemme know your thoughts.

itests/kit/node_unmanaged.go
```diff
@@ -344,11 +344,121 @@ func (tm *TestUnmanagedMiner) OnboardSectorWithPiecesAndMockProofs(ctx context.C
 
 	respCh := make(chan WindowPostResp, 1)
 
-	go tm.wdPostLoop(ctx, sectorNumber, respCh, true)
+	go tm.wdPostLoop(ctx, sectorNumber, respCh, true, tm.sealedCids[sectorNumber], tm.sealedSectorPaths[sectorNumber], tm.cacheDirPaths[sectorNumber])
 
 	return sectorNumber, respCh
 }
 
+func (tm *TestUnmanagedMiner) mkStagedFileWithPieces(_ context.Context, sectorNumber abi.SectorNumber, pt abi.RegisteredSealProof) ([]abi.PieceInfo, string) {
```
rvagg, line 352:
```suggestion
func (tm *TestUnmanagedMiner) mkStagedFileWithPieces(sectorNumber abi.SectorNumber, pt abi.RegisteredSealProof) ([]abi.PieceInfo, string) {
```

if we don't use it, how about we just remove it?

itests/kit/node_unmanaged.go
```diff
@@ -344,11 +344,121 @@ func (tm *TestUnmanagedMiner) OnboardSectorWithPiecesAndMockProofs(ctx context.C
...
+	require.NoError(tm.t, err)
+
+	randomBytes := make([]byte, s.Size())
+	_, err = io.ReadFull(rand.Reader, randomBytes)
+	require.NoError(tm.t, err)
+
+	updatePath := requireTempFile(tm.t, bytes.NewReader(randomBytes), uint64(s.Size()))
+	_ = updatePath.Close()
+	updateDir := filepath.Join(tm.t.TempDir(), fmt.Sprintf("update-%d", sectorNumber))
+	_ = os.MkdirAll(updateDir, 0700)
```
rvagg, line 407: can we check these two errors?
```suggestion
	require.NoError(tm.t, updatePath.Close())
	updateDir := filepath.Join(tm.t.TempDir(), fmt.Sprintf("update-%d", sectorNumber))
	require.NoError(tm.t, os.MkdirAll(updateDir, 0700))
```

itests/manual_onboarding_test.go
```diff
@@ -115,6 +115,21 @@ func TestManualCCOnboarding(t *testing.T) {
 			// Wait till both miners' sectors have had their first post and are activated and check that this is reflected in miner power
 			waitTillActivatedAndAssertPower(ctx, t, minerB, bRespCh, bSectorNum, uint64(defaultSectorSize), withMockProofs)
 			waitTillActivatedAndAssertPower(ctx, t, minerC, cRespCh, cSectorNum, uint64(defaultSectorSize), withMockProofs)
+
+			// Miner B has activated the CC sector -> upgrade it with snapdeals
+			// Note: We can't activate a sector with mock proofs as the WdPost is successfully disputed and so no point
+			// in snapping it as snapping is only for activated sectors
+			if !withMockProofs {
+				respCh := minerB.SnapDealWithRealProofs(ctx, kit.TestSpt, bSectorNum)
```
rvagg, line 123: We have a wdPostLoop running already for minerB don't we?

Maybe when we call `minerB.OnboardCCSectorWithRealProofs(ctx, kit.TestSpt)` in the test, we give it a cancellable context and then cancel that before we start a new loop. The tricky bit would be that we want the original post loop to continue right up until we submit the `ProveReplicaUpdates3`, and the timing between doing that and starting a new wdPostLoop is embedded in `SnapDealWithRealProofs`. It could be enough to pass in a `cancelPreviousPostLoop context.CancelFunc` to `SnapDealWithRealProofs` and get it to call that just after submitting the PRU3 and prior to starting a new loop? What do you think?

aarshkshah1992: As discussed offline, we're figuring out if we need to test WdPosts for snapped sectors explicitly.

rvagg (approved):
maybe we should go ahead and get these all on master so we can iterate from there; I'm happy with this and the stack we have as a starting point that we can mangle as we see the need when adding new pieces

so +1 from me to merge this and the parent PR

PR #12070, opened 2024-06-06
fix:ethapi:Fail fast if user attempts to send to 0x000..

ZenGround0:
## Proposed Changes

This is broken silently now and its not good for UX

## Additional Info

https://filecoinproject.slack.com/archives/CRK2LKYHW/p1717635018083559

## Checklist

- [] New features have usage guidelines and / or documentation updates in
- [ ] CI is green

### DIFF
--- a/chain/types/ethtypes/eth_transactions.go
+++ b/chain/types/ethtypes/eth_transactions.go
@@ -253,6 +253,10 @@ func getFilecoinMethodInfo(recipient *EthAddress, input []byte) (*methodInfo, er
 	var to address.Address
 	var method abi.MethodNum
 
+	var zeroEthAddr EthAddress
+	if recipient != nil && *recipient == zeroEthAddr {
+		return nil, fmt.Errorf("zero address recipient disallowed, please set `To` to null")
+	}
 	if recipient == nil {
 		// If recipient is nil, use Ethereum Address Manager Actor and CreateExternal method
 		method = builtintypes.MethodsEAM.CreateExternal


ZenGround0:
It would be better to actually make this work.  For now I'm making the implicit error (attempt to run invoke contract and get rekt) explicit.  Now users will see rpc server error instead of a weird version of success.

To actually make this work we'd need to mix up things in a difficult way.  There is no longer one canonical serialization of filecoin create external so we'd need to test out both against sig and go with the one that validates.

rvagg (approved):

ZenGround0:
Ok nice this fails explicitly now:
```
Error: 
server returned an error response: error code 1: failed to convert to unsigned msg: failed to get method info: zero address recipient disallowed, please set `To` to null
```
I'm going to try to get it working to handle the 0x000 case too.

chain/types/ethtypes/eth_transactions.go
```diff
@@ -253,6 +253,10 @@ func getFilecoinMethodInfo(recipient *EthAddress, input []byte) (*methodInfo, er
 	var to address.Address
 	var method abi.MethodNum
 
+	var zeroEthAddr EthAddress
+	if recipient != nil && *recipient == zeroEthAddr {
+		return nil, fmt.Errorf("zero address recipient disallowed, please set `To` to null")
+	}
```
Stebalien, line 259: This is used inside message validation (now) and is therefore consensus critical. Given that the user _can_ send to this address (it is a valid account), we're not allowed to do this.

Stebalien (changes requested):
Tooling _shouldn't_ send to 0x0 when creating contracts. if you're seeing that, something is very wrong somewhere else in the API.

Stebalien:
Yeah, I've confirmed:

1. Sends to 0x0 are _valid_.
2. Contract creation transactions send to RLP-encoded `nil` (0x80).

If we're somehow translating this `nil` to `0x0000000...`, we have a bigger problem.

ZenGround0:
I'll keep trying to understand where the 0x0 is coming from.  Hopefully just the tool.

chain/types/ethtypes/eth_transactions.go
ZenGround0: FYI @rvagg this was a pretty bad mistake on my part.  If I'd merged this and we hadn't thought more about it we would have opened up a fork.  So I guess the lesson for you here is to trust me less.

PR #12071, opened 2024-06-06
fix: manual onbaording itest: do not sumit replica update messages during immutable deadlines 

aarshkshah1992:
This fixes a bug in the Manual Onboarding Snapdeals itest where we used to submit Snapdeals messages during immutable deadlines.

See https://github.com/filecoin-project/builtin-actors/blob/6d0725aabbb3ecc35e02cbcd20036901924411e2/actors/miner/src/lib.rs#L3759.

And https://github.com/filecoin-project/lotus/blob/730c96ecafe28cfdffa46c13a64ed98e053fc24f/storage/pipeline/states_replica_update.go#L254


rvagg (approved):
ok, I think I understand the actors logic in this - you have to be at least one challenge window _before_ the proving window for the sector

so I guess this logic works

Issue #11640, opened 2024-02-22
Record processed epochs / tipsets in events index


rvagg:
Arising out of https://github.com/filecoin-project/lotus/pull/11618 - Lotus currently has no way to know whether the earliest events it has are the earliest there are or the earliest that that have been recorded. Ideally requesting "events from height X" should error if we can say for sure "I don't have events for that height", but that's not possible today because we don't know.

* [ ] Add a minor migration to the event index database with a new table to record "height" once it's been fully processed, the migration could either walk back through indexed events and mark those as done, or to be more comprehensive, walk back through the message receipts and mark them off?
* [ ] Error on the `GetActorEvents` and `SubscribeActorEvents` APIs if you request a height that's earlier than we have.

Stebalien:
In terms of the migration... I believe that checking if we have _any_ events at a tipset is relatively inexpensive, and we only need to do that for tipsets that don't have recorded events.

jennijuju:
@Stebalien iirc we want this for the message index too, is that right (when we were discussing tracing api)?

Stebalien:
Yes.

masih:
@rvagg please feel free to assign me other issues that relate to this, considering the need for migration. That should reduce the need for coordination.

Thanks

Issue #12078, opened 2024-06-10
getLogs seems to be missing events when run at head


ashwinphatak:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I did not make any code changes to lotus.

### Lotus component

- [X] lotus Ethereum RPC
- [ ] lotus FVM - Lotus FVM interactions
- [ ] FEVM tooling
- [ ] Other

### Lotus Version

```text
Daemon:  1.27.0-dev+mainnet+git.20640c062+api1.5.0
Local: lotus version 1.27.0-dev+mainnet+git.20640c062
```

### Repro Steps

We don't yet have steps to reproduce. The issue showed up in long running downstream processes and only seems to happen once in a while.

### Describe the Bug

So far we've found at least two instances when lotus didn't return matching events during a getLogs RPC call at head (https://filfox.info/en/tipset/3976539). The same calls returned data when made later (couple of days). These missing logs were only discovered in retrospect because the downstream service started throwing errors when processing later events. The downstream service code remained unchanged during these tests.

This screenshot shows a db table during the original run, during which the block which was processed returned no matching events:

![image](https://github.com/filecoin-project/lotus/assets/14833/1c94fe6e-b32d-4f76-99c0-ac2637b2590c)

This screenshot shows the result of running the same getLogs call after a couple of days:

![image](https://github.com/filecoin-project/lotus/assets/14833/d2d4cd82-c647-40ee-a4d8-f67ad27b8357)

To confirm this issue exists with getLogs only at HEAD, we've left the service running with changes to stay 16 blocks behind head at all times. It's been running for more than two days so far without similar problems.

### Tooling

```text
We're using etherjs to make RPC calls
```

### Configuration Options

```text
config.toml


[Fevm]
  # EnableEthRPC enables eth_ rpc, and enables storing a mapping of eth transaction hashes to filecoin message Cids.
  # This will also enable the RealTimeFilterAPI and HistoricFilterAPI by default, but they can be disabled by config options above.
  #
  # type: bool
  # env var: LOTUS_FEVM_ENABLEETHRPC
  #EnableEthRPC = false

[...]

PR #12080, opened 2024-06-12
fix: events index: record processed epochs and tipsets for events and eth_get_log blocks till requested tipset has been indexed

aarshkshah1992:
Closes https://github.com/filecoin-project/lotus/issues/11640 and #12078

### DIFF
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -61,9 +61,13 @@ var ddls = []string{
 		value BLOB NOT NULL
 	)`,
 
+	createTableEventsSeen,
+
 	createIndexEventEntryIndexedKey,
 	createIndexEventEntryCodecValue,
 	createIndexEventEntryEventId,
+	createIndexEventsSeenHeight,
+	createIndexEventsSeenTipsetKeyCid,
 
 	// metadata containing version of schema
 	`CREATE TABLE IF NOT EXISTS _meta (
@@ -75,6 +79,7 @@ var ddls = []string{
 	`INSERT OR IGNORE INTO _meta (version) VALUES (3)`,
 	`INSERT OR IGNORE INTO _meta (version) VALUES (4)`,
 	`INSERT OR IGNORE INTO _meta (version) VALUES (5)`,
+	`INSERT OR IGNORE INTO _meta (version) VALUES (6)`,
 }
 
 var (
@@ -82,13 +87,14 @@ var (
 )
 
 const (
-	schemaVersion = 5
+	schemaVersion = 6
 
 	eventExists          = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
 	insertEvent          = `INSERT OR IGNORE INTO event(height, tipset_key, tipset_key_cid, emitter_addr, event_index, message_cid, message_index, reverted) VALUES(?, ?, ?, ?, ?, ?, ?, ?)`
 	insertEntry          = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
 	revertEventsInTipset = `UPDATE event SET reverted=true WHERE height=? AND tipset_key=?`
 	restoreEvent         = `UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
+	insertEventsSeen     = `INSERT OR IGNORE INTO events_seen(height, tipset_key, tipset_key_cid) VALUES(?, ?, ?)`
 
 	createIndexEventEmitterAddr  = `CREATE INDEX IF NOT EXISTS event_emitter_addr ON event (emitter_addr)`
 	createIndexEventTipsetKeyCid = `CREATE INDEX IF NOT EXISTS event_tipset_key_cid ON event (tipset_key_cid);`
@@ -98,6 +104,16 @@ const (
 	createIndexEventEntryIndexedKey = `CREATE INDEX IF NOT EXISTS event_entry_indexed_key ON event_entry (indexed, key);`
 	createIndexEventEntryCodecValue = `CREATE INDEX IF NOT EXISTS event_entry_codec_value ON event_entry (codec, value);`
 	createIndexEventEntryEventId    = `CREATE INDEX IF NOT EXISTS event_entry_event_id ON event_entry(event_id);`
+
+	createTableEventsSeen = `CREATE TABLE IF NOT EXISTS events_seen (
+		id INTEGER PRIMARY KEY,
+		height INTEGER NOT NULL,
+		tipset_key BLOB NOT NULL,
+		tipset_key_cid BLOB NOT NULL
+	)`
+
+	createIndexEventsSeenHeight       = `CREATE INDEX IF NOT EXISTS events_seen_height ON events_seen (height);`
+	createIndexEventsSeenTipsetKeyCid = `CREATE INDEX IF NOT EXISTS events_seen_tipset_key_cid ON events_seen (tipset_key_cid);`
 )
 
 type EventIndex struct {
@@ -108,6 +124,7 @@ type EventIndex struct {
 	stmtInsertEntry          *sql.Stmt
 	stmtRevertEventsInTipset *sql.Stmt
 	stmtRestoreEvent         *sql.Stmt
+	stmtInsertEventsSeen     *sql.Stmt
 }
 
 func (ei *EventIndex) initStatements() (err error) {
@@ -134,6 +151,11 @@ func (ei *EventIndex) initStatements() (err error) {
 	ei.stmtRestoreEvent, err = ei.db.Prepare(restoreEvent)
 	if err != nil {
 		return xerrors.Errorf("prepare stmtRestoreEvent: %w", err)
+	}
+
+	ei.stmtInsertEventsSeen, err = ei.db.Prepare(insertEventsSeen)
+	if err != nil {
+		return xerrors.Errorf("prepare stmtInsertEventsSeen: %w", err)
 	}
 
 	return nil
@@ -407,6 +429,58 @@ func (ei *EventIndex) migrateToVersion5(ctx context.Context) error {
 	return nil
 }
 
+func (ei *EventIndex) migrateToVersion6(ctx context.Context) error {
+	now := time.Now()
+
+	tx, err := ei.db.BeginTx(ctx, nil)
+	if err != nil {
+		return xerrors.Errorf("begin transaction: %w", err)
+	}
+	defer func() { _ = tx.Rollback() }()
+
+	stmtCreateTableEventsSeen, err := tx.PrepareContext(ctx, createTableEventsSeen)
+	if err != nil {
+		return xerrors.Errorf("prepare stmtCreateTableEventsSeen: %w", err)
+	}
+	_, err = stmtCreateTableEventsSeen.ExecContext(ctx)
+	if err != nil {
+		return xerrors.Errorf("create table events_seen: %w", err)
+	}
+
+	_, err = tx.ExecContext(ctx, createIndexEventsSeenHeight)
+	if err != nil {
+		return xerrors.Errorf("create index events_seen_height: %w", err)
+	}
+	_, err = tx.ExecContext(ctx, createIndexEventsSeenTipsetKeyCid)
+	if err != nil {
+		return xerrors.Errorf("create index events_seen_tipset_key_cid: %w", err)
+	}
+
+	// INSERT an entry in the events_seen table for all epochs we do have have events for in our DB
+	_, err = tx.ExecContext(ctx, `
+    INSERT INTO events_seen (height, tipset_key, tipset_key_cid)
+    SELECT e.height, e.tipset_key, e.tipset_key_cid
+    FROM event e
+    WHERE NOT EXISTS (
+        SELECT 1 FROM events_seen es
+        WHERE es.height = e.height AND es.tipset_key = e.tipset_key AND es.tipset_key_cid = e.tipset_key_cid
+    )
+`)
+	if err != nil {
+		return xerrors.Errorf("insert events into events_seen: %w", err)
+	}
+
+	err = tx.Commit()
+	if err != nil {
+		return xerrors.Errorf("commit transaction: %w", err)
+	}
+
+	ei.vacuumDBAndCheckpointWAL(ctx)
+
+	log.Infof("Successfully migrated event index from version 4 to version 5 in %s", time.Since(now))
+	return nil
+}
+
 func (ei *EventIndex) vacuumDBAndCheckpointWAL(ctx context.Context) {
 	// During the large migrations, we have likely increased the WAL size a lot, so lets do some
 	// simple DB administration to free up space (VACUUM followed by truncating the WAL file)
@@ -498,6 +572,16 @@ func NewEventIndex(ctx context.Context, path string, chainStore *store.ChainStor
 			version = 5
 		}
 
+		if version == 5 {
+			log.Infof("Upgrading event index from version 5 to version 6")
+			err = eventIndex.migrateToVersion6(ctx)
+			if err != nil {
+				_ = db.Close()
+				return nil, xerrors.Errorf("could not migrate event index schema from version 5 to version 6: %w", err)
+			}
+			version = 6
+		}
+
 		if version != schemaVersion {
 			_ = db.Close()
 			return nil, xerrors.Errorf("invalid database version: got %d, expected %d", version, schemaVersion)
@@ -518,6 +602,13 @@ func (ei *EventIndex) Close() error {
 		return nil
 	}
 	return ei.db.Close()
+}
+
+func (ei *EventIndex) isTipsetProcessed(ctx context.Context, tipsetKey []byte) (bool, error) {
+	row := ei.db.QueryRowContext(ctx, "SELECT COUNT(*) > 0 FROM events_seen WHERE tipset_key = ?", tipsetKey)
+	var exists bool
+	err := row.Scan(&exists)
+	return exists, err
 }
 
 func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, revert bool, resolver func(ctx context.Context, emitter abi.ActorID, ts *types.TipSet) (address.Address, bool)) error {
@@ -550,6 +641,10 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 	if err != nil {
 		return xerrors.Errorf("load executed messages: %w", err)
 	}
+	tsKeyCid, err := te.msgTs.Key().Cid()
+	if err != nil {
+		return xerrors.Errorf("tipset key cid: %w", err)
+	}
 
 	eventCount := 0
 	// iterate over all executed messages in this tipset and insert them into the database if they
@@ -565,11 +660,6 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 					continue
 				}
 				addressLookups[ev.Emitter] = addr
-			}
-
-			tsKeyCid, err := te.msgTs.Key().Cid()
-			if err != nil {
-				return xerrors.Errorf("tipset key cid: %w", err)
 			}
 
 			// check if this event already exists in the database
@@ -649,6 +739,16 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 			}
 			eventCount++
 		}
+	}
+
+	// add an entry to the event_seen table for this tipset
+	_, err = tx.Stmt(ei.stmtInsertEventsSeen).Exec(
+		te.msgTs.Height(),
+		te.msgTs.Key().Bytes(),
+		tsKeyCid.Bytes(),
+	)
+	if err != nil {
+		return xerrors.Errorf("exec insert events seen: %w", err)
 	}
 
 	err = tx.Commit()
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -97,7 +97,7 @@ const (
 	restoreEvent         = `UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
 	revertEventSeen      = `UPDATE events_seen SET reverted=true WHERE height=? AND tipset_key_cid=?`
 	restoreEventSeen     = `UPDATE events_seen SET reverted=false WHERE height=? AND tipset_key_cid=?`
-	insertEventsSeen     = `INSERT OR IGNORE INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, ?)`
+	upsertEventsSeen     = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, ?) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
 	isTipsetProcessed    = `SELECT COUNT(*) > 0 FROM events_seen WHERE tipset_key_cid=?`
 	getMaxHeightInIndex  = `SELECT MAX(height) FROM events_seen`
 
@@ -129,7 +129,7 @@ type EventIndex struct {
 	stmtInsertEntry          *sql.Stmt
 	stmtRevertEventsInTipset *sql.Stmt
 	stmtRestoreEvent         *sql.Stmt
-	stmtInsertEventsSeen     *sql.Stmt
+	stmtUpsertEventsSeen     *sql.Stmt
 	stmtRevertEventSeen      *sql.Stmt
 	stmtRestoreEventSeen     *sql.Stmt
 
@@ -175,9 +175,9 @@ func (ei *EventIndex) initStatements() (err error) {
 		return xerrors.Errorf("prepare stmtRestoreEvent: %w", err)
 	}
 
-	ei.stmtInsertEventsSeen, err = ei.db.Prepare(insertEventsSeen)
+	ei.stmtUpsertEventsSeen, err = ei.db.Prepare(upsertEventsSeen)
 	if err != nil {
-		return xerrors.Errorf("prepare stmtInsertEventsSeen: %w", err)
+		return xerrors.Errorf("prepare stmtUpsertEventsSeen: %w", err)
 	}
 
 	ei.stmtRevertEventSeen, err = ei.db.Prepare(revertEventSeen)
@@ -498,13 +498,8 @@ func (ei *EventIndex) migrateToVersion6(ctx context.Context) error {
 
 	// INSERT an entry in the events_seen table for all epochs we do have events for in our DB
 	_, err = tx.ExecContext(ctx, `
-    INSERT INTO events_seen (height, tipset_key_cid, reverted)
-    SELECT e.height, e.tipset_key_cid, e.reverted
-    FROM event e
-    WHERE NOT EXISTS (
-        SELECT 1 FROM events_seen es
-        WHERE es.height = e.height AND es.tipset_key_cid = e.tipset_key_cid	
-    )
+    INSERT OR IGNORE INTO events_seen (height, tipset_key_cid, reverted)
+    SELECT DISTINCT height, tipset_key_cid, reverted FROM event
 `)
 	if err != nil {
 		return xerrors.Errorf("insert events into events_seen: %w", err)
@@ -766,7 +761,6 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 	}
 
 	eventCount := 0
-	isTipSetRevertRestored := false
 	// iterate over all executed messages in this tipset and insert them into the database if they
 	// don't exist, otherwise mark them as not reverted
 	for msgIdx, em := range ems {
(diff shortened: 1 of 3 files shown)


chain/events/filter/index.go
```diff
@@ -75,20 +79,22 @@ var ddls = []string{
...
 const (
-	schemaVersion = 5
+	schemaVersion = 6
 
 	eventExists          = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
 	insertEvent          = `INSERT OR IGNORE INTO event(height, tipset_key, tipset_key_cid, emitter_addr, event_index, message_cid, message_index, reverted) VALUES(?, ?, ?, ?, ?, ?, ?, ?)`
 	insertEntry          = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
 	revertEventsInTipset = `UPDATE event SET reverted=true WHERE height=? AND tipset_key=?`
 	restoreEvent         = `UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
+	insertEventsSeen     = `INSERT OR IGNORE INTO events_seen(height, tipset_key, tipset_key_cid) VALUES(?, ?, ?)`
```
Stebalien, line 97: Any reason we're storing both the tipset key and cid? I guess this applies to the events table as well.

chain/events/filter/index.go
```diff
@@ -520,6 +609,13 @@ func (ei *EventIndex) Close() error {
 	return ei.db.Close()
 }
 
+func (ei *EventIndex) isTipsetProcessed(ctx context.Context, tipsetKey []byte) (bool, error) {
+	row := ei.db.QueryRowContext(ctx, "SELECT COUNT(*) > 0 FROM events_seen WHERE tipset_key = ?", tipsetKey)
```
Stebalien, line 613: I think we need an additional "reverted" flag if we're going to use this table to fix https://github.com/filecoin-project/lotus/issues/11640. The issue here is:

1. Apply TipSet A at epoch X (is process returns false).
2. Switch to TipSet B at epoch X (is processed returns false).
4. Switch back to TipSetA (is processed returns _true_, but it's not).

When I discussed this with @rvagg the plan was (IIRC) to record:

1. The earliest processed tipset/epoch for backfilling.
2. The last processed tipset (and _only_ the last processed one).

When "catching up", you'd always look at the "path" between the last processed tipset and the current tipset (taking reverts into account).

But that doesn't solve your `getLogs` case. If a user asks for a historical "uncle", there's no easy way to tell if that uncle has been processed (you don't know what chain it's on).

So, I think the solution is to add a reverted flag so, on start, we can find the latest non-reverted tipset and "catch up" from there.

---

Note: We don't have to do this now, but I'd like to clean all this up a bit:

1. Remove the reverted flag from the `events` table, that'll be recorded in the `events_seen` table (maybe call this `tipsets`?).
2. Remove the `tipset_cid`/`tipset_key` from the `events` table as well and reference the correct entry in the `tipsets` table by foreign key.

It makes some of the queries more complex, but the events table is getting fat and we're doing more work than necessary to process reverts.

Stebalien (commented):
I have some opinions but this really needs a look by @rvagg (who has done more thinking on this than I have at this point).

chain/events/filter/index.go
```diff
@@ -407,6 +429,63 @@ func (ei *EventIndex) migrateToVersion5(ctx context.Context) error {
...
+	}
+
+	// INSERT an entry in the events_seen table for all epochs we do have events for in our DB
+	_, err = tx.ExecContext(ctx, `
+    INSERT INTO events_seen (height, tipset_key, tipset_key_cid)
+    SELECT e.height, e.tipset_key, e.tipset_key_cid
+    FROM event e
+    WHERE NOT EXISTS (
+        SELECT 1 FROM events_seen es
+        WHERE es.height = e.height AND es.tipset_key = e.tipset_key AND es.tipset_key_cid = e.tipset_key_cid
```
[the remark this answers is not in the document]
rvagg: an alternative would be to find the min and max and fill in all between those two as being seen, that might yield a a state that's much more likely to be valid; there's no good reason to have blanks between min and max, but there are good reasons to have missing events, (this can been seen particularly on calibnet)

chain/events/filter/index.go
```diff
@@ -407,6 +429,63 @@ func (ei *EventIndex) migrateToVersion5(ctx context.Context) error {
...
+	if err != nil {
+		return xerrors.Errorf("increment _meta version: %w", err)
+	}
+
+	err = tx.Commit()
+	if err != nil {
+		return xerrors.Errorf("commit transaction: %w", err)
+	}
+
+	ei.vacuumDBAndCheckpointWAL(ctx)
```
rvagg, line 483: this needs to be removed from `migrateToVersion5`, we bump it to the last migration so if you're doing >1 migrations we don't do this more than once (and yes, we should do this better now we have so many migrations but it works for now)

chain/events/filter/index.go
rvagg: This is a good question for the rest of the schema actually, and @aarshkshah1992 is just conforming to what's there - we keep both in `event`, we insert them both with new events and we query matching them both for "exists" and when we do a "restore" operation. Maybe we should just ditch tipset_key and only ever use tipset_key_cid? I'm not seeing what value-add both are giving us in here.

chain/events/filter/index.go
rvagg: I'm not sure we need a "reverted" column specifically for this use-case, we just need to know whether a particular tipset has been inserted or not. Even if it flips from applied to reverted and back to applied again, the tipset doesn't change so it's events won't change. What am I not seeing that #11640 would need a "reverted" @Stebalien? It makes sense if we have a "tipset" table where we store all the tipset things, but that's probably too much scope for this.

chain/events/filter/index.go
rvagg: @aarshkshah1992 can you look at the possibility of removing the key and just using the CID for this particular case, we can deal with the other instances later I think.

chain/events/filter/index.go
Stebalien: You're right, it fixes the main issue but not my followup comment (https://github.com/filecoin-project/lotus/issues/11640#issuecomment-1964955238).

1. Main issue: Sometimes a user will call `getLogs` on a tipset we haven't processed yet. This PR fixes that case (or works towards fixing it).
2. Followup comment: Sometimes we'll shut down before indexing some tipset and, when we restart, we need to "catch up" and index any missing tipsets. In this case, my concern is that we'll shutdown right after step 3 above ("Switch back to TipSetA"), restart, and think "oh, I see we've already indexed TipSetA". But we _actually_ need to go and un-mark TipSetA as reverted and mark TipSetB as reverted instead.

chain/events/filter/index.go
aarshkshah1992: Changed this to just use the CID. Whenever we have they key, we also have the CID ofcourse.

chain/events/filter/index.go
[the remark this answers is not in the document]
aarshkshah1992: @rvagg min and max of what ? The problem is that for epochs that do not have events, we simply do not have any entry for them in any of our event tabls.

chain/events/filter/index.go
aarshkshah1992: Fixed this.

chain/events/filter/index.go
aarshkshah1992: I think I agree with Steb.

Being able to answer "What's the latest tipset for which you have events on the cannonical chain" is important. Might even be for this work itself. Still working through it. But have made this change.

chain/events/filter/index.go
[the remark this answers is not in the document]
rvagg: I was thinking min and max height; start from `MIN(height)` and add an entry for every one until `MAX(height)` .. but I suppose you need tipset_key_cid for each of those too so you'd have to go looking for it - but this could become a chain walk operation too -- find min and max, walk chain between and add an entry for each one.

Maybe this isn't worth thinking too hard about since it's a migration but I _think_ doing a walk and including _every_ epoch in that range is much more likely to be accurate than just filling in the ones we have events for. Because we should have an entry in the table even for tipsets that have no events because that's part of what we're solving here - no events in the tipset, vs no events in the database but maybe tipset has events?

chain/events/filter/index.go
aarshkshah1992: This is done.

chain/events/filter/index.go
aarshkshah1992: This is done.

chain/events/filter/index.go
```diff
@@ -520,6 +640,61 @@ func (ei *EventIndex) Close() error {
...
+func (ei *EventIndex) IsHeightProcessed(ctx context.Context, height uint64) (bool, error) {
+	mh, err := ei.GetMaxHeightInIndex(ctx)
+	if err != nil {
+		return false, err
+	}
+	return height <= mh, nil
+}
+
+func (ei *EventIndex) IsTipsetProcessed(ctx context.Context, tipsetKeyCid []byte) (bool, error) {
+	row := ei.db.QueryRowContext(ctx, "SELECT COUNT(*) > 0 FROM events_seen WHERE tipset_key_cid = ?", tipsetKeyCid)
```
rvagg, line 692: this should probably be a prepared statement

chain/events/filter/index.go
```diff
@@ -520,6 +640,61 @@ func (ei *EventIndex) Close() error {
 	return ei.db.Close()
 }
 
+func (ei *EventIndex) SubscribeUpdates() (uint64, chan *EventIndexUpdate) {
```
rvagg, line 643: perhaps you should just return an `Unsubscribe()` or `Cancel()` function from this instead of a uint64 so you don't need a second method for `UnsubscribeUpdates`, and you also don't allow for one caller to cancel the subscription of another

chain/events/filter/index.go
```diff
@@ -530,16 +705,45 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
...
+			return xerrors.Errorf("revert event seen: %w", err)
+		}
+
 		err = tx.Commit()
 		if err != nil {
 			return xerrors.Errorf("commit transaction: %w", err)
 		}
 
+		ei.mu.Lock()
+		tSubs := ei.updateSubs
```
rvagg, line 729: this isn't a full copy, you'll need to go further than this to be protected

```suggestion
		tSubs := make([]int, len(ei.updateSubs))
		copy(tSubs, ei.updateSubs)
```

chain/events/filter/index.go
```diff
@@ -108,6 +127,25 @@ type EventIndex struct {
...
+}
+
+type updateSub struct {
+	ctx    context.Context
+	ch     chan *EventIndexUpdate
+	cancel context.CancelFunc
+}
+
+type EventIndexUpdate struct {
+	Height    uint64
```
rvagg, line 146: this should probably be `abi.ChainEpoch`, I see some casting going on that would be redundant here

chain/events/filter/index.go
```diff
@@ -108,6 +127,25 @@ type EventIndex struct {
...
+	TipsetCid cid.Cid
```
rvagg, line 147:
```suggestion
	TipSetKeyCid cid.Cid
```

chain/events/filter/index.go
```diff
@@ -550,6 +754,10 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 	if err != nil {
 		return xerrors.Errorf("load executed messages: %w", err)
 	}
+	tsKeyCid, err := te.msgTs.Key().Cid()
+	if err != nil {
+		return xerrors.Errorf("tipset key cid: %w", err)
+	}
```
rvagg, line 760: this is done in both branches, pull up to the top and do once?

chain/events/filter/index.go
```diff
@@ -567,11 +775,6 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 				addressLookups[ev.Emitter] = addr
 			}
 
-			tsKeyCid, err := te.msgTs.Key().Cid()
```
rvagg, line 570: good catch btw, for _each_ event, for _each_ element of an event, we're encoding the tipset key 🤦

chain/events/filter/index.go
```diff
@@ -646,16 +849,61 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 				if rowsAffected != 1 {
 					log.Warnf("restored %d events but expected only one to exist", rowsAffected)
 				}
+
+				res, err = tx.Stmt(ei.stmtRestoreEventSeen).Exec(
```
rvagg, line 853: this block seems misplaced - restoring for the tipset inside the internal loop? shouldn't this be with the `stmtInsertEventsSeen` below? you could even do the insert only if `rowsAffected == 0` after this query

chain/events/filter/index.go
rvagg: could make this a private method `updateSubsCopy()` or similar, since it's done twice in here

node/impl/full/eth.go
```diff
@@ -1270,6 +1307,35 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
 	return ethFilterResultFromEvents(ctx, ces, e.SubManager.StateAPI)
 }
 
+func (e *EthEventHandler) waitForHeightProcessed(ctx context.Context, height abi.ChainEpoch) error {
```
rvagg, line 1310: do you think it's safe enough to just rely on the RPC call timeout to time this out rather than having an explicit one in here?

node/impl/full/eth.go
```diff
@@ -1270,6 +1307,35 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
...
+	if err != nil {
+		return xerrors.Errorf("failed to check if event index has events for given height: %w", err)
+	}
+	if b {
+		return nil
+	}
+
+	if height > e.Chain.GetHeaviestTipSet().Height() {
+		return xerrors.New("height is in the future")
+	}
```
rvagg, line 1322: move to the top?

node/impl/full/eth.go
```diff
@@ -1270,6 +1307,35 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
 	return ethFilterResultFromEvents(ctx, ces, e.SubManager.StateAPI)
 }
 
+func (e *EthEventHandler) waitForHeightProcessed(ctx context.Context, height abi.ChainEpoch) error {
+	ei := e.EventFilterManager.EventIndex
+	b, err := ei.IsHeightProcessed(ctx, uint64(height))
```
rvagg, line 1312: There's a race in here between calling `IsHeightProcessed` and `SubscribeUpdates` where the height may be processed between the two calls. You might need a `SubscribeUpdatesIfHeightProcessed` that can do it atomically for you; then you're going to have to make sure you hold a lock internally between the query and getting the channel in place.

A slight variation on this would be to ditch the "is height processed" check entirely, always subscribe, but immediately send the last processed tipset as soon as you've added the subscription (which also needs to be done while holding the lock) so that you always get a `EventIndexUpdate` with the current head which may be what you want, so it's like "is height processed" but we find out while watching the channel, which should have an event waiting for us.

node/impl/full/eth.go
aarshkshah1992: Have added a 3 epohc timeout here just in case.

node/impl/full/eth.go
aarshkshah1992: Fixed this by making the event stateless and making all clients defer to the DB as the source of truth after being notified that the DB has been updated.

node/impl/full/eth.go
aarshkshah1992: Done.

node/impl/full/eth.go
aarshkshah1992: Done.

chain/events/filter/index.go
aarshkshah1992: Done.

chain/events/filter/index.go
aarshkshah1992: Done.

chain/events/filter/index.go
aarshkshah1992: Done.

chain/events/filter/index.go
aarshkshah1992: Done.

chain/events/filter/index.go
aarshkshah1992: Done.

chain/events/filter/index.go
aarshkshah1992: Done.

chain/events/filter/index.go
aarshkshah1992: Done.

node/impl/full/eth.go
```diff
@@ -44,6 +44,11 @@ var ErrUnsupported = errors.New("unsupported method")
 
 const maxEthFeeHistoryRewardPercentiles = 100
 
+var (
+	// wait for 3 epochs
+	eventReadTimeout = 90 * time.Second
```
rvagg, line 49: there's a `EpochDurationSeconds` in chain/actors/builtin

rvagg (approved):
I _think_ this looks ok, but it's slightly risky so I'd be more comfortable having someone else look too, @Stebalien if available would be good

aarshkshah1992:
@rvagg Would love a review from @Stebalien too but which part do you find risky ?

node/impl/full/eth.go
aarshkshah1992: Fixed this.

node/impl/full/eth.go
```diff
@@ -1258,10 +1263,47 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
...
+	pf, err := e.parseEthFilterSpec(ctx, filterSpec)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to parse eth filter spec: %w", err)
+	}
+
+	if pf.tipsetCid == cid.Undef {
+		maxHeight := pf.maxHeight
+		if maxHeight == -1 {
+			maxHeight = e.Chain.GetHeaviestTipSet().Height()
+		}
```
Stebalien, line 1279: We should always check the heaviest tipset, IMO. If we don't have a tipset at or beyond the requested one, fail immediately instead of waiting.

node/impl/full/eth.go
```diff
@@ -1258,10 +1263,47 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
...
+		err := e.waitForHeightProcessed(ctx, maxHeight)
```
Stebalien, line 1280: I kind of assumed the event indexing system would handle this for us. Although the install/take/uninstall pattern makes that hard.

Maybe add a `GetEvents` method to the `EventFilterManager` that waits if necessary then returns the events (and doesn't require the install/uninstall dance)?

node/impl/full/eth.go
```diff
@@ -1258,10 +1263,47 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
...
+		}
+
+		b, err := e.EventFilterManager.EventIndex.IsTipsetProcessed(ctx, pf.tipsetCid.Bytes())
+		if err != nil {
+			return nil, xerrors.Errorf("failed to check if tipset events have been indexed: %w", err)
+		}
+		if !b {
+			return nil, xerrors.Errorf("event index failed to index tipset %s", pf.tipsetCid.String())
+		}
+	}
```
Stebalien, line 1301: We need to check the range (min/max should be sufficient as long as we assume automatic backfilling), not just the end.

node/impl/full/eth.go
Stebalien: Yeah, we should _really_ do it this way. You'll be able to drop all the subscribe stuff as well.

Basically:

1. Add a `sync.Cond` to `EventFilterManager`.
2. First up, check to make sure the latest tipset we have in the chainstore is beyond the requested tipset/height.
3. Take `EventFilterManager.mu`
4. Compare `currentHeight` against the target height.
5. If `currentHeight` is less than the target, wait on the condition and repeat step 4.
6. If greater than or equal to, check if we've actually processed the tipset we care about. If not, abort.
7. Release the lock and collect the relevant events.

Stebalien: Hm. A cond variable is going to be hard if we want to obey the context.

But... I think there's a better way. We can:

1. Have event filters decide if they're done (as they process events).
2. Add a `WaitFinished` (or something) method to event filters that waits for them to declare that they're "done" (can match no more events.

Stebalien:
FYI, https://github.com/filecoin-project/lotus/issues/12111

chain/events/filter/index.go
```diff
@@ -98,6 +109,16 @@ const (
...
+
+	createTableEventsSeen = `CREATE TABLE IF NOT EXISTS events_seen (
+		id INTEGER PRIMARY KEY,
+		height INTEGER NOT NULL,
+		tipset_key_cid BLOB NOT NULL,
+		reverted INTEGER NOT NULL
+	)`
+
+	createIndexEventsSeenHeight       = `CREATE INDEX IF NOT EXISTS events_seen_height ON events_seen (height);`
+	createIndexEventsSeenTipsetKeyCid = `CREATE INDEX IF NOT EXISTS events_seen_tipset_key_cid ON events_seen (tipset_key_cid);`
```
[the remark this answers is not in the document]
aarshkshah1992: I don't really see a use case for it right now. We can add it when there is a need to.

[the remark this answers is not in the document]
Stebalien: Probably fine? Assuming everything works and SQLITE doesn't squawk at you.

chain/events/filter/index.go
```diff
@@ -75,20 +80,26 @@ var ddls = []string{
...
+	schemaVersion = 6
 
 	eventExists          = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
 	insertEvent          = `INSERT OR IGNORE INTO event(height, tipset_key, tipset_key_cid, emitter_addr, event_index, message_cid, message_index, reverted) VALUES(?, ?, ?, ?, ?, ?, ?, ?)`
 	insertEntry          = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
 	revertEventsInTipset = `UPDATE event SET reverted=true WHERE height=? AND tipset_key=?`
 	restoreEvent         = `UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
+	revertEventSeen      = `UPDATE events_seen SET reverted=true WHERE height=? AND tipset_key_cid=?`
+	restoreEventSeen     = `UPDATE events_seen SET reverted=false WHERE height=? AND tipset_key_cid=?`
+	insertEventsSeen     = `INSERT OR IGNORE INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, ?)`
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

chain/events/filter/index.go
```diff
@@ -645,17 +855,66 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 				// this is a sanity check as we should only ever be updating one event
 				if rowsAffected != 1 {
 					log.Warnf("restored %d events but expected only one to exist", rowsAffected)
+				} else {
+					isTipSetRevertRestored = true
```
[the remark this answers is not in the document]
aarshkshah1992: Fixed this.

chain/events/filter/index.go
```diff
@@ -401,9 +465,64 @@ func (ei *EventIndex) migrateToVersion5(ctx context.Context) error {
...
+
+	// INSERT an entry in the events_seen table for all epochs we do have events for in our DB
+	_, err = tx.ExecContext(ctx, `
+    INSERT INTO events_seen (height, tipset_key_cid, reverted)
+    SELECT e.height, e.tipset_key_cid, e.reverted
+    FROM event e
+    WHERE NOT EXISTS (
+        SELECT 1 FROM events_seen es
+        WHERE es.height = e.height AND es.tipset_key_cid = e.tipset_key_cid	
+    )
```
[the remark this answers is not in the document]
aarshkshah1992: Fixed this.

node/impl/full/eth.go
aarshkshah1992: Added a check for this.

node/impl/full/eth.go
aarshkshah1992: I would like to address this as part of https://github.com/filecoin-project/lotus/issues/12116 rather than solving this in a patchy way here. For now, this PR does solve the problem of `eth_getLogs` failing to return events for tipsets on the cannonical chain and I would like to ship this as an intermediate step since we're now anyways going to prioritise https://github.com/filecoin-project/lotus/issues/12116

node/impl/full/eth.go
aarshkshah1992: Assuming autobackfilling works, the presence of events for  `max`  implies the presence of events for `min` no ?

Auto backfilling will be implemented as part of https://github.com/filecoin-project/lotus/issues/12116.

aarshkshah1992:
@Stebalien I have addressed all your review comments. 

Except for https://github.com/filecoin-project/lotus/pull/12080#discussion_r1645152589 which I would like to address as part of the refactor in https://github.com/filecoin-project/lotus/issues/12116 because that really fixes this in a clean way rather than patching the current broken dual write/dual read mechanism we have for (`ChainNotify` writes to filter buffer and index DB -> event APIs read from both filter buffer and index DB).

node/impl/full/eth.go
Stebalien: My proposal should be much simpler than the current subscribe logic, unless I'm missing something. This kind of subscribe pattern tends to have deadlock issues (although I haven't looked at your implementation closely beyond suggesting an alternative).

Stebalien: Nvm. We discussed this and @aarshkshah1992's proposal is better overall.

node/impl/full/eth.go
Stebalien: Backfilling will only go back to the beginning of the local node's history.

aarshkshah1992: Fixed this by checking that we have processed events for the `minHeight` as well.

aarshkshah1992:
@rvagg @Stebalien I'd like to merge this. Based on an offline conversation with @Stebalien -> he agrees this is good to go in it's current form as it does fix the missing event issue with`GetLogs`.

We will prioritise https://github.com/filecoin-project/lotus/issues/12116 next.

chain/events/filter/index.go
```diff
@@ -75,20 +80,27 @@ var ddls = []string{
...
+	schemaVersion = 6
 
 	eventExists          = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
 	insertEvent          = `INSERT OR IGNORE INTO event(height, tipset_key, tipset_key_cid, emitter_addr, event_index, message_cid, message_index, reverted) VALUES(?, ?, ?, ?, ?, ?, ?, ?)`
 	insertEntry          = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
 	revertEventsInTipset = `UPDATE event SET reverted=true WHERE height=? AND tipset_key=?`
 	restoreEvent         = `UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
+	revertEventSeen      = `UPDATE events_seen SET reverted=true WHERE height=? AND tipset_key_cid=?`
+	restoreEventSeen     = `UPDATE events_seen SET reverted=false WHERE height=? AND tipset_key_cid=?`
+	upsertEventsSeen     = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, ?) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
```
rvagg, line 100: `reverted` is always `false` in this call, save a substitution by just setting it eh?

```suggestion
	upsertEventsSeen     = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, false) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
```

node/impl/full/eth.go
rvagg: did you forget to push the fix? still has `90` hardwired here

```suggestion
	eventReadTimeout = builtin.EpochDurationSeconds * time.Second
```

node/impl/full/eth.go
```diff
@@ -1258,10 +1263,60 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
...
+		return nil, xerrors.Errorf("failed to parse eth filter spec: %w", err)
+	}
+
+	if pf.tipsetCid == cid.Undef {
+		maxHeight := pf.maxHeight
+		if maxHeight == -1 {
+			maxHeight = e.Chain.GetHeaviestTipSet().Height()
+		}
+		if maxHeight > e.Chain.GetHeaviestTipSet().Height() {
+			return nil, xerrors.Errorf("max height requested is greater than the heaviest tipset")
```
rvagg, line 1281:
```suggestion
			return nil, xerrors.Errorf("maxHeight requested is greater than the heaviest tipset")
```

node/impl/full/eth.go
```diff
@@ -1258,10 +1263,60 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
...
+
+		err := e.waitForHeightProcessed(ctx, maxHeight)
+		if err != nil {
+			return nil, err
+		}
+
+		// should also have the minHeight in the filter indexed
+		b, err := e.EventFilterManager.EventIndex.IsHeightProcessed(ctx, uint64(pf.minHeight))
+		if err != nil {
+			return nil, xerrors.Errorf("failed to check if event index has events for the minheight: %w", err)
```
rvagg, line 1292:
```suggestion
			return nil, xerrors.Errorf("failed to check if event index has events for the minHeight: %w", err)
```

chain/events/filter/index.go
```diff
@@ -651,11 +862,40 @@ func (ei *EventIndex) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 		}
 	}
 
+	// this statement will mark the tipset as processed and will insert a new row if it doesn't exist
+	// or update the reverted field to false if it does
+	_, err = tx.Stmt(ei.stmtUpsertEventsSeen).Exec(
+		te.msgTs.Height(),
+		tsKeyCid.Bytes(),
+		false,
```
rvagg, line 870:
```suggestion
```

node/impl/full/eth.go
```diff
@@ -1270,6 +1325,53 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
...
+	defer cancel()
+
+	// if the height we're interested in has already been indexed -> there's nothing to do here
+	b, err := ei.IsHeightProcessed(ctx, uint64(height))
+	if err != nil {
+		return xerrors.Errorf("failed to check if event index has events for given height: %w", err)
+	}
+	if b {
+		return nil
+	}
```
rvagg, line 1344: could compact this and the duplicate block below a little

```suggestion
	if b, err := ei.IsHeightProcessed(ctx, uint64(height)); err != nil {
		return xerrors.Errorf("failed to check if event index has events for given height: %w", err)
	} else if b {
		return nil
	}
```

(up to you tho)

node/impl/full/eth.go
```diff
@@ -1270,6 +1325,53 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
...
+
+	for {
+		select {
+		case <-subCh:
+			b, err = ei.IsHeightProcessed(ctx, uint64(height))
```
rvagg, line 1362: why don't we just return the height on the object coming from the channel and check that rather than needing to do a round-trip through the db at each epoch that's processed? we could have thousands of these queued up waiting for a far-future epoch, each one doing a SELECT on each epoch while they wait

rvagg: oh, nevermind, `maxHeight > e.Chain.GetHeaviestTipSet().Height()` takes care of that concern

node/impl/full/eth.go
aarshkshah1992: @rvagg I did but the tests barf on this because the tests have like 2ms block times and it's sometimes not enough for the head to get processed.

node/impl/full/eth.go
aarshkshah1992: Reduced it below at multiple places . Thanks.

node/impl/full/eth.go
rvagg: hah, that's a bit of a dilemma then!

rvagg (approved):

Issue #12081, opened 2024-06-12
Do not enable `read_uncommitted` by default


masih:
The `PRAGMA` for 3 databases across lotus repo enable uncommitted reads:

https://github.com/filecoin-project/lotus/blob/395cdd721e7ead60e4b7853b3023f50daeed697f/chain/ethhashlookup/eth_transaction_hash_lookup.go#L25

https://github.com/filecoin-project/lotus/blob/6bc04c8060d59099f5d4eadf7b1ad716f8799b33/chain/events/filter/index.go#L32

https://github.com/filecoin-project/lotus/blob/6bc04c8060d59099f5d4eadf7b1ad716f8799b33/chain/events/filter/index.go#L32

This can have some performance gain but will cause all sorts of problems, specially in db interactions that require ACID properties where data consistency matters. Based on a quick look we _do_ care about data consistency across all three places listed above. Therefore, I recommend against performing dirty reads.

snissn:
I think we can close this - https://github.com/filecoin-project/lotus/pull/12090/files#diff-fe5e47a357d5465c3a2b1002977a53ad7c0990f76da22137417723577b87b19eL32 @rvagg made the change in pull 12090

masih:
That PR only updates events db pragma. right?

rvagg:
yeah, only events sorry, we can follow up with the others as a separate PR

BigLep:
2024-06-18 triage note: @rvagg we wanted to confirm that https://github.com/filecoin-project/lotus/pull/12098 will address this issue.

rvagg:
yes, it will, they'll all have the same pragma and read_uncommitted is gone

Issue #12089, opened 2024-06-13
SQLite WAL File for Events Growing Until Lotus Restart


snissn:
**Description:**
The SQLite Write-Ahead Logging (WAL) file for storing events in the Lotus client continues to grow indefinitely until the client is restarted. This can lead to excessive disk usage and potential performance degradation. It is not the state file but a write ahead log that should be periodically reset. Restarting lotus clears the file. `events.db-wal` was 14GBs from running over a < 24 hour period and was deleted and reset after restarting lotus. It continues to grow from 0 bytes after restarting lotus

**Steps to Reproduce:**
1. Run the Lotus client.
2. Monitor the size of the SQLite WAL file over time.
3. Observe that the WAL file grows continuously without being truncated.

**Expected Behavior:**
The SQLite WAL file should be periodically checkpointed and truncated to prevent uncontrolled growth.

**Environment:**
- Lotus version: master
- OS: ubuntu

**Additional Information:**
Implementing a mechanism to manually trigger checkpoints or reviewing the configuration settings related to SQLite checkpointing may help mitigate this issue. The following config parameter should force the WAL to be vacuumed automatically  https://sqlite.org/pragma.html#pragma_wal_autocheckpoint. Also as helpful information a query can be sent directly to sqlite to check point via `PRAGMA wal_checkpoint(FULL);` 

**Possible Solutions:**
1. Investigate and fix any bugs preventing automatic checkpointing related to configuration of pragma_wal_autocheckpoint.
2. Implement automatic checkpointing every N minutes as a temporary workaround that executes `PRAGMA wal_checkpoint(FULL);`

rvagg:
I spent some time tinkering with this because I've been running the db since nv22 and my WAL is now over 200G which is definitely not normal. Unfortunately I couldn't figure out a straightforward solution and get it solved! But I do think this is urgent and we should spend some time figuring this out because it likely impacts performance, aside from just being a pain in the backside for node operators.

My understanding and some of the things I've tried:

* Our current settings should automatic checkpoint an _should_ compact the WAL periodically because of it; but it doesn't seem to
* It won't checkpoint and compact if there are open read operations. I did find one potentially unclosed transaction in index.go but dealing with it didn't seem to make a difference. Perhaps there are more subtle bugs with our transactions in there.
* Fiddling with other WAL settings didn't seem to make much difference, including putting a hard max on it!
[...]

PR #12090, opened 2024-06-14
fix: events: sqlite db improvements

rvagg:
* fix two unclosed multi-row queries
* tune options to limit wal growth

Ref: https://github.com/filecoin-project/lotus/issues/12089

--

~This is a draft for now while I test it out. I've manually truncated my current, massive wal file (`sqlite3 /path/to/sqlite/events.db` followed by `PRAGMA wal_checkpoint(TRUNCATE);`) so I'm starting from zero again and will watch its behaviour with this and report back.~

### DIFF
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -26,10 +26,11 @@ var pragmas = []string{
 	"PRAGMA temp_store = memory",
 	"PRAGMA mmap_size = 30000000000",
 	"PRAGMA page_size = 32768",
-	"PRAGMA auto_vacuum = NONE", // not useful until we implement GC
+	"PRAGMA auto_vacuum = NONE",
 	"PRAGMA automatic_index = OFF",
 	"PRAGMA journal_mode = WAL",
-	"PRAGMA read_uncommitted = ON",
+	"PRAGMA wal_autocheckpoint = 256", // checkpoint @ 256 pages
+	"PRAGMA journal_size_limit = 0",   // always reset journal and wal files
 }
 
 // Any changes to this schema should be matched for the `lotus-shed indexes backfill-events` command
@@ -751,6 +752,7 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 		}
 		return xerrors.Errorf("exec prefill query: %w", err)
 	}
+	defer q.Close()
 
 	var ces []*CollectedEvent
 	var currentID int64 = -1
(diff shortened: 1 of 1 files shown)


rvagg:
I think I've cracked it!

This line, right at the beginning of initialising the database checks whether we need to make the database or not:

https://github.com/filecoin-project/lotus/blob/63b95779cebb10dadf385b1e0e13c899b2922d1b/chain/events/filter/index.go#L440

In normal operation, we end up at `q.Next()` which gives us a row that just contains `"_meta"`, but because it's an `*sql.Rows`, which holds on to a read operation until closed, and we never call `q.Next()` again (which would return `false` and close it) or just `q.Close()`, we hold on to a read operation.

I believe this is the read operation that's left open for the entirety of the lifetime of the database so the checkpoint never gets to run and the WAL just grows and grows. This could also explain performance problems too if we have a mammoth WAL that we need to process when doing queries too.

The changes in here force a WAL truncation on checkpoint, and I've lowered the checkpoints to 256 pages. Our page size is unfortunately set quite large for this db (set on db create), so 256 gets us to ~8MiB WAL size before checkpoint and then truncate. The default is 1000, which would land us closer to 32MiB.

Watching the chain as I write this, I'm seeing it truncate every 3-4 epochs where it hits around the 8MiB mark. So it's ~2MiB of events per epoch, although it's extremely variable. There'll probably be epochs where we easily blow past the 8MiB mark.

So, we could increase this number, but compacting the write-ahead log every epoch or 4 doesn't seem like such a bad thing.

rvagg:
Here's the WAL for my mainnet node using this patch, every 30 seconds, compacting every 2-4 epochs.

Screenshot 2024-06-14 at 3 32 23 PM

Stebalien (approved):
:heart:

snissn:
@rvagg does this ticket close https://github.com/filecoin-project/lotus/issues/12081 ?

rvagg:
Continuing to monitor my wal file and the behaviour is not what I expected: mostly it seems to hover around what I showed above, but it has long stretches where it gets stuck at a particular, overly large, size. A few hours at 42MiB; the next day a stretch of ~5h at 44MiB, down again for 2 epochs then up to 37MiB for another ~6h, up from there to 41MiB for a couple of hours, then back down again; later the same day it grew up to 140MiB for a few hours but now it's running normal again.

The most important thing is that it eventually shrinks. However the behaviour of growing and getting stuck at a size well above the set maximum is strange because I can't reconcile it with what the docs suggest and what I know of the access patterns. Like we have long-running read locks that eventually free up after hours; but that doesn't square with how we run this thing. Perhaps this is just internal sqlite behaviour that's not clearly documented.

PR #12092, opened 2024-06-14
feat: itest: minor UnmanagedMiner refactor for reusability and simplicity

rvagg:
* move some utility methods into UnmanagedMiner so they can be used from other itests
* make UnmanagedMiner aware of mock proofs so we can remove a lot of branching

### DIFF
--- a/itests/manual_onboarding_test.go
+++ b/itests/manual_onboarding_test.go
@@ -13,10 +13,10 @@ import (
 	"github.com/filecoin-project/lotus/itests/kit"
 )
 
-const defaultSectorSize = abi.SectorSize(2 << 10) // 2KiB
-
 // Manually onboard CC sectors, bypassing lotus-miner onboarding pathways
 func TestManualSectorOnboarding(t *testing.T) {
+	const defaultSectorSize = abi.SectorSize(2 << 10) // 2KiB
+
 	req := require.New(t)
 
 	for _, withMockProofs := range []bool{true, false} {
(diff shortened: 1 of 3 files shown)


itests/manual_onboarding_test.go
```diff
@@ -103,72 +95,24 @@ func TestManualSectorOnboarding(t *testing.T) {
...
-
-	soi, err := miner.FullNode.StateSectorGetInfo(ctx, miner.ActorAddr, sector, head.Key())
-	req.NoError(err)
-	t.Logf("Miner %s SectorOnChainInfo %d: %+v", miner.ActorAddr.String(), sector, soi)
-
-	_ = miner.FullNode.WaitTillChain(ctx, kit.HeightAtLeast(head.Height()+5))
-
-	t.Log("Checking power after PoSt ...")
-
-	// Miner B should now have power
```
aarshkshah1992, line 156: nit: this comment should change to "Miner should have no power"

aarshkshah1992 (approved):

itests/manual_onboarding_test.go
rvagg: this is removed from here, the code is moved to https://github.com/filecoin-project/lotus/pull/12092/files#diff-607cce01302fd6d23e59fb8699690252d7a55f88256c11a3112e9995fc591889R1015 and the same comment is there, but it's testing `sectorSize` power, so it _should_ have power, right?

PR #12093, opened 2024-06-14
feat: lotus-shed: finality calculator (WIP)

rvagg:
Doesn't quite work, something wrong with the bessel library that makes it panic.

In this current form it just takes a file full of chain tipset counts. I took https://github.com/consensus-shipyard/ec-finality-calculator/blob/main/finality_calc_validator.py and put this just before the `print(finality_calc_validator(chain, e, f, c, s))`:

```python
    filename = "chain.txt"
    with open(filename, 'w') as file:
        for item in chain:
            file.write("%s\n" % item)

    # Print the filename
    print(f"Chain written to: {os.path.abspath(filename)}")
```

Then I can run `lotus-shed finality-calculator /path/to/chain.txt` and 🤞 for a comparable result. But so far I just get a panic at the call to `bessel.I()`. 🤷


rvagg:
Recent F3 dramas reminded me that I had this in limbo (for nearly 2 years?) .. and also that I now have AI that can help track down the tedious maths bug that I didn't have the time or patience to figure out which caused blockage! So here it is, bug discovered (it was silly, negative `k` values were producing garbage values, needed to be normalised, [see this if you care](https://github.com/rvagg/go-skellam-pmf/commit/9cf38e94a40e4bbf2646ac13717072a05a4b1734#diff-5646b1a925f2e5d6992d56f509933d08b90acd7d8715830685054ee7003db009)).

New commit message is:

    feat(shed): ec finality calculator (FRC-0089)
    
    Add `lotus-shed finality-calculator` subcommand that computes the
    probability a tipset gets replaced, based on FRC-0089. Reads chain
    history from a running node or text file and outputs a summary table
    with the shallowest depth achieving the 2^-30 safety target.
    
    Under healthy conditions (~5 blocks/epoch), probabilistic finality is
    typically reached within ~30 epochs (~15 min), compared to the static
    900-epoch (~7.5 hr) EC finality assumption.

In action

image

Follow-up from here would be to move it into lotus-proper and add an API for it. Probably just in /v2, add a `ChainGetTipSetFinalityStatus` API for verbose info, then add a `Tag` for `TipSetSelector` for the other APIs in there that's `"ec-finalized"` _or_ even just integrate it with the F3 flow and have `"finalized"` max out on this or F3; same could apply to Ethereum APIs.

rvagg:
OK, merging this as is, but I'll follow up with another PR that moves the calculator into lotus and proposes some API changes to incorporate this.

jsoares:
Nice catch; AI to the rescue :) 

Coincidentally, we also published a more detailed preprint last week (math unchanged): https://arxiv.org/pdf/2603.01307

rvagg:
Thanks @jsoares, I missed this comment today when I got #13547 up which goes deeper with this work. In there I only reference the FRC, I could also reference the paper I suppose. The one thing I note that's different, in #13547 I'm going with this assertion from the FRC text:

> Filecoin currently use a 900-epoch (7.5 hours) settlement period to provide an intended finality guarantee of 2^{-30}

And use it to say "one-in-a-billion".

But the paper says this:

> Also worthy of note is that an error probability of 10<sup>−10</sup>, corresponding to a once-in-10,000-years event, is attained in fewer than 30 rounds; this is the same error probability that was originally targeted when setting the 900-round soft finality parameter in Filecoin

So now we have competing claims for the origin of the `900`, is it 2^-30 (~10^-9, 'one-in-a-billion') or 10^-10 (~2^-33, 'once-in-10,000-years')? Not strictly important, but I am making claims in the docs in there, particularly: https://github.com/filecoin-project/lotus/blob/rvagg/finality-calculator/documentation/en/finality.md

jsoares:
> Thanks @jsoares, I missed this comment today when I got #13547 up which goes deeper with this work. In there I only reference the FRC, I could also reference the paper I suppose. The one thing I note that's different, in #13547 I'm going with this assertion from the FRC text:
> 
> > Filecoin currently use a 900-epoch (7.5 hours) settlement period to provide an intended finality guarantee of 2^{-30}
> 
> And use it to say "one-in-a-billion".
> 
> But the paper says this:
> 
> > Also worthy of note is that an error probability of 10−10, corresponding to a once-in-10,000-years event, is attained in fewer than 30 rounds; this is the same error probability that was originally targeted when setting the 900-round soft finality parameter in Filecoin
> 
> So now we have competing claims for the origin of the `900`, is it 2^-30 (~10^-9, 'one-in-a-billion') or 10^-10 (~2^-33, 'once-in-10,000-years')? Not strictly important, but I am making claims in the docs in there, particularly: https://github.com/filecoin-project/lotus/blob/rvagg/finality-calculator/documentation/en/finality.md

Another nice catch, but this is an issue with the paper. The target was `2^-30`, even if its origins are somewhat murky. In the paper, we switched to base 10 because 2 becomes awkward to reason about for the range with which we're dealing, and the first integer exponent that **exceeded** the `2^-30` target was `-10`. However, the way we phrased it is wrong; I'll update the paper. Thanks for pointing it out.

PR #12095, opened 2024-06-15
chore: types: remove more items forgotten after markets

ribasushi:


rvagg (commented):
this seems OK to me but I'm just unsure about all the types in api_full.go; I'd like @aarshkshah1992 to have a look

api/api_full.go
```diff
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/google/uuid"
```
aarshkshah1992, line 9: Yeah none of the types removed in this file are used anymore. This is correct.

aarshkshah1992 (approved):

PR #12098, opened 2024-06-17
feat: sqlite: extract common init and migration utilities

rvagg:
All of the versioning and migration logic is shared between the 3 sqlite databases we use the other two have the same pending-query problem as was resolved for _meta in https://github.com/filecoin-project/lotus/pull/12090, so that's addressed here by making it common.

* Use the same `PRAGMA` set across all 3 - this may need to change in future but we can configure that later when we need it.
* Same `_meta` versioning and migration logic. Do it all in a migration loop where each migration is wrapped in a transaction, logging and error handling and the calling instance is responsible for feeding in functions that can perform a migration. Versioning and `_meta` is entirely hidden from the calling instance.
* Two minor breaking changes in here:
  1. `ethhashlookup.NewTransactionHashLookup` now takes a context as its first argument.
  2. `index.NewMsgIndex` now takes a full path to its database as its `path` (not `basePath`) argument to make it consistent with the others.

Needs a fresh set of eyes to make sure I haven't stumbled over something in one of these to make it inconsistent. I'll do another pass of this myself later too, before I try deploying this on my calibnet and then mainnet nodes. There should be no meaningful visible changes for a running instance from this, mostly they just get consistent `PRAGMA`s and possibly some changed logging.

Also closes: https://github.com/filecoin-project/lotus/issues/12081

Will need some rebase work after https://github.com/filecoin-project/lotus/pull/12080 is landed so this should hold off till that's done.

### DIFF
--- a/chain/ethhashlookup/eth_transaction_hash_lookup.go
+++ b/chain/ethhashlookup/eth_transaction_hash_lookup.go
@@ -1,6 +1,7 @@
 package ethhashlookup
 
 import (
+	"context"
 	"database/sql"
 	"errors"
 	"strconv"
@@ -10,20 +11,12 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
+	"github.com/filecoin-project/lotus/lib/sqlite"
 )
 
-var ErrNotFound = errors.New("not found")
+const DefaultDbFilename = "txhash.db"
 
-var pragmas = []string{
-	"PRAGMA synchronous = normal",
-	"PRAGMA temp_store = memory",
-	"PRAGMA mmap_size = 30000000000",
-	"PRAGMA page_size = 32768",
-	"PRAGMA auto_vacuum = NONE",
-	"PRAGMA automatic_index = OFF",
-	"PRAGMA journal_mode = WAL",
-	"PRAGMA read_uncommitted = ON",
-}
+var ErrNotFound = errors.New("not found")
 
 var ddls = []string{
 	`CREATE TABLE IF NOT EXISTS eth_tx_hashes (
@@ -33,17 +26,7 @@ var ddls = []string{
 	)`,
 
 	`CREATE INDEX IF NOT EXISTS insertion_time_index ON eth_tx_hashes (insertion_time)`,
-
-	// metadata containing version of schema
-	`CREATE TABLE IF NOT EXISTS _meta (
-    	version UINT64 NOT NULL UNIQUE
-	)`,
-
-	// version 1.
-	`INSERT OR IGNORE INTO _meta (version) VALUES (1)`,
 }
-
-const schemaVersion = 1
 
 const (
 	insertTxHash = `INSERT INTO eth_tx_hashes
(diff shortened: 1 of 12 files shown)


chain/index/msgindex.go
```diff
@@ -266,23 +218,6 @@ func PopulateAfterSnapshot(lctx context.Context, basePath string, cs ChainStore)
...
 
-// init utilities
-func prepareDB(db *sql.DB) error {
-	for _, stmt := range dbDefs {
-		if _, err := db.Exec(stmt); err != nil {
-			return xerrors.Errorf("error executing sql statement '%s': %w", stmt, err)
-		}
-	}
-
-	for _, stmt := range dbPragmas {
```
aarshkshah1992, line 277: In lib/sqlite.go, we apply the pragma first during `Open` and then exec the DDL during `initDB`.
But here, we exec the DDL first and then apply the pragma. Are we sure that the creation of the pragma has not relevant effects to the execution of the DDL ?

lib/sqlite/sqlite.go
```diff
@@ -0,0 +1,165 @@
...
+
+		log.Infof("Successfully migrated %s database from version %d to %d in %s", name, i-1, i, time.Since(now))
+	}
+
+	if runVacuum {
+		// During the large migrations, we have likely increased the WAL size a lot, so lets do some
+		// simple DB administration to free up space (VACUUM followed by truncating the WAL file)
+		// as this would be a good time to do it when no other writes are happening.
+		log.Infof("Performing %s database vacuum and wal checkpointing to free up space after the migration", name)
+		_, err := db.ExecContext(ctx, "VACUUM")
```
aarshkshah1992, line 154: Is the "VACCUM" and "PRAGMA wal_checkpoint(TRUNCATE) stmt available in Postgress as well ?

lib/sqlite/sqlite.go
```diff
@@ -0,0 +1,165 @@
...
+		return xerrors.Errorf("invalid %s database version: no version found", name)
+	}
+
+	if foundVersion > schemaVersion {
+		return xerrors.Errorf("invalid %s database version: version %d is greater than the number of migrations %d", name, foundVersion, len(versionMigrations))
+	}
+
+	runVacuum := foundVersion != schemaVersion
+
+	for i := foundVersion + 1; i <= schemaVersion; i++ {
```
aarshkshah1992, line 125: Right so if foundversion=schemaversion then this look wont be executed

lib/sqlite/sqlite.go
```diff
@@ -0,0 +1,165 @@
...
+	if q != nil {
+		defer func() { _ = q.Close() }()
+	}
+
+	if errors.Is(err, sql.ErrNoRows) || !q.Next() {
+		// empty database, create the schema including the _meta table and its versions
+		ddls := append(metaDdl(uint64(schemaVersion)), ddls...)
+		for _, ddl := range ddls {
+			if _, err := db.Exec(ddl); err != nil {
+				return xerrors.Errorf("failed to %s database execute ddl %q: %w", name, ddl, err)
```
aarshkshah1992, line 99: I think we also need a `db.Close` here

```
_ = db.Close()
```

aarshkshah1992 (approved):
lgtm

aarshkshah1992 (commented):
some questions. Overall looking great. Would be nice to test this on a devnet/calibnet node once rebased on https://github.com/filecoin-project/lotus/pull/12080.

lib/sqlite/sqlite.go
```diff
@@ -0,0 +1,165 @@
...
+	for i := foundVersion + 1; i <= schemaVersion; i++ {
+		now := time.Now()
+
+		log.Infof("Migrating %s database to version %d...", name, i)
+
+		tx, err := db.BeginTx(ctx, nil)
+		if err != nil {
+			return xerrors.Errorf("failed to start %s database transaction: %w", name, err)
+		}
+		defer func() { _ = tx.Rollback() }()
```
aarshkshah1992, line 134: doing this in a loop is fine as only the tx that errors out will be rolled back when the function returns as all other txns will already have been committed and rollback is a no-op once a tx is committed.

lib/sqlite/sqlite.go
aarshkshah1992: @rvagg Do we need to close the DB here before returning ?

chain/ethhashlookup/eth_transaction_hash_lookup.go
```diff
@@ -103,50 +86,17 @@ func (ei *EthTxHashLookup) DeleteEntriesOlderThan(days int) (int64, error) {
 	return res.RowsAffected()
 }
 
-func NewTransactionHashLookup(path string) (*EthTxHashLookup, error) {
-	db, err := sql.Open("sqlite3", path+"?mode=rwc")
+func NewTransactionHashLookup(ctx context.Context, path string) (*EthTxHashLookup, error) {
+	db, _, err := sqlite.Open(path)
```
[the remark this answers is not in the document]
aarshkshah1992: Ah okay because we will get the `error opening database` if the path does not exist. So in no case will err==nil AND exists=false.

aarshkshah1992 (approved):
lgtm

chain/index/msgindex.go
rvagg: Yes, I did note this difference but judged it to be a mistake. The pragmas shouldn't impact creation _except_ for the page size which is set at db creation; so when these are first created they use the default page size, which has flow-on effects for the other settings that depend on page size.

The one where it matters now is `wal_autocheckpoint = 256` which we are setting - it's a multiple of page size, and the default page size is quite a bit lower than what we are setting, and the default `wal_autocheckpoint` is `1024`. But all this means is that we end up running checkpoints and possible compaction at a slightly smaller WAL size than before.

This may be a slight change of behaviour for users that have this feature turned on and then upgrade, but it shouldn't be noticeable. Also, message index is not on by default and it says this in the config file: `EXPERIMENTAL FEATURE. USE WITH CAUTION`, so I'm not sure who even is running it (I'm asking Glif and Zondax).

lib/sqlite/sqlite.go
rvagg: yeah, I'm aware that this doesn't read very clearly, but I couldn't come up with a better way to express it. I've added some comments.

lib/sqlite/sqlite.go
rvagg: should be done in the calling function; it was in 2 of the 4 calls, but I've added the other 2 and a comment in the method docs
good catch!

chain/ethhashlookup/eth_transaction_hash_lookup.go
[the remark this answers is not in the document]
rvagg: there's only one place this matters, msgindex.go has a reconcilliation process _if_ it exists and it needs to catch up; we could consider similar for the others—like events index where we do some checks for previous epochs to make sure we didn't miss any, but right now there's only one

rvagg:
OK, so on the use of the message index - it is being used by some large node operators, it's not just a hidden feature that nobody uses it.

The change here, of switching the create+pragma order, means that _if_ you have an existing message index, we are going to change the checkpoint size, which will impact how big the write-ahead-log will be before it tries to compact it:

* Currently it's block-size (4096 - because pragma was done after db create) x 1025 (default) = 4MiB
* With this change, for an existing node, it'll be block-size (4096 - until the next migration when we run a vacuum) x 256 = 1MiB
* With this change for a new node, it'll be block-size (32768) x 256 = 8MiB (consistent with all other dbs)

There's also likely some minor performance impacts from increasing the block-size, but it depends on the exact usage and is _likely_ a factor of how big the chunks are that we're writing. It probably makes the events db faster, but for the other 2 dbs they probably don't benefit from the larger block-size, but I doubt there'll be much cost from it.

Issue #12083, opened 2024-06-12
Add finality-related params for `eth_getBlockByNumber` 


eshon:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [X] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Currently `eth_getBlockByNumber` params related to the finality of a block are not completely implemented in the Lotus Eth RPC API.

[Here](https://docs.chainstack.com/reference/ethereum-getblockbynumber) is a more detailed description of the params.

It might be helpful for RPC API users to understand the finality of a block in a way similar to Ethereum semantics so that they can better prepare themselves in the event of a reorg.

### Describe the solution you'd like

`safe` could be used for blocks after a 30 epoch delay (as [explained here](https://github.com/filecoin-project/lotus/issues/12078#issuecomment-2160091533))

snissn:
I see that the for api eth_getBlockByNumber has a param "finalized"  - https://docs.alchemy.com/reference/eth-getblockbynumber

String - Either the hex value of a block number OR One of the following block tags:
- pending - A sample next block built by the client on top of latest and containing the set of transactions usually taken from local mempool. Intuitively, you can think of these as blocks that have not been mined yet.
- latest - The most recent block in the canonical chain observed by the client, this block may be re-orged out of the canonical chain even under healthy/normal conditions.
- safe - The most recent crypto-economically secure block, cannot be re-orged outside of manual intervention driven by community coordination. Intuitively, this block is “unlikely” to be re-orged.
- finalized - The most recent crypto-economically secure block, that has been accepted by >2/3 of validators. Cannot be re-orged outside of manual intervention driven by community coordination. Intuitively, this block is very unlikely to be re-orged.
- earliest - The lowest numbered block the client has available. Intuitively, you can think of this as the first block created.

[...]

PR #12110, opened 2024-06-18
feat: support "safe" param for eth_getBlockByNumber for 30 blocks

snissn:
## Related Issues
https://github.com/filecoin-project/lotus/issues/12083

```
$ curl  -s -X POST localhost:1234/rpc/v1   -H "Content-Type: application/json"   -d '{
    "jsonrpc":"2.0",
    "method":"eth_getBlockByNumber",
    "params":["safe", true],
    "id":1
}'  | jq -r '.result.number' | xargs printf "%d\n"
1714033

$ curl  -s -X POST localhost:1234/rpc/v1   -H "Content-Type: application/json"   -d '{
    "jsonrpc":"2.0",
    "method":"eth_getBlockByNumber",
    "params":["latest", true],
    "id":1
}'  | jq -r '.result.number' | xargs printf "%d\n"
1714063
```

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/types/ethtypes/eth_types.go
+++ b/chain/types/ethtypes/eth_types.go
@@ -28,6 +28,10 @@ import (
 
 var ErrInvalidAddress = errors.New("invalid Filecoin Eth address")
 
+// Research into Filecoin chain behaviour suggests that probabilistic finality generally approaches the intended stability guarantee at, or near, 30 epochs. Although a strictly "finalized" safe recommendation remains 900 epochs.
+// See https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0089.md
+const SafeEpochDelay = abi.ChainEpoch(30)
+
 type EthUint64 uint64
 
 func (e EthUint64) MarshalJSON() ([]byte, error) {
--- a/gateway/proxy_eth.go
+++ b/gateway/proxy_eth.go
@@ -17,6 +17,7 @@ import (
 	"github.com/filecoin-project/go-state-types/big"
 
 	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/chain/events/filter"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
@@ -143,8 +144,9 @@ func (gw *Node) checkBlkParam(ctx context.Context, blkParam string, lookback eth
 		}
 		num = ethtypes.EthUint64(head.Height()) - lookback
 	case "safe":
-		safeEpochDelay := 30 // https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0089.md
-		num = ethtypes.EthUint64(head.Height()) - lookback - ethtypes.EthUint64(safeEpochDelay)
+		num = ethtypes.EthUint64(head.Height()) - lookback - ethtypes.EthUint64(ethtypes.SafeEpochDelay)
+	case "finalized":
+		num = ethtypes.EthUint64(head.Height()) - lookback - ethtypes.EthUint64(build.Finality)
 	default:
 		if err := num.UnmarshalJSON([]byte(`"` + blkParam + `"`)); err != nil {
 			return fmt.Errorf("cannot parse block number: %v", err)
--- a/gateway/proxy_eth.go
+++ b/gateway/proxy_eth.go
@@ -142,6 +142,9 @@ func (gw *Node) checkBlkParam(ctx context.Context, blkParam string, lookback eth
 			break
 		}
 		num = ethtypes.EthUint64(head.Height()) - lookback
+	case "safe":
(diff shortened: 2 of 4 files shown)


gateway/proxy_eth.go
```diff
@@ -142,6 +142,9 @@ func (gw *Node) checkBlkParam(ctx context.Context, blkParam string, lookback eth
 			break
 		}
 		num = ethtypes.EthUint64(head.Height()) - lookback
+	case "safe":
+		safeEpochDelay := 30 // https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0089.md
```
rvagg, line 146: How about we make this a const in chain/types/ethtypes/eth_types.go, it's already used by both files

And we should be clear about it in the comment, other than just pointing to the FRC, just because we're not documenting these properly anywhere. Something like:

> Research into Filecoin chain behaviour suggests that probabilistic finality generally approaches the intended stability guarantee at, or near, 30 epochs. Although a strictly "finalized" safe recommendation remains 900 epochs.
> See https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0089.md

rvagg:
Would be worth doing `finalized` as well in this and then we'd have the complete set, no?

aarshkshah1992:
@snissn "finalised" should be very similar to the "safe" implementation no ? Best to just add it to this PR.

rvagg:
there's a `build.Finality` for that btw

snissn:
Thanks @rvagg @aarshkshah1992 !! 

I will move the safeEpochDelay to a const in eth_types and will use `build.Finality` for the additional case and request a review again when that is ready

snissn:
I've added support for 'finalized' and created a const for the safe delay with the suggested comment
```
$ curl  -s -X POST localhost:1234/rpc/v1   -H "Content-Type: application/json"   -d '{
    "jsonrpc":"2.0",
    "method":"eth_getBlockByNumber",
    "params":["safe", true],
    "id":1
}'  | jq -r '.result.number' | xargs printf "%d\n"
1717680

$ curl  -s -X POST localhost:1234/rpc/v1   -H "Content-Type: application/json"   -d '{
    "jsonrpc":"2.0",
    "method":"eth_getBlockByNumber",
    "params":["latest", true],
    "id":1
}'  | jq -r '.result.number' | xargs printf "%d\n"
1717710

$ curl  -s -X POST localhost:1234/rpc/v1   -H "Content-Type: application/json"   -d '{
    "jsonrpc":"2.0",
    "method":"eth_getBlockByNumber",
    "params":["finalized", true],
    "id":1
}'  | jq -r '.result.number' | xargs printf "%d\n"
1716810
```

chain/types/ethtypes/eth_types.go
```diff
@@ -28,6 +28,10 @@ import (
 
 var ErrInvalidAddress = errors.New("invalid Filecoin Eth address")
 
+// Research into Filecoin chain behaviour suggests that probabilistic finality generally approaches the intended stability guarantee at, or near, 30 epochs. Although a strictly "finalized" safe recommendation remains 900 epochs.
```
rvagg, line 31:
```suggestion
// Research into Filecoin chain behaviour suggests that probabilistic finality
// generally approaches the intended stability guarantee at, or near, 30 epochs.
// Although a strictly "finalized" safe recommendation remains 900 epochs.
``````

very minor nit, but can we break long comments like this at the 80 or 100 chars

**Comment by rvagg** [comment]
Lack of tests is a bit of a bummer on this and there's not an easy slot for a new unit test - but if you could come up with an easy way to do it locally there then that would be great.

Here's an integration test that would exercise it, which is something at least, stick this in `itests/eth_api_test.go` under `TestEthGetGenesis` which has a `EthGetBlockByNumber` call so it's in a similar ballpark:

```go
func TestEthBlockNumberAliases(t *testing.T) {
	blockTime := 2 * time.Millisecond
	kit.QuietMiningLogs()
	ens.InterconnectAll().BeginMining(blockTime)
	ens.Start()

	build.Clock.Sleep(time.Second)

	defer cancel()

	head := client.WaitTillChain(ctx, kit.HeightAtLeast(build.Finality+100))

	// latest should be head-1 (parents)
	latestEthBlk, err := client.EVM().EthGetBlockByNumber(ctx, "latest", true)
	require.NoError(t, err)
	require.GreaterOrEqual(t, uint64(latestEthBlk.Number), uint64(head.Height()-1))

	// safe should be latest-30
	safeEthBlk, err := client.EVM().EthGetBlockByNumber(ctx, "safe", true)
	require.NoError(t, err)
	require.LessOrEqual(t, math.Abs(float64(latestEthBlk.Number-30)-float64(safeEthBlk.Number)), 1.0)

	// finalized should be Finality blocks behind latest
	finalityEthBlk, err := client.EVM().EthGetBlockByNumber(ctx, "finalized", true)
	require.NoError(t, err)
	require.LessOrEqual(t, math.Abs(float64(uint64(latestEthBlk.Number)-uint64(build.Finality))-float64(finalityEthBlk.Number)), 1.0)
}
```

snissn:
Thanks that test looks great and it feels a lot better that we have this test coverage with this PR! 84199c277

rvagg (approved):

Issue #12091, opened 2024-06-14
Flaky TestManualSectorOnboarding - post sector in wrong deadline


rvagg:
```
    node_unmanaged.go:867: Miner f01002: WindowPoST(101): SectorPartition: &{Deadline:0 Partition:0}
    node_unmanaged.go:875: Miner f01002: WindowPoST(101): ProvingDeadline: &{CurrentEpoch:927 PeriodStart:59 Index:47 Open:2879 Close:2939 Challenge:2859 FaultCutoff:2809 WPoStPeriodDeadlines:48 WPoStProvingPeriod:2880 WPoStChallengeWindow:60 WPoStChallengeLookback:20 FaultDeclarationCutoff:70}
```

Comes back via the error channel to:

```
    manual_onboarding_test.go:138: 
        	Error Trace:	/home/runner/work/lotus/lotus/itests/manual_onboarding_test.go:138
        	            				/home/runner/work/lotus/lotus/itests/manual_onboarding_test.go:117
        	Error:      	Received unexpected error:
        	            	Miner(f01002): sector 101 is not in the deadline 0, but 47
        	Test:       	TestManualSectorOnboarding/WithRealProofs
```

logs are full of lots and lots of these:

```
    node_unmanaged.go:611: Sector 101: WindowPoSt submission failed: Miner(f01002): sector 101 is not in the deadline 0, but 47
    node_unmanaged.go:641: Sector 101: WindowPoSt #2 submitted
    node_unmanaged.go:867: Miner f01002: WindowPoST(101): SectorPartition: &{Deadline:0 Partition:0}
    node_unmanaged.go:875: Miner f01002: WindowPoST(101): ProvingDeadline: &{CurrentEpoch:2939 PeriodStart:2939 Index:47 Open:5759 Close:5819 Challenge:5739 FaultCutoff:5689 WPoStPeriodDeadlines:48 WPoStProvingPeriod:2880 WPoStChallengeWindow:60 WPoStChallengeLookback:20 FaultDeclarationCutoff:70}
```

What are we doing wrong here that we get `Deadline:0 Partition:0` but then fail with `sector 101 is not in the deadline 0, but 47`? Is this a snap deal thing, does snapping change the deadline and we're not updating?

rvagg:
Sorry, this meant to include a link to the CI failure with full logs, here it is: https://github.com/filecoin-project/lotus/actions/runs/9510760462/job/26216079595

rvagg:
[...]

PR #12113, opened 2024-06-19
fix: itest: fix wdpost proving epoch calculation

rvagg:
Fixes: https://github.com/filecoin-project/lotus/issues/12091 (I think)

Pulled out of #12105 where I was forced to figure it out and fix it.

### DIFF
--- a/itests/kit/blockminer.go
+++ b/itests/kit/blockminer.go
@@ -100,6 +100,7 @@ func newPartitionTracker(ctx context.Context, t *testing.T, client v1api.FullNod
 
 	parts, err := client.StateMinerPartitions(ctx, minerAddr, dlIdx, types.EmptyTSK)
 	require.NoError(t, err)
+
 	return &partitionTracker{
 		minerAddr:  minerAddr,
 		partitions: parts,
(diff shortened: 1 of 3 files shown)


itests/kit/blockminer.go
```diff
@@ -245,7 +246,9 @@ func (bm *BlockMiner) MineBlocksMustPost(ctx context.Context, blocktime time.Dur
 				dlinfo, err := bm.miner.FullNode.StateMinerProvingDeadline(ctx, minerAddr, ts.Key())
 				require.NoError(bm.t, err)
 				require.NotNil(bm.t, dlinfo, "no deadline info for miner %s", minerAddr)
-				impendingDeadlines = append(impendingDeadlines, minerDeadline{addr: minerAddr, deadline: *dlinfo})
+				if dlinfo.Open <= dlinfo.CurrentEpoch {
```
aarshkshah1992, line 249: This should be `dlinfo.Open` < `dlinfo.CurrentEpoch` right ? It is fine if it's == `dlinfo.CurrentEpoch`

itests/kit/node_unmanaged.go
```diff
@@ -788,6 +783,20 @@ func (tm *TestUnmanagedMiner) calculateNextPostEpoch(
 
 	tm.t.Logf("Miner %s: WindowPoST(%d): ProvingDeadline: %+v", tm.ActorAddr, sectorNumber, di)
 
+	if di.Open > di.CurrentEpoch {
```
aarshkshah1992, line 786: Just because deadline open is > current epoch, how can we conclude that this is a new sector yet to be ticked by cron ?  This could also be the case for sector that have already been ticked by cron right ?

aarshkshah1992: @rvagg Honestly, I think it'd be much easier to reason about this if we wait for one challenge epoch to pass after onboarding a sector and get the new values.

rvagg: no, this is the miner's deadline info not sector info, I _think_ this only happens upon first enrol

rvagg:
I changed this to wait one challenge window and re-collect deadline `Info` because I'm seeing odd behaviour locally with 47 and am wondering if there's a special case there

rvagg:
I have a bit more confidence in this one after tinkering a bit more and grokking actors a bit better. One thing I'm doing now is not hitting the deadline open exactly, but waiting a bit, because I think we might be getting in too quick before cron updates the deadline index in state. It doesn't explain the 47/0 failures specifically but we'll see if that comes back after this. I'll pull it back in to https://github.com/filecoin-project/lotus/pull/12105 and continue with this.

rvagg:
ironically the test failed in CI for this branch, but it's different I think, might just be a timeout https://github.com/filecoin-project/lotus/actions/runs/9579047475/job/26410654516?pr=12113

rvagg:
been re-running in CI and it seems solid; will continue to test in my niporep branch but I think this is good to go

aarshkshah1992 (approved):

PR #12119, opened 2024-06-19
(WIP) Integrate Fast Finality for Filecoin

Kubuxu:
## Checklist

- [ ] CI is green

### DIFF
--- a/chain/lf3/ec.go
+++ b/chain/lf3/ec.go
@@ -1,0 +1,161 @@
+package lf3
+
+import (
+	"context"
+	"sort"
+	"time"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-f3"
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/lotus/chain/actors/builtin/power"
+	"github.com/filecoin-project/lotus/chain/stmgr"
+	"github.com/filecoin-project/lotus/chain/store"
+	"github.com/filecoin-project/lotus/chain/types"
+	"golang.org/x/xerrors"
+)
+
+type ecWrapper struct {
+	ChainStore   *store.ChainStore
+	StateManager *stmgr.StateManager
+	Manifest     f3.Manifest
+}
+
+type tsWrapper struct {
+	inner *types.TipSet
+}
+
+func (ts *tsWrapper) Key() gpbft.TipSetKey {
+	return ts.inner.Key().Bytes()
+}
+
+func (ts *tsWrapper) Beacon() []byte {
+	// TOOD: verify if correct
+	return ts.inner.MinTicket().VRFProof
+}
+
+func (ts *tsWrapper) Epoch() int64 {
+	return int64(ts.inner.Height())
+}
+
+func (ts *tsWrapper) Timestamp() time.Time {
+	return time.Unix(int64(ts.inner.Blocks()[0].Timestamp), 0)
+}
+
+func wrapTS(ts *types.TipSet) f3.TipSet {
+	if ts == nil {
+		return nil
+	}
+	return &tsWrapper{ts}
+}
+
+// GetTipsetByEpoch should return a tipset before the one requested if the requested
+// tipset does not exist due to null epochs
+func (ec *ecWrapper) GetTipsetByEpoch(ctx context.Context, epoch int64) (f3.TipSet, error) {
+	ts, err := ec.ChainStore.GetTipsetByHeight(ctx, abi.ChainEpoch(epoch), nil, true)
+	if err != nil {
+		return nil, xerrors.Errorf("getting tipset by height: %w", err)
+	}
+	return wrapTS(ts), nil
+}
+
+func (ec *ecWrapper) GetTipset(ctx context.Context, tsk gpbft.TipSetKey) (f3.TipSet, error) {
+	tskLotus, err := types.TipSetKeyFromBytes([]byte(tsk))
+	if err != nil {
+		return nil, xerrors.Errorf("decoding tsk: %w", err)
+	}
+
+	ts, err := ec.ChainStore.GetTipSetFromKey(ctx, tskLotus)
+	if err != nil {
+		return nil, xerrors.Errorf("getting tipset by key: %w", err)
+	}
+
+	return wrapTS(ts), nil
+}
+
+func (ec *ecWrapper) GetHead(_ context.Context) (f3.TipSet, error) {
+	return wrapTS(ec.ChainStore.GetHeaviestTipSet()), nil
+}
+
+func (ec *ecWrapper) GetParent(ctx context.Context, tsF3 f3.TipSet) (f3.TipSet, error) {
+	var ts *types.TipSet
+	if tsW, ok := tsF3.(*tsWrapper); ok {
+		ts = tsW.inner
+	} else {
+		tskLotus, err := types.TipSetKeyFromBytes([]byte(tsF3.Key()))
+		if err != nil {
+			return nil, xerrors.Errorf("decoding tsk: %w", err)
+		}
+		ts, err = ec.ChainStore.GetTipSetFromKey(ctx, tskLotus)
+		if err != nil {
+			return nil, xerrors.Errorf("getting tipset by key for get parent: %w", err)
+		}
+	}
+	parentTs, err := ec.ChainStore.GetTipSetFromKey(ctx, ts.Parents())
+	if err != nil {
+		return nil, xerrors.Errorf("getting parent tipset: %w", err)
+	}
+	return wrapTS(parentTs), nil
+}
+
+func (ec *ecWrapper) GetPowerTable(ctx context.Context, tskF3 gpbft.TipSetKey) (gpbft.PowerEntries, error) {
+	tskLotus, err := types.TipSetKeyFromBytes([]byte(tskF3))
+	if err != nil {
+		return nil, xerrors.Errorf("decoding tsk: %w", err)
+	}
+	ts, err := ec.ChainStore.GetTipSetFromKey(ctx, tskLotus)
+	if err != nil {
+		return nil, xerrors.Errorf("getting tipset by key for get parent: %w", err)
+	}
+	log.Infof("collecting power table for: %d", ts.Height())
+	stCid := ts.ParentState()
+
+	sm := ec.StateManager
+
+	powerAct, err := sm.LoadActor(ctx, power.Address, ts)
+	powerState, err := power.Load(ec.ChainStore.ActorStore(ctx), powerAct)
+
+	var powerEntries gpbft.PowerEntries
+	err = powerState.ForEachClaim(func(miner address.Address, claim power.Claim) error {
+		if claim.QualityAdjPower.LessThanEqual(big.Zero()) {
+			return nil
+		}
+
+		// TODO: optimize
+		ok, err := powerState.MinerNominalPowerMeetsConsensusMinimum(miner)
+		if err != nil {
+			return xerrors.Errorf("checking consensus minimums: %w", err)
+		}
+		if !ok {
+			return nil
+		}
+
+		id, err := address.IDFromAddress(miner)
+		if err != nil {
+			return xerrors.Errorf("transforming address to ID: %w", err)
+		}
+
+		pe := gpbft.PowerEntry{
+			ID:    gpbft.ActorID(id),
+			Power: claim.QualityAdjPower.Int,
+		}
+		waddr, err := stmgr.GetMinerWorkerRaw(ctx, sm, stCid, miner)
+		if err != nil {
+			return xerrors.Errorf("GetMinerWorkerRaw failed: %w", err)
+		}
+		if waddr.Protocol() != address.BLS {
+			return xerrors.Errorf("wrong type of worker address")
+		}
+		pe.PubKey = gpbft.PubKey(waddr.Payload())
+		powerEntries = append(powerEntries, pe)
+		return nil
+	})
+	if err != nil {
+		return nil, xerrors.Errorf("collecting the power table: %w", err)
+	}
+
+	sort.Sort(powerEntries)
+	return powerEntries, nil
+}
--- a/chain/lf3/f3.go
+++ b/chain/lf3/f3.go
@@ -1,0 +1,82 @@
+package lf3
+
+import (
+	"context"
+	"time"
+
+	"github.com/filecoin-project/go-f3"
+	"github.com/filecoin-project/go-f3/blssig"
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/build"
+	"github.com/filecoin-project/lotus/chain/stmgr"
+	"github.com/filecoin-project/lotus/chain/store"
+	"github.com/filecoin-project/lotus/node/modules/dtypes"
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
+	logging "github.com/ipfs/go-log/v2"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"github.com/libp2p/go-libp2p/core/host"
+	"go.uber.org/fx"
+	"golang.org/x/xerrors"
+)
+
+type F3 struct {
+	f3 *f3.F3
+}
+
+type F3Params struct {
+	fx.In
+
+	NetworkName  dtypes.NetworkName
+	PubSub       *pubsub.PubSub
+	Host         host.Host
+	ChainStore   *store.ChainStore
+	StateManager *stmgr.StateManager
+	Datastore    dtypes.MetadataDS
+	Wallet       api.Wallet
+}
+
+var log = logging.Logger("f3")
+
+func New(lc fx.Lifecycle, params F3Params) (*F3, error) {
+	logging.SetLogLevel("f3", "DEBUG")
+	manifest := f3.LocalnetManifest()
+	manifest.NetworkName = gpbft.NetworkName(params.NetworkName)
+	manifest.ECDelay = time.Duration(build.BlockDelaySecs) * time.Second
+	manifest.ECPeriod = manifest.ECDelay
+	manifest.BootstrapEpoch = 210
+	manifest.ECFinality = 200
+	ds := namespace.Wrap(params.Datastore, datastore.NewKey("/f3"))
+	ec := &ecWrapper{
+		ChainStore:   params.ChainStore,
+		StateManager: params.StateManager,
+		Manifest:     manifest,
+	}
+	verif := blssig.VerifierWithKeyOnG1()
+	module, err := f3.New(context.TODO(), 1000 /*TODO expose signing*/, manifest, ds,
+		params.Host, params.PubSub, &signer{params.Wallet}, verif, ec, log)
+
+	if err != nil {
+		return nil, xerrors.Errorf("creating F3: %w", err)
+	}
+
+	var liftimeContext context.Context
+	var cancel func()
+	lc.Append(fx.StartStopHook(
+		func(ctx context.Context) {
+			liftimeContext, cancel = context.WithCancel(ctx)
+			go func() {
+				err := module.Run(0, liftimeContext)
+				if err != nil {
+					log.Errorf("running f3: %+v", err)
+				}
+			}()
+		}, func() {
+			cancel()
+		}))
+
+	return &F3{
+		f3: module,
+	}, nil
+}
--- a/chain/lf3/signer.go
+++ b/chain/lf3/signer.go
@@ -1,0 +1,35 @@
+package lf3
+
+import (
+	"context"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/lotus/api"
+	"golang.org/x/xerrors"
+)
+
+type signer struct {
+	wallet api.Wallet
+}
+
+// Signs a message with the secret key corresponding to a public key.
+func (s *signer) Sign(sender gpbft.PubKey, msg []byte) ([]byte, error) {
+	addr, err := address.NewBLSAddress(sender)
+	if err != nil {
+		return nil, xerrors.Errorf("converting pubkey to address: %w", err)
+	}
+	sig, err := s.wallet.WalletSign(context.TODO(), addr, msg, api.MsgMeta{Type: api.MTUnknown})
+	if err != nil {
+		return nil, xerrors.Errorf("error while signing: %w", err)
+	}
+	return sig.Data, nil
+}
+
+// MarshalPayloadForSigning marshals the given payload into the bytes that should be signed.
+// This should usually call `Payload.MarshalForSigning(NetworkName)` except when testing as
+// that method is slow (computes a merkle tree that's necessary for testing).
+// Implementations must be safe for concurrent use.
+func (s *signer) MarshalPayloadForSigning(nn gpbft.NetworkName, p *gpbft.Payload) []byte {
+	return p.MarshalForSigning(nn)
+}
--- a/go.mod
+++ b/go.mod
@@ -10,9 +10,11 @@ replace github.com/filecoin-project/filecoin-ffi => ./extern/filecoin-ffi // pro
 
 replace github.com/filecoin-project/test-vectors => ./extern/test-vectors // provided via a git submodule
 
+replace github.com/filecoin-project/go-f3 => ../go-f3
+
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2
-	github.com/BurntSushi/toml v1.3.0
+	github.com/BurntSushi/toml v1.3.2
 	github.com/DataDog/zstd v1.4.5
 	github.com/GeertJohan/go.rice v1.0.3
 	github.com/Gurpartap/async v0.0.0-20180927173644-4f7f499dd9ee
@@ -26,7 +28,7 @@ require (
 	github.com/dgraph-io/badger/v2 v2.2007.4
 	github.com/docker/go-units v0.5.0
 	github.com/drand/drand v1.5.11
-	github.com/drand/kyber v1.3.0
+	github.com/drand/kyber v1.3.1
 	github.com/dustin/go-humanize v1.0.1
 	github.com/elastic/go-elasticsearch/v7 v7.14.0
 	github.com/elastic/go-sysinfo v1.7.0
@@ -41,6 +43,7 @@ require (
 	github.com/filecoin-project/go-commp-utils v0.1.3
 	github.com/filecoin-project/go-commp-utils/nonffi v0.0.0-20220905160352-62059082a837
 	github.com/filecoin-project/go-crypto v0.0.1
+	github.com/filecoin-project/go-f3 v0.0.0-20240619112353-e4ef2ea3078f
 	github.com/filecoin-project/go-fil-commcid v0.1.0
 	github.com/filecoin-project/go-hamt-ipld/v3 v3.1.0
 	github.com/filecoin-project/go-jsonrpc v0.3.2
@@ -102,7 +105,7 @@ require (
 	github.com/klauspost/compress v1.17.8
 	github.com/koalacxr/quantile v0.0.1
 	github.com/libp2p/go-buffer-pool v0.1.0
-	github.com/libp2p/go-libp2p v0.34.1
+	github.com/libp2p/go-libp2p v0.35.0
 	github.com/libp2p/go-libp2p-kad-dht v0.25.2
 	github.com/libp2p/go-libp2p-pubsub v0.11.0
 	github.com/libp2p/go-libp2p-record v0.2.0
@@ -130,7 +133,7 @@ require (
 	github.com/stretchr/testify v1.9.0
 	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	github.com/triplewz/poseidon v0.0.0-20230828015038-79d8165c88ed
-	github.com/urfave/cli/v2 v2.25.5
+	github.com/urfave/cli/v2 v2.27.2
 	github.com/whyrusleeping/bencher v0.0.0-20190829221104-bb6607aa8bba
 	github.com/whyrusleeping/cbor-gen v0.1.1
 	github.com/whyrusleeping/multiaddr-filter v0.0.0-20160516205228-e903e4adabd7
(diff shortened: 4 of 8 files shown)


chain/lf3/f3.go
```diff
@@ -0,0 +1,131 @@
...
+func (fff *F3) Participate(ctx context.Context, minerIDAddress uint64) error {
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+
+		ch := make(chan *gpbft.MessageBuilder, 4)
```
masih, line 103: Please document the rationale behind `4` as channel buffer size.

chain/lf3/f3.go
```diff
@@ -0,0 +1,131 @@
...
+		fff.Inner.SubscribeForMessagesToSign(ch)
```
masih, line 104: Handle `closer` returned by `SubscribeForMessagesToSign`?

chain/lf3/f3.go
```diff
@@ -0,0 +1,131 @@
...
+		}))
+
+	return fff, nil
+}
+
+// Participate runs the participation loop for givine minerID
+// it is blocking
+func (fff *F3) Participate(ctx context.Context, minerIDAddress uint64) error {
+
+	for {
```
masih, line 96: This loop is going to run hot if for whatever reason retry attempts fail immediately. 
For safety, we probably should add some artificial delay or break out of the outer loop if we hit some number of iterations too fast.

chain/lf3/signer.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/filecoin-project/go-f3/gpbft"
+
+	"github.com/filecoin-project/lotus/api"
+)
+
+type signer struct {
+	wallet api.Wallet
+}
+
+// Signs a message with the secret key corresponding to a public key.
```
masih, line 18:
```suggestion
// Sign signs a message with the secret key corresponding to a public key.
```
Also say something about the implicit assumptions that the public key must be known by wallet and must be BLS?

go.mod
```diff
@@ -41,6 +41,7 @@ require (
 	github.com/filecoin-project/go-commp-utils v0.1.3
 	github.com/filecoin-project/go-commp-utils/nonffi v0.0.0-20220905160352-62059082a837
 	github.com/filecoin-project/go-crypto v0.0.1
+	github.com/filecoin-project/go-f3 v0.0.0-20240624015334-ee517df37076
```
masih, line 44: This commit does not seem to be on `main`.

Ideally we should depend on a concrete release.
At least we should not depend on a commit that is not on `main`.

node/builder_chain.go
```diff
@@ -148,7 +149,10 @@ var ChainNode = Options(
 		Override(RunPeerMgrKey, modules.RunPeerMgr),
 		Override(HandleIncomingMessagesKey, modules.HandleIncomingMessages),
 		Override(HandleIncomingBlocksKey, modules.HandleIncomingBlocks),
+		//Override(F3Key, lf3.New),
```
masih, line 152: Ditto re commented out code.

node/modules/storageminer.go
```diff
@@ -351,6 +351,25 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+func F3Participation(mctx helpers.MetricsCtx, lc fx.Lifecycle, api v1api.FullNode, minerAddress dtypes.MinerAddress) error {
+	ctx := helpers.LifecycleCtx(mctx, lc)
+	go func() {
+		for {
+			ch, err := api.F3Participate(ctx, address.Address(minerAddress))
+			if errors.Is(err, context.Canceled) {
+				return
+			} else if err != nil {
+				log.Errorf("error while trying to participate in F3: %+v", err)
+				time.Sleep(1 * time.Second)
```
masih, line 363: Document the rationale for sleep? or better, parameterise it.

api/api_full.go
```diff
@@ -859,6 +860,17 @@ type FullNode interface {
 	// Note: this API is only available via websocket connections.
 	// This is an EXPERIMENTAL API and may be subject to change.
 	SubscribeActorEventsRaw(ctx context.Context, filter *types.ActorEventFilter) (<-chan *types.ActorEvent, error) //perm:read
+
+	// F3Participate should be called by a miner node to particpate in signing F3 consensus.
+	// The address should be of type ID
+	// F3Participate can only be used through websocket connection
+	// The returned channel will never be closed by the F3
+	// If it is closed without the context being cancelled, the caller should retry.
+	F3Participate(ctx context.Context, minerID address.Address) (<-chan error, error) //perm:admin
```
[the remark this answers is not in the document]
Kubuxu: `address.ID` is enum not a type: https://github.com/filecoin-project/go-address/blob/master/address.go#L66

chain/lf3/f3.go
```diff
@@ -0,0 +1,131 @@
...
+	manifest.ECFinality = 200
+	ds := namespace.Wrap(params.Datastore, datastore.NewKey("/f3"))
+	ec := &ecWrapper{
+		ChainStore:   params.ChainStore,
+		StateManager: params.StateManager,
+		Manifest:     manifest,
+	}
+	verif := blssig.VerifierWithKeyOnG1()
+
+	module, err := f3.New(context.TODO(), 1000 /*TODO expose signing*/, manifest, ds,
```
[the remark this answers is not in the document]
Kubuxu: It is not used anymore, I forgot to remove the param.

chain/lf3/f3.go
Kubuxu: There is no need; abandoning the channel is a valid strategy for closing.

go.mod
Kubuxu: Yes, it is my latest PR, I'll update: https://github.com/filecoin-project/go-f3/pull/367

chain/lf3/f3.go
masih: Then please assign to `_` and explicitly document why closer is not called?

chain/lf3/ec.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+	parentTs, err := ec.ChainStore.GetTipSetFromKey(ctx, ts.Parents())
+	if err != nil {
+		return nil, xerrors.Errorf("getting parent tipset: %w", err)
+	}
+	return wrapTS(parentTs), nil
+}
+
+func (ec *ecWrapper) GetPowerTable(ctx context.Context, tskF3 gpbft.TipSetKey) (gpbft.PowerEntries, error) {
```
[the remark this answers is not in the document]
Kubuxu: PowerEntries are serializable, while PowerTable is not.
Also for the majority of tipsets, we don't need the gpbft representation of the power table, we just need a hash of entires.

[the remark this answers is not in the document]
masih: > PowerEntries are serializable, while PowerTable is not.

Since `PowerTable` should be reproducible from its entries, that seems like a complexity to capture inside the `PowerTable` marshaller.

> for the majority of tipsets, we don't need the gpbft representation of the power table, we just need a hash of entires.

For those cases we should explicitly pass a powertable CID? 


Neither of the comments above are blockers. This just occurs to me as a debt.

chain/lf3/f3.go
Kubuxu: It won't run hot, as in the majority of failure scenarios, it exists. This' for' is only in effect when we get dropped by the eventbus.

go.mod
Kubuxu: We are now pointing to main

chain/lf3/f3.go
Kubuxu: It run hot in tests, fixed it.

node/modules/storageminer.go
```diff
@@ -351,6 +352,44 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
 	return sst, nil
 }
 
+func F3Participation(mctx helpers.MetricsCtx, lc fx.Lifecycle, api v1api.FullNode, minerAddress dtypes.MinerAddress) error {
+	ctx := helpers.LifecycleCtx(mctx, lc)
```
jennijuju, line 356: if f3enabled = false, skip

node/modules/storageminer.go
```diff
@@ -351,6 +352,44 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
 	return sst, nil
 }
 
+func F3Participation(mctx helpers.MetricsCtx, lc fx.Lifecycle, api v1api.FullNode, minerAddress dtypes.MinerAddress) error {
```
rvagg, line 355: why is this needed in miner?

Kubuxu: Keys and libp2p are in Lotus, but Lotus itself has no idea what miners are running that are using it.
So this is the miner telling Lotus: I want you to participate in the F3 network for me.

node/impl/full/f3.go
```diff
@@ -0,0 +1,60 @@
...
+
+var ErrF3Disabled = errors.New("F3 is disabled")
+
+func (f3api *F3API) F3Participate(ctx context.Context, miner address.Address) (<-chan string, error) {
+	// make channel with some buffere to avoid blocking under higher load
+	errCh := make(chan string, 4)
+
+	if f3api.F3 == nil {
+		log.Infof("F3Participate called for %v, F3 is disabled", miner)
+		// we return a channel that will never be closed
```
rvagg, line 30: why not just error so they know it's not possible?

node/modules/storageminer.go
rvagg: got it

node/impl/full/f3.go
masih: I think this is because the implementation treats any error returned as the second return type as retriable, except context cancellation. See [here](https://github.com/filecoin-project/lotus/blob/9a7f0f27b82e0e3f96447f96a1147b44e9836d76/node/modules/storageminer.go#L373). 

Returning a never-closed channel affectively signals to downstream to "do nothing". This is a bit convoluted considering `ErrF3Disabled` exists. 

Assuming I have not missed anything I will go ahead and make this change. Cc @Kubuxu

BigLep:
2024-06-27 nv23 upgrade standup conversation: before merging, expecting @Stebalien and (@aarshkshah1992 or @rvagg)  to review before merge.

api/api_full.go
```diff
@@ -859,6 +860,19 @@ type FullNode interface {
 	// Note: this API is only available via websocket connections.
 	// This is an EXPERIMENTAL API and may be subject to change.
 	SubscribeActorEventsRaw(ctx context.Context, filter *types.ActorEventFilter) (<-chan *types.ActorEvent, error) //perm:read
+
+	// F3Participate should be called by a miner node to particpate in signing F3 consensus.
+	// The address should be of type ID
+	// F3Participate can only be used through websocket connection
```
Stebalien, line 866: Why? IIRC, channels work over HTTP as well.

build/params_calibnet.go
```diff
@@ -152,3 +151,6 @@ const BootstrapPeerThreshold = 4
 const Eip155ChainId = 314159
 
 var WhitelistedBlock = cid.Undef
+
+const F3Enabled = false
+const F3BootstrapEpoch abi.ChainEpoch = 999999999999999
```
Stebalien, line 156: -1?

Stebalien: Or just "max int".

build/params_testground.go
```diff
@@ -133,9 +133,11 @@ var (
...
 	ZeroAddress = MustParseAddress("f3yaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaby2smx7a")
 
-	WhitelistedBlock  = cid.Undef
-	BootstrappersFile = ""
-	GenesisFile       = ""
+	WhitelistedBlock                 = cid.Undef
+	BootstrappersFile                = ""
+	GenesisFile                      = ""
+	F3Enabled                        = false
+	F3BootstrapEpoch  abi.ChainEpoch = -30
```
Stebalien, line 140: Comment on what this means?

node/impl/full/f3.go
```diff
@@ -0,0 +1,59 @@
...
+
+	actorID, err := address.IDFromAddress(miner)
+	if err != nil {
+		return nil, xerrors.Errorf("miner address in F3Participate not of ID type: %w", err)
+	}
+
+	go func() {
+		// Participate takes control of closing the channel
+		f3api.F3.Participate(ctx, actorID, errCh)
+	}()
```
Stebalien, line 43: nit:
```suggestion
	// Participate takes control of closing the channel
	go f3api.F3.Participate(ctx, actorID, errCh)
```

chain/lf3/f3.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/build"
+	"github.com/filecoin-project/lotus/chain/stmgr"
+	"github.com/filecoin-project/lotus/chain/store"
+	"github.com/filecoin-project/lotus/node/modules/dtypes"
+	"github.com/filecoin-project/lotus/node/modules/helpers"
+)
+
+type F3 struct {
+	Inner *f3.F3
```
Stebalien, line 29:
```suggestion
	*f3.F3
```

Stebalien: Unless there's a reason we don't want to expose all of F3.

node/modules/storageminer.go
```diff
@@ -351,6 +353,48 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+		Min:    1 * time.Second,
+		Max:    1 * time.Minute,
+		Factor: 1.5,
+		Jitter: false,
+	}
+	go func() {
+		for {
+			if ctx.Err() != nil {
+				return
+			}
```
Stebalien, line 368:
```suggestion
		for ctx.Err() == nil {
```

node/modules/storageminer.go
```diff
@@ -351,6 +353,48 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+
+			for err := range ch {
+				// we have communication with F3 in lotus, reset the backoff
+				b.Reset()
+				if err != "" {
+					log.Warnf("participating in F3 encountered an error: %v", err)
+				}
+			}
+			log.Info("F3Participate exited, retrying")
+			time.Sleep(b.Duration())
```
Stebalien, line 392: Obey context, ideally.

node/modules/storageminer.go
```diff
@@ -351,6 +353,48 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+			if errors.Is(err, context.Canceled) {
+				log.Errorf("Context cancelled while attampting F3 participation: %+v", err)
+				return
+			} else if errors.Is(err, full.ErrF3Disabled) {
+				log.Errorf("Cannot participate in F3 as it is disabled: %+v", err)
+				return
+			} else if err != nil {
+				log.Errorf("while starting to participate in F3: %+v", err)
+				// use exponential backoff to avoid hotloop
+				time.Sleep(b.Duration())
```
Stebalien, line 380: obey context

node/modules/storageminer.go
```diff
@@ -351,6 +353,48 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+				continue
+			}
+
+			for err := range ch {
```
Stebalien, line 384: Put the for loop in a final else clause and combine the sleep statements?

node/modules/storageminer.go
```diff
@@ -351,6 +353,48 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+				// we have communication with F3 in lotus, reset the backoff
+				b.Reset()
+				if err != "" {
+					log.Warnf("participating in F3 encountered an error: %v", err)
+				}
+			}
+			log.Info("F3Participate exited, retrying")
```
Stebalien, line 391: warn, this shouldn't happen unless the reader is really slow.

Stebalien (commented):
Starting with the nits.

Stebalien (commented):
Making progress

chain/lf3/f3.go
masih: I think we shouldn't expose it at all unless there is an explicit reason. Looks like only GetCert and GetLatestCert is used by node impl. I'll export receivers for those explicitly.

build/params_testground.go
masih: I believe this is meant to be "disable value". I have changed it to -1

node/modules/storageminer.go
masih: I have reworked the logic here to what i believe is the intent. I'd be grateful if you give it another look please.

node/modules/storageminer.go
masih: I have reworked the logic here to what i believe is the intent. I'd be grateful if you give it another look please.

chain/lf3/ec.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	powerState, err := power.Load(ec.ChainStore.ActorStore(ctx), powerAct)
+	if err != nil {
+		return nil, xerrors.Errorf("loading power actor state: %w", err)
+	}
+
+	var powerEntries gpbft.PowerEntries
+	err = powerState.ForEachClaim(func(minerAddr address.Address, claim power.Claim) error {
+		if claim.QualityAdjPower.LessThanEqual(big.Zero()) {
```
Stebalien, line 140: nit:

```suggestion
		if claim.QualityAdjPower.Sign() <= 0 {
```

Stebalien (commented):
Making progress

chain/lf3/f3.go
```diff
@@ -0,0 +1,159 @@
...
+			return nil
+		}
+
+	inner:
+		for {
+			select {
+			case mb, ok := <-msgCh:
+				if !ok {
+					// the broadcast bus kicked us out
+					log.Infof("lost message bus subscription, retrying")
```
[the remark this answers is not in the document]
Stebalien: I've changed this to a warning, but I'm still a bit concerned... On the other hand, there probably isn't much we can do.

chain/lf3/ec.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (ts *tsWrapper) Key() gpbft.TipSetKey {
+	return ts.inner.Key().Bytes()
+}
+
+func (ts *tsWrapper) Beacon() []byte {
+	entries := ts.inner.Blocks()[0].BeaconEntries
+	if len(entries) == 0 {
+		return []byte{}
```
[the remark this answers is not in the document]
Stebalien: Because https://github.com/filecoin-project/go-f3/blob/69e5251b1d1e8ff91c48b5b1c7527ee84e81ebe3/gpbft/message_builder.go#L121-L123? Because if that's the case, there needs to be a comment here.

Stebalien (approved):

Stebalien:
I've addressed most of my own comments, but there are a few outstanding ones.

chain/lf3/f3.go
```diff
@@ -0,0 +1,164 @@
...
+	Datastore    dtypes.MetadataDS
+	Wallet       api.Wallet
+}
+
+var log = logging.Logger("f3")
+
+func New(mctx helpers.MetricsCtx, lc fx.Lifecycle, params F3Params) (*F3, error) {
+	manifest := f3.LocalnetManifest()
+	manifest.NetworkName = gpbft.NetworkName(params.NetworkName)
+	manifest.ECDelay = 2 * time.Duration(build.BlockDelaySecs) * time.Second
```
aarshkshah1992, line 52: So if `EcDelay` and `EcPeriod` are always the same, why do they both exist in `manifest` ? I assume this will change in the future ?

api/api_full.go
masih: I will remove this doc line for now to avoid confusion.

node/impl/full/f3.go
```diff
@@ -0,0 +1,57 @@
...
+	fx.In
+
+	F3 *lf3.F3 `optional:"true"`
+}
+
+var ErrF3Disabled = errors.New("f3 is disabled")
+
+func (f3api *F3API) F3Participate(ctx context.Context, miner address.Address) (<-chan string, error) {
+
+	if f3api.F3 == nil {
```
aarshkshah1992, line 26: I'm having a hard time figuring out where this is initialised. When will this be non nil ?

node/modules/storageminer.go
```diff
@@ -351,6 +353,60 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+		backoffWithContext := func() {
+			timer.Reset(b.Duration())
+			select {
+			case <-ctx.Done():
+				log.Errorf("Context is done while retrying F3 participation: %+v", ctx.Err())
+			case <-timer.C:
+			}
+		}
+
+		for ctx.Err() != nil {
```
aarshkshah1992, line 382: Shouldn;t this be `ctx.Err == nil` ?

node/impl/full/f3.go
aarshkshah1992: Ah it's optional and will not be set if we don't have an instance of `lf3.F3` available which will happen if `build.F3Enabled` is false.

aarshkshah1992: Is that correct ?

node/modules/lp2p/pubsub.go
```diff
@@ -377,6 +379,7 @@ func GossipSub(in GossipIn) (service *pubsub.PubSub, err error) {
 		build.BlocksTopic(in.Nn),
 		build.MessagesTopic(in.Nn),
 		build.IndexerIngestTopic(in.Nn),
+		gpbft.NetworkName(in.Nn).PubSubTopic(),
```
aarshkshah1992, line 382: Is there a way to ensure this is enabled only if `build.F3Enabled` is true ? AFAICT, this will always be enabled right now without any gating.

aarshkshah1992 (commented):
First pass review. Doing one more deeper round now.

chain/lf3/ec.go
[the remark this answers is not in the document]
masih: Yes I believe so. Otherwise CONVERGE message validations would fail. For now I'll comment and leave a TODO.

node/modules/storageminer.go
```diff
@@ -351,6 +353,57 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+func F3Participation(mctx helpers.MetricsCtx, lc fx.Lifecycle, api v1api.FullNode, minerAddress dtypes.MinerAddress) error {
+	ctx := helpers.LifecycleCtx(mctx, lc)
+	b := &backoff.Backoff{
+		Min:    1 * time.Second,
+		Max:    1 * time.Minute,
+		Factor: 1.5,
+		Jitter: false,
+	}
+	go func() {
+		timer := time.NewTimer(0)
```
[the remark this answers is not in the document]
masih: Would it even if reset before first read?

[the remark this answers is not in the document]
masih: I don't believe it does, since `backoffWithContext` always resets the timer before selecting on it. But I'll leave the stop read dance be for safety

```go
func TestTimer_ZeroTimerWithResetDoesNotFireImmediately(t *testing.T) {
	timer := time.NewTimer(0)
	start := time.Now()
	timer.Reset(10 * time.Second)
	<-timer.C
	if time.Since(start) < 10*time.Second {
		t.Fail()
	}
}
```

node/builder_chain.go
```diff
@@ -149,6 +150,8 @@ var ChainNode = Options(
 		Override(HandleIncomingMessagesKey, modules.HandleIncomingMessages),
 		Override(HandleIncomingBlocksKey, modules.HandleIncomingBlocks),
 	),
+
+	If(build.F3Enabled, Override(new(*lf3.F3), lf3.New)),
```
aarshkshah1992, line 154: This will be instantiated even if the node is not a Miner. But AFAICT, only Miners participating in F3 need this. So why not move it to `builder_miner` ?

api/api_full.go
```diff
@@ -859,6 +860,19 @@ type FullNode interface {
...
+	// The returned channel will never be closed by the F3
+	// If it is closed without the context being cancelled, the caller should retry.
+	// The values returned on the channel will inform the caller about participation
+	// Empty strings will be sent if participation succeeded, non-empty strings explain possible errors.
+	F3Participate(ctx context.Context, minerID address.Address) (<-chan string, error) //perm:admin
+	// F3GetCertificate returns a finality certificate at given instance number
+	F3GetCertificate(ctx context.Context, instance uint64) (*certs.FinalityCertificate, error) //perm:read
```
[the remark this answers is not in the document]
masih: No, instance is a monotonically increasing "index" of the GPBFT cycle at which reaching finality is attempted if you like.

build/params_mainnet.go
```diff
@@ -168,3 +168,6 @@ const Eip155ChainId = 314
 
 // WhitelistedBlock skips checks on message validity in this block to sidestep the zero-bls signature
 var WhitelistedBlock = MustParseCid("bafy2bzaceapyg2uyzk7vueh3xccxkuwbz3nxewjyguoxvhx77malc2lzn2ybi")
+
+const F3Enabled = false
+const F3BootstrapEpoch abi.ChainEpoch = -1
```
aarshkshah1992, line 173: Don't we want to enable this codepath in mainnet for NV23 ?

node/modules/storageminer.go
```diff
@@ -351,6 +353,60 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+		if !timer.Stop() {
+			<-timer.C
+		}
+
+		// Backoff while obeying the context.
+		backoffWithContext := func() {
+			timer.Reset(b.Duration())
+			select {
+			case <-ctx.Done():
+				log.Errorf("Context is done while retrying F3 participation: %+v", ctx.Err())
```
aarshkshah1992, line 377: We should return here

aarshkshah1992 (commented):
Done.

aarshkshah1992 (commented):
Finished my review.

node/impl/full/f3.go
masih: yes

aarshkshah1992:
@BigLep I have reviewed this one.

node/builder_chain.go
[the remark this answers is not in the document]
masih: yep

api/api_full.go
```diff
@@ -859,6 +860,18 @@ type FullNode interface {
...
+	// The address should be of type ID
+	// The returned channel will never be closed by the F3
+	// If it is closed without the context being cancelled, the caller should retry.
+	// The values returned on the channel will inform the caller about participation
+	// Empty strings will be sent if participation succeeded, non-empty strings explain possible errors.
+	F3Participate(ctx context.Context, minerID address.Address) (<-chan string, error) //perm:admin
+	// F3GetCertificate returns a finality certificate at given instance number
+	F3GetCertificate(ctx context.Context, instance uint64) (*certs.FinalityCertificate, error) //perm:read
+	// F3GetLatestCertificate returns the latest finality certificate
+	F3GetLatestCertificate(ctx context.Context) (*certs.FinalityCertificate, error) //perm:read
```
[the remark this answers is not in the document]
masih: They will be used very soon once dynamic manifest is integrated for passive testing. I don't think there is much point removing them here and then add back in considering they are already added. Unless it significantly complicates the rc-1 release.

chain/lf3/signer.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/filecoin-project/lotus/api"
+)
+
+type signer struct {
+	wallet api.Wallet
+}
+
+// Sign signs a message with the private key corresponding to a public key.
+// The the key must be known by the wallet and be of BLS type.
+func (s *signer) Sign(sender gpbft.PubKey, msg []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
masih: https://github.com/filecoin-project/go-f3/issues/387

chain/lf3/f3.go
```diff
@@ -0,0 +1,159 @@
...
```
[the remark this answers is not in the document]
masih: The fix has been bubbled up to this PR.

build/params_calibnet.go
```diff
@@ -152,3 +151,6 @@ const BootstrapPeerThreshold = 4
 const Eip155ChainId = 314159
 
 var WhitelistedBlock = cid.Undef
+
```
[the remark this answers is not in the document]
masih: I think the rationale is to do the butterfly net testing first, then enable it. For calibnet we also need a bootstrap PR along with passive testing server on F3 side to dynamically change the manifest.

chain/lf3/f3.go
masih: I believe it will. Cc @Kubuxu

aarshkshah1992 (approved):

chain/lf3/f3.go
Stebalien: (IMO, ECDelay should just be a multiple of ECPeriod, but yeah, it'll probably change)

chain/lf3/f3.go
```diff
@@ -0,0 +1,164 @@
...
+// It is blocking
+func (fff *F3) Participate(ctx context.Context, minerIDAddress uint64, errCh chan<- string) {
+	defer close(errCh)
+
+	for ctx.Err() == nil {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
```
Stebalien, line 101: removeme

chain/lf3/f3.go
```diff
@@ -0,0 +1,164 @@
...
+				log.Errorf("signing message: %+v", err)
+				return err
+			}
+			log.Infof("miner with id %d is sending message in F3", minerIDAddress)
+			fff.inner.Broadcast(ctx, signatureBuilder, payloadSig, vrfSig)
+			return nil
+		}
+
+	inner:
+		for {
```
Stebalien, line 134: nit: `for ctx.Err() == nil`

Stebalien (approved):

aarshkshah1992:
Backported in https://github.com/filecoin-project/lotus/pull/12168.

chain/lf3/ec.go
```diff
@@ -0,0 +1,203 @@
...
+func (ec *ecWrapper) GetHead(_ context.Context) (f3.TipSet, error) {
+	return wrapTS(ec.ChainStore.GetHeaviestTipSet()), nil
+}
+
+func (ec *ecWrapper) GetParent(ctx context.Context, tsF3 f3.TipSet) (f3.TipSet, error) {
+	var ts *types.TipSet
+	if tsW, ok := tsF3.(*f3TipSet); ok {
+		ts = tsW.cast()
+	} else {
+		tskLotus, err := types.TipSetKeyFromBytes(tsF3.Key())
```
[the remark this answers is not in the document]
Kubuxu: Never, unless someone changes something but I don't like to panic if recovery is easy.

PR #12130, opened 2024-06-21
SealProofVariant and lotus-bench NI-PoRep additions

rvagg:
* Introduce a `SealProofVariant` type with standard, synthetic and non-interactive forms so we can do simpler switching between the variants
* Wire up ni-porep to lotus-bench `simple` and add a configurable bash script that can run through the whole sealing process

### DIFF
--- a/cmd/lotus-bench/bench-sectors.sh
+++ b/cmd/lotus-bench/bench-sectors.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+
+# This is an example of how a full sector lifecycle can be benchmarked using `lotus-bench`. The
+# script generates an unsealed sector, runs PC1, PC2, C1, and C2, and prints the duration of each
+# step. The script also prints the proof length and total duration of the lifecycle.
+#
+# Change `flags` to `--non-interactive` to run NI-PoRep, and switch `sector_size` to the desired
+# sector size. The script assumes that the `lotus-bench` binary is in the same directory as the
+# script.
+#
+# Note that for larger sector sizes, /tmp may not have enough space for the full lifecycle.
+
+set -e
+set -o pipefail
+
+tmpdir=/tmp
+
+flags=""
+# flags="--non-interactive"
+sector_size=2KiB
+# sector_size=8MiB
+# sector_size=512MiB
+# sector_size=32GiB
+# sector_size=64GiB
+
+unsealed_file=${tmpdir}/unsealed${sector_size}
+sealed_file=${tmpdir}/sealed${sector_size}
+cache_dir=${tmpdir}/cache${sector_size}
+c1_file=${tmpdir}/c1_${sector_size}.json
+proof_out=${tmpdir}/proof_${sector_size}.hex
+rm -rf $unsealed_file $sealed_file $cache_dir $c1_file
+
+echo "Generating unsealed sector ..."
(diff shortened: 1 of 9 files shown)


cmd/lotus-bench/bench-sectors.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
```
BigLep, line 1: I don't know the team history or view or all the context here, but in general would vote for not seeing this amount of bash getting checked in.

I generally advocate for a "two finger rule" with checked-in bash (i.e., that any bash that extends what I can cover with two fingers on the screen is a signal that instead convert to something else).  This is stemming from a code review and maintenance regard given (my opinion) shell scripting's esoteric and less-ergonomic nature.  

Basically if this is something we care about continuing to work and be modifiable in future, then I would convert it to something else (presumably go) that the team is well-versed in.

rvagg: This is here for the reverse reason! I find this more ergonomic for doing a full seal workflow than what's available in lotus-bench today. I suppose it could be replaced with a `full` but I'm not sure who wants this. I mostly checked this in because I found it useful and didn't want to either discard it or put it in a place where it could get lost; but perhaps it's not useful or interesting to anyone else. I don't know!

BigLep: Got it.  I like the idea of checking it in so we don't lose it.  It seems like we don't know the full utility of this in the long term so can leave it as is.  If this is something we need to use/maintain/support in future, we could make it more robust.  Seems good to me.

rvagg: I'll add some additional notes to this—it may turn out to not be useful and if it's not touched in 6 months then maybe it gets thrown out, that's OK.

rvagg: I've added some docs to the top of the script. We'll see what @aarshkshah1992 thinks of it being in here. I'm fine removing it entirely if you both think it's cruft.

itests/kit/ensemble.go
```diff
@@ -260,7 +260,7 @@ func (n *Ensemble) MinerEnroll(minerNode *TestMiner, full *TestFullNode, opts ..
 		)
 
 		// Will use 2KiB sectors by default (default value of sectorSize).
-		proofType, err := miner.SealProofTypeFromSectorSize(options.sectorSize, n.genesis.version, false)
+		proofType, err := miner.SealProofTypeFromSectorSize(options.sectorSize, n.genesis.version, miner.SealProofVariant_Standard)
```
aarshkshah1992, line 263: Why are we always using `Standard` in the ensemble ? What if tests wanna use NI ?

cmd/lotus-bench/bench-sectors.sh
aarshkshah1992: I had a look at this. Okay to keep it here if it's just to help out people who might want to use this as a reference to understand the full PoRep/NI-PoRep workflow. The script is straight forward to follow.

However, strongly recommend that we translate this to Golang if we're actually using this for benchmarking that workflow.

itests/kit/ensemble.go
rvagg: then someone's going to have to add it - this is for the full lotus-miner and it has no support for dealing with NI, it assumes precommit for a start, so currently it doesn't make sense to parameterise this

aarshkshah1992 (approved):

PR #12131, opened 2024-06-21
fix: ci: update concurrency group settings of the sorted pr checks workflow

galargh:
## Related Issues

n/a

## Proposed Changes

1. Use `workflow_run.head_repository` + `workflow_run.head_branch` in the concurrency group name instead of `workflow_run.id`. This way, the concurrency group should work across all the workflows associated with a single PR.
2. Do not `cancel_in_progress` workflow runs of the sorted PR checks workflow. This should minimise the number of notifications users receive about cancelled workflows.

## Checklist

- [ ] CI is green

### DIFF
--- a/.github/workflows/sorted-pr-checks.yml
+++ b/.github/workflows/sorted-pr-checks.yml
@@ -22,8 +22,8 @@ permissions:
   pull-requests: write
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.inputs.pull_number }}-${{ github.event.workflow_run.id }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event.inputs.pull_number }}-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: false
 
 jobs:
   comment:


rvagg (approved):

rvagg (approved):
@galargh won't both of these be unique within a pull request as well though? where's the additional uniqueness coming from that helps us out here?

galargh:
> @galargh won't both of these be unique within a pull request as well though? where's the additional uniqueness coming from that helps us out here?

Before, we were using `github.event.workflow_run.id` in the concurrency group name. In this context, that would be ID of a run of one of these 4 workflows (`Build`, `Check`, `CodeQL`, `Test`). We run these workflows for each pull request. Each run of theses workflows would end up with a different ID. Thus, the resulting 4 runs of the `Sorted PR Checks` workflow would end up in 4 different concurrency groups.

After, we are using `github.event.workflow_run.head_repository` and `github.event.workflow_run.head_branch`. The values for these for either `Build`, `Check`, `CodeQL` or `Test` would be depending on the pull request that trigger them. More importantly, for 1 PR, each of those 4 workflows would get the same value of `head_repository` and `head_branch`. Thus, the 4 runs of the `Sorted PR Checks` workflow in this case would end up in the same concurrency group and this is how the race condition to check if a comment exists and update or create it is solved.

There is one caveat to this approach, there might be multiple PRs associated with the same `head_repository` and `head_branch`. This is possible to set up in GitHub, but in reality it shouldn't be too common of a pattern.

rvagg (approved):
got it, sgtm

PR #12135, opened 2024-06-21
feat: api: sanity check the "to" address of outgoing messages

Stebalien:
## Proposed Changes

If the "to" address of an outgoing message is a _delegated_ address, verify that it maps to a valid Ethereum address. This isn't a consensus critical change, but it'll help prevent client-side address conversion libraries from directing messages into oblivion (e.g., by mis-translating `0xff0000....` addresses into `f410f...` addresses instead of `f0...` addresses.

Notes:

- We cannot filter these addresses inside pubsub (without a network upgrade) without risking the stability of the pubsub network (all nodes need to agree on which messages are valid and which messages aren't; censorship is treated as an attack).
- We've discussed making it impossible to send funds to "invalid" f4 addresses within the protocol itself, but that has other implications and is a more complicated debate (it means running actor code for method-0 sends in some cases).

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/mempool_test.go
+++ b/itests/mempool_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/big"
 
 	"github.com/filecoin-project/lotus/api"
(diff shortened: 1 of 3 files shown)


rvagg:
This seems fine to me, but here's my understanding of what it's doing: refusing to receive messages for the message pool from the API. One implication is that it would refuse such messages from the large public RPC providers, which would cut off a large % of the source of these messages (I think).

I'm not sure how the pubsub censorship resistance works though so can't really comment on that. Is there any path to receiving and passing them on but just refusing to pack them into blocks? Or is the packing part of the protocol too?

aarshkshah1992:
Added the backport label to this so we have it in the v1.28 release.

aarshkshah1992:
Assigning myself to take this to completion.

Stebalien:
LGTM but I can't approve as I created the PR (and someone else should make sure the logic is fine).

rvagg (approved):

PR #12138, opened 2024-06-24
release:v1.27.1

jennijuju:


jennijuju:
@rvagg - i recall seeing a message either with Vulcanize or blockscout that #11952 in this release requires a migration in the SQLite db but it is not captured in upgrade warning or any release note, and that makes me a bit uncomfortable. could we add a quick note on either a migration guide or if its a small migration - at least what to expect when node operators w/ eth api enabled  should expect when they start a node with v1.27.1 ?

BigLep (approved):
I haven't been close to the changes for this release to give a very good spot check, but this seems right from my scan.  I'll give an "approve" in case this helps unblock and the risk is low.

jennijuju:
Thanks, @BigLep! Given that thisill be included in the next stable release, I will wait for at least another review from the maintainers bef PR marks what wore merging!

.github/workflows/release.yml
```diff
@@ -41,38 +46,61 @@ jobs:
...
       # This enables the job to create and/or update GitHub releases
       contents: write
     runs-on: ubuntu-latest
     needs: [build]
     env:
+<<<<<<< HEAD
       PUBLISH: ${{ startsWith(github.ref, 'refs/tags/') }}
+=======
+      PUBLISH: ${{ startsWith(inputs.ref || github.ref, 'refs/tags/') }}
+>>>>>>> release/v1.27.1
```
rvagg, line 93: got some merge conflicts here

rvagg:
4 differences between this and `release/v1.27.1`

1. The commit conflicts in `.github/workflows/release.yml` that have to be dealt with
2. This has a change that exists on `releases` but isn't on `release/v1.27.1` where `chore: pin golanglint-ci to v1.58.2` appears at the end of the changelog entry for v1.11.1!! I don't think that's a problem for this PR but something to keep an eye on when we merge back in to master.
3. One newline removal in `Makefile` at line 383. Inconsequential but weird.
4. One newline addition in `scripts/version-check.sh`. Inconsequential but weird.

So, good to go with the first one addressed.

jennijuju:
> 4 differences between this and `release/v1.27.1`
> 
> 1. The commit conflicts in `.github/workflows/release.yml` that have to be dealt with
> 2. This has a change that exists on `releases` but isn't on `release/v1.27.1` where `chore: pin golanglint-ci to v1.58.2` appears at the end of the changelog entry for v1.11.1!! I don't think that's a problem for this PR but something to keep an eye on when we merge back in to master.
> 3. One newline removal in `Makefile` at line 383. Inconsequential but weird.
> 4. One newline addition in `scripts/version-check.sh`. Inconsequential but weird.
> 
> So, good to go with the first one addressed.

Thanks! Will fix 1,3,4. And 2 weirded me out as well when I resolved the conflicts. So did a quick check on the code itself and seems it was in place. But notes on the comment on when we merge to master

jennijuju:
(i feel like there should be some checks that catch that not-fully-resolved conflicts but.. yah

rvagg (approved):
no diff to `release/v1.27.1`

PR #12132, opened 2024-06-21
fix: genesis preseal: get it working

aarshkshah1992:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/12121

Need to merge and release Actors and `go-state-types` before merging this.

https://github.com/filecoin-project/builtin-actors/pull/1553
https://github.com/filecoin-project/go-state-types/pull/276

## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] If the PR affects users (e.g., new feature, bug fix, system requirements change), update the CHANGELOG.md and add details to the UNRELEASED section.
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12142, opened 2024-06-25
deps: actor v14 dev & gst v14-rc2

jennijuju:


rvagg:
both included in https://github.com/filecoin-project/lotus/pull/12132 now which has the fix that needs to get this over the line

PR #12143, opened 2024-06-25
chore: test: refactor, fix and improve unmanaged miner

rvagg:
Currently stacked on top of #12130 which should be merged first.

This is from niporep work, extracted directly out of #12105 and includes some necessary bits from there but there are also some generally useful fixes and improvements. One thing I'm working toward is multi-sector aggregates which I haven't _quite_ got to but this refactoring makes it possible. Getting this out of #12105 makes that PR much simpler to deal with.

### DIFF
--- a/itests/kit/node_unmanaged.go
+++ b/itests/kit/node_unmanaged.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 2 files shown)


aarshkshah1992:
@rvagg I am confused. How are these tests green without the Actors bundle containing the NI-PoRep work ?

aarshkshah1992:
Oh okay, because we don't have any NI-PoRep tests in here.

rvagg:
because this doesn't touch niporep in actors, that'll come in the next PR, go-state-types and ffi have all we need for now

rvagg:
> next PR

#12105 - which I'll rebase and squish down to a single commit + actors bundle

aarshkshah1992:
@rvagg Please can you rebase this on master whenever you feel ready ?

itests/kit/node_unmanaged.go
```diff
@@ -440,66 +379,228 @@ func (tm *TestUnmanagedMiner) waitForMutableDeadline(ctx context.Context, sector
...
+	preCommitSealRandEpoch := tm.waitPreCommitSealRandomness(ctx, sectorNumber, proofType)
+
+	// Generate a Pre-Commit for the CC sector -> this persists the proof on the `TestUnmanagedMiner` Miner State
+	tm.generatePreCommit(ctx, sectorNumber, preCommitSealRandEpoch, proofType, pieces)
 
 	// --------------------Create pre-commit for the CC sector -> we'll just pre-commit `sector size` worth of 0s for this CC sector
 
-	// Step 1: Wait for the pre-commitseal randomness to be available (we can only draw seal randomness from tipsets that have already achieved finality)
-	preCommitSealRand := tm.waitPreCommitSealRandomness(ctx, sectorNumber)
+	if !proofType.IsNonInteractive() {
```
aarshkshah1992, line 405: What happens if it is NI ? Will that be addressed in https://github.com/filecoin-project/lotus/pull/12105 ?

itests/kit/node_unmanaged.go
```diff
@@ -938,20 +1012,43 @@ func (tm *TestUnmanagedMiner) generateProveCommit(
 
 	tm.t.Logf("Miner %s: Running proof type %d SealCommitPhase2 for sector %d...", tm.ActorAddr, proofType, sectorNumber)
 
-	sectorProof, err := ffi.SealCommitPhase2(scp1, sectorNumber, actorId)
-	req.NoError(err)
+	var sectorProof []byte
+
+	if proofType.IsNonInteractive() {
+		circuitProofs, err := ffi.SealCommitPhase2CircuitProofs(scp1, sectorNumber)
```
aarshkshah1992, line 1018: So we have the ProveCommit for NI-PoRep here. But what do we replace the `PreCommit` phase with ?

itests/kit/node_unmanaged.go
```diff
@@ -1050,3 +1143,11 @@ func (tm *TestUnmanagedMiner) AssertDisputeFails(ctx context.Context, sector abi
 	require.Contains(tm.t, err.Error(), "failed to dispute valid post")
 	require.Contains(tm.t, err.Error(), "(RetCode=16)")
 }
+
+func (tm *TestUnmanagedMiner) IsUmmutableDeadline(ctx context.Context, deadlineIndex uint64) bool {
```
aarshkshah1992, line 1147: `IsImmutable` ?

itests/kit/node_unmanaged.go
```diff
@@ -440,66 +379,228 @@ func (tm *TestUnmanagedMiner) waitForMutableDeadline(ctx context.Context, sector
...
+	}
+
+	// Step 6: Submit the ProveCommit to the network
+	if proofType.IsNonInteractive() {
+		tm.t.Log("Submitting ProveCommitSector ...")
+
+		var provingDeadline uint64 = 7
+		if tm.IsUmmutableDeadline(ctx, provingDeadline) {
+			// avoid immutable deadlines
+			provingDeadline = 5
```
aarshkshah1992, line 460: What's this with hard coding the "5" and "7" deadlines here ?

itests/kit/node_unmanaged.go
```diff
@@ -440,66 +379,228 @@ func (tm *TestUnmanagedMiner) waitForMutableDeadline(ctx context.Context, sector
...
-		// Step 2: Write empty bytes that we want to seal i.e. create our CC sector
-		tm.makeAndSaveCCSector(ctx, sectorNumber)
+		proveCommit = tm.generateProveCommit(ctx, sectorNumber, proofType, seedRandomness, pieces)
+	}
+
+	return seedEpoch, proveCommit
+}
+
+func (tm *TestUnmanagedMiner) SubmitProveCommit(
+	ctx context.Context,
```
aarshkshah1992, line 439: Yeah so we're completely handling `ProveCommits` for NI-PoRep proofs but I am wondering if something is missing with replacing the PC1/PC2 for these proofs.

itests/kit/node_unmanaged.go
```diff
@@ -440,66 +379,228 @@ func (tm *TestUnmanagedMiner) waitForMutableDeadline(ctx context.Context, sector
...
+	sectorNumber abi.SectorNumber,
+	seedEpoch abi.ChainEpoch,
+	proveCommit []byte,
+	pieceManifest []miner14.PieceActivationManifest,
+) {
+
+	req := require.New(tm.t)
+
+	if proofType.IsNonInteractive() {
+		req.Nil(pieceManifest, "piece manifest should be nil for Non-interactive PoRep")
```
aarshkshah1992, line 450: We should also directly fail the call to `SnapDeal` then if proof type is NI.

aarshkshah1992 (commented):
Mostly lgtm but I'm failing to build a mental model of how PC1/PC2 works here for NI PoReps.

itests/kit/node_unmanaged.go
rvagg: nope, nothing needed here for it - all we do is skip precommit and proceed as if we'd done it and it just works

rvagg: * interactive is: PC1, PC2, `<submit precommit>`, C1, C2, `<submit provecommit>`
* non-interactive is simply: PC1, PC2, C1, C2, `<submit provecommit>`

itests/kit/node_unmanaged.go
rvagg: See above, absolutely nothing, it just flows through PC1, PC2, C1, C2(w/ raw circuit proofs output)

itests/kit/node_unmanaged.go
rvagg: heh, yeah, fixed

itests/kit/node_unmanaged.go
rvagg: I just want a stable deadline to assert against when it lands (see below from here), but if it's immutable then it won't work and this test will be flaky, so I have to move it to another stable deadline. It can't be 6 because we might be in 6 (i.e. 6&7 are immutable), so I jump back to 5 which I know will be mutable since it's 2 epochs removed from now and we're unlikely to accidentally trip forward into an immutable one. I guess >9 would also be somewhat safe but <=5 seems like the safest option.

itests/kit/node_unmanaged.go
rvagg: nope, it's that simple; the only change is that we take "circuit proofs" instead - if we didn't take "circuit proofs" then we'd get an aggregate output from C2, but since we changed `ProveCommitSectorsNI` to only accept aggregate we have to collect all circuits from all sectors and aggregate them all together

you can see this work in https://github.com/filecoin-project/lotus/blob/master/cmd/lotus-bench/bench-sectors.sh, one of the experiments I had to work through to get this all figured out

itests/kit/node_unmanaged.go
rvagg: see https://github.com/filecoin-project/lotus/pull/12105, where we snap in a deal into an ni-porep sector

ni-porep is for CC only, no option to specify pieces or commd, but it's designed to onboard CC so you can snap deals into it later. One of the things covered in https://medium.com/@filoz/simplifying-the-onboarding-pipeline-f3b26ccbb758 is the ideal of SaaS CC sectors -> snap deals to "simplify" the data onboarding process. Perhaps this could end up being the primary data onboarding pathway into the future, even doing away with the need for PCS3

aarshkshah1992 (approved):

Issue #12001, opened 2024-05-16
Tracking Issue: Flaky Tests in GitHub Actions


rjan90:
## Description

This tracking issue is to monitor the investigation and resolution of flaky/failing tests observed only in GitHub Actions. These tests have shown at most 2 failures in 54 runs. (Ref: #11786)

## List of Flaky Tests

```[tasklist]
### Tasks
- [ ] `itest-path_type_filters`
- [ ] `itest-eth_filter`
- [ ] `itest-wdpost`
- [ ] `itest-deals_pricing`
- [ ] `itest-sector_finalize_early`
- [ ] `itest-wdpost_worker_config`
- [ ] `itest-cli`
- [ ] `multicore-sdr`
- [ ] `itest-harmonydb`
```

jennijuju:
@aarshkshah1992 with your remove market PR, will itest-deals_pricing be gone as well?

rjan90:
Some additional notes on a couple of these tests:

- `itest-harmonydb` can probably be removed once Curio has migrated succesfully to their own repo.
- `multicore-sdr` in the [PR adding it](https://github.com/filecoin-project/lotus/pull/6479) it was noted that `"the test works when run individually, but broken when it runs with any other tests as proofs expect the call to FilInitLogFd to be the very first proofs call."`
- `itest-sector_finalize_early` tests finalizing a sector (i.e moving it to the long-term storage), before the ProveCommit-message is sent. The option/config was added iirc, because there where a lot of issues early on that could cause the FinalizeStep (moving a sealed sector to its LTS) to fail, that raised the want for this config. But afaict, the finalize early config is not set by default today.

rjan90:
With the removal of markets in Lotus/Lotus-Miner, these tests has been removed:
- `itest-path_type_filters`
- `itest-deals_pricing`

Ref: https://github.com/filecoin-project/lotus/pull/12099

aarshkshah1992:
Fixed a couple of flaky tests as part of 

https://github.com/filecoin-project/lotus/pull/12203 [eth_filters_itest]
https://github.com/filecoin-project/lotus/pull/12200 [eth_legacy_transaction_itest]

aarshkshah1992:
https://github.com/filecoin-project/lotus/actions/runs/9887807887/job/27310358031?pr=12207

NI-PoRep itest

rvagg:
[...]

PR #12156, opened 2024-06-27
chore: deps: update ffi and go-state-types for NV23

aarshkshah1992:
Update ffi and go-state-types for NV23


extern/filecoin-ffi
aarshkshah1992, line 1: @rvagg I believe this is all that is needed to update ffi.

rvagg: yep, looks good, pointing to e467d2992e3f9bd0 is what we want

aarshkshah1992:
CI failure is a known flaky and has been documented at https://github.com/filecoin-project/lotus/issues/12001.

go.mod
```diff
@@ -33,7 +33,7 @@ require (
 	github.com/elastic/gosigar v0.14.2
 	github.com/etclabscore/go-openrpc-reflect v0.0.36
 	github.com/fatih/color v1.15.0
-	github.com/filecoin-project/filecoin-ffi v0.30.4-0.20220519234331-bfd1f5f9fe38
```
rvagg, line 36: I was scratching my head about this being so old until I remembered about the `replace`, so this isn't technically needed but certainly nice to get it off 2022! can't wait for ffi go binary modules tho.

rvagg (approved):

go.mod
aarshkshah1992: @rvagg I don't think I will be working on ffi go modules anytime soon man. It is a non trivial amount of work and I would like to focus on the ETH/Fevm side of things.

aarshkshah1992: If someone else wants to hack on it -> happy to share my findings/direct them to the docs we've produced for it.

Issue #12165, opened 2024-06-27
Calibration network genesis should have a faucet account for tokens and datacap


aarshkshah1992:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Caliberation network genesis should have a faucet account for tokens and datacap.

### Describe the solution you'd like

Only migrate the `f090` Actor to a keyless Actor if it is a multisig.

BigLep:
Related butterfly issue (https://github.com/filecoin-project/lotus/issues/12160 ), which links to Slack thread with more info.

aarshkshah1992:
@f8-ptrk 

The implementation now ensures that `f090` is a multisig before doing the migration. Please see https://github.com/filecoin-project/go-state-types/pull/281.

rjan90:
We will keep the ticket open until the Calibration upgrade has happened.

rjan90:
Closing ticket as completed now that the calibration upgrade has happened.

Issue #12160, opened 2024-06-27
Butterfly network genesis should have a faucet account for tokens and datacap


aarshkshah1992:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Post FIP-0085, we can no longer use Actor `f090` as the faucet for Butterfly. So we need to update the Butterfly genesis block to contain a new account that can be used as a faucet for tokens and datacap.

### Describe the solution you'd like

The solution should be to update the butterfly bootstrap scripts to create this new account and give it a FIL and datacap balance. Still looking into this will be done.

BigLep:
Relevant slack thread: https://filecoinproject.slack.com/archives/CP50PPW2X/p1719487265531159

rjan90:
- Tested https://faucet.butterfly.fildev.network/funds.html after the reset with Lotus v1.28.0-rc1, and can confirm that I was able to get funds after the NV23 upgrade epoch.

- For getting Datacap through the faucet we need https://github.com/filecoin-project/lotus-infra/issues/1584.
   - To unblock testing we can manually hand out Datacap

rjan90:
Closing this issue as we have confirmed that the Butterfly-faucet works after the NV23 upgrade when resetting with Lotus v1.28.0-rc1

For fixing the Datacap faucet, we need to automatically add RKH address during the reset process of the Butterfly-network, which progress can be tracked in https://github.com/filecoin-project/lotus-infra/issues/1584.
- Its not a blocker for testing/continued rollout of v1.28.0 as we can manually hand out Datacap - although its additional manual labour

PR #12166, opened 2024-06-27
chore: go state types: update deps

aarshkshah1992:
Update `go-state-types` to fix the calibnet and butterfly net faucet issue post NV23 FIP-0085 migration.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,7 +46,7 @@ require (
 	github.com/filecoin-project/go-jsonrpc v0.3.2
 	github.com/filecoin-project/go-padreader v0.0.1
 	github.com/filecoin-project/go-paramfetch v0.0.4
-	github.com/filecoin-project/go-state-types v0.14.0-rc4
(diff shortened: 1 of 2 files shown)


ZenGround0:
Will wait to double check the f090 migration test before approving

aarshkshah1992:
@ZenGround0 

It's green.

BigLep:
@aarshkshah1992 : to be clear, does this address both https://github.com/filecoin-project/lotus/issues/12160 and https://github.com/filecoin-project/lotus/issues/12165 ?

jennijuju (approved):

aarshkshah1992:
@BigLep This does fix both issues. We will now run the `go-state-types`  FIP-0085 migration only if the `090` Actor is a multisig actor. However, it is an account actor for both butterfly and calibnet and so this PR essentially makes FIP-0085 a "no-op" for both butterfly **_and_** calibnet.

PR #12176, opened 2024-07-03
feat:ec:f3: add env var to disable F3

adlrocha:
## Related Issues

Closes https://github.com/filecoin-project/go-f3/issues/388 

## Proposed Changes

This PR:
* Does not run F3 if the `DISABLE_F3` env variable is set.
* Stops F3 in runtime if the `DISABLE_F3` env var is set. The process is currently polling this env variable every 5 minutes, we can poll more often if needed.

## Additional Info

To test it you can run Lotus normally with F3 enabled and then set `DISABLE_F3 = 1  on your environment, this should output a log saying that F3 has been disabled.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/lf3/f3.go
+++ b/chain/lf3/f3.go
@@ -3,6 +3,7 @@ package lf3
 import (
 	"context"
 	"errors"
+	"os"
 	"time"
 
 	"github.com/ipfs/go-datastore"
@@ -46,6 +47,12 @@ type F3Params struct {
 
 var log = logging.Logger("f3")
 
+const (
+	F3DisableEnvKey      = "DISABLE_F3"
+	F3DisableEnvValue    = "_yes_"
+	F3DisableCheckPeriod = 2 * time.Duration(build.BlockDelaySecs)
+)
+
 func New(mctx helpers.MetricsCtx, lc fx.Lifecycle, params F3Params) (*F3, error) {
 	manifest := f3.LocalnetManifest()
 	manifest.NetworkName = gpbft.NetworkName(params.NetworkName)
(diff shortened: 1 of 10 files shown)


node/builder_miner.go
```diff
@@ -141,7 +143,7 @@ func ConfigStorageMiner(c interface{}) Option {
 		Override(new(config.HarmonyDB), cfg.HarmonyDB),
 		Override(new(harmonydb.ITestID), harmonydb.ITestID("")),
 		Override(new(*ctladdr.AddressSelector), modules.AddressSelector(&cfg.Addresses)),
-		If(build.F3Enabled, Override(F3Participation, modules.F3Participation)),
+		If(build.F3Enabled && os.Getenv(lf3.F3DisableEnvKey) != lf3.F3DisableEnvValue, Override(F3Participation, modules.F3Participation)),
```
aarshkshah1992, line 146: @adlrocha I think we need a similar check to disable subscribing to the F3 Gossip topic if this env var is disabled.
I think a  goroutine that does this periodically makes sense there too.

Kubuxu: There is no reason for additional goroutine or anything. If the envvar is set, then F3 won't get constructed by the dependency injection. I'm assuming that we don't change this envvar dynamically during runtime.

masih: I don't know if pubsub would relay messages for topics that are allowed by the subscription filter, _but not subscribed to_. 

In either case I think it is better to not even allow the topic if F3 is disabled. That way nothing can possibly go wrong.

Kubuxu: Yeah, although that is separate from trying to close F3 after envvar is changed. We should probably create `IsF3Enabled()` in `build` use it in the lp2p/pubsub.go and dep injection.

adlrocha: Extended the env var check to pubsub too, I completely missed this one. That being said, there's not much we can do gossipsub-wise without restarting the node if we disable F3 in runtime, as pubsub filtering is a constructor option (AFAIK). Please correct me if I am missing something.

masih: Message relaying will not happen until join/subscribe is called on a topic. Both of those happen deep inside go-f3 once the F3 is instantiated, which means having the topic in allowed list or not does not matter. 

I think this is not a blocker. I defer to @Kubuxu and @aarshkshah1992.

masih (approved):

chain/lf3/f3.go
```diff
@@ -83,6 +90,23 @@ func New(mctx helpers.MetricsCtx, lc fx.Lifecycle, params F3Params) (*F3, error)
 					log.Errorf("running f3: %+v", err)
 				}
 			}()
+			go func() {
+				ticker := time.NewTicker(F3DisableCheckPeriod)
+				for {
+					select {
+					case <-ticker.C:
```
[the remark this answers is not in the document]
adlrocha: You are right, there is no way to disable F3 in runtime through just an env, we would require some config file or a dedicated command to trigger the disable, so this routine does nothing.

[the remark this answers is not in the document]
Kubuxu: It is fine if it is a restart.

masih (approved):

Kubuxu (approved):

PR #12180, opened 2024-07-04
feat: niporep: multi-sector onboarding through UnmanagedMiner

rvagg:
* Test various multi-sector cases with niporep
* UnmanagedMiner can now onboard multiple sectors for one miner and manage PoSt for all of them

This has tests for both real and mock proofs with NI-PoRep, it includes testing 1 and more (up to max 65) sectors in mock proofs, including cases where there are expected sector failures but activation can still continue. They also proceed to confirm Window PoSt on all the sectors and that the miner gets the appropriate power for them all. Then they continue to snap a deal into one of the onboarded sectors (and because of the new PoSt handling, the new snapped sector can be successfully PoSted if need be--although there's no need here beyond just possible accident of snapping and then needing to PoSt before the test ends).

Lots of refactoring in UnmanagedMiner to make this all work, and because I've spent so much time in there there's a fair bit of reorganisation too. Unfortunately that means there's a significant amount of diff churn in here, some of which is just moving things around. There's now only 1 main onboarding method: `TestUnmanagedMiner#OnboardSectors(proofType abi.RegisteredSealProof, withPieces bool, count int, opts ...OnboardOpt) []OnboardedSector`. This handles all cases of onboarding new sectors, including for the existing manual onboarding test. There's also `SnapDeal` to do a single deal snap into a specific sector.

WindowPoSt stuff has been heavily reworked, some of it simplified, although there's new complexity thanks to partition management.

There's an internal `sectorInfo` map for all committed sectors for a miner that retains all properties (we kept a lot of these in separate maps previously). This only gets saved to the miner on commit, but prior to that it gets built up through the precommit and provecommit generation phases. Only after we successfully submit a commit and the sector is acknowledged as successfully activated do we save it to the miner, and then Window PoSt can pick it up, and `SnapDeal` can work with it. There's also `OnboardedSector` which is a subset of sector info that we can return to the caller. I decided not to expose everything but to keep the public bits as simple as is needed for now.

In the NI-PoRep itest, it now has a list of scenarios that it runs through. Currently only the mock proofs ones are run locally or in CI. The real proofs need `LOTUS_RUN_VERY_EXPENSIVE_TESTS=1` and some of them are explicitly skipped so the `skip: true` needs to be commented out. They are mostly useful for now for local testing as they are quite costly and not as useful for a future where we have some way of running `LOTUS_RUN_VERY_EXPENSIVE_TESTS=1` in CI.

### DIFF
--- a/itests/kit/node_unmanaged.go
+++ b/itests/kit/node_unmanaged.go
@@ -6,16 +6,18 @@ import (
 	"crypto/rand"
 	"fmt"
 	"io"
-	"math"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/ipfs/go-cid"
 	libp2pcrypto "github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/stretchr/testify/require"
 	cbg "github.com/whyrusleeping/cbor-gen"
+	"golang.org/x/sync/errgroup"
 
 	ffi "github.com/filecoin-project/filecoin-ffi"
 	"github.com/filecoin-project/go-address"
@@ -23,6 +25,8 @@ import (
 	"github.com/filecoin-project/go-state-types/builtin"
 	miner14 "github.com/filecoin-project/go-state-types/builtin/v14/miner"
 	"github.com/filecoin-project/go-state-types/crypto"
+	"github.com/filecoin-project/go-state-types/dline"
+	"github.com/filecoin-project/go-state-types/exitcode"
 	"github.com/filecoin-project/go-state-types/proof"
 
 	"github.com/filecoin-project/lotus/api"
@@ -32,9 +36,19 @@ import (
 	"github.com/filecoin-project/lotus/chain/wallet/key"
 )
 
+// TODO: this shouldn't be fixed, we should make a new one for each sector/piece
+var fixedPieceCid = cid.MustParse("baga6ea4seaqjtovkwk4myyzj56eztkh5pzsk5upksan6f5outesy62bsvl4dsha")
+
+// 32 bytes of 1's: this value essentially ignored in NI-PoRep proofs, but all zeros is not recommended.
+// Regardless of what we submit to the chain, actors will replace it with 32 1's anyway but we are
+// also doing proof verification call directly when we prepare an aggregate NI proof.
+var niPorepInteractiveRandomness = abi.InteractiveSealRandomness([]byte{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1})
+
 // TestUnmanagedMiner is a miner that's not managed by the storage/infrastructure, all tasks must be manually executed, managed and scheduled by the test or test kit.
 // Note: `TestUnmanagedMiner` is not thread safe and assumes linear access of it's methods
 type TestUnmanagedMiner struct {
+	ctx        context.Context
+	cancelFunc context.CancelFunc
 	t          *testing.T
 	options    nodeOpts
 	mockProofs bool
@@ -44,14 +58,12 @@ type TestUnmanagedMiner struct {
 	sealedSectorDir   string
 	currentSectorNum  abi.SectorNumber
 
-	cacheDirPaths       map[abi.SectorNumber]string
-	unsealedSectorPaths map[abi.SectorNumber]string
-	sealedSectorPaths   map[abi.SectorNumber]string
-	sealedCids          map[abi.SectorNumber]cid.Cid
-	unsealedCids        map[abi.SectorNumber]cid.Cid
-	sealTickets         map[abi.SectorNumber]abi.SealRandomness
+	committedSectorsLk sync.Mutex
+	committedSectors   map[abi.SectorNumber]sectorInfo
 
-	proofType map[abi.SectorNumber]abi.RegisteredSealProof
+	runningWdPostLoop bool
+	postsLk           sync.Mutex
+	posts             []windowPost
 
 	ActorAddr address.Address
 	OwnerKey  *key.Key
@@ -62,27 +74,50 @@ type TestUnmanagedMiner struct {
 	}
 }
 
-type WindowPostResp struct {
-	Posted bool
+// sectorInfo contains all of the info we need to manage the lifecycle of the sector. These
+// properties are private, but could be moved to OnboardedSector if they are useful to calling tests
+// or need to be modified by WithModifyNIActivationsBeforeSubmit.
+type sectorInfo struct {
+	sectorNumber        abi.SectorNumber
+	proofType           abi.RegisteredSealProof
+	cacheDirPath        string
+	unsealedSectorPath  string
+	sealedSectorPath    string
+	sealedCid           cid.Cid
+	unsealedCid         cid.Cid
+	sectorProof         []byte
+	pieces              []abi.PieceInfo
+	sealTickets         abi.SealRandomness
+	sealRandomnessEpoch abi.ChainEpoch
+}
+
+type windowPost struct {
+	Posted []abi.SectorNumber
+	Epoch  abi.ChainEpoch
 	Error  error
 }
 
-func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.Address, mockProofs bool, opts ...NodeOpt) *TestUnmanagedMiner {
-	require.NotNil(t, full, "full node required when instantiating miner")
+func NewTestUnmanagedMiner(ctx context.Context, t *testing.T, full *TestFullNode, actorAddr address.Address, mockProofs bool, opts ...NodeOpt) *TestUnmanagedMiner {
+	req := require.New(t)
+
+	req.NotNil(full, "full node required when instantiating miner")
 
 	options := DefaultNodeOpts
 	for _, o := range opts {
 		err := o(&options)
-		require.NoError(t, err)
+		req.NoError(err)
 	}
 
-	privkey, _, err := libp2pcrypto.GenerateEd25519Key(rand.Reader)
-	require.NoError(t, err)
+	actorIdNum, err := address.IDFromAddress(actorAddr)
+	req.NoError(err)
 
-	require.NotNil(t, options.ownerKey, "owner key is required for initializing a miner")
+	privkey, _, err := libp2pcrypto.GenerateEd25519Key(rand.Reader)
+	req.NoError(err)
+
+	req.NotNil(options.ownerKey, "owner key is required for initializing a miner")
 
 	peerId, err := peer.IDFromPrivateKey(privkey)
-	require.NoError(t, err)
+	req.NoError(err)
 	tmpDir := t.TempDir()
 
 	cacheDir := filepath.Join(tmpDir, fmt.Sprintf("cache-%s", actorAddr))
@@ -93,7 +128,11 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
 	_ = os.Mkdir(unsealedSectorDir, 0755)
 	_ = os.Mkdir(sealedSectorDir, 0755)
 
+	ctx, cancel := context.WithCancel(ctx)
+
 	tm := TestUnmanagedMiner{
+		ctx:               ctx,
+		cancelFunc:        cancel,
 		t:                 t,
 		options:           options,
 		mockProofs:        mockProofs,
@@ -101,18 +140,12 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
 		unsealedSectorDir: unsealedSectorDir,
 		sealedSectorDir:   sealedSectorDir,
 
-		unsealedSectorPaths: make(map[abi.SectorNumber]string),
-		cacheDirPaths:       make(map[abi.SectorNumber]string),
-		sealedSectorPaths:   make(map[abi.SectorNumber]string),
-		sealedCids:          make(map[abi.SectorNumber]cid.Cid),
-		unsealedCids:        make(map[abi.SectorNumber]cid.Cid),
-		sealTickets:         make(map[abi.SectorNumber]abi.SealRandomness),
+		committedSectors: make(map[abi.SectorNumber]sectorInfo),
 
 		ActorAddr:        actorAddr,
 		OwnerKey:         options.ownerKey,
 		FullNode:         full,
-		currentSectorNum: 101,
-		proofType:        make(map[abi.SectorNumber]abi.RegisteredSealProof),
+		currentSectorNum: abi.SectorNumber(actorIdNum * 100), // include the actor id to make them unique across miners and easier to identify in logs
 	}
 	tm.Libp2p.PeerID = peerId
 	tm.Libp2p.PrivKey = privkey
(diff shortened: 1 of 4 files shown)


itests/niporep_manual_test.go
```diff
@@ -29,129 +29,347 @@ import (
...
+					kit.WithRequireActivationSuccess(tcMiner.allOrNothing),
+					kit.WithModifySectorsBeforeCommit(func(sectors []*kit.OnboardedSector) error {
+						head, err := client.ChainHead(ctx) // get the latest chain head before onboarding
+						if err != nil {
+							return err
+						}
+						for j, good := range tcMiner.goodSectorsToOnboard {
+							if !good {
+								// Set the seed epoch too far in the past for this sector to be valid
+								sectors[j].SealRandomnessEpoch = head.Height() - 181*builtin.EpochsInDay
```
rvagg, line 276: I'm belatedly realising this doesn't quite work for my 6good+1bad real-proofs test; I need to alter the strategy here a little

rvagg: fixed

rvagg:
Working

```
--- PASS: TestManualNISectorOnboarding (2080.88s)
    --- PASS: TestManualNISectorOnboarding/mock_proofs,_miners_with_1,_3_(1_bad)_and_6_sectors (63.81s)
    --- PASS: TestManualNISectorOnboarding/mock_proofs,_miner_with_65_sectors (18.37s)
    --- PASS: TestManualNISectorOnboarding/real_proofs,_1_miner_with_1_sector (142.48s)
    --- PASS: TestManualNISectorOnboarding/real_proofs,_1_miner_with_2_sectors (226.80s)
    --- PASS: TestManualNISectorOnboarding/real_proofs,_1_miner_with_3_sectors (255.70s)
    --- PASS: TestManualNISectorOnboarding/real_proofs,_1_miner_with_4_sectors (314.56s)
    --- PASS: TestManualNISectorOnboarding/real_proofs,_1_miner_with_7_sectors (535.01s)
    --- PASS: TestManualNISectorOnboarding/real_proofs,_1_miner_with_7_sectors,_1_bad (524.15s)
```

aarshkshah1992 (commented):
Review 1

itests/kit/node_unmanaged.go
```diff
@@ -93,271 +132,465 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
...
+	}
+
+	// Submit ProveCommit for all sectors
+	exitCodes := tm.submitProveCommit(proofType, sectors, options.requireActivationSuccess, options.modifyNIActivationsBeforeSubmit)
+
+	onboarded := make([]OnboardedSector, count)
+	for i := range sectors {
+		if exitCodes[i] == exitcode.Ok {
+			// only save the sector if it was successfully committed
+			tm.setCommittedSector(sectors[i])
```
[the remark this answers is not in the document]
aarshkshah1992: This is to account for knowingly giving bad sectors in the NI-PoRep tests.

itests/kit/node_unmanaged.go
```diff
@@ -93,271 +132,465 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
...
 }
 
-func (tm *TestUnmanagedMiner) AssertPower(ctx context.Context, raw uint64, qa uint64) {
+// OnboardSectors onboards the specified number of sectors to the miner using the specified proof
+// type. If `withPieces` is true and the proof type supports it, the sectors will be onboarded with
+// pieces, otherwise they will be CC.
+//
+// This method is synchronous but each sector is prepared in a separate goroutine to speed up the
+// process with real proofs.
+func (tm *TestUnmanagedMiner) OnboardSectors(
```
aarshkshah1992, line 191: This pattern of onboard all here -> some are expected to fail and so wont be part of returned onboarded sectors here -> caller then asserts that the returned sectors match their expectation of "good sectors to onboard" did get a bit hard to reason about, especially as there are then code paths here where we check "if exit code == ok {} else {}" but don't fail the test if exit code is expected to be not okay.

I think it will be simpler if caller passes in a list of sector indices here that they expect to fail and we just assert on that here and account/assert on exit codes accordingly.

itests/kit/node_unmanaged.go
```diff
@@ -464,332 +685,424 @@ func (tm *TestUnmanagedMiner) SubmitProveCommit(
...
-	sectorNumber abi.SectorNumber,
-	sealedCid cid.Cid,
-	sealedPath,
-	cacheDir string,
-) (chan WindowPostResp, context.CancelFunc) {
-
-	ctx, cancelFn := context.WithCancel(pctx)
-	respCh := make(chan WindowPostResp, 1)
+func (tm *TestUnmanagedMiner) wdPostLoop() {
+	if tm.runningWdPostLoop {
```
aarshkshah1992, line 804: There should only ever be one wdPostLoop per miner and so this should never happen/something has gone wrong if this happens. I think we should get rid of this variable.

itests/kit/node_unmanaged.go
```diff
@@ -464,332 +685,424 @@ func (tm *TestUnmanagedMiner) SubmitProveCommit(
...
-			case <-ctx.Done():
-			default:
+			if post.Error != nil && tm.ctx.Err() == nil {
+				tm.log("WindowPoSt submission failed for sectors %v at epoch %d: %s", post.Posted, post.Epoch, post.Error)
+			} else if tm.ctx.Err() == nil {
+				postCount++
+				tm.log("WindowPoSt loop completed post %d at epoch %d for sectors %v", postCount, post.Epoch, post.Posted)
 			}
+			tm.postsLk.Lock()
+			tm.posts = append(tm.posts, post)
```
aarshkshah1992, line 831: We should not do this if ` tm.ctx.Err()` is not nil.

Otherwise, there is a race where this is called after the context is cancelled(post submission errors out because of the context cancellation) but we wrongly mark the post as successful by appending it here and error is nil in the appended post.

itests/kit/node_unmanaged.go
```diff
@@ -464,332 +685,424 @@ func (tm *TestUnmanagedMiner) SubmitProveCommit(
...
+	nextDeadlineEpoch := di.PeriodStart + di.WPoStChallengeWindow*abi.ChainEpoch(currentDeadlineIdx+1)
+
+	tm.log("Window PoST waiting until next challenge window, currentDeadlineIdx: %d, nextDeadlineEpoch: %d", currentDeadlineIdx, nextDeadlineEpoch)
+
+	heads, err := tm.FullNode.ChainNotify(ctx)
+	if err != nil {
+		return err
+	}
+
+	for chg := range heads {
```
aarshkshah1992, line 883: Does `ChainNotify` close the returned channel if the context is cancelled. I think it does. But if not, we should handle that here.

itests/kit/node_unmanaged.go
```diff
@@ -464,332 +685,424 @@ func (tm *TestUnmanagedMiner) SubmitProveCommit(
...
 				return
 			}
-			postCount++
-			tm.t.Logf("Sector %d: WindowPoSt #%d submitted", sectorNumber, postCount)
 		}
 	}()
+}
+
+// advancePostLoop waits until we are within the next challenge window for this miner
+func (tm *TestUnmanagedMiner) advancePostLoop() error {
```
aarshkshah1992, line 865: nit: this should be "waitForNextPostDeadline"

aarshkshah1992 (commented):
Review 2

itests/kit/node_unmanaged.go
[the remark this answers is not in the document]
rvagg: yes, exactly - we get a success overall but the batch return tells us which sectors are not OK

adding more comment here

itests/kit/node_unmanaged.go
```diff
@@ -93,271 +132,465 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
...
+	// will also submit a precommit for the sector.
+	for i := 0; i < count; i++ {
+		wg.Add(1)
+		// We hold on to `sector`, adding new properties to it as we go along until we're finished with
+		// this phase, then add it to `sectors`
+		sector := tm.nextSector(proofType)
+		sector.sealRandomnessEpoch = sealRandEpoch
+
+		go func(idx int) {
+			assert := assert.New(tm.t)
```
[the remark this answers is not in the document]
rvagg: ooops!

itests/kit/node_unmanaged.go
```diff
@@ -93,271 +132,465 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
...
+				// Build a sector with non-zero pieces to onboard
+				if tm.mockProofs {
+					sector.pieces = []abi.PieceInfo{{
+						Size:     abi.PaddedPieceSize(tm.options.sectorSize),
+						PieceCID: fixedPieceCid,
+					}}
+				} else {
+					var err error
+					sector, err = tm.mkAndSavePiecesToOnboard(sector)
+					req.NoError(err)
```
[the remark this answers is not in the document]
rvagg: Replaced with the `assert` dance; thanks.

Yeah, so `require` is going to make a fatal and panic, with no opportunity for cleanup, `assert` will just fail and then let you deal with it later. So I've taken the generally recommended approach in this file of anywhere that is called from a goroutine other than the primary test one should either use `assert` or return the error so it can be dealt with properly.

itests/kit/node_unmanaged.go
```diff
@@ -374,88 +607,76 @@ func (tm *TestUnmanagedMiner) waitForMutableDeadline(ctx context.Context, sector
...
-	var seedRandomness abi.InteractiveSealRandomness
-	seedEpoch, seedRandomness = tm.proveCommitWaitSeed(ctx, sectorNumber, proofType)
-
-	proveCommit = []byte{0xde, 0xad, 0xbe, 0xef} // mock prove commit
-	if !tm.mockProofs {
-		proveCommit = tm.generateProveCommit(ctx, sectorNumber, proofType, seedRandomness, pieces)
+			}
+			if len(sector.pieces) > 0 {
+				sectorCopy := sector
+				spci[i].UnsealedCid = &sectorCopy.unsealedCid
```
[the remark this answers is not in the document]
rvagg: trying to be loop-safe for the reference, I think I was originally avoiding a lint error but it seems redundant in the way this ended up.

I've removed this block since it's duplicated

itests/kit/node_unmanaged.go
```diff
@@ -93,271 +132,465 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
...
+			// Generate a PreCommit for the CC sector if required
+			sector, err = tm.generatePreCommit(sector, sealRandEpoch)
+			if assert.NoError(err, "failed to generate PreCommit for CC sector"); tm.t.Failed() {
+				return
+			}
+
+			// Submit the PreCommit to the network if required
+			//TODO: should do this for all sectors in the batch in one go
+			err = tm.preCommitSectors(sealRandEpoch, proofType, []sectorInfo{sector})
+			if assert.NoError(err, "failed to PreCommit CC sector"); tm.t.Failed() {
```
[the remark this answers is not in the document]
rvagg: good call, I think maybe just `if !assert.NoError...` should work instead

itests/kit/node_unmanaged.go
```diff
@@ -93,271 +132,465 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
...
+					sector, err = tm.makeAndSaveCCSector(sector)
+					if assert.NoError(err, "failed to create CC sector"); tm.t.Failed() {
+						return
+					}
+				}
+			}
+
+			// Generate a PreCommit for the CC sector if required
+			sector, err = tm.generatePreCommit(sector, sealRandEpoch)
+			if assert.NoError(err, "failed to generate PreCommit for CC sector"); tm.t.Failed() {
```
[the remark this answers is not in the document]
rvagg: replaced with `if !assert.NoError(...) {`

itests/kit/node_unmanaged.go
```diff
@@ -93,271 +132,465 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
...
+	// For each sector, run PC1, PC2, C1 an C2, preparing for ProveCommit. If the proof needs it we
+	// will also submit a precommit for the sector.
+	for i := 0; i < count; i++ {
+		wg.Add(1)
+		// We hold on to `sector`, adding new properties to it as we go along until we're finished with
+		// this phase, then add it to `sectors`
+		sector := tm.nextSector(proofType)
+		sector.sealRandomnessEpoch = sealRandEpoch
+
+		go func(idx int) {
```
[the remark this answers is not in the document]
rvagg: 👍 done

itests/kit/node_unmanaged.go
rvagg: it started off that way I think I changed it for 2 reasons: fewer arguments for common cases (less a problem now I have Options to pass in) and I wanted to check specific exit codes .. but in the end I think I removed that too!

I'll have a think about it, it felt dirty the first time I tried it but maybe now with options it'll be more pleasant

itests/kit/node_unmanaged.go
rvagg: The current code allows for multiple calls to `OnboardSectors`, although it's not something I've exercised yet, but everything is in place for it to _just work_. At the end of onboarding it runs `tm.wdPostLoop()` and that should be a noop if it doesn't need to start it.

itests/kit/node_unmanaged.go
```diff
@@ -464,332 +685,424 @@ func (tm *TestUnmanagedMiner) SubmitProveCommit(
...
+			tm.postsLk.Lock()
+			tm.posts = append(tm.posts, post)
+			tm.postsLk.Unlock()
 		}
 
-		var postCount int
-		for ctx.Err() == nil {
-			currentEpoch, nextPost, err := tm.calculateNextPostEpoch(ctx, sectorNumber)
-			tm.t.Logf("Activating sector %d, next post %d, current epoch %d", sectorNumber, nextPost, currentEpoch)
+		if err != nil {
```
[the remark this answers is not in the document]
rvagg: weird

itests/kit/node_unmanaged.go
rvagg: 👍 this function did something a bit different originally

itests/kit/node_unmanaged.go
rvagg: yes, it does, I went digging before I did this: https://github.com/filecoin-project/lotus/blob/f08ff209fa33fdd6840758064c294b22048d9aeb/chain/store/store.go#L295-L303

itests/kit/node_unmanaged.go
rvagg: Changed this all to just return `[]abi.SectorNumber` and it'll expect them all to succeed unless you specify a `WithExpectedExitCodes([]exitcode.ExitCode)`. It simplifies a few other things along the way.

itests/kit/node_unmanaged.go
```diff
@@ -464,332 +697,427 @@ func (tm *TestUnmanagedMiner) SubmitProveCommit(
...
-		proofBytes = []byte{0xde, 0xad, 0xbe, 0xef}
-	} else {
-		proofBytes, err = tm.generateWindowPost(ctx, sectorNumber, sealedCid, sealedPath, cacheDir)
+		sp, err := tm.FullNode.StateSectorPartition(tm.ctx, tm.ActorAddr, sectorNumber, head.Key())
 		if err != nil {
-			return fmt.Errorf("Miner(%s): failed to generate window post for sector %d: %w", tm.ActorAddr, sectorNumber, err)
+			return fmt.Errorf("Miner(%s): failed to get sector partition for sector %d: %w", tm.ActorAddr, sectorNumber, err)
+		}
+
+		if di.Index != sp.Deadline {
```
aarshkshah1992, line 1004: This is already checked in the if else below.

itests/kit/node_unmanaged.go
```diff
@@ -464,332 +697,427 @@ func (tm *TestUnmanagedMiner) SubmitProveCommit(
...
+			return fmt.Errorf("Miner(%s): sector %d is not in the deadline %d, but %d", tm.ActorAddr, sectorNumber, sp.Deadline, di.Index)
+		}
+
+		if challengeEpoch == abi.ChainEpoch(-1) {
```
aarshkshah1992, line 1008: This does not need to be in this loop and should be pulled outside. Also, this if else does not make sense. They are not mutually exclusive conditions, right ?

itests/kit/node_unmanaged.go
```diff
@@ -464,332 +697,427 @@ func (tm *TestUnmanagedMiner) SubmitProveCommit(
...
-		return fmt.Errorf("Miner(%s): submitting PoSt for sector %d failed: %s", tm.ActorAddr, sectorNumber, r.Receipt.ExitCode)
+	toSubmitPartitions := make([]uint64, 0)
+	toSubmitSectors := make([]sectorInfo, 0)
+	for partition, sectors := range partitionMap {
+		if len(sectors) == 0 {
+			continue
+		}
+		toSubmitPartitions = append(toSubmitPartitions, partition)
+		toSubmitSectors = append(toSubmitSectors, sectors...)
+		if len(toSubmitPartitions) == miner14.PoStedPartitionsMax {
```
aarshkshah1992, line 1079: So the limit is on the number of partitions we can prove in a single Post but not on the number of sectors that we can prove per partition ?

itests/kit/node_unmanaged.go
```diff
@@ -464,332 +697,427 @@ func (tm *TestUnmanagedMiner) SubmitProveCommit(
...
+
+		// Push all our post messages to the mpool and wait for them later. The blockminer might have
+		// ceased mining while it waits to see our posts in the mpool and if we wait for the first
+		// message to land but it's waiting for a later sector then we'll deadlock.
+		mCid, err := tm.mpoolPushMessage(&miner14.SubmitWindowedPoStParams{
+			ChainCommitEpoch: chainRandomnessEpoch,
+			ChainCommitRand:  chainRandomness,
+			Deadline:         di.Index,
+			Partitions:       pp,
+			Proofs:           []proof.PoStProof{{PoStProof: minerInfo.WindowPoStProofType, ProofBytes: proofBytes}}, // can only have 1
```
aarshkshah1992, line 1060: I see. So we don't need to associate sector proofs with their corresponding partition here. You can just say "here are the partitions I am proving" and "here is one single proof for all the sectors in all those partitions".

itests/niporep_manual_test.go
```diff
@@ -29,129 +29,328 @@ import (
...
+							if !ec.IsSuccess() {
+								// Set the expiration in the past to ensure the sector fails onboarding
+								activations[j].Expiration = head.Height() - 100
+							}
+						}
+						return activations
+					}),
+				)
+
+				req.Len(sectors[i], expectSuccesses)
```
aarshkshah1992, line 284: The `sectors[i]` returned by `OnboardSectors` really contains exit codes for all sectors passed to `OnboardSectors`. So, I am not sure how these can always be equal.

itests/kit/node_unmanaged.go
```diff
@@ -93,271 +128,486 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
...
+		RequireNotificationSuccess: false,
+	}
+	r, err := tm.SubmitMessage(params, 1, builtin.MethodsMiner.ProveReplicaUpdates3)
+	req.NoError(err)
+	req.True(r.Receipt.ExitCode.IsSuccess())
+
+	si.pieces = pieces
+	si.sealedCid = newSealedCid
+	si.unsealedCid = newUnsealedCid
+	tm.setCommittedSector(si)
```
aarshkshah1992, line 402: Right so until this is done, WindowPost loop will use the old `sectorInfo` and will start using the new one as soon as this is in.

itests/kit/node_unmanaged.go
rvagg: Correct, it's fixed to `3` but I'm using the constant for it anyway. The limit for the number of sectors that can be proven is on the number of sectors per partition, and that depends on the sector size: https://github.com/filecoin-project/ref-fvm/blob/844d34a2228a9de7f86c3ca86f9efbbf9ea4215b/shared/src/sector/registered_proof.rs#L175-L188

It's 2 sectors per partition for 2k sectors so that's 6 sectors that can be proven in one post max for 2k.

itests/kit/node_unmanaged.go
rvagg: Yes; and on top of that, when you actually submit it, you have this `Proofs` property which is an array but it can only have 1 in it: https://github.com/filecoin-project/builtin-actors/blob/54236ae89880bf4aa89b0dba6d9060c3fd2aacee/actors/miner/src/lib.rs#L510-L516

I'm not sure what the history is here, but I think you can have multiple for a `VerifyWindowPoSt` call, so you could make the most of that if you wanted to verify your own posts locally.

itests/niporep_manual_test.go
rvagg: it only contains sector numbers now for sectors that returned `exitcode.Ok`, so we should only get the ones successfully onboarded

itests/niporep_manual_test.go
```diff
@@ -29,129 +29,328 @@ import (
...
 			}
+
+			// Verify that ProveCommitSectorsNI rejects messages with invalid parameters
+			verifyProveCommitSectorsNIErrorConditions(ctx, t, miners[0], sealProofType)
 		})
 	}
 }
 
+// This should be run after we've successfully onboarded a sector so we are properly enrolled in
+// cron and the deadline tests work properly.
```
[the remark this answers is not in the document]
rvagg: fair enough, I've put these into a `TestNISectorFailureCases` with just the minimal pieces necessary to run it

itests/kit/node_unmanaged.go
```diff
@@ -93,271 +128,486 @@ func NewTestUnmanagedMiner(t *testing.T, full *TestFullNode, actorAddr address.A
...
+		SectorProofs:               [][]byte{snapProof},
+		UpdateProofsType:           updateProofType,
+		RequireActivationSuccess:   true,
+		RequireNotificationSuccess: false,
+	}
+	r, err := tm.SubmitMessage(params, 1, builtin.MethodsMiner.ProveReplicaUpdates3)
+	req.NoError(err)
+	req.True(r.Receipt.ExitCode.IsSuccess())
+
+	si.pieces = pieces
```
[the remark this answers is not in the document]
rvagg: We're replacing the slice reference on a copy of `sectorInfo`, not modifying the slice itself, if we were messing with the slice then it wouldn't be safe but here we're just telling our local, new `sectorInfo` to point to a difference pieces slice. I'm pretty sure this is safe.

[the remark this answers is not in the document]
aarshkshah1992: Yeah, looking through it again, this is correct.

aarshkshah1992 (approved):

itests/kit/node_unmanaged.go
```diff
@@ -1474,20 +1474,20 @@ func (tm *TestUnmanagedMiner) WaitTillActivatedAndAssertPower(sectors []abi.Sect
...
-
-		if !tm.mockProofs {
+		if !tm.mockProofs { // else it would pass, which we don't want
+			// if the sector is in the current or previous deadline, we can't dispute the PoSt
+			di, err := tm.FullNode.StateMinerProvingDeadline(tm.ctx, tm.ActorAddr, types.EmptyTSK)
+			req.NoError(err)
+			sl, err := tm.FullNode.StateSectorPartition(tm.ctx, tm.ActorAddr, sectorNumber, types.EmptyTSK)
+			req.NoError(err)
+			if sl.Deadline == CurrentDeadlineIndex(di) {
+				tm.log("In current proving deadline for sector %d, waiting for the next deadline to dispute PoSt", sectorNumber)
```
aarshkshah1992, line 1484: @rvagg Also need to ensure it's not in the previous deadline right ?

rvagg: no, there's 2 rules: proving deadline for the sector is closed, and we haven't had more than 2 days elapsed: https://github.com/filecoin-project/builtin-actors/blob/master/actors/miner/src/deadlines.rs#L86-L106

aarshkshah1992 (approved):

PR #12193, opened 2024-07-06
chore: cli: use `embed` pkg to split long template content to file

winniehere:
Splitting long content template into file is more friendly for reading, also for IDE rendering


rvagg (approved):
sgtm, and confirmed that content is identical

PR #12200, opened 2024-07-09
[skip changelog]: fix: flaky eth_legacy_tx_test: wait for async msg indexing

aarshkshah1992:
Fixes the flaky `TestLegacyContractInvocation` test in eth_legacy_tx_test. The message indexing in the eth tx db happens async to submitting the transactions and so we need to wait for the submitted tx to be indexed before we can read it from the tx db.


rvagg (approved):
sgtm

PR #12203, opened 2024-07-09
fix: Eth Tx Events Bloom Filter: fix slice modification bug and flaky test

aarshkshah1992:
This fixes a  bug in the Bloom Filter for events that is part of ETH Tx receipts.  The issue is that all receipts reference the same underlying backing empty array for it's events bloom filter during initialisation. However, modification to the bloom filter then also modifies the backing array... which is then used to initialise other bloom filters.

**_This also fixes the eth_filters flaky itest._**

### DIFF
--- a/chain/types/ethtypes/eth_types.go
+++ b/chain/types/ethtypes/eth_types.go
@@ -188,8 +188,6 @@ type EthBlock struct {
 const EthBloomSize = 2048
 
 var (
-	EmptyEthBloom  = [EthBloomSize / 8]byte{}
-	FullEthBloom   = [EthBloomSize / 8]byte{}
 	EmptyEthHash   = EthHash{}
 	EmptyUncleHash = must.One(ParseEthHash("0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347")) // Keccak-256 of an RLP of an empty array
 	EmptyRootHash  = must.One(ParseEthHash("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")) // Keccak-256 hash of the RLP of null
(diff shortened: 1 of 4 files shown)


node/impl/full/eth_utils.go
```diff
@@ -700,7 +702,7 @@ func newEthTxReceipt(ctx context.Context, tx ethtypes.EthTx, lookup *api.MsgLook
 		BlockNumber:      blockNumber,
 		Type:             ethtypes.EthUint64(2),
 		Logs:             []ethtypes.EthLog{}, // empty log array is compulsory when no logs, or libraries like ethers.js break
-		LogsBloom:        ethtypes.EmptyEthBloom[:],
+		LogsBloom:        LogsBloom,
```
rvagg, line 705: Should we just get rid of `EmptyEthBloom` entirely, because this is the only place it's referenced?

```suggestion
		LogsBloom:        [EthBloomSize / 8]byte{},
```

rvagg: Keeping it does have the nice effect of describing what it's doing. Maybe the better alternative is to turn these into functions:

`newEmptyEthBloom()` and `newFullEthBloom()` and get rid of the globals. Then we also reduce the need for the awkward `copy`

aarshkshah1992: @rvagg Done.

chain/types/ethtypes/eth_types.go
```diff
@@ -203,14 +203,26 @@ func init() {
 	}
 }
 
+func NewEmptyEthBloom() []byte {
+	b := make([]byte, len(EmptyEthBloom))
+	copy(b, EmptyEthBloom[:])
+	return b
```
rvagg, line 209: How about just this, and get rid of `EmptyEthBloom` entirely; there's no need to copy nothing and it's never touched anywhere else.

```suggestion
	return [EthBloomSize / 8]byte{}
```

rvagg (approved):

Issue #12205, opened 2024-07-10
ETH RPC issues


ashwinphatak:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I did not make any code changes to lotus.

### Lotus component

- [X] lotus Ethereum RPC
- [ ] lotus FVM - Lotus FVM interactions
- [ ] FEVM tooling
- [ ] Other

### Lotus Version

```text
Daemon:  1.27.0-dev+mainnet+git.20640c062+api1.5.0
Local: lotus version 1.27.0-dev+mainnet+git.20640c062

and later
```

### Describe the Bug

1. lotus gets stuck (bad block), restarting works temporarily till it hits the same issue
2. snap synced head tracking nodes get slower over time
3. archive node is slow (eth_getLogs especially)
4. null block issue - eth_getLogs throws an error if either `from` or `to` are a null block (doesn't seem to happen with all nodes, needs more analysis)
5. consistency issues at head (https://github.com/filecoin-project/lotus/issues/12078, but same might be the case for other RPC calls like eth_call)

### Tooling

```text
We're using etherjs to make RPC calls
```

### Configuration Options

```text
config.toml


[Fevm]
  # EnableEthRPC enables eth_ rpc, and enables storing a mapping of eth transaction hashes to filecoin message Cids.
  # This will also enable the RealTimeFilterAPI and HistoricFilterAPI by default, but they can be disabled by config options above.
  #
  # type: bool
  # env var: LOTUS_FEVM_ENABLEETHRPC
  #EnableEthRPC = false

  # EthTxHashMappingLifetimeDays the transaction hash lookup database will delete mappings that have been stored for more than x days
  # Set to 0 to keep all mappings
  #
  # type: int
  # env var: LOTUS_FEVM_ETHTXHASHMAPPINGLIFETIMEDAYS
  #EthTxHashMappingLifetimeDays = 0



run script:
```
export LOTUS_CHAINSTORE_ENABLESPLITSTORE="true"
export LOTUS_EVENTS_ENABLEACTOREVENTSAPI="true"
export LOTUS_INDEX_ENABLEMSGINDEX="true"
[...]

PR #12207, opened 2024-07-10
fix: ETH getLogs: fix slowness at head and ignore null blocks

aarshkshah1992:
Temporary patch for the v1.28 release to fix some of the issues mentioned in https://github.com/filecoin-project/lotus/issues/12205.

- Event Index is not aware of null blocks because it is never invoked for null blocks. 
- If user does not explicitly specify "latest", eth_getlogs should inly block will "latest"-1 as the events for the "latest" block will only be available after the block is executed (deferred execution).

The right thing to do here is revisit how all the ETH RPC APIs deal with the "latest" flag and ensure they are "deferred execution-safe". We also need to make the Event Index aware of null blocks(or atleast have a way to ensure that a given height resulted in a null block and so it has already "processed" the events for that height). Let's prioritise this as part of our ETH RPC work.

### DIFF
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -698,11 +698,12 @@ func (ei *EventIndex) GetMaxHeightInIndex(ctx context.Context) (uint64, error) {
 	return maxHeight, err
 }
 
-func (ei *EventIndex) IsHeightProcessed(ctx context.Context, height uint64) (bool, error) {
-	row := ei.stmtIsHeightProcessed.QueryRowContext(ctx, height)
-	var exists bool
-	err := row.Scan(&exists)
-	return exists, err
+func (ei *EventIndex) IsHeightPast(ctx context.Context, height uint64) (bool, error) {
+	maxHeight, err := ei.GetMaxHeightInIndex(ctx)
+	if err != nil {
+		return false, err
+	}
+	return height <= maxHeight, nil
 }
 
 func (ei *EventIndex) IsTipsetProcessed(ctx context.Context, tipsetKeyCid []byte) (bool, error) {
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -1276,8 +1276,14 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
 	if pf.tipsetCid == cid.Undef {
 		maxHeight := pf.maxHeight
 		if maxHeight == -1 {
-			maxHeight = e.Chain.GetHeaviestTipSet().Height()
+			// we can't use the heaviest tipset here as that could block the user on block execution
+			maxHeight = e.Chain.GetHeaviestTipSet().Height() - 1
 		}
+
+		if maxHeight < 0 {
+			return nil, xerrors.Errorf("maxHeight requested is less than 0")
+		}
+
 		if maxHeight > e.Chain.GetHeaviestTipSet().Height() {
 			return nil, xerrors.Errorf("maxHeight requested is greater than the heaviest tipset")
 		}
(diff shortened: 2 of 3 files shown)


node/impl/full/eth.go
```diff
@@ -1286,13 +1292,6 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
 		if err != nil {
 			return nil, err
 		}
-
-		// should also have the minHeight in the filter indexed
-		if b, err := e.EventFilterManager.EventIndex.IsHeightProcessed(ctx, uint64(pf.minHeight)); err != nil {
-			return nil, xerrors.Errorf("failed to check if event index has events for the minHeight: %w", err)
-		} else if !b {
-			return nil, xerrors.Errorf("event index does not have event for epoch %d", pf.minHeight)
```
aarshkshah1992, line 1294: This could just be a null block or an epoch we missed indexing because of en edge case where a node crashes before writing the events to the DB.

The right way to handle this is to introduce

a) Make event Index null block aware
b) Automated backfilling-auto repair in the event Index DB

chain/events/filter/index.go
```diff
@@ -698,11 +698,12 @@ func (ei *EventIndex) GetMaxHeightInIndex(ctx context.Context) (uint64, error) {
 	return maxHeight, err
 }
 
-func (ei *EventIndex) IsHeightProcessed(ctx context.Context, height uint64) (bool, error) {
-	row := ei.stmtIsHeightProcessed.QueryRowContext(ctx, height)
-	var exists bool
-	err := row.Scan(&exists)
-	return exists, err
+func (ei *EventIndex) IsHeightPast(ctx context.Context, height uint64) (bool, error) {
```
aarshkshah1992, line 701: @rvagg 

This changed is to handle:

1) null blocks: The Event Index will never see null blocks - so the best metric for "have you already indexed this height" really is to ensure that we've indexed a height greater than it
2) holes in the event Index caused by hairy edge cases (Event Index sees a block but node crashes before it is indexed). We need auto-repair/auto-backfilling in the Index.

node/impl/full/eth.go
aarshkshah1992: So the best we can do right now is to ensure that we are past the maxHeight in the range. It is not perfect but a strict improvement over what we have till we fix Indexing.

node/impl/full/eth.go
```diff
@@ -1276,8 +1276,14 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
 	if pf.tipsetCid == cid.Undef {
 		maxHeight := pf.maxHeight
 		if maxHeight == -1 {
-			maxHeight = e.Chain.GetHeaviestTipSet().Height()
+			// we can't use the heaviest tipset here as that could block the user on block execution
```
rvagg, line 1279:
```suggestion
			// heaviest tipset doesn't have events because its messages haven't been executed yet
```

node/impl/full/eth.go
```diff
@@ -1276,8 +1276,14 @@ func (e *EthEventHandler) EthGetLogs(ctx context.Context, filterSpec *ethtypes.E
...
+			maxHeight = e.Chain.GetHeaviestTipSet().Height() - 1
 		}
+
+		if maxHeight < 0 {
+			return nil, xerrors.Errorf("maxHeight requested is less than 0")
+		}
+
 		if maxHeight > e.Chain.GetHeaviestTipSet().Height() {
```
rvagg, line 1287: this is an interesting case; if they request this then they'll block for <=30s, maybe we don't want that? :shrug:

node/impl/full/eth.go
rvagg: agreed, but you should leave a note in here, at least a TODO, that this check should ideally be restored in some form

rvagg (approved):
left a couple of small suggestions, I'm mainly interested in leaving good comments in here so we don't forget what needs to be done

node/impl/full/eth.go
aarshkshah1992: I agree. I've fixed this. Looks like `eth_getBlockByNumber` also returns the head -1 tipset for "latest"

```
func getTipsetByBlockNumber(ctx context.Context, chain *store.ChainStore, blkParam string, strict bool) (*types.TipSet, error) {
	if blkParam == "earliest" {
		return nil, fmt.Errorf("block param \"earliest\" is not supported")
	}

	head := chain.GetHeaviestTipSet()
	switch blkParam {
	case "pending":
		return head, nil
	case "latest":
		parent, err := chain.GetTipSetFromKey(ctx, head.Parents())
		if err != nil {
			return nil, fmt.Errorf("cannot get parent tipset")
		}
		return parent, nil
```

node/impl/full/eth.go
aarshkshah1992: Added docs.

aarshkshah1992:
@rvagg Addressed your review here.

rvagg (approved):

aarshkshah1992:
Test is a flaky. Added to https://github.com/filecoin-project/lotus/issues/12001.

PR #12112, opened 2024-06-19
feat: api: Clean API for Miners

snadrus:
This is urgently needed to unblock Curio compatibility with NV23, which is blocking Curio from running on CalibrationNet.

## Related Issues

To help Lotus have more of a clean API, this PR: 
- Enables access to the latest params without depending on lotus/build (which added bulk and risk to Curio). 
- Extracts buildconstants out of lotus/build to enable clean access to that content without the other "build" items.
- "Types" packages (API surface) depend on buildconstants instead of lotus/build
- Rm Curio, including building and testing. (these changes broke it anyway)
- Deprecated access to those vars through lotus/build, but piped them through to change the least code possible

Review is quite easy: the "161 files changed" is mostly deletions. 

## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] If the PR affects users (e.g., new feature, bug fix, system requirements change), update the CHANGELOG.md and add details to the UNRELEASED section.
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12217, opened 2024-07-12
chore!: markets: remove stray unixfs constants, features and references

rvagg:
Noticed we still had some while reviewing #12112, grep revealed more.

This is breaking, in that it removes public constants.

Also some typos golangci-lint picked up for me locally, not sure why they are not being pinged in CI though? Maybe I have a spellchecker installed that it's using.


rvagg:
pushed a commit to remove MinDealDuration & MaxDealDuration

aarshkshah1992 (approved):

Issue #12218, opened 2024-07-12
Flaky test: eth_deploy TestDeployment transaction hash race


rvagg:
Failure example: https://github.com/filecoin-project/lotus/actions/runs/9901634707/job/27354329600#step:10:662

The failure happens here: https://github.com/filecoin-project/lotus/blob/2822404b52a7f7e14ae186b638b2027a41512725/itests/eth_deploy_test.go#L99-L101

So, transaction is submitted, it goes into the messagepool, and we expect `EthGetTransactionByHash` to return immediately with our new message because it's on chain.

The failure happens here: https://github.com/filecoin-project/lotus/blob/2822404b52a7f7e14ae186b638b2027a41512725/node/impl/full/eth.go#L266-L274

We go looking for the hash but can't find it in the translation database.

Normal operation for the eth hash translation database in two separate paths, ([they are both set up here](https://github.com/filecoin-project/lotus/blob/2822404b52a7f7e14ae186b638b2027a41512725/node/modules/ethmodule.go#L80-L86)):

1. On `Apply()`, when messages land on chain, hashes are immediately inserted into the db: https://github.com/filecoin-project/lotus/blob/2822404b52a7f7e14ae186b638b2027a41512725/node/impl/full/txhashmanager.go#L73
2. We have a messagepool watcher that inserts any new message into the database as they are seen: https://github.com/filecoin-project/lotus/blob/2822404b52a7f7e14ae186b638b2027a41512725/node/impl/full/txhashmanager.go#L107-L120

It's this second path that's the problem. We're listening for updates that come in via a goroutine @ https://github.com/filecoin-project/lotus/blob/2822404b52a7f7e14ae186b638b2027a41512725/chain/messagepool/messagepool.go#L1549-L1576 into our own receiving goroutine. So normally it's fast enough, but in some cases there's a race and the insert doesn't happen quick enough for it to be available.

We _could_ fix this by putting a wait in the test that checks the db for the hash, but I think the intention of this test is to show that it's immediately available in the db, which is not unreasonable for Eth APIs I think: you submit a transaction to the node, then you should be able to call `EthGetTransactionByHash` on the submitted hash immediately so you can inspect it deeper with the Lotus APIs. So I can imagine this race hitting users directly.

[...]

PR #12219, opened 2024-07-12
fix: eth: make tx hash immediately available in lookup db after submission

rvagg:
Fixes: https://github.com/filecoin-project/lotus/issues/12218

See #12218 for gory details on why this is necessary. This should fix any case where someone does an eth `sendRawTransaction` and it's not immediately available via other APIs that expect to be able to translate the eth tx hash to a Filecoin message CID (internally or via `EthGetMessageCidByTransactionHash` or `EthGetTransactionHashByCid`).

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -831,6 +831,11 @@ func (a *EthModule) EthSendRawTransaction(ctx context.Context, rawTx ethtypes.Et
 		return ethtypes.EmptyEthHash, err
 	}
 
+	txHash, err := txArgs.TxHash()
+	if err != nil {
+		return ethtypes.EmptyEthHash, err
+	}
+
 	smsg, err := ethtypes.ToSignedFilecoinMessage(txArgs)
 	if err != nil {
 		return ethtypes.EmptyEthHash, err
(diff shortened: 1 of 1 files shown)


node/impl/full/eth.go
```diff
@@ -840,6 +845,12 @@ func (a *EthModule) EthSendRawTransaction(ctx context.Context, rawTx ethtypes.Et
 		return ethtypes.EmptyEthHash, err
 	}
 
+	// make it immediately available in the transaction hash lookup db, even though it will also
+	// eventually get there via the mpool
+	if err := a.EthTxHashManager.TransactionHashLookup.UpsertHash(txHash, smsg.Cid()); err != nil {
+		return ethtypes.EmptyEthHash, err
```
aarshkshah1992, line 851: I don't think we should return an error just because the upsert fails. Otherwise the client will wrongly assume that the txn hasn't been sent to the mpool. Best to just log an error here and return.

aarshkshah1992:
@rvagg 

Even in the current flow, messages that have not landed on chain will be added to this txn -> hash DB on message pool submission (really, it's just a mapping without any guarantees of whether the txn has landed or not). So, this change is fine. However, we should NOT error out the API just because the DB update fails.

node/impl/full/eth.go
rvagg: OK, it's a bit of a no-mans land if it fails but at least there'll be 2 more chances.

aarshkshah1992: Yeah and it's not great to tell clients their tx submission failed but then go around and execute it later.

aarshkshah1992 (approved):

aarshkshah1992:
Please feel free to merge once CI is green.

PR #12232, opened 2024-07-15
lotus v1.27.2-rc2

rvagg:
Let's try out the auto-release builder again.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -318,7 +318,7 @@ jobs:
         env:
           release_id: ${{ fromJSON(steps.after.outputs.release).id }}
           make_latest: ${{ fromJSON(steps.project.outputs.config).latest }}
-          GITHUB_TOKEN: ${{ github.token }}
+          GITHUB_TOKEN: ${{ secrets.TAG_CREATE_GITHUB_TOKEN || github.token }}
         run: |
           gh api -X PATCH /repos/$GITHUB_REPOSITORY/releases/$release_id \
             --field draft=false \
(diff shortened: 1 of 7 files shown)


galargh:
There's one more thing. We have https://github.com/filecoin-project/lotus/settings/rules/575504 set up which restricts who can create tags. In particular, it's going to block creation of a tag on release publish when we merge this PR because we're using the default GitHub token. To overcome this we can:
a. disable the tag creation rule
b. use PAT belonging to eligible user for publishing the releases

For this release, we can either go with a. temporarily or do the publish release step manually after the merge and after the release workflow from the merge finishes/fails (the publish is the last step and nothing else happens after it so if it fail it's OK if we do just that one ourselves). And, in the meantime, I'll set up b. so that we don't have to disable the protection rule.

rvagg:
@galargh is there no way to add GHA to the "bypass" list? That seems like an oversight.

galargh:
Unfortunately, no. The reason they give is that it would allow anyone with push access to bypass the rules. I still think it would be beneficial to have that option since it would allow preventing accidental tag creations.

Anyway, I now set up TAG_CREATE_GITHUB_TOKEN secret and updated the ipdx/release-automation branch so that it uses it.

aarshkshah1992 (approved):

galargh:
https://github.com/filecoin-project/lotus/releases/tag/v1.27.2-rc2 was created by https://github.com/filecoin-project/lotus/actions/runs/9940753120 as expected

Then, docker publish was triggered by the tag creation (which is another reason why we need to use a PAT for publish, really) - https://github.com/filecoin-project/lotus/actions/runs/9941688486

PR #12236, opened 2024-07-15
fix: f3: panic on unset manifest properties

rvagg:
Failing when running off master.

`GpbftConfig` is now set in go-f3's `main` branch when calling `LocalDevnetManifest()`, but `CxConfig` isn't. Perhaps we can have that set as well and then have an updated tag @masih?

```
Jul 15 20:05:10 fletcher lotus-calibnet[385671]: github.com/filecoin-project/go-f3.(*F3).resumeInternal(0xc022820790, {0x8b393d0, 0xc022c30f50})
Jul 15 20:05:10 fletcher lotus-calibnet[385671]:         /home/rvagg/go/pkg/mod/github.com/filecoin-project/go-f3@v0.0.3/f3.go:314 +0x1fa
Jul 15 20:05:10 fletcher lotus-calibnet[385671]: github.com/filecoin-project/go-f3.(*F3).reconfigure(0xc022820790, {0x8b393d0?, 0xc022c30fa0?}, 0xc022c31d10)
Jul 15 20:05:10 fletcher lotus-calibnet[385671]:         /home/rvagg/go/pkg/mod/github.com/filecoin-project/go-f3@v0.0.3/f3.go:254 +0x21c
Jul 15 20:05:10 fletcher lotus-calibnet[385671]: github.com/filecoin-project/go-f3.(*F3).Start(0xc022820790, {0x8b393d0, 0xc022c30fa0})
Jul 15 20:05:10 fletcher lotus-calibnet[385671]:         /home/rvagg/go/pkg/mod/github.com/filecoin-project/go-f3@v0.0.3/f3.go:177 +0x193
Jul 15 20:05:10 fletcher lotus-calibnet[385671]: github.com/filecoin-project/lotus/chain/lf3.New.func1()
Jul 15 20:05:10 fletcher lotus-calibnet[385671]:         /home/rvagg/go/src/github.com/filecoin-project/lotus/chain/lf3/f3.go:79 +0x35
Jul 15 20:05:10 fletcher lotus-calibnet[385671]: go.uber.org/fx/internal/lifecycle.Wrap[...].func1()
```

PR #12237, opened 2024-07-15
chore: f3: Upgrade to latest go-f3 `v0.0.4`

masih:
## Related Issues
#12236

## Proposed Changes
Upgrade to the latest go F3, with fix to default cert exchange config in local devnet.

## Additional Info
None

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/lf3/manifest.go
+++ b/chain/lf3/manifest.go
@@ -11,30 +11,23 @@ import (
 
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/build/buildconstants"
-	"github.com/filecoin-project/lotus/chain/stmgr"
-	"github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/node/modules/dtypes"
 )
 
-func NewManifestProvider(nn dtypes.NetworkName, cs *store.ChainStore, sm *stmgr.StateManager, ps *pubsub.PubSub) manifest.ManifestProvider {
+func NewManifestProvider(nn dtypes.NetworkName, ps *pubsub.PubSub) manifest.ManifestProvider {
 	m := manifest.LocalDevnetManifest()
 	m.NetworkName = gpbft.NetworkName(nn)
 	m.ECDelayMultiplier = 2.
 	m.ECPeriod = time.Duration(build.BlockDelaySecs) * time.Second
 	m.BootstrapEpoch = int64(buildconstants.F3BootstrapEpoch)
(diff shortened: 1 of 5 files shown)


Stebalien (approved):
Note: this also changes the manifest protocol. But we haven't deployed that server yet so we should be fine.

chain/lf3/manifest.go
```diff
@@ -11,30 +11,23 @@ import (
...
 	"github.com/filecoin-project/lotus/build"
 	"github.com/filecoin-project/lotus/build/buildconstants"
-	"github.com/filecoin-project/lotus/chain/stmgr"
-	"github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/node/modules/dtypes"
 )
 
-func NewManifestProvider(nn dtypes.NetworkName, cs *store.ChainStore, sm *stmgr.StateManager, ps *pubsub.PubSub) manifest.ManifestProvider {
+func NewManifestProvider(nn dtypes.NetworkName, ps *pubsub.PubSub) manifest.ManifestProvider {
 	m := manifest.LocalDevnetManifest()
```
rvagg, line 18: @masih both `MinimumPollInterval` and `MaximumPollInterval` are based on `DefaultEcConfig.ECPeriod` in go-f3 but we don't reset them here. _Should_ we? I believe they come out the same here as long as `BlockDelaySecs` is `30`. How much do these matter to align to epoch length?

masih: >  I believe they come out the same here as long as `BlockDelaySecs` is 30

I think regardless of what `BlockDelaySecs` is set to cert exchange min max would be 30s and 2m respectively, because their value is instantiated via `DefaultCxConfig` var using `DefaultEcConfig.ECPeriod`.

For example, this passes for me.

```go
import (
	"testing"
	"time"

	"github.com/filecoin-project/go-f3/manifest"
)

func TestMinMaxPollIntervalIsNotOverridden(t *testing.T) {
	m := manifest.LocalDevnetManifest()
	m.ECPeriod = 1
	if m.CxConfig.MinimumPollInterval != 30*time.Second {
		t.Fatalf("min ain't it")
	}
	if m.CxConfig.MaximumPollInterval != 4*30*time.Second {
		t.Fatalf("max ain't it")
	}
}
```

> How much do these matter to align to epoch length?

Cert exchange internally optimises for the right poll time. As long as min and max are a wide enough interval of time, and min is not less than an epoch we should be good.

@Stebalien does that sound right?

PR #12238, opened 2024-07-15
fix: api itest panic

aarshkshah1992:
1) Fixes flakiness in `testNonGenesisMiner` in `api_test` by de-racing the closing of the miner repo and the sealer reading it in a go-routine. If the go-routine gets scheduled after the miner repo is closed in the test, this causes a panic which fails the test. The right thing to do here is for the sealer to read the repo when it is instantiated and pass the config to the go-routine.

2) Adds some logging to the `testOutOfGasError` itest to see what's going on when it fails (very very rare).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 - This Lotus release includes some correctness improvements to the events subsystem, impacting RPC APIs including `GetActorEventsRaw`, `SubscribeActorEventsRaw`, `eth_getLogs` and the `eth` filter APIs. Part of these improvements involve an events database migration that may take some time to complete on nodes with extensive event databases. See [filecoin-project/lotus#12080](https://github.com/filecoin-project/lotus/pull/12080) for details.
 
+- Breaking change in public APIs `storage/pipeline.NewPreCommitBatcher` and `storage/pipeline.New`. They now have an additional error return to deal with errors arising from fetching the sealing config.
+
 ## New features
 
 - feat: Add trace transaction API supporting RPC method `trace_transaction` ([filecoin-project/lotus#12068](https://github.com/filecoin-project/lotus/pull/12068))
--- a/itests/api_test.go
+++ b/itests/api_test.go
@@ -173,6 +173,8 @@ func (ts *apiSuite) testOutOfGasError(t *testing.T) {
 		buildconstants.BlockGasLimit = originalLimit
 	}()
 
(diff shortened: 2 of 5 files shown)


node/impl/full/gas.go
```diff
@@ -315,6 +315,8 @@ func gasEstimateGasLimit(
 	msg.GasFeeCap = big.Zero()
 	msg.GasPremium = big.Zero()
 
+	log.Infof("gasEstimateGasLimit: msg.GasLimit set to %d, buildconstants.BlockGasLimit: %d", msg.GasLimit, buildconstants.BlockGasLimit)
+
```
rvagg, line 319:
```suggestion
```

I think this is going to be far too noisy running an operational node

rvagg:
This breaks 2 public APIs, which while we may not expect people to be using them, we should at least signal and not just roll this out in a casual patch release.

My preference for signalling breakage is in commit message, which, if we can do consistently, will really help our changelog, so your commit message could be something like:

```
fix!: sealer: handle initialisation error without panic

storage/pipeline.NewPreCommitBatcher and storage/pipeline.New now have an additional
error return to deal with errors arising from fetching the sealing config
```

Also API breakage may be something we shouldn't skip changelog on?

node/impl/full/gas.go
aarshkshah1992: @rvagg Is there a way to only run this on CI ?

rvagg: I guess you could check for the `CI` env var but that seems a little excessive for this path. If you really want it you could make it a Debug, but itests hard-wire log level unfortunately so you wouldn't even see it then.
Does this add much value now that you've addressed the problem? Or do you have enough uncertainty about the fix that you want to keep seeing this?

rjan90:
> Also API breakage may be something we shouldn't skip changelog on?

Yes, we should add this the changelog / write down what the API´s we are breaking - so its clear/understandable what should be added in a `Upgrade Warnings` section of a new release.

aarshkshah1992:
@rvagg  Addressed your review. Please can you review ?

rvagg:
for the record, examples of failures that we're attempting to fix here:

* https://github.com/filecoin-project/lotus/actions/runs/9891743740/job/27322860947
* https://github.com/filecoin-project/lotus/actions/runs/9891029052/job/27320620791
* https://github.com/filecoin-project/lotus/actions/runs/9949190437/job/27485039787

rvagg (approved):

PR #12243, opened 2024-07-16
fix: flaky wdPost dispute itest

aarshkshah1992:
Fixes the flaky wdPost dispute itest.
See https://github.com/filecoin-project/lotus/actions/runs/9468130894/job/26083829500 for an example failed run.

### DIFF
--- a/itests/wdpost_dispute_test.go
+++ b/itests/wdpost_dispute_test.go
@@ -297,7 +297,6 @@ waitForProof:
 
 		deadlines, err := client.StateMinerDeadlines(ctx, maddr, types.EmptyTSK)
 		require.NoError(t, err)
-
 		for dlIdx, dl := range deadlines {
 			nonEmpty, err := dl.PostSubmissions.IsEmpty()
 			require.NoError(t, err)
@@ -319,6 +318,7 @@ waitForProof:
 			Deadline:  targetDeadline,
 			PoStIndex: 0,
 		}
+
 		enc, aerr := actors.SerializeParams(params)
 		require.NoError(t, aerr)
 
--- a/itests/wdpost_dispute_test.go
+++ b/itests/wdpost_dispute_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-bitfield"
+	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/builtin"
 	minertypes "github.com/filecoin-project/go-state-types/builtin/v8/miner"
 	"github.com/filecoin-project/go-state-types/crypto"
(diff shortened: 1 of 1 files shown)


aarshkshah1992:
Raising a separate PR to fix https://github.com/filecoin-project/lotus/actions/runs/9953073160/job/27495762758?pr=12243

itests/wdpost_dispute_test.go
```diff
@@ -330,13 +329,19 @@ waitForProof:
...
 		}
-		_, err := client.MpoolPushMessage(ctx, msg, nil)
+		_, err = client.MpoolPushMessage(ctx, msg, nil)
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "failed to dispute valid post")
 		require.Contains(t, err.Error(), "(RetCode=16)")
 	}
 }
 
+func heightAtLeast(target abi.ChainEpoch) func(set *types.TipSet) bool {
```
rvagg, line 339: isn't this the same as `kit.HeightAtLeast`?

itests/wdpost_dispute_test.go
```diff
@@ -285,36 +286,34 @@ func TestWindowPostDisputeFails(t *testing.T) {
...
+		di, err := client.StateMinerProvingDeadline(ctx, maddr, head.Key())
+		require.NoError(t, err)
+
 		deadlines, err := client.StateMinerDeadlines(ctx, maddr, types.EmptyTSK)
 		require.NoError(t, err)
 		for dlIdx, dl := range deadlines {
 			nonEmpty, err := dl.PostSubmissions.IsEmpty()
 			require.NoError(t, err)
-			if nonEmpty {
+			if nonEmpty && di.Index == uint64(dlIdx) {
```
rvagg, line 303: what's with the naming of this variable? shouldn't it be `isEmpty`?

rvagg: I think also, you could flip this and do `di.Index == uint64(dlIdx)` in an outer `if` and the `IsEmpty` on an inner to avoid the unnecessary call and err check on `IsEmpty`

itests/wdpost_dispute_test.go
```diff
@@ -285,36 +286,34 @@ func TestWindowPostDisputeFails(t *testing.T) {
...
+		head, err := client.ChainHead(ctx)
+		require.NoError(t, err)
+
+		di, err := client.StateMinerProvingDeadline(ctx, maddr, head.Key())
+		require.NoError(t, err)
+
 		deadlines, err := client.StateMinerDeadlines(ctx, maddr, types.EmptyTSK)
 		require.NoError(t, err)
 		for dlIdx, dl := range deadlines {
 			nonEmpty, err := dl.PostSubmissions.IsEmpty()
```
rvagg, line 301: I wonder if maybe the error here is simply that this should be flipped - wait until _one_ deadline is non-empty and that should be enough, no need to compare deadline indexes or anything - the fact that this variable is `nonEmpty` suggests that was the plan, but it should be the inverse -- name it `isEmpty` and wait until it's false and then you're in an epoch where there's a post submission, wait until 5 epochs after the close of it and then dispute.

BigLep:
Slack discussion: https://filecoinproject.slack.com/archives/CP50PPW2X/p1721117282841029

PR #12246, opened 2024-07-16
[skip changelog] chore: types: buildconstants split post-cleanup

ribasushi:
## Related Issues

Another followup to https://github.com/filecoin-project/lotus/pull/12112
🧵 musing about root cause: https://filecoinproject.slack.com/archives/CP50PPW2X/p1721133178627459

## Proposed Changes
Remove multiple uses of deprecated constants. This is a large diff, yet should have exactly zero functional changes.

Ideally as a result of this some parts of the depchain will become lighter, with downstream reaping the same benefits as the team that initiated this split.

P.S. work was done as a result of forming better intuition of current dependency graph


build/buildconstants/drand.go
```diff
@@ -10,3 +16,91 @@ const (
 	DrandIncentinet
 	DrandQuicknet
 )
+
+type DrandConfig struct {
```
aarshkshah1992, line 20: @ribasushi Please can you leave a comment explaining these large DRAND related diffs ?

build/buildconstants/params.go
```diff
@@ -2,18 +2,40 @@ package buildconstants
...
-	BuildCalibnet     = 0x4
-	BuildInteropnet   = 0x5
-	BuildButterflynet = 0x7
+	BuildDefault = iota
+	BuildMainnet
+	Build2k
+	BuildDebug
+	BuildCalibnet
+	BuildInteropnet
+	unused1
```
aarshkshah1992, line 12: This "unused" type seems to be newly introduced. Why do we need it ?

PR #12247, opened 2024-07-17
misc: add waffle to match w/ forest

jennijuju:
Forest has this & it's pretty sweet 😄 can we have the same too?
image


chain/consensus/filcns/upgrades.go
```diff
@@ -63,8 +63,8 @@ import (
 	"github.com/filecoin-project/lotus/node/bundle"
 )
 
-//go:embed FVMLiftoff.txt
```
rvagg, line 66: delete `FVMLiftoff.txt` while you're at it?

jennijuju: done

rvagg (approved):

aarshkshah1992 (approved):

jennijuju:
@rvagg @aarshkshah1992 you guys are 🆒

PR #12251, opened 2024-07-17
fix: lotus-miner: remove provecommit1 method

LexLuthr:
## Proposed Changes

This PR drops the support for ProveCommit1.

## Checklist

- [x] Update CHANGELOG.md or signal that this change does not need it.
- [ ] CI is green


magik6k (commented):
Pretty sure we need to call processPieces with the last arg set to true to convert dealid sectors to ddo-style - https://github.com/filecoin-project/lotus/blob/2ea100928899443c88e92cedd4f1290176076b24/storage/pipeline/states_sealing.go#L854

Also pretty sure we need to convert all precommits to ddo-style, otherwise deals in f05 won't get activated

storage/pipeline/states_sealing.go
```diff
@@ -749,109 +713,71 @@ func (m *Sealing) handleSubmitCommit(ctx statemachine.Context, sector SectorInfo
...
-		}
-	}
+	var preCommit bool
+
 	for _, piece := range sector.Pieces {
 		piece := piece
+		if piece.HasDealInfo() {
+			info := piece.DealInfo()
+			// If we have a dealID then covert to PAM
+			if info.Impl().DealID > 0 {
```
rvagg, line 725: So the two cases, if we have DealInfo, is that `DealID>0` which means we have explicit deal info and need to build a PieceActivationManifest from it ourselves, or the `DealID==0` case which means we've been supplied a PieceActivationManifest and should just use that and ignore the other deal properties?

storage/pipeline/commit_batch.go
```diff
@@ -284,43 +282,6 @@ func (b *CommitBatcher) maybeStartBatch(notif bool) ([]sealiface.CommitBatchRes,
 		return nil, err
 	}
 
-	if individual {
```
rvagg, line 287: It's not entirely clear to me if we ever get to this path now, post-nv22; I think from the larger block above this that if we're post-nv22 that we always end up with `processBatchV2` (btw the comment in there, `ProveCommitSectors2` should probably be changed to `ProveCommitSectors3`) and then clear out those sectors from the list so even when we land in this branch we don't end up calling `processIndividually()` at all.

i.e. this cleanup is mainly a dead-code removal exercise, isn't it? There shouldn't be active use of PCS post-nv22 with lotus-miner using the existing code paths, correct?

storage/pipeline/states_sealing.go
LexLuthr: Piece are not only deal pieces here. They can be fillers for partially filled sectors. Thus the check if we have deal info or not.

storage/pipeline/commit_batch.go
LexLuthr: There was a recent issue from one of the SP where they go method not found error. They somehow reached PCS1 path code. Removal simply clean up things for good leaving nothing to chance.

rvagg: Oh, so we should also push out a lotus-miner release at some point after this as well? We've separated the node and miner release process so we need to explicitly opt to do this.

LexLuthr: Yes. We should ship a new miner release for NV23.

magik6k: This path was active when basefee is below the batch balancer level.

magik6k (approved):
Don't see anything wrong here, looks good!

rvagg:
Continuing the discussion above (https://github.com/filecoin-project/lotus/pull/12251#discussion_r1687422143) about how we invoke this removed code path post nv22. (Permalinks below point to v1.28.0)

In https://github.com/filecoin-project/lotus/blob/28a578f7fef7873c85e2a1789e35772d828c5499/storage/pipeline/commit_batch.go#L240-L245 we have the basefee check, where if the basefee is below the batch balancer level we `individual = true`.

Since nv22, it _always_ goes into this block, regardless of what `individual` is, and any sectors that return from `processBatchV2()` having been processed are removed from the `todo` list: https://github.com/filecoin-project/lotus/blob/28a578f7fef7873c85e2a1789e35772d828c5499/storage/pipeline/commit_batch.go#L246-L278

When we get to `processIndividually()`, we end up without anything in `todo` so don't process anything: https://github.com/filecoin-project/lotus/blob/28a578f7fef7873c85e2a1789e35772d828c5499/storage/pipeline/commit_batch.go#L287-L290

But the catch is in `processBatchV2()`, where we have a check for deal IDs and we don't process those sectors: https://github.com/filecoin-project/lotus/blob/28a578f7fef7873c85e2a1789e35772d828c5499/storage/pipeline/commit_batch.go#L359-L363

So the effect of this is that CC sectors can go in via PCS3, but deal sectors go in via PCS.

There's additional changes in here that relate to deal IDs as they pertain to inspecting pieces. In particular this code inspects pieces, looking for non-filler pieces and always opts out of DDO (`false`). It's only `true` when it's used for ReplicaUpdate. Either way we get a list of deal IDs back for non-CC sectors, which forces us into the PCS onboarding state as long as the basefee condition holds: https://github.com/filecoin-project/lotus/blob/28a578f7fef7873c85e2a1789e35772d828c5499/storage/pipeline/states_sealing.go#L854-L875

jennijuju:
Thanks! would've been great if its in PR description in the first place

TippyFlitsUK:
Started sealing a non-DDO deal on mainnet `1.27.2` to `WaitSeed` stage. Upgraded to `1.28.1` and the sector fails to finalize as expected.

[12251 Test.txt](https://github.com/user-attachments/files/16384538/12251.Test.txt)

PR #12253, opened 2024-07-17
fix(store): correctly break weight ties based on smaller ticket

arajasek:
## Related Issues

@rjan90 and @jennijuju reported seeing this log a lot, which prompted an investigation of the logic. This code does not implement [FIP-0023](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0023.md) correctly. Specifically, if we have the following mintickets:

```
ts1: 1, 2
ts2: 0,3
```

This code will incorrectly return ts1 as the "smaller" tipset, even though the tie should have been broken in favour of ts2 based on `1 > 0`.

## Proposed Changes

Fix this bug, we should return in favour of ts2 if its ticket is smaller. Add test.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/store/store.go
+++ b/chain/store/store.go
@@ -1322,12 +1322,15 @@ func breakWeightTie(ts1, ts2 *types.TipSet) bool {
 	// blocks are already sorted by ticket
 	for i := 0; i < s; i++ {
 		if ts1.Blocks()[i].Ticket.Less(ts2.Blocks()[i].Ticket) {
-			log.Infof("weight tie broken in favour of %s", ts1.Key())
+			log.Infof("weight tie broken in favour of %s against %s", ts1.Key(), ts2.Key())
 			return true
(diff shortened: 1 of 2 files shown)


rvagg (approved):
matches FIP-0023 from my read of it

needs a fiximports though on the new test file

Stebalien:
@ribasushi pointed out that this is likely consensus breaking and I think he's right. Won't this cause us to order blocks within a tipset differently?

Stebalien:
(more importantly, the current behavior is forky and the new behavior is less forky)

arajasek:
> Ah! This is between two tipsets. It's a "soft" consensus breaking change that'll likely resolve within a few tipsets, from what I can tell.

Yeah, the other thing is that this is _already_ the fallback option. Weight is what's actually supposed to decide between tipsets, this is only for when that ties.

Issue #8392, opened 2022-03-29
Make `make gen` faster


magik6k:
Mostly make it actually work in parallel

rvagg:
```
$ make gen
...
real    1m40.957s
```

Linux, nvme, plenty of power and resources.

If I go and `time` prefix every step in the makefile then I can break it down into the various stages:

```
make gen

go run ./gen/inline-gen . gen/inlinegen-data.json
# 0m0.176s
go run ./chain/actors/agen
# 0m0.296s
go fmt ./...
# 0m0.594s
go run ./gen/api
# 0m0.255s
goimports -w api
# 0m0.132s
goimports -w api
# 0m0.127s
go run ./gen/main.go
# 0m5.147s
go generate -x ./...
# 1m1.699s
goimports -w api/
# 0m0.122s
go run ./node/config/cfgdocgen > ./node/config/doc_gen.go
# 0m0.149s
go run ./scripts/fiximports
# 0m5.130s
go build -ldflags="-X=github.com/filecoin-project/lotus/build.CurrentCommit=+git.3daff2ed8.dirty" -o docgen-md ./api/docgen/cmd
# 0m0.307s
./docgen-md "api/api_full.go" "FullNode" "api" "./api" > documentation/en/api-v1-unstable-methods.md
# 0m0.060s
./docgen-md "api/v0api/full.go" "FullNode" "v0api" "./api/v0api" > documentation/en/api-v0-methods.md
# 0m0.048s
./docgen-md "api/api_storage.go" "StorageMiner" "api" "./api" > documentation/en/api-v0-methods-miner.md
# 0m0.052s
./docgen-md "api/api_worker.go" "Worker" "api" "./api" > documentation/en/api-v0-methods-worker.md
# 0m0.055s
go build -ldflags="-X=github.com/filecoin-project/lotus/build.CurrentCommit=+git.3daff2ed8.dirty" -o docgen-openrpc ./api/docgen-openrpc/cmd
# 0m0.304s
./docgen-openrpc "api/api_full.go" "FullNode" "api" "./api" > build/openrpc/full.json
# 0m1.958s
./docgen-openrpc "api/api_storage.go" "StorageMiner" "api" "./api" > build/openrpc/miner.json
# 0m0.792s
./docgen-openrpc "api/api_worker.go" "Worker" "api" "./api" > build/openrpc/worker.json
# 0m0.365s
[...]

PR #12258, opened 2024-07-18
chore: lint: enable method, function & type godoc linting

rvagg:
Making another small dent in the 5 line diff PRs (like https://github.com/filecoin-project/lotus/pull/12257)


rjan90 (approved):

rjan90:
It looks like running make gen / make docsgen-cli is necessary for these tests to pass.

When running `make gen`, could you record the time it takes, and note it down in this [issue](https://github.com/filecoin-project/lotus/issues/8392)? I want to determine if the slowness is specific to my computer or if the process is generally time-consuming. If it is, we should prioritize optimizing it to save time across the board.

rvagg:
sorry, had a gen done locally but forgot to push, it was more complicated than just a gen because there's a template impacted

yes, it takes forever, which is why I forgot about my local fixes because I'd moved on to something else

aarshkshah1992 (approved):

Issue #12255, opened 2024-07-18
Events db query speed


rvagg:
While looking at https://github.com/filecoin-project/lotus/issues/12254 I took the opportunity to look into timing. On my mainnet node with quite a bit of events history (42G events.db):

```
$ time curl http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock": "0x3E8872", "toBlock": "0x3E8872"}],"id":1}'
...
real    0m3.377s
```

When I run this query in a loop and then grab 120s of samples to build a flamegraph, it's all in the sqlite native engine:

<img width="1207" alt="Screenshot 2024-07-18 at 10 37 56 AM" src="https://github.com/user-attachments/assets/95c59b21-4cc3-473f-9914-df392e3a7cdc">

The query that's built, prepared and then executed (with substitutions) is:

```
SELECT
      event.id,
      event.height,
      event.tipset_key,
      event.tipset_key_cid,
      event.emitter_addr,
      event.event_index,
      event.message_cid,
      event.message_index,
      event.reverted,
      event_entry.flags,
      event_entry.key,
      event_entry.codec,
      event_entry.value
    FROM event JOIN event_entry ON event.id=event_entry.event_id
    WHERE event.height>=? AND event.height<=? AND event.reverted=?
    ORDER BY event.height DESC, event_entry._rowid_ ASC
```

First thing to note is the stupid `height>=? and height<=?` for the case where fromBlock=toBlock, we can easily optimise that but a quick test of doing that doesn't show much difference in execution speed.

Considering the indexes needed for this:

* `event.id` is a PRIMARY KEY
* `event_entry.event_id` has its own index
* `event.height` has its own index
* `event.reverted` has its own index
* `event_entry._rowid_` I believe is implicitly indexed and by the time it needs to sort this should be minimal anyway

**BUT** if I make a composite index for the two WHERE columns: `CREATE INDEX IF NOT EXISTS tmp_event_height_reverted ON event (height, reverted);` and then re-run the query:

```
$ time curl http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock": "0x3E8872", "toBlock": "0x3E8872"}],"id":1}'
real    0m0.008s
```

That's a bit dramatic!

[...]

PR #12261, opened 2024-07-18
fix: events: address sqlite index selection performance regressions

rvagg:
_Edit:_ If you want to take advantage of the performance gains from this change without upgrading to a version of Lotus that includes this fix, you can do it manually for yourself in a safe way that will still work when you eventually do upgrade Lotus to include this fix.

With `sqlite3` installed, substituting the correct path for `/path/to/lotus/repo`, run:

```
sqlite3 /path/to/lotus/repo/sqlite/events.db '
  DROP INDEX IF EXISTS event_emitter_addr;
  DROP INDEX IF EXISTS event_reverted;
  DROP INDEX IF EXISTS event_entry_indexed_key;
  DROP INDEX IF EXISTS event_entry_codec_value;'
```

---

Fixes: https://github.com/filecoin-project/lotus/issues/12255

SQLite was found to be avoiding the intended initial indexes for some types of complex queries and opting for minor indexes which don't narrow down the search space enough. Specifically we want queries to first either narrow by height or tipset_key_cid and then apply other criteria. Having alternative indexes when a query such as `height>=X AND height<=Y` are encountered cause SQLite to avoid the height index entirely. By removing additional indexes that could be used during the main query path (prefillFilter), we force SQLite to use the intended indexes and narrow the results without the use of indexes.

---

Some examples and numbers. These approximate the queries generated by `prefillFilter()` for various inputs and are run against my 40GiB events db containing 380,025 epochs. I've got 2 copies of it, one with the schema as in this PR and one with the schema as in master and I'm running and timing these queries directly on the commandline.

### Basic select using `height=`

_(note the current version doesn't even use this, it's always >= & <=)_

```sql
SELECT event.height, hex(event.tipset_key_cid)
FROM event
JOIN event_entry ON event.id=event_entry.event_id
WHERE event.height=4098162 AND event.reverted=false
ORDER BY event.height DESC, event_entry._rowid_ ASC
```

Before: 3.05s, after: 0.0s - the reason for the slowness is that it chooses the `reverted` index instead of `height`, for reasons I can't fathom. This could be fixed by having a compound index on the two, but that fails for other cases.

### Basic select using height>= and height <= for single height

Results in the same output as above.

```sql
SELECT event.height, hex(event.tipset_key_cid)
FROM event
JOIN event_entry ON event.id=event_entry.event_id
WHERE event.height>=4098162 AND event.height<=4098162 AND event.reverted=false
ORDER BY event.height DESC, event_entry._rowid_ ASC
```

Before: 3.10, after: 0.0s

### Basic select using tipset_key_cid

Same results as above but using the tipset instead of height.

```sql
SELECT event.height, hex(event.tipset_key_cid)
FROM event
JOIN event_entry ON event.id=event_entry.event_id
WHERE event.tipset_key_cid=X'0171A0E40220DA1BA218ADC163101801DD8894A7378ABE5FDC4C62DC7056DCC96EC5F039EF64' AND event.reverted=false
ORDER BY event.height DESC, event_entry._rowid_ ASC
```

Before: 3.42, after: 0.0s - it's again using `reverted` for some reason, and then matching on tsk cid which is slower because it's a large bytes match.

### Select at height and match emitter_addr

Again, this isn't quite what master does because we don't have `height=X` there, but close:

```sql
SELECT event.height, hex(event.tipset_key_cid)
FROM event
JOIN event_entry ON event.id=event_entry.event_id
WHERE event.height=4098162 AND event.reverted=false AND event.emitter_addr=X'040A60E1773636CF5E4A227D9AC24F20FECA034EE25A'
ORDER BY event.height DESC, event_entry._rowid_ ASC
```

Before: 3.15s, after: 0.0s - `reverted` again, but if I remove the `reverted` clause here it goes back to the `height` index rather than the `emitter_addr` index.

### Select at height >= and <= and match emitter_addr

Same as above but with `height>=X AND height <=X`

Before: 3.19s, after: 0.0s

### Select at tipset and match emitter_addr

Returns the same results as the above 2 but uses tsk

```sql
SELECT event.height, hex(event.tipset_key_cid)
FROM event
JOIN event_entry ON event.id=event_entry.event_id
WHERE event.tipset_key_cid=X'0171A0E40220DA1BA218ADC163101801DD8894A7378ABE5FDC4C62DC7056DCC96EC5F039EF64' AND event.reverted=false AND event.emitter_addr=X'040A60E1773636CF5E4A227D9AC24F20FECA034EE25A'
ORDER BY event.height DESC, event_entry._rowid_ ASC
```

Before: 3.35s, after: 0.0s

### Select event_entry properties

Now we get interesting, this is the kind of thing you might get through `GetActorEventsRaw` or `eth_getLogs` searching for a particular log output. `prefillFilter()` builds these queries, they're a bit nested and complex but trigger a bunch of different indices:

```sql
SELECT event.height, hex(event.tipset_key_cid),hex(event.emitter_addr),event_entry.indexed,event_entry.codec,event_entry.key,hex(event_entry.value)
FROM event
JOIN event_entry ON event.id=event_entry.event_id, event_entry ee2 ON event.id=ee2.event_id
WHERE event.height>=4095282 AND event.height<=4098162 AND event.reverted=false AND event.emitter_addr=X'040A60E1773636CF5E4A227D9AC24F20FECA034EE25A' AND ee2.indexed=1 AND ee2.key='t1' AND ((ee2.value=X'DDF252AD1BE2C89B69C2B068FC378DAA952BA7F163C4A11628F55A4DF523B3EF' AND ee2.codec=85))
ORDER BY event.height DESC, event_entry._rowid_ ASC
```

Before: 0.73s, after: 0.1s.

`EXPLAIN QUERY PLAN` on this is interesting, it's still avoiding the index we want to hit (`height`) but it's finding one that'll narrow it down enough so the resultset isn't too big to filter further down:

Before:

```
QUERY PLAN
|--SEARCH ee2 USING INDEX event_entry_codec_value (codec=? AND value=?)
|--SEARCH event USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH event_entry USING INDEX event_entry_event_id (event_id=?)
`--USE TEMP B-TREE FOR ORDER BY
```

After:

```
QUERY PLAN
|--SEARCH event USING INDEX event_height (height>? AND height<?)
|--SEARCH ee2 USING INDEX event_entry_event_id (event_id=?)
|--SEARCH event_entry USING INDEX event_entry_event_id (event_id=?)
`--USE TEMP B-TREE FOR RIGHT PART OF ORDER BY
```

---

I'm using `reverted` in here because `eth_*` will use that, but it's not always the case if you use `GetActorEventsRaw` so the speed may vary if it hits alternative indexes.

After all of this I still don't know what criteria SQLite is using to select its index ordering. Perhaps it's to do with the efficiency of the indexes? Bool index is better than int index is better than bytes index? I don't really understand the use of `event_entry_codec_value` in the last one though, but maybe that's a special case because of nesting?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ## ☢️ Upgrade Warnings ☢️
 
-- This Lotus release includes some correctness improvements to the events subsystem, impacting RPC APIs including `GetActorEventsRaw`, `SubscribeActorEventsRaw`, `eth_getLogs` and the `eth` filter APIs. Part of these improvements involve an events database migration that may take some time to complete on nodes with extensive event databases. See [filecoin-project/lotus#12080](https://github.com/filecoin-project/lotus/pull/12080) for details.
+- This Lotus release includes some performance improvements to the events subsystem, impacting RPC APIs including `GetActorEventsRaw`, `SubscribeActorEventsRaw`, `eth_getLogs` and the `eth` filter APIs. Part of these improvements involve an events database migration that may take some time to complete on nodes with extensive event databases. See [filecoin-project/lotus#12261](https://github.com/filecoin-project/lotus/pull/12261) for details.
 
 - Breaking change in public APIs `storage/pipeline.NewPreCommitBatcher`, `sealing.NewCommitBatcher` and `storage/pipeline.New`. They now have an additional error return to deal with errors arising from fetching the sealing config.
 
--- a/chain/events/filter/index.go
+++ b/chain/events/filter/index.go
@@ -71,27 +71,6 @@ const (
 	    UNIQUE(height, tipset_key_cid)
 	)`
 
-	insertEvent          = `INSERT OR IGNORE INTO event(height, tipset_key, tipset_key_cid, emitter_addr, event_index, message_cid, message_index, reverted) VALUES(?, ?, ?, ?, ?, ?, ?, ?)`
-	insertEntry          = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
-	revertEventsInTipset = `UPDATE event SET reverted=true WHERE height=? AND tipset_key=?`
-	restoreEvent         = `UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
-	revertEventSeen      = `UPDATE events_seen SET reverted=true WHERE height=? AND tipset_key_cid=?`
-	restoreEventSeen     = `UPDATE events_seen SET reverted=false WHERE height=? AND tipset_key_cid=?`
-	upsertEventsSeen     = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, false) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
-
-	eventExists = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
-	// QUERY PLAN
-	// `--SEARCH event USING INDEX event_height (height=?)
-	isTipsetProcessed = `SELECT COUNT(*) > 0 FROM events_seen WHERE tipset_key_cid=?`
-	// QUERY PLAN
-	// `--SEARCH events_seen USING COVERING INDEX events_seen_tipset_key_cid (tipset_key_cid=?)
-	getMaxHeightInIndex = `SELECT MAX(height) FROM events_seen`
-	// QUERY PLAN
-	// `--SEARCH events_seen USING COVERING INDEX events_seen_height
-	isHeightProcessed = `SELECT COUNT(*) > 0 FROM events_seen WHERE height=?`
-	// QUERY PLAN
-	// `--SEARCH events_seen USING COVERING INDEX events_seen_height (height=?)
-
 	// When modifying indexes in this file, it is critical to test the query plan (EXPLAIN QUERY PLAN)
 	// of all the variations of queries built by prefillFilter to ensure that the query first hits
 	// an index that narrows down results to an epoch or a reasonable range of epochs. Specifically,
@@ -130,21 +109,45 @@ const (
 	createIndexEventsSeenTipsetKeyCid = `CREATE INDEX IF NOT EXISTS events_seen_tipset_key_cid ON events_seen (tipset_key_cid);`
 )
 
+// preparedStatementMapping returns a map of fields of the preparedStatements struct to the SQL
+// query that should be prepared for that field. This is used to prepare all the statements in
+// the preparedStatements struct but it's also used by testing code to access the raw query strings
+// to ensure that the correct indexes are being used by SELECT queries.
+func preparedStatementMapping(ps *preparedStatements) map[**sql.Stmt]string {
+	return map[**sql.Stmt]string{
+		&ps.insertEvent:          `INSERT OR IGNORE INTO event(height, tipset_key, tipset_key_cid, emitter_addr, event_index, message_cid, message_index, reverted) VALUES(?, ?, ?, ?, ?, ?, ?, ?)`,
+		&ps.insertEntry:          `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`,
+		&ps.revertEventsInTipset: `UPDATE event SET reverted=true WHERE height=? AND tipset_key=?`,
+		&ps.restoreEvent:         `UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`,
+		&ps.revertEventSeen:      `UPDATE events_seen SET reverted=true WHERE height=? AND tipset_key_cid=?`,
+		&ps.restoreEventSeen:     `UPDATE events_seen SET reverted=false WHERE height=? AND tipset_key_cid=?`,
+		&ps.upsertEventsSeen:     `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, false) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`,
+		&ps.eventExists:          `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`, // QUERY PLAN: SEARCH event USING INDEX event_height (height=?)
+		&ps.isTipsetProcessed:    `SELECT COUNT(*) > 0 FROM events_seen WHERE tipset_key_cid=?`,                                                                                               // QUERY PLAN: SEARCH events_seen USING COVERING INDEX events_seen_tipset_key_cid (tipset_key_cid=?)
+		&ps.getMaxHeightInIndex:  `SELECT MAX(height) FROM events_seen`,                                                                                                                       // QUERY PLAN: SEARCH events_seen USING COVERING INDEX events_seen_height
+		&ps.isHeightProcessed:    `SELECT COUNT(*) > 0 FROM events_seen WHERE height=?`,                                                                                                       // QUERY PLAN: SEARCH events_seen USING COVERING INDEX events_seen_height (height=?)
+
+	}
+}
+
+type preparedStatements struct {
+	insertEvent          *sql.Stmt
+	insertEntry          *sql.Stmt
+	revertEventsInTipset *sql.Stmt
+	restoreEvent         *sql.Stmt
+	upsertEventsSeen     *sql.Stmt
+	revertEventSeen      *sql.Stmt
+	restoreEventSeen     *sql.Stmt
+	eventExists          *sql.Stmt
+	isTipsetProcessed    *sql.Stmt
+	getMaxHeightInIndex  *sql.Stmt
+	isHeightProcessed    *sql.Stmt
+}
+
 type EventIndex struct {
 	db *sql.DB
 
-	stmtEventExists          *sql.Stmt
-	stmtInsertEvent          *sql.Stmt
-	stmtInsertEntry          *sql.Stmt
-	stmtRevertEventsInTipset *sql.Stmt
-	stmtRestoreEvent         *sql.Stmt
-	stmtUpsertEventsSeen     *sql.Stmt
-	stmtRevertEventSeen      *sql.Stmt
-	stmtRestoreEventSeen     *sql.Stmt
-
-	stmtIsTipsetProcessed   *sql.Stmt
-	stmtGetMaxHeightInIndex *sql.Stmt
-	stmtIsHeightProcessed   *sql.Stmt
+	stmt *preparedStatements
 
 	mu           sync.Mutex
 	subIdCounter uint64
@@ -158,65 +161,6 @@ type updateSub struct {
 }
 
 type EventIndexUpdated struct{}
-
-func (ei *EventIndex) initStatements() (err error) {
-	ei.stmtEventExists, err = ei.db.Prepare(eventExists)
-	if err != nil {
-		return xerrors.Errorf("prepare stmtEventExists: %w", err)
-	}
-
-	ei.stmtInsertEvent, err = ei.db.Prepare(insertEvent)
-	if err != nil {
-		return xerrors.Errorf("prepare stmtInsertEvent: %w", err)
-	}
-
-	ei.stmtInsertEntry, err = ei.db.Prepare(insertEntry)
-	if err != nil {
-		return xerrors.Errorf("prepare stmtInsertEntry: %w", err)
-	}
-
-	ei.stmtRevertEventsInTipset, err = ei.db.Prepare(revertEventsInTipset)
-	if err != nil {
-		return xerrors.Errorf("prepare stmtRevertEventsInTipset: %w", err)
-	}
-
-	ei.stmtRestoreEvent, err = ei.db.Prepare(restoreEvent)
-	if err != nil {
-		return xerrors.Errorf("prepare stmtRestoreEvent: %w", err)
-	}
-
-	ei.stmtUpsertEventsSeen, err = ei.db.Prepare(upsertEventsSeen)
-	if err != nil {
-		return xerrors.Errorf("prepare stmtUpsertEventsSeen: %w", err)
-	}
-
-	ei.stmtRevertEventSeen, err = ei.db.Prepare(revertEventSeen)
-	if err != nil {
-		return xerrors.Errorf("prepare stmtRevertEventSeen: %w", err)
-	}
-
-	ei.stmtRestoreEventSeen, err = ei.db.Prepare(restoreEventSeen)
-	if err != nil {
-		return xerrors.Errorf("prepare stmtRestoreEventSeen: %w", err)
-	}
-
-	ei.stmtIsTipsetProcessed, err = ei.db.Prepare(isTipsetProcessed)
-	if err != nil {
-		return xerrors.Errorf("prepare isTipsetProcessed: %w", err)
-	}
-
-	ei.stmtGetMaxHeightInIndex, err = ei.db.Prepare(getMaxHeightInIndex)
-	if err != nil {
-		return xerrors.Errorf("prepare getMaxHeightInIndex: %w", err)
-	}
-
-	ei.stmtIsHeightProcessed, err = ei.db.Prepare(isHeightProcessed)
-	if err != nil {
-		return xerrors.Errorf("prepare isHeightProcessed: %w", err)
-	}
-
-	return nil
-}
 
 func NewEventIndex(ctx context.Context, path string, chainStore *store.ChainStore) (*EventIndex, error) {
 	db, _, err := sqlite.Open(path)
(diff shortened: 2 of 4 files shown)


chain/events/filter/index.go
```diff
@@ -67,34 +63,67 @@ var (
...
 	restoreEvent         = `UPDATE event SET reverted=false WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
 	revertEventSeen      = `UPDATE events_seen SET reverted=true WHERE height=? AND tipset_key_cid=?`
 	restoreEventSeen     = `UPDATE events_seen SET reverted=false WHERE height=? AND tipset_key_cid=?`
 	upsertEventsSeen     = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, false) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
-	isTipsetProcessed    = `SELECT COUNT(*) > 0 FROM events_seen WHERE tipset_key_cid=?`
-	getMaxHeightInIndex  = `SELECT MAX(height) FROM events_seen`
-	isHeightProcessed    = `SELECT COUNT(*) > 0 FROM events_seen WHERE height=?`
 
-	createIndexEventEmitterAddr  = `CREATE INDEX IF NOT EXISTS event_emitter_addr ON event (emitter_addr)`
+	eventExists = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
```
rvagg, line 82: I'm just grouping the queries together here, not changing them, so I can show the query plan for them and document the need to care about index selection

chain/events/filter/index.go
```diff
@@ -510,15 +545,17 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
...
 				joinAlias := fmt.Sprintf("ee%d", join)
-				joins = append(joins, fmt.Sprintf("event_entry %s on event.id=%[1]s.event_id", joinAlias))
+				joins = append(joins, fmt.Sprintf("event_entry %s ON event.id=%[1]s.event_id", joinAlias))
 				clauses = append(clauses, fmt.Sprintf("%s.indexed=1 AND %[1]s.key=?", joinAlias))
 				values = append(values, key)
-				subclauses := make([]string, 0, len(vals))
-				for _, val := range vals {
-					subclauses = append(subclauses, fmt.Sprintf("(%s.value=? AND %[1]s.codec=?)", joinAlias))
-					values = append(values, val.Value, val.Codec)
+				if len(vals) > 1 {
```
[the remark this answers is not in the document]
rvagg: oh .. nevermind, we never hit this because `len(vals)` is checked above!

chain/events/filter/index.go
```diff
@@ -510,7 +545,7 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 			if len(vals) > 0 {
```
rvagg, line 545: @aarshkshah1992 in `{Get,Subscribe}ActorEventsRaw` we advertise the `"fields"` param that gets passed in here as `keysWithCodec` like this:

```
	// Matches events with the specified key/values, or all events if empty.
	// If the value is an empty slice, the filter will match on the key only, accepting any value.
```

But I think this is incorrect—"empty slice" is `len(vals)==0` which means we never make a join for the `eeX.key=?` case. Do you agree? If so we should fix this, either by changing docs or moving the `len(vals)>0` down to the `subclauses` section in this block. It can be done in a separate PR as it needs a test.

chain/events/filter/index.go
```diff
@@ -67,34 +63,67 @@ var (
...
+	isHeightProcessed = `SELECT COUNT(*) > 0 FROM events_seen WHERE height=?`
+	// QUERY PLAN
+	// `--SEARCH events_seen USING COVERING INDEX events_seen_height (height=?)
+
+	// When modifying indexes in this file, it is critical to test the query plan (EXPLAIN QUERY PLAN)
+	// of all the variations of queries built by prefillFilter to ensure that the query first hits
+	// an index that narrows down results to an epoch or a reasonable range of epochs. Specifically,
+	// event_tipset_key_cid or event_height should be the first index. Then further narrowing can take
+	// place within the small subset of results.
+	// Unfortunately SQLite has some quicks in index selection that mean that certain query types will
```
aarshkshah1992, line 100: quirks*

chain/events/filter/index.go
```diff
@@ -67,34 +63,67 @@ var (
...
+	// JOIN
+	// 	event_entry ON event.id=event_entry.event_id,
+	// 	event_entry ee2 ON event.id=ee2.event_id
+	// WHERE event.height>=? AND event.height<=? AND event.reverted=? AND event.emitter_addr=? AND ee2.indexed=1 AND ee2.key=?
+	// ORDER BY event.height DESC, event_entry._rowid_ ASC
+	//
+	// ->
+	//
+	// QUERY PLAN
+	// |--SEARCH event USING INDEX event_height (height>? AND height<?)
```
aarshkshah1992, line 118: I think the point was to show that the `event_height` Index does not get used if height ranges are queries. Is this the correct query plan ?

chain/events/filter/index.go
```diff
@@ -55,8 +53,6 @@ var ddls = []string{
 
 	createTableEventsSeen,
 
-	createIndexEventEntryIndexedKey,
```
aarshkshah1992, line 58: Why do we need to remove the `createIndexEventEntryIndexedKey` AND `createIndexEventEntryCodecValue` indices > Where do they get in the way ?

chain/events/filter/index.go
```diff
@@ -480,13 +510,20 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
...
+		if f.minHeight >= 0 && f.minHeight == f.maxHeight {
+			clauses = append(clauses, "event.height=?")
 			values = append(values, f.minHeight)
-		}
-		if f.maxHeight >= 0 {
-			clauses = append(clauses, "event.height<=?")
-			values = append(values, f.maxHeight)
+		} else {
+			if f.maxHeight >= 0 && f.minHeight >= 0 {
+				clauses = append(clauses, "event.height BETWEEN ? AND ?")
```
aarshkshah1992, line 518: Just out of curiosity: Is there a difference between using `>= and <=` AND `BETWEEN` ?

chain/events/filter/index.go
```diff
@@ -496,12 +533,10 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 	}
 
 	if len(f.addresses) > 0 {
-		subclauses := make([]string, 0, len(f.addresses))
 		for _, addr := range f.addresses {
-			subclauses = append(subclauses, "emitter_addr=?")
 			values = append(values, addr.Bytes())
 		}
-		clauses = append(clauses, "("+strings.Join(subclauses, " OR ")+")")
+		clauses = append(clauses, "event.emitter_addr IN ("+strings.Repeat("?,", len(f.addresses)-1)+"?)")
```
aarshkshah1992, line 539: Nice fix for an off by one 👍

chain/events/filter/index.go
```diff
@@ -510,7 +545,7 @@ func (ei *EventIndex) prefillFilter(ctx context.Context, f *eventFilter, exclude
 			if len(vals) > 0 {
 				join++
 				joinAlias := fmt.Sprintf("ee%d", join)
-				joins = append(joins, fmt.Sprintf("event_entry %s on event.id=%[1]s.event_id", joinAlias))
+				joins = append(joins, fmt.Sprintf("event_entry %s ON event.id=%[1]s.event_id", joinAlias))
```
aarshkshah1992, line 548: This is just a cosmetic change, right ?

chain/events/filter/index.go
aarshkshah1992: This is just a cosmetic change, right ? Or did you see a diff with using `IN []` instead of `OR`

chain/events/filter/index_migrations.go
```diff
@@ -135,19 +135,11 @@ func migrationVersion2(db *sql.DB, chainStore *store.ChainStore) sqlite.Migratio
 	}
 }
 
-// migrationVersion3 migrates the schema from version 2 to version 3 by creating two indices:
-// 1) an index on the event.emitter_addr column, and 2) an index on the event_entry.key column.
+// migrationVersion3 migrates the schema from version 2 to version 3. It originally added two
+// indices: 1) an index on the event.emitter_addr column, and 2) an index on the event_entry.key
+// column. However, as of version 7, these indices have been removed as they were found to be a
```
aarshkshah1992, line 140: As per below, the `event_entry.key` Index was removed in migration4. So we should preserve that comment to document history. Only the `event.emitter_addr` Index is being removed in migration 7.

chain/events/filter/index_migrations.go
```diff
@@ -161,10 +153,13 @@ func migrationVersion3(ctx context.Context, tx *sql.Tx) error {
 // And then creating the following indices:
 //  1. an index on the event.tipset_key_cid column
 //  2. an index on the event.height column
-//  3. an index on the event.reverted column
-//  4. an index on the event_entry.indexed and event_entry.key columns
-//  5. an index on the event_entry.codec and event_entry.value columns
+//  3. an index on the event.reverted column (removed in version 7)
+//  4. an index on the event_entry.indexed and event_entry.key columns (removed in version 7)
```
aarshkshah1992, line 157: Better to document that we're removing composite indices here and below and not indices on the individual columns.

chain/events/filter/index_migrations.go
```diff
@@ -236,3 +228,32 @@ func migrationVersion6(ctx context.Context, tx *sql.Tx) error {
...
+//  2. the index on the event.reverted column
+//  3. the index on the event_entry.indexed and event_entry.key columns
+//  4. the index on the event_entry.codec and event_entry.value columns
+//
+// These indices were found to be a performance hindrance as they prevent SQLite from using the
+// intended initial indexes on height or tipset_key_cid in many query variations. Without additional
+// indices to fall-back on, SQLite is forced to narrow down each query via height or tipset_key_cid
+// which is the desired behavior.
+func migrationVersion7(ctx context.Context, tx *sql.Tx) error {
+	for _, create := range []struct {
```
aarshkshah1992, line 244: not: loop variable should be "drop" instead of "create".

aarshkshah1992 (commented):
lgtm

Let's test this on a mainnet node and ideally ask Glif/Mikers to deploy it and get the RPC timings out to see the impact.

chain/events/filter/index.go
aarshkshah1992: Yeah I dug through this and this needs to be fixed. Will raise a PR tomorrow for this. Let's keep this comment unresolved till then.

rvagg: Made it an issue so it doesn't get lost, thanks for confirming: https://github.com/filecoin-project/lotus/issues/12264

chain/events/filter/index.go
rvagg: the sqlite index selection story gets worse, consider the case I outlined above with the secondary join and `AND ((ee2.value=X'DDF252AD1BE2C89B69C2B068FC378DAA952BA7F163C4A11628F55A4DF523B3EF' AND ee2.codec=85))`, if I run a query plan with and without this index:

```
QUERY PLAN
|--SEARCH ee2 USING INDEX event_entry_codec_value (codec=? AND value=?)
|--SEARCH event USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH event_entry USING INDEX event_entry_event_id (event_id=?)
`--USE TEMP B-TREE FOR ORDER BY
```

without:

```
QUERY PLAN
|--SEARCH event USING INDEX event_height (height>? AND height<?)
|--SEARCH ee2 USING INDEX event_entry_event_id (event_id=?)
|--SEARCH event_entry USING INDEX event_entry_event_id (event_id=?)
`--USE TEMP B-TREE FOR RIGHT PART OF ORDER BY
```

it reaches in to event_entry indexes (!!) to find the first index to narrow down the selection; it ends up being not-terrible because the `value` part of the index is relatively unique. But imagine a really popular smart contract, running thousands or millions of times over the life of this database, that's significantly worse than first narrowing down by epoch.

chain/events/filter/index.go
rvagg: I don't think so, I couldn't observe a measurable speed difference when I was trying this, it ended up _feeling_ better to specify it this way to be more explicit (you're welcome to object to an aesthetic change if you disagree).

chain/events/filter/index.go
rvagg: no observed difference, but I haven't tried with a long list; I would hope the db engine would be able to see the same thing here but who knows without exhaustive testing!

chain/events/filter/index.go
rvagg: yep

chain/events/filter/index.go
```diff
@@ -67,34 +63,67 @@ var (
...
 	restoreEventSeen     = `UPDATE events_seen SET reverted=false WHERE height=? AND tipset_key_cid=?`
 	upsertEventsSeen     = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, false) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
-	isTipsetProcessed    = `SELECT COUNT(*) > 0 FROM events_seen WHERE tipset_key_cid=?`
-	getMaxHeightInIndex  = `SELECT MAX(height) FROM events_seen`
-	isHeightProcessed    = `SELECT COUNT(*) > 0 FROM events_seen WHERE height=?`
 
-	createIndexEventEmitterAddr  = `CREATE INDEX IF NOT EXISTS event_emitter_addr ON event (emitter_addr)`
+	eventExists = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
+	// QUERY PLAN
+	// `--SEARCH event USING INDEX event_height (height=?)
```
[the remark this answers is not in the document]
rvagg: it does seem a bit weird, I suppose it makes the call that it's narrowed it down enough with the first index that it would be too costly to do it further with the second; which may be reasonable given that the tsk cid index is on large byte strings.

if I change the order of the conditions then it doesn't make a difference, it seems to prefer to use `height`

my expectation coming into this is that order matters, but it appears to be making decisions on another heuristic, my guess is that it might be something about likely efficiency based on datatype. I'm tempted to go fishing for answers.

chain/events/filter/index.go
rvagg: the query plan is correct, it's updated for the current schema; I've amended the text to say:

```
	// e.g. we want to see that `event_height` is the first index used in this query:
```

it turns out that height ranges are just find if it doesn't have a better alternative. it was the composite index on `(height, reverted)` that was causing problems with ranges - it would use this for `height=? and reverted=false` but then bork on it for `height>=? and height <=? and reverted=false`!! but with a non-composite index it seems to be OK with this, we just don't get to index `reverted`.

chain/events/filter/index.go
```diff
@@ -67,34 +63,68 @@ var (
...
+	getMaxHeightInIndex = `SELECT MAX(height) FROM events_seen`
+	// QUERY PLAN
+	// `--SEARCH events_seen USING COVERING INDEX events_seen_height
+	isHeightProcessed = `SELECT COUNT(*) > 0 FROM events_seen WHERE height=?`
+	// QUERY PLAN
+	// `--SEARCH events_seen USING COVERING INDEX events_seen_height (height=?)
+
+	// When modifying indexes in this file, it is critical to test the query plan (EXPLAIN QUERY PLAN)
+	// of all the variations of queries built by prefillFilter to ensure that the query first hits
+	// an index that narrows down results to an epoch or a reasonable range of epochs. Specifically,
```
BigLep, line 97: Curious: is there a way to mechanize this?  Basically, can we iterate over the queries build by prefillFilter and assert that the output of EXPLAIN QUERY PLAN has event_tipset_key_cid or event_height should as the first index?

BigLep (commented):
Wow - awesome work investigating this.  I have been living in a world of assuming the DB engine would "do the right thing".  Good to know that verification is needed.  Well done!

chain/events/filter/index.go
rvagg: I did start investigating this but didn't take it far enough to have a definitive answer; without full testing I believe this is a `sqlite3` client thing rather than something we can invoke programmatically, but I should check further.

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 
 ## ☢️ Upgrade Warnings ☢️
 
-- This Lotus release includes some correctness improvements to the events subsystem, impacting RPC APIs including `GetActorEventsRaw`, `SubscribeActorEventsRaw`, `eth_getLogs` and the `eth` filter APIs. Part of these improvements involve an events database migration that may take some time to complete on nodes with extensive event databases. See [filecoin-project/lotus#12080](https://github.com/filecoin-project/lotus/pull/12080) for details.
+- This Lotus release includes some performance improvements to the events subsystem, impacting RPC APIs including `GetActorEventsRaw`, `SubscribeActorEventsRaw`, `eth_getLogs` and the `eth` filter APIs. Part of these improvements involve an events database migration that may take a some time to complete on nodes with extensive event databases. See [filecoin-project/lotus#12258](https://github.com/filecoin-project/lotus/pull/12258) for details.
```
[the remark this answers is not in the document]
rvagg: heh, no, it's supposed to be _this_ PR, I guessed wrong

chain/events/filter/index.go
aarshkshah1992: No was just curious. This is lgtm.

aarshkshah1992 (approved):

chain/events/filter/index.go
rvagg: done in #12267

Issue #12254, opened 2024-07-17
Inconsistent logIndex between eth_getLogs and eth_getTransactionReceipt


ilyalukyanov:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
N/A
```

### Repro Steps

1. For a recent transaction that emitted logs call, e.g. `{"method":"eth_getTransactionReceipt","params":["0x29ef3cd631c072f2ab3650f0e20a4b678e11c711786004583aa2f071c8cfb74a"],"id":1,"jsonrpc":"2.0"}`
2. For the block of that transaction call, e.g. `{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock": "0x3E8872", "toBlock": "0x3E8872"}],"id":1}`
3. The same log in both responses has different values of `logIndex`

### Describe the Bug

As per repro steps:
- The request `{"method":"eth_getTransactionReceipt","params":["0x29ef3cd631c072f2ab3650f0e20a4b678e11c711786004583aa2f071c8cfb74a"],"id":1,"jsonrpc":"2.0"}` returns logs with `logIndex` ranging from `0x0` to `0xd`.
- The request `{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock": "0x3E8872", "toBlock": "0x3E8872"}],"id":1}` returns logs with `logIndex` ranging from `0x74` to `0x81`
- The same logs should have the same `logIndex` regardless of which method returned them.
[...]

PR #12262, opened 2024-07-18
fix: eth_getReceipt: logIndex should be relative to the tipset not the message 

aarshkshah1992:
Closes https://github.com/filecoin-project/lotus/issues/12254.


node/impl/full/eth.go
```diff
@@ -427,12 +435,47 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+		// For each message, do a StateSearchMsg to get the msgLookup
+		msgLookup, err := a.StateAPI.StateSearchMsg(ctx, types.EmptyTSK, msg.Cid(), limit, true)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to search for message %s: %w", msg.Cid(), err)
+		}
+
+		if msgLookup != nil && msgLookup.Receipt.EventsRoot != nil {
+			// Load the receipt root from the msgLookup and read the events
+			events, err := a.ChainAPI.ChainGetEvents(ctx, *msgLookup.Receipt.EventsRoot)
+			if err != nil {
```
aarshkshah1992, line 469: @rvagg Should we recompute the events here if the Chainstore does NOT have them or return an error ?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - https://github.com/filecoin-project/lotus/pull/12221: Fix a nil reference panic in the ETH Trace API
 - https://github.com/filecoin-project/lotus/pull/12112: Moved consts from build/ to build/buildconstants/ for ligher curio deps.
 - https://github.com/filecoin-project/lotus/pull/12237: Upgrade to go-f3 `v0.0.4`.
+- https://github.com/filecoin-project/lotus/pull/12262: fix incorrect `logIndex` ordering in response of `EthGetTransactionReceipt`
```
rvagg, line 9:
```suggestion
- https://github.com/filecoin-project/lotus/pull/12262: fix incorrect `logIndex` in response of `EthGetTransactionReceipt`
```

node/impl/full/eth.go
```diff
@@ -427,12 +432,61 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+		}
+		if msgLookup == nil {
+			// This is the best we can do. In theory, we could have just not indexed this
+			// transaction, but there's no way to check that here.
+			return nil, fmt.Errorf("message not found: %s", msg.Cid())
+		}
+
+		if msgLookup.Receipt.EventsRoot != nil {
+			// Load the receipt root from the msgLookup and read the events
+			events, err := a.ChainAPI.ChainGetEvents(ctx, *msgLookup.Receipt.EventsRoot)
```
rvagg, line 470: since this block has to do the same work as in `newEthTxReceipt`, how about you break _after_ you've reached the message with the right CID, capture those `events` and pass them to `newEthTxReceipt` so it doesn't have to do the work too

rvagg (approved):
Looks good to me, with one doc suggestion and one minor refactor suggestion.
I left some thoughts in slack about testing this. Essentially: we should be able to spam the message pool with invocations until we get a tipset that contains >X messages with invocations in it and then we should have our case that exercises this. Currently the `client.EVM().InvokeSolidity` call in the testkit just does one message + wait at a time, we need to invoke multiple in parallel and watch the chain for multiple to land at once. I don't _think_ we need an new contract with multiple emits, but I guess if that's not hard it wouldn't hurt to be able to test the indexing even further.

I will try this out today.

node/impl/full/eth.go
```diff
@@ -427,12 +432,61 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
-	receipt, err := newEthTxReceipt(ctx, tx, msgLookup, a.ChainAPI, a.StateAPI)
+	// Load all messages in the tipset
+	msgs, err := a.Chain.MessagesForTipset(ctx, ts)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to load messages for tipset: %w", err)
+	}
+	logIndexStart := uint64(0)
+	// Iterate over all messages till we get to message with cid c
+	for _, msg := range msgs {
+		if msg.Cid() == c {
```
rvagg, line 453: also need to handle the case where we don't find the message we're after

rvagg:
@aarshkshah1992 ok, I couldn't leave this one alone, the main problem here is that it's searching the wrong tipset for the messages, if you check for `m.Cid() == msg` in the loop it doesn't find it, so you end up with the wrong start index.

So in my hacking, I imported the way that the events index consumes the events, by directly loading the amt, which didn't turn out to be the issue; then I looked at the address resolver thing, which I think is still an issue, but I'm not sure we should be handling it here or not, but I have code to do it anyway, then I went back and forth on figuring out which tipset to load the messages and receipts from.

I think the key take-away I have from looking at this is that `StateSearchMsg` shows you the tipset where the message was executed, so that's the receipts tipset and the one you want for the messages that are executed is the one prior, `Parents()`.

Instead of just using the transaction from the original issue, I figured out how to better replicate this on my splitstore node. If you go to the receiver of the transaction it's this: https://filfox.info/en/address/f410fytoouibgl63cxtkko7jr7ibvdqgrsnj4k63noaq which I suppose is a smart contract that does complicated things. Each transaction for this address has similarly complex flows that show up as interesting stuff in getLogs and transactionReceipt.

So taking the most recent transaction as of writing, I can do:

```sh
time curl http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionReceipt","params":["0xafc40a6dcea6d16066658e9b568981f18dc75b6e9538bf6cf4e6d402c128ab05"],"id":1,"jsonrpc":"2.0"}' | jq .result.logs > /tmp/transaction_receipt.logs
```

Then the same for `eth_getLogs` using the epoch, hex encoded:

```sh
time curl http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock": "0x3e972d", "toBlock": "0x3e972d"}],"id":1}' | jq .result > /tmp/get_logs.logs
```

Then diff the two files.

There is a diff, it's:

```diff
<     "address": "0x8460766edc62b525fc1fa4d628fc79229dc73031",
<     "data": "0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000003b62616679626569617361376f6a633576697a71796434797179336e74637563616b6b626372343677726f333366773572657936356c3668666c68610000000000",
<     "topics": [
<       "0xdb4cfdc795eda16407c08dab26edc583f47e2d10e9e56c250f4536124093e878",
<       "0x0000000000000000000000000000000000000000000000000000000000002c1c",
<       "0x00000000000000000000000053bdfdea92f7a60aef82228926d02878018acb4e"
<     ],
<     "removed": false,
<     "logIndex": "0x61",
<     "transactionIndex": "0x38",
<     "transactionHash": "0xa30d554b092288aa31c5b5c304b5d546a43c88e1778a7a10a67da62c5531196b",
<     "blockHash": "0xa0187194ede8ccb219dd4c52dff45d2c964ff1fe6bc04e6ef209684611b35d5a",
<     "blockNumber": "0x3e972d"
<   },
<   {
```

but that's because `eth_getLogs` is getting all logs for that epoch, and this `transactionHash` is a different transaction in the epoch, the rest are identical, including `transactionIndex`, which seems positive. I think I need to test it on fuller epochs. But you can see already that this first log is `logIndex` `0x61`, so there's a lot more logs before that, probably builtin actor events; so I think that means it's correct.

<details>
<summary>Anyway, here's a diff of my code, it's really rough, but shows it working:</summary>

```diff
diff --git a/node/impl/full/eth.go b/node/impl/full/eth.go
index 64c0aef5e..b3659badd 100644
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"sort"
 	"strconv"
@@ -18,6 +19,7 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-address"
+	amt4 "github.com/filecoin-project/go-amt-ipld/v4"
 	"github.com/filecoin-project/go-jsonrpc"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
@@ -39,6 +41,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
 	"github.com/filecoin-project/lotus/node/modules/dtypes"
+	blockadt "github.com/filecoin-project/specs-actors/actors/util/adt"
 )
 
 var ErrUnsupported = errors.New("unsupported method")
@@ -422,7 +425,7 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
 		c = txHash.ToCid()
 	}
 
-	msgLookup, err := a.StateAPI.StateSearchMsg(ctx, types.EmptyTSK, c, limit, true)
+	msgLookup, err := a.StateAPI.StateSearchMsg(ctx, types.EmptyTSK, c, limit, false)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to lookup Eth Txn %s as %s: %w", txHash, c, err)
 	}
@@ -431,13 +434,22 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
 		// transaction, but there's no way to check that here.
 		return nil, nil
 	}
-
-	ts, err := a.Chain.LoadTipSet(ctx, msgLookup.TipSet)
+	if msgLookup.Message != c {
+		return nil, xerrors.Errorf("lookup returned different message: %s != %s", msgLookup.Message, c)
+	}
+	rcptTs, err := a.Chain.LoadTipSet(ctx, msgLookup.TipSet)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load tipset %s: %w", msgLookup.TipSet, err)
 	}
+	ts, err := a.Chain.LoadTipSet(ctx, rcptTs.Parents())
+	if err != nil {
+		return nil, xerrors.Errorf("failed to load tipset %s: %w", rcptTs.Parents(), err)
+	}
+	if rcptTs.Height() <= ts.Height() {
+		return nil, xerrors.Errorf("failed to get tipset by height: expected height > %d, got %d", ts.Height(), rcptTs.Height())
+	}
 
-	tx, err := newEthTxFromMessageLookup(ctx, msgLookup, -1, a.Chain, a.StateAPI, ts)
+	tx, err := newEthTxFromMessageLookup(ctx, msgLookup, -1, a.Chain, a.StateAPI, rcptTs)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to convert %s into an Eth Txn: %w", txHash, err)
 	}
@@ -447,43 +459,81 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
 	if err != nil {
 		return nil, xerrors.Errorf("failed to load messages for tipset: %w", err)
 	}
+	st := a.Chain.ActorStore(ctx)
+	arr, err := blockadt.AsArray(st, rcptTs.Blocks()[0].ParentMessageReceipts)
+	if err != nil {
+		return nil, xerrors.Errorf("load receipts amt: %w", err)
+	}
+
+	if uint64(len(msgs)) != arr.Length() {
+		return nil, xerrors.Errorf("mismatching message and receipt counts (%d msgs, %d rcts)", len(msgs), arr.Length())
+	}
+
 	logIndexStart := uint64(0)
 	// Iterate over all messages till we get to message with cid c
-	for _, msg := range msgs {
-		if msg.Cid() == c {
+
+	resolver := func(ctx context.Context, emitter abi.ActorID, ts *types.TipSet) (address.Address, bool) {
+		idAddr, err := address.NewIDAddress(uint64(emitter))
+		if err != nil {
+			return address.Undef, false
+		}
+
+		actor, err := a.StateManager.LoadActor(ctx, idAddr, ts)
+		if err != nil || actor.DelegatedAddress == nil {
+			return idAddr, true
+		}
+
+		return *actor.DelegatedAddress, true
+	}
+	var found bool
+	for i, msg := range msgs {
+		if msg.Cid() == msgLookup.Message {
+			found = true
 			break
 		}
 
-		// For each message, do a StateSearchMsg to get the msgLookup
-		msgLookup, err := a.StateAPI.StateSearchMsg(ctx, types.EmptyTSK, msg.Cid(), limit, true)
+		var rct types.MessageReceipt
+		found, err := arr.Get(uint64(i), &rct)
 		if err != nil {
-			return nil, xerrors.Errorf("failed to search for message %s: %w", msg.Cid(), err)
+			return nil, xerrors.Errorf("load receipt: %w", err)
+		}
+		if !found {
+			return nil, xerrors.Errorf("receipt %d not found", i)
 		}
-		if msgLookup == nil {
-			// This is the best we can do. In theory, we could have just not indexed this
-			// transaction, but there's no way to check that here.
-			return nil, fmt.Errorf("message not found: %s", msg.Cid())
+		if rct.EventsRoot == nil {
+			continue
 		}
 
-		if msgLookup.Receipt.EventsRoot != nil {
-			// Load the receipt root from the msgLookup and read the events
-			events, err := a.ChainAPI.ChainGetEvents(ctx, *msgLookup.Receipt.EventsRoot)
-			if err != nil {
-				// Fore-recompute, we must have enabled the Event APIs after computing this
-				// tipset.
-				if _, _, err := a.StateAPI.StateManager.RecomputeTipSetState(ctx, ts); err != nil {
-					return nil, xerrors.Errorf("failed get events: %w", err)
-				}
-				// Try again
-				events, err = a.ChainAPI.ChainGetEvents(ctx, *msgLookup.Receipt.EventsRoot)
-				if err != nil {
-					return nil, xerrors.Errorf("failed get events: %w", err)
-				}
+		evtArr, err := amt4.LoadAMT(ctx, st, *rct.EventsRoot, amt4.UseTreeBitWidth(types.EventAMTBitwidth))
+		if err != nil {
+			return nil, xerrors.Errorf("load events amt: %w", err)
+		}
+
+		var evt types.Event
+		err = evtArr.ForEach(ctx, func(u uint64, deferred *cbg.Deferred) error {
+			if u > math.MaxInt {
+				return xerrors.Errorf("too many events")
+			}
+			if err := evt.UnmarshalCBOR(bytes.NewReader(deferred.Raw)); err != nil {
+				return err
 			}
+			_, ok := resolver(ctx, evt.Emitter, ts)
+			if ok {
+				logIndexStart++
+			}
+			return nil
+		})
 
-			// Update logIndexStart with the number of events
-			logIndexStart += uint64(len(events))
+		if err != nil {
+			return nil, xerrors.Errorf("failed to iterate over events: %w", err)
 		}
+
+		// Update logIndexStart with the number of events
+		// logIndexStart += uint64(evtArr.Len())
+	}
+
+	if !found {
+		return nil, xerrors.Errorf("failed to find message %s in tipset %s", c, ts.Key())
 	}
 
 	receipt, err := newEthTxReceipt(ctx, tx, msgLookup, a.ChainAPI, a.StateAPI, ts, logIndexStart)
```
</details>

aarshkshah1992:
@rvagg I think the best way to go ahead here is to use the Event Index. The whole premise of the bug raised by Axelar is that the reponse of the `getReceipt` API should be consistent with the `getLogs` API and the best way to ensure that is for the `getReceipt` to use the same event source that `getLogs` uses given that we've already fixed the logIndex correctness issue in `getLogs`. 

The complexity involved in wrangling out all the messages here and counting the number of events in each by looking at different tipsets for messages and receipts is another reason to simply use the event Index as it's simpler to reason about (event Index has already handled this complexity by using different tipsets for msgs vs receipts).

If there are data availability issues in the event Index, we should fix the event Index.

aarshkshah1992:
Closing in favour of https://github.com/filecoin-project/lotus/pull/12269.

PR #12267, opened 2024-07-19
test: events: test sqlite indexes used in query plan of all selects and updates

rvagg:
This is a stacked PR ontop of #12261; I'm doing it separately because it moves a bunch of things around that are a bit of a distraction in #12261 and I think it's probably easier to see and review as a separate commit.

This is all about running `EXPLAIN QUERY PLAN` on all selects and updates in the events db and ensuring that an acceptable index is used as the first step in the execution. We want to either narrow down by height or tipset and we have a few indexes that do this across two tables.

This should lock in the perf gains in #12261 such that any future tinkering with the indexes will fail this test and cause the author to consider carefully whether the change is appropriate.

### DIFF
--- a/chain/events/filter/index_test.go
+++ b/chain/events/filter/index_test.go
@@ -5,8 +5,11 @@ import (
 	pseudo "math/rand"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 	"testing"
 
+	"github.com/ipfs/go-cid"
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-address"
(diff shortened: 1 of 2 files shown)


chain/events/filter/index_test.go
```diff
@@ -957,3 +962,84 @@ func TestEventIndexPrefillFilterExcludeReverted(t *testing.T) {
...
+		}()
+		// First response to EXPLAIN QUERY PLAN should show us the use of an index that we want to
+		// encounter first to narrow down the search space - either a height or tipset_key_cid index
+		// - sqlite_autoindex_events_seen_1 is for the UNIQUE constraint on events_seen
+		// - events_seen_height and events_seen_tipset_key_cid are explicit indexes on events_seen
+		// - event_height and event_tipset_key_cid are explicit indexes on event
+		rows.Next()
+		var id, parent, notused, detail string
+		require.NoError(t, rows.Scan(&id, &parent, &notused, &detail), "scan explain query plan for query: "+stmt)
+		require.Regexp(t, regexp.MustCompile(" INDEX (sqlite_autoindex_events_seen_1|event(s_seen)?_(height|tipset_key_cid))( |$)"), detail, "index used for query: "+stmt)
```
BigLep, line 1004: I agree the the regex looks right although I think there is a case for being less clever and just being verbose and listing out the 4 options above as its easier to spot check.  I don't know the team convention though.

chain/events/filter/index_test.go
```diff
@@ -957,3 +962,84 @@ func TestEventIndexPrefillFilterExcludeReverted(t *testing.T) {
...
+		t.Logf("[%s] has plan start: %s", strings.Replace(regexp.MustCompile(`(?m)^\s+`).ReplaceAllString(stmt, " "), "\n", "", -1), detail)
```
BigLep, line 1005: Idea (can skip): In case this helps others, add comment like "removes leading whitespace and newlines from stmt"

BigLep (commented):
Wow - good stuff.  Conceptually makes sense and seems right to me.  Thanks for doing this.

chain/events/filter/index_test.go
rvagg: yeah, it was more explicit originally but I decided to compact it because it's just a log line; I've reverted it now, I agree on being more explicit

chain/events/filter/index_test.go
rvagg: agreed, made it more clear; it's more lines because this isn't easy in Go but it's clearer now

chain/events/filter/index_test.go
```diff
@@ -957,3 +960,87 @@ func TestEventIndexPrefillFilterExcludeReverted(t *testing.T) {
...
+			"sqlite_autoindex_events_seen_1",
+			"events_seen_height",
+			"events_seen_tipset_key_cid",
+			"event_height",
+			"event_tipset_key_cid",
+		}
+		indexUsed := false
+		for _, index := range expectedIndexes {
+			if strings.Contains(detail, " INDEX "+index) {
+				indexUsed = true
```
aarshkshah1992, line 998: @rvagg But what if the detail uses some other index before using an expected index ? We wont fail the test here, right ?

rvagg: no, because we only pull the first row of the results - `row.Next()` above, subsequent rows contain the additional indexes

aarshkshah1992 (approved):

Issue #10287, opened 2023-02-16
ETH API: Faster EthGetTransactionReceipt


Stebalien:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated improvement to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Improvement Suggestion

EthGetTransactionReceipt currently replays the message, but this isn't strictly necessary:

1. We can get the event's CID from the receipt.
2. We can determine what the "total cost" _should_ have been from the gas used, gas limit, fee cap, premium, etc. I.e., we can just re-run ComputeGasOutputs.

rjan90:
So `EthGetTransactionReceipt` is currently not replaying the messages anymore (PR: https://github.com/filecoin-project/lotus/pull/12269). Closing as completed (Using the indexer, pulling the events out of the indexer)

PR #12269, opened 2024-07-19
fix: Eth Event Receipt Logs: use event index for logs

aarshkshah1992:
Closes https://github.com/filecoin-project/lotus/issues/12254.


node/impl/full/eth.go
```diff
@@ -1270,7 +1271,33 @@ func (a *EthModule) EthCall(ctx context.Context, tx ethtypes.EthCall, blkParam e
...
+		return nil, err
+	}
+	logs, err := ethFilterLogsFromEvents(ctx, ces, e.SubManager.StateAPI)
+	if err != nil {
+		return nil, err
+	}
+	var out []ethtypes.EthLog
+	for _, log := range logs {
+		if log.TransactionHash == txHash {
+			out = append(out, log)
```
aarshkshah1992, line 1286: @rvagg Doing it this way so we don't need to add one more "msgCid" index to the events Index given that we just optimised perf based on using the height and tip set cid indices.

This will utilise the `tipset_cid` index we already have on the Event Index.

aarshkshah1992:
Writing tests now.

itests/eth_filter_test.go
```diff
@@ -622,6 +637,123 @@ func TestTxReceiptBloom(t *testing.T) {
...
+	require.Equal(t, 3, len(receipts[1].Logs), "Second transaction should have 3 event logs")
+	require.Equal(t, 3, len(receipts[2].Logs), "Third transaction should have 3 event logs")
+
+	// Iterate over all logs in all receipts and assert that the logIndex field is numbered from 0 to 9
+	expectedLogIndex := 0
+	var receiptLogs []ethtypes.EthLog
+	for _, receipt := range receipts {
+		for _, log := range receipt.Logs {
+			require.Equal(t, ethtypes.EthUint64(expectedLogIndex), log.LogIndex, "LogIndex should be sequential")
+			expectedLogIndex++
```
rvagg, line 731: This assumes that the messages are going to land in the same tipset; I'm not sure that's a safe assumption. You have a slower blocktime here (1s) which helps, but you're still just throwing 3 at the mpool and assuming they get picked up at the same time, what if they're not?

Some ideas:
* Add more than 3, to increase probability of >1 landing in a single tipset; then group receipts by tipset and loop through tipsets and only count expectedLogIndex _per tipset_
* Put this whole block in a loop, (give it ~3 retries? more?) such that if you check the receipts and find that they are all in different tipsets then you bork and repeat the loop until they do, then check that they line up _per tipset_

itests/eth_filter_test.go
```diff
@@ -2349,3 +2495,39 @@ func unpackUint64Values(data []byte) []uint64 {
 	}
 	return vals
 }
+
+func deployContract(ctx context.Context, t *testing.T, client *kit.TestFullNode, ethAddr ethtypes.EthAddress,
```
rvagg, line 2499: what's the difference between this and using `client.EVM().DeployContractFromFilename`?

node/impl/full/eth.go
rvagg: Fair enough.

Also if we're using the filter API then we don't have any other choices - however we could take this further if we wanted to - `EventFilterManager#Install` could take a message CID argument and we could build that into the query without adding an index for it, it'll still hit the tipset index but it'll do this same filtering in the sqlite layer rather than having to bubble all of the data up through the cgo interface, through the event object creation and through the eth translation layer too just to get filtered out here.

Maybe not a bit enough deal to bother with now? But it could be a big difference on tipsets with a _lot_ of events and a lot of messages. Maybe add a TODO here to note that this filtering could be done in the sql layer for "msgCid".

node/impl/full/eth.go
```diff
@@ -1270,7 +1271,33 @@ func (a *EthModule) EthCall(ctx context.Context, tx ethtypes.EthCall, blkParam e
...
 
+func (e *EthEventHandler) GetEthLogsForBlockAndTransaction(ctx context.Context, blockHash *ethtypes.EthHash, txHash ethtypes.EthHash) ([]ethtypes.EthLog, error) {
+	ces, err := e.ethGetEventsForFilter(ctx, &ethtypes.EthFilterSpec{BlockHash: blockHash})
+	if err != nil {
+		return nil, err
+	}
+	logs, err := ethFilterLogsFromEvents(ctx, ces, e.SubManager.StateAPI)
+	if err != nil {
+		return nil, err
+	}
```
rvagg, line 1282:
```suggestion
	logs, err := e. EthGetLogs(ctx, &ethtypes.EthFilterSpec{BlockHash: blockHash})
	if err != nil {
		return nil, err
	}
```

isn't this the same thing? without needing to refactor `EthGetLogs`?

node/impl/full/eth.go
```diff
@@ -1270,7 +1271,33 @@ func (a *EthModule) EthCall(ctx context.Context, tx ethtypes.EthCall, blkParam e
 	return ethtypes.EthBytes{}, nil
 }
 
+func (e *EthEventHandler) GetEthLogsForBlockAndTransaction(ctx context.Context, blockHash *ethtypes.EthHash, txHash ethtypes.EthHash) ([]ethtypes.EthLog, error) {
```
rvagg, line 1274:
```suggestion
func (e *EthEventHandler) getEthLogsForBlockAndTransaction(ctx context.Context, blockHash *ethtypes.EthHash, txHash ethtypes.EthHash) ([]ethtypes.EthLog, error) {
```

make private until it needs to be public?

rvagg:
As per my testing methodology for mainnet in https://github.com/filecoin-project/lotus/pull/12262#issuecomment-2238603285 (using [a recent transaction](https://filfox.info/en/message/bafy2bzacebs5e4t2b3gfsudbuqso5ckkkhwzeepaxhe276wvqm3wy3ojzehzu?t=1) for that same contract):

```
time curl http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionReceipt","params":["0x5822ec9187b621b6bdcdbea41c7b52aa2d1d1c88dd9decb38a8f5c6004b14603"],"id":1,"jsonrpc":"2.0"}' | jq .result.logs > /tmp/transaction_receipt.logs
real    0m0.017s
```

```
time curl http://localhost:1234/rpc/v1 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock": "0x3eb89b", "toBlock": "0x3eb89b"}],"id":1}' | jq '[.result[] | select(.transactionHash == "0x5822ec9187b621b6bdcdbea41c7b52aa2d1d1c88dd9decb38a8f5c6004b14603")]' > /tmp/get_logs.logs
real    0m0.011s
```

```
diff /tmp/transaction_receipt.logs /tmp/get_logs.logs
```

(no output)

So it looks correct. In this case `logIndex` starts at `0x77`.

`eth_getTransactionReceipt` is taking longer here than `eth_getLogs`. Running them repeatedly shows a fairly consistent timing averaging around 0.017s vs 0.011s. I think that can be explained by needing to do a bunch of blockstore loads and a call to `ComputeGasOutputs`.

node/impl/full/eth.go
aarshkshah1992: @rvagg The return type of `EthGetLogs` i.e. `EthFilterResult` looks like:

```
type EthFilterResult struct {
	Results []interface{}
}
``` 

as it's supposed to be a generic type that  can work for the other ETH APIs as well that might simply send logs here. So we can't just use the output of `EthLogs` directly as we'll then have to do the messy wrangling of converting it to an `[]EthGetLogs`

rvagg: ohhh, that all makes sense now, sorry, should have spotted that

itests/eth_filter_test.go
aarshkshah1992: @rvagg The main difference is that `DeployContractFromFilename ` uses Filecoin messages to create the transaction.

I'd like to use ETH messages to create contracts and submit transactions here as we're testing ETH RPC.

node/impl/full/eth.go
aarshkshah1992: Done.

itests/eth_filter_test.go
aarshkshah1992: Done.

node/impl/full/eth.go
aarshkshah1992: Done.

rvagg:
@aarshkshah1992 I just discovered this: https://github.com/filecoin-project/lotus/issues/10287 which has some notes about potential speed-ups. Not necessary for this PR but important to file away in the memory banks because it goes to the speed concerns that are similar to trace_block and trace_filter.

aarshkshah1992:
@rvagg Yeah I am putting together a list of high value work items for RPC and will add it there.

itests/eth_filter_test.go
rvagg: I think you should document this then, or rename it to make it clear why it's here so a future someone doesn't just come and remove it

itests/eth_filter_test.go
```diff
@@ -622,6 +637,144 @@ func TestTxReceiptBloom(t *testing.T) {
...
+
+	maxPriorityFeePerGas, err := client.EthMaxPriorityFeePerGas(ctx)
+	require.NoError(t, err)
+
+	paramsArray := [][]byte{params4Events, params3Events, params3Events}
+
+	var receipts []*api.EthTxReceipt
+	var hashes []ethtypes.EthHash
+
+	nonce := 1
```
rvagg, line 693: let's use `nonce` to cap the maximum number of loops, it could be 10 or something largish, but I think we should cap it rather than relying on a 10m timeout; `require.LessThan(t, nonce, 10, "tried too many times to land messages in same tipset")`

rvagg (approved):
A few suggestions in there

itests/eth_filter_test.go
aarshkshah1992: Done.

itests/eth_filter_test.go
aarshkshah1992: Done.

Issue #11590, opened 2024-01-22
FIP-0074: CLI to invoke the new SettleDealPayments method


arajasek:
SPs will now need to invoke this method in order to settle deal payment. The lotus-miner should expose functionality to do so.

This method will be similar to, and should be placed alongside, the existing `actorWithdrawCmd` ([here](https://github.com/filecoin-project/lotus/blob/e6dc720224e6fdf99e2fe88d101b4e829af8363d/cmd/lotus-miner/actor.go#L243)), which performs a similar operation.

jennijuju:
needs https://github.com/filecoin-project/go-state-types/issues/239

LexLuthr:
@arajasek @jennijuju Can you please confirm if there is an existing idea about how to create the bitfield for the params? I am just trying to figure out which deals to include in params. I can filter out based on provider but how do I ensure that I am not including the deals which are not going to result in anything and just waste the gas. Or is my assumption about the waste of gas wrong here?

ZenGround0:
@LexLuthr are you still planning to work on this?

I would included two levels of granularity.  First just filter by provider id.  The gas costs shouldn't be significant in the common case.  Second users can provide a list of deal ids if they really want fine grained control.

LexLuthr:
@ZenGround0 I won't be able to work on this at this time.

akaladarshi:
Hey @rjan90
Can I work on this?

rjan90:
Hey @akaladarshi,

Feel free to take this on. If you have any questions or need guidance, don't hesitate to ask here on GitHub or in the #fil-lotus-dev channel on Filecoin Slack.

akaladarshi:
Hey @rjan90,

Hey @Phi-rjan,
I have raised a draft PR for now.
Please review it and let me know if this is the correct implementation.

I am little bit confused about payments for the deal, it was mentioned in `FIP-0074` that the gas fee will be paid by the invoker of the function.

PR #12270, opened 2024-07-20
feat: expose settle deal command for lotus miner

akaladarshi:
## Related Issues
Fixes: #11590 

## Proposed Changes
Expose settle-deal command for lotus-miner


rjan90:
@akaladarshi Thanks for working on this. Two initial things based on the failing checks:

- In you /lotus-folder, you need to run `make gen` and `make docsgen-cli` for the gen-check and docs-check to pass.
- Add a entry in the unreleased section of the [Changelog](https://github.com/filecoin-project/lotus/blob/master/CHANGELOG.md) for what this feature does.

akaladarshi:
> @akaladarshi Thanks for working on this. Two initial things based on the failing checks:
> 
> * In you /lotus-folder, you need to run `make gen` and `make docsgen-cli` for the gen-check and docs-check to pass.
> * Add a entry in the unreleased section of the [Changelog](https://github.com/filecoin-project/lotus/blob/master/CHANGELOG.md) for what this feature does.

Hey @rjan90, Thanks for informing.

I just wanted to get an idea if this is what was expected for the implementation, so raised a draft PR.
Once I get confirmation I will clean the PR and fix all the failing CI checks.

LexLuthr (changes requested):
We seem to be missing the case for settling all deal payments.
https://github.com/filecoin-project/lotus/issues/11590#issuecomment-2057438506

rjan90:
@LexLuthr Want to give an updated review of this PR after the code changes for settling all deal payments landed.

cli/spcli/actor.go
```diff
@@ -34,6 +36,112 @@ import (
...
+					return err
+				}
+
+				// if all the deals were successful
+				if settlementReturn.Results.SuccessCount == uint64(len(dealIDs)) {
+					fmt.Printf("Successfully settled %d deals\n", settlementReturn.Results.SuccessCount)
+				} else {
+					fmt.Printf("Settled %d out of %d deals\n", settlementReturn.Results.FailCount, len(dealIDs))
+				}
+			}
```
LexLuthr, line 139: We don't need the if statement as method doesn't exists before NV23. We should directly try to unmarshal the return. 
FYI: go-state-type version != network.Version

cli/spcli/actor.go
```diff
@@ -34,6 +36,112 @@ import (
...
+				var settlementReturn markettypes14.SettleDealPaymentsReturn
+				if err = settlementReturn.UnmarshalCBOR(bytes.NewReader(wait.Receipt.Return)); err != nil {
+					return err
+				}
+
+				// if all the deals were successful
+				if settlementReturn.Results.SuccessCount == uint64(len(dealIDs)) {
+					fmt.Printf("Successfully settled %d deals\n", settlementReturn.Results.SuccessCount)
+				} else {
+					fmt.Printf("Settled %d out of %d deals\n", settlementReturn.Results.FailCount, len(dealIDs))
```
LexLuthr, line 137: Incorrect success count calculation.

```suggestion
					fmt.Printf("Settled %d out of %d deals\n", len(dealIDs) - settlementReturn.Results.FailCount, len(dealIDs))
```

cli/spcli/actor.go
```diff
@@ -34,6 +36,112 @@ import (
...
+				}
+			}
+			return nil
```
LexLuthr, line 140: We should display the settlement amount as well. User needs to know if they spent more gas than the actual settlement amount.

LexLuthr: We should probably also show how many deals have completed their lifetime

LexLuthr (changes requested):
Sorry for the delayed review.
Logic is sound. Just need some changes to the result output statements.

cli/spcli/actor.go
akaladarshi: Done.

Just a question is there any other state the deal could be in apart from `success` and `fail`?
If not then we can do something like this: 
`fmt.Printf("Settled %d out of %d deals\n", settlementReturn.Results.SuccessCount, len(dealIDs))`

cli/spcli/actor.go
akaladarshi: Done.

cli/spcli/actor.go
LexLuthr: Return type is already predefined so it will always stick to those types. Your suggestion seems better than current one.

LexLuthr: I will approve once this final change is finished.

akaladarshi: Done.

LexLuthr (approved):

beck-8:
How can users distinguish whether they have already executed manual settlement? Is there no way to distinguish? Does that mean that if the user has 100,000 transactions, it must be executed atomically, otherwise he doesn't know which part he executes? Will repeated execution waste gas?

PR #12272, opened 2024-07-22
[skip changelog] chore: drand: cleanup legacy and unused configs

ribasushi:
## Related Issues
https://github.com/filecoin-project/lotus/pull/10476

## Proposed Changes

- Remove the network-portion of `mainnet` just like it was done for `incentinet` in https://github.com/filecoin-project/lotus/pull/10476. This will stop network clients being built for the unused network and will resolve spurious errors like:
```
{"level":"info","ts":"2024-07-09T15:36:34.379+1000","logger":"drand","caller":"client/optimizing.go:194","msg":"","optimizing_client":"endpoint down when speed tested","client":"HT
TP(\"https://api.drand.secureweb3.com:6875/\").(+verifier)","err":"decoding response: EOF"}
``` 

- [As per ](https://filecoinproject.slack.com/archives/C047Z15JNEA/p1721656072094789?thread_ts=1720504044.538999&cid=C047Z15JNEA) @AnomalRoil remove unused `devnet` config

### DIFF
--- a/build/buildconstants/drand.go
+++ b/build/buildconstants/drand.go
@@ -30,22 +30,6 @@ type DrandPoint struct {
 }
 
 var DrandConfigs = map[DrandEnum]DrandConfig{
-	DrandMainnet: {
-		Servers: []string{
-			"https://api.drand.sh",
-			"https://api2.drand.sh",
-			"https://api3.drand.sh",
-			"https://drand.cloudflare.com",
-			"https://api.drand.secureweb3.com:6875", // Storswift
-		},
-		Relays: []string{
-			"/dnsaddr/api.drand.sh/",
-			"/dnsaddr/api2.drand.sh/",
-			"/dnsaddr/api3.drand.sh/",
-		},
-		IsChained:     true,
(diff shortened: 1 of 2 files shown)


rvagg:
As mentioned on Slack: I'm concerned that removing these configs impact clients syncing the chain from before the epochs where we've bumped the drand beacon. We're on quicknet now as per nv22 but if you wanted to sync prior to quicknet then I think you might need the config for mainnet? And then there's Incentinet used even earlier in the network's life too. Is it necessary to have access to these to sync? Could we make more headway on this issue by attacking how the clients are instantiated and only do it on demand: https://github.com/filecoin-project/lotus/blob/1b06f5ef881c705607dff637cc7fc683365b1bce/chain/beacon/drand/drand.go#L239 - i.e. we have one client for each network, so we have 3 for mainnet running now and 2 for calibnet.

AnomalRoil (approved):

ZenGround0 (approved):

rvagg:
https://github.com/filecoin-project/lotus/issues/12414 is apparently one result of removing the network config, we've still been using it to look up per-epoch beacons with `StateGetBeaconEntry`.

PR #12286, opened 2024-07-24
feat: api: F3GetF3PowerTable

jennijuju:
add F3GetF3PowerTable to get the f3 power table w/  active f3 participants


jennijuju:
(skipping change log as we dont expect any user to be using these unstable api yet, they are for dev testing atm

Stebalien (approved):

PR #12289, opened 2024-07-24
[skip changelog]feat: p2p: environment variables for disabling DHT query filter and routing table filter

hanabi1224:
## Related Issues

https://github.com/ChainSafe/forest/issues/4576
https://github.com/ChainSafe/forest/pull/4580
https://github.com/ChainSafe/forest/pull/4581

## Proposed Changes

We're setting up local and CI regression tests for Forest bootstrap node to make sure the Kademlia setup works for Lotus. In both test scenarios the bootstrap node has no public IP, and we need to disable the DHT PublicRoutingTableFilter in Lotus. This PR introduces environment variables for doing that.

(Skipping changelog as this is more of a developer-facing feature)

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [x] Update CHANGELOG.md or signal that this change does not need it.
  - If the PR affects users (e.g., new feature, bug fix, system requirements change), update the CHANGELOG.md and add details to the UNRELEASED section.
  - If the change does not require a CHANGELOG.md entry, do one of the following:
    - Add `[skip changelog]` to the PR title
    - Add the label `skip/changelog` to the PR
- [x] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [x] Tests exist for new functionality or change in behavior
- [x] CI is green

PR #12292, opened 2024-07-24
feat: p2p: allow overriding bootstrap nodes with  environmemnt variable

hanabi1224:
## Related Issues

https://github.com/ChainSafe/forest/issues/4576
https://github.com/ChainSafe/forest/pull/4580
https://github.com/ChainSafe/forest/pull/4581
#12289

## Proposed Changes

Allow overriding bootstrap nodes at runtime with network specific environment variables

## Checklist

- [x] Update CHANGELOG.md or signal that this change does not need it.
- [x] CI is green


rvagg:
Is there a good reason to make new env vars for the different network types? Wouldn't a single `LOTUS_P2P_BOOTSTRAPPERS` suffice for all of them?

hanabi1224:
> Is there a good reason to make new env vars for the different network types? Wouldn't a single `LOTUS_P2P_BOOTSTRAPPERS` suffice for all of them?

Hi @rvagg I was assuming a developer could run different networks on the same machine(not necessarily at the same time), maybe devnet with LOTUS_P2P_BOOTSTRAPPERS_DEVNET then calibnet, in this case if it's a shared variable, they will need to always change the value whenever they switch the network. Does that make sense?

rvagg:
Not really, this seems like the kind of env var you'd want to localise to the instance itself, not set in the global environment, so it gets targeted to whatever instance you're running. We also don't have a precedent of doing this, I don't believe there's ever been a need (`git grep .Getenv.*LOTUS | awk -F'"' '{print $2}' | sort | uniq`). It might make sense if you could have two network types in the same lotus instance, but you can't, you have one path through the config generation, in fact, it's specific to a compiled binary.

Unless you have a specific workflow in mind how you'd consume this that would make it particularly awkward to switch for the different network types?

hanabi1224:
@rvagg  changed to using a single `LOTUS_P2P_BOOTSTRAPPERS`

rvagg (approved):

PR #12299, opened 2024-07-24
chore: release: merge release/v1.28.0 into releases

rvagg:
Merge of release/v1.28.0 into releases prefixed with 3 reverts of the auto-release CI changes so this lands a little more cleanly and we don't end up with the auto-release tooling on this branch before we want it.

My understanding was that we wanted the auto-release flow for v1.28.0; but I think I got that wrong and we wanted to enable it _after_ v1.28.0. So we can add them in the next release. For now they're simply reverted.

This is an alternative to https://github.com/filecoin-project/lotus/pull/12298 that doesn't involve rebasing and reverts those CI related commits:

```
$ git diff --stat v1.28.0..rvagg/releases/v1.28.0
 CHANGELOG.md | 42 ++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 42 insertions(+)
```


BigLep (approved):
Looks good - thanks!

jennijuju (approved):

PR #12311, opened 2024-07-27
[skip changelog] chore: logging: switch runtime Print() spots to use the logger

ribasushi:
## Proposed Changes
While examining [a fullsync log](https://github.com/filecoin-project/lotus/pull/12272#issuecomment-2244366061) noticed that the actor bundle loader prints directly to STDOUT. Grepping briefly through the codebase uncovered only 2 spots like that, so fix them both.

### DIFF
--- a/node/bundle/bundle.go
+++ b/node/bundle/bundle.go
@@ -3,11 +3,11 @@ package bundle
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"io"
 	"os"
 
 	"github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log/v2"
 	"github.com/ipld/go-car"
 	"golang.org/x/xerrors"
 
@@ -18,6 +18,8 @@ import (
 	"github.com/filecoin-project/lotus/build/buildconstants"
 	"github.com/filecoin-project/lotus/chain/actors"
 )
+
+var log = logging.Logger("bundle")
 
 func LoadBundleFromFile(ctx context.Context, bs blockstore.Blockstore, path string) (cid.Cid, error) {
 	f, err := os.Open(path)
@@ -55,7 +57,7 @@ func LoadBundles(ctx context.Context, bs blockstore.Blockstore, versions ...acto
 			// All manifests are registered on start, so this must succeed.
 			return xerrors.Errorf("unknown actor version v%d", av)
 		}
-		fmt.Printf("manifest cid: %s\n", manifestCid)
+		log.Infof("manifest cid: %s", manifestCid)
 
 		if haveManifest, err := bs.Has(ctx, manifestCid); err != nil {
 			return xerrors.Errorf("blockstore error when loading manifest %s: %w", manifestCid, err)


chain/consensus/filcns/upgrades.go
```diff
@@ -2521,9 +2519,7 @@ func UpgradeActorsV14(ctx context.Context, sm *stmgr.StateManager, cache stmgr.M
 		return cid.Undef, xerrors.Errorf("migrating actors v14 state: %w", err)
 	}
 
-	if term.IsTerminal(int(os.Stderr.Fd())) {
-		fmt.Print(upgradeSplash)
```
rvagg, line 2525: I think this one's fine .. we probably should have made it check `os.Stdout.Fd()` though; it just means that it's not going anywhere for anyone not running it in a console. which is great. If it goes into the log then it messes up people's logs with the big ugly waffle splash. Alternatively we could print the splash to stderr because it really is a throw-away cosmetic thing.

chain/consensus/filcns/upgrades.go
```diff
@@ -2519,7 +2521,9 @@ func UpgradeActorsV14(ctx context.Context, sm *stmgr.StateManager, cache stmgr.M
 		return cid.Undef, xerrors.Errorf("migrating actors v14 state: %w", err)
 	}
 
-	log.Info("\n\n" + upgradeSplash)
+	if term.IsTerminal(int(os.Stderr.Fd())) {
+		fmt.Fprint(os.Stderr, upgradeSplash)
```
masih, line 2525: FWIW, I would vote for removing the cosmetic splash in favour of consistently logging the system behaviour. 

In that, as a system administrator for a given version I want to be able to infer sequence of events as they happened whether I am looking at CLI output or logs. 

Assuming there is no concrete reason for the splash other than eye candy the most straight forward way I can think of to achieve this is to remove splash sugar, and simply log.

Having bored you fine people to tears, I have no strong opinions on this.

masih (approved):

rvagg:
You've made this more controversial than it should be and playing divide and conquer among maintainers is not appreciated.

Please revert the changes entirely in chain/consensus/filcns/upgrades.go and if you really feel strongly enough to bikeshed about a stdout print that's already gated by a tty check then feel free to open a separate PR.

The other change is reasonable.

rvagg (changes requested):
remove chain/consensus/filcns/upgrades.go changes

rvagg (approved):

rvagg:
> @masih and @ZenGround0 step in and say "this is not cool, do the old thing"

As per @masih:

> FWIW, I would vote for removing the cosmetic splash in favour of consistently logging the system behaviour.

I don't read his text or @ZenGround0's 👍 as a vote for your original change but for removing the splash entirely.

That key misunderstanding - either yours or mine - is why I'm asking this to be scaled back because this is not worth anyone's time doing here, for the sake of a splash, that's gated by a tty check and doesn't touch the logs, yet your change makes 36 lines of ascii art go into everyone's logs: https://github.com/filecoin-project/lotus/blob/4cc7061c8f7084a89062b21eafecffea896b1f34/chain/consensus/filcns/UpgradeSplash.txt

PR #12312, opened 2024-07-27
[skip changelog] chore: blockstore_gc: minor cleanups

ribasushi:
## Proposed Changes
Two standalone fixes around GC - a forgotten `flush()` and lack of `ctx` propagation.

The refator separating the iteration into its own function allows "walking the store with a function", moving a lot of the complexity a step down, leaving [just the logic in place for easier read](https://github.com/filecoin-project/lotus/blob/d6fe66d8280a7c/blockstore/badger/blockstore.go#L393-L416).

cc @ZenGround0 to review, as this used to be their domain

### DIFF
--- a/blockstore/badger/blockstore.go
+++ b/blockstore/badger/blockstore.go
@@ -12,7 +12,7 @@ import (
 
 	"github.com/dgraph-io/badger/v2"
 	"github.com/dgraph-io/badger/v2/options"
-	"github.com/dgraph-io/badger/v2/pb"
+	badgerstruct "github.com/dgraph-io/badger/v2/pb"
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	ipld "github.com/ipfs/go-ipld-format"
@@ -242,7 +242,7 @@ func (b *Blockstore) unlockMove(state bsMoveState) {
 // are persisted to the new blockstore; if a failure occurs aboring the move,
 // then they must be peristed to the old blockstore.
 // In short, the blockstore must not lose data from new writes during the move.
-func (b *Blockstore) movingGC() error {
+func (b *Blockstore) movingGC(ctx context.Context) error {
 	// this inlines moveLock/moveUnlock for the initial state check to prevent a second move
 	// while one is in progress without clobbering state
 	b.moveMx.Lock()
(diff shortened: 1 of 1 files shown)


blockstore/badger/blockstore.go
```diff
@@ -389,37 +389,66 @@ func symlink(path, linkTo string) error {
...
-		batch := to.NewWriteBatch()
-		defer batch.Cancel()
+// doCopy copies a badger blockstore to another
+func (b *Blockstore) doCopy(ctx context.Context, from, to *badger.DB) (defErr error) {
+	batch := to.NewWriteBatch()
+	defer func() {
+		if defErr == nil {
+			defErr = batch.Flush()
+		}
+		if defErr != nil {
```
rvagg, line 399:
```suggestion
		} else {
```

rvagg: or just use an early return to make this even cleaner

blockstore/badger/blockstore.go
```diff
@@ -627,7 +656,18 @@ func (b *Blockstore) Flush(context.Context) error {
 	b.lockDB()
 	defer b.unlockDB()
 
-	return b.db.Sync()
+	// fsync the new db first
+	if b.dbNext != nil {
+		if err := b.dbNext.Sync(); err != nil {
+			return err
```
rvagg, line 662: should we at least continue and try and sync the db this was actually called on? perhaps we collect this error and return a multierr if we get one from both

rvagg: `go.uber.org/multierr` is used more often and it'd be my preferred option, so just a `multierr.Combine(err1, err2)` should be fine here, and it'll play fine with `nil`s too.

blockstore/badger/blockstore.go
rvagg: I don't mind too much, the main point was the `== nil` along with `!= nil`, if that's gone then you can do what you like here; the `else` would be fine too

blockstore/badger/blockstore.go
```diff
@@ -389,37 +390,66 @@ func symlink(path, linkTo string) error {
...
+		if defErr == nil {
+			defErr = batch.Flush()
+		}
+		if defErr != nil {
+			batch.Cancel()
+		}
+	}()
 
-		for _, kv := range list.Kv {
-			if kv.Key == nil || kv.Value == nil {
```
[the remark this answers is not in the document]
ZenGround0: Cause its here: https://github.com/filecoin-project/lotus/pull/12312/files#diff-a021eaf5f05caa0ebd58294daec4cfd559e3b07a8cdc8efb3769f73cbcd64947R433 !

ZenGround0 (approved):

Issue #12293, opened 2024-07-24
 Meta Issue: Fixing high impact correctness and performance problems in ETH RPC API for snapshot synced nodes


aarshkshah1992:
This issue aims to be a meta-issue to capture and track work that needs to be done to enhance correctness, performance, and stability of the ETH RPC API on snapshot synced nodes. Note that improving performance for ETH RPC API  on archival nodes is out of scope for this issue and will be addressed by a future issue.

Our goal is to improve the developer experience (DX) for key partners, including:

1. ETH Subgraph Providers (e.g., Vulcanize)
2. EVM Explorers (e.g., Blockscout)
3. Cross-chain Bridges (e.g., Axelar)
4. DApp Developers using existing ETH tooling which relies on ETH RPC API

### Correctness and data availability issues in the chain state Indexes used by the ETH RPC API

Currently, we maintain three primary indices on the chain state, which are essential for both correctness and performance of multiple ETH RPC APIs.

#### Transaction Index 

- Maps an ETH transaction hash to a Filecoin message CID.
- This mapping is one-way; if it's not in the index, there's no way to go from an ETH transaction hash to a Filecoin message CID.
- APIs depend on this index to lookup the Filecoin message CID for a given ETH transaction hash, and then lookup the message by CID to get the actual transaction data.
- If an entry is not found in this Index for a given transaction hash, the subsequent lookup for the message will be slow and likely fail if the query pertains to an ETH transaction.

#### Message Index 

- Maps a Message("transaction") to it's tipset and block number
- Heavily used by ETH RPC to lookup the tipset containing a given message CID to get the full message/receipts
- A miss on this Index leads to a painfully slow search through the state store for the requested message

#### Event Index 

- Associates each tipset/block with the event logs for all transactions contained within.
- This index is crucial for the functionality of the ETH Event APIs, which are extensively utilized by subgraph providers and bridges.
- Issues with data availability in this index can lead to costly recomputation of tipsets to regenerate events or result in the Events API failing to return events that should be available—assuming the corresponding tipset is present in the chain state, these events should be reliably indexed.

#### All of the above indices suffer from some or all of the following problems that need to be fixed:

- They're not reset and rehydrated when a node syncs from a snapshot
- They're not automatically backfilled on node startup
[...]

PR #12316, opened 2024-07-27
feat: add support for badger version 4 while keeping version 2 support

snissn:
wip 

TODO at the moment:
- [ ] clean up Options changes and ensure no features are lost there

## Checklist

- [ ] CI is green


ZenGround0 (commented):
Overall it looks quite good.  I'm going to take some more time to internalize more exactly what is happening.

blockstore/splitstore/markset_badger.go
```diff
@@ -443,7 +443,7 @@ func closeBadgerDB(db *badger.DB, path string, persist bool) error {
 // badger logging through go-log
 type badgerLogger struct {
```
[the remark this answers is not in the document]
snissn: I think the need for this type is the same overall problem that v2 and v4 even if they have overlap the go compiler sees them as different types

node/modules/blockstore.go
```diff
@@ -56,7 +56,16 @@ func BadgerHotBlockstore(lc fx.Lifecycle, r repo.LockedRepo) (dtypes.HotBlocksto
 		return nil, err
 	}
 
-	opts, err := repo.BadgerBlockstoreOptions(repo.HotBlockstore, path, r.Readonly())
+	c, err := r.Config()
+	if err != nil {
+		return nil, err
+	}
+	cfg, ok := c.(*config.FullNode)
```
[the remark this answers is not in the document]
snissn: Your point feels valid! It is really confusing and expect a few tweaks will be needed. I had a solution more like the way you are suggesting and will try to review the code tomorrow to see if I can improve the injection of the version variable

BigLep:
Is there some context we can look to on why we're doing this change?  (I'm not saying we shouldn't do it, but I want to make sure the rationale is captured.)

node/repo/blockstore_opts.go
```diff
@@ -1,63 +1,12 @@
...
-	badgerbs "github.com/filecoin-project/lotus/blockstore/badger"
+	versions "github.com/filecoin-project/lotus/blockstore/badger/versions"
 )
 
 // BadgerBlockstoreOptions returns the badger options to apply for the provided
 // domain.
-func BadgerBlockstoreOptions(domain BlockstoreDomain, path string, readonly bool) (badgerbs.Options, error) {
-	opts := badgerbs.DefaultOptions(path)
-
-	// Due to legacy usage of blockstore.Blockstore, over a datastore, all
```
ZenGround0, line 15: Speaking as someone who has struggled with learning about our use of badger these comments are quite valuable and shouldn't be dropped.  Even if some of them are out of date its better to have them around so we know what the original intention was behind these values.  I would copy them over [here](https://github.com/filecoin-project/lotus/pull/12316/files#diff-c8625dbecfeb5154905c055e61a49cd1405c44670fdd05da7f221f9902468e20R157) and maybe on the v4 function too.

blockstore/badger/versions/badger_v2.go
```diff
@@ -0,0 +1,314 @@
...
+
+func (b *BadgerV2) Load(r io.Reader, maxPendingWrites int) error {
+	return b.DB.Load(r, maxPendingWrites)
+}
+
+func (b *BadgerV2) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
+	return nil, fmt.Errorf("AllKeysChan is not implemented")
+}
+
+func (b *BadgerV2) DeleteBlock(context.Context, cid.Cid) error {
```
ZenGround0, line 185: I'm a bit confused about what's going on here.  Are these not implemented because they are never called?  If that's the case would it make more sense to just leave them out of the interface?

blockstore/badger/versions/badger_v4.go
```diff
@@ -0,0 +1,304 @@
...
+		if len(kvs) == 0 {
+			return nil
+		}
+		return iter(kvs)
+	}
+	return stream.Orchestrate(ctx)
+}
+
+func (b *BadgerV4) DefaultOptions(path string, readonly bool) Options {
+	var opts Options
```
ZenGround0, line 149: I'm a bit concerned that users will try to run badger v4 but then the difference in default options will make them think something is wrong.  I guess it depends on whether we think people should upgrade to latest version without being badger power users.  What was your thinking behind dropping all of the default settings for badger V2 for V4?

Stebalien:
One of the biggest issues we have with badger is performance in _large_ datastores. That's harder to test, but it would be nice to see how this performs at 1TiB+.

> [For historical reasons](https://github.com/filecoin-project/lotus/blob/v1.28.1/node/repo/blockstore_opts.go#L15-L18) all current keys stored in badger are of the form /blocks/{{base32-enc-of-entire-mh}}, which means 66 bytes per key. This is about 1.6~1.7TiB (!!!!) just for the index on a full archival node. At the very least multihash need to be stored as-is in binary form, with no prefix, giving you about 36 bytes per key.

Completely agree, although I think it's 38 bytes for blake2b. But 38 bytes versus 69 is still nearly a 2x savings.

This should be a relatively straight-forward optimization and is likely worth it.

> If you go further and apply [something like this](https://github.com/ribasushi/ltsh/commit/77caa938b4c1ece) you can safely shrink the key size down to only 16 bytes.

Unfortunately, checking if we have a collision isn't quite enough, we need to be able to make progress even if there's a collision (otherwise we could end up with halts). This is doable, although non-trivial. Personally, I'd:

1. Use the _second_ half of the key (16 bytes) as the value.
2. Store the prefix of the key in the value, concatenated with the actual block.
3. If there's a collision on _read_, try again with the full key.
4. If there's a collision on _write_, store the new block with the full key.

Which brings us down to ~half a TiB of memory instead of 2TiB for a full archival node (if they want fast reads). The real question is.... is that enough of a gain to excuse the extra complexity? @snissn what do you think? How much effort are you willing to put into this?

----

Importantly, I don't think we should block this change on additional improvements. But, at a minimum, making the first change and using binary keys seems worth it.

Stebalien:
@ribasushi pointed out that we only store the multihash, so it's 36 not 38 bytes.

rvagg:
Won't key collision handling mean that every Put needs a Has before it? Internally badger is doing its own Has for every Put, so we'd end up doing even more work.

Stebalien:
> Won't key collision handling mean that every Put needs a Has before it? Internally badger is doing its own Has for every Put, so we'd end up doing even more work.

Yes, but we already do a `Has` because badger is  _versioned_ under the covers and will just keep writing new values before eventually GCing. To avoid writing a ton of duplicate data, we check before we write.

We're not _currently_ doing this under a transaction but we'd need to in order to make this safe. On the other hand, writes always use transactions anyways so... that probably won't introduce much overhead.

BigLep:
@snissn : when you're back, I'm interested to hear your thoughts on additional improvements here that are being suggested and what the rollout strategy will be.  I'm a newbie here, but I'm wondering things like how does someone roll forward, how do they rollback, etc.

snissn:
@BigLep it would really depend on your use case. 
> how does someone roll forward, how do they rollback, etc.
Badger directories are folders so you could just retain a copy of your folder. Upgrading from v2 to v4 you could save a copy of your old badger data folder and roll it back if there was a concern. The use case for many node operators however is that one would take a recent mainnet dump and import it into a V4 badger and it would work just like the v2 badger instance would have. 

So if someone is starting from nothing the migration is very easy - ie just import the chain onto an empty folder with the configuration changed from badger v2 to badger v4.

snissn:
@ribasushi @Stebalien 

I appreciate the effort to improve the on-disk efficiency of key data, and I agree it's an important improvement. However, I believe incorporating those changes into this PR might not be the best approach. This PR primarily focuses on the badger database engine import, while the changes suggested by @ribasushi to improve the on disk efficiency of key data mostly has code changes in `blockstore/badger/blockstore.go`. The code in this PR makes barely any changes to `blockstore/badger/blockstore.go`. Incorporating these changes, while valuable, are quite distinct and could complicate the review process if combined.

I am more than willing to support the initiative to improve key efficiency in a separate PR. This way, we can ensure both enhancements are implemented effectively without delaying either.

node/modules/blockstore.go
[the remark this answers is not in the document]
snissn: fixed here 50cabd0da

blockstore/badger/blockstore_test.go
```diff
@@ -87,20 +107,20 @@ func newBlockstore(optsSupplier func(path string) Options) func(tb testing.TB) (
...
 
-func testMove(t *testing.T, optsF func(string) Options) {
+func testMove(t *testing.T, optsF func(string, bool, int) versions.Options) {
 	ctx := context.Background()
 	basePath := t.TempDir()
 
 	dbPath := filepath.Join(basePath, "db")
 
-	db, err := Open(optsF(dbPath))
+	db, err := Open(optsF(dbPath, false, BADGER_VERSION_2))
```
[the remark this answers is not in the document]
snissn: you pointed out a bug. thank you. fix here cab56a839

node/repo/blockstore_opts.go
snissn: thanks for this. i've restored the comments and deleted dead code that wasn't being used c4611608f

blockstore/badger/versions/badger_v4.go
snissn: this is an artifact of some confusion I had implementing the PR. I've cleaned it up and removed this code c4611608f. thanks for the call out 👍

blockstore/badger/versions/badger_v2.go
snissn: Thanks - I think that these are not needed. I was stuck in mud metaphorically with the node builder system and may have temporarily had a configuration that tried to convert the badger object to a block object. I have removed DeleteBlock and AllKeysChan from the interface and the v2 and v4 implementation and optimistically as I write this now it appears everything is compiling properly and working as intended with them removed 4de5f718c123105dd98596b74c7dff302ff761af and 52604e948e28b80829f4fe131eec2a249aa40d26

blockstore/badger/versions/badger_v4.go
```diff
@@ -0,0 +1,277 @@
...
+	return item.Item.Version()
+}
+
+func (item *BadgerV4Item) ValueCopy(dst []byte) ([]byte, error) {
+	return item.Item.ValueCopy(dst)
+}
+
+func (item *BadgerV4Item) ValueSize() int64 {
+	return item.Item.ValueSize()
+}
```
Stebalien, line 277: We're already embedding `Item`. Do we need to re-define these functions as well? I think we already meet the required interface, right?

Stebalien: (this and the other cases)

blockstore/badger/blockstore.go
```diff
@@ -683,13 +588,14 @@ func (b *Blockstore) Has(ctx context.Context, cid cid.Cid) (bool, error) {
 		defer KeyPool.Put(k)
 	}
 
-	err := b.db.View(func(txn *badger.Txn) error {
+	err := b.db.View(func(txn badger.Txn) error {
 		_, err := txn.Get(k)
 		return err
 	})
 
+	errKeyNotFound := b.db.GetErrKeyNotFound()
```
Stebalien, line 596: This isn't great. Other options:

1. Re-write errors inside these functions. We'd need to be careful to make sure we don't miss any cases...
2. Horrible but... we _can_ `badger2.ErrKeyNotFound = badger4.ErrKeyNotFound`. Really, it's not _that_ horrible (unless someone is stashing these errors early and using those stashed version for comparison).
3. Change the APIs so we don't need to check against special errors?
4. Provide `IsKeyNotFound` etc. helpers that abstract over different badger versions? Also kind of nasty.

I can't think of a great solution, tbh.

node/config/types.go
```diff
@@ -484,6 +484,7 @@ type Pubsub struct {
 type Chainstore struct {
 	EnableSplitstore bool
 	Splitstore       Splitstore
+	BadgerVersion    int
```
rvagg, line 487:
```suggestion
	// EXPERIMENTAL FEATURE. USE WITH CAUTION
	// BadgerVersion switches the version of the Badger database engine. The default is 2 which has
	// been well-tested and widely deployed. Switching this to version 4 will enable the new version
	// of Badger, but the blockstore will be incompatible with Lotus running with version 2. Switching
	// versions can only be done when the blockstore is empty and will be repopulated from a snapshot
	// or chain sync. It cannot be upgraded or downgraded with existing data and there is currently
	// no automatic migration tooling.
	// This is an experimental feature and should not be used in production.
	BadgerVersion    int
	EnableSplitstore bool
	Splitstore       Splitstore
```

moving it up is going to help readability since `Splitstore` is a large nested block

blockstore/badger/versions/badger.go
```diff
@@ -0,0 +1,50 @@
...
+			db = BadgerDB(&BadgerV4{dbV4})
+		}
+	case 2:
+		var dbV2 *badgerV2.DB
+		dbV2, err = badgerV2.Open(opts.V2Options)
+		if err == nil {
+			db = BadgerDB(&BadgerV2{dbV2})
+		}
+	default:
+		return nil, errors.New("unsupported badger version")
```
rvagg, line 43:
```suggestion
		return nil, fmt.Errorf("unsupported badger version: %v", opts.BadgerVersion)
```

blockstore/badger/versions/badger.go
```diff
@@ -0,0 +1,50 @@
...
+func (b *BadgerLogger) Warningf(format string, args ...interface{}) {
+	b.Skip2.Warnf(format, args...)
+}
+func OpenBadgerDB(opts Options) (BadgerDB, error) {
+	var db BadgerDB
+	var err error
+
+	version := opts.BadgerVersion
+
+	switch version {
```
rvagg, line 29:
```suggestion
	switch opts.BadgerVersion {
```

blockstore/badger/versions/badger.go
```diff
@@ -0,0 +1,50 @@
...
+	*zap.SugaredLogger // skips 1 caller to get useful line info, skipping over badger.Options.
+
+	Skip2 *zap.SugaredLogger // skips 2 callers, just like above + this logger.
+}
+
+// Warningf is required by the badger logger APIs.
+func (b *BadgerLogger) Warningf(format string, args ...interface{}) {
+	b.Skip2.Warnf(format, args...)
+}
+func OpenBadgerDB(opts Options) (BadgerDB, error) {
```
rvagg, line 23:
```suggestion
}

func OpenBadgerDB(opts Options) (BadgerDB, error) {
```

node/config/types.go
snissn: thank you! I appreciate the addition of this comment

blockstore/badger/versions/badger_v4.go
snissn: I noticed that [Item.Version()](https://pkg.go.dev/github.com/dgraph-io/badger/v4#Item.Version) was unused and I removed it. 6d5d67918. 

Item is used in blockstore when we use the badger iterator. Item.Key for example is used in `Blockstore.AllKeysChan ()` as

```
for iter.Rewind(); iter.Valid(); iter.Next() {
    ...
    k := iter.Item().Key()
    ...
}
```
 I think we need a specific function for v2 Item and v4 Item as we have it in this PR.

ZenGround0 (approved):
Once tests are passing this LGTM

blockstore/badger/versions/badger.go
snissn: thanks! 624779f27

blockstore/badger/blockstore.go
snissn: I looked into `badger2.ErrKeyNotFound = badger4.ErrKeyNotFound` and the current implementation feels like the most maintainable solution option to me

snissn:
thanks @ZenGround0 I think this is in a good place to be merged. I don't have merge access but please use the squash option to bring this PR in as a single commit when it is merged.

documentation/en/default-lotus-config.toml
```diff
@@ -149,6 +149,19 @@
...
 
 [Chainstore]
+  # EXPERIMENTAL FEATURE. USE WITH CAUTION
+  # BadgerVersion switches the version of the Badger database engine. The default is 2 which has
+  # been well-tested and widely deployed. Switching this to version 4 will enable the new version
+  # of Badger, but the blockstore will be incompatible with Lotus running with version 2. Switching
+  # versions can only be done when the blockstore is empty and will be repopulated from a snapshot
+  # or chain sync. It cannot be upgraded or downgraded with existing data and there is currently
+  # no automatic migration tooling.
+  # This is an experimental feature and should not be used in production.
```
BigLep, line 159: I think it would be good to link to the tracking issue where someone can go for finding updates on what is remaining for moving this out of experimental.

BigLep (commented):
I'm no Lotus veteran so apology in advance for any newbie questions or mistakes.

A few things:

## User communication
1. I think we need to expand the changelog entry.  I put down some potential text.
2. Is there a stronger way to communicate this is experimental?  I'm wondering if the config key and env variable can be prefixed with "Experimental".  That gives a clear signal and then also gives maintainers liberty to change config keys/semantics for when it's not experimental.
3. I think it would be good to have a "Add badger v4 support" tracking issue.  A first task can be adding experimental support (which this PR does) but it can also list any followup tasks we need to do to fully add Badger v4 support.  It also serves as a good place to store the rationale for this work, benchmarks, etc. rather than having it embedded in PR comments.
4. I think it would be good have @rjan90 review/approve, at least from the user communication/rollout perspective.

## KV store improvements
1. I understand the suggestion to not overload this PR - sounds good.  Do we have a backlog item tracking the space improvements suggestions?
2. No pressure, but is this something you're planning to do a fast-followup with?
3. My understanding of @ribasushi 's point is that if we're already going to be asking users at some point to do a migration when they update to badger v4, we should bundle in other things that also require a migration (e.g., key changes).  I'm wondering: how is this going to play out in time?  I'm assuming the user message will be: anytime you change the value of one of these flags (LOTUS_CHAINSTORE_BADGERVERSION or LOTUS_CHAINSTORE_KEY_OPTIMIZATION) one needs to setup a new datastore.  If we ever want to change the default values then we'll need to provide a migraiton path (ideally automated while keeping a backup).  I just want to make sure we're thinking ahead for how we minimize user impact and account for the total work to fully complete this so we can remove old code paths.  (I think this is the kind of discussion/planning that would be good in an overarching issue.)

snissn:
@BigLep Thank you for your feedback and suggestions! I appreciate your guidance as I navigate these areas—I'm still learning, so your input is invaluable.

### User Communication

1. **Expanding the Changelog Entry**  
   I understand where you're coming from. After reviewing the general format of the `CHANGELOG.md`, it seems that this PR aligns with the typical entries. However, I find your notes very insightful. As we gather positive feedback on the Badger upgrade from the community, it would be great to collaborate on enhancing our documentation strategy to make the rollout as smooth as possible.

2. **Communicating Experimental Features**  
   I agree that it's crucial to signal the experimental nature of this feature clearly. I believe the current setup already makes this quite evident, especially with the recent changes by @rvagg, [which include prominent warnings](https://github.com/filecoin-project/lotus/pull/12316/commits/d2de85e481011c3ef0ef81d9e30cdd3f6a4f5c3d). The config.toml that's generated by `lotus config default` now has this text:
```
[Chainstore]
  # EXPERIMENTAL FEATURE. USE WITH CAUTION
  # BadgerVersion switches the version of the Badger database engine. The default is 2 which has
  # been well-tested and widely deployed. Switching this to version 4 will enable the new version
  # of Badger, but the blockstore will be incompatible with Lotus running with version 2. Switching
  # versions can only be done when the blockstore is empty and will be repopulated from a snapshot
  # or chain sync. It cannot be upgraded or downgraded with existing data and there is currently
  # no automatic migration tooling.
  # This is an experimental feature and should not be used in production.
  #
  # type: int
  # env var: LOTUS_CHAINSTORE_BADGERVERSION
  #BadgerVersion = 2
```
 But I’m always open to refining this if needed!

3. **Tracking Issue for Badger v4 Support**  
   That’s a great idea. I think creating a tracking issue will be very beneficial once we have some feedback on Badger v4’s performance and reliability. This way, we can plan our next steps more effectively.

4. **Involving @rjan90**  
   Definitely, having @rjan90 review this would be valuable, particularly from the perspective of user communication and rollout.

### KV Store Improvements

1. **Tracking Space Improvements**  
   I’m not sure if there’s a backlog item for this, but I can look into it.

2. **Follow-up on Space Improvements**  
   While I’m not opposed to following up quickly, I’d prefer to keep this discussion focused on the current PR for now.

3. **Migration Planning**  
   You raise an excellent point about the importance of minimizing user impact during migrations. I think this is something we should consider carefully, especially as it relates to the timing and communication around these changes. Depending on the user’s setup, the approach might vary significantly, so planning ahead is essential to ensure a smooth transition for everyone.

rvagg:
I'm also concerned about the key efficiency improvements coming on top of this as a separate PR after this is merged. It puts two separate hard-breaking changes in master. If we're doing both then we should do them both pretty closely together, and ideally at the same time. At a minimum I'd like this _not_ to go into a release without the key efficiency improvements also going with it. Unless we come to an agreement that we're not going to try and achieve that any time soon. Reading sentiment and adding my own to it suggests we would like to get it done asap since we have this opportunity with the breakage here.

I agree that tying that functionality to this PR makes it too large and less focused, and it's unfair to lump that burden on @snissn here, but maybe this is a case for a feature branch where we can review and land multiple pieces before they make it into master and accidentally into separate releases.

BigLep:
Thanks @snissn .  Responding to a few things:

## User Communication

> Expanding the Changelog Entry: it seems that this PR aligns with the typical entries

By the time we get to doing a release, certain newly-added features have text added in.  https://github.com/filecoin-project/lotus/blob/master/CHANGELOG.md#v1281--2024-07-24 is an example with F3.  I accept that F3 is bigger than a Badger update, but we want to avoid having to do documentation editorializing in the crunch of trying to get a release out.  I think it's something that needs to be thought about and done as part of the code writing process.

> Communicating Experimental Features

I agree that the config.toml docs have good disclaimers, but someone doesn't see that when they use environment variables and it also doesn't a clear signal that things may change before we make it non-experimental.  If we get this into the name (`LOTUS_EXPERIMENTAL_CHAINSTORE_BADGERVERSION` / `Experimental.BadgerVersion`) this is very clear even if they don't read the doc comments and gives a clear break for when we release a non experimental version.  

(I realize this is a larger discussion about how experimental features are handled in general in Lotus.  I don't know the history here, and so this is something that would be good to have Lotus maintainers chime in on.  My bias is coming from what I saw be clear for users in projects like Kubo ([example](https://github.com/ipfs/kubo/blob/master/docs/experimental-features.md#http-gateway-over-libp2p)).)

> Tracking Issue for Badger v4 Support

I don't think we need/should wait for some on Badger v4’s performance and reliability.  This is a big change, and thus Ideally I don't think this PR should have been opened without an issue explaining the work and providing a central place to capture decision points, findings, next steps, etc.  

## KV Store Improvements
It sounds like maintainers need to:
1. Create a backlog item
2. Determine how interlinked this badger update needs to be with the KV store improvements (i.e., feature branch where both workstream happen, allow them to land in master separately but off by default)
3. Determine a rollout strategy.  I think some of the key things here are:
   1. Is there one config flag (ENABLE_AWESOME_DATASTORE_IMPROVEMENTS) which couples Badger update + key shrinkage improvements or multiple config flags (one for each).
   2. What would be the path for getting rid of old code paths?  Would we do some sort of automated migration on startup of a new Lotus version, provide written manual directions that need to be followed, etc.

If there are other maintainers at the 2024-08-14 maintainer call today, I'll see if I can get answers or input on any of the above and ensure there is clarity on the next steps for landing this PR (and where to land it to).

BigLep:
Below are some notes I took from Lotus maintainer conversation on 2024-08-14.  Others can fill in if I'm misrepresenting:

1. It's great to see experimentation and work to make the datastore better as this a known major pain point for users.  We want to see contributions here and want to see this area of the code get better.  Thank you @snissn for working to move the ball forward!  It would be great to see other experiments by the community here as well (e.g., Postgres-backed datastore).
2. The snag we're running into is whether the disruptiveness of this change is worth it.  If this didn't require a user to do a manual migration, this would be easy to approve and merge.  After a limited time window, we could remove old code, keep the new code, and users would benefit while not increasing the maintenance burden on maintainers.
3. To determine whether this change is worth it, we'd also want to look at:
   - How much does it impact splitstore compaction speed?
   - How much does v4 impact runtime CPU and memory?
   - Are execution time of tispets impacted in a material way? (although this isn't really a user concern)
   It sounded like there was maybe an A/B-testing framework that you have used in the past with Lotus @snissn?
   I know you provided the snapshort import improvement times, but that is a pure Go pathway, and we don't think is representative of other important scenarios.
4. Related to the previous point, we want to make sure we're clear on: "who is the user that is really benefiting enough to warrant the time for this feature?"  Yes faster import snapshot times is nice, but our understanding is that that hasn't been a make or break issue for users.  (Please correct if you've observed otherwise.). This is why we raised the snapshot compaction time measurement, as improving that would be really nice for all.
5. Because of the user and maintenance burden that this change brings, a larger narrative of the benefits and the rollouts steps is needed so we can look at it to help answer is it worth it.  I can understand that help from maintainers may be needed to help generate that and think this through.  Given maintainers are tied up with other areas currently but also to help this PR land rather than stay open while additional testing and followups are done, we think @rvagg's idea of a feature branch makes sense here.  Maybe we name it `feat/faster-datastore-with-badger`?  This is also where other improvements like key shrinkage could land.

Thanks again @snissn .

rvagg:
`feat/faster-datastore-with-badger` made, this PR could be retargeted.

Re feature branch and how to proceed, here's some additional thoughts on how we can be productive with that:

* We should review PRs into that branch like we do with `master` because any eventual merge into master is likely going to be very large and maybe impossible given its eventual scope. Ongoing reviews for incremental changes would help alleviate that burden by building trust in the branch over time.
* Likewise, keeping the branch updated will be helpful.
* As @BigLep suggests, we need to make sure that this moves the needle enough to warrant the additional maintenance and support burden of having yet another permutation of possible Lotus configuration in the wild. So we need to work on quantifying that a bit more, perhaps there's enough here across the scenarios that matter to make it worth a push.
* A migration from 2 -> 4 would certainly help matters and may become necessary for a path forward here so we don't have perpetual v2 support if we deem v4 worthwhile.
* Getting feedback from people using this would be really helpful, we have such variety of node users out there that there may be a subset of users for who this is a significant enough win which justifies further pushes to make this worth pushing further toward release.
* The key efficiency improvements may make this worthwhile, since we can bundle a hard-break in blockstore with significant enough improvements to warrant the pain: do smaller keys make the in-memory indexes small enough to make a dent on concerns of larger node operators? Archival node operators in particular have the problem with badger that the indexes don't fit in memory and badger apparently insists on swapping them in and out when it needs to go fishing.
* As @Stebalien was pointing out in our discussion about this recently, for average users, most blockstore usage comes via the Rust layer; which is an inversion of the original Lotus design: Badger was a reasonable option when cgo was being avoided, but now we have Rust code that needs to call back into Go just to load data from the blockstore. This means that the ultimate "fix" for Lotus' blockstore woes exists in a non-Go blockstore that Go has the burden of calling in to. This may not make life better for archival node operators but there may be an alternative path there where they get an external blockstore that they can scale via traditional means (e.g. postgres).

ZenGround0:
@BigLep @rvagg glad to see you guys thinking about integration wholistically and seriously.

However I think we are overthinking it.  A very standard development usage of lotus is 1) sync from snapshot 2) run for a while 3) lose sync tear down go back to 1.  And SPs are good at spinning lotus up and down.  We're really overestimating the utility of database migration from my pov.  I think it would be a waste of work to write a migration personally.   

I don't think we need to block on making v4 even better, this is an experimental feature, why can't we just break v4 badger support in the future by making breaking improvements?  The feature is clearly marked experimental.  This feature is totally independent of the default so if someone wants to try v4 badger why stop them?

I don't think we need to do so much work quantifying this experimental feature until we commit to working on it ourselves.  It makes sense to do this investigation before we commit to working on this.   We're going to get organic feedback from people in the network only if we make the change available.

What about saying "if this causes more than 2 unhappy user reports we revert this" to limit maintenance overhead? 

And hey maybe I'm underthinking it.  To me it just feels like we're indefinitely stopping merging a complete-enough open source contribution without great reason.

ZenGround0:
A little added context, @snissn is trying to get more performance testing information from RPC providers and getting an experimental feature on master makes the process of convincing them to test a lot easier.

trruckerfling:
Great discussion above. It seems we all agree early user feedback is needed. Can we merge this as an experimental feature so we can start getting early user feedback?  

We can decide on migration as needed after that early feedback comes in. Concretely, Ankr and Protofire seemed keen to try this when we last spoke a month ago and Mike can share this with them once merged

BigLep:
Thanks all for the comments here.  Some thoughts below synthesized from conversations I've had more recently (meaning I can stand behind anything written here but any utility was likely due to others):

It seems like the target user here we're hoping to benefit here is RPC providers.  A couple of things on this:
1. There is other/separate ongoing work right seeking to benefit this userbase as well in https://github.com/filecoin-project/lotus/issues/12293 .  This was deemed to be the most impactful single area to apply limited maintainer time.
2. Does the RPC provider user care about having to restart from a snapshot?  I'm assuming not, but I wanted to check as that helps gage how important we need to weigh the "breakage" this causes if it's enabled.  I saw @ZenGround0 say " And SPs are good at spinning lotus up and down".  I wanted to make sure that applies to RPC providers too.

Maintainers are aware of other ways to improve RPC provider experience as well beyond this badger update, including things like:
1. disable fsync on the chainstore: https://github.com/filecoin-project/lotus/pull/7965
2. reduce the index size (discssed in this PR)
3. disable mmap

It is encouraging to see people pick up the torch on the blockstore problem.  The concern is that we're poking at the edges which still consumes non-trivial time and bandwidth for marginal improvements to something that’s just structurally bad.  The worry is that minor incremental fixes will increase the permutations of how people run lotus, which will make maintainer work harder in the long run.  For example, "are you running badger 2 or 4? are you running the version with fsync defaulting to off or have you set it to on? is mmap enabled on your node? do you have eth API enabled? are you running with historical events? are you saving messages to the blockstore? what’s your splitstore mode?” 

If doing any combination of things can be shown to make a meaningful difference to the life of node operators, then that would be great news and would give weight to schedule more time to take a proper look at this and think holistically on integrating it back into master/main in a non-experimental way.  Without better info, there isn't maintainer conviction to see this moving the needle enough on the “lotus sucks” meter, or adjusting the “filecoin is a hard chain to run” sentiment, or dispelling the “forest can do the same work with a fraction of the resources” meme.

In knowing that having a holistic strategy to dealing with the blockstore problem is likely what's needed but there's no one committed in 2024Q3 to make that happen, I still think the dedicated feature branch path is a good compromise.  I think the couple of RPC providers like Ankr and Protofire who are keen to experiment should be able.  Hopefully we can get some good news from those experiments to help inform priorities later on in the year?  Does that work @snissn and keep you sufficiently unblocked?

snissn:
Hi - I have updated the PR to target `feat/faster-datastore-with-badger`.

I hope that helps!!

BigLep:
Circling back to next steps now that I'm back...

My understanding is that @snissn is doing this work currently as an extra/side project outside of the context of a particular team's priorities and would like to get feedback of how this change performs in a production setting.

It's to this end that the idea of a feature branch was suggested (and has been created).  

Regardless of the branch though, the more pertinent question is how the branch stays updated.  Given Lotus maintainers aren't focused on this more holistic blockstore effort currently ([comment above](https://github.com/filecoin-project/lotus/pull/12316#issuecomment-2294382788)), Lotus maintainers will leave it up to those interested in experimenting here to keep up with `master` or the latest `release/` production release.  

(In retrospect it may have been better not to do this as a "feature branch" as that maybe gives the impression that this feature is more official than it is currently.  But I guess having a feature branch enables actually merging this PR vs. leaving it open indefinitely. 🤷)

Whether in a `feat` branch or `mikers/BadgerVersions`, I think it's important to communicate with anyone who is trying this functionality out that:
1. There currently isn't commitment or plan for how this gets into `master`.  I wouldn't want someone taking time to experiment here to think that we're on the verge of getting this into an official release.
2. Data that will help with this decision (but is not the only consideration) is a production operator showing CPU, memory, and compaction speed of some nodes using v2 vs. those using v4.   
  - Ideally this would be done as a in-parallel A/B test.  
  - (If there are other impactful metrics that operators observe, those are welcome too.)
  - Let's capture this info in the currently-missing issue justifying/explaining this work

@rvagg: if you're good with the above, can you please take on merging?

trruckerfling:
Following up here:
- user testing
   - @snissn spoke with [Ankr](https://filecoinproject.slack.com/archives/C03K2KZEZB3/p1724366999501759) and [Protofire](https://filecoinproject.slack.com/archives/C017HM9BJ8Z/p1724366889017739) to agree to test this and share feedback (thanks all!)
   - have also tagged on maintainer questions from @BigLep @rvagg and added yall so free to follow there
   - ideally, this gives us clarity into v4 upgrade need and readiness
- feature branch maintenance
  - this PR has been retargeted into feature branch as requested. Unfortunately, we can't commit bandwidth to maintain this branch
  - depending on user testing feedback, we can help with updates to this PR as needed

BigLep:
Next steps per 2024-08-28 sync between FilOz and Fil-B teams: 
* @rvagg to review so this can be merged into a feature branch
* Strong signal is then needed from infra providers that this indeed worth pursuing more to get it into `master`.  (This has been discussed in more detail earlier in the PR).  

Additional notes:
* @snissn is the only person so far who has raised their hand to try and drive it forward with infra providers, but Lotus maintainers will keep this feature in mind as they also engage with infra providers as well.
* Lotus maintainers haven't committed to keep this feature branch updated.

rvagg:
I happened to be thinking about this one a couple of days ago when I was working on our fevm archival node. Noting in my head just how hard a migration would be but also that it would be really interesting to measure performance difference at that scale.

What I'd really like out of this PR is a migration tool that lets me upgrade at will. If I could have a `lotus-shed`, or even `lotus` command that I could point to an old repo and a new repo location to perform a full migration. If I could have that, I'd love to try allocating a new volume for a duplicate db and then migrating so I can run either (or both!) and measure some of the differences.

PR #12319, opened 2024-07-29
feat: Lotus Send CLI: Lotus send should work with ETH address receipients

aarshkshah1992:
This fixes the Lotus send CLI to work with ETH address recipients (both `0xff...` corresponding to Filecoin `f0` addresses and `0xabcd` corresponding to Filecoin `f410f` addresses).

Have tested this locally on a local devnet.

```
./lotus send --from-eth-addr 0x4582e0a8f02f45dfb6cca081b1cb82fcf0508c34 0xff000000000000000000000000000000000003ed 10

f4 addr:  t410fiwbobkhqf5c57nwmuca3ds4c7tyfbdbun3hxwmi
Sending message from: t410fiwbobkhqf5c57nwmuca3ds4c7tyfbdbun3hxwmi
Sending message to: t01005
Using Method: 3844450837
bafy2bzacec3lcrfevp4wsuya65t2nien2cjselbmqd3cnr2w32hukbwxg57zy
```

```
./lotus send 0xff000000000000000000000000000000000003ed 10
Sending message from: t3uhfeeespjbuhde6vfxvgyuhoj4danwv3fdsk3m3t46p2kpegm76jpvukgnlb2w7aucldnmq2j4un3nlxs73a
Sending message to: t01005
Using Method: 0
bafy2bzacedwvhrfdopwr7azpjcjvjw22orirulikewdpw6zgc3tg5bgspv3vu
```

```
./lotus send 0x4582e0a8f02f45dfb6cca081b1cb82fcf0508c34 100
Sending message from: t3uhfeeespjbuhde6vfxvgyuhoj4danwv3fdsk3m3t46p2kpegm76jpvukgnlb2w7aucldnmq2j4un3nlxs73a
Sending message to: t410fiwbobkhqf5c57nwmuca3ds4c7tyfbdbun3hxwmi
Using Method: 0
bafy2bzacebnj45yttbtcd75rajfwkqpc27tsru6brfhufolnjtvhsixltitqo
```

```
./lotus send --from-eth-addr 0x4582e0a8f02f45dfb6cca081b1cb82fcf0508c34 0xff000000000000000000000000000000000003ec  0.2
f4 addr:  t410fiwbobkhqf5c57nwmuca3ds4c7tyfbdbun3hxwmi
Sending message from: t410fiwbobkhqf5c57nwmuca3ds4c7tyfbdbun3hxwmi
Sending message to: t01004
Using Method: 3844450837
bafy2bzacedk25ku4sny272z6erhha4bx5pymbfued4wu6434z4yoaka2n6rvu
```

```
./lotus send --from-eth-addr 0x4582e0a8f02f45dfb6cca081b1cb82fcf0508c34 0x28dfd8668534e9caa8fb0b77577c8b78831d0d07 0.2
f4 addr:  t410fiwbobkhqf5c57nwmuca3ds4c7tyfbdbun3hxwmi
Sending message from: t410fiwbobkhqf5c57nwmuca3ds4c7tyfbdbun3hxwmi
Sending message to: t410ffdp5qzufgtu4vkh3bn3vo7elpcbr2dih4tfslca
Using Method: 3844450837
bafy2bzacedrwb75aagtwl74krto7aym5ato7vwqtg6cc4wlgzvmf3wiwrp7ik
```

```
 ./lotus send --from-eth-addr 0x4582e0a8f02f45dfb6cca081b1cb82fcf0508c34 t01005 10
f4 addr:  t410fiwbobkhqf5c57nwmuca3ds4c7tyfbdbun3hxwmi
Sending message from: t410fiwbobkhqf5c57nwmuca3ds4c7tyfbdbun3hxwmi
Sending message to: t01005
Using Method: 3844450837
bafy2bzaceaijk7lflnkt3l6s6irnxsmmoennfj57jpjgn3zi4e53t6qbokkae
```

```
./lotus send --from t410ffdp5qzufgtu4vkh3bn3vo7elpcbr2dih4tfslca t01005 0.2
Sending message from: t410ffdp5qzufgtu4vkh3bn3vo7elpcbr2dih4tfslca
Sending message to: t01005
Using Method: 3844450837
bafy2bzaceb5o4uqbktypwpiabxmqycxku6lcz5yace4hxkvlkhyd424shpdhg
```

```
./lotus send --from t410ffdp5qzufgtu4vkh3bn3vo7elpcbr2dih4tfslca 0xff000000000000000000000000000000000003ed  0.1
Sending message from: t410ffdp5qzufgtu4vkh3bn3vo7elpcbr2dih4tfslca
Sending message to: t01005
Using Method: 3844450837
bafy2bzacecqggccw43b455eady3fdv2apk7tybfi5ua2qrcaox33eo3la4yna
```

### DIFF
--- a/cli/send.go
+++ b/cli/send.go
@@ -131,9 +131,7 @@ var SendCmd = &cli.Command{
 			}
 			fmt.Println("f4 addr: ", faddr)
 			params.From = faddr
-		}
-
-		if params.From == address.Undef {
+		} else {
 			defaddr, err := srv.FullNodeAPI().WalletDefaultAddress(ctx)
 			if err != nil {
 				return fmt.Errorf("failed to get default address: %w", err)
--- a/cli/send.go
+++ b/cli/send.go
@@ -90,7 +90,20 @@ var SendCmd = &cli.Command{
 
 		params.To, err = address.NewFromString(cctx.Args().Get(0))
 		if err != nil {
-			return ShowHelp(cctx, fmt.Errorf("failed to parse target address: %w", err))
+			// could be an ETH address
+			ea, err := ethtypes.ParseEthAddress(cctx.Args().Get(0))
+			if err != nil {
+				return ShowHelp(cctx, fmt.Errorf("failed to parse target address; address must be a valid FIL address or an ETH address: %w", err))
+			}
+			// this will be either "f410f..." or "f0..."
+			params.To, err = ea.ToFilecoinAddress()
+			if err != nil {
+				return ShowHelp(cctx, fmt.Errorf("failed to convert ETH address to FIL address: %w", err))
+			}
+			// ideally, this should never happen
+			if !(params.To.Protocol() == address.ID || params.To.Protocol() == address.Delegated) {
+				return ShowHelp(cctx, fmt.Errorf("ETH addresses can only map to a FIL addresses starting with f410f or f0"))
+			}
 		}
 
 		val, err := types.ParseFIL(cctx.Args().Get(1))
@@ -119,6 +132,17 @@ var SendCmd = &cli.Command{
 			fmt.Println("f4 addr: ", faddr)
 			params.From = faddr
 		}
+
+		if params.From == address.Undef {
+			defaddr, err := srv.FullNodeAPI().WalletDefaultAddress(ctx)
+			if err != nil {
+				return fmt.Errorf("failed to get default address: %w", err)
+			}
+			params.From = defaddr
+		}
+
+		_, _ = fmt.Fprintf(cctx.App.Writer, "Sending message from: %s\n", params.From.String())
+		_, _ = fmt.Fprintf(cctx.App.Writer, "Sending message to: %s\n", params.To.String())
 
 		if cctx.IsSet("params-hex") {
 			decparams, err := hex.DecodeString(cctx.String("params-hex"))
@@ -166,6 +190,7 @@ var SendCmd = &cli.Command{
 		} else {
 			params.Method = abi.MethodNum(cctx.Uint64("method"))
 		}
+		_, _ = fmt.Fprintf(cctx.App.Writer, "Using Method: %d\n", params.Method)
 
 		if cctx.IsSet("gas-premium") {
 			gp, err := types.BigFromString(cctx.String("gas-premium"))
--- a/cli/send.go
+++ b/cli/send.go
@@ -141,9 +141,6 @@ var SendCmd = &cli.Command{
 			params.From = defaddr
 		}
 
-		_, _ = fmt.Fprintf(cctx.App.Writer, "Sending message from: %s\n", params.From.String())
-		_, _ = fmt.Fprintf(cctx.App.Writer, "Sending message to: %s\n", params.To.String())
-
 		if cctx.IsSet("params-hex") {
 			decparams, err := hex.DecodeString(cctx.String("params-hex"))
 			if err != nil {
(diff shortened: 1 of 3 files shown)


cli/send.go
```diff
@@ -167,6 +188,8 @@ var SendCmd = &cli.Command{
 			params.Method = abi.MethodNum(cctx.Uint64("method"))
 		}
 
+		_, _ = fmt.Fprintf(cctx.App.Writer, "Sending message from: %s\nSending message to: %s\nUsing Method: %d\n", params.From.String(), params.To.String(), params.Method)
```
rvagg, line 191: failing unit test because of this now, see CI

```
        	Error Trace:	/home/runner/work/lotus/lotus/cli/send_test.go:115
        	Error:      	Not equal: 
        	            	expected: "bafy2bzacecno6zh4xdnhzugo6fnenefhbmoia3fw72wbqnn55jqoov7xbnob4\n"
        	            	actual  : "Sending message from: f410faaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaonc6iji\nSending message to: f01\nUsing Method: 3844450837\nbafy2bzacecno6zh4xdnhzugo6fnenefhbmoia3fw72wbqnn55jqoov7xbnob4\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1,4 @@
        	            	+Sending message from: f410faaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaonc6iji
        	            	+Sending message to: f01
        	            	+Using Method: 3844450837
        	            	 bafy2bzacecno6zh4xdnhzugo6fnenefhbmoia3fw72wbqnn55jqoov7xbnob4
        	Test:       	TestSendEthereum/simple
```

cli/send.go
```diff
@@ -120,6 +133,14 @@ var SendCmd = &cli.Command{
 			params.From = faddr
 		}
 
+		if params.From == address.Undef {
```
rvagg, line 136:
```suggestion
		} else {
```

Perhaps there's a way that one of the above options can produce an `address.Undef`, if that's the case then we should probably error in those branches. Having this free-floating feels like it's going to catch a possible error case that should be reported to the user and we pick a default that may not be intended.

rvagg:
Seems pretty straightforward; in terms of enabling what we want to work properly at least.

You followed up with each of those sends and confirmed that the recipient received it? Is this something we could capture in a simple itest, not necessarily using the CLI but using the send logic for the various cases.

cli/send.go
aarshkshah1992: Fixed.

aarshkshah1992:
@rvagg 

- Fixed test and accepted your change
- Yes for each of those cases I verified that the recipient did indeed get the funds.
- Writing automated tests for each of those cases will probably have to use itests that construct the same message that Lotus send constructs here and then submitting them to the mpool and asserting balances. But that is bigger than the change I'm making here.

rvagg (approved):

cli/send.go
```diff
@@ -128,7 +147,7 @@ var SendCmd = &cli.Command{
 			params.Params = decparams
 		}
 
-		if ethtypes.IsEthAddress(params.From) {
+		if ethtypes.IsEthAddress(params.From) || ethtypes.IsEthAddress(params.To) {
```
Stebalien, line 150: I think this is going to cause some issues.

1. There are almost certainly users using native accounts to send messages to EVM smart contracts, explicitly specifying the method number.
2. Implicitly calling `InvokeContract` is the _nice_ thing to do, but that's a significant change of behavior.

Ideally this would have this method number handling if and only if the user passes a literal `0x...` address in the to field. If they pass a normal Filecoin address, they should get the normal Filecoin method handling.

rvagg: pulled this up into a new issue https://github.com/filecoin-project/lotus/issues/12326

PR #12323, opened 2024-07-30
feat: eth: FilecoinAddressToEthAddress can take a complex argument

rvagg:
This is just a demonstration of what we could do with the existing API to extend it to have additional parameters to allow for more complex specification. If you supply something other than an f410 or f0 then you need to supply it with a `"block"` argument too which tells the node what tipset to use to look the address up. Optionally we could just accept all the fX address types and default to something, `"safe"` (probably 30 behind head), but this also allows you to specify `"finalized"` or some specific height, using the standard eth block number method of supplying it as a hex string.

Thread in slack @ https://filecoinproject.slack.com/archives/CP50PPW2X/p1722328160720429

### DIFF
--- a/build/openrpc/full.json
+++ b/build/openrpc/full.json
@@ -6922,19 +6922,28 @@
         },
         {
             "name": "Filecoin.FilecoinAddressToEthAddress",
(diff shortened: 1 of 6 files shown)


build/openrpc/full.json
```diff
@@ -6922,19 +6922,28 @@
...
                     "name": "p1",
-                    "description": "address.Address",
+                    "description": "ethtypes.StateAddressSpec",
                     "summary": "",
                     "schema": {
                         "examples": [
                             "f01234"
                         ],
                         "additionalProperties": false,
+                        "properties": {
```
rvagg, line 6938: this is my biggest concern here, we'd need a way to rectify the openrpc spec for this gnarly beast; but I think we need to be messing with this for some of the correctness work we have on our plate anyway

Issue #12308, opened 2024-07-25
FilecoinAddressToEthAddress RPC API call does not support f1/f2/f3


eshon:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Latest via Glif Nodes RPC API
```

### Repro Steps

1. Run '...'
2. Do '...'
3. See error '...'
...

### Describe the Bug

Various EVM dapps like Drips are intergrating which might need to be able to translate f1/f2/f3 into 0xFF000... equivalents.

It would be good for the Lotus RPC API to support this and it seems like an efficient lookup. It does not have the same lookup costs as https://github.com/filecoin-project/lotus/issues/12184.

### Logging Information

```text
Response for any f1/f2/f3 is "invalid Filecoin Eth address"
```

jennijuju:
note there will be a finality/deterministic concern with whats proposed. maybe should only take f0 from the user and user is responsible for giving the right & confirmed f0

eshon:
@jennijuju - isn't Fast Finality coming soon?

jennijuju:
Yah then we should really consider add “safe, finalized, future” options and checks.

aarshkshah1992:
@jennijuju already hit the nail on the head. The biggest concern here is finality. If the "f0" address assigned to the "f1/f2/f3" address gets re-orged and users end up sending funds to the "0xff..." ETH address which is derived from the "f0" address -> users can lose funds.

[...]

PR #10027, opened 2023-01-16
feat: rpc: correct `eth_subscribe` implementation

magik6k:
## Related Issues
Depends on https://github.com/filecoin-project/go-jsonrpc/pull/88
Closes https://github.com/filecoin-project/lotus/issues/9985
Closes https://github.com/filecoin-project/ref-fvm/issues/1573

## Proposed Changes
* Change eth_subscribe impl to use reverse calls (really reverse notifs)
* Implement eth_subscribe on gateway

## Additional Info
Not tested at all; Would be great to at least test lotus+lotus-gateway setup, and look at what is sent in wireshark to make sure everything looks as it should. Adding an itest would be nice to

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, market, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [x] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12324, opened 2024-07-30
feat: ETH API: Add V1 API to convert Filecoin Address to ETH address

aarshkshah1992:
Closes https://github.com/filecoin-project/lotus/issues/12308.

Introduces a new `FilecoinAddressToEthAddressV1` API that allows all Filecoin address types (f0/1/2/3/4) to ETH addresses based on the user's reorg tolerance.

Deprecates the existing `FilecoinAddressToEthAddress` API that only supports `f0`/`f4` addresses in favour of `FilecoinAddressToEthAddressV1`.


node/impl/full/eth.go
```diff
@@ -207,6 +208,56 @@ func (a *EthAPI) FilecoinAddressToEthAddress(ctx context.Context, filecoinAddres
...
+	case address.SECP256K1, address.Actor, address.BLS:
+		// Valid protocols
+	default:
+		// Ideally, this should never happen but is here for sanity checking.
+		return ethtypes.EthAddress{}, xerrors.Errorf("invalid filecoin address protocol: %s", filecoinAddress.String())
+	}
+
+	// Set default block parameter to "finalized" if not provided
+	if blkParam == "" || len(blkParam) == 0 {
+		blkParam = "finalized"
```
aarshkshah1992, line 230: @rvagg Going with "finalized" as the default here as we want the default to be the safest option here. 

The other thing we can do here is to get rid of the default and fail the request if client hasn't specified a `blkParam` for a f1/f2/f3 address.

itests/eth_api_test.go
```diff
@@ -92,6 +98,122 @@ func TestFilecoinAddressToEthAddress(t *testing.T) {
...
+
+	idAddr, err = client.StateLookupID(ctx, multisigAddress, types.EmptyTSK)
+	require.NoError(t, err)
+
+	ethAddr, err = ethtypes.EthAddressFromFilecoinAddress(idAddr)
+	require.NoError(t, err)
+	require.Equal(t, ethAddr, apiEthAddr)
+}
+
+func TestFilecoinAddressToEthAddressV1Finalised(t *testing.T) {
```
aarshkshah1992, line 181: @rvagg This test tests for the "finalized" blkparam for an f1 address.

node/impl/full/eth.go
rvagg: finalized seems fine to me, and _should_ be awesome post-f3.

node/impl/full/eth.go
```diff
@@ -207,6 +208,56 @@ func (a *EthAPI) FilecoinAddressToEthAddress(ctx context.Context, filecoinAddres
...
+	ts, err := getTipsetByBlockNumber(ctx, a.Chain, blkParam, true)
+	if err != nil {
+		return ethtypes.EthAddress{}, xerrors.Errorf("failed to get tipset for block %s: %w", blkParam, err)
+	}
+
+	// Lookup the ID address
+	idAddr, err := a.StateManager.LookupIDAddress(ctx, filecoinAddress, ts)
+	if err != nil {
+		return ethtypes.EthAddress{}, xerrors.Errorf(
+			"failed to lookup ID(f0) address for given Filecoin address %s: "+
```
rvagg, line 243: it's a `t0` on calibnet, although I'm not sure of the best logic to swap that out here dynamically but I think you can call out to `address.CurrentNetwork`, check whether it's `address.Testnet` or `address.Mainnet` then use either `address.TestnetPrefix` or `address.MainnetPrefix` (although I bet the best way is to do something like `address.NewIDAddress(0)`, `.String()` it and cut off the first 2 chars).

rvagg:
Implementation seems fine to me, but we can't call this `FilecoinAddressToEthAddressV1`, that's not going to work. For a start it's the second iteration of this API, not the first, but also this is on the /v1/ RPC API, so that's just weird. Making it `V2` hanging off a /v1/ is going to be extra weird.

We have a couple of problems to wrestle with:

1. We probably should stop evolving the /v1/ API as @jennijuju mentioned on Slack, but I'm not sure that helps us here because do we really want to make the start of a /v2/ a prerequisite for getting this done?
2. We don't want to break any /v1/ APIs
3. We can't add optional arguments to these calls, it's not how the RPC system (or at least our go-jsonrpc) works

I proposed two alternative solutions:

* Use the name `EthAddressFromFilecoinAddress`, which mirrors the existing `EthAddressToFilecoinAddress`, perhaps we then drop `FilecoinAddressToEthAddress` from the /v2/ API. Unfortunately this would be adding another method to /v1/ but I don't see a way around that atm.
* Make the existing `FilecoinAddressToEthAddress` handle this, but do it in a backward compatible way by hijacking the JSON encoding, which I proposed in: https://github.com/filecoin-project/lotus/pull/12323

I'd like to hear @Stebalien's thoughts on this though.

rvagg:
@aarshkshah1992 I just stumbled upon this, after tinkering on the gateway: https://github.com/filecoin-project/ref-fvm/issues/1573

For `eth_subscribe` the API has an optional parameter: when you say `"newHeads"` it's just 1 param, but when you say `"logs"` you supply a second optional parameter: https://docs.infura.io/api/networks/ethereum/json-rpc-methods/subscription-methods/eth_subscribe

It was implemented in https://github.com/filecoin-project/lotus/pull/10027.

That PR has a lot of noise to support the subscribe data flow so it needs a whole reverse implementation for the client. But the relevant bit is the replacement of the method's params with `params jsonrpc.RawParams`. If the method signature is `(ctx context.Context, params jsonrpc.RawParams)` then go-jsonrpc will just give you the `json.RawMessage` to decode for yourself.

There are 3 methods that do this in the codebase so far:
* `EthFeeHistory`: https://github.com/filecoin-project/lotus/blob/f6978f01725fc8f8ef72cdf83d15aa57b8e076db/node/impl/full/eth.go#L695
* `EthEstimateGas`: https://github.com/filecoin-project/lotus/blob/f6978f01725fc8f8ef72cdf83d15aa57b8e076db/node/impl/full/eth.go#L1244
* `EthSubscribe`: https://github.com/filecoin-project/lotus/blob/f6978f01725fc8f8ef72cdf83d15aa57b8e076db/node/impl/full/eth.go#L1875

All with the same pattern.

So, I think you can just do this for `FilecoinAddressToEthAddress` and add your optional parameter on the end and we have a non-breaking change.

aarshkshah1992:
@rvagg 

Also tested this on a devnet node

```
curl -X POST -H "Content-Type: application/json" --data '{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "Filecoin.FilecoinAddressToEthAddress",
  "params": ["t410fghja63ghyzd54cbswfapglsutx5e64xnm7mazwi"]
}' http://localhost:1234/rpc/v1
{"id":1,"jsonrpc":"2.0","result":"0x31d20f6cc7c647de0832b140f32e549dfa4f72ed"}
```

```
curl -X POST -H "Content-Type: application/json" --data '{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "Filecoin.FilecoinAddressToEthAddress",
  "params": ["t3thonru4djgcwi2vuuymzqkbymiv6tixs5igehh3ht46tjgqyjifkuz2sbziovsgshr2vttjh34q6bzavelea","latest"]
}' http://localhost:1234/rpc/v1
{"id":1,"jsonrpc":"2.0","result":"0xff00000000000000000000000000000000000064"}
```

api/api_full.go
```diff
@@ -726,8 +726,33 @@ type FullNode interface {
...
+	// "f0" ID addresses are not permanent and can be affected by chain reorganizations. To account for this,
+	// the API includes a `blkNum` parameter, which specifies the block number that is used to determine the tipset state to use for converting an
+	// "f1"/"f2"/"f3" address to an "f0" address. This parameter functions similarly to the `blkNum` parameter in the existing `EthGetBlockByNumber` API.
+	// See https://docs.alchemy.com/reference/eth-getblockbynumber for more details.
+	//
+	// Parameters:
+	// - ctx: The context for the API call.
+	// - filecoinAddress: The Filecoin address to convert.
+	// - blkNum: The block number or state for the conversion. Defaults to "finalized" for maximum safety.
+	//   Possible values: "pending", "latest", "finalized", "safe", or a specific block number.
```
rvagg, line 749:
```suggestion
	//   Possible values: "pending", "latest", "finalized", "safe", or a specific block number represented as hex.
```

api/api_full.go
```diff
@@ -726,8 +726,33 @@ type FullNode interface {
 	EthAccounts(ctx context.Context) ([]ethtypes.EthAddress, error) //perm:read
 	// EthAddressToFilecoinAddress converts an EthAddress into an f410 Filecoin Address
 	EthAddressToFilecoinAddress(ctx context.Context, ethAddress ethtypes.EthAddress) (address.Address, error) //perm:read
-	// FilecoinAddressToEthAddress converts an f410 or f0 Filecoin Address to an EthAddress
-	FilecoinAddressToEthAddress(ctx context.Context, filecoinAddress address.Address) (ethtypes.EthAddress, error) //perm:read
+
+	// `FilecoinAddressToEthAddress` is the recommended API for converting any Filecoin address to an EthAddress.
```
rvagg, line 730:
```suggestion
	// `FilecoinAddressToEthAddress` converts any Filecoin address to an EthAddress.
```

api/api_full.go
```diff
@@ -726,8 +726,33 @@ type FullNode interface {
...
+	//
+	// This method supports all Filecoin address types:
+	// - "f0" and "f4" addresses: Converted directly.
+	// - "f1", "f2", and "f3" addresses: First converted to their corresponding "f0" ID address, then to an EthAddress.
+	//
+	// Requirements:
+	// - For "f1", "f2", and "f3" addresses, they must be instantiated on-chain, as "f0" ID addresses are only assigned to actors when they are created on-chain.
```
rvagg, line 737: I wonder if we should go into detail about this? I remember being confused about this the first time I encountered it; it's perfectly logical to me now but for the uninitiated they may need to know a bit more about what it means to be on chain

build/openrpc/full.json
```diff
@@ -6922,21 +6922,29 @@
...
+                                "description": "Number is a number",
+                                "type": [
+                                    "number"
+                                ]
+                            }
                         ],
-                        "additionalProperties": false,
                         "type": [
-                            "object"
+                            "array"
```
rvagg, line 6947: we're going to need to do something about these specs for correctness

chain/types/ethtypes/eth_types.go
```diff
@@ -826,6 +826,44 @@ func GetContractEthAddressFromCode(sender EthAddress, salt [32]byte, initcode []
 	return ethAddr, nil
 }
 
+type FilecoinAddressToEthAddressParams struct {
+	FilecoinAddress address.Address
+	BlkParam        *string
```
rvagg, line 831:
```suggestion
	FilecoinAddress address.Address `json:"address"`
	Block        *string `json:"block,omitempty"`
```

Are we following a pattern with `BlkParam`? seems odd to name it

this `json` tagging isn't really critical since this is internal, just nice to be consistent

rvagg: actually forget this, it's never touched

chain/types/ethtypes/eth_types.go
```diff
@@ -826,6 +826,44 @@ func GetContractEthAddressFromCode(sender EthAddress, salt [32]byte, initcode []
...
+		}
+	default:
+		return xerrors.Errorf("expected 1 or 2 params, got %d", len(params))
+	}
+
+	return nil
+}
+
+func (e *FilecoinAddressToEthAddressParams) MarshalJSON() ([]byte, error) {
+	if e.BlkParam != nil {
```
rvagg, line 861:
```suggestion
	if e.BlkParam != nil && e.BlkParam != "" {
```

itests/eth_api_test.go
```diff
@@ -52,44 +60,147 @@ func TestEthAddressToFilecoinAddress(t *testing.T) {
...
-	require.NoError(t, err)
+		return jsonrpc.RawParams(params)
+	}
 
+	// Test for f4 address
+	filecoinDelegatedAddr, err := client.WalletNew(ctx, types.KTDelegated)
+	require.NoError(t, err)
+	ethAddr, err := ethtypes.EthAddressFromFilecoinAddress(filecoinDelegatedAddr)
+	require.NoError(t, err)
+	apiEthAddr, err := client.FilecoinAddressToEthAddress(ctx, mkParams(filecoinDelegatedAddr, ""))
```
rvagg, line 89: if you look at the other uses of `RawParams` in itests there's a use of lib/result/result.Wrap that makes this a bit easier and avoids the extra helper function:

```go
	history, err := client.EthFeeHistory(ctx, result.Wrap[jsonrpc.RawParams](
		json.Marshal([]interface{}{5, "0x10"}),
	).Assert(require.NoError))
```

node/impl/full/eth.go
```diff
@@ -203,8 +204,63 @@ func (a *EthAPI) EthAddressToFilecoinAddress(ctx context.Context, ethAddress eth
...
+		return ethtypes.EthAddress{}, xerrors.Errorf("failed to get tipset for block %s: %w", blkParam, err)
+	}
+
+	// Lookup the ID address
+	idAddr, err := a.StateManager.LookupIDAddress(ctx, filecoinAddress, ts)
+	if err != nil {
+		return ethtypes.EthAddress{}, xerrors.Errorf(
+			"failed to lookup ID address for given Filecoin address %s: "+
+				"ensure that the address has been instantiated on-chain and sufficient epochs have passed since instantiation to confirm to the given 'blkParam' %s. "+
+				"Error: %w",
```
rvagg, line 250:
```suggestion
			"failed to lookup ID address for given Filecoin address %s ("+
				"ensure that the address has been instantiated on-chain and sufficient epochs have passed since instantiation to confirm to the given 'blkParam': \"%s\"): %w",
```

Mainly making this not require a full-stop and capital `"Error"`.

Although, I also think you should pull that entire bracket part out and make it conditional, and you could have 3 cases for it:

1. they have a `blkParam` that's not hex, so is one of the text ones, in which case this message is 👌 
2. they have a `blkParam` but it's a hex, in which case you could have an alternate message about how it's not found at that block
3. they have no `blkParam`, in which case they get told to wait for `"finalized"` but they never said such a thing, you could reword it to inform about the default

itests/eth_api_test.go
aarshkshah1992: Done.

aarshkshah1992:
@rvagg Addressed your review. Please can you 🟢  ?

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@
 
 - feat: Add trace filter API supporting RPC method `trace_filter` ([filecoin-project/lotus#12123](https://github.com/filecoin-project/lotus/pull/12123)). Configuring `EthTraceFilterMaxResults` sets a limit on how many results are returned in any individual `trace_filter` RPC API call.
 
+- feat: `FilecoinAddressToEthAddress` RPC can now return ETH addresses for all Filecoin address types ("f0"/"f1"/"f2"/"f3") based on client's re-org tolerance.
```
rvagg, line 22:
```suggestion
- feat: `FilecoinAddressToEthAddress` RPC can now return ETH addresses for all Filecoin address types ("f0"/"f1"/"f2"/"f3") based on client's re-org tolerance. This is a breaking change if you are using the API via the go-jsonrpc library or by using Lotus as a library, but is a non-breaking change when using the API via any other RPC method as it adds an optional second argument.
```

rvagg (approved):
I'm a tiny bit hesitant because it's quite disruptive when you see how you have to deal with the API in Go via Lotus directly, but I'm going to make the assumption that the majority of callers of this method are doing it via plain RPC or some other RPC wrapper.

aarshkshah1992:
@rvagg IIUC, most clients do indeed consume this via RPC so I think the cost is worth it as it allows to get away without creating a new API and confusing RPC users.

Issue #11153, opened 2023-08-09
Reject Eth subscriptions & filters through the gateway over HTTP


Stebalien:
Currently, the lotus gateway forgets all eth subscriptions and event filters between HTTP calls. These APIs only work over the websocket transport.

This happens because:

1. We track filter and subscription IDs in the top-level HTTP request's context.
2. Even if we didn't do that, we couldn't guarantee that all HTTP requests hit the same API server (e.g., given a load balancer), so the websocket transport is the only reliable way to use these APIs anyways.

See https://github.com/filecoin-project/lotus/blob/9310cd6036d9318af4c12d1b1954ef4347912544/gateway/handler.go#L89-L96

**Proposal:** In the relevant Eth API methods (all those either taking or returning an `EthFilterID` and/or an `EthSubscriptionID`), check if the request is being made via a websocket and, if not, reject it with an error explaining that the API is only supported over the websocket transport.

**Motivation:** Users aren't confused because this just silently refuses to work. Instead, users will get a nice error message telling them to use a websocket if they need to use this API.

Stebalien:
Fixed in https://github.com/filecoin-project/lotus/pull/12327.

PR #12327, opened 2024-07-31
feat!: gateway: fix rate limiting, better stateful handling

rvagg:
Ref: https://github.com/filecoin-project/go-jsonrpc/pull/118
Ref: https://github.com/filecoin-project/lotus/issues/11589
Ref: https://github.com/filecoin-project/lotus/issues/11153

---

API changes and breakage:

* Plain HTTP connections now can't use stateful ethereum APIs (they couldn't really before, but they allowed addition of filters and subs in a dangerous way), you have to be using WebSockets
* `gateway.NewNode` now only takes 1 argument (the api) and a list of options to override defaults
* `gateway.NewRateLimiterHandler` and `gateway.NewConnectionRateLimiterHandler` have been replaced with `gateway.NewRateLimitHandler`.
* The handlers returned by both `gateway.NewRateLimitHandler` and the primary `gateway.Handler` return an `http.Handler` augmented with a `Shutdown(ctx)` method to be used for graceful cleanup of resources.
* The default for the config option `Events.FilterTTL` has been reduced from 24h to 1h. It's a measure of "how long since this thing was last used", and 24h is way too long for that IMO.

Fix:

* `--per-conn-rate-limit` was previously applied as a global rate limiter,
  effectively making it have the same impact as `--rate-limit`. This change fixes
  the behaviour such that `--per-conn-rate-limit` is applied as a API call
  limiter within a single connection (i.e. a WebSocket connection). The rate
  is specified as tokens-per-second, where tokens are relative to the expense
  of the API call being made.

Improve:

* Auto-clean-up of stateful resources when a WebSockets connection ends, so we close down subscriptions and remove filters if the user didn't do so themselves.
* Added `--eth-max-filters-per-conn` to override the default (16) and it now applies to the combined total of eth subs and filters, not separately
* Replace the `--conn-per-minute` limit mechanism to a standard `rate.Limiter` instead of the custom rate limit implementation. There is a slight change of behaviour with this in that I allow a burst, which the original doesn't. It also has a different cleanup mechanism with a separate goroutine dedicated to this task instead of a goroutine per request that decrements and may cleanup if zero.

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -61,6 +61,7 @@ func Handler(
 	m := mux.NewRouter()
 
 	opts = append(opts, jsonrpc.WithReverseClient[lapi.EthSubscriberMethods]("Filecoin"), jsonrpc.WithServerErrors(lapi.RPCErrors))
+
 	serveRpc := func(path string, hnd interface{}) {
 		rpcServer := jsonrpc.NewServer(opts...)
 		rpcServer.Register("Filecoin", hnd)
@@ -106,7 +107,11 @@ type statefulCallHandler struct {
 }
 
 func (h statefulCallHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	r = r.WithContext(context.WithValue(r.Context(), statefulCallTrackerKey, newStatefulCallTracker()))
+	tracker := newStatefulCallTracker()
+	defer func() {
+		go tracker.cleanup()
+	}()
+	r = r.WithContext(context.WithValue(r.Context(), statefulCallTrackerKey, tracker))
 	h.next.ServeHTTP(w, r)
 }
 
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -21,20 +21,48 @@ import (
 	"github.com/filecoin-project/lotus/node"
 )
 
-type perConnLimiterKeyType string
-
-const perConnLimiterKey perConnLimiterKeyType = "limiter"
-
+type perConnectionAPIRateLimiterKeyType string
 type filterTrackerKeyType string
 
-const statefulCallTrackerKey filterTrackerKeyType = "statefulCallTracker"
+const (
+	perConnectionAPIRateLimiterKey   perConnectionAPIRateLimiterKeyType = "limiter"
+	statefulCallTrackerKey           filterTrackerKeyType               = "statefulCallTracker"
+	connectionLimiterCleanupInterval                                    = 30 * time.Second
+)
 
-// Handler returns a gateway http.Handler, to be mounted as-is on the server.
-func Handler(gwapi lapi.Gateway, api lapi.FullNode, rateLimit int64, connPerMinute int64, opts ...jsonrpc.ServerOption) (http.Handler, error) {
+// ShutdownHandler is an http.Handler that can be gracefully shutdown.
+type ShutdownHandler interface {
+	http.Handler
+
+	Shutdown(ctx context.Context) error
+}
+
+var _ ShutdownHandler = &statefulCallHandler{}
+var _ ShutdownHandler = &RateLimitHandler{}
+
+// Handler returns a gateway http.Handler, to be mounted as-is on the server. The handler is
+// returned as a ShutdownHandler which allows for graceful shutdown of the handler via its
+// Shutdown method.
+//
+// The handler will limit the number of API calls per minute within a single WebSocket connection
+// (where API calls are weighted by their relative expense), and the number of connections per
+// minute from a single host.
+//
+// Connection limiting is a hard limit that will reject requests with a 429 status code if the limit
+// is exceeded. API call limiting is a soft limit that will delay requests if the limit is exceeded.
+func Handler(
+	gwapi lapi.Gateway,
+	api lapi.FullNode,
+	perConnectionAPIRateLimit int,
+	perHostConnectionsPerMinute int,
+	opts ...jsonrpc.ServerOption,
+) (ShutdownHandler, error) {
+
 	m := mux.NewRouter()
 
+	opts = append(opts, jsonrpc.WithReverseClient[lapi.EthSubscriberMethods]("Filecoin"), jsonrpc.WithServerErrors(lapi.RPCErrors))
 	serveRpc := func(path string, hnd interface{}) {
-		rpcServer := jsonrpc.NewServer(append(opts, jsonrpc.WithReverseClient[lapi.EthSubscriberMethods]("Filecoin"), jsonrpc.WithServerErrors(lapi.RPCErrors))...)
+		rpcServer := jsonrpc.NewServer(opts...)
 		rpcServer.Register("Filecoin", hnd)
 		rpcServer.AliasMethod("rpc.discover", "Filecoin.Discover")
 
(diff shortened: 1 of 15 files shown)


itests/gateway_test.go
```diff
@@ -185,52 +190,72 @@ func TestGatewayMsigCLI(t *testing.T) {
...
-func startNodesWithFunds(
-	ctx context.Context,
-	t *testing.T,
-	blocktime time.Duration,
-	lookbackCap time.Duration,
-	stateWaitLookbackLimit abi.ChainEpoch,
-) *testNodes {
-	nodes := startNodes(ctx, t, blocktime, lookbackCap, stateWaitLookbackLimit)
+type startOptions struct {
+	blocktime                 time.Duration
```
masih, line 209:
```suggestion
	blockTime                 time.Duration
```

itests/gateway_test.go
```diff
@@ -260,19 +293,48 @@ func startNodes(
 
 	// Create a gateway client API that connects to the gateway server
 	var gapi api.Gateway
-	gapi, closer, err = client.NewGatewayRPCV1(ctx, "ws://"+srv.Listener.Addr().String()+"/rpc/v1", nil)
+	var _closer jsonrpc.ClientCloser
```
masih, line 296: Subjective: The `_` in name conventionally makes me think this should not be set explicitly (a style used in named return types). I would rename this to something else if `_` in name is there just to avoid clashing with the wrapper.

itests/gateway_test.go
```diff
@@ -185,52 +190,72 @@ func TestGatewayMsigCLI(t *testing.T) {
...
 
-	// The full node starts with a wallet
-	fullWalletAddr, err := nodes.full.WalletDefaultAddress(ctx)
-	require.NoError(t, err)
+type startOption func(*startOptions)
 
-	// Get the lite node default wallet address.
-	liteWalletAddr, err := nodes.lite.WalletDefaultAddress(ctx)
-	require.NoError(t, err)
+func applyStartOptions(opts ...startOption) startOptions {
```
masih, line 220: Subjective: I would name this `newStartOptions` because that's what it does. "apply"ing is an implementation detail that happens as part of construction.

itests/gateway_test.go
```diff
@@ -297,3 +359,209 @@ func sendFunds(ctx context.Context, fromNode *kit.TestFullNode, fromAddr address
...
+	ctx := context.Background()
+	tokensPerSecond := 10
+	requestsPerMinute := 30 // http requests
+	nodes := startNodes(ctx, t,
+		withNodeOpts(kit.DisableEthRPC()),
+		withPerConnectionAPIRateLimit(tokensPerSecond),
+		withPerHostRequestsPerMinute(requestsPerMinute),
+	)
+
+	time.Sleep(time.Second)
```
masih, line 376: Why? Can we avoid this with `require.Eventually`?

itests/gateway_test.go
```diff
@@ -297,3 +359,209 @@ func sendFunds(ctx context.Context, fromNode *kit.TestFullNode, fromAddr address
...
+				req.NotNil(result["result"])
+				height, ok := result["result"].(map[string]interface{})["Height"].(float64)
+				req.True(ok)
+				req.Greater(int(height), 0)
+			} else {
+				req.Equal(http.StatusTooManyRequests, response.StatusCode)
+				req.LessOrEqual(i, requestsPerMinute+1)
+				failed = true
+			}
+		}()
```
masih, line 422: Subjective: Refactoring this block into a set of functions would make it easier to read.

itests/gateway_test.go
```diff
@@ -297,3 +359,209 @@ func sendFunds(ctx context.Context, fromNode *kit.TestFullNode, fromAddr address
...
+	allowPad := time.Duration(float64(tokensPerLoop) / float64(tokensPerSecond) * float64(time.Second)) // add padding to account for slow test runs
+	t.Logf("expected end: %s, now: %s, allowPad: %s, actual delta: %s", expectedEnd, time.Now(), allowPad, time.Since(expectedEnd))
+	req.WithinDuration(expectedEnd, time.Now(), allowPad)
+
+	client := &http.Client{}
+	url := fmt.Sprintf("http://%s/rpc/v1", nodes.gatewayAddr)
+	jsonPayload := []byte(`{"method":"Filecoin.ChainHead","params":[],"id":1,"jsonrpc":"2.0"}`)
+	var failed bool
+	for i := 0; i < requestsPerMinute*2 && !failed; i++ {
+		func() {
```
masih, line 400: I am curious, why the function wrapping?

itests/gateway_test.go
```diff
@@ -297,3 +359,209 @@ func sendFunds(ctx context.Context, fromNode *kit.TestFullNode, fromAddr address
...
+	_, err = nodes.lite.EthNewPendingTransactionFilter(ctx)
+	require.ErrorContains(t, err, "too many subscriptions and filters for this connection")
+	_, err = nodes.lite.EthNewFilter(ctx, &ethtypes.EthFilterSpec{})
+	require.ErrorContains(t, err, "too many subscriptions and filters for this connection")
+	_, err = nodes.lite.EthSubscribe(ctx, res.Wrap[jsonrpc.RawParams](json.Marshal(ethtypes.EthSubscribeParams{EventType: "newHeads"})).Assert(req.NoError))
+	require.ErrorContains(t, err, "too many subscriptions and filters for this connection")
+
+	t.Logf("Shutting down the lite node")
+	req.NoError(nodes.lite.Stop(ctx))
+	nodes.rpcCloser() // once the websocke connection is closed, the server should clean up the filters for us
```
masih, line 545:
```suggestion
	nodes.rpcCloser() // once the websocket connection is closed, the server should clean up the filters for us.
```

itests/gateway_test.go
```diff
@@ -297,3 +359,209 @@ func sendFunds(ctx context.Context, fromNode *kit.TestFullNode, fromAddr address
...
+
+	time.Sleep(time.Second) // unfortunately we have no other way to check for completeness of shutdown and cleanup
```
masih, line 547: Can you explain how the sleep here assures completeness of shutdown and cleanup?

itests/gateway_test.go
```diff
@@ -297,3 +359,209 @@ func sendFunds(ctx context.Context, fromNode *kit.TestFullNode, fromAddr address
...
+
+	t.Logf("Checking that all filters and subs were cleared up by directly talking to full node")
+
+	ok, err = nodes.full.EthUnsubscribe(ctx, subId2) // unsub on full node, already done
+	req.NoError(err)
+	req.False(ok) // already unsubbed because of auto-cleanup
```
masih, line 553: unsubbed as in unsubscribed?

masih (commented):
Two potential blockers:

* DoS attack vector due to unbounded limiters per host.
* Repeated timer instantiation in loop instead of reusing ticker.

Other than that, I left a bunch of comments/suggestions. 

Thank you for working on this @rvagg 🍻

gateway/handler.go
```diff
@@ -21,20 +21,49 @@ import (
...
+// minute from a single host.
+//
+// Connection limiting is a hard limit that will reject requests with a 429 status code if the limit
+// is exceeded. API call limiting is a soft limit that will delay requests if the limit is exceeded.
+func Handler(
+	gwapi lapi.Gateway,
+	api lapi.FullNode,
+	perConnectionAPIRateLimit int,
+	perHostConnectionsPerMinute int,
+	opts ...jsonrpc.ServerOption,
```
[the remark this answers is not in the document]
rvagg: fair enough, may as well do it in here since we're breaking anyway

gateway/handler.go
```diff
@@ -61,104 +90,156 @@ func Handler(gwapi lapi.Gateway, api lapi.FullNode, rateLimit int64, connPerMinu
...
-	/*ah := &auth.Handler{
-		Verify: nodeApi.AuthVerify,
-		Next:   mux.ServeHTTP,
-	}*/
-
-	rlh := NewRateLimiterHandler(m, rateLimit)
-	clh := NewConnectionRateLimiterHandler(rlh, connPerMinute)
-	return clh, nil
+	handler := &statefulCallHandler{m}
+	if perConnectionAPIRateLimit > 0 && perHostConnectionsPerMinute > 0 {
```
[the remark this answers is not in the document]
rvagg: actually this is wrong, should be `||`, either can be >0 to take effect

gateway/proxy_eth.go
```diff
@@ -104,7 +106,7 @@ func (gw *Node) checkEthBlockParam(ctx context.Context, blkParam ethtypes.EthBlo
 		var num ethtypes.EthUint64
 		if blkParam.PredefinedBlock != nil {
 			if *blkParam.PredefinedBlock == "earliest" {
-				return fmt.Errorf("block param \"earliest\" is not supported")
+				return xerrors.New("block param \"earliest\" is not supported")
```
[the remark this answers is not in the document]
rvagg: I'm going to leave this one alone in this PR because the same pattern is repeated elsewhere in the code that I haven't touched

gateway/proxy_eth.go
```diff
@@ -592,6 +607,7 @@ func (gw *Node) EthUnsubscribe(ctx context.Context, id ethtypes.EthSubscriptionI
 
 	ok, err := gw.target.EthUnsubscribe(ctx, id)
 	if err != nil {
+		// don't delete the subscription, it's "stuck" so should still count towards the limit
```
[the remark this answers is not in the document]
rvagg: chasing these calls down into eth.go, I think the only case here is that of "filter not found"; and in the case of EthSubscribe specifically, there's not really even a case where we get that

BUT there is one important caveat - the backend node is different to the one where the filter was originally installed, either because the node was restarted (in which case the error shouldn't count toward the limit in the gateway for the user), or because we have a problem with sticky sessions for this user - some reverse proxy is involved that has shunted this request to the wrong gateway or from the gateway to the wrong node (that latter one is a bit weird, but not impossible with a funky setup).

In the case of non-sticky sessions, then the gateway operator might be in a bit of trouble. Now we have a logger in here maybe it'd be worth printing a warning here.

itests/gateway_test.go
rvagg: tighter handling of deferred body close, I don't really want to leave multiple dangling till the very end and deferred is too nice a feature to not use; got a better suggestion?

gateway/handler.go
```diff
@@ -61,104 +90,156 @@ func Handler(gwapi lapi.Gateway, api lapi.FullNode, rateLimit int64, connPerMinu
...
-	connPerMinute int64
-	handler       http.Handler
+func (h *RateLimitHandler) getLimits(host string) *hostLimiter {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	entry, exists := h.limiters[host]
+	if !exists {
+		var limiter *rate.Limiter
+		if h.perHostConnectionsPerMinute > 0 {
```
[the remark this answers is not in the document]
rvagg: you're right, this function is a hold-over from a combination of the previous functionality and a form this PR took before this final form and now it's just weird, I've removed it

gateway/handler.go
```diff
@@ -61,104 +90,156 @@ func Handler(gwapi lapi.Gateway, api lapi.FullNode, rateLimit int64, connPerMinu
...
+	go h.cleanupExpiredLimiters(ctx)
+	return h
 }
 
-type ConnectionRateLimiterHandler struct {
-	mu            sync.Mutex
-	ipmap         map[string]int64
-	connPerMinute int64
-	handler       http.Handler
+func (h *RateLimitHandler) getLimits(host string) *hostLimiter {
```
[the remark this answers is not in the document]
rvagg: removed, inlined the code

gateway/handler.go
```diff
@@ -61,104 +90,156 @@ func Handler(gwapi lapi.Gateway, api lapi.FullNode, rateLimit int64, connPerMinu
...
-	r = r.WithContext(context.WithValue(r.Context(), statefulCallTrackerKey, newStatefulCallTracker()))
+type hostLimiter struct {
+	limiter    *rate.Limiter
+	lastAccess time.Time
+}
 
-	h.handler.ServeHTTP(w, r)
+type RateLimitHandler struct {
+	cancelFunc                  context.CancelFunc
+	mu                          sync.Mutex
```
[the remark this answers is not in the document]
rvagg: yeah, good point, renamed to `limitersLk`

gateway/node.go
```diff
@@ -31,14 +32,20 @@ import (
 	"github.com/filecoin-project/lotus/node/modules/dtypes"
 )
 
+var log = logger.Logger("gateway")
+
 const (
 	DefaultLookbackCap            = time.Hour * 24
 	DefaultStateWaitLookbackLimit = abi.ChainEpoch(20)
 	DefaultRateLimitTimeout       = time.Second * 5
+	DefaultEthMaxFiltersPerConn   = 16
```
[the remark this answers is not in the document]
rvagg: documented and renamed the other options cause they are poorly named:

* `DefaultMaxLookbackDuration`
* `DefaultMaxMessageLookbackEpochs`

gateway/proxy_eth.go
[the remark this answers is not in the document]
rvagg: added log

gateway/node.go
```diff
@@ -174,22 +182,88 @@ var (
...
+
 // NewNode creates a new gateway node.
-func NewNode(api TargetAPI, sHnd *EthSubHandler, lookbackCap time.Duration, stateWaitLookbackLimit abi.ChainEpoch, rateLimit int64, rateLimitTimeout time.Duration) *Node {
-	var limit rate.Limit
-	if rateLimit == 0 {
-		limit = rate.Inf
-	} else {
-		limit = rate.Every(time.Second / time.Duration(rateLimit))
+func NewNode(api TargetAPI, opts ...Option) *Node {
+	options := &Options{
```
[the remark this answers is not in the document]
rvagg: One problem with this is that the per-host limit needs a new limiter for each host, so you'd need to supply a limiter-factory to to make it consistent. It would be nice to offer the flexibility though because a limiter has more than just one parameter.
Unless you can come up with a way to make the interface consistent between `Node` and `Handler` in terms of supplying limiter params I'll keep it as it is.

itests/gateway_test.go
rvagg: we have a `WaitTillChain` that I can use here, just to make sure the chain is ticking

itests/gateway_test.go
rvagg: I've pulled out a client rpc call here that I can reuse across the two separate tests, let me know if you think I should go further

itests/gateway_test.go
rvagg: expanded on the comment in here:

```
	// Once the websocket connection is closed, the server should clean up the filters for us.
	// Unfortunately we have no other way to check for completeness of shutdown and cleanup.
	// * Asynchronously the rpcCloser() call will end the client websockets connection to the gateway.
	// * When the gateway recognises the end of the HTTP conneciton, it will asynchronously make calls
	//   to the fullnode to clean up the filters.
	// * The fullnode will then uninstall the filters and we can finally move on to check it directly
	//   that this has happened.
	// This should happen quickly, but we have no way to synchronously check for it. So we just wait a bit.
```

i.e. multiple layers of async over 2 separate HTTP connections and no direct insight to internal state of either gateway or fullnode. I also can't think of a good "eventually" that's idempotent enough to use here.

gateway/proxy_eth.go
```diff
@@ -498,7 +498,7 @@ func (gw *Node) EthNewPendingTransactionFilter(ctx context.Context) (ethtypes.Et
 		return ethtypes.EthFilterID{}, err
 	}
 
-	return addUserFilterLimited(ctx, func() (ethtypes.EthFilterID, error) {
+	return gw.addUserFilterLimited(ctx, "EthNewPendingTransactionFilter", func() (ethtypes.EthFilterID, error) {
```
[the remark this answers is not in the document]
rvagg: I'm not sure that's going to improve maintainability because it'll separate the method names from the methods themselves. This way they are at least close.

rvagg:
@masih thanks for the thorough review. I've gone through all your feedback and marked as resolved the easy bits here but left as unresolved the things that may either need follow-up or have responses to questions. Would you mind having another look please?

gateway/node.go
```diff
@@ -174,22 +182,88 @@ var (
...
+type Options struct {
+	subHandler             *EthSubHandler
+	lookbackCap            time.Duration
+	stateWaitLookbackLimit abi.ChainEpoch
+	rateLimit              int
+	rateLimitTimeout       time.Duration
+	ethMaxFiltersPerConn   int
+}
+
+type Option func(*Options)
```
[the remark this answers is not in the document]
masih: TL;DR: That approach seems fine, thanks Rod.

---
Per type options tends to end up biting me back, and I usually go with a simple Option at package level even if there are options that would be ignored by some types. That's a generalised subjective take on my part.

I see; so all these exported APIs will probably never be used as libraries. If they were then I would personally prefer adding the error now and avoid breaking code later. But for this specific case I find myself bikeshedding.

This perhaps goes into the bucket of stuff to revisit if/when we approach "fillib".

gateway/handler.go
```diff
@@ -21,20 +21,49 @@ import (
...
+
+// Handler returns a gateway http.Handler, to be mounted as-is on the server. The handler is
+// returned as a ShutdownHandler which allows for graceful shutdown of the handler via its
+// Shutdown method.
+//
+// The handler will limit the number of API calls per minute within a single WebSocket connection
+// (where API calls are weighted by their relative expense), and the number of connections per
+// minute from a single host.
+//
+// Connection limiting is a hard limit that will reject requests with a 429 status code if the limit
```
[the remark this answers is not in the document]
masih: Side note: If we use `http.StatusTooManyRequests`, it gets converted to link in godoc. I usually try to use links in docs wherever possible.

gateway/handler.go
```diff
@@ -61,104 +90,156 @@ func Handler(gwapi lapi.Gateway, api lapi.FullNode, rateLimit int64, connPerMinu
...
 
-func (h *ConnectionRateLimiterHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if h.connPerMinute == 0 {
-		h.handler.ServeHTTP(w, r)
-		return
-	}
+func (h *RateLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	host, _, err := net.SplitHostPort(r.RemoteAddr)
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
```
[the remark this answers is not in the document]
masih: Mm, you are probably right.

itests/gateway_test.go
masih: I see the code is full of time.Sleeps from priror work. Avoiding those seems like a separate body of work.

Thank you for the comment; that helps me as a reader a lot 👍

masih (approved):

gateway/handler.go
```diff
@@ -61,104 +90,156 @@ func Handler(gwapi lapi.Gateway, api lapi.FullNode, rateLimit int64, connPerMinu
...
+type hostLimiter struct {
+	limiter    *rate.Limiter
+	lastAccess time.Time
+}
 
-	h.handler.ServeHTTP(w, r)
+type RateLimitHandler struct {
+	cancelFunc                  context.CancelFunc
+	mu                          sync.Mutex
+	limiters                    map[string]*hostLimiter
```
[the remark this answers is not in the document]
rvagg: I'm not actually sure anyone uses this! You'd be better to deploy this with nginx in front of it and use that for standard rate limiting and only using lotus-gateway to deal with the complexities of lotus calls, like the API rate limiting—but even that, you probably should gate users with API keys and handle them that way.

Basically: I'm not sure any of this rate limiting stuff is used in practice anywhere or is all that useful, I'm fixing what's there in case someone is using it!

itests/gateway_test.go
```diff
@@ -297,3 +364,211 @@ func sendFunds(ctx context.Context, fromNode *kit.TestFullNode, fromAddr address
...
+	tokensPerLoop := 4
+	start := time.Now()
+	for i := 0; i < loops; i++ {
+		_, err := nodes.lite.ChainHead(ctx)
+		req.NoError(err)
+	}
+	tokensUsed := loops * tokensPerLoop
+	expectedEnd := start.Add(time.Duration(float64(tokensUsed) / float64(tokensPerSecond) * float64(time.Second)))
+	allowPad := time.Duration(float64(tokensPerLoop*2) / float64(tokensPerSecond) * float64(time.Second)) // add padding to account for slow test runs
+	t.Logf("expected end: %s, now: %s, allowPad: %s, actual delta: %s", expectedEnd, time.Now(), allowPad, time.Since(expectedEnd))
```
rvagg, line 398: I witnessed a flaky on this test, so I loosened it up a bit; I extended the loops to 25 to smooth out the bumps and doubled the padding. Mostly I've witnessed the expected end be within about 50ms of actual on my machine.

rvagg: https://github.com/filecoin-project/lotus/actions/runs/10300409613/job/28509756058?pr=12327

```
    gateway_test.go:398: expected end: 2024-08-08 10:32:18.112474055 +0000 UTC m=+21.555100385, now: 2024-08-08 10:32:18.167164355 +0000 UTC m=+21.609790693, allowPad: 800ms, actual delta: 54.690384ms
```

rvagg:
@masih || @aarshkshah1992 would you mind having one last look over this before I merge please? @masih the final commit should be the delta between your review and now and contains mostly things you raised.

masih (approved):

PR #12328, opened 2024-07-31
fix: send cli: only override method if recipient is a "0x" address

### LINKED ISSUE #12326: `lotus send` shouldn't default to InvokeContract just because IsEthAddress(To)
Pulling out [this comment](https://github.com/filecoin-project/lotus/pull/12319/files#r1697351622) from #12319 (which is merged) so we don't lose track of it.

> * There are almost certainly users using native accounts to send messages to EVM smart contracts, explicitly specifying the method number.
>  * Implicitly calling InvokeContract is the nice thing to do, but that's a significant change of behavior.
>
> Ideally this would have this method number handling if and only if the user passes a literal 0x... address in the to field. If they pass a normal Filecoin address, they should get the normal Filecoin method handling.

This seems reasonable, so we should probably replace `IsEthAddress(To)` with a bool that gets set when we originally do a `ParseEthAddress` and it succeeds.

I'm also wondering why we strictly disable `--method` for these; does it make sense to lift that restriction and caveat emptor?

aarshkshah1992:
Closes #12326 .

### DIFF
--- a/cli/send.go
+++ b/cli/send.go
@@ -88,6 +88,7 @@ var SendCmd = &cli.Command{
 		ctx := ReqContext(cctx)
 		var params SendParams
 
+		is0xRecipient := false
 		params.To, err = address.NewFromString(cctx.Args().Get(0))
 		if err != nil {
 			// could be an ETH address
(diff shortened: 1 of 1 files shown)


Stebalien:
Yep, lgtm!

Issue #11744, opened 2024-03-19
Option to back-fill events by re-execution of messages


rvagg:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

`lotus-shed indexes backfill-events` will walk through specified epochs and extract events from the AMT referenced by `receipt.EventsRoot` and put them in the index for you. But, it can't do anything if you haven't been recording the events in your blockstore.

If both `Fevm.EnableEthRPC` (existing config option, defaults to `false`) or `Events.EnableActorEventsAPI` (new with 1.26.0, defaults to `false`) are set to `false` then `EnableStoringEvents` isn't set, which prevents both decoding and storing of events on FVM execution.

* https://github.com/filecoin-project/lotus/blob/2e75f3b796dd1396baa37d73ea99dc3407b4da8d/chain/vm/fvm.go#L526-L531 - events don't get decoded
* https://github.com/filecoin-project/lotus/blob/2e75f3b796dd1396baa37d73ea99dc3407b4da8d/chain/consensus/compute_state.go#L233-L236 - `events` don't get populated
* https://github.com/filecoin-project/lotus/blob/2e75f3b796dd1396baa37d73ea99dc3407b4da8d/chain/consensus/compute_state.go#L367-L374 - event blocks and the corresponding AMT don't get stored in the blockstore

Because the current backfill operation relies on `ChainGetEvents` which itself relies on being able to load the event AMT from its root CID on the message receipt _from the blockstore_. Which won't be there if you haven't configured `Fevm.EnableEthRPC=true` or `Events.EnableActorEventsAPI=true`.

### Describe the solution you'd like

[...]

Issue #12116, opened 2024-06-19
Refactoring event Indexing and simplifying the ETH events RPC flow (and fix a bunch of known issues with it)


aarshkshah1992:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [X] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

The current Chain Notify <> Event Index <> Event Filter Management <> ETH RPC Events API is racy, causes missed events, is hard to reason about and has known problems such as lack of automated backfilling, returning empty events for tipsets on the cannonical chain even though the tipset has events, not using the event Index as the source of truth etc. This issue aims to propose a new architecture/flow for event indexing and filtering to fix all of the above.

### Describe the solution you'd like

- After thinking deeply about this and going through the code -> I think our Chain Notify <-> Events Indexing <> Event Filter management <> ETH RPC Events API flow is racy/needs some refactoring to really fix the "events availability" problem for clients among some of the other issues we've run into.
- Really, if the event index is enabled , then the event index should be the source of truth for all events that are served to clients via the Event RPC APIs
- Given that the event index is needed for the correct functioning of the  `EthGetLogs` , `EthGetFilterChanges`  and `EthGetFilterLogs`  anyways, I'd posit that we can really simplify and streamline things if these along with the `EthSubscribe`  API use the event Index as the source of truth
- But that is not the case currently. We do a dual write from the head changes to both the event index and the filter buffers:
See  https://github.com/filecoin-project/lotus/blob/6f821c3285e801a76d76d39444919c61acfbf627/chain/events/filter/event.go#L333 and then https://github.com/filecoin-project/lotus/blob/6f821c3285e801a76d76d39444919c61acfbf627/chain/events/filter/event.go#L340
[...]

PR #12330, opened 2024-07-31
refactor: concurrently process lotus shed backfill events and recompute state

akaladarshi:
## Related Issues
Fixes: #11744 

## Proposed Changes
- Concurrently backfill the events in lotus-shed backfill-events cmd.
- Add `StateRecomputeTipset` in `StateAPI` which recomputes the tipset incase events are not available

### DIFF
--- a/api/api_full.go
+++ b/api/api_full.go
@@ -543,6 +543,11 @@ type FullNode interface {
 	// Messages in the `apply` parameter must have the correct nonces, and gas
 	// values set.
 	StateCompute(context.Context, abi.ChainEpoch, []*types.Message, types.TipSetKey) (*ComputeStateOutput, error) //perm:read
+
+	// StateRecomputeTipset recomputes the state of the given tipset, without trying to lookup a pre-computed result
+	// in the chainstore.
+	StateRecomputeTipset(ctx context.Context, tsk types.TipSetKey) (cid.Cid, error) //perm:read
+
 	// StateVerifierStatus returns the data cap for the given address.
 	// Returns nil if there is no entry in the data cap table for the
 	// address.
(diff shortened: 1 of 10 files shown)


akaladarshi:
@rvagg As I was going through SQLite, I noticed that it doesn’t support concurrent writes. It only supports single writer and multiple readers using `WAL` mode, which doesn’t serve our purpose.

Could you provide some suggestions regarding this? One possible solution I can think of is storing data in memory and writing it later.

When I say write later I mean we can have a channel to read the processed events and write to the DB using that channel (just a thought not sure about the feasibility)

api/api_full.go
```diff
@@ -543,6 +543,11 @@ type FullNode interface {
 	// Messages in the `apply` parameter must have the correct nonces, and gas
 	// values set.
 	StateCompute(context.Context, abi.ChainEpoch, []*types.Message, types.TipSetKey) (*ComputeStateOutput, error) //perm:read
+
+	// StateRecomputeTipset recomputes the state of the given tipset, without trying to lookup a pre-computed result
+	// in the chainstore.
+	StateRecomputeTipset(ctx context.Context, tsk types.TipSetKey) (cid.Cid, error) //perm:read
```
rvagg, line 549: Is this the only way to achieve this via the rpc? do we not expose enough of the pieces to recompute without adding a new API? We should have a bias _against_ adding new RPC APIs if we can help it.

rvagg:
Yes, so the multiple-writer problem is precisely why this needs to be removed from lotus-shed and integrated into lotus itself. Currently we have people using this command but wrapping it in a retry loop so that if it fails a write (because of concurrent) then it just retries. But that risks the other side of this where the lotus daemon itself could fail a write but do so relatively silently so you have a hole in your events because you're backfilling older ones!

But the workaround is something like:
1. turn on events but turn off historical events in your daemon config
2. start the daemon
3. run lotus-shed to backfill
4. stop the daemon
5. turn on historical events
6. start the daemon

And as long as steps 3 and 4 don't run over an epoch boundary and you've backfilled right up to the current epoch, then it _should_ be good.

But obviously this sucks as a method and is not something we can really recommend.

rvagg:
https://github.com/filecoin-project/lotus/issues/12116 has some of the background of this discussion btw

I'm wondering here whether we should just pivot this PR to do the work inside lotus daemon. At least make a start at the backfill task. I wrote up some thoughts on what could be done here: https://github.com/filecoin-project/lotus/issues/12116#issuecomment-2261718066 do you think you'd be interested in trying to make a start on that? We (you @aarshkshah1992 and I) should chat more about this I think. Maybe have a video call but I'll let @aarshkshah1992 react here first.

api/api_full.go
akaladarshi: I was also not sure about adding a new API as `StateCompute` was already there, that's why I raised a draft PR to gather everyone's thoughts, I will try to find some other way to do it by using existing APIs.

Issue #12334, opened 2024-08-02
lotus-shed backfill events doesn't touch events_seen


rvagg:
We forgot to update backfilling when we landed https://github.com/filecoin-project/lotus/pull/12080. If you run backfill-events after that migration (1.28) then it's not going to update the table that we need to determine if we've processed that tipset.

We're going to nuke the lotus-shed backfilling, see conversation in https://github.com/filecoin-project/lotus/pull/12330 and https://github.com/filecoin-project/lotus/issues/12116, but for now it's still used and it should work.

We need to copy this logic: https://github.com/filecoin-project/lotus/blob/6bdbbc024b30da92c9160ded6457d970ebbfa241/chain/events/filter/index.go#L870-L879

Into the bottom of this closure: https://github.com/filecoin-project/lotus/blob/6bdbbc024b30da92c9160ded6457d970ebbfa241/cmd/lotus-shed/indexes.go#L179

PR #12335, opened 2024-08-02
fix: lotus-shed: store processed tipset after backfilling events

akaladarshi:
## Related Issues
Fixes: #12334 

## Proposed Changes
store processed tipset in `event-seen` table after backfilling the events in `lotus-shed backfill-event`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - https://github.com/filecoin-project/lotus/pull/12292: feat: p2p: allow overriding bootstrap nodes with environmemnt variable
 - https://github.com/filecoin-project/lotus/pull/12319: feat: `lotus send CLI`: allow sending to ETH addresses
 - https://github.com/filecoin-project/lotus/pull/12332: fix: ETH RPC: receipts: use correct txtype in receipts
-- https://github.com/filecoin-project/lotus/pull/12335: fix: store processed tipset after backfilling events
+- https://github.com/filecoin-project/lotus/pull/12335: fix: lotus-shed: store processed tipset after backfilling events
 
 ## New features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - https://github.com/filecoin-project/lotus/pull/12292: feat: p2p: allow overriding bootstrap nodes with environmemnt variable
 - https://github.com/filecoin-project/lotus/pull/12319: feat: `lotus send CLI`: allow sending to ETH addresses
 - https://github.com/filecoin-project/lotus/pull/12332: fix: ETH RPC: receipts: use correct txtype in receipts
+- https://github.com/filecoin-project/lotus/pull/12335: fix: store processed tipset after backfilling events
 
 ## New features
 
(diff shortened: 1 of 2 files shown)


akaladarshi:
@rvagg, I am not able to add `skip/changelog` label into this PR. 

Can you help me with that.

aarshkshah1992:
@akaladarshi Please can you update the `CHANGELOG.MD` file to record this bug fix ?

aarshkshah1992 (approved):
lgtm. Just need to update the `ChangeLog`. Thanks a lot.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - https://github.com/filecoin-project/lotus/pull/12292: feat: p2p: allow overriding bootstrap nodes with environmemnt variable
 - https://github.com/filecoin-project/lotus/pull/12319: feat: `lotus send CLI`: allow sending to ETH addresses
 - https://github.com/filecoin-project/lotus/pull/12332: fix: ETH RPC: receipts: use correct txtype in receipts
+- https://github.com/filecoin-project/lotus/pull/12335: fix: store processed tipset after backfilling events
```
rvagg, line 20:
```suggestion
- https://github.com/filecoin-project/lotus/pull/12335: fix: lotus-shed: store processed tipset after backfilling events
```

Since it's an adjacent tool and not lotus itself.

cmd/lotus-shed/indexes.go
```diff
@@ -196,6 +202,11 @@ var backfillEventsCmd = &cli.Command{
 			var eventsAffected int64
 			var entriesAffected int64
 
+			tsKeyCid, err := currTs.Key().Cid()
```
rvagg, line 205: Hard to see this while reviewing on my phone but I think this is only being set once here but we need to do it each time we update it so we're inserting the right one. `currTs` should follow the epoch as we move through the history and its CID should be the thing we insert.

rvagg: nope, my phone deceived me, this is a good optimisation

cmd/lotus-shed/indexes.go
```diff
@@ -299,6 +305,16 @@ var backfillEventsCmd = &cli.Command{
 				}
 			}
 
+			// this statement will mark the tipset as processed and will insert a new row if it doesn't exist
+			// or update the reverted field to false if it does
```
rvagg, line 309:
```suggestion
			// mark the tipset as processed
```

rvagg (approved):
I think I'm good with this except for the minor fix in the changelog - I'm also going to rename the PR to the same and we should do likewise when we squash merge this so it's in the commit log like this.

cmd/lotus-shed/indexes.go
akaladarshi: If you see [processHeight](https://github.com/akaladarshi/lotus/blob/76d29d44b3a93f54627e945ebabdb2dec49cc029/cmd/lotus-shed/indexes.go#L356) is called first then we are updating the [currTs](https://github.com/akaladarshi/lotus/blob/76d29d44b3a93f54627e945ebabdb2dec49cc029/cmd/lotus-shed/indexes.go#L363), and It's not an async call so `currTs` won't be updated till `processHeight` is finished.

Let me know if I am missing something.

CHANGELOG.md
akaladarshi: Done

rvagg (approved):

PR #12345, opened 2024-08-05
[skip changelog] chore: types: remove all references to FFI/rust implementations of CommP/sha2-256-254

ribasushi:


rvagg:
🥳 🎆 🥳 

1 2 3 4 5

🥳 🎆 🥳 

Congratulations! You win ... something

rvagg:
@ribasushi would you mind rebasing on `rvagg/non-storage-verifier` please, it's a bit hard to review this as you pulled in some `master` commits. I've rebased mine to current master but now we have overlapping commits.

rvagg:
fwiw I don't find per-commit reviewing very helpful or productive and would rather a changeset be complete enough to review as a body

rvagg:
this seems ok to me but goes deeper than I was doing in and touches storage/ stuff, so I'd like @magik6k's sign-off on this, and hopefully #11581 too would be great so we can get this flattened and merged.

rjan90:
Could you update the PR title to match the: `<type>: <area>: <subject>` format? 

```
Types: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test, chore
Area: e.g. api, chain, state, mempool, multisig, networking, paych, proving, sealing, wallet, deps
```

rvagg:
I think it's going to have to be against master eventually, I'm having a hard enough time getting reviews on #11581 and adding this to the mix is going to narrow down the number of potential reviewers even further so I'm inclined to tackle merging #11581 first and dealing with this second.

rvagg:
Sorry, I squashed my noisy commits in https://github.com/filecoin-project/lotus/pull/11581 when merging, then I put #12358 on top of it and it got merged. So now this is against master but has conflicts and shows the changes in #11581 again. Sorry to mess you around on this @ribasushi, but I should be able to review this now and merge since it's not my PR. I just need to do some double-checking of the logic in here which is a bit more complicated than #11581 was.

chain/proofs/proofs.go
```diff
@@ -19,48 +19,24 @@ func GenerateUnsealedCID(proofType abi.RegisteredSealProof, pieces []abi.PieceIn
...
+	pcid, psz, err := commp.PieceAggregateCommP(proofType, pieces)
+	if err != nil {
+		return cid.Undef, err
 	}
 
-	// Next we need to generate the unsealed CID by merkleizing the pieces; even though this function
-	// has the same name, it does a different job and won't give us the desired results if we hadn't
-	// done the piece-padding first.
-	return nonffi.GenerateUnsealedCID(proofType, allPieces)
+	return commp.ZeroPadPieceCommitment(pcid, psz.Unpadded(), upssize)
```
rvagg, line 33: ok, so this works because even though `PieceAggregateCommP` doesn't pad out a sector, `ZeroPadPieceCommitment` will defer to go-fil-comp-hashhash's `PadCommP` do to its padding using the static `stackedNulPadding` to fill out the missing portions of the tree

it's quite a departure from the approach here, but appears to work fine and has been tested in production elsewhere.

My only quibble is the "experimental" notice in go-fil-commp-hashhash which I'd love for you to remove - you're an admin of the package and can make that call.

storage/sealer/mock/mock.go
```diff
@@ -732,7 +731,8 @@ func (m mockVerifProver) VerifyWindowPoSt(ctx context.Context, info prooftypes.W
 }
 
 func (m mockVerifProver) GenerateDataCommitment(pt abi.RegisteredSealProof, pieces []abi.PieceInfo) (cid.Cid, error) {
-	return ffiwrapper.GenerateUnsealedCID(pt, pieces)
+	pcid, _, err := commp.PieceAggregateCommP(pt, pieces)
```
rvagg, line 734: note for other reviewer here: the `GenerateUnsealedCID` originally referenced here was the same as the nonffi version which was renamed to `PieceAggregateCommP` for v2; it's different to what's called `GenerateUnsealedCID` here in that it doesn't pad out a sector for you, see note in chain/proofs/proofs.go above.

storage/sealer/ffiwrapper/sealer_cgo.go
```diff
@@ -24,12 +23,13 @@ import (
...
 	rlepluslazy "github.com/filecoin-project/go-bitfield/rle"
-	commpffi "github.com/filecoin-project/go-commp-utils/ffiwrapper"
-	"github.com/filecoin-project/go-commp-utils/zerocomm"
+	"github.com/filecoin-project/go-commp-utils/v2"
+	"github.com/filecoin-project/go-commp-utils/v2/zerocomm"
 	commcid "github.com/filecoin-project/go-fil-commcid"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/proof"
 
+	"github.com/filecoin-project/lotus/chain/proofs"
```
rvagg, line 32: slight irony of this change in that I originally copied the functions being now referenced in `proofs` from this file in #11581; the logic of this PR is that we're deprecating their use here and we're not using ffi for them so they don't really belong here now

rvagg (approved):
I'm OK with this but don't want to merge it unilaterally. I'd really like a @filecoin-project/curio person to cast an eye over it since it touches storage/ pieces

masih (approved):

magik6k (approved):
Stared at this for a while and see no reason why it shouldn't work. If something doesn't work we'll notice rather quickly

PR #12346, opened 2024-08-06
feat: events: add `lotus-shed indexes inspect-events` health-check

rvagg:
Works read-only so can be done on a live events db. Only checks for problems that backfill could (_should be able to_) address.

cc @akaladarshi

### DIFF
--- a/cmd/lotus-shed/indexes.go
+++ b/cmd/lotus-shed/indexes.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"math"
 	"path"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/lotus-shed/indexes.go
```diff
@@ -387,6 +392,192 @@ var backfillEventsCmd = &cli.Command{
...
+				if seenHeight != int(ts.Height()) {
+					problems = append(problems, fmt.Sprintf("events_seen height mismatch (%d)", seenHeight))
+				}
+				if seenReverted != 0 {
+					problems = append(problems, "events_seen marked as reverted")
+				}
+			}
+
+			var actualEvents int
+			if err := stmtEventCount.QueryRow(tsKeyCid.Bytes()).Scan(&actualEvents); err != nil {
```
aarshkshah1992, line 532: This is fine because the events Index maps events/entries against their inclusion tipset and not the execution tipset.

aarshkshah1992 (approved):

PR #12348, opened 2024-08-06
fix: api: add datacap balance to circ supply

jennijuju:
while datacap actor is not expected to be receiving funds, however, it did https://filfox.info/en/message/bafy2bzacebszh56u7lqkooszlmoszfumiq4zztgavsbgnxs2t7cmimpp656cs?t=1 . which is uncovered for circulating supply APIs internal accounting.

### DIFF
--- a/chain/stmgr/supply.go
+++ b/chain/stmgr/supply.go
@@ -417,7 +417,8 @@ func (sm *StateManager) GetCirculatingSupply(ctx context.Context, height abi.Cha
 			a == builtin.BurntFundsActorAddr ||
 			a == builtin.SaftAddress ||
 			a == builtin.ReserveAddress ||
-			a == builtin.EthereumAddressManagerActorAddr:
+			a == builtin.EthereumAddressManagerActorAddr ||
+			a == builtin.DatacapActorAddr:
 
 			unCirc = big.Add(unCirc, actor.Balance)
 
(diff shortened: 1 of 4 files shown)


chain/stmgr/supply.go
```diff
@@ -443,7 +443,8 @@ func (sm *StateManager) GetCirculatingSupply(ctx context.Context, height abi.Cha
 			builtin.IsPaymentChannelActor(actor.Code) ||
 			builtin.IsEthAccountActor(actor.Code) ||
 			builtin.IsEvmActor(actor.Code) ||
-			builtin.IsPlaceholderActor(actor.Code):
+			builtin.IsPlaceholderActor(actor.Code) ||
+			builtin.IsDatacapActor(actor.Code):
```
Stebalien, line 447: We should probably exclude this actor, right? The funds cannot be withdrawn.

Stebalien: Add this to the `unCirc` check above:

`a == builtin.DatacapActorAddr`

jennijuju: I wasn’t sure… cuz it is not supposed to receive funds but it can . I don’t have a strong opinion - your suggestion would make senses cuz I don’t think the actor has a withdraw fil function. Will confirm and make the change

Stebalien: (I hope...)

jennijuju: ah okay good to know! thanks  & updated

jennijuju:
post change output

```
./lotus state  circulating-supply
Exact circulating supply:  23576960.0003615724929024 FIL
```

Stebalien (approved):

Issue #11007, opened 2023-06-26
Enable automatic backfilling of sqlite databases (msgindex, events, txhash)


fridrik01:
Due to the importance of having these indices up-to-date from a performance perspective (API providers), we should make sure that we always have these databases fully backfilled.

Currently we don't automatically backfill the msgindex/events/txhas sqlite databases but instead rely on the user knowing to backfill manually. Having these databases fully backfilled is important to the performance of several RPC methods which otherwise have a huge latency (over 10+ minutes on archival nodes).  

Instead, on Lotus startup, we should just always check if we have missing data (handling also cases where config was changed from disabled to enabled for an index) and then handle the backfilling in the background automatically. Only in some rare edge cases should user need to manually run backfill via lotus-shed, for example if they are migrating dbs and/or editing the raw sqlite data.

rvagg:
This probably needs something like `MaxBackfillLookback` for the case of an archival node turning this on late in the game; maybe you don't want to go back to FEVM and reprocess all tipsets to get events that you don't have.

BigLep:
This is being subsumed by work in https://github.com/filecoin-project/lotus/issues/12453

PR #12353, opened 2024-08-07
feat(tx_index): fill transaction index gap on enabling automatic backfill

akaladarshi:
## Related Issues
Fixes part of: #11007 

## Proposed Changes
- Adds `EnableAutomaticBackFill` in index config in `fullNode`
- Adds `MaxAutomaticBackFillTxIndexHeight` in index config in `fullnode`
- Tries to fill the gap between the last populated from `current tipset tx hashes` till:
  - we reach `MaxAutomaticBackFillTxIndexHeight` or
  - we reach a conflict in the `eth tx hash index` database
- Adds `PopulateAfterSnapshot` for eth tx hash index to populate data for while loading from snapshot

### DIFF
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -71,7 +71,7 @@ type ApisConfig struct {
 }
 
 type JournalConfig struct {
-	//Events of the form: "system1:event1,system1:event2[,...]"
+	// Events of the form: "system1:event1,system1:event2[,...]"
 	DisabledEvents string
 }
 
@@ -632,6 +632,9 @@ type IndexConfig struct {
 	// EXPERIMENTAL FEATURE. USE WITH CAUTION
 	// EnableMsgIndex enables indexing of messages on chain.
 	EnableMsgIndex bool
+
+	// EnableAutomaticBackFill enables automatic index back-filling
+	EnableAutomaticBackFill bool
 }
 
 type HarmonyDB struct {
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -633,8 +633,11 @@ type IndexConfig struct {
 	// EnableMsgIndex enables indexing of messages on chain.
 	EnableMsgIndex bool
 
-	// EnableAutomaticBackFill enables automatic index back-filling
-	EnableAutomaticBackFill bool
+	// EnableAutomaticBackFillTxIndex enables automatic index back-filling
+	EnableAutomaticBackFillTxIndex bool
+
+	// Maximum number of blocks to process during a single back-fill operation
+	MaxAutomaticBackFillTxIndexHeight uint64
 }
 
 type HarmonyDB struct {
--- a/node/impl/full/txhashmanager.go
+++ b/node/impl/full/txhashmanager.go
@@ -26,7 +26,7 @@ func (m *EthTxHashManager) Revert(_ context.Context, _, _ *types.TipSet) error {
 
 // FillIndexGap populates the Ethereum transaction hash lookup database with missing entries
 // by processing blocks until we reach the maximum number of automatic back-fill epochs or the message is already indexed.
-func (m *EthTxHashManager) FillIndexGap(currHead *types.TipSet, ctx context.Context, maxAutomaticBackFillBlocks abi.ChainEpoch) error {
+func (m *EthTxHashManager) FillIndexGap(ctx context.Context, currHead *types.TipSet, maxAutomaticBackFillBlocks abi.ChainEpoch) error {
 	log.Info("Start back-filling transaction index from current head: %d", currHead.Height())
 
 	var (
--- a/node/impl/full/txhashmanager.go
+++ b/node/impl/full/txhashmanager.go
@@ -2,11 +2,11 @@ package full
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/crypto"
-
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/build/buildconstants"
 	"github.com/filecoin-project/lotus/chain/ethhashlookup"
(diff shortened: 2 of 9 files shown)


akaladarshi:
@aarshkshah1992 As per our discussion about waiting till the observer gets the latest head and process that,
I have a few ideas, let's discuss them once we verify this is correct.

node/modules/ethmodule.go
```diff
@@ -78,7 +84,7 @@ func EthModuleAPI(cfg config.FevmConfig) func(helpers.MetricsCtx, repo.LockedRep
 				}
 
 				// Tipset listener
-				_ = ev.Observe(&ethTxHashManager)
+				ev.Observe(&ethTxHashManager)
```
aarshkshah1992, line 88: We don't need this change.

node/config/types.go
```diff
@@ -621,6 +621,9 @@ type IndexConfig struct {
 	// EXPERIMENTAL FEATURE. USE WITH CAUTION
 	// EnableMsgIndex enables indexing of messages on chain.
 	EnableMsgIndex bool
+
+	// EnableAutomaticBackFill enables automatic index back-filling
+	EnableAutomaticBackFill bool
```
aarshkshah1992, line 626: For now, we need to rename this to `EnableAutomaticBackFillTxIndex` as we're only doing it for the TX Index.

node/impl/full/txhashmanager.go
```diff
@@ -23,6 +23,49 @@ func (m *EthTxHashManager) Revert(ctx context.Context, from, to *types.TipSet) e
...
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Finished filling transaction index gap. Total processed block: %d", totalProcessedBlock)
+
+	return nil
+}
+
+// PopulateExistingMappings walk back from the current head to the minimum height and populate the eth transaction hash lookup database
```
aarshkshah1992, line 68: nit: check grammar

node/impl/full/txhashmanager.go
```diff
@@ -23,6 +23,49 @@ func (m *EthTxHashManager) Revert(ctx context.Context, from, to *types.TipSet) e
 	return nil
 }
 
+// FillIndexGap back-fills the gap between the current head and the last populated transaction index
```
aarshkshah1992, line 26: nit: "last populated transaction index in* the Index"

node/impl/full/txhashmanager.go
```diff
@@ -23,6 +23,49 @@ func (m *EthTxHashManager) Revert(ctx context.Context, from, to *types.TipSet) e
...
+	lastTxHash, lastTxCid, err := m.TransactionHashLookup.GetLastTransaction()
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Start filling transaction index gap from: %s", lastTxHash)
+
+	totalProcessedBlock := 0
+	ts := m.StateAPI.Chain.GetHeaviestTipSet()
+	for _, block := range ts.Blocks() {
```
aarshkshah1992, line 37: I think we need to keep going back in the chain to backfill. This loop terminates after backfilling messages in the heaviest tipset only.

node/impl/full/txhashmanager.go
```diff
@@ -23,6 +23,49 @@ func (m *EthTxHashManager) Revert(ctx context.Context, from, to *types.TipSet) e
...
+			}
+
+			// process the message
+			m.ProcessSignedMessage(ctx, msg)
+		}
+
+		totalProcessedBlock++
+	}
+
+	ts, err = m.StateAPI.Chain.GetTipSetFromKey(ctx, ts.Parents())
```
aarshkshah1992, line 58: This needs to move to the loop above

node/impl/full/txhashmanager.go
```diff
@@ -23,6 +23,49 @@ func (m *EthTxHashManager) Revert(ctx context.Context, from, to *types.TipSet) e
...
+		if err != nil {
+			// If we can't find the messages, we've either imported from snapshot or pruned the store
+			log.Debug("exiting message mapping population at epoch ", ts.Height())
+			return nil
+		}
+
+		for _, msg := range msgs {
+			if msg.Cid() == lastTxCid {
+				// We've reached the last indexed transaction
+				return nil
```
aarshkshah1992, line 48: log completion and then return. Add logging around how many entires were inserted, which tipset it started from and which one it ended at etc.

aarshkshah1992 (commented):
Round 1 of review. @akaladarshi Please can we also add logic to populate this from a snapshot just like we do for the `MsgIndex` ?

node/impl/full/txhashmanager.go
rvagg: We need to be careful on two fronts with going backward up the chain for two reasons:

1. the last processed message CID may not be on the canonical chain anymore so you may never encounter it; and
2. the last processed message may not even be in the blockstore due to a splitstore compaciton or some other _happening_ (e.g. you've copied across someone else's txhash.db or restored an old one from disk, whatever)

I suspect `ChainGetPath` might be a good API to try here: get the tipset of the `lastTxHash`, get the latest tipset, find the path between. I'm just not sure what the error conditions are for that - how big a gap is too large, and how do we deal with splitstore GC problems (I suspect "find me the tipset for `lastTxHash` would error in that case if it was too old, in which case what's the strategy then?).

akaladarshi: @rvagg I have been going through the `ChainGetPath` code, but I didn't see it returning a `too old` error, it only returns a `not found` error.

If it is not found then nodes have to import the snapshot.
Is there any other way to get the state?

Also `ChainGetPath` returns both `revert` and `apply` changes, so do we have to filter out `apply` changes and start backfilling?

akaladarshi:
@aarshkshah1992 For the observer part, I thought we could do something like this:

- Create a `backfill chan`
- Provide the `backfill chan` to the observer 
  - Once it gets  a new block from `ChainNotify` is sends it to the `backfill chan`
   - Wait till the `backfill chan` is closed (means backfill is done)
- Provide the same `backfill chan` to the `FillIndexGap` function in `EthTxManager`, also make it async
  - It will wait till it receives the new block through the `backfill chan` from the observer
  - Starts backfilling 
  - Once backfilling is done close the `backfill chan`  so the observer can proceed as usual

node/impl/full/txhashmanager.go
aarshkshah1992: Discuss this with @akaladarshi  async. We worked through multiple solutions here. He will post a note about it.

node/config/types.go
akaladarshi: Done.

Issue #10520, opened 2023-03-20
Add Eth API Block Cache


fridrik01:
We can cache recently [requested blocks at the Eth API layer](https://github.com/filecoin-project/lotus/blob/4c80e179ae2ef72cef0b74755ce6f5bce54dec5b/node/impl/full/eth.go#L230), saving quite a bit of time and state accesses. We’ll probably want to cache both the “full tx” version and the “tx hash” version. This is likely the most important cache for the Eth API.

We can consider using this cache to help calculate fee history, although the fee history cache may end up being significantly larger.

PR #12359, opened 2024-08-08
feat: ETH RPC: Use Block Cache for `EthGetBlockByHash`

aarshkshah1992:
Closes https://github.com/filecoin-project/lotus/issues/10520.

Part of https://github.com/filecoin-project/lotus/issues/12293.

### DIFF
--- a/documentation/en/default-lotus-config.toml
+++ b/documentation/en/default-lotus-config.toml
@@ -241,6 +241,16 @@
   # env var: LOTUS_FEVM_ETHTRACEFILTERMAXRESULTS
   #EthTraceFilterMaxResults = 500
 
+  # EthBlkCacheSize specifies the size of the cache used for caching Ethereum blocks.
+  # This cache enhances the performance of the eth_getBlockByHash RPC call by minimizing the need to access chain state for
+  # recently requested blocks that are already cached.
+  # The default size of the cache is 500 blocks.
+  # Note: Setting this value to 0 disables the cache.
+  #
+  # type: int
+  # env var: LOTUS_FEVM_ETHBLKCACHESIZE
+  #EthBlkCacheSize = 500
+
 
 [Events]
   # DisableRealTimeFilterAPI will disable the RealTimeFilterAPI that can create and query filters for actor events as they are emitted.
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -140,8 +140,8 @@ type EthModule struct {
 	EthTraceFilterMaxResults uint64
 	EthEventHandler          *EthEventHandler
 
-	EthBlkCache   *arc.ARCCache[cid.Cid, *ethtypes.EthBlock]
-	EthBlkTxCache *arc.ARCCache[cid.Cid, *ethtypes.EthBlock]
+	EthBlkCache   *arc.ARCCache[cid.Cid, *ethtypes.EthBlock] // caches blocks by their CID but blocks only have the transaction hashes
+	EthBlkTxCache *arc.ARCCache[cid.Cid, *ethtypes.EthBlock] // caches blocks along with full transaction payload by their CID
 
 	ChainAPI
 	MpoolAPI
@@ -322,7 +322,7 @@ func (a *EthModule) EthGetBlockByHash(ctx context.Context, blkHash ethtypes.EthH
 	// Fetch the tipset using the block hash
 	ts, err := a.Chain.GetTipSetByCid(ctx, cid)
 	if err != nil {
-		return ethtypes.EthBlock{}, xerrors.Errorf("failed to load tipset by CID %s: %w", blkHash.String(), err)
+		return ethtypes.EthBlock{}, xerrors.Errorf("failed to load tipset by CID %s: %w", cid, err)
 	}
 
 	// Generate an Ethereum block from the Filecoin tipset
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/golang-lru/arc/v2"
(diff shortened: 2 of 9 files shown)


aarshkshah1992:
Tested this on a mainnet node

**_With Caching enabled_** (fetch all txdata as well)

**_First request_**

```
real	0m0.060s
```

```
time curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getBlockByHash","params":["0xd5742c7ead47c4224dd1e7a877532fc52fdc3094bad36b85b66661b2c2cf2401", true],"id":1}' http://localhost:1234/rpc/v1

real	0m0.034s
```


**_Second request_**

```
time curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getBlockByHash","params":["0xd5742c7ead47c4224dd1e7a877532fc52fdc3094bad36b85b66661b2c2cf2401", true],"id":1}' http://localhost:1234/rpc/v1

real	0m0.034s
user	0m0.021s
sys	0m0.007s
```

node/impl/full/eth.go
```diff
@@ -138,6 +139,10 @@ type EthModule struct {
 	EthTxHashManager         *EthTxHashManager
 	EthTraceFilterMaxResults uint64
 	EthEventHandler          *EthEventHandler
+
+	EthBlkCache   *arc.ARCCache[cid.Cid, *ethtypes.EthBlock]
```
akaladarshi, line 143: Can you please add some comments here.

so we can get clarity on why the same data is being stored on two different caches.

node/impl/full/eth.go
```diff
@@ -296,11 +301,41 @@ func (a *EthModule) EthGetBlockTransactionCountByHash(ctx context.Context, blkHa
...
+			log.Errorw("nil value in eth block cache", "hash", blkHash.String())
+			cache.Remove(cid)
+		}
+	}
+
+	// Fetch the tipset using the block hash
+	ts, err := a.Chain.GetTipSetByCid(ctx, cid)
 	if err != nil {
-		return ethtypes.EthBlock{}, xerrors.Errorf("error loading tipset %s: %w", ts, err)
+		return ethtypes.EthBlock{}, xerrors.Errorf("failed to load tipset by CID %s: %w", blkHash.String(), err)
```
akaladarshi, line 325: xerrors.Errorf("failed to load tipset by CID %s: %w", cid.String(), err)

akaladarshi (commented):
LGTM.
Just a couple of comments.

Stebalien:
Do you have any before metrics?

documentation/en/default-lotus-config.toml
```diff
@@ -241,6 +241,16 @@
   # env var: LOTUS_FEVM_ETHTRACEFILTERMAXRESULTS
   #EthTraceFilterMaxResults = 500
 
+  # EthBlkCacheSize specifies the size of the cache used for caching Ethereum blocks.
+  # This cache enhances the performance of the eth_getBlockByHash RPC call by minimizing the need to access chain state for
+  # recently requested blocks that are already cached.
```
Stebalien, line 246: I would consider having a separate "recency" cache for the latest N blocks (just to _ensure_ that those requests are very fast). I'm thinking, e.g., this could be tied into the gateway's lookback limit so that all blocks a user _can_ request are also cached.

But this isn't critical and should definitely happen in a followup PR (if ever).

node/impl/full/eth.go
Stebalien: Also, there's no need to explicitly call `String()`.

Stebalien (commented):
Seems like a reasonable approach.

Stebalien (approved):

node/impl/full/eth.go
```diff
@@ -296,11 +301,41 @@ func (a *EthModule) EthGetBlockTransactionCountByHash(ctx context.Context, blkHa
...
+
+	// Attempt to retrieve the Ethereum block from cache
+	cid := blkHash.ToCid()
+	if cache != nil {
+		if ethBlock, found := cache.Get(cid); found {
+			if ethBlock != nil {
+				return *ethBlock, nil
+			}
+			// Log and remove the nil entry from cache
+			log.Errorw("nil value in eth block cache", "hash", blkHash.String())
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien No just a sanity check, can remove if meh

aarshkshah1992:
@Stebalien 

**_Without any caching_**


1st fetch

```
time curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getBlockByHash","params":["0xd5742c7ead47c4224dd1e7a877532fc52fdc3094bad36b85b66661b2c2cf2401", true],"id":1}' http://localhost:1234/rpc/v1

real	0m0.073s
```

2nd fetch
```

time curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getBlockByHash","params":["0xd5742c7ead47c4224dd1e7a877532fc52fdc3094bad36b85b66661b2c2cf2401", true],"id":1}' http://localhost:1234/rpc/v1
real	0m0.065s

```

Stebalien:
2x savings actually aren't that large here... I wonder if that's just because the network is quiet right now.

On the other hand, the cache is pretty small so it probably isn't harmful.

rvagg (approved):
This seems fine but I am a little concerned about the large memory footprint of lotus just getting bigger, turn on the eth API and you have yet another blob in memory. Although thankfully it only grows if you use it.
I'm not sure exactly how much these take up but the JSON output of one of your example blocks is over 500Kb, so maybe they end up being something like 200Kb in memory, so perhaps 100Mb here when we cache 500 of them.

PR #12366, opened 2024-08-09
docs: expand `CONTRIBUTING.md` and update `README.md`

rjan90:
## Related Issues
Closes: #12360

## Proposed Changes
- Expanded `CONTRIBUTING.md` with detailed guidelines that was previously in `README.md`
- Removed the detailed contribution guidelines in `README.md` and instead link to the `CONTRIBUTING.md`
- Update links in `pull_request_template.md`

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title conforms with [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/README.md#pr-title-conventions)
- [x] Update CHANGELOG.md or signal that this change does not need it per [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/README.md#changelog-management)
- [x] New features have usage guidelines and / or documentation updates in
  - [x] [Lotus Documentation](https://lotus.filecoin.io)
  - [x] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12372, opened 2024-08-09
docs: updating CONTRIBUTING.md with table of contents and TOC guidance

BigLep:
This is followup to https://github.com/filecoin-project/lotus/pull/12366#discussion_r1711780636

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,15 @@
+<!-- omit from toc -->
 # Contributing to Lotus
 
 Lotus is a universally open project and welcomes contributions of all kinds: code, docs, and more. We appreciate your interest in contributing!
+
+- [Before Contributing](#before-contributing)
+- [Implementing Changes](#implementing-changes)
+- [PR Title Conventions](#pr-title-conventions)
+- [CHANGELOG Management](#changelog-management)
+- [Markdown Conventions](#markdown-conventions)
+  - [Table Of Contents](#table-of-contents)
+- [Getting Help](#getting-help)
 
 ## Before Contributing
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -43,6 +52,15 @@ If the change does not require a CHANGELOG.md entry, do one of the following:
...
+
+### Table Of Contents
+For longer documents, we tend to inline a table of contents (TOC).  This is slightly annoying with Markdown on GitHub because the actual table of contents needs to be checked-in rather than just giving a directive to the renderer like []`[[_TOC_]]` on GitLab](https://docs.gitlab.com/ee/user/markdown.html#table-of-contents).  (Yes GitHub UI does enable someone to see an auto-generated outline if they click the "outline" button on the top right-hand corner when viewing a Markdown file, but that isn't necessarily obvious to users, requires an extra click, and doesn't enable any customization.)
+
+[There is tooling](https://stackoverflow.com/questions/11948245/markdown-to-create-pages-and-table-of-contents) that can help with auto-generating and updating a TOC and allow customization.  Maintainers aren't wedded to a particular format/standard/tool but have gone with [VS Code Markdown All in One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) because it has what we needed and is the most popular VS Code plugin for this functionality (at least as of 202408).
+
+
```
rjan90, line 63:
```suggestion
## Markdown Conventions
We optimize our markdown files for viewing on GitHub. That isn't to say other syntaxes can't be used, but that is the flavor we focus on and at the minimum don't want to break.

### Table Of Contents
For longer documents, we tend to inline a table of contents (TOC). This is slightly annoying with Markdown on GitHub because the actual table of contents needs to be checked-in rather than just giving a directive to the renderer like []`[[_TOC_]]` on GitLab](https://docs.gitlab.com/ee/user/markdown.html#table-of-contents).  (Yes GitHub UI does enable someone to see an auto-generated outline if they click the "outline" button on the top right-hand corner when viewing a Markdown file, but that isn't necessarily obvious to users, requires an extra click, and doesn't enable any customization.)

[There is tooling](https://stackoverflow.com/questions/11948245/markdown-to-create-pages-and-table-of-contents) that can help with auto-generating and updating a TOC and allow customization.  Maintainers aren't wedded to a particular format/standard/tool but have gone with [VS Code Markdown All in One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) because it has what we needed and is the most popular VS Code plugin for this functionality (at least as of 202408).

```

rjan90 (approved):

CONTRIBUTING.md
rjan90: Just cleaned up some double spaces, a double "we" in the `but that is the flavor we focus` sentence, as well as removing the additional empty line at 63.

Issue #12343, opened 2024-08-05
[WIP] Lotus Node v1.29.0 Release


BigLep:
# Lotus Node 1.29.0 Release

[//]: # (Below are non-visible steps intended for the issue creator)
[//]: # (Start an issue with title "[WIP] Lotus Node|Miner v1.29.0" and adjust the title for whether it's a Node or Miner release.)
[//]: # (Copy in the content of https://github.com/filecoin-project/lotus/blob/master/documentation/misc/RELEASE_ISSUE_TEMPLATE.md)
[//]: # (Find/Replace "X.Y.Z" with the actual values.)
[//]: # (Copy/paste the "Release Checklist > RC1" section to "Release Checklist > Stable \(non-RC\) Release" and apply the "diff" called out there.)
[//]: # (Find/Replace "RCX" with "RC1".)
[//]: # (If this isn't a release tied to a network upgrade, remove all items with "\(network upgrade\)")
[//]: # (Adjust the "Meta" section values.)
[//]: # (Apply the `tpm` label to the issue)
[//]: # (Create the issue)
[//]: # (Pin the issue on GitHub)

## 😶‍🌫 Meta
* Scope: Node
* Is this linked with a network upgrade, and thus mandatory? No
* Related network upgrade version: n/a

## 🚢 Estimated shipping date

[//]: # (If/when we know an exact date, remove the "week of".)
[//]: # (If a date week is an estimate, annotate with "estimate".)

| Candidate | Date | Release URL |
|-----------|------|-------------|
| RC1 | Week of 2024-08-19 | https://github.com/filecoin-project/lotus/releases/tag/v1.29.0-rc1 |
| Stable (non-RC) | Week of 2024-09-01 | https://github.com/filecoin-project/lotus/releases/tag/v1.29.0 |

## 🪢 Dependencies for releases
> [!NOTE]
> 1. This is the set of changes that need to make it in for a given RC.  This is effectively the set of changes to cherry-pick from master.  
> 2. They can be checked as done once they land in `master`.  
> 3. They are presented here for quick reference, but backporting is tracked in each `Release Checklist`.

[//]: # (Copy/paste this for each RC, and increment "X")
### RC1
- [x] https://github.com/filecoin-project/lotus/pull/12327

### Stable (non-RC)
- [x] ~https://github.com/filecoin-project/lotus/issues/12414~
   - Noted in the changelog that we are releasing v1.29.0 without this fix.

### Before RC1
[...]

PR #12322, opened 2024-07-30
docs: update LOTUS_RELEASE_FLOW.MD document

rjan90:
## Related Issues
#12072

## Proposed Changes
Update the LOTUS_RELEASE_FLOW document to account for the new release flow.

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [x] Update CHANGELOG.md or signal that this change does not need it.
  - If the PR affects users (e.g., new feature, bug fix, system requirements change), update the CHANGELOG.md and add details to the UNRELEASED section.
  - If the change does not require a CHANGELOG.md entry, do one of the following:
    - Add `[skip changelog]` to the PR title
    - Add the label `skip/changelog` to the PR
- [x] New features have usage guidelines and / or documentation updates in
  - [x] [Lotus Documentation](https://lotus.filecoin.io)
  - [x] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12373, opened 2024-08-09
docs: update RELEASE_ISSUE_TEMPLATE.md base on v1.29.0 release execution

BigLep:
This is accumulating improvements based on lessons learned from executing https://github.com/filecoin-project/lotus/issues/12343

### DIFF
--- a/documentation/misc/RELEASE_ISSUE_TEMPLATE.md
+++ b/documentation/misc/RELEASE_ISSUE_TEMPLATE.md
@@ -3,7 +3,7 @@
 # Lotus Node|Miner X.Y.Z Release
 
 [//]: # (Below are non-visible steps intended for the issue creator)
-[//]: # (Start an issue with title "[WIP] Lotus Node|Miner vX.Y.Z" and adjust the title for whether it's a Node or Miner release.)
+[//]: # (Start an issue with title "[WIP] Lotus Node|Miner vX.Y.Z Release" and adjust the title for whether it's a Node or Miner release.)
 [//]: # (Copy in the content of https://github.com/filecoin-project/lotus/blob/master/documentation/misc/RELEASE_ISSUE_TEMPLATE.md)
 [//]: # (Find/Replace "X.Y.Z" with the actual values.)
 [//]: # (Copy/paste the "Release Checklist > RCX" section to "Release Checklist > Stable \(non-RC\) Release" and apply the "diff" called out there.)


rjan90:
Seems like we found an edge-case in the pr-title-check, where the regex pattern doesn't account/allow for periods (.) in the scope. I pushed a change here: https://github.com/filecoin-project/lotus/pull/12373/commits/de60b086d5039dbd4659933a26cfd3f61c83cda8. Just to confirm that this will fix it.

rvagg:
I think I'd blame the message rather than the regex here, it probably should be something like a bare `docs:`, like `docs: v1.29.0 release template improvements`, or `docs(release):`, like `docs(release): template improvements from v1.29.0

rjan90:
@BigLep Sorry that I kind of hijacked this PR to refactor the PR Title Check workflow to reduce noise, as I got bothered by it when working on the UPDATE_RELEASE_FLOW draft PR: https://github.com/filecoin-project/lotus/pull/12322.

In hindsight I should have opened a separate PR for this, but the monday "jetlag" got the better of me! I will open a new PR for accumulating improvements based on lessons learned from executing https://github.com/filecoin-project/lotus/issues/12343 once this lands.

BigLep:
> I think I'd blame the message rather than the regex here, it probably should be something like a bare `docs:`, like `docs: v1.29.0 release template improvements`, or `docs(release):`, like `docs(release): template improvements from v1.29.0

That's fare.  I'll adjust to `docs(release):` and put the version number in the description itself.

BigLep:
> Sorry that I kind of hijacked this PR to refactor the PR Title Check workflow to reduce noise

No problem - I'll create a new PR with the release issue improvements.  (I now need to also add the change to doing "docs(release): ..."

BigLep (commented):
While we're here, how about we remove "synchronize" from line 5.  Pushing  commits shouldn't need to trigger this workflow.

Also, given we've done a few rounds of tweaking this, I'm wondering if we would have been better off going with something off the shelf that has already been tweaked...

rjan90:
> Also, given we've done a few rounds of tweaking this, I'm wondering if we would have been better off going with something off the shelf that has already been tweaked...

I agree. What we potentially miss with going with something off the shelf, is the period where both styles are allowed (i.e `fix(releases):` and `fix: releases:`)

If we encounter more issues with it, I´m inclined to go with the off the shelf option, just so that we do not spend more time on fixing it.

BigLep:
> What we potentially miss with going with something off the shelf, is the period where both styles are allowed (i.e `fix(releases):` and `fix: releases:`)

The top three tools I saw all allow passing in a custom regex.  That said, looking at it more, I don't think other tools meet all our needs.  I just added a comment in the original PR: https://github.com/filecoin-project/lotus/pull/12340#issuecomment-2284434397

So, basically lets hold the course and finish fine tuning this for our needs.  Thanks!

BigLep:
New PR for doing the release process improvements: https://github.com/filecoin-project/lotus/pull/12378

rjan90 (approved):

PR #12388, opened 2024-08-14
[WIP] A new `ChainIndexer` that replaces the fragmented MsgIndex, EthTxHashIndex and EventsIndex

aarshkshah1992:
This PR implements a new `ChainIndexer` in Lotus that indexes tipsets, messages, transaction hashes, and events by consuming the `ChainNotify` API.

 It aims to replace and subsume the existing `MsgIndex`, `EventsIndex`, and `EthTxHashIndex`, which are currently fragmented across multiple databases and have several known issues documented in [filecoin-project/lotus#12293](https://github.com/filecoin-project/lotus/issues/12293).

###  Key Features
The `ChainIndexer` offers the following key features:

- Indexes all necessary state in a single database. 
- We now index a "tipset" and the relevant state changes caused by a tipset which makes the Indexed state easy to reason about and more aligned with how we persist state in the Chainstore. 
- Implements snapshot hydration.
- Implements  automated config driven Index garbage collection (GC).
- Provides automated backfilling.
- Offers simplified configuration.
- Wraps the asynchronous indexing in a synchronous READ API for RPC endpoints to consume to avoid off by N errors caused by async Indexing.

**Here are the next implementation steps going ahead to get this PR in a ready for review/ready for testing state.**

### Switch RPC APIs to use the Chain Index
- The Filecoin and ETH RPC APIs will switch to using the `ChainIndexer` instead of the `MsgIndex`, `EthTxHashIndex` and `EventsIndex`.
- The `EventFilterManager` will read events from the `ChainIndexer` and prefill all registered filters rather than depending on the Indexer to do the pre-filling of filters.
- The `ChainIndexer` will listen to Mpool message addition updates to index the corresponding ETH Tx Hash. The `EthTxHashManager` will no longer be used for this.

### Read APIs Should Account for the Async Nature of Indexing
- All APIs that read from the index will wait for the current head in the chainstore to be indexed if their first read attempt fails and then retry the read before returning a response to the user.
  - Note that we're only waiting for an already produced tipset in our chainstore to be indexed here; we're not waiting for a new tipset to be produced. This should work well in practise but if it times out, it points to another underlying problem in our chainstore <-> indexing pathway that should be investigated.
- This is a workaround to handle asynchronous indexing in Filecoin.
- For events, it should be noted that indexing the current head `T` only indexes events in `T-1` because of deferred execution.

### ETH RPC APIs Should Only Expose Executed Tipsets and Messages
- As part of this work, we should review all ETH RPC APIs to ensure they only expose and respond to requests for executed tipsets and transactions. This is because Filecoin uses a deferred execution model, unlike Ethereum:
 - In Filecoin, messages included in tipset `T` are executed in tipset `T + 1`.
 - In Ethereum, messages included in tipset `T` are also executed in tipset `T`.
- Exposing messages and tipsets that have not been executed yet via ETH RPC APIs in Lotus causes errors when users ask for the corresponding execution state, receipts, or events for those tipsets/messages because they do not yet exist.
- There have been proposals for an even more conservative implementation where ETH RPC APIs only expose "finalized" tipsets and messages post-F3. However, it remains to be determined how well this would work in practice, given that clients might end up waiting for up to 3 epochs (90 seconds) for already included messages in the worst case. 
- For now, we should go ahead with exposing all(even non-finalised) executed tipsets/messages and revisit the implementation after F3 ships.

### Removing Re-orged Tipsets That Are No Longer Part of the Canonical Chain
- The `ChainIndexer` will periodically prune all permanently re-orged/reverted tipsets from the index. It can do this by simply pruning all tipsets at a height less than `(current head - finality policy - some buffer)`.
- The use of foreign key-based cascading deletes in the DDL will greatly simplify this implementation. By simply deleting a tipset from the index, all associated indexed state will be deleted from the DB. See [SQLite Foreign Keys](https://www.sqlite.org/foreignkeys.html) for more information.

### Garbage Collection
- Garbage collection (GC) will be configuration-driven. Users can specify how much history they want to retain, and the `ChainIndexer` can perform periodic GC based on this configuration.
- GC should be straightforward in the `ChainIndexer` because of the use of `FOREIGN KEY ON CASCADE DELETES`, as described in [SQLite Foreign Keys](https://www.sqlite.org/foreignkeys.html).
 - When a tipset is deleted, all associated indexed state will be automatically deleted from the DB due to the cascading delete behavior.
 
### Snapshot Hydration
- When a node is synced from a snapshot, the index should be completely deleted, and a new index should be hydrated from the snapshot.
- It's important to note that snapshots don't contain events. In order to hydrate events in the index, messages in the tipset will have to be re-executed.

### Automated Backfilling
- When a Lotus node starts up, it performs the following steps:
   - Looks up the latest non-reverted tipset in the `ChainIndex` for which the corresponding state exists in the statestore.
    - Instantiates the `Observer` with that tipset as the current head.
   - Starts the `Observer`.
- This process ensures that the `ChainIndexer` will observe the `(Apply, Revert)` path between its last non-reverted indexed tipset and the current heaviest tipset in the chainstore before processing real-time updates, effectively performing automated backfilling.
- One challenge arises for a niche use case when an RPC provider toggles the Indexing flag to ON after keeping it OFF for an extended period or for the first time. In such cases, the backfilling backlog could interfere with indexing real-time tipset changes, potentially impacting RPC queries that primarily target state at or near the head.
- To address this issue, a configuration option can be exposed that allows such users to disable automated backfilling if their primary focus is serving RPC queries for new tipsets after enabling indexing.

### Simplify Indexing Config
- The current indexing configuration in Lotus is extremely complex, with partial indexing options that make it difficult for node operators to understand the state they are indexing or should be indexing. 
- To improve the user experience and simplify the implementation, the current config will be replaced with a simple "Indexing ON/Indexing OFF" switch.  Users either index everything that Lotus needs to provide fast RPC responses or index nothing. 
- The niche use case of "Index X but do not Index Y" will no longer be supported.

### Migration from Old Indices to the New ChainIndex
- Develop a `lotus-shed` utility that allows users to migrate existing indices to the new `ChainIndexer` database. This command should only be executed when the Lotus node is offline to ensure data consistency and avoid potential conflicts.
- When a Lotus node starts up, it should bypass any migration or backfilling processes and directly begin indexing new tipsets in the `ChainIndexer`. This approach offers several benefits:
 1. Users can migrate the historical index at their own pace without incurring a performance penalty during node startup.
 2. The node can quickly respond to queries for new tipsets since indexing for these tipsets commences as soon as the node is operational.
- By decoupling the migration process from the node startup, users gain flexibility in managing the transition to the new indexing system while maintaining optimal performance for real-time tipset indexing.

### Solid Unit Tests, itests and testing on a calibnet node
- Goes without saying


chain/index/msgindex.go
```diff
@@ -23,15 +23,6 @@ const DefaultDbFilename = "msgindex.db"
 
 var log = logging.Logger("msgindex")
 
-var ddls = []string{
-	`CREATE TABLE IF NOT EXISTS messages (
-     cid VARCHAR(80) PRIMARY KEY ON CONFLICT REPLACE,
```
aarshkshah1992, line 28: We will completely remove the msgIndex once this PR lands.

chain/index/ddls.go
```diff
@@ -0,0 +1,98 @@
...
+		message_index INTEGER NOT NULL,
+		events_processed INTEGER NOT NULL,
+		UNIQUE (tipset_key_cid, message_cid)
+	)`,
+
+	`CREATE TABLE IF NOT EXISTS event (
+		event_id INTEGER PRIMARY KEY,
+		message_id INTEGER NOT NULL,
+        event_index INTEGER NOT NULL,
+        emitter_addr BLOB NOT NULL,
```
rvagg, line 23: space/tab issue here btw

chain/index/ddls.go
```diff
@@ -0,0 +1,98 @@
...
+		indexed INTEGER NOT NULL,
+		flags BLOB NOT NULL,
+		key TEXT NOT NULL,
+		codec INTEGER,
+		value BLOB NOT NULL,
+		FOREIGN KEY (event_id) REFERENCES event(event_id) ON DELETE CASCADE
+	)`,
+
+	`CREATE TABLE eth_tx_hash (
+		eth_tx_hash TEXT PRIMARY KEY,
```
rvagg, line 40: would a fixed-length `CHAR` buy us some efficiency here? I think we should be able to know with certainty what the length of every txhash is, right?

rvagg: ah, probably not, I think sqlite might fake its fixed-length fields

chain/index/chain_indexer.go
```diff
@@ -0,0 +1,723 @@
...
+	// if we've already indexed events for this tipset, mark them as unreverted and return
+	res, err := tx.Stmt(ci.stmts.stmtEventsUnRevert).ExecContext(ctx, msgTsKeyCidBytes)
+	if err != nil {
+		return xerrors.Errorf("error unreverting events for tipset: %w", err)
+	}
+	rows, err := res.RowsAffected()
+	if err != nil {
+		return xerrors.Errorf("error unreverting events for tipset: %w", err)
+	}
+	if rows > 0 {
```
rvagg, line 222: ok, so we're using this as our indicator of whether we've seen these events before and won't bother checking individual events like we do now, this seems sensible I think and avoids overhead, although it has more confidence that we have a consistent state!

chain/index/chain_indexer.go
```diff
@@ -0,0 +1,723 @@
...
+
+func (ci *ChainIndexer) GetMsgInfo(ctx context.Context, msg_cid cid.Cid) (MsgInfo, error) {
+	msgInfo, err := ci.lookupMsg(ctx, msg_cid)
+	if err == nil {
+		return msgInfo, nil
+	}
+
+	if err == sql.ErrNoRows {
+		// wait till head is processed and retry
+		if err := ci.waitTillHeadIndexed(ctx); err != nil {
```
rvagg, line 503: The msg could be in the mpool, right? that's about the only other place that legitimate messages might be. This call will block for up to 30s, is that a reasonable thing to do in this case? Shouldn't we just bail and say "not found" earlier under the assumption that we'll pick it up in the mpool (when you eventually implement that)?

chain/index/chain_indexer.go
```diff
@@ -0,0 +1,723 @@
...
+		if err != nil {
+			return nil, xerrors.Errorf("read events: %w", err)
+		}
+
+	}
+
+	return ems, nil
+}
+
+func (ci *ChainIndexer) GetMsgInfo(ctx context.Context, msg_cid cid.Cid) (MsgInfo, error) {
```
rvagg, line 495: hopefully the linter is going to have issues with your underscore in `msg_cid`

chain/index/chain_indexer.go
```diff
@@ -0,0 +1,723 @@
...
+
+	return MsgInfo{
+		Message: msg_cid,
+		Epoch:   abi.ChainEpoch(height),
+		TipSet:  tsKeyCid,
+	}, nil
+}
+
+func (ci *ChainIndexer) waitTillHeadIndexed(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
rvagg, line 538: should we just use `EpochDurationSeconds` here?

rvagg:
The "waitTill" up to 30s concerns me a little in these APIs. I _think_ the problem we're trying to solve is the async updating of the index, correct? So we might be querying something that's just happened but it hasn't propagated to the db. But these waits could be up to 30s to find out if something just doesn't exist.

I wonder if there's an alternative strategy here in checking with the status of the chain synchronously. We could do something like:

1. Check if what we want is there (like currently); return if it is, continue if it's not
2. Find out what the heaviest tipset is with `GetHeaviestTipSet` which I believe is updated synchronously with execution
3. Do a "wait for tipset" on that particular tipset
4. Re-do the query and return that result

Then we deal with the async updating but only have to wait until what has already been processed has propagated. Does that work?

chain/index/chain_indexer.go
aarshkshah1992: @rvagg Eventually, mpool msgs will also be in this DB once I wire it up. This is to handle the case where user is asking for something in the HEAD.

chain/index/chain_indexer.go
aarshkshah1992: Sorry, trying out some new tooling. Will fix all this.

chain/index/chain_indexer.go
aarshkshah1992: @rvagg Honestly, this has nothing to do with an epoch. I think we can change the timeout to 5seconds. The idea is that the Index should reflect the heaviest block we have in our statestore that's it. We're not waiting for a new block to come in here.

aarshkshah1992:
> The "waitTill" up to 30s concerns me a little in these APIs. I _think_ the problem we're trying to solve is the async updating of the index, correct? So we might be querying something that's just happened but it hasn't propagated to the db. But these waits could be up to 30s to find out if something just doesn't exist.
> 
> I wonder if there's an alternative strategy here in checking with the status of the chain synchronously. We could do something like:
> 
> 1. Check if what we want is there (like currently); return if it is, continue if it's not
> 2. Find out what the heaviest tipset is with `GetHeaviestTipSet` which I believe is updated synchronously with execution
> 3. Do a "wait for tipset" on that particular tipset
> 4. Re-do the query and return that result
> 
> Then we deal with the async updating but only have to wait until what has already been processed has propagated. Does that work?

@rvagg That is the plan and what I'm trying to do here i.e. ensure that the Index has Indexed the current heaviest tipset in the chainstore before retrying a read if the first read attempt fails. We don't need to wait for a new block to come in here.  I will update the code/docs/timeout to make this more clearer.

chain/index/ddls.go
```diff
@@ -0,0 +1,98 @@
...
+// TODO: Allow for msg_cid <> eth_tx_hash for unconfirmed mpool messages
+var ddls = []string{
+	`CREATE TABLE IF NOT EXISTS tipset_message (
+		message_id INTEGER PRIMARY KEY,
+		tipset_key_cid BLOB NOT NULL,
+		height INTEGER NOT NULL,
+		reverted INTEGER NOT NULL,
+		message_cid BLOB NOT NULL,
+		message_index INTEGER NOT NULL,
+		events_processed INTEGER NOT NULL,
```
rvagg, line 15: what's the purpose of this field btw?

aarshkshah1992: @rvagg Eventually, want the event read API to be differentiate between "events processed and no events" vs "yet to process events". This is a simple denormalised field that will help with that.

aarshkshah1992: For a tipset T, events in T will be processed when we index a tipset U such that Parent(U)=T.

rvagg: but do we use this anywhere? I'm not seeing it, what's the plan?

aarshkshah1992: @rvagg Yeah lemme add a "ReadEventsAPI" to show it'll be used and wire it into the `EventFilterManager`.

aarshkshah1992: This PR was to show the general direction I am taking. Next one will be code complete.

rjan90:
> Please update the PR title to match https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#pr-title-conventions

Maybe we should allow the [WIP] prefix in PR-titles as well, so that this PR-checker is not so noisy?

aarshkshah1992:
@Stebalien In case you feel like taking a look.

BigLep:
> Maybe we should allow the [WIP] prefix in PR-titles as well, so that this PR-checker is not so noisy?

@rjan90 : I think ideally we'd allow [WIP] in the title when the PR is in draft but then have a tighter check when it's out of draft?

BigLep:
Thanks for the awesome writeup @aarshkshah1992 about where this is going and what is remaining.  Very thorough but easy to understand.  Good stuff!  I'm excited to see this land.

aarshkshah1992:
@github-actions No sorry, cannot bro ❤️

aarshkshah1992:
The goal of this PR was to implement a POC for the `ChainIndexer` to flush out the problem space and get initial feedback.

This PR has now been subsumed by https://github.com/filecoin-project/lotus/pull/12421.

The description has been ported over to an issue created for the `ChainIndexer` work at https://github.com/filecoin-project/lotus/issues/12453.

PR #12389, opened 2024-08-14
fix(pubsub): Switch to a last-seen cache strategy

Stebalien:
## Proposed Changes

We were using the default first-seen strategy which means we'd loop messages every 2 minutes (when they expire from the cache). This new strategy keeps the message in the cache until we haven't seen it for 2 minutes, ensuring that it eventually drops off the network (with high likelihood).

The downside to this change, and a reason to potentially not include it, is that it may significantly harm message propegation. Specifically:

1. Node sends a message.
2. Node's peers propegate it.
3. Peer re-sends it.
4. Nodes peers (the same peers) have already seen the message so they _don't_ propagate it. But some other nodes on the network didn't get it, so the message is censored.

In practice, we've configured our re-broadcast interval to be greater than the cache time so this may not be an issue? But this still isn't a straight-forward merge.

## Additional Info

The correct fix is message expiration: https://github.com/libp2p/go-libp2p-pubsub/issues/573, but that requires deeper protocol changes.

## Checklist

- [ ] CI is green

### DIFF
--- a/node/modules/lp2p/pubsub.go
+++ b/node/modules/lp2p/pubsub.go
@@ -9,6 +9,7 @@ import (
 
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	pubsub_pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	"github.com/libp2p/go-libp2p-pubsub/timecache"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/minio/blake2b-simd"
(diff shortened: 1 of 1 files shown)


rjan90:
What steps are needed for this PR to proceed, assuming we still want to move forward? During yesterdays triage meeting, we discussed this briefly, and a key question was how to properly test this change. Should we consider converting it back to a draft while we address this?

masih:
@Stebalien  is it worth making this into a manifest config to safely test it or you think it's harmless enough to merge?

Stebalien:
(I meant to close this ages ago)

Issue #12377, opened 2024-08-12
Implement a lotus-shed command to garbage collect all the indices not available in the state store


akaladarshi:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [X] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Before unifying all the indexes (msg, tx, event) into a single DB,  it’s crucial to garbage-collect (GC) any indices that are no longer referenced in the state store. This step will ensure that pruned or obsolete data isn't carried over during the migration, leading to a cleaner and more efficient database.

### Describe the solution you'd like

Introduce a lotus-shed command that performs garbage collection on all indices not referenced by the current state. This command should effectively remove unnecessary indexes, optimizing the database before the unification of the indexes.

### Additional context
This has been discussed [here](https://github.com/filecoin-project/lotus/issues/10807#issuecomment-2274278435) as well and it will be helpful to @aarshkshah1992, as he is working on unifying the DB for the indexes.

rvagg:
@akaladarshi @aarshkshah1992 we need to figure out what the most efficient way to figure out what our oldest state tree is. We have all the tipsets in the chain store but I suppose if we were to iterate over the state roots and try and load just the state root block then we'd eventually find one that doesn't load.

aarshkshah1992:
@rvagg

As discussed offline, I think the best approach here is for the user to specify "start with tipset X and gc everything backwards till epoch Y".   That atleast enables majority of our users who run a snapshot synced node with splitstore enabled (sync once -> then "run forever")to GC events they know they don't have the state for anymore (these users usually retain state for the last 1 or 2 days). 

Once we have this in place, we can think about GC strategies for other use cases.

BigLep:
This is subsumed by work happening in https://github.com/filecoin-project/lotus/issues/12453

PR #12393, opened 2024-08-15
feat(index): add lotus shed commands to prune all indexes

akaladarshi:
## Related Issues
Fixes: #12377 

## Proposed Changes
- Add a `lotus-shed` command to prune the indexes, other 
  - Add a `prune-txhash` command to prune tx hash index
  - Add a `prune-events` command
  - refactor `prune-msgindex` command to prune the msg index
  - Add `prune-all-indexes` command to prune all the indexes (msg, events, tx hash)at once
 
- Use `older-than` flag to get the height to start pruning indexes older than the provided height

### DIFF
--- a/cmd/lotus-shed/indexes.go
+++ b/cmd/lotus-shed/indexes.go
@@ -899,12 +899,11 @@ var backfillTxHashCmd = &cli.Command{
 
 var pruneTxHashCmd = &cli.Command{
 	Name:  "prune-txhash",
-	Usage: "Prune the txhash.db for txs older than a given height",
+	Usage: "Prune the txhash.db for txs older than provided days",
 	Flags: []cli.Flag{
 		&cli.IntFlag{
-			Name:  "older-than",
-			Value: 0,
-			Usage: "height to start pruning txhashes older than this epoch, 0 means start from head",
+			Name:  "older-than-days",
+			Usage: "prune tx hashes older than this number of days, 0 means prune all",
 		},
 	},
 	Action: func(cctx *cli.Context) error {
@@ -914,15 +913,17 @@ var pruneTxHashCmd = &cli.Command{
 		}
 		defer srv.Close() //nolint:errcheck
 
-		api := srv.FullNodeAPI()
-		ctx := lcli.ReqContext(cctx)
-
-		basePath, startHeight, err := parsePruneArgs(ctx, cctx, api)
+		basePath, err := homedir.Expand(cctx.String("repo"))
 		if err != nil {
-			return xerrors.Errorf("error parsing prune args: %w", err)
+			return xerrors.Errorf("failed to get base path: %w", err)
 		}
 
-		if err := pruneTxIndex(basePath, startHeight); err != nil {
+		days := cctx.Int("older-than-days")
+		if days < 0 {
+			return xerrors.Errorf("invalid number of days to prune: %d", days)
+		}
+
+		if err := pruneTxIndex(basePath, days); err != nil {
 			return xerrors.Errorf("error pruning txindex: %w", err)
 		}
 
@@ -971,7 +972,11 @@ var pruneAllIndexesCmd = &cli.Command{
 		&cli.IntFlag{
 			Name:  "older-than",
 			Value: 0,
-			Usage: "height to start pruning events older than this epoch, 0 means start from head",
+			Usage: "height to start pruning msg index and events older than this epoch, 0 means start from head",
+		},
+		&cli.IntFlag{
+			Name:  "older-than-days",
+			Usage: "prune txhashes older than this number of days",
 		},
 	},
 	Action: func(cctx *cli.Context) error {
@@ -1000,7 +1005,12 @@ var pruneAllIndexesCmd = &cli.Command{
 		})
 
 		g.Go(func() error {
-			if err := pruneTxIndex(basePath, startHeight); err != nil {
+			days := cctx.Int("older-than-days")
+			if days <= 0 {
+				return xerrors.Errorf("invalid number of days to prune: %d", days)
+			}
+
+			if err := pruneTxIndex(basePath, days); err != nil {
 				return xerrors.Errorf("error pruning txindex: %w", err)
 			}
 
@@ -1100,8 +1110,8 @@ func pruneEventsIndex(basePath string, startHeight int64) error {
 	return nil
 }
 
-// pruneTxIndex is a helper function that prunes the txindex.db for a number of epochs starting from a specified height
-func pruneTxIndex(basePath string, startHeight int64) error {
+// pruneTxIndex is a helper function that prunes the txindex.db for txs older than a given number of days
+func pruneTxIndex(basePath string, days int) error {
 	log.Infof("pruning txindex")
 
 	dbPath := path.Join(basePath, indexesDBDir, ethhashlookup.DefaultDbFilename)
@@ -1116,7 +1126,7 @@ func pruneTxIndex(basePath string, startHeight int64) error {
 		}
 	}()
 
-	rowsAffected, err := ethhashlookup.DeleteEntriesOlderThan(db, abi.ChainEpoch(startHeight))
+	rowsAffected, err := ethhashlookup.DeleteEntriesOlderThan(db, days)
 	if err != nil {
 		return xerrors.Errorf("failed to delete entries: %w", err)
 	}
--- a/cmd/lotus-shed/indexes.go
+++ b/cmd/lotus-shed/indexes.go
@@ -653,7 +653,7 @@ var backfillMsgIndexCmd = &cli.Command{
 			return err
 		}
 
-		defer closer()
+		defer closer() //nolint:errcheck
 		ctx := lcli.ReqContext(cctx)
 
 		curTs, err := api.ChainHead(ctx)
@@ -753,7 +753,7 @@ var pruneMsgIndexCmd = &cli.Command{
 		api := srvc.FullNodeAPI()
 		closer := srvc.Close
 
-		defer closer()
+		defer closer() //nolint:errcheck
 		ctx := lcli.ReqContext(cctx)
 
 		basePath, startHeight, err := parsePruneArgs(ctx, cctx, api)
@@ -790,7 +790,8 @@ var backfillTxHashCmd = &cli.Command{
 		if err != nil {
 			return err
 		}
-		defer closer()
+
+		defer closer() //nolint:errcheck
 
 		ctx := lcli.ReqContext(cctx)
 
--- a/cmd/lotus-shed/indexes.go
+++ b/cmd/lotus-shed/indexes.go
@@ -31,33 +31,36 @@ import (
 
 const (
 	// same as in chain/events/index.go
-	eventExists                = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
-	eventCount                 = `SELECT COUNT(*) FROM event WHERE tipset_key_cid=?`
-	entryCount                 = `SELECT COUNT(*) FROM event_entry JOIN event ON event_entry.event_id=event.id WHERE event.tipset_key_cid=?`
-	insertEvent                = `INSERT OR IGNORE INTO event(height, tipset_key, tipset_key_cid, emitter_addr, event_index, message_cid, message_index, reverted) VALUES(?, ?, ?, ?, ?, ?, ?, ?)`
-	insertEntry                = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
-	upsertEventsSeen           = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, false) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
-	tipsetSeen                 = `SELECT height,reverted FROM events_seen WHERE tipset_key_cid=?`
-	selectEventIdsFromHeight   = `SELECT id, height FROM event WHERE height < ?`
-	selectEventIdsBetweenRange = `SELECT id, height FROM event WHERE height < ? AND height >= ?`
+	eventExists      = `SELECT MAX(id) FROM event WHERE height=? AND tipset_key=? AND tipset_key_cid=? AND emitter_addr=? AND event_index=? AND message_cid=? AND message_index=?`
+	eventCount       = `SELECT COUNT(*) FROM event WHERE tipset_key_cid=?`
+	entryCount       = `SELECT COUNT(*) FROM event_entry JOIN event ON event_entry.event_id=event.id WHERE event.tipset_key_cid=?`
+	insertEvent      = `INSERT OR IGNORE INTO event(height, tipset_key, tipset_key_cid, emitter_addr, event_index, message_cid, message_index, reverted) VALUES(?, ?, ?, ?, ?, ?, ?, ?)`
+	insertEntry      = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
+	upsertEventsSeen = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, false) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
+	tipsetSeen       = `SELECT height,reverted FROM events_seen WHERE tipset_key_cid=?`
 )
 
 // delete queries
 const (
 	deleteMsgIndexFromStartHeight = `DELETE FROM messages WHERE epoch < ?`
-	deleteMsgIndexBetweenRange    = `DELETE FROM messages WHERE epoch < ? AND epoch >= ?`
 
 	deleteTxHashIndexByCID = `DELETE FROM eth_tx_hashes WHERE cid = ?`
 
-	deleteEventsByIDs        = `DELETE FROM event WHERE id IN (%s)`
-	deleteEventEntriesByIDs  = `DELETE FROM event_entry WHERE event_id IN (%s)`
-	deleteEventSeenByHeights = `DELETE FROM events_seen WHERE height IN (%s)`
+	deleteEventFromStartHeight   = `DELETE FROM event WHERE height < ?;`
+	deleteEventEntriesByEventIDs = `DELETE FROM event_entry WHERE event_id IN (SELECT id FROM event WHERE height < ?);`
+	deleteEventsSeenByHeight     = `DELETE FROM events_seen WHERE height < ?;`
 )
 
 // database related constants
 const (
 	databaseType   = "sqlite"
 	databaseDriver = "sqlite3"
+
+	eventTable       = "event"
+	eventEntryTable  = "event_entry"
+	eventsSeenTable  = "events_seen"
+	msgIndexTable    = "messages"
+	txHashIndexTable = "eth_tx_hashes"
 )
 
 func withCategory(cat string, cmd *cli.Command) *cli.Command {
@@ -73,8 +76,10 @@ var indexesCmd = &cli.Command{
 		withCategory("msgindex", backfillMsgIndexCmd),
 		withCategory("msgindex", pruneMsgIndexCmd),
 		withCategory("txhash", backfillTxHashCmd),
+		withCategory("txhash", pruneTxHashCmd),
 		withCategory("events", backfillEventsCmd),
 		withCategory("events", inspectEventsCmd),
+		withCategory("events", pruneEventsCmd),
 		withCategory("indexes", pruneAllIndexesCmd),
 	},
 }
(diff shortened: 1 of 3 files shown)


akaladarshi:
@aarshkshah1992 We can also create separate commands for pruning the `events`, and `txhash` table separately.

Also raising draft PR to get suggestions on couple of things:
- Instead of going to each event one by one, this PR directly deletes based on the height and epochs
- Use of go routines

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +904,362 @@ var backfillTxHashCmd = &cli.Command{
...
+
+// pruneAllIndexesCmd is a command that prunes all the indexes
+var pruneAllIndexesCmd = &cli.Command{
+	Name:  "prune-all-indexes",
+	Usage: "Prune all the index for a number of epochs starting from a specified height",
+	Flags: []cli.Flag{
+		&cli.IntFlag{
+			Name:  "from",
+			Value: 0,
+			Usage: "the tipset height to start backfilling from (0 is head of chain)",
```
rvagg, line 916: "pruning", not "backfilling"?

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +904,362 @@ var backfillTxHashCmd = &cli.Command{
...
+	)
+	for rows.Next() {
+		var (
+			id     int64
+			height int64
+		)
+		if err := rows.Scan(&id, &height); err != nil {
+			return "", "", "", xerrors.Errorf("error scanning event id: %w", err)
+		}
+		eventIds = append(eventIds, id)
```
rvagg, line 1099: oh, this could get big, probably not ideal for an sql query

rvagg:
I think this is going to be pretty inefficient doing the select and building a really long targeted delete, can we not just use a range delete? I'm also not sure it's worth the bother doing a full range with these (i.e. the "epochs" arg probably doesn't get used by anyone). I think you'd just want to have an "older than" (in fact, you should rename "from" to be something like that, `--older-than`?).

So, can we just bundle the heavy lifting directly into the queries?

```
DELETE FROM event_entry WHERE event_id IN (SELECT id FROM event WHERE height < ?);
DELETE FROM event  WHERE height < ?;
DELETE FROM events_seen WHERE height < ?;
```

akaladarshi:
> I think this is going to be pretty inefficient doing the select and building a really long targeted delete, can we not just use a range delete? I'm also not sure it's worth the bother doing a full range with these (i.e. the "epochs" arg probably doesn't get used by anyone). I think you'd just want to have an "older than" (in fact, you should rename "from" to be something like that, `--older-than`?).
> 
> So, can we just bundle the heavy lifting directly into the queries?
> 
> ```
> DELETE FROM event_entry WHERE event_id IN (SELECT id FROM event WHERE height < ?);
> DELETE FROM event  WHERE height < ?;
> DELETE FROM events_seen WHERE height < ?;
> ```

- Yeah, you're correct range queries will be more efficient, will update it.
   - I was thinking in the direction of how it was entered (event-> events
   _entries -> event_seen), but we can delete them directly (much more efficient)

- As per the discussion we (me and @aarshkshah1992 ) had, we decided to include the `epoch` part, but as you're saying it won't be required as much then queries will become simpler. @aarshkshah1992 please share your views as well before I get into to the changes.

- As for the `--older-than` flag, it makes much more sense than `from`, but the other prune command `prune-msgindex` was using the `from` flag so I wanted to maintain consistency. 
  - So should I update there as well?
  - Also I was thinking we can have separate pruning commands for each index, what are your thoughts on it @rvagg ?

rvagg:
I don't think it need symmetry with the backfill commands, and honestly I find 'from' a little bit awkward for those anyway (I added that "going backwards" in there because of that awkwardness, I kind of wish they were something like `--start=4100000 --end=-10000`; but it's not too important I think.

Separate sub-commands makes sense to me rather than all in one. You may only have one of these indexes enabled anyway.

aarshkshah1992:
@akaladarshi I agree with Rodd. Looking at this now, separate pruning cmds for all Indices where we "prune everything older than epoch X" makes a lot of sense.

akaladarshi:
> rather

I was inclined more towards having a single command to prune all the indexes at once and having individual commands as well.

From a code perspective, there won't be much change as I have already separated each index pruning into its function. It's a matter of calling those functions (individually or together).

akaladarshi:
> @akaladarshi I agree with Rodd. Looking at this now, separate pruning cmds for all Indices where we "prune everything older than epoch X" makes a lot of sense.



> @akaladarshi I agree with Rodd. Looking at this now, separate pruning cmds for all Indices where we "prune everything older than epoch X" makes a lot of sense.

I was inclined more towards having a single command to prune all the indexes at once and having individual commands as well.

From a code perspective, there won't be much change as I have already separated each index pruning into its function. It's a matter of calling those functions (individually or together).

cmd/lotus-shed/indexes.go
akaladarshi: Done.

cmd/lotus-shed/indexes.go
akaladarshi: Updated.

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +886,375 @@ var backfillTxHashCmd = &cli.Command{
...
+			if err != nil {
+				return err
+			}
+
+			startHeight += int64(currTs.Height())
+
+			if startHeight == 0 {
+				return xerrors.Errorf("bogus start height %d", startHeight)
+			}
+		}
```
rvagg, line 927: how about you refactor this into a utility function since this block is repeated 4 times: `basePath, startHeight, err := parsePruneArgs(cctx)`

rvagg:
This is neat, thanks @akaladarshi. I have a suggestion, but I don't want to impose this as a requirement, just if you think it's interesting and worth doing: a `--dry-run` for each of these that does a query instead of a delete for the given args. You could `SELECT count(*) FROM events WHERE height < ?` _and_ the same for `>= ?` so you can report something like "prune would delete X and leave Y events".

I think this should even work to do both in one query, for each table:

```
SELECT
    SUM(CASE WHEN height < ? THEN 1 ELSE 0 END) AS delete_count,
    SUM(CASE WHEN height >= ? THEN 1 ELSE 0 END) AS leave_count
FROM events
```

This is just a nice feature, I think though running this on my own node and I don't think I want to be pruning at the moment (I'm building up a big db which is helpful for debugging) but I would like to see what it would do, so that's what I came up with.

akaladarshi:
> This is neat, thanks @akaladarshi. I have a suggestion, but I don't want to impose this as a requirement, just if you think it's interesting and worth doing: a `--dry-run` for each of these that does a query instead of a delete for the given args. You could `SELECT count(*) FROM events WHERE height < ?` _and_ the same for `>= ?` so you can report something like "prune would delete X and leave Y events".
> 
> I think this should even work to do both in one query, for each table:
> 
> ```
> SELECT
>     SUM(CASE WHEN height < ? THEN 1 ELSE 0 END) AS delete_count,
>     SUM(CASE WHEN height >= ? THEN 1 ELSE 0 END) AS leave_count
> FROM events
> ```
> 
> This is just a nice feature, I think though running this on my own node and I don't think I want to be pruning at the moment (I'm building up a big db which is helpful for debugging) but I would like to see what it would do, so that's what I came up with.

That sounds like a good-to-have feature, I will start working on it once I am done with testing this PR locally with the calibration testnet node before merging and migrating of indexes to common DB.

cmd/lotus-shed/indexes.go
akaladarshi: Done.

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +900,339 @@ var backfillTxHashCmd = &cli.Command{
...
+			fmt.Println("request cancelled")
+			return nil
+		default:
+		}
+
+		for _, blockheader := range currTs.Blocks() {
+			blkMsgs, err := api.ChainGetBlockMessages(ctx, blockheader.Cid())
+			if err != nil {
+				log.Infof("failed to get block messages at height: %d", currTs.Height())
+				continue // should we break here?
```
akaladarshi, line 1149: @rvagg should we break here or keep going till the first tipset?

Is there a possibility, that a block is missing in between?

While testing I saw one block was missing, after that, it kept going on like this.


Testing logs:
`2024-08-21T20:21:39.118+0530    INFO    lotus-shed      lotus-shed/indexes.go:1148      failed to get block messages at height: 1867646

2024-08-21T20:21:39.122+0530    INFO    lotus-shed      lotus-shed/indexes.go:1148      failed to get block messages at height: 1867641

2024-08-21T20:21:39.127+0530    INFO    lotus-shed      lotus-shed/indexes.go:1148      failed to get block messages at height: 1867633

2024-08-21T20:21:39.134+0530    INFO    lotus-shed      lotus-shed/indexes.go:1148      failed to get block messages at height: 1867623`

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +900,339 @@ var backfillTxHashCmd = &cli.Command{
...
+	var totalRowsAffected int64
+	rowsAffected, err := res.RowsAffected()
+	if err != nil {
+		return xerrors.Errorf("error getting rows affected: %w", err)
+	}
+
+	totalRowsAffected += rowsAffected
+	log.Infof("pruned %s table, rows affected: %d", eventTable, rowsAffected)
+
+	res, err = tx.Exec(deleteEventEntriesByEventIDs, startHeight)
```
akaladarshi, line 1069: Here calling `res.RowsAffected()` is returning 0 even though in DB data is being deleted.

My guess is that SQLite not able to figure out how many rows are affected because there is a subquery inside it.

What are your thoughts on this @rvagg.

cmd/lotus-shed/indexes.go
```diff
@@ -682,68 +736,34 @@ var backfillMsgIndexCmd = &cli.Command{
...
-		api, closer, err := lcli.GetFullNodeAPI(cctx)
+		srvc, err := lcli.GetFullNodeServices(cctx)
 		if err != nil {
 			return err
 		}
 
-		defer closer()
-		ctx := lcli.ReqContext(cctx)
+		api := srvc.FullNodeAPI()
+		closer := srvc.Close
```
rvagg, line 754: maybe just keep the pattern the same across these, you have some churn in naming and you have an extra `closer` here; the one below in `pruneTxHashCmd` is clean:

```go
		srv, err := lcli.GetFullNodeServices(cctx)
		if err != nil {
			return err
		}
		defer srv.Close() //nolint:errcheck

		api := srv.FullNodeAPI()
		ctx := lcli.ReqContext(cctx)
```

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +900,339 @@ var backfillTxHashCmd = &cli.Command{
...
+
+			return nil
+		})
+
+		return g.Wait()
+	},
+}
+
+// pruneEventsIndex is a helper function that prunes the events index.db for a number of epochs starting from a specified height
+func pruneEventsIndex(_ context.Context, basePath string, startHeight int64) error {
```
rvagg, line 1027: just ditch the context arg if you have no way of using it, I know it's nice to have symmetry but if you don't need the symmetry for any tactical reason then just keep it clean

cmd/lotus-shed/indexes.go
```diff
@@ -36,11 +40,58 @@ const (
...
+	deleteTxHashIndexByCID = `DELETE FROM eth_tx_hashes WHERE cid = ?`
+
+	deleteEventFromStartHeight   = `DELETE FROM event WHERE height < ?;`
+	deleteEventEntriesByEventIDs = `DELETE FROM event_entry WHERE event_id IN (SELECT id FROM event WHERE height < ?);`
+	deleteEventsSeenByHeight     = `DELETE FROM events_seen WHERE height < ?;`
+)
+
+// database related constants
+const (
+	databaseType   = "sqlite"
```
rvagg, line 56: I found this naming confusing when parsing the path resolution below, perhaps something like `indexesDir`? i.e. it's the name of a directory and isn't used to refer to the type of database anywhere

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +900,339 @@ var backfillTxHashCmd = &cli.Command{
...
+	db, err := sql.Open(databaseDriver, dbPath+"?_txlock=immediate")
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		err := db.Close()
+		if err != nil {
+			fmt.Printf("ERROR: closing db: %s", err)
+		}
```
rvagg, line 1040: prefer compact form where you can get away with it, IMO it helps with Go's nasty eye-sprawl

```suggestion
		if err := db.Close(); err != nil {
			fmt.Printf("ERROR: closing db: %s", err)
		}
```

cmd/lotus-shed/indexes.go
rvagg: you just need to swap this with `deleteEventFromStartHeight` - this query relies on the entries being in the `event` table because of the sub-select, but you've just deleted the ones that it's trying to match; so clean up `event_entry` first, then `event` and you're all good

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +900,339 @@ var backfillTxHashCmd = &cli.Command{
...
+	}
+
+	var totalRowsAffected int64
+	rowsAffected, err := res.RowsAffected()
+	if err != nil {
+		return xerrors.Errorf("error getting rows affected: %w", err)
+	}
+
+	totalRowsAffected += rowsAffected
+	log.Infof("pruned %s table, rows affected: %d", eventTable, rowsAffected)
```
rvagg, line 1067: not critical: I suggest keeping track of these numbers separately and then logging _after_ you've committed the transaction, because until you do, they're not really gone; if you have a rollback then these log messages will be incorrect

rvagg: if you're concerned about progress logging, print a message before each execution: `log.Infof("pruning `%s` table", eventTable)`; then at least the user knows its working if this takes a long time for some reason (it shouldn't unless it's on a really slow disk)

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +900,339 @@ var backfillTxHashCmd = &cli.Command{
...
+
+	rowsAffected, err = res.RowsAffected()
+	if err != nil {
+		return xerrors.Errorf("error getting rows affected: %w", err)
+	}
+
+	err = tx.Commit()
+	if err != nil {
+		return xerrors.Errorf("error committing tx: %w", err)
+	}
```
rvagg, line 1097:
```suggestion
	if err = tx.Commit(); err != nil {
		return xerrors.Errorf("error committing tx: %w", err)
	}
```

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +900,339 @@ var backfillTxHashCmd = &cli.Command{
...
+	db, err := sql.Open(databaseDriver, dbPath)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		err := db.Close()
+		if err != nil {
+			fmt.Printf("ERROR: closing db: %s", err)
+		}
```
rvagg, line 1210:
```suggestion
		if err := db.Close(); err != nil {
			fmt.Printf("ERROR: closing db: %s", err)
		}
```

cmd/lotus-shed/indexes.go
rvagg: Oh, this one is tricky! because we don't have an epoch to delete against we can't do it bulk, as you've discovered, but we also can't do it this way because we kind of want the inverse: we want to delete entries for messages where we _don't_ have the message in the blockstore ourselves. So it's almost like: collect all messages, then delete everything not in that list. Which is impractical because that could get very large. The other problem with using the chain to do this is that there are going to be messages not on the canonical chain that we need to consider (i.e. there's been a reorg, so walking back from some tip doesn't lead us to tipsets that were previously considered canonical but are no longer), perhaps we want to include them, perhaps we should delete them because they are non-canonical, it's actually not even clear what you'd want to do because it depends on how the lookup API is being used. But conceptually what you'd want is something like: nothing in this txhashdb that isn't also in my blockstore. Because if I can't get to the original message anyway then there's no real help in being able to translate it from an eth hash.

BUT; just doing a quick look at the db, it's the one that we already have GC in already, and we achieve that by timestamp. If you turn on `EthTxHashMappingLifetimeDays` then it'll clean up according to the timestamp it was inserted at.

So I think the strategy needs to change for this table: we just can't do it on epoch unfortunately; we either need to not offer it here at all, or we need to come up with a timestamp and run something like the existing functionality: https://github.com/filecoin-project/lotus/blob/dbef5ded1585bcdcf84bbe09f3908210f9775ef6/chain/ethhashlookup/eth_transaction_hash_lookup.go#L131

We _can_ get a timestamp. So, if you want to keep this pruning here, you could keep the existing `api.ChainGetTipSetByHeight(ctx, abi.ChainEpoch(startHeight), types.EmptyTSK)` to get the tipset, look at the first block and get its timestamp (e.g. [like this](https://github.com/filecoin-project/lotus/blob/dbef5ded1585bcdcf84bbe09f3908210f9775ef6/blockstore/splitstore/splitstore_compact.go#L131) and then work out how many seconds ago it is and then you can run the delete using `datetime()` maths. Something like `DELETE FROM eth_tx_hashes WHERE insertion_time < datetime('now', '-X seconds')`. Note how sqlite datetime maths works:

```
sqlite> select datetime('now'), datetime('now', '-10 seconds'), datetime('now', '-86400 seconds');
2024-08-22 03:38:03|2024-08-22 03:37:53|2024-08-21 03:38:03
```

This is one area where we could apply some reuse of existing code: in `eth_transaction_hash_lookup.go`, we could expose a method to reuse the existing delete statement, and since we only run this once per hour on the existing GC we can get rid of the preparedstatement too and expose a public function to run it on any instance of a db, here's a quick go at what what may look like over in that file:

```go
func (ei *EthTxHashLookup) DeleteEntriesOlderThan(days int) (int64, error) {
	if ei.db == nil {
		return 0, xerrors.New("db closed")
	}
	epochs := abi.ChainEpoch(build.BlockDelaySecs * builtin.SecondsInDay * uint64(days))
	return DeleteEntriesOlderThan(ei.db, epochs)
}

// DeleteEntriesOlderThan deletes entries older than the given number of epochs from now.
func DeleteEntriesOlderThan(db *sql.DB, epochs abi.ChainEpoch) (int64, error) {
	secondsAgo := int64(epochs) * int64(build.BlockDelaySecs)
	res, err := db.Exec(deleteOlderThan, "-"+strconv.FormatInt(secondsAgo, 10)+" seconds")
	if err != nil {
		return 0, err
	}
	return res.RowsAffected()
}
```

cmd/lotus-shed/indexes.go
akaladarshi: As I mentioned the data is still being deleted but, it's not returning the number of rows affected.

Anyway, I will update it as you mentioned.

rvagg: But how do you know it's being deleted? you've lost the entried they join with on the `event` table, have you tried a `select count(*) from event` and `select count(*) from event_index` to see that they're actually deleted? Because of the subquery working on rows that were previously deleted, you should not have the join and therefore not get deletion.

cmd/lotus-shed/indexes.go
akaladarshi: Done.

cmd/lotus-shed/indexes.go
akaladarshi: Done.

cmd/lotus-shed/indexes.go
akaladarshi: Updated.

cmd/lotus-shed/indexes.go
akaladarshi: Done.

cmd/lotus-shed/indexes.go
akaladarshi: I didn't realise this until after I ran the node. Initially, I thought it was DB-type.

Anyway, I have updated it.

cmd/lotus-shed/indexes.go
akaladarshi: Done.

cmd/lotus-shed/indexes.go
akaladarshi: Done.

cmd/lotus-shed/indexes.go
akaladarshi: You're correct, I was checking it the wrong way.

cmd/lotus-shed/indexes.go
akaladarshi: Updated as per your suggestion of reusing code.

BigLep:
@akaladarshi : what are the next steps for this PR so it can be merged?

akaladarshi:
> @akaladarshi : what are the next steps for this PR so it can be merged?

@BigLep I am not sure about this PR as of now because we have removed the existing indexes in favour of #12421.

I think @aarshkshah1992 or @rvagg are the best people who can comment on this.

BigLep:
@aarshkshah1992 : should this be closed in light of ChainIndexer work?

BigLep:
I've requested review from you @aarshkshah1992 .  I assume this can be closed, but will let you comment.

PR #12405, opened 2024-08-19
fix(ci): don't PR or changelog check for draft PRs

rvagg:
I was just noticing in https://github.com/filecoin-project/lotus/pull/12388 how unnecessary the GHA spam is for a Draft, so this should prevent that until you mark as ready-for-review, I think.

### DIFF
--- a/.github/workflows/pr-title-check.yml
+++ b/.github/workflows/pr-title-check.yml
@@ -2,13 +2,19 @@ name: PR Title Check
 
 on:
   pull_request_target:
-    types: [opened, edited, synchronize, reopened]
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
+      - ready_for_review
 
 permissions:
   pull-requests: write
 
 jobs:
   check-pr-title:
+    if: github.event.pull_request.draft == false
     name: Check PR Title
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-title-check.yml
```diff
@@ -2,13 +2,19 @@ name: PR Title Check
 
 on:
   pull_request_target:
-    types: [opened, edited, synchronize, reopened]
+    types:
+      - opened
+      - edited
+      - synchronize
```
BigLep, line 8:
```suggestion
```

My understanding is that this is triggered when more commits are pushed.  We shouldn't need to rerun this notification just because there are more commits.

BigLep (approved):
Thanks for improving.

rvagg:
This is still running the PR check, is that because of `pull_request_target`?

BigLep:
> This is still running the PR check, is that because of pull_request_target?

I'm not quite following.  @rjan90 added pull_request_target so it runs when a contributor opens a PR from a fork.

I think the reason you keep getting alerted about this is because of the `syncrhonize` event (and that its removal hasn't been merged yet).  

(I'm no expert here though...)

rvagg:
My point being - I intentionally broke both rules here, no changelog and bad PR title, and it's a draft. But it's complaining about the PR title being wrong, and not the changelog. My expectation is that because I'm modifying the action itself that it would run my edits, but I'm wondering if `pull_request_target` means it runs the original action from master rather than my branch, while `pull_request` is running the version off my branch.

rjan90 (approved):

PR #12419, opened 2024-08-28
feat: nv24 skeleton

kamuik16:
## Related Issues
Closes #12351 

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title conforms with [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#pr-title-conventions)
- [x] Update CHANGELOG.md or signal that this change does not need it per [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#changelog-management)
- [x] New features have usage guidelines and / or documentation updates in
  - [x] [Lotus Documentation](https://lotus.filecoin.io)
  - [x] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [x] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12423, opened 2024-08-30
fix(test): Add OpenCL deps to `Cache Dependencies` workflow

rjan90:
## Related Issues
[Do not merge] 

The `Cache Dependencies` test is failing in https://github.com/filecoin-project/lotus/pull/12419.

## Checklist

- [ ] CI is green


rvagg:
Here's what I think is going on:

* This PR is still failing because you're installing too late, if you moved it up above ~line 219 then it should pass
* On line 216 we already do this by running .github/actions/install-system-dependencies which has these dependencies on it
* `restore_fetch_params` is `false` because we only care about proof params to determine whether to rebuild or not, only if extern/filecoin-ffi/parameters.json changes do we rebuild.
* `make deps` can be run if **either** `restore_fetch_params` or `restore_make_deps` is `true`, but `restore_make_deps` is `true` because it's looking at the extern/filecoin-ffi/ commit hash. But, we don't have dependencies to do this!

I think the fix here is to remove the block you added, but add in this to the `if` checks for both the `install-system-dependencies` (line 215) and `install-go` (line 217) jobs: ` || steps.restore_make_deps.outputs.cache-hit != 'true'`. That way, when we hit `make deps` because `restore_make_deps` is `true`, we have what we need.

BigLep:
@rjan90 : I assume we're using this PR to track this issue.  

Has all of Rod's feedback/suggestions been incorporated, or is there more we need to here?

BigLep:
FIL Slack conversations on this:
* IPDX: https://filecoinproject.slack.com/archives/C06CY12V83S/p1725272519636799
* fil-lotus-dev: https://filecoinproject.slack.com/archives/CP50PPW2X/p1725023492104539

.github/workflows/test.yml
```diff
@@ -186,7 +186,7 @@ jobs:
           echo -e "path<<EOF\n$CACHE_PATH\nEOF" | tee -a $GITHUB_OUTPUT
       - id: make_deps
         env:
-          CACHE_KEY: ${{ runner.os }}-${{ runner.arch }}-make-deps-${{ hashFiles('./.git/modules/extern/filecoin-ffi/HEAD') }}
+          CACHE_KEY: ${{ runner.os }}-${{ runner.arch }}-make-deps-${{ hashFiles('./.git/modules/extern/filecoin-ffi/HEAD') }}-portable
```
rvagg, line 189: when you change the build step, change this key slightly or else you'll end up using an older build, I stuck "-portable" on here for that reason

chain/actors/builtin/multisig/multisig.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/ipfs/go-cid"
+	"github.com/minio/blake2b-simd"
```
rvagg, line 7: how did this get in here? it breaks the build, kind of sus

rvagg:
https://github.com/filecoin-project/lotus/pull/12440 seems to address the problem here - instead of using the install-filcrypto script (invoked via the `make` chain) we use the build-release script in filecoin-ffi directly. Which means that there's some specific build steps in there that aren't available when we use the standard install-filcrypto.

Fixing it this way seems reasonable for now but it's a pretty manual hack for something that we should be able to do with `make` and an env var or two. So we should figure out how to address it in filecoin-ffi.

rvagg:
https://github.com/filecoin-project/filecoin-ffi/pull/480

rjan90:
I have fixed the itest-migration in https://github.com/filecoin-project/lotus/pull/12423/commits/742635a1d371286fb31a71ace9fb464692bfd540. Only thing missing before is:

- Cleaning up the commits bit so this can be rebase merged.
- Address the CodeQL failure: https://github.com/filecoin-project/lotus/pull/12423/checks?check_run_id=29992864615. I think it can be dismissed (iirc, this is something we see every time we generate the skeleton iirc). But I would like others to confirm.
- Include https://github.com/filecoin-project/lotus/pull/12451#issuecomment-2345059697 in here?

rvagg:
You could just fix the codeql check, in multisig/actor.go.template at the top of `txnParams`:

```go
	if id > uint64(math.MaxInt64) {
		return nil, xerrors.Errorf("transaction ID %d is out of range for int64", id)
	}
```

it's only used for actors v0 so it's pretty silly to have to mess with this each time, in fact just pulling in `msig{{.latestVersion}}` for it is silly! maybe we should just push that into message0.go and leave the rest alone

rvagg:
> Include https://github.com/filecoin-project/lotus/pull/12451#issuecomment-2345059697 in here?

nothing to include, that's all done here and that PR can be closed when this is merged

rvagg:
https://github.com/filecoin-project/lotus/pull/12454

rjan90:
> Cleaning up the commits a bit so this can be rebase merged.

I tried to remove a couple of the additions and reversion that were in consecutive commits. But rebasing this PR now is hard because there are recorded submodule modifications. So I think the easier path to get a clean commit history, is to cherry-pick the needed commits over to: https://github.com/filecoin-project/lotus/pull/12419

rjan90:
Closing as the relevant changes here has been added/cherry-picked into: #12455

PR #7420, opened 2021-10-01
Update proofs to v10.0.0

magik6k:
See https://github.com/filecoin-project/rust-fil-proofs/blob/master/CHANGELOG.md#1000---2021-09-30

This release should make computing proofs a lot faster, especially when using the new CUDA feature (`FFI_USE_CUDA=1 FFI_BUILD_FROM_SOURCE=1 ...`)

PR #12424, opened 2024-08-31
[skip changelog] chore: tests: fix failures on recent macos

ribasushi:
On modern darwin `os.ReadDir("/dev/fd")` always results in an error. Because of this storage/sealer/ffiwrapper/sealer_test.go invariably fails.

Instead of fixing the FD enumerator, remove the check entirely, as it has been disabled since https://github.com/filecoin-project/lotus/pull/7420/files

### DIFF
--- a/storage/sealer/ffiwrapper/sealer_test.go
+++ b/storage/sealer/ffiwrapper/sealer_test.go
@@ -278,8 +278,6 @@ func getGrothParamFileAndVerifyingKeys(s abi.SectorSize) {
 //
 // go test -run=^TestDownloadParams
 func TestDownloadParams(t *testing.T) {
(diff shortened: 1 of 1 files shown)


rvagg (approved):
heh, ok, we lose some logging here and also some nagging

@vmx should we expect any leaky fds? do you think this check might be ancient and irrelevant enough that complete removal is reasonable?

rvagg:
Yeah, the main thing about removing the logging for me is the loss of instutional knowledge that this _could be a problem_. With this code gone, I'm unaware that this ever is/was a thing. What I'd like to know is if this is a real problem now, or just historical baggage we can happily forget.

magik6k (approved):
This code has served it's purpose a long time ago when this was a problem

rvagg:
👌 thanks @magik6k

PR #12431, opened 2024-09-04
feat(eth): add EthSendRawTransactionUntrusted, use via gateway eth_sendRawTransaction

rvagg:
Also improves the test coverage for `MpoolPush*` variants.


Stebalien:
> panic: reflect.Set: value of type *api.Gateway is not assignable to type full.EthModuleAPI

rvagg:
Yeah, ok, much more complicated if we don't want to expose this on the `GatewayAPI`. So instead I've moved this new one from `EthModuleAPI` to `EthAPI`, but then it entangles us with some other dependencies so I had to reorg some other pieces to get it to all wire up properly. But this should be good now.

node/impl/full/eth.go
```diff
@@ -170,8 +169,10 @@ type EthAPI struct {
 	Chain        *store.ChainStore
 	StateManager *stmgr.StateManager
 
+	EthTxHashManager
```
Stebalien, line 172:
```suggestion
	EthTxHashManager EthTxHashManager
```

(no need to export these methods, same with the `MpoolAPI`)

Stebalien (approved):

Issue #12127, opened 2024-06-20
[DX Streamline] Establish flaky test eradication program


BigLep:
## Done Criteria
1. There is a metric that shows how much lotus is plagued by test flakiness.  Eradication of flaky tests should have an impact on this metric.
2. It is self-service for anyone to find this metric.
3. It is self-service for anyone to find flaky tests and how flaky they are.
4. There is a public SOP/runbook on how a maintainer should handle when a flaky test is discovered.  

## Why Important
Flaky tests are insidious.  We have them in our codebase, and it has two primary problems:
1. Slows down the velocity we can merge because judgement needs to be applied for whether it's safe to merge and we need to rely on those with admin permissions (e.g., maintainers) to merge.
2. It's compounding.  Once we have some, it's easier for more to slip in, which makes it harder to dig out later.

We want to have the standard of "we only merge if CI is green", and that is much easier bar to hold if flakiness has been eradicated.

## User/Customer
Maintainers and contributors.

## Notes
* This issue can be completed even before all flaky tests are eradicated.  This issue is done when we have credible visibility and plan to get there.
* It's acknowledged that handling flaky tests is not a new effort for Lotus development.  This issue is more about building on existing efforts and getting more robust so that we win this battle.
* Lotus has the ["impact/test-flakiness" label](https://github.com/filecoin-project/lotus/issues?q=is%3Aissue+label%3Aimpact%2Ftest-flakiness+) for historically identified flaky tests.
* Handling flaky tests is nothing unique to Lotus.  There is lots of prior art and tooling.  Some examples / discussion I'm aware of:
   * go-libp2p: they have a score badge on [their readme](https://github.com/libp2p/go-libp2p/blob/master/README.md) which links to deeper analysis at https://marcopolo.github.io/FlakyTests/
   * Observable notebooks have been created concerning this topic: https://observablehq.com/@observablehq/integration-test-flakiness 
   * Github projects: https://github.com/F-Secure/flaky-tests-detection
   * Articles on the topic:
      * https://engineering.fb.com/2020/12/10/developer-tools/probabilistic-flakiness/
      * https://dl.acm.org/doi/10.1145/3377813.3381370

BigLep:
I'm pasting in some comments that had from Slack DMs:

---
Valuable comments from @marcopolo who lead this effort for go-libp2p:

Jotting a couple thoughts:

1. Observable notebook: A fork from https://observablehq.com/@observablehq/integration-test-flakiness

[...]

PR #12434, opened 2024-09-05
fix(ci): modify JUnit output file to properly name packages for BuildPulse

rvagg:
When you go test ./itests/path_to_test.go then your binary is named "command-line-arguments" thanks to a Go quirk. This hack is to just fix that directly in the JUnit output before we upload it so we have proper naming.

Ref: https://github.com/filecoin-project/lotus/issues/12127


rjan90 (approved):

rvagg:
https://app.buildpulse.io/@filecoin-project/lotus/builds/41478876?q=TestStatefulCallHandling&filter=all

I don't think this worked

rvagg:
Oh, it does work, we just have two different things to play with here.

Screenshot 2024-09-05 at 4 26 22 PM

We have two variables to play with here, this fix changes the "classname" but we also have the testsuite name. See `name="command-line-arguments"` and `classname="command-line-arguments"`.

Should we just change them both? We have the opportunity to do something different in each of them. e.g. one could be "itest" and the other the name of the itest.

I think maybe just changing them both to the same might be enough.

Thoughts @rjan90 @BigLep ?

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites tests="4" failures="0" errors="0" time="29.431164">
        <testsuite tests="4" failures="0" time="23.593000" name="command-line-arguments" timestamp="2024-09-05T16:20:32+10:00">
                <properties>
                        <property name="go.version" value="go1.21.13 linux/amd64"></property>
                </properties>
                <testcase classname="command-line-arguments" name="TestGatewayWalletMsig" time="4.150000"></testcase>
                <testcase classname="command-line-arguments" name="TestGatewayMsigCLI" time="3.150000"></testcase>
                <testcase classname="command-line-arguments" name="TestGatewayRateLimits" time="11.590000"></testcase>
                <testcase classname="command-line-arguments" name="TestStatefulCallHandling" time="4.570000"></testcase>
        </testsuite>
</testsuites>
```

rvagg:
going with suite name "itests" and the classname the name of the itest file minus .go

rvagg:
Screenshot 2024-09-05 at 7 22 19 PM

That's better, now we have both fields occupied by something meaningful. `itests` is the actual package name anyway so it'll be consistent with the other tests. The other one is the filename, so it's a bit synthetic but quite informative.

Took me 4 `fixup!`s to get it right though!

BigLep (approved):
Thanks - this looks good to me.

A couple of thoughts:

> We have two variables to play with here

I think what you've done here is better than what we get from unit tests which don't have the `command-line-argument` problem.  Great.
Example: https://app.buildpulse.io/@filecoin-project/lotus/tests/22587403651
image

> Modifying the .xml file directly before uploading (rather than doing command-line foo before running `go test`

Per above, I do think this gives cleaner output in BuildPulse.  One callout is that command-line-arguments still shows up in the GitHub Action logs.

Example: 
https://github.com/filecoin-project/lotus/actions/runs/10721811712/job/29731273734
image

I assume this isn't a problem though as there hasn't been a complaint before and you have enough context about where the failure is since we have one job per itest file.

rvagg:
> One callout is that command-line-arguments still shows up in the GitHub Action logs

Yes and it still shows up when I do it locally as well, I just can't find a work-around for it because when you target a test file specifically you apparently lose the package information where the builder needs it; hence the need for this distant hack.

> better than what we get from unit tests

OK, so I've applied it a little bit more comprehensively so now it rewrites the suite name on all of them. But there's a catch - when it ends up executing `go test ./package/path/...` like it does for `unit-rest`, `unit-node`, etc., then the junit.xml output has a `<testsuite>` block for each Go package. Within those are the individual `<testcase>`s. If we rewrite the name of the suites such that they are duplicates, then BuildPulse discards what it thinks are duplicates (learned from trying it out). So we'd either need to reduce them all down to a single `<testsuite>` per run (which is either quite a complex regex or some cmdline xml filtering) or not have duplicates.

So instead, I've done this: the suite name is now trimmed of the package prefix `github.com/filecoin-project/lotus/` and this is replaced with the name of the test run, and we leave the class name as the package, so we end up with:

Screenshot 2024-09-06 at 1 31 13 PM

And they're slightly nicer when in the list view:

Screenshot 2024-09-06 at 1 31 45 PM

We could flip this, which would make the list view more verbose, but would make the title nicer in the per-test view.

I haven't played enough with it to know how useful these suite and class categorisations are. Fitting Java concepts to Go is a bit of a stretch here since the "class" isn't really the granulation we have.

@BigLep have you toyed in BuildPulse enough to have an opinion? Maybe we should ask their advice?

.github/workflows/test.yml
```diff
@@ -275,6 +275,21 @@ jobs:
...
+          # assigns the name and classname attributes to "command-line-arguments". Others get the
+          # package name for both.
+          if [[ "${{ matrix.name }}" == itest-* ]]; then
+            PACKAGE_NAME=$(basename "$PACKAGES" .go)
+            sed -i 's/ name="command-line-arguments"/ name="itests"/g' "$REPORTS_PATH/$NAME.xml"
+            sed -i 's/classname="command-line-arguments"/classname="'"$PACKAGE_NAME"'"/g' "$REPORTS_PATH/$NAME.xml"
+          else
+            sed -i 's# name="github.com/filecoin-project/lotus/\(.*\)"# name="'${{ matrix.name }}':\1"#g' "$REPORTS_PATH/$NAME.xml"
+          fi
+          cat "$REPORTS_PATH/$NAME.xml"
```
rvagg, line 292: check out the actions run output if you want to see what we're dealing with here

BigLep:
> @BigLep have you toyed in BuildPulse enough to have an opinion? Maybe we should ask their advice?

@rvagg : I think the improvements you made here are good and seem like the right tradeoffs.  Let's go with them, and adjust in future if needed.

.github/workflows/test.yml
BigLep: Ack - this makes sense to me.  I'll resolve this so we can merge.

PR #12439, opened 2024-09-10
chore(build): trim and compress genesis CARs in build

rvagg:
Ref: https://github.com/filecoin-project/lotus-infra/issues/1627
Ref: https://github.com/filecoin-project/lotus-infra/pull/1628

* Remove WASM actors from butterflynet.car
* Compress with zstd

Resulting `lotus` binary is 10.7MiB smaller, 251M -> 241M.

Draft for now because it needs to be confirmed working.


masih (approved):

rjan90:
Can confirm I was able to join a newly reset butterfly-network with a stripped out and compressed genesis CAR-file which was added as a Butterfly-asset in: https://github.com/filecoin-project/lotus/pull/12443.

Screenshot 2024-09-10 at 11 35 30

rvagg:
Cool, let's combine and merge then, although we probably want a changelog entry. Would you mind doing that in your PR @rjan90? I'm off for the night.

PR #12440, opened 2024-09-10
chore(ci): maximally portable ffi

rvagg:
another go at https://github.com/filecoin-project/lotus/pull/12423


BigLep (commented):
I'm not up on what's going on here, but will our "6 months from now" selves understand what's gong on, especially if we're doing something outside of how filecoin-ffi is usually consumed/built?

rvagg:
No, that's why we should solve this with https://github.com/filecoin-project/filecoin-ffi/pull/480

PR #12445, opened 2024-09-11
chore(deps): update go-state-types, adapt to QAP API changes

rvagg:
Updates go-state-types and incorporates https://github.com/filecoin-project/go-state-types/pull/308 which removes "deal weight" entirely from QAP calculation since it's unnecessary. So, we inherit a signature change here for `QAPowerForWeight` which drops the deal weight argument.

This is on top of https://github.com/filecoin-project/lotus/pull/12384 because it touches pledge calculation too. I'll leave this as draft until that merges. We also don't have a tagged gst with this change, and it might be best to leave this commit until @rjan90 cuts a second prerelease and pulls that in for nv24.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # UNRELEASED
 
+## Improvements
+
+- The miner actor builtin `QAPowerForWeight` no longer accepts the unused "dealWeight" parameter, the function signature now only takes 3 arguments: sectorSize, sectorDuration, verifiedWeight. ([filecoin-project/lotus#12445](https://github.com/filecoin-project/lotus/pull/12445))
+
 ## Bug Fixes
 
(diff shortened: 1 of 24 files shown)


rjan90:
>  We also don't have a tagged gst with this change, and it might be best to leave this commit until @rjan90 cuts a second prerelease and pulls that in for nv24.

I can cut GST v0.15.0-rc1 to make the GST update here a bit cleaner. However, since there is still some time before we start the Lotus nv24 train, I suggest leaving this unmerged until we have decided if we want to ship a Lotus minor version with the indexing revamp (pending discussion we will have in a weeks time or so).

BigLep:
@rvagg : 
1. Is the next step here to get @ZenGround0 to review?  If so, he will probably need a Slack ping.
2. Is there a certain lotus release you want to have this in?  (e.g., 1.32.0 for nv25)

rvagg:
@BigLep:

1. Yep, probably
2. Whatever I can get this in that's not a patch release, it's not a major change but it would be a minor breaking API change if someone's reaching in to the builtins package to use them directly

PR #12421, opened 2024-08-29
feat: a new ChainIndexer to index tipsets, messages and events

aarshkshah1992:
For https://github.com/filecoin-project/lotus/issues/12453.
Subsumes https://github.com/filecoin-project/lotus/pull/12388 to implement the `ChainIndexer`.

### TODO
- [x] Profile Sqlite Index usage  and read queries on the Index
- [x] Clean up Config
- [x] Smoke testing on a calibnet node (test GC, snapshot hydration, automated backfilling, etc)
- [x] Review from @rvagg 
- [x] Write unit & itests for the `ChainIndexer` (tests are at https://github.com/filecoin-project/lotus/pull/12521)

## Migration 

Index _"migration"_ will effectively involve re-indexing the chain state here. This is because the data in the old Indices isn't reliable (we're already aware of multiple bugs in the old Indices which motivated this PR in the first place) and the old Indices have never been gc'd/kept in sync with the chain state on nodes with splitstore enabled.

### For Snapshot synced nodes with Splitstore enabled

- Most splitstore enabled nodes only keep 2-3 days of chain state around. The best option here is to simply backfill/re-index the chain state on node startup by setting `ReConcileEmptyIndex` to true and setting the `MaxReconcileTipsets` as needed.

### For archival nodes

- If archival nodes want to backfill the index as part of daemon startup, they can always bootup the node with `ReConcileEmptyIndex` set to true and `MaxReconcileTipsets` set to infinity. This will drastically slow down the node startup  but will ensure that the node starts up with it's entire history indexed.

- The other option for users is to use the `Backfill` RPC API  with a range of height they want backfilled. However, we need to handle race conditions here where backfilling races with new tipsets coming in and need to ensure that backfilling does not end up overwriting updates caused by new tipsets being indexed.

### For looking up `MsgInfo`

```
~/.lotus-calibnet/sqlite% sqlite3 chainindex.db "EXPLAIN QUERY PLAN SELECT tipset_key_cid, height FROM tipset_message WHERE message_cid='a' AN
D reverted=0"
QUERY PLAN
`--SEARCH tipset_message USING INDEX idx_message_cid (message_cid=?)
```
- There should be a handful of rows with `reverted != 0` and only one with `reverted=0` for a given `message_cid` so no need to index by `reverted` as well here.

### For looking up an eth tx hash

```
[...]

PR #12447, opened 2024-09-11
feat: add support for eth_getBlockReceipts

### LINKED ISSUE #12429: Support eth_getBlockReceipts
I _think_ the output of this is the same as `eth_getTransactionReceipts` but you query for the block number instead of transaction hash and the result is just an array with all of the `eth_getTransactionReceipts` results.

This is an efficient way to get all details of a tipset rather than querying for all messages and constructing it yourself. Just need to be careful about doing proper de-duplication.

virajbhartiya:
## Related Issues
Fixes #12429  

## Proposed Changes
Implimented `eth_getBlockReceipts` method which accepts a block number and returns all the transaction receipts for it

## Additional Info
[QuickNode](https://www.quicknode.com/docs/ethereum/eth_getBlockReceipts)
[Infura](https://docs.infura.io/api/networks/blast/json-rpc-methods/eth_getblockreceipts)

## Checklist

- [ ] CI is green


CHANGELOG.md
```diff
@@ -6,7 +6,9 @@
 
 ## New features
 
-* Add `EthSendRawTransactionUntrusted` RPC method to be used for the gateway when accepting `EthSendRawTransaction` and `eth_sendRawTransaction`. Applies a tighter limit on the number of messages in the queue from a single sender and applies additional restrictions on nonce increments. ([filecoin-project/lotus#12431](https://github.com/filecoin-project/lotus/pull/12431))
+- Add `eth_getBlockReceipts` RPC method to fetch all the receipts for a given block. ([filecoin-project/lotus#12447](https://github.com/filecoin-project/lotus/pull/12447))
```
rvagg, line 9: check the changelog here, you've added a ton of lint-style changes in here - some of these additional changes are fine, but don't belong in this PR, maybe you have a markdown linter installed in your editor that's doing these? Keep the change limited to the single addition of an entry for the change you're making.

itests/eth_deploy_test.go
```diff
@@ -140,6 +140,12 @@ func TestDeployment(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, receipt)
 
+	// Verify that the receipt is correct.
+	receipts, err := client.EthGetBlockReceipts(ctx, ethtypes.EthBlockNumberOrHash{BlockHash: &receipt.BlockHash})
+	require.NoError(t, err)
+	require.Len(t, receipts, 1)
```
rvagg, line 146: I would love to be able to get a version of this with >1 transactions in a block, we have some other tests that do a multi-message submission and wait for them to land, it would be good if we could replicate that behaviour here in a non-flaky way, have a fish around and see if you can find something, otherwise let me know if you have trouble and I might be able to either find something similar or cook something up

node/impl/full/eth.go
```diff
@@ -2135,3 +2137,40 @@ func (g gasRewardSorter) Swap(i, j int) {
...
+	for i, msg := range messages {
+		msgLookup, err := a.StateAPI.StateSearchMsg(ctx, ts.Key(), msg.Cid(), api.LookbackNoLimit, true)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to search for message: %w", err)
+		}
+		if msgLookup == nil {
+			continue
+		}
+
+		tx, err := newEthTxFromMessageLookup(ctx, msgLookup, i, a.Chain, a.StateAPI)
```
rvagg, line 2162: I think a bunch of the cost in this call could be avoided by pulling up some of its functionality into this loop, or refactoring out a subset of that functionality that avoids loading the tipset again, loading the parent tipset to get the block hash and initialising the state tree. Those operations are all common across messages and you could do them once at the top of the loop and reuse inside, or use them to call a new function that's shared with `newEthTxFromMessageLookup`. `newEthTxFromSignedMessage` is the most important part of the call that you care about, most of the rest is common across all messages.

Something like this perhaps:

```diff
diff --git a/node/impl/full/eth_utils.go b/node/impl/full/eth_utils.go
index a9645ce9d..957b1bea0 100644
--- a/node/impl/full/eth_utils.go
+++ b/node/impl/full/eth_utils.go
@@ -643,19 +643,18 @@ func newEthTxFromMessageLookup(ctx context.Context, msgLookup *api.MsgLookup, tx
                }
        }
 
-       blkHash, err := ethtypes.EthHashFromCid(parentTsCid)
+       st, err := sa.StateManager.StateTree(ts.ParentState())
        if err != nil {
-               return ethtypes.EthTx{}, err
+               return ethtypes.EthTx{}, xerrors.Errorf("failed to load message state tree: %w", err)
        }
 
-       smsg, err := getSignedMessage(ctx, cs, msgLookup.Message)
-       if err != nil {
-               return ethtypes.EthTx{}, xerrors.Errorf("failed to get signed msg: %w", err)
-       }
+       return newEthTx(ctx, cs, st, parentTs.Height(), parentTsCid, msgLookup.Message, txIdx)
+}
 
-       st, err := sa.StateManager.StateTree(ts.ParentState())
+func newEthTx(ctx context.Context, cs *store.ChainStore, st *state.StateTree, blockHeight abi.ChainEpoch, parentTsCid cid.Cid, msgCid cid.Cid, txIdx int) (ethtypes.EthTx, error) {
+       smsg, err := getSignedMessage(ctx, cs, msgCid)
        if err != nil {
-               return ethtypes.EthTx{}, xerrors.Errorf("failed to load message state tree: %w", err)
+               return ethtypes.EthTx{}, xerrors.Errorf("failed to get signed msg: %w", err)
        }
 
        tx, err := newEthTxFromSignedMessage(smsg, st)
@@ -664,10 +663,15 @@ func newEthTxFromMessageLookup(ctx context.Context, msgLookup *api.MsgLookup, tx
        }
 
        var (
-               bn = ethtypes.EthUint64(parentTs.Height())
+               bn = ethtypes.EthUint64(blockHeight)
                ti = ethtypes.EthUint64(txIdx)
        )
 
+       blkHash, err := ethtypes.EthHashFromCid(parentTsCid)
+       if err != nil {
+               return ethtypes.EthTx{}, err
+       }
+
        tx.BlockHash = &blkHash
        tx.BlockNumber = &bn
        tx.TransactionIndex = &ti
```

And then you can use `newEthTx` from within this loop.

itests/fevm_test.go
```diff
@@ -1053,3 +1053,25 @@ func TestFEVMErrorParsing(t *testing.T) {
...
+
+	// Deploy a contract to generate a transaction
+	_, contractAddr := client.EVM().DeployContractFromFilename(ctx, "contracts/SimpleCoin.hex")
+
+	// Get the latest block
+	latestBlock, err := client.EthGetBlockByNumber(ctx, "latest", true)
+	require.NoError(t, err)
+
+	// Get block receipts
+	receipts, err := client.EthGetBlockReceipts(ctx, ethtypes.EthBlockNumberOrHash{BlockHash: &latestBlock.Hash})
```
rvagg, line 1069: this isn't working because nothing was invoked on "latest", it's probably even a null round

you need to trigger something to happen, then use the tipset that it was submitted on to get the receipts

it's probably easiest just to look for places where `EthGetTransactionReceipt` is currently used and piggy-back that

virajbhartiya:
I have made the changes mentioned above, I just couldn't figure out get more than 1 transaction block, I am researching a bit more on that.

CHANGELOG.md
```diff
@@ -3136,7 +3137,7 @@ Contributors
 
 # 1.16.1 / 2022-07-07
 
-This is an OPTIONAL PATCH releases for storage providers who have failed to publish `SubmitWindowedPoSt` due to out of gas error. The error log looks like `/wdpost_run.go:xxx	estimating gas	{"error": "estimating gas used: message execution failed: exit SysErrOutOfGas(7)...`.
```
rvagg, line 3139: still got some stray edits in the changelog, this line and down at 3187-

rvagg:
@virajbhartiya check out https://github.com/filecoin-project/lotus/pull/12269 - I think that what we want is already there, `TestMultipleEvents` has a loop to try and get 3 transactions into the same tipset. I think you should just be able to add something on the end of that test with `eth_getBlockReceipts`.

You also need to `go mod tidy` and possibly update your local submodules (`make clean deps` - although we currently don't have filecoin-ffi builds published [reason unknown yet] so if you do that then you're in for a manual Rust compile which is a bit tedious - alternatively back out the merge-main that you did so you don't pull in the commits on master that have the filecoin-ffi update).

node/impl/full/eth.go
```diff
@@ -2135,3 +2137,57 @@ func (g gasRewardSorter) Swap(i, j int) {
...
+		return nil, xerrors.Errorf("failed to get tipset: %w", err)
+	}
+
+	messages, err := a.Chain.MessagesForTipset(ctx, ts)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get messages for tipset: %w", err)
+	}
+
+	// Get the next tipset where the messages are actually executed
+	nextTs, err := a.Chain.GetTipsetByHeight(ctx, ts.Height()+1, ts, false)
```
rvagg, line 2153:
```suggestion
	nextTs, err := a.Chain.GetTipsetByHeight(ctx, ts.Height()+1, nil, false)
```

The tipset argument is intended to serve as a starting point, by providing `nil` you're saying "just find it for me starting from the head". Providing `ts` says "find me the _next_ tipset, by starting at this one and walking backward", which won't work.

BigLep:
I re-requested @rvagg review given https://filecoinproject.slack.com/archives/CP50PPW2X/p1726515846011179

rvagg:
Some feedback:

* We still have some CHANGELOG.md differences beyond just the addition of the new line, you now have some deletions. Perhaps try checking out the version from master to your own branch and re-inserting the single change manually (`git checkout origin/master -- CHANGELOG.md`).
* Same thing with go.sum, you've got some additional changes in there that are causing problems with dependencies for running the tests. Just check it out from master to overwrite your changes.
* As per the discussion on Slack, we're going to need to avoid the `Height()+1` thing: https://filecoinproject.slack.com/archives/CP50PPW2X/p1726457493521879

This is quite a diversion from the original task, that's my bad for not looking far enough into the future at the fact that we're being asked for the tipset with the messages but what we need is on the next tipset where those messages are executed. This is getting complicated, and I'm happy to guide you through the process but please let me know if this gets bigger in scope than what you're happy to work on and I can either hop in and contribute or take over.

The main problem we have is that this API can be asked for an arbitrary tipset. If it just took block "numbers" (epochs) then we might be able to get away with it, but it can also be a block hash, which is a tipset CID. Unfortunately, that tipset may not be on the canonical chain, it could be part of a reorg that shunts it off to a fork that's discarded, but it's still valid as long as we have it. So `Height()+1` is going to get us the canonical chain, but not necessarily the next tipset in the chain past the specific tipset that was requested.

So, what we need to do is re-execute the current tipset to get the next so that we have receipts and for execution of the messages we care about, and therefore the logs too. In most cases, it will already have been executed, and there is an execution cache, so it's going to be cheap most of the time. If you look in `EthGetBlockByHash` it does a similar thing. Trace the call trough `newEthBlockFromFilecoinTipSet` and you'll see an `executeTipset` which does what we need - essentially take a given tipset (which we have), execute its messages, get its receipts and then we have what we need to get the events.

Unfortunately it gets trickier from there because we need to go from the raw events to the eth logs. This might take some refactoring to expose the pieces we need and it's also going to be impacted a little by https://github.com/filecoin-project/lotus/pull/12421, but I think we can ignore that for now. In `EventFilterManager` we have a function [`loadExecutedMessages`](https://github.com/filecoin-project/lotus/blob/9c266bf96ba941fe8e5c3cf33fbbc28f6f51ac8a/chain/events/filter/event.go#L435) that can extract the raw events for us from a tipset. Back in eth.go we have `ethFilterLogsFromEvents` which can convert them to what we need for the output in eth form.

virajbhartiya:
I have reverted the changes made in `CHANGELOG.md` and `go.sum`, as per the discussion on Slack made the necessary changes to avoid using `Height()+1`. Please check it out.  

I'm trying to implement `loadExecutedMessages` in `EthGetBlockReceipts`. I got some ideas on how to implement it from your explanation. I'll give it a try. Please help out with it.

aarshkshah1992:
Hey @Kubuxu Why did we close this issue ?

aarshkshah1992:
The implementation for this has yet to land so re-opening it.

rvagg:
@virajbhartiya just retarget your PR to `master` and reopen the PR, we don't use a `main` branch in lotus but that's what your PR originally targeted.

rvagg:
nevermind, I see we are continuing in https://github.com/filecoin-project/lotus/pull/12478

PR #12457, opened 2024-09-13
chore(eth): minor optimisation in EthHashFromCid

rvagg:
Very minor, but it bothers me each time I have to look at it so I may as well fix it.

`BenchmarkEthHashFromCid-32       5629081               216.2 ns/op`

vs

`BenchmarkEthHashFromCid-32      29226618                41.58 ns/op`

### DIFF
--- a/chain/types/ethtypes/eth_types.go
+++ b/chain/types/ethtypes/eth_types.go
@@ -529,7 +529,13 @@ func handleHexStringPrefix(s string) string {
 }
 
 func EthHashFromCid(c cid.Cid) (EthHash, error) {
-	return ParseEthHash(c.Hash().HexString()[8:])
+	b := c.Hash()
+	if len(b) != EthHashLength+4 {
+		return EthHash{}, fmt.Errorf("CID hash length is not 32 bytes")
+	}
+	var h EthHash
+	copy(h[:], b[4:])
+	return h, nil
 }
 
 func ParseEthHash(s string) (EthHash, error) {
--- a/chain/types/ethtypes/eth_types.go
+++ b/chain/types/ethtypes/eth_types.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/ipfs/go-cid"
 	"github.com/multiformats/go-multihash"
+	mh "github.com/multiformats/go-multihash"
 	"github.com/multiformats/go-varint"
 	"golang.org/x/crypto/sha3"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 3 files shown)


rvagg:
1. I could return no error here, none is generated
2. I could do a check that there are actually 4 bytes here to avoid possible panic

rvagg:
`BenchmarkEthHashFromCid-32      32204678                37.94 ns/op`

Added proper error checking.

chain/types/ethtypes/eth_types.go
```diff
@@ -529,7 +529,13 @@ func handleHexStringPrefix(s string) string {
 }
 
 func EthHashFromCid(c cid.Cid) (EthHash, error) {
-	return ParseEthHash(c.Hash().HexString()[8:])
+	b := c.Hash()
+	if len(b) != EthHashLength+4 {
+		return EthHash{}, fmt.Errorf("CID hash length is not 32 bytes")
+	}
```
Stebalien, line 535: I'd still check to make sure it's the right type. Really, I'd:

1. Use `cid.Prefix` to compute the expected byte-prefix. Store that in a const/global var.
2. Use `c.Bytes()` to get the bytes.
3. Use `bytes.CutPrefix` to remove/check for this prefix.
4. Check that the remaining digest is 32 bytes long.

That'll avoid _all_ parsing.

Stebalien: Hm. I guess we weren't even checking that before...

Stebalien: E.g., see https://github.com/filecoin-project/lotus/blob/e0c6047bf20cbe9d68b595a9a4d0080c218d605b/chain/sub/incoming.go#L40-L47

rvagg: cool, I didn't know about `CutPrefix`, but also this does make it stricter than it currently is, we're OK with that? It may not make sense to do it any other way but you can currently feed it a CID with the wrong codec and hash function as long as the digest is what you want and it'll work.

rvagg:
Latest change, using `cid#Bytes` and `CutPrefix` gets us to almost nothing:

`BenchmarkEthHashFromCid-32      280525276                4.142 ns/op`

rvagg:
Reaching new heights of how many force pushes on a trivial change I can achieve in a very short period of time..

chain/types/ethtypes/eth_types.go
Stebalien: > we're OK with that?

IMO, the current lax-ness is a bug.

Stebalien (approved):

chain/types/ethtypes/eth_types.go
Stebalien: The tests disagree with me... is that an issue with the tests?

Stebalien: Yeah, we're using a fake CID.

rvagg: defaked

rvagg: well, better faked

PR #12460, opened 2024-09-13
feat(f3): checkpoint tipsets that are finalized by F3

masih:
## Related Issues
* Part of https://github.com/filecoin-project/go-f3/issues/603
* Depends on https://github.com/filecoin-project/go-f3/pull/642

## Proposed Changes
Once a decision is received from F3, checkpoint it in chain-store. As part of checkpointing, refactor the F3 tipset wrapper to reduce type casting.

## Checklist

- [x] CI is green

### DIFF
--- a/build/buildconstants/params_2k.go
+++ b/build/buildconstants/params_2k.go
@@ -200,3 +200,7 @@ var F3Enabled = true
 const ManifestServerID = "12D3KooWHcNBkqXEBrsjoveQvj6zDF3vK5S9tAfqyYaQF1LGSJwG"
 
 var F3BootstrapEpoch abi.ChainEpoch = 1000
+
+// F3Consensus set whether F3 should checkpoint tipsets finalized by F3. This
+// flag has no effect if F3 is not enabled.
+const F3Consensus = true
--- a/chain/lf3/ec.go
+++ b/chain/lf3/ec.go
@@ -30,6 +30,9 @@ type ecWrapper struct {
 	ChainStore   *store.ChainStore
 	Syncer       *chain.Syncer
 	StateManager *stmgr.StateManager
+
+	// Checkpoint sets whether to checkpoint tipsets finalized by F3 in ChainStore.
+	Checkpoint bool
 }
 
 type f3TipSet struct {
@@ -209,6 +212,9 @@ func (ec *ecWrapper) getPowerTableLotusTSK(ctx context.Context, tsk types.TipSet
 }
 
 func (ec *ecWrapper) Finalize(ctx context.Context, key gpbft.TipSetKey) error {
+	if !ec.Checkpoint {
+		return nil // Nothing to do; checkpointing is not enabled.
+	}
 	tsk, err := toLotusTipSetKey(key)
 	if err != nil {
 		return err
@@ -242,9 +248,5 @@ func (ec *ecWrapper) getTipSetFromF3TSK(ctx context.Context, key gpbft.TipSetKey
 }
 
 func toLotusTipSetKey(key gpbft.TipSetKey) (types.TipSetKey, error) {
-	tsk, err := types.TipSetKeyFromBytes(key)
-	if err != nil {
-		return types.TipSetKey{}, xerrors.Errorf("decoding tpiset key: %w", err)
-	}
-	return tsk, nil
+	return types.TipSetKeyFromBytes(key)
 }
--- a/chain/lf3/ec.go
+++ b/chain/lf3/ec.go
@@ -12,6 +12,7 @@ import (
 	"github.com/filecoin-project/go-f3/gpbft"
 	"github.com/filecoin-project/go-state-types/abi"
 
+	"github.com/filecoin-project/lotus/chain"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/power"
 	"github.com/filecoin-project/lotus/chain/stmgr"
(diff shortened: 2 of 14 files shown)


chain/lf3/ec.go
```diff
@@ -20,52 +20,55 @@ import (
...
 }
 
-var _ ec.TipSet = (*f3TipSet)(nil)
-
-type f3TipSet types.TipSet
-
-func (ts *f3TipSet) cast() *types.TipSet {
-	return (*types.TipSet)(ts)
+type f3TipSet struct {
+	*types.TipSet
```
Kubuxu, line 36: I don't love the pointer chasing but 🤷 doesn't matter much.

masih: Happy to revert it. I thought it reads better than casting but that's entirely subjective on my part.

Kubuxu: it reads better for sure

chain/lf3/ec.go
```diff
@@ -219,3 +203,36 @@ func (ec *ecWrapper) getPowerTableLotusTSK(ctx context.Context, tsk types.TipSet
 	sort.Sort(powerEntries)
 	return powerEntries, nil
 }
+
+func (ec *ecWrapper) Finalize(ctx context.Context, key gpbft.TipSetKey) error {
+	ts, err := ec.getTipSetFromF3TSK(ctx, key)
+	if err != nil {
+		return err
+	}
+	if err = ec.ChainStore.SetCheckpoint(ctx, ts); err != nil {
```
Stebalien, line 212: Looking at the current `SetCheckpoint` code, we have a few issues:

1. `SetCheckpoint` won't sync. I think we need to check if we have the tipset and, if not, call `Sync` (via the syncer) first. I.e., if `getTipSEtFromF3TSK` fails, we need to sync.
2. `SetCheckpoint` will (currently) refuse to work if it's older than finality. It claims that doing so would be "too slow but... tbh, I think we should just re-write that code to call `GetTipsetByHeight`. Then it'll get to use the tipset cache/index and should be pretty fast. Otherwise, we'll have problems when bootstrapping F3: if bootstrap fails for a few rounds then tries to agree on base, it'll fail to finalize base because base will be more than finality epochs in the past.

masih: Further to 1:1 discussion, used `Syncer.SyncCheckpoint` to work around the potential need to sync missing tipsets and captured a separate issue to improve checkpinting performance by using `GetTipsetByHeight` [here](https://github.com/filecoin-project/lotus/issues/12475). 🙏

chain/lf3/ec.go
```diff
@@ -20,52 +21,56 @@ import (
...
+	case header == nil, len(header.BeaconEntries) == 0:
+		// This should never happen in practice, but set beacon to a non-nil 32byte slice
+		// to force the message builder to generate a ticket. Otherwise, messages that
+		// require ticket, i.e. CONVERGE will fail validation due to the absence of
+		// ticket. This is a convoluted way of doing it.
+
+		// TODO: investigate if this is still necessary, or how message builder can be
+		//       adapted to behave correctly regardless of beacon value, e.g. fail fast
+		//       instead of building CONVERGE with empty beacon.
+		return emptyBeacon
```
Stebalien, line 63: I'd just... re-create this every time. Safer and we shouldn't hit this case anyways.

chain/lf3/ec.go
```diff
@@ -75,57 +80,38 @@ func (ec *ecWrapper) GetTipsetByEpoch(ctx context.Context, epoch int64) (ec.TipS
...
 	}
 
-	return wrapTS(ts), nil
+	return &f3TipSet{TipSet: ts}, nil
 }
 
-func (ec *ecWrapper) GetHead(_ context.Context) (ec.TipSet, error) {
-	return wrapTS(ec.ChainStore.GetHeaviestTipSet()), nil
+func (ec *ecWrapper) GetHead(context.Context) (ec.TipSet, error) {
+	return &f3TipSet{TipSet: ec.ChainStore.GetHeaviestTipSet()}, nil
```
Stebalien, line 96: IIRC, there's an issue where `GetHeaviestTipSet` can return nil if we have some form of datastore issue? I'd handle that case here and return an error, just to be safe.

chain/lf3/ec.go
```diff
@@ -219,3 +205,44 @@ func (ec *ecWrapper) getPowerTableLotusTSK(ctx context.Context, tsk types.TipSet
...
+	if err != nil {
+		return nil, xerrors.Errorf("getting tipset from key: %w", err)
+	}
+	return ts, nil
+}
+
+func toLotusTipSetKey(key gpbft.TipSetKey) (types.TipSetKey, error) {
+	tsk, err := types.TipSetKeyFromBytes(key)
+	if err != nil {
+		return types.TipSetKey{}, xerrors.Errorf("decoding tpiset key: %w", err)
```
Stebalien, line 245: Does `TipSetKeyFromBytes` not annotate the error?

masih: It passes up the cid.CidFromBytes error which may be too generic for debugging purposes. Happy to remove this if you think it is superfluous.

Kubuxu (commented):
We need to disable the finalize on the mainnet network right now.

build/buildconstants/params_2k.go
```diff
@@ -200,3 +200,7 @@ var F3Enabled = true
 const ManifestServerID = "12D3KooWHcNBkqXEBrsjoveQvj6zDF3vK5S9tAfqyYaQF1LGSJwG"
 
 var F3BootstrapEpoch abi.ChainEpoch = 1000
+
+// F3CheckpointFinalizedTipSet set whether F3 should checkpoint tipsets finalized
+// by F3. This flag has no effect if F3 is not enabled.
+const F3CheckpointFinalizedTipSet = true
```
Stebalien, line 206: Nit: I'd just call this `F3Consensus`.

chain/lf3/ec.go
Stebalien: Maybe modify `TipSetKeyFromBytes`?

Stebalien (approved):

PR #12450, opened 2024-09-11
feat: migration("re-indexing"), backfilling and diasgnostics tooling for the `ChainIndexer`

aarshkshah1992:
# ChainIndexer Migration and Diagnostics Tooling

This PR implements the _"migration"_ (really _re-indexing_ / _backfilling_), and diagnostics tooling for the `ChainIndexer` implemented in [PR #12450](https://github.com/filecoin-project/lotus/pull/12450), and is part of the work for  [#12453](https://github.com/filecoin-project/lotus/issues/12453). This tooling takes the form of both RPC APIs on the daemon and `lotus-shed` CLI commands.

## Re-indexing Process

The re-indexing tool enables clients to index their entire existing ChainState in the `ChainIndexer`. This process is necessary due to the removal of the existing `MsgIndex`, `EthTxIndex`, and `EventIndex` from Lotus.

### Why Re-index Instead of Migrate?

We've chosen to re-index rather than migrate data from existing indices for two primary reasons:

1. Known issues: The existing indices have multiple known problems, and migrating could perpetuate incorrect index entries.
2. Lack of garbage collection: Existing indices contain many entries for which the corresponding tipset messages/events no longer exist in the ChainStore due to splitstore GC.

Instead, we're re-indexing the `Chainstore`/`Chainstate` on the node into the `ChainIndexer`. This ensures that all re-indexed entries have gone through the indexing logic of the new `ChainIndexer` and that the Index is in sync/reflects the actual contents of the `Chainstore`/`Chainstate` post re-indexing.

## Diagnostics Tooling

This PR introduces diagnostic tools for detecting corrupt Index entries at specific epochs or epoch ranges.

While this PR implements functionality for optionally backfilling missing Index entries, it does not yet include the capability to "repair" corrupted Indexed entries. The repair functionality will be introduced in a subsequent PR. This approach allows us to first gather and analyze user reports, helping us understand the types and causes of corrupted Indexed entries(and if all they exist in the new `ChainIndexer`) before implementing repair mechanisms.

## Core API

The fundamental building block for this tooling is the following RPC API:

```go
type IndexValidation struct {
	TipsetKey string
	Height    uint64

	TotalMessages  uint64
	TotalEvents    uint64
	EventsReverted bool

	Backfilled bool
}

func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.ChainEpoch, backfill bool) (*types.IndexValidation, error)
```

_**This API has the following features:**_
[...]

PR #12462, opened 2024-09-14
feat: chain indexer todos [skip changelog]

akaladarshi:
This PR is part of ChainIndexer tooling [PR](https://github.com/filecoin-project/lotus/pull/12450):
 - It addresses the TODO's and comments

### DIFF
--- a/chain/index/api.go
+++ b/chain/index/api.go
@@ -57,7 +57,7 @@ func (si *SqliteIndexer) getTipsetCountsAtHeight(ctx context.Context, height abi
 }
 
 func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.ChainEpoch, backfill bool) (*types.IndexValidation, error) {
-	if !si.started && si.closed {
+	if !si.started && si.isClosed() {
 		return nil, xerrors.Errorf("ChainValidateIndex can only be called after the indexer has been started and not closed")
 	}
 
@@ -131,56 +131,53 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
 		return nil, xerrors.Errorf("index corruption: non-reverted tipset at height %d has key %s, but canonical chain has %s", epoch, indexedTsKeyCid, expectedTsKeyCid)
 	}
 
-	indexedNonRevertedMsgCount, indexedNonRevertedEventsCount, hasRevertedEvents, err := si.getIndexedTipSetData(ctx, expectedTs.Key())
+	// indexedTsKeyCid and expectedTsKeyCid are the same, so we can use `expectedTs` to fetch the indexed data
+	indexedData, err := si.getIndexedTipSetData(ctx, expectedTs.Key())
 	if err != nil {
 		return nil, xerrors.Errorf("failed to get tipset message and event counts at height %d: %w", expectedTs.Height(), err)
 	}
 
-	if err = si.verifyIndexedData(ctx, expectedTs, indexedNonRevertedMsgCount); err != nil {
+	if err = si.verifyIndexedData(ctx, expectedTs, indexedData); err != nil {
 		return nil, xerrors.Errorf("failed to verify indexed data at height %d: %w", expectedTs.Height(), err)
 	}
 
 	return &types.IndexValidation{
 		TipsetKey:      expectedTs.Key().String(),
 		Height:         uint64(expectedTs.Height()),
-		TotalMessages:  indexedNonRevertedMsgCount,
-		TotalEvents:    indexedNonRevertedEventsCount,
-		EventsReverted: hasRevertedEvents,
+		TotalMessages:  uint64(indexedData.nonRevertedMessageCount),
+		TotalEvents:    uint64(indexedData.nonRevertedEventCount),
+		EventsReverted: indexedData.hasRevertedEvents,
 	}, nil
 }
 
-// verifyIndexedData verifies that the indexed data for a tipset is correct
-// by comparing the number of messages in the chainstore to the number of messages indexed
-
-// TODO: verify indexed events too (to verify the events we need to load the next tipset (ts+1) and verify the events are the same)
-func (si *SqliteIndexer) verifyIndexedData(ctx context.Context, ts *types.TipSet, indexedMsgCount uint64) (err error) {
-	msgs, err := si.cs.MessagesForTipset(ctx, ts)
-	if err != nil {
-		return xerrors.Errorf("failed to get messages for tipset: %w", err)
-	}
-
-	msgCount := uint64(len(msgs))
-	if msgCount != indexedMsgCount {
-		return xerrors.Errorf("tipset message count mismatch: chainstore has %d, index has %d", msgCount, indexedMsgCount)
-	}
-
-	return nil
+type indexedTipSetData struct {
+	nonRevertedMessageCount int
+	nonRevertedEventCount   int
+	hasRevertedEvents       bool
 }
 
-func (si *SqliteIndexer) getIndexedTipSetData(ctx context.Context, tsKey types.TipSetKey) (messageCount uint64, eventCount uint64, hasRevertedEvents bool, err error) {
-	tsKeyBytes := tsKey.Bytes()
+// getIndexedTipSetData fetches the indexed tipset data for a tipset
+func (si *SqliteIndexer) getIndexedTipSetData(ctx context.Context, tsKey types.TipSetKey) (*indexedTipSetData, error) {
+	cid, err := tsKey.Cid()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get tipset key cid: %w", err)
+	}
+	tsKeyBytes := cid.Bytes()
 
+	var data indexedTipSetData
 	err = withTx(ctx, si.db, func(tx *sql.Tx) error {
-		if err = tx.Stmt(si.stmts.getNonRevertedTipsetMessageCountStmt).QueryRowContext(ctx, tsKeyBytes).Scan(&messageCount); err != nil {
+		if err = tx.Stmt(si.stmts.getNonRevertedTipsetMessageCountStmt).QueryRowContext(ctx, tsKeyBytes).Scan(&data.nonRevertedMessageCount); err != nil {
 			return xerrors.Errorf("failed to query non reverted message count: %w", err)
 		}
 
-		if err = tx.Stmt(si.stmts.getNonRevertedTipsetEventCountStmt).QueryRowContext(ctx, tsKeyBytes).Scan(&eventCount); err != nil {
+		if err = tx.Stmt(si.stmts.getNonRevertedTipsetEventCountStmt).QueryRowContext(ctx, tsKeyBytes).Scan(&data.nonRevertedEventCount); err != nil {
 			return xerrors.Errorf("failed to query non reverted event count: %w", err)
 		}
 
-		if eventCount > 0 {
-			if err = tx.Stmt(si.stmts.hasRevertedEventsStmt).QueryRowContext(ctx, tsKeyBytes).Scan(&hasRevertedEvents); err != nil {
+		// we are only fetching non-reverted events, so if there are no non-reverted events,
+		// then we need to check if the events were reverted.
+		if data.nonRevertedEventCount == 0 {
+			if err = tx.Stmt(si.stmts.hasRevertedEventsStmt).QueryRowContext(ctx, tsKeyBytes).Scan(&data.hasRevertedEvents); err != nil {
 				return xerrors.Errorf("failed to check for reverted events: %w", err)
 			}
 		}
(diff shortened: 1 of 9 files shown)


chainindex/api.go
```diff
@@ -57,9 +57,10 @@ func (si *SqliteIndexer) getTipsetCountsAtHeight(ctx context.Context, height abi
 }
 
 func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.ChainEpoch, backfill bool) (*types.IndexValidation, error) {
-	if !si.started {
-		return nil, xerrors.Errorf("ChainValidateIndex can only be called after the indexer has been started")
+	if !si.started && si.closed {
```
aarshkshah1992, line 60: You need to take a read lock before reading this. See how we do it elsewhere.

chainindex/api.go
```diff
@@ -126,17 +127,71 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
+	msgCount := uint64(len(msgs))
+	if msgCount != indexedMsgCount {
+		return xerrors.Errorf("tipset message count mismatch: chainstore has %d, index has %d", msgCount, indexedMsgCount)
+	}
+
+	return nil
+}
+
+func (si *SqliteIndexer) getIndexedTipSetData(ctx context.Context, tsKey types.TipSetKey) (messageCount uint64, eventCount uint64, hasRevertedEvents bool, err error) {
+	tsKeyBytes := tsKey.Bytes()
```
aarshkshah1992, line 171: let's return a struct here with these fields instead of returning multiple values here.

aarshkshah1992, line 171: This is incorrect. You need to use the tipsetKey.Cid.Bytes path here to get the key to look up the DB.

chainindex/api.go
```diff
@@ -126,17 +127,71 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
+		TotalMessages:  indexedNonRevertedMsgCount,
+		TotalEvents:    indexedNonRevertedEventsCount,
+		EventsReverted: hasRevertedEvents,
 	}, nil
 }
 
+// verifyIndexedData verifies that the indexed data for a tipset is correct
+// by comparing the number of messages in the chainstore to the number of messages indexed
+
+// TODO: verify indexed events too (to verify the events we need to load the next tipset (ts+1) and verify the events are the same)
```
aarshkshah1992, line 155: Please can you implement this TODO to verify events as well ?

aarshkshah1992: You can use the `loadExecutedMessages` function we have for this. Then you can just read messages and events together.

chainindex/api.go
```diff
@@ -126,17 +127,71 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
+		if err = tx.Stmt(si.getNonRevertedTipsetMessageCountStmt).QueryRowContext(ctx, tsKeyBytes).Scan(&messageCount); err != nil {
+			return xerrors.Errorf("failed to query non reverted message count: %w", err)
+		}
+
+		if err = tx.Stmt(si.getNonRevertedTipsetEventCountStmt).QueryRowContext(ctx, tsKeyBytes).Scan(&eventCount); err != nil {
+			return xerrors.Errorf("failed to query non reverted event count: %w", err)
+		}
+
+		// if there are events, we need to check if there are reverted events
+		if eventCount > 0 {
```
aarshkshah1992, line 183: @akaladarshi I think this is incorrect. `eventCount ` will be >0 ONLY if non-reverted events exist. So, we should check for reverted events ONLY if `eventCount`==0.

chainindex/api.go
akaladarshi: I was little skeptical about events, because we need to do something like `currTipset.Height() + 1` to get execution tipset then fetch data for it through `loadExecuteMessages`.

So should I go ahead with this?

aarshkshah1992: @akaladarshi 

Hmm..good point.

Yeah let's do it. So, here are the steps:

1) Load tipset `tsX` at user supplied epoch `e`
2) Load tipset `tsY` at epoch `e+1`
3) Ensure parent of ``tsY` == `tsX` (if there is a mismatch here, return an error because it means chain has forked and user needs to retry)
4) Load events and messages and validate the counts.

chainindex/api.go
akaladarshi: Done.

chainindex/api.go
akaladarshi: Done.

akaladarshi: Done.

chainindex/api.go
akaladarshi: Updated.
Added a function `isClosed`, so we can reuse it.

chain/index/ddls.go
```diff
@@ -82,5 +82,8 @@ func preparedStatementMapping(ps *preparedStatements) map[**sql.Stmt]string {
 		&ps.hasNullRoundAtHeightStmt:              "SELECT NOT EXISTS(SELECT 1 FROM tipset_message WHERE height = ?)",
 		&ps.getNonRevertedTipsetAtHeightStmt:      "SELECT tipset_key_cid FROM tipset_message WHERE height = ? AND reverted = 0",
 		&ps.countTipsetsAtHeightStmt:              "SELECT COUNT(CASE WHEN reverted = 1 THEN 1 END) AS reverted_count, COUNT(CASE WHEN reverted = 0 THEN 1 END) AS non_reverted_count FROM (SELECT tipset_key_cid, MAX(reverted) AS reverted FROM tipset_message WHERE height = ? GROUP BY tipset_key_cid) AS unique_tipsets",
+		&ps.getNonRevertedTipsetMessageCountStmt:  "SELECT COUNT(*) FROM tipset_message WHERE tipset_key_cid = ? AND reverted = 0",
+		&ps.getNonRevertedTipsetEventCountStmt:    "SELECT COUNT(*) FROM event WHERE message_id IN (SELECT message_id FROM tipset_message WHERE tipset_key_cid = ? AND reverted = 0)",
```
aarshkshah1992, line 86: @akaladarshi What if the tipset is NOT reverted but the events are reverted ? This query only ensures that the tipset is NOT reverted. Does it have any implications on the correctness of the code /API ?

chain/index/api.go
```diff
@@ -57,9 +57,10 @@ func (si *SqliteIndexer) getTipsetCountsAtHeight(ctx context.Context, height abi
 }
 
 func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.ChainEpoch, backfill bool) (*types.IndexValidation, error) {
-	if !si.started {
-		return nil, xerrors.Errorf("ChainValidateIndex can only be called after the indexer has been started")
+	if !si.started && si.isClosed() {
```
aarshkshah1992, line 60: OR closed. Not AND, right ?

chain/index/api.go
```diff
@@ -126,17 +127,124 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
 		return nil, xerrors.Errorf("failed to cast tipset key cid: %w", err)
 	}
 
-	if indexedTsKeyCid != expectedTsKeyCid {
+	if !indexedTsKeyCid.Equals(expectedTsKeyCid) {
 		return nil, xerrors.Errorf("index corruption: non-reverted tipset at height %d has key %s, but canonical chain has %s", epoch, indexedTsKeyCid, expectedTsKeyCid)
 	}
 
+	// indexedTsKeyCid and expectedTsKeyCid are the same, so we can use `expectedTs` to fetch the indexed data
+	indexedData, err := si.getIndexedTipSetData(ctx, expectedTs.Key())
```
aarshkshah1992, line 135: Just pass ` expectedTs` here and use the toTipsetKeyCidBytes function we have to find the database key.

chain/index/api.go
```diff
@@ -57,9 +57,10 @@ func (si *SqliteIndexer) getTipsetCountsAtHeight(ctx context.Context, height abi
 }
 
 func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.ChainEpoch, backfill bool) (*types.IndexValidation, error) {
-	if !si.started {
-		return nil, xerrors.Errorf("ChainValidateIndex can only be called after the indexer has been started")
+	if !si.started && si.isClosed() {
+		return nil, xerrors.Errorf("ChainValidateIndex can only be called after the indexer has been started and not closed")
```
aarshkshah1992, line 61: have different if statements for both conditions so the error message can be more granular.

chain/index/api.go
```diff
@@ -126,17 +127,124 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
+		TotalMessages:  uint64(indexedData.nonRevertedMessageCount),
+		TotalEvents:    uint64(indexedData.nonRevertedEventCount),
+		EventsReverted: indexedData.hasRevertedEvents,
 	}, nil
 }
 
+type indexedTipSetData struct {
+	nonRevertedMessageCount int
+	nonRevertedEventCount   int
+	hasRevertedEvents       bool
```
aarshkshah1992, line 156: Can you add a test for these fields to `TestEthGetLogsBasic` in `eth_filtest_test.go` ?

We already call this API there and so this should be easy. See 

```
iv, err := client.ChainValidateIndex(ctx, abi.ChainEpoch(0), false)
	require.NoError(err)
	require.NotNil(iv)

	fmt.Printf("index validation: %v\n", iv)

	iv, err = client.ChainValidateIndex(ctx, abi.ChainEpoch(22), false)
	require.NoError(err)
	require.NotNil(iv)
	fmt.Printf("index validation: %v\n", iv)
```

chain/index/api.go
```diff
@@ -159,9 +267,17 @@ func (si *SqliteIndexer) backfillMissingTipset(ctx context.Context, ts *types.Ti
...
 	return &types.IndexValidation{
-		TipsetKey:  ts.Key().String(),
-		Height:     uint64(ts.Height()),
-		Backfilled: true,
+		TipsetKey:      ts.Key().String(),
+		Height:         uint64(ts.Height()),
+		Backfilled:     true,
+		TotalMessages:  indexedData.nonRevertedMessageCount,
+		TotalEvents:    indexedData.nonRevertedEventCount,
+		EventsReverted: indexedData.hasRevertedEvents,
```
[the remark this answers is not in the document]
akaladarshi: you mean rename the `IndexValidation` fields?

chain/index/api.go
```diff
@@ -126,17 +127,124 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
+	if !indexedTsKeyCid.Equals(expectedTsKeyCid) {
 		return nil, xerrors.Errorf("index corruption: non-reverted tipset at height %d has key %s, but canonical chain has %s", epoch, indexedTsKeyCid, expectedTsKeyCid)
 	}
 
+	// indexedTsKeyCid and expectedTsKeyCid are the same, so we can use `expectedTs` to fetch the indexed data
+	indexedData, err := si.getIndexedTipSetData(ctx, expectedTs.Key())
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get tipset message and event counts at height %d: %w", expectedTs.Height(), err)
+	}
+
```
aarshkshah1992, line 139: ensure `indexedData` is not nil before proceeding here.

chain/index/api.go
```diff
@@ -126,17 +127,124 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
+		return xerrors.Errorf("failed to get tipset key cid: %w", err)
+	}
+
+	msgs, err := si.cs.MessagesForTipset(ctx, ts)
+	if err != nil {
+		return xerrors.Errorf("failed to get messages for tipset: %w", err)
+	}
+
+	msgCount := len(msgs)
+	if msgCount != indexedData.nonRevertedMessageCount {
```
aarshkshah1992, line 206: You don't need to call `MessagesForTipset` here and compare number of messages since you are calling `loadExecutedMessages` below, right ? You can just compare then.

chain/index/api.go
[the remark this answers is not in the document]
aarshkshah1992: Yes. Let's remove `EventsReverted` from `IndexValidation`. We can add it later if we need it.

chain/index/api.go
aarshkshah1992: Look at what the `loadExecutedMessages ` in it's implementation. Reading the implementation code will give you a better  understanding of how it all works together.

chain/index/api.go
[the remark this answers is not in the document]
akaladarshi: Then, I think we can also remove the query for checking, if the events were reverted.

```go

if data.nonRevertedEventCount == 0 {
	if err = tx.Stmt(si.stmts.hasRevertedEventsStmt).QueryRowContext(ctx, tsKeyCidBytes).Scan(&data.hasRevertedEvents); err != nil {
		return xerrors.Errorf("failed to check for reverted events: %w", err)
	}
}

```

chain/index/api.go
akaladarshi: Just checked, you're correct it's the same thing, I will update.

chain/index/api.go
akaladarshi: Ideally, it will only be nil if error is there and if there is error and `indexedData` is not nil then that means issue is in DB, so we should return.

chain/index/api.go
```diff
@@ -126,17 +127,124 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
+func (si *SqliteIndexer) getIndexedTipSetData(ctx context.Context, tsKey types.TipSetKey) (*indexedTipSetData, error) {
+	cid, err := tsKey.Cid()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get tipset key cid: %w", err)
+	}
+	tsKeyBytes := cid.Bytes()
+
+	var data indexedTipSetData
+	err = withTx(ctx, si.db, func(tx *sql.Tx) error {
+		if err = tx.Stmt(si.stmts.getNonRevertedTipsetMessageCountStmt).QueryRowContext(ctx, tsKeyBytes).Scan(&data.nonRevertedMessageCount); err != nil {
```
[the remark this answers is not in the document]
akaladarshi: I was using transaction so I get the consistent view of DB for querying.

chain/index/api.go
aarshkshah1992: Yes but no harm in checking it and returning an error of it is nil here so user's node dosen't crash if there exists a bug that causes a nil value here.

chain/index/api.go
akaladarshi: Done.

chain/index/api.go
akaladarshi: Done.

chain/index/api.go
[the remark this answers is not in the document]
akaladarshi: Done.

chain/index/api.go
akaladarshi: Done.

chain/index/api.go
akaladarshi: Done.

chain/index/api.go
akaladarshi: The test is failing right now because `loadExecutedMessage` is not loading any messages, and we are using it to verify the indexed data.

chain/index/indexer.go
```diff
@@ -227,7 +225,7 @@ func (si *SqliteIndexer) indexSignedMessage(ctx context.Context, tx *sql.Tx, msg
 
 func (si *SqliteIndexer) Apply(ctx context.Context, from, to *types.TipSet) error {
 	si.closeLk.RLock()
```
rvagg, line 227: double lock here because you do it in `isClosed` too

chain/index/ddls.go
rvagg: would it be corruption if one were reverted and the other not? is that something we should also be checking for?

chain/index/api.go
```diff
@@ -126,17 +131,107 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
 	}, nil
 }
 
+type indexedTipSetData struct {
+	nonRevertedMessageCount int
+	nonRevertedEventCount   int
+}
+
+// getIndexedTipSetData fetches the indexed tipset data for a tipset
+func (si *SqliteIndexer) getIndexedTipSetData(ctx context.Context, ts *types.TipSet) (*indexedTipSetData, error) {
```
rvagg, line 166: how about you just return a mostly-filled `IndexValidation` from this that can be built on or merged from here? alternatively just 2 ints since that's all its doing

chain/index/api.go
```diff
@@ -126,17 +131,107 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
+		return nil
+	})
+
+	return &data, err
+}
+
+// verifyIndexedData verifies that the indexed data for a tipset is correct
+// by comparing the number of messages and events in the chainstore to the number of messages and events indexed.
+// NOTE: Events are loaded from the executed messages of the tipset at the next epoch (ts.Height() + 1).
+func (si *SqliteIndexer) verifyIndexedData(ctx context.Context, ts *types.TipSet, indexedData *indexedTipSetData) (err error) {
```
rvagg, line 191: ah I see why you want a struct for this, but it's still just 2 ints and it might even be clearer to provide them as named arguments here

chain/index/api.go
```diff
@@ -126,17 +131,107 @@ func (si *SqliteIndexer) ChainValidateIndex(ctx context.Context, epoch abi.Chain
...
+	tsKeyCid, err := ts.Key().Cid()
+	if err != nil {
+		return xerrors.Errorf("failed to get tipset key cid: %w", err)
+	}
+
+	// get the tipset where the messages of `ts` will be executed (deferred execution)
+	executionTs, err := si.cs.GetTipsetByHeight(ctx, ts.Height()+1, nil, false)
```
rvagg, line 198: see discussion in https://github.com/filecoin-project/lotus/pull/12447#issuecomment-2354470036, same problem here I think, particularly as you get close to the tip of the chain

aarshkshah1992:
Merging this and continuing all work as part of the parent PR. Will address all review comments there.

chain/index/api.go
aarshkshah1992: @rvagg Which is why we error out here if the parent of this tipset is not the original tipset. We can then ask the user to retry.

aarshkshah1992: I don't think we want to re-execute the tipset/regenerate events here.

Issue #12453, opened 2024-09-12
A new ChainIndexer  that subsumes that existing MsgIndex, EventIndex and TransactionIndex


aarshkshah1992:
## Summary
This issue is for the implementation of a new `ChainIndexer` in Lotus that will replace and subsume the existing `MsgIndex`, `EventsIndex`, and `EthTxHashIndex`, which are currently fragmented across multiple databases and have several known issues documented in [filecoin-project/lotus#12293](https://github.com/filecoin-project/lotus/issues/12293).

## Key Features
The `ChainIndexer` offers the following key features:

- Indexes all necessary state in a single database. 
- We now index a "tipset" and the relevant state changes caused by a tipset which makes the Indexed state easy to reason about and more aligned with how we persist state in the Chainstore. 
- Implements snapshot hydration.
- Implements  automated config driven Index garbage collection (GC).
- Provides automated backfilling.
- Offers simplified configuration.
- Wraps the asynchronous indexing in a synchronous READ API for RPC endpoints to consume to avoid off by N errors caused by async Indexing.

Note: while the ChainIndexer is primarily focused on events and ETH RPC usecases, it also benefits pre-FEVM as well.  For example, `StateSearchMsg` and its various dependents will now have a shortcut to find the message.

## Implementation Items

```[tasklist]
### Tasks
- [x] Land `ChainIndexer` implementation at https://github.com/filecoin-project/lotus/pull/12421
- [x] Land https://github.com/filecoin-project/lotus/pull/12463
- [x] Land Migration/Backfilling/Diagnostics tooling for the `ChainIndexer` at https://github.com/filecoin-project/lotus/pull/12450
- [x] Land https://github.com/filecoin-project/lotus/pull/12485
- [x] https://github.com/filecoin-project/lotus/issues/12489 (Land https://github.com/filecoin-project/lotus/pull/12504)
- [ ] Work with Glif to deploy the `ChainIndexer` on their FEVM archival node and re-index the entire archived chain history in the `ChainIndexer`. OKU should then be able to use that node for their app. A green light from Glif & OKU is what we need here to then ship this out to all users.
- [x] Write a document explaining the mechanics of re-indexing/backfilling/diagnostics/inspection of the `ChainIndexer` and explain how RPC providers can use the RPCs/CLI commands being built out in https://github.com/filecoin-project/lotus/pull/12450 for all of those use cases. (See https://github.com/filecoin-project/lotus/pull/12600).
[...]

PR #12463, opened 2024-09-14
feat: recompute tipset to generate missing events if event indexing is enabled

aarshkshah1992:
For https://github.com/filecoin-project/lotus/issues/12453.

If we're unable to find events for a given receipt, we recompute the tipset to generate events and re-read the events true to enable auto-repair for events. This will work during reconciliation as well.

### DIFF
--- a/chain/index/indexer.go
+++ b/chain/index/indexer.go
@@ -21,6 +21,7 @@ var _ Indexer = (*SqliteIndexer)(nil)
 
 // IdToRobustAddrFunc is a function type that resolves an actor ID to a robust address
 type IdToRobustAddrFunc func(ctx context.Context, emitter abi.ActorID, ts *types.TipSet) (address.Address, bool)
+type tipsetExecutorFnc func(ctx context.Context, ts *types.TipSet) (cid.Cid, cid.Cid, error)
 
 type preparedStatements struct {
 	insertEthTxHashStmt                   *sql.Stmt
@@ -53,6 +54,7 @@ type SqliteIndexer struct {
 	cs ChainStore
 
 	idToRobustAddrFunc IdToRobustAddrFunc
+	tipsetExecutorFnc  tipsetExecutorFnc
 
 	stmts *preparedStatements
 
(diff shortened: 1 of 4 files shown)


chain/index/indexer.go
```diff
@@ -21,6 +21,7 @@ var _ Indexer = (*SqliteIndexer)(nil)
 
 // IdToRobustAddrFunc is a function type that resolves an actor ID to a robust address
 type IdToRobustAddrFunc func(ctx context.Context, emitter abi.ActorID, ts *types.TipSet) (address.Address, bool)
+type tipsetExecutorFnc func(ctx context.Context, ts *types.TipSet) (cid.Cid, cid.Cid, error)
```
rvagg, line 24: Leaky abstraction here, since the two CIDs are always discarded. A couple of options to make this a bit nicer:

1. Just use an `error` return and make the wrapper function 1 line longer to discard the others (even though currently it doesn't even need to be a wrapper, just `indexer.SetTipsetExecutorFnc(sm.RecomputeTipSetState)` would do).
2. Just call it `recomputeTipSetState` and admit that this is all we're expecting here.
3. Make `SqliteIndexer` depend on `StateManager` and load it in via fx to avoid jumping through hoops and to be completely honest.


(also, `Fnc` -> `Func`?)

rvagg:
I'm OK with this in principle since the only path to calling this would be from `backfillIndex` -> `indexTipsetWithParentEvents`. Getting there via `Apply` should not encounter it because the amt should be there.

Aside from the review above, the only other comment I have is that it was kind of difficult tracing the calls with a `git grep` down into this because it jumps around through 3 files to get to the top calls, that makes me a little sus about how clean the abstractions are, but I don't have any good suggestions, just an observation.

chain/index/indexer.go
aarshkshah1992: @rvagg Fixed this by using a  `recomputeTipSetState` function here that simply returns an error and passing in a wrapper to hide the AMT cids during construction.

aarshkshah1992:
@rvagg I have addressed your review. I agree that there is some indirection involved in tracing the injection -> use flow for the repair function but can't think of a quick cleaner way here too.

rvagg (approved):

aarshkshah1992:
@rvagg Please let me know if I have your ✔️ on this PR and I can merge it to the parent branch.

PR #12474, opened 2024-09-17
feat: lotus-shed tooling for chain indexer

akaladarshi:
This PR is a part of: https://github.com/filecoin-project/lotus/pull/12450

PR changes:
- Adds `lotus-shed indexes validate-chainindex` to inspect and backfill the chain indexer through `ChainValidateIndex` API.

### DIFF
--- a/cmd/lotus-shed/indexes.go
+++ b/cmd/lotus-shed/indexes.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ipfs/go-cid"
 	"github.com/mitchellh/go-homedir"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/xerrors"
@@ -21,6 +22,7 @@ import (
 	"github.com/filecoin-project/go-state-types/exitcode"
 
 	lapi "github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/index"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
 	lcli "github.com/filecoin-project/lotus/cli"
@@ -37,6 +39,13 @@ const (
 	tipsetSeen       = `SELECT height,reverted FROM events_seen WHERE tipset_key_cid=?`
 
 	getEthTxHashCountForTipset = `SELECT COUNT(*) FROM eth_tx_hash WHERE message_cid IN (SELECT message_cid FROM tipset_message WHERE tipset_key_cid = ? AND reverted = 0)`
+	getTotalEventEntries       = `
+		SELECT COUNT(*)
+		FROM event_entry ee
+		JOIN event e ON ee.event_id = e.event_id
+		JOIN tipset_message tm ON e.message_id = tm.message_id
+		WHERE tm.tipset_key_cid = ? AND tm.message_cid IS NOT NULL
+	`
 )
 
 func withCategory(cat string, cmd *cli.Command) *cli.Command {
@@ -55,6 +64,7 @@ var indexesCmd = &cli.Command{
 		withCategory("events", backfillEventsCmd),
 		withCategory("events", inspectEventsCmd),
 		withCategory("chainindex_backfill", backfillChainIndexCmd),
+		withCategory("chainindex_inspect", inspectChainIndexCmd),
 	},
 }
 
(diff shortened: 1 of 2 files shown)


akaladarshi:
@aarshkshah1992 

`inspect-chainindex` cmd uses `chainValidateIndex` API and also fetches manually from the chain indexer to inspect the data by comparing with chain store data.

Should I keep it that way or remove it because `ChainIndexValidate` API is already verifying the indexed data.

aarshkshah1992:
@akaladarshi Let's remove it because `ChainIndexValidate ` is already validating the indexed data.

akaladarshi:
> @akaladarshi Let's remove it because `ChainIndexValidate ` is already validating the indexed data.

Ok sure.

Also right now, in inspect we are just doing counts mostly, should we check the data as well?

aarshkshah1992:
@akaladarshi But let's do that once we have this `lotus-shed` command implementation in place.

cmd/lotus-shed/indexes.go
```diff
@@ -34,6 +35,15 @@ const (
 	insertEntry      = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
 	upsertEventsSeen = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, false) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
 	tipsetSeen       = `SELECT height,reverted FROM events_seen WHERE tipset_key_cid=?`
+
+	getEthTxHashCountForTipset = `SELECT COUNT(*) FROM eth_tx_hash WHERE message_cid IN (SELECT message_cid FROM tipset_message WHERE tipset_key_cid = ? AND reverted = 0)`
```
akaladarshi: @aarshkshah1992 

should we return data related to `eth_tx_hash` from `ChainValidateIndex` API in `IndexValidation` as well ?.

aarshkshah1992: @akaladarshi After thinking this through, I think we should return data for ETH TX hashes, **however** for now, let's just continue validating counts like we are currently doing. We can validate the actual contents as well as part of a future work stream but I think what we have is good for now..

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +881,178 @@ var backfillTxHashCmd = &cli.Command{
...
+		}()
+
+		api := srv.FullNodeAPI()
+		ctx := lcli.ReqContext(cctx)
+
+		fromEpoch := cctx.Int("from")
+		if fromEpoch == 0 {
+			curTs, err := api.ChainHead(ctx)
+			if err != nil {
+				return err
```
aarshkshah1992, line 939: need an informative error message

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +881,178 @@ var backfillTxHashCmd = &cli.Command{
...
+			}
+			fromEpoch = int(curTs.Height()) - 1
+		} else {
+			fromEpoch = fromEpoch - 1
```
aarshkshah1992, line 943: why do we need this ?

aarshkshah1992: Only do this is `fromEpoch` >= `head.Height()`

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +881,178 @@ var backfillTxHashCmd = &cli.Command{
...
+		}
+
+		if fromEpoch <= 0 {
+			return fmt.Errorf("invalid from epoch: %d", fromEpoch)
+		}
+
+		toEpoch := cctx.Int("to")
+		if toEpoch > fromEpoch {
+			return fmt.Errorf("to epoch must be less than from epoch")
```
aarshkshah1992, line 952: errors.New()

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +881,178 @@ var backfillTxHashCmd = &cli.Command{
...
+			func() string {
+				if backfill {
+					return "backfill and inspect chainindex"
+				}
+				return "inspect chainindex"
+			}(),
+			fromEpoch, toEpoch)
+
+		// starting from the FromEpoch-1 and going down to the ToEpoch
+		// this is because `ChainValidateIndex` fetches the tipset.height()+1, which might not be available in case of chain head
```
aarshkshah1992, line 973: this comment is not clear. Wdym by "`ChainValidateIndex` fetches the tipset.height()+1" ?

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +881,178 @@ var backfillTxHashCmd = &cli.Command{
...
+			if ctx.Err() != nil {
+				return ctx.Err()
+			}
+
+			indexValidateResp, err := api.ChainValidateIndex(ctx, abi.ChainEpoch(epoch), backfill)
+			if err != nil {
+				if failfast {
+					return fmt.Errorf("failed to validate index for epoch %d: %w", epoch, err)
+				}
+				log.Warnf("Error validating index for epoch %d: %v", epoch, err)
```
aarshkshah1992, line 984: use %w for errors.

cmd/lotus-shed/indexes.go
```diff
@@ -879,3 +881,178 @@ var backfillTxHashCmd = &cli.Command{
...
+				continue
+			}
+
+			if !indexValidateResp.TipSetKey.IsEmpty() || indexValidateResp.Height != uint64(epoch) && indexValidateResp.Backfilled == backfill {
```
aarshkshah1992, line 988: I dont think we need this validation block here at all. The `ChainIndexValidationAPI` guruantees that it will return a valid `indexValidateResp` if error is nil.

cmd/lotus-shed/indexes.go
akaladarshi: This was done because `ChainnValidateIndex` fetches event of `ts` from `ts+1`, So in case of `fromEpoch == head.height()`, we can't get `ts.Height() + 1`.

So due to nature of `ChainnValidateIndex` by default `fromEpoch` will not be included in backfilling, and to maintain consistency I did `fromEpoch -1`.

PR #12478, opened 2024-09-17
feat: add support for eth_getBlockReceipts

### LINKED ISSUE #12429: Support eth_getBlockReceipts

virajbhartiya:
## Related Issues
Fixes #12429  

## Proposed Changes
Implimented `eth_getBlockReceipts` method which accepts a block number and returns all the transaction receipts for it

## Additional Info
[QuickNode](https://www.quicknode.com/docs/ethereum/eth_getBlockReceipts)
[Infura](https://docs.infura.io/api/networks/blast/json-rpc-methods/eth_getblockreceipts)

## Checklist

- [ ] CI is green

### DIFF
--- a/api/api_full.go
+++ b/api/api_full.go
@@ -768,11 +768,12 @@ type FullNode interface {
 	EthGetTransactionHashByCid(ctx context.Context, cid cid.Cid) (*ethtypes.EthHash, error)                                                     //perm:read
 	EthGetMessageCidByTransactionHash(ctx context.Context, txHash *ethtypes.EthHash) (*cid.Cid, error)                                          //perm:read
 	EthGetTransactionCount(ctx context.Context, sender ethtypes.EthAddress, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthUint64, error) //perm:read
-	EthGetTransactionReceipt(ctx context.Context, txHash ethtypes.EthHash) (*EthTxReceipt, error)
-	EthGetBlockReceipts(ctx context.Context, blkParam ethtypes.EthBlockNumberOrHash) ([]*EthTxReceipt, error)                                  //perm:read
-	EthGetTransactionReceiptLimited(ctx context.Context, txHash ethtypes.EthHash, limit abi.ChainEpoch) (*EthTxReceipt, error)                 //perm:read
-	EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error)    //perm:read
-	EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNum ethtypes.EthUint64, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error) //perm:read
+	EthGetTransactionReceipt(ctx context.Context, txHash ethtypes.EthHash) (*EthTxReceipt, error)                                               //perm:read
+	EthGetBlockReceipts(ctx context.Context, blkParam ethtypes.EthBlockNumberOrHash) ([]*EthTxReceipt, error)                                   //perm:read
+	EthGetBlockReceiptsLimited(ctx context.Context, blkParam ethtypes.EthBlockNumberOrHash, limit abi.ChainEpoch) ([]*EthTxReceipt, error)      //perm:read
+	EthGetTransactionReceiptLimited(ctx context.Context, txHash ethtypes.EthHash, limit abi.ChainEpoch) (*EthTxReceipt, error)                  //perm:read
+	EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error)     //perm:read
+	EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNum ethtypes.EthUint64, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error)  //perm:read
 
 	EthGetCode(ctx context.Context, address ethtypes.EthAddress, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthBytes, error)                                  //perm:read
 	EthGetStorageAt(ctx context.Context, address ethtypes.EthAddress, position ethtypes.EthBytes, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthBytes, error) //perm:read
--- a/api/api_full.go
+++ b/api/api_full.go
@@ -768,10 +768,11 @@ type FullNode interface {
 	EthGetTransactionHashByCid(ctx context.Context, cid cid.Cid) (*ethtypes.EthHash, error)                                                     //perm:read
 	EthGetMessageCidByTransactionHash(ctx context.Context, txHash *ethtypes.EthHash) (*cid.Cid, error)                                          //perm:read
 	EthGetTransactionCount(ctx context.Context, sender ethtypes.EthAddress, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthUint64, error) //perm:read
-	EthGetTransactionReceipt(ctx context.Context, txHash ethtypes.EthHash) (*EthTxReceipt, error)                                               //perm:read
-	EthGetTransactionReceiptLimited(ctx context.Context, txHash ethtypes.EthHash, limit abi.ChainEpoch) (*EthTxReceipt, error)                  //perm:read
-	EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error)     //perm:read
-	EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNum ethtypes.EthUint64, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error)  //perm:read
+	EthGetTransactionReceipt(ctx context.Context, txHash ethtypes.EthHash) (*EthTxReceipt, error)
+	EthGetBlockReceipts(ctx context.Context, blkParam ethtypes.EthBlockNumberOrHash) ([]*EthTxReceipt, error)                                  //perm:read
+	EthGetTransactionReceiptLimited(ctx context.Context, txHash ethtypes.EthHash, limit abi.ChainEpoch) (*EthTxReceipt, error)                 //perm:read
+	EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error)    //perm:read
+	EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNum ethtypes.EthUint64, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error) //perm:read
 
 	EthGetCode(ctx context.Context, address ethtypes.EthAddress, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthBytes, error)                                  //perm:read
 	EthGetStorageAt(ctx context.Context, address ethtypes.EthAddress, position ethtypes.EthBytes, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthBytes, error) //perm:read
--- a/chain/events/filter/event.go
+++ b/chain/events/filter/event.go
@@ -16,6 +16,7 @@ import (
 	"github.com/filecoin-project/go-state-types/abi"
 	blockadt "github.com/filecoin-project/specs-actors/actors/util/adt"
 
+	"github.com/filecoin-project/lotus/chain/actors/adt"
 	cstore "github.com/filecoin-project/lotus/chain/store"
 	"github.com/filecoin-project/lotus/chain/types"
 )
@@ -437,37 +438,43 @@ func (m *EventFilterManager) LoadExecutedMessages(ctx context.Context, msgTs, rc
 		return nil, xerrors.Errorf("read messages: %w", err)
 	}
 
-	st := m.ChainStore.ActorStore(ctx)
+	store := m.ChainStore.ActorStore(ctx)
 
-	arr, err := blockadt.AsArray(st, rctTs.Blocks()[0].ParentMessageReceipts)
+	arr, err := blockadt.AsArray(store, rctTs.Blocks()[0].ParentMessageReceipts)
 	if err != nil {
 		return nil, xerrors.Errorf("load receipts amt: %w", err)
 	}
 
-	if uint64(len(msgs)) != arr.Length() {
-		return nil, xerrors.Errorf("mismatching message and receipt counts (%d msgs, %d rcts)", len(msgs), arr.Length())
-	}
-
-	ems := make([]executedMessage, len(msgs))
-
-	for i := 0; i < len(msgs); i++ {
-		ems[i].msg = msgs[i]
-
-		var rct types.MessageReceipt
-		found, err := arr.Get(uint64(i), &rct)
+	receipts := make([]types.MessageReceipt, arr.Length())
+	for i := uint64(0); i < arr.Length(); i++ {
+		found, err := arr.Get(i, &receipts[i])
 		if err != nil {
 			return nil, xerrors.Errorf("load receipt: %w", err)
 		}
 		if !found {
 			return nil, xerrors.Errorf("receipt %d not found", i)
 		}
-		ems[i].rct = &rct
+	}
 
-		if rct.EventsRoot == nil {
+	return LoadExecutedMessages(ctx, store, msgs, receipts)
+}
+
+func LoadExecutedMessages(ctx context.Context, store adt.Store, msgs []types.ChainMsg, receipts []types.MessageReceipt) ([]executedMessage, error) {
+	if len(msgs) != len(receipts) {
+		return nil, xerrors.Errorf("mismatching message and receipt counts (%d msgs, %d rcts)", len(msgs), len(receipts))
+	}
+
+	ems := make([]executedMessage, len(msgs))
+
+	for i := 0; i < len(msgs); i++ {
+		ems[i].msg = msgs[i]
+		ems[i].rct = &receipts[i]
+
+		if receipts[i].EventsRoot == nil {
 			continue
 		}
 
-		evtArr, err := amt4.LoadAMT(ctx, st, *rct.EventsRoot, amt4.UseTreeBitWidth(types.EventAMTBitwidth))
+		evtArr, err := amt4.LoadAMT(ctx, store, *receipts[i].EventsRoot, amt4.UseTreeBitWidth(types.EventAMTBitwidth))
 		if err != nil {
 			return nil, xerrors.Errorf("load events amt: %w", err)
 		}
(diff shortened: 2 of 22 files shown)


virajbhartiya:
As far as my understanding of deferred execution, it refers to the scheduling or delaying of actions within the Filecoin protocol in smart contracts or while processing messages. Also, Filecoin nodes maintain a replicated state machine, and state transitions may be deferred while validation of proofs or consensus confirmations haven't been met.

For chain forks, it occur when two or more valid blocks are mined at the same height but from different miners, creating divergent chains. I am still reading about the Expected Consensus mechanism but it chooses which fork to go forward with based on which chain has the highest aggregate power of the miners that have contributed to that chain.

api/api_full.go
```diff
@@ -768,10 +768,11 @@ type FullNode interface {
 	EthGetTransactionHashByCid(ctx context.Context, cid cid.Cid) (*ethtypes.EthHash, error)                                                     //perm:read
 	EthGetMessageCidByTransactionHash(ctx context.Context, txHash *ethtypes.EthHash) (*cid.Cid, error)                                          //perm:read
 	EthGetTransactionCount(ctx context.Context, sender ethtypes.EthAddress, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthUint64, error) //perm:read
-	EthGetTransactionReceipt(ctx context.Context, txHash ethtypes.EthHash) (*EthTxReceipt, error)                                               //perm:read
-	EthGetTransactionReceiptLimited(ctx context.Context, txHash ethtypes.EthHash, limit abi.ChainEpoch) (*EthTxReceipt, error)                  //perm:read
-	EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error)     //perm:read
-	EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNum ethtypes.EthUint64, txIndex ethtypes.EthUint64) (ethtypes.EthTx, error)  //perm:read
+	EthGetTransactionReceipt(ctx context.Context, txHash ethtypes.EthHash) (*EthTxReceipt, error)
```
rvagg, line 771:
```suggestion
	EthGetTransactionReceipt(ctx context.Context, txHash ethtypes.EthHash) (*EthTxReceipt, error) //perm:read

```

rvagg: these `perm` comments influence the authenticated rpc layer, saying what permisisons are needed to access the API

chain/events/filter/event.go
```diff
@@ -308,10 +308,9 @@ type EventFilterManager struct {
 	AddressResolver  func(ctx context.Context, emitter abi.ActorID, ts *types.TipSet) (address.Address, bool)
 	MaxFilterResults int
 	EventIndex       *EventIndex
-
-	mu            sync.Mutex // guards mutations to filters
-	filters       map[types.FilterID]EventFilter
-	currentHeight abi.ChainEpoch
+	mu               sync.Mutex // guards mutations to filters
```
rvagg, line 311:
```suggestion

	mu               sync.Mutex // guards mutations to filters
```

chain/events/filter/event.go
```diff
@@ -432,7 +431,7 @@ func (m *EventFilterManager) Remove(ctx context.Context, id types.FilterID) erro
 	return nil
 }
 
-func (m *EventFilterManager) loadExecutedMessages(ctx context.Context, msgTs, rctTs *types.TipSet) ([]executedMessage, error) {
+func (m *EventFilterManager) LoadExecutedMessages(ctx context.Context, msgTs, rctTs *types.TipSet) ([]executedMessage, error) {
```
rvagg, line 434: the lint on this line is reasonable, I think you can export `ExecutedMessage` as it is, it has accessors on the private properties anyway

gateway/node.go
```diff
@@ -328,3 +328,7 @@ func (gw *Node) limit(ctx context.Context, tokens int) error {
 	}
 	return nil
 }
+
+func (gw *Node) EthGetBlockReceipts(ctx context.Context, blkParam ethtypes.EthBlockNumberOrHash) ([]*api.EthTxReceipt, error) {
+	return gw.target.EthGetBlockReceipts(ctx, blkParam)
+}
```
rvagg, line 334: put this in proxy_eth.go

also, see in there we have `EthGetTransactionReceipt`, we need to add a rate limit at the top checking `gw.limit(ctx, stateRateLimitTokens)`

and also, you can see in there why `EthGetTransactionReceiptLimited` exists - to pass the `maxMessageLookbackEpochs` through to the function. You should be able to follow that same pattern and add a `EthGetBlockReceiptsLimited`:

* add in api/api_full.go
* add in api/api_gateway.go
* add in the top of gateway/node.go
* add in gateway/proxy_eth.go along with `EthGetBlockReceipts`
* in node/impl/full/eth.go rename your current `EthGetBlockReceipts` to `EthGetBlockReceiptsLimited` with a `api.LookbackNoLimit` limit like `EthGetTransactionByHash` does

rvagg: The gateway/proxy* methods exist to impose limits for anyone exposing these APIs to the public, like glif etc. so you can impose rate limiting and also limitations on how far back in state you can look. I think glif does 1 day by default.

node/impl/full/eth.go
```diff
@@ -2135,3 +2141,93 @@ func (g gasRewardSorter) Swap(i, j int) {
...
+	}
+
+	// Get the parent tipset CID
+	parentTsCid, err := ts.Parents().Cid()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get parent tipset cid: %w", err)
+	}
+
+	// Load executed messages and extract events
+	executedMsgs, err := a.EthEventHandler.EventFilterManager.LoadExecutedMessages(ctx, ts, ts)
```
rvagg, line 2170: ah, we have a problem here - you can't pass in `ts` to this, you need `ts+1` which you don't have because you synthetically generated the next state, but if you look inside `LoadExecutedMessages` you'll see that these two parameters are only needed to gather `msgs` and reference `ParentMessageReceipts`, which you have as `receipts`. So what you probably want to do is make a new function internal to `EventFilterManager` that wraps `LoadExecutedMessages` and make it take msgs and receipts as arguments. You could even detach it from `EventFilterManager` entirely so you don't need to reach into that dependency (it also makes this independent of the events db, which is an interesting property).

Here's a quick go at what that might look like:

```diff
diff --git a/chain/events/filter/event.go b/chain/events/filter/event.go
index fa17d235e..2ce31e127 100644
--- a/chain/events/filter/event.go
+++ b/chain/events/filter/event.go
@@ -16,6 +16,7 @@ import (
        "github.com/filecoin-project/go-state-types/abi"
        blockadt "github.com/filecoin-project/specs-actors/actors/util/adt"
 
+       "github.com/filecoin-project/lotus/chain/actors/adt"
        cstore "github.com/filecoin-project/lotus/chain/store"
        "github.com/filecoin-project/lotus/chain/types"
 )
@@ -438,37 +439,43 @@ func (m *EventFilterManager) loadExecutedMessages(ctx context.Context, msgTs, rc
                return nil, xerrors.Errorf("read messages: %w", err)
        }
 
-       st := m.ChainStore.ActorStore(ctx)
+       store := m.ChainStore.ActorStore(ctx)
 
-       arr, err := blockadt.AsArray(st, rctTs.Blocks()[0].ParentMessageReceipts)
+       arr, err := blockadt.AsArray(store, rctTs.Blocks()[0].ParentMessageReceipts)
        if err != nil {
                return nil, xerrors.Errorf("load receipts amt: %w", err)
        }
 
-       if uint64(len(msgs)) != arr.Length() {
-               return nil, xerrors.Errorf("mismatching message and receipt counts (%d msgs, %d rcts)", len(msgs), arr.Length())
-       }
-
-       ems := make([]executedMessage, len(msgs))
-
-       for i := 0; i < len(msgs); i++ {
-               ems[i].msg = msgs[i]
-
-               var rct types.MessageReceipt
-               found, err := arr.Get(uint64(i), &rct)
+       receipts := make([]types.MessageReceipt, arr.Length())
+       for i := 0; i < len(receipts); i++ {
+               found, err := arr.Get(uint64(i), &receipts[i])
                if err != nil {
                        return nil, xerrors.Errorf("load receipt: %w", err)
                }
                if !found {
                        return nil, xerrors.Errorf("receipt %d not found", i)
                }
-               ems[i].rct = &rct
+       }
+
+       return LoadExecutedMessages(ctx, store, msgs, receipts)
+}
+
+func LoadExecutedMessages(ctx context.Context, store adt.Store, msgs []types.ChainMsg, receipts []types.MessageReceipt) ([]executedMessage, error) {
+       if len(msgs) != len(receipts) {
+               return nil, xerrors.Errorf("mismatching message and receipt counts (%d msgs, %d rcts)", len(msgs), len(receipts))
+       }
+
+       ems := make([]executedMessage, len(msgs))
 
-               if rct.EventsRoot == nil {
+       for i := 0; i < len(msgs); i++ {
+               ems[i].msg = msgs[i]
+               ems[i].rct = &receipts[i]
+
+               if receipts[i].EventsRoot == nil {
                        continue
                }
 
-               evtArr, err := amt4.LoadAMT(ctx, st, *rct.EventsRoot, amt4.UseTreeBitWidth(types.EventAMTBitwidth))
+               evtArr, err := amt4.LoadAMT(ctx, store, *receipts[i].EventsRoot, amt4.UseTreeBitWidth(types.EventAMTBitwidth))
                if err != nil {
                        return nil, xerrors.Errorf("load events amt: %w", err)
                }
@@ -491,7 +498,6 @@ func (m *EventFilterManager) loadExecutedMessages(ctx context.Context, msgTs, rc
                if err != nil {
                        return nil, xerrors.Errorf("read events: %w", err)
                }
-
        }
 
        return ems, nil
```

It also gets us close, but not quite, to being able to remove some duplicate code for loading events which you'll also see in node/impl/full/chain.go `ChainGetEvents`. But that can be a job for some other time.

node/impl/full/eth.go
```diff
@@ -2135,3 +2141,93 @@ func (g gasRewardSorter) Swap(i, j int) {
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to load executed messages: %w", err)
+	}
+
+	// Convert executed messages to events
+	events := make([]*filter.CollectedEvent, 0, len(executedMsgs))
+	for _, em := range executedMsgs {
+		for _, ev := range em.Events() {
+			addr, err := address.NewIDAddress(uint64(ev.Emitter))
```
rvagg, line 2179: We're going to need to check this, does it produce what we want? I suspect not, you'll find out if you compare the output in the itests with `EthGetTransactionReceipt` I think.

there's a translation layer inside the event subsystem to go from id addresses to robust addresses, you can see it in action here: https://github.com/filecoin-project/lotus/blob/9b86cc5d60183251a8d2e8d8999d5d2661e00063/chain/events/filter/event.go#L104-L114

The resolver itself is https://github.com/filecoin-project/lotus/blob/9b86cc5d60183251a8d2e8d8999d5d2661e00063/node/modules/actorevent.go#L134-L147

You could just potentially use the resolver in `EventFilterManager` to do this, it's exposed as `AddressResolver`, just be sure to cache the look-ups like `CollectEvents` does because you're likely to have a lot of duplicates in a single set.

node/impl/full/eth.go
```diff
@@ -2135,3 +2141,93 @@ func (g gasRewardSorter) Swap(i, j int) {
...
+		if err != nil {
+			return nil, xerrors.Errorf("failed to create Eth receipt: %w", err)
+		}
+
+		// Add logs to the receipt
+		receipt.Logs = filterLogsByTxHash(logs, receipt.TransactionHash)
+
+		ethReceipts = append(ethReceipts, &receipt)
+	}
+
```
rvagg, line 2220: As a sanity check, could you keep track of how many `receipt.Logs` end up being filtered and then comparing it to `len(logs)` just to make sure we got the same number—if we didn't then there's a problem and we should treat that as an error (which would need further investigation if someone ever encountered it).

itests/fevm_test.go
```diff
@@ -1053,3 +1053,29 @@ func TestFEVMErrorParsing(t *testing.T) {
...
+func TestEthGetBlockReceipts(t *testing.T) {
+	ctx, cancel, client := kit.SetupFEVMTest(t)
+	defer cancel()
+
+	// Deploy a contract to generate a transaction
+	_, contractAddr := client.EVM().DeployContractFromFilename(ctx, "contracts/SimpleCoin.hex")
+
+	// Invoke a function to ensure there is a transaction
+	_, _, err := client.EVM().InvokeContractByFuncName(ctx, client.DefaultKey.Address, contractAddr, "someFunction()", []byte{})
+	require.NoError(t, err)
```
rvagg, line 1066: you might need to check out the other uses of `SimpleCoin` to make sure you interact with it appropriately, I think we're failing here

itests/eth_filter_test.go
```diff
@@ -748,6 +754,24 @@ func TestMultipleEvents(t *testing.T) {
 	expectedLogIndex := 0
 	var receiptLogs []ethtypes.EthLog
 	for _, receipt := range receipts {
+		require.Equal(t, ethtypes.EthUint64(0x1), receipt.Status)
+
+		// Test GetBlockReceipts
+		blockReceipts, err := client.EthGetBlockReceipts(ctx, ethtypes.EthBlockNumberOrHash{BlockHash: &receipt.BlockHash})
+		require.NoError(t, err)
```
rvagg, line 761: the error in CI here is due to the misuse of `LoadExecutedMessages` - you're passing `ts, ts` so it looks messages up in the first `ts` and receipts in the second `ts` and the number of receipts doesn't match the number of messages, so:

```
        	            	failed to load executed messages: mismatching message and receipt counts (3 msgs, 0 rcts)
```

Hence the need to refactor because you have the messages `ts` but not the receipts one because you're only executing messages, not assembling a full new tipset with blocks etc.

itests/eth_deploy_test.go
```diff
@@ -140,6 +140,12 @@ func TestDeployment(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, receipt)
 
+	// Verify that the receipt is correct.
+	receipts, err := client.EthGetBlockReceipts(ctx, ethtypes.EthBlockNumberOrHash{BlockHash: &receipt.BlockHash})
+	require.NoError(t, err)
```
rvagg, line 145: same proble, here

```
        	            	failed to load executed messages: mismatching message and receipt counts (1 msgs, 0 rcts)
```

rvagg:
Getting very close I think!

fwiw you can run the itests locally but they're best done specifically targeted, e.g. `go test ./itests/fevm_test.go -run TestEthGetBlockReceipts -v`. It should be quick enough to run on most machines. You should try it out on the ones you're modifying to see how close you are. Currently I need to Approve and Run CI for you here as a first-time contributor, unfortunately.

gateway/node.go
virajbhartiya: Could you please explain a bit more about the limit? Currently, I am passing `abi.ChainEpoch` as the limit to the function. Is that correct or should it be something else, for tests, should I pass the limit as 1 or 0 to retrieve all the receipts or should it be something else?

itests/eth_filter_test.go
virajbhartiya:
```go
fromAddr := client.DefaultKey.Address // Get the address to pass
inputData := inputDataFromFrom(ctx, t, client, fromAddr)

// Invoke the sendCoin function to ensure there is a transaction
_, _, err := client.EVM().InvokeContractByFuncName(ctx, client.DefaultKey.Address, contractAddr, "sendCoin(address,uint256)", inputData)
require.NoError(t, err) // Ensure the function call is successful
```

itests/fevm_test.go
virajbhartiya:
```go
inputData, err := hex.DecodeString("0000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000066162636465660000000000000000000000000000000000000000000000000000")
require.NoError(t, err)

// Invoke the sendCoin function to ensure there is a transaction
_, _, err = client.EVM().InvokeContractByFuncName(ctx, client.DefaultKey.Address, contractAddr, "sendCoin(address,uint256)", inputData)
require.NoError(t, err) // Ensure the function call is successful
```
Is this the correct implementation for it?

itests/eth_filter_test.go
virajbhartiya: I tried to refactor it to
```go	
st, _, err = a.StateManager.TipSetState(ctx, ts)
if err != nil {
	return nil, xerrors.Errorf("failed to load executed messages: %w", err)
}

executedMsgs, err := a.EthEventHandler.EventFilterManager.LoadExecutedMessages(ctx, ts, st)
```

But I am unable to get the right type for st
```
cannot use st (variable of type cid.Cid) as *types.TipSet value in argument to 
```

Could you please help me out with this?

gateway/node.go
rvagg: `gw.limit()` is a rate limiter based on the complexity of the query:

```
	basicRateLimitTokens  = 1
	walletRateLimitTokens = 1
	chainRateLimitTokens  = 2
	stateRateLimitTokens  = 3
```

the APIs that need to go digging around in state, (more than just basic chain queries), consume the most tokens; this API would be one of those, so just use `gw.limit(ctx, stateRateLimitTokens)` to "burn up" 3 tokens for that query. The internals of the rate limiter have a mechanism to account for those and either slow a user down or block them if they consume too many of these tokens in a short period.

For the `limit` argument to `EthGetBlockReceiptsLimited` from `EthGetBlockReceipts`, just use `api.LookbackNoLimit` and then copy the logic you see in `EthGetTransactionReceiptLimited` from proxy_eth.go - where it gets capped at `gw.maxMessageLookbackEpochs`. What that does is lets the gateway user run it with `--api-max-lookback` and a number of epochs, such as `2880`. So when you end up resolving the epoch that the `EthGetBlockReceiptsLimited` call (in eth.go), you can compare that to the `limit` parameter that gets passed in and reject it if the block is too far in the past.

itests/eth_filter_test.go
rvagg: see the suggestion above: https://github.com/filecoin-project/lotus/pull/12478#discussion_r1764336196

you need to be passing in the messages and receipts, which you have, not letting `LoadExecutedMessages` figure it out because it's currently designed to take 2 tipsets, but you don't have the second one because you're not actually creating a tipset, you're just executing messages to get receipts; so bypass all of that by refactoring out the need to supply tipsets so you just supply messages and their receipts.

If you apply the diff in my suggestion above you should get closer to this.

itests/fevm_test.go
aarshkshah1992: Had a chat with @virajbhartiya and he will use the `MultipleEvents` contract to generate events here and test this out.

virajbhartiya: I was learning about `Cid` in IPFS, and from what I understand, it's a unique identifier created by hashing the contents of a file stored on IPFS. This makes the file immutable, as any change to the content generates a new Cid. Cid functions as a content-based address, allowing nodes to locate the file on the network. It’s similar to how domain names function in web2, but instead of pointing to a location, it points directly to the data itself. Am I going in the right direction?  @aarshkshah1992

virajbhartiya:
I have made the changes suggested above, and also implemented `MultipleEvents` in the test, can you please give it a look @aarshkshah1992 @rvagg

aarshkshah1992:
@virajbhartiya Merging this into a feature branch and taking it over to fill in the missing bits. Looks mostly good here.

Thanks for all your hard work here !

Issue #9324, opened 2022-09-16
`boostd dagstore initialize-all` crashes Boost - error coming from reader in lotus storage


rjan90:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [X] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [X] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
boostd version 1.4.0+git.810afec
Based on the boostd version, the SP is running either Lotus v1.17.0 or v1.17.1
```

### Describe the Bug

This is was an error reported initially by @stuberman in the Boost-repo, but the crash is coming from the reader in the lotus storage. [Issue-report here.](https://github.com/filecoin-project/boost/discussions/814)

When running `boostd dagstore initialize-all --concurrency=3` or even `boostd dagstore initialize-all --concurrency=1` then boostd crashes after a few minutes. See log below:

### Logging Information

```text
[...]

PR #12491, opened 2024-09-19
fix(miner): make UnpadReader be a correct Reader

magik6k:
## Related Issues
Fixes https://github.com/filecoin-project/lotus/issues/9324

## Proposed Changes
The fr32 unpad reader currently requires that the output is power-of-two-unpadded size, which is an extremely sketchy assumption and seems to cause lots of pain in retrievals.

There is already quite a lot of code in piece-provider which ensures that reads are initialized at correct offsets (multiple of 127-unpadded-bytes), but the read sizes are only ensured by the fact that the output reader is wrapped in a bufio.Reader, which won't always buffer writes with the required size, leading to hard to debug panics.

This PR aims to make the UnpadReader behave more like any normal reader, removing a giant footgun in the retrieval code.

## Additional Info
I realize that this code should probably eventually be external to lotus, but right now both boost and curio depend on this for decoding fr32 data. Should be extracted but let's fix the panics first.

## Checklist

- [ ] CI is green

### DIFF
--- a/storage/sealer/fr32/readers.go
+++ b/storage/sealer/fr32/readers.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"math/bits"
 
+	pool "github.com/libp2p/go-buffer-pool"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-state-types/abi"
@@ -14,6 +15,8 @@ type unpadReader struct {
 
 	left uint64
 	work []byte
+
+	stash []byte
 }
 
 func BufSize(sz abi.PaddedPieceSize) int {
@@ -31,6 +34,10 @@ func NewUnpadReaderBuf(src io.Reader, sz abi.PaddedPieceSize, buf []byte) (io.Re
 		return nil, xerrors.Errorf("bad piece size: %w", err)
 	}
 
+	if abi.PaddedPieceSize(len(buf)).Validate() != nil {
+		return nil, xerrors.Errorf("bad buffer size")
+	}
+
 	return &unpadReader{
 		src: src,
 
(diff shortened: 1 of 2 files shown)


LexLuthr (approved):

rjan90:
Looks like there are a couple of lint errors related to not checking the error return value of the `rand.Read()` that needs to be addressed.

rvagg:
Would be good to also address @ribasushi's findings in https://filecoinproject.slack.com/archives/C06GD1SS56Y/p1727046562422669 - set numcpus low enough and this reliably borks; although it's the same before and after this patch. Something odd going on when you get below 8 CPUs.

tediou5:
> Would be good to also address @ribasushi's findings in https://filecoinproject.slack.com/archives/C06GD1SS56Y/p1727046562422669 - set numcpus low enough and this reliably borks; although it's the same before and after this patch. Something odd going on when you get below 8 CPUs.

After #12884 was merged, the issue has been resolved, so I believe we can move forward with this PR.

rvagg:
@tediou5 would you mind taking this one over? When https://github.com/filecoin-project/github-mgmt/pull/141 is merged you should even be able to take over this branch as is if you like; @magik6k can speak up if he objects but I suspect not because they have alternatives in Curio now.

tediou5:
> @tediou5 would you mind taking this one over? When [filecoin-project/github-mgmt#141](https://github.com/filecoin-project/github-mgmt/pull/141) is merged you should even be able to take over this branch as is if you like; @magik6k can speak up if he objects but I suspect not because they have alternatives in Curio now.

Of course! :) But now that the previous issues are resolved, I hope someone can help with testing and reviewing.

BigLep:
@tediou5 : I think if it looks good to you and you were able to verify locally, you are good to merge.  Or is there something specific you want others to test?

tediou5:
> I think if it looks good to you and you were able to verify locally, you are good to merge.

@BigLep This fix works for me, but I don’t have merge permission.

rvagg:
@tediou5 you should do, you were added to `@filecoin-project/lotus-contributors` which has write permissions

PR #12505, opened 2024-09-23
feat: eth_getBlockReceipts API implementation

aarshkshah1992:
Closes https://github.com/filecoin-project/lotus/issues/12429.
This is a port over of the work done by @virajbhartiya at https://github.com/filecoin-project/lotus/pull/12478.
Just adding in the last bit and taking it to completion here as we need it for the upcoming Lotus release.

PR #12514, opened 2024-09-26
fix(eth): proper gas calculation for eth_getBlockReceipts

rvagg:
Ref: https://github.com/filecoin-project/lotus/pull/12505

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -588,10 +588,7 @@ func (a *EthModule) EthGetBlockReceiptsLimited(ctx context.Context, blockParam e
 		return nil, xerrors.Errorf("failed to load state tree: %w", err)
 	}
 
-	baseFee, err := a.StateManager.ChainStore().ComputeBaseFee(ctx, ts)
-	if err != nil {
-		return nil, xerrors.Errorf("failed to compute base fee: %w", err)
-	}
+	baseFee := ts.Blocks()[0].ParentBaseFee
 
 	ethReceipts := make([]*api.EthTxReceipt, 0, len(msgs))
 	for i, msg := range msgs {
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -527,7 +527,12 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash 
 		return nil, xerrors.Errorf("failed to convert %s into an Eth Txn: %w", txHash, err)
 	}
 
-	receipt, err := newEthTxReceipt(ctx, tx, msgLookup, a.ChainAPI, a.EthEventHandler)
+	parentTs, err := a.Chain.GetTipSetFromKey(ctx, msgLookup.TipSet)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to load parent tipset: %w", err)
+	}
+
+	receipt, err := newEthTxReceipt(ctx, tx, parentTs.Blocks()[0].ParentBaseFee, msgLookup.Receipt, a.EthEventHandler)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to create Eth receipt: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


rvagg:
Oh, the failure is interesting

```
    eth_legacy_transactions_test.go:362: 
        	Error Trace:	/home/runner/work/lotus/lotus/itests/eth_legacy_transactions_test.go:362
        	Error:      	Not equal: 
        	            	expected: 229215
        	            	actual  : 231755
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -4,3 +4,3 @@
        	            	   abs: (big.nat) (len=1) {
        	            	-   (big.Word) 229215
        	            	+   (big.Word) 231755
        	            	   }
        	Test:       	TestLegacyContractInvocation

```

why is that not happy? I shouldn't have changed EthGetTransactionReceipt ..

rvagg:
ok, 🤞 let's try that change, I didn't go back to the parents to get the parentbasefee

node/impl/full/eth.go
```diff
@@ -575,23 +588,21 @@ func (a *EthModule) EthGetBlockReceiptsLimited(ctx context.Context, blockParam e
 		return nil, xerrors.Errorf("failed to load state tree: %w", err)
 	}
 
+	baseFee, err := a.StateManager.ChainStore().ComputeBaseFee(ctx, ts)
```
aarshkshah1992, line 591: @rvagg I was a bit confused about this one. So the base fee a user will pay for their tx is determined at message inclusion and not execution. If yes, this makes sense. We don,t need the execution tipset at all for this calculation.

node/impl/full/eth.go
```diff
@@ -527,7 +527,20 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
 		return nil, xerrors.Errorf("failed to convert %s into an Eth Txn: %w", txHash, err)
 	}
 
-	receipt, err := newEthTxReceipt(ctx, tx, msgLookup, a.ChainAPI, a.EthEventHandler)
+	ts, err := a.Chain.GetTipSetFromKey(ctx, msgLookup.TipSet)
```
aarshkshah1992, line 530: Right so `msgLookup.Tipset` is the "execution tipset" of the message and NOT the "inclusion tipset".

node/impl/full/eth.go
rvagg: ya, I go tthis one wrong, it should be the ts ParentBaseFee, see latest

node/impl/full/eth.go
rvagg: yeah, and the base fee comes off the inclusion one

or something like that

node/impl/full/eth.go
```diff
@@ -527,7 +527,20 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+		return nil, xerrors.Errorf("failed to lookup tipset %s when constructing the eth txn receipt: %w", msgLookup.TipSet, err)
+	}
+
+	// The tx is located in the parent tipset
+	parentTs, err := a.Chain.LoadTipSet(ctx, ts.Parents())
+	if err != nil {
+		return nil, xerrors.Errorf("failed to lookup tipset %s when constructing the eth txn receipt: %w", ts.Parents(), err)
+	}
+
+	baseFee := parentTs.Blocks()[0].ParentBaseFee
```
aarshkshah1992, line 541: So this gives the base fee after executing the parent tipset of ` parentTs` which is the inclusion tipset for `tx` as `msgLookup.TipSet` is the execution tipset and `parentTs` = `Parent(msgLookup.TipSet)`.

This is correct because we want the base fee the user paid for their tx which is the network base fee before the inclusion tipset was mined.

node/impl/full/eth.go
```diff
@@ -575,23 +588,18 @@ func (a *EthModule) EthGetBlockReceiptsLimited(ctx context.Context, blockParam e
 		return nil, xerrors.Errorf("failed to load state tree: %w", err)
 	}
 
+	baseFee := ts.Blocks()[0].ParentBaseFee
```
aarshkshah1992, line 591: Same logic as above. Works because `ts` here is the "inclusion" tipset for the messages.

aarshkshah1992 (approved):

node/impl/full/eth.go
rvagg: yes, but
> base fee before the inclusion tipset was mined
this is the bit I'm not super confident about, but it _seems_ to be correct based on how everything is wired up today

Issue #12465, opened 2024-09-16
ChainIndexer should be able to GC at a cadence of epochs instead of days


aarshkshah1992:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

This is for https://github.com/filecoin-project/lotus/pull/12421#discussion_r1760555985.

At present, `ChainIndexer` can only gc `eth_tx_hash` at a cadence of days. However, we should change this so it has a cadence of epochs.

### Describe the solution you'd like

Update the ChainIndex GC logic and DDL queries so it can remove entries from the `eth_tx_hash` table at a cadence of epochs.

aarshkshah1992:
@akaladarshi Wanna pick up this one ?

akaladarshi:
> @akaladarshi Wanna pick up this one ?

Yeah sure, I can pick this up.

BigLep:
I updated the issue title to be more accurate (I think).  Please fix if I got it wrong.

aarshkshah1992:
@akaladarshi Please feel free to raise a PR on top of https://github.com/filecoin-project/lotus/pull/12450.

rjan90:
Closing ticket as completed by: https://github.com/filecoin-project/lotus/pull/12516

PR #12516, opened 2024-09-26
refactor: use epochs to gc eth tx hashes from chain indexer

akaladarshi:
Fixes issue: #12465, it changes:
- Update the remove `eth_tx_hash` query to delete before epoch instead of days
- refactor chain indexer `gc` function.


aarshkshah1992:
@akaladarshi Will review this on Monday. Thanks for raising this !

aarshkshah1992:
@akaladarshi Please can you raise this PR against https://github.com/filecoin-project/lotus/pull/12521  and change the GC test we already have for this change ?

chain/index/gc_test.go
```diff
@@ -11,57 +11,66 @@ import (
...
-	})
+	// all tipsets will be in future
+	tsKeyCid1, _, _ := insertRandomTipsetAtHeight(t, si, 1, false, genesisTime)
+	tsKeyCid2, _, _ := insertRandomTipsetAtHeight(t, si, 10, false, genesisTime)
+	tsKeyCid3, _, _ := insertRandomTipsetAtHeight(t, si, 50, false, genesisTime)
 
 	si.gc(ctx)
 
-	// tipset at height 1 and 10 should be removed
+	// tipset at height 1 data should be removed
```
aarshkshah1992, line 32: dont understand this comment. what's "height 1 data" ?

chain/index/gc_test.go
```diff
@@ -11,57 +11,66 @@ import (
...
+	err = si.stmts.getNonRevertedTipsetEventCountStmt.QueryRow(tsKeyCid2.Bytes()).Scan(&count)
+	require.NoError(t, err)
+	require.Equal(t, 0, count)
+
+	err = si.stmts.getNonRevertedTipsetEventEntriesCountStmt.QueryRow(tsKeyCid2.Bytes()).Scan(&count)
+	require.NoError(t, err)
+	require.Equal(t, 0, count)
+
 	// tipset at height 50 should not be removed
 	err = si.db.QueryRow("SELECT COUNT(*) FROM tipset_message WHERE height = 50").Scan(&count)
```
aarshkshah1992, line 60: lot of duplication here. Can we wrap this and the two sql statements below into a function and re-use it everywhere ?

chain/index/gc.go
```diff
@@ -75,17 +74,17 @@ func (si *SqliteIndexer) gc(ctx context.Context) {
...
 	// Also GC eth hashes
 
-	// Convert gcRetentionEpochs to number of days
-	gcRetentionDays := si.gcRetentionEpochs / (builtin.EpochsInDay)
-	if gcRetentionDays < 1 {
-		log.Infof("skipping gc of eth hashes as retention days is less than 1")
-		return
-	}
+	currHeadTime := time.Unix(int64(head.MinTimestamp()), 0)
+	retentionDuration := time.Duration(si.gcRetentionEpochs*builtin.EpochDurationSeconds) * time.Second
```
aarshkshah1992, line 78: Can you add a comment explaining this calculation here ?

Why do we need time.Duration(si.gcRetentionEpochs*builtin.EpochDurationSeconds) * time.Second ?

aarshkshah1992:
@akaladarshi Some quick comments.

chain/index/gc.go
```diff
@@ -75,17 +76,24 @@ func (si *SqliteIndexer) gc(ctx context.Context) {
...
 	// Also GC eth hashes
 
-	// Convert gcRetentionEpochs to number of days
-	gcRetentionDays := si.gcRetentionEpochs / (builtin.EpochsInDay)
-	if gcRetentionDays < 1 {
-		log.Infof("skipping gc of eth hashes as retention days is less than 1")
-		return
-	}
+	// Calculate the retention duration based on the number of epochs to retain.
+	// retentionDuration represents the total duration (in nano seconds) for which data should be retained before considering it for garbage collection.
```
aarshkshah1992, line 80: how is it in nanoseconds ?

chain/index/gc.go
```diff
@@ -75,17 +76,24 @@ func (si *SqliteIndexer) gc(ctx context.Context) {
...
+	// graceDuration represents the additional duration (in nano seconds) to retain data after the retention duration.
+	// Since time.Duration expects a nanosecond value, we multiply the total seconds by time.Second to convert it to nanoseconds.
+	retentionDuration := time.Duration(si.gcRetentionEpochs*builtin.EpochDurationSeconds) * time.Second
+	graceDuration := time.Duration(graceEpochs*builtin.EpochDurationSeconds) * time.Second
+
+	// Calculate the total duration to retain data.
+	totalRetentionDuration := retentionDuration + graceDuration
+	currHeadTime := time.Unix(int64(head.MinTimestamp()), 0)
+	// gcTime is the time that is (gcRetentionEpochs + graceEpochs) in nano seconds before currHeadTime
+	gcTime := currHeadTime.Add(-totalRetentionDuration)
```
aarshkshah1992, line 90: if this is less than or equal to 0, return without doing anything. Also please can we had a test for when `gcTime ` <= 0 ?

chain/index/gc.go
```diff
@@ -75,17 +76,24 @@ func (si *SqliteIndexer) gc(ctx context.Context) {
...
+
+	log.Infof("gc'ing eth hashes before time %s", gcTime.UTC().String())
 
-	log.Infof("gc'ing eth hashes older than %d days", gcRetentionDays)
-	res, err = si.stmts.removeEthHashesOlderThanStmt.ExecContext(ctx, "-"+strconv.Itoa(int(gcRetentionDays))+" day")
+	res, err = si.stmts.removeEthHashesBeforeTimeStmt.ExecContext(ctx, gcTime.Unix())
```
aarshkshah1992, line 94: why do we need to do `.Unix()` here ?

akaladarshi: I just wanted to provide as accurate time as possible, so I choose unix.

chain/index/gc.go
akaladarshi: It will be seconds 👍🏾 (I will update it)
so what is happening here is we are converting EpochDurationSeconds(30) to its time duration by multiplying with time.seconds.

akaladarshi: So `time.Duration` takes data in `nanoseconds`, 
so if I just multiply `time.Duration(rententionEpoch * EpochsDurationSeconds)`, it will give 600 (rententionEpoch = 20, EpochsDurationSeconds=30) but that will be in **nanoseconds**, to get proper time in seconds we have to multiply with `time.Seconds`

aarshkshah1992:
Thanks !

PR #12520, opened 2024-09-27
feat(eth): fix EthGetTransactionCount for pending block parameter

### LINKED ISSUE #10357: EthGetTransactionCount ignores input
The blkParam argument is ignored by EthGetTransactionCount, which instead always returns nonce based on the mempool. We should:

- only call into the mempool if "pending" is supplied
- get this value from State otherwise
- consider dropping the TSK param passed to the mempool's GetNonce method

virajbhartiya:
## Related Issues
Fixes #10357 

## Proposed Changes
1. Added a condition to check for the "pending" block parameter.
2. Implemented mempool nonce retrieval for "pending" blocks.
3. Included error handling for the mempool query.
4. Ensured the mempool is only accessed when necessary (i.e., for "pending" blocks).

## Checklist

- [ ] CI is green

### DIFF
--- a/itests/fevm_test.go
+++ b/itests/fevm_test.go
@@ -1252,8 +1252,8 @@ func TestEthGetTransactionCount(t *testing.T) {
 		// Check pending transaction count immediately after submission
 		pendingCount, err := client.EVM().EthGetTransactionCount(ctx, ethAddr, ethtypes.NewEthBlockNumberOrHashFromPredefined("pending"))
 		require.NoError(t, err)
-		require.GreaterOrEqual(t, pendingCount, ethtypes.EthUint64(i+1),
-			fmt.Sprintf("Pending transaction count should be at least %d, but got %d", i+1, pendingCount))
+		require.True(t, pendingCount == i || pendingCount == i+1,
+			fmt.Sprintf("Pending transaction count should be either %d or %d, but got %d", i, i+1, pendingCount))
 
 		// Wait for the transaction to be mined
 		_, err = client.EVM().WaitTransaction(ctx, lastHash)
--- a/itests/fevm_test.go
+++ b/itests/fevm_test.go
@@ -1252,7 +1252,7 @@ func TestEthGetTransactionCount(t *testing.T) {
 		// Check pending transaction count immediately after submission
 		pendingCount, err := client.EVM().EthGetTransactionCount(ctx, ethAddr, ethtypes.NewEthBlockNumberOrHashFromPredefined("pending"))
 		require.NoError(t, err)
-		require.True(t, pendingCount == i || pendingCount == i+1,
+		require.True(t, int(pendingCount) == i || int(pendingCount) == i+1,
 			fmt.Sprintf("Pending transaction count should be either %d or %d, but got %d", i, i+1, pendingCount))
 
 		// Wait for the transaction to be mined
--- a/itests/fevm_test.go
+++ b/itests/fevm_test.go
@@ -1252,7 +1252,8 @@ func TestEthGetTransactionCount(t *testing.T) {
 		// Check pending transaction count immediately after submission
 		pendingCount, err := client.EVM().EthGetTransactionCount(ctx, ethAddr, ethtypes.NewEthBlockNumberOrHashFromPredefined("pending"))
 		require.NoError(t, err)
-		require.Equal(t, ethtypes.EthUint64(i+1), pendingCount, "Pending transaction count should increment immediately")
+		require.GreaterOrEqual(t, pendingCount, ethtypes.EthUint64(i+1),
+			fmt.Sprintf("Pending transaction count should be at least %d, but got %d", i+1, pendingCount))
 
 		// Wait for the transaction to be mined
 		_, err = client.EVM().WaitTransaction(ctx, lastHash)
--- a/itests/fevm_test.go
+++ b/itests/fevm_test.go
@@ -1249,15 +1249,25 @@ func TestEthGetTransactionCount(t *testing.T) {
 		client.EVM().SignTransaction(tx, key.PrivateKey)
 		lastHash = client.EVM().SubmitTransaction(ctx, tx)
 
-		// Wait for the transaction to be mined
-		_, err := client.EVM().WaitTransaction(ctx, lastHash)
+		// Check pending transaction count immediately after submission
+		pendingCount, err := client.EVM().EthGetTransactionCount(ctx, ethAddr, ethtypes.NewEthBlockNumberOrHashFromPredefined("pending"))
 		require.NoError(t, err)
+		require.Equal(t, ethtypes.EthUint64(i+1), pendingCount, "Pending transaction count should increment immediately")
+
+		// Wait for the transaction to be mined
+		_, err = client.EVM().WaitTransaction(ctx, lastHash)
+		require.NoError(t, err)
+
+		// Check latest transaction count after the transaction is mined
+		latestCount, err := client.EVM().EthGetTransactionCount(ctx, ethAddr, ethtypes.NewEthBlockNumberOrHashFromPredefined("latest"))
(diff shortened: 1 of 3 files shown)


virajbhartiya:
@aarshkshah1992 I have added tests in `fevm_tests.go`. `EthGetTransactionCount` seemed to be tested also in `TestGetCodeAndNonce` but I have written a standalone test for it. Referenced https://github.com/filecoin-project/lotus/blob/3a90ce4107e203ea55f36b1e4baab3da85c6b320/itests/eth_bytecode_test.go#L22-L95 for implementing it

node/impl/full/eth.go
```diff
@@ -461,22 +461,37 @@ func (a *EthModule) EthGetTransactionHashByCid(ctx context.Context, cid cid.Cid)
 
 func (a *EthModule) EthGetTransactionCount(ctx context.Context, sender ethtypes.EthAddress, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthUint64, error) {
 	addr, err := sender.ToFilecoinAddress()
+
 	if err != nil {
-		return ethtypes.EthUint64(0), nil
+		return ethtypes.EthUint64(0), xerrors.Errorf("invalid address: %w", err)
```
rvagg, line 466: this will need to be checked against eth implementation behaviour - if you supply an invalid address, will it error or return nothing?

itests/fevm_test.go
```diff
@@ -1199,3 +1199,82 @@ func deployContractWithEth(ctx context.Context, t *testing.T, client *kit.TestFu
...
+		client.EVM().SignTransaction(tx, key.PrivateKey)
+		lastHash = client.EVM().SubmitTransaction(ctx, tx)
+
+		// Wait for the transaction to be mined
+		_, err := client.EVM().WaitTransaction(ctx, lastHash)
+		require.NoError(t, err)
+	}
+
+	// Get the latest transaction count
+	latestCount, err := client.EVM().EthGetTransactionCount(ctx, ethAddr, ethtypes.NewEthBlockNumberOrHashFromPredefined("latest"))
```
rvagg, line 1258: minor suggestion: you could also check on each transaction submission that it increments since you're waiting for it as well; perhaps also look at `"pending"` before you wait, but that's probably going to be flaky.

node/impl/full/eth.go
virajbhartiya: If an invalid address is provided (e.g., wrong length, non-hexadecimal characters, etc), eth returns an error.

rvagg: ACK, and it looks like the other methods in here that do this conversion also error here, so 👍

node/impl/full/eth.go
```diff
@@ -461,22 +461,37 @@ func (a *EthModule) EthGetTransactionHashByCid(ctx context.Context, cid cid.Cid)
 
 func (a *EthModule) EthGetTransactionCount(ctx context.Context, sender ethtypes.EthAddress, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthUint64, error) {
 	addr, err := sender.ToFilecoinAddress()
+
```
rvagg, line 464:
```suggestion
```

itests/fevm_test.go
```diff
@@ -1199,3 +1199,92 @@ func deployContractWithEth(ctx context.Context, t *testing.T, client *kit.TestFu
...
+			MaxPriorityFeePerGas: types.NanoFil,
+			GasLimit:             int(gaslimit),
+		}
+		client.EVM().SignTransaction(tx, key.PrivateKey)
+		lastHash = client.EVM().SubmitTransaction(ctx, tx)
+
+		// Check pending transaction count immediately after submission
+		pendingCount, err := client.EVM().EthGetTransactionCount(ctx, ethAddr, ethtypes.NewEthBlockNumberOrHashFromPredefined("pending"))
+		require.NoError(t, err)
+		require.Equal(t, ethtypes.EthUint64(i+1), pendingCount, "Pending transaction count should increment immediately")
```
rvagg, line 1255: Sorry, my bad, but I think this is at risk of being flaky - there's a small chance it gets into a block and out of the mpool between submission and GetTransactionCount; so how about instead of an exact check, you do a "between" `i` and `i+1`; either as two asserts, or as one `require.True` with that particular check and `fmt.Sprintf()` the message so we can see what it is.

virajbhartiya: I feel this should work @rvagg 

```go
pendingCount, err := client.EVM().EthGetTransactionCount(ctx, ethAddr, ethtypes.NewEthBlockNumberOrHashFromPredefined("pending"))
require.NoError(t, err)
require.GreaterOrEqual(t, pendingCount, ethtypes.EthUint64(i+1), fmt.Sprintf("Pending transaction count should be at least %d, but got %d", i+1, pendingCount))
```

rvagg: If you're doing a `>=` then you'd want the lower of possible values, which would be `i`, but you should also check for the weird cases where it's bigger than its should be, hence the need for an upper bound.

The lower bound of `i` is needed (I theorise) because you push the message into the mpool, but time has moved on between that action and the `GetTransactionCount`. Likely it's almost always going to remain in the mpool for that next call, but it's possible that a new block was minted between the two calls and it happened to move from the mpool into a new block. It's unlikely, but this is the nature of racy tests that you have to account for, running with `-race` may even pick it up occasionally. The aim should be to remove probability from the equation wherever possible and thereby remove the possibility of flaky tests.

So. `count` is almost always going to be `i+1` but it may also possibly be `i` in some cases.

Optionally just remove this check, it's nice to have in there so we can make sure `pending` doesn't do weird things, but if you feel it's getting too verbose then it's not a huge loss.

aarshkshah1992 (approved):
good stuff !

aarshkshah1992:
@rvagg Happy for me to merge this ?

itests/fevm_test.go
```diff
@@ -1199,3 +1199,93 @@ func deployContractWithEth(ctx context.Context, t *testing.T, client *kit.TestFu
...
+			GasLimit:             int(gaslimit),
+		}
+		client.EVM().SignTransaction(tx, key.PrivateKey)
+		lastHash = client.EVM().SubmitTransaction(ctx, tx)
+
+		// Check pending transaction count immediately after submission
+		pendingCount, err := client.EVM().EthGetTransactionCount(ctx, ethAddr, ethtypes.NewEthBlockNumberOrHashFromPredefined("pending"))
+		require.NoError(t, err)
+		require.GreaterOrEqual(t, pendingCount, ethtypes.EthUint64(i+1),
+			fmt.Sprintf("Pending transaction count should be at least %d, but got %d", i+1, pendingCount))
```
rvagg, line 1256:
```suggestion
		require.True(t, pendingCount == i || pendingCount == i+1,
			fmt.Sprintf("Pending transaction count should be either %d or %d, but got %d", i, i+1, pendingCount))
```

this is what I meant by being precise, the problem with "GreaterOrEqual" is that you're not going to catch weird cases where something might throw the `pending` result out, maybe it ends up being `i*2` or `99999999` for some odd reason, you wouldn't catch that.

virajbhartiya: thanks @rvagg , understood!

virajbhartiya:
I am unable to understand why `TestGatewayRateLimits` failed, it's passing when I run it locally

rvagg:
The gateway test is on me, unfortunately: https://github.com/filecoin-project/lotus/issues/12001#issuecomment-2344993355

Most of the flakies are catalogued in #12001 if you want to look them up. There are a few more not documented in there too: https://github.com/filecoin-project/lotus/issues?q=is%3Aissue+is%3Aopen+flaky

So you can ignore that one.

rvagg (approved):

rvagg:
🥳 thanks @virajbhartiya !

PR #12525, opened 2024-09-30
feat(actors): pack script can copy from local builds

rvagg:


rjan90 (approved):
That is a convenient way to get local builds compared to the env-var

PR #12529, opened 2024-09-30
feat: fix GetBlock for null rounds by returning nil

### LINKED ISSUE #10909: getBlock return an Error
### Checklist

- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.

### Lotus component

- [X] lotus JSON-RPC API
- [ ] Other

### Lotus Version

```text
All versions
```

### Repro Steps

1. curl --data '{"jsonrpc": "2.0", "id":1, "method": "eth_getBlockByNumber", "params": ["0x47C87",true] }' -H 'Content-Type:application/json' https://api.node.glif.io/rpc/v1

### Describe the Bug

I in https://api.hyperspace.node.glif.io/rpc/v1 call getBlock (294023) interface, return an Error: Error: requested epoch was a null round, Recommended return:
{
...prevBlock,
parentHash: prevBlock.hash,
blockNumber: prevBlock.blockNumber+1,
}

### Logging Information

```text
N/A
```

virajbhartiya:
## Related Issues
fixes #10909

## Proposed Changes
fix GetBlock for non-existent blocks by returning `nil`

## Additional Info
https://github.com/ethereum/go-ethereum
https://docs.infura.io/api/networks/ethereum/json-rpc-methods/eth_getblockbynumber

## Checklist

- [ ] CI is green


virajbhartiya:
@aarshkshah1992 Can you please give this a view

chain/store/store.go
```diff
@@ -863,10 +863,16 @@ func (cs *ChainStore) Contains(ctx context.Context, ts *types.TipSet) (bool, err
...
+			return nil // Block not found
+		}
 		blk, err = types.DecodeBlock(b)
 		return err
 	})
-	return blk, err
+	if err != nil {
+		return nil, err
+	}
+	return blk, nil
```
rvagg, line 875:
```suggestion
	if blk == nil {
		return nil, nil
	}
	return blk, nil
```

small nit, let's not return the typed `nil`, it can have surprising effects in the call chain where comparisons might come in to play

rvagg:
Can you add something in itests for this?

chain/store/store.go
```diff
@@ -863,10 +863,19 @@ func (cs *ChainStore) Contains(ctx context.Context, ts *types.TipSet) (bool, err
 func (cs *ChainStore) GetBlock(ctx context.Context, c cid.Cid) (*types.BlockHeader, error) {
```
aarshkshah1992, line 863: @virajbhartiya I don't think this is the right way to fix this. `GetBlock` is used across APIs and we shouldn't change it's behaviour.

This issue is specifically for the `EthGetBlockByNumber` API to return a (nil, nil) to the user for null rounds i.e. epochs for which the chain did not produce any block. See https://filecoinproject.slack.com/archives/CP50PPW2X/p1726047559393929 for what null rounds are.

aarshkshah1992: The fix is very simply. See https://github.com/filecoin-project/lotus/blob/13fb3223fcc9212dd7349c26cc782d0a5dc377ca/node/impl/full/eth_utils.go#L91.

`getTipsetByBlockNumber` already returns an `ErrNullRound` (would be good learning to understand how/why it determines an epoch to be a null round). If `getblockByNumber` sees this error -> it should return no block and no error to the user.

Let's also do some digging in the itests to see how we can identify and query for null rounds to assert this behaviour.

virajbhartiya: Oh okay, thanks for clarifying, I'll revert the changes made to `getBlock` and fix `ethGetBlockByNumber` to return accordingly if the epoch is a null round

virajbhartiya:
I have reverted back to the original state, I'll make the changes in `getTipsetByBlockNumber`

chain/store/store.go
virajbhartiya: This is how I thought of implementing the test
```go
func TestEthGetBlockByNumberNullRound(t *testing.T) {
	ctx, cancel, client := kit.SetupFEVMTest(t)
	defer cancel()

	// Find a null round
	head, err := client.ChainHead(ctx)
	require.NoError(t, err)

	var nullRound int64
	for i := head.Height(); i > 0; i-- {
		ts, err := client.ChainGetTipSetByHeight(ctx, i, types.EmptyTSK)
		require.NoError(t, err)
		if len(ts.Blocks()) == 0 {
			nullRound = int64(i)
			break
		}
	}

	if nullRound == 0 {
		t.Skip("No null round found in recent history, skipping test")
	}

	// Get the block for the null round
	block, err := client.EthGetBlockByNumber(ctx, strconv.FormatUint(uint64(nullRound), 10), false)
	require.NoError(t, err)
	require.Equal(t, ethtypes.EthBlock{}, block, "Expected empty block for null round")
}
```
But while running it is unable to find a null round. Is there some way to force a null round to test it out?
@aarshkshah1992 @rvagg

aarshkshah1992: @virajbhartiya You can see how null rounds can be injected at https://github.com/filecoin-project/lotus/blob/13fb3223fcc9212dd7349c26cc782d0a5dc377ca/itests/eth_transactions_test.go#L431.

I just searched through the itests folder to find this.

virajbhartiya:
Null rounds essentially gaps in the blockchain where no blocks are successfully mined for a particular epoch. These occur naturally in the protocol with a designed 0.7% probability due to the network's winner selection mechanism, though they can also happen if all Storage Providers who were selected as potential block miners for that epoch fail to produce or publish blocks. During a null round, while no blockchain data is created, the network's epoch clock continues to advance. This creates a gap in the chain's sequence. When a null round occurs, the next successful block will link back to the last valid tipset before the null round, effectively skipping over the empty epoch. This ensures network progression even when block production temporarily fails. And also null rounds differ from empty tipsets in that they don't exist in the chain state at all. 

Null rounds ->  when no block gets mined at a given epoch
Empty tipsets -> valid tipsets that contain no messages but have a CID and exist in the chain state

aarshkshah1992:
@virajbhartiya Great ! One more question -> given an epoch -> how can determine if it's a null round or not in Lotus ?

virajbhartiya:
That can be done by 
Trying to retrieve a tipset for your target epoch, if no tipset exists for that epoch, it's likely a null round and even if you get a tipset back it needs to be verified its height matches the target epoch. If the returned tipset's height doesn't match your target, it's a null round

Since null rounds don't exist in the chain state, we need to look for gaps in the chain. These gaps can be found by failing to find a tipset at that epoch

virajbhartiya:
To be able to return null values when a block is a null round

```diff
- func (a EthModule) EthGetBlockByNumber(ctx context.Context, blkParam string, fullTxInfo bool) (ethtypes.EthBlock, error)
+ func (a EthModule) EthGetBlockByNumber(ctx context.Context, blkParam string, fullTxInfo bool) (*ethtypes.EthBlock, error)
```
which will result in
```diff
- return newEthBlockFromFilecoinTipSet(ctx, ts, fullTxInfo, a.Chain, a.StateAPI)
+ return &block, nil
```

So the function of the definition of `EthGetBlockByNumber` will have to be changed everywhere. Is this okay or there's a better way of doing this?

aarshkshah1992:
@virajbhartiya Yes, we need to change the function signature. Document this in the `CHANGELOG` when you write the change log for this.

virajbhartiya:
I have updated `EthGetBlockByNumber` to return `nil` for a null round and also written itest for it and updated the ones using the function.

node/impl/full/eth.go
```diff
@@ -341,12 +341,29 @@ func (a *EthModule) EthGetBlockByHash(ctx context.Context, blkHash ethtypes.EthH
...
-		return ethtypes.EthBlock{}, err
+		if err == ErrNullRound {
+			// Return nil for null rounds
+			return nil, nil
+		}
+		return nil, xerrors.Errorf("failed to get tipset: %w", err)
+	}
+
+	// Check if the tipset is valid
+	if ts == nil || ts.Height() == 0 || len(ts.Blocks()) == 0 {
```
aarshkshah1992, line 356: why do we need these checks ?

itests/eth_transactions_test.go
```diff
@@ -454,8 +454,9 @@ func TestGetBlockByNumber(t *testing.T) {
 	}
 
 	// Fail when trying to fetch a null round.
-	_, err = client.EthGetBlockByNumber(ctx, (ethtypes.EthUint64(nullHeight)).Hex(), true)
-	require.Error(t, err)
+	block, err := client.EthGetBlockByNumber(ctx, (ethtypes.EthUint64(nullHeight)).Hex(), true)
```
aarshkshah1992, line 457: formatting

node/impl/full/eth.go
aarshkshah1992: Let's get rid of these.

virajbhartiya: - `ts == nil`: Ensures the tipset exists.
- `ts.Height() == 0`: Checks if it's the genesis block.
- `len(ts.Blocks()) == 0`: Ensures the tipset contains at least one block.

node/impl/full/eth.go
```diff
@@ -341,12 +341,29 @@ func (a *EthModule) EthGetBlockByHash(ctx context.Context, blkHash ethtypes.EthH
...
+		return nil, nil
 	}
-	return newEthBlockFromFilecoinTipSet(ctx, ts, fullTxInfo, a.Chain, a.StateAPI)
+
+	// Create an Ethereum block from the Filecoin tipset
+	block, err := newEthBlockFromFilecoinTipSet(ctx, ts, fullTxInfo, a.Chain, a.StateAPI)
```
aarshkshah1992, line 361: Why is `a.StateAPI` on a new line ? I think you need to fix the new line wrap in your editor.

node/impl/full/eth.go
```diff
@@ -341,12 +341,29 @@ func (a *EthModule) EthGetBlockByHash(ctx context.Context, blkHash ethtypes.EthH
 	return blk, nil
 }
 
-func (a *EthModule) EthGetBlockByNumber(ctx context.Context, blkParam string, fullTxInfo bool) (ethtypes.EthBlock, error) {
+func (a *EthModule) EthGetBlockByNumber(ctx context.Context, blkParam string, fullTxInfo bool) (*ethtypes.EthBlock, error) {
+	// Get the tipset for the specified block number
```
aarshkshah1992, line 345: for the specified block param.

itests/eth_api_test.go
```diff
@@ -362,6 +362,9 @@ func TestFilecoinAddressToEthAddressFinalised(t *testing.T) {
 func TestEthGetGenesis(t *testing.T) {
 	blockTime := 100 * time.Millisecond
 	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs(), kit.ThroughRPC())
+	require.NotNil(t, client, "Client should not be nil")
+	require.NotNil(t, ens, "Ensemble should not be nil")
+
```
aarshkshah1992, line 367: dont need these two

itests/eth_api_test.go
```diff
@@ -370,15 +373,28 @@ func TestEthGetGenesis(t *testing.T) {
...
+	// Check if the block is nil (null round)
+	if ethBlk == nil {
+		t.Log("Genesis block is a null round, skipping EthBlock checks")
+	} else {
+		require.NotNil(t, ethBlk, "EthBlock should not be nil")
+	}
+
 	genesis, err := client.ChainGetGenesis(ctx)
 	require.NoError(t, err)
+	require.NotNil(t, genesis, "Genesis block should not be nil")
```
aarshkshah1992, line 385: why do we need this ?

itests/eth_api_test.go
```diff
@@ -370,15 +373,28 @@ func TestEthGetGenesis(t *testing.T) {
...
 
 	genesisCid, err := genesis.Key().Cid()
 	require.NoError(t, err)
+	require.NotEmpty(t, genesisCid, "Genesis CID should not be empty")
```
aarshkshah1992, line 389: why do we need this ?

itests/eth_api_test.go
```diff
@@ -370,15 +373,28 @@ func TestEthGetGenesis(t *testing.T) {
...
 
 	genesisHash, err := ethtypes.EthHashFromCid(genesisCid)
 	require.NoError(t, err)
-	require.Equal(t, ethBlk.Hash, genesisHash)
+	require.NotEmpty(t, genesisHash, "Genesis hash should not be empty")
```
aarshkshah1992, line 393: dont need this

itests/eth_api_test.go
```diff
@@ -370,15 +373,28 @@ func TestEthGetGenesis(t *testing.T) {
...
+
+	if ethBlk != nil {
+		require.Equal(t, ethBlk.Hash, genesisHash)
```
aarshkshah1992, line 396: ??

itests/eth_block_hash_test.go
```diff
@@ -65,18 +65,26 @@ func TestEthBlockHashesCorrect_MultiBlockTipset(t *testing.T) {
 		require.NoError(t, err)
 
 		ethBlockA, err := n2.EthGetBlockByNumber(ctx, hex, true)
-		// Cannot use static ErrFullRound error for comparison since it gets reserialized as a JSON RPC error.
-		if err != nil && strings.Contains(err.Error(), "null round") {
-			require.Less(t, ts.Height(), abi.ChainEpoch(i), "did not expect a tipset at epoch %d", i)
+		if err != nil {
+			if strings.Contains(err.Error(), "null round") {
```
aarshkshah1992, line 69: why have we changed this code ?

itests/eth_block_hash_test.go
```diff
@@ -65,18 +65,26 @@ func TestEthBlockHashesCorrect_MultiBlockTipset(t *testing.T) {
...
+				continue
+			}
+			require.NoError(t, err)
+		}
+
+		// Check if the tipset height is less than or equal to the current epoch
+		require.LessOrEqual(t, ts.Height(), abi.ChainEpoch(i), "tipset height should not exceed the current epoch")
+
+		// Skip Null rounds
+		if ethBlockA == nil {
```
aarshkshah1992, line 79: ??

aarshkshah1992 (commented):
@virajbhartiya Reviewed.

itests/eth_api_test.go
```diff
@@ -408,11 +424,12 @@ func TestEthBlockNumberAliases(t *testing.T) {
 
 	head := client.WaitTillChain(ctx, kit.HeightAtLeast(policy.ChainFinality+100))
 	// latest should be head-1 (parents)
-	var latestEthBlk ethtypes.EthBlock
+	var latestEthBlk *ethtypes.EthBlock
 	for {
 		var err error
 		latestEthBlk, err = client.EVM().EthGetBlockByNumber(ctx, "latest", true)
 		require.NoError(t, err)
+
```
aarshkshah1992, line 432: dont need this

itests/eth_api_test.go
```diff
@@ -430,7 +447,7 @@ func TestEthBlockNumberAliases(t *testing.T) {
 	// safe should be latest-30
 	safeEthBlk, err := client.EVM().EthGetBlockByNumber(ctx, "safe", true)
 	require.NoError(t, err)
-	diff = int64(latestEthBlk.Number-30) - int64(safeEthBlk.Number)
+	diff = int64(latestEthBlk.Number) - int64(30) - int64(safeEthBlk.Number)
```
aarshkshah1992, line 450: ??

node/impl/full/eth.go
virajbhartiya: Understood

itests/eth_api_test.go
virajbhartiya: I'll remove the specific checks for genesis blocks in the changes that I have made

itests/eth_block_hash_test.go
virajbhartiya: Oh sorry didn't mean to push this, I was testing out something else

aarshkshah1992 (approved):
lgtm

itests/fevm_test.go
```diff
@@ -1289,3 +1290,80 @@ func TestEthGetTransactionCount(t *testing.T) {
...
+	kit.SendFunds(ctx, t, client, filAddr, types.FromFil(10))
+
+	// Test getting the latest block
+	latest, err := client.EthBlockNumber(ctx)
+	require.NoError(t, err)
+	latestBlock, err := client.EthGetBlockByNumber(ctx, "latest", true)
+	require.NoError(t, err)
+	require.NotNil(t, latestBlock)
+	require.Equal(t, latest, latestBlock.Number)
+
```
aarshkshah1992, line 1315: @virajbhartiya The above is racy. What if a new block is created between the calls to `EthBlockNumber` and `EthGetBlockByNumbe` by the test miner ? These calls will return different blocks then.

I think just asking for the "latest" to `EthGetBlockByNumber` and making sure it's not nil and no error is fine.

itests/fevm_test.go
```diff
@@ -1289,3 +1290,80 @@ func TestEthGetTransactionCount(t *testing.T) {
...
+	// Test getting a specific block by number
+	specificBlock, err := client.EthGetBlockByNumber(ctx, latest.Hex(), true)
+	require.NoError(t, err)
+	require.NotNil(t, specificBlock)
+	require.Equal(t, latest, specificBlock.Number)
```
aarshkshah1992, line 1320: remove this `require.Equal(t, latest, specificBlock.Number)` check because it suffers from the same problem as above. What if a new block is minded by the test miner before you call `EthGetBlockByNumber(latest.hex())` ?

aarshkshah1992 (commented):
@virajbhartiya Last set of comments.

itests/fevm_test.go
virajbhartiya: Oh yes I didn't take into consideration about blocks being mined between calls. I'll fix this!

virajbhartiya:
@aarshkshah1992 I have made the changes suggested in you review and modified the tests.

aarshkshah1992 (approved):
lgtm

rvagg:
https://github.com/filecoin-project/lotus/pull/12641

sorry, reverts aren't fun to have done to you

PR #12531, opened 2024-09-30
feat: implement ticket based F3 participation lease

masih:
## Related Issues
* Fixes https://github.com/filecoin-project/go-f3/issues/599

## Proposed Changes
Implemented enhanced ticket-based participation system for F3 consensus in `F3Participate`. This update introduces a new design where participation tickets grant a temporary lease, allowing storage providers to sign as part of the F3 consensus mechanism. This design ensures that tickets are checked for validity and issuer alignment, handling errors
robustly. If there's an issuer mismatch, the system advises miners to retry with the existing ticket. If the ticket is invalid or expired, miners are directed to obtain a new ticket via `F3GetOrRenewParticipationTicket`.

## Checklist

- [x] CI is green


chain/lf3/f3.go
```diff
@@ -133,43 +136,44 @@ func (fff *F3) runSigningLoop(ctx context.Context) {
...
 				}
 			}
 		}
 	}
 }
 
+func (fff *F3) GetOrRenewParticipationToken(ctx context.Context, minerID uint64, previous api.F3ParticipationToken) (api.F3ParticipationToken, error) {
+	select {
+	case <-ctx.Done():
+		return api.F3ParticipationToken{}, ctx.Err()
```
Stebalien, line 163: Any reason to check the context?

api/api_full.go
```diff
@@ -936,6 +956,39 @@ type FullNode interface {
...
+// subject, and the expiration instance.
+type F3ParticipationLease struct {
+	// Session uniquely identifies the session in which this lease is issued. This
+	// information can then be used to infer if the issuer of the token has restarted
+	// since the lease was issued.
+	Session uuid.UUID
+	// Issuer is the identity of the node that issued the lease.
+	Issuer peer.ID
+	// Subject is the actor ID of the lease holder.
+	Subject uint64
```
Stebalien, line 978: I'd just call it `MinerID` to reduce confusion.

api/api_full.go
```diff
@@ -936,6 +956,39 @@ type FullNode interface {
...
+	// ExpireAfterInstance is the GPBFT instance ID, beyond which this lease expires.
+	// See F3Progress.
+	ExpireAfterInstance uint64
```
Stebalien, line 981: I _think_ we need a "not before" as well, right? I.e., we need to grant a lease between instances A and B.

node/impl/full/f3.go
```diff
@@ -23,67 +22,74 @@ type F3API struct {
...
-		return false, ErrF3Disabled
+		return false, api.ErrF3Disabled{}
 	}
 	return f3api.F3.IsRunning(), nil
 }
 
 func (f3api *F3API) F3GetECPowerTable(ctx context.Context, tsk types.TipSetKey) (gpbft.PowerEntries, error) {
 	if f3api.F3 == nil {
-		return nil, ErrF3Disabled
+		return nil, api.ErrF3Disabled{}
```
Stebalien, line 76: What if we:

1. Make this error type private.
2. Expose a public instance of it (`var ErrF3Disabled errF3Disabled`)?

I _think_ we'd still be able to use it with jsonrpc and it would make this code a little cleaner.

Stebalien (commented):
Some code nits and a few comments, but I'm a bit brain-dead from debugging tests all day so I'll have to leave the full review for tomorrow (hopefully @Kubuxu can get to it before I do).

node/impl/full/f3.go
masih: That was my initial thought but was unsure if it would play nice with RPC errors. Going to give this a go.

api/api_full.go
masih: I had that initially but removed it for simplicity. Without it I think the design would work correctly and miners renew within the last 2 instances.

Without a strict boundary i think the token design as is would still work correctly. Because, each node will only give one lease out per miner, and would only renew if the old one is expired or was issued by itself.

With it the API is more defensive against miners that call to get a token too often, but the entire implementation here is based on the assumption that miner and node are trusted.

Right?

chain/lf3/f3.go
masih: To respect the context that is passed through even if successive calls don't take a context? I think that's better than ignoring the context entirely. But I have no strong opinions.

Stebalien: We're not blocking, right? IMO, checking the context makes sense if you have a long-running loop, block, etc. Otherwise, it's up to the caller to deal with it.

api/api_full.go
Stebalien: My concern is that "expired" is relative. E.g.:

1. Miner sees that the latest instance is X according to lotus A.
2. Miner gives a lease to lotus B, but lotus B thinks that the latest instance is X-1.

masih: Got it thanks

I think this won't cause problems because the miners renew lease within the last 2 instances remaining in the lease. But that's not enforced by the Lotus node API. 

I'll enforce it 👍

Stebalien: Oh, I see. IMO, this API is backwards: the _Miner_ should be the one in control, not the node. 

That is:

1. Miner calls "pre-participate" and gets a peer id, session id (unnecessary?), and the node's current instance.
2. Miner calls "participate" with a not-before instance, not-after instance, and a peer ID.

Importantly, the Miner decides the instance range to lease out, not the node.

Stebalien: Hm. Well, as long as the "not before" is encoded in the ticket this should be fine. But I still think the miner should be able to specify the "not before" time when getting the ticket, at least.

masih: > the Miner should be the one in control, not the node.

Why?  I mean that's valid too but why is that better than node deciding the lease range?

Stebalien: But in that case, you're right. Given that this is the lease as specified by the _node_, there's no reason to have a "not before".

masih: > miner should be able to specify the "not before"

Thinking about it more, I agree we should add this. Because, the F3 progress across nodes may not be exactly the same. I think i also need to encode currentInstance in the ticket response in couple of places.


While we are on the subject, do we care about the session at all? Right now it is a noop apart from the informative log line to indicate that the node must have restarted.

api/api_full.go
```diff
@@ -936,6 +957,39 @@ type FullNode interface {
...
+type F3ParticipationTicket []byte
+
+// F3ParticipationLease defines the lease granted to a storage provider for
+// participating in F3 consensus, detailing the session identifier, issuer,
+// subject, and the expiration instance.
+type F3ParticipationLease struct {
+	// Session uniquely identifies the session in which this lease is issued. This
+	// information can then be used to infer if the issuer of the ticket has restarted
+	// since the lease was issued.
+	Session uuid.UUID
```
Kubuxu, line 975: Do we need this to be `uuid.UUID` (it is a new dependency for API users)? Could it be just a string?

Kubuxu:
I think I would like to see the ability of the Miner to set the maximum term of the lease. Fundamentally, the Miner is in control of what they give out.

api/api_full.go
masih: > it is a new dependency for API users

`Common.Session` API already depends on it. 

For the purposes of this PR it's primarily for an informative log line when a node restarts. Otherwise has no effect on the leasing functionality. Happy to remove it and add later if we are not bothered about it.

chain/lf3/f3.go
masih: On the subjective matters: It is my opinion that if context exist it should be respected. e.g. if an expired context is passed to a function, regardless of whether it is blocking or not, it should suspend execution.

But I will revert the changes here and ignore the context since it seems this repo does not follow that practice.

api/api_full.go
masih: Removed Session. We can add later if there is a need for it beyond logging.

masih:
> the ability of the Miner to set the maximum term

Added, where the miner sets the validity term, as well as an upper maximum dictated by node for safety.

Kubuxu (approved):

chain/lf3/participation_lease.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (l *leaser) participate(ticket api.F3ParticipationTicket) (api.F3ParticipationLease, error) {
+	currentInstance, _, _ := l.progress()
+	lease, err := l.validate(currentInstance, ticket)
+	if err != nil {
+		return api.F3ParticipationLease{}, err
+	}
+	l.mutex.Lock()
+	l.leases[lease.MinerID] = lease
```
Stebalien, line 85: I think we need to look at the current lease. E.g., in case wires get crossed and an old call to participate gets stuck somewhere.

Storing an _expired_ lease is fine, we'll clean it up later. But clobbering a _good_ lease with an older lease is probably not something we want to do.

node/modules/storageminer.go
```diff
@@ -357,76 +356,187 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+	for ctx.Err() == nil {
+		if ticket, err := p.tryGetF3ParticipationTicket(ctx); err != nil {
+			return err
+		} else if lease, participating, err := p.tryF3Participate(ctx, ticket); err != nil {
+			return err
+		} else if !participating {
+			continue
+		} else if err := p.awaitLeaseExpiry(ctx, lease); err != nil {
+			return err
 		}
```
Stebalien, line 392: Do we really want to exit this loop if we hit an error?

node/modules/storageminer.go
```diff
@@ -357,76 +356,187 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
-			newLease := time.Now().Add(leaseTime)
+func (p *f3Participator) tryGetF3ParticipationTicket(ctx context.Context) (api.F3ParticipationTicket, error) {
+	p.backoff.Reset()
+	for ctx.Err() == nil {
+		switch ticket, err := p.node.F3GetOrRenewParticipationTicket(ctx, p.participant, p.previousTicket, p.leaseTerm); {
+		case ctx.Err() != nil:
+			return api.F3ParticipationTicket{}, ctx.Err()
+		case errors.Is(err, api.ErrF3Disabled):
+			log.Errorw("Cannot participate in F3 as it is disabled.", "err", err)
+			return api.F3ParticipationTicket{}, xerrors.Errorf("acquiring F3 participation ticket: %w", err)
```
Stebalien, line 406: Won't this be the case while bootstrapping, switching networks, etc?

node/modules/storageminer.go
Stebalien: Or are we saying that all errors are unrecoverable and that these APIs should only return errors when recovery is impossible?

node/modules/storageminer.go
```diff
@@ -357,76 +356,187 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+		// to 13.
+		checkProgressMaxAttempts = 13
+		// checkProgressInterval defines the duration between progress checks in normal operation mode.
+		// This interval is used when there are no errors in retrieving the current progress.
+		checkProgressInterval = 10 * time.Second
+		// leaseTerm The number of instances the miner will attempt to lease from nodes.
+		leaseTerm = 5
+	)
+
+	go func() {
```
Stebalien, line 520: It would be really nice (especially for testing) if we could wait for this to stop on exit (e.g., with a waitgroup plus an exit hook).

node/modules/storageminer.go
```diff
@@ -357,76 +356,187 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
-			timer.Reset(leaseTime / 2)
-			select {
-			case <-ctx.Done():
-				return
-			case <-timer.C:
-			}
+func (p *f3Participator) backOffFor(ctx context.Context, d time.Duration) {
+	if !p.timer.Stop() {
+		<-p.timer.C
+	}
```
Stebalien, line 487: This will deadlock if the timer already fired and we already read from the channel.

Stebalien: I believe we just need to create a new timer each time (and don't store it anywhere).

node/modules/storageminer.go
```diff
@@ -357,76 +356,187 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+	p.backoff.Reset()
+	for ctx.Err() == nil {
+		switch progress, err := p.node.F3GetProgress(ctx); {
+		case errors.Is(err, api.ErrF3Disabled):
+			log.Errorw("Cannot await F3 participation lease expiry as F3 is disabled.", "err", err)
+			return xerrors.Errorf("awaiting F3 participation lease expiry: %w", err)
+		case err != nil:
+			if p.backoff.Attempt() > float64(p.maxCheckProgressAttempts) {
+				log.Errorw("Too many failures while attempting to check F3  progress. Restarting participation.", "attempts", p.backoff.Attempt(), "err", err)
+				return nil
```
Stebalien, line 464: returning nil?

Stebalien (changes requested):

chain/lf3/participation_lease.go
masih: > in case wires get crossed and an old call to participate gets stuck somewhere
> clobbering a good lease with an older lease

If we reach this line it means the the lease is "valid":
* decodes
* has not expired
* conforms to max leased instances accepted by node (that is controlled by Miner, see earlier feedback)

What specific edge case you have in mind Steb?

node/modules/storageminer.go
masih: > Do we really want to exit this loop if we hit an error?

Yes because `awaitLeaseExpiry` exits with error only if either [F3 is disabled](https://github.com/filecoin-project/lotus/blob/3d12ca5c53c5df74bb9daa0ea9be27b283e2469c/node/modules/storageminer.go#L460) or [context has an error](https://github.com/filecoin-project/lotus/blob/3d12ca5c53c5df74bb9daa0ea9be27b283e2469c/node/modules/storageminer.go#L477).

> APIs should only return errors when recovery is impossible?

The APIs only return error when recovery is not possible, yes.

node/modules/storageminer.go
masih: I am not sure. 

> bootstrapping
> switching networks

Good point: Do we disable F3 or pause the inner runner? 🤔 

I expect that the F3 is enabled when the miner starts. Does/Should Lotus support runtime reconfiguration at runtime for `F3Enabled` flag?

Looking at [the API implementation](https://github.com/filecoin-project/lotus/blob/3d12ca5c53c5df74bb9daa0ea9be27b283e2469c/node/impl/full/f3.go) the `api.ErrF3Disabled` is returned when `lf3.F3` is nil, which translates to "Not injected by DI" I believe, which means `F3Enabled` must have been false. If Lotus does not support runtime update for this config then I think the execution path for both node and the miner is that should be enabled before the process starts. Which means `lf3.F3` will not be nil prior to bootstrapping or at the time of network switch. Which means `api.ErrF3Disabled` will not be returned. 

Right?

node/modules/storageminer.go
masih: > This will deadlock if the timer already fired and we already read from the channel.

Despite the fact that backoff is never called from multiple goroutines?

> I believe we just need to create a new timer each time (and don't store it anywhere).

Will do, at the price of higher GC which i guess doesn't matter considering the backoff call frequency.

node/modules/storageminer.go
masih: Yes, this is intentional to abandon current lease and start participation from scratch. I chose the [max attempts](https://github.com/filecoin-project/lotus/blob/3d12ca5c53c5df74bb9daa0ea9be27b283e2469c/node/modules/storageminer.go#L503-L512) value such that we can be confident the abandoned lease is expired when the participation is restarted with a new lease.

The rationale for restarting participation to begin with is that this error can happen if lotus node is down due to operational issues: maybe it is being restarted etc. Returning error here means the miner will drop out of participation entirely until it is restarted. which I believe is not what we want. Unless I have missed something?

node/modules/storageminer.go
Stebalien: Ah, I see. You're right.

node/modules/storageminer.go
Stebalien: This code sketches me out.

1. We should never make decisions based on the instance the lotus node we _happen_ to be talking to is at.
2. We should never rely on "enough time".

Instead, we should always:

1. Keep track of the last instance we agreed to participate in.
2. Make sure any lotus node we try to renew the ticket on has that information (e.g., has the old ticket).

Stebalien: Ah, I see. We'll _always_ keep the previous ticket anyways. In that case, this seems totally reasonable.

node/modules/storageminer.go
```diff
@@ -357,78 +357,193 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+			checkProgressInterval,
+			leaseTerm,
+		)
+
+		switch err := participator.participate(ctx); {
+		case err == nil, errors.Is(err, context.Canceled):
+			log.Infof("Stopped participating in F3")
+		default:
+			log.Errorw("F3 participation stopped abruptly", "err", err)
 		}
```
Stebalien, line 537: can we wait for `OnStart` here? It would also be kind of nice to write this more like a service (with start/stop methods).

node/modules/storageminer.go
```diff
@@ -357,78 +357,193 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+	go func() {
+		defer wg.Done()
+
+		participator := newF3Participator(
+			node,
+			participant,
+			&backoff.Backoff{
+				Min:    1 * time.Second,
+				Max:    1 * time.Minute,
+				Factor: 1.5,
```
Stebalien, line 525: Is there any way we could base these off of, e.g., the block time?

node/modules/storageminer.go
```diff
@@ -357,78 +357,193 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
+	// called from multiple goroutines.
+	timer := time.NewTimer(d)
+	select {
+	case <-ctx.Done():
+		return
+	case <-timer.C:
+	}
+}
+
+func F3Participation(mctx helpers.MetricsCtx, lc fx.Lifecycle, node v1api.FullNode, participant dtypes.MinerAddress) error {
```
Stebalien, line 495: Can we use this in the itest?

node/modules/storageminer.go
```diff
@@ -357,78 +357,193 @@ func SectorStorage(mctx helpers.MetricsCtx, lc fx.Lifecycle, lstor *paths.Local,
...
-			select {
-			case <-ctx.Done():
-				return
-			case <-timer.C:
-			}
+func (p *f3Participator) backOffFor(ctx context.Context, d time.Duration) {
+	// Create a timer every time to avoid potential risk of deadlock or the need for
+	// mutex despite the fact that f3Participator is never (and should never) be
+	// called from multiple goroutines.
+	timer := time.NewTimer(d)
```
Stebalien, line 487: We should defer stop this for better garbage collection (IMO).

chain/lf3/participation_lease.go
Stebalien: IIRC, we allow slightly overlapping leases, right? We could end up in a situation where we:

1. Try to renew a lease for T1-T5. That gets stuck.
2. Try to renew the lease, now from T2-T6. That goes through.
3. The lease for T1-T5 goes through, replacing the longer lease.

I'm not sure if this is possible in the current architecture, but that's the kind of thing I'm a bit worried about.

masih: I see, interesting.

Over TCP connections I I am not sure if that's possible, unless there is some built-in retry in json rpc or at infra level that would blindly retry requests on behalf of nodes? I would then wonder how the response delivery would work. 

In any case, better safe than sorry. Thanks for pointing that out Steven

Stebalien: 1. Miner makes the first request, it gets all the way to the point where we'd record the lease then pauses.
2. Connection fails (miner reboots, etc.).
3. Miner makes the second request.
4. Lotus serves the second request.
5. The first request continues.

node/modules/storageminer.go
masih: I am curious, why would that matter?

node/modules/storageminer.go
masih: I _think_ [we already do](https://github.com/filecoin-project/lotus/blob/63595bb73df0b673eeaa31a0d7dcba7d2823d077/itests/f3_test.go#L192)?

node/modules/storageminer.go
Stebalien: > Despite the fact that backoff is never called from multiple goroutines?

 In this case, it would have deadlocked precisely the second time we call it because it would have expired the first time and we would have read from the channel. So `Stop()` would have returned false (nothing to stop) and reading from the channel would have blocked forever.

node/modules/storageminer.go
masih: @Stebalien could you check my [handwriting](https://github.com/filecoin-project/lotus/pull/12531/commits/ec3b3f03950659872dbb69acd5fbcf19bf091a9f) on this one please?

node/modules/storageminer.go
Stebalien: Oh, I see. I thought this was new.

node/modules/storageminer.go
Stebalien: I'm mostly worried about test scenarios where we might run things at 100ms blocktimes. In that case, a min of 1s os _way_ too long.

Stebalien: Note: this is about as far from critical as it gets.

node/modules/storageminer.go
```diff
@@ -512,33 +516,35 @@ func F3Participation(mctx helpers.MetricsCtx, lc fx.Lifecycle, node v1api.FullNo
...
+					participant,
+					&backoff.Backoff{
+						Min:    1 * time.Second,
+						Max:    1 * time.Minute,
+						Factor: 1.5,
+					},
+					checkProgressMaxAttempts,
+					checkProgressInterval,
+					leaseTerm,
+				)
```
Stebalien, line 536: I'd actually _construct_ this outside, but either way is fine.

node/modules/storageminer.go
```diff
@@ -512,33 +516,35 @@ func F3Participation(mctx helpers.MetricsCtx, lc fx.Lifecycle, node v1api.FullNo
...
+
+				switch err := participator.participate(ctx); {
+				case err == nil, errors.Is(err, context.Canceled):
```
Stebalien, line 539: nit: I generally avoid looking at the error itself and instead check `ctx.Err` and/or `ctx.Done`.

1. Services will occasionally fail, e.g., because some internal context gets canceled (shouldn't happen, but does).
2. Canceling the context can sometimes cause other follow-on errors which should generally be ignored (although I guess we do want to, e.g., learn about shutdown errors if they involve failing to record state, etc.).

chain/lf3/participation_lease.go
masih: I see thanks, you are right.

The latest commit should fix this such that the leaser never accepts a ticket with start instance before the previous lease. Please let know if I have missed anything :pray:

Stebalien (approved):
LGTM modulo a few nits that you're free to ignore, and some failing lints.

node/modules/storageminer.go
masih: The `f3Participator` construct is new. it refactors the logic that was in a loop inside `F3Participation` into its own type.

node/modules/storageminer.go
masih: > I'm mostly worried about test scenarios where we might run things at 100ms blocktimes. In that case, a min of 1s os _way_ too long.

ah that makes sense

Stebalien (approved):

PR #12553, opened 2024-10-04
fix: handle error and revert data in EthEstimateGas and EthCall

### LINKED ISSUE #10311: Eth API: On revert, return data in the error
### Checklist

### Lotus component

- [ ] lotus client
- [X] lotus JSON-RPC API
- [ ] Other

### Improvement Suggestion

JSON-RPC supports an optional `data` field in errors to return additional information. Furthermore, it looks like Ethereum clients _use_ this field to return "revert reasons".

Specifically, we have an error type defined here:

https://github.com/filecoin-project/go-jsonrpc/blob/787a96afc5da60084c97b597b61f0fad18c65a3d/handler.go#L68-L72

```go
type respError struct {
	Code    ErrorCode       `json:"code"`
	Message string          `json:"message"`
	Meta    json.RawMessage `json:"meta,omitempty"`
}
```

We should extend this to:

```go
type respError struct {
	Code    ErrorCode       `json:"code"`
	Message string          `json:"message"`
	Meta    json.RawMessage `json:"meta,omitempty"`
	Data    json.RawMessage `json:"data,omitempty"`
}
```

And provide a way to return said data from a JSON-RPC method. Likely the best way is to define `interface { ErrorData() interface{} }`.

virajbhartiya:
## Related Issues
Closes #10311

## Proposed Changes
Modified `EthCall` and `EthEstimateGas` to handle errors and revert data

## Checklist

- [ ] CI is green

### DIFF
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -177,6 +177,15 @@ type EthAPI struct {
 	EthEventAPI
 }
 
+type EthCallError struct {
+	Message string
+	Data    []byte
+}
+
+func (e *EthCallError) Error() string {
+	return e.Message
+}
+
 var ErrNullRound = errors.New("requested epoch was a null round")
 
 func (a *EthModule) StateNetworkName(ctx context.Context) (dtypes.NetworkName, error) {
@@ -1462,10 +1471,14 @@ func (a *EthModule) EthEstimateGas(ctx context.Context, p jsonrpc.RawParams) (et
 		// guts of EthCall). This will give us an ethereum specific error with revert
 		// information.
 		msg.GasLimit = buildconstants.BlockGasLimit
-		if _, err2 := a.applyMessage(ctx, msg, ts.Key()); err2 != nil {
+		invokeResult, err2 := a.applyMessage(ctx, msg, ts.Key())
+		if err2 != nil {
 			err = err2
 		}
-		return ethtypes.EthUint64(0), xerrors.Errorf("failed to estimate gas: %w", err)
+		return ethtypes.EthUint64(0), &EthCallError{
+			Message: fmt.Sprintf("failed to estimate gas: %s", err),
+			Data:    []byte(parseEthRevert(invokeResult.MsgRct.Return)),
+		}
 	}
 
 	expectedGas, err := ethGasSearch(ctx, a.Chain, a.Stmgr, a.Mpool, gassedMsg, ts)
@@ -1613,7 +1626,10 @@ func (a *EthModule) EthCall(ctx context.Context, tx ethtypes.EthCall, blkParam e
 
 	invokeResult, err := a.applyMessage(ctx, msg, ts.Key())
 	if err != nil {
-		return nil, err
+		return nil, &EthCallError{
+			Message: err.Error(),
+			Data:    []byte(parseEthRevert(invokeResult.MsgRct.Return)),
+		}
 	}
 
 	if msg.To == builtintypes.EthereumAddressManagerActorAddr {


virajbhartiya:
@aarshkshah1992 I am using https://github.com/filecoin-project/lotus/blob/0e7292acc708a23ddf238dae3c27767cc6728513/node/impl/full/eth_utils.go#L333 to get the reverted reason for the Data field.

node/impl/full/eth.go
```diff
@@ -177,6 +177,15 @@ type EthAPI struct {
 	EthEventAPI
 }
 
+type EthCallError struct {
+	Message string
+	Data    []byte
+}
+
+func (e *EthCallError) Error() string {
+	return e.Message
```
aarshkshah1992, line 186: But how will the `go-jsonrpc` library extract the data here to then send it to the client ?

aarshkshah1992:
Please see https://filecoinproject.slack.com/archives/CP50PPW2X/p1727757229039759?thread_ts=1727699414.185139&cid=CP50PPW2X

aarshkshah1992:
Best way to test this is to write an itest that deploys a contract that does a divide by 0 (so there's a revert) and see what error you get when you call `EthCall` in your itests for the block which has the execution data for that contract. I think you'll then see what I'm talking about. The client should see the revert reason in the `data` field of the error.

BigLep:
@virajbhartiya: I assume you'll re-request review when you're ready for review again.

virajbhartiya:
Hey @BigLep, yes I'll request a re-review, currently I'm running into a few issues while testing out on devnet so working on fixing those.

akaladarshi:
@aarshkshah1992 me and @virajbhartiya was discussing about returning `execution reverted` as a message and adding reason behind it as data.

The changes I have pushed are not complete, need to get your thoughts on it, then will finish the changes.

aarshkshah1992:
@akaladarshi Would be great to first get this working e2e on a calibnet node and we can then discuss the exact semantics of the error message.

Also, thanks to both for this !

akaladarshi:
> node

@aarshkshah1992 here is the `eth_estimateGas` response after the changes

Screenshot 2024-10-11 at 5 20 04 PM


Here is the response from ethereum node for same data

Screenshot 2024-10-11 at 5 24 44 PM

aarshkshah1992:
@akaladarshi @virajbhartiya Great stuff ! That looks correct. Can you also confirm that we ONLY get this data field for contract reverts and that the response for all other errors is the same as it was before ? Once you confirm that, I think we can implement this for the other API as well. Thanks a lot.

CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
 # UNRELEASED
 
 ## New features
+
+* Add `EthSendRawTransactionUntrusted` RPC method to be used for the gateway when accepting `EthSendRawTransaction` and `eth_sendRawTransaction`. Applies a tighter limit on the number of messages in the queue from a single sender and applies additional restrictions on nonce increments. ([filecoin-project/lotus#12431](https://github.com/filecoin-project/lotus/pull/12431))
```
aarshkshah1992, line 7: You probably need to rebase on master again for all of these to disappear.

chain/types/ethtypes/errors.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// ErrorData returns the error data.
+func (e *ExecutionRevertedError) ErrorData() interface{} {
+	return e.Data
+}
+
+// ErrorCode returns the JSON error code for a revert.
+func (e *ExecutionRevertedError) ErrorCode() int {
+	return 3
```
aarshkshah1992, line 34: Why is this hard coded to 3 ?

aarshkshah1992: Don't we have a `Code` field on `ExecutionRevertedError` ? Shouldn't we use that ?

chain/types/ethtypes/errors.go
```diff
@@ -0,0 +1,44 @@
...
+	"golang.org/x/xerrors"
+)
+
+var ErrExecutionReverted = xerrors.New("execution reverted")
+
+// ExecutionRevertedError is an error that occurs when a transaction reverts.
+type ExecutionRevertedError struct {
+	Message string
+	Code    int
+	Meta    []byte
```
aarshkshah1992, line 15: Don't see this `Meta` field being used anywhere. It's neither written to or read from. Should be accepting it in the constructor and setting it ?

go.mod
```diff
@@ -12,6 +12,8 @@ replace github.com/filecoin-project/test-vectors => ./extern/test-vectors // pro
 
 replace github.com/filecoin-project/filecoin-ffi => ./extern/filecoin-ffi // provided via a git submodule
 
+replace github.com/filecoin-project/go-jsonrpc => github.com/virajbhartiya/go-jsonrpc v0.0.0-20241011111701-53eab64ec154
```
aarshkshah1992, line 15: We will need to change this once the `go-jsonrpc` PR lands and we have a release for it.

node/impl/full/eth_utils.go
```diff
@@ -31,6 +31,10 @@ import (
 	"github.com/filecoin-project/lotus/chain/vm"
 )
 
+const (
+	errcodeDefault = -32000
```
aarshkshah1992, line 35: What's this for ? Why did we have to add it ?

itests/fevm_test.go
```diff
@@ -7,14 +7,17 @@ import (
 	"encoding/binary"
 	"encoding/hex"
```
aarshkshah1992, line 8: Why do we have such a large diff in this file ? Can we get rid of all the whitespace changes and only change the test code we need to ? Reviewing diffs like this is tricky.

node/impl/full/eth.go
```diff
@@ -1433,9 +1433,12 @@ func (a *EthModule) applyMessage(ctx context.Context, msg *types.Message, tsk ty
 	}
 
 	if res.MsgRct.ExitCode.IsError() {
-		reason := parseEthRevert(res.MsgRct.Return)
-		return nil, xerrors.Errorf("message execution failed: exit %s, revert reason: %s, vm error: %s", res.MsgRct.ExitCode, reason, res.Error)
+		return nil, ethtypes.NewExecutionRevertedWithDataError(
+			errcodeDefault,
```
aarshkshah1992, line 1437: Shouldn't we pass the exit code on the receipt here ? Why are we passing `errcodeDefault` here ?

node/impl/full/eth.go
```diff
@@ -1612,6 +1623,13 @@ func ethGasSearch(
 }
 
 func (a *EthModule) EthCall(ctx context.Context, tx ethtypes.EthCall, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthBytes, error) {
+	var result ethtypes.EthBytes
+	defer func() {
+		if r := recover(); r != nil {
```
aarshkshah1992, line 1628: What is this for ? Why do we need it ?

node/impl/full/eth.go
```diff
@@ -1628,13 +1646,15 @@ func (a *EthModule) EthCall(ctx context.Context, tx ethtypes.EthCall, blkParam e
 	}
 
 	if msg.To == builtintypes.EthereumAddressManagerActorAddr {
-		// As far as I can tell, the Eth API always returns empty on contract deployment
 		return ethtypes.EthBytes{}, nil
 	} else if len(invokeResult.MsgRct.Return) > 0 {
-		return cbg.ReadByteArray(bytes.NewReader(invokeResult.MsgRct.Return), uint64(len(invokeResult.MsgRct.Return)))
+		result, err = cbg.ReadByteArray(bytes.NewReader(invokeResult.MsgRct.Return), uint64(len(invokeResult.MsgRct.Return)))
```
aarshkshah1992, line 1651: What is this change doing ? I don't see us returning an `ExecutionRevertedError` anywhere.

node/impl/full/eth.go
```diff
@@ -1612,6 +1623,13 @@ func ethGasSearch(
 }
 
 func (a *EthModule) EthCall(ctx context.Context, tx ethtypes.EthCall, blkParam ethtypes.EthBlockNumberOrHash) (ethtypes.EthBytes, error) {
```
aarshkshah1992, line 1625: Where are we returning an `ExecutionRevertedError` in this function ?

aarshkshah1992:
@virajbhartiya @akaladarshi 

Reviewed. Thanks a lot. Also need to rebase this on master and fix the conflicts.

node/impl/full/eth.go
akaladarshi: @aarshkshah1992 
We are returning the `ExecutionRevertedError` from `applyMessage` function directly, so we don't to do it separately.

In `EstimateGas` we are checking separately because of `GasEstimateMessageGas` function also returns an error.

https://github.com/virajbhartiya/lotus/blob/b1c146a3a2f2f9ca0a1c56a7f0ef7bf438a7db59/node/impl/full/eth.go#L1435

node/impl/full/eth.go
akaladarshi: It was added just to add more context in the error:
 `return nil, xerrors.Errorf("failed to read msg recipient: %w", err)`
 
`ExecutionRevertedError` will be returned by the `applyMessage` directly as it was already returning the `execution reverted` error.

itests/fevm_test.go
akaladarshi: It's a linter fix, everywhere there is whitespace after `//`

node/impl/full/eth.go
akaladarshi: @aarshkshah1992 actually there was some confusion about what error code to return because, in ethereum there two different kind of error codes for the same message (execution reverted), In `EthCall` and `EstimateGas` it returns error code `3` and generally while executing message it returns `-3200`.

I will the default error code to `3` for  `EthCall` and `EstimateGas` which we were already returning in `ErrorCode()`

node/impl/full/eth_utils.go
akaladarshi: Will remove it in favor of error code `3` for `EthCall` and `EstimateGas`, but in other places for execution reverted we need to return this(-32000) code.

chain/types/ethtypes/errors.go
akaladarshi: We can remove this as we have Data field now.

chain/types/ethtypes/errors.go
akaladarshi: Yeah this (3) should be default error code for `EthCall` and `EstimateGas`. Will update it.

node/impl/full/eth_utils.go
aarshkshah1992: @akaladarshi But then why was it not a pre-existing code ? How did we handle this before you raised this PR ?

akaladarshi: I have to look into it little bit what exact error code FVM is returning, that's why I asked it in slack (do we have same error code as ethereum): https://filecoinproject.slack.com/archives/CP50PPW2X/p1728633105115299,

Will check this and update it accordingly.

chain/types/ethtypes/errors.go
akaladarshi: Yeah that make sense, So let me confirm once what kind of error codes FVM returns and I will update it accordingly.

node/impl/full/eth_utils.go
aarshkshah1992: Irrespective of what it is returning, let's not change the error codes here. We only want the data field to be in like with ETH here -> everything else should stay the same right ?

aarshkshah1992:
Let's address the current batch of comments for which we have solid answers too @akaladarshi.

For stuff that is still unclear/ambiguous -> we can then request a review from Stebalien 👍

node/impl/full/eth.go
akaladarshi: Removed.

akaladarshi:
@aarshkshah1992 So I have removed all the hardcoded error codes, and I am directly using the FVM returned error code.

Now the only thing we need to answer is whether we should register the `ExecutionRevertedError` with an error code like [this](https://github.com/virajbhartiya/lotus/blob/22d418a7132eb99dbcb4527c47f90b12edd4b9b0/api/api_errors.go#L53).

 - If we do then we can remove the `ErrorWithCode` interface from `go-jsonrpc`.
 - If not then we need to keep it in order to get the code for the `ExecutionRevertedError` error

aarshkshah1992:
@Stebalien Please can you take a look at just the one comment above i.e. https://github.com/filecoin-project/lotus/pull/12553#issuecomment-2414106748 ?

aarshkshah1992:
@akaladarshi Please can you fix the CI ? Looks like it's all red.

virajbhartiya:
@aarshkshah1992 I was fixing a test, I'll check why the CI is failing

Stebalien:
> @Stebalien Please can you take a look at just the one comment above i.e. https://github.com/filecoin-project/lotus/pull/12553#issuecomment-2414106748?

Personally, I'd:

1. Register the error using the existing error registration logic in _this_ PR.
2. If you're up for it, make a new patch that switches to an interface-based version with `ErrorWithCode`. IMO, that's a much better design but it's a larger refactor. I'd rather not use _both_ mechanisms at the same time.

aarshkshah1992:
@Stebalien 

```
Register the error using the existing error registration logic in this PR.
```

Which PR ? Your comment is missing a link.

itests/fevm_test.go
rvagg: if this is from your editor doing this automatically then it's typically good practice to disable that--it's fine to make lint fixes but large diffs like that should be done as a separate PR and not come in with a PR where the changes are unrelated, or at least make them minimal (e.g. it's OK to fix up some of these around the site where you're changing things, but usually not the whole file if you can help it.)

chain/types/ethtypes/errors.go
```diff
@@ -0,0 +1,43 @@
...
+	return e.Data
+}
+
+// ErrorCode returns the JSON error code for a revert.
+func (e *ExecutionRevertedError) ErrorCode() int {
+	return e.Code
+}
+
+// NewExecutionRevertedWithDataError returns an ExecutionRevertedError with the given code and data.
+func NewExecutionRevertedWithDataError(code int, data string) *ExecutionRevertedError {
```
rvagg, line 37: constructor name should match struct name with a `New` in front

chain/types/ethtypes/errors.go
```diff
@@ -0,0 +1,43 @@
+package ethtypes
+
+import (
+	"fmt"
+
+	"golang.org/x/xerrors"
+)
+
+var ErrExecutionReverted = xerrors.New("execution reverted")
```
rvagg, line 9: You've made 2 error objects in this file (this one, and the `struct` that conforms to the interface) and the only time you use this one is for its string in this file and once in the itest. I'd just either use this string, `"execution reverted"` inline in the constructor, or make it a private `const` here. It's OK to expect `ErrorContains` `"execution reverted"` in the test without linking it to an existing const, in fact that's desirable in a test so make sure what you think is a const is actually what you want it to be.

chain/types/ethtypes/errors.go
```diff
@@ -0,0 +1,43 @@
...
+	Message string
+	Code    int
+	Data    string
+}
+
+// Error implements the error interface.
+func (e *ExecutionRevertedError) Error() string {
+	if e.Message == "" {
+		return fmt.Sprintf("json-rpc error %d", e.Code)
+	}
```
rvagg, line 22: do we expect to ever hit this?

maybe we should do away with `Message` entirely and just always return this, make it `("execution reverted: %d", e.Code)`

rvagg: and if we don't hit this code, I wonder what the point of `Code` is, do we use that anywhere?

itests/fevm_test.go
akaladarshi: So generally I try to do `go fmt` in files after changing things that's why sometimes I get large diff, anyway, I will disable it and revert these whitespace changes so we can keep it clean.

chain/types/ethtypes/errors.go
akaladarshi: Yeah so If you look at the `ErrorCode()` function it returns the code, which is basically for not registered errors.

If you look at this comment above I think you will get the idea: https://github.com/filecoin-project/lotus/pull/12553#issuecomment-2414106748
Whatever we decide based on it, we will either remove the code or keep it.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Reduce size of embedded genesis CAR files by removing WASM actor blocks and compressing with zstd. This reduces the `lotus` binary size by approximately 10 MiB. ([filecoin-project/lotus#12439](https://github.com/filecoin-project/lotus/pull/12439))
 - Add ChainSafe operated Calibration archival node to the bootstrap list ([filecoin-project/lotus#12517](https://github.com/filecoin-project/lotus/pull/12517))
 - Fix hotloop in F3 pariticpation API ([filecoin-project/lotus#12575](https://github.com/filecoin-project/lotus/pull/12575))
+- Return data with `eth_call` and `eth_estimateGas` APIs `execution reverted` error ([filecoin-project/lotus#12553](https://github.com/filecoin-project/lotus/pull/12553))
```
rvagg, line 10:
```suggestion
- Return a `Data` field from RPC when `eth_call` and `eth_estimateGas` APIs encounter `execution reverted` errors ([filecoin-project/lotus#12553](https://github.com/filecoin-project/lotus/pull/12553))
```

api/api_errors.go
```diff
@@ -110,3 +115,22 @@ func (errF3ParticipationTicketStartBeforeExisting) Error() string {
...
+
+// Error returns the error message.
+func (e *ErrExecutionRevertedWithData) Error() string { return e.message }
+
+// ErrorData returns the error data.
+func (e *ErrExecutionRevertedWithData) ErrorData() interface{} { return e.data }
+
+// NewErrExecutionRevertedWithData creates a new ErrExecutionRevertedWithData.
+func NewErrExecutionRevertedWithData(data string) *ErrExecutionRevertedWithData {
+	return &ErrExecutionRevertedWithData{
```
rvagg, line 132:
```suggestion
func (e ErrExecutionRevertedWithData) Error() string { return e.message }

// ErrorData returns the error data.
func (e ErrExecutionRevertedWithData) ErrorData() interface{} { return e.data }

// NewErrExecutionRevertedWithData creates a new ErrExecutionRevertedWithData.
func NewErrExecutionRevertedWithData(data string) ErrExecutionRevertedWithData {
	return ErrExecutionRevertedWithData{
```

Does jsonrpc bork still on non-interface? Not a big deal but if we can avoid pointer receivers where they are not necessary.

akaladarshi: @rvagg They are necessary as of now because we are registering the errors as `RPCErrors.Register(EExecutionRevertedWithData, new(*ErrExecutionRevertedWithData))
}` so `go-jsonrpc` will expect error to be returned as a pointer type for  `c, ok := e.byType[reflect.TypeOf(err)]; `.

I am trying to figure out a way, so we can get the code without worrying if the error is pointer type or not.

rvagg: no, don't bother, pointer is OK if you're forced to downstream

api/api_errors.go
```diff
@@ -110,3 +118,44 @@ func (errF3ParticipationTicketStartBeforeExisting) Error() string {
...
+	Data    string
+}
+
+// Error returns the error message.
+func (e *ErrExecutionRevertedWithData) Error() string { return e.Message }
+
+// FromJSONRPCError converts a JSONRPCError to ErrExecutionRevertedWithData.
+func (e *ErrExecutionRevertedWithData) FromJSONRPCError(jerr jsonrpc.JSONRPCError) error {
+	if jerr.Code != EExecutionRevertedWithData {
+		return nil
```
rvagg, line 133: let's error in this case - because what will happen here is that the user will get a `ErrExecutionRevertedWithData` error that's completely empty - and this _is_ an error because something got wired up wrong

api/api_errors.go
```diff
@@ -110,3 +118,44 @@ func (errF3ParticipationTicketStartBeforeExisting) Error() string {
...
+	}
+
+	data, ok := jerr.Data.(string)
```
rvagg, line 136: I thought we'd need to unmarshall this from a json binary string? Are we getting a quoted string out of this?

api/api_errors.go
rvagg: as per my comment in https://github.com/filecoin-project/go-jsonrpc/pull/123 - we should actually handle an error return - reverting to the default message type but complaining about it in logs seems appropriate

CHANGELOG.md
akaladarshi: Done

api/api_errors.go
akaladarshi: Done.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - `lotus-shed indexes inspect-indexes` now performs a comprehensive comparison of the event index data for each message by comparing the AMT root CID from the message receipt with the root of a reconstructed AMT. Previously `inspect-indexes` simply compared event counts, comparing AMT roots confirms all the event data is byte-perfect. ([filecoin-project/lotus#12570](https://github.com/filecoin-project/lotus/pull/12570))
 - Expose APIs to list the miner IDs that are currently participating in F3 via node. ([filecoin-project/lotus#12608](https://github.com/filecoin-project/lotus/pull/12608))
 - Implement new `lotus f3` CLI commands to list F3 participants, dump manifest and check the F3 status. ([filecoin-project/lotus#12617](https://github.com/filecoin-project/lotus/pull/12617))
+- Return a `Data` field from RPC when `eth_call` and `eth_estimateGas` APIs encounter `execution reverted` errors([filecoin-project/lotus#12553](https://github.com/filecoin-project/lotus/pull/12553))
```
rvagg, line 14:
```suggestion
- Return a `"data"` field on the `"error"` returned from RPC when `eth_call` and `eth_estimateGas` APIs encounter `execution reverted` errors. ([filecoin-project/lotus#12553](https://github.com/filecoin-project/lotus/pull/12553))
```

sorry, my bad, I know I've made a suggestion here before.

api/api_errors.go
```diff
@@ -4,9 +4,27 @@ import (
...
 
+	"golang.org/x/xerrors"
+
 	"github.com/filecoin-project/go-jsonrpc"
 )
 
+var invalidExecutionRevertedWithDataMsg = xerrors.New("invalid execution reverted with data error")
+
+// Default messages for errors.
+const (
```
rvagg, line 15: I don't think these are going to be appreciated, there's already 3 places that some of the errors in this file appear or have constants, it's probably overkill to pull these strings up to constants.

api/api_errors.go
```diff
@@ -74,39 +96,80 @@ func ErrorIsIn(err error, errorTypes []error) bool {
...
-	return "ticket starts before existing lease"
+	return ticketStartBeforeExistingDefaultMsg
 }
 
 type errF3NotReady struct{}
 
-func (errF3NotReady) Error() string { return "f3 isn't yet ready to participate" }
+func (errF3NotReady) Error() string { return f3NotReadyDefaultMsg }
+
+type ErrExecutionRevertedWithData struct {
```
rvagg, line 136: Now the dust has settled, what do you think about this just being `ErrExecutionReverted`? We don't have one already and we can just set the expectation that you always instantiate these with `data`.

rvagg (changes requested):
pretty close, some minor, mostly cosmetic suggestions

Also, I'm down to 2 failures in my local branch of https://github.com/filecoin-project/fevm-contract-tests and I think one of them is about a missing `data` field for a reverted contract, so I believe this will solve that.

api/api_errors.go
akaladarshi: Yeah, This make sense, `withData` was added because we explicitly were returning `ErrorData()`, but now that's not the case. 

I will rename it.

go.mod
```diff
@@ -12,6 +12,8 @@ replace github.com/filecoin-project/test-vectors => ./extern/test-vectors // pro
 
 replace github.com/filecoin-project/filecoin-ffi => ./extern/filecoin-ffi // provided via a git submodule
 
+replace github.com/filecoin-project/go-jsonrpc => github.com/virajbhartiya/go-jsonrpc v0.0.0-20241023032238-90c8605569be
```
rvagg, line 15: you have a v0.7.0 now: https://github.com/filecoin-project/go-jsonrpc/releases/tag/v0.7.0

rvagg (changes requested):
just the go-jsonrpc dependency bump and we can merge this

akaladarshi:
@virajbhartiya We need to update the version of `go-jsronrpc` to the latest release.

rvagg (approved):

rvagg:
:ok_hand: very nice work @akaladarshi & @virajbhartiya!

rvagg:
Flagging that I'm having some problems with this testing fevm-contract-tests, I'm not sure we're doing a better job of conforming to expected go-ethereum output with this.

Most interestingly, there's a relevant chai assertion from Waffle that's now failing a lot with the Uniswap tests, `revertedWith(string)`: https://github.com/TrueFiEng/Waffle/blob/238c11ccf9bcaf4b83c73eca16d25243c53f2210/waffle-chai/src/matchers/revertedWith.ts#L6

So I'm getting failures like this:

```
     AssertionError: Expected transaction to be reverted with Unable to reenter, but other exception was thrown: ProviderError: execution reverted
```

Coming out of an assertion like this:

```js
await expect(reentrant.swapToReenter(pool.address)).to.be.revertedWith('Unable to reenter')
```

Which I believe is unmodified from the upstream uniswap tests.

I think that it's worked previously because we included the revert reason in the string and it matches the hardhat way of doing reverts (see `decodeHardhatError`, which is pretty liberal in its matching). But when you have a `data` field I think you're dealing with this: https://github.com/TrueFiEng/Waffle/blob/238c11ccf9bcaf4b83c73eca16d25243c53f2210/waffle-provider/src/revertString.ts#L28 which performs a full decode of a hex string: https://github.com/TrueFiEng/Waffle/blob/238c11ccf9bcaf4b83c73eca16d25243c53f2210/waffle-provider/src/revertString.ts#L37-L45

Maybe we should be returning the hex string rather than decoding it ourselves in https://github.com/filecoin-project/lotus/blob/2387b1895d44c42cf01bab7182778db1c14fa952/node/impl/full/eth_utils.go#L333, or alternatively do both.

rvagg:
Recording some investigation progress as I try and get to the bottom of this.

I modified `NewErrExecutionReverted` locally to try and restore some of the original info in the error message, decoding the data string "reason" and also putting in the exit code like the original (although this isn't formatted quite the same, I need to try that next).

```go
      err := &ErrExecutionReverted{
              Message: fmt.Sprintf("execution reverted (code=%d, reason=%s, vm error=%s)", exitCode, reason, error),
              Data:    fmt.Sprintf("0x%x", data),
        }
```

But modifying `extern/fevm-uniswap-v3-core/node_modules/@ethereum-waffle/chai/dist/cjs/matchers/revertedWith.js` to log the actual error object we're looking at for a reverted, it comes out with this:

```
{
  "error": {
    "code": 11,
    "message": "execution reverted (code=33, reason=message failed with backtrace:\n00: f01004 (method 3844450837) -- contract reverted (33)\n (RetCode=33), vm error=Error(OLD))",
    "data": "0x586408c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000034f4c44000000000000000000000000000
0000000000000000000000000000000"
  },
  "id": 46,
  "jsonrpc": "2.0"
}
```

Which has our decoded error message in it, but the point of `data` was that it should be able to find it in there.

I am dealing with an older waffle version though, `@ethereum-waffle/chai` version `3.4.4` for fevm-uniswap-tests, but I think that tracks with the upstream version where this would be passing so I'm not sure what the problem is yet.

The code in 3.4.4 that's doing the matching is:

```js
            const reasonsList = error.results && Object.values(error.results).map((o) => o.reason);
            const message = (error instanceof Object && 'message' in error) ? error.message : JSON.stringify(error);
            const isReverted = reasonsList
                ? reasonsList.some((r) => r === revertReason)
                : message.includes('revert') && message.includes(revertReason);
```

We don't have a `results` and `message.includes('revert') && message.includes(revertReason);` should pass for this now, but it's not?

rvagg:
It seems to be all about the `"data"` field, once you add one then it fails to recognise it as a valid revert error. I've tried tinkering with the code, message and data field, and it's only when I remove the data field that I get it to pass. So the next step is to figure out _why_, is it decoding this wrong? Do we need a different format than this?

---

fail:

```json
{
  "error": {
    "code": 1,
    "message": "message execution failed: exit 33, revert reason: message failed with backtrace:\n00: f01004 (method 3844450837) -- contract reverted (33)\n (RetCode=33), vm error: Error(O
LD)",
    "data": "0x586408c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000034f4c440000000000000000000000000000000000000000000000000000000000"
  },
  "id": 46,
  "jsonrpc": "2.0"
}
```

fail:

```json
{
  "error": {
    "code": 11,
    "message": "execution reverted (code=33, reason=message failed with backtrace:\n00: f01004 (method 3844450837) -- contract reverted (33)\n (RetCode=33), vm error=Error(OLD))",
    "data": "0x586408c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000034f4c440000000000000000000000000000000000000000000000000000000000"
  },
  "id": 46,
  "jsonrpc": "2.0"
}
```

pass:

```json
{
  "error": {
    "code": 1,
    "message": "message execution failed: exit 33, revert reason: message failed with backtrace:\n00: f01004 (method 3844450837) -- contract reverted (33)\n (RetCode=33), vm error: Error(OLD)"
  },
  "id": 46,
  "jsonrpc": "2.0"
}
```

pass:

```json
{
  "error": {
    "code": 11,
    "message": "message execution failed: exit 33, revert reason: message failed with backtrace:\n00: f01004 (method 3844450837) -- contract reverted (33)\n (RetCode=33), vm error: Error(OLD)"
  },
  "id": 46,
  "jsonrpc": "2.0"
}
```

pass:

```json
{
  "error": {
    "code": 11,
    "message": "execution reverted (code=33, reason=message failed with backtrace:\n00: f01004 (method 3844450837) -- contract reverted (33)\n (RetCode=33), vm error=Error(OLD))"
  },
  "id": 46,
  "jsonrpc": "2.0"
}
```

rvagg:
Got it, `0x586408c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000034f4c440000000000000000000000000000000000000000000000000000000000` is cbor, it packs a byte slice which is `0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000034f4c440000000000000000000000000000000000000000000000000000000000'` when presented raw. When we return that then the eth libraries can read it.

Basically, we use cbor to pass these things around, but cbor is foreign to eth-land so we need to unpack.

rvagg:
addressed in https://github.com/filecoin-project/lotus/pull/12675

PR #12560, opened 2024-10-06
fix(doc): prefer worded month in `LOTUS_RELEASE_FLOW`

DemoYeti:


rjan90:
Thanks for the PR, but I will reject this given that this is just a stylistic change, and does not contain functional or technical improvements. We usually follow the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format.

Issue #12570, opened 2024-10-10
ChainIndexer: make epoch validation more comprehensive


rvagg:
Currently we are just comparing counts of things, assuming that if we have the right number of messages, events and event entries, then the epoch is properly indexed. This isn't necessarily true and it would be good to validate in more detail. Events are particularly hard since there is a lot of data to be fetched and compared. We already have the data from the blockstore though, so that's not an additional expense (we need to load an entire AMT just to get its length so we know what size to compare).

SQLite has some `sha3` functionality. It's not clear to me if these are available programatically or if they are only available on the `sqlite3` cli. If they are available programatically then we could generate hashes of specific columns in the database and compare that digest to a locally generated digest of the same fields. A select over event fields: emitter, event_index, message (cid?), and the associated event_entry fields: indexed, flags, key, codec, and maybe even value to do the whole lot, and then a sha3 of the results. Attempting to reconstruct this locally would be an interesting exercise.

Unfortunately this would tie us to SQLite functionality, so if we made the database pluggable we'd want to abstract this away somehow so that we can do something similar for another database.

Or, alternatively: just do a bulk select and compare everything in one go.

rvagg:
Another proposal that was raised at one point for this was to recalculate the AMT root of each of the messages' events and compare that to what the receipt says. That way we wouldn't even need to load the AMT from the blockstore and it may end up being more efficient. In theory we _should_ have everything we need to do a lossless reconstruction. The only catch I see is that in ChainIndexer, as in the prior event storage code, we skip over any event where we can't look up the address for the actor ID. I don't know if in practice we have any epochs where this has ever happened, but it's a possibility in the code that such an event won't be stored.

PR #12571, opened 2024-10-10
feat(events): compare-amt option for lotus-shed indexes inspect-events

rvagg:
Instead of relying just on entry counts, compare the regenerated AMT root using just what we have in the db with the message receipt event root. This should tell us precisely that we have what we should or not.

Ref: https://github.com/filecoin-project/lotus/issues/12570

--

I mainly just wanted to see if this was practical for the discussion in #12570, and it seems to be. The only problem is address lookup. Some of my tipsets don't resolve f410 addresses. I get back 1,500 tipsets, to epoch 4339494 and get a `actor not found` error so I'm not sure what to do about that. I've turned that into a "problem" rather than an error and left a TODO to figure out if it can be resolved. I assumed that we could get this mapping at the tipset, but I obviously don't understand something about addresses.

With addresses sorted out, this could work for ChainIndexer too. It's more expensive than comparing lengths (0.5s compared to 0.3s to do 20 epochs from head in a quick test), which is a bit of a surprise because to compare lengths you need to load the original AMT from the blockstore, but the expense comes from doing the address lookups via the API, which may be slightly cheaper when run in-daemon rather than via the RPC like lotus-shed does.

### DIFF
--- a/cmd/lotus-shed/indexes.go
+++ b/cmd/lotus-shed/indexes.go
@@ -10,16 +10,21 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ipfs/go-cid"
+	cbor "github.com/ipfs/go-ipld-cbor"
 	"github.com/mitchellh/go-homedir"
 	"github.com/urfave/cli/v2"
+	cbg "github.com/whyrusleeping/cbor-gen"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-address"
+	amt4 "github.com/filecoin-project/go-amt-ipld/v4"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/crypto"
 	"github.com/filecoin-project/go-state-types/exitcode"
 
 	lapi "github.com/filecoin-project/lotus/api"
+	bstore "github.com/filecoin-project/lotus/blockstore"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
 	lcli "github.com/filecoin-project/lotus/cli"
@@ -34,6 +39,10 @@ const (
 	insertEntry      = `INSERT OR IGNORE INTO event_entry(event_id, indexed, flags, key, codec, value) VALUES(?, ?, ?, ?, ?, ?)`
 	upsertEventsSeen = `INSERT INTO events_seen(height, tipset_key_cid, reverted) VALUES(?, ?, false) ON CONFLICT(height, tipset_key_cid) DO UPDATE SET reverted=false`
 	tipsetSeen       = `SELECT height,reverted FROM events_seen WHERE tipset_key_cid=?`
+
+	// these queries are used to extract just the information used to reconstruct the event AMT from the database
+	selectEventIdAndEmitter = `SELECT id, emitter_addr FROM event WHERE tipset_key_cid=? and message_cid=? ORDER BY event_index ASC`
+	selectEventEntries      = `SELECT flags, key, codec, value FROM event_entry WHERE event_id=? ORDER BY _rowid_ ASC`
 )
 
 func withCategory(cat string, cmd *cli.Command) *cli.Command {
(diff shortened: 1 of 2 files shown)


cmd/lotus-shed/indexes.go
```diff
@@ -483,43 +498,186 @@ var inspectEventsCmd = &cli.Command{
...
+							var actorId abi.ActorID
+							if id, ok := addrIdCache[addr]; ok {
+								actorId = id
+							} else {
+								if addr.Protocol() != address.ID {
+									idAddr, err := api.StateLookupID(ctx, addr, ts.Key())
+									if err != nil {
+										// TODO: fix this? we should be able to resolve all addresses
+										return cid.Undef, fmt.Sprintf("failed to resolve address (%s), could not compare amt", addr.String()), nil
+									}
```
rvagg, line 595: @Stebalien could you have a look at this and suggest if we can do anything about this? We're reconstructing the event AMT from what we've stored, and we store ID addresses or DelegatedAddresses if it is one. I seem to be able to use `StateLookupID` at the tipset to resolve most of those but it regularly gets hung up on particular addresses it can't resolve. I need an `ActorID` for each of these to reconstruct the event AMT.

rvagg:
4358 epochs in my splitstore that it can do this comparison on, and only 12 fail with these address lookup errors:

```
✗ Epoch 4339494 (bafy2bzacedi67cc4hnoxtyfwcxjfqpx6vtwylejf6copjljphwrv4lazxojoo): failed to resolve address (f410frpwhuxbvcpa7sk3iz6mnjnsix7je6zs46qc7pja), could not compare amt
✗ Epoch 4339491 (bafy2bzacecpv4bqtleojyj7hi7dusnkuykn6kibkbd56c4ob46l42x66lwopi): failed to resolve address (f410fxh3uvls6o65egzapjlwkr5iv56kk7pu7uf7m4ci), could not compare amt
✗ Epoch 4339483 (bafy2bzaceb7dv7xb34kktigsjewsuzinlyuib5ava6p6zstqnbst3x2fylb5y): failed to resolve address (f410f35pbojlm5onzmzw55oocnwgbfo5md7za7bx4iaa), could not compare amt
✗ Epoch 4339481 (bafy2bzacedveo3p7dazs3yklnkikfve4aou6gdlcycjnx4ysfsrookbngk6wo): failed to resolve address (f410f7px5a2sebnbvatcaadtceafxrnckxaw4mavydbq), could not compare amt
✗ Epoch 4339475 (bafy2bzacec3ffum3odwicjpubvg7j7afg26et6gpzhfixg4iscrzrgmwunxf6): failed to resolve address (f410fwnajqo2ozwf36bunolkn4tuzd6gcjamldsnumwq), could not compare amt
✗ Epoch 4339472 (bafy2bzacebikp6g4qje5rt3ydn54spuokrtcybe7doqfwfsi2p45bv7uwh3ly): failed to resolve address (f410fpzr5ifoqmsqlwgio2gyfzfqzldazh6mv7yxsjwq), could not compare amt
✗ Epoch 4338989 (bafy2bzaceb3v7vnki2bqze6b4qqkg4y2ijv3rlrkhclvzcmnpmrvhgejnnvx2): failed to resolve address (f410fgrholovr2ab54n6nnmerklhkcigmk3mczzgkrza), could not compare amt
✗ Epoch 4338987 (bafy2bzacebvx5txuxekq2inesre7bwqchcd3qdpwjrouisp6p7ceqan2awmqk): failed to resolve address (f410fqlqx24dg6cn5lfmf5twixhvilnwfsig4lf4rx3q), could not compare amt
✗ Epoch 4338985 (bafy2bzacebfh2hdtc5vk3oaywtmrcxyowraeqcdw5v5gqzgesboc6pgd4odr4): failed to resolve address (f410f3f6lkjrxmpszyn62dg6ihtvfgpe4lgxzqqfqz3i), could not compare amt
✗ Epoch 4338984 (bafy2bzaceafev3gi7bpywpti3yc6ahkj7vqaggjnugcex4t3hcgdpb4tjo276): failed to resolve address (f410f6roex63vvdy4hwvvmsmt4msl3ts42hlzompxlhq), could not compare amt
✗ Epoch 4338983 (bafy2bzacec7ahfjagwbrm5lzeskct3mxgmdcoszwwpdumpekz7msolict3qgw): failed to resolve address (f410f7rxno56bkace75glk33iyhi2urlzdyqfspsrb6q), could not compare amt, failed to resolve address (f410fdcu75eaa5wxp6qmvp7qmobwa2i6i76lqpj2tcty), could not compare amt
✗ Epoch 4338951 (bafy2bzacebnxhcqokrfa7n5trn444ivipurvqyahrxnasbzcpvw5ppogn3gam): failed to resolve address (f410fopdzv2y2xb6lai2oi3nt3xlpbxpki2a3g5fcima), could not compare amt, failed to resolve address (f410fbcbcxvugqeblszsaghnbwcpdkiwgfvsqzolyeia), could not compare amt
```

jennijuju:
I think this is also close to  the “validation tool” in my head & I think it provides a more reliable response compare to the count checks (this should be a part of the reconciliation process or recommended for index providers to run on a regular basis)

cmd/lotus-shed/indexes.go
Stebalien: Deferred execution strikes again, never trust a tipset.

 `LookupIDAddress` uses the parent tipset for performance reasons so you're probably trying to resolve IDs for actors that were created in that tipset.

I'd strongly consider modifying `LookupIDAddress` to use the correct state. Really, I'd audit all calls to `sm.parentState` and fix them to call `stmgr.TipSetState`. This used to be a huge perf issue but we'll now try pretty hard to find the correct state-tree root by walking the chain.

Stebalien: Note: that won't help you with _uncle_ blocks. For those, you'll end up re-computing (although you will cache that new state-cid in memory).

A possible optimization is to _try_ looking up the ID in the parent tipset before trying to compute the new tipset. That'll cost more in the worst-case scenario, but may be better in most cases.

rvagg: Ahhh of course, and in this case it's easy enough because (a) we're only dealing with the canonical chain and (b) I have the previous tipset (walking backward) handy so I can just feed that in to `LookupIDAddress` instead.

So making that change gets me all ✔️ for my current chain, no address problems.

rvagg:
Cleaned this up a bit and now it's faster than just doing the counts because I managed to reduce the number of API calls. So, I've just made this the normal behaviour, not bothering with counts. I left the counting code in there so it can provide some informative output in the case of a mismatch but there's not really a good reason to keep doing it that way if we can be accurate.

cmd/lotus-shed/indexes.go
```diff
@@ -572,14 +639,148 @@ var inspectEventsCmd = &cli.Command{
...
+		for rows2.Next() {
+			var flags []byte
+			var key string
+			var codec uint64
+			var value []byte
+			if err := rows2.Scan(&flags, &key, &codec, &value); err != nil {
+				return cid.Undef, false, "", xerrors.Errorf("failed to scan row: %w", err)
+			}
+			entry := types.EventEntry{
+				Flags: flags[0],
```
aarshkshah1992, line 729: Why do we only read the first byte of the flag here ?

cmd/lotus-shed/indexes.go
```diff
@@ -572,14 +639,148 @@ var inspectEventsCmd = &cli.Command{
...
+// and a fatal error if one occurred.
+func amtRootForEvents(
+	ctx context.Context,
+	api lapi.FullNode,
+	tsKeyCid cid.Cid,
+	prevTsKey types.TipSetKey,
+	stmtSelectEventIdAndEmitter, stmtSelectEventEntries *sql.Stmt,
+	message lapi.Message,
+	addrIdCache map[address.Address]abi.ActorID,
+) (cid.Cid, bool, string, error) {
```
aarshkshah1992, line 661: I don't really see us setting the returned string to a meaningful value here. Perhaps we can remove it in favour of only keeping the error ?

cmd/lotus-shed/indexes.go
```diff
@@ -483,87 +492,145 @@ var inspectEventsCmd = &cli.Command{
...
+					stmtSelectEventIdAndEmitter,
+					stmtSelectEventEntries,
+					messages[msgIndex],
+					addrIdCache,
+				)
+				if err != nil {
+					return err
+				}
+				if has && !hasEvents {
+					hasEvents = true
```
aarshkshah1992, line 567: Why can't this just be `hasEvents=has` without any if guards ?

cmd/lotus-shed/indexes.go
```diff
@@ -483,87 +492,145 @@ var inspectEventsCmd = &cli.Command{
...
+				}
+
+				if problem != "" {
+					problems = append(problems, problem)
+				} else if amtRoot != *receipt.EventsRoot {
+					problems = append(problems, fmt.Sprintf("events root mismatch for event %d", eventIndex))
+					// also provide more information about the mismatch
+					if err := checkEventAndEntryCounts(); err != nil {
```
aarshkshah1992, line 575: The `checkEventAndEntryCounts` only matches on counts. I think we should also log a "counts are same but roots dont match" or something similar to point out this specific problem to the user.

aarshkshah1992 (commented):
nice

cmd/lotus-shed/indexes.go
```diff
@@ -483,87 +492,145 @@ var inspectEventsCmd = &cli.Command{
...
-					continue
+			checkEventAndEntryCounts := func() error {
+				// compare by counting events, using ChainGetEvents to load the events from the chain
+				expectEvents, expectEntries, err := chainEventAndEntryCountsAt(ctx, currTs, receipts, api)
+				if err != nil {
+					return err
 				}
-				events, err := api.ChainGetEvents(ctx, *receipt.EventsRoot)
+				if expectEvents > 0 {
+					hasEvents = true
```
aarshkshah1992, line 520: `hasEvents` is set in multiple places but it's only real use is in logging here:

```
if err := stmtTipsetSeen.QueryRow(tsKeyCid.Bytes()).Scan(&seenHeight, &seenReverted); err != nil {
				if err == sql.ErrNoRows {
					if hasEvents {
						problems = append(problems, "not in events_seen table")
					} else {
						problems = append(problems, "zero-event epoch not in events_seen table")
					}
```

Can we simplify how `hasEvents` is set -> set it only once and preferably during initialisation.

cmd/lotus-shed/indexes.go
rvagg: It's just a `uint8`, so we only need the first byte: https://github.com/filecoin-project/lotus/blob/d4b8af99b201053a7ae40397052fe75522c867ac/chain/types/event.go#L23

I'm not sure why we store it as a `BLOB` though, maybe it was imagined it would be extended at some point? I doubt it.

cmd/lotus-shed/indexes.go
rvagg: cause it's run for each message in the tipset, we just want to set this if it encounters at least one event in one of the messages

cmd/lotus-shed/indexes.go
rvagg: Yeah, that's what will happen here - the reason I call `checkEventAndEntryCounts` is that the error for that tipset will be something like "✗ [1234567] (bafyfoo): events root mismatch for event 5, expected 10 events but got 14, expected 100 event entries but got 199". If the counts match then it'd just be "✗ [1234567] (bafyfoo): events root mismatch for event 5".

But looking at this error I realise it's a little bit wrong, it's not "event X" it's "message X" and "X" isn't really useful, so I'm replacing it with:

```go
problems = append(problems, fmt.Sprintf("events root mismatch for message %s", messages[msgIndex].Cid))
```

cmd/lotus-shed/indexes.go
rvagg: I've removed one of the places where it's set, so it's only set in one place now (that was a remnant from a previous refactor that allowed for this code to go either counting or amt checking, but now it's just amt checking with counting if we need it). But we still need it to be set the first time we encounter a message with >0 events and we don't know that ahead of time. The only other way I can think to do that would be to check all of the `EventsRoot` values for each receipt and see if they are either `nil` or point to the 0-length AMT (I'm actually not sure of the fvm behaviour here - does it set that to `nil` when there are no events or set it to the CID of the 0-length AMT?). But there's even a loop in there, so I may as well do the `has` return on the AMT builder and use that.

aarshkshah1992: I see. Yeah makes sense.

aarshkshah1992 (approved):

PR #12575, opened 2024-10-10
fix(f3): fix hot loop in F3 participation

Kubuxu:
Due to a bug in the `awaitLeaseExpiry`, we would retry immediately, which then would cause a hot loop.
Fix the bug and add a bit of hot-loop protection.


masih (approved):

Kubuxu (commented):
LGTM

PR #12581, opened 2024-10-10
feat(f3): log when gaining/losing participation leases

Stebalien:
## Proposed Changes

Log whenever we start/stop participating for a storage miner in F3. Given that we use overlapping leases, we expect that we'll start participating for a miner on start then continue to participate over time.

This could get a bit noisy for setups that constantly switch between different lotus nodes but there's not much I can do about that and this will really help with debugging.

## Checklist

- [ ] CI is green


rvagg (approved):
ok, but it'd be nice to get `t` in there for testnet prefix; we need a `MustNewAddress(id)` for this kind of thing.

Stebalien:
> ok, but it'd be nice to get t in there for testnet prefix; we need a MustNewAddress(id) for this kind of thing.

I just removed the f0 prefix. I agree we should have a better way to do this, but I don't want to have to deal with creating/formatting an address _just_ for logging. And an ID is sufficient here.

Issue #12584, opened 2024-10-11
eth_getLogs should not show removed events in some cases


rvagg:
TheGraph seems to have an expectation that eth_getLogs doesn't return reverted events, at least past a particular point ([it recommends running against a finalised chain for this reason I believe](https://github.com/graphprotocol/graph-node/blob/fbb45890891368f418f26355f579e7a7c72db57e/chain/ethereum/src/ethereum_adapter.rs#L832-L838)). So, it's running into trouble with a response like this for an `eth_getLogs` with `{ "fromBlock":"0x3E2831", "toBlock":"0x3E2831" }` on a node that experienced a reorg at epoch 4073521 which resulted in reverted logs:

<details>
<summary>JSON log output</summary>

```json
{
    "id": 74,
    "jsonrpc": "2.0",
    "result": [
        {
            "address": "0xf7c27a7c5b33933880ce8f4054c4d340020181b2",
            "data": "0x000000000000000000000000000000000000000000000000000000000000002a000000000000000000000000000000000000000000000000000000000000000046494c00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000066f5f5800000000000000000000000000000000000000000000000007068fb1598aa0000000000000000000000000000000000000000000000000000000000000000251c",
            "topics": [
                "0x8a73e656799f8fbd4ea08e9420225176f98f85ef9a2a11a5a011580b65c8f145",
                "0x00000000000000000000000010a15f515e68ce7bc5884cc6a7254846d1a910fb"
            ],
            "removed": true,
            "logIndex": "0x1e",
            "transactionIndex": "0x40",
            "transactionHash": "0x8e188f03e039a9f5a245cc1175434da357792eed0614776816a9b15e2e1e8774",
            "blockHash": "0x9bdcfa3dcf978b56901e94e2b6ac6e220d27f957e7113d614e522a47e78496cb",
            "blockNumber": "0x3e2831"
        },
        {
            "address": "0xf7c27a7c5b33933880ce8f4054c4d340020181b2",
            "data": "0x000000000000000000000000000000000000000000000000000000000000002a000000000000000000000000000000000000000000000000000000000000000046494c00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000066f5f5800000000000000000000000000000000000000000000000007068fb1598aa0000000000000000000000000000000000000000000000000000000000000000251c",
            "topics": [
[...]

PR #12585, opened 2024-10-11
feat: improve semantics for returning reverted events to event API users

aarshkshah1992:
For https://github.com/filecoin-project/lotus/issues/12584.

-Event APIs should only returns reverted events if client has queries by block hash rather than epoch.
- The Subscribe API should always return reverted events for an accurate observation of real time changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ## Bug Fixes
 - Fix a bug in the `lotus-shed indexes backfill-events` command that may result in either duplicate events being backfilled where there are existing events (such an operation *should* be idempotent) or events erroneously having duplicate `logIndex` values when queried via ETH APIs. ([filecoin-project/lotus#12567](https://github.com/filecoin-project/lotus/pull/12567))
+- Event APIs (Eth events and actor events) should only return reverted events if client queries by specific block hash / tipset. Eth and actor event subscription APIs should always return reverted events to enable accurate observation of real-time changes. ([filecoin-project/lotus#12585](https://github.com/filecoin-project/lotus/pull/12585))
+
+## Improvements
 
 ## Improvements
 
--- a/node/impl/full/eth.go
+++ b/node/impl/full/eth.go
@@ -1727,15 +1727,12 @@ func (e *EthEventHandler) ethGetEventsForFilter(ctx context.Context, filterSpec 
 		}
 	}
 
-	// Create a temporary filter
-	excludeReverted := pf.tipsetCid == cid.Undef
-	f, err := e.EventFilterManager.Install(ctx, pf.minHeight, pf.maxHeight, pf.tipsetCid, pf.addresses, pf.keys, excludeReverted)
+	// Fill a filter and collect events
+	f, err := e.EventFilterManager.Fill(ctx, pf.minHeight, pf.maxHeight, pf.tipsetCid, pf.addresses, pf.keys)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to install event filter: %w", err)
 	}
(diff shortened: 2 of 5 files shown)


node/impl/full/eth.go
```diff
@@ -1957,7 +1958,8 @@ func (e *EthEventHandler) EthNewFilter(ctx context.Context, filterSpec *ethtypes
 		return ethtypes.EthFilterID{}, err
 	}
 
-	f, err := e.EventFilterManager.Install(ctx, pf.minHeight, pf.maxHeight, pf.tipsetCid, pf.addresses, pf.keys, true)
```
rvagg, line 1960: change in functionality here - you couldn't use `eth_newFilter` to install a filter that got you a reverted tipset but now you can; but I think this is reasonable and maybe we consider this a bug being fixed

node/impl/full/eth.go
```diff
@@ -2123,7 +2125,7 @@ func (e *EthEventHandler) EthSubscribe(ctx context.Context, p jsonrpc.RawParams)
 			}
 		}
 
-		f, err := e.EventFilterManager.Install(ctx, -1, -1, cid.Undef, addresses, keysToKeysWithCodec(keys), true)
+		f, err := e.EventFilterManager.Install(ctx, -1, -1, cid.Undef, addresses, keysToKeysWithCodec(keys), false)
```
rvagg, line 2128: I guess this is a bugfix too? https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub#logs is the only explicit mention of this I can find when searching but it does clearly say that you should expect reverts on an `eth_subscribe`.

aarshkshah1992: Yes the fact that this was `true` surprised me too.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 # UNRELEASED
 
+- https://github.com/filecoin-project/lotus/pull/12585
+  - Event APIs(ETH Events API and Actor Events API) should only return reverted events if client queries by block hash.
+  - Both event Subscibr APIs should always return reverted events to enable accurate observation of real time changes.
```
rvagg, line 7:
```suggestion
- Event APIs (Eth events and actor events) should only return reverted events if client queries by specific block hash / tipset. Eth and actor event subscription APIs should always return reverted events to enable accurate observation of real-time changes. ([filecoin-project/lotus#12585](https://github.com/filecoin-project/lotus/pull/12585))
```

CHANGELOG.md
```diff
@@ -91,7 +95,7 @@ This is a Lotus Node only release, which includes a variety of new features, imp
 
 ## 📝 Changelog
 
-See https://github.com/filecoin-project/lotus/compare/v1.28.2...release/v1.29.0 for the set of changes since the last release.
```
rvagg, line 94: lots of trailing newline fixes in this diff, did you mean to do this? I don't mind too much but it is noisy if you weren't intending to and your editor was just being clever.

node/impl/full/eth.go
aarshkshah1992: Filters also exist to observe real time changes to chain state and so reverts should be a part of it.

CHANGELOG.md
aarshkshah1992: I think this is ZED.

aarshkshah1992: Will look into how to avoid these, dont know why it's doing this.

aarshkshah1992:
Fixing `TestGetActorEventsRaw`.

rvagg (approved):
I messed around with the commits, cleaned up the changelog, and it's good for a rebase merge now as is once tests pass

PR #12586, opened 2024-10-11
fix(events): handle subscribe and get/fill cases separately

rvagg:
After reviewing https://github.com/filecoin-project/lotus/pull/12585 I realised the remaining problem was this:

* subscribe-type calls always wants future reverts but only historical reverts when a tipset is specified
* get-type calls only want reverts when a tipset is specified

But we only have one "excludeReverted" argument to `Install()`. My first go at this was changing that argument to `subscribe` and using that to differentiate the behaviour and doing the `tipsetKeyCid == cid.Undef` check just for `prefillFilter`. But then I realised that for the non-subscribe calls we only ever install+remove the filter immediately, so the logic is kind of unnecessary if we just introduce a second method to deal with that case.

So now we have a `Fill()` in addition to an `Install()`, with the former not installing the filter, it just fills it if there's anything to fill and applies the `tipsetKeyCid == cid.Undef` directly itself. Then `Install()` can use `Fill()` for its historic stuff and then install the filter and ignore the reversion exclusion entirely because you always want those reverts in that case.

* `eth_subscribe` can't get historic events, so it doesn't touch `prefillFilter`
* `SubscribeActorEventsRaw()` can get historic events, but you can also specify a tipset (!) which is kind of silly, but perhaps you want to subscribe to the next tipset, which you have, but there aren't events for it yet .. that should work now, including reversions of it.
* `eth_newFilter` is the interesting case because it's a hybrid API that can be used a bit like `eth_subscribe` but for polling, but it can also be used for just getting historic events like `eth_getLogs` so it needs this dual functionality supplied by `Install()`.

### DIFF
--- a/chain/events/filter/event.go
+++ b/chain/events/filter/event.go
@@ -46,11 +46,10 @@ type eventFilter struct {
 	keysWithCodec map[string][]types.ActorEventBlock // map of key names to a list of alternate values that may match
 	maxResults    int                                // maximum number of results to collect, 0 is unlimited
 
-	mu              sync.Mutex
-	collected       []*CollectedEvent
-	lastTaken       time.Time
-	ch              chan<- interface{}
-	excludeReverted bool
+	mu        sync.Mutex
+	collected []*CollectedEvent
+	lastTaken time.Time
+	ch        chan<- interface{}
 }
 
 var _ Filter = (*eventFilter)(nil)
@@ -84,9 +83,6 @@ func (f *eventFilter) ClearSubChannel() {
 }
 
 func (f *eventFilter) CollectEvents(ctx context.Context, te *TipSetEvents, revert bool, resolver AddressResolver) error {
-	if f.excludeReverted && revert {
-		return nil
-	}
 	if !f.matchTipset(te) {
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


chain/events/filter/event.go
```diff
@@ -84,9 +83,6 @@ func (f *eventFilter) ClearSubChannel() {
 }
 
 func (f *eventFilter) CollectEvents(ctx context.Context, te *TipSetEvents, revert bool, resolver AddressResolver) error {
```
aarshkshah1992, line 85: Good stuff removing this completely from API calls that only need historical events.

aarshkshah1992 (approved):

chain/events/filter/event.go
rvagg: yeah, I realised there's a race in here - you could get an `Apply()` or `Revert()` between installing and removing it .. it'd be very very unlikely, and I don't think it'd be a big harm, but with this new functionality, if that were to happen, you could potentially get reverted events showing up in your query for historical events!

aarshkshah1992:
@rvagg Just need to fix the lint error.

PR #12588, opened 2024-10-11
fix: bug in exclude reverted logic for eth_getLogs

aarshkshah1992:
Patches a bug introduced earlier in https://github.com/filecoin-project/lotus/pull/12586.


rjan90 (approved):

aarshkshah1992:
@rvagg in case you wanna take a look later.

PR #12592, opened 2024-10-11
fix(f3): don't return an error on stop if our context was canceled

Stebalien:
## Proposed Changes

Don't return an error on stop if our context was canceled. Otherwise, the node's `stop` function returns an error and the splitstore and window post dispute tests randomly fail.

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title conforms with [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#pr-title-conventions)
- [x] Update CHANGELOG.md or signal that this change does not need it per [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#changelog-management)
- [x] New features have usage guidelines and / or documentation updates in
  - [x] [Lotus Documentation](https://lotus.filecoin.io)
  - [x] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [x] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12593, opened 2024-10-12
chore: libp2p: LOTUS_P2P_BOOTSTRAPPERS="" is valid setting

ribasushi:
Allows one to run
```
LOTUS_P2P_BOOTSTRAPPERS="" ./lotus daemon
```

and then connect to a chosen bootstrapper manually

### DIFF
--- a/build/bootstrap.go
+++ b/build/bootstrap.go
@@ -21,24 +21,36 @@ const (
 var bootstrapfs embed.FS
 
 func BuiltinBootstrap() ([]peer.AddrInfo, error) {
+
+	// do not connect to anything for itests
 	if DisableBuiltinAssets {
 		return nil, nil
 	}
+
+	var bsList, bsListFin []string
+
+	// envvar always takes precedence
 	if bootstrappers, found := os.LookupEnv(BootstrappersOverrideEnvVarKey); found {
 		log.Infof("Using bootstrap nodes overridden by environment variable %s", BootstrappersOverrideEnvVarKey)
-		return addrutil.ParseAddresses(context.TODO(), strings.Split(strings.TrimSpace(bootstrappers), ","))
-	}
-	if buildconstants.BootstrappersFile != "" {
+		bsList = strings.Split(bootstrappers, ",")
+	} else if buildconstants.BootstrappersFile != "" {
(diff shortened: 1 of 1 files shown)


rvagg:
I'm a bit dubious about the need for this, can you explain a concrete use-case where this makes sense?

Also since #12592 is merged, could you rebase and we can look at CI again, because those failures are not known flakies.

rjan90:
> The use case is testing.

@ribasushi Why can't you use the `lotus daemon --bootstrap` flag, if the plan is to test without bootstrappers?

rjan90:
> If it is that much of a problem - just close the PR, I'll carry the fix locally.

I was hoping to understand the differences between using the `--bootstrap` flag and the environment variable for this purpose. There's no need for frustration; I'm simply looking for clarification.

Stebalien:
This introduces a footgun because bash doesn't error on undefined variables. E.g., the following would disable bootstrapping and should likely just fail with an error:

```
export LOTUS_P2P_BOOTSTRAPPERS="$UNDEFINED_OR_MISSPELLED_VAR"
```

 I'd rather just stick with the current mechanism of `--bootstrap=false`. Not protecting bootstrappers if `--bootstrap=false` would also be reasonable but note: protection isn't the same thing as "peering", we won't automatically re-connect.

Closing for now given that "for testing" isn't really enough of a motivation in this case. We can reconsider if we find this causing an actual problem.

Stebalien:
> I obviously disagree with this outcome and find it beneath this devteam.

If you disagree, please provide clear motivation (more than just "for testing"). Please _don't_ attack the team reviewing your PRs.

Stebalien:
I've re-read this entire interaction multiple times and have no idea what you're talking about:

1. PR was submitted with no motivation provided.
2. Motivation was requested (as is _normal_ for all proposed changes). It's not obvious that "empty" is a valid/reasonable value here.
3. The provided motivation was a vague "for testing".
4. The counter-argument I provided is that this change introduces footgun. You elected to not continue the discussion (on slack) so I closed it.

Where is the pettiness? Why do you say that the team _maintaining_ this software doesn't have the right to ask for motivation for changes?

PR #12597, opened 2024-10-14
fix: determine F3 participants relative to current network name

### LINKED ISSUE #12519: Flaky F3 itests
Happening with some frequency, here's some recent examples:

* TestF3_PauseAndRebootstrap: https://github.com/filecoin-project/lotus/actions/runs/11046427121/job/30685760550?pr=12505
* TestF3_Rebootstrap: https://github.com/filecoin-project/lotus/actions/runs/11063981852/job/30741029746?pr=12518#step:9:25233

masih:
## Related Issues
Fixes #12519 

## Proposed Changes
When manifest changes, depending on the timing it is possible for newly generated valid leases to get removed if the sign message loop attempts to sign messages that are as a result of progressing previous network.

Here is an example scenario in a specific order that was causing itests to fail:
 * participants get a lease for network A up to instance 5
 * network A progresses to instance 6
 * manifest changes the network name to B
 * participants get a new lease for network B up to instance 5
 * sign loop receives a message from network A, instance 6
 * `getParticipantsByInstance` lazily removes leases since it only
   checks the instance.
 * the node ends up with no participants, and stuck.

To fix this:
 1) check if participants asked for are within the current network, and if not refuse to participate.
 2) check network name, as well as instance, to lazily remove expired leases.

To aid debugging failing tests in the future add option to print progress of all nodes at every eventual assertion, disabled by default.

## Additional Info

Look closely at the commits in this PR. The commits introduce a dedicated CI job that repeatedly runs the flaky tests (50 times) to assert that they are indeed fixed. The job is then removed in later commits.

These commits are left here for the benefit of the reviewer as [proof of the pudding](https://github.com/filecoin-project/lotus/actions/runs/11369403828/job/31626763159?pr=12597).

## Checklist

- [x] CI is green


chain/lf3/participation_lease.go
```diff
@@ -112,15 +112,26 @@ func (l *leaser) participate(ticket api.F3ParticipationTicket) (api.F3Participat
...
 }
 
-func (l *leaser) getParticipantsByInstance(instance uint64) []uint64 {
+func (l *leaser) getParticipantsByInstance(network gpbft.NetworkName, instance uint64) []uint64 {
 	l.mutex.Lock()
 	defer l.mutex.Unlock()
+	currentManifest, _ := l.status()
+	currentNetwork := currentManifest.NetworkName
+	if currentNetwork != network {
+		log.Warnf("no participants for network: current network (%s) does not match requested network (%s) at instance %d", currentNetwork, network, instance)
```
Stebalien, line 121: IMO, there's no point in warning here. It just means the signing request is old.

itests/f3_test.go
```diff
@@ -275,14 +301,23 @@ func setupWithStaticManifest(t *testing.T, manif *manifest.Manifest, testBootstr
...
 		return e.ms.Run(ctx)
 	})
 
+	// Update initial manifest params to shorten the timeouts and backoff for
+	// testing, and assert it is consistently applied to all nodes.
+	e.m.Gpbft.Delta = 250 * time.Millisecond
+	e.m.Gpbft.DeltaBackOffExponent = 1.3
+	e.m.Gpbft.RebroadcastBackoffBase = manif.Gpbft.Delta * 2
+	e.m.Gpbft.RebroadcastBackoffMax = manif.Gpbft.RebroadcastBackoffBase * 2
+	e.m.Gpbft.RebroadcastBackoffExponent = manif.Gpbft.DeltaBackOffExponent
```
Stebalien, line 317: What about creating a `newTestManifest` and using that instead of `lf3.NewManifest` so we don't have to override these here.

Stebalien (approved):

itests/f3_test.go
masih: Doing it this way has a subtle side effect: it would be relative to whatever the default initial manifest is set to.

Happy to go the other route, though i think we still need to assert it is applied to all nodes on test environment instantiation.

Stebalien: There shouldn't be a default initial manifest, right?

Stebalien: I.e., the "default initial manifest" is https://github.com/filecoin-project/lotus/blob/f6d4fc7ccb0a53fd49cf01a82ab3a9528345afab/itests/f3_test.go#L212

itests/f3_test.go
```diff
@@ -275,14 +326,16 @@ func setupWithStaticManifest(t *testing.T, manif *manifest.Manifest, testBootstr
...
 	errgrp.Go(func() error {
 		defer func() {
 			require.NoError(t, manifestServerHost.Close())
 		}()
 		return e.ms.Run(ctx)
 	})
 
+	// Assure manifest is picked up by all nodes.
+	e.ms.UpdateManifest(manif)
+	e.waitTillManifestChange(manif, 20*time.Second)
```
Stebalien, line 338: This shouldn't be necessary. All nodes are constructed with `manif` as their initial manifest.

masih: I thought I assert it anyway for sanity. Happy to remove.

Stebalien: I'd kind of prefer to test it without it.

Stebalien (approved):

Kubuxu (approved):

PR #12602, opened 2024-10-15
chore: update triplewz/poseidon

magik6k:
## Related Issues
https://github.com/triplewz/poseidon/pull/2

## Proposed Changes
Curio pulls in the new version of the Poseidon library which switched to generic fr.Element. We need that to support SnapDeal decode in pure Go, and can't use the older version.

## Checklist

- [x] CI is green


go.mod
```diff
@@ -132,7 +133,7 @@ require (
 	github.com/sirupsen/logrus v1.9.2
 	github.com/stretchr/testify v1.9.0
 	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
-	github.com/triplewz/poseidon v0.0.1
+	github.com/triplewz/poseidon v0.0.2-0.20240407130934-5265fab9d889
```
Stebalien, line 136: Any chance we could ask for an upstream release?

storage/sealer/commitment/commr.go
```diff
@@ -25,7 +25,7 @@ func CommR(commC, commRLast [32]byte) ([32]byte, error) {
 	input_b.SetBytes(commRLast[:])
 	input := []*big.Int{input_a, input_b}
 
-	cons, err := poseidon.GenPoseidonConstants(3)
+	cons, err := poseidon.GenPoseidonConstants[*fr.Element](3)
```
Stebalien, line 28: Confirming that there _is_ a golden test for this here: https://github.com/filecoin-project/lotus/blob/f6d4fc7ccb0a53fd49cf01a82ab3a9528345afab/storage/sealer/commitment/commr_test.go#L9-L35.

Stebalien (approved):
I'd prefer a released version, but this seems reasonable. From what I can tell, this is _only_ used in sealing, we have a separate flow to _check_ the proofs. That means this change should be covered by the itests.

But just to double check: @magik6k have you tested the lotus-miner sealing flow and/or confirmed that it's tested (enough) in CI? I now some proof types aren't generally tested because they're too slow to test in CI.

go.mod
magik6k: Not sure how to get to the maintainer, @vmx might know though.

storage/sealer/commitment/commr.go
magik6k: Correct, that test is exactly that

go.mod
magik6k: (I wouldn't block this PR on the upstream tag, this is actively blocking us on updating Curio to latest NV, and Curio is now most of calibnet)

magik6k:
> From what I can tell, this is only used in sealing, we have a separate flow to check the proofs. That means this change should be covered by the itests.

It's only used with "External PC2 binary" mode of sealing, which IIUC isn't adopted pretty much anywhere (that was used for brownfield supraseal pc2, which while faster wasn't very stable). We also use the new version of this library in Curio for many other parts of the proving system, and it definitely works fine there.

rjan90:
> It's only used with "External PC2 binary" mode of sealing, which IIUC isn't adopted pretty much anywhere (that was used for brownfield supraseal pc2, which while faster wasn't very stable).

Can confirm that the "External PC2 binary" mode of sealing is not well adopted, because it has the downside of only working with CC-sectors currently: and switching between external PC2 binary / internal PC modes depending if it is data or not in the sectors, would be a very painful experience.

go.mod
Stebalien: I've filed an issue, no reason to wait.

Stebalien (approved):

PR #12603, opened 2024-10-16
chore(events): improve perf for parallel event filter matching

rvagg:
Blocked until #12421 is merged and we can do this against master.

I started writing up a `good-first-issue` ticket for this but it eneded up being easier to write the code than describe what needed doing.

1. Cache address look-ups for the given tipset across filters
2. Run the filters in an errgroup
3. Removed the `bool` return argument and just use `address.Undef`, I believe there's no case where this is an acceptable value

Filters can be installed via `eth_newFilter`, `eth_subscribe` and `SubscribeActorEventsRaw`. All filters installed have `CollectEvents` run on them for each new tipset and revert. So this can get expensive if you're an RPC provider who allows users to install these things and watch for live events.

### DIFF
--- a/chain/events/filter/event.go
+++ b/chain/events/filter/event.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/ipfs/go-cid"
 	cbg "github.com/whyrusleeping/cbor-gen"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-address"
@@ -28,7 +29,10 @@ func isIndexedValue(b uint8) bool {
 	return b&(types.EventFlagIndexedKey|types.EventFlagIndexedValue) > 0
 }
 
-type AddressResolver func(context.Context, abi.ActorID, *types.TipSet) (address.Address, bool)
+// AddressResolver is a function that resolves an actor ID to an address. If the
+// actor ID cannot be resolved to an address, the function must return
+// address.Undef.
+type AddressResolver func(context.Context, abi.ActorID, *types.TipSet) address.Address
 
 type EventFilter interface {
 	Filter
@@ -77,9 +81,6 @@ func (f *eventFilter) CollectEvents(ctx context.Context, te *TipSetEvents, rever
 		return nil
 	}
 
-	// cache of lookups between actor id and f4 address
-	addressLookups := make(map[abi.ActorID]address.Address)
-
 	ems, err := te.messages(ctx)
 	if err != nil {
 		return xerrors.Errorf("load executed messages: %w", err)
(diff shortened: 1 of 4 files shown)


BigLep:
@rvagg : is this something you're hoping to see land for 1.31.0?

rvagg:
Yes, but I'd like chainindexer to land first, this can go in separately. Like #12623 I think they deserve separate call-outs in the CHANGELOG.

rjan90:
@rvagg should we land this now, given that we are looking at release v1.31.0-rc1 tomorrow?

chain/events/filter/event.go
```diff
@@ -28,7 +29,10 @@ func isIndexedValue(b uint8) bool {
 	return b&(types.EventFlagIndexedKey|types.EventFlagIndexedValue) > 0
 }
 
-type AddressResolver func(context.Context, abi.ActorID, *types.TipSet) (address.Address, bool)
+// AddressResolver is a function that resolves an actor ID to an address. If the
+// actor ID cannot be resolved to an address, the function should return
```
masih, line 33:
```suggestion
// actor ID cannot be resolved to an address, the function must return
```
...because the code correctness depends on it?

chain/events/filter/event.go
```diff
@@ -319,11 +314,18 @@ func (m *EventFilterManager) Apply(ctx context.Context, from, to *types.TipSet)
...
 
-	// TODO: could run this loop in parallel with errgroup if there are many filters
+	tsAddressResolver := m.createTipSetCachedAddressResolver()
+
+	g, ctx := errgroup.WithContext(ctx)
 	for _, f := range m.filters {
-		if err := f.CollectEvents(ctx, tse, false, m.AddressResolver); err != nil {
-			return err
-		}
+		f := f
```
masih, line 321: Redundant now that this module depends on Go [1.22](https://github.com/filecoin-project/lotus/blob/ffe5b280804cea2a3bf1500da922b1811635d693/go.mod#L3).
```suggestion
```

chain/events/filter/event.go
```diff
@@ -344,16 +346,41 @@ func (m *EventFilterManager) Revert(ctx context.Context, from, to *types.TipSet)
...
 
-	// TODO: could run this loop in parallel with errgroup if there are many filters
+	tsAddressResolver := m.createTipSetCachedAddressResolver()
+
+	g, ctx := errgroup.WithContext(ctx)
 	for _, f := range m.filters {
-		if err := f.CollectEvents(ctx, tse, true, m.AddressResolver); err != nil {
-			return err
-		}
+		f := f
```
masih, line 353: Ditto.

chain/events/filter/event.go
```diff
@@ -344,16 +346,41 @@ func (m *EventFilterManager) Revert(ctx context.Context, from, to *types.TipSet)
...
+	}
+
+	if err := g.Wait(); err != nil {
+		return err
 	}
 
 	return nil
 }
 
+func (m *EventFilterManager) createTipSetCachedAddressResolver() func(ctx context.Context, emitter abi.ActorID, ts *types.TipSet) address.Address {
```
masih, line 366: Subjective: 
* This could be a function (in FP sense): takes an address resolver and returns a cached version of it.

Then:
* One could wrap the struct field resolver with it at EventFilterManager.
* This would avoid multiple calls to `createTipSetCachedAddressResolver` in favour of singular cached address resolver for the lifetime of event filter manager (assuming this is desirable for performance reasons, though watch out for unbounded map size growth).

masih (approved):

chain/events/filter/event.go
rvagg: Well, I'm only caching `abi.ActorID => address.Address`, not including the tipset, so I only want it cached for the particular tipset and then discarded since the mapping may vary between tipsets. We could potentially add tipset to the cache, make it an LRU to conserve some memory, but I feel like we have too many permanent caches across the codebase already that contribute to memory bloat.

rvagg: I've turned it from a method to a function that takes a resolver and added some docs to it to clarify its intended usage.

masih (approved):

PR #12609, opened 2024-10-17
feat(cli): lotus chain head --height to print epoch number

rvagg:
A simpler version of what's currently possible with:

	lotus chain list --format '<height>' --count 1


rjan90 (approved):
Needs a make gen / make docsgen-cli, but else is fine

PR #12614, opened 2024-10-18
fix(chainindex): recompute tipset when we find no receipts

rvagg:
Can't backfill from snapshot without doing this earlier, snapshots don't have receipts either.

### DIFF
--- a/chain/index/events.go
+++ b/chain/index/events.go
@@ -144,9 +144,35 @@ func loadExecutedMessages(ctx context.Context, cs ChainStore, recomputeTipSetSta
 
 	st := cs.ActorStore(ctx)
 
+	var recomputed bool
+	recompute := func() error {
+		tskCid, err2 := rctTs.Key().Cid()
+		if err2 != nil {
+			return xerrors.Errorf("failed to compute tipset key cid: %w", err2)
+		}
+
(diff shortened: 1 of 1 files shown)


aarshkshah1992:
My only contention here would be if it's needed/if the Indexer is the right place to do this. 

The right way to solve this is to include receipts in the snapshots. Making every snapshot synced node recompute tipsets in the snapshot to re-generate the receipts sounds excessive.

rvagg:
> The right way to solve this is to include receipts in the snapshots

Unlikely to happen, everyone's been working to make snapshots _smaller_, and with f3 they're likely to trim them even further. Adding at least one more block per message is not going to be a small addition, and the user-base for consuming these is going to be very tiny. Who else is going to backfill events like this other than RPC providers who have spun up a new node and want to maximise their history depth? I'm doubtful they even do anything like this today—we don't have recompute in the current backfill so I'd really like to know what they are doing now.

aarshkshah1992:
@rvagg 

```
we don't have recompute in the current backfill so I'd really like to know what they are doing now.
```

I am fairly confident that that they return "not founds/errors" right now for the snapshot tipsets. 
Given that most RPC providers restart only once every _few_ weeks, they probably get away with it.

aarshkshah1992:
@rvagg Okay, I see your point. Let's go ahead with this change. But let's do it when user kicks off a manual backfill (which is what your PR is doing anyways) -> not when the snapshot is populated.

PR #12615, opened 2024-10-18
fix: support Forrest parity testing for F3 participation APIs

masih:
## Proposed Changes
For parity testing, Forest nodes need to decode an F3 participation ticket that is issued by Lotus. Lotus so happens to use CBOR to encode tickets, which it then decodes to issue a lease.

The lease issuer is of type `peer.ID` which is an alias of type `string` that may contain non UTF-8 characters. If this type is used directly in CBOR encoding then it gets encoded as string, which works fine in Golang but not Rust, which in turn results in decoding issues in Forest.

To avoid this use `[]byte` as the type to encode the issuer public key. In Lotus, the value will be the binary marshalling of `peer.ID` that issued the ticket.

While at it, also fix two issues that were brought up during discussion:

* A miner must not be able to ask for a ticket to participate in zero instances. Validate and return error if instances is set to zero.

* Use CBOR tuple encoding for a slightly better efficient wire encoding of ticket. `cborgen.Write*` APIs explicitly create a given file; hence the need to separate the file to which tuple kinds are generated since the existing encoding for types in `api/cbor_gen.go` use maps.

## Checklist

- [x] CI is green


api/api_full.go
```diff
@@ -981,7 +980,7 @@ type F3ParticipationLease struct {
 	// Network is the name of the network this lease belongs to.
 	Network gpbft.NetworkName
 	// Issuer is the identity of the node that issued the lease.
-	Issuer peer.ID
+	Issuer []byte
```
Stebalien, line 983: Can we use a custom type alias of PeerID that gets encoded as a string (base58 encoded) in json, but otherwise gets encoded as bytes? Not super critical, but it's nice being able to read it as a string.

Alternatively, encode it as a base58 string always?

See the OpenRPC example json for why I'm making this comment.

masih: Going with base58 encoding everywhere 👍

Kubuxu (approved):

PR #12618, opened 2024-10-20
feat: implement methods for `eth_getTransactionByBlockNumberAndIndex` and `eth_getTransactionByBlockHashAndIndex`

### LINKED ISSUE #10640: Support eth_getTransactionByBlockNumberAndIndex and eth_getTransactionByBlockHashAndIndex
These are currently unimplemented and some ecosystem partners seem to require them.

virajbhartiya:
Closes #10640

## Proposed Changes
Implement methods for `eth_getTransactionByBlockNumberAndIndex` and `eth_getTransactionByBlockHashAndIndex

## Checklist

- [ ] CI is green


node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
-func (a *EthAPI) EthGetTransactionByBlockNumberAndIndex(context.Context, ethtypes.EthUint64, ethtypes.EthUint64) (ethtypes.EthTx, error) {
-	return ethtypes.EthTx{}, ErrUnsupported
+func (a *EthAPI) EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNum ethtypes.EthUint64, index ethtypes.EthUint64) (ethtypes.EthTx, error) {
+	ts, err := getTipsetByBlockNumber(ctx, a.Chain, strconv.FormatUint(uint64(blkNum), 10), false)
+	if err != nil {
+		return ethtypes.EthTx{}, xerrors.Errorf("failed to get tipset: %w", err)
+	}
+	return a.getTransactionByTipsetAndIndex(ctx, ts, index)
+}
+func (a *EthAPI) getTransactionByTipsetAndIndex(ctx context.Context, ts *types.TipSet, index ethtypes.EthUint64) (ethtypes.EthTx, error) {
```
rvagg, line 590:
```suggestion
}

func (a *EthAPI) getTransactionByTipsetAndIndex(ctx context.Context, ts *types.TipSet, index ethtypes.EthUint64) (ethtypes.EthTx, error) {
```

rvagg:
These should go in api_gateway.go along with the other Eth methods in there - they are reasonable to expose publicly. Once you do that you'll get some compile errors. Go to gateway/proxy_eth.go and have a look at how the methods that take a `blkNum` argument handle it - you want to `checkTipsetHeight` at some point, either directly or via `checkEthBlockParam`--I think we only need the latter if we're supporting the string predefined variants (`'latest'` etc.) (are we with these?). The main point is that users of the gateway need to be able to set an epoch lookback limit on queires and queries that resolve to tipsets older than that limit should error.

itests/fevm_test.go
```diff
@@ -1436,3 +1436,78 @@ func TestEthGetBlockByNumber(t *testing.T) {
...
+
+	// Use the block hash from the receipt
+	blockHash := receipt.BlockHash
+
+	// Get the block by its hash
+	block, err := client.EthGetBlockByHash(ctx, blockHash, false)
+	require.NoError(t, err)
+	require.NotNil(t, block)
+
+	ethTx, err := client.EthGetTransactionByBlockHashAndIndex(ctx, block.Hash, ethtypes.EthUint64(0))
```
rvagg, line 1466: This is potentialy flaky because it's not certain that you're going to find the transaction you want at position `0` here - it could be that one a miner that's submitting windowposts to keep the chain alive submits something at this same block, so you can't be sure unfortunately.

I think the better approach for these tests would be similar to `TestMultipleEvents` in eth_filter_test.go which sets up a scenario where it's sure that all transactions got into the same tipset before it proceeds to check them. So in here, you'd do submit your transaction (ideally multiple so you can test more than `0`), confirm that the tipset is shaped the way you want, and then proceed to verify. You could inspect a tipset to see the messages included in the way you want, then once you've got what you need then do your queries with your new API and continue.

It would also be OK if you did both of these APIs in a single test case too, it'll be cheaper to run that way too!

node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+		if err != nil {
+			log.Errorf("Failed to create pending Ethereum transaction: %v", err)
+			return ethtypes.EthTx{}, xerrors.Errorf("failed to create pending Ethereum transaction: %w", err)
+		}
+		return pendingTx, nil
+	}
+
+	log.Debugf("Message found in chain: %+v", msgLookup)
+
+	tx, err := newEthTxFromMessageLookup(ctx, msgLookup, -1, a.Chain, a.StateAPI)
```
rvagg, line 639: If possible, we should avoid a `StateSearchMsg`, and I think that we could bypass the need for it by jumping straight to `newEthTx`. `MsgLookup` is useful if you're finding a message that's isolated from a tipset, but in this case you have the tipset already and its list of messages so you shouldn't have to go hunting for it.

itests/fevm_test.go
rvagg: You should also test the case where there's a tipset but no messages, what should that return? And where the index is >= number of messages. Let's ensure we have the error cases covered and get roughly the message we expect (you can use `ErrorContains` to check just a portion of the message you care about).

node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
 	return &receipt, nil
 }
 
-func (a *EthAPI) EthGetTransactionByBlockHashAndIndex(context.Context, ethtypes.EthHash, ethtypes.EthUint64) (ethtypes.EthTx, error) {
-	return ethtypes.EthTx{}, ErrUnsupported
+func (a *EthAPI) EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, index ethtypes.EthUint64) (ethtypes.EthTx, error) {
+	ts, err := a.Chain.GetTipSetByCid(ctx, blkHash.ToCid())
```
aarshkshah1992, line 575: This should return a `*ethtypes.EthTx` so we can return nil if there is no such TX.

node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+	if err != nil {
+		return ethtypes.EthTx{}, xerrors.Errorf("failed to find tipset for block hash %s: %w", blkHash, err)
+	}
+
+	return a.getTransactionByTipsetAndIndex(ctx, ts, index)
 }
 
-func (a *EthAPI) EthGetTransactionByBlockNumberAndIndex(context.Context, ethtypes.EthUint64, ethtypes.EthUint64) (ethtypes.EthTx, error) {
-	return ethtypes.EthTx{}, ErrUnsupported
+func (a *EthAPI) EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNum ethtypes.EthUint64, index ethtypes.EthUint64) (ethtypes.EthTx, error) {
```
aarshkshah1992, line 583: This should return a *ethtypes.EthTx so we can return nil if there is no such TX.

node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+	ts, err := getTipsetByBlockNumber(ctx, a.Chain, strconv.FormatUint(uint64(blkNum), 10), false)
```
aarshkshah1992, line 584: The. block number will be hexadecimal or one of the block param tags mentioned here:

See https://docs.infura.io/api/networks/arbitrum/json-rpc-methods/eth_gettransactionbyblocknumberandindex

image

aarshkshah1992: I think rather than taking ` blkNum ethtypes.EthUint64` as a param here, you should take a `blkParam` string just like we take in the other functions where user can pass in a block number OR a block param tag ("latest", "pending" etc and pass that to `getTipsetByBlockNumber` here).

node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+func (a *EthAPI) getTransactionByTipsetAndIndex(ctx context.Context, ts *types.TipSet, index ethtypes.EthUint64) (ethtypes.EthTx, error) {
+	msgs, err := a.Chain.MessagesForTipset(ctx, ts)
+	if err != nil {
+		log.Errorf("Failed to get messages for tipset: %v", err)
+		return ethtypes.EthTx{}, xerrors.Errorf("failed to get messages for tipset: %w", err)
+	}
+
+	if uint64(index) >= uint64(len(msgs)) {
+		log.Errorf("Transaction index %d out of bounds (total messages: %d)", index, len(msgs))
+		return ethtypes.EthTx{}, xerrors.Errorf("transaction index out of bounds")
```
aarshkshah1992, line 599: error message should also say how many transactions are in the tipset so user can see why it's out of bounds.

aarshkshah1992 (commented):
@virajbhartiya Reviewed. good stuff.

virajbhartiya:
@rvagg @aarshkshah1992 I have addressed the above reviews. Both the tests have been merged into one. The `EthGetTransactionByBlockNumberAndIndex` can now accept either hexadecimal values or "latest", "pending", etc. I have also made the changes in `getTransactionByTipsetAndIndex` by removing the MsgLookup since the tipset is already there in the list of messages.  Thanks for pointing me out to the resources!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - `lotus chain head` now supports a `--height` flag to print just the epoch number of the current chain head ([filecoin-project/lotus#12609](https://github.com/filecoin-project/lotus/pull/12609))
 - `lotus-shed indexes inspect-indexes` now performs a comprehensive comparison of the event index data for each message by comparing the AMT root CID from the message receipt with the root of a reconstructed AMT. Previously `inspect-indexes` simply compared event counts, comparing AMT roots confirms all the event data is byte-perfect. ([filecoin-project/lotus#12570](https://github.com/filecoin-project/lotus/pull/12570))
 - Expose APIs to list the miner IDs that are currently participating in F3 via node. ([filecoin-project/lotus#12608](https://github.com/filecoin-project/lotus/pull/12608))
+- Implement `EthGetTransactionByBlockNumberAndIndex` and `EthGetTransactionByBlockHashAndIndex` methods. ([filecoin-project/lotus#12618](https://github.com/filecoin-project/lotus/pull/12618))
```
rvagg, line 13:
```suggestion
- Implement `EthGetTransactionByBlockNumberAndIndex` (`eth_getTransactionByBlockNumberAndIndex`) and `EthGetTransactionByBlockHashAndIndex` (`eth_getTransactionByBlockHashAndIndex`) methods. ([filecoin-project/lotus#12618](https://github.com/filecoin-project/lotus/pull/12618))
```

may as well maximise the recognition factor for people used to the proper names

gateway/proxy_eth.go
```diff
@@ -200,6 +200,14 @@ func (gw *Node) EthGetBlockByNumber(ctx context.Context, blkNum string, fullTxIn
 	return gw.target.EthGetBlockByNumber(ctx, blkNum, fullTxInfo)
 }
 
+func (gw *Node) EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, txIndex ethtypes.EthUint64) (*ethtypes.EthTx, error) {
+	return gw.target.EthGetTransactionByBlockHashAndIndex(ctx, blkHash, txIndex)
+}
+
+func (gw *Node) EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkNum string, txIndex ethtypes.EthUint64) (*ethtypes.EthTx, error) {
+	return gw.target.EthGetTransactionByBlockNumberAndIndex(ctx, blkNum, txIndex)
```
rvagg, line 208: needs rate limiting and tipset lookback checking

see `EthGetBlockByNumber` in here for how to do this, it can mostly be copied from there

gateway/proxy_eth.go
```diff
@@ -200,6 +200,14 @@ func (gw *Node) EthGetBlockByNumber(ctx context.Context, blkNum string, fullTxIn
 	return gw.target.EthGetBlockByNumber(ctx, blkNum, fullTxInfo)
 }
 
+func (gw *Node) EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, txIndex ethtypes.EthUint64) (*ethtypes.EthTx, error) {
+	return gw.target.EthGetTransactionByBlockHashAndIndex(ctx, blkHash, txIndex)
```
rvagg, line 204: needs rate limiting and tipset lookback checking

see `EthGetBlockTransactionCountByHash` in here for how to do this

node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+
+	if validBlockTags[blkParam] {
+		blockNumber = blkParam
+	} else {
+		blkNum, err := ethtypes.EthUint64FromHex(blkParam)
+		if err != nil {
+			return nil, fmt.Errorf("invalid block number format: %w", err)
+		}
+		blockNumber = strconv.FormatUint(uint64(blkNum), 10)
+	}
```
rvagg, line 601:
```suggestion
	var blockNumber string

	switch blkParam {
	case "latest", "pending", "earliest", "final":
		blockNumber = blkParam
	default:
		blkNum, err := ethtypes.EthUint64FromHex(blkParam)
		if err != nil {
			return nil, xerrors.Errorf("invalid block number format: %w", err)
		}
		blockNumber = strconv.FormatUint(uint64(blkNum), 10)
	}
```

also note the use of `xerrors` here - it's used in lotus so we get error chaining, which `fmt.Errorf` doesn't support

rvagg: I'm not sure we need all this checking anyway, doesn't `getTipsetByBlockNumber` do it for us? wouldn't it just be the same as EthGetBlockByNumber?

node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+
+	ts, err := getTipsetByBlockNumber(ctx, a.Chain, blockNumber, true)
+	if err != nil {
+		if err == ErrNullRound {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to get tipset for block %s: %w", blockNumber, err)
```
rvagg, line 608: xerrors, same for the other error returns from here down

node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+	}
+
+	if ts == nil {
```
rvagg, line 611:
```suggestion
	} else if ts == nil {
```

node/impl/full/eth.go
```diff
@@ -570,12 +571,78 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+	}
+
+	msg := msgs[index]
+
+	cid, err := ts.Key().Cid()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get tipset key cid: %w", err)
+	}
+
+	tx, err := newEthTx(ctx, a.Chain, nil, ts.Height(), cid, msg.Cid(), int(index))
```
rvagg, line 640: I think you'll need the `stateTree`, since this API can get any message it's going to encounter `SigTypeSecp256k1` and `SigTypeBLS` messages too. In fact this might be worth checking in an itest because currently it's just looking at eth (delegated) transactions.

itests/fevm_test.go
```diff
@@ -1436,3 +1436,97 @@ func TestEthGetBlockByNumber(t *testing.T) {
...
+		receipts = append(receipts, receipt)
+	}
+
+	// Ensure all transactions are in the same tipset
+	blockHash := receipts[0].BlockHash
+	blockNumber := receipts[0].BlockNumber
+	for _, receipt := range receipts[1:] {
+		require.Equal(t, blockHash, receipt.BlockHash)
+		require.Equal(t, blockNumber, receipt.BlockNumber)
+	}
```
rvagg, line 1475: this is going to be flaky, because they legitimately may not be, you're depending on `SubmitTransaction` to queue up and all be accepted in the same tipset which you don't have a guarantee of.

You'll need to put this, and the 3 loops before it, into a loop, and repeat until they are all in the same tipset - you could just do that in the `WaitTransaction` loop - get the first one, then check its block against the subsequent ones. If they don't match, `continue`, otherwise `break`.

node/impl/full/eth.go
```diff
@@ -175,6 +175,7 @@ type EthAPI struct {
 
 	EthModuleAPI
 	EthEventAPI
+	StateAPI StateAPI
```
rvagg, line 178: what do we need this for?

virajbhartiya: Oh, it was left to be removed, in my previous implementation it was being used for MsgLookup which is not required anymore. I'll remove it

node/impl/full/eth.go
aarshkshah1992: I think `getTipsetByBlockNumber ` should just work.

gateway/proxy_eth.go
aarshkshah1992: @virajbhartiya You have done something similar in past PRs as well. Would reccomend to look back at the work you've done for a sanity check to see what's missing in the PRs you raise.

node/impl/full/eth.go
aarshkshah1992: @virajbhartiya Basically, if you follow the implementation of the ` newEthTx` -> you'll see that it can return "ETH transactions" even for messages whose address is NOT delegated that Filecoin native messages. But to do that -> we need to pass the state tree here. You can see how to get the state tree from other ETH RPC APIs.

aarshkshah1992: You can then find an itest somewhere which submits Filecoin transactions and fetch those using these two APIs and ensure it works.

virajbhartiya:
@rvagg @aarshkshah1992 I have addressed the comments above. used `Xerrors` isntead of `fmt`. Added rate and lookback limits for both the methods, and added tests for both - `SigTypeSecp256k1` and `SigTypeBLS` messages.

itests/fevm_test.go
```diff
@@ -1436,3 +1436,107 @@ func TestEthGetBlockByNumber(t *testing.T) {
...
+	for _, receipt := range receipts {
+		blockHashes = append(blockHashes, receipt.BlockHash)
+		blockNumbers = append(blockNumbers, receipt.BlockNumber)
+	}
+
+	for _, hash := range txHashes {
+		var ethTx *ethtypes.EthTx
+		var err error
+		for _, blockHash := range blockHashes {
+			ethTx, err = client.EthGetTransactionByBlockHashAndIndex(ctx, blockHash, ethtypes.EthUint64(0))
```
rvagg, line 1485: @virajbhartiya this still isn't quite doing what we need, your fundamental problem is back when you deploy the contracts.

Have a close look at TestMultipleEvents in eth_filter_test.go, in this loop: https://github.com/filecoin-project/lotus/blob/02a8b972db82eee0ff109366f57122caf7790128/itests/eth_filter_test.go#L694-L741

Note particularly the `allInSameTipset` check at the end and the fact that the loop will continue until it gets that correct. It'll keep on submitting 3 transactions (from `paramsArray`) until it can see that they all landed in the same tipset.

Your test as it is now is submitting _and_ waiting for each transaction, so they're guaranteed to land on separate tipsets so you only ever get to test index=0. Instead, you should:

1. Set up a `for {}` loop
2. Run through your `numTx` loop, only doing `SubmitTransaction` for each of them
3. Wait for each of the hashes you submitted in another loop, so we know they all landed
4. Check all of the receipts that their `BlockHash` is the same.
5. If they are all the same, break, if they are different, continue the loop.

Then you should have `numTx` transactions (messages) in a single block (tipset).

The second problem you have is that you can't guarantee that they all landed in the order you expect, _or_ that they are the only ones to land in the tipset! There is a miner in the background that you're not interacting with that's keeping the chain going and it could have got something in that tipset as well. So what I think you should do is use the `receipts[index].TransactionIndex` when you're calling `EthGetTransactionByBlockHashAndIndex` rather than a fixed number. But note also your "Index out of range" has this problem too - you don't know how many should be in that tipset, there may be more than you submitted. I think for that one you should just request a high, improbable index, like `100`.

virajbhartiya: We need multiple transaction in the same `tipset`, since the function is designed to iterate through transactions in a single tipset - testing with just one transaction per tipset wouldn't verify that the indexing mechanism works correctly when there are multiple transactions to index through. Multiple transactions allow testing different index positions (0, 1, 2, etc.)
This verifies the function can correctly retrieve transactions at different positions within a block. so looping it in `for {}` and checking `allInSameTipset` makes sure that all land in the same `tipset`.

node/impl/full/eth.go
```diff
@@ -570,12 +570,66 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
 	return &receipt, nil
 }
 
-func (a *EthAPI) EthGetTransactionByBlockHashAndIndex(context.Context, ethtypes.EthHash, ethtypes.EthUint64) (ethtypes.EthTx, error) {
-	return ethtypes.EthTx{}, ErrUnsupported
+func (a *EthAPI) EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, index ethtypes.EthUint64) (*ethtypes.EthTx, error) {
+	ts, err := a.Chain.GetTipSetByCid(ctx, blkHash.ToCid())
+	if err != nil {
+		return nil, xerrors.Errorf("block not found")
```
rvagg, line 576:
```suggestion
		return nil, xerrors.Errorf("failed to get tipset by cid: %w", err)
```

using xerrors with `%w` gives us nicer stack traces, we just need to remember to include the err

node/impl/full/eth.go
```diff
@@ -570,12 +570,66 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+	}
+
+	return a.getTransactionByTipsetAndIndex(ctx, ts, index)
+}
+
+func (a *EthAPI) EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkParam string, index ethtypes.EthUint64) (*ethtypes.EthTx, error) {
+
```
rvagg, line 583:
```suggestion
```

node/impl/full/eth.go
```diff
@@ -570,12 +570,66 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+	ts, err := getTipsetByBlockNumber(ctx, a.Chain, blkParam, true)
+	if err != nil {
+		if err == ErrNullRound {
+			return nil, nil
+		}
```
rvagg, line 588: See https://github.com/filecoin-project/lotus/issues/12633 that I just wrote up today, let's not introduce another case of `null` for null rounds here since only 1 of our current APIs do this and apparently (TIL) that's not in line with what an Eth dev would expect `null` to mean, we should just return the right error here.

itests/fevm_test.go
```diff
@@ -1436,3 +1436,125 @@ func TestEthGetBlockByNumber(t *testing.T) {
...
+		for i := 1; i < len(receipts); i++ {
+			if receipts[i].BlockHash != receipts[0].BlockHash {
+				allInSameTipset = false
+				break
+			}
+		}
+
+		if allInSameTipset {
+			break
+		}
```
rvagg, line 1508: optional, but you could `t.Logf` here to say you're looping because they didn't land in the same tipset - that may help with a pathological case of the test timing out because we couldn't get them to land (I don't know if we'd ever experience that, so it's probably unnecessary)

rvagg (changes requested):
looking good! almost there, just some minor suggestions left

itests/fevm_test.go
virajbhartiya: I'll add `t.Logf("Retrying because transactions didn't land in the same tipset")`

node/impl/full/eth.go
virajbhartiya: I'll add `return nil, xerrors.Errorf("block number %s was a null round", blkParam)` and same for `EthGetTransactionByBlockHashAndIndex` as well

node/impl/full/eth.go
```diff
@@ -570,12 +570,68 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
+	return a.getTransactionByTipsetAndIndex(ctx, ts, index)
+}
+
+func (a *EthAPI) EthGetTransactionByBlockNumberAndIndex(ctx context.Context, blkParam string, index ethtypes.EthUint64) (*ethtypes.EthTx, error) {
+	ts, err := getTipsetByBlockNumber(ctx, a.Chain, blkParam, true)
+	if err != nil {
+		if err == ErrNullRound {
+			return nil, xerrors.Errorf("block number %s was a null round", blkParam)
+		}
+		return nil, xerrors.Errorf("failed to get tipset for block %s: %w", blkParam, err)
```
rvagg, line 591:
```suggestion
		return nil, xerrors.Errorf("failed to get tipset for block %s: %w", blkParam, err)
```

let's address this as part of https://github.com/filecoin-project/lotus/issues/12633

for this one, we just need to be sure that we pass the `ErrNullRound` on as it is - your special case branch is returning a new error that doesn't include it (`%w` would have been OK, but we also may as well just return it.

As per https://github.com/filecoin-project/lotus/issues/12633, the important part is that the text contained in the existing `ErrNullRound` is still included, regardless of what else we include, so we either return a fresh `ErrNullRound` (and perhaps decorate it further with the block number, as long as it just extends, and doesn't change, the existing message), so a `%w` does that for us.

node/impl/full/eth.go
```diff
@@ -570,12 +570,68 @@ func (a *EthModule) EthGetTransactionReceiptLimited(ctx context.Context, txHash
...
 }
 
-func (a *EthAPI) EthGetTransactionByBlockHashAndIndex(context.Context, ethtypes.EthHash, ethtypes.EthUint64) (ethtypes.EthTx, error) {
-	return ethtypes.EthTx{}, ErrUnsupported
+func (a *EthAPI) EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, index ethtypes.EthUint64) (*ethtypes.EthTx, error) {
+	ts, err := a.Chain.GetTipSetByCid(ctx, blkHash.ToCid())
+	if err != nil {
+		if err == ErrNullRound {
+			return nil, xerrors.Errorf("block hash %s was a null round", blkHash)
+		}
```
rvagg, line 578:
```suggestion
```

See below, let's address these all as part of https://github.com/filecoin-project/lotus/issues/12633

rvagg (changes requested):
all good except for those `ErrNullRound` inclusions, remove those and we'll merge this

virajbhartiya:
@rvagg Removed the null round checks for both thee functions

rvagg (approved):
great work!

PR #12619, opened 2024-10-21
fix(chainindexer): backfilling should halt when chain state data is missing and not backfill parents

aarshkshah1992:
For https://github.com/filecoin-project/lotus/pull/12421#discussion_r1805860944 AND https://filecoinproject.slack.com/archives/CP50PPW2X/p1729486344066509?thread_ts=1729485320.287299&cid=CP50PPW2X

**_Tested it on a calibnet node:_**

Screenshot 2024-10-21 at 11 56 53 AM

### DIFF
--- a/chain/index/api.go
+++ b/chain/index/api.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 
 	"github.com/ipfs/go-cid"
+	ipld "github.com/ipfs/go-ipld-format"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-state-types/abi"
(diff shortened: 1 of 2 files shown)


chain/index/api.go
```diff
@@ -285,33 +286,22 @@ func (si *SqliteIndexer) verifyIndexedData(ctx context.Context, ts *types.TipSet
...
-
-			return nil
+			return si.indexTipsetWithParentEvents(ctx, tx, ts, executionTs)
 		})
 	}
 
 	if err := backfillFunc(); err != nil {
-		return nil, xerrors.Errorf("failed to backfill tipset: %w", err)
+		if ipld.IsNotFound(err) {
+			return nil, xerrors.Errorf("failed to backfill tipset at epoch %d: chain store does not contain data", ts.Height())
```
rvagg, line 302:
```suggestion
			return nil, xerrors.Errorf("failed to backfill tipset at epoch %d: chain store does not contain data: %w", ts.Height(), err)
```

It might be wise to still return the original error, just in case there's something unexpected going on and we need to debug?

cmd/lotus-shed/chain_index.go
```diff
@@ -135,8 +136,10 @@ number of failed RPC calls. Otherwise, it will exit with a zero status.
 			_, _ = fmt.Fprintf(cctx.App.Writer, "%s starting chainindex validation; from epoch: %d; to epoch: %d; backfill: %t; log-good: %t\n", currentTimeString(),
 				fromEpoch, toEpoch, backfill, logGood)
 		}
-
 		totalEpochs := fromEpoch - toEpoch + 1
+		haltNoData := false
+		haltHeight := -1
```
rvagg, line 141: we could just use `haltHeight` as the signal without `haltNoData` by checking for `-1`

rvagg (approved):
seems fine to me, suggestions in line

PR #12631, opened 2024-10-23
chore: update network skeleton docs

kamuik16:
## Proposed Changes
Updates Docs to include the reference to the PRs to the latest network skeleton upgrade (24). 

cc @rjan90

Also there were two Filecoin-FFI section, so I removed one of those, lmk if I'm wrong.


BigLep (commented):
Good catch on the duplicate filecoin-ffi section.  I assume that was me...  Thanks for fixing.  

I'll let @rjan90 respond on if there is any reason not to update the doc to point to nv24 (more recent) example than the nv23 one.  I'm not aware of reason, but I think it would come down to if nv24 steps didn't follow the desired process for some reason.

kamuik16:
> Good catch on the duplicate filecoin-ffi section. I assume that was me... Thanks for fixing.
> 
> I'll let @rjan90 respond on if there is any reason not to update the doc to point to nv24 (more recent) example than the nv23 one. I'm not aware of reason, but I think it would come down to if nv24 steps didn't follow the desired process for some reason.

I had done the complete nv24 skeleton upgrade by following all the steps only, but happy to wait for @rjan90 approval.

rjan90 (approved):
Thanks! Looks good to me

Issue #12594, opened 2024-10-14
ChainValidateIndex should validate the actual content of the events and not just match on the number of events


aarshkshah1992:
See https://github.com/filecoin-project/lotus/pull/12450#discussion_r1794502554.

The `ChainValidateIndex` currently only compares the counts of the events in the Index with the number of events for  the epoch in the Index to validate the Indexed events.

However, we should also compare the actual content of the events in the chain state AMT with the Indexed events for a more comprehensive check.

See https://github.com/filecoin-project/lotus/issues/12570 and https://github.com/filecoin-project/lotus/pull/12571 for exploration of more complete checking.

BigLep:
@aarshkshah1992 : I'm not close to the details, but https://github.com/filecoin-project/lotus/pull/12571 is related right?

rvagg:
yes, https://github.com/filecoin-project/lotus/pull/12571 is the implementation for lotus-shed, all the hard logic is explored in there so a lot of that code can just be copied, like the original counting code was copied from there

aarshkshah1992:
Closed by https://github.com/filecoin-project/lotus/pull/12632.

PR #12632, opened 2024-10-23
feat(chainindex): compare events AMT root between Index and chain state for validation

aarshkshah1992:
Closes https://github.com/filecoin-project/lotus/issues/12570 AND https://github.com/filecoin-project/lotus/issues/12594.


rvagg (approved):
Cool, the only thing I'm a bit uncertain about is the event root always nil when there are no events, or can it be the empty amt cid? I suppose we'll see.

Issue #12633, opened 2024-10-23
eth_getBlockByNumber should return error for null round


rvagg:
`eth_getBlockByNumber` uniquely handles null rounds by returning a `nil` response (`null` via RPC), whereas all the other `eth_` APIs that take a block number will return an error.

go-ethereum's behaviour seems to be that `null` indicates that the number isn't available, or at least can't be found in the number<>block mapping, which sounds like it'd be the right thing for us to do for null rounds, but users take this to mean that the block isn't _yet_ available but will become at some point if they keep on polling for it because they just got ahead of the chain.

From @ilyalukyanov:

> In EVM null  means not found. Given all block numbers are sequential, that specifically means that a block in the future was requested. Therefore:
> * If a block isn't found, any block after it would also be not found
> * If a block isn't found, but you poll it for long enough, eventually the chain will arrive at it and it will be found.

We have special handling for null rounds in this API that returns `nil`: https://github.com/filecoin-project/lotus/blob/02a8b972db82eee0ff109366f57122caf7790128/node/impl/full/eth.go#L346-L354

We should return a consistent error for all of these cases, currently we don't. All cases of `getTipsetByBlockNumber`:

* `EthGetBlockByNumber`: return `nil`
* `EthFeeHistory`: return error: `bad block parameter X: requested epoch was a null round`
* `FilecoinAddressToEthAddress`: `failed to get tipset for block X: requested epoch was a null round`
* `EthTraceBlock`: `failed to get tipset: requested epoch was a null round`
* `EthTraceReplayBlockTransactions`: `failed to get tipset: requested epoch was a null round`

We should make these consistent.

1. Make `ErrNullRound` dynamic and let it include the requested block number/string: `"block number 0x1234 was a null round"`
1. In each of these cases, `errors.Is` for `ErrNullRound` and return the `err` directly (or `errors.As` and return the instance?)
2. Where it's not a `ErrNullRound` return a nested error description, but let's make it consistent eh?

rvagg:
Thanks to @eshon to alerting me to https://github.com/graphprotocol/graph-node/issues/4729#issuecomment-1650257347

It turns out that the `ErrNullRound` string is a hard-wired expectation in TheGraph: https://github.com/graphprotocol/graph-node/pull/5294/files#diff-b31a71a95f48aa7d0f11b5ea3501befc2560d749d73872e9eebe8003de85d921R1090-R1099

[...]

PR #12641, opened 2024-10-25
Revert "feat: fix GetBlock for null rounds by returning nil (#12529)"

rvagg:
This reverts commit c6745545041939b97962e376ee2593609bb3d5b5.

Discussion in https://github.com/filecoin-project/lotus/issues/12633 about the subtleties of dealing with this problem - returning nil or an error. This could be an unfixable wac-a-mole, but the fact that we had a complaint pretty quickly on an RC for just this API shows that we need to tread more cautiously.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 # UNRELEASED
 
 ## New features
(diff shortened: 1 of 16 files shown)


rvagg:
I've left `TestEthGetBlockByNumber` in here because @virajbhartiya added a bunch of good coverage in there, I've just changed some of the expectations based on this reversion.

rvagg:
I'm also ignoring the request to fix the title and I'm going to PR the bot to accept `Revert "...standard commit in here"` since it's the common git format for reversions.

aarshkshah1992 (approved):

PR #12655, opened 2024-10-29
feat: return an error for null rounds from RPC methods

virajbhartiya:
## Related Issues
Fixes [#12633](https://github.com/filecoin-project/lotus/issues/12633)

## Checklist

- [ ] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 # UNRELEASED
 
 ## New features
-- Return a consistent error when encountering null rounds in ETH RPC method calls. ([filecoin-project/lotus#12633](https://github.com/filecoin-project/lotus/pull/12633))
+- Return a consistent error when encountering null rounds in ETH RPC method calls. ([filecoin-project/lotus#12655](https://github.com/filecoin-project/lotus/pull/12655))
 - Reduce size of embedded genesis CAR files by removing WASM actor blocks and compressing with zstd. This reduces the `lotus` binary size by approximately 10 MiB. ([filecoin-project/lotus#12439](https://github.com/filecoin-project/lotus/pull/12439))
 - Add ChainSafe operated Calibration archival node to the bootstrap list ([filecoin-project/lotus#12517](https://github.com/filecoin-project/lotus/pull/12517))
 - `lotus chain head` now supports a `--height` flag to print just the epoch number of the current chain head ([filecoin-project/lotus#12609](https://github.com/filecoin-project/lotus/pull/12609))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # UNRELEASED
 
 ## New features
+- Return a consistent error for encountering null rounds in RPC method calls. ([filecoin-project/lotus#12633](https://github.com/filecoin-project/lotus/pull/12633))
 - Reduce size of embedded genesis CAR files by removing WASM actor blocks and compressing with zstd. This reduces the `lotus` binary size by approximately 10 MiB. ([filecoin-project/lotus#12439](https://github.com/filecoin-project/lotus/pull/12439))
 - Add ChainSafe operated Calibration archival node to the bootstrap list ([filecoin-project/lotus#12517](https://github.com/filecoin-project/lotus/pull/12517))
 - `lotus chain head` now supports a `--height` flag to print just the epoch number of the current chain head ([filecoin-project/lotus#12609](https://github.com/filecoin-project/lotus/pull/12609))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 # UNRELEASED
 
 ## New features
-- Return a consistent error for encountering null rounds in RPC method calls. ([filecoin-project/lotus#12633](https://github.com/filecoin-project/lotus/pull/12633))
+- Return a consistent error when encountering null rounds in ETH RPC method calls. ([filecoin-project/lotus#12633](https://github.com/filecoin-project/lotus/pull/12633))
 - Reduce size of embedded genesis CAR files by removing WASM actor blocks and compressing with zstd. This reduces the `lotus` binary size by approximately 10 MiB. ([filecoin-project/lotus#12439](https://github.com/filecoin-project/lotus/pull/12439))
 - Add ChainSafe operated Calibration archival node to the bootstrap list ([filecoin-project/lotus#12517](https://github.com/filecoin-project/lotus/pull/12517))
 - `lotus chain head` now supports a `--height` flag to print just the epoch number of the current chain head ([filecoin-project/lotus#12609](https://github.com/filecoin-project/lotus/pull/12609))
--- a/api/api_errors.go
+++ b/api/api_errors.go
@@ -204,3 +204,7 @@ func (e *ErrNullRound) ToJSONRPCError() (jsonrpc.JSONRPCError, error) {
 		Data:    e.Epoch,
 	}, nil
 }
+func (e *ErrNullRound) Is(target error) bool {
+	_, ok := target.(*ErrNullRound)
+	return ok
+}
--- a/api/api_errors.go
+++ b/api/api_errors.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"errors"
+	"fmt"
 	"reflect"
 
 	"golang.org/x/xerrors"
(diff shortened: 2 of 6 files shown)


itests/fevm_test.go
```diff
@@ -1659,3 +1660,100 @@ func TestEthEstimateGas(t *testing.T) {
...
+		require.NoError(t, err)
+		if ts.Height() == nullHeight {
+			nullHeight--
+		} else {
+			break
+		}
+	}
+
+	nullBlockHex := fmt.Sprintf("0x%x", nullHeight)
+
```
rvagg, line 1697:
```suggestion

	client.WaitTillChain(ctx, kit.HeightAtLeast(nullHeight+2))
```

Again, let's get some distance from the null rounds so we're sure we're looking in an encoded history

itests/fevm_test.go
```diff
@@ -1659,3 +1660,100 @@ func TestEthEstimateGas(t *testing.T) {
...
+		ts, err := client.ChainGetTipSetByHeight(ctx, nullHeight, types.EmptyTSK)
+		require.NoError(t, err)
+		if ts.Height() == nullHeight {
+			nullHeight--
+		} else {
+			break
+		}
+	}
+
+	nullBlockHex := fmt.Sprintf("0x%x", nullHeight)
```
rvagg, line 1696:
```suggestion
	nullBlockHex := fmt.Sprintf("0x%x", int(nullHeight))
```

this is the real bug you're encountering - `%x` is doing a string version of the argument, but it's an `abi.ChainEpoch` which has its own `String()` method. So we'll just make sure `Sprintf` knows it's just an `int`. If you print your output you'll see it's not what you want and instead it's encoding the string version of the epoch as the bytes of that string, which becomes a much bigger hex number than you want it to be.

itests/fevm_test.go
```diff
@@ -1659,3 +1660,100 @@ func TestEthEstimateGas(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := tc.testFunc()
+			t.Logf("error: %v", err)
+			require.Error(t, err)
+
+			// Test errors.Is
+			require.True(t, errors.Is(err, new(api.ErrNullRound)),
+				"error should be or wrap ErrNullRound")
```
rvagg, line 1747:
```suggestion
			require.ErrorIs(t, err, new(api.ErrNullRound), "error should be or wrap ErrNullRound")
```

itests/fevm_test.go
```diff
@@ -1659,3 +1660,100 @@ func TestEthEstimateGas(t *testing.T) {
...
+
+			// Test errors.As and verify message
+			var nullRoundErr *api.ErrNullRound
+			require.True(t, errors.As(err, &nullRoundErr),
+				"error should be convertible to ErrNullRound")
```
rvagg, line 1752:
```suggestion
			require.ErrorAs(t, err, &nullRoundErr, "error should be convertible to Er
rNullRound")
```

itests/fevm_test.go
```diff
@@ -1659,3 +1660,100 @@ func TestEthEstimateGas(t *testing.T) {
...
+		name     string
+		testFunc func() error
+	}{
+		{
+			name: "EthGetBlockByNumber",
+			testFunc: func() error {
+				_, err := client.EthGetBlockByNumber(ctx, nullBlockHex, true)
+				if err == nil {
+					return errors.New("expected error for null round")
+				}
```
rvagg, line 1708:
```suggestion
```

api/api_errors.go
```diff
@@ -160,3 +165,42 @@ func NewErrExecutionReverted(reason string) *ErrExecutionReverted {
...
+	return nil
+}
+
+func (e *ErrNullRound) ToJSONRPCError() (jsonrpc.JSONRPCError, error) {
+	return jsonrpc.JSONRPCError{
+		Code:    ENullRound,
+		Message: e.Message,
+		Data:    e.Epoch,
+	}, nil
+}
```
rvagg, line 206:
```suggestion
}

// Is performs a non-strict type check, we only care if the target is an ErrNullRound
// and will ignore the contents (specifically there is no matching on Epoch).
func (e *ErrNullRound) Is(target error) bool {
       _, ok := target.(*ErrNullRound)
       return ok
}
```

We're going to need our own `Is` implementation because `errors.Is` does a direct match, which is fine for sentinel errors but this isn't a sentinel. In addition, we're going to do it non-strict. A standard form of this might check that `Epoch` was the same, but all we care about is this as as if it's a sentinel error.

api/api_errors.go
```diff
@@ -160,3 +165,46 @@ func NewErrExecutionReverted(reason string) *ErrExecutionReverted {
...
+func (e *ErrNullRound) Is(target error) bool {
```
rvagg, line 207:
```suggestion
// Is performs a non-strict type check, we only care if the target is an ErrNullRound
// and will ignore the contents (specifically there is no matching on Epoch).
func (e *ErrNullRound) Is(target error) bool {
```

@virajbhartiya this is worth a comment, since we are not doing a strict matching

itests/fevm_test.go
```diff
@@ -1659,3 +1660,93 @@ func TestEthEstimateGas(t *testing.T) {
...
+			// Test errors.Is
+			require.ErrorIs(t, err, new(api.ErrNullRound), "error should be or wrap ErrNullRound")
+
+			// Test errors.As and verify message
+			var nullRoundErr *api.ErrNullRound
+			require.ErrorAs(t, err, &nullRoundErr, "error should be convertible to ErrNullRound")
+
+			expectedMsg := fmt.Sprintf("requested epoch was a null round (%d)", nullHeight)
+			require.Equal(t, expectedMsg, nullRoundErr.Error())
+			require.Equal(t, abi.ChainEpoch(nullHeight), abi.ChainEpoch(nullRoundErr.Epoch))
```
rvagg, line 1749: see lint failure here, one of these is an unnecessary conversion, or you could do `EqualValues` if you want to not bother with conversion

virajbhartiya:
@rvagg the `TestEthFeeHistory` was failing because of now inclusion of null rounds, I have made most of the changes and the refactored `calculateExpectations` but it is still failing and I am unable to figure out why. Also addressed other comments

node/impl/full/eth_utils.go
```diff
@@ -88,7 +88,7 @@ func getTipsetByBlockNumber(ctx context.Context, chain *store.ChainStore, blkPar
 			return nil, fmt.Errorf("cannot get tipset at height: %v", num)
 		}
 		if strict && ts.Height() != abi.ChainEpoch(num) {
-			return nil, ErrNullRound
+			return nil, api.NewErrNullRound(int64(num))
```
rvagg, line 91: how about you just make it take an `abi.ChainEpoch` instead since that's what it would always be anyway

node/impl/full/eth.go
```diff
@@ -1167,13 +1164,9 @@ func (a *EthModule) EthTraceBlock(ctx context.Context, blkNum string) ([]*ethtyp
 }
 
 func (a *EthModule) EthTraceReplayBlockTransactions(ctx context.Context, blkNum string, traceTypes []string) ([]*ethtypes.EthTraceReplayBlockTransaction, error) {
-	if len(traceTypes) != 1 || traceTypes[0] != "trace" {
```
rvagg, line 1170: @virajbhartiya why was this block removed? was this a mistake?

virajbhartiya: Oh right that was a mistake, that wasn't supposed to be removed

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 # UNRELEASED
 
 ## New features
+- Return a consistent error for encountering null rounds in RPC method calls. ([filecoin-project/lotus#12633](https://github.com/filecoin-project/lotus/pull/12633))
```
rvagg, line 6:
```suggestion
- Return a consistent error when encountering null rounds in ETH RPC method calls. ([filecoin-project/lotus#12633](https://github.com/filecoin-project/lotus/pull/12633))
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 # UNRELEASED
 
 ## New features
+- Return a consistent error when encountering null rounds in ETH RPC method calls. ([filecoin-project/lotus#12633](https://github.com/filecoin-project/lotus/pull/12633))
```
rvagg, line 6:
```suggestion
- Return a consistent error when encountering null rounds in ETH RPC method calls. ([filecoin-project/lotus#12655](https://github.com/filecoin-project/lotus/pull/12655))
```

rvagg (approved):
👌 great work

PR #12657, opened 2024-10-29
fix(chainindex): retry transaction if database connection is lost

aarshkshah1992:
Retry transaction if database connection is lost.

### DIFF
--- a/chain/index/helpers.go
+++ b/chain/index/helpers.go
@@ -144,7 +144,7 @@ func withTx(ctx context.Context, db *sql.DB, fn func(*sql.Tx) error) error {
 }
 
 func isRetryableError(err error) bool {
-	return err != nil && strings.Contains(err.Error(), "database is locked")
+	return err != nil && (strings.Contains(err.Error(), "database is locked") || strings.Contains(err.Error(), "statement is closed"))
 }
 
 func isIndexedFlag(b uint8) bool {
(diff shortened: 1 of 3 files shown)


chain/index/helpers.go
```diff
@@ -144,7 +144,7 @@ func withTx(ctx context.Context, db *sql.DB, fn func(*sql.Tx) error) error {
 }
 
 func isRetryableError(err error) bool {
-	return err != nil && strings.Contains(err.Error(), "database is locked")
+	return err != nil && (strings.Contains(err.Error(), "database is locked") || strings.Contains(err.Error(), "statement is closed"))
```
rvagg, line 147: I'm not convinced this is the right approach, why is the statement getting closed? it's not a network connection so shouldn't suffer from network problems. This seems like it's masking a real problem.

aarshkshah1992:
@rvagg Do you mean to say that there should never be a transient conn error with Sqllite ?

rvagg:
I can't imagine why. Unless the docs suggest otherwise? It's basically an open file descriptor?

chain/index/helpers.go
aarshkshah1992: @rvagg I see what you mean. Have reverted this change.

aarshkshah1992:
@rvagg I see what you mean. I have reverted that change. 

I have a strong suspicion based on meandering through the code that the RPC context cancellation from the client side (say a transient RPC connection failure) while a statement is executing is causing this.

I've added some logging to look this up the next time we run into this.

PR #12658, opened 2024-10-30
chore: update GST, Filecoin-FFI and Actors to final versions NV24

rjan90:
## Proposed Changes
- Update GST to https://github.com/filecoin-project/go-state-types/releases/tag/v0.15.0
- Update Filecoin-FFI to https://github.com/filecoin-project/filecoin-ffi/releases/tag/v1.30.0
- Update Builtin-actors to https://github.com/filecoin-project/builtin-actors/releases/tag/v15.0.0

## Checklist

- [ ] CI is green


rvagg (approved):
OK, so the cbor-gen bump propagated here, so it's worth looking at what that gives us: https://github.com/whyrusleeping/cbor-gen/compare/v0.1.2...v0.2.0

* The big thing in that bump was the optional sorting of struct fields, which we don't want and shouldn't touch. I think this was probably a bluesky thing and it shouldn't impact our code paths.
* The thing that does touch us is @Stebalien 's rejection of negative bigints, which _should_ be fine, mainly because we have our own bigint encoding that does these as bytes, so we shouldn't go anywhere near that code (I'm not sure that fix has anything to do with filecoin? maybe it does for f3?).
  - Our bigint byte format: https://github.com/filecoin-project/go-state-types/blob/3524467d26605ec24d4704c7d5d3376598a3d814/big/int.go#L221
  - Used by a custom `MarshalCBOR`: https://github.com/filecoin-project/go-state-types/blob/3524467d26605ec24d4704c7d5d3376598a3d814/big/int.go#L284-L297
* There were a few perf improvements in there.

Overall this should be fine, maybe we should have pushed this earlier through an RC but the risk seems very low to me.

PR #12665, opened 2024-11-01
fix(api): make stick sessions actually work and make them non-racy

Stebalien:
## Related Issues

We apparently have a way to specify that all "related" requests should go to the same node. However:

1. It didn't work at all. All future requests would go to the first successful node from the first request. Because that's how stack variables work.
2. It was racy if the context was re-used concurrently. But only the first time, see point 1.

## Proposed Changes

Use an atomic, get rid of a few of the odd decisions, and write a test.

## Checklist

- [x] CI is green

### DIFF
--- a/cli/util/api.go
+++ b/cli/util/api.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"reflect"
 	"strings"
+	"sync/atomic"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cli/util/api.go
```diff
@@ -231,7 +232,7 @@ type contextKey string
 
 // OnSingleNode is not thread safe
```
magik6k, line 233:
```suggestion
// OnSingleNode is thread safe
```

magik6k (approved):
Reads a little more correctly now

cli/util/api.go
Stebalien: Ok, honestly, that makes me kind of angry.

1. `OnSingleNode` _was_ threadsafe.
2. Using a context derived from it was not.
3. This was horribly dangerous. And by design? REALLY?!?

PR #12667, opened 2024-11-01
fix(f3): when returning a partially used lease, trim its ValidityTerm

Kubuxu:
Otherwise the early renew logic will never trigger.

### DIFF
--- a/chain/lf3/participation_lease.go
+++ b/chain/lf3/participation_lease.go
@@ -104,6 +104,8 @@ func (l *leaser) participate(ticket api.F3ParticipationTicket) (api.F3Participat
 	if found {
 		// short-circuite for reparticipation.
 		if currentLease == newLease {
+			newLease.ValidityTerm = newLease.ToInstance() - instant.ID
+			newLease.FromInstance = instant.ID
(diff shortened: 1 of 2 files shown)


Stebalien (approved):
... I thought I had tested this. How is anything working?

Stebalien:
Oh, I see. The lease will eventually expire. We just won't get the overlap behavior.

chain/lf3/participation_lease.go
```diff
@@ -104,6 +104,8 @@ func (l *leaser) participate(ticket api.F3ParticipationTicket) (api.F3Participat
 	if found {
 		// short-circuite for reparticipation.
 		if currentLease == newLease {
+			newLease.ValidityTerm = newLease.ToInstance() - instant.ID
+			newLease.FromInstance = instant.ID
```
masih, line 108: New lease is not stored in lease map. Should it be?

Stebalien: It shouldn't be stored because the lease in the lease map is the decoded ticket. It's... confusing and we should probably do something about it.

Issue #12636, opened 2024-10-24
MaxFilterResults is applied before filtering for eth logs


rvagg:
Noticed by @ArseniiPetrovich: 

```
$ curl -s -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"fromBlock":"0x42B4AC","toBlock":"0x42B4D4"}],"id":2,"jsonrpc":"2.0"}' http://127.0.0.1:1234/rpc/v1 | jq '.result[].blockNumber' | uniq | wc -l
32
$ curl -s -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"fromBlock":"0x42B4AC","toBlock":"0x42B4D5"}],"id":2,"jsonrpc":"2.0"}' http://127.0.0.1:1234/rpc/v1 | jq '.result[].blockNumber' | uniq | wc -l
32
```

and

```
$ curl -s -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"fromBlock":"0x42B4AC","toBlock":"0x42B4D4"}],"id":2,"jsonrpc":"2.0"}' http://127.0.0.1:1234/rpc/v1 | jq '.result[].blockNumber' | head -1
"0x42b4ac"
ubuntu@ip-192-168-16-104:~$ curl -s -X POST -H "Content-Type: application/json" --data '{"method":"eth_getLogs","params":[{"fromBlock":"0x42B4AC","toBlock":"0x42B4D5"}],"id":2,"jsonrpc":"2.0"}' http://127.0.0.1:1234/rpc/v1 | jq '.result[].blockNumber' | head -1
"0x42b4ae"
```

i.e. extending the range of epochs while leaving the start of the range static drops events from the front of the list.

This is because we apply `MaxFilterResults` max (the invisible cap) before filtering for eth events, not after.

For that same block range:

```
$ curl -s -X POST -H "Content-Type: application/json" --data '{"method":"Filecoin.GetActorEventsRaw","params":[{"fromHeight":4371628,"toHeight":4371668}],"id":2,"jsonrpc":"2.0"}' http://127.0.0.1:1234/rpc/v1 | jq '.result|length'
10000
[...]

PR #12671, opened 2024-11-06
fix: add max results error and default raw codec for eth_getLogs

akaladarshi:
## Related Issues
#12636 

## Proposed Changes
- Adds raw codec while filtering ethereum logs
- Error out if we hit max result 

## Checklist

- [ ] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,58 +83,58 @@ For the set of changes since the last stable release:
 
 ## 👨‍👩‍👧‍👦 Contributors
 
-| Contributor | Commits | Lines ± | Files Changed |
-|-------------|---------|---------|---------------|
-| Krishang | 2 | +34106/-0 | 109 |
-| Rod Vagg | 86 | +10643/-8291 | 456 |
-| Masih H. Derkani | 59 | +7700/-4725 | 298 |
-| Steven Allen | 55 | +6113/-3169 | 272 |
-| kamuik16 | 7 | +4618/-1333 | 285 |
-| Jakub Sztandera | 10 | +3995/-1226 | 94 |
-| Peter Rabbitson | 26 | +2313/-2718 | 275 |
-| Viraj Bhartiya | 5 | +2624/-580 | 50 |
-| Phi | 7 | +1337/-1519 | 257 |
-| Mikers | 1 | +1274/-455 | 23 |
-| Phi-rjan | 29 | +736/-600 | 92 |
-| Andrew Jackson (Ajax) | 3 | +732/-504 | 75 |
-| LexLuthr | 3 | +167/-996 | 8 |
-| Aarsh Shah | 12 | +909/-177 | 47 |
-| web3-bot | 40 | +445/-550 | 68 |
-| Piotr Galar | 6 | +622/-372 | 15 |
-| aarshkshah1992 | 18 | +544/-299 | 40 |
-| Steve Loeppky | 14 | +401/-196 | 22 |
-| Frrist | 1 | +403/-22 | 5 |
-| Łukasz Magiera | 4 | +266/-27 | 13 |
-| winniehere | 1 | +146/-144 | 3 |
-| Jon | 1 | +209/-41 | 4 |
-| Aryan Tikarya | 2 | +183/-8 | 7 |
-| adlrocha | 2 | +123/-38 | 21 |
-| dependabot[bot] | 11 | +87/-61 | 22 |
-| Jiaying Wang | 8 | +61/-70 | 12 |
-| Ian Davis | 2 | +60/-38 | 5 |
-| Aayush Rajasekaran | 2 | +81/-3 | 3 |
-| hanabi1224 | 4 | +46/-4 | 5 |
-| Laurent Senta | 1 | +44/-1 | 2 |
-| jennijuju | 6 | +21/-20 | 17 |
-| parthshah1 | 1 | +23/-13 | 1 |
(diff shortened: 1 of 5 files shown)


chain/index/events.go
```diff
@@ -563,6 +566,9 @@ func makePrefillFilterQuery(f *EventFilter) ([]any, string, error) {
 				clauses = append(clauses, "("+strings.Join(subclauses, " OR ")+")")
 			}
 		}
+	} else {
+		clauses = append(clauses, "ee.codec=?")
+		values = append(values, uint64(multicodec.Raw))
```
rvagg, line 571: This won't work because we call this method from both the eth APIs and also `GetActorEventsRaw` and `SubscribeActorEventsRaw` and these APIs want any codec if they don't specify a `KeysWithCodec`.

I think what you might need to do is extend `EventFilter` to add a `OnlyRaw`, or `WithCodec` or something like that so that when this gets called from the Eth APIs, that can be set so we can restrict to Raw only.

rvagg: yeah, for now just `Codec` is fine, "only used if KeysWithCodec is not set"

CHANGELOG.md
```diff
@@ -27,7 +27,7 @@
 - Try harder in the F3 participation loop to participate using the same lotus node ([filecoin-project/lotus#12664](https://github.com/filecoin-project/lotus/pull/12664)).
 - The mining loop will now correctly "stick" to the same upstream lotus node for all operations pertaining to mining a single block ([filecoin-project/lotus#12665](https://github.com/filecoin-project/lotus/pull/12665)).
 - Make the ordering of event output for `eth_` APIs and `GetActorEventsRaw` consistent, sorting ascending on: epoch, message index, event index and original event entry order. ([filecoin-project/lotus#12623](https://github.com/filecoin-project/lotus/pull/12623))
-
+- Return error after hitting max results in event filtering and add default `raw` codec for eth_getLogs. ([filecoin-project/lotus#12671](https://github.com/filecoin-project/lotus/pull/12671))
```
rvagg, line 30:
```suggestion

- Return error after hitting max results in event filtering and add default `raw` codec for eth_getLogs. ([filecoin-project/lotus#12671](https://github.com/filecoin-project/lotus/pull/12671))
```

chain/index/events.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 )
 
+var ErrMaxResultsReached = xerrors.New("max results limit reached, results truncated")
```
rvagg, line 25:
```suggestion
var ErrMaxResultsReached = fmt.Errorf("filter matches too many events, try a more restricted filter")
```

node/impl/full/eth.go
```diff
@@ -1768,12 +1768,19 @@ func (e *EthEventHandler) ethGetEventsForFilter(ctx context.Context, filterSpec
...
+
 	ef := &index.EventFilter{
 		MinHeight:     pf.minHeight,
 		MaxHeight:     pf.maxHeight,
 		TipsetCid:     pf.tipsetCid,
 		Addresses:     pf.addresses,
 		KeysWithCodec: pf.keys,
+		Codec:         codec,
 		MaxResults:    e.EventFilterManager.MaxFilterResults,
 	}
```
rvagg, line 1785:
```suggestion
	ef := &index.EventFilter{
		MinHeight:     pf.minHeight,
		MaxHeight:     pf.maxHeight,
		TipsetCid:     pf.tipsetCid,
		Addresses:     pf.addresses,
		KeysWithCodec: pf.keys,
		Codec:         multicodec.Raw,
		MaxResults:    e.EventFilterManager.MaxFilterResults,
	}
```

you can avoid the conditional, it's always `Raw` regardless. Look in `keysToKeysWithCodec` and you'll see even there if they supply a key then it's going to be set to `Raw` anyway.

chain/index/interface.go
```diff
@@ -47,6 +47,8 @@ type EventFilter struct {
 
 	KeysWithCodec map[string][]types.ActorEventBlock // map of key names to a list of alternate values that may match
 	MaxResults    int                                // maximum number of results to collect, 0 is unlimited
+
+	Codec uint64 // optional codec filter, only used if KeysWithCodec is not set
```
rvagg, line 51:
```suggestion
	Codec multicodec.Code // optional codec filter, only used if KeysWithCodec is not set
```

akaladarshi: Done.

chain/index/events.go
akaladarshi: Done.

CHANGELOG.md
akaladarshi: Done.

CHANGELOG.md
```diff
@@ -78,58 +78,58 @@ For the set of changes since the last stable release:
 
 ## 👨‍👩‍👧‍👦 Contributors
 
-| Contributor | Commits | Lines ± | Files Changed |
```
rvagg, line 81: @akaladarshi you need to fix your editor to stop making automatic changes like this, could you please reset the changelog to just make the one line change? otherwise this looks good to go!

rvagg (changes requested):

CHANGELOG.md
akaladarshi: I have fixed it.

rvagg (approved):
nice!

PR #12675, opened 2024-11-06
fix(eth): present revert "data" as plain bytes

rvagg:
decode the cbor return value for reverts and present that, as is expected by Ethereum tooling


rvagg:
https://github.com/filecoin-project/lotus/pull/12553 has details of my investigation into this, I'll also be following up with PRs to fevm-uniswap-tests and fevm-contract-tests to match this.

aarshkshah1992 (approved):

PR #12680, opened 2024-11-07
fix: getMsgInfo returns an ErrNotFound when there are no rows

### LINKED ISSUE #12674: ChainIndexer: GetMsgInfo should return ErrNotFound when not found
```
2024-11-06T21:13:14.735+1100    WARN    statemgr        stmgr/searchwait.go:175 error searching message index: error looking up message in index: sql: no rows in result set
```

Lots of this kind of thing in the logs, but it shouldn't be logged: https://github.com/filecoin-project/lotus/blob/8865a9d0931bbb6167944b5a4cd6867290aabea2/chain/stmgr/searchwait.go#L171-L176

`GetMsgInfo` should return an `ErrNotFound` when there are no rows, here: https://github.com/filecoin-project/lotus/blob/8865a9d0931bbb6167944b5a4cd6867290aabea2/chain/index/read.go#L51-L54

Needs tests as well.

virajbhartiya:
## Related Issues
Closes #12674

## Proposed Changes
GetMsgInfo returns an ErrNotFound when there are no rows.

## Checklist

- [ ] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Implement `EthGetTransactionByBlockNumberAndIndex` (`eth_getTransactionByBlockNumberAndIndex`) and `EthGetTransactionByBlockHashAndIndex` (`eth_getTransactionByBlockHashAndIndex`) methods. ([filecoin-project/lotus#12618](https://github.com/filecoin-project/lotus/pull/12618))
 
 ## Bug Fixes
+- GetMsgInfo returns an ErrNotFound when there are no rows ([filecoin-project/lotus#12680](https://github.com/filecoin-project/lotus/pull/12680))
 - Fix a bug in the `lotus-shed indexes backfill-events` command that may result in either duplicate events being backfilled where there are existing events (such an operation *should* be idempotent) or events erroneously having duplicate `logIndex` values when queried via ETH APIs. ([filecoin-project/lotus#12567](https://github.com/filecoin-project/lotus/pull/12567))
 - Event APIs (Eth events and actor events) should only return reverted events if client queries by specific block hash / tipset. Eth and actor event subscription APIs should always return reverted events to enable accurate observation of real-time changes. ([filecoin-project/lotus#12585](https://github.com/filecoin-project/lotus/pull/12585))
 - Add logic to check if the miner's owner address is delegated (f4 address). If it is delegated, the `lotus-shed sectors termination-estimate` command now sends the termination state call using the worker ID. This fix resolves the issue where termination-estimate did not function correctly for miners with delegated owner addresses. ([filecoin-project/lotus#12569](https://github.com/filecoin-project/lotus/pull/12569))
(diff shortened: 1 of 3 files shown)


chain/index/read_test.go
```diff
@@ -57,23 +57,74 @@ func TestGetMsgInfo(t *testing.T) {
...
+		for {
+			mi, err = s.GetMsgInfo(ctx, msgCid)
+			if err == nil {
+				break
+			}
+			if ctx.Err() != nil {
+				t.Fatalf("timed out waiting for message to appear: %v", err)
+			}
+			time.Sleep(50 * time.Millisecond)
+		}
```
rvagg, line 113: see `require.Eventually()` to do this with a bit more robustness

although, I'm not sure this test case adds much, what unique coverage are we adding with this?

virajbhartiya: I initially thought this might be needed if the message appeared after a delay which is I had added not, not sure if needed or not

virajbhartiya:
@rvagg I have done a string comparison to check for the error and return `ErrNotFound` Please let me know if that's the way to go about it or there's a better way for this

chain/index/read.go
```diff
@@ -49,6 +50,9 @@ func (si *SqliteIndexer) GetMsgInfo(ctx context.Context, messageCid cid.Cid) (*M
 	var height int64
 
 	if err := si.queryMsgInfo(ctx, messageCid, &tipsetKeyCidBytes, &height); err != nil {
+		if strings.Contains(err.Error(), "no rows in result set") {
```
rvagg, line 53: I think `if err == sql.ErrNoRows {` should have been OK for this, no? You had it in your original version, was it not working? `errors.Is(err, sql.ErrNoRows)` if we're afraid about it being wrapped.

virajbhartiya: It wasn't working as intended while testing out `require.Eventually` test case, which is why I changed it, I'll revert it and if everything works as needed I'll push the changes

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 - Make `EthTraceFilter` / `trace_filter` skip null rounds instead of erroring. ([filecoin-project/lotus#12702](https://github.com/filecoin-project/lotus/pull/12702))
 - Event APIs (`GetActorEventsRaw`, `SubscribeActorEventsRaw`, `eth_getLogs`, `eth_newFilter`, etc.) will now return an error when a request matches more than `MaxFilterResults` (default: 10,000) rather than silently truncating the results. Also apply an internal event matcher for `eth_getLogs` (etc.) to avoid builtin actor events on database query so as not to include them in `MaxFilterResults` calculation. ([filecoin-project/lotus#12671](https://github.com/filecoin-project/lotus/pull/12671))
+- GetMsgInfo returns an ErrNotFound when there are no rows ([filecoin-project/lotus#12680](https://github.com/filecoin-project/lotus/pull/12680))
```
rvagg, line 10:
```suggestion
- `ChainIndexer#GetMsgInfo` returns an `ErrNotFound` when there are no rows. ([filecoin-project/lotus#12680](https://github.com/filecoin-project/lotus/pull/12680))
```

PR #12681, opened 2024-11-07
feat(docs): add workflow to generate documentation on PR merge

### LINKED ISSUE #12233: Run `make docsgen-cli` on merge
### Done Criteria
There is a mechanism to ensure that `make docsgen-cli` is run automatically on merge to the master branch.

### Why Important
The `docsgen-cli` command auto-generates CLI documentation. In instances where contributors push a single line commit during a review, it might cause the `docsgen-cli` check to fail. This requires the person writing the PR to re-run make docsgen-cli, often context switching and waiting XX-minutes (TODO: write down time it takes) for the command to finish.

Additionally, most new contributors to Lotus are not aware of the need to run this command and running this automatically on merge can help alleviate some chores needed to do when contributing to the codebase.

### User/Customer
Lotus contributors and maintainers.

### Notes

virajbhartiya:
## Related Issues
Closes #12233 

## Proposed Changes
Add a workflow that run `make docsgen-cli` whenever the PR is merged

## Checklist

- [ ] CI is green


.github/workflows/docsgen-cli.yml
```diff
@@ -0,0 +1,31 @@
+name: Generate Documentation
+
```
rvagg, line 2:
```suggestion

permissions:
  contents: write

```

I don't recall what our org settings are but this may be needed to be able to push changes from an action; doesn't hurt to have it either way

.github/workflows/check.yml
```diff
@@ -26,21 +26,19 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          submodules: 'recursive'
+          submodules: "recursive"
```
rvagg, line 29: you should check your editor and remove these auto-fix rules that you have, it's generally not good manners to go editing things unrelated to what you're intending to change without good reason to do so

rvagg:
@virajbhartiya can you resolve the conflict with `master` here please? You'll see that your `"` substitutions are getting in the way, so resolving those would be good.

virajbhartiya:
Hey @rvagg, I have resolved the conflicts in the PR

virajbhartiya:
@rvagg any suggestions on why is the workflow failing?

rvagg:
Failing because you need to install system dependencies for the build

in check.yml, you'll see:

```yml
      - uses: ./.github/actions/install-system-dependencies
      - uses: ./.github/actions/install-go
      - uses: ./.github/actions/make-deps
```

these are required to perform the actions you need, so you'll have to do it again in your new job

but, having said that, it's not cheap, so maybe this should all be in a single job so that's all done once and it's just the docsgen-cli bit that's done separately.

So how about this:

1. rename check.yml to check-and-gen.yml
2. add a `docsgen-cli` section to it with your new stuff, so that happens in the same flow

I'm still not convinced this is a great idea, it might get in the way. So we'll need to have a collective chat with others before we proceed.

virajbhartiya:
Hey @rvagg thanks for the review, I have currently modified it accordingly, renamed check.yml to check-and-gen.yml and added it in the same file. We can discuss regarding this on slack wtih other how to go ahead with this PR

.github/workflows/check.yml
virajbhartiya: For some reason, my linter was automatically replacing the quotes, fixed it

.github/workflows/check-and-gen.yml
```diff
@@ -70,3 +68,37 @@ jobs:
...
+          make docsgen-cli || {
+            echo "Error: Documentation generation failed"
+            exit 1
+          }
+
+      - name: Commit and push if documentation changed
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
+            git config user.name "github-actions[bot]"
+            git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
BigLep, line 98: I have no expertise here, but maybe add a comment on where this comes from?  

Is https://github.com/orgs/community/discussions/26560 the canonical link?

.github/workflows/check-and-gen.yml
```diff
@@ -70,3 +68,37 @@ jobs:
...
+            git add .
+            git commit -m "docs: update API documentation via docsgen-cli"
```
BigLep, line 100: If we're adding a commit, I think it would be good to tie what PR this commit is asociated with. 
(We haven't developed team standards yet on whether commits should have a PR reference, but I think this is coming and we may as well do it here.)
Maybe @rvagg has input?
(This isn't blocking though.)

.github/workflows/check-and-gen.yml
virajbhartiya: Yeah that's correct, I'll add a comment pointing to the GitHub discussion for future references

.github/workflows/check-and-gen.yml
BigLep: This can be disregarded.  I understand that this commit will be added before the PR is merged, and since  most PRs will be squashed, this goes away.  I'll resolve this.

BigLep:
I know there have been verbal conversation on this one.  Did we write out the next steps?

rvagg:
Didn't write it out, but next steps are to get this completed and merged and then experience it live and decide whether we hate it or love it and whether it's a QoL improvement on balance.

.github/workflows/check-and-gen.yml
```diff
@@ -70,3 +68,38 @@ jobs:
       - uses: ./.github/actions/install-go
       - run: go mod tidy -v
       - run: git diff --exit-code
+  generate-docs:
+    name: Generate Documentation
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release/'))
```
rvagg, line 74: I'm unsure about this and whether a merge into one of these branches would trigger the job as well, we only want it to run on the PR branch, strictly. So this line worries me a little and it needs verification, somehow, or we need a subject-matter expert to help.

.github/workflows/check-and-gen.yml
```diff
@@ -70,3 +68,38 @@ jobs:
...
+
+      - name: Commit and push if documentation changed
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
+           # Bot email is configured from: https://github.com/orgs/community/discussions/26560
+            git config user.name "github-actions[bot]"
+            git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+            git add .
+            git commit -m "docs: update API documentation via docsgen-cli"
+            git push
```
rvagg, line 102: Another thing I'm wondering is if this is going to work to fork-branches, if this is run a a filecoin-project GHA runner, will we always have permissions to push back into the PR branch? Do we need to ensure we have a "can edit branches" thing turned on in our repo? Do we need to do anything else to make this work?

.github/workflows/check-and-gen.yml
```diff
@@ -33,8 +33,6 @@ jobs:
       - uses: ./.github/actions/make-deps
       - run: make gen
```
rvagg, line 34: Probably too big a leap for now, but I'd actually like to move `docsgen` out of `make gen` so we can run it here. Maybe for now we just do docsgen-cli and if it works OK we'll consider expanding it.

.github/workflows/check-and-gen.yml
rvagg: @BigLep maybe something to pull IPDX into to help review.

BigLep: @galargh : do you know offhand whether this will only run on pushes to PR branches and not on master or release branches?

galargh: That's the opposite of what this conditional says.

The conditional is: accept a workflow run which was triggered by a push event to either `master` or `release/*` branch.

.github/workflows/check-and-gen.yml
galargh: As currently configured, this will be run on push events to `master` or `release/*`. But it won't work because they are protected. To overcome this, one would have to use a personal access token for a user that can disregard the branch protection rules.

galargh (commented):
FYI, https://github.com/ipdxco/github-as-code/blob/master/.github/workflows/fix.yml is a working example of a workflow that runs on PRs and writes to PR branches.

BigLep:
@rvagg : do you think we can do the next steps in https://github.com/filecoin-project/lotus/pull/12681#issuecomment-2499426438 this week to run the experiment?  (Or have new thoughts come in and we defer for now.)

BigLep:
@virajbhartiya : is the final feedback something you'll be able to incorporate or does someone else need to take this?

.github/workflows/check-and-gen.yml
virajbhartiya, line 74: @BigLep So as for the changes what we need is to run on `pull_request` instead of the current condition so it runs only on PR branches or am I missing something?

BigLep: @virajbhartiya : that's right.  We want this to run on the PR branches only.

rvagg:
I hope you don't mind @virajbhartiya but I've squashed this down to one commit, rebased on master, added a little fix to make sure this is only done on PRs and have force pushed all of that to your branch

rvagg:
https://github.com/filecoin-project/lotus/pull/12854

rvagg:
OK, I'm pushing my latest attempt at this and giving up for now, it's driving me mad.

* https://github.com/filecoin-project/lotus/pull/12856 is a PR testing it out, and that's just from _this_ repo, I gave up on testing it from a fork but that would need to be done too.
* It makes a merge commit, I think because actions/checkout does a detached head but I thought `ref:` was supposed to deal with that, argh!
* The push after commit wipes out all of the "Checks" so to find the actions runs you have to hunt through the commits and look for the ❌ or ✅ and get to it via that; this isn't OK and would be cause enough to bail on this whole project if this is what you have to put up with if an action pushes to a branch that it's checking.
* I really wanted to see a comment on the PR come out of it but didn't get past all the other frustration to make my commenter work, but that's probably not too hard, see error in https://github.com/filecoin-project/lotus/actions/runs/13024797399/job/36331973937?pr=12856

As of now, @virajbhartiya's `docsgen-workflow` branch has my latest. If this doesn't get picked up again soon I think we should just give up. `docsgen-cli` is much faster now than it once was and it may not be a bad thing to even bundle it with `docsgen`. Asking the user to do it is less onerous than it was when we started this.

BigLep:
@galargh : if you are up for timeboxing 1 hour to land this, that's great.   I wouldn't want to spend more than an hour here though as we are ok to close it too.

galargh:
> @galargh : if you are up for timeboxing 1 hour to land this, that's great. I wouldn't want to spend more than an hour here though as we are ok to close it too.

This is a minimal example of a workflow committing to pull requests (including public, non-org forks where the user allowed maintainer edits) - https://github.com/galorgh/.github/blob/master/.github/workflows/playground3.yml

It requires the use of `pull_request_target` to obtain credentials that are allowed writing to forks. One should be very careful when operating the the pull request target context because it has access to secrets and powerful credentials.

The pushed commit doesn't trigger any more workflows by design. This means it is not suitable for repos with required checks.

The way to overcome the limitations of this approach is to use custom credentials - either bot account's PAT or designated GItHub App's secrets.

rvagg:
> The pushed commit doesn't trigger any more workflows by design. This means it is not suitable for repos with required checks.

@galargh the problem I was having was that the pushed commit would make the previous workflows disappear in the original PR, making it pretty useless. It makes sense, because the checks were against the previous head but we just updated the head, but it's not the effect I was after.

Currently my vote is on bailing on this approach and just making the user deal with it. `make gen` has become faster to run, but I believe it still means you have to install dependencies and jump through all those hoops, which is unfortunate.

BigLep:
> Currently my vote is on bailing on this approach and just making the user deal with it. `make gen` has become faster to run, but I believe it still means you have to install dependencies and jump through all those hoops, which is unfortunate.

Ack.  I'm going to close this for now then.  This work was spawned from https://github.com/filecoin-project/lotus/issues/12233 which came from an operational event and this was identified as a way to potentially move quicker.  Per above, given the command has become faster, we at least moved in a better direction.

PR #12691, opened 2024-11-12
feat(cli): enhance ChainListCmd with JSON output and <epoch> placeholder support

Kubuxu:
This PR updates the `ChainListCmd` with:
 - JSON output via `<json_tipset>` for compact tipset representation.
 - Support for `<epoch>` as a placeholder.
 - Streamlined format flag usage documentation.


rvagg (approved):
nice, although just a `--json` that's mutually exclusive with `--format` might make this eaiser

Kubuxu:
I would expect `--json` to probably return block headers, which is not what I wanted here. I needed a nice way to resolve epoch to JSON tipset in API format to call `louts-shed rpc` with it.

PR #12699, opened 2024-11-18
fix(test): unflake TestEthBlockNumberAliases

rvagg:
Noticed as flaky: https://github.com/filecoin-project/lotus/actions/runs/11851664737/job/33028556808 - that one's likely to do with null round accounting (lack thereof) but there's also the stable `head` problem that's not accounted for with the `"safe"` and `"finalized"` forms in this test.

### DIFF
--- a/itests/eth_api_test.go
+++ b/itests/eth_api_test.go
@@ -20,7 +20,6 @@ import (
 	inittypes "github.com/filecoin-project/go-state-types/builtin/v8/init"
 	"github.com/filecoin-project/go-state-types/exitcode"
 
-	"github.com/filecoin-project/lotus/build"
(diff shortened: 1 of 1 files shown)


masih (approved):
Looks good. I would set up a temporary CI job to run this test repeatedly just to make sure it is indeed unflaked.

Kubuxu (approved):

rvagg:
I've re-run the test in CI manually >5 times rather than futz around with making a special CI job for this.

masih:
> futz around with making a special CI job for this

In case you'd find it useful: https://gist.github.com/masih/993149988f1cd3c2cb13ee46e86d8a5c

PR #12701, opened 2024-11-19
docs(indexer): more clarity for ChainIndexer configuration setup

rvagg:
Feedback already suggesting that our lack of automagic config switching is leading to some confusion. I'd still rather avoid automatically turning on ChainIndexer in these cases so we'll try with more clear error messages and see if we continue to get reports.

### DIFF
--- a/documentation/en/default-lotus-config.toml
+++ b/documentation/en/default-lotus-config.toml
@@ -223,6 +223,7 @@
 [Fevm]
   # EnableEthRPC enables eth_ RPC methods.
   # Note: Setting this to true will also require that ChainIndexer is enabled, otherwise it will cause an error at startup.
+  # Set EnableIndexer in the ChainIndexer section of the config to true to enable the ChainIndexer.
   #
   # type: bool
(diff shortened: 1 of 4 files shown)


virajbhartiya (commented):
I am not sure but will it be a good idea to add a commented-out config for chainidnexer the way `EnableEthRPC = true` is set to true but commented in the config file by default so that it is easier for a new user.

rjan90 (approved):

rjan90:
Needs a `make gen/make docsgen` it seems, but apart from that, lgtm.

aarshkshah1992 (approved):

rvagg:
> commented-out config for chainidnexer

unfortunately the default config is generated from settings automatically, and their values are meant to reflect what they are currently set to - so they are a kind of documentation for the user, demonstrating what the current settings are, so they have to change them if they want them.

rvagg:
ahh, actually needs a `docsgen-cli` because of the config file output, `gen` and `docsgen` don't cut it (I ran them already!)

PR #12703, opened 2024-11-19
feat(shed): lotus-shed sectors dump-sectors

rvagg:
Most of the fields in SectorOnChainInfo for _all_ of the records on chain (using the power actor to list miners and then listing all of their sectors). No filtering so far, output as CSV. Currently doesn't print the two CIDs. This came up as part of the FIP-0096 (or whatever number this will get https://github.com/filecoin-project/FIPs/pull/1079) discussions looking at termination fees and I wanted to collect some data for it.

### DIFF
--- a/cmd/lotus-shed/state-stats.go
+++ b/cmd/lotus-shed/state-stats.go
@@ -5,8 +5,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
+	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"sync"
@@ -29,6 +30,8 @@ import (
 
 	"github.com/filecoin-project/lotus/api"
 	"github.com/filecoin-project/lotus/blockstore"
+	badgerbs "github.com/filecoin-project/lotus/blockstore/badger"
+	"github.com/filecoin-project/lotus/blockstore/splitstore"
 	"github.com/filecoin-project/lotus/chain/actors"
 	"github.com/filecoin-project/lotus/chain/actors/builtin"
 	"github.com/filecoin-project/lotus/chain/consensus"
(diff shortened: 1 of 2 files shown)


cmd/lotus-shed/state-stats.go
```diff
@@ -226,41 +229,70 @@ func loadChainStore(ctx context.Context, repoPath string) (*StoreHandle, error)
...
 	}
 
-	bs, err := lr.Blockstore(ctx, repo.UniversalBlockstore)
+	cold, err := lkrepo.Blockstore(ctx, repo.UniversalBlockstore)
 	if err != nil {
-		return nil, fmt.Errorf("failed to open blockstore: %w", err)
+		return nil, xerrors.Errorf("failed to open universal blockstore %w", err)
 	}
 
+	path, err := lkrepo.SplitstorePath()
```
ZenGround0, line 242: What is going on here ? What was breaking that made you change this?

ZenGround0: When you open universal blockstore it should be wrapped around all of the splitstore right?

ZenGround0 (commented):
The main shed command looks fine but Id like to understand why you are changing 
 the loadChainStore function to explicitly assume splitstore things.

cmd/lotus-shed/state-stats.go
rvagg: @ZenGround0: when using the original I couldn't get it to run - when it initialised and figured out what the latest head is (from the metadata store) it would then try and load that tipset but fail with not-found errors. Switching to this makes it work. My assumption is that this works fine for a non-splitstore setup, but breaks when using a splitstore, and nobody running splitstore has been using the commands that use this code until I tried it.

Both patterns are found throughout lotus-shed, I copied a lot of this new stuff from elsewhere. I'd like to understand it a bit better to make it consistent throughout, maybe just make a helper function for this. But I haven't taken the time to bother going that deep yet. If you have more of a clue then I'm all ears!

ZenGround0: The reason I'm concerned is that I am almost certain this used to work with a splitstore setup.  I was running state stat stuff back in 2023 when splitstore hotstore started working again.  Not found errors on current state lookup are also unexpected.  I guess you can merge this since these are only shed commands but it is a scary feeling to accept that the universal blockstore just doesn't work.

ZenGround0 (approved):

cmd/lotus-shed/state-stats.go
rvagg: ok, this is good to know, worth investigating

Issue #12706, opened 2024-11-19
Use urfav Command struct directly to generate docsgen-cli


masih:
The docsgen-cli builds binaries first, then executes them to get the usage, recursively for all sub commands. The process is less than optimal:
* logic to parse the usage is error prone and makes implicit assumptions (e.g. https://github.com/filecoin-project/lotus/issues/12705)
* The `make docsgen-cli` target is slower than we want (even though it was made about 5X faster by porting over to go), and the majority of time is spent on building binaries.

We already have Go structs for commands. We should be using those (along with urfav's built-in helper generators) to generate the docs. This way the docs generation would be _far_ faster and significantly less error prone.

To do this however, we should refactor the `cli` package to make some commands exported so that they are accessible from the docsgen-cli script.

PR #12717, opened 2024-11-25
feat(docs): generate cli docs directly from the struct

akaladarshi:
## Related Issues
Fixes: #12706


CHANGELOG.md
```diff
@@ -89,58 +90,58 @@ For the set of changes since the last stable release:
 
 ## 👨‍👩‍👧‍👦 Contributors
 
-| Contributor | Commits | Lines ± | Files Changed |
```
rvagg, line 92: too much changelog auto-formatting again, this needs to be undone

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ## Improvements
 
 - The miner actor builtin `QAPowerForWeight` no longer accepts the unused "dealWeight" parameter, the function signature now only takes 3 arguments: sectorSize, sectorDuration, verifiedWeight. ([filecoin-project/lotus#12445](https://github.com/filecoin-project/lotus/pull/12445))
+- Generate the cli docs directly from the structs instead using binaries. ([filecoin-project/lotus#12717](https://github.com/filecoin-project/lotus/pull/12717))
```
rvagg, line 9:
```suggestion
- Generate the cli docs directly from the code instead compiling and executing binaries' `help` output. ([filecoin-project/lotus#12717](https://github.com/filecoin-project/lotus/pull/12717))
```

cli/miner/actor_test.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 func TestWorkerKeyChange(t *testing.T) {
-	//stm: @OTHER_WORKER_KEY_CHANGE_001
+	// stm: @OTHER_WORKER_KEY_CHANGE_001
```
rvagg, line 28: Undo change, probably a code editor auto-format. I think these `stm:` commands needed to be carefully formatted. One day we'll probably just remove them all but for now let's leave them alone.

scripts/docsgen-cli/doc_generator.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// Generate generates documentation for the CLI app
+func (g *DocGenerator) Generate(name string) error {
+	mdPath := filepath.Join(g.outputDir, fmt.Sprintf("cli-%s.md", name))
+	file, err := os.Create(mdPath)
+	if err != nil {
+		return fmt.Errorf("failed to create markdown file: %w", err)
+	}
+	defer file.Close()
```
rvagg, line 35: note lint

documentation/en/default-lotus-config.toml
```diff
@@ -1,380 +0,0 @@
-[API]
```
rvagg, line 1: you're going to need to replicate this too, but it should be pretty easy, the `lotus config default` is fairly simple: https://github.com/filecoin-project/lotus/blob/07f2f6990b55b4a4b93ee94609172c85335e92fd/cmd/lotus/config.go#L32-L40

documentation/en/cli-lotus-miner.md
```diff
@@ -10,65 +11,69 @@ VERSION:
...
+   help, h                        Shows a list of commands or help for one command
 
 GLOBAL OPTIONS:
-   --actor value, -a value                  specify other actor to query / manipulate
-   --color                                  use color in display output (default: depends on output being a TTY)
-   --miner-repo value, --storagerepo value  Specify miner repo path. flag(storagerepo) and env(LOTUS_STORAGE_PATH) are DEPRECATION, will REMOVE SOON (default: "~/.lotusminer") [$LOTUS_MINER_PATH, $LOTUS_STORAGE_PATH]
-   --vv                                     enables very verbose mode, useful for debugging the CLI (default: false)
-   --help, -h                               show help
-   --version, -v                            print the version
+   --actor, -a                    specify other actor to query / manipulate (default: "")
```
rvagg, line 34: 3. the option parameters are missing here, and they're important, `--actor value` is the only way to use `--actor`; which is different to `--color` which doesn't take a parameter.

documentation/en/cli-lotus-miner.md
```diff
@@ -10,65 +11,69 @@ VERSION:
...
+   --pre-sealed-metadata          specify the metadata file for the presealed sectors (default: "")
+   --nosync                       don't check full-node sync status (default: "false")
+   --symlink-imported-sectors     attempt to symlink to presealed sectors instead of copying them into place (default: "false")
+   --no-local-storage             don't use storageminer repo for sector storage (default: "false")
+   --gas-premium                  set gas premium for initialization messages in AttoFIL (default: "0")
+   --from                         select which address to send actor creation message from (default: "")
+   --confidence                   --confidence value	number of block confirmations to wait for (default: 0)
+   --help, -h                     show help (default: "false")
+
+COMMANDS:
```
rvagg, line 70: 4. `COMMANDS:` is below the `OPTIONS:` here

documentation/en/cli-lotus-miner.md
```diff
@@ -92,26 +99,27 @@ USAGE:
...
+   --miner-api                    2345 (default: "")
+   --enable-gpu-proving           enable use of GPU for mining operations (default: "true")
+   --nosync                       don't check full-node sync status (default: "false")
+   --manage-fdlimit               manage open file limit (default: "true")
+   --help, -h                     show help (default: "false")
+
 ```
 
 ## lotus-miner stop
+
```
rvagg, line 111: fwiw these spaces are fine, I'm not a fan of squishing everything up so these are good by me

documentation/en/cli-lotus-miner.md
```diff
@@ -77,13 +82,15 @@ USAGE:
    lotus-miner init restore [command options] [backupFile]
 
 OPTIONS:
-   --nosync                don't check full-node sync status (default: false)
-   --config value          config file (config.toml)
-   --storage-config value  storage paths config (storage.json)
-   --help, -h              show help
+   --nosync                       don't check full-node sync status (default: "false")
```
rvagg, line 85: 5. `"false"` vs `false` - figure out what the quoting rule is, could be strings only?

documentation/en/cli-lotus-miner.md
```diff
@@ -77,13 +82,15 @@ USAGE:
...
+   --config                       config file (config.toml) (default: "")
```
rvagg, line 86: 6. don't show default where one isn't set - you'll need to figure out the rule for that (is it the zero value? then how come `false` shows up sometimes but not others? :shrug:

documentation/en/cli-lotus-miner.md
```diff
@@ -92,26 +99,27 @@ USAGE:
...
 ## lotus-miner stop
+
 ```
 NAME:
    lotus-miner stop - Stop a running lotus miner
 
 USAGE:
    lotus-miner stop [command options] [arguments...]
 
-OPTIONS:
```
rvagg, line 110: 7. probably should add an implicit `--help` to everything.

documentation/en/cli-lotus-miner.md
```diff
@@ -165,173 +175,19 @@ USAGE:
...
+this command must be within this base path
 
 OPTIONS:
-   --system value [ --system value ]  limit to log system
-   --help, -h                         show help
-```
+   --offline                      create backup without the node running (default: "false")
+   --help, -h                     show help (default: "false")
 
-### lotus-miner log alerts
```
rvagg, line 284: why did this one come out of order? do you know why urfave puts it here but yours doesn't?

rvagg:
Looks like a lot of the logic for this lives in https://github.com/urfave/cli/blob/main/help.go and the templates in here (unfortunately not exported!) are here: https://github.com/urfave/cli/blob/main/template.go

Use this as an exercise in minimising the diff, that's your challenge - how can you get to near zero red/green lines in the diff here for the documentation output, aiming to only make changes where the existing script is getting it _wrong_ (mostly missing some things I believe).

BigLep:
@akaladarshi : I assume you'll re-request review when you're ready for it to be looked at again.

cli/miner/actor_test.go
akaladarshi: Done

documentation/en/cli-lotus-miner.md
```diff
@@ -10,65 +11,69 @@ VERSION:
...
-   help, h       Shows a list of commands or help for one command
-   CHAIN:
-     actor  manipulate the miner actor
-     info   Print miner info
-   STORAGE:
-     sectors  interact with sector store
-     proving  View proving information
-     storage  manage sector storage
-     sealing  interact with sealing pipeline
+   init                           Initialize a lotus miner repo
```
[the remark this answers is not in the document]
akaladarshi: Done.

documentation/en/cli-lotus-miner.md
akaladarshi: Fixed.

documentation/en/cli-lotus-miner.md
akaladarshi: Fixed.

documentation/en/cli-lotus-miner.md
akaladarshi: Done.

documentation/en/cli-lotus-miner.md
akaladarshi: Done.

documentation/en/cli-lotus-miner.md
```diff
@@ -165,173 +175,19 @@ USAGE:
...
-      warn
-      error
-
-      Environment Variables:
-      GOLOG_LOG_LEVEL - Default log level for all log systems
-      GOLOG_LOG_FMT   - Change output log format (json, nocolor)
-      GOLOG_FILE      - Write logs to file
-      GOLOG_OUTPUT    - Specify whether to output to file, stderr, stdout or a combination, i.e. file+stderr
-
+Online backups:
```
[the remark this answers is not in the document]
akaladarshi: Done.

documentation/en/cli-lotus-miner.md
```diff
@@ -31,15 +33,19 @@ COMMANDS:
...
-   --miner-repo value, --storagerepo value  Specify miner repo path. flag(storagerepo) and env(LOTUS_STORAGE_PATH) are DEPRECATION, will REMOVE SOON (default: "~/.lotusminer") [$LOTUS_MINER_PATH, $LOTUS_STORAGE_PATH]
-   --vv                                     enables very verbose mode, useful for debugging the CLI (default: false)
-   --help, -h                               show help
-   --version, -v                            print the version
+   --actor value, -a              specify other actor to query / manipulate
+   --color                        use color in display output (default: depends on output being a TTY)
+   --panic-reports value           (default: "~/.lotusminer") [LOTUS_PANIC_REPORT_PATH]
+   --repo value                    (default: "~/.lotus") [LOTUS_PATH]
+   --miner-repo value, --storagerepo Specify miner repo path. flag(storagerepo) and env(LOTUS_STORAGE_PATH) are DEPRECATION, will REMOVE SOON (default: "~/.lotusminer") [LOTUS_MINER_PATH, LOTUS_STORAGE_PATH]
+   --vv                           enables very verbose mode, useful for debugging the CLI (default: false)
```
rvagg, line 41: `--storagerepo` should be `--storagerepo value`? and then the spacing for this longest line should be the same as the others

documentation/en/cli-lotus.md
```diff
@@ -10,11 +11,12 @@ VERSION:
    1.31.1-dev
 
 COMMANDS:
-   daemon   Start a lotus daemon process
-   backup   Create node metadata backup
-   config   Manage node config
-   version  Print version
-   help, h  Shows a list of commands or help for one command
+   daemon        Start a lotus daemon process
```
rvagg, line 14: do we have a minimum width setting here that needs adjusting? looks like we have 4 extra spaces than we need

documentation/en/cli-lotus-miner.md
```diff
@@ -31,15 +33,19 @@ COMMANDS:
...
 GLOBAL OPTIONS:
-   --actor value, -a value                  specify other actor to query / manipulate
-   --color                                  use color in display output (default: depends on output being a TTY)
-   --miner-repo value, --storagerepo value  Specify miner repo path. flag(storagerepo) and env(LOTUS_STORAGE_PATH) are DEPRECATION, will REMOVE SOON (default: "~/.lotusminer") [$LOTUS_MINER_PATH, $LOTUS_STORAGE_PATH]
-   --vv                                     enables very verbose mode, useful for debugging the CLI (default: false)
-   --help, -h                               show help
-   --version, -v                            print the version
+   --actor value, -a              specify other actor to query / manipulate
+   --color                        use color in display output (default: depends on output being a TTY)
+   --panic-reports value           (default: "~/.lotusminer") [LOTUS_PANIC_REPORT_PATH]
```
rvagg, line 38: ah, this one is missing because `Hidden: true` - better build that into your code too, if a flag is `Hidden`, then exclude it from the list

documentation/en/cli-lotus-miner.md
rvagg: Oh, I get why this is shuffled now -- the original is following the order as they appear in the `COMMANDS:` output, but yours is following the naive order they appear in the `Commands` slice. I think we should follow that same order—either retain it when you print `COMMANDS:` or just follow this algorithm:

1. Iterate over the `Commands` slice and print out any that _don't_ have a `Category`
2. Iterate over the `Commands` slice again and print out any that _do_ have a `Category`

I don't know if there's a sorting on `Category` name applied, maybe it won't matter for now.

cli/pprof.go
```diff
@@ -11,47 +11,51 @@ import (
...
-var PprofCmd = &cli.Command{
-	Name:   "pprof",
-	Hidden: true,
-	Subcommands: []*cli.Command{
-		PprofGoroutines,
-	},
+func PprofCmd() *cli.Command {
+	return &cli.Command{
+		Name:   "pprof",
+		Hidden: true,
```
rvagg, line 17: you need to obey this so it doesn't show up in docs, same as flags, check for `Hidden` and skip if `true`

documentation/en/cli-lotus.md
```diff
@@ -1114,7 +1178,23 @@ OPTIONS:
    --help, -h    show help
 ```
 
+### lotus mpool clear
```
rvagg, line 1181: this is new because it's `Hidden` - because it's both deprecated and pretty unsafe to invoke so we don't want to encourage it

rvagg:
Very close @akaladarshi, good work so far. I haven't done an in-depth review of the actual code, I'm mainly looking at the output and assuming that the code is good, I'll look over that next.

cli/clicommands/cmd.go
```diff
@@ -13,19 +13,19 @@ var Commands = []*cli.Command{
 	lcli.WithCategory("basic", lcli.MultisigCmd),
 	lcli.WithCategory("basic", lcli.FilplusCmd),
 	lcli.WithCategory("basic", lcli.PaychCmd),
-	lcli.WithCategory("developer", lcli.AuthCmd),
+	lcli.WithCategory("developer", lcli.AuthCmd()),
```
rvagg, line 16: why do these need to become functions?

scripts/docsgen-cli/flag_formatter.go
```diff
@@ -0,0 +1,76 @@
...
+	return &FlagFormatter{flag: flag}
+}
+
+// Format formats the flag string with optional environment variables
+func (ff *FlagFormatter) Format() string {
+	names, usage, value, envVars := ff.GetData()
+
+	var formatted string
+	if value != "" {
+		formatted = fmt.Sprintf("   %-30s %s (default: %s)", names, usage, value)
```
rvagg, line 26:
```suggestion
		defpfx := " "
		if usage == "" {
			defpfx = ""
		}
		formatted = fmt.Sprintf("   %-30s %s%s(default: %s)", names, usage, defpfx, value)
```

something like this would deal with the case where there's no usage string; alternatively you could extend `usage` by one space if it's not `""`.
I'm not sure this is strictly needed when you account for the `Hidden` flags but it would fix current alignment problems.

rvagg: Also this `-30` is getting in the way when the `names` is too long. Like when `--miner-repo value, --storagerepo` comes in to play which is longer than you want.

Here's what I think I'd do given your current code structure so as not to have to make too many changes: make the this number a property of `FlagFormatter` that you pass in when you construct it (or make a default value you can override if you like) (you'll probably need to `fmt.Sprintf` the format itself to insert it). Then you do two passes iterating through your flags - the first pass looks at the maximum length of the `names` from `GetData` and then the second pass sets the maximum length if it's greater than 30 and does the actual print.

akaladarshi:
@rvagg I think I found a minimum solution for this we can directly use the urface cli package for generating, Initially I though we need to print everything (hidden), but this will directly print the correct format value.

```go
     customAppData := func() map[string]interface{} {
		return map[string]interface{}{
			"ExtraInfo": g.app.ExtraInfo,
		}
	}

	cli.HelpPrinterCustom(header, cli.AppHelpTemplate, g.app, customAppData())
```

akaladarshi:
> @rvagg I think I found a minimum solution for this we can directly use the urface cli package for generating, Initially I though we need to print everything (hidden), but this will directly print the correct format value.
> 
> ```go
>      customAppData := func() map[string]interface{} {
> 		return map[string]interface{}{
> 			"ExtraInfo": g.app.ExtraInfo,
> 		}
> 	}
> 
> 	cli.HelpPrinterCustom(header, cli.AppHelpTemplate, g.app, customAppData())
> ```

Removed the above change in favour of `app.Run(args[]string)`, because it was not printing the default values of flags. Anyway `app.Run` internally uses `cli.HelpPrinter` to print the commands.

cli/clicommands/cmd.go
akaladarshi: Previously, `lotus` and `lotus-miner` shared global command variables, causing conflicts when running app.Run("lotus-miner", "-h"). This led to jumbled help output like `lotus-miner lotus auth`.

Makefile
```diff
@@ -348,11 +348,8 @@ snap: lotus lotus-miner lotus-worker
 	snapcraft
 	# snapcraft upload ./lotus_*.snap
 
-# separate from gen because it needs binaries
-docsgen-cli: lotus lotus-miner lotus-worker
+docsgen-cli:
 	$(GOCC) run ./scripts/docsgen-cli
-	./lotus config default > documentation/en/default-lotus-config.toml
-	./lotus-miner config default > documentation/en/default-lotus-miner-config.toml
```
rvagg, line 355: @akaladarshi if you want to take a shortcut on this task, we could scope it just to doing what you've done so far - generate the docs from the structs, but leave the existing workflow that requires building the binaries in place, i.e. restore this Makefile target to the way it was. The remaining problem is the need to generate the default config files which currently need the binaries to do it, but we could defer that to a separate PR and get this one landed.

cli/clicommands/cmd.go
rvagg: See below, I went for a bit of a trek to find out why this is the case and came up with an easier approach than rewriting all of these.

scripts/docsgen-cli/main.go
```diff
@@ -45,89 +46,21 @@ func main() {
...
+		"lotus-miner":  miner.App(),
 	}
 
-	// Skip the first new line since the docs do not start with a newline at the very
-	// top.
-	if depth != 0 {
-		if _, err := file.WriteString("\n"); err != nil {
-			return err
+	for name, app := range cliApps {
+		generator := NewDocGenerator(outputDir, app)
```
rvagg, line 57:
```suggestion
	for name, app := range cliApps {
		// reset HelpName manually because shared Command objects get modified
		// by the first app they are added to
		for _, cmd := range app.Commands {
			cmd.HelpName = fmt.Sprintf("%s %s", app.HelpName, cmd.Name)
		}
		generator := NewDocGenerator(outputDir, app)
```

this is all that's needed to reset the problem that you're working around by making the `Command`s in functions: https://github.com/urfave/cli/blob/v2-maint/app.go#L232-L240

Unfortunately you can't just reset `HelpName` and re-run `app.Setup()` because it's a strictly run-once function, which is pretty annoying. But since we don't do anything sophisticated in `HelpName` I think it's good enough to just rewrite it here. Then you can undo all of the functions you had to make (sorry), but hopefully it's just a case of checking thise files out from master again.

rvagg: actually probably wont' be that easy because you had to put them in a new package, but hopefully it's straightforward enough to just un-`func` those shared cmds

Makefile
```diff
@@ -348,11 +348,11 @@ snap: lotus lotus-miner lotus-worker
 	snapcraft
 	# snapcraft upload ./lotus_*.snap
 
-# separate from gen because it needs binaries
-docsgen-cli: lotus lotus-miner lotus-worker
+docsgen-cli:lotus lotus-miner lotus-worker
```
rvagg, line 351:
```suggestion
docsgen-cli: lotus lotus-miner lotus-worker
```

Makefile
```diff
@@ -348,11 +348,11 @@ snap: lotus lotus-miner lotus-worker
 	snapcraft
 	# snapcraft upload ./lotus_*.snap
 
-# separate from gen because it needs binaries
 docsgen-cli: lotus lotus-miner lotus-worker
 	$(GOCC) run ./scripts/docsgen-cli
 	./lotus config default > documentation/en/default-lotus-config.toml
 	./lotus-miner config default > documentation/en/default-lotus-miner-config.toml
+
```
rvagg, line 355:
```suggestion
```

rvagg:
Squashed down to a single commit and rebased onto current master in this branch.

rvagg:
After a `make clean deps`, the current `master` has `make docsgen-cli` running at 25s. This branch does the same in 17s. The catch is that `docsgen-config` is a separate target and still requires the lotus and lotus-miner binaries (just not lotus-worker), so `make docsgen-cli docsgen-config`, being now a two-step process that first requires a partial compile (for `go run`) and a full compile for the two binaries needed for config, so it ends up ebing 30s, i.e. longer. But at least you can choose, and the config files are just another batch of work to be done that I want to separate from this PR.

@rjan90 please have a look at the changes I made in the head commit here, I've put `docsgen-config` as a separate required target to be run on releases and I also put it in CI.

rvagg:
I think follow-up work here to make the config output should be straightforward but it would be worth testing whether it should be part of the same binary or not, it might be faster to keep it in the same thing rather than have two binaries it has to setup and execute.

rvagg:
OK, nevermind about the follow-up task, I just bundled it in here and got rid of the need to `lotus config > out` at all, so now it's all part of the docsgen-cli app so it runs relatively quickly in the same binary. It's still not super-fast because it does need to load and compile a lot of code in the various paths that the CLI app touches, it just doesn't need to form a proper binary with all of the inclusions that a full `make` does.

`make docsgen-cli` after a `clean deps` takes 17s now. So we've shaved off approximately 1/3rd of the time for this task. I've got rid of the `docsgen-config` task I introduced.

@rjan90 if you wouldn't mind reviewing that would be appreciated. You're welcome to focus on the two additional commits I added since they are minimal and I already reviewed the larger initial commit.

documentation/en/cli-lotus.md
```diff
@@ -2208,10 +2208,10 @@ OPTIONS:
 
 ```
 NAME:
-   lotus log list - List log systems
+   lotus-miner log list - List log systems
 
 USAGE:
-   lotus log list [command options] [arguments...]
+   lotus-miner log list [command options] [arguments...]
```
rjan90, line 2214: I do not think this is correct? If I understand the intent of the changes/PR in general, it seems like we will still have `lotus auth api-info`, `lotus log list`, `lotus log set-level` and `lotus log alerts` commands, and therefor having these being written as lotus-miner CLIs in the Lotus CLI seems wrong?

rvagg: yeah, you're right, good catch; the annoyance is that the command objects are used in both places and they get a certain amount of internal state built up over use, and we use them twice so this name gets sticky; I'll figure out a fix

rjan90 (approved):

PR #12719, opened 2024-11-25
feat(api): v2 API integrating with F3

rvagg:
WIP but opening this for discussion purposes.

* To explore how we could evolve the existing APIs I've started a `/v2` API set here, although I'm not committing to pursuing that as _the_ solution, it just seems like an appropriate thing to do as long as we're messing with a bunch of standard methods. We could take a minimal approach of adding just the basics to begin with and then adding more as they prove to be necessary. The [Common Node API](https://github.com/filecoin-project/FIPs/pull/1027) seems like a good foundation to start with.
* I have two, probably mutually exclusive approaches in here, there's also other ways we could do similar things.
  1. The main one I've gone with is to introduce an `EpochSelector` string type that's either `"latest"` or `"finalized"` and then in places it can either be:
    * An optional argument where none previously existed (e.g. `ChainHead()` or `ChainHead("finalized")`).
    * An overload for some methods where `TipSetKey` currently exists - so you can supply either a tsk or one of these strings in that position.
    * An overload for some methods where `ChainEpoch` currently exists - so you can supply an integer or one of these strings in that position.
  2. Allowed `"latest"` and `"finalized"` to be JSON decoded into a value `TipSetKey`. `"latest"` comes out as an `EmptyTSK` which is the same as providing `[]` now. `"finalized"` comes out as a sentinel `FinalizedTSK`, which is a special TSK comprising `["bafkqaclgnfxgc3djpjswi"]` (raw identity CID with the string `"finalized"`). This approach would allow us to leave the existing APIs unmolested and even work within `/v1` transparently. Similar treatment could be done for `abi.ChainEpoch` but that's a little bit trickier because of the need for a sentinel value, what would that be? Also the optional argument would be a breaking Go API change if we wanted to add something to `ChainHead` and `ChainNotify`.


api/v2api/full.go
```diff
@@ -0,0 +1,142 @@
...
+	// ChainNotify returns channel with chain head updates. Takes an optional
+	// EpochSelector argument to specify the point in the chain to follow, which
+	// can be either "latest" or "finalized". This defaults to "latest" if not
+	// specified.
+	//
+	// The first message is guaranteed to be of length == 1, and type ==
+	// "current".
+	//
+	// The optional argument is decoded as a types.OptionalEpochSelectorArg.
+	ChainNotify(ctx context.Context, p jsonrpc.RawParams) (<-chan []*api.HeadChange, error) //perm:read
```
masih, line 24: Just so I learn: how popular/critical is it to have a "push" notification API? e.g. what would immediately break if this is converted to a pull model?

rvagg: Anything that returns a channel is usable via WebSockets so push data to you, otherwise you're polling, probably with `ChainHead`, but the nice thing about `ChainNotify` is that you get a constant stream of changes so you have a consistent chain including reverts and applies as it reorgs so you just process them as they come rather than trying to figure it out.

The alternative of polling `ChainHead` you'd get different tipsets when you poll and have to react to those changes in whatever way your local state demands. If you want to replicate the full `ChainNotify` behaviour you might need to follow up with `ChainGetPath` to get the changes between what you currently know in your local state as the chain head and what `ChainHead` is telling you what it is now, that way you could rebuild a `ChainNotify` but it'd be gnarly.

`ChainNotify` is a pretty critical API internally for Lotus, and there's at least one place that lotus-miner uses it (and therefore likely Curio too) and I wouldn't be surprised to see other external applications we're not aware of.

masih: Thanks, Rod.

Is there anything that inherently depends on the push model though? For example, is there something critical about the end-to-end latency that `ChainNotify` users care about/cannot be provided in a "pull" model?

I am specifically interested in its criticality for full node API users (i.e. excluding internal use by lotus node itself).

rvagg: I'm not sure. It's certainly a handy API. You can spin it up on a WebSocket connection and react to the changes as they come in.

I'm just not sure how many users are reliant on changes right at the head, I try and avoid it where possible because it's chaos! Maybe exchanges care? I don't really understand the certainty model that exchanges operate under.

You can even use `ChainNotify` just for a singular "tell me when the next epoch has arived" so I can get on with something, that's a pretty typical use internally and I can imagine that being useful externally too. I just can't point to an external use-case for it.

https://github.com/filecoin-project/lotus/blob/1b276e402caeb0f139bd881d76cdcdd206221f55/itests/fevm_test.go#L1409-L1413

masih: The context for my silly questions: It's my sense that such push APIs are 1) hard to get correct for external use, and 2) hard to scale (e.g. in the usecase of exchanges, archival nodes etc.).

Hence the conversation on whether we can avoid this API entirely in v2, in favour of a "Smart Client" implementation that affectively offers the same functionality but with a far smaller maintenance/correctness burden for Lotus Node implementation.

masih:
For the record: I continue to believe that the push style `ChainNotify` API is expensive to implement/maintain with unclear benefits to the users (e.g. they still have to implement reconnect on connection loss, delivery guarantees are AFAIK not spelt out and so on). 

My recommendation would be to move to a pull model unless there is a very clear usecase that requires concretely defined "event delivery guarantees". This needs a deeper Product Research. If you are a user of this API/building atop it: please reach out.

rvagg:
We can opt to not put some of these methods onto the gateway. Currently /v0/ and /v1/ both expose `ChainNotify` but we can easily opt to not add it to `/v2` and see if anyone squeals; that would be an interesting signal of usage. @masih you might also want to drop a note inline in https://github.com/filecoin-project/FIPs/pull/1027 about your concerns; excluding it from the Common Node API might be reasonable.

masih:
> you might also want to drop a note inline in [filecoin-project/FIPs#1027](https://github.com/filecoin-project/FIPs/pull/1027) about your concerns

[Done](https://github.com/filecoin-project/FIPs/pull/1027/files#r1867424000). Thanks for pointing out that FIP.

PR #12727, opened 2024-11-27
chore(gateway): add new methods

dumikau:


api/api_gateway.go
```diff
@@ -34,6 +35,12 @@ import (
 //  * Generate openrpc blobs
 
 type Gateway interface {
+	ChainGetNode(ctx context.Context, p string) (*IpldObject, error)
```
rvagg, line 38: why this one @dumikau ?

rvagg: tbh I can't see this one making it to the v2 API that's under development now without a really good use-case, are people asking for it?

dumikau: Yes, this method was added by request of a user from Neti-Soft

api/api_gateway.go
```diff
@@ -34,6 +35,12 @@ import (
...
+	StateMinerPartitions(ctx context.Context, m address.Address, dlIdx uint64, tsk types.TipSetKey) ([]Partition, error)
+	StateMinerSectors(context.Context, address.Address, *bitfield.BitField, types.TipSetKey) ([]*miner.SectorOnChainInfo, error)
+	StateMinerFaults(context.Context, address.Address, types.TipSetKey) (bitfield.BitField, error)
+	StateMinerAvailableBalance(context.Context, address.Address, types.TipSetKey) (types.BigInt, error)
+	StateMarketParticipants(context.Context, types.TipSetKey) (map[string]MarketBalance, error)
```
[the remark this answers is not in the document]
dumikau: Yes, all these methods were added by requests of different users

eshon:
- I checked with the requestor teams, and the only one still needed is `StateMinerAvailableBalance` which is safe to add.
The other StateMiner* calls can be removed from this PR.

- As for `ChainGetNode`, `ChainReadObj` would be a better option because it goes to the blockstore and not the network, so a workaround is being discussed.

rvagg:
I'm not completely against `ChainGetNode` if people _really_ want it. I'd just consider it a legacy API, so not ideal to encourage use of. But it is more efficient at getting to your data than `ChainReadObj` if you know exactly where you need to navigate to through a series of blocks.

rjan90:
> Thank you for submitting the PR and contributing to lotus! Lotus maintainers need more of your input before merging it, please address the suggested changes or reply to the comments or this PR will be closed in 72 hours

Closing since it has been over 72 hours, and stalebot was unable to close.

Issue #12709, opened 2024-11-20
Request for eth logs or events within a range doesn't return an error if they are not indexed


rvagg:
This was flagged in the initial ChainIndexer review but I think we need to do something about it. To describe the problem, consider the following from my node which has default ChainIndexer prune settings:

```sh
$ sqlite3 ./chainindex/chainindex.db 'select min(height) from tipset_message'
4438537
$ node -pe '"0x" + (4438530-10).toString(16)'
0x43b9f8
$ curl -X POST 'http://127.0.0.1:1236/rpc/v1' -H 'Content-Type: application/json' --data '{"jsonrpc":"2.0","id":1,"method":"Filecoin.EthGetLogs","params":[{"fromBlock":"0x43b9f8","toBlock": "0x43b9f8"}]}'
{"error":{"code":1,"message":"failed to get events for filter: failed to get events for filter from chain indexer: not found in index"},"id":1,"jsonrpc":"2.0"}
$ # ✅
$ curl -X POST 'http://127.0.0.1:1236/rpc/v1' -H 'Content-Type: application/json' --data '{"jsonrpc":"2.0","id":1,"method":"Filecoin.GetActorEventsRaw","params":[{"fromHeight":4438527,"toHeight":4438527}]}'
{"error":{"code":1,"message":"get events for filter: not found in index"},"id":1,"jsonrpc":"2.0"}
$ # ✅
$ curl -X POST 'http://127.0.0.1:1236/rpc/v1' -H 'Content-Type: application/json' --data '{"jsonrpc":"2.0","id":1,"method":"Filecoin.EthGetLogs","params":[{"fromBlock":"0x43b9f8","toBlock": "0x43b9f9"}]}'
{"id":1,"jsonrpc":"2.0","result":[]}
$ # ❌
$ curl -X POST 'http://127.0.0.1:1236/rpc/v1' -H 'Content-Type: application/json' --data '{"jsonrpc":"2.0","id":1,"method":"Filecoin.GetActorEventsRaw","params":[{"fromHeight":4438527,"toHeight":4438528}]}'
{"id":1,"jsonrpc":"2.0","result":null}
$ # ❌
```

1. When you request events for a specific epoch that isn't indexed, you'll get an error, this is good (also see https://github.com/filecoin-project/lotus/issues/12555)
2. When you request a range of epochs, it doesn't check whether it has them, you won't get an error and it'll act like there's just no events there

[...]

PR #12728, opened 2024-11-27
feat: check eth events indexed in range

akaladarshi:
## Related Issues
#12709 

## Proposed Changes
- check if the events within ranged is indexed


chain/index/events.go
```diff
@@ -280,6 +282,39 @@ func (si *SqliteIndexer) checkTipsetIndexedStatus(ctx context.Context, f *EventF
...
 
+func (si *SqliteIndexer) checkRangeIndexedStatus(ctx context.Context, f *EventFilter) error {
+	startCid, err := si.getTipsetKeyCidByHeight(ctx, f.MinHeight)
+	if err != nil {
+		if errors.Is(err, ErrNotFound) {
+			// Null round for the start of the range is acceptable
+			return nil
+		}
+		return xerrors.Errorf("failed to get tipset key cid for start height: %w", err)
+	}
```
rvagg, line 293: In the `ErrNotFound` case, I'd like it to increment by `1` until it finds one <= `MaxHeight` that's not a null round.

In the special case where there are only null rounds between `MinHeight` and `MaxHeight`, pretend that it's indexed and don't return an error, they'll just get no results.

So you'll need a loop here, and a local copy of `f.MinHeight` to increment.

chain/index/events.go
```diff
@@ -280,6 +282,39 @@ func (si *SqliteIndexer) checkTipsetIndexedStatus(ctx context.Context, f *EventF
...
+			if head == nil || f.MaxHeight > head.Height() {
+				return xerrors.Errorf("range end is in the future: %w", err)
+			}
+			return ErrNotFound // End is unindexed
+		}
+		return xerrors.Errorf("failed to get tipset key cid for end height: %w", err)
+	}
+
+	if exists, err := si.isTipsetIndexed(ctx, startCid); err != nil || !exists {
+		return xerrors.Errorf("start tipset not indexed: %w", err)
```
rvagg, line 309:
```suggestion
		return xerrors.Errorf("filter start tipset is not indexed: %w", err)
```

chain/index/events.go
```diff
@@ -280,6 +282,39 @@ func (si *SqliteIndexer) checkTipsetIndexedStatus(ctx context.Context, f *EventF
...
+
+	endCid, err := si.getTipsetKeyCidByHeight(ctx, f.MaxHeight)
+	if err != nil {
+		if errors.Is(err, ErrNotFound) {
+			// Check if the range ends in the future
+			head := si.cs.GetHeaviestTipSet()
+			if head == nil || f.MaxHeight > head.Height() {
+				return xerrors.Errorf("range end is in the future: %w", err)
+			}
+			return ErrNotFound // End is unindexed
```
rvagg, line 303: We're back in the null round case, I think that we should walk backward to find the first non-null lower than `MaxHeight`. Basically the same as the approach outlined above for `MinHeight` but in the other direction. You may find a way to abstract the algorithm to make it a reusable function, or it may be easier to just write it all out.

Basically: let's contract the range if the edges contain null rounds, walk _inward_ until we find a min and max that are not null rounds. Be careful not to iterate beyond the other bound (don't increment from Min beyond the Max value, and don't decrement from Max beyond the Min value). The special case is when there are _only_ null rounds between them, don't error, just return. There's another special case of contracting to min==max, you could easily do an `if` block with `startCid.Equals(endCid)` around the second `isTipsetIndexed` call to deal with that (although it's not critical since it'd just mean two queries for the same thing if it wasn't handled).

BigLep:
@akaladarshi : I assume you'll re-request review when it's ready.

chain/index/events.go
```diff
@@ -236,48 +238,98 @@ func loadExecutedMessages(ctx context.Context, cs ChainStore, recomputeTipSetSta
...
-			}
+// checkRangeIndexedStatus verifies if a range of heights is indexed.
+// It checks for the existence of non-null rounds at the range boundaries.
+func (si *SqliteIndexer) checkRangeIndexedStatus(ctx context.Context, f *EventFilter) error {
+	minHeight := f.MinHeight
+	maxHeight := f.MaxHeight
 
-			return xerrors.Errorf("failed to get tipset key cid by height: %w", err)
+	// Find the first non-null round in the range
+	startCid, err := si.findFirstNonNullRound(ctx, &minHeight, maxHeight)
```
rvagg, line 248: The use of the pointer to update it here is clever, but a little too non-obvious from a maintenance perspective, there's no obvious notice here it gets altered. I think you'd be better off returning a new `minHeight` along with the cid. You'll have to `var startCid, endCid cid.Cid` and not use `:=` for these calls though.

rvagg: `[]byte` sorry, not `cid.Cid`

which is another point - not being a CID, let's be careful with naming of these things, `startKeyBytes` perhaps?

chain/index/events.go
```diff
@@ -236,48 +238,98 @@ func loadExecutedMessages(ctx context.Context, cs ChainStore, recomputeTipSetSta
...
-		// This function distinguishes between two scenarios:
-		// 1. Missing events: The requested tipset is not present in the Index (an error condition).
-		// 2. Valid case: The tipset exists but contains no events (a normal situation).
-		// Currently, this distinction is only made for the common use case where a user requests events for a single tipset.
-		// TODO: Implement this functionality for a range of tipsets. This is expensive and not a common use case so it's deferred for now.
+		return xerrors.Errorf("failed to find last non-null round: %w", err)
+	}
+
+	// If all rounds are null, consider the range valid
+	if endCid == nil {
```
rvagg, line 268: oo, interesting case, you should never get here, right? If you can't come up with a plausible way to get here then I think we should error here and say something about this being "unexpected" (we want someone to report this error if they ever stumble upon it)

chain/index/events.go
```diff
@@ -236,48 +238,98 @@ func loadExecutedMessages(ctx context.Context, cs ChainStore, recomputeTipSetSta
...
 		return nil
 	}
 
-	// If we couldn't determine a specific tipset, return ErrNotFound
-	if tipsetKeyCid == nil {
-		return ErrNotFound
+	// Check indexing for start and end tipsets
+	if err := si.checkTipsetByKeyCid(ctx, startCid, minHeight); err != nil {
+		return err
```
rvagg, line 274: let's think through what the error is going to look like to the user here if they started with a null round height:

* Request with minHeight `X`
* Error says: `failed to check if tipset at height X+1 is indexed`

It's a bit awkward, but you could do a check for `minHeight != f.MinHeight` and in that case return a custom error wrapping the one you got, something like `xerrors.Errorf("checking first non-null tipset after %d: %w", f.MinHeight, err)`

rvagg: same with the `endCid` case too

rvagg: actually, now I think it through more, let's not bother, the error case should be rare I think, so not worth the extra code

chain/index/events.go
```diff
@@ -236,48 +238,98 @@ func loadExecutedMessages(ctx context.Context, cs ChainStore, recomputeTipSetSta
...
+	if err != nil {
+		return xerrors.Errorf("failed to check if tipset at height %d is indexed: %w", height, err)
+	}
+
+	if exists {
+		return nil // null round
+	}
+
+	return ErrNotFound // tipset is not indexed
+}
```
rvagg, line 296: I'm not really sure the `height` is necessary here, but the main thing I want to do is clarify what this is for. How about we rename this one `checkTipsetIndexedStatus` since you've dropped it form its parent


```suggestion
func (si *SqliteIndexer) checkTipsetIndexedStatus(ctx context.Context, tipsetKeyCid []byte, height abi.ChainEpoch) error {
	exists, err := si.isTipsetIndexed(ctx, tipsetKeyCid)
	if err != nil {
		return xerrors.Errorf("failed to check if tipset at height %d is indexed: %w", height, err)
	} else if exists {
		return nil // has been indexed
	}

	return ErrNotFound
}
```

chain/index/events.go
```diff
@@ -236,48 +238,98 @@ func loadExecutedMessages(ctx context.Context, cs ChainStore, recomputeTipSetSta
...
-	switch {
-	case f.TipsetCid != cid.Undef:
-		tipsetKeyCid = f.TipsetCid.Bytes()
-	case f.MinHeight >= 0 && f.MinHeight == f.MaxHeight:
-		tipsetKeyCid, err = si.getTipsetKeyCidByHeight(ctx, f.MinHeight)
-		if err != nil {
-			if err == ErrNotFound {
-				// this means that this is a null round and there exist no events for this epoch
-				return nil
-			}
```
rvagg, line 255: you've lost all of this logic in the process of refactoring, so we have a couple of problems now:

1. We're not testing the case where the filter specifies a specific `TipsetCid` - which we can do directly without needing to look at the range (case 1 above)
2. We're not taking the shortcut path for the case where the filter specifies a min==max, that one is easy because it's just one tipset to test (case 2 above)
3. Your new code just jumps straight in and assumes that `f.MinHeight` and `f.MaxHeight` are useful, but that may not be the case. They're allowed to have `-1` values, which signifies that neither have been specified (we do that in https://github.com/filecoin-project/lotus/blob/12d76bd9f11b14f0c89c69784616a5e9cb120058/node/impl/full/eth.go#L1844). Which is why the `>= 0` check existed in the original case.

So I think you'll need to restore some form of the original `switch` to accommodate those two cases above and then for the third case where `f.MinHeight >= 0 && f.MaxHeight >= 0`, and maybe `&& f.MinHeight < f.MaxHeight` just for good measure, just to be defensive about possible inputs.

chain/index/events.go
```diff
@@ -236,48 +238,98 @@ func loadExecutedMessages(ctx context.Context, cs ChainStore, recomputeTipSetSta
...
+func (si *SqliteIndexer) findFirstNonNullRound(ctx context.Context, minHeight *abi.ChainEpoch, maxHeight abi.ChainEpoch) ([]byte, error) {
+	for height := *minHeight; height <= maxHeight; height++ {
+		cid, err := si.getTipsetKeyCidByHeight(ctx, height)
+		if err == nil {
+			*minHeight = height // Update the minHeight to the found height
+			return cid, nil
+		}
+		if !errors.Is(err, ErrNotFound) {
+			return nil, xerrors.Errorf("failed to get tipset key cid for height %d: %w", height, err)
+		}
```
rvagg, line 308: Let's switch this for readability, since programmers are used to seeing errors handled first (but not always, just the majority of the time), we introduce a cognitive hiccup when we switch things around; but if we keep it close to the expected flow then it's easier to skim

```suggestion
		cid, err := si.getTipsetKeyCidByHeight(ctx, height)
		if err != nil {
			if !errors.Is(err, ErrNotFound) {
				return nil, xerrors.Errorf("failed to get tipset key cid for height %d: %w", height, err)
			}
			// else null round, keep searching
			continue
		}

		*minHeight = height // Update the minHeight to the found height
		return cid, nil
```

BigLep:
@akaladarshi : do you think you'll be able to incorporate feedback by EOD 2024-12-18?  Not required, but I think it would be nice to get merged before folks start disappearing for winter break.

akaladarshi:
> @akaladarshi : do you think you'll be able to incorporate feedback by EOD 2024-12-18?  Not required, but I think it would be nice to get merged before folks start disappearing for winter break.

Hey @BigLep, I am OOO this week will be back next week.

rvagg:
Squash merging in to a local branch `akaladarshi/check-logs-index-range`, I'll finish this off and do a new PR and retain the squashed commit for the original work so provenance isn't lost.

PR #12740, opened 2024-11-28
feat: enforce dependency on released versions of packages

### LINKED ISSUE #7131: CI: Depend On Releases
## Done Criteria
A CI check catches if go.mod is being updated with any non-released version.  This includes checking direct and transitive dependencies.  There should be a way for maintainers to override this check (e.g., PR label or code comment that links to explanation).

## Why Important
1. Security / reliability - While there's no guarantee that a released version doesn't have bugs or issues, it seems fair to assume that non-released versions have even more.  For example, https://github.com/filecoin-project/lotus/issues/12467 was triggered because of a bug in non-released library that lotus was depending on when the latest released version didn't have the bug.
2. (bonus) Faster builds
3. (bonus) Makes Lotus a better citizen when it's imported by other projects.
 
## User/Customer
1. All consumers if it makes Lotus more secure/stable.
2. Maintainers and contributors gain from the faster builds

## Notes
1. Our "changelog check" workflow could be used for inspiration.  Any case where we are letting in a non-released version, we want it to be intentional and documented.
2. It's clear that a true mechanism is needed.  [At least of 202409 we have 150+ out of 670+ dependencies using a commit hash rather than a released version.](https://github.com/filecoin-project/lotus/issues/7131#issuecomment-2359417825) This is something maintainers have wanted to change for 3+ years, but without the mechanism this is alive and well.
3. This issue can be marked as done once we prevent new-non-released versions from sneaking in without intentional approval.  It's a separate item to clean up "the sins of the past".

galargh:
## Related Issues

Resolves #7131

## Proposed Changes

This PR implements dependency on released versions enforcement as described in #7131.

## Additional Info

In this iteration of the workflow, the information about the violations is described in the workflow run details. Here's an example from the following run - https://github.com/filecoin-project/lotus/actions/runs/12073744091?pr=12740

Screenshot 2024-12-05 at 00 42 20

The list of approved unreleased dependencies is stored inline, in the workflow, see https://github.com/filecoin-project/lotus/blob/36da6130e408a99c6026b44367fa57c1f03fbde5/.github/workflows/dependency-check.yml#L20-L77

## Checklist

- [x] CI is green


CONTRIBUTING.md
```diff
@@ -58,6 +58,14 @@ If the change does not require a CHANGELOG.md entry, do one of the following:
 
 Note that this is enforced with https://github.com/filecoin-project/lotus/blob/master/.github/workflows/changelog.yml
 
+## Dependency Management
+
+We enforce that each dependency on an unreleased version of a package, as long as said package has any released versions (i.e. it is not a `v0.0.0`), is explicitly documented in the `ALLOWED` list stored in [.github/workflows/dependency-check.yml](.github/workflows/dependency-check.yml).
+If you are adding such a dependency, please add it to the `ALLOWED` list.
+Please note that this requirement applies both to direct and indirect dependencies.
+
+Note that this is enforced with https://github.com/filecoin-project/lotus/blob/master/.github/workflows/dependency-check.yml
```
BigLep, line 67:
```suggestion
## Dependency Management

We strive to use release dependencies because:
1. Security / reliability - While there's no guarantee that a released version doesn't have bugs or issues, it seems fair to assume that non-released versions have even more. For example, https://github.com/filecoin-project/lotus/issues/12467 was triggered because of a bug in non-released library that lotus was depending on when the latest released version didn't have the bug.
2. Faster builds
3. Makes Lotus a better citizen when it's imported by other projects.

We enforce that each dependency on an unreleased version of a packages is explicitly documented in the `ALLOWED` list stored in [.github/workflows/dependency-check.yml](.github/workflows/dependency-check.yml).
* If you are adding such a dependency, please add it to the `ALLOWED` list.
* This requirement applies both to direct and indirect dependencies.
* This requirement applies to packages that have released versions (i.e., is not a `v0.0.0`), 
* This is enforced with https://github.com/filecoin-project/lotus/blob/master/.github/workflows/dependency-check.yml
* This enforcement was initially done per [#7131](https://github.com/filecoin-project/lotus/issues/7131)
```

The main things I wanted to achieve are:
1. Document "the why"
2. Point to the back story so an interested party (or us when trying to recall this 6 months from now) can find it

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,101 @@
...
+              }
+            ]
+        run: |
+          echo "unreleased<<EOF" >> $GITHUB_OUTPUT
+          go list -m -json all |
+            jq -s '
+              (
+                map({Path: .Path, Version: .Version}) |
+                map(select(.Version | test("^v\\d+\\.\\d+\\.\\d+-")?)) |
+                map(select(.Version | test("^v0\\.0\\.0-")? | not))
```
BigLep, line 85: Go newbie question: is it possible to have a `v0.0.0` dependency if the the dependended-module actually has done a release?

Basically, per [here](https://github.com/filecoin-project/lotus/issues/7131#issuecomment-2359417825) we're excluding 100+ dependencies because they are v0.0.0 in our `go.mod`, but is it possible that we should actually be depending on non-v0.0.0 versions for some of them now?  My worry is that v0.0.0 dependencies get to fly under the radar and there is no forcing function to get them to "fly above the radar".

BigLep (commented):
Thanks @galargh .  This is an improvement I believe.  I would like other maintainers to give the approval like @rvagg , @Stebalien , and/or @rjan90 .  

A couple of things standing out to me:
1. I like that we're capturing the "reason" in our exception list.  
2. I wonder though if it would be better to inline the exception list / reason inline `go.mod`:

```go
require (
    github.com/yugabyte/pgx/v5 v5.5.3-yb-2     //  @unrelease-version-exceptoin-reason: "fill this in with the reason"
)
```

The presence of `@unrelease-version-exceptoin-reason` signals that this should have an exception and it also provides a reason.

Pros I see:
* All dependency info is in one file

Cons:
* There is more code to write for extracting which dependencies should have an exception.  Basically, we can't use `go list -m -json all` and do easy `jq` commands like you have since the dependency comments aren't going to be included with with `go list` commands.
* Not one quick place to check for all dependencies that have an exception (although this is easy to get with a grep)

I also want to recognize that it was @Stebalien who suggested having a [separate config file](https://github.com/filecoin-project/lotus/issues/7131#issuecomment-2428091019) like this.

3. Every network upgrade we temporarily introduce some dev dependencies (e.g., go-state-types).  Would we want to have the way to communicate a regex for a given dependency? 

```json
              {
                "Path": "github.com/filecoin-project/go-state-types",
                "VersionRegex": "v0.\d+.\d+-dev",
                "Reason": "Temporary dependency bubbling as part of a network upgrade release."
              },
```

Pros:
* One less step during network upgrades of having to update the exclusion list.

Cons:
* Complicates the checking logic (i.e., can't use the concise `jq` commands you have now).

@galargh : please don't act on any of my feedback yet.  Lets get input from @Stebalien , @rvagg , and/or @rjan90 first before any action is taken (if any action is to be taken).

Thanks!

Stebalien:
> I wonder though if it would be better to inline the exception list / reason inline go.mod:

That would be nice but we'll have to be a bit careful to avoid bad interactions with tooling. From my basic testing it looks like `go mod tidy` preserves comments, although comments on indirect dependencies turn into `// indirect; my comment`.

> Every network upgrade we temporarily introduce some dev dependencies (e.g., go-state-types). Would we want to have the way to communicate a regex for a given dependency?

I would at least allow a regex. Forcing the user to edit this file every time they want to update a dependency is going to be rather annoying.

Alternatively, we could support a set of flags (to avoid having to type out the same regular expressions over and over):

- "AllowUntaggedRelease" - allows arbitrary untagged versions.
- "AllowZeroVersion" - allows untagged versions if and only if there aren't any other tagged releases (we have 40 of these).
- "AllowDevRelease" - allows pre-release tags.

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,101 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+      - id: dependencies
+        env:
+          ALLOWED: |
+            [
+              {
+                "Path": "github.com/filecoin-project/go-data-transfer/v2",
```
rvagg, line 23: this one's been banished frm the codebase already

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,101 @@
...
+                "Version": "v2.0.0-rc7",
+                "Reason": "unknown"
+              },
```
rvagg, line 26: this one's been banished from the codebase already

```suggestion
```

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,101 @@
...
+              {
+                "Path": "github.com/filecoin-project/go-state-types",
+                "Version": "v0.16.0-dev",
```
rvagg, line 29: regex here would be good I think

.github/workflows/dependency-check.yml
rvagg: oh, wait, I see it in the `go list` list .. why tho?

rvagg:
I don't mind this, having an override is nice, having to put in notes is good too but it would be great to figure out what should go in for all of the ones listed so far.

galargh:
In https://github.com/filecoin-project/lotus/pull/12740/commits/e18cc37e68a8a292367f41c39ff0cc18c2f2c097, I was able to implement extracting the dependencies directly from the `go.mod` file and parsing special comment - `dependency-check-ignore`.

In https://github.com/filecoin-project/lotus/pull/12740/commits/01b45ef1636ed22769875c8f4b012117c7ba346f I added these special comments to the currently offending dependencies to showcase how they would look like. I left the reason as `unknown` since I lacked the context to fill it out myself.

CONTRIBUTING.md
galargh: Added in https://github.com/filecoin-project/lotus/pull/12740/commits/6e0bc3517c5cd68fd7c633ed57bbbf1e69dee404 :)

Stebalien:
Ooh. That's nice.

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    name: Dependency Check
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+      - id: all
+        run: |
+          echo "dependencies<<EOF" >> $GITHUB_OUTPUT
+          # Extract all dependencies from go.mod (include indirect dependencies and comments)
```
BigLep, line 21:
```suggestion
          # Extract all dependencies from go.mod (include indirect dependencies and comments).
          # `go list` isn't used because:
          # 1. it lists ALL the transitive dependencies, even those that are unused and don't make it to the go.mod file
          # 2. It doesn't extract the inline `dependency-check-ignore` comments.
```

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,49 @@
...
+          sed -n '/require (/,/)/{/require (/!{/)/!p;};}' go.mod | sed 's/^[[:space:]]*//' | tee -a $GITHUB_OUTPUT
```
BigLep, line 22:
```suggestion
          # Extract the lines from 'require (' to the first ')' including those lines in the go.mod file.
          sed -n '/require (/,/)/p' go.mod |

          # Remove the 'require (' line.
          sed '/require (/d' |

          # Remove the ')' line.
          sed '/^)/d' |

          # Remove leading whitespace from each line.
          sed 's/^[[:space:]]*//' |

          # Append the result to the file specified by the GITHUB_OUTPUT environment variable.
          tee -a $GITHUB_OUTPUT
```

I know what you have is correct, but I worry it will require someone to use GPT to unpack/understand it quickly.  (I may be miscalibrated here and maybe the syntax is common place.  I got scared off with the nested nature.). Above is an attempt to keep it simple and clear.

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,49 @@
...
+          echo "EOF" >> $GITHUB_OUTPUT
+      - id: unreleased
+        env:
+          DEPENDENCIES: ${{ steps.all.outputs.dependencies }}
+        run: |
+          echo "dependencies<<EOF" >> $GITHUB_OUTPUT
+          # Find all dependencies that use prerelease versions (exclude v0.0.0 versions)
```
BigLep, line 29:
```suggestion
          # Find all dependencies that use prerelease versions (i.e., exclude vX.Y.Z and v0.0.0 versions).
```

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,49 @@
...
+          DEPENDENCIES: ${{ steps.unreleased.outputs.dependencies }}
+        run: |
+          echo "dependencies<<EOF" >> $GITHUB_OUTPUT
+          # Find all unreleased dependencies without a dependency-check-ignore comment
+          grep -Pv 'dependency-check-ignore:\s' <<< "$DEPENDENCIES" | tee -a $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+      - if: steps.unexplained.outputs.dependencies != ''
+        env:
+          MESSAGE: |
+            A new unreleased dependency was discovered in this PR. Please do one of the options in [dependency management conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#dependency-management)
```
BigLep, line 43:
```suggestion
            A new unreleased and unexplained dependency was discovered in this PR. Please do one of the options in [dependency management conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#dependency-management).
```

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,49 @@
...
+
+            Unreleased dependencies:
```
BigLep, line 45:
```suggestion
            Unexplained unreleased dependencies:
```

CONTRIBUTING.md
```diff
@@ -58,6 +58,19 @@ If the change does not require a CHANGELOG.md entry, do one of the following:
...
+We strive to use release dependencies because:
+1. Security / reliability - While there's no guarantee that a released version doesn't have bugs or issues, it seems fair to assume that non-released versions have even more. For example, https://github.com/filecoin-project/lotus/issues/12467 was triggered because of a bug in non-released library that lotus was depending on when the latest released version didn't have the bug.
+2. Faster builds
+3. Makes Lotus a better citizen when it's imported by other projects.
+We enforce that each dependency on an unreleased version of a package is explicitly documented in the `go.mod` file via an inline comment of the form `dependency-check-ignore: <reason>`.
+* If you are adding such a dependency, please add a suitable comment to the `go.mod` file as well.
+* This requirement applies both to direct and indirect dependencies.
+* This requirement applies to packages that have released versions (i.e., is not a `v0.0.0`),
+* This is enforced with https://github.com/filecoin-project/lotus/blob/master/.github/workflows/dependency-check.yml
+* This enforcement was initially done per [#7131](https://github.com/filecoin-project/lotus/issues/7131)
```
BigLep, line 72:
```suggestion

We enforce that each dependency on an unreleased version of a package is explicitly documented in the `go.mod` file via an inline comment of the form `dependency-check-ignore: <reason>`.
* If you are adding such a dependency, please add a suitable comment to the `go.mod` file as well.
* This requirement applies both to direct and indirect dependencies.
* This requirement applies to packages that have released versions (i.e., is not a `v0.0.0`).
* This is enforced with https://github.com/filecoin-project/lotus/blob/master/.github/workflows/dependency-check.yml
* This enforcement was initially done per [#7131](https://github.com/filecoin-project/lotus/issues/7131).
```

Minor space/grammar nits.

BigLep (changes requested):
Thanks for the updates.  I like the the inline `dependency-check-ignore`.

A few thoughts:
1. There is boiler plate reading in and out the GITHIB_OUTPUT for each step, but it doesn't look like breaking into steps actually gives us better presentation when I look at https://github.com/filecoin-project/lotus/actions/runs/12237619056/job/34133776622?pr=12740
image

Would it be cleaner to just do it all in one step and print out along the way?

2. My only other main concern is that we don't have a `dependency-check-ignore-regex` and it if it's going to be annoying to maintainers like @rjan90 who bubble up go-state-types `vX.Y.z-dev` versions periodically.  That is the one area I can see where this new check could be inducing undesired work.  I am fine to wait to see if that actually turns out to be a nuisance, but my worry with the current dependency-check.yml is that this will be a non-trivial addition since we have so far been relying on shell scripting and shell scripting maybe gets ugly to parse regexes to then apply to a given line. (I don't have experience here doing this in shell script land - it's just my gut feel).   Things that would allay my concern are comments from maintainers saying that they don't need the regex or a sketch for how we'd quickly add it in future if we do want it.

3. For other reviewers, I don't think we should block this PR on the `dependency-check-ignore: UNKNOWN` cases.  I'm all for figuring these out and lets do it if someone can quickly figure it out (less than 30 minutes total for all of them).  If not, it should get backlogged since this PR isn't making things worse (it's just exposing the truth of the current situation).

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,49 @@
...
+          # Extract all dependencies from go.mod (include indirect dependencies and comments)
+          sed -n '/require (/,/)/{/require (/!{/)/!p;};}' go.mod | sed 's/^[[:space:]]*//' | tee -a $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+      - id: unreleased
+        env:
+          DEPENDENCIES: ${{ steps.all.outputs.dependencies }}
+        run: |
+          echo "dependencies<<EOF" >> $GITHUB_OUTPUT
+          # Find all dependencies that use prerelease versions (exclude v0.0.0 versions)
+          grep -Pv 'v0\.0\.0|v[0-9]+\.[0-9]+\.[0-9]+([^-]|$)' <<< "$DEPENDENCIES" | tee -a $GITHUB_OUTPUT
```
rvagg, line 30: first match in this seems redundant, no? `v0.0.0` is matched by the second part, or am I missing something?

From what I can tell, `[^-]` is making allowance for the two  `+incompatible` packages, is that correct? Or is there something more broad? Having an open ended "if it doesn't have a `-` then it can have anything you like" is probably not ideal here.

`+incompatible` is a [Golang-specific thing](https://go-review.googlesource.com/c/go/+/124384) and I think we could just accommodate that directly. So maybe it becomes something like:

```suggestion
          grep -Pv 'v[0-9]+\.[0-9]+\.[0-9]+(\+incompatible)?$' <<< "$DEPENDENCIES" | tee -a $GITHUB_OUTPUT
```

rvagg: oh, nmind re `v0.0.0`, it's open-ended on the back so it's for the non-tagged packages, we can be more specific I think because `v0.0.0` can also be a valid version and you could have `-whatever` on the back of it.

```suggestion
          grep -Pv 'v0\.0\.0-[0-9]{14}-[0-9a-f]{7,}$|v[0-9]+\.[0-9]+\.[0-9]+(\+incompatible)?$' <<< "$DEPENDENCIES" | tee -a $GITHUB_OUTPUT
```

go.mod
```diff
@@ -52,7 +52,7 @@ require (
 	github.com/filecoin-project/go-jsonrpc v0.7.0
 	github.com/filecoin-project/go-padreader v0.0.1
 	github.com/filecoin-project/go-paramfetch v0.0.4
-	github.com/filecoin-project/go-state-types v0.16.0-rc1
+	github.com/filecoin-project/go-state-types v0.16.0-rc1 // dependency-check-ignore: unknown
```
rvagg, line 55:
```suggestion
	github.com/filecoin-project/go-state-types v0.16.0-rc1 // dependency-check-ignore: preparing for nv25
```

rvagg:
inline `ignore` comments are nice, this is looking good

.github/workflows/dependency-check.yml
BigLep: @rvagg : can you speak to this comment?  (I don't know the Go world to know if this is a problem.)

rvagg: I'm pretty sure it's possible although I can't point to an instance. It's one of the reasons I was suggesting we be more specific in the new form @ https://github.com/filecoin-project/lotus/pull/12740#discussion_r1877013484 - regex the `v0.0.0-date-commit` form and avoid any other variations.

rvagg:
@galargh see https://github.com/filecoin-project/lotus/pull/12774 for my proposed changes btw, not sure if you saw my ping there

.github/workflows/dependency-check.yml
BigLep: This is resolved by https://github.com/filecoin-project/lotus/pull/12774 .  That PR checks that any v0.0.0 dependency doesn't have a tagged version and if it doesn, it must have the `dependency-check-ignore`.

BigLep:
I am good with this merging, once https://github.com/filecoin-project/lotus/pull/12774 is merged in as well.  At that point, all comments have been addressed.  Thanks again @galargh and @rvagg !

galargh:
> There is boiler plate reading in and out the GITHIB_OUTPUT for each step, but it doesn't look like breaking into steps actually gives us better presentation when I look at https://github.com/filecoin-project/lotus/actions/runs/12237619056/job/34133776622?pr=12740

I added step names to all the steps. https://github.com/filecoin-project/lotus/actions/runs/12314231229/job/34369897245?pr=12740 should read better now :) I used the comments that we included for the step names as they explain what happens inside pretty well.

> My only other main concern is that we don't have a dependency-check-ignore-regex and it if it's going to be annoying to maintainers like @rjan90 who bubble up go-state-types vX.Y.z-dev versions periodically. That is the one area I can see where this new check could be inducing undesired work. I am fine to wait to see if that actually turns out to be a nuisance, but my worry with the current dependency-check.yml is that this will be a non-trivial addition since we have so far been relying on shell scripting and shell scripting maybe gets ugly to parse regexes to then apply to a given line. (I don't have experience here doing this in shell script land - it's just my gut feel). Things that would allay my concern are comments from maintainers saying that they don't need the regex or a sketch for how we'd quickly add it in future if we do want it.

To be fair, I'm not sure if the regex support is necessary with the inline comments implementation. For example, now that we have a `dependency-check-ignore` comment next to `go-state-types` dependency, it will stay there until it is explicitly removed. In particular, it would survive any updates made through `go get go-state-types@some-new-unreleased-version`.

However, if we end up needing to add any more logic to the check, I think we should switch to a friendlier programming language, which should be pretty straightforward at this point.

> @galargh see https://github.com/filecoin-project/lotus/pull/12774 for my proposed changes btw, not sure if you saw my ping there

@rvagg Thanks! That's a great addition to the check 🚀  I think we can either a) merge as-is if you just rebase on top of this branch, or b) I could rework it in the style of the rest of the check myself (and possibly use a proper programming language while at it). @BigLep If we go with option b, I'd need a couple more hours to get it fully done if that's alright.

BigLep:
@galargh: 

> To be fair, I'm not sure if the regex support is necessary with the inline comments implementation. For example, now that we have a dependency-check-ignore comment next to go-state-types dependency, it will stay there until it is explicitly removed. In particular, it would survive any updates made through go get go-state-types@some-new-unreleased-version.

You're right.  What you have is good and I don't think more is needed.

---
I think we just get the "released versions after v0.0.0"  support from https://github.com/filecoin-project/lotus/pull/12774 in and we ship this.  Thanks!

BigLep:
Per slack DM with @galargh, Lotus maintainers will take on merging https://github.com/filecoin-project/lotus/pull/12774 into this PR (dealing with merge conflicts) and then merging the PR.  I will plan on doing it next week unless someone gets to it first.  Thanks all.

rvagg (approved):

BigLep:
Thanks for handling this @rvagg.  One less thing for me today - appreciated!

galargh:
Thanks for merging this 🙇

PR #12741, opened 2024-11-28
ci: automate release issue creation

### LINKED ISSUE #12477: lotus-releaser: release issue creation
## Done Criteria
A release engineer can answer some questions and get the corresponding release issue created to cover these realistic scenarios:
1. minor release for a network upgrade 
2. minor release for a non-network upgrade
3. normal patch release against current minor release
4. emergency patch release against current minor release
5. emergency patch release against previous minor release

There should also be a Node vs. Miner dimension.

## Why Important
https://github.com/filecoin-project/lotus/blob/master/documentation/misc/RELEASE_ISSUE_TEMPLATE.md is non-trivial at this point and requires various manual step depending on the release type one is doing.  This takes time each release and is somewhat error prone.  

This is heightened during high-severity events like we had with https://github.com/filecoin-project/lotus/issues/12467 .  In that case, releasers went off script because the desired script couldn't be generated quickly enough.  We ended up not following https://github.com/filecoin-project/lotus/blob/master/LOTUS_RELEASE_FLOW.md in that we didn't first merge changes to master and the changelog entries were varied between the patches.

This also serves as a initial simple beach head to incrementally add more tooling/automation to the release process  so we can start reducing the number of steps in the checklist.  

## User/Customer
Maintainers / releasers

## Notes
1. By "emergency patch release" I mean likely just one commit and no RC.
2. I think this is the data input the scripting needs to take to get the permutations above:
    1. Is this for node or miner?
    2. Is this for a network upgrade? If so, what network version?
    3. Is this a minor release?
    4. Is this a patch release? If yes, what previous release are you patching?
    5. Is this an emergency patch? (If yes, we can skip RC steps)
3. While the initial hope is just to spit out a URL that a release engineer can click to get the release pre-populated, I think we can build towards something like https://github.com/ipfs/kuboreleaser so that more of the steps are automated and standardized.

galargh:
## Related Issues

Resolves #12477

## Proposed Changes

This PR implements the release issue creation automation.

Firstly, it turns the `RELEASE_ISSUE_TEMPLATE.md` into a Go template. This allowed us to use variable substitution and conditional blocks, which were the prerequisites for release issue rendering. We used variable substitution inline. We hid the conditional block starts/ends in HTML comment tags (`<!--`, `-->`). We kept the modifications to a minimum and ensured that the template could still be copied manually and updated using almost the same steps as before.

Secondly, we added a new command - `create-issue` - to the `cmd/release` tool. The command accepts a number of flags, values of which are then used to render the issue template. Before creating a new issue, the command checks whether if it already exists. It requires GITHUB_TOKEN with write access to issues to be present in the environment. The command doesn't operate in interactive mode. It expects all the data to be provided via flags instead.

Finally, we created a new GitHub Actions workflow which accepts exactly the same inputs as the command and calls it to create the new release issue. It can be triggered on demand.

## Additional Info

An example of the release issue created using `release create-issue --type node --tag 1.2.3 --level patch --network-upgrade 23`:  https://github.com/galargh/.github/issues/73

## Checklist

- [x] CI is green

### DIFF
--- a/.github/workflows/create-release-issue.yml
+++ b/.github/workflows/create-release-issue.yml
@@ -4,16 +4,27 @@ on:
   workflow_dispatch:
     inputs:
       type:
-        description: 'What's the type of the release? (Options: node, miner, both)"'
+        description: "What's the type of the release?"
         required: true
+        type: choice
+        options:
+          - node
+          - miner
+          - both
       tag:
         description: "What's the tag of the release? (e.g., 1.30.1)"
         required: true
       level:
-        description: "What's the level of the release? (Options: major, minor, patch)"
+        description: "What's the level of the release?"
         required: true
+        default: 'warning'
+        type: choice
+        options:
+          - major
+          - minor
+          - patch
       network-upgrade:
-        description: "What's the version of the network upgrade this release is related to? (e.g., 25)"
+        description: "What's the version of the network upgrade this release is related to? (e.g. 25)"
         required: false
       discussion-link:
         description: "What's a link to the GitHub Discussions topic for the network upgrade?"
--- a/.github/workflows/create-release-issue.yml
+++ b/.github/workflows/create-release-issue.yml
@@ -54,11 +54,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/install-go
-      - run: go build -o release ./cmd/release
       - env:
           GITHUB_TOKEN: ${{ github.token }}
         run: |
-          ./release create-issue \
+          go run cmd/release/main.go create-issue \
             --create-on-github true \
             --type "${{ github.event.inputs.type }}" \
             --tag "${{ github.event.inputs.tag }}" \
--- a/.github/workflows/create-release-issue.yml
+++ b/.github/workflows/create-release-issue.yml
@@ -4,16 +4,16 @@ on:
   workflow_dispatch:
     inputs:
       type:
-        description: 'Release type'
+        description: 'What's the type of the release? (Options: node, miner, both)"'
         required: true
       tag:
-        description: "What's the tag of the release? (e.g. 1.30.1)"
+        description: "What's the tag of the release? (e.g., 1.30.1)"
         required: true
       level:
-        description: "What's the level of the release? (major, minor, patch)"
+        description: "What's the level of the release? (Options: major, minor, patch)"
         required: true
       network-upgrade:
-        description: "What's the version of the network upgrade this release is related to? (e.g. 23)"
+        description: "What's the version of the network upgrade this release is related to? (e.g., 25)"
         required: false
       discussion-link:
         description: "What's a link to the GitHub Discussions topic for the network upgrade?"
@@ -22,22 +22,10 @@ on:
         description: "What's a link to the Lotus CHANGELOG entry for the network upgrade?"
         required: false
       rc1-date:
-        description: "What's the expected shipping date for RC1 (YYYY-MM-DD)?"
-        required: false
-      rc1-precision:
-        description: "How precise is the RC1 date? (day, week)"
-        required: false
-      rc1-confidence:
-        description: "How confident is the RC1 date? (estimated, confirmed)"
+        description: "What's the expected shipping date for RC1?"
         required: false
       stable-date:
-        description: "What's the expected shipping date for the stable release (YYYY-MM-DD)?"
-        required: false
-      stable-precision:
-        description: "How precise is the stable release date? (day, week)"
-        required: false
-      stable-confidence:
-        description: "How confident is the stable release date? (estimated, confirmed)"
+        description: "What's the expected shipping date for the stable release?"
         required: false
 
 defaults:
@@ -60,6 +48,7 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         run: |
           ./release create-issue \
+            --create-on-github true \
             --type "${{ github.event.inputs.type }}" \
             --tag "${{ github.event.inputs.tag }}" \
             --level "${{ github.event.inputs.level }}" \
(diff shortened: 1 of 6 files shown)


BigLep:
@galargh : this is awesome!  I spent quite a bit of time with this today when working to generate the 1.32.0 release issue.  A made a set of changes based on this experience in https://github.com/filecoin-project/lotus/pull/12749 which is targeting this branch.

galargh:
@BigLep Lovely! Thank you for that. I now merged your follow-up into this PR. Let me know if there's anything else we should take care of in this iteration.

cmd/release/main.go
```diff
@@ -1,11 +1,21 @@
...
+	"html/template"
+	"net/url"
 	"os"
 	"os/exec"
+	"regexp"
+	"strconv"
 	"strings"
 
+	masterminds "github.com/Masterminds/semver/v3"
+	sprig "github.com/Masterminds/sprig/v3"
```
BigLep, line 17: I don't know the go conventions, but given we have two things coming from "masterminds", should we call this `masterminds_semver` ?

BigLep (approved):
Looks good to me as I already engaged a lot with this, but it would be good to have @rjan90 or @rvagg approve so they're aware.  Thanks!

cmd/release/main.go
rvagg: it's not a big deal, but `semver` and `sprig` would probably be best, in fact they should be the default for both of these, no? so they may not need the names at all

cmd/release/main.go
```diff
@@ -163,6 +175,237 @@ func main() {
...
+					// releaseType gets special formatting
+					if releaseType == "node" {
+						releaseType = "Node"
+					} else if releaseType == "miner" {
+						releaseType = "Miner"
+					} else if releaseType == "both" {
+						releaseType = "Node and Miner"
+					} else {
+						return fmt.Errorf("invalid value for the 'type' flag. Allowed values are 'node', 'miner', and 'both'")
+					}
```
rvagg, line 252:
```suggestion
					releaseType := c.String("type")
					switch releaseType {
					case "node":
					    releaseType = "Node"
					case "miner":
					    releaseType = "Miner"
					case "both":
					    releaseType = "Node and Miner"
					default:
					    return fmt.Errorf("invalid value for the 'type' flag. Allowed values are 'node', 'miner', and 'both'")
					}
```

cmd/release/main.go
```diff
@@ -163,6 +175,237 @@ func main() {
...
+					data["CreateOnGitHub"] = createOnGitHub
+					data["Type"] = releaseType
+					data["Tag"] = releaseVersion.String()
+					data["NextTag"] = releaseVersion.IncPatch().String()
+					data["Level"] = releaseLevel
+					data["NetworkUpgrade"] = networkUpgrade
+					data["NetworkUpgradeDiscussionLink"] = discussionLink
+					data["NetworkUpgradeChangelogEntryLink"] = changelogLink
+					data["RC1DateString"] = rc1Date
+					data["StableDateString"] = stableDate
```
rvagg, line 325:
```suggestion
data := map[string]interface{}{
	"CreateOnGitHub": createOnGitHub,
	"Type": releaseType,
	"Tag": releaseVersion.String(),
	"NextTag": releaseVersion.IncPatch().String(),
	"Level": releaseLevel,
	"NetworkUpgrade": networkUpgrade,
	"NetworkUpgradeDiscussionLink": discussionLink,
	"NetworkUpgradeChangelogEntryLink": changelogLink,
	"RC1DateString": rc1Date,
	"StableDateString": stableDate,
}
```

cmd/release/main.go
```diff
@@ -163,6 +175,237 @@ func main() {
...
+
+Body:
+-----
+%s
+
+URL to create issue:
+-------------------
+%s
+`
+						fmt.Printf(debugFormat, issueTitle, issueBody, issueURL)
```
rvagg, line 378: Typically in urfave/cli you should `_, _ = fmt.Fprintf(c.App.Writer, "tmpl", arg1, arg2)`, and you also have access to `c.App.ErrWriter`. They point to stdout and stderr by default but it makes the app code more flexible.

rvagg (approved):
seems ok to me, some Go cleanup suggestions left inline

cmd/release/main.go
BigLep: The alias for spring is my Go newbieness.  
I believe masterminds semver is named because another semver library is used below.

cmd/release/main.go
BigLep: Thanks.  Go newbie at work here.

cmd/release/main.go
BigLep: Thank you.  I meant to look up if there was a more succinct way to do this.  Thanks.

.github/workflows/create-release-issue.yml
```diff
@@ -0,0 +1,71 @@
...
+    name: Create Release Issue
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/install-go
+      - run: go build -o release ./cmd/release
+      - env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          ./release create-issue \
```
rvagg, line 61: could be a single `go run cmd/release/main.go create issue`

not critical, but we also do it this way in release.yml

cmd/release/main.go
BigLep: I have removed the `sprig` label.

.github/workflows/create-release-issue.yml
BigLep: I updated this - thanks.

cmd/release/main.go
BigLep: Thanks.  I updated this.

PR #12747, opened 2024-12-02
fix(sync): do not allow to expand checkpointed tipsets

Kubuxu:
Resolves a bug in sync by preventing checkpoint expansion.

### DIFF
--- a/chain/sync.go
+++ b/chain/sync.go
@@ -531,7 +531,7 @@ func (syncer *Syncer) Sync(ctx context.Context, maybeHead *types.TipSet) error {
 	if maybeHead.Height() == hts.Height() {
 		// check if maybeHead is fully contained in headTipSet
 		// if that is the case, there is nothing for us to do
-		// we need to exit out early, otherwise checkpoint-fork logic will reject it
+		// we need to exit out early, otherwise checkpoint-fork logic might wrongly reject it
 		fullyContained := true
 		for _, c := range maybeHead.Cids() {
 			if !hts.Contains(c) {
--- a/chain/sync.go
+++ b/chain/sync.go
@@ -530,6 +530,7 @@ func (syncer *Syncer) Sync(ctx context.Context, maybeHead *types.TipSet) error {
 
 	if maybeHead.Height() == hts.Height() {
 		// check if maybeHead is fully contained in headTipSet
+		// meaning we already synced all the blocks that are a part of maybeHead
 		// if that is the case, there is nothing for us to do
 		// we need to exit out early, otherwise checkpoint-fork logic might wrongly reject it
 		fullyContained := true
--- a/chain/sync.go
+++ b/chain/sync.go
@@ -845,6 +845,14 @@ loop:
 	if err != nil {
 		return nil, xerrors.Errorf("failed to load next local tipset: %w", err)
 	}
+
+	if chkpt := syncer.store.GetCheckpoint(); !ignoreCheckpoint && chkpt != nil {
+		// don't allow us to expand the tipset beyond the checkpoint
+		if known.Equals(chkpt) && base.IsChildOf(knownParent) {
+			return nil, xerrors.Errorf("tispet expanding fork: %w", ErrForkCheckpoint)
+		}
+	}
+
 	if base.IsChildOf(knownParent) {
 		// common case: receiving a block that's potentially part of the same tipset as our best block
 		return blockSet, nil
--- a/chain/sync.go
+++ b/chain/sync.go
@@ -520,12 +520,30 @@ func (syncer *Syncer) Sync(ctx context.Context, maybeHead *types.TipSet) error {
 
 	hts := syncer.store.GetHeaviestTipSet()
 
+	// noop pre-checks
 	if hts.ParentWeight().GreaterThan(maybeHead.ParentWeight()) {
 		return nil
 	}
 	if syncer.Genesis.Equals(maybeHead) || hts.Equals(maybeHead) {
 		return nil
 	}
+
+	if maybeHead.Height() == hts.Height() {
+		// check if maybeHead is fully contained in headTipSet
+		// if that is the case, there is nothing for us to do
+		// we need to exit out early, otherwise checkpoint-fork logic will reject it
+		fullyContained := true
+		for _, c := range maybeHead.Cids() {
+			if !hts.Contains(c) {
+				fullyContained = false
+				break
+			}
+		}
+		if fullyContained {
+			return nil
+		}
+	}
+	// end of noop prechecks
 
 	if err := syncer.collectChain(ctx, maybeHead, hts, false); err != nil {
 		span.AddAttributes(trace.StringAttribute("col_error", err.Error()))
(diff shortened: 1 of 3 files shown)


chain/sync.go
```diff
@@ -845,6 +845,14 @@ loop:
 	if err != nil {
 		return nil, xerrors.Errorf("failed to load next local tipset: %w", err)
 	}
+
+	if chkpt := syncer.store.GetCheckpoint(); !ignoreCheckpoint && chkpt != nil {
```
Stebalien, line 849: Is this not going to be a bit racy? I guess maybe that's ok? Will we try to re-sync otherwise?

In general, I'm not sure why we're not just doing this with heights. I.e., if `known` has a height less than or equal to the checkpoint, don't try to sync that chain?

BigLep:
@rvagg : it came up during 2024-12-02 F3 sync that it would be good if you're able to review this as this is one of the blockers for a 1.32.0 RC (release issue coming...)

chain/sync.go
Kubuxu: You are correct. That would work great as well.
The majority of the thing here was trying to get the test working.

chain/sync.go
```diff
@@ -845,6 +845,14 @@ loop:
...
+		// don't allow us to expand the tipset beyond the checkpoint
+		if known.Equals(chkpt) && base.IsChildOf(knownParent) {
+			return nil, xerrors.Errorf("tispet expanding fork: %w", ErrForkCheckpoint)
+		}
+	}
```
rvagg, line 854:
```suggestion
	if !ignoreCheckpoint {
		if chkpt := syncer.store.GetCheckpoint(); chkpt != nil && base.Height() <= chkpt.Height() {
			// don't allow us to expand the tipset beyond the checkpoint
			return nil, xerrors.Errorf("tispet at or below checkpoint height: %w", ErrForkCheckpoint)
		}
	}
```

Kubuxu: This wasn't as trivial, needed to make one more change.

masih (approved):

chain/sync.go
```diff
@@ -520,13 +520,31 @@ func (syncer *Syncer) Sync(ctx context.Context, maybeHead *types.TipSet) error {
...
 		return nil
 	}
 	if syncer.Genesis.Equals(maybeHead) || hts.Equals(maybeHead) {
 		return nil
 	}
 
+	if maybeHead.Height() == hts.Height() {
+		// check if maybeHead is fully contained in headTipSet
+		// if that is the case, there is nothing for us to do
+		// we need to exit out early, otherwise checkpoint-fork logic will reject it
```
Stebalien, line 534: I think this logic is right, but I'm not sure about the comment. The checkpoint issue is when `hts` is fully contained in `maybeHead`, not the other way around.

Here the issue is that we'll try to sync the wrong chain from our peers (when we later try to expand it?) and then _fail_, refusing to checkpoint to the smaller head. At least I think that's the issue...

Also, what happens if `maybeHead` is `hts+1`? `+2`?

Kubuxu: If our head is [A, B], and tipset [A] gets in here, it should be accepted, but the new checkpoint rejection logic will reject it.
If it is something built on [A, B] it is okay, checked in a test.

chain/sync.go
```diff
@@ -520,13 +520,31 @@ func (syncer *Syncer) Sync(ctx context.Context, maybeHead *types.TipSet) error {
 
 	hts := syncer.store.GetHeaviestTipSet()
 
+	// noop pre-checks
 	if hts.ParentWeight().GreaterThan(maybeHead.ParentWeight()) {
 		return nil
 	}
```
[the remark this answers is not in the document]
Kubuxu: As far as I can see, checkpointing will force switch the head, bypassing this check, at which point the checkpoint is the head.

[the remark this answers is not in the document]
Stebalien: We sync then checkpoint, not the other way around, IIRC?

chain/sync.go
Stebalien: It should be rejected because it's lighter than `[A, B]`. But this has nothing to do with checkpoints.

The _checkpointing_ issue is that, if we try to switch from `[A, B]` to `[A]`, we can fail in `Sync` in some test scenarios (that was the test issue _I_ was running into, at least) even though we already _have_ `[A]`. That will _prevent_ us from checkpointing.

So, I'm saying that this logic will fix the test issue but I'm still pretty sure the comment is wrong (and I'm not sure if this is the _best_ way to fix the failure).

Kubuxu: > It should be rejected because it's lighter than [A, B]. But this has nothing to do with checkpoints.

It will be accepted as we only look at ParentWeight. This is how tipset expansion naturally works. Without checkpoint given [A, B], incoming [C], expansion to [A, B, C] is natural.

Yes, there is another issue where if we are on [A, B] and we want to checkpoint [A], we will initiate fetch of [A]. This is more of a wart than a consensus-affecting bug.

1. Without this early escape, given head=checkpoint=[A, B], we would reject the sync attempt to [A], which should be accepted as it is a valid block that is part of our tipset (rejecting sync attempt of A might lead to marking [A] as bad block which we for sure don't want).
2. I've adjusted the test to push tipset [A,B] to the node and then checkpoint [A] to show that it works.
3. What this code fixes is rejecting the expansion from checkpointed [A] to [A, B], which is tested in [here](https://github.com/filecoin-project/lotus/pull/12747/files#diff-1444b0b32efb1403c113d6340498ebd7a2940b944011d25909c1871fb3922b04R1070)

Kubuxu: > The checkpoint issue is when hts is fully contained in maybeHead, not the other way around.

Yes, and that is resolved by flat-out rejecting tipsets that try to merge at or below checkpoint, but then, as explained in 1., we need to allow/or just ignore subsets of `hts`  while `hts = checkpoint` is our head, so we don't reject them.

Kubuxu: I slightly adjusted the wording of the comment

chain/sync.go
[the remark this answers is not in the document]
Kubuxu: The checkpoint logic uses `collectChain` directly.

chain/sync.go
Stebalien: Oh, I see... Can we fix this where we actually mark the tipset as bad? I.e.:

1. We check to see if it syncs away from the checkpoint.
2. Mark all blocks as bad that _aren't_ a part of the checkpoint?

Or am I still misunderstanding this issue.

Kubuxu: Blocks not part of the checkpoint will still get marked as bad. It is just the blocks that are part of the `checkpoint = head` that bypass sync.

Stebalien: Also, I _think_ the current fix has a race:

1. We start syncing `[A]`.
2. We checkpoint `[A, B]`.
3. We finish syncing `[A]` and mark `[A]` as bad?

Kubuxu: They don't get added as bad. I would prefer not to touch that logic in this PR, though.
The `collectChain` errors don't get marked as bad automatically.

chain/sync.go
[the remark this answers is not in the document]
Stebalien: We call `collectChain` before calling `SetCheckpoint`: https://github.com/filecoin-project/lotus/blob/5791b4a9786bc259df2c1cc2775fd9aaf62d25fb/chain/checkpoint.go#L38-L40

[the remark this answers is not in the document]
Stebalien: Resolved sync. We call `collectChain` _directly_, bypassing this code entirely.

Stebalien (approved):

PR #12762, opened 2024-12-05
feat(f3): resolve finality for eth APIs according to F3

rvagg:
From CHANGELOG:

> **Ethereum APIs meet F3!** When F3 is enabled and running, Ethereum APIs that accept block descriptors `"finalized"` and `"safe"` will use F3 to determine the block to select instead of the default 30 block delay for `"safe"` and 900 block delay for `"finalized"`.

* Introduced an `lf3.F3API` interface that I can mock and replace the default `lf3.F3`
* Added an `IsEnabled()` to the F3 interface, `lf3.F3` returns `true`
* Added a `lf3.DisabledF3` implementation that gets loaded when F3 isn't enabled, it returns `false` for `IsEnabled()`. All methods now have an error return and they'll return `api.ErrF3Disabled`; so this simplifies `node/impl/full/f3.go` which just has to pass through the calls instead of checking whether `fff` is `nil`.
* Added a `MockF3API` that can be used as a replacement to modify some of its behaviour (focused on the changes in here fore now).
* Added a `TipSetResolver`, which for now, just exposes `ResolveEthBlockSelector` for the Ethereum APIs. But this will be the extension point for the additional non-eth API work (I'll rebase https://github.com/filecoin-project/lotus/pull/12719 ontop of this). This replaces the eth utility function `getTipsetByBlockNumber` and integrates with F3.

So, now when you resolve `"safe"` or `"finalized"`, we first query F3, then:

1. If F3 is disabled or not running, fall back to EC behaviour (`"safe"` = head-30-1, `"finalized"` = head-900-1) (we always -1 for the Eth APIs because the queries are transaction-centric and care about transaction execution).
2. If F3 reports its latest epoch is later than EC finality, we fall back to EC behaviour
3. If F3 is closer to head than the default EC variant (31, or 901), use the F3 epoch (-1) instead

@Kubuxu @masih PTAL

@Stebalien I'm pinging you because I wouldn't mind you additional sanity checking on the eth tipset-1 behaviour in here.

### DIFF
--- a/chain/tsresolver/tipset_resolver.go
+++ b/chain/tsresolver/tipset_resolver.go
@@ -1,0 +1,191 @@
+package tsresolver
+
+import (
+	"context"
+	"errors"
+
+	logging "github.com/ipfs/go-log/v2"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-f3/certs"
+	"github.com/filecoin-project/go-f3/manifest"
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/actors/policy"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/chain/types/ethtypes"
+)
+
+var log = logging.Logger("chain/tsresolver")
+
+const (
+	EthBlockSelectorEarliest  = "earliest"
+	EthBlockSelectorPending   = "pending"
+	EthBlockSelectorLatest    = "latest"
+	EthBlockSelectorSafe      = "safe"
+	EthBlockSelectorFinalized = "finalized"
+)
+
+type TipSetLoader interface {
+	GetHeaviestTipSet() (ts *types.TipSet)
+	LoadTipSet(ctx context.Context, tsk types.TipSetKey) (*types.TipSet, error)
+	GetTipsetByHeight(ctx context.Context, h abi.ChainEpoch, anchor *types.TipSet, prev bool) (*types.TipSet, error)
+}
+
+type F3 interface {
+	GetManifest(ctx context.Context) (*manifest.Manifest, error)
+	GetLatestCert(ctx context.Context) (*certs.FinalityCertificate, error)
+}
+
+type TipSetResolver interface {
+	ResolveEthBlockSelector(ctx context.Context, selector string, strict bool) (*types.TipSet, error)
+}
+
+type tipSetResolver struct {
+	loader TipSetLoader
+	f3     F3
+}
+
+var _ TipSetResolver = (*tipSetResolver)(nil)
+
+func NewTipSetResolver(loader TipSetLoader, f3 F3) TipSetResolver {
+	return &tipSetResolver{
+		loader: loader,
+		f3:     f3,
+	}
+}
+
+// ResolveEthBlockSelector resolves an Ethereum block selector string to a TipSet.
+//
+// The selector can be one of:
+//   - "pending": the chain head
+//   - "latest": the TipSet with the latest executed messages (head - 1)
+//   - "safe": the TipSet with messages executed at least eth.SafeEpochDelay (30) epochs ago or the
+//     latest F3 finalized TipSet
+//   - "finalized": the TipSet with messages executed at least policy.ChainFinality (900) epochs ago
+//     or the latest F3 finalized TipSet
+//   - a decimal block number: the TipSet at the given height
+//   - a 0x-prefixed hex block number: the TipSet at the given height
+//
+// If a specific block number is specified and `strict` is true, an error is returned if the block
+// number resolves to a null round, otherwise in the case of a null round the first non-null TipSet
+// immediately before the null round is returned.
+func (tsr *tipSetResolver) ResolveEthBlockSelector(ctx context.Context, selector string, strict bool) (*types.TipSet, error) {
+	switch selector {
+	case EthBlockSelectorEarliest:
+		return nil, xerrors.Errorf(`block param "%s" is not supported`, EthBlockSelectorEarliest)
+
+	case EthBlockSelectorPending:
+		return tsr.loader.GetHeaviestTipSet(), nil
+
+	case EthBlockSelectorLatest:
+		return tsr.loader.LoadTipSet(ctx, tsr.loader.GetHeaviestTipSet().Parents())
+
+	case EthBlockSelectorSafe:
+		return tsr.loadFinalizedTipSet(ctx, ethtypes.SafeEpochDelay)
+
+	case EthBlockSelectorFinalized:
+		return tsr.loadFinalizedTipSet(ctx, policy.ChainFinality)
+
+	default:
+		// likely an 0x hex block number or a decimal block number
+		return tsr.resolveEthBlockNumberSelector(ctx, selector, strict)
+	}
+}
+
+// loadFinalizedTipSet will use F3 to load and return the latest finalized tipset.
+//
+//   - If it can't load a finalized tipset from F3, it will return the tipset fallbackDelay epochs
+//     behind the chain head (-1).
+//   - If the finalized tipset from F3 is older than fallbackDelay epochs, it will return the tipset
+//     fallbackDelay epochs behind the chain head (-1).
+func (tsr *tipSetResolver) loadFinalizedTipSet(ctx context.Context, fallbackDelay abi.ChainEpoch) (*types.TipSet, error) {
+	head := tsr.loader.GetHeaviestTipSet()
+	latestHeight := head.Height() - 1 // always one behind for Eth compatibility due to deferred execution
+	fallbackHeight := latestHeight - fallbackDelay
+
+	f3TipSet, err := tsr.maybeF3FinalizedTipSet(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if f3TipSet != nil && f3TipSet.Height() > fallbackHeight {
+		// return the parent of the finalized tipset, which will be >= fallbackHeight
+		return tsr.loader.LoadTipSet(ctx, f3TipSet.Parents())
+	} // else F3 is disabled, not running, or behind the default safe or finalized height
+
+	ts, err := tsr.loader.GetTipsetByHeight(ctx, fallbackHeight, head, true)
+	if err != nil {
+		return nil, xerrors.Errorf("loading tipset at height %v: %w", fallbackHeight, err)
+	}
+	return ts, nil
+}
+
+// maybeF3FinalizedTipSet returns the latest F3 finalized tipset if F3 is enabled and the latest
+// certificate indicates that the tipset is finalized on top of the EC chain. If F3 is disabled or
+// or the current manifest is not finalizing tipsets on top of the EC chain, or we can't load a
+// certificate with a finalized tipset, it returns nil. This method will only return errors that
+// arise from dealing with the chain store, not from F3.
+func (tsr *tipSetResolver) maybeF3FinalizedTipSet(ctx context.Context) (*types.TipSet, error) {
+	// TODO: switch the order of GetManifest and GetLatestCert the former won't panic on
+	// a fresh F3 instance: https://github.com/filecoin-project/lotus/issues/12772
+	cert, err := tsr.f3.GetLatestCert(ctx)
+	if err != nil {
+		if !errors.Is(err, api.ErrF3Disabled) {
+			log.Warnf("loading latest F3 certificate: %s", err)
+		}
+		return nil, nil
+	}
+	if cert == nil {
+		return nil, nil
+	}
+
+	if manifest, err := tsr.f3.GetManifest(ctx); err != nil {
+		log.Warnf("loading F3 manifest: %s", err)
+		return nil, nil
+	} else if !manifest.EC.Finalize {
+		// F3 is not finalizing tipsets on top of EC, ignore it
+		return nil, nil
+	}
+
+	if len(cert.ECChain) == 0 || cert.ECChain.IsZero() {
+		// defensive; finality certs are supposed to have a guarantee of at least one tipset
+		return nil, nil
+	}
+
+	tsk, err := types.TipSetKeyFromBytes(cert.ECChain.Head().Key)
+	if err != nil {
+		return nil, xerrors.Errorf("decoding tipset key reported by F3: %w", err)
+	}
+
+	finalizedTipSet, err := tsr.loader.LoadTipSet(ctx, tsk)
+	if err != nil {
+		return nil, xerrors.Errorf("loading tipset reported as finalized by F3 %s: %w", tsk, err)
+	}
+
+	return finalizedTipSet, nil
+}
+
+func (tsr *tipSetResolver) resolveEthBlockNumberSelector(ctx context.Context, selector string, strict bool) (*types.TipSet, error) {
+	var num ethtypes.EthUint64
+	if err := num.UnmarshalJSON([]byte(`"` + selector + `"`)); err != nil {
+		return nil, xerrors.Errorf("cannot parse block number: %v", err)
+	}
+
+	head := tsr.loader.GetHeaviestTipSet()
+	if abi.ChainEpoch(num) > head.Height()-1 {
+		return nil, xerrors.New("requested a future epoch (beyond 'latest')")
+	}
+
+	ts, err := tsr.loader.GetTipsetByHeight(ctx, abi.ChainEpoch(num), head, true)
+	if err != nil {
+		return nil, xerrors.Errorf("cannot get tipset at height: %v", num)
+	}
+
+	if strict && ts.Height() != abi.ChainEpoch(num) {
+		return nil, api.NewErrNullRound(abi.ChainEpoch(num))
+	}
+
+	return ts, nil
+}
--- a/itests/eth_api_test.go
+++ b/itests/eth_api_test.go
@@ -13,6 +13,9 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-f3/certs"
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/go-f3/manifest"
 	"github.com/filecoin-project/go-jsonrpc"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
@@ -22,6 +25,7 @@ import (
 
 	"github.com/filecoin-project/lotus/build/buildconstants"
 	"github.com/filecoin-project/lotus/chain/actors/policy"
+	f3mock "github.com/filecoin-project/lotus/chain/lf3/mock"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/chain/types/ethtypes"
 	"github.com/filecoin-project/lotus/chain/wallet/key"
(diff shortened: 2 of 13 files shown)


itests/eth_api_test.go
```diff
@@ -396,42 +399,102 @@ func TestNetVersion(t *testing.T) {
...
+		f3Enabled              bool
+		f3DelayEpochs          int
+		expectedSafeEpoch      int
+		expectedFinalizedEpoch int
+	}{
+		{
+			name:                   "f3 disabled",
+			f3Enabled:              false,
+			expectedSafeEpoch:      31,
+			expectedFinalizedEpoch: 901,
```
rvagg, line 414: this is the baseline that we want to keep stable; it's the behaviour of the original test so we're essentially asserting that without f3 involved we're replicating the existing behaviour and not changing anything

itests/eth_api_test.go
```diff
@@ -396,42 +399,102 @@ func TestNetVersion(t *testing.T) {
...
+			f3DelayEpochs:          5,
+			expectedSafeEpoch:      6,
+			expectedFinalizedEpoch: 6,
+		},
+		{
+			name:                   "f3 enabled, far from head",
+			f3Enabled:              true,
+			f3DelayEpochs:          1000,
+			expectedSafeEpoch:      31,
+			expectedFinalizedEpoch: 901,
```
rvagg, line 428: back to original behaviour

itests/eth_api_test.go
```diff
@@ -396,42 +399,102 @@ func TestNetVersion(t *testing.T) {
...
+				{"finalized", abi.ChainEpoch(tc.expectedFinalizedEpoch)},
+			} {
+				t.Run(query.param, func(t *testing.T) {
+					head, err := client.ChainHead(ctx)
+					req.NoError(err)
+
+					var blk ethtypes.EthBlock
+					for {
+						// loop here until we get all of the operations performed in a single epoch, if the
+						// operations span multiple epochs our numbers will be off
```
rvagg, line 468: how we have to make things pseudo-atomic in itests - we need it all to happen within a single epoch but don't have a good mechanism to force it; thankfully it's not a big deal to just keep on retrying until beforeEpich==afterEpoch

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("loading tipset reported as finalized by F3 %s: %w", tsk, err)
+	}
+
+	return finalizedTipSet, nil
+}
+
+func (tsr *tipSetResolver) resolveEthBlockNumberSelector(ctx context.Context, selector string, strict bool) (*types.TipSet, error) {
+	var num ethtypes.EthUint64
+	if err := num.UnmarshalJSON([]byte(`"` + selector + `"`)); err != nil {
```
rvagg, line 145: most of the hard work for this function happen in `EthUint64#UnmarshalJSON`: https://github.com/filecoin-project/lotus/blob/dbf41383b37f1155dd0a28b10da2909226adb171/chain/types/ethtypes/eth_types.go#L51-L75

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	ts, err := tsr.loader.GetTipsetByHeight(ctx, abi.ChainEpoch(num), head, true)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get tipset at height: %v", num)
+	}
+
+	if strict && ts.Height() != abi.ChainEpoch(num) {
+		return nil, api.NewErrNullRound(abi.ChainEpoch(num))
+	}
```
rvagg, line 161: Some of our APIs can't be sloppy and if you request a specific block then you should get it; we've standardised returning an `api.ErrNullRound` that passes through the RPC layer in this case. It's a departure from expected Ethereum tooling behaviour simply because they don't have null rounds to deal with, but we've decided that it's better to return an error than an actual `null` or an alternative tipset, and consumers have adapted to this. There's an itest that asserts this behaviour holds: https://github.com/filecoin-project/lotus/blob/dbf41383b37f1155dd0a28b10da2909226adb171/itests/fevm_test.go#L1664

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, nil
+	}
+
+	cert, err := tsr.f3.GetLatestCert(ctx)
+	if err != nil {
+		log.Debugf("loading latest F3 certificate: %s", err)
+		return nil, nil
+	}
+
+	tsk, err := types.TipSetKeyFromBytes(cert.ECChain.Head().Key)
```
rvagg, line 130: in the F3 code there's this in `Head()`:

```
	if c.IsZero() {
		return nil
	}
```

I probably should account for this case, what does it mean for this to be the case? Is it a bootstrap thing?

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+
+		head := tsr.loader.GetHeaviestTipSet()
+		latestHeight := head.Height() - 1 // always one behind for Eth compatibility due to deferred execution
+		defaultHeight := latestHeight - defaultDelay
+
+		if f3TipSet, err := tsr.getF3FinalizedTipSet(ctx); err != nil {
+			return nil, err
+		} else if f3TipSet != nil && f3TipSet.Height() > defaultHeight {
+			// return the parent of the finalized tipset (deferred execution, eth cares about t)
```
masih, line 100: Questions:
* we have already decremented the height by 1 in line 94; why get parent of tipset finalised by F3?
* I thought deferred execution in the context of eth only matters for `head`? I am not clear on this one, apologies.

masih (approved):
No blockers. I recommend getting error construction to be consistent. Left some comments on what might be surprising from DX/maintainer perspective.

chain/tsresolver/tipset_resolver.go
Kubuxu: Thinking about the consequences of deferred execution:
If tipset T is finalized, then the parent state root is finalized, as are the messages executed in the next tipset. However, that execution state, in theory, is not finalized.

So, both interpretations are valid.

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+
+	default:
+		// likely an 0x hex block number or a decimal block number
+		return tsr.resolveEthBlockNumberSelector(ctx, selector, strict)
+	}
+}
+
+func (tsr *tipSetResolver) getF3FinalizedTipSet(ctx context.Context) (*types.TipSet, error) {
+	if !tsr.f3.IsEnabled() {
+		return nil, nil
```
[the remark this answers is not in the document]
Kubuxu: It is internal call site, it essentially translates F3Disabled into, this is known and valid scenario so nothing went wrong.

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	if running, _ := tsr.f3.IsRunning(); !running {
+		return nil, nil
+	}
```
Kubuxu, line 122: We should have an additional check here for "is F3 running with finalization/static manifest".

chain/tsresolver/tipset_resolver.go
Kubuxu: We have a gpbft protocol case where the chain can be empty. Finality certs are always guaranteed to have at least one tipset in them.

itests/eth_api_test.go
```diff
@@ -396,42 +399,102 @@ func TestNetVersion(t *testing.T) {
...
+				{"latest", 1}, // head - 1
+				{"safe", abi.ChainEpoch(tc.expectedSafeEpoch)},
+				{"finalized", abi.ChainEpoch(tc.expectedFinalizedEpoch)},
+			} {
+				t.Run(query.param, func(t *testing.T) {
+					head, err := client.ChainHead(ctx)
+					req.NoError(err)
+
+					var blk ethtypes.EthBlock
+					for {
```
Kubuxu, line 466: nit: This is getting quite deep; I would suggest extracting it somehow.

chain/tsresolver/tipset_resolver.go
masih: @rvagg, here is my take on this comment:

> is F3 running with finalization

This one means if "finalization" is enabled in F3 manifest, which in Lotus lingo translates to "Should F3 checkpoint the finalized tipsets?". It is specified [here](https://github.com/filecoin-project/go-f3/blob/d4f080decf5cd19c037b736263aedbf589faf8bf/manifest/manifest.go#L151-L152).

If finalisation in disabled in F3, then the API should be using EC finality.

> /static manifest

The F3 manifest integration in lotus is dynamic by default but up to a point, beyond which it will revert exclusively to a static manifest that "activates" F3 as the consensus mechanism for the network with uniform parameters across all nodes and no longer controllable by the F3 engineering team.

In the static manifest finalisation is enabled. If we are in a state where F3 is running, and we are in static manifest (i.e. F3 is active) then use F3 to resolve a tipset's finality.


@Kubuxu please correct me if I have missed anything.

rvagg: OK, thanks! I the static/dynamic thing has more nuance than I was aware of, I probably should go read the spec to understand that a bit more deeply.

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+//     latest F3 finalized TipSet
+//   - "finalized": the TipSet with messages executed at least policy.ChainFinality (900) epochs ago
+//     or the latest F3 finalized TipSet
+//   - a decimal block number: the TipSet at the given height
+//   - a 0x-prefixed hex block number: the TipSet at the given height
+//
+// If a specific block number is specified and `strict` is true, an error is returned if the block
+// number resolves to a null round, otherwise in the case of a null round the first non-null TipSet
+// immediately before the null round is returned.
+func (tsr *tipSetResolver) ResolveEthBlockSelector(ctx context.Context, selector string, strict bool) (*types.TipSet, error) {
```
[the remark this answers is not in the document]
rvagg: I'm not sure I agree with this since it's not a closed set of possible values: it's an awkward union of a set of fixed strings (defined as constants in here for clarity) and the various string forms that can be decoded by `ethtypes.EthUint64`. This method is more like a "figure out what this string I've been given means" rather than "take one of these and transform it into one of those". Can you talk more about how you think it would help by defining it as a type?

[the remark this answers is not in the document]
masih: > "figure out what this string I've been given means"

I take this function to do more than that because it is exported? But I see the type is not exported. 

> how you think it would help by defining it as a type?

The function clearly looks for well known string values. Using a type is an executable definition of that intent for me as a reader of code.

If this not exported anywhere then this is bikeshedding. Marking as resolved as it is a non blocker for me and a matter of oppinion.

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (tsr *tipSetResolver) resolveEthBlockNumberSelector(ctx context.Context, selector string, strict bool) (*types.TipSet, error) {
+	var num ethtypes.EthUint64
+	if err := num.UnmarshalJSON([]byte(`"` + selector + `"`)); err != nil {
+		return nil, xerrors.Errorf("cannot parse block number: %v", err)
+	}
+
+	head := tsr.loader.GetHeaviestTipSet()
+	if abi.ChainEpoch(num) > head.Height()-1 {
+		return nil, errors.New("requested a future epoch (beyond 'latest')")
```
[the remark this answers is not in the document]
masih: I find this inconsistency unfortunate.

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+		// likely an 0x hex block number or a decimal block number
+		return tsr.resolveEthBlockNumberSelector(ctx, selector, strict)
+	}
+}
+
+func (tsr *tipSetResolver) getF3FinalizedTipSet(ctx context.Context) (*types.TipSet, error) {
+	if !tsr.f3.IsEnabled() {
+		return nil, nil
+	}
+	if running, _ := tsr.f3.IsRunning(); !running {
```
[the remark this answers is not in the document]
rvagg: 👍

I see this now on the godocs:

> IsRunning returns true if gpbft is running Used mainly for testing purposes

Does that mean it doesn't have much utility after launch? Should an API consumer just be relying on error types because "it's not running" is an abnormal state?

chain/tsresolver/tipset_resolver.go
rvagg: 1. I think there is a calculation error here with comparing to a value that's already decremented
2. The tipset-1 thing is how we've baked in this behaviour to the existing Eth APIs, but I've been wondering whether F3 might be a chance to revisit current practice for `safe` and `finalized`. I'm going to have to go back and do a quick audit of the APIs that consume this to figure out what guarantees we need and whether the fixed list of messages is enough, or whether we need them to have been executed. Events are important for many of these APIs and they are attached to the next tipset, which is why head-1 is an important case.

chain/tsresolver/tipset_resolver.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, fmt.Errorf(`block param "%s" is not supported`, EthBlockSelectorEarliest)
+
+	case EthBlockSelectorPending:
+		return tsr.loader.GetHeaviestTipSet(), nil
+
+	case EthBlockSelectorLatest:
+		// head - 1 because we're always one behind for Eth compatibility due to deferred execution
+		return tsr.loader.LoadTipSet(ctx, tsr.loader.GetHeaviestTipSet().Parents())
+
+	case EthBlockSelectorSafe, EthBlockSelectorFinalized:
```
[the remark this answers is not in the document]
rvagg: Current form is a result of refactoring from something that did separate it out more!

But the fact that you're flagging this means it probably doesn't read as good as it _feels_ to me post-refactor, so I'll have another shuffle and see if I can make it cleaner.

chain/tsresolver/tipset_resolver.go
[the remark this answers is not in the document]
rvagg: I'd personally prefer to do away with `xerrors`, but I'd also like the stdlib to be better at dealing with chained errors. Hence, I'm mostly at peace with the current half-way state.

chain/tsresolver/tipset_resolver.go
rvagg: OK, went digging to understand a bit more of the dynamic/static situation and ended up having to dig into code. From what I can tell, there's no clear way of telling that we're not operating under a dynamic manifest, so the only check we can do here (the only check we need?) is to ask whether `manifest.EC.Finalize` is `true`, and if it's not, then ignore F3.

Does that sound right?

chain/tsresolver/tipset_resolver.go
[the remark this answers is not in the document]
rvagg: I ruminated and yielded, exclusively `xerrors` now

chain/tsresolver/tipset_resolver.go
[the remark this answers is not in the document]
rvagg: it is exported on the interface and it's used as exported

I think there's a refactor here that will align this better, we have a `EthBlockNumberOrHash` that _should_ have served this purpose but it looks like it wasn't properly integrated when `"finalized"` was originally introduced, so repurposing it for this is a bit awkward and may break some things; I don't want to do that right now because it's not at all simple.

Opened an issue describing the shape of the problem here: https://github.com/filecoin-project/lotus/issues/12766

Thanks for raising.

chain/tsresolver/tipset_resolver.go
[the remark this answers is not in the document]
rvagg: I have a follow-up question to this, will comment inline again on the latest iteration

chain/tsresolver/tipset_resolver.go
rvagg: I'm going to defer (har har) dealing with the -1 question for `"safe"` and `"finalized"` for a possible follow-up PR, since this is only replicating existing behaviour. I've dealt with the other concern (the decrement was fine, but I've tried to make the code more clear and added a comment)

chain/tsresolver/tipset_resolver.go
rvagg: I believe this is all handled now but I wouldn't mind a sanity check from both of you (@masih & @Kubuxu) in this function in particular.

It's now called `maybeF3FinalizedTipSet` to indicate that you may not get something useful from it. If F3 can't help us then we just move on to the defaults.

1. We're not going to error here unless it's from dealing with the chain store, an error from F3 is worthy of a WARN log, and only if the error isn't `ErrF3Disabled`.
2. If `!manifest.EC.Finalize`, we don't bother continuing
3. Some basic defensive checks on the certificate
4. All good? load and return that tipset, it's the latest that F3 has to offer.

rvagg: no, sadly this doesn't work:

```
runtime error: invalid memory address or nil pointer dereference
github.com/filecoin-project/go-jsonrpc.doCall.func1
	/home/runner/go/pkg/mod/github.com/filecoin-project/go-jsonrpc@v0.7.0/handler.go:237
runtime.gopanic
	/home/runner/work/_tool/go/1.22.7/x64/src/runtime/panic.go:770
runtime.panicmem
	/home/runner/work/_tool/go/1.22.7/x64/src/runtime/panic.go:261
runtime.sigpanic
	/home/runner/work/_tool/go/1.22.7/x64/src/runtime/signal_unix.go:881
github.com/filecoin-project/lotus/chain/lf3.(*F3).GetManifest
	/home/runner/work/lotus/lotus/chain/lf3/f3.go:205
github.com/filecoin-project/lotus/chain/tsresolver.(*tipSetResolver).maybeF3FinalizedTipSet
	/home/runner/work/lotus/lotus/chain/tsresolver/tipset_resolver.go:131
```

So `GetManifest` as it exists today isn't safe to call when it hasn't got a manifest. This is in an itest, letting it load the default `F3` module. I think maybe it needs to have a check in here for `nil`, but what should it return in that case? https://github.com/filecoin-project/lotus/blob/a46832f330215e047302d607753a59299a51eb45/chain/lf3/f3.go#L203-L206

rvagg: So we're querying it before the manifest provider can give it a manifest, I think `F3#GetManifest` should either return `nil, nil` or an error I could `errors.Is` on to deal with this case.

rvagg: Yeah, it initialises with its `manifest` `atomic.Pointer` empty, and it only gets set after `Start()` is called and `manifestProvider.ManifestUpdates()` successfully sends it a new manifest. https://github.com/filecoin-project/go-f3/blob/6c3d2deca27e361b68680156eef99271d82c29a6/f3.go#L170

rvagg:
FYI this is getting more of a refactor. It's been suggested that (for now at least), we should only expose the differences on a /v2/ API so /v1/ users don't experience any change even when F3 goes live so that it's an opt-in thing. I'll need to do some work to introduce the /v2/ in this PR and try and keep the code duplication minimal, which will be a challenge.

BigLep:
@rvagg : given https://github.com/filecoin-project/lotus/pull/12762#issuecomment-2538687834 and that I don't think this is active for you currently, I"m going to move this back to draft so it's clear that it's not ready for review.  Please correct me if that's wrong.

rvagg:
uh. yeah, and I moved a ton of refactoring into a PR that has since landed, so this one is now out of date and isn't even close to being landable! I'm going to have to close this and start again I think but I don't want to lose reference to it so I'll leave it open until I pick it up.

BigLep:
@rvagg : can this be closed in favor of https://github.com/filecoin-project/lotus/pull/13026 ?

Issue #12758, opened 2024-12-05
New flaky: TestSyncCheckpointHead


rvagg:
Likely related to https://github.com/filecoin-project/lotus/pull/12747

https://github.com/filecoin-project/lotus/actions/runs/12173107368/job/33952946823#step:9:18130

rvagg:
Another one: https://github.com/filecoin-project/lotus/actions/runs/12171328368/job/33948060614#step:9:17726

rvagg:
https://github.com/filecoin-project/lotus/actions/runs/12173107368/job/33956317854#step:9:18206

seems to be _very_ flaky

I can get it to fail locally just running it once `go test ./chain/ -run TestSyncCheckpointHead`

PR #12763, opened 2024-12-05
fix(sync): mark blocks which diverge from checkpoint as bad

Kubuxu:
Resolves flake in `TestSyncCheckpointHead`

Not certain why this was flaky, it would mean that sometimes the forking
tipset would not hit this case.

Resolves: #12758


rvagg (approved):
nice

PR #12768, opened 2024-12-10
chore: remove IPNI advertisement relay over pubsub via Lotus node

masih:
## Proposed Changes
The initial implementation of IPNI used GossipSub as a way to propagate IPNI advertisement chain. To do this the propagation had to be relayed through the Lotus node due to strict Filecoin GossipSub validation rules.

Since then IPNI has moved on to roll out its own sync protocol that works over HTTP, and HTTP-over-libp2p. This move has been the official way of advertising content to IPNI federation over a year now.

Therefore, remove the ad relay over pubsub via Lotus node as it is now considered to have reached its EOL as a mechanism for advertising to IPNI.

## Checklist

- [x] CI is green


willscott (approved):

rvagg (approved):
We're confident that there are no potentially usable (in-use) versions of Boost that may be still reliant on this?

masih:
> We're confident that there are no potentially usable (in-use) versions of Boost that may be still reliant on this?

If there is, I think we want to break them. It has been a year and the time has come to remove support for the deprecated protocol.

willscott:
@simlecode  correct

simlecode:
@willscott  Does this mean that we can't use [gossip pubsub](https://github.com/ipni/index-provider/blob/main/engine/engine.go#L222) to announce deals?

willscott:
That is correct. I believe this feature was considered unmaintained by lotus and boost maintainers, and as such could not be considered reliable or a method by which publishers were able to have confidence that announcements actually made it to indexers.

masih:
@simlecode Just to clarify what "announce deals" means:

* The term "announce" or "announcement" has an explicit meaing in IPNI: it is a simple HTTP PUT call to an indexer to say "Hey there is something new that you can get from me". It is only a notification. You cannot send that over libp2p today. It's only supported over HTTP. If there is enough demand we can consider adding it as it's a small body of work to do so. Please capture an issue at https://github.com/ipni/specs


* There is then "Advertisement Publishing" which is what a SP does to make the chain of IPNI ads accessible to an indexer. This is a pull model where an indexer (having leant from announcements that there is something new) reaches out to a SP to pull the actual ads that contain hashes, provider info and so on. The advertisement publishing is supported over both HTTP and libp2p today. To further clarify, this is a feature that lives in SP client software: Boost, Curio and so on, and no longer lives in Lotus.

PR #12769, opened 2024-12-10
chore: use errors.New to replace fmt.Errorf with no parameters

ChengenH:
1. use errors.New to replace fmt.Errorf with no parameters
2. Fix spelling errors in words

### DIFF
--- a/cli/chain.go
+++ b/cli/chain.go
@@ -1252,11 +1252,11 @@ var ChainExportRangeCmd = &cli.Command{
 		}
 
 		if head.Height() < tail.Height() {
-			return errors.New("Height of --head tipset must be greater or equal to the height of the --tail tipset")
+			return errors.New("height of --head tipset must be greater or equal to the height of the --tail tipset")
 		}
 
 		if !cctx.Bool("internal") {
-			return errors.New("Non-internal exports are not implemented")
+			return errors.New("non-internal exports are not implemented")
 		}
 
 		err = api.ChainExportRangeInternal(ctx, head.Key(), tail.Key(), lapi.ChainExportConfig{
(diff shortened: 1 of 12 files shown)


rvagg (changes requested):
Thanks for the contribution. As nice as this is, it's not a hard standard that we enforce in this codebase, in fact we use `xerrors` more than `errors`, but even that's not consistent _except_ where we're wrapping errors.

This PR just opens the door to a larger change-set that I don't think I want to encourage as small PRs because it'll just be highlighting the inconsistency rather than solving it:

```
$ git grep -E 'fmt\.Errorf\("[^%]*"\)' -- "*.go" | wc -l
556
```

However, the changes other than the `errors.New()` edits in here are OK, would you mind scaling it back to non-capital letter error string starts and the typo fixes?

ChengenH:
> Thanks for the contribution. As nice as this is, it's not a hard standard that we enforce in this codebase, in fact we use `xerrors` more than `errors`, but even that's not consistent _except_ where we're wrapping errors.感谢您的贡献。尽管这很好，但我们在这个代码库中强制执行它并不是一个硬性标准，事实上我们使用 `xerrors` 的次数多于`错误`，但即使这样也不一致，_除非_我们包装错误。
> 
> This PR just opens the door to a larger change-set that I don't think I want to encourage as small PRs because it'll just be highlighting the inconsistency rather than solving it:这个 PR 只是为更大的变更集打开了大门，我认为我不想作为小 PR 来鼓励它，因为它只会突出不一致而不是解决它：
> 
> ```
> $ git grep -E 'fmt\.Errorf\("[^%]*"\)' -- "*.go" | wc -l
> 556
> ```
> 
> However, the changes other than the `errors.New()` edits in here are OK, would you mind scaling it back to non-capital letter error string starts and the typo fixes?但是，除错误外的更改`。这里的 New（）` 编辑是可以的，您介意将其缩小到非大写字母错误字符串开头和拼写错误修复吗？

image
I think some specific terms can start with a capital letter.

rvagg:
Yes, they can, I was just suggesting the ones you found in cli/chain.go

ChengenH:
> Yes, they can, I was just suggesting the ones you found in cli/chain.go

I understand, I will close this PR and open a new one.

PR #12770, opened 2024-12-10
feat: optimize ForEachClaim to return only eligible claims

Kubuxu:
Avoids double AMT traversal

Depends on: https://github.com/filecoin-project/go-state-types/pull/333


rvagg (approved):
a comment on this method might be helpful? up to you tho

go.mod
```diff
@@ -52,7 +52,7 @@ require (
 	github.com/filecoin-project/go-jsonrpc v0.7.0
 	github.com/filecoin-project/go-padreader v0.0.1
 	github.com/filecoin-project/go-paramfetch v0.0.4
-	github.com/filecoin-project/go-state-types v0.16.0-rc1
+	github.com/filecoin-project/go-state-types v0.16.0-rc1.0.20241210151506-27c2d581aa19
```
masih, line 55: We need to bubble up a release before merging this.

masih (approved):

go.mod
masih: Looks like the release is already there (rc2). Great. We just need this updated and we are good to go I think 👍

PR #12774, opened 2024-12-12
feat(ci): go list check versions available for untagged dependencies

rvagg:
This is on top of https://github.com/filecoin-project/lotus/pull/12740, I'm experimenting with using `go list -versions` to inspect `v0.0.0` dependencies (as per discussion with @BigLep). We should be able to say "you're using an untagged version, but it has a version, that's naughty.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -78,7 +78,7 @@ require (
 	github.com/gorilla/websocket v1.5.3
 	github.com/gregdhill/go-openrpc v0.0.0-20220114144539-ae6f44720487
 	github.com/hako/durafmt v0.0.0-20200710122514-c0fb7b4da026
-	github.com/hannahhoward/go-pubsub v0.0.0-20200423002714-8d62886cc36e
+	github.com/hannahhoward/go-pubsub v0.0.0-20200423002714-8d62886cc36e // dependency-check-ignore: TODO: needs upgrading to tagged version
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/hashicorp/golang-lru/arc/v2 v2.0.7
 	github.com/hashicorp/golang-lru/v2 v2.0.7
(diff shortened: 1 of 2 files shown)


rvagg:
OK, this works now and reports:

```
v0.0.0 dependencies with available tags:
github.com/hannahhoward/go-pubsub
howett.net/plist
```

first one is a direct dependency and the second is an indirect; I've added an ignore to both, pubsub could be upgraded but I won't do it here, the other one is used by github.com/elastic/go-sysinfo which has an update but still doesn't use a tagged version of plist.

go.mod
```diff
@@ -352,7 +352,7 @@ require (
 	gopkg.in/cheggaaa/pb.v1 v1.0.28 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	howett.net/plist v0.0.0-20181124034731-591f970eefbb // indirect
+	howett.net/plist v0.0.0-20181124034731-591f970eefbb // indirect; indirect dependency-check-ignore: required by github.com/elastic/go-sysinfo
```
BigLep, line 355:
```suggestion
	howett.net/plist v0.0.0-20181124034731-591f970eefbb // indirect; dependency-check-ignore: required by github.com/elastic/go-sysinfo
```

extra `indirect` is redundant?

BigLep (approved):

go.mod
rvagg: yeah, that can be fixed, but this PR is full of conflicts now the base has shifted so I'm going to leave it to @galargh to resolve if he pulls this in, otherwise he can write it the way he wants

BigLep:
I'll handle merge conflicts this week per https://github.com/filecoin-project/lotus/pull/12740#issuecomment-2542081817

PR #12777, opened 2024-12-12
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters

### DIFF
--- a/api/cbor_gen.go
+++ b/api/cbor_gen.go
@@ -3,6 +3,7 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"math"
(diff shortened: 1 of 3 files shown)


rjan90:
Hey @RiceChuan! Thanks for the contribution - I think it is worth highlighting this comment made by Rod on similar changes earlier in the week in this PR: https://github.com/filecoin-project/lotus/pull/12769#pullrequestreview-2494256469

> As nice as this is, it's not a hard standard that we enforce in this codebase, in fact we use xerrors more than errors, but even that's not consistent except where we're wrapping errors.
> 
> This PR just opens the door to a larger change-set that I don't think I want to encourage as small PRs because it'll just be highlighting the inconsistency rather than solving it:

As there are no other changes in this PR then the `errors.New()` changes, I will close this issue.

PR #12778, opened 2024-12-12
feat(gateway): add F3GetCertificate & F3GetLatestCertificate to gateway

rvagg:
Only applies a rate limiting cost of "basic" which is 1 token because I believe these are pretty efficient calls.

Doesn't apply a lookback time limit, you can fetch any arbitrary instance's certificate from a public gateway. go-f3's "certstore" is keyed by instance ID which is what's being queried here.

### DIFF
--- a/itests/gateway_test.go
+++ b/itests/gateway_test.go
@@ -558,3 +558,35 @@ func TestStatefulCallHandling(t *testing.T) {
 		req.ErrorContains(err, "filter not found")
 	}
 }
+
+func TestGatewayF3(t *testing.T) {
+	// Test that disabled & not-running F3 calls properly error
+
+	kit.QuietMiningLogs()
+
+	t.Run("not running", func(t *testing.T) {
+		ctx := context.Background()
+		nodes := startNodes(ctx, t)
+
+		cert, err := nodes.lite.F3GetLatestCertificate(ctx)
+		require.ErrorContains(t, err, "F3 is not running")
+		require.Nil(t, cert)
+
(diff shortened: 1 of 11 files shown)


masih (approved):

rvagg:
Updated with an itest that makes sure these don't misbehave when it's not running or is disabled. It's a bit too complex to combine the awkward F3 itest setup with the awkward gateway itest setup to test it running. I started, but the F3 itest setup is already huge without pulling more bits in.

itests/gateway_test.go
```diff
@@ -558,3 +558,35 @@ func TestStatefulCallHandling(t *testing.T) {
 		req.ErrorContains(err, "filter not found")
 	}
 }
+
+func TestGatewayF3(t *testing.T) {
+	// Test that disabled & not-running F3 calls properly error
+
+	kit.QuietMiningLogs()
+
+	t.Run("F3 not running", func(t *testing.T) {
```
masih, line 567: Subjective: 
* Avoid including words that are inferred from the outer test name. In this case _I_ would drop `F3`
* Pick names such that when read in conjuction with outer test names they read like a user story (a poor man's lightweight behaviour driven test).
* Use `_` to separate context in which test being run to avoid test name conflict across tests in the same package, e.g. `TestGateway_F3`

masih (approved):

itests/gateway_test.go
rvagg: 👍 changed inner test names, but opted to not use `_` because it's not the style of the current file or many of the existing itests

Issue #12772, opened 2024-12-11
F3#GetManifest can panic


rvagg:
https://github.com/filecoin-project/lotus/blob/9bfd0b4295f7d51dbe306c5d20a0502705918056/chain/lf3/f3.go#L187-L189

It's possible currently for the inner call, `f3.F3#Manifest()` to return a `nil`. Apparently this shouldn't be possible, it should always fall back to the static manifest which is available at compile time: https://github.com/filecoin-project/lotus/pull/12762/files#r1878043815

Unfortunately the current architecture of go-f3's `F3` is that it's initialised with a `nil` manifest, then it has to first `Start()` and then has to get its first update from the `ManifestProvider` before it's set. So calling `GetManifest()` here before that happens can result in a panic, which can happen in itests when F3 is not properly set up.

This should either be guarded here, or fixed in go-f3.

masih:
Note, manifest can only be nil in dynamic manifest provider: the mechanism that makes passive testing possible.  In dynamic manifest provider, there is an option which sets the _initial manifest_, i.e. the one to use if no other manifest is broadcasted via the manifest server. I see [this option](https://github.com/filecoin-project/lotus/blob/be74be07e62554464f9307443bbb84df1635b895/chain/lf3/manifest.go#L54) is indeed set in Lotus _if_ static manifest is non-nil.

My question is then: under what condition static manifest is set to nil? @rvagg do you have a test that reproduces this in a tagged release?

rvagg:
```go
func TestManifestNil(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
	defer cancel()

	client, _, ens := kit.EnsembleMinimal(t, kit.MockProofs())
	ens.InterconnectAll().BeginMining(2 * time.Millisecond)
	ens.Start()

	client.WaitTillChain(ctx, kit.HeightAtLeast(10))
	manifest, err := client.F3GetManifest(ctx)

	require.NoError(t, err)
	require.NotNil(t, manifest)
}
```

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x21e667c]

goroutine 31 [running]:
testing.tRunner.func1.2({0x45de640, 0x9df7370})
        /snap/go/10751/src/testing/testing.go:1631 +0x24a
testing.tRunner.func1()
[...]

PR #12781, opened 2024-12-13
fix(f3): test API during inactive-F3 modes and make consistent and safe

rvagg:
Closes: https://github.com/filecoin-project/lotus/issues/12772

* Added an itest to test all of the F3 API method returns when both explicitly disabled F3 and F3 not properly started
* Fixed the `GetManifest` panic
* Made error messages more consistent (go-f3 has error messages that start with `"F3` but we have some `"f3`)
  * `GetManifest` now copies the `"no known network manifest"` error message that `GetF3PowerTable` returns (from inside go-f3) during the not-running state
  * `client.F3GetOrRenewParticipationTicket` now returns `nil` as the zero value of `api.F3ParticipationTicket` which is a slice, like the inner `lf3/leaser.getOrRenewParticipationTicket` returns, so we don't get `nil` in one case and `[]` in another

One minor special case is `F3GetECPowerTable` returns a sensible value when F3 isn't ready, so it's only when it's properly disabled that it gives an error. This seems reasonable even if it feels mildly inconsistent.

### DIFF
--- a/api/api_errors.go
+++ b/api/api_errors.go
@@ -98,7 +98,7 @@ func (ErrActorNotFound) Error() string { return "actor not found" }
 
 type errF3Disabled struct{}
 
-func (errF3Disabled) Error() string { return "f3 is disabled" }
+func (errF3Disabled) Error() string { return "F3 is disabled" }
 
 type errF3ParticipationTicketInvalid struct{}
 
@@ -124,7 +124,7 @@ func (errF3ParticipationTicketStartBeforeExisting) Error() string {
 
 type errF3NotReady struct{}
 
-func (errF3NotReady) Error() string { return "f3 isn't yet ready to participate" }
+func (errF3NotReady) Error() string { return "F3 isn't yet ready to participate" }
 
 // ErrExecutionReverted is used to return execution reverted with a reason for a revert in the `data` field.
 type ErrExecutionReverted struct {
--- a/api/api_errors.go
+++ b/api/api_errors.go
@@ -98,7 +98,7 @@ func (ErrActorNotFound) Error() string { return "actor not found" }
 
 type errF3Disabled struct{}
 
-func (errF3Disabled) Error() string { return "F3 is disabled" }
+func (errF3Disabled) Error() string { return "f3 is disabled" }
 
 type errF3ParticipationTicketInvalid struct{}
 
@@ -124,7 +124,7 @@ func (errF3ParticipationTicketStartBeforeExisting) Error() string {
 
 type errF3NotReady struct{}
 
-func (errF3NotReady) Error() string { return "F3 isn't yet ready to participate" }
+func (errF3NotReady) Error() string { return "f3 isn't yet ready to participate" }
 
 // ErrExecutionReverted is used to return execution reverted with a reason for a revert in the `data` field.
 type ErrExecutionReverted struct {
--- a/chain/lf3/f3.go
+++ b/chain/lf3/f3.go
@@ -194,7 +194,7 @@ func (fff *F3) GetManifest(ctx context.Context) (*manifest.Manifest, error) {
 	}
 	cert0, err := fff.inner.GetCert(ctx, 0)
 	if err != nil {
-		return m, xerrors.Errorf("getting certificate for instance 0: %w", err)
+		return m, nil // return manifest without power table
 	}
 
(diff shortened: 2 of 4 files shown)


api/api_errors.go
```diff
@@ -98,7 +98,7 @@ func (ErrActorNotFound) Error() string { return "actor not found" }
 
 type errF3Disabled struct{}
 
-func (errF3Disabled) Error() string { return "f3 is disabled" }
+func (errF3Disabled) Error() string { return "F3 is disabled" }
```
masih, line 101: Capitalised error message is against the common Go practices that I am aware of.

I don't know how consistently this is followed in Lotus repo. But I can see even in this PR, error message casing is inconsistent.

chain/lf3/f3.go
```diff
@@ -184,20 +184,23 @@ func (fff *F3) GetLatestCert(ctx context.Context) (*certs.FinalityCertificate, e
 	return fff.inner.GetLatestCert(ctx)
 }
 
-func (fff *F3) GetManifest(ctx context.Context) *manifest.Manifest {
+func (fff *F3) GetManifest(ctx context.Context) (*manifest.Manifest, error) {
 	m := fff.inner.Manifest()
+	if m == nil {
+		return nil, xerrors.Errorf("no known network manifest")
```
masih, line 190: * `xerrors.New`?
* Do we want a specific type for this error?

masih (approved):

api/api_errors.go
rvagg: As per comment at the top of the PR, this is to align with go-f3 (see https://github.com/filecoin-project/go-f3/blob/1b17c6c40971e9ff408afa8b8cd7fb90f79e587c/f3.go#L119-L127); I wouldn't mind keeping these lower-case but it's harder to ix in go-f3 than just make it aligned here, and capitalising acronyms is OK and this is kind of an acronym so we're in one of those cases where I reckon it's OK to go either way

rvagg: And re lotus conventions - we pretty consistently stick to the lower-case error convention; there's only a few stray cases left in the repo

chain/lf3/f3.go
masih: The sub API level errors don't matter really. But at API level I think Lotus should explicitly define its own error types, independent of its dependencies.

For this specific case the net effect is that the participant will retry anyway so up to you if you'd like to mae this an explicit error type or leave it as is. Either works i think.

api/api_errors.go
masih: FWIW i have captured https://github.com/filecoin-project/go-f3/issues/795

rvagg: ok, if you're making as a bug over there I'll lower-case it here and then we can have them aligned when we eventually get an update 👍

chain/lf3/f3.go
rvagg: Raised in https://github.com/filecoin-project/go-f3/issues/795#issuecomment-2541104898 that I think it would be good to be able to type check against so we could transform them more reliably for Lotus. It would be pretty nice if we could do `errors.Is` checks on all of the errors in this list but I don't really want to achieve that by string comparisons of `Error()` values from what go-f3 passes on.

rvagg:
@masih can you have a look at https://github.com/filecoin-project/lotus/blob/2666c3e40f23adcc5ae236c3af7808acc99a36e1/itests/f3_test.go#L355-L357 please - this is now failing. Previous to this PR, `F3GetManifest` is only going to error if F3 is fully disabled, so I'm not sure what this error check was for. Am I safe to remove the error check and just return `false` when the manifest is `nil`? That would be equivalent to the existing behaviour but I'm not sure whether I'm exposing something of interest in the process of doing this.

masih:
> is only going to error if F3 is fully disabled, so I'm not sure what this error check was for. Am I safe to remove the error check and just return `false` when the manifest is `nil`?

Yes, returning false on either error or mismatching Manifest should be fine.

Looks like the cause of tests failing is mismatching manifest though.

rvagg:
Ah, got it, I accidentally introduced a new error condition in `GetManifest` of "can't get latest certificate so can't return you a manifest" .. which isn't appropriate if we have a manifest but haven't quite started. Nice to have test coverage for it!

PR #12723, opened 2024-11-25
misc(f3): tool for explicit power table generation

Kubuxu:

PR #12782, opened 2024-12-13
fix(shed): lotus-shed f3 certs command -n arg not -N

rvagg:
Ref: https://github.com/filecoin-project/lotus/pull/12723

This is my bad, I made a suggestion to change `"N"` to `"n"` in the cmd args list but not the usage of it. @Kubuxu which way do you want to go with this? Are you happy to switch all to`"n"`?


masih (approved):

Kubuxu:
No preference really

PR #12796, opened 2024-12-19
chore(eth): refactor eth API module into separate pieces in new pkg

rvagg:
As I start on the `/v2` API the eth module is my first hurdle: I was `/v1` to remain as it is now, but `/v2` to have the same Eth endpoints _but_ be F3 aware, so `"finalized"` and `"safe"` get adjusted according to F3. Otherwise they should be the same. Because almost everything is in a single eth.go, it makes it a bit hard to do a nice pluggable solution where I can reuse most of the code and just have replacements for a resolver. So I decided to rip apart the eth.go monolith into smaller components, in their own package, with interfaces for their dependencies rather than the concrete types.

There's very little change in the code even though the diff is big, it's mostly just moving things around. I took a few opportunities to clean up some awkward messes, and I was forced to do some new mildly awkward new things.

* `github.com/filecoin-project/lotus/node/impl/eth` is a new package, with what used to be in `github.com/filecoin-project/lotus/node/impl/full/eth*.go` but logically split up.
* The new eth package requires on fewer pieces of the rest of the codebase, in particular interfaces have been introduced for the main functional dependencies like `ChainStore` and `StateManager`.
* `github.com/filecoin-project/lotus/node/impl/gasutils` contains some components that were in `node/impl/full/gas.go` but are shared between Filecoin and Eth APIs.
* The construction and startup utilities in node/modules/ethmodule.go are now in node/modules/eth.go but have been trimmed because we need to do less manual wiring now.
* There's some pieces I haven't touched, like the filter stuff and subscription manager.
* Made consistent use of `xerrors` in the new eth package.


node/modules/eth.go
```diff
@@ -0,0 +1,249 @@
...
+
+				return *actor.DelegatedAddress
+			},
+
+			MaxFilterResults: cfg.MaxFilterResults,
+		}
+
+		params.Lifecycle.Append(fx.Hook{
+			OnStart: func(context.Context) error {
+				ev, err := events.NewEvents(ctx, &params.EventHelperAPI)
```
Stebalien, line 238: Can we rename this context to be `lctx` so it's clear that it's for lifetimes?

node/modules/eth.go
```diff
@@ -0,0 +1,249 @@
...
+	*stmgr.StateManager
+	EventHelperAPI
+	*messagepool.MessagePool
+	full.StateAPI
+	full.ChainAPI
+}
+
+func MakeEthEventsExtended(cfg config.EventsConfig, enableEthRPC bool) func(EthEventsParams) (eth.EthEventsExtended, error) {
+	return func(params EthEventsParams) (eth.EthEventsExtended, error) {
+		ctx := helpers.LifecycleCtx(params.MetricsCtx, params.Lifecycle)
```
Stebalien, line 112:
```suggestion
		lctx := helpers.LifecycleCtx(params.MetricsCtx, params.Lifecycle)
```

Stebalien (approved):

rvagg:
Additional fixes applied after a sync zoom chat with @Stebalien about this:

* Clarified naming - everything that defines an external API has an `API` suffix in the name, and I've removed that from some interfaces I was using to define internal things (like `ChainStore`). It's much clearer now when you view it through the lens of what's intended for RPC and what's not.
* Context naming - `lctx` for the lifecycle context on init.
* Added some docs around the `GatewayEthSend` special-case. But also, we might look at alternative ways to improve this in the future, perhaps with some more creative/newer fx use.
* Misc cleanups from things I noticed while walking back through the code, again.

Issue #10554, opened 2023-03-23
Import snapshot directly into hotstore for nodes running splitstore


ZenGround0:
### Checklist

- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Coldstore shouldn't track unwanted state in discard mode.  Today we waste ~100 G of disk space no matter what when syncing from snapshot 

### Describe the solution you'd like

import-snaphsot daemon flag should import the chain into the hotstore if the node is configured to run with splitstore.

Basically repo should create a blockstore using hotstore instead of universal store location if applicable.  We might want to make a new blockstore location that automatically does this or we could manually check the repo config during import.

### Describe alternatives you've considered

We could also make it easier to purge the coldstore of snapshot state by improving chain prune cold-compact but it would be better to prevent this problem in the first place than to keep it around and also create a solution

ZenGround0:
Looking at this more today the problem seems to be that 
1. Node startup depends on initial state in universal blockstore 
2. Splitstore startup depends on node startup

So we have to unravel these dependencies if we want to save the disk space

PR #12803, opened 2024-12-25
fix(splitstore): merry christmas lotus! Remove ~120 G from lotus datastore

ZenGround0:
## Related Issues

#10554 

## Proposed Changes

As part of the christmas spirit I wanted to do something that everyone in the network would see as a gift.  This year I decided to get us the gift of 100 fewer gigs in the lotus datastore

Basically we just hold onto two copies of the state.  One original in the cold store and one hot copy.  As far as I can tell there is zero reason to do this. 

## Additional Info

Somebody tell me if this breaks something?  I'm running pretty good on mainnet right now.  A little more testing to take this out of draft 

## Checklist

- [ ] CI is green


blockstore/splitstore/splitstore_warmup.go
```diff
@@ -136,14 +58,111 @@ func (s *SplitStore) doWarmup(curTs *types.TipSet) error {
...
 	// also save the compactionIndex, as this is used as an indicator of warmup for upgraded nodes
 	err = s.ds.Put(s.ctx, compactionIndexKey, int64ToBytes(s.compactionIndex))
 	if err != nil {
 		return xerrors.Errorf("error saving compaction index: %w", err)
 	}
-
 	return nil
 }
+
+// the legacy warmup procedure before we wrote snapshots directly to the hotstore
```
ZenGround0, line 71: TODO put this behind a splitstore config to support the now rare case where people want to start a hot store off of a cold store

node/repo/fsrepo.go
```diff
@@ -423,39 +426,59 @@ func (fsr *fsLockedRepo) Close() error {
...
+	}
+	readonly := fsr.readonly
+	opts, err := BadgerBlockstoreOptions(HotBlockstore, path, readonly)
+	if err != nil {
+		return nil, nil, err
+	}
+	bs, err := badgerbs.Open(opts)
+	if err != nil {
+		return nil, nil, err
+	}
```
ZenGround0, line 479: This change is probably too much of a hack and needs some more thought.  We probably want unified guarantees across domains.  Do we even need the one time call thing here for universal bstore?  Maybe we can get rid of that.  Alternatively maybe we can reuse our already opened hotstore when we start of the splitstore instead of closing it down first.

ZenGround0: Went with the second option which turns out to be quite clean

node/repo/interface.go
```diff
@@ -64,7 +64,7 @@ type LockedRepo interface {
 	// The supplied context must only be used to initialize the blockstore.
 	// The implementation should not retain the context for usage throughout
 	// the lifecycle.
-	Blockstore(ctx context.Context, domain BlockstoreDomain) (blockstore.Blockstore, error)
+	Blockstore(ctx context.Context, domain BlockstoreDomain) (blockstore.Blockstore, func() error, error)
```
rvagg, line 67: would be good to document what the returned `func` is in the comment, I had to go to the bottom of the only function that implements it to find the `closer`

node/repo/fsrepo.go
```diff
@@ -416,46 +419,88 @@ func (fsr *fsLockedRepo) Close() error {
...
-
-	fsr.bsOnce.Do(func() {
-		path := fsr.join(filepath.Join(fsDatastore, "chain"))
-		readonly := fsr.readonly
-
-		if err := os.MkdirAll(path, 0755); err != nil {
-			fsr.bsErr = err
-			return
-		}
+func (fsr *fsLockedRepo) Blockstore(ctx context.Context, domain BlockstoreDomain) (blockstore.Blockstore, func() error, error) {
```
rvagg, line 434: document the returned func here, or type it

node/repo/interface.go
rvagg: or type it, `type CloserFn func() error`

cmd/lotus-shed/splitstore.go
```diff
@@ -353,6 +355,40 @@ func deleteSplitstoreKeys(lr repo.LockedRepo) error {
...
+func printSplitstoreKeys(lr repo.LockedRepo) error {
+	ds, err := lr.Datastore(context.TODO(), "/metadata")
+	if err != nil {
+		return xerrors.Errorf("error opening datastore: %w", err)
+	}
+	if closer, ok := ds.(io.Closer); ok {
+		defer closer.Close() //nolint
+	}
+
+	res, err := ds.Query(context.Background(), query.Query{Prefix: "/splitstore"})
```
rvagg, line 367: @ZenGround0 can you point me to where this data is originally written? I can't find it. Or is this legacy data?

rvagg: oh, found them: https://github.com/filecoin-project/lotus/blob/0e4857a11ba57c8c29e27dc063f65da535bd5ff0/blockstore/splitstore/splitstore.go#L30-L51

cmd/lotus-shed/deal-label.go
```diff
@@ -52,7 +52,7 @@ var dealLabelCmd = &cli.Command{
 
 		defer lkrepo.Close() //nolint:errcheck
 
-		bs, err := lkrepo.Blockstore(ctx, repo.UniversalBlockstore)
+		bs, _, err := lkrepo.Blockstore(ctx, repo.UniversalBlockstore)
```
rvagg, line 55: should we be using the closer in all of these?

node/modules/blockstore.go
```diff
@@ -45,46 +40,32 @@ func DiscardColdBlockstore(lc fx.Lifecycle, bs dtypes.UniversalBlockstore) (dtyp
...
-
+			return closer()
+		},
+	})
 	return bs, nil
 }
 
-func SplitBlockstore(cfg *config.Chainstore) func(lc fx.Lifecycle, r repo.LockedRepo, ds dtypes.MetadataDS, cold dtypes.ColdBlockstore, hot dtypes.HotBlockstore) (dtypes.SplitBlockstore, error) {
-	return func(lc fx.Lifecycle, r repo.LockedRepo, ds dtypes.MetadataDS, cold dtypes.ColdBlockstore, hot dtypes.HotBlockstore) (dtypes.SplitBlockstore, error) {
+func SplitBlockstore(cfg *config.Chainstore) func(lc fx.Lifecycle, r repo.LockedRepo, ds dtypes.MetadataDS, cold dtypes.ColdBlockstore, hot dtypes.HotBlockstore) (dtypes.SplitBlockstore, func() error, error) {
```
rvagg, line 56:
```suggestion
func SplitBlockstore(cfg *config.Chainstore) func(lc fx.Lifecycle, r repo.LockedRepo, ds dtypes.MetadataDS, cold dtypes.ColdBlockstore, hot dtypes.HotBlockstore) (dtypes.SplitBlockstore, error) {
```

I think this is both unnecessary and could cause a problem. It ends up passing `func(...) (dtypes.SplitBlockstore, func() error, error)` to `fx.Provide` which takes all arguments except the last (if it's an `error`) to be types to feed into the DI system to be available to match to anything that wants it. So we end up giving it both a `dtypes.SplitBlockstore` _and_ a (noop) `func() error`. I'm not sure if fx is going to use it if it's not got a named type, but it's in the DI system at least and perhaps someone ends up adding some code that requires a `func() error` and magically gets this noop one and then can't figure out why it's not doing what they expect.

rvagg:
Well, it's already been quite an education trying to get my head around what this is doing and why it would work. Some things I've discovered that I didn't really understand previously:

* `./datastore/chain` is the 120G that we're "saving"
* We're likely saving a fair bit less than that because: (a) we still need all the chain, regardless of where it's stored, and (b) it's only the churned state that would have been GCd that we'd be saving, there's a lot of static state that either never changes or won't change for a long time that we won't be saving—but the savings would add up over time for a splitstore node.
  - My `chain` is 120G but my `splitstore` is 62G. If we consider that a typical snapshot gives me approximately what splitstore's compaction gives me in terms of historical state, then I'm only wasting somewhere below 62G (chain growth is included in there so it _is_ less), because a lot of that 120G is still useful - the chain, and static state. And I've been running this splitstore node since May.
* This original behaviour looks to me like it comes from the original work being split between these two PRs: https://github.com/filecoin-project/lotus/pull/5695 & https://github.com/filecoin-project/lotus/pull/4992 - the former introducing the "split" blockstore that was intended to be able to have multiple backing stores for different purposes (a pipe dream ever-noticeable through Lotus code), this PR handled the `Import` case. Then the second PR introduced the hot store with GC, but it didn't touch the import case.
* One of my main questions was whether this would impact the chain history that we need to have available - does moving it out of the universal store and into the hot store make it a candidate for GC? The answer is no: `SplitStore#compact()` is quite clear that it walks the chain, protects the block headers and even the tipsetkeys, but it only walks 4 finalities of stateroots (I wish this number were configurable, that'd be so much more helpful than the other splitstore config values).
* `SplitStore` almost always will look in hot first and then look in cold, so it's this second read where we find most of the historic stuff that got imported when you started.

So I think this should be perfectly fine to do.

But, I wonder if it might be good to ship this along with a super-simple lotus-shed that you can run on a closed repo that would manually copy everything from the existing cold store into the hot store. Then you can take advantage of this without a resync. If that's too tricky then it's probably not worth it but it might be neat to ship this with that option.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 # UNRELEASED
 
+- Sync snapshots directly to hotstore.  Save 120 G of lotus disk space. ([filecoin-project/lotus#12800](https://github.com/filecoin-project/lotus/pull/12803))
```
rvagg, line 12: As per my comments, "120 G" is overselling it by quite a bit, we need to nuance this and probably not even put a number on it.

rvagg:
Another question I have about this is that might be important: is it going to increase CPU usage for hotstore compaction? Do we take shortcuts if we find that a CID is in the coldstore and essentially skip over it, while being more rigorous about hotstore objects? I don't _think_ so but it's not crystal clear to me that this is the case.

CHANGELOG.md
ZenGround0: Technically correct @rvagg but this is not really the christmas spirit. 🎅  😢 

I really don't buy that this is significantly overselling it. Its just the difference in startup sizes immediately after syncing from snapshot and a lot of us do that pretty often.  Plus it looks like its also exactly the high water mark which matters because that's the value that matters when you need to get bigger disks.

I guess we should have a graph of hotstore size over time to quantify it.  After 8 months your hot store has leaned down as the state has replaced itself but its a slow process.

ZenGround0: And hey I said "~120" !

ZenGround0:
> My chain is 120G but my splitstore is 62G. If we consider that a typical snapshot gives me approximately what splitstore's compaction gives me in terms of historical state, then I'm only wasting somewhere below 62G (chain growth is included in there so it is less), because a lot of that 120G is still useful - the chain, and static state. And I've been running this splitstore node since May.

It looks like maybe you were missing that when you sync from snapshots you don't get your nice 62 G in hotstore but 120G in hotstore (+ the 120G never gone from universal bstore).  I got 120G from measuring disk and it will be the same for everyone (at current state sizes).  You are down to 62G as your hotstore has slowly replaced the useless churn over many months.

ZenGround0:
Actually @rvagg your numbers aren't squaring up with my understanding of the system. Snapshot size is 1 state + 2.2 finalities of churn.  After compaction we should have 1 state + 4 finalities of churn.  So the hotstore should never be below 120 G.  Furthermore the splitstore is a totally separate datastore from the universal blockstore.  This is why in lotus today the hotstore must get warmed up by copying the entire snapshot into the hotstore.  In discussion above there is an assumption that data in the universal store is "useful" but I don't think that's the case.  Reading from the splitstore doesn't check the universal blockstore first for example.

Since you're seeing 60 G splitstore dir clearly I'm off about something here, do you see what I'm missing?

CHANGELOG.md
ZenGround0: I think you just missed the full copy of all cold objects to hot in the warmup procedure: https://github.com/filecoin-project/lotus/blob/master/blockstore/splitstore/splitstore_warmup.go#L89-L109

rvagg:
yeah, ok, warmup is something else... so I understand it less than I thought I did. I'll have to budget some time to dig further next week

rvagg:
haven't forgotten about this one, just trying to make backlog space

BigLep:
2025-06-25 maintainer conversation: ball is in @rvagg's court to understand splitstore better and see if can understand if there were reasons that more wasn't put in the hotstore to begin with.  

This was useful to jump into this blackbox, but there's nothing pushing us right now to finish it off yet.  As a result will move back to draft/in-progress.

PR #12808, opened 2025-01-06
feat(eth): check events are indexed within in requested range

rvagg:
This is a continuation of @akaladarshi's excellent work in #12728, finishing it up and making it mergeable. Original commit squashed from that PR is retained here, with just a small extra one by me added on top.

### DIFF
--- a/chain/index/ddls.go
+++ b/chain/index/ddls.go
@@ -91,7 +91,6 @@ func preparedStatementMapping(ps *preparedStatements) map[**sql.Stmt]string {
 		&ps.getMsgCidFromEthHashStmt:                  "SELECT message_cid FROM eth_tx_hash WHERE tx_hash = ? LIMIT 1",
 		&ps.insertEthTxHashStmt:                       "INSERT INTO eth_tx_hash (tx_hash, message_cid) VALUES (?, ?) ON CONFLICT (tx_hash) DO UPDATE SET inserted_at = CURRENT_TIMESTAMP",
 		&ps.insertTipsetMessageStmt:                   "INSERT INTO tipset_message (tipset_key_cid, height, reverted, message_cid, message_index) VALUES (?, ?, ?, ?, ?) ON CONFLICT (tipset_key_cid, message_cid) DO UPDATE SET reverted = 0",
-		&ps.hasTipsetStmt:                             "SELECT EXISTS(SELECT 1 FROM tipset_message WHERE tipset_key_cid = ?)",
 		&ps.updateTipsetToNonRevertedStmt:             "UPDATE tipset_message SET reverted = 0 WHERE tipset_key_cid = ?",
 		&ps.updateTipsetToRevertedStmt:                "UPDATE tipset_message SET reverted = 1 WHERE tipset_key_cid = ?",
 		&ps.removeTipsetsBeforeHeightStmt:             "DELETE FROM tipset_message WHERE height < ?",
--- a/chain/index/events.go
+++ b/chain/index/events.go
@@ -17,13 +17,14 @@ import (
 	"github.com/filecoin-project/go-address"
 	amt4 "github.com/filecoin-project/go-amt-ipld/v4"
 	"github.com/filecoin-project/go-state-types/abi"
-	"github.com/filecoin-project/lotus/chain/types"
 	blockadt "github.com/filecoin-project/specs-actors/actors/util/adt"
+
+	"github.com/filecoin-project/lotus/chain/types"
 )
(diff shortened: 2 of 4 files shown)


BigLep:
@aarshkshah1992 : can you please look at this week so we close this out?

aarshkshah1992:
Reviewing this today. Apologies for the delay folks !

chain/index/events.go
```diff
@@ -280,7 +282,92 @@ func (si *SqliteIndexer) checkTipsetIndexedStatus(ctx context.Context, f *EventF
 	return ErrNotFound // Tipset is not indexed
 }
 
-// getTipsetKeyCidByHeight retrieves the tipset key CID for a given height.
+// checkRangeIndexedStatus verifies if a range of tipsets specified by the given height range is
+// indexed. It checks for the existence of non-null rounds at the range boundaries.
+func (si *SqliteIndexer) checkRangeIndexedStatus(ctx context.Context, minHeight abi.ChainEpoch, maxHeight abi.ChainEpoch) error {
+	head := si.cs.GetHeaviestTipSet()
+	if minHeight > head.Height() || maxHeight > head.Height() {
+		return ErrRangeInFuture
```
aarshkshah1992, line 290: better to have the error mention the latest height we have in our index

chain/index/events.go
```diff
@@ -280,7 +282,92 @@ func (si *SqliteIndexer) checkTipsetIndexedStatus(ctx context.Context, f *EventF
...
+	for height := minHeight; height <= maxHeight; height++ {
+		cid, err := si.getTipsetKeyCidByHeight(ctx, height)
+		if err != nil {
+			if !errors.Is(err, ErrNotFound) {
+				return nil, 0, xerrors.Errorf("failed to get tipset key cid for height %d: %w", height, err)
+			}
+			// else null round, keep searching
+			continue
+		}
+		minHeight = height // Update the minHeight to the found height
```
aarshkshah1992, line 347: why do we need to update `minHeight` to height here ? Can't we just return `height` ? IIUC, we're not passing a pointer to this function and so not really updating the original `minHeight` param

chain/index/events.go
```diff
@@ -280,7 +282,92 @@ func (si *SqliteIndexer) checkTipsetIndexedStatus(ctx context.Context, f *EventF
...
+	if startCid == nil {
+		return nil
+	}
+
+	// Find the last non-null round in the range
+	endCid, endHeight, err := si.findLastNonNullRound(ctx, maxHeight, minHeight)
+	if err != nil {
+		return xerrors.Errorf("failed to find last non-null round: %w", err)
+	}
+	// If all rounds are null, consider the range valid
```
aarshkshah1992, line 308: this comment is wrong

chain/index/events.go
aarshkshah1992: Aha I see why. This is to zero down to range of non-null epochs. Yeah makes sense.

aarshkshah1992 (approved):
just a few nits but looks solid

chain/index/ddls.go
```diff
@@ -91,7 +91,6 @@ func preparedStatementMapping(ps *preparedStatements) map[**sql.Stmt]string {
 		&ps.getMsgCidFromEthHashStmt:                  "SELECT message_cid FROM eth_tx_hash WHERE tx_hash = ? LIMIT 1",
 		&ps.insertEthTxHashStmt:                       "INSERT INTO eth_tx_hash (tx_hash, message_cid) VALUES (?, ?) ON CONFLICT (tx_hash) DO UPDATE SET inserted_at = CURRENT_TIMESTAMP",
 		&ps.insertTipsetMessageStmt:                   "INSERT INTO tipset_message (tipset_key_cid, height, reverted, message_cid, message_index) VALUES (?, ?, ?, ?, ?) ON CONFLICT (tipset_key_cid, message_cid) DO UPDATE SET reverted = 0",
-		&ps.hasTipsetStmt:                             "SELECT EXISTS(SELECT 1 FROM tipset_message WHERE tipset_key_cid = ?)",
```
aarshkshah1992, line 94: we removed it as not needed anywhere.

rvagg: removed here because it's a duplicate from line 67, it's still used

chain/index/events.go
rvagg: No, you're right, this isn't necessary. It originally was in his first version because it was passed as a pointer that got updated but I suggested we not do that and this is an artefact of that change.

Issue #12790, opened 2024-12-16
Some Lotus CLI commands don't enforce number of arguments


rvagg:
Ref: https://github.com/filecoin-project/lotus/issues/12788

https://github.com/filecoin-project/lotus/blob/7f2068efc36b5f2d2665f7f1b25650d870d1f870/cli/wallet.go#L735-L739

`lotus wallet add` does a `< 1` check on `NArg()`, but it really should be `!= 1` to match the usage `<amount>`. This would pick up any mis-use, like supplying the options too late (thinking that it's got GNU CLI flexibility when it's unfortunately Go's frustrating typical BSD CLI inflexibility). I haven't looked but it's likely that there are more in here with this same problem.

Marking this job as `good first issue` because it's really just a fairly straightforward audit; walking through all of the commands attached to the `lotus` CLI (we'll scope it to that for now) and checking that `NArgs` is being properly enforced to match the usage (or vice versa).

Ideally the CLI package should be able to understand its usage string and enforce that, or maybe we could come up with our own usage string parser (e.g. `"<foo> [bar] [baz]" should have >=1 && <=3`); but that could be a later-job and for now just getting them matched would be good.

ameeetgaikwad:
would like to take this up!

rjan90:
Hey @ameeetgaikwad! I was just wondering if you had any progress on the fix for issue, or any questions related to it?

ameeetgaikwad:
hey @rvagg i got what you are saying.
one doubt, if the args length is 0, do i have to put a check for it as well?
```
if cctx.Narg() > 0 {
return IncorrectNumArgs(cctx)
}
```

rvagg:
sorry, I'm a bit behind on my backlog so am starting to catch up to this; I think the answer is yes, we want to be strict so as to not let the user stumble by sending empty commands into the void where they think it'll do something but it actually won't

but again we can limit the blast-radius of this by focusing on just the `lotus` and `lotus-miner` command, or even just `lotus`; certainly there are far too many in `lotus-shed` to bother with and that's a playground anyway, `lotus-worker is probably important too

ameeetgaikwad:
Let me know if any more changes are needed

> sorry, I'm a bit behind on my backlog so am starting to catch up to this; I think the answer is yes, we want to be strict so as to not let the user stumble by sending empty commands into the void where they think it'll do something but it actually won't
> 
[...]

PR #12814, opened 2025-01-07
refactor: refactoring Narg checks

ameeetgaikwad:
## Related Issues

#12790 
## Checklist

- [ ] CI is green

### DIFF
--- a/cli/filplus.go
+++ b/cli/filplus.go
@@ -545,7 +545,7 @@ var filplusRemoveExpiredAllocationsCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
 			return IncorrectNumArgs(cctx)
 		}
 
@@ -638,7 +638,7 @@ var filplusRemoveExpiredClaimsCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
 			return IncorrectNumArgs(cctx)
 		}
 
--- a/cli/spcli/sectors.go
+++ b/cli/spcli/sectors.go
@@ -1336,7 +1336,7 @@ func TerminateSectorCmd(getActorAddress ActorAddressGetter) *cli.Command {
 			},
 		},
 		Action: func(cctx *cli.Context) error {
-			if cctx.NArg() < 1 {
+			if cctx.NArg() != 1 {
 				return lcli.ShowHelp(cctx, fmt.Errorf("at least one sector must be specified"))
 			}
 
--- a/cmd/lotus-miner/proving.go
+++ b/cmd/lotus-miner/proving.go
@@ -304,7 +304,7 @@ var provingRecoverFaultsCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
 			return lcli.ShowHelp(cctx, xerrors.Errorf("must pass at least 1 sector number"))
 		}
 
--- a/cmd/lotus-miner/sealing.go
+++ b/cmd/lotus-miner/sealing.go
@@ -410,7 +410,7 @@ var sealingDataCidCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 || cctx.NArg() > 2 {
+		if cctx.NArg() != 1 || cctx.NArg() > 2 {
 			return lcli.ShowHelp(cctx, xerrors.Errorf("expected 1 or 2 arguments"))
 		}
 
--- a/cli/wallet.go
+++ b/cli/wallet.go
@@ -733,7 +733,7 @@ var walletMarketAdd = &cli.Command{
 		afmt := NewAppFmt(cctx.App)
 
 		// Get amount param
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
 			return IncorrectNumArgs(cctx)
 		}
 		f, err := types.ParseFIL(cctx.Args().First())


rvagg:
@ameeetgaikwad I think we need to be a bit more complete with this, have a look for instances of `ctx.Args` (specifically `cctx.Args()` usually but `ctx.Args` is a good shortening that should catch variations), and then walk back in the commands to see if there is a check there. Doing this in `cli` and doing a quick browse of the initial results I can see `slash-consensus` seems to need a check for `2` but doesn't have one. evm.go has its own custom error method (`if argc := cctx.Args().Len(); argc != 1 {`) which we probably should make consistent and use the `IncorrectNumArgs` function (I'm not sure there's a good reason to do it this way, seems like copypasta code to me). I've only looked through the first few files and haven't even touched miner or worker. It would be good here to be a bit more exhaustive than just fixing the obvious ones.

ameeetgaikwad:
> @ameeetgaikwad I think we need to be a bit more complete with this, have a look for instances of `ctx.Args` (specifically `cctx.Args()` usually but `ctx.Args` is a good shortening that should catch variations), and then walk back in the commands to see if there is a check there. Doing this in `cli` and doing a quick browse of the initial results I can see `slash-consensus` seems to need a check for `2` but doesn't have one. evm.go has its own custom error method (`if argc := cctx.Args().Len(); argc != 1 {`) which we probably should make consistent and use the `IncorrectNumArgs` function (I'm not sure there's a good reason to do it this way, seems like copypasta code to me). I've only looked through the first few files and haven't even touched miner or worker. It would be good here to be a bit more exhaustive than just fixing the obvious ones.

I'll go further deep into it!

cli/filplus.go
```diff
@@ -545,7 +545,7 @@ var filplusRemoveExpiredAllocationsCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
```
rvagg, line 548: no, this one's correct already, the first hint is in `ArgsUsage` where you see `...`, then further down there's a `Slice()` and then an iteration over however many there are

cli/multisig.go
```diff
@@ -89,7 +89,7 @@ var msigCreateCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
```
rvagg, line 92: `ArgsUsage` has a `...` then later we do a `for _, a := range cctx.Args().Slice() {`

rvagg:
@ameeetgaikwad as per my latest comments, could you do a quick scan over your edits and double-check that you're not editing ones that can take >=X args rather than just ==X

ameeetgaikwad:
@rvagg so I shouldn't be editing >=X or <=X args right? only the one which has ==X. Is that correct?

cli/filplus.go
ameeetgaikwad: can you elaborate on this? Didn't quite understood it

cli/chain.go
```diff
@@ -399,6 +399,10 @@ var ChainSetHeadCmd = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
+		if cctx.NArg() != 1 {
+			return IncorrectNumArgs(cctx)
+		}
+
```
rvagg, line 405: This one can be removed, see the block just below, this command has a subtle usage - you can either set it to `--genesis`, set an `--epoch X`, _or_ provide a tipset.
Not a very graceful command, we should probably make it more standard, maybe a `--tipset` instead of using the arg. But for now this can be removed and we'll leave it alone.

cli/disputer.go
```diff
@@ -150,6 +154,10 @@ var disputerStartCmd = &cli.Command{
 
 		ctx := ReqContext(cctx)
 
+		if cctx.NArg() != 1 {
+			return IncorrectNumArgs(cctx)
+		}
+
```
rvagg, line 160: oh this is an interesting one! you've stumbled upon a command with very bad settings, let's be a bit more radical here:

1. nargs should be strictly `0`, it doesn't actually accept an unflagged argument at all
2. In `Flags` above, `"start-epoch"` should be renamed to `"height"` because that's what it actually reads for that 
property
3. `"max-fee"` needs to be listed in `Flags`, it's optional and could be described as `"maximum fees to spend on a dispute message"`
4. `"from"` needs to be lited in `Flags`, it's optional and could be described as `"the address to send dispute messages from if not the default wallet address"`

While you're in there, `"max-fee"` and `"from"` `Flags` can be copied as they are into `disputerMsgCmd` which also takes those flags but doesn't list them.

cli/evm.go
```diff
@@ -217,8 +217,8 @@ var EvmDeployCmd = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
-		if argc := cctx.Args().Len(); argc != 1 {
-			return xerrors.Errorf("must pass the contract init code")
+		if cctx.NArg() != 1 {
+			return IncorrectNumArgs(cctx)
 		}
```
rvagg, line 222: This is good, but we're losing information here; before this change it tells you what's missing but afterward it just tells you "incorrect number of arguments". So let's improve things: in `cli/helper.go`, just below the `IncorrectNumArgs` function, add a new one:

```go
func IncorrectNumArgsWithHint(cctx *ufcli.Context, hint string) error {
	return ShowHelp(cctx, fmt.Errorf("incorrect number of arguments, got %d, %s", cctx.NArg(), hint))
}
```

Then you can do this:

```suggestion
		if cctx.NArg() != 1 {
			return IncorrectNumArgsWithHint(cctx, "contract init code required")
		}
```

cli/evm.go
```diff
@@ -338,8 +338,8 @@ var EvmInvokeCmd = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
-		if argc := cctx.Args().Len(); argc != 2 {
-			return xerrors.Errorf("must pass the address and calldata")
+		if cctx.NArg() != 2 {
+			return IncorrectNumArgs(cctx)
```
rvagg, line 342:
```suggestion
			return IncorrectNumArgsWithHint(cctx, "address and calldata required")
```

cli/f3.go
```diff
@@ -316,8 +316,8 @@ var f3SubCmdCerts = &cli.Command{
 				f3FlagOutput,
 			},
 			Before: func(cctx *cli.Context) error {
-				if count := cctx.NArg(); count > 1 {
-					return fmt.Errorf("too many arguments: expected at most 1 but got %d", count)
+				if cctx.NArg() != 1 {
+					return IncorrectNumArgs(cctx)
 				}
```
rvagg, line 321: This one allows zero too, look at how it's used below (`if cctx.Args().Present()`). Let's use the hint again:

```suggestion
				if cctx.NArg() > 1 {
					return IncorrectNumArgsWithHint(cctx, "expected 1 at most")
				}
```

cli/f3.go
```diff
@@ -397,8 +397,8 @@ Examples:
 				f3FlagReverseOrder,
 			},
 			Before: func(cctx *cli.Context) error {
-				if count := cctx.NArg(); count > 1 {
-					return fmt.Errorf("too many arguments: expected at most 1 but got %d", count)
+				if cctx.NArg() > 1 {
+					return IncorrectNumArgs(cctx)
```
rvagg, line 401: same as above

cli/filplus.go
```diff
@@ -256,7 +256,7 @@ var filplusListAllocationsCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() > 1 {
+		if cctx.NArg() != 1 {
```
rvagg, line 259: this one allows zero as well

rvagg: see `if cctx.NArg() == 1 {` block below that has an early return but behaves differently otherwise

cli/filplus.go
```diff
@@ -401,7 +401,7 @@ var filplusListClaimsCmd = &cli.Command{
...
```
rvagg, line 404: same with this one, allows zero

cli/log.go
```diff
@@ -81,6 +81,10 @@ var LogSetLevel = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
+		if cctx.NArg() != 1 {
+			return IncorrectNumArgs(cctx)
+		}
+
 		if !cctx.Args().Present() {
 			return fmt.Errorf("level is required")
 		}
```
rvagg, line 90: you've essentially duplicated the check here, one way or the other would be fine - either leave this one alone, or replace that existing `if` block with a new `IncorrectNumArgsWithHint` that preserves the hint (i.e. here it tells you what's missing, but in your version it just tells you that you've provided the wrong number of args)

cli/multisig.go
```diff
@@ -214,6 +214,10 @@ var msigInspectCmd = &cli.Command{
 
 		store := adt.WrapStore(ctx, cbor.NewCborStore(blockstore.NewAPIBlockstore(api)))
 
+		if cctx.NArg() != 1 {
+			return IncorrectNumArgs(cctx)
+		}
+
```
rvagg, line 220: this is being done too late, do it at the top like the rest, but also note that it's already there with a `cctx.Args().Present()` check. You could replace it with what you have (just moved up) and preserve the message by using `IncorrectNumArgsWithHint`

cli/multisig.go
```diff
@@ -399,7 +403,7 @@ var msigProposeCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 3 {
+		if cctx.NArg() != 3 {
```
rvagg, line 406: check the block directly below this, this command has a bit of nuance, let's just leave it alone

cli/multisig.go
```diff
@@ -521,7 +525,7 @@ var msigApproveCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 2 {
+		if cctx.NArg() != 2 {
```
rvagg, line 528: ditto, but there's 2 more blocks below here; so these commands already strictly check their number of arguments, they just have nuance to the way they report the misuse, let's leave it alone

cli/multisig.go
```diff
@@ -658,7 +662,7 @@ var msigCancelCmd = &cli.Command{
...
```
rvagg, line 665: and again, there's 2 blocks below this that handle other cases up to 6 arguments, let's leave it alone

cli/net.go
```diff
@@ -275,6 +275,10 @@ var NetDisconnect = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
+		if cctx.NArg() != 1 {
```
rvagg, line 278: nope, check how `cctx.Args()` is used below, it can be any number; it'll work fine with zero, but we probably should report misuse in that case so you could do `== 0`, and even do the `Hint` variant with `"must provide at least one peerID"`

cli/net.go
```diff
@@ -416,6 +420,10 @@ var NetFindPeer = &cli.Command{
 
 		ctx := ReqContext(cctx)
 
+		if cctx.NArg() != 1 {
+			return IncorrectNumArgs(cctx)
+		}
```
rvagg, line 425: You should be doing these checks first up in the command, it's the cheapest check to perform and it's good to avoid unnecessary work if you're just going to error out after doing it. Note here in particular you're doing it _after_ we call `GetFullNodeAPI` which sets up a network connection to the full node and instantiates a full client wrapper. Not super expensive but if you're just going to error then we should avoid it. So arg checks first.

But then, if you notice, at the top there is already a check, it just isn't handled properly. In particular it should return an error so the command has a non-zero exit code at least. You could move your check up and use a `IncorrectNumArgsWithHint` saying `"must provide a peerID"`.

cli/net.go
```diff
@@ -592,6 +600,10 @@ var NetBlockAddPeer = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
+		if cctx.NArg() < 1 {
```
rvagg, line 603: good check, but move it up

cli/net.go
```diff
@@ -618,6 +630,10 @@ var NetBlockAddIP = &cli.Command{
...
```
rvagg, line 633: ditto, move it up

cli/net.go
```diff
@@ -634,6 +650,10 @@ var NetBlockAddSubnet = &cli.Command{
...
```
rvagg, line 653: ditto

cli/net.go
```diff
@@ -660,6 +680,10 @@ var NetBlockRemovePeer = &cli.Command{
...
```
rvagg, line 683: ditto

cli/net.go
```diff
@@ -686,6 +710,10 @@ var NetBlockRemoveIP = &cli.Command{
...
```
rvagg, line 713: ditto

cli/net.go
```diff
@@ -702,6 +730,10 @@ var NetBlockRemoveSubnet = &cli.Command{
...
```
rvagg, line 733: ditto

cli/net.go
```diff
@@ -958,6 +990,10 @@ var NetProtectAdd = &cli.Command{
...
```
rvagg, line 993: ditto

rvagg: actually, no, if you follow `decodePeerIDsFromArgs` you'll see that it already errors on no arguments, so let's remove this one

cli/net.go
```diff
@@ -988,6 +1024,10 @@ var NetProtectRemove = &cli.Command{
...
```
rvagg, line 1027: remove, same as the previous, `decodePeerIDsFromArgs` does the work for us

cli/spcli/actor.go
```diff
@@ -265,6 +265,10 @@ func ActorSetAddrsCmd(getActor ActorAddressGetter) *cli.Command {
 
 			ctx := lcli.ReqContext(cctx)
 
+			if cctx.NArg() != 1 {
```
rvagg, line 268: too late, but it's already handled at the top, see the way it deals with `args`, let's remove this one

cli/spcli/actor.go
```diff
@@ -746,6 +750,10 @@ func ActorProposeChangeWorkerCmd(getActor ActorAddressGetter) *cli.Command {
...
```
rvagg, line 753: already handled at the top, let's just leave this one or you could replace it with `IncorrectNumArgsWithHint` and preserve the message

cli/spcli/actor.go
```diff
@@ -992,6 +1000,10 @@ func ActorConfirmChangeWorkerCmd(getActor ActorAddressGetter) *cli.Command {
...
```
rvagg, line 1003: ditto

cli/spcli/sectors.go
```diff
@@ -1336,7 +1336,7 @@ func TerminateSectorCmd(getActorAddress ActorAddressGetter) *cli.Command {
 			},
 		},
 		Action: func(cctx *cli.Context) error {
-			if cctx.NArg() < 1 {
+			if cctx.NArg() != 1 {
```
rvagg, line 1339: no, this one can take >1 arguments too, see the line starting with `sectorNumbers := lo.Map(cctx.Args().Slice(),`, undo this change

cli/sync.go
```diff
@@ -118,6 +118,10 @@ var SyncMarkBadCmd = &cli.Command{
 		defer closer()
 		ctx := ReqContext(cctx)
 
+		if cctx.NArg() != 1 {
```
rvagg, line 121: you've duplicated the check just below here; move the check to the top and use `IncorrectNumArgsWithHint` to preserve the message

cli/sync.go
```diff
@@ -149,6 +153,10 @@ var SyncUnmarkBadCmd = &cli.Command{
...
```
rvagg, line 156: this duplicates a block below (`if !cctx.Args().Present()`), but you could bring it up to the top of the function with the message and use `IncorrectNumArgsWithHint` to preserve it

cli/sync.go
```diff
@@ -180,6 +188,10 @@ var SyncCheckBadCmd = &cli.Command{
...
```
rvagg, line 191: dupliates a block just below it, again you could use `IncorrectNumArgsWithHint` and bring it to the top if you like

cli/sync.go
```diff
@@ -222,6 +234,10 @@ var SyncCheckpointCmd = &cli.Command{
...
```
rvagg, line 237: No, let's leave this one alone, it's a bit more subtle. Read the code below. If `--epoch` is set it tries to use that, otherwise it tries to parse the tipset from the rest of the args, which could be 1 or more - a tipset is made up of at least one CID and this command lets you pass in as many as required.

cmd/lotus-miner/proving.go
```diff
@@ -304,7 +304,7 @@ var provingRecoverFaultsCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
```
rvagg, line 307: no, this one iterates over `cctx.Args().Slice()` so it can take more than 1 arg too, undo this

cmd/lotus-miner/sealing.go
```diff
@@ -410,7 +410,7 @@ var sealingDataCidCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 || cctx.NArg() > 2 {
+		if cctx.NArg() != 1 || cctx.NArg() > 2 {
```
rvagg, line 413: no, this check won't work, `!=1||>2` is only going to catch `1` but it can be 2, leave this one alone

cmd/lotus-shed/sectors.go
```diff
@@ -88,7 +88,7 @@ var terminateSectorPenaltyEstimationCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
```
rvagg, line 91: no, see `cctx.Args().Slice()` below, undo this one and we don't need to be messing with the lotus-shed commands

cmd/lotus-shed/sync.go
```diff
@@ -38,7 +38,7 @@ var syncValidateCmd = &cli.Command{
 		defer closer()
 		ctx := lcli.ReqContext(cctx)
 
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
```
rvagg, line 41: same as sectors.go, undo and leave it alone

cmd/lotus-shed/sync.go
```diff
@@ -73,7 +73,7 @@ var syncScrapePowerCmd = &cli.Command{
 	Usage:     "given a height and a tipset, reports what percentage of mining power had a winning ticket between the tipset and height",
 	ArgsUsage: "[height tipsetkey]",
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
```
rvagg, line 76: ditto

cmd/lotus-shed/sync.go
```diff
@@ -88,7 +88,7 @@ var syncScrapePowerCmd = &cli.Command{
 		defer closer()
 		ctx := lcli.ReqContext(cctx)
 
-		if cctx.NArg() < 1 {
+		if cctx.NArg() != 1 {
```
rvagg, line 91: ditto

cli/multisig.go
ameeetgaikwad: should I replace the `showHelp` with `IncorrectNumArgsWithHint` the one we wrote above

rvagg: no, this one's got too much nuance, leave multisig.go alone, it's got fairly sophisticated handling already

ameeetgaikwad:
I have some doubts regarding some comments. I have mentioned them.

Also let me know if I have missed some scope of this refactor. I'll refactor those folders as well

ameeetgaikwad:
Is there anything else we need to change here?

rjan90:
Closing this as outdated

rvagg:
Reopening, I'll help get this landed, @ameeetgaikwad has spent quite a bit of time on this already and I really should help get it over the line. Unfortunately I haven't been able to find the time on my end to review but the scope here has been reduced from original so we should be able to sort this out.

ameeetgaikwad:
hey @rvagg thanks!

ameeetgaikwad:
anything i can help here @rvagg

PR #12818, opened 2025-01-08
fix quotation marks in documentation files

xiaobei0715:
Incorrect use of quotation marks can lead to certain editors failing to accurately render text content, which may result in garbled output or other formatting issues that could confuse readers. Hope it helps.


rjan90:
Thanks for submitting a PR to Lotus. However, according to our [contributing guidelines](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md), we try to avoid changes that are primarily stylistic in nature:

> Please refrain from submitting PRs to adapt existing code to subjective preferences. The changeset should contain functional or technical improvements/enhancements, bug fixes, new features, or some other clear material contribution. Simple stylistic changes are likely to be rejected in order to reduce code churn.

While you mention that "incorrect use of quotation marks can lead to certain editors failing to accurately render text content", but I have yet too se any issues mentioning this problem, and therefore I will close it. If there is a clearer issue at hand, where people are encountering rendering issues because of this, I think we can open it again.

PR #12828, opened 2025-01-14
feat(shed): lotus-shed miner locked-vested to find erroneously locked funds

rvagg:
Ref: https://github.com/filecoin-project/builtin-actors/issues/1594

Current output:

```
Scanning actors (looking for v24 miners [bafk2bzacecnl2hqe3nozwo7al7kdznqgdrv2hbbbmpcbcwzh3yl4trog433hc] at epoch 4616980)............................................................................................................................................................................................................................................................................................................................................
Total actors: 3320119
Total miners: 686515
Miners with locked funds: 2073
Total locked funds: 355780.058629719966731966 FIL
```


rvagg:
@ZenGround0 this is probably overcounting because it's comparing to _now_ but the unlocking only happens on deadline tick. Maybe it should count from `WPoStProvingPeriod` less than now?

rvagg:
If I do that, the number of miners reduces but the total FIL is still large, which makes sense since we're just discounting funds that are about to vest:

```
Total actors: 3320124
Total miners: 686518
Miners with locked funds: 1530
Total locked funds: 342826.386784283280495497 FIL
```

ZenGround0 (approved):
It would be useful to have an additional command that dumps the vesting table of a given miner if you're in the mood to add that here.  But either way it will now be easy to build that with this as an example.

rvagg:
@ZenGround0 good idea, done in this PR (change title to match). I also put a "stale epoch" of now minus one deadline, which I think matches how cron works—if it hasn't cleared by then, then it's probably not going to clear. Can you have a look at this logic? Maybe I should `-1` from this too?

```
Total actors: 3320186
Total miners: 2026
Miners with locked vested funds: 2026
Total locked vested funds: 355059.650786527951408884 FIL
```

It can also list miners and the miners in this list and the amount locked. The largest is `f01825045` with `11797`, and he's relatively old, from 3956678 to 4463558 where it stops.

BigLep:
I moved this back to "awaiting review" from @ZenGround0 .

ZenGround0:
> I moved this back to "awaiting review" 

Approved

PR #12812, opened 2025-01-07
chore: add `help` target to Makefile to display the usage of all targets

caseylove:
## Proposed Changes
Add a new target `help`.

The `help` target will help new developers contribute to this project easily.

### Screenshot of `make help`:

img_v3_02ia_165c14b1-f5ac-4765-bbca-cd1469c3c6eg

## Checklist

Before you mark the PR ready for review, please make sure that:

- [ ] Commits have a clear commit message.
- [ ] PR title conforms with [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#pr-title-conventions)
- [ ] Update CHANGELOG.md or signal that this change does not need it per [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#changelog-management)
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12836, opened 2025-01-21
fix(make): restore docsgen-cli target

rvagg:
https://github.com/filecoin-project/lotus/pull/12812 undid a change that we made in https://github.com/filecoin-project/lotus/pull/12717, this commit restores it


masih (approved):

rjan90 (approved):
Ahh, thanks for catching 🙏

PR #12856, opened 2025-01-29
chore: break docsgen-cli to see if it gets generated for me

rvagg:
Testing #12681


rvagg:
https://github.com/filecoin-project/lotus/pull/12681#issuecomment-2620722976

PR #12858, opened 2025-01-29
feat(ci): add stale branch workflow

rjan90:
## Proposed Changes
Add a workflow that marks stale branches and deletes stale branches:
- Branches are marked as stale after 180 days without no commits.
- Branch that have been inactive for 360 days is deleted.
- Has no protection rules.
- Exclude any branch that starts with 'release/miner' as these have no branch protection rules ([ref](https://github.com/filecoin-project/lotus/branches/all?query=release%2Fminer))
- Is not the default branch

## Checklist

- [ ] CI is green

### DIFF
--- a/.github/workflows/stale-branches
+++ b/.github/workflows/stale-branches
@@ -20,5 +20,5 @@ jobs:
         repo-token: '${{ secrets.GITHUB_TOKEN }}'
         days-before-stale: 180
         days-before-delete: 360
-        branches-filter-regex: ''^(?!release/miner)'
+        branches-filter-regex: ''^(?!release)'
         dry-run: false
--- a/.github/workflows/stale-branches
+++ b/.github/workflows/stale-branches
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Stale Branches
-      uses: crs-k/stale-branches@v5.0.0
+      uses: crs-k/stale-branches@c6e09a3de1046d68b21eccdca23321d0ec277964 # v7.0.0
       with:
         repo-token: '${{ secrets.GITHUB_TOKEN }}'
         days-before-stale: 180 #Number of days a branch has been inactive before it is considered stale and a 'stale branch 🗑️' issue is opened.
--- a/.github/workflows/stale-branches
+++ b/.github/workflows/stale-branches
@@ -18,7 +18,16 @@ jobs:
       uses: crs-k/stale-branches@v5.0.0
       with:
         repo-token: '${{ secrets.GITHUB_TOKEN }}'
-        days-before-stale: 90
-        days-before-delete: 360
-        branches-filter-regex: ''^(?!release)'
-        dry-run: false
+        days-before-stale: 180 #Number of days a branch has been inactive before it is considered stale and a 'stale branch 🗑️' issue is opened.
+        days-before-delete: 360 #Number of days a branch has been inactive before it is deleted.
+        comment-updates: false #comment with updated information will be added to existing issues each workflow run.
+        max-issues: 20 #the number of 'stale branch 🗑️' issues that can exist. Also, max number of branches that can be deleted per run.
+        tag-committer: true #when an issue is opened, the last committer will be tagged in the comments.
+        stale-branch-label: 'stale branch 🗑️' #Label applied to issues created for stale branches. Must be unique to this workflow.
+        compare-branches: 'info' #compares each branch to the repo's default branch. When set to info, additional output describes if the current branch is ahead, behind, diverged, or identical to the default branch.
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-branches
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  stale_branches:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Stale Branches
+      uses: crs-k/stale-branches@v5.0.0
+      with:
+        repo-token: '${{ secrets.GITHUB_TOKEN }}'
+        days-before-stale: 180
```
BigLep, line 21:
```suggestion
        days-before-stale: 90
```

Use 3 months to conform with GitHub's UI: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/viewing-branches-in-your-repository ?

(Or maybe I'm confused on what the utility is of marking branches of stale.)

BigLep (approved):
I'm good to merge, but a couple of questions coming to mind:
1. What does denoting a branch as "stale" do?
2. What kind of communication is there that a branch is about to be deleted?

.github/workflows/stale-branches
```diff
@@ -0,0 +1,24 @@
...
+        days-before-delete: 360
+        branches-filter-regex: ''^(?!release/miner)'
```
rvagg, line 23: is this regex right? we'd want everything starting with `release` right, so `releases`, `release/vx.y.z`, `release/miner/vx.y.z` so just chopping off the `/miner` part of this should be good?

.github/workflows/stale-branches
```diff
@@ -0,0 +1,24 @@
...
+permissions:
+  issues: write
+  contents: write
+
+jobs:
+  stale_branches:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Stale Branches
+      uses: crs-k/stale-branches@v5.0.0
```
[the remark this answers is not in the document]
rjan90: By default it should exclude branches with branch protection rules - see the default of `include-protected-branches` input here https://github.com/marketplace/actions/stale-branches.

If we want to be extra careful and verify that is the case, we can run this in dry-mode for a bit, and observe its behaviour.

.github/workflows/stale-branches
rjan90: My thought was that release branches is already protected by branch protection rules, so with this pattern we could verify that that branch protection pattern also works. But chopping off the `/miner` should be fine, and there is definetly other branches that has protection rules on them, which we can verify that with. Updated here: https://github.com/filecoin-project/lotus/pull/12858/commits/1e66f3e00530c6edd461befa7a5b2c0636fe8319

rvagg:
let's land it with dry-run and observe the output the first time it runs and see if there are any surprises

rjan90:
Agree on landing it with dry-run initially so that we can inspect the logs of the workflow firs. After diving deeper into the workflow today to answer some of Steve's questions, it also seems like an initial manual cleanup of the branches would be the best approach before landing this workflow, as well as some additional parameter tweaks:

- The difference between GitHub's definition of a stale branch and this workflows definition of a stale branch, is that once a branch is defined as stale by this bot (the `days-before-stale` param):
   - It creates a GH issue in the repo and tags the author saying that their branch is stale, and that it will be closed in (`days-before-stale` - `days-before-delete`)
   - If the author deletes, or add branch protection, their stale branch on their own, the issue will be closed automatically. If the author does not delete their stale branch, their branch will be deleted after the XY-amount of days from issue creation to deletion.

Given that we have 600+ branches, and most of these branches already are stale, it could quickly get a bit spammy without a initial cleanup, although the max amount of of stale branch issues that can exist at the same time is 20.

I will start with the manual cleanup, and then tweak the parameters.

BigLep:
All sounds good @rjan90 .  Also, I think it probably makes sense to document what some of the default are in the workflow (e.g., protoected branches excluded) or just add the value explicitly in so it's self documenting.

rjan90:
I have updated the workflow with comments and explicitly set defaults to make it more self-descriptive. The 'dry-run' option has been set to true, allowing us to observe the behavior first.

Additionally, I have completed an initial cleanup of old branches, reducing the total from 647 to 161.

.github/workflows/stale-branches
```diff
@@ -0,0 +1,33 @@
...
```
galargh, line 18: There's v7 of this action available now - https://github.com/crs-k/stale-branches/releases/tag/v7.0.0

I'd also suggest pinning to the commit id of the version

```suggestion
      uses: crs-k/stale-branches@c6e09a3de1046d68b21eccdca23321d0ec277964 # v7.0.0
```

BigLep (approved):

PR #12863, opened 2025-02-03
chore: log failure to run CLI command

masih:
## Proposed Changes
Prior code only conditionally logged the output when `LOTUS_DEV` env var was present. This doesn't make sense for a number of reasons:

* In case of an error starting up `lotus daemon` via systemctl for example the logs would miss the error. One then has to search through system journal to see what happened.

* There are no other places that I can find where `LOTUS_DEV` env var is used. The commit that introduced this condition is very old with no clear commit message to shed light into the rationale.

For the first reason alone, the changes here remove that condition, and log the message when the logging output is not standard err or out. This way we at least cover the case where lotus is run in a production environment and would avoid partial logging of errors.

## Checklist

- [ ] CI is green

### DIFF
--- a/cli/helper.go
+++ b/cli/helper.go
@@ -8,6 +8,7 @@ import (
 	"os/signal"
 	"syscall"
 
+	logging "github.com/ipfs/go-log/v2"
 	ufcli "github.com/urfave/cli/v2"
 )
 
@@ -46,11 +47,13 @@ func RunApp(app *ufcli.App) {
 	}()
 
 	if err := app.Run(os.Args); err != nil {
-		if os.Getenv("LOTUS_DEV") != "" {
-			log.Warnf("%+v", err)
-		} else {
-			fmt.Fprintf(os.Stderr, "ERROR: %s\n\n", err) // nolint:errcheck
+		if cfg := logging.GetConfig(); !(cfg.Stdout || cfg.Stderr) {
+			// To avoid printing the error twice while making sure that log file contains the
+			// error, check the config and only print it if the output isn't stderr or
+			// stdout.
+			log.Errorw("Failed to start application", "err", err)
 		}
+		_, _ = fmt.Fprintf(os.Stderr, "ERROR: %s\n\n", err)
 		var phe *PrintHelpErr
 		if errors.As(err, &phe) {
 			_ = ufcli.ShowCommandHelp(phe.Ctx, phe.Ctx.Command.Name)


Stebalien:
If I run a random CLI command and it fails, will I now see the error twice? Once as a log line, and once as a message?

masih:
> If I run a random CLI command and it fails, will I now see the error twice? Once as a log line, and once as a message?

Depends on what the logging configuration is. For example, log could be going to a file. But certainly possible for one to see it twice.

I am happy to acomodate an alternative way to do this, as long as: Logs capture the exit error of the application (in this case `lotus daemon`). As an operator, discovering that it does not was a big surprise to me.

Stebalien:
I'm not sure the best way to solve this. Right now, the user gets the error on the console twice where the error log is pretty verbose. E.g., running `lotus send f01010101 10` outputs:


> 2025-02-03T18:22:30.931-0800    ERROR   cli     cli/helper.go:49        Failed to start application     {"err": "could not get API info for FullNode: repo directory does not exist. Make sure your configuration is correct", "errVerbose": "could not get API info for FullNode:\n    github.com/filecoin-project/lotus/cli/util.GetRawAPIMulti\n        /home/steb/src/github.com/filecoin-project/lotus/cli/util/api.go:152\n  - repo directory does not exist. Make sure your configuration is correct"}
> ERROR: could not get API info for FullNode: repo directory does not exist. Make sure your configuration is correct

Where it would otherwise output just:

> ERROR: could not get API info for FullNode: repo directory does not exist. Make sure your configuration is correct

---

However, I do get the desire to see failures in `lotus daemon` etc. show up in the log. Options I can see are:

1. Log if and only if the log output is also being sent to a file or URL.
2. Log the error from `lotus daemon` _only_.

I'd prefer the first and that should be as simple as calling `go-log.GetConfig()` and checking if either `config.File` or `config.URL` are non-empty. 

(IMO, if you're using systemd, you should generally just dump everything to systemd and we should _probably_ have better support for syslog log levels, but that's a different can of worms)

masih:
> (IMO, if you're using systemd, you should generally just dump everything to systemd and we should _probably_ have better support for syslog log levels, but that's a different can of worms)

Indeed. Fiddling around with `GetConfig` seems hacky but it's the path of least resistance. The root cause IMHO is the duality of CLI output. I would handle it all with logging, and have a custom formatter settable by global flags with a reasonable default for the common case.

masih:
@Stebalien This is ready for another look whenever you get a chance. Thanks

Stebalien (approved):

PR #12864, opened 2025-02-03
feat(ci): run unit test jobs on arm64 machines additionally

### LINKED ISSUE #12860: Add CI runners for linux and mac arm64
## Done Criteria
Lotus unit tests run on linux and mac arm64 runners.

## Why Important
We do support it as a platform for nodes and do have users running on it.  We should confirm we do don't break support.

## User/Customer
* Maintainers to make sure platform support isn't borken
* arm64 users to have higher confidence that Lotus changes will continue to work on their platform of choice

## Notes
* This was brought up in #fil-lotus-dev as part of "linux arm64 runners are now available to open source projects" discussion.  See https://filecoinproject.slack.com/archives/CP50PPW2X/p1738057693640399

galargh:
## Related Issues

Resolves #12860

## Proposed Changes

Start running `unit-*` jobs on arm64 machines in addition to x86_64 ones. 

## Additional Info

In order to accommodate this change, I moved the test discovery code to a dedicated Go CLI. The important change in the structure of metadata is that now each test group can be associated with many target runners.

What's more, I split the test group discovery into part 1 which lists all the test groups, and part 2 which returns the metadata (e.g. runner information, etc.) for a given test group name. Thanks to this the outputs of the discovery job can be smaller and we minimise the risk of hitting the maximum output size limit.

## Checklist

- [ ] CI is green

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,6 +40,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
+        # We need to cache for each architecture we support: x86_64 and arm64
         runner: [ubuntu-latest, ubuntu-24.04-arm]
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,16 +24,16 @@ jobs:
     name: Discover Test Groups
     runs-on: ubuntu-latest
     outputs:
-      groups: ${{ steps.test.outputs.groups }}
+      test_group_execution_contexts: ${{ steps.list_test_group_execution_contexts.outputs.test_group_execution_contexts }}
     steps:
       - uses: actions/checkout@v4
         with:
           submodules: 'recursive'
           fetch-depth: 0
-      - id: test
+      - id: list_test_group_execution_contexts
         run: |
-          echo "groups<<EOF" >> $GITHUB_OUTPUT
-          go run ./cmd/ci/main.go --json list-test-groups | jq -r '.msg' | tee -a $GITHUB_OUTPUT
+          echo "test_group_execution_contexts<<EOF" >> $GITHUB_OUTPUT
+          go run ./cmd/ci/main.go --json list-test-group-execution-contexts | jq -r '.msg' | tee -a $GITHUB_OUTPUT
           echo "EOF" >> $GITHUB_OUTPUT
   cache:
     name: Cache Dependencies
@@ -122,7 +122,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include: ${{ fromJson(needs.discover.outputs.groups) }}
+        include: ${{ fromJson(needs.discover.outputs.test_group_execution_contexts) }}
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,12 +37,10 @@ jobs:
           echo "EOF" >> $GITHUB_OUTPUT
   cache:
     name: Cache Dependencies
-    runs-on: ubuntu-latest
-    outputs:
-      fetch_params_key: ${{ steps.fetch_params.outputs.key }}
-      fetch_params_path: ${{ steps.fetch_params.outputs.path }}
-      make_deps_key: ${{ steps.make_deps.outputs.key }}
-      make_deps_path: ${{ steps.make_deps.outputs.path }}
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -100,6 +98,23 @@ jobs:
         with:
           key: ${{ steps.make_deps.outputs.key }}
           path: ${{ steps.make_deps.outputs.path }}
+      - env:
+          fetch_params_key: ${{ steps.fetch_params.outputs.key }}
+          fetch_params_path: ${{ steps.fetch_params.outputs.path }}
+          make_deps_key: ${{ steps.make_deps.outputs.key }}
+          make_deps_path: ${{ steps.make_deps.outputs.path }}
+          file: jobs.cache.${{ runner.os }}.${{ runner.arch }}.outputs.json
+        run: |
+          jq -n '{
+            "fetch_params_key": env.fetch_params_key,
+            "fetch_params_path": env.fetch_params_path,
+            "make_deps_key": env.make_deps_key,
+            "make_deps_path": env.make_deps_path
+          }' | tee -a "$file"
+      - uses: actions/upload-artifact@v4
+        with:
+          name: jobs.cache.${{ runner.os }}.${{ runner.arch }}.outputs
+          path: jobs.cache.${{ runner.os }}.${{ runner.arch }}.outputs.json
   test:
     needs: [discover, cache]
     name: Test (${{ matrix.name }}) ${{ toJson(matrix.runner) }}
@@ -122,18 +137,37 @@ jobs:
           echo "EOF" >> $GITHUB_OUTPUT
       - name: Install gotestsum
         run: go install gotest.tools/gotestsum@latest
+      - id: artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: jobs.cache.${{ runner.os }}.${{ runner.arch }}.outputs
+      - id: cache
+        env:
+          file: jobs.cache.${{ runner.os }}.${{ runner.arch }}.outputs.json
+        run: |
+          echo "make_deps_key=$(jq -r .make_deps_key "$file")" | tee -a $GITHUB_OUTPUT
+          echo "make_deps_path<<EOF" | tee -a $GITHUB_OUTPUT
+          jq -r .make_deps_path "$file" | tee -a $GITHUB_OUTPUT
+          echo "EOF" | tee -a $GITHUB_OUTPUT
+
+          echo "fetch_params_key=$(jq -r .fetch_params_key "$file")" | tee -a $GITHUB_OUTPUT
+          echo "fetch_params_path<<EOF" | tee -a $GITHUB_OUTPUT
+          jq -r .fetch_params_path "$file" | tee -a $GITHUB_OUTPUT
+          echo "EOF" | tee -a $GITHUB_OUTPUT
+
+          rm "$file"
       - name: Restore cached make deps outputs
         uses: actions/cache/restore@v4
         with:
-          key: ${{ needs.cache.outputs.make_deps_key }}
-          path: ${{ needs.cache.outputs.make_deps_path }}
+          key: ${{ steps.cache.outputs.make_deps_key }}
+          path: ${{ steps.cache.outputs.make_deps_path }}
           fail-on-cache-miss: true
       - if: ${{ fromJson(steps.group.outputs.metadata).needs_parameters }}
         name: Restore cached fetch params outputs
         uses: actions/cache/restore@v4
         with:
-          key: ${{ needs.cache.outputs.fetch_params_key }}
-          path: ${{ needs.cache.outputs.fetch_params_path }}
+          key: ${{ steps.cache.outputs.fetch_params_key }}
+          path: ${{ steps.cache.outputs.fetch_params_path }}
           fail-on-cache-miss: true
       # TODO: Install statediff (used to be used for conformance)
       - name: Create temporary directory for reports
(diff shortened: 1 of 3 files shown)


BigLep:
@galargh : I assume you'll mark it ready for review when it is.

galargh:
> @galargh : I assume you'll mark it ready for review when it is.

Yes, it is ready for review now :)

cmd/ci/main.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+	return ""
+}
+
+func getRunners(name string) [][]string {
+	if os.Getenv("GITHUB_REPOSITORY_OWNER") != "filecoin-project" {
+		return [][]string{{"ubuntu-latest"}}
+	}
+
+	namesToRunners := map[string][][]string{
```
masih, line 132: I am not sure how much I like this TBH. Editing yaml for this kind of data seems much more convenient than string hardcoded into `map[string][][]string`.

From a design perspective, the code here is entirely CI specific. It should live somewhere inside `.github` folder?

I do think the long term goal is to desolve this into itests themselves in form of annotations or struct tags on "Test Suit" (see testify).

galargh: > I do think the long term goal is to desolve this into itests themselves in form of annotations or struct tags on "Test Suit" (see testify).

I'd treat this as a mid-step towards that desired goal. It's going to be much easier to parse and manipulate the data from itests here. And then, the hardcoded info inside this tool will disappear.

.github/workflows/test.yml
```diff
@@ -31,148 +31,16 @@ jobs:
...
-          groups="$(jq -n --argjson g "$groups" --argjson r "$runners" '$g | map(. + {"runner": (.name as $n | $r | .[$n]) })')"
-
-          # Apply the needs_parameters flag to the groups
-          groups="$(jq -n --argjson g "$groups" --argjson p "$parameters" '$g | map(. + {"needs_parameters": ([.name] | inside($p)) })')"
-
-          # Output the groups
-          echo "groups=$groups"
-          echo "groups=$(jq -nc --argjson g "$groups" '$g')" >> $GITHUB_OUTPUT
+          echo "groups<<EOF" >> $GITHUB_OUTPUT
+          go run ./cmd/ci/main.go --json list-test-groups | jq -r '.msg' | tee -a $GITHUB_OUTPUT
```
BigLep, line 36: Is the data model still the same?

My understanding of before:

```json
[
  { "name": "group1", "runner": "ubuntu-latest" },
  { "name": "group2", "runner": "ubuntu-24.04-arm" }
]
```

now:
```json
{
    "name": "itest-direct_data_onboard_verified",
    "runner": [
      "ubuntu-latest"
    ]
  }
```

Basically `runner` is an array now.  (Given its an array, lets change to runners to be clear?)

cmd/ci/main.go
```diff
@@ -0,0 +1,354 @@
...
+	"path/filepath"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+)
+
+type TestGroup struct {
+	Name   string   `json:"name"`
+	Runner []string `json:"runner"`
```
BigLep, line 15: `runners` instead of `runner` since there may be multiple?

cmd/ci/main.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+type TestGroupMetadata struct {
```
BigLep, line 18: I have found the data model and naming a bit confusing.  In my mind something name  "TestGroup" should be:

```go
type TestGroup struct {
	Name   string   `json:"name"`
	Runners []string `json:"runners"` // ← plural
	Packages          []string `json:"packages"`
	NeedsParameters   bool     `json:"needs_parameters"`
	SkipConformance   bool     `json:"skip_conformance"`
	TestRustProofLogs bool     `json:"test_rust_proof_logs"`
	Format            string   `json:"format"`
	GoTestFlags       string   `json:"go_test_flags"`
}
```

"Runners" is just as much metadata as the other fields.

IIUC though, we can't just get a big dump of all the "test groups" (name + runners + other metadata) because we're going to exceed some github limits between stetps.

As a result, it looks like the strategy that is being taken is:
1. get all the <testGroupName,runner> pairs and then 
2. for each pair, you want to be able to query to get the additional metadata.

That is understandable.  I think I would just update the names to reflect this clear.

Ideas:

`list-test-groups` → `list-test-group-ids`
I think it needs to return `<testGroupName,runner>` not a an array of runners for the current github workflow code to work when looking at:

```json
matrix:
        include: ${{ fromJson(needs.discover.outputs.groups) }}
```

Maybe then have
```go
type TestGroupId struct {
	Name   string   `json:"name"`
	Runner string `json:"runner"`
}
```

`get-test-group-metadata` seems fine but for completeness the returned object would ideally include the runners as well (although I realize that isn't needed so we could comment why it isn't included)

cmd/ci/main.go
BigLep: Yeah, I'd love to see more of this metadata live closer to the tests themselves with annotations, but I understand that is a bigger step.  This to me has better readability than jq commands in yaml files we have now.  I don't see this making things worse and is actually improving what we have now, so 👍 from me.

The idea of moving this into the .github directory seems good/reasonable.

.github/workflows/test.yml
```diff
@@ -31,148 +31,16 @@ jobs:
...
-    runs-on: ubuntu-latest
-    outputs:
-      fetch_params_key: ${{ steps.fetch_params.outputs.key }}
-      fetch_params_path: ${{ steps.fetch_params.outputs.path }}
-      make_deps_key: ${{ steps.make_deps.outputs.key }}
-      make_deps_path: ${{ steps.make_deps.outputs.path }}
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
```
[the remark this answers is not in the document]
galargh: No, we only care about the architecture here really. We use what we cache here on self-hosted runners as well.

.github/workflows/test.yml
```diff
@@ -270,22 +179,22 @@ jobs:
...
           LOTUS_SRC_DIR: ${{ github.workspace }}
           REPORTS_PATH: ${{ steps.reports.outputs.path }}
-          SKIP_CONFORMANCE: ${{ matrix.skip_conformance || '1' }}
-          TEST_RUSTPROOFS_LOGS: ${{ matrix.test_rustproofs_logs || '0' }}
-          FORMAT: ${{ matrix.format || 'standard-verbose' }}
-          PACKAGES: ${{ join(matrix.packages, ' ') }}
+          SKIP_CONFORMANCE: ${{ fromJson(steps.group.outputs.metadata).skip_conformance && '1' || '0' }}
+          TEST_RUSTPROOFS_LOGS: ${{ fromJson(steps.group.outputs.metadata).test_rustproofs_logs && '1' || '0' }}
+          FORMAT: ${{ fromJson(steps.group.outputs.metadata).format || 'standard-verbose' }}
+          PACKAGES: ${{ join(fromJson(steps.group.outputs.metadata).packages, ' ') }}
```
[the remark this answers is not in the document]
galargh: It doesn't matter much here because we literally execute a script to get these values, but it's a [good security practice](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable) for some types of workflows so I tend to stick with it in general.

.github/workflows/test.yml
galargh: Runner has always been an array. Before, it was also:

```
[
  { "name": "group1", "runner": ["ubuntu-latest"] },
  { "name": "group2", "runner": ["ubuntu-24.04-arm"] }
]
```

We needed them to be arrays because self-hosted runners have more than one label.

cmd/ci/main.go
galargh: It's still one. I'll make `[]string` into a `Runner` type so that it's clearer.

cmd/ci/main.go
galargh: > I think it needs to return <testGroupName,runner> not a an array of runners for the current github workflow code to work when looking at:

This bit is already true ;)

I'll update the naming 👍

cmd/ci/main.go
```diff
@@ -0,0 +1,354 @@
...
+
+func findIntegrationTestGroups() ([]TestGroup, error) {
+	groups := []TestGroup{}
+
+	err := filepath.Walk("itests", func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !info.IsDir() && strings.HasSuffix(info.Name(), "_test.go") {
```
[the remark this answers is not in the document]
galargh: No, it won't. Should we rename it or update the condition? Or is it not supposed to be executed?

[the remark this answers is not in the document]
BigLep: @rjan90 : is https://github.com/filecoin-project/lotus/blob/master/itests/multisig/suite.go supposed to be run as part of CI?  It looks like it's currently being skipped and we're wondering if it should be included now.

cmd/ci/main.go
```diff
@@ -10,11 +10,23 @@ import (
...
 
-type TestGroup struct {
-	Name   string   `json:"name"`
-	Runner []string `json:"runner"`
+type TestGroupExecutionContext struct {
+	Name   string `json:"name"`
+	Runner Runner `json:"runner"`
 }
 
+type Runner []string
```
BigLep, line 18: Maybe add the explanatory comment for why Runner is `[]string` rather than just `string`?

cmd/ci/main.go
```diff
@@ -24,53 +36,252 @@ type TestGroupMetadata struct {
...
+			createPackagePath("journal", "..."),
+			createPackagePath("lib", "..."),
+			createPackagePath("paychmgr", "..."),
+			createPackagePath("tools", "..."),
 		},
 	}
 
-	if strings.HasPrefix(name, "itest-") {
-		return []string{strings.Join([]string{".", "itests", strings.Join([]string{strings.TrimPrefix(name, "itest-"), "test.go"}, "_")}, string(os.PathSeparator))}
+	if strings.HasPrefix(testGroupName, "itest-") {
```
BigLep, line 276: (nit) but maybe move this if to the top of the function to avoid the interleaving of concerns that we have now

nonItest map declaration
itest handling
nonItest handling

cmd/ci/main.go
```diff
@@ -24,53 +36,252 @@ type TestGroupMetadata struct {
...
+
+	return groups
+}
+
+func getRunners(testGroupName string) []Runner {
+	if os.Getenv("GITHUB_REPOSITORY_OWNER") != "filecoin-project" {
+		return []Runner{linux_x64}
+	}
+
+	testGroupNamesToRunners := map[string][]Runner{
```
BigLep, line 173: nit: `testGroupNameToRunner` (since the map key is a single test group name?

.github/workflows/test.yml
BigLep: Ack - ok.  Good to know.  My bad.  Resolving.

cmd/ci/main.go
BigLep: `TestGroupExecutionContext` sounds good

Resolving

.github/workflows/test.yml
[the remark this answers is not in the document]
BigLep: Got it. So this job needs to run on x86_64 and arm, which is accomplished by `[ubuntu-latest, ubuntu-24.04-arm]`.  Sounds good.  

```suggestion
        # We need to cache for each architecture we support: x86_64 and arm
        runner: [ubuntu-latest, ubuntu-24.04-arm]
```

.github/workflows/test.yml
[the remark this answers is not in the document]
BigLep: Got it - makes sense.  I'm good to practice the right behavior here.  

Should we adjust this line below to be consistent?

```
-- ${{ fromJson(steps.group.outputs.metadata).go_test_flags || '' }}
```

BigLep (approved):
This is seeming good to me, and certainly not making anything worse.  My remaining comments are small nits.

I resolved all previous conversations that I believe are resolved now.

Before merging, it would be good to get input on https://github.com/filecoin-project/lotus/pull/12864#discussion_r1949708575

Also, is there an easy way to sanity check that we aren't reducing our coverage with this change (i.e., we didn't accidentally miss a test group execution context)?

cmd/ci/main.go
BigLep: The test metadata conversation also relates to https://github.com/filecoin-project/lotus/pull/11782

galargh: I do quite like all the command line tools living in the `cmd` directory myself. It's just a personal preference, though. I can definitely move them to the `.github` folder if you prefer. Would you like me to move the `release` tool there as well as it's mainly used in the CI?

.github/workflows/test.yml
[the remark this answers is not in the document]
galargh: Ah, yes! This is, in fact, an exception to that rule because we need the "words" in the `go_test_flags` string to be interpreted independently rather than one long string.

galargh:
> Also, is there an easy way to sanity check that we aren't reducing our coverage with this change (i.e., we didn't accidentally miss a test group execution context)?

I'll do a 1:1 comparison with `main` before hitting merge on this.

cmd/ci/main.go
rvagg: I'm fine with `cmd` for this one since we already have `release` in there, we can debate location for both in a separate PR I think.

galargh:
I have confirmed that we run the exact same set of jobs in this PR (+ the arm jobs, of course) as we do on `master`.

We also have one conversation that I think still needs a follow-up (about `suite.go`) - https://github.com/filecoin-project/lotus/pull/12864#discussion_r1947095853 - but I don't think it should block this PR so I'm gonig to merge it now.

cmd/ci/main.go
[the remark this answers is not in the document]
rjan90: Hi! I've reviewed the Slack and GitHub history and couldn't find any reason to skip this. Therefore, I suggest we include it now.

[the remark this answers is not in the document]
BigLep: From @galargh:

> I did check on that multisig/suite.go file again. We’re correctly skipping it after all.
> It only defines the RunMultisigTests function which is correctly called from the multisig_test.go which we do run (https://github.com/filecoin-project/lotus/blob/7b16d37ca3d7abcfa5eff7d2c2851d4d2c997c7c/itests/multisig_test.go#L37).
> So no further action needed.

I'm resolving this comment as a result.

PR #12870, opened 2025-02-04
chore: fix comments in .go

comfsrt:
api/api_full.go
incosistency - inconsistency

api/api_storage.go
onde r- under


rvagg (commented):
👍 but please run `make gen` on top of this, both of these godocs get compiled into openrpc docs that get checked in as well

rvagg:
Interesting, I didn't think `make gen` got far enough into it to need all the ffi stuff, can you provide your error output? this might be something we can address to make it easier.
Also, the README points you to https://lotus.filecoin.io/lotus/install/macos/ where you can find some simple `brew` instructions to install pkg-info, hwloc and friends.

rjan90:
Ran `make gen` for you in #12903 @comfsrt - will therefore close this PR.

PR #12878, opened 2025-02-07
chore(blockstore): add an LRU cache to the API blockstore wrapper

rvagg:
I've been relying on `NewAPIBlockstore` a lot lately for lotus-shed commands, miscellaneous things to inspect state mainly, where I don't want to stop my node to mount it as the blockstore. But realised that a lot of the fetches are for the same small blocks because we have a lot of things like THE empty partition, THE empty deadline, THE empty HAMT. It turns out that putting a simple LRU in there makes it so much faster to do simple iteration and checking when you're hitting these things.

This is mainly used in lotus-shed, but it has a few uses in the `lotus` command, specifically for some things under `lotus filplus`, `lotus miner info` and `lotus multisig`. The `spcli` commands also use this a little.

### DIFF
--- a/blockstore/api.go
+++ b/blockstore/api.go
@@ -3,6 +3,7 @@ package blockstore
 import (
 	"context"
 
+	lru "github.com/hashicorp/golang-lru/v2"
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	"golang.org/x/xerrors"
@@ -15,14 +16,22 @@ type ChainIO interface {
 }
 
 type apiBlockstore struct {
-	api ChainIO
+	api   ChainIO
+	cache *lru.Cache[cid.Cid, []byte]
 }
 
 // This blockstore is adapted in the constructor.
 var _ BasicBlockstore = (*apiBlockstore)(nil)
 
 func NewAPIBlockstore(cio ChainIO) Blockstore {
-	bs := &apiBlockstore{api: cio}
+	lc, err := lru.New[cid.Cid, []byte](1024)
+	if err != nil {
+		panic(err)
+	}
+	bs := &apiBlockstore{
+		api:   cio,
+		cache: lc,
+	}
 	return Adapt(bs) // return an adapted blockstore.
 }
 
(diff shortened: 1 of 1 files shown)


blockstore/api.go
```diff
@@ -15,14 +16,22 @@ type ChainIO interface {
...
+	api   ChainIO
+	cache *lru.Cache[cid.Cid, []byte]
 }
 
 // This blockstore is adapted in the constructor.
 var _ BasicBlockstore = (*apiBlockstore)(nil)
 
 func NewAPIBlockstore(cio ChainIO) Blockstore {
-	bs := &apiBlockstore{api: cio}
+	lc, err := lru.New[cid.Cid, []byte](1024)
```
masih, line 27: I recommend leaving an in-line comment about the rationale for picking this size.

blockstore/api.go
```diff
@@ -15,14 +16,22 @@ type ChainIO interface {
...
+	if err != nil {
+		panic(err)
```
masih, line 29: Subjective: I would leave a comment about why this panic. It's cheap to do and reduces cognitive cycles for future readers: "`lru.New` only returns an error the size is less than 1. There is no point returning this error since the size is configured as a constant."

blockstore/api.go
```diff
@@ -31,31 +40,44 @@ func (a *apiBlockstore) DeleteBlock(context.Context, cid.Cid) error {
...
 
 func (a *apiBlockstore) Get(ctx context.Context, c cid.Cid) (blocks.Block, error) {
+	if bb, ok := a.cache.Get(c); ok {
+		return blocks.NewBlockWithCid(bb, c)
+	}
 	bb, err := a.api.ChainReadObj(ctx, c)
 	if err != nil {
 		return nil, err
 	}
+	a.cache.Add(c, bb)
```
masih, line 57: Use `ContainsOrAdd` for better performance in a highly concurrent environment. That would reduce unnecessary churn in LRU.

blockstore/api.go
```diff
@@ -31,31 +40,44 @@ func (a *apiBlockstore) DeleteBlock(context.Context, cid.Cid) error {
...
 
 func (a *apiBlockstore) GetSize(ctx context.Context, c cid.Cid) (int, error) {
+	if bb, ok := a.cache.Get(c); ok {
+		return len(bb), nil
+	}
 	bb, err := a.api.ChainReadObj(ctx, c)
 	if err != nil {
 		return 0, err
 	}
+	a.cache.Add(c, bb)
```
masih, line 69: Ditto re `ContainsOrAdd`.

blockstore/api.go
```diff
@@ -31,31 +40,44 @@ func (a *apiBlockstore) DeleteBlock(context.Context, cid.Cid) error {
...
 	bb, err := a.api.ChainReadObj(ctx, c)
 	if err != nil {
 		return nil, err
 	}
+	a.cache.Add(c, bb)
 	return blocks.NewBlockWithCid(bb, c)
 }
 
 func (a *apiBlockstore) GetSize(ctx context.Context, c cid.Cid) (int, error) {
+	if bb, ok := a.cache.Get(c); ok {
```
masih, line 62: Double checking that we do mean to call `Get` instead of `Peek`, i.e., should the calls to `GetSize` affect cache eviction?

blockstore/api.go
```diff
@@ -31,31 +40,44 @@ func (a *apiBlockstore) DeleteBlock(context.Context, cid.Cid) error {
...
 	bb, err := a.api.ChainReadObj(ctx, c)
 	if err != nil {
 		return 0, err
 	}
+	a.cache.Add(c, bb)
 	return len(bb), nil
 }
 
 func (a *apiBlockstore) Put(ctx context.Context, block blocks.Block) error {
+	a.cache.Add(block.Cid(), block.RawData())
```
masih, line 74: Here we probably want to update the recentness? If not, ditto re `ContainsOrAdd`.

blockstore/api.go
```diff
@@ -31,31 +40,44 @@ func (a *apiBlockstore) DeleteBlock(context.Context, cid.Cid) error {
...
 	return a.api.ChainPutObj(ctx, block)
 }
 
 func (a *apiBlockstore) PutMany(ctx context.Context, blocks []blocks.Block) error {
 	for _, block := range blocks {
+		a.cache.Add(block.Cid(), block.RawData())
```
masih, line 80: Same comment as add in `Put`.

blockstore/api.go
```diff
@@ -68,6 +90,4 @@ func (a *apiBlockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error)
 	return nil, xerrors.New("not supported")
 }
 
-func (a *apiBlockstore) HashOnRead(enabled bool) {
-	return
-}
+func (a *apiBlockstore) HashOnRead(enabled bool) {}
```
masih, line 93:
```suggestion
func (a *apiBlockstore) HashOnRead(bool) {}
```

masih (approved):
No blockers.

blockstore/api.go
rvagg: eh, probably a good thing to do but I bet if I made this method panic then nothing would break

Issue #12838, opened 2025-01-21
`EthGetBlockReceiptsLimited` does not use limit


elmattic:
### Checklist

- [x] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [x] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [x] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [x] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [x] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
./lotus version
Daemon:  1.31.0-rc1+calibnet+git.6cd798259.dirty+api1.5.0
Local: lotus version 1.31.0-rc1+calibnet+git.6cd798259.dirty
```

### Repro Steps

1. Run '...'
2. Do '...'
3. See error '...'
...

### Describe the Bug

https://github.com/filecoin-project/lotus/blob/master/node/impl/eth/transaction.go#L448

Upon reviewing this function's implementation, I noticed that the `limit` parameter is not being used anywhere.

This is not necessarily a problem, but rather a potential question about what the specification should be.

### Logging Information

```text
n/a
```

rvagg:
Whoops! Thanks for picking this up @elmattic; we'll get this done.

rvagg:
(unassigning myself to leave the door open but in reality I'll probably end up doing)

kamuik16:
> (unassigning myself to leave the door open but in reality I'll probably end up doing it)

I can do it.

ameeetgaikwad:
would like to take this up @rvagg

BigLep:
Go for it @ameeetgaikwad.  Assigned!

BigLep:
@ameeetgaikwad : just checking in on how this is going.

ameeetgaikwad:
@BigLep yes! Pushing my EOD

[...]

PR #12883, opened 2025-02-09
fix(eth): apply limit in EthGetBlockReceiptsLimited

ameeetgaikwad:
## Related Issues

 #12838 
## Checklist

- [ ] CI is green


rjan90:
OK, that was a bit excessive from the PR title GH-actions workflow, I'll investigate how to make it less spammy and improve its behavior.

BigLep:
@ameeetgaikwad : I don't know the codebase to be able to speak to the feasibility, but did you look at adding a unit test?

rvagg:
@ameeetgaikwad check out the lint error from CI, your editor should be able to pick this one up too if you have it configured correctly.

https://github.com/filecoin-project/lotus/blob/b1376cfbd2c3316548cf46ecb5626ade196da4d4/gateway/node.go#L272

This is the error we're kind of aiming to replicate, but it's a slightly different since we're not doing this directly in the gateway. Also, let's use the epoch (`ts.Height()`) number rather than tipset key because this is an Ethereum API where tipsets make less sense, so the error can tell them that the number is too far back. Let's make it:

`"the requested block number %d is older than the allowed limit of %d and cannot be retrieved"`

rvagg:
Regarding testing, you should be able to piggy-back an existing itest to run this. Try this:

```diff
diff --git a/itests/fevm_test.go b/itests/fevm_test.go
index 7bbab0c64..babae9eb4 100644
--- a/itests/fevm_test.go
+++ b/itests/fevm_test.go
@@ -1185,6 +1185,17 @@ func TestEthGetBlockReceipts(t *testing.T) {
        gethBlockReceipts, err := client.EthGetBlockReceipts(ctx, req)
        require.NoError(t, err)
        require.Len(t, gethBlockReceipts, 3)
+
+       t.Run("EthGetBlockReceiptsLimited", func(t *testing.T) {
+               // just to be sure we're far enough in the chain for the limit to work
+               client.WaitTillChain(ctx, kit.HeightAtLeast(10))
+               // request epoch 2
+               bn := ethtypes.EthUint64(5)
+               // limit to 5 epochs lookback
+               blockReceipts, err := client.EthGetBlockReceiptsLimited(ctx, ethtypes.EthBlockNumberOrHash{BlockNumber: &bn}, 5)
+               require.ErrorContains(t, err, "older than the allowed")
+               require.Nil(t, blockReceipts, "should not return any receipts")
+       })
 }
 
 func deployContractWithEth(ctx context.Context, t *testing.T, client *kit.TestFullNode, ethAddr ethtypes.EthAddress,
```

Then you can run it with `go test ./itests/fevm_test.go -v -run TestEthGetBlockReceipts` to see it work 🤞

ameeetgaikwad:
> @ameeetgaikwad : I don't know the codebase to be able to speak to the feasibility, but did you look at adding a unit test?

yes i tried adding tests for `EthGetBlockReceiptsLimited ` function. But when I run the tests, it gives this error: `EthGetBlockReceiptsLimited is not in the openrpc spec ⁠`

ameeetgaikwad:
> @ameeetgaikwad check out the lint error from CI, your editor should be able to pick this one up too if you have it configured correctly.
> 
> https://github.com/filecoin-project/lotus/blob/b1376cfbd2c3316548cf46ecb5626ade196da4d4/gateway/node.go#L272
> 
> This is the error we're kind of aiming to replicate, but it's a slightly different since we're not doing this directly in the gateway. Also, let's use the epoch (`ts.Height()`) number rather than tipset key because this is an Ethereum API where tipsets make less sense, so the error can tell them that the number is too far back. Let's make it:
> 
> `"the requested block number %d is older than the allowed limit of %d and cannot be retrieved"`

ok ser! will look into this

BigLep:
@ameeetgaikwad : I know maintainers have been slow in responding here.  We're going to have some breathing room from releases and network upgrades soon.  Is this something you're still up for landing?  If so, have you looked at the list/itest failures?

node/impl/eth/transaction.go
```diff
@@ -412,7 +412,8 @@ func (e *ethTransaction) EthGetTransactionReceiptLimited(ctx context.Context, tx
 	if msgLookup == nil {
 		// This is the best we can do. In theory, we could have just not indexed this
 		// transaction, but there's no way to check that here.
-		return nil, nil
+		return nil, xerrors.Errorf("transaction receipt is too old (limit: %d epochs)", limit)
```
rvagg, line 415: This is a good addition, but you need to be careful of the case where `limit == api.LookbackNoLimit`, so do a check in here for `limit > api.LookbackNoLimit` like you do below and return `nil, nil` if the caller didn't actually pass a limit - because this is the case of just directly calling `EthGetTransactionReceipt()` directly above this function, and why the existing fevm_test.go integration test is failing in CI.

When you have it sorted, run `go test ./itests/fevm_test.go`, it'll take a little while but it'll run through all of the fevm tests and give you output if there's any errors.

node/impl/eth/transaction.go
```diff
@@ -451,6 +452,10 @@ func (e *ethTransaction) EthGetBlockReceiptsLimited(ctx context.Context, blockPa
 		return nil, xerrors.Errorf("failed to get tipset: %w", err)
 	}
 
+	if limit > api.LookbackNoLimit && ts.Height() < e.chainStore.GetHeaviestTipSet().Height()-abi.ChainEpoch(limit) {
```
rvagg, line 455: see the CI error, we now have `make lint` in the Makefile (you may need to rebase, or merge current master in to get it), run that and you should see the lint errors before you push

rvagg:
sorry for the delay here, heads-down in other work but this is very close, just a couple of things to attend to and I've given some advice above about how to ensure it's good to go locally before pushing and getting CI to complain

node/impl/eth/transaction.go
ameeetgaikwad: Alright. Will do

ameeetgaikwad:
> sorry for the delay here, heads-down in other work but this is very close, just a couple of things to attend to and I've given some advice above about how to ensure it's good to go locally before pushing and getting CI to complain

nothing for now. Working on the required changes

node/impl/eth/transaction.go
```diff
@@ -397,7 +397,11 @@ func (e *ethTransaction) EthGetTransactionReceiptLimited(ctx context.Context, tx
 	} else if msgLookup == nil {
 		// This is the best we can do. In theory, we could have just not indexed this
 		// transaction, but there's no way to check that here.
+		if limit > api.LookbackNoLimit {
+			return nil, xerrors.Errorf("transaction receipt is too old (limit: %d epochs)", limit)
+		}
 		return nil, nil
+
```
rvagg, line 404:
```suggestion
```

node/impl/eth/transaction.go
```diff
@@ -397,6 +397,9 @@ func (e *ethTransaction) EthGetTransactionReceiptLimited(ctx context.Context, tx
 	} else if msgLookup == nil {
 		// This is the best we can do. In theory, we could have just not indexed this
 		// transaction, but there's no way to check that here.
+		if limit > api.LookbackNoLimit {
+			return nil, xerrors.Errorf("transaction receipt is too old (limit: %d epochs)", limit)
```
rvagg, line 401:
```suggestion
			return nil, xerrors.Errorf("transaction receipt not found or may be too old (limit: %d epochs)", limit)
```

node/impl/eth/transaction.go
```diff
@@ -397,6 +397,9 @@ func (e *ethTransaction) EthGetTransactionReceiptLimited(ctx context.Context, tx
 	} else if msgLookup == nil {
 		// This is the best we can do. In theory, we could have just not indexed this
 		// transaction, but there's no way to check that here.
+		if limit > api.LookbackNoLimit {
+			return nil, xerrors.Errorf("transaction receipt not found or may be too old (limit: %d epochs)", limit)
+		}
```
rvagg, line 402:
```suggestion
		// This is the best we can do. We may just not have indexed this transaction, or we may have a
		// limit applied and not searched far back enough, but we don't have a way to go. Because
		// Ethereum tooling expects an empty response for transaction-not-found, we don't have a way of
		// differentiating between "can't find" and "doesn't exist".
```

Sorry @ameeetgaikwad, even though you did the hard work here https://github.com/filecoin-project/lotus/pull/12999 came up in the meantime and clarified that we need to return `nil` because it may be that it doesn't exist, even with the limit. So we'll put this comment in here for now and perhaps revisit it if we run into troubles with people not being able to find old transactions.

node/impl/eth/transaction.go
```diff
@@ -436,6 +438,10 @@ func (e *ethTransaction) EthGetBlockReceiptsLimited(ctx context.Context, blockPa
 		return nil, xerrors.Errorf("failed to get tipset: %w", err)
 	}
 
+	if limit > api.LookbackNoLimit && ts.Height() < e.chainStore.GetHeaviestTipSet().Height()-limit {
+		return nil, xerrors.Errorf("tipset %s is too old to fetch receipts for", ts.Key())
```
rvagg, line 442:
```suggestion
		return nil, xerrors.Errorf("tipset %s is older than the allowed lookback limit", ts.Key())
```

PR #12884, opened 2025-02-10
fix(fr32): correctly calculate the todo size under low core counts

tediou5:
## Related Issues

Fixes https://github.com/filecoin-project/lotus/issues/9324

## Proposed Changes

Currently, the **unpadReader**'s work size is calculated using **MTTresh * mtChunkCount(sz)**. When the core count is low, this may cause the todo length to exceed `len(work)`. 

In the current implementation, the length of `out` is actually expanded automatically by Go's underlying mechanisms, making adjustments quite troublesome.

```go
// io.go
func ReadAll(r Reader) ([]byte, error) {
  b := make([]byte, 0, 512)
  for {
    n, err := r.Read(b[len(b):cap(b)])
    b = b[:len(b)+n]
    if err != nil { /* ... */ }

    if len(b) == cap(b) {
      // Add more capacity (let append pick how much).
      b = append(b, 0)[:len(b)] // <-- here
    }
  }
}
```

There are two feasible solutions:

1. Increase the minimum **mtChunkCount** to 16.
2. Ensure that **todo** does not exceed the size of **work**.

I'm not entirely sure if other conditions could also lead to similar issues, so for now, I've chosen the second approach. I added a check when calculating **todo** to prevent overflow:

```go
todo := min(abi.PaddedPieceSize(outTwoPow), abi.PaddedPieceSize(len(r.work)))
```

Additionally I’ve also fixed the CI.

## Checklist

- [ ] CI is green

### DIFF
--- a/storage/sealer/fr32/readers.go
+++ b/storage/sealer/fr32/readers.go
@@ -92,7 +92,7 @@ func (r *unpadReader) readInner(out []byte) (int, error) {
 		return 0, xerrors.Errorf("output must be of valid padded piece size: %w", err)
 	}
 
-	todo := abi.PaddedPieceSize(outTwoPow)
+	todo := min(abi.PaddedPieceSize(outTwoPow), abi.PaddedPieceSize(len(r.work)))
 	if r.left < uint64(todo) {
 		todo = abi.PaddedPieceSize(1 << (63 - bits.LeadingZeros64(r.left)))
 	}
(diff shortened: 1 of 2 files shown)


rvagg:
I've found it hard to context switch deep enough to grok either this or the previous attempt to edit this code so it's hard for me to weigh in.

> There's less data to read each time, which affects performance, I guess.

This _seems_ right but I don't have time to validate it. I think if we just go ahead and roll this out, we'd find out about problems with it if we started hearing about CommP mismatches. We might just have to 🤞 and go ahead unless we have someone else who has the brainspace to drop in here and grok what this code is doing and give a proper 👍 . Maybe Kuba could do that when he's back from his break?

rvagg (approved):

PR #12886, opened 2025-02-11
feat(build): add `make lint` target

rvagg:
Sick of looking up the command manually... Also good to be able to tell contributors about this, e.g. https://github.com/filecoin-project/lotus/pull/12883

I started looking at upgrading the version, but there's dragons in there that are best left for another PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,18 +285,10 @@ unittests:  ## Run unit tests
 	@$(GOCC) test $(shell go list ./... | grep -v /lotus/itests)
 .PHONY: unittests
 
-install-linter:
-	@if ! command -v golangci-lint >/dev/null 2>&1; then \
-		echo "Installing golangci-lint $(GOLANGCI_LINT_VERSION)" && \
-		go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION); \
-	elif ! golangci-lint --version | grep -q "$(GOLANGCI_LINT_VERSION) "; then \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -283,6 +285,20 @@ unittests:  ## Run unit tests
...
+		echo "Installing golangci-lint $(GOLANGCI_LINT_VERSION)" && \
+		go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION); \
+	elif ! golangci-lint --version | grep -q "$(GOLANGCI_LINT_VERSION) "; then \
+		echo "Updating golangci-lint to $(GOLANGCI_LINT_VERSION)" && \
+		go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION); \
+	fi
+.PHONY: install-linter
+
+lint: install-linter
+	golangci-lint run --timeout 10m --concurrency 4
```
masih, line 299: * Use `go run` directly? It will install if not present.
* While at it, add `go mod tidy` and `go vet ./...`?
```suggestion
lint:
	go mod tidy
	go vet ./...
	go run github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION) run --timeout 10m --concurrency 4
```

masih (approved):

Makefile
rvagg: very nice

PR #12887, opened 2025-02-11
chore(lint): upgrade golangci-lint & fix new lint errors

rvagg:
Draft because it's not finished. I'm bailing for now because there's so much to fix. If someone else wants to pick this up in the meantime then you'd be welcome to take it.

The new `G115` gosec errors (integer overflow on conversion) are the most annoying and I'm tempted to disable it entirely because they're not going to be easy to work through, but it's not a bad error to check for so we should probably at least try to deal with them.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,7 +10,7 @@ linters:
     - gosec
     - unconvert
     - staticcheck
-    - exportloopref
+    - copyloopvar
     - unused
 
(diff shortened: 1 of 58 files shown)


rvagg:
Most of this got done on `master` already, the only thing worth doing from this original PR is the `copyloopvar` linter and the simple fixes from that. So this is now pretty simple and ready for review.

PR #12891, opened 2025-02-11
chore: use the correct environment variable for proof params path

masih:
The actual environment variable that sets the proof params fetch path is `FIL_PROOFS_PARAMETER_CACHE` according to `go-paramfetch` implementation. See:

* https://github.com/filecoin-project/go-paramfetch/blob/62ad5888f87076c9589d2b63af904b1f89a91b86/paramfetch.go#L31

Fix it in `Dockerfile` to avoid redundant download of params if the wrong env var is set.


rjan90:
Gave a head-up to Curio [here in Slack](https://filecoinproject.slack.com/archives/C06GD1SS56Y/p1739282345875469), as I see they are also using `FILECOIN_PARAMETER_CACHE` here: https://github.com/filecoin-project/curio/blob/82c295b69d37f91eb0015a90e5b1d16ccfb4c128/Dockerfile#L71-L74

rjan90 (approved):

PR #12895, opened 2025-02-13
chore: disable F3 manifest autofusing on butterflynet for testing purposes

masih:
To be able to test F3 on butterflynet without having to build a bespoke image of lotus or reset the butterflynet change the F3 bootstrap epoch to -1. This disables the autofusing of static manifest.


rjan90 (approved):
SGTM!

PR #12912, opened 2025-02-21
chore: update go-multiaddr to v0.15

MarcoPolo:
Updates the go-multiaddr dependency and this library's usage of go-multiaddr. Rationale for this breaking change can be found in the draft release notes of go-multiaddr: https://github.com/multiformats/go-multiaddr/releases/. But the tl;dr is that this breaking change gets us away from the easy-to-misuse Mutliaddr interface that has led to many nil-ptr panics.

Downstream users like Lotus need this update in drand so that they themselves can update.

Requires:
- [ ] https://github.com/libp2p/go-libp2p/pull/3145
- [ ] https://github.com/libp2p/go-libp2p-pubsub/pull/594
- [ ] https://github.com/libp2p/go-libp2p-kad-dht/pull/1030


rjan90:
Closing this as outdated - we have bumped go-multiaddrs to v0.15 in: https://github.com/filecoin-project/lotus/pull/13173

PR #12915, opened 2025-02-24
chore: parameterise balance in `lotus-seed genesis add-miner`

masih:
Allow the balance to be specified in `lotus-seed genesis add-miner` via `--balance` flag.


rjan90 (approved):
I edited the title from `lotus-shed add-miner` --> `lotus-seed genesis add-miner`. Else it looks good to me

PR #12916, opened 2025-02-24
chore: publish containers for `arm/64` platform and enable build cache

masih:
Publish containers for `arm64` architecture. Because, ARM-based cloud instances are much cheaper to run than other architectures.

While at it, enable caching for docker push step using github actions.


masih:
Fanatically slow docker build. Not worth it; closing.

PR #12919, opened 2025-02-25
feat(miner): remove batch balancer-related functionality

### LINKED ISSUE #12902: FIP 0100: remove batch balancer-related functionality
## Done Criteria 
1. Remove batch balancer, associated config and branching logic
2. Aggregate where possible, batch otherwise where possibleAggregation constraints remain:

* For ProveCommitAggregate and ProveCommitSectors3, the minimum sectors for aggregation is **4** and the maximum is **819**
* For ProveCommitSectorsNI (Non-Interactive) the minimum sectors for aggregation is **1** and the maximum is **65**

Decision matrix should now be:

* PreCommit: prefer batching wherever possible
* Commit:
  * Prefer aggregation on commit where there are at least 4 sectors available for a single commit
  * Otherwise batch where possible

## Why Important
See https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0100.md#change-motivation 

## Notes

tediou5:
## Related Issues

close #12902

## Proposed Changes

Removed BatchPreCommitAboveBaseFee, AggregateAboveBaseFee, and the related checks.

The MinCommitBatch is actually constrained by the SNARK circuits, so it will remain at 4 without change:

  - You need 4 or more sectors to ensure that the verification time for aggregated proofs is smaller than for batched proofs.

  - You need 13 sectors or more to ensure that the aggregated proof size is smaller than the batched proof size.

The MaxCommitBatch limit comes from the original ([FIP-0013](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0013.md#scale-and-limits-1)), and it seems there's not much incentive to modify it, so these two parameters will remain unchanged.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/actors/policy/policy.go.template
+++ b/chain/actors/policy/policy.go.template
@@ -37,6 +37,7 @@ const (
 	SealRandomnessLookback         = ChainFinality
 	PaychSettleDelay               = paych{{.latestVersion}}.SettleDelay
 	MaxPreCommitRandomnessLookback = builtin{{.latestVersion}}.EpochsInDay + SealRandomnessLookback
+	DeclarationsMax                = 3000
 )
 
 var (
@@ -283,6 +284,9 @@ func GetAddressedSectorsMax(nwVer network.Version) (int, error) {
 	}
 }
 
+// GetDeclarationsMax is deprecated
+// 
+// DEPRECATED: remove after nv25 (FIP 0100)
 func GetDeclarationsMax(nwVer network.Version) (int, error) {
 	v, err := actorstypes.VersionForNetwork(nwVer)
 	if err != nil {
@@ -294,6 +298,8 @@ func GetDeclarationsMax(nwVer network.Version) (int, error) {
 			{{if (eq . 0)}}
 				// TODO: Should we instead error here since the concept doesn't exist yet?
 				return miner{{.}}.AddressedPartitionsMax, nil
+			{{else if (ge . 16)}}
+				return DeclarationsMax, nil	
 			{{else}}
 				return miner{{.}}.DeclarationsMax, nil
 			{{end}}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
   - `DealIDs` has now been removed from the public API's `SectorOnChainInfo` (was deprecated in FIP-0079)
   - Removed `--only-cc` from `spcli sectors extend` command
   - Change circulating supply calculation for calibnet, butterflynet and 2k for nv25 upgrade; see ([filecoin-project/lotus#12938](https://github.com/filecoin-project/lotus/pull/12938)) for more information
+- feat(miner): remove batch balancer-related functionality ([filecoin-project/lotus#12919](https://github.com/filecoin-project/lotus/pull/12919))
 
 # UNRELEASED v.1.32.0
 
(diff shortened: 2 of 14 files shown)


BigLep:
@tediou5 : no pressure, but is this ready for review?  If so, lets move it out of draft to make it clear.  Thanks.

BigLep:
@tediou5 : thanks for working on this.  We wanted to make sure you saw https://github.com/filecoin-project/lotus/issues/12902#issuecomment-2686372897

tediou5:
> @tediou5 : thanks for working on this. We wanted to make sure you saw [#12902 (comment)](https://github.com/filecoin-project/lotus/issues/12902#issuecomment-2686372897)

@BigLep Got it, let me check the aggregation part of the handling.

tediou5:
@BigLep Yes, I think it's ready for review now. I'd like to get some feedback as well since I'm not very familiar with this part of the code, and I'm a bit worried I might have missed something.

Stebalien:
Please run `make gen` and commit the results. 

Also, this appears to now be submitting batches that are too large (See the failing tests):

> batch of 1000 too large, max 256

rvagg:
Sorry for the delay in bringing up this important point: I think we're going to have to retain all of this functionality but gate it behind a network version check, at least until the feature is live (nv25?), then we can remove it. The problem is that if we roll out a version of lotus-miner with all of this removed, we're either going to throw users to the wolves of not having this functionality in place during a period where there _is_ the batch balancer in operation (nv24), or advise them to upgrade just after the upgrade that activates FIP-0100.

So instead, we should:

* Retain the config settings but mark them clearly as `DEPRECATED` and note they will be removed in the next release
* Gate everything with a call to `StateNetworkVersion` and do a `if nv < network.Version25` around everything that involves the batch balancer.

e.g.

* https://github.com/filecoin-project/lotus/blob/a474d459b2b3be1d888f710e455c843b14e580dc/storage/wdpost/wdpost_run.go#L229-L230 for 1.1 proofs
* https://github.com/filecoin-project/lotus/blob/a474d459b2b3be1d888f710e455c843b14e580dc/storage/pipeline/piece/piece_info.go#L81-L84 for PCS3
* https://github.com/filecoin-project/lotus/blob/a474d459b2b3be1d888f710e455c843b14e580dc/storage/pipeline/states_sealing.go#L41 remnant of the DDO PCS3 pipeline that hasn't been removed

^ we should go through and remove all of these too, in a separate PR

BigLep:
@tediou5 : do you think you'll be able to make the updates so we can get this merged by the end of the week (2025-03-14) so an RC can be cut at the beginning of next week (2025-03-17)?

tediou5:
> @tediou5 : do you think you'll be able to make the updates so we can get this merged by the end of the week (2025-03-14) so an RC can be cut at the beginning of next week (2025-03-17)?

@BigLep I think I can get it done. Let me give it a try.

tediou5:
@Stebalien @rvagg @BigLep Restored the previous logic and now execute it conditionally based on the network version.

tediou5 (commented):
Thank you for your review.

BigLep:
@tediou5 : I'm trying to follow where this is at.  Has all feedback been incorporated except for where you have questions?

Stebalien (approved):
LGTM. We're going to need to do this the right way without hard-coded constants, but this PR has the MVP for the upgrade.

chain/actors/policy/policy.go.template
```diff
@@ -1,6 +1,8 @@
 package policy
 
 import (
+	"math"
+
```
Stebalien, line 5:
```suggestion
	"math"

```

rvagg (approved):

tediou5:
> LGTM. We're going to need to do this the right way without hard-coded constants, but this PR has the MVP for the upgrade.

Yes, I agree. I prefer to estimate gas so that messages are only constrained by MaxBlockGasLimit rather than a fixed constant. But I think I'll submit this in a separate PR.

PR #12923, opened 2025-02-27
feat: allow genesis timestamp to be specified in `lotus-seed genesis new`

masih:
Add the ability to specify the genesis timestamp in `lotus-seed genesis new` command in RFC-3339 format.

### DIFF
--- a/cmd/lotus-seed/genesis.go
+++ b/cmd/lotus-seed/genesis.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


rjan90 (approved):
Looks good to me. For my own understanding, what issue/missing feature did you encounter when not having this ability?

masih:
> what issue/missing feature did you encounter when not having this ability

The need to fiddle with jq like [this](https://github.com/filecoin-project/lotus-infra/blob/main/ansible/setup_fildev_network.bash#L195-L199) othrwise.

Issue #12897, opened 2025-02-13
Node crashed, and stop syncing with EOF


emmanuelm41:
### Checklist

- [x] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [x] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [x] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [x] I did not make any code changes to lotus.

### Lotus component

- [x] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon: v1.31.1
```

### Repro Steps

1. Run the node with ChainIndexer running
2. Wait until some random block fails, and crashes the node entirely.

### Describe the Bug

Once the node starts, it crashes for some reason in some random block. So far this happened twice. The first one was fix by upgrading the node from v1.31.0 to v1.31.1. The second one is still preventing our node to start. This is a full archival node. 

### Logging Information

```text
2025-02-13T16:23:32.379Z	INFO	badgerbs	v2@v2.2007.4/value.go:1158	Replaying file id: 42814 at offset: 723470378
2025-02-13T16:23:32.379Z	INFO	badgerbs	v2@v2.2007.4/value.go:1178	Replay took: 6.132µs
[...]

PR #12930, opened 2025-03-04
fix: allow users to optionally configure node startup even if index reconciliation fails

aarshkshah1992:
For https://github.com/filecoin-project/lotus/issues/12897.

There chain state has some missing messages and this PR enables them to start the node even if index reconciliation fails.

### DIFF
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -640,6 +640,8 @@ type ChainIndexerConfig struct {
 	// - This ensures that the index is always in a consistent state with the chain before the node starts.
 	//
 	// Default: false
+	// // WARNING: Only set to true if you are okay with an index that may be out of sync with the chain.
+	// This can lead to inaccurate or missing data in RPC responses that depend on the indexer.
 	AllowIndexReconciliationFailure bool
 }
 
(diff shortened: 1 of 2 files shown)


rvagg:
This is pretty blunt. @emmanuelm41 can you consider this change and comment on whether it makes your life any better? This gives you a flag that lets you start the node up, log an error, but the reconciliation is aborted entirely. It doesn't address the messages being missing (that's still a mystery for a full archival node, we really need to get to the bottom of that but it's separate to the chainindexer tripping over it). It means there may be a small gap in what's been indexed, although it can always manually run a backfill over epochs you're concerned about with `lotus index validate-backfill`.

aarshkshah1992:
@emmanuelm41 

- Based on the error message at https://github.com/filecoin-project/lotus/issues/12897, that node's chain state is missing some messages at epoch 4702087.  
- This PR enables you to start your node with the `AllowIndexReconciliationFailure` flag enabled (false by default, you need to set it to true) which means that your node will boot up even if it fails to reconcile the Index with the chain state.

- Note the `validate-backfill` CLI will also fail for epoch 4702087. 
- Ideally, what you want to do is `validate-backfill` from the latest epoch upto epoch **_4702087_** and then `validate-backfill` from epoch **_4702086_** upto genesis in the background while the node is running so you don't need the reconciliation at all to backfill the Index. 

Let me know if you have any questions.

rvagg (approved):

PR #12574, opened 2024-10-10
fix: set address current network for non-mainnet since default is changed

masih:
## Related Issues
* https://github.com/filecoin-project/go-address/commit/e08728680fb

## Proposed Changes
Default `address.CurrentNetwork` was changed in an upstream dependency. Update the build constants for non-mainnet to explicitly set it to `Testnet`.

Add tests to assert it.
## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [ ] PR title conforms with [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#pr-title-conventions)
- [ ] Update CHANGELOG.md or signal that this change does not need it per [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#changelog-management)
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12932, opened 2025-03-05
feat: nv25 and nv26 - fip-0100 (and some FIP-0098)

rvagg:
Dev builds of builtin-actors and go-state-types. Still needs a bit more work in actors and we still need a migration and invariants in go-state-types and we need a new ref-fvm to go with it. But we can start looking at work in here too.


rvagg:
Sadly I think I've introduced a flaky that I'll need to make more robust. After 18 runs locally of https://github.com/filecoin-project/lotus/pull/12952 I got a fail:

```
                Error Trace:    /home/rvagg/go/src/github.com/filecoin-project/lotus/itests/daily_fees_test.go:222
                                                        /home/rvagg/go/src/github.com/filecoin-project/lotus/itests/daily_fees_test.go:625
                Error:          Not equal:
                                expected: 9047612
                                actual  : 8583626
                Test:           TestDailyFees
                Messages:       expected miner to have burned 9047612, but it's burned 8583626
```

We seem to have run the calculation on the wrong side of a deadline so I'll have to figure out how to be more precise.

rjan90:
Been running some migration benchmarks based on this branch here: https://github.com/filecoin-project/lotus/issues/12959#issuecomment-2729459407. Both in "offline" and "online" modes.

Stebalien:
Note: I've confirmed that the v16 bundle matches https://github.com/filecoin-project/builtin-actors/commit/cd4378471b64fbabdaa257a77bd962f8702c1984 (one commit behind the target commit due to https://github.com/filecoin-project/builtin-actors/issues/1656, but it's good enough for an RC).

build/buildconstants/params_calibnet.go
```diff
@@ -30,80 +30,77 @@ var ActorDebugging = false
 const BootstrappersFile = "calibnet.pi"
 const GenesisFile = "calibnet.car.zst"
 
-const UpgradeBreezeHeight = -1
+const UpgradeBreezeHeight abi.ChainEpoch = -1
```
Stebalien, line 33: Are the types necessary?

Stebalien: Ah, is this to help unify with `params_2k` to ensure that if this builds, that builds?

build/openrpc/full.json
```diff
@@ -21479,7 +21484,7 @@
                                 "SealedCID": {
                                     "/": "bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4"
                                 },
-                                "DealIDs": [
+                                "DeprecatedDealIDs": [
```
Stebalien, line 21487: https://github.com/filecoin-project/go-state-types/pull/374

chain/actors/builtin/miner/miner.go
```diff
@@ -10,6 +10,7 @@ import (
 	actorstypes "github.com/filecoin-project/go-state-types/actors"
 	"github.com/filecoin-project/go-state-types/big"
 	minertypes13 "github.com/filecoin-project/go-state-types/builtin/v13/miner"
+	minertypes16 "github.com/filecoin-project/go-state-types/builtin/v16/miner"
 	minertypes "github.com/filecoin-project/go-state-types/builtin/v9/miner"
```
Stebalien, line 14: Can we unify these to v16?

cli/spcli/sectors.go
```diff
@@ -768,7 +768,7 @@ func SectorsExtendCmd(getActorAddress ActorAddressGetter) *cli.Command {
 				if !found {
 					return xerrors.Errorf("sector %d is not active", id)
 				}
-				if len(si.DealIDs) > 0 && cctx.Bool("only-cc") {
+				if len(si.DeprecatedDealIDs) > 0 && cctx.Bool("only-cc") {
 					continue
 				}
```
Stebalien, line 773: Isn't this check kind of broken now?

cli/state.go
```diff
@@ -1505,16 +1505,24 @@ var StateSectorCmd = &cli.Command{
...
+		edr := big.Zero()
+		if si.ExpectedDayReward != nil {
+			edr = *si.ExpectedDayReward
+		}
+		fmt.Println("ExpectedDayReward: ", types.FIL(edr))
+		esp := big.Zero()
+		if si.ExpectedStoragePledge != nil {
+			esp = *si.ExpectedStoragePledge
+		}
+		fmt.Println("ExpectedStoragePledge: ", types.FIL(esp))
```
Stebalien, line 1525: I'd consider just not printing these fields in this case.

cli/state.go
```diff
@@ -1505,16 +1505,24 @@ var StateSectorCmd = &cli.Command{
 		if si.SectorKeyCID != nil {
 			fmt.Println("SectorKeyCID: ", si.SectorKeyCID)
 		}
-		fmt.Println("DealIDs: ", si.DealIDs)
+		fmt.Println("DealIDs (deprecated): ", si.DeprecatedDealIDs)
```
Stebalien, line 1508: I'd just drop this entire field.

cli/state.go
Stebalien: Ideally, we'd just ignore these fields entirely after nv25 (i.e., get the network version for the given tipset).

cmd/lotus-shed/sectors.go
```diff
@@ -710,7 +710,7 @@ var dumpSectorOnChainInfoCmd = &cli.Command{
 					maddr,
 					sector.SectorNumber,
 					sector.SealProof,
-					len(sector.DealIDs),
+					len(sector.DeprecatedDealIDs),
```
Stebalien, line 713: this is broken now, I'd just remove it entirely.

cmd/lotus-shed/terminations.go
```diff
@@ -164,7 +164,7 @@ var terminationsCmd = &cli.Command{
 					}
 
 					for _, sector := range sectors {
-						for _, deal := range sector.DealIDs {
+						for _, deal := range sector.DeprecatedDealIDs {
```
Stebalien, line 167: Also broken, remove it.

Stebalien: (or fix it, but that involves digging into the market actor.

build/buildconstants/params_mainnet.go
```diff
@@ -113,22 +122,19 @@ const UpgradeWaffleHeight abi.ChainEpoch = 4154640
...
 
-// This fix upgrade only ran on calibrationnet
-const UpgradeWatermelonFix2Height abi.ChainEpoch = -2
+// This epoch, 90 days after Teep is the completion of FIP-0100 where actors will start applying
+// the new daily fee to pre-Teep sectors being extended.
+var UpgradeTockHeight = UpgradeTeepHeight + builtin.EpochsInDay*90
 
-// This fix upgrade only ran on calibrationnet
-const UpgradeCalibrationDragonFixHeight abi.ChainEpoch = -3
+var UpgradeTeepInitialFilReserved = InitialFilReserved // FIP-0100: no change for mainnet
```
Stebalien, line 137: should also be constants on mainnet.

build/buildconstants/params_calibnet.go
rvagg: It's not essential, it's just consistent, it doesn't get properly checked because of the build excludes. I'll back it out if you object but the inconsistency got to me after it tripped me up with one particular codesite (can't remember now, but it was do to with epoch arithmetic where it didn't know it should be a `ChainEpoch`.

build/buildconstants/params_mainnet.go
```diff
@@ -43,7 +43,7 @@ const UpgradeSmokeHeight abi.ChainEpoch = 51000
 const UpgradeIgnitionHeight abi.ChainEpoch = 94000
 const UpgradeRefuelHeight abi.ChainEpoch = 130800
 
-const UpgradeAssemblyHeight abi.ChainEpoch = 138720
+var UpgradeAssemblyHeight abi.ChainEpoch = 138720
```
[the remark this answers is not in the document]
rvagg: https://github.com/filecoin-project/lotus/blob/8775c5b6071f7b98c916ab745fcf077f27d9323a/itests/kit/ensemble.go#L180-L188

because of this:

https://github.com/filecoin-project/lotus/blob/8775c5b6071f7b98c916ab745fcf077f27d9323a/chain/stmgr/supply.go#L246-L251

build/buildconstants/params_mainnet.go
rvagg: can't because `InitialFilReserved` isn't, because we calculate it from whole FIL

build/buildconstants/params_calibnet.go
Stebalien: No, I've been hit by it before as well.

build/buildconstants/params_mainnet.go
Stebalien: Ah, I see. But what about the heights?

chain/actors/builtin/miner/miner.go
rvagg: Yes, but it's costly: https://github.com/filecoin-project/lotus/pull/12960

build/buildconstants/params_mainnet.go
rvagg: no, because we have `LOTUS_DISABLE_TEEP` set; but I assume that once we bump to nv27 that we can go back and make those consts, maybe that's on the checklist? @rjan90 ?

cli/spcli/sectors.go
rvagg: removed `--only-cc` and put an error in when it's invoked

cmd/lotus-shed/terminations.go
rvagg: removed the cmd, if someone wants it back then it's going to look quite different anyway

rvagg:
OK, big 'ol update here:
* go-state-types latest, which removes `DeprecatedDealIds` from `SectorOnChainInfo` in the JSON API response (👌)
* Removed `spcli sectors extend --only-cc` cause it relies on `DealIds` which hasn't been useful since FIP-0076
* Refactored the itest so it's a bit easier to read and follow now (IMO), and shorter and I think I've removed the flake by being much more precise with counting proving deadlines for sectors so we can calculate fees properly (deadline maths is hard)
* Updated CHANGELOG with much more detail about this PR

This review comment: https://github.com/filecoin-project/lotus/pull/12932#discussion_r1999225369 - has resulted in this draft PR: https://github.com/filecoin-project/lotus/pull/12960 - it's not small and needs a little bit more work, I'm not sure we want to do it for this network upgrade but if we do I think it should be a separate PR and we leave the duplication alone in here.

build/buildconstants/params_mainnet.go
Stebalien: AH.

build/buildconstants/params_mainnet.go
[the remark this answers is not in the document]
Stebalien: I didn't notice that, that's awful. Setting anything in there is awful but at least `EquivocationDelaySecs` is safe-ish.

I'm going to go comment over there now.

itests/kit/ensemble.go
```diff
@@ -177,6 +177,15 @@ func NewEnsemble(t *testing.T, opts ...EnsembleOpt) *Ensemble {
...
 
+	// See FIP-0100; we manually set these to the 2k network settings for itests not run with -tags 2k
+	buildconstants.UpgradeAssemblyHeight = -1 // so GetVMCirculatingSupplyDetailed() properly calculates FilReserved
+	// 700M FIL is the correct value for test networks, with UpgradeAssemblyHeight in the past, this
+	// value is used in supply.go to calculate the mining reserve in circulation which is added to
+	// circulating supply.
+	buildconstants.InitialFilReserved = types.MustParseFIL("700000000 FIL").Int
+	// Bump it at Teep upgrade to simulate FIP-0100 and get CS close to ~700M FIL like in mainnet.
+	buildconstants.UpgradeTeepInitialFilReserved = types.MustParseFIL("1400000000 FIL").Int
+
```
Stebalien, line 188: We really shouldn't do this. We shouldn't set `EquivocationDelaySecs` either, but we were trying to quickly fix a terribly broken thing at the time (it was, IIRC, also pretty necessary in order to get the tests to work.

1. Why do we _care_ what the initial FIL reserved in tests is? At a minimum, there should be no need to alter `InitialFilReserved` unless I'm mistaken.
2. Do we actually hit `UpgradeAssemblyHeight` in unit tests?
3. In terms of `UpgradeTeepInitialFilReserved`, can we adjust it in the test that actually needs it to change (restoring it at the bottom)?

Alternatively... if we set these in `build_mainnet_test.go`, do they get set when running all tests? That may be an alternative that's less hidden (although I'd still try to avoid, e.g., altering `InitialFilReserved` if at all possible.

chain/actors/builtin/miner/miner.go
Stebalien: Commented there.

Stebalien (approved):
LGTM modulo my one remaining open comment. But I'm OK with merging this even with that outstanding.

itests/kit/ensemble.go
rvagg: The aim here is to make our unit tests run in a more realistic environment. Today they all operate in pre-Assembly mode, which probably isn't terrible because it just removes reserve accounting from the calculation, but what I was attempting to do here is push the tests to operate in conditions that are not too dissimilar to mainnet. We can only do that by pushing past Assembly (`138720`) and then making sure CS calculations don't resolve to `0` (because `InitialFilReserved` is 300M when it's actually set at 700M for the test genesis). `UpgradeAssemblyHeight=-1` and setting `InitialFilReserved` to 700M starts us off at a reserve disbursed of `0` and then Teep throws us into a CS that's close to mainnet. Because Teep is now the default `TestNetworkVersion` they're all running with a CS that's close to mainnet.

What do you mean by `build_mainnet_test.go`, are you talking about `params_mainnet_test.go` introduced in https://github.com/filecoin-project/lotus/pull/12574 or making a new file? I'm not sure that helps us with this since it's only touched if run directly and we run all these itests directly by referencing the file.

I think a `-tag 2k` would make this all go away, but if we do that in CI it's unlikely to also be done when people run it locally and will just end up being confusing.

build/buildconstants/params_calibnet.go
```diff
@@ -113,6 +110,15 @@ const UpgradeTeepHeight = 2520574
...
 var UpgradeTuktukPowerRampDurationEpochs = uint64(builtin.EpochsInDay * 3)
 
+// 2025-03-25T23:00:00Z
+const UpgradeTeepHeight abi.ChainEpoch = 2520574
+
+var UpgradeTeepInitialFilReserved = wholeFIL(1_200_000_000) // FIP-0100: 300M -> 1.2B FIL
+
+// This epoch, 90 days after Teep is the completion of FIP-0100 where actors will start applying
+// the new daily fee to pre-Teep sectors being extended.
+var UpgradeTockHeight abi.ChainEpoch = UpgradeTeepHeight + builtin.EpochsInDay*90
```
rjan90, line 120: I think we landed on that the Tock upgrade should be 7 days after the Teep upgrade?
```suggestion
// This epoch, 7 days after Teep is the completion of FIP-0100 where actors will start applying
// the new daily fee to pre-Teep sectors being extended.
var UpgradeTockHeight abi.ChainEpoch = UpgradeTeepHeight + builtin.EpochsInDay*7
```

rvagg: good catch, thanks!

rvagg:
I think this is good to merge (rebase merge pls, I think the commit history is fairly clean). I wouldn't mind @Stebalien having a look at my response above to his last outstanding review. We could deal with it, but we're coping with discrepancies between itest environment and mainnet even though it runs in mainnet mode but with 2k sectors (i.e. params_mainnet) and we have a weird genesis with an initial reserve of 700M. So those changes are to adapt to that and bring it close to mainnet.

itests/kit/ensemble.go
Stebalien: > I was attempting to do here is push the tests to operate in conditions that are not too dissimilar to mainnet.

I see. In that case, we really need to find a better way to configure these things. But I can live with it for now.

My objection is that this is an unseen side-effect.

- This probably won't cause issues for anyone in production because they're unlikely to use the testing framework, but it's still really nasty.
- This can definitely cause issues for anyone who reads these variables before constructing the node.
- This makes `t.Parallel` impossible. That ship already partially sailed, but this makes the issue worse.

I'm OK landing this for now, but we've really got to fix this. The wider issue is that we should be using these constants to construct some form of "config" object which we should be using internally.

> InitialFilReserved to 700M starts us off at a reserve disbursed of 0 and then Teep throws us into a CS that's close to mainnet. Because Teep is now the default TestNetworkVersion they're all running with a CS that's close to mainnet.

This I disagree with. FIL Reserved should match _mainnet_ and I don't want our tests running with a fake circulating supply that doesn't match the actual tokens available.

I don't think having "similar" fees to mainnet matters at all in unit tests.

> What do you mean by build_mainnet_test.go, are you talking about params_mainnet_test.go introduced in https://github.com/filecoin-project/lotus/pull/12574 or making a new file? I'm not sure that helps us with this since it's only touched if run directly and we run all these itests directly by referencing the file.

`_test.go` files get built when testing. I've used them in the past to add test-only extensions/alterations. I _know_ that works when testing within a package but I _thought_ it worked when testing across packages? But I can't remember.

> I think a -tag 2k would make this all go away, but if we do that in CI it's unlikely to also be done when people run it locally and will just end up being confusing.

Yeah, please don't do that.

rvagg: > `_test.go` files get built when testing

but with itests we `go test ./itests/daily_fees_test.go` both locally and in CI so in that case I don't think it touches anything else that isn't directly included; so I don't think that's an option.

Stebalien: Yeah, ok, that doesn't work.

Stebalien (approved):

itests/kit/ensemble.go
rvagg: for future reference, we resolved this in `fix: reduce amount of buildconstant hacking in itests` - now we only have the one extra constant to mess with in the ensemble globally, we soak up FilBase when making initial accounts in genesis so we have 300M reserved and we move the teep reserve bump into the test itself.

PR #12935, opened 2025-03-06
fix: delegate integer operations to go-state-types

Stebalien:
## Proposed Changes

Delegates most calls to go-state-types and treats "nil" as "zero" (we encode nil as zero, so we might as well treat it as such when doing math).

## Additional Info

@rvagg just in case we want to do it this way.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/types/bigint.go
+++ b/chain/types/bigint.go
@@ -31,46 +31,51 @@ func BigFromBytes(b []byte) BigInt {
 }
 
 func BigFromString(s string) (BigInt, error) {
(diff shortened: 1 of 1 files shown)


Stebalien:
Honestly, I think we should merge this no matter how we want to handle nil/zeros.

rvagg (approved):
Agreed, this is good regardless, but let's see if that test failure was a flaky or not first

Stebalien:
Looks like they're all passing now.

Issue #12136, opened 2024-06-22
[DX Streamline] LOTUS_RUN_EXPENSIVE_TESTS periodically outside of PR workflow


BigLep:
## Done Criteria
1. LOTUS_RUN_EXPENSIVE_TESTS are run periodically (e.g., nightly) against master automatically.  They aren't run as part of the normal PR flow because of how long they take.
2. There is notification via github issue or Slack if there are any failures so that maintainers can try and identify the culprit change.
3. There is a self-service URL to find the previous results of LOTUS_RUN_EXPENSIVE_TESTS. 

## Why Important
There are "expensive" tests that test key functionality.  This is getting exasperated by NI-PoRep tests which can take 10+ minutes on CI.  We want to avoid compromising on test quality because of long CI runs.  We want the test coverage but not to be slowed down by it for every CI run.

## User/Customer
Contributors and maintainers

## Notes
1. This was discussed in #fil-lotus-dev on 2024-06-20 in https://filecoinproject.slack.com/archives/CP50PPW2X/p1718859677050149

rvagg:
Currently `LOTUS_RUN_EXPENSIVE_TESTS=1` is the same as `CI=true`, which is why I proposed that `LOTUS_RUN_VERY_EXPENSIVE_TESTS` introduces a new level that we would run periodically.

rvagg:
`master` now includes one test that's behind `LOTUS_RUN_VERY_EXPENSIVE_TESTS`, so this has become a slightly higher priority now

galargh:
I think it'd be good if there was also a way to request the run of very expensive tests on a PR (e.g. through a label) and if they run when the test itself was modified

PR #12234, opened 2024-07-15
ci: tests: adds tooling to run very expensive tests

laurentsenta:
## Related Issues
Closes https://github.com/filecoin-project/lotus/issues/12136

## Proposed Changes

- [x] Keep the `LOTUS_RUN_EXPENSIVE_TESTS` flag unchanged, but make it explicit in CI
- [x] Adds the `LOTUS_RUN_VERY_EXPENSIVE_TESTS` flag based on a few signals (wip)
    - [x] If a PR contains the label `need/very-expensive-tests`
    - [x] During a nightly run
- [x] Adds a `very_expensive_tests` array in the `test.yml` settings (similar to how test are configured for custom runners, etc).
    - [x] Tests in that array will run with `-timeout=60m`

## Additional Info

Currently the test using VERY EXPENSIVE is `niporep_manual_test.go`.

Example of a regular run:
https://github.com/filecoin-project/lotus/actions/runs/9908341740/job/27439500093?pr=12225
You'll find `skipping test` logs in the `gotestsum` step.

Adding the `need/very-expensive-tests` label should toggle those on.
 
## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title is in the form of of `<PR type>: <area>: <change being made>`
  - example: ` fix: mempool: Introduce a cache for valid signatures`
  - `PR type`: fix, feat, build, chore, ci, docs, perf, refactor, revert, style, test
  - `area`, e.g. api, chain, state, mempool, multisig, networking, paych, proving, sealing, wallet, deps
- [x] Update CHANGELOG.md or signal that this change does not need it.
  - If the PR affects users (e.g., new feature, bug fix, system requirements change), update the CHANGELOG.md and add details to the UNRELEASED section.
  - If the change does not require a CHANGELOG.md entry, do one of the following:
    - Add `[skip changelog]` to the PR title
    - Add the label `skip/changelog` to the PR
- [x] New features have usage guidelines and / or documentation updates in
  - [x] [Lotus Documentation](https://lotus.filecoin.io)
  - [x] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [x] Tests exist for new functionality or change in behavior
- [x] CI is green

PR #12939, opened 2025-03-06
feat: set up very expensive tests to run in CI

galargh:
## Related Issues

Closes https://github.com/filecoin-project/lotus/issues/12136

## Proposed Changes

In this PR I set up a new workflow responsible for running very expensive tests. I also enable running expensive tests in our regular test workflow.

## Additional Info
This is an updated version of https://github.com/filecoin-project/lotus/pull/12234

In this iteration, the very expensive tests workflow is separate to the test workflow. This is so that we don't have to rerun all the tests on a label change.

The way the new workflow works is that when one adds a `need/very-expensive-tests` label to a PR, the workflow will start, remove the label, and execute the tests.

The new workflow will only execute the tests that have very expensive components to them. The information about that is currently stored in the `ci` cmd tool. It will be moved to the test files themselves in the future once we move more towards the self-identification setup.

This is not ready to be merged because the only very expensive test we have at the moment is failing with `signal: killed`. Here's a link to the most recent failure: https://github.com/filecoin-project/lotus/actions/runs/13706584703

Question: should the workflow comment on a PR with the very expensive test results? I think it might be useful since if one modifies labels after the very expensive tests are run, then the link to the very expensive tests run would disappear from the PR.

## Checklist

- [ ] CI is green


.github/workflows/test.yml
```diff
@@ -20,201 +16,6 @@ permissions:
...
-      - if: success() || failure()
-        uses: actions/upload-artifact@v4
-        with:
-          name: ${{ matrix.name }}-${{ runner.os }}-${{ runner.arch }}
-          path: |
-            ${{ steps.reports.outputs.path }}/${{ matrix.name }}.xml
-            ${{ steps.reports.outputs.path }}/${{ matrix.name }}.json
-        continue-on-error: true
+    name: Test
+    uses: ./.github/workflows/reusable-test.yml
```
BigLep, line 21:
```suggestion
    uses: ./.github/workflows/reusable-test.yml
    with:
      run_very_expensive_tests: false
```

I know that's the default, but it makes it explicit that this workflow isn't running with expensive tests.

cmd/ci/main.go
```diff
@@ -63,6 +63,12 @@ func main() {
 			{
 				Name:  "list-test-group-execution-contexts",
 				Usage: "List all test group execution contexts",
+				Flags: []cli.Flag{
+					&cli.BoolFlag{
+						Name:  "very-expensive-tests-run",
+						Usage: "Whether the groups not containing any very expensive tests should be filtered out",
```
BigLep, line 69:
```suggestion
						Usage: "Whether to only include the groups with very expensive tests",
```

BigLep: Does this mean that if the boolean flag is not set that our test run includes all tests (i.e., not-very-expensive-tests + very-expensive-tests)?

BigLep (commented):
Sweet - thanks for doing this!

> the very expensive tests workflow is separate to the test workflow. This is so that we don't have to rerun all the tests on a label change.

Makes sense - sounds good.

> remove the label

Why do we remove `need/very-expensive-tests`?  I assume because this is the way to manually trigger the expensive test run for the PR.

I guess in a perfect world if the `need/very-expensive-tests` label was present, the expensive test groups would become a required CI check and would automatically get run whenever other tests are run.

> The information about that is currently stored in the ci cmd tool. It will be moved to the test files themselves in the future once we move more towards the self-identification setup.

Sounds good.

> Question: should the workflow comment on a PR with the very expensive test results? I think it might be useful since if one modifies labels after the very expensive tests are run, then the link to the very expensive tests run would disappear from the PR.

I think PR's labeled with `need/very-expensive-tests` should have the comment in the PR.  (Ideally it would behave the same as every other test, except this PR just has extra tests being run as a result of the label.). 

For a PR, I don't think we should be cutting issues if very expensive tests fail.  I think we should only cut an issue if the very expensive cron job fails.

galargh:
> Why do we remove need/very-expensive-tests? I assume because this is the way to manually trigger the expensive test run for the PR.
> 
> I guess in a perfect world if the need/very-expensive-tests label was present, the expensive test groups would become a required CI check and would automatically get run whenever other tests are run.
>

I was trying to avoid the need to rerun the very expensive tests whenever unrelated labels are changed because these tests are expensive to run. That's also the reason why I didn't want to bundle them with the other tests. The problem is that we cannot configure the workflow to trigger only when a specific label changes. It's all or nothing.

However, I did get another idea for how we might approach it. Maybe we could drop the removal of the label and, instead, if an unrelated label is modified, we would look up the last run of the very expensive tests and just copy the result from it. WDYT?

> I think PR's labeled with need/very-expensive-tests should have the comment in the PR. (Ideally it would behave the same as every other test, except this PR just has extra tests being run as a result of the label.).

If we implement that new idea, we wouldn't have to worry about the comments.

> For a PR, I don't think we should be cutting issues if very expensive tests fail. I think we should only cut an issue if the very expensive cron job fails.

👍

cmd/ci/main.go
galargh: Yes, but it's worth noting that very-expensive-tests is a sub-group of not-very-expensive-tests.

rvagg:
Looking at the bad test: `LOTUS_RUN_VERY_EXPENSIVE_TESTS=1 go test ./itests/niporep_manual_test.go -run TestManualNISectorOnboarding/real_proofs,_1_miner_with_7_sectors,_1_bad -v`

At the peak, only about 700 epochs in, which is where the test is getting killed (quite early, this isn't timeout related) we get to ~32GiB of RAM, so I assume this is the limitation we're dealing with for these runners?

```
3836338 rvagg     20   0 4226.9g  31.6g   2.7g S  3064  16.8   7:55.08 /tmp/go-build2459112198/b001/itests.test -test.testlogfile=/tmp/go-build2459112198/b001/test+
```

It goes down from there, this is the initial part where it's generating the initial proof so it's a lot of work. There's only this one case that's in un here and it's the `real_proofs` that make it "very expensive". What options do we have on the runner side here because I'm not sure how much lattitude we have on the software side because I'm suspecting this is almost all in the proofs code; but I can explore that further if we have no options. Could we add some small swap just for this case if we're strictly limited to 32GiB? We shouldn't need much at all, I think we're probably just pushing a little over 32GiB for the system in this run.

rvagg:
btw once we get into it, it goes way down:

```
3836338 rvagg     20   0 4185.4g   3.7g   1.3g S  53.3   2.0 228:35.20 /tmp/go-build2459112198/b001/itests.test -test.testlogfile=/tmp/go-build2459112198/b001/test+
```

but it does take a long time to run this test because of the amount of time needed to tick through all the epochs required. In my test run it's got up to 15,478 epochs 556.76s; on a fairly fast system.

BigLep:
@galargh : apologies for the delay here.  A few things:

For getting the current `VERY_EXPENSIVE_TEST` to pass, it seems like we need more memroy (slightly more than 32GB).  Can this be accomplished with beefier runners or can we enable some swap space on the runners we have ([more info](https://github.com/filecoin-project/lotus/pull/12939#issuecomment-2764994040))?  

In terms of the interaction model:
1. Periodic running where failures cut an issue - it seems like we're covered here ✅
2. Selectively running at times a manual step in the Lotus release process or during a PR that we think warrants an explicit change - It sounds like the model we're looking at is triggering a a test run by applying a label.  I am understanding the interaction better where applying the PR "run expensive tests" label queues up the test run and when the test run completes, the results are linked as a comment in the PR and the PR "run expensive tests" label is removed.  The PR label is removed so that there is a way to explicitly re-request "run expensive tests" on the PR again (e.g., after pushing additional commits).  I think that works and if that's a lot easier from an implementation regard, lets go with it.  I don't want to block this PR further on best being the enemy of better.  
   - I was envisioning a world where whenever we'd run our tests normally there would be a step to check if the PR "run expensive tests" label was present on the PR and if so also run expensive tests and if not skip the expensive tests.  But per above, if that isn't easily feasible, lets skip that.

galargh:
I moved the tests to a slightly bigger runner and it looks like we're through :) I also updated the very expensive tests workflow so that it behaves similarly to our regular one. The one difference is, it will also get triggered whenever a label is added to a PR.

.github/workflows/very-expensive-test.yml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      # WARN: This will run very expensive test every time ANY label is added
+      - labeled
+  schedule:
+    - cron: '0 0 * * *' # Runs nightly at 0AM UTC
```
rvagg, line 12: any idea what it's going to cost to run this daily? given the current very expensive and how long they take to run on the current hardware, are we looking at cents or dollars here?

.github/workflows/reusable-test.yml
```diff
@@ -0,0 +1,233 @@
...
+          PACKAGES: ${{ join(fromJson(steps.group.outputs.metadata).packages, ' ') }}
+          TIMEOUT: ${{ inputs.run_very_expensive_tests && '60m' || '10m' }}
+          MONITOR_FREE_MEMORY: ${{ runner.debug }}
+        run: |
+          if [[ "$MONITOR_FREE_MEMORY" == "1" ]]; then
+            while true; do
+              free -m
+              sleep 5
+            done &
+          fi
```
rvagg, line 200: this stays? I assume it helped you get to the memory issues? should it be removed now?

rvagg (approved):
I wish it were easier to compare the old test.yml and reusable-test.yml. I tried checking old and new out locally and manually diffing them but the changes are huge so maybe formatting was changed? I ended up side-by-side in the GitHub UI to confirm they're mostly the same. I find these files really dense and hard to parse so keeping diffs minimal where possible helps.

.github/workflows/reusable-test.yml
galargh: It will only come one when a job is rerun with debug logging enabled. I think it could be useful to leave it in but I have nothing against removing it either.

```suggestion
        run: |
```

.github/workflows/very-expensive-test.yml
galargh: I'd say about $1 per very expensive test run.

galargh:
This is the diff of `test.yml` from `master` vs `resuable-test.yml` - https://github.com/filecoin-project/lotus/commit/6cc0c97d46a21a4f7ced55bff679071dd6218681

rvagg:
ACK, thanks, this is good to go I think

.github/workflows/very-expensive-test.yml
```diff
@@ -0,0 +1,43 @@
+name: Very Expensive Test
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      # WARN: This will run very expensive test every time ANY label is added
+      - labeled
```
BigLep, line 10:
```suggestion
      # WARN: This will run very expensive test workflow every time ANY label is added (but we short circuit if there isn't the `need/very-expensive-test` label.
      - labeled
```

Isn't that right?  There's no real major downsides of running additionally and then short-circuiting right (at which point this is a good callout but not necessarily a major WARN to me).

.github/workflows/very-expensive-test.yml
BigLep: Ok, sounds good.  Let go with it for now and we can adjust.  It definitely helps to find out soon after a bad commit if we broke something...

.github/workflows/very-expensive-test.yml
```diff
@@ -0,0 +1,43 @@
...
+    uses: ./.github/workflows/reusable-test.yml
+    with:
+      run_very_expensive_tests: true
+  issue:
+    name: Issue
+    if: failure() && github.event_name == 'schedule'
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create issue
```
BigLep, line 32: And what happens if it keeps failing each day?  Will additional comments be added?  Basically it would be nice to see if this is still continuing to occur or if it was a one-off.

BigLep (approved):
Minor comments, but looks good to me.  I'll let you merge in case you want to monitor anything afterwards.

.github/workflows/very-expensive-test.yml
galargh: The callout was more about the situation in which `need/very-expensive-test` label is present in the PR and then we add more labels. In that situation, whenever a new label is added, the tests are rerun. I don't think that's a blocker but I thought worth a callout.

.github/workflows/very-expensive-test.yml
galargh: As for now, it creates an issue per failure. We can certainly modify it to comment instead if an issue exists already and is open. Would that work?

BigLep: > We can certainly modify it to comment instead if an issue exists already and is open. Would that work?

Yeah, lets do that.  I think that will be preferred by maintainers.

.github/workflows/very-expensive-test.yml
BigLep: Got it - makes sense.  We can resolve.

PR #12944, opened 2025-03-10
chore(build): update filecoin-ffi to a released version

Stebalien:
## Proposed Changes

Update filecoin-ffi to a released version. That way we can pull releases from GitHub instead of building every time. This updates the FFI to v1.32.0-rc2.

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title conforms with [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#pr-title-conventions)
- [x] Update CHANGELOG.md or signal that this change does not need it per [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#changelog-management)
- [x] New features have usage guidelines and / or documentation updates in
  - [x] [Lotus Documentation](https://lotus.filecoin.io)
  - [x] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [x] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #12949, opened 2025-03-12
fix: f3: use F3InitialPowerTableCID in contract provided manifests

Kubuxu:
Resolves: https://github.com/filecoin-project/go-f3/issues/916


rjan90 (commented):
I think the extern/filecoin-ffi changes was unintended here? At least it seems that ffi-commit is older then changes introduced here: https://github.com/filecoin-project/lotus/pull/12944

Kubuxu:
fixed

masih (approved):

rjan90 (approved):

PR #12950, opened 2025-03-13
fix(index): avoid deadlock on close

Stebalien:
## Related Issues

See the test failure in #12942.

## Proposed Changes

Avoid a deadlock when closing the indexer.

We can't hold the write lock while waiting on close because `isClosed` was taking the read lock (and called while closing).

Instead, I just got rid of the locks entirely. Now, we just check the context. Really, I'm not sure why we even have such an "is closed" check. I _think_ what we actually want is to lock to prevent closing until we've finished all atomic operations, but that's a larger change and I don't have enough understanding of this code to easily make that change.

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/index/indexer.go
+++ b/chain/index/indexer.go
@@ -82,9 +82,6 @@ type SqliteIndexer struct {
 
 	started bool
 
-	closeLk sync.RWMutex
-	closed  bool
-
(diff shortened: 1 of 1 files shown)


rvagg (approved):
Yeah, OK, we have the context locally so may as well use it.

PR #12960, opened 2025-03-17
chore!: unify miner, market and verifreg builtin wrappers to v16

rvagg:
https://github.com/filecoin-project/lotus/pull/12932#discussion_r1999225369

This is the cost of unifying the miner types, it has flow-on effects for market and verifreg. It cleans some things up and there's probably a lot of opportunities in here to do more disentanglement (I haven't, this is just a quick rework of all the broken bits).

Is it worth it?


Stebalien:
Yeah, let's punt this for now. But I think the _right_ way to do this is to alias the state types in the abstraction (chain/actors/builtin/...) and avoid importing `go-state-types/builtin/vN` outside of said abstractions.

rvagg:
Agreed, but last time I proposed doing something like that it involved pulling in a lot more types than we reference and I got push-back. But I'm happy to go further than this.

magik6k:
> Is it worth it?

Just flagging that this will also cause pain in boost, curio, cidgravity-gateway, and probably many other projects which use the current aliases. Really that situation is quite unfortunate, and I'd also like to one day fix this, but I wouldn't do this without very strong motivation.

PR #12968, opened 2025-03-20
feat: integrate & test FIP-0098 additions

rvagg:
Ref: https://github.com/filecoin-project/go-state-types/pull/376
Ref: https://github.com/filecoin-project/builtin-actors/pull/1639

I've added some extras in our state types abstractions here, I'd like to add more and reduce (ideally remove) the direct uses of go-state-types versioned actor types we have through the codebase. (See https://github.com/filecoin-project/lotus/pull/12960 for some of how annoying this is). My goal in the itest was to not refer to go-state-types at all while testing all of those additions. We should be able to approach almost all other uses of those types throughout the code in the same way, but we're pretty sloppy.

Also removed some batch balancer stuff.

### DIFF
--- a/chain/actors/builtin/miner/actor.go.template
+++ b/chain/actors/builtin/miner/actor.go.template
@@ -2,32 +2,40 @@ package miner
 
 import (
 	"github.com/ipfs/go-cid"
-	actorstypes "github.com/filecoin-project/go-state-types/actors"
-	"github.com/filecoin-project/go-state-types/big"
-	"github.com/filecoin-project/go-state-types/network"
-	"github.com/filecoin-project/lotus/chain/actors"
 	cbg "github.com/whyrusleeping/cbor-gen"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-state-types/abi"
+	actorstypes "github.com/filecoin-project/go-state-types/actors"
+	"github.com/filecoin-project/go-state-types/big"
+	gstbuiltin "github.com/filecoin-project/go-state-types/builtin"
+	minertypes13 "github.com/filecoin-project/go-state-types/builtin/v13/miner"
+{{- range .versions}}
+	{{- if (ge . 16)}}
+		minertypes{{.}} "github.com/filecoin-project/go-state-types/builtin/v{{.}}/miner"
+		smoothing{{.}} "github.com/filecoin-project/go-state-types/builtin/v{{.}}/util/smoothing"
+	{{- end -}}
+{{end}}
+	minertypes "github.com/filecoin-project/go-state-types/builtin/v9/miner"
 	"github.com/filecoin-project/go-state-types/cbor"
 	"github.com/filecoin-project/go-state-types/dline"
-	"github.com/filecoin-project/go-state-types/proof"
-
-	"github.com/filecoin-project/lotus/chain/actors/adt"
-	"github.com/filecoin-project/lotus/chain/types"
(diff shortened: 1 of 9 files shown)


BigLep:
@rvagg : I assume this should be backported?

@rjan90 : would you ideally want to get this merged before the calibnet upgrade so migration pathways are flexed with this code?  (I assume we'd just encourage a few calibnodes to use `master` with this change rather than cutting another RC at this point.)

chain/actors/builtin/miner/actor.go.template
```diff
@@ -319,3 +332,37 @@ func AllCodes() []cid.Cid {
...
+	v, err := actorstypes.VersionForNetwork(nwVer)
+	if err != nil {
+		return big.Zero(), err
+	}
+	switch v {
+	{{- range .versions}}
+	case actorstypes.Version{{.}}:
+		{{- if (ge . 16)}}
+			return minertypes{{.}}.PledgePenaltyForContinuedFault(
+				smoothing{{.}}.FilterEstimate{
```
Stebalien, line 351: later: we should consider moving `FilterEstimate` to a shared package.

chain/actors/builtin/miner/actor.go.template
```diff
@@ -319,3 +332,37 @@ func AllCodes() []cid.Cid {
...
+					VelocityEstimate: rewardEstimate.VelocityEstimate,
+				},
+				smoothing{{.}}.FilterEstimate{
+					PositionEstimate: networkQaPowerEstimate.PositionEstimate,
+					VelocityEstimate: networkQaPowerEstimate.VelocityEstimate,
+				},
+				qaSectorPower,
+			), nil
+		{{- else}}
+				return big.Zero(), nil
```
Stebalien, line 362: This is a lie.

Stebalien: Honestly, the pledge penalty is the same for all older versions. We could just call into v16 for all versions.

Stebalien (approved):
One suggestion but otherwise this looks correct.

chain/actors/builtin/miner/actor.go.template
rvagg: yep, on my list

chain/actors/builtin/miner/actor.go.template
rvagg: great, cause I didn't want to backport the function to all the previous versions

rvagg:
I also added a `miner.PledgePenaltyForTermination` utility wrapper for the v16 implementation while I was in there, but this one will error for everything prior to v16 since that's the simplified one we introduced.

chain/actors/builtin/miner/actor.go.template
```diff
@@ -334,35 +336,75 @@ func AllCodes() []cid.Cid {
...
+        faultFee abi.TokenAmount,
+) (abi.TokenAmount, error) {
+    v, err := actorstypes.VersionForNetwork(nwVer)
+    if err != nil {
+        return big.Zero(), err
+    }
+
+    if v < actorstypes.Version16 {
+        return big.Zero(), xerrors.Errorf("not supported for network version %d", v)
+    }
```
Stebalien, line 398: Is this necessary? We already have a default below.

rvagg: I guess not, I was imagining the different language would be more helpful than it actually is

removed

Stebalien (approved):

rvagg:
I shouldn't have merged this without a new go-state-types, I'll get that updated now in a new PR

Issue #12767, opened 2024-12-10
Investigate wrong address prefix for calibnet node


rvagg:
Reported on Slack: https://filecoinproject.slack.com/archives/CEGN061C5/p1733812367148489

Glif are currently running `v1.32.0-rc1` and this happens:

```
$ curl -s -X POST 'https://api.calibration.node.glif.io/rpc/v0' -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"Filecoin.ChainGetParentMessages","params":[{"/":"bafy2bzaced3dhjm77nnxe5nsqs66m3yxeohyb2mvsccplz7pintkb4jdo67s6"}],"id":1}' | jq '.result[0].Message.From'
"f1ac6ndwj6nghqbmtbovvnwcqo577p6ox2pt52q2y"
```

I'm running 1.30.0 still and I get the expected output:

```
$ curl -s -X POST 'http://localhost:1235/rpc/v0' -H "Content-Type: application/json" --data '
{"jsonrpc":"2.0","method":"Filecoin.ChainGetParentMessages","params":[{"/": "bafy2bzaced3dhjm77nnxe5nsqs66m3yxeohyb2mvsccplz7pintkb4jdo67s6"}],"id":1}' | jq '.result[0].Messag
e.From'
"t1ac6ndwj6nghqbmtbovvnwcqo577p6ox2pt52q2y"
```

Need to investigate why `address.CurrentNetwork` is not set to `address.Testnet` or why `address.Testnet` isn't resulting in an `address.TestnetPrefix`. Is this a 1.32.0 thing? Or might there have been a compile problem of some kind?

rvagg:
Certainly related, but they look OK to me; the fact that a node will successfully sync calibnet means it was built with the calibnet build tag so none of these paths should even be relevant.

I'll upgrade my node this week and see if it's the same to determine whether this is actually a code problem and go from there.

rvagg:
I just compiled and deployed `1.32.0-rc2+calibnet+git.a9718c841+api1.5.0` and it's still correct on my end:

```
$ curl -s -X POST 'http://localhost:1235/rpc/v0' -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"Filecoin.ChainGetParentMessages","params":[{"/":"bafy2bzacedxj55j56mv2qb7r6zlopxtdq4nh2ast2xpralakro5lixljuuage"}],"id":1}' | jq '.result[0].Message.From'
[...]

PR #12969, opened 2025-03-20
feat: improve `lotus-gateway` build for network selection

### LINKED ISSUE #12967: Makefile target to build calibnet version of lotus-gateway
Resolution to an ongoing issue with glif's Calibnet endpoint where it would return `f` addresses for Calibnet calls. @dumikau figured out the problem: it's isolated to lotus-gateway and it's because it's not built with the `calibnet` tag.

```makefile
lotus-gateway-calibnet: GOFLAGS+=-tags=calibnet
lotus-gateway-calibnet: lotus-gateway  ## Build for calibnet network
```

This fixes the problem, we should probably just add that or something like it.

rjan90:
## Related Issues
Closes #12967 #12767 

## Proposed Changes

- Add a marker file .calibnet-mode that gets created when running `make calibnet`
- Make `lotus-gateway` check for that markerfile
   - If the file exists, it builds with calibnet tags
   - If the file doesn’t exist, it builds for mainnet
- Make sure the marker file gets removed when running `make clean`

## Additional Info
We discussed other approaches as well, but we wanted to keep current state intact and not add `lotus-gateway` to `make calibnet` since it is already overloaded, as having to build extra binaries is not ideal. And additionally we wanted to keep consistency, and prefer to build `lotus-gateway` instead of introducing a special `lotus-gateway-calib`.

## Checklist

- [ ] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 > * [CHANGELOG_1.2x.md](./documentation/changelog/CHANGELOG_1.2x.md) - v1.20.0 to v1.29.2
 
 # UNRELEASED
+- feat: improve `lotus-gateway` build for network selection ([filecoin-project/lotus#12969](https://github.com/filecoin-project/lotus/pull/12969))
 - Allow users to start node even if Index Reconciliation fails (https://github.com/filecoin-project/lotus/pull/12930)
 - Exposed `StateGetNetworkParams` in the Lotus Gateway API ([filecoin-project/lotus#12881](https://github.com/filecoin-project/lotus/pull/12881))
 - **BREAKING**: Removed `SupportedProofTypes` from `StateGetNetworkParams` response as it was unreliable and didn't match FVM's actual supported proofs ([filecoin-project/lotus#12881](https://github.com/filecoin-project/lotus/pull/12881))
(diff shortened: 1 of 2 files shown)


rjan90:
Testing it:

- Running `make calibnet` I can see that there is now a marker file:
```
ls -la | grep .calibnet-mode
-rw-rw-r--  1 filoz filoz         0 Mar 20 07:24 .calibnet-mode
```
- Running `make lotus-gateway` now outputs which tags its building `lotus-gateway` with:
```
make lotus-gateway
rm -f lotus-gateway
Building lotus-gateway with calibnet tags
go: downloading golang.org/x/time v0.11.0 
```
- Checking version:
```
./lotus-gateway -h
NAME:
   lotus-gateway - Public API server for lotus

USAGE:
   lotus-gateway [global options] command [command options]

VERSION:
   1.32.1-dev+calibnet+git.d2be3a825
```

BigLep:
Requested review from @rvagg given he was the opener of the issues being fixed.

rvagg:
I don't love this because of the marker file - which creates persistent and surprising state that's going to cause unintended consequences later (it also needs to be in .gitignore so nobody accidentally commits it). Ideally, the Makefile's state only persists for its execution lifetime and the remaining state comes from source files. Marker files being a last resort if you can't detect some state information about your source.

Anyway, after a few rounds with my mate Claude, we arrived at an alternative approach: https://github.com/filecoin-project/lotus/pull/12972 - what do you think about that one instead?

PR #12972, opened 2025-03-21
feat(build): add calibnet-* etc. targets for per-network builds

rvagg:
Closes: https://github.com/filecoin-project/lotus/issues/12767

Alternative attempt to https://github.com/filecoin-project/lotus/pull/12969 that doesn't involve a senti

`make calibnet` builds everything you need for a full devnet package, you would now do `make calibnet-lotus`, `make calibnet-lotus-gateway` etc. to build individual binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,23 +70,58 @@ CLEAN+=build/.update-modules
 deps: $(BUILD_DEPS)  ## Install build dependencies
 .PHONY: deps
 
-build-devnets: build lotus-seed lotus-shed  ## Build binaries for development networks
+# Network-specific flags
+DEBUG_FLAGS=-tags=debug
+TWOK_FLAGS=-tags=2k
+CALIBNET_FLAGS=-tags=calibnet
+BUTTERFLYNET_FLAGS=-tags=butterflynet
+INTEROPNET_FLAGS=-tags=interopnet
+
+# Network-specific pattern rules
+debug-%:
+	$(MAKE) $* GOFLAGS="$(GOFLAGS) $(DEBUG_FLAGS)"
+2k-%:
+	$(MAKE) $* GOFLAGS="$(GOFLAGS) $(TWOK_FLAGS)"
+calibnet-%:
+	$(MAKE) $* GOFLAGS="$(GOFLAGS) $(CALIBNET_FLAGS)"
+butterflynet-%:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -70,23 +70,58 @@ CLEAN+=build/.update-modules
...
+	@printf "\033[33m'make 2k' builds all devnet binaries. Use 'make 2k-<binary>' targets for individual binaries.\033[0m\n"
+	@printf "Example: make 2k-lotus 2k-lotus-miner\n\n"
+	$(MAKE) 2k-lotus 2k-lotus-miner 2k-lotus-worker 2k-lotus-seed 2k-lotus-shed
+.PHONY: 2k
+
+calibnet:
+	@printf "\033[33m'make calibnet' builds all devnet binaries. Use 'make calibnet-<binary>' targets for individual binaries.\033[0m\n"
+	@printf "Example: make calibnet-lotus calibnet-lotus-miner\n\n"
+	$(MAKE) calibnet-lotus calibnet-lotus-miner calibnet-lotus-worker calibnet-lotus-seed calibnet-lotus-shed
+.PHONY: calibnet
```
rjan90, line 112:
```suggestion
	$(MAKE) calibnet-lotus calibnet-lotus-miner calibnet-lotus-worker calibnet-lotus-seed calibnet-lotus-shed calibnet-lotus-gateway
.PHONY: calibnet
```

rvagg: The reason I didn't add this in here is because it matches the current behaviour, so `make calibnet` still does the same thing. I'm also making an assumption here that the whole "devnets" workflow is used by some upstream projects to build docker containers or other setups so wanted to leave it alone until we know we can just ditch it because it's kind of crufty.

rjan90: I see. So for someone like Glif, that would want to build the lotus-gateway for their Calibnet nodes, they would have to run `calibnet-lotus-gateway` ?

rvagg: yes, @dumikau do you want to weigh in here, as the devops expert and talk us through what the ideal for this might be?

dumikau: I don't really care which command builds the binaries since we use multi-stage build for our images and copy the chosen binaries over to the runtime image anyway. The only preference I have is that for binaries to have the same name (`lotus-gateway` for both mainnet and calibnet) so our automations are uniform.

rjan90: That is great! Binaries will still have the same `lotus-gateway` name

rjan90 (approved):

PR #12973, opened 2025-03-21
chore(deps): bump go-state-types to v0.16.0-rc8

rvagg:
Was supposed to be done in https://github.com/filecoin-project/lotus/pull/12968


rvagg:
Also noticed a flake in the daily_fees_test: https://github.com/filecoin-project/lotus/actions/runs/13984190634/job/39154961267?pr=12973 so I've pushed a change which fixes a problem in the test, although I'm not sure it'll fix that particular flake since the discrepancy is in the wrong direction.

rjan90 (approved):

Issue #11997, opened 2024-05-15
New method to access ProviderSectors HAMT from market actor


LexLuthr:
We need a way to access information in https://github.com/filecoin-project/go-state-types/blob/ad909a3d6ebef5ebc5688b4310e954e27da2203b/builtin/v13/market/market_state.go#L66. This can be very useful to verify F05 deals post NV23 if sealer is using ProveCommitSector3 method to commit all sectors.

rvagg:
@LexLuthr I've moved this to the lotus repo as the API is going to have to be here rather than in go-state-types.

LexLuthr:
> @LexLuthr I've moved this to the lotus repo as the API is going to have to be here rather than in go-state-types.

Thank you. Look forward to have this soon. Along with https://github.com/filecoin-project/lotus/pull/11936, we will have all we need to track market deal reliably.

beck-8:
Will there be any progress here soon?

rvagg:
No, sorry, but it shouldn't be too hard to do. Would someone like to start off by proposing the method signatures and documentation that would slot into `api/api_full.go`? I can probably tackle it once we have agreement from you (as users) on what we're actually implementing.

LexLuthr:
StateMarketSectorList(ctx, minerAddress) (map[sector number][]abi.DealID, err)
StateMarketSector(ctx, minerAddress, sectorNumber) ([]abi.DealID, err)

I do a confusion here though. If a sector only contains f05 deals, should I assume that []abi.DealID would be in same order as present in physical sector or do we lose the order?
What happens when DDO is in the mix for the above question?

rvagg:
Labelling this `good first issue` although it's not easy unless you've got some existing experience with the way Lotus plumbs its API pieces together.

Also related to https://github.com/filecoin-project/lotus/issues/12791 so this might help sort that issue out.

PR #12978, opened 2025-03-25
feat: expose market ProviderSectors access on state-types abstraction

rvagg:
Closes: https://github.com/filecoin-project/lotus/issues/11997

See the itest for how this is used.

~_Depends on https://github.com/filecoin-project/go-state-types/pull/378_~


LexLuthr (approved):
Looks good! Thank you for the example usage. This should be very handy.

ZenGround0 (approved):
Nothing looking controversial here

PR #12985, opened 2025-03-27
chore: make fiximports run in parallel

rvagg:
fiximports has been driving me a little insane lately. For some reason it's started taking up to ~40s each execution and I can't figure out why. I tried switched from running go from a snap to just installing it from a package, I tried clearing my go build cache, I tried removing all the x/tools sources I had. Maybe I made a difference somewhere in there, but I'm still ~20s.

When measuring, the time cost is all in `imports.Process`, so something in there isn't having a happy time.

After a couple of iterations of trying to fix it I've landed on the following approach which gives me the best results:

1. Find all of the files we need to process
2. Read all file contents in (+ do the quick imports newline squash)
3. For each mutation of `imports.LocalPrefix`, process all files in parallel—so we still do `imports.Process` twice but we do the files in parallel and don't need to worry about `imports.LocalPrefix` mutation
4. Write out any files that have been changed

Before:

```
$ time make fiximports 
go run ./scripts/fiximports

real    0m20.004s
user    0m15.952s
sys     0m12.062s
```

After:

```
$ time make fiximports 
go run ./scripts/fiximports

real    0m1.859s
user    0m16.487s
sys     0m16.172s
```

Almost too good to be true? Seems to work just fine though. @masih would you mind taking a good look at this when you have some time please?

### DIFF
--- a/scripts/fiximports/main.go
+++ b/scripts/fiximports/main.go
@@ -3,12 +3,14 @@ package main
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
+	"sync"
+	"sync/atomic"
 
 	"golang.org/x/tools/imports"
 )
@@ -25,7 +27,18 @@ var (
 	consecutiveNewlinesRegex = regexp.MustCompile(`\n\s*\n`)
 )
 
+type fileContent struct {
+	path     string
+	original []byte
+	current  []byte
+	changed  bool
+}
+
 func main() {
+	numWorkers := runtime.NumCPU()
+
+	// Collect all the filenames that we want to process
+	var files []string
 	if err := filepath.Walk(".", func(path string, info fs.FileInfo, err error) error {
 		switch {
 		case err != nil:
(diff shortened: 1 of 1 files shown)


masih:
Seen https://github.com/filecoin-project/lotus/pull/11695#discussion_r1525474298 ?

rvagg:
No, but

> I think the thing to do is to simplify our formatting to two groups of SDK and no SDK

Yes, that was my initial thought as I was working on this because it's that global that stops simple parallelism. Figured I'd try for least-change though and this yields pretty good results.

scripts/fiximports/main.go
```diff
@@ -39,49 +52,152 @@ func main() {
...
+		}
 	}
-	written, err := target.Write(replacement)
-	if err != nil {
-		return err
+	close(changedFiles)
+
+	for w := 0; w < numWorkers; w++ {
+		wg.Add(1)
+		go func() {
```
masih, line 181: You can simplify this code quite a bit using error group?

scripts/fiximports/main.go
```diff
@@ -39,49 +52,152 @@ func main() {
...
+	for i := range fileContents {
+		if fileContents[i] != nil { // shouldn't be nil, but just in case
+			filesChan <- i
 		}
 	}
-	return nil
+	close(filesChan)
+
+	for w := 0; w < numWorkers; w++ {
+		wg.Add(1)
```
masih, line 138: Ditto re using error group.

scripts/fiximports/main.go
```diff
@@ -39,49 +52,152 @@ func main() {
...
-		fmt.Printf("Error fixing go imports: %v\n", err)
+		_, _ = fmt.Fprintf(os.Stderr, "Error walking directory: %v\n", err)
 		os.Exit(1)
 	}
+
+	// Read all file contents in parallel
+	fileContents := readFilesParallel(files, numWorkers)
+
+	// Because we have multiple ways of separating imports, we have to imports.Process for each one
+	// but imports.LocalPrefix is a global, so we have to set it for each group and process files
```
masih, line 66: This was the reason I did not make the script execute in parallel. Per group parallelism makes sense 👍

masih (approved):
No blockers. 

I think the logic can be simplified quite a bit using error groups.

scripts/fiximports/main.go
rvagg: My first naive attempt when I dove in here tried for complete parallelism but I was quickly educated about why you didn't do it in the first place!

rvagg:
Yeah, errgroups made it much better, thanks. PTAL @masih

scripts/fiximports/main.go
```diff
@@ -39,49 +52,112 @@ func main() {
...
+				current:  collapsed,
+				changed:  !bytes.Equal(content, collapsed),
+			}
+			return nil
+		})
 	}
-	written, err := target.Write(replacement)
-	if err != nil {
-		return err
+
```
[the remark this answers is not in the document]
rvagg: Reconsider in light of this projecting having a Go 1.23 minimum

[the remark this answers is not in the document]
rvagg: hah, not interactive, disappointing

masih (approved):

PR #12999, opened 2025-03-31
fix(eth): return nil for eth transactions not found

rvagg:
Originally reported by @dumikau. I'm pretty sure the "not found" path for Eth calls should be to return nothing, but the usual path is: (1) search in ChainIndexer for the Eth hash and use that if it has it, then (2) to a naive conversion from the hash to a CID and use that (`txHash.ToCid()`). That second path is usually not what the user intends, and if we don't have it in the ChainIndexer then it's probably not landed on chain yet. But in the current case for `eth_getTransactionReceipt` we're doing a `StateSearchMsg` on the CID we get from a `txHash.ToCid()` and that will return an `ipld: count not find` when it gets down in to `GetCMessage` message trying to fetch it from the blockstore. Then we end up with this confusion message with a CID that bears no relationship to the hash the user asked for (from their perspective) and it's just confusing. I think the right thing to do is return `nil` which I _believe_ is what Eth tooling dows.

```
Error could not coalesce error (error={ "code": 1, "message": "failed to lookup Eth Txn 0x2ea4388c7f8c43433500c9559bb219db2dc52f76a2f12e70c57bbafec2f8acbf as bafy2bzaceaxkioemp6gegqzvadevlg5sdhns3rjpo2rpcltqyv53v7wc7cwl6: failed to load message: ipld: could not find bafy2bzaceaxkioemp6gegqzvadevlg5sdhns3rjpo2rpcltqyv53v7wc7cwl6" }, payload={ "id": 298, "jsonrpc": "2.0", "method": "eth_getTransactionReceipt", "params": [ "0x2ea4388c7f8c43433500c9559bb219db2dc52f76a2f12e70c57bbafec2f8acbf" ] }, code=UNKNOWN_ERROR, version=6.13.4)
```


BigLep:
@masih : can you please take a look at this so we can get it merged and as a slight way to learn more about the ETH APIs?

masih (approved):

Issue #12990, opened 2025-03-27
Design `v2` "non-Eth" (chain, F3, state) RPC APIs that are F3-aware


BigLep:
## Done Criteria
There is a public document that lists proposed:
1. APIs being added
2. APIs being removed
3. APIs being modified
as a result of F3.  We should be able to share this with a partner who will use them and give feedback on them

The groups in scope are:
* [Chain](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#Chain)
* [F3](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#F3)
* [State](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#State)

[Eth](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#Eth) is not in scope because we already know what to do from an implementation regard (and that is happening in https://github.com/filecoin-project/lotus/issues/12988)

## Why Important
1. Ensure we implement the right thing
2. Enable integraters to plan for how they'll adapt with the advent of F3 

## Notes
1. Exchanges (e.g., Coinbase through Zondax) are a target audience of this work.
2. There is existing work in this area to draw from: https://www.notion.so/filecoindev/WIP-API-updates-w-F3-106dc41950c1801789bef5eca507efa5 .  That said it isn't currently crisp on our current proposal.  We don't want readers to have to pick through to try and guess what our proposal is.
3. A key challenge to solve is a clean way for the user to specify the tipset they want to use.  We have to introduce a new API paradigm to users but we should be careful not to break too hard so everyone has to rewrite stuff.  We could lean on the Eth paradigm to make bridging easier which is why proposals to date have been a mash-up of Filecoin and Eth tipset selectors.
4. This design document doesn't block implementation work.  There is code to write that will be independent of the exact APIs we settle on. For example, the tipset selection needs to be plumbing through everywhere we execute the backends of the APIs so we don’t have to have two versions of everything. v2 and v1 (and therefore v0) should share code paths, but the “which tipset” argument is now different. Refactoring to promote reuse is going to be tricky given the current internal architecture and the way we use the DI (uber/ux) tooling to assemble Lotus.
[...]

PR #13003, opened 2025-04-02
feat: implement ChainGetTipSet in Lotus v2 APIs

masih:
Introduce the first API for Lotus v2, focusing on `ChainGetTipSet` within the `Chain` group. Define `TipSetSelector` for advanced tipset retrieval options, and create a compact JSON-RPC call format.

Gracefully accommodate both EC and F3 finalized tipsets based on node configuration, where:
* EC finalized tipset is returned when F3 is turned off, has no finalized tipset or F3 isn't ready.
* F3 finalized is returned otherwise.

Support three categories of selectors under `TipSetSelector`:
* By tag: either "latest" or "finalized."
* By height: epoch, plus optional fallback to previous non-null tipset.
* By tipset key.

The selection falls back to tag "latest" if the user specifies no selection criterion.

The JSON-RPC format is designed to use JSON Object as the parameters passed to the RPC call to remain compact, and extensible. Examples include:

* Latest Tipset: ```json {"jsonrpc":"2.0","method":"Filecoin.ChainGetTipSet","id":1} ```
* Finalized Tipset: ```json {"jsonrpc":"2.0","method":"Filecoin.ChainGetTipSet","params":{"tag":"finalized"},"id":1} ```

* Tipset at specific height: ```json {"jsonrpc":"2.0","method":"Filecoin.ChainGetTipSet","params":{"height":{"at":1413,"previous":true}},"id":1} ```

The open API spec and Markdown generation is excluded from this work to reduce the PR verbosity and will be included in a follow-up PR once the basic API primitives are reviewed.

Note that all v2 APIs are experimental and may change with no notice.

Relates to: https://github.com/filecoin-project/lotus/pull/12719


chain/types/tipset_selector.go
```diff
@@ -0,0 +1,128 @@
...
+
+// TipSetHeight is a criterion that selects a tipset At given height.
+//
+// In a case where the tipset at given height is null, and Previous is true,
+// it'll select the previous non-null tipset instead. Otherwise, it returns the
+// null tipset at the given height.
+type TipSetHeight struct {
+	At       abi.ChainEpoch `json:"at,omitempty"`
+	Previous bool           `json:"previous,omitempty"`
+}
```
rvagg, line 128: The only thing missing from this is what I was calling "anchor", when you specify a height, with the current APIs, you can also specify the chain to look at. It might become less of a problem post-F3 but it doesn't fully go away. 

Internally, `ChainStore#GetTipsetByHeight(ctx context.Context, h abi.ChainEpoch, ts *types.TipSet, prev bool)` does all of the work here, currently using `TipSetHeight`, that `ts` parameter will have to be `nil` which defaults to the current heaviest tipset. But you ought to be able to anchor it to a known tipset that you believe is the chain you care about.

In an F3 world, perhaps you also want to be able to specify `"finalized"` here too? I'm not sure if that buys us much though since heaviest is going to build on top of `"finalized"`, but we'd get an error if the height is greater than F3-finalized at least?

rvagg:
I can live with this I think. I don't love the verbosity of the nesting but at least you've come up with short descriptors, `"tag"` and `"at"`, I like that 👍. Good stuff.

rvagg:
Unrelated flaky itest recorded @ https://github.com/filecoin-project/lotus/issues/13005

chain/types/tipset_selector.go
masih: Right, adding. Thanks.

It should be trivial to make the anchor extensible. I'll write it with that in mind and perhaps add TipSetTag as an anchor to see what it would look like.

masih: Added. This now looks like:
* `{"jsonrpc":"2.0","method":"Filecoin.ChainGetTipSet","params":{"height":{"at":111,"anchor":{"tag":"finalized"}}},"id":1}`
* `{"jsonrpc":"2.0","method":"Filecoin.ChainGetTipSet","params":{"height":{"at":111,"anchor":{"key":[...]}}},"id":1}`

Kubuxu:
Initial quick skim:
- Height is a bit of a miss-nomer, we don't track height but an epoch in filecoin as height originally refers to the height of the hash tree.
- Why the generics complexity logic in the type? Why not use the TipSetCriterion directly with additional validation checks in the API?

masih:
Discussed at stand up:

> Height is a bit of a miss-nomer, we don't track height but an epoch in filecoin as height originally refers to the height of the hash tree.

It is. From internal notion doc discussions I got the sense that we want to switch to "height" as the new terminology. Happy to revert back to Epoch, or move on with Epoch. Either works.

@Kubuxu kindly pointed out that the Go SDK to interact with selectors is slightly cumbersome. I am open to suggestions on this one and I think we can iterate on it as non-breaking API changes later. For now I am keen to pin down the primitives for the Lotus V2 API first and then iterate.


> Why the generics complexity logic in the type? Why not use the TipSetCriterion directly with additional validation checks in the API?

The generics is used solely as Type Constraint to `NewTipSetSelector`. We can't use `TipSetCriterion` as argument to `ChainGetTipSet` directly, because it is not a type alias of `jsonrpc.RawParams` without losing the ability to accept requests with no params specified. This is a limitation inherited by the implementation of jsonrpc library used by Lotus.

During the discussion, the choice of default param as "Heaviest" if nothing is specified came up: 
* @Kubuxu mentioned that maybe we should explicitly ask for what the user wants and have no default if no selector is specified. 
* I have no strong opinions on the defaults. I do see that explicit is better ( which is _the_ rationale behind changes here compared to prior iterations). However, we would lose the "short-and-sweet" aspect of API calls by users in other languages, a non-functional requirement for the v2 API as I understand it. 


@Kubuxu please add anything that I might have missed.

@rvagg thoughts?

Kubuxu:
Discussed it a bit with Masih.
1. The weird typing is required if we want to have `Filecoin.ChainGetTipSet` with no params to work across JSOR-RPC
2. IDK if I like that you will be required to type,
```
ts, err := api.ChainGetTipSet(ctx, types.NewTipSetSelector(types.TipSetHeight{
	At:       123,
	Previous: true,
	Anchor: &types.TipSetAnchor{
		Tag: &types.TipSetTags.Finalized,
		Key: &types.TipSetKey{},
	},
}))```
is something what we want

File: itests/api_v2_test.go [hunk 2]
```diff
@@ -0,0 +1,291 @@
...
+		wantErr    string
+	}{
+		{
+			name:       "no selector",
+			request:    `{"jsonrpc":"2.0","method":"Filecoin.ChainGetTipSet","id":1}`,
+			wantTipSet: heaviest,
+		},
+		{
+			name:       "latest tag",
+			request:    `{"jsonrpc":"2.0","method":"Filecoin.ChainGetTipSet","params":{"tag":"latest"},"id":1}`,
```
BigLep, line 102:
```suggestion
			request:    ChainGetTipSetRequestBody(`{"tag":"latest"}`),
```

nit: factor out the boiler plate?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 
 # UNRELEASED
 
+- feat: implement ChainGetTipSet in Lotus v2 APIs
```
BigLep, line 12:
```suggestion
- feat: add experimental v2 APIs that are "F3 aware".  (TODO: expand this section significantly to cover where someone learns about the new APIs, how they enable them, and what expectations they should have around them - i.e., they may change)
```

I know the user documentation isn't the focus.  I'm thinking to leave a TODO so it's clear for whoever does the release that this feature needs to be expanded in the changelog (in case it hadn't gotten expanded earlier).

BigLep (commented):
No blocking comments from me.  Just following along.

I wanted to confirm we are good with how calling these APIs looks like from other languages (particularly JS).

BigLep:
> I'm 90% that the latest on empty is not what we want. Split across "empty invalid" and empty returning finalized.

(Not an expert opinion here), but I agree.  I think it's important to help builders "avoid running into trouble unknowinglyg".  We can do this by either (as Kuba says):
1. Be explicit - no default
2. Use finalized

itests/api_v2_test.go
masih: I explicitly want the test to show the raw JSON-rpc request on the wire. See comment above the test table.

masih:
Going ahead with no default.

> IDK if I like that you will be required to type

Let's see what things look like after explicit param is needed and we can iterate. Probably faster to sit down and go through it synchronously to distill actionables on this one.

masih:
@Kubuxu  Do you want to get rid of the default anchor for getting tipset by height too or leave it as optional?

masih:
@Kubuxu 
* A selector is now strictly required; no defaults.
* We can't get rid of the type alias to `jsonrpc.RawParams` without making request body more ugly:
  *  The jsonrpc library lotus uses would implicitly switch to using JSON array as parms if the type of an argument is not `jsonrpc.RawParams`. 
  * This means request params on the wire would have to have an array wrapping, e.g.: `{"jsonrpc":"2.0","method":"Filecoin.ChainGetTipSet","params":[{"tag":"finalized"}],"id":1}` instead of `{"jsonrpc":"2.0","method":"Filecoin.ChainGetTipSet","params":{"tag":"finalized"},"id":1}`.
  * The choices are to: use raw params type alias, live with the redundant wrapping, or fix go-jsonrpc to do things differently.
    * I don't think it makes sense to use the mix-and-match style params experimented with in https://github.com/filecoin-project/lotus/pull/12719 because for me that conflicts with "Explicit APIs" we desire. I vote is to keep the parameters keyed if it needs to be extensible.
  * My pick is to use the the params type alias as it is the most straightforward way to get what we want while keeping the API extensible if we want to add anything else.
* I have not touched the Go API for constructing selectors. This one is mostly bike shedding and best done synchronously.

Kubuxu:
the `params` being array is fully expected. It is a single positional argument, which is an object
I haven't seen the named parameters used anywhere in the JSON-RPC spec.
They are allowed by not that popular.

Kubuxu:
Additionally, most of our tooling assumes positional parameters. IMO not worth the hassle for something that can be trivial abstracted out in the client if someone wants it.

masih:
> Additionally, most of our tooling assumes positional parameters. 

What tooling exactly?

> can be trivial abstracted out in the client if someone wants it.

This is true for everything. I thought we wanted to make the wire format concise. Either works I don't have strong feelings about this.

Kubuxu:
> What tooling exactly?

our JSON RPC lib, lotus-shed rpc, other languages JSON-RPC bindings

> This is true for everything. I thought we wanted to make the wire format concise. Either works I don't have strong feelings about this.

IMO the aim is more about making the API easy and logical to use. I agree that named parameters in this case could be better but I don't think the current tradeoff is correct. If, in the future, we have proper support for named parameters in our json-rpc lib, we could add it here as well.

masih:
> our JSON RPC lib, lotus-shed rpc, other languages JSON-RPC bindings

All those libraries use auto-generated code. I don't see any of them implicitly assuming that params is an array. 



> I agree that named parameters in this case could be better but I don't think the current tradeoff is correct.

Sorry, are you suggesting that object wrapped in array is not the correct wire format trade-off?

Kubuxu:
From JS/TS it might be better (I looked at one JS lib).
`lotus-shed rpc` is manual, some probably the same

> Sorry, are you suggesting that object wrapped in array is not the correct wire format trade-off?

My argument is that named parameters are not right tradeoff right now, given our tooling situation.

masih:
Let's discuss further in stand up.

masih:
@Kubuxu  this PR is ready for another look whenever you get a chance. It makes all the changes we discussed at standup. Please let me know if there is anything that I might have missed.

chain/types/tipset_selector.go
```diff
@@ -0,0 +1,162 @@
...
+// In a case where the tipset at given height is null, and Previous is true,
+// it'll select the previous non-null tipset instead. Otherwise, it returns the
+// null tipset at the given height.
+//
+// The Anchor may optionally be specified as TipSetTag, or TipSetKey. If
+// specified, the selected tipset is guaranteed to be a child of the tipset
+// specified by the anchor at the given height. Otherwise, the "finalized" TipSetTag
+// is used as the Anchor.
+type TipSetHeight struct {
+	At       abi.ChainEpoch `json:"at,omitempty"`
```
rvagg, line 110: So if you omit `"at"` then you'll default to `0` and `Validate()` is going to be fine with that. This is not terrible, but I wonder whether we should be distinguishing an omitted value from `0` here because `0` is almost certainly not going to be what you want out of the API.

rvagg:
Looks really good, I'll have to try and restart my Eth API work ontop of this for /v2 to remind myself about all the module hackery needed to make this work. But I like how the `TipSetSelector` has worked out here and my initial take is that the the `ChainGetTipSet` v2 wrapper is nice and clean, although I wonder how we're going to keep it simple when we need to deal with selecting a tipset for all the other methods (thinking especially the `State*` ones) that currently take an explicit tipset or `EmptyTSK`. Perhaps it's the v1 APIs that have to get wrapped to suit the internal v2 methods where a `TipSetSelector` is what we accept on the internal functions.

masih:
> how we're going to keep it simple when we need to deal with selecting a tipset for all the other methods (thinking especially the `State*` ones

Thanks for pointing that out. I will pick up a `State` API next to flesh out any complexities before making `Chain*` more featureful.

> Perhaps it's the v1 APIs that have to get wrapped to suit the internal v2 methods

This has been my thinking so far: to eventually turn v1 APIs a wrapper to v2.

documentation/en/api-v2-unstable-methods.md
```diff
@@ -0,0 +1,56 @@
...
+The "Finalized" tag returns the tipset that is considered finalized based on
+the consensus protocol of the current node, either Filecoin EC Finality or
+Filecoin Fast Finality (F3). The finalized tipset selection gracefully falls
+back to EC finality in cases where F3 isn't ready or not running.
+
+In a case where no selector is provided, an error is returned. The selector
+must be explicitly specified.
+
+For more details, refer to the types.TipSetSelector and
+types.NewTipSetSelector.
```
BigLep, line 24: Is someone looking at this in the future at https://docs.filecoin.io/reference/json-rpc/chain or https://github.com/filecoin-project/lotus/blob/master/api/README.md going to know how to navigate to `chain/types/tipset_selector.go` to find this info?

For example, I assume if in the 2025-04-09 meeting with Zondax if we just show this Markdown that they don't have enough context to know about the flexibility of TipSetSelector.

documentation/en/api-v2-unstable-methods.md
```diff
@@ -0,0 +1,56 @@
...
+Perms: read
+
+Inputs:
+```json
+[
+  {
+    "tag": "finalized"
+  }
+]
+```
```
BigLep, line 46: Can/should we show other example inputs (e.g., "latest", "height", "key")?

BigLep (commented):
Sweet to see the doc generation happening.

I left a couple of comments when wearing the lens of "learn about new APIs from the generated docs".  (I'm not up on all the limitations of our custom doc generation system.)

documentation/en/api-v2-unstable-methods.md
masih: The generation code only supports a single example. Changing that goes beyond the scope of this PR. I am happy to take a look in a separate PR.

documentation/en/api-v2-unstable-methods.md
rvagg: heh, yeah that README probably should be nuked, I didn't even realise it existed

masih: https://github.com/filecoin-project/lotus/pull/13017

masih:
@Kubuxu @rvagg can I get a review on this PR so that we can distill what's missing from this PR to land please?

documentation/en/api-v2-unstable-methods.md
BigLep: Agreed that is out of scope from this PR and is a bigger undertaking not to be taken in now.  (Thanks for helping me learn about the limitations of our current docsgen system.)

documentation/en/api-v2-unstable-methods.md
BigLep: We spoke about this during 2025-04-08 standup.  It's really unfortunate that our docsgen process is not better for consumers, particularly consumers of other languages besides Go.

For this particular issue at hand of "how does someone learn about the rich `TipSetSelector` type"?
* They won't be able to get that from our generated `documentation/en/api-v2-unstable-methods.md`.
* A Go developer can learn about this type if they look at the godocs
* Maybe the thing to do is to create a side document (Notion, hackmd) where we can state the API contract and richly format it while we're trying to get feedback.  (It's a later bigger item to get Lotus APIs and docs to be contract driven.)

documentation/en/api-v2-unstable-methods.md
```diff
@@ -0,0 +1,56 @@
+# Groups
+* [Chain](#Chain)
+  * [ChainGetTipSet](#ChainGetTipSet)
+## Chain
```
BigLep, line 4: I didn't see a quick existing way to add a comment to something like `https://www.notion.so/filecoindev/Lotus-F3-aware-APIs-1cfdc41950c180ae97fef580e79427d5` in `docsgen.go` before or after the table of contents.  If it is easy, maybe we do so to tie togther docs work happening outside of the repo.  No worries if not.  Feel free to skip/resolve this comment.

documentation/en/api-v2-unstable-methods.md
BigLep: If it helps, I created https://www.notion.so/filecoindev/Lotus-F3-aware-APIs-1cfdc41950c180ae97fef580e79427d5 for housing any API docs we want to have outside the repo while we're in "actively getting user feedback mode".

Anyways, I think this conversation can be resolved as there's nothing more to be actioned here (but I left it open in case we want to make clear that this problem still exists and will need large followup work at some later point to correct.()

masih: I have created a sub-page in notion under Lotus F3 Aware APIs to document the v2 API design/implementation so far: https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658

It is not finished yet but should help overcome the shortcomings of the generated docs in the meantime.

documentation/en/api-v2-unstable-methods.md
masih: Added links to notion doc in generated markdown: https://github.com/filecoin-project/lotus/pull/13003/commits/338572287f3c73d4d3112e37f7e528d137cd9650

Kubuxu (approved):
SGWM

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 # UNRELEASED
 
 - fix(eth): always return nil for eth transactions not found ([filecoin-project/lotus#12999](https://github.com/filecoin-project/lotus/pull/12999))
+- feat: add experimental v2 APIs that are "F3 aware."  (TODO: expand this section significantly to cover where someone learns about the new APIs, how they enable them, and what expectations they should have around them—i.e., they may change)
```
rvagg, line 13: also TODO, when we revisit this, we need the link to this PR; I'll try and remember to add it when I follow-up with Eth stuff if it's not done already by then

rvagg (commented):
sorry I was slow on the draw here, but a post-merge LGTM from me

CHANGELOG.md
masih: Help Me ~Obi-Wan Kenobi~ Generative AI, You’re My Only Hope.

PR #13006, opened 2025-04-04
feat!: actors bundle v16.0.1 & special handling for calibnet

rvagg:
Not tested yet.

* Includes [builtin-actors v16.0.1](https://github.com/filecoin-project/builtin-actors/releases/tag/v16.0.1) bundles
* Includes a repack of the v16.0.1 tarball to include some of the v16.0.0 bundles:

```
$ tar tf build/actors/v16.tar.zst 
builtin-actors-butterflynet.car
builtin-actors-butterflynet-v16.0.0.car
builtin-actors-calibrationnet.car
builtin-actors-calibrationnet-v16.0.0.car
builtin-actors-caterpillarnet.car
builtin-actors-devnet.car
builtin-actors-devnet-v16.0.0.car
builtin-actors-mainnet.car
builtin-actors-testing.car
builtin-actors-testing-fake-proofs.car
```

* Adds a `UpgradeTockFixHeight`, set to `-1` in most places, particularly mainnet. It's only intended for calibnet, but can be run on butterflynet and devnet too since the v16.0.0 bundles are there.
* Includes a `upgradeActorsV16Fix` simple-migration which just does the actors code switch where `UpgradeTockFixHeight` is needed.
* Adds a special case to the vanilla v16 upgrade such that if `UpgradeTockFixHeight` is set, we do a v16.0.0 upgrade rather than the v16.0.1 one.

So, if this works as intended, it should be able to roll through the existing calibnet (and current butterflynet if that's still running) history with the right actors and switch when it's supposed to.

Unfortunately it's a little more complicated than previous ones because we're switching out the entire actors bundle. Perhaps we could have got away with just doing the evm actor, but all the CIDs changed because we updated a dependency so for completeness I'd prefer to go the whole hog.


gen/bundle/bundle.go
```diff
@@ -89,14 +96,3 @@ func main() {
 		panic(err)
 	}
 }
-
-func getOldGitTagFromEmbeddedMetadata(m *build.BuiltinActorsMetadata) string {
-	for _, v := range build.EmbeddedBuiltinActorsMetadata {
-		// if we agree on the manifestCid for the previously embedded metadata, use the previously set tag
-		if m.Version == v.Version && m.Network == v.Network && m.ManifestCid == v.ManifestCid {
-			return m.BundleGitTag
```
rvagg, line 97: This was a bug, should have been `v.BundleGitTag` which is why the `BundleGitTag`s have been progressively disappearing from builtin_actors_gen.go each time we upgrade. I decided to just roll it in to the above code while I was fixing it.

rjan90:
> (and current butterflynet if that's still running)

The Butterfly network is still running (and is on nv26 as well). So this could be a good test to deploy to reduce risk for the Calibnet upgrade.

```
lotus info
Network: butterflynet
StartTime: 238h17m25s (started at 2025-03-25 13:07:01 +0000 UTC)
Chain: [sync ok] [basefee 100 aFIL] [epoch 46140]
Peers to: [publish messages 2] [publish blocks 2]

lotus state actor-cids
Network Version: 26
Actor Version: 16
Manifest CID: bafy2bzacedn2h6huw7v2elmmdmpv4phdv4wjwgct7kcrrtdgz7jkjdm6uwa6k
```

Kubuxu (approved):
from reading SGWM but I'm not deep enough in migration code to be 100% confident

build/buildconstants/params_calibnet.go
```diff
@@ -119,6 +121,9 @@ var UpgradeTeepInitialFilReserved = wholeFIL(1_200_000_000) // FIP-0100: 300M ->
 // the new daily fee to pre-Teep sectors being extended. This is 90 days on mainnet.
 var UpgradeTockHeight abi.ChainEpoch = UpgradeTeepHeight + builtin.EpochsInDay*7
 
+// 2025-04-07T23:00:00Z
+const UpgradeTockFixHeight abi.ChainEpoch = 2558014
```
rjan90, line 125: Confirming that the epoch here correct, and that we have [confirmation from the Forest team](https://filecoinproject.slack.com/archives/C05P37R9KQD/p1743750721560579?thread_ts=1743750513.797599&cid=C05P37R9KQD) about the epoch as well.

rvagg:
Pushed a fix after writing a `TestMigrationTeepTockFix` integration test that walks through 24->25->26->TockFix and tests state at each point. There were a few errors in my migration and the way I was structuring things. I also included the `"testing"` and `"testing-fake-proofs"` v16.0.0 bundles in the v16 tarball because that's what we use in itests (TIL).

itests/migration_test.go
```diff
@@ -1150,3 +1151,102 @@ func preFip0081StateMinerInitialPledgeForSector(ctx context.Context, t *testing.
...
+		epoch       abi.ChainEpoch
+		av          actorstypes.Version
+		getManifest func(actorstypes.Version) cid.Cid
+	}{
+		{0, actorstypes.Version15, getManifestForVersion},
+		{nv25Epoch, actorstypes.Version16, func(actorstypes.Version) cid.Cid { return buggyMetadata.ManifestCid }},
+		{nv26Epoch, actorstypes.Version16, func(actorstypes.Version) cid.Cid { return buggyMetadata.ManifestCid }},
+		{teepFixEpoch, actorstypes.Version16, getManifestForVersion},
+	} {
+		t.Run(fmt.Sprintf("upgrade-%d-at-%d", upgIdx, upgrade.epoch), func(t *testing.T) {
```
rvagg, line 1224: @ZenGround0 if your contract deploy code is straightforward enough it could be inserted here with a `true`/`false` in the test case struct to say whether it should pass or fail (which would be `true`, `false`, `false`, `true`). The most recent simple contract success/fail assertion like that we did is in fevm_test.go in `TestTstore` that tests before and after the nv25 upgrade that `TSTORE` works.

rjan90: Will merge now. Following [up here](https://filecoinproject.slack.com/archives/C08LNJV0J1Z/p1743776714411149?thread_ts=1743772884.385399&cid=C08LNJV0J1Z), that these test cases can land once the US team members get online

PR #13026, opened 2025-04-10
feat(eth)!: add Eth APIs to /v2 + minor improvements & fixes

rvagg:
~Very draft for now, mainly focused on keeping /v1 working without molesting it too much. The beginnings of /v2 are in here already, it's just not wired up to the main struct yet .. or tested.~


rvagg:
Got it all wired up now. I've tried to improve the state of some of the things in here but it's quite tricky.

The most interesting part of this as far as v2 is concerned is in the `TipSetResolver` in `node/impl/eth/tipsetresolver.go` which behaves differently if it's given an `F3Backend` or not. Then in the DI system we construct a `V1` and `V2` variant of this, and all our Eth modules have both `V1` and `V2` forms which take the right kind of `TipSetResolver`.

Misc things I've done in here that apply to v1 as well that I'll need to note in the changelog:

* Moved `api.EthTxReceipt` to `ethtypes.EthTxReceipt` where all its friends are. I'm not sure why it was there, all alone, I'm going to assume a mistake. I've left an alias and a deprecation notice.
* I've changed the way that `EthGetTransactionByHashLimited`, `EthGetTransactionReceiptLimited` and `EthGetBlockReceiptsLimited` are handled on the gateway. Like was already done for `EthSendRawTransactionUntrusted` I've made it so they're not usable on the gateway. They weren't really doing any harm on the gateway (unlike `EthSendRawTransactionUntrusted` but they just don't belong as first-class citizens, so I could remove them from `api.Gateway` although internally they still exist thanks to DI but they "not supported" error.
* `EthGetBlockTransactionCountByNumber` now takes a `blkNum string` instead of a `blkNum ethtypes.EthUint64`. It's an odd-man-out in the APIs by taking strictly a Uint64 and not allowing labels (`"finalized"` etc.). I believe this was an oversight originally, `eth_getBlockTransactionCountByNumber` is supposed to take a flexible input type.
* _I will also give similar treatment to `EthGetLogs` because the `"fromBlock"` and `"toBlock"` parameters are supposed to take more labels than we allow, but it won't be a breaking change._

Still need to add more test coverage and I also need to exercise the behaviour of F3 for `"safe"` and `"latest"` for both v1 and v2.

rvagg:
Making good progress here but the tests are getting tedious. Half of the impacted endpoints are being tested so far.

rvagg:
Calling this _Ready for Review_ although I still need to write the CHANGELOG entry (so I'm leaving that failing). I've expanded support for tags in the /v1 API in the process of doing this so I need to make clear what's changing, not just for /v2 additions. But the code and tests I think are good to go.

api/api_full.go
```diff
@@ -1452,20 +1452,4 @@ type HotGCOpts struct {
...
-	Status            ethtypes.EthUint64   `json:"status"`
-	ContractAddress   *ethtypes.EthAddress `json:"contractAddress"`
-	CumulativeGasUsed ethtypes.EthUint64   `json:"cumulativeGasUsed"`
-	GasUsed           ethtypes.EthUint64   `json:"gasUsed"`
-	EffectiveGasPrice ethtypes.EthBigInt   `json:"effectiveGasPrice"`
-	LogsBloom         ethtypes.EthBytes    `json:"logsBloom"`
-	Logs              []ethtypes.EthLog    `json:"logs"`
-	Type              ethtypes.EthUint64   `json:"type"`
-}
+type EthTxReceipt = ethtypes.EthTxReceipt // Deprecated: use ethtypes.EthTxReceipt instead
```
rvagg, line 1455: unsure why this was even in here, it should be with its friends

api/api_gateway.go
```diff
@@ -35,15 +35,11 @@ import (
 //  * Generate openrpc blobs
 
 type Gateway interface {
-	MpoolPending(context.Context, types.TipSetKey) ([]*types.SignedMessage, error)
```
rvagg, line 38: changes in here are just reorgs to get them into their groups and close to their neighbours, I've not been thorough, just cleaning up the obvious bits out of place

api/api_full.go
```diff
@@ -837,23 +837,23 @@ type FullNode interface {
 	// EthBlockNumber returns the height of the latest (heaviest) TipSet
 	EthBlockNumber(ctx context.Context) (ethtypes.EthUint64, error) //perm:read
 	// EthGetBlockTransactionCountByNumber returns the number of messages in the TipSet
-	EthGetBlockTransactionCountByNumber(ctx context.Context, blkNum ethtypes.EthUint64) (ethtypes.EthUint64, error) //perm:read
+	EthGetBlockTransactionCountByNumber(ctx context.Context, blkNum string) (ethtypes.EthUint64, error) //perm:read
```
rvagg, line 840: this is the largest "breaking" change, but we don't expect most people to use this API in Go and the JSON unmarshal expects strings in both cases anyway

build/buildconstants/params_shared_vals.go
```diff
@@ -103,3 +103,8 @@ const InitialBaseFee int64 = 100e6
 const MinimumBaseFee int64 = 100
 const PackingEfficiencyNum int64 = 4
 const PackingEfficiencyDenom int64 = 5
+
```
rvagg, line 106: change here should dedupe with https://github.com/filecoin-project/lotus/pull/13034

node/impl/eth/api.go
```diff
@@ -25,6 +28,143 @@ var (
 
 var log = logging.Logger("node/eth")
 
+// EthFilecoin -------------------------------------------------------------------------------------
```
rvagg, line 31: everything in here except `TipSetResolver` is just moving these interfaces out of their respective impl files to here, not a strictly necessary change, just came about during my circular refactoring journey trying to improve things; unclear whether I actually did improve things though!

node/impl/eth/events.go
```diff
@@ -609,6 +578,10 @@ func ethTxHashFromMessageCid(ctx context.Context, c cid.Cid, cs ChainStore) (eth
 // * "block" instead of "height"
 // * strings that can have "latest" and "earliest" and nil
 // * hex strings for actual heights
+//
+// Note: Ethereum supports more than "latest" and "earliest" (e.g. "finalized" and "safe"), but
+// currently we do not. The use-case isn't as clear for this but it's possible to add in the future
+// if needed.
```
rvagg, line 584: I decided not to bother for now, maybe if we're ever asked, but it's worth noting the discrepancy

node/impl/eth/tipsetresolver.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	return ts, nil
+}
+
+func (tsr *tipSetResolver) GetTipsetByBlockNumberOrHash(ctx context.Context, blkParam ethtypes.EthBlockNumberOrHash) (*types.TipSet, error) {
+	head := tsr.cs.GetHeaviestTipSet()
+
+	predefined := blkParam.PredefinedBlock
+	if predefined != nil {
+		return tsr.getTipSetByLabel(ctx, *predefined)
```
rvagg, line 157: this bit is new if you compare to the old `getTipsetByEthBlockNumberOrHash`

node/impl/eth/transaction.go
```diff
@@ -114,101 +94,44 @@ func (e *ethTransaction) EthBlockNumber(ctx context.Context) (ethtypes.EthUint64
 	return ethtypes.EthUint64(parent.Height()), nil
 }
 
-func (e *ethTransaction) EthGetBlockTransactionCountByNumber(ctx context.Context, blkNum ethtypes.EthUint64) (ethtypes.EthUint64, error) {
-	ts, err := e.chainStore.GetTipsetByHeight(ctx, abi.ChainEpoch(blkNum), nil, false)
+func (e *ethTransaction) EthGetBlockTransactionCountByNumber(ctx context.Context, blkParam string) (ethtypes.EthUint64, error) {
+	ts, err := e.tipsetResolver.GetTipsetByBlockNumber(ctx, blkParam, true)
 	if err != nil {
-		return ethtypes.EthUint64(0), xerrors.Errorf("error loading tipset %s: %w", ts, err)
```
rvagg, line 120: the rule is now - don't wrap errors that come out of the `TipSetResolver` because they may be an `ErrNullRound` which won't travel through go-jsonrpc when wrapped, plus the extra wrapped text just makes the error weird and confusing

node/impl/eth/transaction.go
```diff
@@ -489,6 +401,27 @@ func (e *ethTransaction) EthGetBlockReceiptsLimited(ctx context.Context, blockPa
 	return ethReceipts, nil
 }
 
+func (e *ethTransaction) getCidForTransaction(ctx context.Context, txHash *ethtypes.EthHash) (cid.Cid, error) {
```
rvagg, line 404: this and `getBlockByTipset` were just moved

node/impl/eth/utils.go
```diff
@@ -703,53 +646,65 @@ func encodeAsABIHelper(param1 uint64, param2 uint64, data []byte) []byte {
...
-		parent, err := cs.GetTipSetFromKey(ctx, head.Parents())
-		if err != nil {
-			return nil, xerrors.New("cannot get parent tipset")
-		}
-		return parent, nil
-	case "safe":
-		latestHeight := head.Height() - 1
-		safeHeight := latestHeight - ethtypes.SafeEpochDelay
-		ts, err := cs.GetTipsetByHeight(ctx, safeHeight, head, true)
+// decodePayload is a utility function which decodes the payload using the given codec
```
rvagg, line 649: this and `decodeParams` were just moved into utils, they're used by more than one module here

node/impl/full/eth.go
```diff
@@ -3,84 +3,70 @@ package full
...
 
-type EthModuleAPI interface {
-	EthBlockNumber(ctx context.Context) (ethtypes.EthUint64, error)
-	EthAccounts(ctx context.Context) ([]ethtypes.EthAddress, error)
-	EthGetBlockTransactionCountByNumber(ctx context.Context, blkNum ethtypes.EthUint64) (ethtypes.EthUint64, error)
-	EthGetBlockTransactionCountByHash(ctx context.Context, blkHash ethtypes.EthHash) (ethtypes.EthUint64, error)
-	EthGetBlockByHash(ctx context.Context, blkHash ethtypes.EthHash, fullTxInfo bool) (ethtypes.EthBlock, error)
-	EthGetBlockByNumber(ctx context.Context, blkNum string, fullTxInfo bool) (ethtypes.EthBlock, error)
-	EthGetTransactionByHash(ctx context.Context, txHash *ethtypes.EthHash) (*ethtypes.EthTx, error)
+// gatewayWithTrusted exists just to do type assertions on api.Gateway, but we know it won't have
```
rvagg, line 15: yes, this is :vomiting_face: but it's part of the DI dance for gateway which is very hard to resolve without it - the "true" gateway API we want to expose is `api.Gateway`, but to make the gateway reusable as a version of all of the modules it's supposed to be able to implement we need to tack these extra pieces on which end up returning "unsupported" when you try to call them

rvagg:
Good to go now, lots of comments inline and some explainers left as comments here. The diff looks big but the changes are mostly around `TipSetResolver` and the mechanics needed to wire it up differently for separate instantiations of the Eth modules for v1 and v2 APIs. The CHANGELOG contains descriptions of the other changes that may not be so obvious.

node/impl/eth/tipsetresolver.go
```diff
@@ -0,0 +1,207 @@
...
+	// for Ethereum APIs we focus on the message inclusion tipset so we have to go back one, in contrast
+	// to the Filecoin native height selectors
+	ts, err = tsr.cs.LoadTipSet(ctx, ts.Parents())
+	if err != nil {
+		return nil, xerrors.Errorf("loading latest f3 cert tipset %s: %w", tsk, err)
+	}
+	return ts, nil
+}
+
+func (tsr *tipSetResolver) getECTipSetByTag(ctx context.Context, tag string) (*types.TipSet, error) {
```
masih, line 77: Meta comment: As a reader I find the name of this function confusing. It claims to get tipset by tag, yet it only checks for "safe". Speaking subjectively, I would have probably restructured the code such that there is "getLatestSafeTipSet" which is called from controlled execution paths higher up that do indeed check for tags other than safe.

node/impl/eth/transaction.go
masih: This might deserve an in-line comment in code.

masih (approved):
No blockers. A lot of refactoring here which I probably would have separated into their own PR, nevertheless thank you for hammering through it all.

Kubuxu (approved):
SGTM, Masih caught things I would comment on

node/modules/eth.go
```diff
@@ -38,47 +98,76 @@ type EthTransactionParams struct {
...
+			go func() {
+				start := time.Now()
+				log.Infoln("Start prefilling GetTipsetByHeight cache")
+				_, err := params.ChainStore.GetTipsetByHeight(params.MetricsCtx, abi.ChainEpoch(0), params.ChainStore.GetHeaviestTipSet(), false)
+				if err != nil {
+					log.Warnf("error when prefilling GetTipsetByHeight cache: %w", err)
+				}
+				log.Infof("Prefilling GetTipsetByHeight done in %s", time.Since(start))
+			}()
+			return nil
```
Stebalien, line 126: Can we pass in a context that gets canceled by `OnStop`? I guess this is existing code...

node/impl/eth/tipsetresolver.go
rvagg: yeah; I've refactored a bit so we don't pass the tag around anymore:

```go
getLatestF3Cert(ctx context.Context) (*certs.FinalityCertificate, error)
getFinalizedF3TipSetFromCert(ctx context.Context, cert *certs.FinalityCertificate) (*types.TipSet, error)
getSafeF3TipSet(ctx context.Context) (*types.TipSet, error)
getFinalizedF3TipSet(ctx context.Context) (*types.TipSet, error)
getSafeECTipSet(ctx context.Context) (*types.TipSet, error)
getFinalizedECTipSet(ctx context.Context) (*types.TipSet, error)
getTipSetByTag(ctx context.Context, tag string) (*types.TipSet, error)
GetTipsetByBlockNumber(ctx context.Context, blkParam string, strict bool) (*types.TipSet, error)
GetTipsetByBlockNumberOrHash(ctx context.Context, blkParam ethtypes.EthBlockNumberOrHash) (*types.TipSet, error)
GetTipSetByHash(ctx context.Context, blkParam ethtypes.EthHash) (*types.TipSet, error)
```

node/impl/eth/tipsetresolver.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+}
+
+func (tsr *tipSetResolver) GetTipsetByBlockNumber(ctx context.Context, blkParam string, strict bool) (*types.TipSet, error) {
+	if slices.Contains(handledLabels, blkParam) {
+		return tsr.getTipSetByLabel(ctx, blkParam)
+	}
+
+	var num ethtypes.EthUint64
+	err := num.UnmarshalJSON([]byte(`"` + blkParam + `"`))
```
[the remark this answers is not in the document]
rvagg: pulled out a helper function for it to ethtypes

node/impl/eth/tipsetresolver.go
```diff
@@ -0,0 +1,207 @@
...
+	return nil, xerrors.New("invalid block param")
+}
+
+func (tsr *tipSetResolver) GetTipSetByHash(ctx context.Context, blkParam ethtypes.EthHash) (*types.TipSet, error) {
+	ts, err := tsr.cs.GetTipSetByCid(ctx, blkParam.ToCid())
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get tipset by hash: %v", err)
+	}
+	if ts == nil {
+		return nil, xerrors.Errorf("failed to find tipset for block hash %s", blkParam)
```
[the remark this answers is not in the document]
rvagg: Great question .. so I went hunting for the Ethereum answer and found this: https://github.com/ethereum/go-ethereum/blob/cb21177aa8b15861067c430f12439e40fc1ad124/eth/api_backend.go#L165-L168

So I think if you request a tipset that doesn't exist then you should get `nil`, so that's also for the "not found" err case as well.

It's the block version of this for transactions, essentially: https://github.com/filecoin-project/lotus/pull/12999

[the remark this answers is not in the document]
rvagg: Broken out into https://github.com/filecoin-project/lotus/issues/13043, I started tackling some of them but I've backed my changes out, I think doing a half fix is going to be worse than a full one because of the need for pointers on a lot of these (e.g. returning `ethtypes.EthBlock{}, nil` is going to give the user a blob of JSON with empty fields but they should get `nil`)

api/api_gateway.go
```diff
@@ -143,8 +145,10 @@ type Gateway interface {
 	EthGetTransactionByBlockHashAndIndex(ctx context.Context, blkHash ethtypes.EthHash, index ethtypes.EthUint64) (*ethtypes.EthTx, error)
 	GetActorEventsRaw(ctx context.Context, filter *types.ActorEventFilter) ([]*types.ActorEvent, error)
 	SubscribeActorEventsRaw(ctx context.Context, filter *types.ActorEventFilter) (<-chan *types.ActorEvent, error)
-	ChainGetEvents(context.Context, cid.Cid) ([]types.Event, error)
 
 	F3GetCertificate(ctx context.Context, instance uint64) (*certs.FinalityCertificate, error)
 	F3GetLatestCertificate(ctx context.Context) (*certs.FinalityCertificate, error)
+
```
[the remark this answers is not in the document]
rvagg: these are just being moved so they're not as jumbled, `Version` is weirdly only implemented for gateway for a specific purpose https://github.com/filecoin-project/lotus/pull/6618 and it's not there for openrpc-specific purposes

node/impl/eth/tipsetresolver.go
```diff
@@ -0,0 +1,207 @@
...
+	tsk, err := types.TipSetKeyFromBytes(cert.ECChain.Head().Key)
+	if err != nil {
+		return nil, xerrors.Errorf("decoding latest f3 cert tipset key: %w", err)
+	}
+	ts, err := tsr.cs.LoadTipSet(ctx, tsk)
+	if err != nil {
+		return nil, xerrors.Errorf("loading latest f3 cert tipset %s: %w", tsk, err)
+	}
+	// for Ethereum APIs we focus on the message inclusion tipset so we have to go back one, in contrast
+	// to the Filecoin native height selectors
```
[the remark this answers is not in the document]
rvagg: yeah, this is a change of behaviour from existing v1 but I think you're right

node/modules/eth.go
rvagg: yeah, we have `LifecycleCtx` for this, fixed

node/impl/eth/tipsetresolver.go
```diff
@@ -0,0 +1,207 @@
...
+
+var _ TipSetResolver = (*tipSetResolver)(nil)
+
+type tipSetResolver struct {
+	cs ChainStore
+	f3 lf3.F3Backend // can be nil if disabled or this is a v1 resolver
+}
+
+func NewV1TipSetResolver(cs ChainStore) TipSetResolver {
+	return &tipSetResolver{cs: cs, f3: nil}
```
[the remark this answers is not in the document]
rvagg: yep, done, but it was rough

node/impl/eth/tipsetresolver.go
```diff
@@ -21,117 +21,153 @@ import (
...
 			return nil, xerrors.New("failed to get parent tipset")
 		}
 		return parent, nil
-	case "safe":
+	case ethtypes.BlockTagSafe:
 		if tsr.f3 != nil {
-			return tsr.getTipSetByTag(ctx, "safe")
+			ts, err := tsr.getSafeF3TipSet(ctx)
+			if err != nil && tsr.useF3ForFinality {
+				return nil, err
```
Stebalien, line 148: IMO, "safe" should always fallback on EC even if we're using F3 for finality, but I don't have strong opinions on this.

Stebalien (approved):

rvagg:
This is not ready to be merged, latest discussion about what's needed here requires me to make the following changes:

* **NO** behaviour change for `"safe"` for /v1 API as long as F3's manifest is `EC.Finalize=false`; i.e. we are in passive testing mode. We should not be using F3 to change any /v1 behaviour until it's finalizing.
* Change the CHANGELOG docs for v2 APIs to be much clearer about the fact that these APIs are going to return F3 results based on the passive testing we're doing and they should not be relied upon for any serious notion of "finalized" or "safe" because what F3 is returning is not impacting EC yet. We should also call out how you can determine when F3 is finalizing by getting the manifest from the API.

BigLep:
Lets talk during 2025-04-22 standup on how we get this merged.

My thoughts:
1. Lets not touch `/v1` at all.  I understand the desire to reduce difference between what "finalized" and "safe" mean in `/v1` and `/v2`, but that is added scope in my opinion.  I think we should be laser focused on providing a way for builders to start using F3-aware APIs as soon as possible.
2. Having clear docs here makes sense.  To be clear, our messaging is:
  - v2 APIs are highly experimental and can change at any point currently without concern for breakage.
  - v2 APIs are F3 aware, regardless if F3 is being used for finalizing.  Basically if the F3 subsystem is enabled, then F3 will be consulted regardless of the value of `manifest.EC.Finalize`
  - These are the steps for seeing whether the F3 subsystem is being used to finalize: ...

masih:
> v2 APIs are F3 aware, regardless if F3 is being used for finalizing.

To refine this slightly: V2 APIs are F3 aware as long as F3 is enabled, regardless of F3 activation on mainnet.  It really should be obvious that none can rely on F3 in mainnet until it is ...activated!

rvagg:
> Lets not touch /v1 at all

I'm find with this and will adjust, although the process of getting here has made some improvements to both forms of the eth APIs which I'll document but I'll leave `"safe"` and `"finalized"` alone for now.

But, we should note the original reason we started touching them here and not lose track of this as a future area of work. The `200` epoch lookback reflects a pessimistic view that if F3 isn't operational or delayed for some reason, then something must not be going well, so we take a very conservative view that EC has the potential for some major reorganisation if F3 is experiencing a hiccup. In that case, we're still going to be offering `30` to `/v1` users, which is probably not great. So the proposal here was to match the pessimistic `200` of `/v2` but give them finality as a consolation so in normal behaviour their `"safe"` is actually sooner than it was before.

BigLep:
> Lets not touch /v1 at all

I was saying this on principle but hadn't looked at the PR.  I looked at it more today and I see there is more nuance.  Apologies if I'm causing whiplash here - my intent was to simplify the conversation so we could hopefully land something sooner without a lot of discussion.

Going through the changelog:

> `/v2 "finalized"` tag maps directly to the F3 finalized epoch (much closer to the chain head).

👍 

> `/v2 "safe"` tag maps to the F3 finalized epoch or 200 epochs behind head, whichever is more recent.

👍 

> `/v1 "finalized"` tag continues to use a fixed 900-epoch delay from the head.

👍 

> `"safe"` tag uses a 30-epoch delay *or* the F3 finalized epoch (whichever is more recent). *After* F3 is fully active on the network, the behavior of `"safe"` on `/v1` will align with `/v2`'s `"safe"`.

The rationale makes sense and maybe this is where we want to end eventually?  If the code is easier to reason about if we have consistency between /v1 and /v2 then I'm more apt to leave what you have or at least just simplify to assuming F3 is active since it will be shortly.  I can also see the mindset of "don't change the semantics of v1 and we potentially improve it later".

> Both: Previously, `"finalized"` and `"safe"` tags referred to epochs `N-1`. This `-1` offset has been removed in both V1 and V2.

👍 

> Both: `eth_getBlockTransactionCountByNumber` now accepts standard Ethereum block specifiers (hex numbers _or_ labels like `"latest"`, `"safe"`, `"finalized"`).

👍 

> Both: Methods accepting `BlockNumberOrHash` now support all standard labels (`"pending"`, `"latest"`, `"safe"`, `"finalized"`). This includes `eth_estimateGas`, `eth_call`, `eth_getCode`, `eth_getStorageAt`, `eth_getBalance`, `eth_getTransactionCount`, and `eth_getBlockReceipts`.

👍 

> Both: Removed internal `Eth*Limited` methods (e.g., `EthGetTransactionByHashLimited`) from the supported gateway API surface.

👍 

> Both: Improved error handling: block selection endpoints now consistently return `ErrNullRound` (and corresponding JSONRPC errors) for null tipsets.

👍

rvagg:
Pushed changes that roll back the use of F3 in /v1 _at all_ but note this may change in the future. I was going to roll back the `-1` offset too but decided that it may as well be addressed now since I'm touching all of this and we have new opinions on tipset vs parent for /v2, so I left that.

Worked on the docs a bit too, but feel free to hack at that to make it clearer.

CHANGELOG.md
```diff
@@ -13,13 +13,35 @@
...
+
+See [Filecoin v2 APIs](https://filoznotebook.notion.site/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658) for an in-depth overview. ([filecoin-project/lotus#13003](https://github.com/filecoin-project/lotus/pull/13003)), ([filecoin-project/lotus#13027](https://github.com/filecoin-project/lotus/pull/13027)), ([filecoin-project/lotus#13034](https://github.com/filecoin-project/lotus/pull/13034))
+
+**F3-aware Ethereum APIs via `/v2` endpoint and improvements to existing `/v1` APIs** ([filecoin-project/lotus#13026](https://github.com/filecoin-project/lotus/pull/13026))<br>
+
+Lotus now offers two versions of its Ethereum-compatible APIs (`eth_`, `trace_`, etc.) with different finality handling:
+* **`/v2` APIs (New & Experimental):** These APIs consult the F3 subsystem (if enabled) for finality information.
+  * `"finalized"` tag maps to the F3 finalized epoch.
```
BigLep, line 33: What happens in these scenarios:
1. F3 isn't enabled
2. F3 isn't providing a finality certificate
3. Last F3-finalized tipset is more than 900 tipsets behind
?  
I assume we want to fall back to EC in all these cases.

(Some of these cases are covered in the diagram in https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658?pvs=4#1d0dc41950c1802991aae0020b4d8e05 - but the diagram doesn't cover when  the last F3-finalized tipset is more than 900 tipsets behind

rvagg:
> I see /v2 also has support for

.. I've updated the CHANGELOG to have a bit more detail, I was intending it to be implicit that they're all covered. Everything in https://github.com/filecoin-project/lotus/blob/master/api/eth_aliases.go is covered, those are standard Ethereum calls (or Erigon), plus our couple of custom extra addressing ones that we don't alias.

CHANGELOG.md
masih: For scenario 3 we should return EC finalized. I'll get this updated on Filecoin v2 and reflect on here.

CHANGELOG.md
```diff
@@ -13,13 +13,35 @@
...
+
+### Experimental v2 APIs with F3 awareness
+
+The Lotus V2 APIs introduce a powerful new TipSet selection mechanism that significantly enhances how applications interact with the Filecoin blockchain. The design reduces API footprint, seamlessly handles both traditional Expected Consensus (EC) and the new F3 protocol, and provides graceful fallbacks.
+
+> [!NOTE]
+> V2 APIs are highly experimental and subject to change without notice.
+
+> [!IMPORTANT]
+> The `/v2` APIs rely on F3 data even if F3 is not yet finalizing the chain (i.e., `EC.Finalize` is `false` in the live F3 manifest used by all participants). To determine if F3 is actively finalizing, call the `F3GetManifest` API and check if `Manifest.EC.Finalize` is `true`. Only when `EC.Finalize` is `true` will the `/v2` `"finalized"` and `"safe"` tags accurately reflect the chain's finality according to F3. This also applies to Ethereum APIs discussed below.
```
masih, line 25: By the time we make a release F3 would hopefully be active on mainnet. Which makes a lot of this kind of notices redundant. But until then, thanks for adding these.

CHANGELOG.md
masih: My vote is to land this PR then iterate. As is this one is pretty giant.

masih (approved):

Kubuxu (approved):

node/impl/eth/tipsetresolver.go
masih: Deferring further discussion to https://github.com/filecoin-project/lotus/issues/13060

CHANGELOG.md
masih: Captured issue here: https://github.com/filecoin-project/lotus/issues/13062

PR #12822, opened 2025-01-11
feat(exec): dump intermediate cache blocks from FVM exec in StateReplay

rvagg:
* Plumb through dump_cache from fvm4 to access intermediate blocks:
  - https://github.com/filecoin-project/filecoin-ffi/pull/512
  - https://github.com/filecoin-project/ref-fvm/pull/2101
* Enable cache dumping in StateReplay with LOTUS_REPLAY_DUMP_CACHED_BLOCKS
* Add optional "Blocks" field InvocResult
* Handle ExecutionEvent::Log's and add "Logs" field to ExecutionTrace
* Dump intermediate cache blocks to CAR in /tmp when they appear while using `lotus-shed msg --exec-trace`.

As with https://github.com/filecoin-project/filecoin-ffi/pull/512, I may bail on some or all of this, it's a little bit janky (but it is for debugging!). But the `ExecutionEvent::Log` thing probably needs to stay, so that may become a new PR.

PR #13027, opened 2025-04-10
feat: implement initial Lotus V2 `State` APIs

masih:
Implement the Lotus V2 APIs for :

* `StateGetActor` - enhanced to accept tipset selector.
* `StateGetID` - enhanced to accept tipset selector / renamed from `LookupID` in v1 for consistency.

Tests are added to exercise the end-to-end API call through JSON rpc with the raw wire format of the JSON-RPC request. Additional factory functions are added for `TipSetLimit`, a new concept to specify a limit for tipsets either as absolute height or relative to some tipset selector. The future State message search APIs in v2 aim to use Limit in upcoming PRs.


rvagg:
This is fine, except for the `StateCompute` and `StateSimulate`, I'm very hesitant to support proliferation of these APIs.

Rather than pollute this issue I broke out a new issue to dump my current thoughts over @ https://github.com/filecoin-project/lotus/issues/13028

I also followed up looking at how useful `TipSetLimit` is, and I don't think it's all that helpful in this case because our own uses of `StateCompute` have tipset==height, so we'd have to find external consumers for who this is useful I think: https://github.com/filecoin-project/lotus/issues/13028#issuecomment-2795479294

So maybe we should hold off on the `StateCompute` and `StateSimulate` for now so we can do a more thoughtful job of replacing our current re-execution APIs.

CHANGELOG.md
```diff
@@ -9,20 +9,37 @@
...
 
-- fix(eth): always return nil for eth transactions not found ([filecoin-project/lotus#12999](https://github.com/filecoin-project/lotus/pull/12999))
-- feat: add experimental v2 APIs that are "F3 aware."  (TODO: expand this section significantly to cover where someone learns about the new APIs, how they enable them, and what expectations they should have around them—i.e., they may change)
+- fix(eth): always return nil for eth transactions not
+  found ([filecoin-project/lotus#12999](https://github.com/filecoin-project/lotus/pull/12999))
+- feat: add Lotus v2 experimental APIs with F3 awareness
+  The Lotus V2 APIs introduce a powerful new TipSet selection mechanism that significantly enhances how applications
+  interact with the Filecoin blockchain. The design reduces API footprint, seamlessly handles both traditional Expected
+  Consensus and the new F3 protocol, and provides graceful fallbacks.
+  See [Filecoin v2 APIs](https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658) for an
```
BigLep, line 18:
```suggestion
  See [Filecoin v2 APIs](https://filoznotebook.notion.site/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658?pvs=4) for an
```

Switched to a publicly accessible link

CHANGELOG.md
BigLep, line 1: Lots of additional line wrapping changes in this file that I assume we want to avoid bringing in with this change.

api/v2api/full.go
```diff
@@ -49,4 +52,108 @@ type FullNode interface {
...
+	// The TipSetSelector parameter provides flexible options for selecting the base tipset:
+	//   - TipSetSelectors.Latest: the most recent tipset with the heaviest weight
+	//   - TipSetSelectors.Finalized: the most recent finalized tipset
+	//   - TipSetSelectors.Height(epoch, previous, anchor): tipset at the specified height
+	//   - TipSetSelectors.Key(key): tipset with the specified key
+	//
+	// See types.TipSetSelector documentation for additional details.
+	//
+	// Experimental: This API is experimental and may change without notice.
+	StateCompute(context.Context, []*types.Message, types.TipSetSelector) (*api.ComputeStateOutput, error) //perm:read
```
BigLep, line 126: I have no calling experience with this API, but as described I was thinking the `TypeSetSelector` should be before the `Message[]` in following with the procedure (load the tipset, apply the messages).  Although maybe clients think about this more as "compute state for these messages against this tipset"?  (Feel free to ignore the comment)

chain/types/tipset_selector_test.go
```diff
@@ -79,3 +81,156 @@ func TestTipSetSelector_Marshalling(t *testing.T) {
...
+		},
+		{
+			name:        "both height and distance",
+			jsonInput:   `{"height": 123, "distance": 456}`,
+			expectError: true,
+		},
+		{
+			name:          "empty object",
+			jsonInput:     `{}`,
+			expectedLimit: types.TipSetLimits.Unlimited,
```
BigLep, line 110: I'm not up on other conventions with the API but would it be more clear to have this be more self documenting?

`{"distance":"unlimited"}`

node/impl/full/state_v2.go
```diff
@@ -0,0 +1,84 @@
...
+
+type StateModuleV2 struct {
+	State StateAPI
+	Chain ChainModuleV2
+
+	fx.In
+}
+
+// TODO: Discussion: do we want to optionally strict the type of selectors that
+//       can be supplied here to avoid foot-guns? For example, only accept TipSetKey
```
BigLep, line 32: What foot guns are you anticipating?

BigLep (commented):
Things that came to mind while following along.  Nothing blocking here.

BigLep:
> So maybe we should hold off on the `StateCompute` and `StateSimulate` for now so we can do a more thoughtful job of replacing our current re-execution APIs.

I was wondering something similar.  Do we focus on shipping some clear cut wins initially (e.g., ETH API group and read APIs for Chain/State) before gettng bogged down in the more complicated cases.

rvagg:
Now that the topic has been opened up, I'm thinking that instead of modifying the /v1 API as I'm doing in https://github.com/filecoin-project/lotus/pull/12822 maybe we should just start working on a /v2 replacement for these APIs. But it's a low priority and I don't think needs to be part of an initial release of this work, unless we're getting signal from elsewhere that this is an important API?

CHANGELOG.md
masih: My bad, I formatted it by accident.

chain/types/tipset_selector_test.go
masih: That would make `distance` a mixed type field which I think would be more trouble than it's worth.

node/impl/full/state_v2.go
masih: [This](https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658#1d0dc41950c180dc8eb9d1afc46cc641) specifically, where a naïve programmer would repeatedly use tipset selectors with no concrete key for calls that should care about cross consistency.

masih:
> But it's a low priority and I don't think needs to be part of an initial release of this work, unless we're getting signal from elsewhere that this is an important API?

`StateCompute` is the API that was mentioned by Zondax as "used a lot. Which is why I added it to the initial pass at `State` v2 APIs.

masih:
> modifying the /v1 API as I'm doing in https://github.com/filecoin-project/lotus/pull/12822 maybe we should just start working on a /v2 replacement for these APIs.

This is consistent with my take, in that if there is not a fire to put out in v1 APIs I would say the energy is best spent on getting v2 the best that it can be. I am also conscious that I don't know as much on the context of the work involved in modifying v1 APIs.

api/v2api/full.go
masih: I will be removing state compute from this PR to add it in a separate PR later. But to address the comment raised:

The way I read these stubs in my head is : Do X Given Y at TipSet. From that angle tipset is less important to the operation that is being asked for _when we are not explicitly asking for tipset_. Therefore, I have set tipset selector to be the last argument in APIs that need to know at which tipset to do something but their primary function is not to get tipset etc.

I can totally see that some other ordering as you have kindly pointed out @BigLep could make sense. I am also aware that the way my head works is not popular, so I defer to the popular vote. 

@Kubuxu @rvagg and anyone interested, please speak up if you have opinions on the ordering as we would have a lot of APIs that take tipset selectors in v2.

chain/types/tipset_selector_test.go
masih: Resolving as tipset limit is now removed from this PR. Happy to discuss this further if/when it is re-introduced.

api/v2api/full.go
rvagg: Yeah, I agree with @masih and I think we'll be well served by trying to massage our verbs so that the "at TipSet" comes last. Our existing APIs have a strong pattern of it being last, I even have muscle-memory now for writing `curl` requests with `[]` last in my parameters array!

But that does remind me, maybe we ought to consider making sure `[]` still works as an alias for `"latest"`? Right now, it's the equivalent of supplying `types.EmptyTSK`. I think the equivalent with `TipSetSelector` would be `{"key":[]}` which is more verbose.

node/impl/full/state_v2.go
rvagg: I want to be able to call this (and other) APIs with `"latest"`, that's pretty typical because you often have good reason to believe that (a) the actor stably exists and (b) you're not expecting to use the return for anything critical. I suppose switching to `"finalized"` as the default usage pattern would be ideal after activation. But if we allow those tags, what else is there to restrict?

rvagg (approved):

api/v2api/full.go
masih: > But that does remind me, maybe we ought to consider making sure [] still works as an alias for "latest"?

Happy to accommodate this, except that:
* it would default to `finalized` instead of `latest` 
* prior feedback asked for explicit API arguments (discussed in F3 standup with @Kubuxu and summarised [here](https://github.com/filecoin-project/lotus/pull/13003#issuecomment-2778113370))

I have no strong opinions on this. For me the fact that we are stuck on using by-position params instead of by-name  makes the wire format ugly enough that I have stopped spending too much time polishing the by-position variant - discussed in length at #13003.

rvagg: ok, well no urgency, let's roll with this and I'll see how much I hate the verbosity cause I use curl more than a few times every week

masih:
Updated the [Filecoin v2 API docs](https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658) to include the new APIs introduced here.

Merging and happy to address any further comments retrospectively.

api/v2api/full.go
BigLep: > Yeah, I agree with @masih and I think we'll be well served by trying to massage our verbs so that the "at TipSet" comes last. Our existing APIs have a strong pattern of it being last, I even have muscle-memory now for writing curl requests with [] last in my parameters array!

Sounds good!

> For me the fact that we are stuck on using by-position params instead of by-name makes the wire format ugly enough that I have stopped spending too much time polishing the by-position variant - discussed in length at https://github.com/filecoin-project/lotus/pull/13003.

This made me realize that we should document well the design rationale for why we don't allow named parameters.  I have started a section here (https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658?pvs=4#1d2dc41950c180019135f921e171e984 ), but I think I missed the verbal conversation on this to be able to fill in the holes.

masih: I made it clear in that section that the rationale is lifted directly from PR convo. 

It does not fully rationalise it for me. but hey.

PR #13034, opened 2025-04-11
feat: implement `safe` tag in tipset selection

masih:
Implement the ability to select a tipset by tag `safe`. The safe tag would select the earliest between finalised and heaviest - 200.


build/parameters.go
```diff
@@ -82,3 +82,5 @@ const Eip155ChainId = buildconstants.Eip155ChainId // Deprecated: Use buildconst
 var WhitelistedBlock = buildconstants.WhitelistedBlock // Deprecated: Use buildconstants.WhitelistedBlock instead
 
 const Finality = policy.ChainFinality // Deprecated: Use policy.ChainFinality instead
+
+const SafeHeightDistance = 200
```
BigLep, line 86: If we are going to let the concept of safe be exposed from more areas than just the eth APIs, I think we should make it consistent.  Basically, is there a reason for ETH APIs to use one value and a different value here?

Ideally can we just use https://github.com/filecoin-project/lotus/blob/811d25c3e29bbda4d188d78f6b41c5056c378ccf/chain/types/ethtypes/eth_types.go#L46 or have that use this build parameter?  And if it's defined both places, cross link it (i.e., if you change this value here please also change over there).

BigLep (commented):
Everything looks good to me other than introducing a different safe value for non-ETH APIs

node/impl/full/chain_v2.go
```diff
@@ -57,13 +58,38 @@ func (cm *ChainModuleV2) ChainGetTipSet(ctx context.Context, selector types.TipS
...
+}
+
+func (cm *ChainModuleV2) getLatestSafeTipSet(ctx context.Context) (*types.TipSet, error) {
+	finalized, err := cm.getLatestFinalizedTipset(ctx)
+	if err != nil {
+		return nil, xerrors.Errorf("getting latest finalized tipset: %w", err)
+	}
+	heaviest := cm.Chain.GetHeaviestTipSet()
+	switch {
+	case finalized == nil:
```
Kubuxu, line 80: This doesn't seem right

build/parameters.go
masih: Happy to change it, but if exchanges don't use the distance of 30 then I have left wondering how useful this tag would be in the context we discussed with Zondax.

Of course exchanges can use a different build param and pick whatever number they want. 

Another key question is: Despite what's being picked for Eth APIs, what do we want the official `safe` distance to be for Filecoin v2 APIs?

Having reviewed FRC-0089, at least we a have math behind the distance 30 which is more than I can say about the distance 200.

node/impl/full/chain_v2.go
```diff
@@ -57,13 +58,38 @@ func (cm *ChainModuleV2) ChainGetTipSet(ctx context.Context, selector types.TipS
...
+		return heaviest, nil
```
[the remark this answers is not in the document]
Kubuxu: Reason being that if finalized is nil for whatever reason, now we are returning heaviest as safe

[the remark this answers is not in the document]
masih: My bad fixing.

Kubuxu (approved):

build/parameters.go
masih: I have reached out to Zondax to ask about 200 vs 30 for filecoin tipsets.

rvagg:
`30` for Eth as it's encoded now comes partly from FRC-0089 related work. That FRC notes that the 900 epochs are meant to provide a 2^-30 guarantee of finality at 900 epochs, hence that choice for "finalized" today. Here's a [brief thread](https://filecoinproject.slack.com/archives/C02D73MHM63/p1718693790469889) of me looking at this by running the finality calculator from the FRC against recent (at that time) epochs where the chain was relatively healthy with a win count of ~5.

Here's what looks like with that 2^-30 guarantee line in green:

image

log scale:

image

When looking at this, I didn't observe it go beyond 25 epochs to get to this level, even when the chain got a bit choppy. But I don't think I went back and looked at particularly problematic points in the life of the chain either so it may get worse. But from a "not necessarily final, but most probably safe", 30 seemed like a reasonable choice.

There was a page somewhere that listed the current bridges and their choices of finality time, they ranged from 60 to 200 IIRC, that also gives an indication of what is considered "_almost_ final" but mostly I think that just comes from our recommendations anyway. @eshon do you have any more info on this?

IMO 30 is a good compromise, it's not finalized, but having a "safe" that's distinct from "finalized" should suggest that it's not going to be 100%, you'd pick "finalized" if you wanted that. Hopefully with F3 working properly this becomes mostly irrelevant.

rvagg:
I have a semi-working FRC-0089 finality calculator in Go but got stuck with some of the math library dependencies so never managed to finish it, at least I got frustrated enough to put it down. I could dig that code up if someone wants to help get it over the line.

rvagg (approved):
yep, good by me, I'll even adjust ETH APIs for /v2 to follow this behaviour when f3 isn't available, I'll leave /v1 alone for now

build/parameters.go
rvagg: @masih this should be in `build/buildconstants/params_shared_vals.go`, don't even leave an alias for it, the rest of these are deprecated so external consumers get a slim package to import if they only want constants.

rvagg: I'm putting this into my Eth branch and decided to document it, so in that file I have:

```go
// SafeHeightDistance is the distance from the current head that we call "safe" for the purpose of
// assuming finality when relying on EC. It's not a substitute for ChainFinality, but a compromise
// where it's acceptable to have a nonzero margin of error in the finality assumption.
const SafeHeightDistance = 200
```

Feel free to take and edit, I'll rebase ontop of this when it's merged.

eshon:
@rvagg - This is a comparison of high TPS chains that can be sorted by finality... https://chainspect.app/dashboard

(Bridge finality will depend on the finality of the chains being used but AxelarScan can probably update confirmations of Filecoin bridged txns based on updates to the Filecoin Eth RPC API.)

I am much more concerned that we are going to confuse exchanges greatly if we message them about F3 too soon because we won't be able to share expected confirmation times as in the [Kraken Docs on confirmation times](https://support.kraken.com/hc/en-us/articles/203325283-Cryptocurrency-deposit-processing-times), and vigilant network monitoring will be required for a while until F3 participation is tied to BRs.

It would be bad PR if F3 halting became a news item that we had to explain to exchanges. Perhaps FilOz should delay F3 messaging for a couple months until further analysis can be done on its performance.

build/parameters.go
masih: No definitive consensus on this, which leaves me to pick a more conservative approach and revisit once more data is collected. Leaving this to be 200.

masih:
Echoing my response to the review comment above:
* 200 is the most reasonable and conservative safe distance value compared to smaller alternatives.
* To truly understand the fitness of this constant for its purpose more data is needed, specifically empirical measurements of F3 on mainnet. This is being performed as part of passive testing right now.
* The entire v2 API is marked as experimental and these constants can and will change.

Based on the rationale above, I am merging this PR and postponing further discussion on the value of "safe" constant until there is more data.

build/parameters.go
masih: Apologies. Due to cached UI i missed your comment. Thanks for cleaning this up in your PR 🙏

rvagg: done in https://github.com/filecoin-project/lotus/pull/13026, copied even more of your test cases for my own versions of them while I was at it

masih: thanks

PR #13049, opened 2025-04-22
[skip changelog]  chore(FIP-0100): remove deprecated code

tediou5:
## Checklist

- [ ] CI is green


BigLep:
@Stebalien : can you please look given @rvagg is out?

BigLep:
(This is followup to https://github.com/filecoin-project/lotus/pull/12919 now that we've upgraded to nv25)

BigLep:
@rvagg : I assume you'll look later this week when you come back online.

storage/pipeline/commit_batch.go
```diff
@@ -233,13 +233,6 @@ func (b *CommitBatcher) maybeStartBatch(notif bool) ([]sealiface.CommitBatchRes,
 
 	individual := (total < cfg.MinCommitBatch) || (total < miner.MinAggregatedSectors) || blackedOut() || !cfg.AggregateCommits
 
-	// TODO: remove after nv25 (FIP 0100)
```
rvagg, line 236: @tediou5 there's also line 388 in this file, using `policy.AggregateProveCommitNetworkFee`, that's now a noop that could be removed.

storage/pipeline/commit_batch.go
```diff
@@ -233,13 +233,6 @@ func (b *CommitBatcher) maybeStartBatch(notif bool) ([]sealiface.CommitBatchRes,
 
 	individual := (total < cfg.MinCommitBatch) || (total < miner.MinAggregatedSectors) || blackedOut() || !cfg.AggregateCommits
```
rvagg, line 234: this line concerns me a bit and I think we should consider tweaking it

1. We could remove all the `blackedOut` code, it's for Skyr and I don't think we need backward compatibility - if we have failing tests because of it we could just remove those tests; I don't think we have a policy of lotus miner being backward compatible, just lotus.
2. I'd like to propose we remove `AggregateCommits` from the config entirely. I can't think of a good reason it should still exist.

But, what I don't have a good handle on is the way that `AggregateCommits` is used on line 210 where we hold off on a batch if it's `true`. I worry that the `AggregateCommits` setting might be confusing two things - whether to use aggregate proofs and how to handle the number of commits in a single PCS3.

@tediou5 would you mind taking a look at this? I'd prefer us to not change the logic that determines when to send a commit message but we want it to default to aggregate proofs whenever possible. I think those two might be entangled.

tediou5: Of course, removing AggregateCommits is a good idea. Let me check the surrounding code.

tediou5: @rvagg I’ve looked through the code and nothing seems to be blocking. There’s just one small issue: if a user really does want to submit only a single ProveCommit (I know that’s unlikely, but it could happen), then with aggregation still forced on they’d have to wait up to 24 hours or manually run publish-now. Do you think that’s acceptable? Also, lotus-shed still contains mathProveCommitAggFeesCmd and mathPreCommitAggFeesCmd; I think we could remove those two methods as well, right?

rjan90: > if a user really does want to submit only a single ProveCommit (I know that’s unlikely, but it could happen), then with aggregation still forced on they’d have to wait up to 24 hours or manually run publish-now.

I´m inclined to say that we are fine with this condition, given that most Lotus-Miner users know the `lotus-miner sectors batching commit --publish-now` cmd

rjan90:
@rvagg could you look over the latest changes here, and see if things look okay so that we can get it merged?

BigLep:
@rvagg : repinging in case you have bandwidth now for this.

rvagg (approved):
yeah, let's do it! so sorry for the delay @tediou5 and thanks for the hard thinky work here

rvagg:
I've added a CHANGELOG entry in here and rebased on master. @rjan90 I'll let  you adjust changelog through the release process if it doesn't suit your preferences.

PR #13050, opened 2025-04-22
chore: sysctl template, service burst limits

Kubuxu:


scripts/98-udp-buf.conf
```diff
@@ -0,0 +1,2 @@
+net.core.rmem_max=7500000
```
rvagg, line 1:
```suggestion
# This configuration file increases the UDP buffer size limits for the system.
net.core.rmem_max=7500000
```

rvagg (approved):

Issue #13062, opened 2025-04-24
Fall back to EC when F3 finalized tipset is further than 900 epochs behind


masih:
In an edge case when F3 falls too far behind head, fall back to EC finality if the F3 finalized tipset is  > 900 epochs behind.

This needs to be reflected in V2 APIs for both Filecoin and Eth.

For context, see https://github.com/filecoin-project/lotus/pull/13026#discussion_r2057358515.

PR #13066, opened 2025-04-24
feat: fall back to EC if F3 finalized tipeset is older than 900 epochs

masih:
In Lotus v2 APIs, in an event where F3 finalized tipset is too far behind EC, > 900 epochs return EC finalized tipset.

Part of: https://github.com/filecoin-project/lotus/issues/13062


BigLep (commented):
Conceptually seems right/good to me and good to see a test.  Will leave to someone else who commits code to the repo to approve.

Kubuxu (approved):
It feels weird that this code is so duplicated
I'm cautious about needing finality resolution in more places. Once the third duplicate appears, let's refactor it out.

PR #13067, opened 2025-04-24
ci: migrate workflows to checkout v4

dizer-ti:
Bumps checkout to v4 for future-proofing against Node 20 runner updates. Workflows compile the same.


masih (approved):
Thank you for the contribution 👍

PR #13071, opened 2025-04-25
feat: implement encoding.TextMarshaler/Unmarshaler for EthUint64

Kubuxu:


chain/types/ethtypes/eth_types.go
```diff
@@ -55,34 +56,52 @@ const (
...
-			return err
-		}
-		eint := EthUint64(parsedInt)
-		*e = eint
-		return nil
-	} else if eint, err := strconv.ParseUint(string(b), 10, 64); err == nil {
-		*e = EthUint64(eint)
-		return nil
+	// if string literal, remove quotes
+	if b[0] == '"' && b[len(b)-1] == '"' {
```
Stebalien, line 97: Use `json.Unmarshal`. This won't handle:

1. Strings < 2 bytes.
2. JSON escape sequences.

Etc.

Stebalien (changes requested):

BigLep:
@Kubuxu : I'm marking this as draft until feedback is incorporated.  Then mark it ready for review so it's clear that others should look.

Issue #13036, opened 2025-04-11
Expose initial F3-aware `/v2/` APIs through lotus-gateway


BigLep:
For https://github.com/filecoin-project/lotus/issues/13004 we intend to release some experimental `/v2` APIs that are F3 aware so that users can start using F3 in some way.

The minimum set we're expecting are:
- https://github.com/filecoin-project/lotus/issues/12988
- [ChainGetTipSet](https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658#1d0dc41950c181cba489f82d6879357d)
- [StateGetActor](https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658#1d2dc41950c180b5a11cde4c7d97be87)
- [StateGetId](https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658?pvs=4#1d2dc41950c1803cbea0e957ccf22fe3)

Once we have those, they should also be exposed through lotus-gateway so that RPC providers like protofire can easily expose them.

BigLep:
This request came from Protofire in private correspondence: https://filecoinproject.slack.com/archives/C03TM2QDTHN/p1744212364990709?thread_ts=1744186355.432429&cid=C03TM2QDTHN

Kubuxu:
I looked at taking a quick stab at it, but it wasn't a quick stab.

Kubuxu:
The problem is that gateway doesn't handle v0/v1/v2 at all currently/

PR #13075, opened 2025-04-28
feat: expose `/v2` APIs through Lotus Gateway

masih:
Integrate the existing v2 APIs into Lotus Gateway such that they are accessible via `/rpc/v2` endpoint similar to full node APIs except through the gateway reverse proxy along with rate limiting and handling of stateful connections over web socket.

The prior gateway implementation did not have a concept of per-uri reverse proxy handling. The changes here separate reverse proxy implementation in gateway from the gateway node to introduce separate handling of v1 and v2 APIs. The work introduces a top-level `v2api.Gateway` interface to capture the APIs exposed by the v2 gateway and generate any relevant documentation based on the interface.

Specifically, the work here separates the handling of stateful connections between v1 and v2 to avoid any cross-contamination between Eth calls to either API. However, the total rate limiting has stayed shared across both URIs.

The underused mock interface in v1 is now removed in favor of the generated gomock interfaces to reduce the LOC in `gateway` package. The interface was barely used by the tests and would have resulted in unnecessary LOC to implement v2.

I have held back on unrelated refactoring of `gateway` package. The code can use further restructuring, but for now the focus is to expose existing /v2 APIs over the gateway.

The implementation of `Version` in /v2 APIs is also left as a TODO since it requires implementation at the top level API and bubbles up into `Common` APIs.

Fixes https://github.com/filecoin-project/lotus/issues/13036


Kubuxu (approved):
SGTM although I haven't reviewed the proxy files in detail

masih:
Thanks Kuba.

> the proxy files

Those are mostly a fork of v1 behaviour with minor adjustments to use selectors (i.e. v2 APIs ) for eth block hash to tipset resolution as well as separate stateful connection handling.

masih:
Landing this and will iterate in the background on improving test/further refactoring. Happy to reflect on any comments retrospectively.

PR #13107, opened 2025-05-07
feat(f3): prefetch the power table of every parent tipset

Stebalien:
Given that we lag F3 by 4 epochs, this means we SHOULD already have the appropriate power table in the cache before we try to agree on it. This is very "best effort" and won't be very effective during catch-up.

Honestly, the _right_ way to do this is likely to have F3 pre-fetch and queue up future tipsets, but that's more complicated.

follows #13106

### DIFF
--- a/chain/lf3/ec.go
+++ b/chain/lf3/ec.go
@@ -15,6 +15,7 @@ import (
 	"github.com/filecoin-project/go-f3/gpbft"
 	"github.com/filecoin-project/go-state-types/abi"
 
+	"github.com/filecoin-project/lotus/build/buildconstants"
 	"github.com/filecoin-project/lotus/chain"
 	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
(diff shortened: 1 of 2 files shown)


Kubuxu:
This has a nasty side effect of causing non-useful work while F3 is behind. Pre-fetch will override itself before the work becomes useful.

Stebalien:
Hm. I thought our cache was bigger. One option is to only cache so far beyond the last requested tipset but... that defeats the point of doing this in lotus (trying to keep it simple).

Given that we already lag F3 by 4 epochs, the right approach is (IMO) to do this prefetching in F3 itself by:

1. Fetching the chain we need immediately.
2. Spawning an async task to pre-fetch up to the max tipsets per-instance power-tables (or doing this in some task worker).

(and fetch power tables in parallel in general).

Kubuxu:
Yeah, that is my plan. Thanks for the code in any case.

PR #13112, opened 2025-05-08
fix(test): make immutable deadline check more conservative

rvagg:
Fixes: https://github.com/filecoin-project/lotus/issues/13098


BigLep:
@aarshkshah1992 @ZenGround0 : please take a look so we can get this merged.

ZenGround0 (approved):
Ok this actually makes sense given how the test works.  The fact that this method is unused anywhere else and the behavior is so specific to this test makes me suspect its going to be cleaner to just put MaybeImmutableDeadline inline.

PR #13118, opened 2025-05-10
feat(ci): only markdown action to skip some CI workflows

### LINKED ISSUE #13069: Pure markdown documentation updates skip CI workflows like tests
## Done Criteria
Making a pure markdown change doesn't trigger various build/test CI workflows.  Specifically, these workflows can be skipped if the PR only has markdown document changes:
1. build.yml
2. check.yml
3. docer.yml
4. test.yml

The boiler plate should be minimized as much as possible.  For example, the logic for this "was only markdown modified check" should be factored to its own action rather than inlined into each of the above files.

Ideally this "was only markdown modified" check action is run at the beginning of each workflow and if true, the workflow short circuits without failing.  This can mean completing successfully or being cancelled.

## Why Important
1. It reduces the turnaround time for pure documentation changes.
2. It incurs unnecessary CI costs. 

## User/Customer
Maintainers, particularly anyone writing docs.

## Notes
1. Changes to source code comments only should still run all the checks.  It is only pure markdown changes that are exempt.
2. Getting a workflow to short-circuit / early-exit appears not to have firstclass support in GitHub actions per https://github.com/actions/runner/issues/662.  https://stackoverflow.com/questions/60589373/how-to-force-job-to-exit-in-github-actions-step discusses some options, but it looks like each step in the workflow will need to look at the result of the "was only markdown modified" check.  
3. This is a great AI agent task.  I started to just do that, but then pulled back because:
   1. I wanted to confirm other maintainers agree with this issue.
   2. I wanted to have a better record of the decision / discussion in case it becomes relevant for other repos.
   3. A well scoped / documented issue should be mostly all the AI Agent needs anyways to do the work.  
4. I have a vague recollection that there was an issue (or maybe slack discussion) about this in the past, but I couldn't find it.  I finally was motivated to take action after doing v2 API documentation PRs.

kamuik16:
## Related Issues

Closes #13069 

## Proposed Changes

- Add a new composite action Only Markdown Changed to detect when a PR modifies only .md files.
- Introduce a planner job in each workflow that runs this action and outputs only_docs.
- Update build, test, check, and docer jobs to run only if only_docs != 'true'.
- Skip all heavy CI steps automatically on docs-only PRs, cutting CI runtime and costs.

## Checklist

- [ ] CI is green

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,9 +20,24 @@ permissions:
   contents: read
 
 jobs:
+  planner:
+    name: Detect docs-only changes
+    runs-on: ubuntu-latest
+    outputs:
+      only_docs: ${{ steps.check.outputs.only_markdown_changes }}
+    steps:
+      - name: Check for markdown-only changes
+        id: check
+        uses: ./.github/actions/only-markdown
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
   build:
     name: Build
+    needs: planner
+    if: ${{ needs.planner.outputs.only_docs != 'true' }}
     runs-on: ubuntu-latest
+
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/actions/only-markdown/action.yml
+++ b/.github/actions/only-markdown/action.yml
@@ -1,0 +1,39 @@
+name: Only Markdown Changed
+description: Check whether every changed file in this PR has the .md extension.
+
+inputs:
+  github_token:
+    description: GitHub token (to fetch the diff)
+    required: true
+
+outputs:
+  only_markdown_changes:
+    description: 'True if every changed file has the .md extension'
+    value: ${{ steps.check.outputs.only_markdown_changes }}
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: 'recursive'
+        fetch-depth: 0
+
+    - name: List changed files
+      id: diff
+      shell: bash
+      run: |
+        BASE=$(git merge-base origin/${{ github.event.pull_request.base.ref }} HEAD)
+        git diff --name-only $BASE HEAD > changed.txt
+        echo "files=$(paste -sd ',' changed.txt)" >> $GITHUB_OUTPUT
+
+    - name: Determine if only .md files changed
+      id: check
+      shell: bash
+      run: |
+        nonmd=$(grep -vE '\.md$' changed.txt || true)
+        if [ -z "$nonmd" ]; then
+          echo "only_markdown_changes=true" >> $GITHUB_OUTPUT
+        else
+          echo "only_markdown_changes=false" >> $GITHUB_OUTPUT
+        fi
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -20,8 +20,22 @@ permissions:
   contents: read
 
 jobs:
+  planner:
+    name: Detect docs-only changes
+    runs-on: ubuntu-latest
+    outputs:
+      only_docs: ${{ steps.check.outputs.only_markdown_changes }}
+    steps:
+      - name: Check for markdown-only changes
+        id: check
+        uses: ./.github/actions/only-markdown
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
   check-gen:
     name: Check (gen-check)
+    needs: planner
+    if: ${{ needs.planner.outputs.only_docs != 'true' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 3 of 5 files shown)


rvagg:
@kamuik16 you've got an outdated filecoin-ffi here - I think you branched just at the wrong moment and now have the wrong ffi; rebase on to master and make the ffi change go away and 🤞 your tests should pass

kamuik16:
> @kamuik16 you've got an outdated filecoin-ffi here - I think you branched just at the wrong moment and now have the wrong ffi; rebase on to master and make the ffi change go away and 🤞 your tests should pass

Thanks @rvagg, merged master into this one.

.github/workflows/build.yml
```diff
@@ -20,9 +20,28 @@ permissions:
...
+  planner:
+    name: Detect docs-only changes
+    runs-on: ubuntu-latest
+    outputs:
+      only_docs: ${{ steps.check.outputs.only_markdown_changes }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
+          fetch-depth: 0
```
BigLep, line 32: Why is this needed given `.github/actions/only-markdown/action.yml` is doing the checkout as well?

.github/workflows/build.yml
```diff
@@ -20,9 +20,28 @@ permissions:
   contents: read
 
 jobs:
+  planner:
+    name: Detect docs-only changes
+    runs-on: ubuntu-latest
+    outputs:
+      only_docs: ${{ steps.check.outputs.only_markdown_changes }}
```
BigLep, line 27: Rather than `only_docs`, I wonder if we call this `continue`.  

Would that just lets us say

`if: ${{ needs.planner.outputs.continue }}`

to keep it concise and to keep all the business logic inside the planner job of whether to continue (rather than leaking out this decision to whether the planner says there are only docs changes).

Not strongly held opinion, just a thought.  this can always be adjusted further.

BigLep (commented):
I'll leave to others to give approval.  I wish we didn't need to make checks in every job of of a workflow, but per https://github.com/filecoin-project/lotus/issues/13069 I didn't see any way to accomplish this, so your approach seems right.

.github/workflows/build.yml
kamuik16: If this is not checked out, it shows this error: `Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/lotus/lotus/.github/actions/only-markdown'. Did you forget to run actions/checkout before running your local action?`

kamuik16: I have removed the checkout code from the composite action, thanks!

.github/workflows/build.yml
kamuik16: Done, addressed.

kamuik16:
Hey @BigLep, @rvagg, I've addressed the initial comments, looking forward for your further opinions here.

.github/workflows/build.yml
BigLep: Got it.  And to be clear, it doesn't work to just put the checkout code in the composite action along and remove it from the workflow?  I'm assuming not.

BigLep (commented):
I wish we could reduce the boiler plate a bit more, but this seems like the best that can be done under the constraints.  

Do you have an example PR that tests this to demonstrate that it works?

It seems good to me, but will leave to @rvagg to give the approval/merge.

kamuik16:
> Thanks @kamuik16.
> 
> I wish we could reduce the boiler plate a bit more, but this seems like the best that can be done under the constraints.
> 
> Do you have an example PR that tests this to demonstrate that it works?
> 
> It seems good to me, but will leave to @rvagg to give the approval/merge.

Yeah, this is the best way to achieve this under the constraints. 

> Do you have an example PR that tests this to demonstrate that it works?

For this, this PR itself can be an example (where non .md files changes and build/tests were ran), but I need to think some other way to test only for .md changes.

kamuik16:
> > but I need to think some other way to test only for .md changes.
> 
> If this isn't easy, I wouldn't worry about this. In the worst case, it's not working as desired for markdown only changes and we need to do followups.

I have tested the commands locally, and it works perfectly fine, and as you said, worst case, the build and test runs, if that happens, which I'm pretty sure won't, I'll fix it.

kamuik16:
Hey @rvagg @BigLep, friendly bump to get review on this :)

BigLep (approved):
Approving so this doesn't go more stale and because I think the potential harm here is low.

BigLep:
I'm going to merge this PR since there are fixes that were already in place to fix the failing tests.  Given this PR is coming from a fork, I can't directly update that fork with the fixes.  Rather than have more back and forth, I think this is safe to just merge.

PR #13121, opened 2025-05-10
Return `method not supported` via Gateway when /v2 is absent

### LINKED ISSUE #13119: Unversioned FULLNODE_API_INFO on a client resolving to a v2 endpoint
Unversioned FULLNODE_API_INFO on a client resolving to a v2 endpoint. Should default to v1 still as lite clients (and anyone else) can't do what they need through v2.

masih:
When there is no /v2 api offered by the full-node gracefully fallback to returning `method not supported` via Gateway.

Fixes #13119

### DIFF
--- a/cli/lotus/daemon.go
+++ b/cli/lotus/daemon.go
@@ -386,10 +386,11 @@ var DaemonCmd = &cli.Command{
 
 			gapiv2, closerV2, err := lcli.GetGatewayAPIV2(cctx)
 			if err != nil {
-				return err
+				log.Warnf("Unable to connect to v2 API. Using method not supported for /rpc/v2 in gateway", "err", err)
+				gapiv2 = &v2api.GatewayStruct{ /* Returns "method not supported" for everything */ }
+			} else {
+				defer closerV2()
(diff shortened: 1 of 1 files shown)


cli/lotus/daemon.go
```diff
@@ -386,10 +386,11 @@ var DaemonCmd = &cli.Command{
 
 			gapiv2, closerV2, err := lcli.GetGatewayAPIV2(cctx)
 			if err != nil {
-				return err
+				log.Warnf("Unable to connect to v2 API. Using method not supported for /rpc/v2 in gateway", "err", err)
```
rvagg, line 389:
```suggestion
				log.Warnw("Unable to connect to v2 API. Using method not supported for /rpc/v2 in gateway", "err", err)
```

rvagg (approved):
tested, works well, just needs the log fix and maybe this should also get a CHANGELOG entry so we can ship a release that may have only this fix (can have more, but if we did just this change then it should say something)

rvagg (approved):

PR #13134, opened 2025-05-21
feat(f3): integrate MapReduce cache

Kubuxu:
This PR integrates MapReduce cache. The replace directives are temporary.

The mess in go-state-types integration is non-optimal, but I could not find a better solution.


masih (commented):
The integration here looks straightforward (ec.go) 👍

extern/go-hamt-ipld
```diff
@@ -0,0 +1 @@
+../../go-hamt-ipld
```
masih, line 1: Meant to remove this?

extern/go-state-types
```diff
@@ -0,0 +1 @@
+../../go-state-types
```
masih, line 1: ditto

masih (approved):

PR #13136, opened 2025-05-21
chore: fix broken architecture link in create-miner documentation

VolodymyrBg:
Replaced the outdated reference to architecture-2.md with the correct path to architecture/architecture.md in the documentation/en/create-miner.md file. This resolves a broken link error and ensures users are directed to the current Lotus architecture documentation. No other content was changed.


masih (approved):
Thank you

Issue #12718, opened 2024-11-25
Disable payment channel manager by default


rvagg:
On node startup, even for a light node, the payment channel manager starts up and watches the chain for paych-related messages to react to. Payment channels get minimal (to no) use currently on mainnet so we should be safe to introduce a configuration option to disable the payment channel manager and turn it on by default.

See https://github.com/filecoin-project/lotus/blob/0d1d7c29c21b41de79be67ef4dcedb4f973bf7c2/node/builder_chain.go#L120-L126

rvagg:
Yeah, sorry, bad wording - turn it off by default, stop it humming in the background.

A few of us are also discussing how we might trim the code down around the paych functionality, removing most, or even all of it from Lotus.

ameeetgaikwad:
is anyone working on this?

rvagg:
Nobody is working on it; you're welcome to take a stab at this @ameeetgaikwad, it might be interesting for you dig through and find all the ways that it's plumbed into a running node. You might want to put some debug logging (just a printf would do it) in the payment channels manager to see it running when you run a node (e.g. in an itest), and then you should be able to see that go away when you stop it running. I expect at least the rate limiting tests in the gateway itest to fail when it's disabled because it takes into account the chatty manager that keeps on polling chain state.

@dd45e640b42e6da7da96faee3996ef7c the proposal here is to simply disable something that isn't being used, it's not a critique of payment channels; in fact the critique of payment channels _is_ that they aren't being used. The original design goal was good—we want to enable easy payments and everyone thinks that getting more FIL moving for payments is positive, but it seems that payment channels were too complex, or were simply the wrong fit.

The next iteration of getting payments to be simpler is probably going to be FWS, you can see the background in this video: https://www.youtube.com/watch?v=Yz5-9oP9d0Q and right now there's a ton of work going in to making this a reality. There's even active work on a payments facilitator you can look at: https://github.com/FilOzone/fws-payments

PR #13139, opened 2025-05-24
feat(paych): add EnablePaymentChannelManager config option

sumitvekariya:
## add EnablePaymentChannelManager config option to disable payment channel manager by default

This change introduces a new configuration option EnablePaymentChannelManager to control whether the payment channel manager is started on node startup.

- Added EnablePaymentChannelManager boolean field to FullNode config struct
- Set default value to false (disabled by default) since payment channels get minimal use on mainnet
- Added conditional wiring in node/builder_chain.go to only start payment channel manager when config option is enabled
- Updated config documentation with clear description of the option

This saves resources for light nodes and most full nodes that don't use payment channel functionality, while still allowing users to enable it when needed by setting EnablePaymentChannelManager: true in their config.

## Related Issues
https://github.com/filecoin-project/lotus/issues/12718

## Proposed Changes
- **Config Addition**: New `EnablePaymentChannelManager` boolean field in `FullNode` config struct
- **Default Behavior**: Payment channel manager disabled by default (`false`)
- **Conditional Service Wiring**: Modified dependency injection in `node/builder_chain.go` to only wire payment channel services when enabled
- **Documentation**: Updated config auto-generated documentation
- **Changelog**: Added entry to UNRELEASED section following contribution conventions

## Additional Info
- **Backward Compatibility**: Full backward compatibility maintained - users can enable payment channels by setting `EnablePaymentChannelManager: true`
- **Resource Savings**: Eliminates unnecessary resource usage for nodes that don't use payment channels
- **Testing**: All existing payment channel tests pass, config tests updated with new default

## Checklist

- [x] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
   - default to PreCommit batching
   - default to ProveCommit aggregation
   - remove config options: AggregateCommits, AggregateAboveBaseFee, BatchPreCommitAboveBaseFee
+- feat(paych): add EnablePaymentChannelManager config option to disable payment channel manager by default ([filecoin-project/lotus#13139](https://github.com/filecoin-project/lotus/pull/13139))
 
 # Node v1.33.0 / 2025-05-08
 The Lotus v1.33.0 release introduces experimental v2 APIs with F3 awareness, featuring a new TipSet selection mechanism that significantly enhances how applications interact with the Filecoin blockchain. This release candidate also adds F3-aware Ethereum APIs via the /v2 endpoint.  All of the /v2 APIs implement intelligent fallback mechanisms between F3 and Expected Consensus and are exposed through the Lotus Gateway.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
   - default to PreCommit batching
   - default to ProveCommit aggregation
   - remove config options: AggregateCommits, AggregateAboveBaseFee, BatchPreCommitAboveBaseFee
-- feat(paych): add EnablePaymentChannelManager config option to disable payment channel manager by default ([filecoin-project/lotus#13139](https://github.com/filecoin-project/lotus/pull/13139))
+- feat(paych): add EnablePaymentChannelManager config option and disable payment channel manager by default ([filecoin-project/lotus#13139](https://github.com/filecoin-project/lotus/pull/13139))
 
 # Node v1.33.0 / 2025-05-08
 The Lotus v1.33.0 release introduces experimental v2 APIs with F3 awareness, featuring a new TipSet selection mechanism that significantly enhances how applications interact with the Filecoin blockchain. This release candidate also adds F3-aware Ethereum APIs via the /v2 endpoint.  All of the /v2 APIs implement intelligent fallback mechanisms between F3 and Expected Consensus and are exposed through the Lotus Gateway.
--- a/node/builder_chain.go
+++ b/node/builder_chain.go
@@ -118,13 +118,6 @@ var ChainNode = Options(
 	Override(new(wallet.Default), From(new(*wallet.LocalWallet))),
 	Override(new(api.Wallet), From(new(wallet.MultiWallet))),
 
-	// Service: Payment channels
-	Override(new(paychmgr.PaychAPI), From(new(modules.PaychAPI))),
-	Override(new(*paychmgr.Store), modules.NewPaychStore),
-	Override(new(*paychmgr.Manager), modules.NewManager),
-	Override(HandlePaymentChannelManagerKey, modules.HandlePaychManager),
-	Override(SettlePaymentChannelsKey, settler.SettlePaymentChannels),
-
 	// Markets (storage)
 	Override(new(*market.FundManager), market.NewFundManager),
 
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 > * [CHANGELOG_1.2x.md](./documentation/changelog/CHANGELOG_1.2x.md) - v1.20.0 to v1.29.2
```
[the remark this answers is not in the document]
sumitvekariya: agreed, removed all the 3 changes below line 11

rvagg:
Odd DI errors assembling lotus:

```
        	            	    missing type:
        	            	    	- *paychmgr.Manager (did you mean to Provide it?)
```

In theory, `paychmgr.Manager` shouldn't be needed with that chunk of code moved into a conditional in builder_chain.go, only this: `Override(new(*paychmgr.Manager), modules.NewManager)` and this: `Override(HandlePaymentChannelManagerKey, modules.HandlePaychManager)` should be caring about it as far as I can tell. So there must be some downstream dependency that depends on it indirectly. I'm just not seeing it. This will need some investigation and experimentation. @sumitvekariya you're welcome to tinker further with this to try and get to the bottom of it. The uber/fx framework (for DI / Direct Injection) we use in builder_chain.go is basically defining how to make the components and then it's up to the DI framework then uses that knowledge to piece together what it needs for a final Lotus instance. The error is suggesting that something still needs a `*paychmgr.Manager` but we've branched out the part where we tell it how to make one. So we need to figure out who needs it still.

sumitvekariya:
> Odd DI errors assembling lotus:
> 
> ```
>         	            	    missing type:
>         	            	    	- *paychmgr.Manager (did you mean to Provide it?)
> ```
> 
> In theory, `paychmgr.Manager` shouldn't be needed with that chunk of code moved into a conditional in builder_chain.go, only this: `Override(new(*paychmgr.Manager), modules.NewManager)` and this: `Override(HandlePaymentChannelManagerKey, modules.HandlePaychManager)` should be caring about it as far as I can tell. So there must be some downstream dependency that depends on it indirectly. I'm just not seeing it. This will need some investigation and experimentation. @sumitvekariya you're welcome to tinker further with this to try and get to the bottom of it. The uber/fx framework (for DI / Direct Injection) we use in builder_chain.go is basically defining how to make the components and then it's up to the DI framework then uses that knowledge to piece together what it needs for a final Lotus instance. The error is suggesting that something still needs a `*paychmgr.Manager` but we've branched out the part where we tell it how to make one. So we need to figure out who needs it still.

Thanks for the guidance @rvagg! I found and fixed the issue. The problem was indeed a downstream dependency - the DI framework was having trouble with the field ordering in the config struct.

## Solution

The fix was **reordering the `EnablePaymentChannelManager` field** in the `FullNode` config struct. Moving it from the end to an earlier position (after `Chainstore`, before `Fevm`) resolved the DI issue.

## Status
✅ Build now succeeds - no more missing `*paychmgr.Manager` error  
✅ `make gen` passes  
✅ Payment channels properly disabled when `EnablePaymentChannelManager = false`

The solution was simpler than expected - no changes to `PaychAPI` were needed, just correcting the field positioning in the config struct.

rvagg:
No, sadly, ordering in config isn't going to make a difference since it's loaded entirely first and then used later.

```
        	Error Trace:	/home/runner/work/lotus/lotus/itests/kit/ensemble.go:498
        	            				/home/runner/work/lotus/lotus/itests/daily_fees_test.go:123
        	            	starting node:
        	            	    github.com/filecoin-project/lotus/node.New
        	            	        /home/runner/work/lotus/lotus/node/builder.go:371
        	            	  - missing dependencies for function "reflect".makeFuncStub
        	            	    	/opt/hostedtoolcache/go/1.23.7/x64/src/reflect/asm_amd64.s:28:
        	            	    missing type:
        	            	    	- *paychmgr.Manager (did you mean to Provide it?)
```

rvagg:
OK, I figured it out. The real problem is that our `FullNodeAPI` which implements all our external JSONRPC interface (the interface in api_full.go) requires a `PaychAPI` (confusingly there are multiple things with this name!) which implements those methods. Now by taking away a thing that implements those methods, `FullNodeAPI` can't assemble in a way that satisfies its interface because we don't make a `paych.PaychAPI` available for the DI system to assemble.

So this is what I've done:

* Turned this `PaychAPI` into an interface
* Switched the old `PaychAPI` to `PaychImpl`
* Introduced a new `DisabledPaychAPI` that returns errors for all of them
* Loaded in `DisabledPaychAPI` when `EnablePaymentChannelManager` is turned off
* Moved the `EnablePaymentChannelManager` into a new sub-section of the configuration named `PaymentChannels` so it's a bit neater than just a new top-level option

I also rebased on top of the current master and squashed all of the existing commits into one.

rvagg:
I've renamed the duplicate `PaychAPI` types in here so there's less confusion, if I don't do it now it may never get done! Most of this code is 5 years old and hasn't been touched since then.

rvagg (approved):
I'm going to have to ask for another review of this from someone else since most of the changeset is now mine. But I think it's good to go, and it's in 3 clear commits that I think can be rebase-merged as is.

node/builder_chain.go
```diff
@@ -361,6 +355,18 @@ func ConfigFullNode(c interface{}) Option {
 				Override(InitChainIndexerKey, modules.InitChainIndexer(cfg.ChainIndexer)),
 			),
 		),
+
+		If(cfg.PaymentChannels.EnablePaymentChannelManager,
+			Override(new(paychmgr.ManagerNodeAPI), From(new(modules.PaychManagerNodeAPI))),
```
ZenGround0, line 360: I'm seeing this is a bit different than the removal up above.  Is this part of an internal refactor?

ZenGround0: Looks like paychmgr.ManagerNodeAPI added

ZenGround0: and API instantiation comes thorugh PaychImpl now

ZenGround0: Ok nice explained here: https://github.com/filecoin-project/lotus/pull/13139#issuecomment-3003200684

ZenGround0 (approved):
I love this. Very clean and a great simplification of the lotus process.  Thanks all for getting this done.

PR #13147, opened 2025-05-29
feat(gateway): add per-request logging to gateway with --request-logging

rvagg:
Logs POST bodies too, so probably not ideal for environments where performance is critical, or where you think there might be data leakage (although the gateway is intended for non-sensitive traffic).

This was just a private branch feature but I found it super useful for debugging and decided there's a good chance I'm going to want this again. I also tried logging to stdout, but having it interleaved with error logs ended up being more useful. Unfortunately the JSON bodies get escaped to be reproduced as JSON output, but it's not terrible.

Another alternative I considered was providing a filename, so it ends up a bit like a classic HTTP server log, `--request-logging=access.log`, but it goes a bit further than HTTP server logging with the request body capture so it may be best to not give the impression that this is the right thing to do.

```json
{"level":"info","ts":"2025-05-28T22:58:32.768+1000","logger":"gateway","caller":"gateway/handler.go:380","msg":"request","remote_ip":"192.168.1.12","method":"POST","url":"/rpc/v1","body":"{\"id\":\"0f6d72ba-f240-4ef9-83c8-9ecc42f5a635\",\"jsonrpc\":\"2.0\",\"method\":\"eth_getBalance\",\"params\":[\"0x0542c82a785fab8b53c7c59aaad7093d87265f3a\",\"0x29414a\"]}"}
{"level":"info","ts":"2025-05-28T22:58:32.769+1000","logger":"gateway","caller":"gateway/handler.go:380","msg":"request","remote_ip":"192.168.1.12","method":"POST","url":"/rpc/v1","body":"{\"id\":\"e764c4b6-234c-44c0-b59c-5b75407488fe\",\"jsonrpc\":\"2.0\",\"method\":\"eth_call\",\"params\":[{\"to\":\"0x0e690d3e60b0576d01352ab03b258115eb84a047\",\"data\":\"0x01ffc9a780
ac58cd00000000000000000000000000000000000000000000000000000000\"},\"0x29414a\"]}"}                                                                                                          {"level":"warn","ts":"2025-05-28T22:58:32.773+1000","logger":"rpc","caller":"go-jsonrpc@v0.7.0/handler.go:421","msg":"error in RPC call to 'eth_call': message execution failed (exit=[33],
revert reason=[message failed with backtrace:\n00: f0162143 (method 3844450837) -- contract reverted at 86 (33)\n01: f0162143 (method 6) -- contract reverted at 349 (33)\n (RetCode=33)], vm error=[none])"}
```

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -1,7 +1,9 @@
 package gateway
 
 import (
+	"bytes"
 	"context"
+	"io"
 	"net"
 	"net/http"
 	"strings"
@@ -42,6 +44,7 @@ type ShutdownHandler interface {
 var _ ShutdownHandler = (*statefulCallHandler)(nil)
 var _ ShutdownHandler = (*RateLimitHandler)(nil)
 var _ ShutdownHandler = (*CORSHandler)(nil)
+var _ ShutdownHandler = (*LoggingHandler)(nil)
 
 // handlerOptions holds the options for the Handler function.
 type handlerOptions struct {
@@ -49,6 +52,7 @@ type handlerOptions struct {
 	perHostConnectionsPerMinute int
 	jsonrpcServerOptions        []jsonrpc.ServerOption
 	enableCORS                  bool
+	enableRequestLogging        bool
 }
 
 // HandlerOption is a functional option for configuring the Handler.
@@ -87,6 +91,13 @@ func WithJsonrpcServerOptions(options ...jsonrpc.ServerOption) HandlerOption {
 func WithCORS(enable bool) HandlerOption {
 	return func(opts *handlerOptions) {
 		opts.enableCORS = enable
+	}
+}
+
+// WithRequestLogging sets whether to enable request logging.
+func WithRequestLogging(enable bool) HandlerOption {
+	return func(opts *handlerOptions) {
+		opts.enableRequestLogging = enable
 	}
 }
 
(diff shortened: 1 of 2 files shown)


rjan90 (approved):

BigLep:
@rvagg : are you able to land this week?

rvagg:
_Maybe_

rvagg:
Rebased and some of the minor suggestions addressed 🤞

gateway/handler.go
```diff
@@ -335,6 +351,61 @@ func (h *CORSHandler) Shutdown(ctx context.Context) error {
...
+			logFields = append(logFields, "body", bodyStr)
+			// Reconstruct the body for downstream handlers: combine what we read and the rest
+			rest := io.MultiReader(bytes.NewReader(buf), r.Body)
+			r.Body = io.NopCloser(rest)
+		}
+	}
+
+	log.Infow("request", logFields...)
+
+	h.next.ServeHTTP(w, r)
```
masih, line 393: What comes `next` after this handler? I don't think the next handler will have access to the full body since the first 1024 bytes are read here for POST requests.

rvagg: it'll be the go-jsonrpc handler that's `next`; above I replace `r.Body` with a `MultiReader` that replaces the 1024 bytes; do you think that's not going to be adequate? or do you have a suggestion for how to handle this more elegantly?

masih: I missed the multireader apologies. That would work 👍 

Not sure if it is more elegant but another pattern I have seen is to read the body into context with a well known key, which is then looked up by convention by downstream handlers. I don't think it is useful here since the chain of handlers are small and fairly simple.

masih (approved):

Issue #13116, opened 2025-05-09
Flaky: itest TestAPIV2_ThroughRPC


rvagg:
e.g. https://github.com/filecoin-project/lotus/actions/runs/14933637490/job/41955873252?pr=13115

I'm pretty sure this is just a race where the chain moves on while observing it. I think the `api_v2_test.go` tests need a variant of the `stableExecute` in `eth_api_f3_test.go` to ensure that a test run doesn't straddle tipsets: https://github.com/filecoin-project/lotus/blob/54bd2c74d6aac95debef2ec33c1615f8fd40c07d/itests/eth_api_f3_test.go#L826-L849

rjan90:
Seems like this is consistently failing in the most recent PRs merged now: 
- https://github.com/filecoin-project/lotus/actions/runs/15056544047/job/42323604174?pr=13130
- https://github.com/filecoin-project/lotus/actions/runs/14999165784/job/42141250338?pr=13127
- https://github.com/filecoin-project/lotus/actions/runs/15021696488/job/42212260189?pr=13126

PR #13148, opened 2025-05-29
fix(test): address flaky TestAPIV2_ThroughRPC

rvagg:
Closes: #13116

This was mostly AI, needs some clearer review than I can afford right at this moment, but seems good on the surface. It also should have that job run in CI a bunch of times to confirm that it's actually fixed.


rvagg:
Rebased on top of https://github.com/filecoin-project/lotus/pull/13149 so I'd like for that one to land before this one so it'll be removed from here.

masih (commented):
👍

PR #13165, opened 2025-06-09
[WIP] ci: execute some workflows conditionally

galargh:
## Checklist

- [ ] CI is green


.github/workflows/build.yml
```diff
@@ -2,10 +2,14 @@ name: Build
...
 on:
   pull_request:
+    paths-ignore:
+      - '**/*.md'
   push:
     branches:
       - master
       - release/*
+    paths-ignore:
+      - '**/*.md'
```
galargh: I would consider including this filter only on `pull_request` event. It's nice to always do a full build on the default branch as then you can more easily go back through history to find failure points.

.github/workflows/check.yml
```diff
@@ -20,26 +20,21 @@ permissions:
   contents: read
 
 jobs:
-  planner:
-    name: Detect docs-only changes
+  pre-check:
```
galargh, line 23: This should be executed on `pull_request` events (and maybe `push` events) only. When triggering the workflow on `workflow_dispatch`, we always want full checks.

.github/workflows/check.yml
```diff
@@ -93,3 +88,17 @@ jobs:
       - uses: ./.github/actions/install-go
       - run: go mod tidy -v
       - run: git diff --exit-code
+  post-check:
```
galargh, line 94: This is needed because some branch protection rules mark check jobs as required. Before merging, we should replace the individual check jobs in those rules with this aggregate post-check check.

.github/workflows/build.yml
BigLep: Got it - sounds good.

.github/workflows/check.yml
```diff
@@ -20,26 +20,23 @@ permissions:
...
-
+          filters: |
+            files:
+              - '!**/*.md'
   check-gen:
     name: Check (gen-check)
-    needs: planner
-    if: ${{ needs.planner.outputs.continue }}
+    needs: [pre-check]
+    if: ${{ needs.pre-check.outputs.check == 'true' }}
```
BigLep: `check == 'true'` reads weird to me for two reasons:

1. Rather than `needs.pre-check.outputs.check == 'true'` can we just do `needs.pre-check.outputs.check`?
2. "check" isn't very descriptive.  That is why we arrived on `continue` which I can understand isn't great either.  Maybe `continue-with-steps`?

BigLep (approved):

.github/workflows/check.yml
BigLep: Updated version reads better.  Resovling.

PR #13174, opened 2025-06-12
feat(f3): migrate f3 datastore to separate datastore

Kubuxu:
Migrate the f3 datastore from being namespaced in the metadata metadata
datastore to having its own dedicated datastore.

Depends on https://github.com/filecoin-project/lotus/pull/13144

### DIFF
--- a/chain/lf3/f3.go
+++ b/chain/lf3/f3.go
@@ -1,13 +1,17 @@
 package lf3
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/autobatch"
 	"github.com/ipfs/go-datastore/namespace"
+	"github.com/ipfs/go-datastore/query"
 	logging "github.com/ipfs/go-log/v2"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/host"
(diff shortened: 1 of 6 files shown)


chain/lf3/f3.go
```diff
@@ -56,32 +60,119 @@ type F3 struct {
...
+	Wallet        api.Wallet
+	Config        *Config
+	LockedRepo    repo.LockedRepo
 
 	Net api.Net
 }
 
 var log = logging.Logger("f3")
 
+var migrationKey = datastore.NewKey("/migration")
```
masih: Use a key with a less chance of accidental conflict with a pre-existing one? e.g. add `f3` somewhere in there?
```suggestion
var migrationKey = datastore.NewKey("/f3-migration/1")
```
```suggestion
var migrationKey = datastore.NewKey("/migration")
```

chain/lf3/f3.go
```diff
@@ -56,32 +60,119 @@ type F3 struct {
...
+	qr, err := source.Query(ctx, query.Query{})
+	if err != nil {
+		return xerrors.Errorf("starting source wildcard query: %w", err)
+	}
+
+	// batch size of 2000, at the time of writing, F3 datastore has 150,000 keys taking ~170MiB
+	// meaning that a batch of 2000keys would be ~2MiB of memory
+	batch := autobatch.NewAutoBatching(target, 2000)
+	var numMigrated int
+	for {
```
masih:
```suggestion
	for ctx.Err() == nil {
```

chain/lf3/f3.go
```diff
@@ -56,32 +60,119 @@ type F3 struct {
...
+	}
+
+	log.Infow("starting migration of f3 datastore", "tag", migrationVal)
+	qr, err := source.Query(ctx, query.Query{})
+	if err != nil {
+		return xerrors.Errorf("starting source wildcard query: %w", err)
+	}
+
+	// batch size of 2000, at the time of writing, F3 datastore has 150,000 keys taking ~170MiB
+	// meaning that a batch of 2000keys would be ~2MiB of memory
```
masih:
```suggestion
	// meaning that a batch of 2000 keys would be ~2MiB of memory
```

chain/lf3/f3.go
```diff
@@ -56,32 +60,119 @@ type F3 struct {
...
+	batch := autobatch.NewAutoBatching(target, 2000)
```
masih, line 131: Questions: 
* How long would this take?
* what's the distribution of value sizes in source? relatively flat I imagine?

chain/lf3/f3.go
```diff
@@ -56,32 +60,119 @@ type F3 struct {
...
+	var numMigrated int
+	for {
+		res, ok := qr.NextSync()
+		if !ok {
+			break
+		}
+		if err := batch.Put(ctx, datastore.NewKey(res.Key), res.Value); err != nil {
+			_ = qr.Close()
```
masih, line 139: would it make sense to put this in a `defer` statement?

masih (approved):
LGTM, left a couple of question around operational dynamics of migration.

chain/lf3/f3.go
Kubuxu: The metadata datastore is prefixed, and the new one is empty. The only conflict could be f3 itself as it takes ownership of this datastore.

chain/lf3/f3.go
Kubuxu: It takes 1-2s
I assume quite uniform

chain/lf3/f3.go
Kubuxu: I initially did, but wanted a checked and escalated Close on the clean path.

PR #13176, opened 2025-06-13
chore: bump the pubsub validation queue length to 4096

masih:
Increase the pubsub validation queue length to 4096, to accommodate the increased rate of message exchange as a result of F3. Note that there's
 no topic-specific option for this setting, hence the bump at top-level
 pubsub instantiation.

 Relates to: https://github.com/libp2p/go-libp2p-pubsub/issues/621

### DIFF
--- a/node/modules/lp2p/pubsub.go
+++ b/node/modules/lp2p/pubsub.go
@@ -252,6 +252,12 @@ func GossipSub(in GossipIn) (service *pubsub.PubSub, err error) {
 		// Gossipsubv1.1 configuration
 		pubsub.WithFloodPublish(true),
 		pubsub.WithMessageIdFn(HashMsgId),
+		// Bump the validation queue to accommodate the increase in gossipsub message
+		// exchange rate as a result of f3. The size of 4096 should offer enough headroom
+		// for slower F3 validation while avoiding: 1) avoid excessive memory usage, 2)
+		// dropped consensus related messages and 3) cascading effect among other topics
+		// since this config isn't topic-specific.
+		pubsub.WithValidateQueueSize(4096),
 		pubsub.WithPeerScore(
 			&pubsub.PeerScoreParams{
 				AppSpecificScore: func(p peer.ID) float64 {


masih:
Many thanks to @dennis-tra at Probelab for pointing this out.

masih:
Marked as draft to look into max message size (default 1MiB in go-libp2p-pubsub) unconfigured in Lotus.

masih:
We know from Dennis that 2048 would be effective but not as much as we ideally want. Not sure what size would be a good compromise.

dennis-tra:
Great @masih! I know from our Hermes deployment that 2048 still results in some discarded messages, so bumping to 4096 sounds reasonable to make sure that we don't miss any.

Kubuxu:
@dennis-tra FYI, we optimised the validator (it could take a while to process the first message of an instance) in more recent go-f3 commits and lotus (go-f3 is not yet merged), so we expect maybe that to help.

dennis-tra:
I'm at `github.com/filecoin-project/go-f3 v0.8.4`. I just checked and the latest version is v0.8.6. However, I copied the validation bits to [Hermes](https://github.com/probe-lab/hermes):

- granite: https://github.com/probe-lab/hermes/blob/803213c8273756afa8038521d836ea0ed49dedce/fil/pubsub.go#L213
- chainexchange: https://github.com/probe-lab/hermes/blob/803213c8273756afa8038521d836ea0ed49dedce/fil/pubsub.go#L259

I can't remember exactly but I think the validation functions weren't exposed back when I worked on that.

> go-f3 is not yet merged

I can't see an open PR, so what's the latest state regarding the validation logic that I should look into?

Kubuxu:
Here is Lotus with latest go-f3 https://github.com/filecoin-project/lotus/pull/13144
I'm going to bump it to 0.8.6 today and merge.

masih:
@Kubuxu I ran a bunch of experiments on our test node and it seems like 256 is the smallest queue size with almost no impact on validation queue getting full. observed a dozen of such instances while running the test for ~30 minutes. I think this is an acceptable level of message loss and a massive improvement from the default of 32.

BigLep:
For completeness if maintainers want the full context, this started out of an internal discussion in https://filecoinproject.slack.com/archives/C08715YFQH5/p1749815281170279

Issue #13167, opened 2025-06-09
API v2 request to Filecoin.GetChainTipset tipset with safe selector results in negative height.


parthshah1:
### Checklist

- [x] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [x] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [x] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [x] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [ ] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [x] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
Daemon: lotus version 1.33.1-dev+2k+git.fb24904a9.dirty (The patch is for local drand nodes).
```

### Repro Steps

This behavior was seen while running Antithesis simulation, where we called Filecoin.ChainGetTipSet with selector safe, it returns these logs:
2025-06-09T00:39:17.328Z	WARN	rpc	go-jsonrpc@v0.8.0/handler.go:425	error in RPC call to 'Filecoin.ChainGetTipSet': getting latest finalized tipset:
    github.com/filecoin-project/lotus/node/impl/full.(*ChainModuleV2).getLatestSafeTipSet
        node/impl/full/chain_v2.go:77
  - height -819 is negative:
    github.com/filecoin-project/lotus/chain/store.(*ChainStore).GetTipsetByHeight
        chain/store/store.go:1266

Because safeHeight is defined by height - policy.Chainfinality, which is I believe 900. So it returns height is negative. 

### Describe the Bug

JsonRPC request to v2 API returns negative if called before 900 epochs aka policy.ChainFinality. 

### Logging Information

```text
2025-06-09T00:39:17.328Z	WARN	rpc	go-jsonrpc@v0.8.0/handler.go:425	error in RPC call to 'Filecoin.ChainGetTipSet': getting latest finalized tipset:
[...]

PR #13180, opened 2025-06-17
fix: assert that EC finalized tipset height is never negative in v2 APIs

masih:
Fix an edge case where if the head epoch is below the chain finality policy (defaulting to 900) EC finalized head height would become negative.

Fixes: #13167


rvagg:
oh, v2 tests are failing with this, `TestAPIV2_ThroughRPC/ChainGetTipSet/height_with_anchor_to_latest`

PR #13189, opened 2025-06-30
feat(spcli): correctly handle the batch logic of `lotus-miner actor settle-deal`

beck-8:
## Related Issues

https://filecoinproject.slack.com/archives/C03CKDLEWG1/p1751257247291959?thread_ts=1745286086.505189&cid=C03CKDLEWG1
## Checklist

- [ ] CI is green


BigLep:
@beck-8 : looks like there's a small lint error to fix.

beck-8:
I originally wanted to change it on the original foundation (lotus-miner). I don't think anyone will need to check the details of the message. Now everyone who uses this tool urgently wants to take out the locked amount.

However, considering that someone may really ask to check the payment, I changed the logic to lotus-shed. This is my idea. Do you have any suggestions?

beck-8:
Or should I ask if we really need this part of logic? Or put this small part elsewhere in the future?
https://github.com/filecoin-project/lotus/blob/cfb4c8ec5578c721001fc1eb5a7e4e11988bf22f/cli/spcli/actor.go#L141-L171

beck-8:
I got another method that is very friendly to human beings. However, it may not be backward compatible with the code.
https://github.com/beck-8/lotus/blob/b34e8928fbf7a9dbde0ffb4fba188d553c4de688/cmd/lotus-shed/market.go#L529-L543

rvagg:
Actually @beck-8, hold off from putting in all that work. I think maybe you should just explain, both briefly in the CHANGELOG and a little more verbose in the `Description` of the new command what it does and why it's useful instead of the `lotus-miner` variant. I think maybe you've got a very narrow use-case here.

The alternative is to just go ahead and adapt your use-case to the `lotus-miner` one, make one command but make it do what you want to do on top of its existing flexibility -- or make a judgement about whether its existing flexibility is even very useful. I see complaints about it, if you just want to fix it then we can get additional SP feedback before doing so but we shouldn't feel like these things are not open to changing.

beck-8:
In fact, no matter which one, SP will still complain, because StateMarketDeals is as high as 29G, and the premise of executing settlement is to wait for a few minutes to pull this data.

beck-8:
I will give more explanations later. I know that the use case is relatively narrow, but I have no other good way to help SP. Some experienced people will use `lotus-miner sectors status` to extract deal, but it only applies to certain environments of some people.

As we discussed at that time, I put forward my ideas and cases. But you don't have to merge this PR. You can close it and restructure it, but you should know my idea.

beck-8:
1. Quickly obtain the relationship between sectors and deals
2. Reasonably handle batches and send messages.

I suggest exporting this method or transferring it to a reasonable place, and then `GetMinerAllDeals` -> `GetMarketDealIDs`
https://github.com/sumitvekariya/lotus/blob/7f2a7b7179dd75b5e379ac78e2af1f9caca04d58/cli/state.go#L1752

BigLep:
My 2025-07-10 understanding of where this PR is at:
* Related conversation: https://filecoinproject.slack.com/archives/C03CKDLEWG1/p1745286086505189
* @beck-8's core goal is to make this API faster.  It's currently too slow because it has to load and transfer all the deals.
* Maintainers want to avoid a duplicated `lotus-shed settle-deal` when there’s a `lotus-miner settle-deal`
* @beck-8 and @tediou5 will work to land this so it has a faster API and that it's leveraged in `lotus-miner settle-deal`
* For now while it's still being worked on, it's in draft.

beck-8:
I know this may not meet some norms. I just want to express the effect I want.

I may merge the command into lotus-miner later (still some logic will be removed)

beck-8:
Test completed, currently meets my expectations.
```bash
./lotus-miner --actor f01703655 actor settle-deal --really-do-it --max-deals 50 --from f3tdlvj4lzhdnes5qzg2w6cghsotka7kjeaiho3qpffctyqlncgls5oxlnclgk4mevivj3acvx46sumyg5ng5a
There are a total of 3335 deals, and about 66 messages will be sent out quickly.
Requested deal settlement in message: bafy2bzacea2j2oscgzjkcrc4pcyh4whja346tf2fkqmp7eqcajvpvegciauqg
Requested deal settlement in message: bafy2bzaceausxqeos6e5qglqqwdfn7yq3pu74opupun4muy6msovmjutdcyw2
Requested deal settlement in message: bafy2bzacecro7en6ij2zik6uvmih6kkukdfbi2p6rkz3stt3yok5xcmuhuyz4
Requested deal settlement in message: bafy2bzaceddzbutjwhg53n3ard7ktomuts4fwr5jevv5slhnasxnev3grikwy
Requested deal settlement in message: bafy2bzaced7by4ol2lnwqs3aabgkcuitad22xpxd46xttxn6dyqnjrdjly4ck
Requested deal settlement in message: bafy2bzaceb7p4gug4vcw23g4i7iogykr3qe4biibjfu37zur7in25teiymppa
Requested deal settlement in message: bafy2bzacebpec5kfuk2ofnq66qlbl4jd6pmtbfip4xuzk5g64ekxg3ivkcjgk
Requested deal settlement in message: bafy2bzacede35a4owjk7eark63vvucibklz2yfurs6p4aeclkqquujfl4cy3a
Requested deal settlement in message: bafy2bzacebafpufpphriwrh7zrxzqcliwcsefbdw5bfzbhhfivvw5kubafnfo
Requested deal settlement in message: bafy2bzaced7ttpkttehptrylyrwgemj7qeuam2iucfs5l2tldgzkordzu7tye
Requested deal settlement in message: bafy2bzacea5puvzm6ncj6gvzq4nblpsgtbbp2nwdbm5ug66j4sxjxktbqhlq4
Requested deal settlement in message: bafy2bzacedxpsykecnfhibggweaa5tphb2pn7oxcv3dg63jzbipvk42bmw5x6
Requested deal settlement in message: bafy2bzaceat3z7hrkocyverij2sre2trp3l7ti7cueaps4mnr5py5hmu666yg
ERROR: mpool push: failed to push message: failed to check balance: not enough funds including pending messages (required: 0.422663959546500762 FIL, balance: 0.3917174615362471 FIL): validation failure
```
---
```go
			&cli.IntFlag{
				Name:  "max-deals",
				Usage: "the maximum number of deals contained in each message",
				Value: 50,
			},
```
`--max-deals 500` `--max-deals 100` out of gas, so set 50

beck-8:
I don't know how to deal with the conflict in `UNRELEASED.md`. If I'm wrong, please help me correct it directly.

rjan90:
@beck-8 I resolved the conflict in the `CHANGELOG.md` for you

cli/spcli/actor.go
```diff
@@ -107,68 +134,71 @@ func ActorDealSettlementCmd(getActor ActorAddressGetter) *cli.Command {
...
+				})
 
-			if wait.Receipt.ExitCode.IsError() {
-				return xerrors.Errorf("Failed to execute withdrawal message %s: %w", wait.Message, wait.Receipt.ExitCode.Error())
 			}
-
-			var settlementReturn markettypes14.SettleDealPaymentsReturn
-			if err = settlementReturn.UnmarshalCBOR(bytes.NewReader(wait.Receipt.Return)); err != nil {
+			err = g.Wait()
+			if err != nil {
```
tediou5: I think, instead of directly changing the previous behavior, could we control this with a flag here?

tediou5 (changes requested):
Thank you for your pr.

From my perspective, this PR contains two parts:
1.  An optimization for the `GetDeals` method.
2.  Batch submission for `settle-deal`.

Regarding the first part, I'm in full agreement. This will greatly improve the quality of life for SPs.

However, for the second part, I have some concerns:
1.  I might be lacking some background knowledge on this, but is it necessary to batch `settle-deal` transactions? This shouldn't be a particularly heavy transaction.
2.  The logic after the transaction is modified here to exit immediately without waiting. While I can understand this behavior, it's a breaking change.

I think, would it be better to preserve the original behavior of waiting for the packaging to complete, but provide a flag to skip it directly?

Additionally, I suggest splitting this PR into two parts. This PR should only contain the `GetDeals` optimization, which we can merge directly. For the batching part, please submit it as a new, separate PR where we can have a more detailed discussion and work on the implementation.

cli/state.go
```diff
@@ -1770,7 +1771,7 @@ var StateSysActorCIDsCmd = &cli.Command{
 }
 
 // GetMarketDealIDs retrieves deal IDs for a sector from the market actor's ProviderSectors HAMT
-func GetMarketDealIDs(ctx context.Context, api v0api.FullNode, maddr address.Address, sid abi.SectorNumber, tsKey types.TipSetKey) ([]abi.DealID, error) {
+func GetMarketDealIDs(ctx context.Context, api v1api.FullNode, maddr address.Address, sid abi.SectorNumber, tsKey types.TipSetKey) ([]abi.DealID, error) {
```
tediou5: You can use WrapperV1Full to wrap v1api. WrapperV1Full implements v0api.FullNode, this way you won't have to modify this.

cli/spcli/actor.go
```diff
@@ -107,68 +139,92 @@ func ActorDealSettlementCmd(getActor ActorAddressGetter) *cli.Command {
...
-			}
+			sort.Slice(dealIDs, func(i, j int) bool {
+				return dealIDs[i] < dealIDs[j]
+			})
+			dealChucks := lo.Chunk(dealIDs, cctx.Int("max-deals"))
 
-			res := smsg.Cid()
+			for _, deals := range dealChucks {
+				dealParams := bitfield.NewFromSet(deals)
+				var err error
```
tediou5: This seems unnecessary.

cli/spcli/actor.go
```diff
@@ -107,68 +139,92 @@ func ActorDealSettlementCmd(getActor ActorAddressGetter) *cli.Command {
...
-				Method: marketactor.Methods.SettleDealPaymentsExported,
-				Params: params,
-			}, nil)
-			if err != nil {
-				return err
-			}
+			sort.Slice(dealIDs, func(i, j int) bool {
+				return dealIDs[i] < dealIDs[j]
+			})
+			dealChucks := lo.Chunk(dealIDs, cctx.Int("max-deals"))
```
tediou5, line 170: if skip-wait-msg, I think we could print a helpful message here or below.

cli/spcli/actor.go
beck-8: I don't understand this.

cli/spcli/actor.go
```diff
@@ -107,68 +140,94 @@ func ActorDealSettlementCmd(getActor ActorAddressGetter) *cli.Command {
...
+				}
 
-			fmt.Printf("Requested deal settlement in message %s\nwaiting for it to be included in a block..\n", res)
+				smsg, err := api.MpoolPushMessage(ctx, &types.Message{
+					To:     marketactor.Address,
+					From:   fromAddr,
+					Value:  types.NewInt(0),
+					Method: marketactor.Methods.SettleDealPaymentsExported,
+					Params: params,
+				}, nil)
```
tediou5: You can construct batch_msgs first and then push them all at once with a single call to MpoolBatchPushMessage. The function will take care of gas estimation and funds check and add each message to the mempool, which will make your code cleaner.

tediou5:
I've submitted the change to switch to using `MpoolBatchPushMessage`, also combine logs to prevent interleaving caused by concurrency.

beck-8:
It looks fine, very good, take off 🛫

tediou5 (approved):
Looks good to me, thanks!

tediou5:
lets go 🛫!

beck-8:
I want to add a flag to indicate whether the deal is expired. By default, only expired deals will be processed.

tediou5:
> I want to add a flag to indicate whether the deal is expired. By default, only expired deals will be processed.

Makes sense to me, let's open a new PR to add this.

beck-8:
Just mark here: this logic doesn't completely cover all cases. Some deals are SPs but are not included in the `StateMinerSectors` sector.
Therefore, the user needs to manually specify the `dealId` found

https://filecoinproject.slack.com/archives/CR98WERRN/p1763087229904099

PR #13212, opened 2025-07-15
feat(shed): lotus-shed miner fees-inspect improvements

rvagg:
- Proper CSV output
- Add --output to write to a file
- Add resume capability from last processed epoch when using --output
- --tipset only required if not resuming, now means "start"
- --count is optional, will default to chain head
- Always walk from oldest to newest so keeping a file updated is easier
- Nice progress indicator when writing to a file
- Improve efficiency comment for inspectMiner function

This does break the command for existing users, but since that's probably just me I think that's OK. The main breakage is in the ordering of the output and the fact that `--tipset` is that start and we go forward for `--count` or until head.

### DIFF
--- a/cmd/lotus-shed/miner-fees.go
+++ b/cmd/lotus-shed/miner-fees.go
@@ -2,8 +2,13 @@ package main
 
 import (
 	"bytes"
+	"encoding/csv"
 	"fmt"
+	"io"
+	"os"
 	"sort"
+	"strconv"
+	"strings"
 
 	cbor "github.com/ipfs/go-ipld-cbor"
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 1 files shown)


cmd/lotus-shed/miner-fees.go
```diff
@@ -433,6 +453,75 @@ var minerFeesInspect = &cli.Command{
...
+				}
+				reader := csv.NewReader(f)
+				var lastLine []string
+				for {
+					line, err := reader.Read()
+					if err == io.EOF {
+						break
+					}
+					if err != nil {
+						f.Close()
```
rjan90:
```suggestion
					if err != nil {
						_ = f.Close()
```

rjan90 (approved):
> This does break the command for existing users, but since that's probably just me I think that's OK.

That is probably fine. Looks good to me as long as tests pass

rvagg:
This is so much better than my previous workflow .. catch up in progress using an existing data file (which I've had to reformat a bit to make compatible):

```
Last processed epoch: 5058269
Resuming from epoch 5058270
Processing 84176 epochs from 5058270 to 5142445
Processing epoch 5058278 [443/84176] (0%)...
```

PR #13245, opened 2025-08-06
docs: fix broken link

JacksonEi:
https://github.com/filecoin-project/lotus/blob/master/cmd/lotus-miner/init.go - broken
https://github.com/filecoin-project/lotus/blob/master/cli/miner/init.go - new

https://github.com/filecoin-project/lotus/blob/master/build/params_shared.go - broken
https://github.com/filecoin-project/lotus/blob/master/build/params_shared_vals.go - new


documentation/en/create-miner.md
```diff
@@ -55,13 +55,13 @@ The main options that define a miner are the owner and worker addresses associat
...
+At the network level, the message (to be included in a block) is sent through the [message pool](https://filecoin-project.github.io/specs/#systems__filecoin_blockchain__message_pool) to the miners. The message pool is an abstraction over `libp2p`'s [PubSub protocol](https://docs.libp2p.io/reference/glossary/#pubsub), a communication pattern where peers subscribe to a topic (an identifying string) and other peers publish to it (broadcasting to all the subscribed peers it has registered). (Particularly in Lotus we use the [GossipSub](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/README.md) implementation of that protocol.) In this case the topic we publish to is [`/fil/msgs`](https://github.com/filecoin-project/lotus/blob/master/build/params_shared_vals.go), to which the miners will be subscribed and will receive, validate and process the message (see [`HandleIncomingMessages`](https://github.com/filecoin-project/lotus/blob/master/node/modules/services.go) in the [miner node construction](https://github.com/filecoin-project/lotus/blob/master/node/builder.go) for more details).
```
rjan90:
```suggestion
At the network level, the message (to be included in a block) is sent through the [message pool](https://filecoin-project.github.io/specs/#systems__filecoin_blockchain__message_pool) to the miners. The message pool is an abstraction over `libp2p`'s [PubSub protocol](https://docs.libp2p.io/reference/glossary/#pubsub), a communication pattern where peers subscribe to a topic (an identifying string) and other peers publish to it (broadcasting to all the subscribed peers it has registered). (Particularly in Lotus we use the [GossipSub](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/README.md) implementation of that protocol.) In this case the topic we publish to is [`/fil/msgs`](https://github.com/filecoin-project/lotus/blob/master/build/params_shared_funcs.go), to which the miners will be subscribed and will receive, validate and process the message (see [`HandleIncomingMessages`](https://github.com/filecoin-project/lotus/blob/master/node/modules/services.go) in the [miner node construction](https://github.com/filecoin-project/lotus/blob/master/node/builder.go) for more details).
```

This should be pointed to `params_shared_funcs.go` instead of `params_shared_vals.go, because the https://github.com/filecoin-project/lotus/blob/0edb14c29a54f08215a85940ec3b8a3709afac1e/build/params_shared_funcs.go#L18 is what the documentation is referring to.

PR #13274, opened 2025-08-19
fix(api): eth_getCode and eth_getStorageAt operate on current tipset state

wjmelements:
Reviewer @rvagg
Fixes #13247

I use `TipSetState` and `LoadActorRaw` instead of `ParentState` within `LoadActor`, which roughly matches the behavior of `EthCall`.

When looking for other methods that might have this problem in lookup.go, I noticed that `EthGetBalance` seems to already do `TipSetState` and `LoadActorRaw`.

#### Test Plan
Deploy a contract and await the receipt
Once receipt returns non-null, immediately kill lotus-miner
Query daemon with eth_getCode on latest
Before this change, eth_getCode outputs 0x (because ErrActorNotFound)
After this change, eth_getCode outputs the deployed bytecode

Apply a tx on the contract that sets nonzero storage from zero.
Once receipt  returns non-null, immediately kill lotus-miner
Query daemon with eth_getStorageAt on latest
Before this change, eth_getStorageAt outputs zero
After this change, eth_getStorageAt outputs nonzero

#### Changes
* add func `CallOnState` to pass a stateCid parameter to `Call`
* eth_getCode and eth_getStorageAt use state from `TipSetState` instead of `ParentState`, and `LoadActorRaw` instead of `LoadActor`.

PR #13275, opened 2025-08-21
fix(api): eth_getTransactionCount operate on current tipset state

wjmelements:
Reviewer @rvagg
Similar to #13274
#### Test Plan
Run local devnet miner and daemon.
Send transaction and await receipt, then kill miner.
Query eth_getTransactionCount of sender.
Before this change it returns zero.
After this change it returns nonzero
#### Changes
* use LoadActorRaw to specify TipSetState cid

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 # UNRELEASED
 - feat(net): add LOTUS_ENABLE_MESSAGE_FETCH_INSTRUMENTATION=1 to turn on metrics and debugging for local vs bitswap message fetching during block validation ([filecoin-project/lotus#13221](https://github.com/filecoin-project/lotus/pull/13221))
 - chore(docs): mark v0 API as "deprecated" and v1 as "stable" ([filecoin-project/lotus#13264](https://github.com/filecoin-project/lotus/pull/13264))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 # UNRELEASED
 - feat(net): add LOTUS_ENABLE_MESSAGE_FETCH_INSTRUMENTATION=1 to turn on metrics and debugging for local vs bitswap message fetching during block validation ([filecoin-project/lotus#13221](https://github.com/filecoin-project/lotus/pull/13221))
 - chore(docs): mark v0 API as "deprecated" and v1 as "stable" ([filecoin-project/lotus#13264](https://github.com/filecoin-project/lotus/pull/13264))
-- fix(api): `eth_getCode` and `eth_getStorageAt` now return state after the specified block rather than before it ([filecoin-project/lotus#13247](https://github.com/filecoin-project/lotus/issues/13247))
+- fix(api): `eth_getCode`, `eth_getStorageAt`, and `eth_getTransactionCount` now return state after the specified block rather than before it ([filecoin-project/lotus#13247](https://github.com/filecoin-project/lotus/issues/13247))
```
rvagg: Just add a second link to this PR in here.

wjmelements: It looks a little silly because the rest of the line is the same as the previous line, but it seems that this is the pattern because there is one PR per line.

PR #13279, opened 2025-08-25
fix: properly handle all RPC API retry errors

rvagg:
Ref: https://github.com/filecoin-project/curio/pull/605

Discovered in Curio but we have the same problem.

Doing some dedupe while I'm at it.

### DIFF
--- a/cli/util/api.go
+++ b/cli/util/api.go
@@ -242,7 +242,7 @@ func OnSingleNode(ctx context.Context) context.Context {
 // makeNodeProxy creates a proxy that load balances across multiple nodes with automatic retry and failover
 func makeNodeProxy[T any](ins []T, outs []interface{}) {
 	if len(ins) == 0 {
-		log.Error("called with empty node list")
+		log.Error("makeNodeProxy called with empty node list")
 		return
 	}
 
@@ -299,7 +299,7 @@ func makeNodeProxy[T any](ins []T, outs []interface{}) {
 				})
 
 				// If retry failed and we don't have a valid result, construct an error response
-				if rerr != nil && (result == nil || len(result) != field.Type.NumOut()) {
+				if rerr != nil && len(result) != field.Type.NumOut() {
 					log.Errorf("retry failed for %s, err: %s", field.Name, rerr)
 					// Construct a response with zero values and the error
 					result = make([]reflect.Value, field.Type.NumOut())


cli/util/api.go
```diff
@@ -239,14 +239,13 @@ func OnSingleNode(ctx context.Context) context.Context {
 	return context.WithValue(ctx, contextKey("retry-node"), new(atomic.Int32))
 }
 
-func FullNodeProxy[T api.FullNode](ins []T, outstr *api.FullNodeStruct) {
+// makeNodeProxy creates a proxy that load balances across multiple nodes with automatic retry and failover
+func makeNodeProxy[T any](ins []T, outs []interface{}) {
 	if len(ins) == 0 {
-		log.Errorf("FullNodeProxy called with empty node list")
+		log.Error("called with empty node list")
```
wjmelements: I don't think you intended to remove the method name here. Perhaps should be 
```
"makeNodeProxy called with empty node list"
```

cli/util/api.go
```diff
@@ -298,80 +297,33 @@ func FullNodeProxy[T api.FullNode](ins []T, outstr *api.FullNodeStruct) {
...
+				// If retry failed and we don't have a valid result, construct an error response
+				if rerr != nil && (result == nil || len(result) != field.Type.NumOut()) {
+					log.Errorf("retry failed for %s, err: %s", field.Name, rerr)
+					// Construct a response with zero values and the error
+					result = make([]reflect.Value, field.Type.NumOut())
+					for i := 0; i < field.Type.NumOut()-1; i++ {
+						result[i] = reflect.Zero(field.Type.Out(i))
+					}
+					// Set the last element to the error
+					result[field.Type.NumOut()-1] = reflect.ValueOf(rerr)
```
wjmelements, line 310: Why the last element instead of the first element?

Can this fail elsewhere because the error has the wrong typing?

I would expect the error to be in a different field than the result. For example in jsonrpc a success looks like
```
{
    "id": 1,
    "result": "0x0"
}
```
while an error would look like
```
{
    "id": 1,
    "error": "out of gas"
}
```

cli/util/api.go
```diff
@@ -298,80 +297,33 @@ func FullNodeProxy[T api.FullNode](ins []T, outstr *api.FullNodeStruct) {
 					e := result[len(result)-1].Interface().(error)
 					return result, e
 				})
+
+				// If retry failed and we don't have a valid result, construct an error response
+				if rerr != nil && (result == nil || len(result) != field.Type.NumOut()) {
```
wjmelements: I think in `go` that `result == nil` is equivalent to `len(result) == 0` for slice types, so I wonder how this behaves when `field.Type.NumOut()` is `0`.
Is that possible? If it is not possible, then you can remove the nil check. If it is possible, then it seems that the error wouldn't be contained in the response.

cli/util/api.go
rvagg: well, there we re two identical error logs with method names in them, now they are one, and the logger will output this callsite anyway :shrug: I'll put it back

cli/util/api.go
rvagg: Good questions.

Why last? Because we're stuck in Go-land here, not JSONRPC-land, we've done the conversion from JSONRPC to Go reflection execution already over here: https://github.com/filecoin-project/go-jsonrpc/blob/bd8f54a9270e128cca30fb691019e011b02ca2ff/client.go#L553-L582 - the normal error path where the server returns an `"error"` goes through `processResponse` and hits `if resp.Error != nil`, setting the error on our Go function. In that case (which is different to the case we have here because we're not doing JSONRPC, just this funky reflection for multi-backend balancing), we allow a couple of return type signatures, where there may or may not be an error return: https://github.com/filecoin-project/go-jsonrpc/blob/bd8f54a9270e128cca30fb691019e011b02ca2ff/util.go#L32-L59 - but in this case we _know_ that error will always be in that position because we define them in api/api_full.go (v1) and api/v2api/full.go (v2), each one of them has at least an error return, always last. In go-jsonrpc there is also the `processError` case, those come from internal parsing or handling errors (those are the `ErrClient` errors). The errors we're mainly going to encounter here though end up coming from this block: https://github.com/filecoin-project/go-jsonrpc/blob/bd8f54a9270e128cca30fb691019e011b02ca2ff/client.go#L211-L260, some of which are the ones we want to retry on cause they don't indicate something fatal, `RPCConnectionError`, but others are just generally bad and not worth retrying because we'll likely get the same response from whoever we talk to.

I think you'd probably want to spend some time looking in go-jsonrpc which sits at the outside boundary for us, particularly if the upstream JSON-RPC folks are thinking about making changes.

cli/util/api.go
rvagg: the `== nil` isn't strictly necessary here, I was just being explicit in expressing intent because there's both a `nil` and a `[]` version of the result that's possible here

will remove because obviously it doesn't make it clearer

wjmelements: how do we distinguish between nil and [] ?

rvagg: we don't, and we don't _really_ need to, `field.Type.NumOut()` is always going to be >= 1, so if we have `len(result) == 0` then something's gone just as wrong as if `result == nil`; both bad, it's an error, deal with it as an error case with `rerr`, which we know isn't nil.

Issue #13129, opened 2025-05-15
Add support for F3 aware snapshots


BigLep:
## Done Criteria
Snapshot importing and exporting can handle snapshots that are augmented with F3 information that was speced in [FRC-108](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0108.md) and is being designed and worked on in https://github.com/ChainSafe/forest/issues/5627

## Why Important
Nodes starting from a snapshot don't need to rely on the cert exchange protocol to catchup because we expect this will get slower with time.  At least as of 202508, catchup takes around ~8 hours without this.

## User/Customer
Node operators who start from a snapshot.

## Notes
1. This depends on
   1. https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0108.md
   2. https://github.com/filecoin-project/go-f3/issues/480
   3. https://github.com/ChainSafe/forest/issues/56282. 
2. Lotus snapshot export support is needed so that Forest can continue to validate that Lotus and Forest snapshot formats match.  
3. Adopting [FRC-108](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0108.md) is in [scope for nv27](https://github.com/filecoin-project/core-devs/discussions/196).

BigLep:
Slack thread where Forest discussed their implementation and e2e tests: https://filecoinproject.slack.com/archives/C0556MSR945/p1754415315072999 .  Hopefully this can expedite Lotus development.
- https://github.com/ChainSafe/forest/pull/5835
- https://github.com/ChainSafe/forest/pull/5886

BigLep:
Note on the impact of this work: Lotus nodes restoring from a v1 snapshot would need (as of 2025-08-19) to sync 228,254 finality certs.  Certificate Exchange sync happens at the rate of 8-9certs/s. As a result, without v2 snapshot support, a Lotus node being restored by snapshot on 2025-10-01 should expect to take ~8 hours.

Kubuxu:
Sample snapshot car with F3 data: https://forest-snapshots.fra1.cdn.digitaloceanspaces.com/f3/forest_snapshot_calibnet_2025-08-07_height_2908399_f3_555573.forest.car.zst
PR in go-car that introduces BlockReader.NextReader, which allows for reading larger blocks (incrementally) https://github.com/ipld/go-car/pull/603

[...]

PR #13282, opened 2025-08-27
feat: support for snapshot v2 in FRC-0108

tediou5:
## Related Issues

close: #13129

## Additional Info

When exporting the F3Snapshot, I preferred not to introduce a temporary file for F3Data storage, so I've written it directly to memory. A memory footprint of a few hundred MB should be acceptable. However, please feel free to comment if this is a concern, and I can modify this logic later (I'm aware the Forest implementation writes to a temporary file first).

Also, the snapshot export now defaults to V2; otherwise, the API would need to be modified to add a parameter for version specification.

c5625aad724e2549896e2c433dfc49e927cd11e0: intro SnapshotMetadata struct and run make gen
fb02ccbce5bcee470b3c644c686b654fb6aed38f: import and export for snapshotV2, 

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/store/snapshot.go
+++ b/chain/store/snapshot.go
@@ -17,7 +17,6 @@ import (
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/namespace"
-	cbor "github.com/ipfs/go-ipld-cbor"
 	format "github.com/ipfs/go-ipld-format"
 	"github.com/ipld/go-car"
 	carutil "github.com/ipld/go-car/util"
@@ -32,10 +31,9 @@ import (
 	"github.com/filecoin-project/go-state-types/abi"
 
 	bstore "github.com/filecoin-project/lotus/blockstore"
+	"github.com/filecoin-project/lotus/build/buildconstants"
 	"github.com/filecoin-project/lotus/chain/actors/builtin"
-	init_ "github.com/filecoin-project/lotus/chain/actors/builtin/init"
 	"github.com/filecoin-project/lotus/chain/actors/policy"
-	"github.com/filecoin-project/lotus/chain/state"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/filecoin-project/lotus/node/modules/dtypes"
 )
@@ -168,7 +166,7 @@ func (cs *ChainStore) exportToCar(ctx context.Context, ts *types.TipSet, inclRec
 	})
 }
 
-func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (head *types.TipSet, genesis *types.BlockHeader, f3TempFile *os.File, err error) {
+func (cs *ChainStore) Import(ctx context.Context, f3Ds dtypes.F3DS, r io.Reader) (head *types.TipSet, genesis *types.BlockHeader, err error) {
 	// TODO: writing only to the state blockstore is incorrect.
 	//  At this time, both the state and chain blockstores are backed by the
 	//  universal store. When we physically segregate the stores, we will need
@@ -177,7 +175,7 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (head *types.TipS
 
 	br, err := carv2.NewBlockReader(r)
 	if err != nil {
-		return nil, nil, nil, xerrors.Errorf("loadcar failed: %w", err)
+		return nil, nil, xerrors.Errorf("loadcar failed: %w", err)
 	}
 
 	s := cs.StateBlockstore()
@@ -198,27 +196,27 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (head *types.TipS
 	var buf []blocks.Block
 
 	if len(roots) == 0 {
-		return nil, nil, nil, xerrors.Errorf("no roots in snapshot car file")
+		return nil, nil, xerrors.Errorf("no roots in snapshot car file")
 	}
 
 	if len(roots) == V2SnapshotRootCount {
 		var metadata SnapshotMetadata
 		blk, err := br.Next()
 		if err != nil {
-			return nil, nil, nil, xerrors.Errorf("failed to read snapshot metadata block: %w", err)
+			return nil, nil, xerrors.Errorf("failed to read snapshot metadata block: %w", err)
 		}
 
 		if err := metadata.UnmarshalCBOR(bytes.NewReader(blk.RawData())); err != nil {
 			// Only one cid in roots, but not metadata, maybe it's a genesis block
 			log.Infof("failed to unmarshal snapshot metadata block: %v, attempting to parse the block as a genesis block instead", err)
 			if err := tailBlock.UnmarshalCBOR(bytes.NewReader(blk.RawData())); err != nil {
-				return nil, nil, nil, xerrors.Errorf("failed to unmarshal genesis block: %w", err)
+				return nil, nil, xerrors.Errorf("failed to unmarshal genesis block: %w", err)
 			}
 			if len(tailBlock.Parents) > 0 {
 				nextTailCid = tailBlock.Parents[0]
 			} else {
 				// note: even the 0th block has a parent linking to the cbor genesis block
-				return nil, nil, nil, xerrors.Errorf("current block (epoch %d cid %s) has no parents", tailBlock.Height, tailBlock.Cid())
+				return nil, nil, xerrors.Errorf("current block (epoch %d cid %s) has no parents", tailBlock.Height, tailBlock.Cid())
 			}
 
 			buf = append(buf, blk)
@@ -227,26 +225,20 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (head *types.TipS
 				// import f3 snapshot
 				cid, f3Reader, _, err := br.NextReader()
 				if err != nil {
-					return nil, nil, nil, xerrors.Errorf("failed to read F3Data reader: %w", err)
+					return nil, nil, xerrors.Errorf("failed to read F3Data reader: %w", err)
 				}
 				if cid != *metadata.F3Data {
-					return nil, nil, nil, xerrors.Errorf("F3Data CID mismatch")
+					return nil, nil, xerrors.Errorf("F3Data CID mismatch")
 				}
 
-				// Save f3 data to temporary file
-				f3TempFile, err = os.CreateTemp("", "f3-snapshot-*.tmp")
-				if err != nil {
-					return nil, nil, nil, xerrors.Errorf("failed to create temporary file for F3 snapshot: %w", err)
-				}
+				f3r := bufio.NewReader(f3Reader)
 
-				// Copy f3Reader to temporary file
-				if _, err := io.Copy(f3TempFile, f3Reader); err != nil {
-					return nil, nil, nil, xerrors.Errorf("failed to copy F3 snapshot to temporary file: %w", err)
-				}
+				prefix := F3DatastorePrefix()
+				f3DsWrapper := namespace.Wrap(f3Ds, prefix)
 
-				// Reset file position to beginning for later reading
-				if _, err := f3TempFile.Seek(0, 0); err != nil {
-					return nil, nil, nil, xerrors.Errorf("failed to seek to beginning of F3 temporary file: %w", err)
+				log.Info("Importing F3Data to datastore")
+				if err := certstore.ImportSnapshotToDatastore(ctx, f3r, f3DsWrapper); err != nil {
+					return nil, nil, xerrors.Errorf("failed to import f3Data to datastore: %w", err)
 				}
 			}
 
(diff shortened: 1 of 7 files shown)


tediou5:
@BigLep The code has not been tested yet, as I haven't had a chance to write the test cases. I'm submitting it now due to time constraints, but if you're interested, you can start with a preliminary review. There isn't much code, and it's mostly based on the existing Forest implementation.

chain/store/snapshot.go
```diff
@@ -80,10 +112,39 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (head *types.TipS
...
+		cid, f3Reader, _, err := br.NextReader()
+		if err != nil {
+			return nil, nil, xerrors.Errorf("failed to read F3Data reader: %w", err)
+		}
+		if cid != metadata.F3Data {
+			return nil, nil, xerrors.Errorf("F3Data CID mismatch")
+		}
+
+		reader := bufio.NewReader(f3Reader)
+		if err := certstore.ImportSnapshotToDatastore(ctx, reader, cs.metadataDs); err != nil {
```
Kubuxu: The certs are not stored in `metadataDs`, there is a separate datastore for F3.

tediou5: Thanks for your review. So, can I do it like this?
```go
f3Ds := namespace.Wrap(cs.metadataDs, datastore.NewKey("/f3"))
```

Kubuxu:
Have you been able to perform some manual testing on this?

tediou5:
> Have you been able to perform some manual testing on this?

Not yet. It has only passed the unit tests so far. I've also added an Import/Export test case that includes `F3Data`.

tediou5:
[9940ecb](https://github.com/filecoin-project/lotus/pull/13282/commits/9940ecb91f219e9ba3ff5553fd3e1866ddc478a2): I am manually testing the import of a [v2 snapshot](https://forest-snapshots.fra1.cdn.digitaloceanspaces.com/f3/forest_snapshot_calibnet_2025-08-07_height_2908399_f3_555573.forest.car.zst) provided by Forest. During the test, I found that Forest serializes `SnapshotMetadata` as a Map, while I was previously using a Tuple. I have now aligned my implementation with Forest's behavior.

I've also encountered an issue in go-f3 where `certstore.ImportSnapshotToDatastore` fails with the error: `failed to decode finality certificate: incomplete block, 19325 bytes expected, 2405 bytes got`, and I am currently investigating it.

tediou5:
318512e: During testing, I discovered that our version of `go-f3` is different from the one used in Forest. The `ImportSnapshotToDatastore` error I mentioned above disappeared after I upgraded the dependency.

Strictly speaking, this version bump should be in a separate PR. I've only included it here to make testing easier. If you'd like, please let me know, and I can submit a separate PR for the upgrade (or you can notify me to rebase after it's upgraded separately).


5047a4a: My mistake, the field name was inconsistent. It has been fixed.

cli/lotus/daemon.go
```diff
@@ -617,8 +617,13 @@ func ImportChain(ctx context.Context, r repo.Repo, fname string, snapshot bool)
 		ir = zr
 	}
 
+	f3Ds, err := lr.Datastore(ctx, "/f3")
```
hanabi1224: Could we reuse `func F3Datastore`?

hanabi1224:
> During testing, I discovered that our version of go-f3 is different from the one used in Forest. The ImportSnapshotToDatastore error I mentioned above disappeared after I upgraded the dependency.

Yes, `go-f3@0.8.10` contains a bug fix and some significant performance improvements

> I found that Forest serializes SnapshotMetadata as a Map

This aligns with CARv1 header which is serialized as a map instead of tuple

BigLep:
@Kubuxu : can you please support getting this landed today (including cutting a go-f3 0.8.10 release)?

chain/store/snapshot.go
```diff
@@ -35,15 +39,50 @@ func (cs *ChainStore) UnionStore() bstore.Blockstore {
 	return bstore.Union(cs.stateBlockstore, cs.chainBlockstore)
 }
 
-func (cs *ChainStore) Export(ctx context.Context, ts *types.TipSet, inclRecentRoots abi.ChainEpoch, skipOldMsgs bool, w io.Writer) error {
+func (cs *ChainStore) Export(ctx context.Context, ts *types.TipSet, inclRecentRoots abi.ChainEpoch, skipOldMsgs bool, f3Ds dtypes.F3DS, w io.Writer) error {
+	store, err := certstore.OpenStore(ctx, f3Ds)
```
Kubuxu: This almost certainly will have a prefix applied to the keys: https://github.com/filecoin-project/go-f3/blob/main/store.go#L23

chain/store/snapshot.go
```diff
@@ -80,15 +119,62 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (head *types.TipS
...
+				cid, f3Reader, _, err := br.NextReader()
+				if err != nil {
+					return nil, nil, xerrors.Errorf("failed to read F3Data reader: %w", err)
+				}
+				if cid != *metadata.F3Data {
+					return nil, nil, xerrors.Errorf("F3Data CID mismatch")
+				}
+
+				reader := bufio.NewReader(f3Reader)
+				if err := certstore.ImportSnapshotToDatastore(ctx, reader, f3Ds); err != nil {
```
Kubuxu: Same here, a manifest specific namespacing gets applied to the datastore https://github.com/filecoin-project/go-f3/blob/main/store.go#L23

chain/store/snapshot.go
```diff
@@ -35,15 +39,50 @@ func (cs *ChainStore) UnionStore() bstore.Blockstore {
...
+	}
+
+		log.Warnf("failed to export latest f3 snapshot: %v; continuing snapshot export without F3 data", err)
+	}
+
+	buffer := bytes.NewBuffer(nil)
+	metadata := SnapshotMetadata{
+		Version:       SnapshotVersion2,
+		HeadTipsetKey: ts.Cids(),
+	}
```
Kubuxu, line 65: I'm not sure if we are ok with exporting only v2 snapshots. I think it is ok but highlighting it.

chain/store/snapshot.go
Kubuxu: Additionally, because this is used on live node, I don't think we should open the certstore a second time. There are no guarantees of a consistent state this way.

chain/store/snapshot.go
```diff
@@ -35,15 +39,50 @@ func (cs *ChainStore) UnionStore() bstore.Blockstore {
...
 	}
 
 	if err := car.WriteHeader(h, w); err != nil {
 		return xerrors.Errorf("failed to write car header: %s", err)
 	}
+	if err := carutil.LdWrite(w, mCid.Bytes(), buffer.Bytes()); err != nil {
+		return xerrors.Errorf("failed to write metadata block to car output: %w", err)
+	}
+	if f3Cid != cid.Undef {
+		if err := carutil.LdWrite(w, f3Cid.Bytes(), f3Buffer.Bytes()); err != nil {
```
Kubuxu: Ideally, we should be able to avoid buffering the whole f3 state in memory. In the absence of that, being able to not export f3 state via an option would be ideal, but that creates a need for API changes

Hmm, tough. I'm cautious of creating a time bomb where, as the f3 state grows, the memory required for the export process increases over time.

Kubuxu:
Structurally, the chainstore doesn't seem an optimal place for all of this, but we can stuff it in there, given the timeline.

> Strictly speaking, this version bump should be in a separate PR
Not really, it is fine if the go-f3 version bump happens in here due to dependency.

As far as I can see, due to the datastore location mismatch, it will not work right now, but after that, it should.

Kubuxu:
Rlease PR is up for go-f3, should be released in 15ish minutes: https://github.com/filecoin-project/go-f3/pull/1050

Kubuxu:
A definitive manual test for verifying the feature's working:
After importing a F3-enabled snapshot into an empty node, F3 resumes operation from the instance in the snapshot.
This can be checked by running the node and issuing `lotus f3 status` and checking the instance number.

chain/store/snapshot.go
tediou5: Yes, reading directly into memory does carry risks. Perhaps it would be safer to write to a temporary file first.

chain/store/snapshot.go
tediou5: I'm not deeply familiar with go-f3, but at this point, this appears to be the only viable method I've identified to obtain the certstore I require. I'll investigate it further when I return tomorrow and will also address the prefix issue you raised.

Kubuxu: go-f3 exposes the certstore: https://github.com/filecoin-project/go-f3/blob/main/f3.go#L131
you would have to expose it in `chain/lf3/f3.go.F3`: https://github.com/filecoin-project/lotus/blob/master/chain/lf3/f3.go#L53

tediou5:
@Kubuxu Refactored `Import` and `Export`:

1.  Both `Import` and `Export` now write the `F3Snapshot` to a temporary file first.
2.  `Export` has been split into `ExportV1` and `ExportV2`. `ExportV1` retains the `FilecoinSnapshotV1` format and is used by the `ChainExport` node API to ensure backward compatibility. `lotus-shed` now defaults to exporting the V2 format. Furthermore, because the `Import` process requires loading the `ChainStore` to get the `NetworkName` for deriving the `f3ds` prefix, the `ExportV2` function now returns the `F3DataTempFile`, and a corresponding `ImportF3Data` method has been added.

The `f3ds` prefix is now calculated correctly, which has been confirmed through manual testing.

Manual testing of the Forest snapshot import was successful.

chain/store/snapshot.go
tediou5: After some consideration during the refactor, I'd prefer not to change the API right now. Therefore, I've reverted ChainExport to export the V1 format, while shed will export in V2. Adding a dedicated ChainExportV2 later on is probably a better approach.

chain/store/snapshot.go
```diff
@@ -252,10 +329,65 @@ func (cs *ChainStore) Import(ctx context.Context, f3Ds dtypes.F3DS, r io.Reader)
...
+
+func (cs *ChainStore) ImportF3Data(ctx context.Context, f3Ds dtypes.F3DS, f3TempFile *os.File) error {
+	defer func() {
+		f3TempFile.Close()
+		os.Remove(f3TempFile.Name())
+	}()
+
+	f3Reader := bufio.NewReader(f3TempFile)
+
+	prefix, err := cs.F3DatastorePrefix(ctx)
```
Kubuxu: Instead of doing this, I would reccomend using: https://github.com/filecoin-project/lotus/blob/a720900f547b70edf15fcb501810f777a6150f27/build/buildconstants/shared_funcs.go#L56

Not the cleanest, but it is stand-alone.
Although I can be convinced otherwise.

chain/store/snapshot.go
```diff
@@ -161,15 +227,26 @@ func (cs *ChainStore) Import(ctx context.Context, f3Ds dtypes.F3DS, r io.Reader)
...
-				if err := certstore.ImportSnapshotToDatastore(ctx, reader, f3Ds); err != nil {
-					return nil, nil, xerrors.Errorf("failed to import f3Data to datastore: %w", err)
+				// Save f3 data to temporary file
+				f3TempFile, err = os.CreateTemp("", "f3-snapshot-*.tmp")
+				if err != nil {
+					return nil, nil, nil, xerrors.Errorf("failed to create temporary file for F3 snapshot: %w", err)
+				}
+
+				// Copy f3Reader to temporary file
+				if _, err := io.Copy(f3TempFile, f3Reader); err != nil {
```
Kubuxu: Is there a reason to copy to a temporary file for import? I think we can straight up import from the reader (as it doesn't buffer).

Kubuxu: but on the other hand, this works, so I would like to hear your opinion as someone who dug deeper into this

Kubuxu:
The last concrete blocker is the tests passing

chain/store/snapshot.go
tediou5: I haven't tried getting the `NetworkName` via `F3Manifest()`.

As you can see, I'm currently getting it from the `InitActor`, which means I have to complete the `ChainStore` import before I can retrieve the `NetworkName`; otherwise, nothing can be fetched from an empty repo.
In my case, I'm unable to bypass this reader, so I need to copy it to a file first.

I'll try it tomorrow. If `F3Manifest()` works as expected, I can adjust both of these parts together.

BigLep (commented):
I think this change should have a changelog entry.  Feel free to correct, but I added on in https://github.com/filecoin-project/lotus/pull/13282/commits/b077ba825e765dc3a555c4b19fa534951cee2ee5

cli/lotus/daemon.go
tediou5: It's having some difficulty, it's not easy to get the required parameters.

tediou5:
@Kubuxu Thanks for your review. The unit tests have been fixed. I've also switched to using `F3Manifest()`, which is great and simplified the code a lot.

Kubuxu (approved):

chain/store/snapshot.go
BigLep: Resolving since we have export of v1 and v2

chain/store/snapshot.go
BigLep: Resolving since approval was given.

chain/store/snapshot.go
BigLep: Resolving since we export v1 and v2

chain/store/snapshot.go
BigLep: Resolving since now uses temporary files

chain/store/snapshot.go
```diff
@@ -80,15 +122,62 @@ func (cs *ChainStore) Import(ctx context.Context, r io.Reader) (head *types.TipS
...
+			buf = append(buf, blk)
+		} else {
+			if metadata.F3Data != nil {
+				// import f3 snapshot
+				cid, f3Reader, _, err := br.NextReader()
+				if err != nil {
+					return nil, nil, xerrors.Errorf("failed to read F3Data reader: %w", err)
+				}
+				if cid != *metadata.F3Data {
+					return nil, nil, xerrors.Errorf("F3Data CID mismatch")
```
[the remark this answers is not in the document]
BigLep: Resolving since wasn't deemed important enough to address in previous review rounds.

cli/lotus/daemon.go
BigLep: My understanding is this comment has been effectively addressed by centralizing the prefix logic with `F3DatastorePrefix()`, which is usable in more contexts.  Resolving.

chain/store/snapshot.go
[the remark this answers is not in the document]
BigLep: Resolving since mustn't be a priority if not flagged during the human review rounds.

BigLep:
I'm going to squash and merge since comments have been incorporated / resolved and approval was given.  Thank you @tediou5 !

PR #13298, opened 2025-09-03
feat(eth): use F3 for "finalized" and "safe" resolution in v1 APIs

rvagg:
As per chat with @BigLep, this is how we backport F3 finalisation rules to v1 Eth APIs if we want to go with that path. It also comes with a future clean-up task that we should open an issue for if we merge this.

---

Same rules as v2:

* "finalized" and "safe" resolve to whatever F3 has finalized
* Fall-back for "safe" is to 200 epochs behind head when F3 isn't operating
* Fall-back for "finalized" is 900 epochs behind head when F3 isn't operating

Run with LOTUS_ETH_V1_DISABLE_F3_FINALITY_RESOLUTION=1 to revert to the old resolution rules.

Future clean-up required:

* Remove LOTUS_ETH_V1_DISABLE_F3_FINALITY_RESOLUTION option
* Remove builder_chain.go split paths for v1 and v2 Eth module building - reuse same modules across v1 and v2.
* Simplify eth_api_f3_test.go to remove special rules for v1 vs v2 - test same rules

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
   - snapshot export now defaults to v2 format with embedded F3 finality certificates, dramatically reducing F3 catchup time from ~8 hours
   - transparently imports both v1 and v2 snapshot formats
   - to export v1 snapshots, use `lotus chain export --skip-old-msgs --recent-stateroots=2001 --snapshot-version=1 <filename>`
-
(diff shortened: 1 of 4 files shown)


Kubuxu (approved):
Seems like a good way to approach it

BigLep:
Cleanup work being tracked in https://github.com/filecoin-project/lotus/issues/13315

PR #13299, opened 2025-09-03
feat(f3): expose simple ChainGetFinalizedTipSet that just returns a tipset

rvagg:
Alias for F3GetLatestCertificate -> ECChain.TipSets[-1] -> TipSet

As discussed with @BigLep here's what a user-focused F3 API in /v1 as opposed to the miner software or developer focused APIs and does away with the [indirection complexity](https://www.notion.so/filecoindev/Exchange-and-RPC-providers-guide-to-F3-Fast-Finality-105dc41950c180c49175e4b3afcc5923?source=copy_link#238dc41950c18018b433d7eee2031360) when trying to explain how to use F3 with /v2 APIs. An option on the table for us to consider in the range of possibilities for helping increase F3 adoption.

***Edit: renamed and added EC fallback, see below.***


BigLep:
> I assume it'll be a separate PR to make /v1 ETH APIs F3 aware.

Doh, my bad.  I see https://github.com/filecoin-project/lotus/pull/13298

eshon:
Can this just be called `GetFinalizedTipset` and if F3 is not running or behind it returns EC finality at 900?

Kubuxu (commented):
SGTM code wise, eshon's suggestion seems good as well

BigLep:
> Can this just be called `GetFinalizedTipset` and if F3 is not running or behind it returns EC finality at 900?

I am good with doing this instead.  I can see reasons for it vs. the `F3GetFinalizedTipset`.  No strong opinions here, but I would nudge towards `GetFinalizedTipset` given the intent to "just serve the user who wants to know a finalized tipset".

---
Note to self on followup actions I imagine taking after this and https://github.com/filecoin-project/lotus/pull/13298 land:
1. Update https://www.notion.so/filecoindev/Exchange-and-RPC-providers-guide-to-F3-Fast-Finality-105dc41950c180c49175e4b3afcc5923
2. Update https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658
3. Propose this as a new API to Common Node API https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0104.md

Kubuxu:
It is either `F3GetFinalizedTipset` or `ChainGetFinalizedTipset` the functionality difference is more important to me

BigLep:
My understanding is there are two options:
1. What is currently implemented: `F3GetFinalizedTipset` which returns the latest finalized F3 tipset or if F3 is not running or has not finalized any rounds yet, an error is returned.
2. @eshon proposal: `ChainGetFinalizedTipset` which returns the latest finalized F3 tipset or if F3 is not running or behind more than 900 epochs it returns EC finality at head-900

What are you voting for @Kubuxu?

Kubuxu:
Option two seems like better UX, but I don't have a strong opinion. If @eshon thinks that it will have better adoption, we should go for it unless there are implementation concerns.

BigLep:
I'm good with option 2 as well.  Are you @rvagg ?

rvagg:
yep, option 2 it is, I like it, `ChainGetFinalizedTipSet` (capital `S` at the end there) and fall-back to 900. I'll make it so.

rvagg:
Done, with comprehensive tests for the various states of F3 enablement or erroring. If there are any problems with F3, including it falling behind EC, then we revert to EC for finality, otherwise we return the F3 finalised tipset.

> feat(f3): expose simple ChainGetFinalizedTipSet API on v1 (and gateway) that just returns the latest F3 finalized tipset, or falls back to EC finality if F3 is not operational on the node. This API can be used for follow-up state calls that clamp to a specific tipset to have assurance of state finality.

itests/f3_test.go
```diff
@@ -419,3 +427,158 @@ func setupWithStaticManifest(t *testing.T, manif *manifest.Manifest, testBootstr
 
 	return e
 }
+
+// TestAPIChainGetFinalizedTipSet tests the behavior of ChainGetFinalizedTipSet
+// under various F3 states, including F3 disabled, nominal operation,
+// F3 not ready, F3 internal errors, and implausible F3 certificates.
+// This is modelled off similar tests in api_v2_test.go for v2/ChainGetTipSet.
```
BigLep, line 429: What about the scenario where where F3 finalized tipset is more than 900 epochs behind head?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
   - snapshot export now defaults to v2 format with embedded F3 finality certificates, dramatically reducing F3 catchup time from ~8 hours
   - transparently imports both v1 and v2 snapshot formats
   - to export v1 snapshots, use `lotus chain export --skip-old-msgs --recent-stateroots=2001 --snapshot-version=1 <filename>`
+- feat(f3): expose simple ChainGetFinalizedTipSet API on v1 (and gateway) that just returns the latest F3 finalized tipset, or falls back to EC finality if F3 is not operational on the node. This API can be used for follow-up state calls that clamp to a specific tipset to have assurance of state finality. ([filecoin-project/lotus#13299](https://github.com/filecoin-project/lotus/pull/13299))
```
BigLep:
```suggestion
- feat(f3): expose simple ChainGetFinalizedTipSet API on v1 (and gateway) that just returns the latest F3 finalized tipset, or falls back to EC finality if F3 is not operational on the node or if the F3 finalized tipset is further back than EC finalized tipset. This API can be used for follow-up state calls that clamp to a specific tipset to have assurance of state finality. ([filecoin-project/lotus#13299](https://github.com/filecoin-project/lotus/pull/13299))
```

This is what I was assuming based on https://github.com/filecoin-project/lotus/pull/13299#issuecomment-3250042363.

I'll defer to your judgement though

BigLep (approved):

itests/f3_test.go
rvagg: covered in the `f3 is older than ec falls back to ec` variant in here, we finalize at 910 and expect 900

node/impl/full/chain.go
```diff
@@ -249,6 +255,74 @@ func (m *ChainModule) ChainGetTipSetAfterHeight(ctx context.Context, h abi.Chain
...
 }
 
+func (m *ChainModule) ChainGetFinalizedTipSet(ctx context.Context) (*types.TipSet, error) {
+	head := m.Chain.GetHeaviestTipSet()
+	ecFinalityHeight := head.Height() - policy.ChainFinality
+
+	// Try F3-based finality first.
+	if ts := m.tryF3Finality(ctx, ecFinalityHeight); ts != nil {
+		return ts, nil
+	}
```
ribasushi, line 265: `tryF3Finality` should bubble up "hard" errors, so that the highlighted block of code can distinguish between "F3 is not running/behind" vs "node is under some novel type of attack, with invalid F3 or unexpectedly divergent F3/EC"

Kubuxu: I disagree. This should continue working to reduce the impact on people who use it. 900 epochs is a totally valid fallback.

In cases of an attack on F3, this is a fallback we would recommend and use.

ribasushi: While typing this I see that the PR was merged. So my comments are likely moot. Nevertheless leaving them here for posterity.

----

@Kubuxu notionally you are totally correct - the exposed finality should match the "internal finality", that is whatever the chainsync algorithm would do. The problem is that holistically `ChainGetFinalizedTipSet` is not the only finality-aware API - there are literally hundreds of them. Given `tryF3Finality` is a newly written function, which does not seem a copy of something that already exists - the likelihood of divergence is high

The more in-depth thing to point out in review would be 
> This looks fishy, the decision path should be unified, not unlike the internal `GetHeaviestTipSet`. Then all of the ETH v1/2 and this new  API call would reliably return the same thing

I did not say this earlier, because it is untenable to implement a cleanup pass in the short window FilOz committed themselves to. So the second best thing was the practical "have the API error instead of potentially diverging under uncharted circumstances"

BigLep:
> Note to self on followup actions I imagine taking after this and #13298 land:
> 
> 1. Update https://www.notion.so/filecoindev/Exchange-and-RPC-providers-guide-to-F3-Fast-Finality-105dc41950c180c49175e4b3afcc5923

✅ Done

> 2. Update https://www.notion.so/filecoindev/Filecoin-V2-APIs-1d0dc41950c1808b914de5966d501658

✅ Done per https://github.com/filecoin-project/lotus/pull/13352

> 3. Propose this as a new API to Common Node API https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0104.md

✅ Done per https://github.com/filecoin-project/FIPs/pull/1199

node/impl/full/chain.go
BigLep: @ribasushi : I went ahead and merged given the timelines and because I saw Kuba had responded to your comment and had given his approval.  I am tracking the followup / cleanup work to this PR in https://github.com/filecoin-project/lotus/issues/13315 . If you still feel strongly about this change, feel free to propose it there.

PR #13328, opened 2025-09-09
fix: resolve quote nesting in network-specific build targets

strahe:
Fixes quote nesting bug in network-specific build targets when LDFLAGS is set.

  Commands like `LDFLAGS=-L/usr/local/cuda/lib64 make calibnet-lotus` failed due to nested double quotes in GOFLAGS construction:
  - `GOFLAGS+=-ldflags="$(ldflags)"` + `GOFLAGS="$(GOFLAGS) $(FLAGS)"`
  - Result: `GOFLAGS="-ldflags="..." -tags=calibnet"` (parsing error)

  Changed to single quotes to avoid nesting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,15 +79,15 @@ INTEROPNET_FLAGS=-tags=interopnet
 
 # Network-specific pattern rules
 debug-%:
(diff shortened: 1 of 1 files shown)


rvagg (approved):
thanks!

PR #13330, opened 2025-09-10
fix: use single quotes in GOFLAGS ldflags to match network targets

rjan90:
## Related Issues
Fixes quote nesting issue introduced by PR #13328.

## Proposed Changes
PR #13328 changed network-specific targets to use single quotes around GOFLAGS assignments, but the base GOFLAGS definition still used double quotes around $(ldflags), causing quote nesting when LDFLAGS is set, and failure to run the make <network> commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ ifneq ($(strip $(LDFLAGS)),)
 	ldflags+=-extldflags=$(LDFLAGS)
 endif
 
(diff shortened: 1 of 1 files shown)


rjan90:
> can you confirm make lotus still works with that since it uses unquoted GOFLAGS

Can confirm that `make lotus` still works for me on MacOS and go version go1.23.10

Issue #13346, opened 2025-09-16
v2 snapshot export errors


BigLep:
As reported by @hanabi1224 in https://filecoinproject.slack.com/archives/CPFTWMY7N/p1757556028301299

I saw some errors when exporting a v2 snapshot, any ideas what could go wrong? The ~/.lotus repo is in sync

```
➜  v2 lotus-shed -v && lotus-shed export --recent-stateroots 900 --skip-old-msgs lotus_v2.car
lotus-shed version 1.34.0-rc2+calibnet+git.c41477c95
2025-09-11T09:57:17.088+0800    ERROR   lotus-shed      lotus-shed/main.go:186  loading tipset:
    github.com/filecoin-project/lotus/chain/store.(*ChainStore).loadHead
        /home/me/git/lotus/chain/store/store.go:237
  - get block bafy2bzacebxzep4losvn7vmhotyknyxpqjlriyncrxj5w5jcyc7ntgjvc7dx6:
    github.com/filecoin-project/lotus/chain/store.(*ChainStore).LoadTipSet.func1
        /home/me/git/lotus/chain/store/store.go:917
  - ipld: could not find bafy2bzacebxzep4losvn7vmhotyknyxpqjlriyncrxj5w5jcyc7ntgjvc7dx6

➜  ~ lotus chain list --count 3
3007653: (Sep 11 09:59:30) [ bafy2bzacebspb35y2qtuc6zac775lb5h7rdg5vkogx5z7mxkgu4yfzin7sy6i: t0135498,bafy2bzacecdg532qeiflifdwkaaxcymsfycpyg2e6osodnyf4cn2f6b5bgt5o: t0143103,bafy2bzacedc663rwpxb7bk4f6ltfqrcxrusb34amjpz4zysjvyooxkbhg4tki: t017840, ]
3007654: (Sep 11 10:00:00) [ bafy2bzacebm5qm6kfjedeq5r3ujyqchea6rml3ec2pnpzxgxwt72egdvlo774: t0155502,bafy2bzaceaqxxpagfeafe5qxwlejpu3mgfwwljtavttuyfe2gtlwwslzhflrc: t0143103, ]
3007655: (Sep 11 10:00:30) [ bafy2bzacedvgqzndkog22hkt7ogqyd53bsob2qlyfjpx7uvkk2kijo6gfci7e: t0143103, ]
```

This is blocking him from testing importing Lotus-generated v2 snapshot into Forest.

BigLep:
[...]

PR #13347, opened 2025-09-18
fix(shed): open ChainStore from splitstore on export

tediou5:
## Related Issues

close: #13346

## Proposed Changes

When using `lotus-shed export`, we previously only opened the `ChainStore` from the `UniversalBlockstore`. This caused issues with not being able to retrieve the latest tipsets.

Now, we open it from the `splitstore`, and I have manually tested this locally. The export and re-import process now works correctly.

## Checklist

- [ ] CI is green

### DIFF
--- a/cmd/lotus-shed/export.go
+++ b/cmd/lotus-shed/export.go
@@ -33,6 +33,8 @@ import (
 	"github.com/filecoin-project/go-state-types/abi"
 
 	"github.com/filecoin-project/lotus/blockstore"
+	badgerbs "github.com/filecoin-project/lotus/blockstore/badger"
+	"github.com/filecoin-project/lotus/blockstore/splitstore"
 	"github.com/filecoin-project/lotus/chain/store"
 	lcli "github.com/filecoin-project/lotus/cli"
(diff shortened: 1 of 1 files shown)


rvagg:
I'd like @ZenGround0 to consider this one. I found myself doing something similar at one point but @ZenGround0 has made the point that it _shouldn't_ matter, you shouldn't be having to do this manual work at this point. But then again, this does seem to be a necessary hurdle in some cases.
But also, splitstore isn't mandatory for users, there are valid cases where splitstore isn't being used at all. I've got a node running right now with:

```
[Chainstore]
  EnableSplitstore = false
```

tediou5:
> you shouldn't be having to do this manual work at this point

@ZenGround0 I'm not familiar enough with this area. Please feel free to make suggestions.

> But also, splitstore isn't mandatory for users, there are valid cases where splitstore isn't being used at all.

@rvagg Should this be configurable? We could add a parameter to specify if Splitstore is enabled.

Kubuxu (approved):

rjan90:
With approval from Kuba here, I will go ahead and merge

PR #13389, opened 2025-10-15
fix(eth): properly return vm error in all gas estimation methods

rvagg:
I realised this is why we don't have consistent output on reverts, sometimes we get the `vm error` and sometimes not.

Same failure case, slightly different pathways:

```
failed to estimate gas: failed to estimate gas: message execution failed: exit 33, reason: message failed with backtrace:
00: f0169791 (method 3844450837) -- contract reverted at 75 (33)
01: f0169791 (method 6) -- contract reverted at 4535 (33)
02: f0169800 (method 3844450837) -- contract reverted at 75 (33)
03: f0169800 (method 6) -- contract reverted at 18957 (33)
 (RetCode=33)
```

```
failed to estimate gas: message execution failed (exit=[33], revert reason=[message failed with backtrace:
00: f0169791 (method 3844450837) -- contract reverted at 75 (33)
01: f0169791 (method 6) -- contract reverted at 4535 (33)
02: f0169800 (method 3844450837) -- contract reverted at 75 (33)
03: f0169800 (method 6) -- contract reverted at 18957 (33)
 (RetCode=33)], vm error=[0x42d750dc0000000000000000000000000542c82a785fab8b53c7c59aaad7093d87265f3a0000000000000000000000009050bdd028ba54a01777cd8c6d0a59d5bc52fdbf])
```


chain/types/ethtypes/eth_types.go
```diff
@@ -1278,3 +1278,73 @@ type EthTxReceipt struct {
...
 }
+
+const errorFunctionSelector = "\x08\xc3\x79\xa0" // Error(string)
+const panicFunctionSelector = "\x4e\x48\x7b\x71" // Panic(uint256)
+
+// panicErrorCodes maps Solidity panic codes to human-readable descriptions.
+var panicErrorCodes = map[uint64]string{
+	0x00: "Panic()",
+	0x01: "Assert()",
+	0x11: "ArithmeticOverflow()",
```
wjmelements, line 1289: this can also be Underflow, unfortunately. Perhaps `ArithmeticOverflowOrUnderflow()` would prevent confusion

wjmelements (approved):

chain/types/ethtypes/eth_types.go
rvagg: I'm not sure who's consuming these strings and whether they are currently baked into some software. This wouldn't be a bad thing to change, but we'd just have to give warning in release notes. Maybe in a follow-up PR? I'll leave that one up to you but for now I'll just leave things as they are.

PR #13394, opened 2025-10-18
feat: add ordering by nonce flag to get sequential messages

### LINKED ISSUE #13246: `list-messages` order should be sequential
### Checklist

- [x] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the lotus forum](https://github.com/filecoin-project/lotus/discussions/categories/ideas) and select the category as `Ideas`.
- [x] I **have** a specific, actionable, and well motivated feature request to propose.

### Lotus component

- [x] lotus daemon - chain sync
- [ ] lotus JSON-RPC API
- [ ] Other

It would be convenient for some tooling if the messages of `./lotus state list-messages --cids --from <>` were sequentially ordered. From spot checking I think they have an ordering; I think they are in reverse block order and then ordered within the block. Without this I have to do extra steps to order them this way.

I would like for `list-messages --from` to be ordered by nonce, which would also be the order they were executed.

It would also be acceptable for my use case if they were in reverse-order.

### Additional context

_No response_

DarkLord017:
## Related Issues
Fixes #13246 

## Checklist

- [ ] CI is green

### DIFF
--- a/cli/state.go
+++ b/cli/state.go
@@ -904,7 +904,6 @@ var StateListMessagesCmd = &cli.Command{
 				if cctx.Bool("cids") && !obn {
 					fmt.Println(c.String())
 					continue
-
 				}
 
 				m, err := api.ChainGetMessage(ctx, c)


DarkLord017:
@wjmelements is there any way to test this , like instead of running it e2e

cli/state.go
```diff
@@ -888,21 +892,56 @@ var StateListMessagesCmd = &cli.Command{
...
+						}{c, m.Nonce, m})
+					}
 				}
-				b, err := json.MarshalIndent(m, "", "  ")
-				if err != nil {
-					return err
+				sort.Slice(sortedMsgs, func(i, j int) bool {
+					return sortedMsgs[i].Nonce < sortedMsgs[j].Nonce
+				})
+				for _, sm := range sortedMsgs {
```
rvagg: You now have duplicate code here, you could be doing all of this once, probably appending to a slice defined before the loop (with a named `struct` type would be good), then doing a second loop to print what's been selected; or perhaps just making a `printMsg` function you can reuse that handles the "cids" thing and the JSONification. But I think that the loop is simple and cheap enough that you could just do it twice and reduce the duplication down to essentially nothing here

DarkLord017: Now check @rvagg

cli/state.go
```diff
@@ -872,6 +876,13 @@ var StateListMessagesCmd = &cli.Command{
 
 		windowSize := abi.ChainEpoch(100)
 
+        obn := cctx.Bool("order-by-nonce") && !froma.Empty()
+
+		var orderedMessages []struct {
+			Nonce uint64
+			Msg   []byte
+		}
```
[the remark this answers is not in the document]
DarkLord017: should i do this @rvagg @wjmelements

cli/state.go
```diff
@@ -898,11 +909,21 @@ var StateListMessagesCmd = &cli.Command{
...
 					return err
 				}
-				fmt.Println(string(b))
+
+				if obn {
+					orderedMessages = append(orderedMessages, struct {
+						Nonce uint64
+						Msg   []byte
+					}{Nonce: m.Nonce, Msg: b})
+					continue
```
rvagg:
```suggestion
```

since you have an `else`

cli/state.go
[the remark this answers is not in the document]
rvagg: nah this is fine

rvagg:
@DarkLord017 would you updating from master (merge from master or rebase on master) and then editing the CHANGELOG.md to note the addition of this flag. Look at other changelog entries for the pattern and make sure to reference this PR.

rvagg:
oh, and `make gen` needs to be run too, when we edit the CLI tools we need to edit the docs that describe how to use them and they come out of the code

DarkLord017:
ya sure ! doing that

DarkLord017:
done @rvagg

rvagg (changes requested):
Sorry @DarkLord017, I was just reviewing the original request and it did mention the desire to `--cids` with this. Which is a bit more tricky because there's currently a `continue` for `--cids` so what you'd need to do is take `obn` into account when doing that and not `continue` if `obn` so you can fetch the message and collect the nonce, and then do CID printing at the end after sorting if `obn`, like you do (but just cid)

rvagg:
also if you wouldn't mind resolving the conflict in CHANGELOG.md that I just made for you (sorry)

cli/state.go
```diff
@@ -889,20 +901,31 @@ var StateListMessagesCmd = &cli.Command{
 			}
 
 			for _, c := range msgs {
-				if cctx.Bool("cids") {
+				if cctx.Bool("cids") && !obn {
 					fmt.Println(c.String())
 					continue
+
```
rvagg:
```suggestion
```

PR #13395, opened 2025-10-20
feat(api): chain export v2 support

tediou5:
## Related Issues

Follow-up to #13282

## Proposed Changes

Add a `version` parameters to the ChainExport API to specify the version of the exported snapshot (1 or 2).
The lotus chain cmd exports the V2Snapshot by default. Use the `--version 1` option to export a V1 snapshot.

## Checklist

- [ ] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 - chore: update benchmark tests to use testing.B.Loop for improved performance ([filecoin-project/lotus#13385](https://github.com/filecoin-project/lotus/pull/13385))
 - fix(eth): properly return vm error in all gas estimation methods ([filecoin-project/lotus#13389](https://github.com/filecoin-project/lotus/pull/13389))
 - chore: all actor cmd support --actor ([filecoin-project/lotus#13391](https://github.com/filecoin-project/lotus/pull/13391))
-- feat(api): ChainExport API supports exporting a specified Snapshot version.
+- feat(api): ChainExport API supports exporting a specified Snapshot version ([filecoin-project/lotus#13395](https://github.com/filecoin-project/lotus/pull/13395))
 
 # Node and Miner v1.34.1 / 2025-09-15
 


tediou5:
@rjan90 This pr needs a changelog, I will add it later

tediou5:
I did a manual test of import and export on cali-network with the latest snapshot and it works fine

CHANGELOG.md
```diff
@@ -16,14 +16,15 @@
 - chore: update benchmark tests to use testing.B.Loop for improved performance ([filecoin-project/lotus#13385](https://github.com/filecoin-project/lotus/pull/13385))
 - fix(eth): properly return vm error in all gas estimation methods ([filecoin-project/lotus#13389](https://github.com/filecoin-project/lotus/pull/13389))
 - chore: all actor cmd support --actor ([filecoin-project/lotus#13391](https://github.com/filecoin-project/lotus/pull/13391))
+- feat(api): ChainExport API supports exporting a specified Snapshot version.
```
rvagg:
```suggestion
- feat(api): ChainExport API supports exporting a specified Snapshot version ([filecoin-project/lotus#13395](https://github.com/filecoin-project/lotus/pull/13395))
```

api/api_full.go
```diff
@@ -207,7 +207,8 @@ type FullNode interface {
 	// back to genesis, the entire genesis state, and the most recent 'nroots'
 	// state trees.
 	// If oldmsgskip is set, messages from before the requested roots are also not included.
-	ChainExport(ctx context.Context, nroots abi.ChainEpoch, oldmsgskip bool, tsk types.TipSetKey) (<-chan []byte, error) //perm:read
+	//
```
rvagg: ?

api/api_full.go
```diff
@@ -207,7 +207,8 @@ type FullNode interface {
...
+	ChainExport(ctx context.Context, nroots abi.ChainEpoch, oldmsgskip bool, tsk types.TipSetKey, version uint64) (<-chan []byte, error) //perm:read
```
rvagg, line 210: this is going to be an annoyingly breaking change .. of the type we've kind of agreed to avoid if at all possible in v1

tediou5: Yes, this is a change that needs careful consideration, so I'm putting it in a separate PR now.
We will need it eventually, and using lotus-shed to export is not always convenient (need to take the node offline). But I agree that we can discuss it more. I just submitted it first so that you can see the diff

BigLep:
@tediou5 : what are the next steps here?  Are you going to drive this forward?

api/api_full.go
rvagg: We _could_ just switch version depending on tipset, at some epoch we start exporting v2 and prior to that we export v1. As long as we document it clearly, and those exports work in active implementations then I don't think this would be a major issue?

Unless we still have use for v1 exports that I'm not aware of. @LesnyRumcajs might be able to fill us in here or bring in someone who can.

LesnyRumcajs: From Forest's perspective, it's fine to have a breaking change here.

1. the `ChainExport` is not a part of the Common Node API, see https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0104.md and so there don't have to be very strong guarantees about breaking changes.
2. As it is, this method is rarely used via raw JSON-RPC and more with a wrapper CLI, e.g., `lotus chain export`. I don't expect any RPC provider to expose it. That said, maybe some tooling does it in a raw way.
3. In Forest, we don't even match the API for this method `Filecoin.ChainExport` and in fact, don't use it at all internally, preferring [Forest.ChainExport](https://github.com/ChainSafe/forest/blob/5c043d9b80dc247cac3e909813a2d64bfafc267f/src/rpc/methods/chain.rs#L499). Again, this is due to it being strictly internal RPC and not supposed to be exposed in any way for obvious reasons.

tediou5: > switch version depending on tipset

This is a simpler approach; it's always better to avoid breaking changes.

tediou5:
> what are the next steps here? Are you going to drive this forward?

This isn't my current focus, but @rvagg 's suggestion is simple, I should be pushing a commit soon.

PR #13455, opened 2025-12-09
fix: fr32: Correct multithreaded Unpad chunk boundary alignment and simplify unpadReader

magik6k:
#### Summary

This PR fixes a data corruption bug in the fr32 multithreaded `Unpad()` function and simplifies the `unpadReader` implementation. The bug caused data loss and corruption when processing data sizes that don't evenly divide into chunk-aligned thread work units.

2025-12-05-225323_1517x104_scrot

#### The Bug

The `mt()` function in `fr32.go` parallelizes `Pad()`/`Unpad()` operations across multiple threads. Each thread is assigned a range of bytes to process:

```go
threadBytes := abi.PaddedPieceSize(padLen / int(threads))
```

**The problem**: `threadBytes` is not guaranteed to be a multiple of 128 bytes (one fr32 chunk).

For example, with `padLen = 8455808` bytes (66061 chunks) and `threads = 16`:
- `threadBytes = 8455808 / 16 = 528488` bytes
- `528488 / 128 = 4128.5` — **not a whole number of chunks**

The `unpad()` function processes only complete 128-byte chunks:
```go
chunks := len(in) / 128  // = 4128 (integer division)
for chunk := 0; chunk < chunks; chunk++ { ... }
```

This means each thread only processes `4128 * 128 = 528384` bytes, leaving **104 bytes unprocessed** at each thread boundary. With 16 threads, up to 1664 bytes could be silently lost.

#### When the bug occurs

The bug manifests when ALL of these conditions are met:

1. **Data size exceeds `MTTresh` (512KB)** — triggers multithreaded path
2. **`padLen / threads` is not a multiple of 128** — creates misaligned thread boundaries
3. **Thread count doesn't evenly divide the chunk count** — causes fractional chunks per thread

This is why existing tests (like `TestRoundtrip16MRand` with 16MiB data) passed — 16MiB divides evenly into chunk-aligned work units. The bug only appeared with sizes like ~8.5MiB that create fractional chunks per thread.

#### Symptoms observed

- Zeros appearing in output (at `0x7fffe0` to `0x800000` in the original report)
- Padded data appearing in unpadded output (padding bits `0x40, 0x80, 0xc0, 0x00` pattern visible)
- Data shifted by a few bits after the corruption point
- Corruption occurring at regular intervals (~512KB boundaries, matching `MTTresh`)

#### The Fix

**1. Fix `mt()` chunk boundary alignment** (`fr32.go`):

```go
// Before: threadBytes could be fractional chunks
threadBytes := abi.PaddedPieceSize(padLen / int(threads))

// After: align to 128-byte chunk boundaries
chunksPerThread := (padLen / int(threads)) / 128
threadBytes := abi.PaddedPieceSize(chunksPerThread * 128)

// Last thread handles any remainder
if thread == int(threads)-1 {
    end = abi.PaddedPieceSize(padLen)
}
```

**2. Simplify `unpadReader`** (`readers.go`):

Rewrote using a simple `bufio.Reader`-style pattern:
- Separate `padbuf` and `unpadbuf` buffers (eliminates any memory aliasing concerns)
- Simple `fill()` → `Read()` pattern
- No complex stash/pool mechanisms that were prone to edge-case bugs

#### Testing

- Added regression tests specifically for the boundary conditions
- All existing fr32 tests continue to pass
- New test `TestUnpadReaderSizeMismatch_ExactByteLoss` specifically verifies data integrity around 8MiB boundaries with non-power-of-2 aligned sizes

#### Impact

This bug could cause silent data corruption when:
- Reading piece data through `PieceProvider.ReadPiece()`  
- Any code path using `fr32.Unpad()` with data sizes > 512KB that create misaligned thread boundaries

The corruption would manifest as incorrect bytes in the middle of the data stream, potentially causing:
- Failed piece verification
- Invalid CommP calculations
- Corrupted unsealed sector data

#### Why this "worked"

I don't think anyone seriously used /piece retrievals and /ipfs retrievals usually fetched individual blocks which were much smaller than the ~8MB threshold where this happened. This is just a theory, this might have been causing an untold number of unexplained issues,

### DIFF
--- a/storage/sealer/fr32/readers_test.go
+++ b/storage/sealer/fr32/readers_test.go
@@ -577,7 +577,6 @@ func TestUnpadReaderSizeMismatch_OffsetRead(t *testing.T) {
 
 	startOffsetPadded := startOffsetUnpadded.Padded()
 	endOffsetPadded := endOffsetUnpadded.Padded()
-	rangePadded := endOffsetPadded - startOffsetPadded
 
 	// Extract the padded range from the full padded data
 	paddedRange := fullPadded[startOffsetPadded:endOffsetPadded]
@@ -587,7 +586,7 @@ func TestUnpadReaderSizeMismatch_OffsetRead(t *testing.T) {
 
 	// BUG PATTERN: Create unpadReader with full piece size but only range available
 	// Note: We use a valid piece size that encompasses our range
-	declaredSize := abi.PaddedPieceSize(rangePadded)
+	declaredSize := endOffsetPadded - startOffsetPadded
 	// Round up to valid piece size (power of 2)
 	for declaredSize&(declaredSize-1) != 0 {
 		declaredSize = declaredSize & (declaredSize - 1)


storage/sealer/fr32/readers.go
magik6k, line 1: Rewrote to have simpler logic inspired by bufio.Reader

storage/sealer/fr32/fr32.go
magik6k, line 1: Actual issue was here

storage/sealer/fr32/fr32_test.go
```diff
@@ -149,6 +150,134 @@ func TestRoundtrip16MRand(t *testing.T) {
...
+		// 66061 chunks = 8455808 padded bytes
+		// With 16 threads: 8455808/16 = 528488 bytes per thread
+		// 528488/128 = 4128.5 - NOT aligned! This was the original bug case.
+		{"66061_chunks_8MiB_boundary", 66061},
+
+		// Various sizes that create misaligned thread boundaries
+		{"prime_chunks_1009", 1009 * 8},  // ~1MB, prime-ish number of chunks
+		{"odd_chunks_8193", 8193},        // Just over 8192 (power of 2)
+		{"odd_chunks_65537", 65537},      // Just over 65536 (power of 2)
+		{"odd_chunks_100003", 100003},    // Large prime-ish
```
ZenGround0: Nit: Its truly prime so why "ish"?

ZenGround0 (approved):
Bug fix and tests LGTM.  We should probably push out a patch release with this ASAP

storage/sealer/fr32/fr32_test.go
magik6k: Yeah was just guessing, I guess you checked.

rvagg (approved):

storage/sealer/fr32/readers_test.go
```diff
@@ -243,3 +244,550 @@ func TestUnpadReaderPartialPiece(t *testing.T) {
...
+
+	// Extract the padded range from the full padded data
+	paddedRange := fullPadded[startOffsetPadded:endOffsetPadded]
+
+	// Create reader with only the range data
+	underlyingReader := bytes.NewReader(paddedRange)
+
+	// BUG PATTERN: Create unpadReader with full piece size but only range available
+	// Note: We use a valid piece size that encompasses our range
+	declaredSize := abi.PaddedPieceSize(rangePadded)
```
tediou5: CI failed: unnecessary conversion (unconvert).

maybe like this?

```go
	startOffsetPadded := startOffsetUnpadded.Padded()
	endOffsetPadded := endOffsetUnpadded.Padded()

	// BUG PATTERN: Create unpadReader with full piece size but only range available
	// Note: We use a valid piece size that encompasses our range
	declaredSize := endOffsetPadded - startOffsetPadded
```

tediou5 (approved):

rjan90:
Merging now given that this has 3 approvals, and will prep it for a patch release as soon as possible

storage/sealer/fr32/fr32.go
rjan90: Resolving this to get the merging to happen

PR #13460, opened 2025-12-23
Feat: gas metrics

magik6k:
## Proposed Changes
Add some useful metrics

## Checklist

- [ ] CI is green

### DIFF
--- a/chain/consensus/compute_state.go
+++ b/chain/consensus/compute_state.go
@@ -2,7 +2,6 @@ package consensus
 
 import (
 	"context"
-	"fmt"
 	"sort"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


rvagg:
Would be interesting stats, but I don't think we can merge this as is cause there's a ton of work going on in hot(ish) paths: adds a couple of sorts and 2x `stats.Record` per `ApplyBlocks`, and in messagepool we iterate the entire pool (which can be thousands) every message every head change with 3x `stats.Record` per message and 4 sorts and 2 passes for the weighted medians, all while holding `mp.lk` (RLock) too.

I don't think there's a `stats.OnlyRunIfEnabled(func(){})` style thing or a boolean we can use, `stats.Record` is just meant to be a noop when there's no view, but we register all the view up-front regardless _and_ a lot of this work is done in preparation for a `stats.Record`. But we could do something like hide all of this behind an environment variable, `LOTUS_ENABLE_DETAILED_GAS_METRICS` and then the concerns mostly go away.

Issue #13005, opened 2025-04-03
Flaky: itest TestDailyFees


rvagg:
Failure @ https://github.com/filecoin-project/lotus/actions/runs/14224775295/job/39861585611?pr=13003

This block of code is running too late: https://github.com/filecoin-project/lotus/blob/592cf456ac1068b1b89bdc0564f100750d979346/itests/daily_fees_test.go#L488-L496

I _think_ we've hit perfectly on the upgrade epoch to trigger that, or at least straddling it. Consider the following code: https://github.com/filecoin-project/lotus/blob/592cf456ac1068b1b89bdc0564f100750d979346/itests/daily_fees_test.go#L361-L374

The error is coming out of trying to decode the miner actor and getting the v15 format but expecting the v16.

Needs some combination of:

* Time it better and be more careful
* Being specific about the tipset argument to these calls, `types.EmptyTSK` isn't great for that string of calls if the chain keeps on moving while we're checking, perhaps that helper function could even take a tipsetkey as an argument
* Delay the nv25 upgrade epoch a bit more since it's apparently too soon for these CI machines to get everything else done

rvagg:
Separate failure for the same test: https://github.com/filecoin-project/lotus/actions/runs/15777259375/job/44474460618#step:12:4834

```
daily_fees_test.go:335: *** Check consistency of deadline fee records and expiration queue fee deduction records
    daily_fees_test.go:353: 
        	Error Trace:	/home/runner/work/lotus/lotus/itests/daily_fees_test.go:353
        	            				/home/runner/work/lotus/lotus/itests/daily_fees_test.go:633
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 1
        	Test:       	TestDailyFees
        	Messages:   	expected total fees 6376500, got 4480786
```

Triggered @ the end of the test https://github.com/filecoin-project/lotus/blob/739ad9fad9f1ef281af4b8ccccc57a07b8fc3697/itests/daily_fees_test.go#L633-L634

I don't have a good explanation for this one. Maybe the `provingWindowsSince` calculation has an edge-case? It's supposed to calculate how many proving windows where fees were due should have passed for each sector. Perhaps we're doing that calculation right on the cron epoch and it's including the window just passed as having been paid but it's not quite done yet.

[...]

PR #13463, opened 2026-01-05
fix(itests): use chain-provided deadline info to avoid period boundary race

rvagg:
Aiming to fix flaky itests, in particular TestDailyFees

May close https://github.com/filecoin-project/lotus/issues/13005


rvagg:
nah, I'm going to work on this a bit more. I think this is really only going to fix `WaitTillActivatedAndAssertPower` which I don't believe is the main culprit in the TestDailyFees test anyway, those failures are about failure to submit window post I think

PR #13467, opened 2026-01-08
fix(eth): use error code 3 for EExecutionReverted

rvagg:
BREAKING: RPC error codes changed - EActorNotFound (from 3 to 11), EExecutionReverted (from 11 to 3). Mismatched client/server versions will deserialize these errors as the wrong Go type, breaking errors.Is/As checks.

Swap EExecutionReverted and EActorNotFound positions so that EExecutionReverted gets code 3, the de facto Ethereum standard for execution reverted errors. This enables proper error decoding by Ethereum tooling like viem, ethers.js, and other RPC clients.

Error code 3 was introduced in geth v1.9.15 and is now expected by most Ethereum ecosystem tooling for automatic ABI decoding of revert reasons from the error data field.

Ref: https://github.com/ethereum/go-ethereum/pull/21083
Ref: https://github.com/ethereum/go-ethereum/pull/31456
Ref: https://github.com/ethereum/go-ethereum/issues/21886
Ref: https://github.com/wevm/viem/blob/6c11882938673e939830027f47ed73e907f2b90e/src/utils/errors/getContractError.ts
Ref: https://www.quicknode.com/docs/ethereum/error-references
Ref: https://docs.metamask.io/services/reference/ethereum/json-rpc-methods/eth_call/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 ## 👌 Improvements
 - docs: fix outdated link in documentation ([#13436](https://github.com/filecoin-project/lotus/pull/13436))
 - docs: fix dead link in documentation ([#13437](https://github.com/filecoin-project/lotus/pull/13437))
+- feat(cli): implement FRC-0102 signing envelope for wallet sign/verify ([filecoin-project/lotus#13471](https://github.com/filecoin-project/lotus/pull/13471))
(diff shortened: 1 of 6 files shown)


BigLep:
@LesnyRumcajs : I see you have a corresponding change in Forest.  Are you able to give a quick review of this change?  It realistically may be the only eyes we get on this work anytime soon...

rvagg:
I've not had any objections from elsewhere about this so with a review we can just pull the trigger and make it happen.

LesnyRumcajs:
@BigLep FYI the PR requires review/approval from a maintainer with write permissions.

Issue #12896, opened 2025-02-13
Remove the restriction on the "from" field when calling from a contract or identical address


eshon:
### Checklist

- [x] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [x] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [x] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [x] I did not make any code changes to lotus.

### Lotus component

- [ ] lotus daemon - chain sync
- [x] lotus fvm/fevm - Lotus FVM and FEVM interactions
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt/WinningPoSt)
- [x] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
v1.31.1
```

### Repro Steps

The following `eth_call` (via RPC) does **not** work on Calibration testnet but works on Eth Sepolia. 
However it works on Calibration testnet when the "from" field is removed.

`{
    "jsonrpc": "2.0",
    "method":"eth_call",
[...]

PR #13470, opened 2026-01-13
fix(eth): allow eth_call and eth_estimateGas from contract and non-existent addresses

### LINKED ISSUE #12441: Lotus FEVM API Incompatibility with normal Ethereum mainnet behavior, causing issues with integration partners
### Checklist

### Lotus component

- [X] lotus Ethereum RPC
- [ ] lotus FVM - Lotus FVM interactions
- [ ] FEVM tooling
- [ ] Other

### Lotus Version

```text
Running the public glif nodes
```

### Repro Steps

Curl request to `eth_estimateGas` that fails with actor not found:

```
curl 'https://api.calibration.node.glif.io/rpc/v1' \
  -H 'Accept: application/json' \
  -H 'Accept-Language: en-US,en;q=0.9,es;q=0.8' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -H 'Cookie: __stripe_mid=778e36cb-7efb-46da-986b-8e8e1146f17e7602bc' \
  -H 'Origin: chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: none' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36' \
  --data-raw '{"id":588633187,"jsonrpc":"2.0","method":"eth_estimateGas","params":[{"from":"0xb1040d3ce131b3204e5229c80a8d5ae271b2ef09","to":"0x7fa33ed9fce1e9c620cef778c5601286cd68e027","data":"0x5c19a95c000000000000000000000000b1040d3ce131b3204e5229c80a8d5ae271b2ef09"}]}'
```

returns:

```
{"jsonrpc":"2.0","error":{"code":1,"message":"failed to estimate gas: ApplyWithGasOnState failed: call raw get actor: resolution lookup failed (t410fweca2phbggzsatssfheavdk24jy3f3yjxzll22i): resolve address t410fweca2phbggzsatssfheavdk24jy3f3yjxzll22i: actor not found"},"id":588633187}
```

This is an incompatibility with the same call on infura ethereum mainnet, which returns:

```
{"jsonrpc":"2.0","id":3525842942,"result":"0xbea9"}
```

### Describe the Bug

Integration partners call `eth_estimateGas` to estimate the gas of a given transaction. In some circumstances, a web app may estimate the gas of a transaction even if the account has no balance to pay for the transaction fee. Ethereum mainnet seems to ignore the `from` address' balance when calling `eth_estimateGas` . Subsequently, when the same user tries to submit the transaction with the same `from` address, the call fails with insufficient funds (or maybe in our case, actor not found).

In some cases, integrators have to patch filecoin specific functionality into their otherwise chain agnostic frontend 

### Tooling

```text
Curl
```

### Configuration Options

```text
This is the public glif api, i'm not sure its exact config options
```

nijoe1:
## Related Issues

Fixes compatibility with Geth's `eth_call` and `eth_estimateGas` behavior where callers can simulate transactions from contract addresses or non-existent addresses. This is required for tools like Hardhat, Foundry, and other Ethereum development frameworks that rely on this behavior for gas estimation and call simulation.

## Proposed Changes

**API Layer:**
- Add `ErrSenderValidationFailed` error type to distinguish sender validation failures from other errors

**VM Layer:**
- Add `ApplyMessageSkipSenderValidation` method to VM interface
- Implement skip-sender-validation for LegacyVM (bypasses actor existence, account type, nonce, and balance checks)
- FVM delegates validation to state manager level

**State Manager:**
- Add `CallWithGasSkipSenderValidation` for gas estimation from any address
- Add `ApplyOnStateWithGasSkipSenderValidation` for eth_call simulation
- Implement `createSyntheticSenderActor` to create ephemeral EthAccount actors for non-existent senders
- Implement `maybeModifySenderForSimulation` to temporarily modify EVM/Placeholder actors to EthAccount for simulation

**Eth/Gas API:**
- `EthEstimateGas` now falls back to skip-sender-validation when sender validation fails
- `EthCall` uses skip-sender-validation for all calls
- Preserve `ErrExecutionReverted` errors for proper JSON-RPC codec handling
- Add telemetry metrics: `eth_call_count`, `eth_estimate_gas_count`, `eth_estimate_gas_skip_sender`

**Key behaviors:**
- State modifications are isolated via buffered blockstore and discarded after simulation
- Synthetic actors have zero balance (value transfers fail as expected)
- Only EVM and Placeholder actors can be modified for simulation (system actors are rejected)

## Additional Info

This matches Geth's `SkipAccountChecks` behavior in `state_transition.go`. The implementation ensures:
- No state leakage to chain (all modifications are ephemeral)
- Proper error propagation for reverts with full error data
- Backwards compatibility for existing EOA-based calls

fixes: #12441 #12896 

## Checklist

- [ ] Update CHANGELOG.md or signal that this change does not need it per [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#changelog-management) [skip changelog]
- [x] CI is green


rvagg:
Really quick skim-review to start with, will try and make time to go deeper. Overall this looks pretty good @nijoe1, great work!

* I don't think this fully fixes https://github.com/filecoin-project/lotus/issues/12896, there's a deeper account type check that's going to hold that one up. That one gets caught up here ultimately: https://github.com/filecoin-project/ref-fvm/blob/269c858c1162ecceb20e30c207ba4e99ad03e98b/fvm/src/executor/default.rs#L407-L428 - needs to be an account actor, so the sender can exist but if it's not an account actor then it'll bork. The failures that are resulting in that report are from patterns that exist in EVM-land where they construct messages _from_ smart contracts and expect them to estimate correctly, that pattern doesn't work for us. So the plumbing to override that in ref-fvm is gonig to be more complex.
* I'm not sure about the synthetic actor creation going on here to mutate state directly before using it. That probably works but I think we need to work through potential problems with that model, you might be opening up new edge cases. You're passing in a state root that's been mutated by manual means, no longer matches the original requested state root, and may not conform to all of the expected invariants of the state. The more "correct" way is to just insert an implicit message earlier than the message you're executing that creates the actor - so the original state is unmolested but gets mutated through the message system, then you don't have to worry about getting the state structure right, the account creation logic in actors will do that for you.

nijoe1:
> * I don't think this fully fixes [Remove the restriction on the "from" field when calling from a contract or identical address #12896](https://github.com/filecoin-project/lotus/issues/12896), there's a deeper account type check that's going to hold that one up. That one gets caught up here ultimately: https://github.com/filecoin-project/ref-fvm/blob/269c858c1162ecceb20e30c207ba4e99ad03e98b/fvm/src/executor/default.rs#L407-L428 - needs to be an account actor, so the sender can exist but if it's not an account actor then it'll bork. The failures that are resulting in that report are from patterns that exist in EVM-land where they construct messages _from_ smart contracts and expect them to estimate correctly, that pattern doesn't work for us. So the plumbing to override that in ref-fvm is gonig to be more complex.

This case is handled [here](https://github.com/filecoin-project/lotus/blob/c0d0c9797008beef7213bb2bb016a39c83becb6a/chain/stmgr/call.go#L455) where it modifies in memory the Contract.code to look like an EthAccount.code

To check out if the tests are not sufficient in real world scenarios I cloned today the foc-devnet and connected my local lotus repo pointing to this PR code changes. Then using foundry I implemented the same set of tests as the ones defined [here](https://github.com/filecoin-project/lotus/blob/c0d0c9797008beef7213bb2bb016a39c83becb6a/itests/eth_skip_sender_test.go) and all passed using the local lotus rpc on devnet where on calibration failed

rvagg:
OK I think `maybeModifySenderForSimulation` is going to be the wrong approach for the same reason - you're modifying an existing actor, mutating state, which changes the system that you're running the transaction over. By changing the nature of the calling actor you change the outcome of the execution and that could cause either subtle or very obvious side effects that are entirely different to an actual execution. Part of the point of estimateGas is that you should be able to run a transaction and see its results on top of a particular state, but you no longer have that same state. The right answer for that case is to change the restrictiveness of the call to not perform the checks that are getting in the way.

nijoe1:
> OK I think `maybeModifySenderForSimulation` is going to be the wrong approach for the same reason - you're modifying an existing actor, mutating state, which changes the system that you're running the transaction over. By changing the nature of the calling actor you change the outcome of the execution and that could cause either subtle or very obvious side effects that are entirely different to an actual execution. Part of the point of estimateGas is that you should be able to run a transaction and see its results on top of a particular state, but you no longer have that same state. The right answer for that case is to change the restrictiveness of the call to not perform the checks that are getting in the way.

That's true. I added locally one test where 

```
ContractA -- calls --> contractB
contractB -- calls-a-getter-from -- contractA
returns the value got from contractA
```

And that fails when doing eth_call because we change contractA code 

Which is the best approach to test local changes made on `ref-fvm` and `filecoin-ffi?` on lotus?

Seems that I found an ugly workaround to  do that where skip-sender-validation tests pass while implementing your above improvements but many other tests fail. That's why I am asking if there is a standard pattern to connect lotus with `filecoin-ffi - ref-fvm`

rvagg:
Sorry this is where it gets complicated. I was going to point you to https://github.com/filecoin-project/lotus/blob/master/documentation/misc/Builtin-actors_Development.md but there's a TODO at the bottom to document the things we need here. You need to clone ref-fvm, filecoin-ffi and do some messing around with Cargo.toml in filecoin-ffi to point to your local copy, or a branch on GitHub that you've modified, then you need to rebuild the CGO bindings in filecoin-ffi to make updates if you're changing the API (I'm assuming you're going to pass a flag down into ref-fvm that tells it to skip validation checks), then you can either symlink ffi into lotus or update the submodule to point to your branch.

Then you'll need PRs against all 3 and we'll have to verify they work and review and merge them up the stack and back to here. It's a little messy but AI should be your friend in figuring out how to do the complicated bits like the horrible plumbing and CGO stuff in filecoin-ffi. Also look through git history to find examples of this. https://github.com/filecoin-project/filecoin-ffi/pull/512 is pretty close and recent, plumbing a boolean option down into ref-fvm, see the linked ref-fvm PR there and also the corresponding change in lotus (somewhere).

nijoe1:
I made the workflow work and pass all the tests including fixes to all your above comments concerns!

I might need some time to figure out how to cross-reference make PRs to each repo as the one that you shared!
I will share more hopefully by EOW! 

Thanks @rvagg

rvagg:
@nijoe1 you'll probably have trouble getting CI happy when you do the separate PRs, it's possible, with Cargo.toml changes, but we can also deal with them individually and I could review by checking out your work and trying it for myself. Do your best and we'll work with it.

nijoe1:
@rvagg I made the requested changes around the synthetic-actor and maybe-modify-actor with only one additional function into [filecoin-ffi](https://github.com/nijoe1/filecoin-ffi/tree/feat/apply-message-for-simulation).

SgtPooki:
Any update here? I would love to get this in to unblock some dealbot use-cases.

BigLep:
Per conversation with @rvagg, this is a good item to do after nv28.

rvagg:
Sorry for the long delay, I've picked this up, rebased, and modified it in #13724; I couldn't force-push back to this branch because that option was turned off for this PR. More explanation over there, including why we don't actually need ffi updated to unblock this. I'd appreciate review over there pls.

PR #13483, opened 2026-02-03
fix(eth): trace_filter returns [] for null round ranges

rvagg:
When all blocks in a trace_filter range are null rounds, the API was returning JSON `null` instead of `[]` because the results slice was never initialised.

It turns out this API is the odd one out, `trace_block`, `eth_getLogs`, `eth_getBlockReceipts`, handle these cases more elegantly, returning `[]`. But because in Go, the difference between JSON marshalling a `null` vs a `[]` is subtle, this one got left out.


redpanda-f (approved):
good and necessary fix 😄 :

makes API deterministic
matches Ethereum expectations
removes nil ambiguity

PR #13469, opened 2026-01-13
fix: update F3-calibnet name and bootstrap epoch Calibnet

rjan90:
## Related Issues
https://github.com/filecoin-project/go-f3/issues/1054

## Proposed Changes
fix: update F3-calibnet name and epoch

## Checklist

Before you mark the PR ready for review, please make sure that:

- [x] Commits have a clear commit message.
- [x] PR title conforms with [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#pr-title-conventions)
- [x] Update CHANGELOG.md or signal that this change does not need it per [contribution conventions](https://github.com/filecoin-project/lotus/blob/master/CONTRIBUTING.md#changelog-management)
- [ ] New features have usage guidelines and / or documentation updates in
  - [ ] [Lotus Documentation](https://lotus.filecoin.io)
  - [ ] [Discussion Tutorials](https://github.com/filecoin-project/lotus/discussions/categories/tutorials)
- [ ] Tests exist for new functionality or change in behavior
- [ ] CI is green

PR #13491, opened 2026-02-06
fix: drain F3 data reader when skipping F3 import during snapshot import

rjan90:
## Related Issues
When a snapshot contains F3 data but the F3 manifest is unavailable or InitialPowerTable is undefined, the import code skips F3 processing but fails to consume the f3Reader. Since the CAR block reader returned this reader from NextReader(), the underlying stream position only advances when the reader is fully consumed. The next br.Next() call then misinterprets F3 payload bytes as CID bytes, causing:

  "invalid cid: expected 1 as the cid version number, got: 132"
  
I got the above error, when trying to start a node on Calibnet with the F3 Calibnet restart changes in #13469 

## Proposed Changes
Drain the reader with io.Copy(io.Discard, f3Reader) in both skip branches so the CAR stream correctly advances past the F3 data block.

## Checklist

- [ ] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 
 # UNRELEASED
 
+## 🐛 Bug Fixes
+- fix: drain F3 data reader when skipping F3 import during snapshot import ([filecoin-project/lotus#13491](https://github.com/filecoin-project/lotus/pull/13491))
+
 ## 👌 Improvements
 - docs: fix outdated link in documentation ([#13436](https://github.com/filecoin-project/lotus/pull/13436))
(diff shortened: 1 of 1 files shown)


chain/store/snapshot.go
rvagg: discard logic is repeated twice, you could restructure all of this to do it once:

```
if f3Manifest != nil && f3Manifest.InitialPowerTable.Defined() {
  // certstore.ImportSnapshotToDatastore bit, currently in the final else block
} else {
  if f3Manifest == nil {
    // log
  } else {
    // other log
  }
  // io.Copy(io.Discard, f3Reader) bit here
}
```

PR #13534, opened 2026-03-03
feat: estimate gas premium from mempool

wjmelements:
Reviewer @rvagg
This premium recommendation change does not require FIP-0115 and can be released early.
I'm still running simulations to see if the system can be improved.
The gas market simulation covers:
* suggesting premium for incoming transactions
* message selection
* base fee adjustment
### Changes
* GasEstimateGasPremiumFromMempool
* tests
* gas market simulation tool
### Simulation
I built a tool for simulating the gas market.
#### Method
The simulation runs over a configurable number of epochs.
There are a configurable number of incoming transactions per epoch.
They have a distribution of possible gas limits and deadlines.
They query GasEstimateGasPremium and then GasEstimateFeeCap and then join the mempool.
The epoch concludes with tq-proposer message selection.
The metrics track how many transactions met their deadlines with a breakdown by block.
#### Results
I ran the simulation over 2000 epochs with a starting base fee of 1000.
Briefly, an explanation of the keys in the tables:
* gpe: average gas per epoch, as a multiple of the block gas limit; also 1/10 of the number of submitted transactions per epoch
* baseFee@2000: the base fee at the end of the simulation
* delayed: how many transactions were late **and** not yet confirmed at the end of the simulation
* ever-late: How many transactions were ever late, broken down by their nblocksincl (nb)
* gas%: average gaslimit utilization as a percentage of 5 proposer blocks

I also ran for gpe as high as 5 but it seems impossible to digest much faster than 2.8.

TQ proposals seem to have a maximum average digestion rate between 2.8 and 3.0 due to nearby TQs.

##### `GasEstimateGasPremiumFromMempool`
```
  ┌─────┬──────────────┬─────────────────────┬──────────────────┬───────┬───────┬───────┬───────┐
  │ gpe │ baseFee@2000 │ delayed (nb1/2/10)  │    ever-late     │ nb1%  │ nb2%  │ nb10% │ gas%  │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.2 │ 115          │ 4 (2/2/0)           │ 747/44K (1.7%)   │ 9.6%  │ 2.8%  │ 0.2%  │ 43.9% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.4 │ 119          │ 6 (3/1/2)           │ 1100/48K (2.3%)  │ 12.0% │ 3.7%  │ 0.5%  │ 47.9% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.6 │ 191          │ 7 (2/0/5)           │ 1598/52K (3.1%)  │ 14.4% │ 4.9%  │ 1.0%  │ 52.0% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.8 │ 943          │ 42 (5/12/25)        │ 2132/56K (3.8%)  │ 14.0% │ 5.8%  │ 1.8%  │ 55.9% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 3.0 │ 8905         │ 1776 (101/261/1414) │ 4188/60K (7.0%)  │ 18.6% │ 8.6%  │ 4.8%  │ 57.7% │
  └─────┴──────────────┴─────────────────────┴──────────────────┴───────┴───────┴───────┴───────┘
  ```
  
  ##### Simple (`BaseFee / 8 + noise`)
Tool option `--flat-premium`
  ```
  ┌─────┬──────────────┬─────────────────────┬──────────────────┬───────┬───────┬───────┬───────┐
  │ gpe │ baseFee@2000 │ delayed (nb1/2/10)  │    ever-late     │ nb1%  │ nb2%  │ nb10% │ gas%  │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.2 │ 937          │ 0 (0/0/0)           │ 768/44K (1.7%)   │ 13.3% │ 2.3%  │ 0.0%  │ 43.6% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.4 │ 953          │ 0 (0/0/0)           │ 1013/48K (2.1%)  │ 15.2% │ 3.1%  │ 0.0%  │ 48.0% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.6 │ 993          │ 0 (0/0/0)           │ 1563/52K (3.0%)  │ 17.7% │ 5.5%  │ 0.2%  │ 51.7% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.8 │ 1001         │ 12 (3/6/3)          │ 3376/56K (6.0%)  │ 27.1% │ 11.0% │ 1.7%  │ 56.3% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 3.0 │ 1001         │ 2169 (225/461/1483) │ 6717/60K (11.2%) │ 28.4% │ 14.8% │ 7.7%  │ 57.3% │
  └─────┴──────────────┴─────────────────────┴──────────────────┴───────┴───────┴───────┴───────┘
  ```
  ##### Historical
  Tool option `--legacy-premium`
  With 100k premium floor (150k for nb2, 200k for nb1)
  ```
  ┌─────┬──────────────┬─────────────────────┬──────────────────┬───────┬───────┬───────┬───────┐
  │ gpe │ baseFee@2000 │ delayed (nb1/2/10)  │    ever-late     │ nb1%  │ nb2%  │ nb10% │ gas%  │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.2 │ 817134       │ 1 (1/0/0)           │ 490/44K (1.1%)   │ 8.3%  │ 1.5%  │ 0.0%  │ 43.6% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.4 │ 844674       │ 0 (0/0/0)           │ 821/48K (1.7%)   │ 10.7% │ 3.2%  │ 0.0%  │ 48.0% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.6 │ 852502       │ 2 (0/2/0)           │ 1497/52K (2.9%)  │ 16.1% │ 5.5%  │ 0.3%  │ 51.7% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 2.8 │ 883219       │ 39 (16/14/9)        │ 3077/56K (5.5%)  │ 24.2% │ 9.1%  │ 1.9%  │ 56.3% │
  ├─────┼──────────────┼─────────────────────┼──────────────────┼───────┼───────┼───────┼───────┤
  │ 3.0 │ 955230       │ 1867 (166/337/1364) │ 5345/60K (8.9%)  │ 25.7% │ 12.6% │ 5.5%  │ 57.4% │
  └─────┴──────────────┴─────────────────────┴──────────────────┴───────┴───────┴───────┴───────┘
  ```

### CODE REVIEW

**Comment by wjmelements** [comment]
force pushed a rebase onto #13531

File: node/impl/gasutils/gasutils.go [hunk 1]
```diff
@@ -243,83 +204,271 @@ func GasEstimateFeeCap(ctx context.Context, cstore ChainStoreAPI, msg *types.Mes
...
+	ctx context.Context,
+	cstore ChainStoreAPI,
+	mpool MessagePoolAPI,
+	nblocksincl uint64,
+	gaslimit int64,
+	tsKey types.TipSetKey,
+) (types.BigInt, error) {
 	if nblocksincl == 0 {
 		nblocksincl = 1
 	}
```
wjmelements, line 232: Should probably also upper-bound this because it's a DoS vector

node/impl/gasutils/gasutils.go
```diff
@@ -243,83 +204,271 @@ func GasEstimateFeeCap(ctx context.Context, cstore ChainStoreAPI, msg *types.Mes
...
 
-	for i := uint64(0); i < nblocksincl*2; i++ {
-		if ts.Height() == 0 {
-			break // genesis
+	// Fraction of the current epoch remaining, used to model incoming transactions.
+	// For subsequent simulated blocks (future epochs) we use 1.0, since a full
+	// epoch of additional transactions is expected.
+	fractionRemaining := epochFractionRemaining(ts)
+
+	for {
```
wjmelements, line 286: check ctx.Done

node/impl/gasutils/gasutils.go
wjmelements: 3dc8015f5ded08952b5c4ca34ff9c3cd8aacb92b

node/impl/gasutils/gasutils.go
wjmelements: 032dfc041

node/impl/gasutils/gasutils.go
```diff
@@ -243,83 +204,278 @@ func GasEstimateFeeCap(ctx context.Context, cstore ChainStoreAPI, msg *types.Mes
...
+	// changes during multi-block simulation. originalGasLimit is preserved so that step 4
+	// can add a fixed amount of modeled gas per epoch, ensuring linear rather than
+	// exponential growth in modeled competition over multiple simulated blocks.
+	type rawMsg struct {
+		gasLimit         int64
+		originalGasLimit int64
+		gasFeeCap        abi.TokenAmount
+		gasPremium       abi.TokenAmount
+	}
+	pendingMsgs, _ := mpool.Pending(ctx)
```
rvagg: This is the most costly part of this process, `Pending` takes a read-lock and copies all messages in the mpool for us, so avoiding doing this so frequently would be nice.

Here's an idea: capture this _once_ per epoch and serve the same batch for that entire epoch. Or if you think the mpool dynamism is going to be large enough to warrant quantising smaller than 30 seconds then you could bound it on time.

```go
type mempoolSnapshot struct {
    tsk   types.TipSetKey
    once  sync.Once
    msgs  []rawMsg
}

var current atomic.Pointer[mempoolSnapshot]
```

The `once` serves as an effective mutex for parallel calls.

When `current.tsk != tsk`, swap a new one in. Alternatively just keep a timestamp and quantise at whatever number of seconds you think is acceptable. So repeated calls to this don't end up incurring the lock + copy each time.

Something like this:

```go
snap := current.Load()                                                                                                                                                                                                                                                                                     
if snap == nil || snap.tsk != tsk {
    snap = &mempoolSnapshot{tsk: tsk}                                                                                                                                                                                                                                                                      
    current.Store(snap)                                                                                                                                                                                                                                                                                    
}
snap.once.Do(func() {
    pendingMsgs, _ := mpool.Pending(ctx)
    // dedup + build []rawMsg
    snap.msgs = distribution
})
// now  you have snap.msgs to work with
```

node/impl/gasutils/gasutils.go
```diff
@@ -243,83 +204,278 @@ func GasEstimateFeeCap(ctx context.Context, cstore ChainStoreAPI, msg *types.Mes
...
-	at := buildconstants.BlockGasTarget * int64(blocks) / 2        // 50th
-	at += buildconstants.BlockGasTarget * int64(blocks) / (2 * 20) // move 5% further
-	prev1, prev2 := big.Zero(), big.Zero()
-	for _, price := range prices {
-		prev1, prev2 = price.Price, prev1
-		at -= price.Limit
-		if at < 0 {
+// premiumAtGasPosition returns the gas premium at a given cumulative gas position (from the top)
+// within the top BlockGasLimit gas of a distribution sorted descending by price.
+func premiumAtGasPosition(distribution []GasMeta, position int64) abi.TokenAmount {
```
rvagg: this is tested but not used, is it something that used to be invoked?

wjmelements: I think it might have been something the Claude made before I told it to use something else, and then it didn't remove it. If it's dead code I'll remove it.

wjmelements: e44f9e229

LinZexiao:
This simulation tool provides an intuitive evaluation of the FIP-0115 algorithm, and **the chosen metrics effectively capture the trade-off between network resource utilization and user transaction costs**. To further refine the evaluation, I suggest the following improvements:

#### 1. Scenario Expansion: Simulating Traffic Downswings and Volatility
The current simulations primarily focus on performance under static message pressure. I recommend introducing a **dynamic traffic model** to observe the algorithm’s behavior when message pressure **suddenly drops (downswing)** or fluctuates significantly.
*   **Objective**: To verify if the algorithm's Premium suggestions lag due to "mempool inertia," which could trigger unnecessary oscillations in the $BaseFee$.

#### 2. Enhanced Metrics: Volatility and Cumulative Expenditure
I suggest adding the following two core metrics to the simulation output:
*   **BaseFee Volatility**: Measures the standard deviation or period-over-period change rate of the $BaseFee$. Lower volatility indicates better **price predictability** for users.
*   **Total Cumulative Expenditure**: The aggregate sum of costs for all included messages ( $\sum GasUsed \times GasPrice$ ) across the entire simulation period. While a snapshot of the $BaseFee$ (e.g., at epoch 2000) might be skewed by transient fluctuations, **cumulative throughput** provides a more robust measure of how efficiently the algorithm utilizes network capacity over a full cycle.

node/impl/gasutils/gasutils.go
wjmelements: > This is the most costly part of this process, Pending takes a read-lock and copies all messages in the mpool for us, so avoiding doing this so frequently would be nice.

In geth I've done something similar every 100ms and it wasn't a problem. Perhaps there is some optimization that can be done in the mempool.

> once per epoch

No, the best gas oracles are real-time and update every transaction.

> then you could bound it on time.

could remake the snapshot once per mutation

wjmelements:
> Total Cumulative Gas Throughput

This is already measured, and is reflected as the percentage `gas%` in the result column as a percentage of maximum theoretical throughput.

It seems to cap out around 57.5%, perhaps because of inherent TQ overlap.

LinZexiao:
> > Total Cumulative Gas Throughput累计总天然气吞吐量
> 
> This is already measured, and is reflected as the percentage `gas%` in the result column as a percentage of maximum theoretical throughput.这已经测量过了，并在结果列中以 `gas%` 占最大理论吞吐量的百分比）的形式反映出来。
> 
> It seems to cap out around 57.5%, perhaps because of inherent TQ overlap.似乎最高值在 57.5% 左右，可能是由于固有的 TQ 重叠造成的。

I apologize, my previous state was incorrect, and I have now corrected it. I think the Average Gas Price effectively reflects the actual amount of tokens users spend to get their messages included on-chain.

wjmelements:
> Total Cumulative Expenditure

This is a good idea. Can also track how much FIL was burned vs how much was paid to proposers. (assumes gaslimit = gasUsed)

LinZexiao:
Additionally, some extreme scenarios may be taken into consideration, such as a surge of large messages (e.g., SettleDealPaymentsExported). To account for this, we could shift the mean simulated message gas limit from 10% to 75% of the block limit.

node/impl/gasutils/gasutils.go
wjmelements: Same for `MpoolPending`

wjmelements: I decided between three solutions for caching MessagePool.Pending:
1. subscribe to the pubsub and actively maintain a consistent snapshot (nonblocking, linear work per update)
2. push changes to a lock-free map (fastest and nonblocking, but reads are not linearizable)
3. generate a consistent snapshot on-demand with `RLock`, invalidating it whenever the mempool changes (blocking, linear work per request)

All requests have to do linear work because they will be returning a copy. It seems there isn't a way to return `const` like in C, so it cannot be proven that consumers aren't unintentionally messing with your snapshot.

For this particular pull request, the best solution is the lock-free map (2), because consistent reads are less important; it is okay if some of the mempool data is out of date some of the time. However, it would break the correctness of MpoolPending, and I want these similar use cases to share logic.

The consistent snapshot (1) is expensive because of the cost of eagerly producing a snapshot that nobody has requested.

So I am preferring the on-demand snapshot (3) which is an improvement over the current behavior in the case where there are many requests between updates, alleviating the worst case.

wjmelements: https://github.com/filecoin-project/lotus/pull/13542

Kubuxu (commented):
Seems to me a bit overly complex, but just a bit. I'm glad we are not running an actual selection process to move forward an epoch, because that could be cost-prohibitive.

Have you tested the results on a live mainnet node for comparison with the old method?
I need a bit more time to think.

wjmelements:
> I'm glad we are not running an actual selection process to move forward an epoch, because that could be cost-prohibitive.

Claude actually suggested doing that!

> Have you tested the results on a live mainnet node for comparison with the old method?

No, I've only run the simulations. Testing live would require a transaction source; maybe I could spend some FIL on this. I think the prices it suggests will be much lower. I'll verify that they can still confirm.

wjmelements:
> I think the prices it suggests will be much lower. I'll verify that they can still confirm.

I setup a mainnet lotus node to test this. I compared `eth_maxPriorityFeePerGas` from a public lotus node (glif) using history to the mempool approach. The history-based approach fluctuated more wildly block to block, sometimes doubling its suggestion, alternating between 100k and 200k, while the mempool approach was steadier around 125k.

[priorityfee.log](https://github.com/user-attachments/files/25982548/priorityfee.log)

I also submitted a [transaction](https://filfox.info/en/message/bafy2bzacebhqbjouscm7hth4mints65crcg3zahi4fcrcbqa4htrusikd7qqm) during this time and it confirmed in the next block.

However I still find the evidence from the simulation more compelling.

wjmelements:
I noticed some irregularities in the results of `eth_maxPriorityFeePerGas` this weekend. I will investigate them tomorrow.

wjmelements:
> I noticed some irregularities in the results of `eth_maxPriorityFeePerGas` this weekend. I will investigate them tomorrow.

I think the behavior is correct, but during the investigation I identified some ways to improve:

* target inclusion in a non-greedy block (previously suggested by @LinZexiao)
* measure total gas premiums paid in the simulation
* Disregard mempool transactions that are blocked from inclusion because of a nonce gap

Kubuxu:
> However I still find the evidence from the simulation more compelling

Of course, testing it in live is more of a smoke test to check if there are no unexpected factors.

wjmelements:
Per a request from @magik6k I have collected some [tx history data](https://github.com/user-attachments/files/26360985/tx-history.csv). Here it is:
gas_premium_by_bucket

The current gas market simulation generates more midsized transactions, while the actual distribution has almost exclusively tiny transactions.
gas_limit_sim_vs_real

I will change the incoming transaction modeling so that it matches this observed distribution.

wjmelements:
Next I am working on the requested "replay" simulation. In order for this to be fair I would need to record actual mempool arrivals. I'll set that up today and run it for at least 24 hours.

wjmelements:
> Next I am working on the requested "replay" simulation. In order for this to be fair I would need to record actual mempool arrivals. I'll set that up today and run it for at least 24 hours.

It took longer to collect data than expected due to two disruptive major internet outages yesterday and the day before. I have a 24 hour period recorded now though so I'll be doing the replay analysis today.

LinZexiao:
Is this PR going to be merged before time for NV28 calibnet ( 2026-04-27 , according to upgrading plan) ?

wjmelements:
> Is this PR going to be merged before time for NV28 calibnet ( 2026-04-27 , according to upgrading plan) ?

I haven't had time yet to work through [the last few issues](https://github.com/filecoin-project/lotus/pull/13534#issuecomment-4071211680), but this doesn't block nv28.

wjmelements:
> Disregard mempool transactions that are blocked from inclusion because of a nonce gap

I'm currently optimizing for this on a separate branch in order to avoid doing `GetActorAfter` for every sender. While there is already a nonce cache, my plan is to put startNonce and startBalance into msgSet. Such an optimization should already benefit message validation (`Add`) and message selection, so I plan to merge it separately and ahead of this change.

Issue #8385, opened 2022-03-28
cannot connect to infura using lite node


flyworker:
### Checklist

- [X] This is **not** a security-related bug/issue. If it is, please follow please follow the [security policy](https://github.com/filecoin-project/lotus/security/policy).
- [X] This is **not** a question or a support request. If you have any lotus related questions, please ask in the [lotus forum](https://github.com/filecoin-project/lotus/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/lotus/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/lotus/issues) and the [lotus forum](https://github.com/filecoin-project/lotus/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/lotus/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to lotus.

### Lotus component

- [X] lotus daemon - chain sync
- [ ] lotus miner - mining and block production
- [ ] lotus miner/worker - sealing
- [ ] lotus miner - proving(WindowPoSt)
- [ ] lotus miner/market - storage deal
- [ ] lotus miner/market - retrieval deal
- [ ] lotus miner/market - data transfer
- [ ] lotus client
- [ ] lotus JSON-RPC API
- [ ] lotus message management (mpool)
- [ ] Other

### Lotus Version

```text
1.4.3
```

### Describe the Bug

cannot connect to infura using lite node
when lite node connect with infura filecoin node with project ID and secret,it shows error listen head changes errored: listenHeadChanges ChainNotify call failed: RPC error (-32004): method 'Filecoin.ChainNotify' not supported in this mode (no out channel support)

### Logging Information

```text
2022-03-28T05:26:37.357-0400    ERROR   events  events/observer.go:58   listen head changes errored: listenHeadChanges ChainNotify call failed: RPC error (-32004): method 'Filecoin.ChainNotify' not supported in this mode (no out channel support)
[...]

PR #13578, opened 2026-04-10
feat(cliutil): accept non-JWT API tokens

beck-8:
Allows non-JWT tokens from third-party RPC providers (e.g. Glif) to be parsed correctly in TOKEN:ADDRESS .

## Related Issues
https://filecoinproject.slack.com/archives/CP50PPW2X/p1775732893778079
https://github.com/filecoin-project/lotus/issues/8385

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 
 ## 👌 Improvements
 
+- feat(cliutil): accept non-JWT API tokens in `TOKEN:ADDRESS`, enabling use of third-party RPC providers (e.g. Glif) that issue opaque API keys ([filecoin-project/lotus#13578](https://github.com/filecoin-project/lotus/pull/13578))
+
 # Node v1.35.1 / 2026-03-31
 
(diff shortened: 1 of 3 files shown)


rjan90 (commented):
Change looks good to me! But maybe I can get a quick sanity check from either @rvagg or @Kubuxu ?

beck-8:
Ignoring this referenced, I temporarily ported it to curio testing.

rvagg:
This is fine as the narrowest possible fix for the problem. There's more we can do here but this is a reasonable unblock. Important to note though that it doesn't fully help with the multiaddr styly: you have to supply URL style to get `wss` (we hard-code `ws://`, see line 60), but that's a pre-existing problem and you can get around it by not using multiaddr. **But** you could fix it here too if you wanted to, detect protocol and adjust accordingly. Something like this in `DialArgs`:

```go
scheme := "ws"
multiaddr.ForEach(ma, func(c multiaddr.Component) bool {
    switch c.Protocol().Code {
    case multiaddr.P_WSS, multiaddr.P_TLS:
    scheme = "wss"
    }
    return true
})
return scheme + "://" + addr + "/rpc/" + version, nil
```

One ask before approve / merge is that it would be good to add some tests for `ParseApiInfo`. We don't have existing tests but now would be a good time to spit out some basic tests. We could cover:

- `jwt.tok.en:/ip4/127.0.0.1/tcp/1234` -> token + multiaddr
- `opaque:/dns/host/tcp/443/wss` -> token + multiaddr
- `opaque:wss://host` -> token + URL (this case)
- `/ip4/127.0.0.1/tcp/1234` -> bare multiaddr
- `wss://host` -> bare URL
- `ws://host:port/path` -> bare URL with path

beck-8:
We actually still have some historical baggage here. The multiaddr http class will be converted to ws. We can't touch him here.

rvagg (approved):
solid

PR #13581, opened 2026-04-17
Revert "fix(basefee)!: add fip-0115 feature flag (`LOTUS_FEES_FIP0115HEIGHT`)"

wjmelements:
Reviewer @zenground0 @kubuxu
CC @jennijuju @LesnyRumcajs
FIP-0115 is accepted and scheduled for inclusion in nv28 (Firehorse).
#### Changes
* Revert https://github.com/filecoin-project/lotus/pull/13550

### DIFF
--- a/chain/store/basefee.go
+++ b/chain/store/basefee.go
@@ -54,10 +54,10 @@ func (cs *ChainStore) ComputeBaseFee(ctx context.Context, ts *types.TipSet) (abi
 	if buildconstants.UpgradeBreezeHeight >= 0 && ts.Height() > buildconstants.UpgradeBreezeHeight && ts.Height() < buildconstants.UpgradeBreezeHeight+buildconstants.BreezeGasTampingDuration {
 		return abi.NewTokenAmount(100), nil
 	}
-	if cs.fip0115BaseFeeHeight >= 0 && ts.Height() >= cs.fip0115BaseFeeHeight {
-		return cs.ComputeNextBaseFeeFromPremiums(ctx, ts)
+	if ts.Height() < buildconstants.UpgradeFireHorseHeight {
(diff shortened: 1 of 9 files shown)


chain/store/basefee.go
```diff
@@ -54,10 +54,10 @@ func (cs *ChainStore) ComputeBaseFee(ctx context.Context, ts *types.TipSet) (abi
 	if buildconstants.UpgradeBreezeHeight >= 0 && ts.Height() > buildconstants.UpgradeBreezeHeight && ts.Height() < buildconstants.UpgradeBreezeHeight+buildconstants.BreezeGasTampingDuration {
 		return abi.NewTokenAmount(100), nil
 	}
-	if cs.fip0115BaseFeeHeight >= 0 && ts.Height() >= cs.fip0115BaseFeeHeight {
-		return cs.ComputeNextBaseFeeFromPremiums(ctx, ts)
+	if ts.Height() < buildconstants.UpgradeFireHorseHeight {
```
rvagg, line 57: I think this should probably be `<=`, most upgrade behaviour changes take effect the epoch after the upgrade, there is some inconsistency here but I think the deferred execution model suggests that state upgrade at epoch, behaviour change next epoch is what we should aim for.

wjmelements: This is the spec, and implemented all all clients

wjmelements: The tipset here is the previous block, so the basefee change does begin at epoch +1. You are asking to change it to +2.

rjan90: Okay, I will merge this PR as is to unblock Butterfly-testing. If there is any need for followup fixes, we can land that in a separate PR.

rjan90 (approved):

PR #13617, opened 2026-05-08
fix(eth): scope receipt logs by message, cap event filter across tipsets

### LINKED ISSUE #13616: pdp: transactions cannot be confirmed when another tx in the same block emits many events
## Setup
- Curio PDP node, calibnet, branch `pdpv0` at `6d77aecf`
- Self-hosted Lotus `1.36.0-rc1+calibnet+git.f0f80cadf`, default config

## What I'm seeing

`eth_getTransactionReceipt` fails for legitimate Curio-sent txs (`pdp-addpieces`, `pdp-proving-period`):

```
filter matches too many events, try a more restricted filter
```

For affected `txHash`es the failure is **permanent**. Receipts have not become fetchable after 12+ hours; Curio's `MessageWatcherEth` retries every chain head and gets the same error each time, producing ~6000 errors/hour on a single PDP node, flat over the 12-hour observation window.

A `LOG1`-in-a-loop pattern observable on calibnet right now causes individual txs to exceed `Events.MaxFilterResults` (default 10000). Once one of those txs lands in a block, `eth_getTransactionReceipt` fails for every other tx in that same block.

## Operator impact

On one calibnet PDP node during a 5-hour attack window:

| send_reason         | sent | permanently stuck |
|---------------------|------|-------------------|
| pdp-addpieces       |   62 |  26 (42%)         |
| pdp-proving-period  | 1756 | 484 (28%)         |

- Each stuck `pdp-addpieces` = a permanently-failed Synapse-SDK / dealbot deal (5-min `addPieces` timeout already long expired).
- Each stuck `pdp-proving-period` = a stalled proving cycle for that dataset; missed deadlines lead to on-chain faults.
- Curio keeps polling the stuck `txHash`es every chain head; the error volume does not decay.
- Each future recurrence of a high-event-emitting tx adds a new permanent stuck batch.

## Acknowledgement

@rvagg in Slack:
> the shape of the fix for this is roughly: don't apply any max results limit where a query for logs is scoped to a specific block or transaction (depending on the call type, some are only per-block like eth_getLogs); eth_getTransactionReceipt shouldn't be limited cause the user can't do anything else, there's no scoping, it's just "give me all of this one tx" and that should be reasonable, if messy

Workaround for self-hosted operators (also from @rvagg): set `MaxFilterResults = 0` in config under `[Events]`. Verified to drain the stuck backlog on my calibnet PDP node.

Filing this as a tracker for the upstream fix; happy to provide additional log/DB excerpts or test the patch.

rvagg:
eth_getTransactionReceipt narrows event lookup to the requested message at the index level, so a transaction emitting many events no longer blocks receipt retrieval for other transactions in the same block. MaxFilterResults is now a hard cap only when a query's events span more than one tipset; single-tipset queries (TipsetCid, BlockHash) bypass it.

Fixes: https://github.com/filecoin-project/lotus/issues/13616

---

I want to get some RPC API operators involved in this because it does have implications for response forms. The current implementation will just error when you try to eth_getTransactionReceipt or eth_getBlockReceipts where there are more than `MaxFilterResults` logs in the collection (tx for the former, all tx's in the block for the latter, and MaxFilterResults default is 10,000). This removes the cap entirely for queries scoped to a single tipset, and narrows the receipt path further to events for the requested transaction, so it'll faithfully return what's there. `MaxFilterResults` applies where a query spans tipsets, so we always give you a way to get transactions, but if it's expensive we'll force you to get more granular. But, this means the maximum number of logs is going to be bigger than now so the JSON serialised form will be much larger, meaning much larger response bodies.

We have an example on calibnet now that's a useful reference thanks to @wjmelements:

```sh
$ curl -s -X POST -H "Content-Type: application/json" --data \
  '{"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["0x547fd9f825662cf2d587a122e04bc0e3a7dd124813bcd9383ee76cde0fe45157"],"id":1}' \
  http://localhost:1235/rpc/v1 > receipt.json
$ du -h receipt.json
84M     receipt.json
```

This msg used 8B gas, and has 216,000 logs in a single transaction. A lot of the noise in the json is not raw log content, so there's not quite so much variation in what you log but a single tx could probably fairly easily push toward the 300,000 log mark within the 10B gas limit, maybe more. Potentially multiple transactions in a tipset like this, which means very large JSON response bodies. Maybe 120M JSON body worst case for a single transaction. Manage to pack multiple into a tipset and it'd get interesting.

Current Lotus lets you not even serve that response. This PR lets you serve it but doesn't give you a direct opt-out. `MaxFilterResults` only applies when someone wants to do multi-tipset eth_getLogs.

ISTM that leaning on proxy-level response body size limiting might be the appropriate mitigation for this. But we could also put back in some kind of protection here at the Lotus level if you just want to not serve such large responses and tell the caller to go figure something else out. Maybe for unpaid RPC endpoints that's the right call. Is Lotus the right place to do that?

Some options:

1. Status quo as per this PR, proxy-level response body limiting is the knob to tune here
2. Add some kind of tunable response-body cap, probably in lotus-gateway (might be tricky since we defer to go-jsonrpc but maybe we can implicate it too to help)
3. Add a new per-event call cap, similar to how `MaxFilterResults` works today but a new variable that'll refuse to serve responses that involve large numbers of events
4. A variation of number 3 where we introduce a boolean that makes `MaxFilterResults` as it is now apply in the current Lotus form, effectively overriding the changes in this PR.

### DIFF
--- a/chain/index/events.go
+++ b/chain/index/events.go
@@ -414,7 +414,14 @@ func (si *SqliteIndexer) GetEventsForFilter(ctx context.Context, f *EventFilter)
 
 		var ces []*CollectedEvent
 		var currentID int64 = -1
+		var lastHeight abi.ChainEpoch = -1
+		var tipsetsSeen int
 		var ce *CollectedEvent
+
+		// Queries narrowed to a single message or tipset bypass MaxResults. For range queries,
+		// the cap is enforced as a hard limit once events come from more than one tipset; a
+		// range whose events all live in one tipset may exceed MaxResults.
+		enforceMaxResults := f.MaxResults > 0 && f.TipsetCid == cid.Undef && f.MsgCid == cid.Undef
 
 		for q.Next() {
 			select {
@@ -457,24 +464,27 @@ func (si *SqliteIndexer) GetEventsForFilter(ctx context.Context, f *EventFilter)
 				return nil, xerrors.Errorf("read prefill row: %w", err)
 			}
 
-			// The query will return all entries for all matching events, so we need to keep track
-			// of which event we are dealing with and create a new one each time we see a new id
+			// The query returns all entries for all matching events; create a new CollectedEvent each time we see a new id.
 			if row.id != currentID {
-				// Unfortunately we can't easily incorporate the max results limit into the query due to the
-				// unpredictable number of rows caused by joins
-				// Error here to inform the caller that we've hit the max results limit
-				if f.MaxResults > 0 && len(ces) >= f.MaxResults {
-					return nil, ErrMaxResultsReached
+				rowHeight := abi.ChainEpoch(row.height)
+				if rowHeight != lastHeight {
+					tipsetsSeen++
+					lastHeight = rowHeight
 				}
 
 				currentID = row.id
 				ce = &CollectedEvent{
 					EventIdx: row.eventIndex,
 					Reverted: row.reverted,
-					Height:   abi.ChainEpoch(row.height),
+					Height:   rowHeight,
 					MsgIdx:   row.messageIndex,
 				}
 				ces = append(ces, ce)
+
+				// Hard cap once events span multiple tipsets; a single contributing tipset may bust.
+				if enforceMaxResults && tipsetsSeen > 1 && len(ces) > f.MaxResults {
+					return nil, ErrMaxResultsReached
+				}
 
 				if row.emitterAddr == nil {
 					ce.EmitterAddr, err = address.NewIDAddress(row.emitterID)
@@ -612,6 +622,11 @@ func makePrefillFilterQuery(f *EventFilter) ([]any, string, error) {
 		// unless asking for a specific tipset, we never want to see reverted historical events
 		clauses = append(clauses, "e.reverted=?")
 		values = append(values, false)
+	}
+
+	if f.MsgCid != cid.Undef {
+		clauses = append(clauses, "tm.message_cid=?")
+		values = append(values, f.MsgCid.Bytes())
 	}
 
 	if len(f.Addresses) > 0 {
--- a/chain/index/events.go
+++ b/chain/index/events.go
@@ -433,11 +433,6 @@ func (si *SqliteIndexer) GetEventsForFilter(ctx context.Context, f *EventFilter)
 			lastEmitterAddrBytes string          // valid when !lastEmitterIsID && set
 		)
 
-		// Queries narrowed to a single message or tipset bypass MaxResults. For range queries,
-		// the cap is enforced as a hard limit once events come from more than one tipset; a
-		// range whose events all live in one tipset may exceed MaxResults.
-		enforceMaxResults := f.MaxResults > 0 && f.TipsetCid == cid.Undef && f.MsgCid == cid.Undef
-
 		// Reused across rows; declaring inside the loop allocates fresh slots each iteration.
 		var row struct {
 			id           int64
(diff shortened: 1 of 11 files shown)


rvagg:
cc @LesnyRumcajs your input here would be helpful too, it's going to impact Forest in different ways and you have examples to test with now and you're being used to serve these endpoints so you can probably form some opinions of how you might want to give tunable knobs to node operators and what the limits might be.

One unfortunate negative for Lotus is that most of our APIs are buffer-and-serialise rather than being streambable. So a response to these queries requires memory accumulation and then serialisation when that's done. Which isn't awesome, but the JSON is always going to be a lot larger than the in-memory bytes needed to represent it.

rvagg:
`eth_getBlockReceipts` for the containing block for the above transaction (`0xed34b6ff143116496bc37ab02d43f81a22216fcf146fda3a2c7a4875763da94a`) is also interesting because it has two of those transactions in it, 432K logs, 166MB of JSON. This takes ~11s on my calibnet node to generate and serialise a response for using the patch in this PR.

rvagg:
I found a couple of memoisation opportunities that reduce the time to process worst-case examples we're playing with to less than 40% of the time, both the per-tx and the per-block calls, much neater now. See latest commit. JSON serialisation cost is the biggest hurt now.

rvagg:
Found some more low-hanging fruit to optimise for these large-transaction cases and we're down by ~75% of the time to produce a response for the cases currently being considered (~25%). Latest commit.

chain/index/events.go
```diff
@@ -457,56 +480,72 @@ func (si *SqliteIndexer) GetEventsForFilter(ctx context.Context, f *EventFilter)
...
 			if row.id != currentID {
-				// Unfortunately we can't easily incorporate the max results limit into the query due to the
-				// unpredictable number of rows caused by joins
-				// Error here to inform the caller that we've hit the max results limit
-				if f.MaxResults > 0 && len(ces) >= f.MaxResults {
-					return nil, ErrMaxResultsReached
+				rowHeight := abi.ChainEpoch(row.height)
+				if rowHeight != lastHeight {
+					tipsetsSeen++
+					lastHeight = rowHeight
```
wjmelements, line 483: We check both `tipsetsSeen` (for `tipsetsSeen > 1`) and `f.TipsetCid == cid.Undef`? It seems redundant. Is there a case we are trying to handle where the filter might contain multiple tipsets but we only want to `ErrMaxResultsReached` if it actually has results from multiple tipsets?

wjmelements (approved):
lgtm

chain/index/events.go
rvagg: no, good catch; it's mostly a readability thing, having an `enforceMaxResults` that's able to spell out how it's enforced is nice, but it's another variable and a redundant check; `tipsetsSeen>1` mostly covers the condition by itself.

so the check below becomes `if f.MaxResults > 0 && tipsetsSeen > 1 && len(ces) > f.MaxResults`

rvagg:
to the question about the limiting path in the core change in here, here's the map of how we hit that collector:

* `TipsetCid` only: `eth_getLogs(BlockHash)`, `eth_getBlockReceipts`, `GetActorEventsRaw{TipSetKey}` - all msgs - should be unlimited
* `TipsetCid` + `MsgCid`: `eth_getTransactionReceipt` (per-tx receipt path) - single tipset, single msg - should be unlimited
* Range: `eth_getLogs(fromBlock..toBlock)` - many messages, many tipsets - should only be limited once the range is beyond one tipset
* There is also a possible range + `MsgCid` that would work through this code and it would be unlimited once it hits the msg, cause it's just that msg, but we don't need/use that anywhere

Kubuxu (approved):

dumikau:
Hey! No objections from our side, we're pretty much okay with responses of any size.

rvagg:
nice; this is in the latest lotus release and has been working great for me on my local node

PR #13618, opened 2026-05-10
fix(eth): return indexed block logsBloom

shrirajpawar4:
## Related Issues

  Fixes #13546

  ## Proposed Changes

  - Add a `tipset_bloom` table to the chain index DB for block-level ETH log blooms.
  - Add a migration for existing chain index DBs.
  - Compute and store tipset-level blooms while indexing actor events.
  - Store empty blooms for indexed tipsets with no events.
  - Rebuild/store blooms when previously reverted events are unreverted.
  - Delete bloom rows on revert.
  - Delete old bloom rows during chain index GC.
  - Expose `GetTipsetBloom` on the chain indexer.
  - Use indexed blooms when constructing ETH block responses for:
    - `eth_getBlockByHash`
    - `eth_getBlockByNumber`
    - new-head subscription block payloads
  - Preserve compatibility by falling back to the previous full bloom when no indexed bloom is available.
  - Extend `TestTxReceiptBloom` to cover receipt bloom, empty block bloom, block-by-hash bloom, and block-by-number bloom.


chain/index/events.go
```diff
@@ -86,6 +82,19 @@ func (si *SqliteIndexer) indexEvents(ctx context.Context, tx *sql.Tx, msgTs *typ
 	if err != nil {
 		return xerrors.Errorf("failed to load executed messages: %w", err)
 	}
+	blockBloom := ethtypes.NewEmptyEthBloom()
+
+	if rows > 0 {
+		log.Debugf("unreverted %d events for tipset: %s", rows, msgTs.Key())
+		if err := si.buildTipsetBloom(ctx, ems, executionTs, blockBloom); err != nil {
```
rvagg: this is new cost in here that will go into the chainstore, but we have the events all in the db that we could build from; our cheap early-return is now much mor expensive - there is work to do but it shouldn't be this expensive; investigate if the bloom can be rebuilt from what we have in the index

chain/index/events.go
```diff
@@ -149,9 +160,93 @@ func (si *SqliteIndexer) indexEvents(ctx context.Context, tx *sql.Tx, msgTs *typ
...
+		return xerrors.Errorf("failed to store tipset bloom: %w", err)
+	}
+
+	return nil
+}
+
+func (si *SqliteIndexer) buildTipsetBloom(ctx context.Context, ems []executedMessage, executionTs *types.TipSet, blockBloom ethtypes.EthBytes) error {
+	addressLookups := make(map[abi.ActorID]address.Address)
+	for _, em := range ems {
+		for _, event := range em.evs {
```
rvagg, line 172: this is largely a duplicate of what we have in `indexEvents` but specifically for bloom building, could we come up with a better abstraction that avoids having the two loops, doing essentially the same thing, but one limited to bloom? maybe a multi-purpose loop func with a callback, or a loop func with args that dictate what it does and doens't do

chain/index/events.go
```diff
@@ -149,9 +160,93 @@ func (si *SqliteIndexer) indexEvents(ctx context.Context, tx *sql.Tx, msgTs *typ
...
+		return
+	}
+
+	ethtypes.EthBloomSet(blockBloom, ethAddr[:])
+	for _, topic := range topics {
+		ethtypes.EthBloomSet(blockBloom, topic[:])
+	}
+}
+
+func ethLogBloomTopics(entries []types.EventEntry) ([]ethtypes.EthHash, bool) {
```
rvagg: this has a lot of duplication of the work done in `ethLogFromEvent` already, how about we move an abstraction for doing this into chain/types/ethtypes/ alongside `EthBloomSet` and just do it in one place

chain/index/events.go
```diff
@@ -149,9 +160,93 @@ func (si *SqliteIndexer) indexEvents(ctx context.Context, tx *sql.Tx, msgTs *typ
...
+			addEventToBloom(blockBloom, addr, event.Entries)
+		}
+	}
 	return nil
 }
 
+func addEventToBloom(blockBloom ethtypes.EthBytes, emitterAddr address.Address, entries []types.EventEntry) {
+	ethAddr, err := ethtypes.EthAddressFromFilecoinAddress(emitterAddr)
+	if err != nil {
+		return
```
rvagg, line 259: I think we're already filtering for delegated addresses by the time we get here so a failure here would be of interest; let's add a `log.Warnw` in this early return and see if it surfaces anything; weird data should get our attention, if it's noisy we can undo it later.

shrirajpawar4:
@Kubuxu can you take a look please

Kubuxu:
I started reviewing it, but I need to wrap my head around a few things.

chain/index/events.go
```diff
@@ -73,8 +74,16 @@ func (si *SqliteIndexer) indexEvents(ctx context.Context, tx *sql.Tx, msgTs *typ
 	if err != nil {
 		return xerrors.Errorf("failed to get rows affected by unreverting events for tipset: %w", err)
 	}
+	blockBloom := ethtypes.NewEmptyEthBloom()
+
 	if rows > 0 {
 		log.Debugf("unreverted %d events for tipset: %s", rows, msgTs.Key())
+		if err := si.buildTipsetBloomFromIndex(ctx, tx, msgTsKeyCidBytes, blockBloom); err != nil {
```
Kubuxu, line 88: This will unnecessarily rebuild the boom every time when going through this path.
We should check if the bloom entry exists and rebuild and upsert it only if it doesn't.

Kubuxu (approved):
Apart from that one comment, LGTM.
I haven't really worked on this part of the code much before, but it looks good.

chain/index/events.go
hugomrdias: @shrirajpawar4 please address this comment

hugomrdias (approved):

Kubuxu (approved):

rjan90:
Pushed a commit for addressing the changelog conflict, thanks for working on this @shrirajpawar4 - will merge this once checks passes now

PR #13628, opened 2026-05-15
fix(gateway): align v2 RPC surface with v1

rvagg:
Some misalignments discovered in what v2 exposes for eth vs v1, this closes the gap and enforces it with some tests.


rjan90 (approved):

rjan90:
Code looks good. But just saw now that this requires a `make gen` it seems like

PR #13636, opened 2026-05-25
feat(mempool): raise MaxUntrustedActorPendingMessages from 10 to 100

beck-8:
## Related Issues
None. The current value has been unchanged since the untrusted push path was introduced in 3c7246196 (2020).

## Proposed Changes
- Raise `MaxUntrustedActorPendingMessages` from `10` to `100` in `chain/messagepool/messagepool.go`.
- Update the corresponding comment on `PushUntrusted`.
- Add a `CHANGELOG.md` entry under `## 👌 Improvements`.

## Additional Info
The per-actor cap on the untrusted push path — used by `lotus-gateway` and therefore every public RPC relaying `MpoolPush` from external clients — has been `10` since this path was introduced in 2020.

That value is too tight for normal usage. A sender pushing at modest cadence can easily have more than 10 messages in flight when basefee spikes briefly or block inclusion stalls, after which every subsequent send is rejected with `ErrTooManyPendingMessages`. The failure surfaces as a hard error on the send path rather than backpressure.

**Impact**
- Cap raised 10x but still an order of magnitude **below** the trusted-path cap (`MaxActorPendingMessages = 1000`), preserving the asymmetry that protects gateway operators from external spam.
- `maxNonceGap = 0` for untrusted messages is **unchanged** — pending messages must still form a contiguous nonce sequence from the on-chain nonce, so an attacker cannot park arbitrary future nonces.
- Worst-case additional per-actor memory: ~90 extra ` SignedMessage ` entries, negligible against the existing global pool bounds.
- No API or wire-format changes.

## Checklist

- [x] Tests exist for new functionality or change in behavior (no test references the literal `10`; behavior is unchanged apart from the constant)
- [ ] CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 ## 👌 Improvements
 
 * Silence libp2p config log spam ([filecoin-project/lotus#13612](https://github.com/filecoin-project/lotus/pull/13612))
+* feat(mempool): raise the default per-actor cap on the untrusted push path (`MaxUntrustedActorPendingMessages`) from 10 to 100, reducing spurious `ErrTooManyPendingMessages` rejections for normal-cadence senders relaying through `lotus-gateway` / public RPC ([filecoin-project/lotus#13636](https://github.com/filecoin-project/lotus/pull/13636))
 
 # Node and Miner v1.36.0 / 2026-05-13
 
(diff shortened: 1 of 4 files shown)


beck-8:
The problem I actually encountered was: I often encountered `too many pending messages for actor` when using glif rpc (run pdp node), regardless of whether it is a paid plan or not. This limit is too low.

rvagg (approved):

